Issue #579, opened 2019-04-28
Use a wizard for creating end devices


kschiffer:
**Summary:** 
The add device form (added in #573) is currently not composing fields based on constraints (except for ABP / OTAA selection). For example, it could hide the fields that are only relevant for specific LoRaWAN Versions, or rename fields accordingly (eg. `NwkSKey ` vs. `FNwkSIntKey`).

**Why do we need this?**
For better UX, avoid faulty inputs

**What is already there? What do you see now?**
The add device form, with conditional fields based on OTAA/ABP

**What is missing? What do you want to see?**
More sophisticated checks and constraints applied into the form, preventing faulty inputs

**Environment:**
Console in browser

**How do you propose to implement this?**
Likely hook into field change events and compose fields accordingly

**Can you do this yourself and submit a Pull Request?**
Yes.

bafonins:
The main problem with the current implementation of the device form is that everything is coupled together. This causes
1. Complex and long validation schema
2. No straightforward way to disable certain fields based on the stack configuration
3. No straightforward way to change the form fields/field titles accoring to selected values
4. Unnecessary complexity in the js sdk for batched creation/update/deletion as well as creation rollback logic 

I propose to implement the device form as the multi-step form. This can looks something like this:
<img width="711" alt="Screenshot 2019-08-26 at 11 29 16" src="https://user-images.githubusercontent.com/16374166/63681375-f1d9ab80-c7f5-11e9-9b51-be65ee570ef7.png">
<img width="711" alt="Screenshot 2019-08-26 at 11 31 57" src="https://user-images.githubusercontent.com/16374166/63681376-f2724200-c7f5-11e9-8ee8-fce173b1963b.png">
<img width="711" alt="Screenshot 2019-08-26 at 11 34 16" src="https://user-images.githubusercontent.com/16374166/63681378-f2724200-c7f5-11e9-9da3-eff3571aa41b.png">

Such approach addresses all issues mentioned above:
1. Instead of one complex schema we define a small one for each step.
2. Simply disable the whole step if a certain reponsible component is not available in the stack. What is more, we can inform the user about this via description/notification.
3. Adapt each step depending on the previously submitted values. For example, 
4. No need for batched requests. The user becomes responsible for creating device in different components. However we might want to keep batched deletion for convenience.

Editing devices can be implemented as a stack of accordions:
[...]

PR #754, opened 2019-05-24
Change app/network key types in Device Overview

### LINKED ISSUE #753: App/Network key C-format option
#### Summary

Would be nice to have the option to display AppKey and Network Key in C-format, the same way as DevEUI and JoinEUI.

#### Why do we need this?

To be able to easily copy and paste the keys to the device.

AppKey and Network Key without an option of C-style representation, DevEUI and JoinEUI with.

AppKey and Network Key with an option of C-style representation.
Screenshot 2019-05-24 at 13 34 01

#### Environment

Safari

Add the same buttons as for DevEUI and JoinEUI

bafonins:
#### Summary

Closes #753 
This PR adds changed the type of `AppKey` and `NwkKey` inspectors to `byte` as well as makes the `NwkKey` label consistent with the Device Add Page (`Network Key` was used before)

Screenshot 2019-05-24 at 13 52 56

### DIFF
--- a/pkg/webui/console/views/device-overview/index.js
+++ b/pkg/webui/console/views/device-overview/index.js
@@ -105,7 +105,7 @@ class DeviceOverview extends React.Component {
         sensitive: false,
         subItems: [
           { key: sharedMessages.appKey, value: root_keys.app_key.key, type: 'byte', sensitive: true },
-          { key: sharedMessages.networkKey, value: root_keys.nwk_key.key, type: 'byte', sensitive: true },
+          { key: sharedMessages.nwkKey, value: root_keys.nwk_key.key, type: 'byte', sensitive: true },
         ],
       })
     }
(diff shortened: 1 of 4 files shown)


bafonins:
@htdvisser could you verify that this functionality is indeed needed?

johanstokking:
Note that `NwkKey` should be consistent with `AppKey`

furtiman:
@johanstokking should it be an idea to suggest settings according to the MAC version chosen while creating the device?

So if I choose 1.1, I get to fill these fields, and if I choose 1.0.4 I get the keys options according to the current spec? Otherwise the explanation of JoinEUI/AppEUI will be needed, and some other question will definitely arise from the users, since 1.1 spec is not used at the moment

kschiffer:
@furtiman This is considered for after Console MVP. See #579.

Issue #631, opened 2019-05-03
Fail downlink based on GS errors


johanstokking:
#### Summary

NS should fail absolute timed downlink that's too late to schedule.

#### Why do we need this?

So that a queued downlink item that can never be scheduled anymore by a GS does not keep halting the downlink flow until the queue is reset.

#### What is already there? What do you see now?

Currently, when you schedule a downlink message with an absolute time in the past or too soon in the future for any GS to schedule, the downlink task fails. However, the downlink item stays first in the queue, so it gets retried and keep failing until the queue is replaced with an item that can be scheduled.

GS now returns an `Aborted` error with error details per downlink path and Rx window that was specified. The error details code can be:

- `ResourceExhausted`: no more resources available, i.e. scheduling conflict
- `FailedPrecondition`: this is not going to work, i.e. too late – this is what should cause failure
- `Unavailable`: try again later, i.e. no clock sync; NS may retry later on this downlink path when there is clock sync (i.e. once there's uplink)
- `Aborted`: try another downlink path, i.e. this gateway doesn't have absolute time

#### What is missing? What do you want to see?

Based on errors returned by GS, NS should decide to fail the downlink and inform AS about that. Basically, when NS sees `FailedPrecondition` on all scheduling attempts, it can fail the downlink.

#### How do you propose to implement this?

NS should decide whether to fail the downlink based on GS error details codes. Suggesting the following;

- If all attempts are either `ResourceExhausted`, `FailedPrecondition`, `Aborted`, fail the downlink
- If any attempt is something else (i.e. `Unavailable`), try again later, like now

#### Can you do this yourself and submit a Pull Request?

If the proposed solution is acceptable for @rvolosatovs, then GS doesn't need a change. I'm open for suggestions that may involve specific GS errors.

rvolosatovs:
Sounds good to me!
This only concerns the "absolute time" downlink, right?
What about other cases?
Should NS *only* retry downlink if at least one of the errors in `Unavailable`?

johanstokking:
The proposed solution is for any timed downlink so: class A and class B/C with absolute time.

And indeed, it should only retry if there's any `Unavailable`.

Do we want to fail the join-accept too?

johanstokking:
There seems to have been regression causing an endless scheduling loop;

```
[...]

Issue #18, opened 2019-01-29
Retry downlink on Tx failure


johanstokking:
**Summary:** 

Retry downlink on Tx failure.

**Why do we need this?**

To perform a best-effort of (un)confirmed downlink.

**What is already there? What do you see now?**

The GS scheduler already detects and avoids downlink conflicts.

We already have a retry on the end device to NS level for confirmed downlink.

**What is missing? What do you want to see?**

Regardless of our amazing GS downlink scheduler, the gateway may still send a Tx acknowledgment with error code. In that case, even for unconfirmed downlink, the GS should try again.

**How do you propose to implement this?**

Once a downlink fail acknowledgement is received, unset all pending MAC state fields and:
1. call `updateDataDownlinkTask` https://github.com/TheThingsNetwork/lorawan-stack/blob/36d9597c89dfab1ef96d56bdd0e508a518b875c2/pkg/networkserver/downlink.go#L138 for data downlink fail. If possible, recreating `MACState.QueuedResponses` would be best
2. recreate `QueuedJoinAccept` and add a downlink task https://github.com/TheThingsNetwork/lorawan-stack/blob/36d9597c89dfab1ef96d56bdd0e508a518b875c2/pkg/networkserver/grpc_gsns.go#L1185-L1193 for join accept fail

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/794 by @egourlao

johanstokking:
Blocked by #76

nejraselimovic:
@johanstokking @rvolosatovs the blocking issue (#76) was closed recently. How important is this issue? Can we schedule it for Q2 or Q3?

johanstokking:
Indeed, this can now be implemented in NS.

johanstokking:
Missing details are actual reliable metrics on how many TX failures are reported by gateways as compared to successful transmissions. Also we need to understand what the TX failure is, and if we can mitigate them already in Gateway Server.

In The Things Stack, the gateway is under control by the Gateway Server. We do not expect gateways to do any scheduling, conflict mitigation or resolution, duty cycle enforcement, etc etc. There are a few indeterministic aspects, like timing (`TOO_LATE`), which we may mitigate as well with adjusting the schedule downlink late setting and RTT interpretation (take p90 instead of p50).

So, I'll get some data and report back here to see what we do with this issue.

johanstokking:
[...]

PR #756, opened 2019-05-24
Various downlink processing fixes and improvements

rvolosatovs:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/631
References https://github.com/TheThingsNetwork/lorawan-stack/issues/18

#### Changes

- Generate AS uplinks along with the downlink(that also fixes the issue of NS sending duplicates in case Class B/C device Rx1 is tried before Rx2)
- Send AS uplinks once the registry transaction is done
- Improve error handling
- Allow exiting early from registry `Set`-type transactions by returning `<non-nil>, nil, nil`
- Notify AS when downlink is expired
- Drop inapplicable application downlinks, schedule the first applicable one, if found
- Avoid retrying non-retryable errors


pkg/joinserver/redis/registry.go
```diff
@@ -160,6 +160,9 @@ func (r *DeviceRegistry) set(tx *redis.Tx, uid string, gets []string, f func(pb
 	if stored == nil && pb == nil {
 		return nil, nil
 	}
+	if pb != nil && len(sets) == 0 {
+		return stored, nil
```
johanstokking: Why not `pb = stored` like in all other Redis registries? We should keep this identical

pkg/networkserver/downlink.go
```diff
@@ -184,83 +197,111 @@ func (ns *NetworkServer) generateDownlink(ctx context.Context, dev *ttnpb.EndDev
...
-	var appDown *ttnpb.ApplicationDownlink
-	if len(cmdBuf) <= fOptsCapacity && len(dev.QueuedApplicationDownlinks) > 0 {
-		down := dev.QueuedApplicationDownlinks[0]
+	var start int
+	for _, down := range dev.QueuedApplicationDownlinks {
 		if len(down.FRMPayload) > int(maxDownLen) {
-			logger.Warn("Application downlink present, but the payload is too long, inform Application Server")
-			go func() {
-				ok, err := ns.handleASUplink(ctx, dev.EndDeviceIdentifiers.ApplicationIdentifiers, &ttnpb.ApplicationUp{
+			logger.Debug("Drop downlink with payload exceeding band regulations")
```
johanstokking: Add some log context on two levels;

1. On the message level, i.e. in the loop (like frame counter)
2. On the particular error level, i.e. in this case `frm_payload_size`

pkg/networkserver/downlink.go
```diff
@@ -184,83 +197,111 @@ func (ns *NetworkServer) generateDownlink(ctx context.Context, dev *ttnpb.EndDev
...
+						ApplicationDownlink: *down,
+						Error:               *ttnpb.ErrorDetailsToProto(errApplicationDownlinkTooLong),
+					},
+				},
+			})
+			start++
+			continue
+		}
+		if down.FCnt <= dev.Session.LastNFCntDown && dev.MACState.LoRaWANVersion.Compare(ttnpb.MAC_V1_1) < 0 {
+			logger.Debug("Drop downlink with too low FCnt")
```
johanstokking: Idem

pkg/networkserver/downlink.go
```diff
@@ -184,83 +197,111 @@ func (ns *NetworkServer) generateDownlink(ctx context.Context, dev *ttnpb.EndDev
...
+					},
+				},
+			})
+			start++
+			continue
+		}
+		switch dev.MACState.DeviceClass {
+		case ttnpb.CLASS_A:
+			if down.GetClassBC() != nil {
+				logger.Debug("Drop class B/C downlink for device in class A mode")
```
johanstokking: All devices are class A devices, even if class B or C is available. If application downlink indicates `class_b_c`, it should only be scheduled as class B/C downlink; as class A downlink it should stay in the queue, not dropped or failed.

rvolosatovs: So it should be skipped then?

pkg/joinserver/redis/registry.go
rvolosatovs: Because this is a different kind of function, this one returns the result and error, whilst others return just error.
We still have to return something here, sure we could make it
```go
pb = stored
return pb, nil
```
But that's redundant, since we can just `return stored, nil`

pkg/networkserver/downlink.go
johanstokking: Not skipped in terms of that we go to the next downlink; downlink should be transmitted in-order.

We should skip the downlink slot in class A if the first queued downlink is for class B/C only.

pkg/networkserver/errors.go
```diff
@@ -20,7 +20,9 @@ import (
 
 var (
 	errCIDOutOfRange             = errors.DefineInvalidArgument("cid_out_of_range", "CID must be in range from {min} to {max}")
+	errClassBCDownlinkForClassA  = errors.DefineInvalidArgument("class_bc_for_class_a", "class B/C downlink queued for class A device")
```
johanstokking: Not used anymore

pkg/networkserver/downlink.go
```diff
@@ -62,36 +62,76 @@ func deviceClassCTimeout(dev *ttnpb.EndDevice, defaults ttnpb.MACSettings) time.
...
+	pairs := []interface{}{
+		"confirmed", down.Confirmed,
+		"f_cnt", down.FCnt,
+		"f_port", down.FPort,
+		"frm_payload_len", len(down.FRMPayload),
+		"priority", down.Priority,
+		"session_key_id", down.SessionKeyID,
+	}
+	if down.GetClassBC() != nil {
+		pairs = append(pairs, "class_bc", true)
```
johanstokking: This (and onward) is technically `class_b_c`; that's also the field name

PR #388, opened 2019-03-28
Basic Station LNS support

KrishnaIyer:
**Summary:**

Closes #74. 

**Changes:**

- Support LNS Uplink and Downlink messages
- Adapt Downstream to reflect #358 
- Testing

**Notes for Reviewers:**

This PR Covers the traffic cases for the BasicStation protocol.

PR #762, opened 2019-05-27
BasicStation Support

### LINKED ISSUE #74: Support Basic Station LNS protocol
**Summary:** 

Support the [Basic Station protocol](https://doc.sm.tc/station/).

**Why do we need this?**

Mini Hub and new reference standard packet forwarder support.

Structure to support various gateway protocols in Gateway Server. Now implemented are gRPC, UDP and MQTT.

Support for Basic Station in the Gateway Server.

- Add `pkg/gatewayserver/io/basicstation` with [LNS Protocol](https://doc.sm.tc/station/tcproto.html) implementation. This is a HTTP websockets server
- The gateway connects first to query the endpoint to actually connect to (see [here](https://doc.sm.tc/station/tcproto.html#querying-the-lns-connection-end-point-discovery)). This means that we expose two websockets servers on two separate ports
- Then, the gateway connects to the router; the GS sends the frequency plan first, and then there's uplink/downlink traffic
- The Basic Station has it's own scheduler, i.e. `TxRequest` can be transmitted as-is to the gateway
- Add a test case in `pkg/gatewayserver/gatewayserver_test.go`

TODO
- [x] Discovery 
- [x] Fetching config including sx1301_conf
- [x] Upstream
- [x] Downstream
- [x] Check header for Auth tokens
- [ ] Support new listeners on 1887/8887.

~1. @johanstokking to prepare branch with layout~
~2. @johanstokking to get support, the Linux based test forwarder and/or test script~
3. @KrishnaIyer to implement the LNS protocol

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/1412 by @johanstokking

KrishnaIyer:
#### Summary

Closes #74 

#### Changes

- A simple PR that merges all the changes to `feature/74-lns` to `master`.

#### Notes for Reviewers

No changes are made in this PR. It's only to merge to master.
If anything else is to be added to the release notes section below, please indicate that.

#### Release Notes

- Added native support to the Basic Station LNS protocol in the Gateway Server.


KrishnaIyer:
blocked on #796

KrishnaIyer:
@johanstokking: We can finally merge this 💯

pkg/gatewayserver/gatewayserver.go
```diff
@@ -150,6 +151,12 @@ func New(c *component.Component, conf *Config) (gs *GatewayServer, err error) {
 	}
 
 	hooks.RegisterUnaryHook("/ttn.lorawan.v3.NsGs", rpclog.NamespaceHook, rpclog.UnaryNamespaceHook("gatewayserver"))
+	bsCtx := gs.FillContext(ctx)
```
htdvisser: This doesn't look quite right. `FillContext` is intended to fill request contexts, not to extend the component context.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,350 @@
...
+
+func (s *srv) handleDiscover(c echo.Context) error {
+	ctx := c.Request().Context()
+	logger := log.FromContext(ctx).WithFields(log.Fields(
+		"endpoint", "discover",
+		"remote_addr", c.Request().RemoteAddr,
+	))
+	ws, err := s.upgrader.Upgrade(c.Response(), c.Request(), nil)
+	if err != nil {
+		logger.WithError(err).Debug("Failed to upgrade request to websocket connection")
```
htdvisser: Feels a bit strange to log "failures" (as you call them) as debug level here and below. They should be warn. Also, I would write "could not" instead of "failed to", because this is something the server can't do because the client screwed up, not because the server failed.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,350 @@
...
+		return err
+	}
+	var req messages.DiscoverQuery
+	if err := json.Unmarshal(data, &req); err != nil {
+		logger.WithError(err).Debug("Failed to parse discover query message")
+		return err
+	}
+
+	if req.EUI.IsZero() {
+		writeDiscoverError(s.ctx, ws, "Invalid request")
```
htdvisser: Is this "Invalid request" because of specification? Otherwise I would be more specific in telling why the request was invalid.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,350 @@
...
+		return errEmptyGatewayEUI
+	}
+
+	ids := ttnpb.GatewayIdentifiers{
+		EUI: &req.EUI.EUI64,
+	}
+	ctx, ids, err = s.server.FillGatewayContext(ctx, ids)
+	if err != nil {
+		logger.WithError(err).Debug("Failed to fill gateway context")
+		writeDiscoverError(ctx, ws, "Router not provisioned")
```
htdvisser: Again, is this because of spec? Otherwise include the err.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,350 @@
...
+
+func (s *srv) handleTraffic(c echo.Context) error {
+	var latestUpstreamXTime int64
+	id := c.Param("id")
+	auth := c.Request().Header.Get(echo.HeaderAuthorization)
+	ctx := c.Request().Context()
+	if auth != "" {
+		md := metadata.New(map[string]string{
+			"id":            id,
+			"authorization": fmt.Sprintf("Bearer %s", auth),
```
htdvisser: To be future-proof we may want to check that `auth` doesn't already contain `Bearer`

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	logger := log.FromContext(ctx).WithFields(log.Fields(
+		"endpoint", "traffic",
+		"remote_addr", c.Request().RemoteAddr,
+	))
+
+	ctx, ids, err := s.server.FillGatewayContext(ctx, ttnpb.GatewayIdentifiers{GatewayID: id})
+	if err != nil {
+		logger.WithError(err).Debug("Failed to fill gateway context")
```
htdvisser: This kind of failure should not be logged as debug

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,350 @@
...
+		return err
+	}
+
+	if auth == "" {
```
htdvisser: I don't get this. If there is no auth, we just give the gateway link rights?
This would at least deserve a comment explaining why.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,350 @@
...
+	defer ws.Close()
+
+	go func() {
+		for {
+			select {
+			case <-conn.Context().Done():
+				return
+			case down := <-conn.Down():
+				dlTime := time.Now()
+				dnmsg := messages.FromDownlinkMessage(ids, *down, int64(s.tokens.Next(down.CorrelationIDs, dlTime)), dlTime, latestUpstreamXTime)
```
htdvisser: Unprotected concurrent read of `latestUpstreamXTime`

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,350 @@
...
+				msg, err := dnmsg.MarshalJSON()
+				if err != nil {
+					logger.WithError(err).Error("Failed to marshal downlink message")
+					continue
+				}
+
+				logger.Info("Send downlink message")
+				if err := ws.WriteMessage(websocket.TextMessage, msg); err != nil {
+					logger.WithError(err).Error("Failed to send downlink message")
+					conn.Disconnect(err)
```
htdvisser: Shouldn't this be a return, while having the Disconnect in a defer?

pkg/gatewayserver/io/basicstationlns/messages/messages.go
```diff
@@ -0,0 +1,214 @@
...
+	FrequencyRange []int                   `json:"freq_range"`
+	DataRates      DataRates               `json:"DRs"`
+	SX1301Config   []pfconfig.SX1301Config `json:"sx1301_conf"`
+
+	// These are debug options to be unset in production gateways
+	NoCCA       bool `json:"nocca"`
+	NoDutyCycle bool `json:"nodc"`
+	NoDwellTime bool `json:"nodwell"`
+
+	// TODO: Roundtrip time monitoring
```
htdvisser: TODOs are written as `// TODO: Some Description (https://link/to/issue)`
Also, it looks like this is already done, right?

pkg/gatewayserver/io/basicstationlns/messages/messages.go
```diff
@@ -0,0 +1,214 @@
...
+
+	s := strings.Split(band.ID, "_")
+	if len(s) < 2 {
+		return RouterConfig{}, errFrequencyPlan
+	}
+	cfg.Region = fmt.Sprintf("%s%s", s[0], s[1])
+	if len(fp.Radios) == 0 {
+		return RouterConfig{}, errFrequencyPlan
+	}
+	// TODO: Handle FP with multiple radios if necessary
```
htdvisser: Missing TODO link to issue

pkg/gatewayserver/io/basicstationlns/messages/messages.go
```diff
@@ -0,0 +1,214 @@
...
+	cfg.FrequencyRange = []int{
+		int(fp.Radios[0].TxConfiguration.MinFrequency),
+		int(fp.Radios[0].TxConfiguration.MaxFrequency),
+	}
+
+	// TODO: https://github.com/TheThingsNetwork/lorawan-stack/issues/761
```
htdvisser: Missing TODO description

pkg/gatewayserver/io/basicstationlns/messages/messages.go
```diff
@@ -0,0 +1,214 @@
...
+	sx1301Conf, err := pfconfig.BuildSX1301Config(&fp)
+	if err != nil {
+		return RouterConfig{}, err
+	}
+
+	// These fields are not defined in the v1.5 ref design https://doc.sm.tc/station/gw_v1.5.html#rfconf-object and would cause a parsing error.
+	sx1301Conf.Radios[0].TxFreqMin = 0
+	sx1301Conf.Radios[0].TxFreqMax = 0
+
+	// TODO: Extend this for > 8ch gateways https://github.com/TheThingsNetwork/lorawan-stack/issues/761
```
htdvisser: TODO issue link between ()

pkg/gatewayserver/io/basicstationlns/messages/messages.go
```diff
@@ -0,0 +1,214 @@
...
+	cfg.SX1301Config = append(cfg.SX1301Config, *sx1301Conf)
+
+	//Add the MuxTime for RTT measurement
```
htdvisser: We write a space after `//`

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
KrishnaIyer: The protocol does not define error messages nor does it process the text. It's directly logged.
These are just adapted from the example python bridge. 
https://github.com/lorabasics/basicstation/blob/c29b8502f8c715daecec6666835da6e981dc820a/src/tc.c#L221

(PS: This error would never occur in the first place, since the station itself has error handling for invalid EUIs and would not start if the config is wrong.)

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
KrishnaIyer: Here also, the actual error message is not processed but just logged. 
The current way of handling it in the station is to retry for a certain no of times and then shutdown.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
KrishnaIyer: Yes. Indeed. `FillGatewayContext` will assert the rights later down the road. For gateways without an auth but a valid EUI, it works the same as UDP (where we provide rights ourselves). Will add the comment.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
KrishnaIyer: Great catch. This needed an update to the exit handling in the upstream routine as well. 👍

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
htdvisser: What I wrote in https://github.com/TheThingsNetwork/lorawan-stack/pull/762#pullrequestreview-249869639. Fix it both here and in other places.

pkg/gatewayserver/gatewayserver.go
KrishnaIyer: Hmm ok. This is again from the original plumbing code.

If that's the case, then we should directly pass the component context here and use this function on the request context in the echo handler right?

cc: @johanstokking

johanstokking: Yes

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
johanstokking: Let's not trigger log warnings from garbage requests from clients.

Regarding the message; it doesn't say the server failed, it says an action failed. The blame is in the cause.

We do not (consistently) differentiate "could not" vs "failed to" to indicate external vs internal. I also don't see that feasible to implement because it fully depends on the cause. And we should not be switching on cause codes to print the "right" log message.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
johanstokking: I disagree about the message but the log level can be higher here.

pkg/gatewayserver/io/basicstationlns/messages/messages.go
johanstokking: FTR; I approved this in other PRs as they were targeting a feature branch. But this should not go to `master` indeed

pkg/gatewayserver/gatewayserver.go
KrishnaIyer: FTR: no need for the above. `FillContext` is already called by `FillGatewayContext`.

KrishnaIyer:
@johanstokking: Pinging you here for a review request.

api/lorawan.proto
```diff
@@ -242,6 +242,13 @@ message DataRate {
   };
 }
 
+
+message RequestInfo{
```
johanstokking: Spacing

johanstokking: What does "RequestInfo" mean on the V3 API level?

johanstokking: It doesn't seem to be used. Is this ready to review?

KrishnaIyer: wtf? Where did this even come from? On it.

KrishnaIyer: This was a commit that was left over from an old rebase mistake. It's dropped now.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,363 @@
...
+					Rights: []ttnpb.Right{ttnpb.RIGHT_GATEWAY_LINK},
+				},
+			},
+		})
+	}
+
+	uid := unique.ID(ctx, ids)
+	ctx = log.NewContextWithField(ctx, "gateway_uid", uid)
+
+	fp, err := s.server.GetFrequencyPlan(ctx, ids)
```
johanstokking: This call is intended for gateways themselves requesting the frequency plan, like gRPC frontend. It's quite an expensive call goes it hits IS (including auth check) and the frequency plan store (by default GitHub).

This call is also performed by `Connect()` so let's do it once. To make FP available here, please expose it through a getter call on `Connection`

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,363 @@
...
+				dnmsg := messages.FromDownlinkMessage(ids, *down, int64(s.tokens.Next(down.CorrelationIDs, dlTime)), dlTime, atomic.LoadInt64(&latestUpstreamXTime))
+				msg, err := dnmsg.MarshalJSON()
+				if err != nil {
+					logger.WithError(err).Error("Failed to marshal downlink message")
+					continue
+				}
+
+				logger.Info("Send downlink message")
+				if err := ws.WriteMessage(websocket.TextMessage, msg); err != nil {
+					logger.WithError(err).Error("Failed to send downlink message")
```
johanstokking: I wouldn't call this an `Error` but a `Warning` as it's (probably) a premature client disconnection

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,363 @@
...
+					return
+				}
+			}
+		}
+	}()
+
+	for {
+		select {
+		case <-conn.Context().Done():
+			return nil
```
johanstokking: It would be nice if we can simply `return conn.Context().Err()` here and gracefully handle context cancellation by the caller

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,363 @@
...
+					if err := conn.HandleTxAck(&txAck); err != nil {
+						logger.WithField("diid", txConf.Diid).Warn("Failed to handle Tx acknowledgement")
+					}
+				} else {
+					logger.WithField("diid", txConf.Diid).Debug("Tx acknowledgement either does not correspond to a downlink message or arrived too late")
+				}
+				recordRTT(conn, receivedAt, txConf.RefTime)
+
+			case messages.TypeUpstreamProprietaryDataFrame:
+				return errMessageTypeNotImplemented.WithAttributes("type", typ)
```
johanstokking: So we break the entire traffic handling when a gateway receives a proprietary message?

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,363 @@
...
+			case messages.TypeUpstreamRemoteShell:
+				return errMessageTypeNotImplemented.WithAttributes("type", typ)
```
johanstokking: Can't we safely ignore this?

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,363 @@
...
+			case messages.TypeUpstreamTimeSync:
+				return errMessageTypeNotImplemented.WithAttributes("type", typ)
```
johanstokking: Idem

pkg/gatewayserver/io/basicstationlns/basicstationlns_test.go
```diff
@@ -0,0 +1,1030 @@
...
+			Query:    messages.DiscoverQuery{},
+			Response: messages.DiscoverResponse{Error: "Empty router EUI provided"},
+		},
+		{
+			Query:    struct{}{},
+			Response: messages.DiscoverResponse{Error: "Empty router EUI provided"},
+		},
+		{
+			Query: struct {
+				EUI string `json:"route"`
```
johanstokking: Is the test here that `route` is not `router`?

pkg/gatewayserver/io/basicstationlns/basicstationlns_test.go
```diff
@@ -0,0 +1,1030 @@
...
+				EUI string `json:"router"`
+			}{EUI: `"01:02:03:04:05:06:07:08:09"`},
+		},
+		{
+			Query: struct {
+				EUI string `json:"router"`
+			}{EUI: `"01:02:03:04:05:06:07-08"`},
+		},
+	} {
+		t.Run(fmt.Sprintf("InvalidQuery/%d", i), func(t *testing.T) {
```
johanstokking: Why are these invalid? Please add comment(s)

pkg/gatewayserver/io/basicstationlns/basicstationlns_test.go
```diff
@@ -0,0 +1,1030 @@
...
+		t.Fatalf("Connection failed: %v", err)
+	}
+	defer conn.Close()
+
+	for _, tc := range []struct {
+		Name                 string
+		VersionQuery         interface{}
+		ExpectedRouterConfig interface{}
+	}{
+		// These tests will fail without unmarshalling of Sx1301_conf https://github.com/TheThingsNetwork/lorawan-stack/issues/808
```
johanstokking: So we can't test the router config without this issue? Seems quite important to me

pkg/gatewayserver/io/basicstationlns/basicstationlns_test.go
```diff
@@ -0,0 +1,1030 @@
...
+					t.Fatalf("Failed to unmarshal response `%s`: %v", string(res), err)
+				}
+				a.So(response, should.Resemble, tc.ExpectedRouterConfig)
+			case <-time.After(timeout):
+				t.Fatalf("Read message timeout")
+			}
+		})
+	}
+}
+func TestTraffic(t *testing.T) {
```
johanstokking: Newline

pkg/gatewayserver/io/basicstationlns/messages/downstream.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Chosen fixed values.
+	dnmsg.Priority = 25
+	dnmsg.RxDelay = 1
+
+	// Fix the Tx Parameters since we don't use the gateway scheduler.
+	dnmsg.Rx1DR = int(scheduledMsg.DataRateIndex)
+	dnmsg.Rx1Freq = int(scheduledMsg.Frequency)
+
+	//Add the MuxTime for RTT measurement
```
johanstokking: Spacing

pkg/gatewayserver/io/basicstationlns/messages/downstream.go
```diff
@@ -0,0 +1,117 @@
...
+	dnmsg.MuxTime = float64(dlTime.Unix()) + float64(dlTime.Nanosecond())/(1e9)
```
johanstokking: `float64(dlTime.UnixNano())/float64(time.Second)` ?

pkg/gatewayserver/io/basicstationlns/messages/downstream.go
```diff
@@ -0,0 +1,117 @@
...
+	// Always use the Basic Station ClassB mode for absolute time scheduling.
+	if scheduledMsg.Time != nil {
+		dnmsg.DeviceClass = uint(ttnpb.CLASS_B)
+		dnmsg.GpsTime = scheduledMsg.Time.Unix()
+		return dnmsg
+	}
+
+	dnmsg.DeviceClass = uint(ttnpb.CLASS_A)
+	// Estimate the xtime based on the timestamp; xtime = timestamp - (rxdelay+1).
+
```
johanstokking: Swap L86 and L87

pkg/gatewayserver/io/basicstationlns/messages/downstream.go
```diff
@@ -0,0 +1,117 @@
...
+	t := time.Unix(int64(scheduledMsg.Timestamp), 0)
+	offset := time.Duration((dnmsg.RxDelay)) * time.Second * (1e6)
+	dnmsg.XTime = latestXTime&0x7FFFFFFF00000000 | (t.Add(-offset).Unix())
```
johanstokking: Remove last parentheses

pkg/gatewayserver/io/basicstationlns/messages/downstream.go
```diff
@@ -0,0 +1,117 @@
...
+		dnmsg.DeviceClass = uint(ttnpb.CLASS_B)
+		dnmsg.GpsTime = scheduledMsg.Time.Unix()
+		return dnmsg
+	}
+
+	dnmsg.DeviceClass = uint(ttnpb.CLASS_A)
+	// Estimate the xtime based on the timestamp; xtime = timestamp - (rxdelay+1).
+
+	t := time.Unix(int64(scheduledMsg.Timestamp), 0)
+	offset := time.Duration((dnmsg.RxDelay)) * time.Second * (1e6)
```
johanstokking: Why do we need another `1e6` ? What is the unit of `offset`?

johanstokking: This needs documentation

pkg/gatewayserver/io/basicstationlns/messages/downstream.go
johanstokking: Also we're not accounting for rollover here. This may make `dnmsg.XTime` smaller than `latestXTime`

johanstokking:
BTW, you can use `pkg/gatewayserver/scheduling.RolloverClock` for the timing, because `ConcentratorTime == xtime`, both `int64` also. You just need to add a variant of `Sync()` that takes the `ConcentratorTime` instead of `uint32`. And mind the locking; `RolloverClock` needs a mutex to work with.

pkg/gatewayserver/io/basicstationlns/basicstationlns_test.go
KrishnaIyer: Yes 👍 . Added comments

pkg/gatewayserver/io/basicstationlns/messages/downstream.go
KrishnaIyer: Nice

pkg/gatewayserver/io/basicstationlns/basicstationlns_test.go
KrishnaIyer: router config is tested and verified on the TTIG. The only reason why un-marshaling of SX1301_conf is not implemented is due to performance reasons (of the un-marshaler). I don't think we need to block this PR for that.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,363 @@
...
+		writeDiscoverError(s.ctx, ws, "Empty router EUI provided")
+		return errEmptyGatewayEUI
+	}
+
+	ids := ttnpb.GatewayIdentifiers{
+		EUI: &req.EUI.EUI64,
+	}
+	ctx, ids, err = s.server.FillGatewayContext(ctx, ids)
+	if err != nil {
+		writeDiscoverError(ctx, ws, fmt.Sprintf("Failed to fetch gateway: %s", err.Error()))
```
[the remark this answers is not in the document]
KrishnaIyer: https://github.com/TheThingsNetwork/lorawan-stack/pull/762#discussion_r294211470

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -903,15 +1037,8 @@ func TestGatewayServer(t *testing.T) {
...
-							a.So(ok, should.BeTrue)
-							if conn.HasScheduler() {
-								settings := msg.DownlinkMessage.GetScheduled()
-								a.So(settings, should.NotBeNil)
-							} else {
-								request := msg.DownlinkMessage.GetRequest()
-								a.So(request, should.NotBeNil)
-							}
+							settings := msg.DownlinkMessage.GetScheduled()
+							a.So(settings, should.NotBeNil)
```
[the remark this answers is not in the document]
KrishnaIyer: This diff was added in one of the earlier PRs when if `ptc.HasScheduler` was removed and left-indented.

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -663,13 +792,17 @@ func TestGatewayServer(t *testing.T) {
...
 						case upCh <- tc.Up:
 						case <-time.After(timeout):
 							t.Fatalf("Failed to send message to upstream channel")
 						}
-						uplinkCount += len(tc.Up.UplinkMessages)
+						if ptc.SupportsStatus {
+							uplinkCount += len(tc.Up.UplinkMessages)
+						} else {
+							// If the protocol does not support Status Messages, then count only the uplinks.
+							uplinkCount += len(tc.Forwards)
```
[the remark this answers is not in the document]
KrishnaIyer: https://github.com/TheThingsNetwork/lorawan-stack/pull/388#issuecomment-489311853

pkg/gatewayserver/gatewayserver.go
```diff
@@ -235,14 +241,17 @@ var (
...
 	}
 
 	uid := unique.ID(ctx, ids)
-	logger := log.FromContext(ctx).WithField("gateway_uid", uid)
-	ctx = events.ContextWithCorrelationID(ctx, fmt.Sprintf("gs:conn:%s", events.NewCorrelationID()))
+	logger := log.FromContext(ctx).WithFields(log.Fields(
+		"protocol", frontend.Protocol(),
+		"gateway_uid", uid,
+	))
+	ctx = events.ContextWithCorrelationID(ctx, fmt.Sprintf("gateway_conn:%s", events.NewCorrelationID()))
```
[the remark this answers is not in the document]
KrishnaIyer: regression of correlation IDs?

KrishnaIyer:
FTR: There are no changes made in this PR (except log levels). All the requested changes so far are from already approved PRs to the feature branch.

pkg/gatewayserver/gatewayserver.go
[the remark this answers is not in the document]
johanstokking: Yes

pkg/gatewayserver/gatewayserver_test.go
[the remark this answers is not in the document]
johanstokking: How is that related, i.e. how are status messages related to `tc.Up.UplinkMessages` and `tc.Forwards`?

pkg/gatewayserver/io/basicstationlns/basicstationlns_test.go
johanstokking: Performance reasons?

johanstokking:
> FTR: There are no changes made in this PR (except log levels). All the requested changes so far are from already approved PRs to the feature branch.

Yes that's good, but it's good to review again now that everything comes together

pkg/gatewayserver/gatewayserver_test.go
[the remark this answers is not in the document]
KrishnaIyer: We discussed this in the other PR:
> Basic Station doesn't know what to do with the GatewayStatus messages. If there's no conditional check, the gateway status test case will fail. Same with UplinkCount. The other frontends forward the status messages and BS doesn't. The uplink count will be one lesser than the other frontends.

pkg/gatewayserver/io/basicstationlns/basicstationlns_test.go
KrishnaIyer: from @htdvisser :
> 1. Because we didn't actually have a need to unmarshal it
> 2. Because the stdlib encoding/json package would be inefficient at unmarshaling it

This is why the linked issue was created. I can simply use an "inefficient" marshaller for now but only if it's really needed.

KrishnaIyer:
I think this is ready for review @johanstokking.

pkg/gatewayserver/gatewayserver_test.go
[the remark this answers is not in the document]
johanstokking: OK but;

> How is that related, i.e. how are status messages related to `tc.Up.UplinkMessages` and `tc.Forwards`?

The `tc.Forwards` says which indices of `tc.Up.UplinkMessages` should be forwarded. The others are invalid and should be dropped.

The `uplinkCount` gets compared to `stats.UplinkCount` and that is only about uplink messages.

I don't see how this is related to gateway status?

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,354 @@
...
+	if err := ws.WriteMessage(websocket.TextMessage, data); err != nil {
+		logger.WithError(err).Warn("Failed to write discover response message")
+		return err
+	}
+	logger.Debug("Sent discover response message")
+	return nil
+}
+
+func (s *srv) handleTraffic(c echo.Context) error {
+	var latestUpstreamXTime int64
```
johanstokking: Why not use `RolloverClock`?

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
johanstokking: This is not addressed

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,354 @@
...
+	}()
+
+	for {
+		select {
+		case <-conn.Context().Done():
+			return conn.Context().Err()
+		default:
+			_, data, err := ws.ReadMessage()
+			if err != nil {
+				logger.WithError(err).Error("Failed to read message")
```
johanstokking: Warning

pkg/gatewayserver/io/basicstationlns/messages/downstream.go
```diff
@@ -0,0 +1,118 @@
...
+		dnmsg.DeviceClass = uint(ttnpb.CLASS_B)
+		dnmsg.GpsTime = scheduledMsg.Time.Unix()
+		return dnmsg
+	}
+
+	dnmsg.DeviceClass = uint(ttnpb.CLASS_A)
+
+	// Estimate the xtime based on the timestamp; xtime = timestamp - (rxdelay+1).
+	// The calculated offset is in microseconds.
+	t := time.Unix(int64(scheduledMsg.Timestamp), 0)
```
johanstokking: `scheduledMsg.Timestamp` is microseconds, while `time.Unix()` takes seconds

pkg/gatewayserver/io/basicstationlns/messages/downstream.go
```diff
@@ -0,0 +1,118 @@
...
+	offset := time.Duration((dnmsg.RxDelay)) * time.Second * 1e6
```
johanstokking: I still don't get this unit; this becomes millions of seconds? Note that `time.Second == 1e6` already

pkg/gatewayserver/io/basicstationlns/messages/downstream.go
```diff
@@ -0,0 +1,118 @@
...
+	dnmsg.XTime = int64(uint64(latestXTime)&0xFFFFFFFF00000000 | uint64(t.Add(-offset).Unix()))
```
johanstokking: This still doesn't account for rollover. Please use `RolloverClock` and cast `ConcentratorTime` to `int64` (lossless at it is `int64` already and accounts for rollover)

KrishnaIyer: Hmm. I don't get that actually. The upper byte not used as an incremental counter but as a session id (whose value is random).
This is separate from the actual time value in the lower byte. 
So a rollover will only occur if the whole 64bits are interpreted as a single unit which is not the case here. As long as the time value (lower 32) is appended to the stored upper bytes, there shouldn't be a rollover issue right?

johanstokking: See offline communication

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
KrishnaIyer: As discussed offline, this won't be necessary since we don't need to account for rollover.

pkg/gatewayserver/io/basicstationlns/messages/downstream.go
KrishnaIyer: Cleaned up this part. The problem was converting `timestamp` into `time.Time` and using Unix seconds whereas all the involved entities are in microseconds.

KrishnaIyer:
rebase and ready for (hopefully) final review @johanstokking!

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+	logger.Debug("Sent discover response message")
+	return nil
+}
+
+func (s *srv) handleTraffic(c echo.Context) error {
+	var latestUpstreamXTime int64
+	id := c.Param("id")
+	var auth string
+	auth = c.Request().Header.Get(echo.HeaderAuthorization)
```
johanstokking: `auth :=`

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
johanstokking: For your future self and other devs; please make this `sessionID int16` that's already `xtime >> 48`

In which situation will this value change while the gateway is connected?

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,362 @@
...
+			case down := <-conn.Down():
+				dlTime := time.Now()
+				scheduledMsg := down.GetScheduled()
+				latestXTime := atomic.LoadInt64(&latestUpstreamXTime)
+
+				// Convert to microseconds
+				concentratorTimeMs := conn.GetConcentratorTime(scheduledMsg.Timestamp) / 1000
+
+				// The first 16 bits of XTime gets the session ID from the upstream latestXTime and the other 48 bits are concentrator timestamp accounted for rollover.
+				xTime := int64(uint64(latestXTime)&0xFFFF000000000000 | uint64(concentratorTimeMs)&0x0000FFFFFFFFFFFF)
```
johanstokking: Make this as easy and clear as possible;

```go
xTime := int64(sessionID) << 48 | conn.GetConcentratorTime(scheduledMsg.Timestamp) / int64(time.Microsecond)
```

pkg/gatewayserver/io/basicstationlns/messages/downstream.go
```diff
@@ -0,0 +1,107 @@
...
+	dnmsg.Rx1Freq = int(scheduledMsg.Frequency)
+
+	// Add the MuxTime for RTT measurement
+	dnmsg.MuxTime = float64(dlTime.UnixNano()) / float64(time.Second)
+
+	// The GS controls the scheduling and hence for the gateway, its always Class A.
+	dnmsg.DeviceClass = uint(ttnpb.CLASS_A)
+
+	// Estimate the xtime based on the timestamp; xtime = timestamp - (rxdelay+1). The calculated offset is in microseconds.
+	dnmsg.XTime = xTime - int64(dnmsg.RxDelay*(1e6))
```
johanstokking:
```go
dnmsg.RxDelay*time.Second/time.Microsecond
```

Compiler optimizes this away

pkg/gatewayserver/io/basicstationlns/messages/downstream.go
```diff
@@ -0,0 +1,107 @@
...
+	DevEUI      string  `json:"DevEui"`
+	DeviceClass uint    `json:"dC"`
+	Diid        int64   `json:"diid"`
+	Pdu         string  `json:"pdu"`
+	RxDelay     int     `json:"RxDelay"`
+	Rx1DR       int     `json:"RX1DR"`
+	Rx1Freq     int     `json:"RX1Freq"`
+	Priority    int     `json:"priority"`
+	XTime       int64   `json:"xtime"`
+	GpsTime     int64   `json:"gpstime"`
```
johanstokking: Isn't this optional?

pkg/gatewayserver/io/basicstationlns/messages/util.go
```diff
@@ -0,0 +1,140 @@
...
+
+	"go.thethings.network/lorawan-stack/pkg/band"
+	"go.thethings.network/lorawan-stack/pkg/errors"
+	"go.thethings.network/lorawan-stack/pkg/ttnpb"
+	"go.thethings.network/lorawan-stack/pkg/types"
+)
+
+var errDataRateIndex = errors.Define("data_rate_index", "data rate index is out of range")
+var errDataRate = errors.DefineNotFound("data_rate", "data rate not found")
+var errUID = errors.DefineInvalidArgument("uid", "invalid uid `{uid}`")
```
johanstokking: Group vars

pkg/gatewayserver/io/basicstationlns/messages/util.go
```diff
@@ -0,0 +1,140 @@
...
+	"reflect"
+	"strings"
+
+	"go.thethings.network/lorawan-stack/pkg/band"
+	"go.thethings.network/lorawan-stack/pkg/errors"
+	"go.thethings.network/lorawan-stack/pkg/ttnpb"
+	"go.thethings.network/lorawan-stack/pkg/types"
+)
+
+var errDataRateIndex = errors.Define("data_rate_index", "data rate index is out of range")
```
johanstokking: `DefineInvalidArgument`

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
KrishnaIyer: Yeah alright. the reason for keeping the existing int64 is to continue using the `atomic` RW. Updated to use a regular mutex.

johanstokking: But do we need a mutex? Is this value going to change during a connection, or is it just a value that needs to become available through uplink?

johanstokking: Refs https://github.com/TheThingsNetwork/lorawan-stack/pull/762#discussion_r301114566

KrishnaIyer: According to the code, it is (currently) constant per connection. But there are 2 reasons why it's better to keep updating this value on every uplink:
1. Constant session ID per connection is [current](https://github.com/lorabasics/basicstation/blob/c29b8502f8c715daecec6666835da6e981dc820a/src/timesync.c#L379) behavior. The fact that the example python bridge [updates this value every uplink](https://github.com/lorabasics/basicstation/blob/c29b8502f8c715daecec6666835da6e981dc820a/examples/station2pkfwd/router.py#L135) suggests, that this may change in the future.
2. Even if we update the variable `sessionID` only once per connection, it might be read in the DL routine concurrently while this is happening. Even if it's rare, I still think it's better to protect the concurrent read in this case.

KrishnaIyer:
blocked on #986

pkg/gatewayserver/gatewayserver.go
```diff
@@ -171,6 +173,36 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 	}
 
 	hooks.RegisterUnaryHook("/ttn.lorawan.v3.NsGs", rpclog.NamespaceHook, rpclog.UnaryNamespaceHook("gatewayserver"))
+	if conf.BasicStation.FallbackFrequencyPlanID != "" {
+		ctx = frequencyplans.WithFallbackID(ctx, conf.BasicStation.FallbackFrequencyPlanID)
```
htdvisser: Can we scope this `ctx` to the BS frontend so that it doesn't influence frontends that we may add in the future?

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,387 @@
...
+func (*srv) Protocol() string { return "basicstation" }
+
+// New creates the Basic Station front end.
+func New(ctx context.Context, server io.Server) *echo.Echo {
+	webServer := echo.New()
+	webServer.Logger = web.NewNoopLogger()
+	webServer.HTTPErrorHandler = errorHandler
+	webServer.Use(
+		middleware.ID(""),
+		echomiddleware.BodyLimit("16M"),
```
htdvisser: This can definitely be smaller than 16M

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
htdvisser: For the record: I'm still not quite happy with this, but ok

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,387 @@
...
+	ws, err := s.upgrader.Upgrade(c.Response(), c.Request(), nil)
+	if err != nil {
+		logger.WithError(err).Debug("Failed to upgrade request to websocket connection")
+		return err
+	}
+	defer ws.Close()
+
+	_, data, err := ws.ReadMessage()
+	if err != nil {
+		logger.WithError(err).Warn("Failed to read message")
```
htdvisser: If we're following @johanstokking's comment above

> Let's not trigger log warnings from garbage requests from clients.

Then we should be consistent and do the same here

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,387 @@
...
+		return err
+	}
+	var req messages.DiscoverQuery
+	if err := json.Unmarshal(data, &req); err != nil {
+		logger.WithError(err).Warn("Failed to parse discover query message")
```
htdvisser: And here

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,387 @@
...
+	}()
+
+	for {
+		select {
+		case <-conn.Context().Done():
+			return conn.Context().Err()
+		default:
+			_, data, err := ws.ReadMessage()
+			if err != nil {
+				logger.WithError(err).Warn("Failed to read message")
```
htdvisser: And here

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,387 @@
...
+				conn.Disconnect(err)
+				return nil
+			}
+
+			typ, err := messages.Type(data)
+			if err != nil {
+				logger.WithError(err).Warn("Failed to parse message type")
```
htdvisser: And here

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,387 @@
...
+			logger = logger.WithFields(log.Fields(
+				"upstream_type", typ,
+			))
+			receivedAt := time.Now()
+
+			switch typ {
+			case messages.TypeUpstreamVersion:
+				var version messages.Version
+				if err := json.Unmarshal(data, &version); err != nil {
+					logger.WithError(err).Warn("Failed to unmarshal version message")
```
htdvisser: And here

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,387 @@
...
+				}
+				if err := ws.WriteMessage(websocket.TextMessage, data); err != nil {
+					logger.WithError(err).Warn("Failed to send router configuration")
+					return err
+				}
+
+			case messages.TypeUpstreamJoinRequest:
+				var jreq messages.JoinRequest
+				if err := json.Unmarshal(data, &jreq); err != nil {
+					logger.WithError(err).Warn("Failed to unmarshal join-request message")
```
htdvisser: And here

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,387 @@
...
+					return nil
+				}
+				up, err := jreq.ToUplinkMessage(ids, fp.BandID, receivedAt)
+				if err != nil {
+					logger.WithError(err).Warn("Failed to parse join-request message")
```
htdvisser: And here

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,387 @@
...
+				if err := conn.HandleUp(up); err != nil {
+					logger.WithError(err).Warn("Failed to handle uplink message")
+				}
+				recordRTT(conn, receivedAt, jreq.RefTime)
+				atomic.StoreInt32(&sessionID, int32(jreq.UpInfo.XTime>>48))
+
+			case messages.TypeUpstreamUplinkDataFrame:
+				var updf messages.UplinkDataFrame
+				if err := json.Unmarshal(data, &updf); err != nil {
+					logger.WithError(err).Warn("Failed to unmarshal uplink data frame")
```
htdvisser: And here

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,387 @@
...
+					return nil
+				}
+				up, err := updf.ToUplinkMessage(ids, fp.BandID, receivedAt)
+				if err != nil {
+					logger.WithError(err).Warn("Failed to parse uplink data frame")
```
htdvisser: And here

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,387 @@
...
+				if err := conn.HandleUp(up); err != nil {
+					logger.WithError(err).Warn("Failed to handle uplink message")
+				}
+				recordRTT(conn, receivedAt, updf.RefTime)
+				atomic.StoreInt32(&sessionID, int32(updf.UpInfo.XTime>>48))
+
+			case messages.TypeUpstreamTxConfirmation:
+				var txConf messages.TxConfirmation
+				if err := json.Unmarshal(data, &txConf); err != nil {
+					logger.WithError(err).Warn("Failed to unmarshal Tx acknowledgement frame")
```
htdvisser: And here

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,387 @@
...
+					if err := conn.HandleTxAck(&txAck); err != nil {
+						logger.WithField("diid", txConf.Diid).Warn("Failed to handle Tx acknowledgement")
+					}
+				} else {
+					logger.WithField("diid", txConf.Diid).Debug("Tx acknowledgement either does not correspond to a downlink message or arrived too late")
+				}
+				recordRTT(conn, receivedAt, txConf.RefTime)
+
+			case messages.TypeUpstreamProprietaryDataFrame, messages.TypeUpstreamRemoteShell, messages.TypeUpstreamTimeSync:
+				logger.WithField("message_type", typ).Warn("Message type not implemented")
```
htdvisser: And here

pkg/gatewayserver/io/basicstationlns/messages/messages.go
```diff
@@ -0,0 +1,213 @@
...
+	if len(fp.Radios) == 0 {
+		return RouterConfig{}, errFrequencyPlan
+	}
+	// TODO: Handle FP with multiple radios if necessary (https://github.com/TheThingsNetwork/lorawan-stack/issues/761)
+	cfg.FrequencyRange = []int{
+		int(fp.Radios[0].TxConfiguration.MinFrequency),
+		int(fp.Radios[0].TxConfiguration.MaxFrequency),
+	}
+
+	// TODO: Dynamically fill this field based on https://github.com/TheThingsNetwork/lorawan-stack/issues/761
```
htdvisser: Based on what?
Link between ()

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
KrishnaIyer: What do you recommend?

htdvisser: I'm okay with leaving it the way it is for this PR, just saying that it could be smaller

PR #765, opened 2019-05-28
[Quickfix] Add prerequisites with version info to frontend documentation

kschiffer:
#### Summary
@furtiman experienced some issue due to not having the correct node and npm version installed. This PR adds a note informing about which versions are necessary.

#### Changes

- Add prerequisites section to frontend section of the development docs

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -193,6 +193,13 @@ The Console is the official management application of the stack. It can be used 
 The OAuth app provides the necessary frontend for the OAuth provider of the stack. It is used e.g. to display the authorization screen that users get prompted with when they want to authorize a third-party app to access the stack.
 
 ### Building the frontend
+
+#### Prerequisites
+In order to build the frontend, you'll need the following:
+* Node JS, version 10.x (see [Development Environment](#development-environment))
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -193,6 +193,13 @@ The Console is the official management application of the stack. It can be used
 The OAuth app provides the necessary frontend for the OAuth provider of the stack. It is used e.g. to display the authorization screen that users get prompted with when they want to authorize a third-party app to access the stack.
 
 ### Building the frontend
+
+#### Prerequisites
+In order to build the frontend, you'll need the following:
+* Node JS, version 10.x (see [Development Environment](#development-environment))
```
bafonins: @furtiman What version causes issues? As far as I remember you had `10.12.*` ?

furtiman: @bafonins The one I had was 12.3.*

https://nodejs.org/en/ - They say that indeed the LTS is 10.x

kschiffer: Ok, so 10 should be good then. @bafonins, maybe we should downgrade then rather, might introduce compatibility issues otherwise.

PR #769, opened 2019-05-28
[Quickfix] Adjust the Application General Settings form

bafonins:
#### Summary

Screenshot 2019-05-28 at 19 19 19

#### Changes

- Remove `required` from the application name field
- Add `autoFocus` to the application name field

#### Notes for Reviewers

The name field is not required, see [application.proto L36](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/api/application.proto#L36)


bafonins:
> I don't think we should use the autoFocus here

well, we do this with other forms and it actually makes sense. if the user is navigating via the keyboard it is convenient to focus on the first available field

PR #776, opened 2019-05-29
Consistent fieldmask handling

rvolosatovs:
#### Summary

Various registry improvements/fixes
Refs #756 

#### Changes

- Allow exiting early from `Set` without "touching"(extracted from #756) by returning `non-nil, nil, nil`
- Ensure consistent implicit fieldmask behavior:
   - Assume `created_at`, `updated_at` and `ids` to be implicit on GET operations in device registry implementations
   - Assume `ids` to be implicit on CREATE operations in components
- Set `ids.dev_addr` in AS and JS
- Validate fields in `Set` RPCs


pkg/applicationserver/grpc_deviceregistry.go
```diff
@@ -34,13 +35,39 @@ func (r asEndDeviceRegistryServer) Get(ctx context.Context, req *ttnpb.GetEndDev
...
 		return nil, err
 	}
-	return r.registry.Set(ctx, req.EndDevice.EndDeviceIdentifiers, req.FieldMask.Paths, func(dev *ttnpb.EndDevice) (*ttnpb.EndDevice, []string, error) {
-		return &req.EndDevice, req.FieldMask.Paths, nil
+
+	if ttnpb.HasAnyField(req.FieldMask.Paths, "ids.dev_addr") && req.EndDevice.DevAddr != nil && !req.EndDevice.DevAddr.IsZero() {
+		return nil, errInvalidFieldValue.WithAttributes("field", "ids.dev_addr")
+	}
+
+	gets := append(req.FieldMask.Paths[:0:0], req.FieldMask.Paths...)
```
johanstokking: Why

pkg/applicationserver/io/web/grpc_webhooks.go
```diff
@@ -67,9 +75,18 @@ func (s webhookRegistryRPC) Set(ctx context.Context, req *ttnpb.SetApplicationWe
 	if err := rights.RequireApplication(ctx, req.ApplicationIdentifiers, ttnpb.RIGHT_APPLICATION_TRAFFIC_READ); err != nil {
 		return nil, err
 	}
-	return s.webhooks.Set(ctx, req.ApplicationWebhookIdentifiers, req.FieldMask.Paths,
+	gets := appendImplicitWebhookGetPaths(append(req.FieldMask.Paths[:0:0], req.FieldMask.Paths...)...)
```
johanstokking: Idem

pkg/applicationserver/grpc_deviceregistry.go
rvolosatovs: We need to ensure `req.FieldMask.Paths` is not mutated on https://github.com/TheThingsNetwork/lorawan-stack/pull/776/files#diff-9d4e6149ff655290e12bcfb9931c439aR53 and https://github.com/TheThingsNetwork/lorawan-stack/pull/776/files#diff-9d4e6149ff655290e12bcfb9931c439aR56 and the only way to guarantee this is by copying

rvolosatovs: E.g. consider https://play.golang.org/p/NKxIKiAEttg
Assuming registry implementation(s) only `append` and `len(req.FieldMask.Paths) == cap(req.FieldMask.Paths)` this in not strictly necessary, but I think it's a good and quite cheap safety measure to ensure correctness.

johanstokking: I see, and it is a good safety measure, but aren't we protecting it in the wrong place?

I think that the registry `Set()` should not modify the getter. When it does, i.e. by appending implicit fields, it should do that to a copy.

rvolosatovs:
Tested on `ff1705` using this script: https://github.com/rvolosatovs/lorawan-stack/blob/bf8083c1bbbfad521db722a2b5fd638cdbadf90e/shell.nix#L226-L271
(create app, link app, create device, get device, schedule clock example, schedule 3 downlinks)

pkg/applicationserver/grpc_deviceregistry.go
```diff
@@ -34,13 +35,49 @@ func (r asEndDeviceRegistryServer) Get(ctx context.Context, req *ttnpb.GetEndDev
 	return r.registry.Get(ctx, req.EndDeviceIdentifiers, req.FieldMask.Paths)
 }
 
+var errInvalidFieldMask = errors.DefineInvalidArgument("field_mask", "invalid field mask")
+var errInvalidFieldValue = errors.DefineInvalidArgument("field_value", "invalid value of field `{field}`")
```
johanstokking: We typically group this stuff in `var ( ... )` blocks

pkg/applicationserver/io/web/redis/registry.go
```diff
@@ -144,16 +146,19 @@ func (r WebhookRegistry) Set(ctx context.Context, ids ttnpb.ApplicationWebhookId
 				return nil
 			}
 		} else {
-			if pb.ApplicationWebhookIdentifiers != ids {
-				return errInvalidIdentifiers
-			}
-
 			pb.UpdatedAt = time.Now().UTC()
-			sets = append(sets, "updated_at")
+			sets = append(append(make([]string, 0, 2+len(sets)),
```
johanstokking: Why `2+` ?

This level of micro-optimizations doesn't make sense really. To me, this has more downsides (i.e. cluttered code base) than benefits (saving some cpu cycles if Go doesn't optimize this already).

PR #785, opened 2019-06-03
Add default values to gateway message count

### LINKED ISSUE #766: Show 0 when no messages are sent/received
#### Summary

Show `0` when no messages are sent/received

#### Steps to Reproduce

1. Create and connect a gateway that hasn't sent a message yet
2. Go to the gateway page

#### What do you see now?

An empty value for downlink messages:

Screenshot 2019-05-28 at 15 50 18

`0`

No

bafonins:
#### Summary

Closes #766 
Apparently default values are not returned from the api

@johanstokking please check that its fixed

### DIFF
--- a/pkg/webui/console/containers/gateway-statistics/index.js
+++ b/pkg/webui/console/containers/gateway-statistics/index.js
@@ -119,8 +119,8 @@ class GatewayStatistic extends React.PureComponent {
       return null
     }
 
-    const uplinkCount = statistics.uplink_count
-    const downlinkCount = statistics.downlink_count
+    const uplinkCount = statistics.uplink_count || 0
+    const downlinkCount = statistics.downlink_count || 0
 
     return (
       <React.Fragment>


kschiffer:
I'm ok with defaulting to 0 but can we safely assume this value when it is not provided by the API, or should we rather go with `n/a`?

bafonins:
@johanstokking

johanstokking:
Yes, this fallback is required. Generated protos that are returned have an `omitempty` tag (see [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/ttnpb/gateway.pb.go#L1410)) which means that those fields are not there at all when they're empty (0 for numerics).

htdvisser:
@kschiffer @bafonins

Since this problem is very difficult to solve in the backend (with the current specification of protocol buffers and the current versions of protobuf, grpc, grpc-gateway and jsonpb) it might be an idea to have the SDK set these "zero values" before passing them up to the UI. The API description JSON already has type information, so any number field that was requested (in the field mask) can be set to 0 if not present in the response.

PR #807, opened 2019-06-11
Application Server Subscription Uplink Context

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/issues/1441

#### Changes

- Application Server subscriptions now carry the context with them when they are delivered to the frontends.

#### Notes for Reviewers

None.

### DIFF
--- a/pkg/applicationserver/io/io.go
+++ b/pkg/applicationserver/io/io.go
@@ -41,6 +41,12 @@ type Server interface {
 	DownlinkQueueList(context.Context, ttnpb.EndDeviceIdentifiers) ([]*ttnpb.ApplicationDownlink, error)
 }
 
+// ContextApplicationUp represents an ttnpb.ApplicationUp with an attached context.
+type ContextApplicationUp struct {
+	context.Context
+	*ttnpb.ApplicationUp
+}
+
 // Subscription is a subscription to an application or integration managed by a frontend.
 type Subscription struct {
 	ctx       context.Context
@@ -49,7 +55,7 @@ type Subscription struct {
 	protocol string
 	ids      *ttnpb.ApplicationIdentifiers
 
-	upCh chan *ttnpb.ApplicationUp
+	upCh chan ContextApplicationUp
 }
 
 // NewSubscription instantiates a new application or integration subscription.
@@ -60,7 +66,7 @@ func NewSubscription(ctx context.Context, protocol string, ids *ttnpb.Applicatio
 		cancelCtx: cancelCtx,
 		protocol:  protocol,
 		ids:       ids,
-		upCh:      make(chan *ttnpb.ApplicationUp, bufferSize),
+		upCh:      make(chan ContextApplicationUp, bufferSize),
 	}
 }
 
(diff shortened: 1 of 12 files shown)


pkg/applicationserver/io/io.go
```diff
@@ -41,6 +41,12 @@ type Server interface {
 	DownlinkQueueList(context.Context, ttnpb.EndDeviceIdentifiers) ([]*ttnpb.ApplicationDownlink, error)
 }
 
+// ContextApplicationUp represents an ttnpb.ApplicationUp with an attached context.
+type ContextApplicationUp struct {
```
johanstokking: `ContextualApplicationUp` is a bit nicer and consistent with the codebase

pkg/applicationserver/io/io.go
```diff
@@ -41,6 +41,12 @@ type Server interface {
 	DownlinkQueueList(context.Context, ttnpb.EndDeviceIdentifiers) ([]*ttnpb.ApplicationDownlink, error)
 }
 
+// ContextApplicationUp represents an ttnpb.ApplicationUp with an attached context.
```
johanstokking: With its context

pkg/applicationserver/io/io.go
```diff
@@ -49,7 +55,7 @@ type Subscription struct {
 	protocol string
 	ids      *ttnpb.ApplicationIdentifiers
 
-	upCh chan *ttnpb.ApplicationUp
+	upCh chan ContextApplicationUp
```
johanstokking: Let's keep this a pointer to make the channel payload as small as possible

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -129,31 +129,31 @@ func (c *connection) setup(ctx context.Context) error {
...
 					continue
 				}
-				buf, err := c.format.FromUp(up)
+				buf, err := c.format.FromUp(up.ApplicationUp)
 				if err != nil {
 					logger.WithError(err).Warn("Failed to marshal upstream message")
 					continue
 				}
-				logger.Info("Publish upstream message")
+				logger.WithField("device_uid", unique.ID(up.Context, up.EndDeviceIdentifiers)).Debug("Publish upstream message")
```
johanstokking: Can you add the field to the logger once you have the contextual up? This is going to be interesting in other log entries as well

pkg/applicationserver/io/io.go
```diff
@@ -82,19 +88,22 @@ var errBufferFull = errors.DefineResourceExhausted("buffer_full", "buffer is ful
...
 // SendUp sends an upstream message.
 // This method returns immediately, returning nil if the message is buffered, or with an error when the buffer is full.
-func (s *Subscription) SendUp(up *ttnpb.ApplicationUp) error {
+func (s *Subscription) SendUp(ctx context.Context, up *ttnpb.ApplicationUp) error {
 	select {
 	case <-s.ctx.Done():
 		return s.ctx.Err()
-	case s.upCh <- up:
+	case s.upCh <- &ContextualApplicationUp{
+		Context:       ctx,
```
htdvisser: It is important to note that this context is likely to be canceled shortly after `SendUp` returns, so if this (or a derived) context is used any future requests (gRPC, HTTP), those may return a `context.Canceled` error without even making the request.

cc: @johanstokking

pkg/applicationserver/io/web/webhooks.go
```diff
@@ -268,7 +268,7 @@ func (w *webhooks) NewSubscription() *io.Subscription {
 			case <-w.ctx.Done():
 				return
 			case msg := <-sub.Up():
-				if err := w.handleUp(w.ctx, msg); err != nil {
+				if err := w.handleUp(msg.Context, msg.ApplicationUp); err != nil {
```
htdvisser: This for example could be problematic, since `msg.Context` may already be canceled at this point.

PR #813, opened 2019-06-13
Fix LNS credentials blob

### LINKED ISSUE #812: Basic Station CUPS sends invalid LNS credentials blob
#### Summary

Basic Station CUPS sends invalid LNS credentials blob to the Station.

#### Steps to Reproduce

1. Configure `cups-boot.trust` and `cups-boot.key` of Basic Station PF to a V3 instance.
2. Connect to GCS via CUPS Server auth (with token) 
3. GCS sends LNS Credentials blob without Client cert (only {trust,key})
4. Check the output of the parsing the LNSCredentials.

#### What do you see now?

Issue 1:
```
Failed to parse tc credentials: ASN.1 SEQ expected for trust 0x4e
```

Issue 2:
```
 [AIO:ERRO] tc contains malformed auth token - expecting: {header: value\r\n}*
```

Successful parsing

#### Environment

`ttn-lw-stack v3.0.3`

1. Four bytes of zeros have to be used in the place of empty cert in the blob {trust,cert,key} as per the [spec](https://doc.sm.tc/station/cupsproto.html#http-post-response).
2. Auth token needs to be prefixed with "Authorization: " (or something similar) and should be suffixed with LFCR.

Yups

KrishnaIyer:
#### Summary

Closes #812 

#### Changes

- Add an empty creds block.
- Add proper auth suffix and prefix.

#### Release Notes

- Fix Basic Station CUPS LNS credentials blob.


KrishnaIyer:
@htdvisser: Can you also check?

PR #817, opened 2019-06-14
Improve api-key/collaborator rights handling

### LINKED ISSUE #189: Checkbox group component
**Summary:** 

Implement the checkbox group component

**Why do we need this?**

We need some convenient way to manage lists of checkboxes.

- The `Field` component
- The `Checkbox` component

A solution that can be integrated into the console. Something like this:
image
Also see [User-Friendly Multi-select](http://doejo.com/blog/solutions-for-implementing-user-friendly-multi-select/)

The component must have:
1. Controllabel height and width
2. The 'select all` checkbox
3. Integration to our `Form` component
4. Display group related errors/validation, e.g. that at least on checkbox should be checked from the list

The component could have:
1. The search functionality
2. Additional list of current selections 

tba

**Environment:**

Browser

Implement it

bafonins:
#### Summary

References #189 
This PR adds the `<RightsGroup/>` component.

Screenshot 2019-06-14 at 15 07 17
Screenshot 2019-06-14 at 15 07 24
Screenshot 2019-06-14 at 15 07 44
Screenshot 2019-06-14 at 15 08 04

#### Changes

- Add the `<RightsGroup />` component
- Add the indeterminate styling for `<Checkbox />`
- Refactor application api key add/edit, application collaborator add/edit pages
- Refactor gateway api key add/edit,
- Extend rights selectors

#### Notes for Reviewers

This PR does not fully implement #189, but improves the way the users can manage rights. We can implement the styling later.
Note, I didnt refactor the app selectors, because as far as I know you have them ready and will make a PR later. If its not the case I can also adjust the app selectors in this PR as well.


pkg/webui/console/components/rights-group/rights-group.styl
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-.id
-  h2()
+.select-all
+  margin-bottom: $ls.xs
```
kschiffer: The hierarchical relationship between the `All Rights` checkbox and their subordainates needs to be made clearer visually. I think applying a left pading to the checkbox group will help here:
```
    border-normal('left')
    padding-left: $ls.m
```
image

pkg/webui/console/components/rights-group/index.js
```diff
@@ -0,0 +1,164 @@
...
+  return {
+    allSelected: selectedCheckboxesCount === totalCheckboxesCount,
+    indeterminate: selectedCheckboxesCount !== 0 && selectedCheckboxesCount !== totalCheckboxesCount,
+  }
+}
+
+@bind
+class RightsGroup extends React.Component {
+
+  state = {}
```
kschiffer:
```suggestion

```
?

kschiffer:
Meta: Note that your opening comment wrongfully marks this PR to close #189:
image

pkg/webui/console/components/rights-group/index.js
bafonins:
```
react-dom.development.js:506 Warning: `RightsGroup` uses `getDerivedStateFromProps` but its initial state is undefined. This is not recommended. Instead, define the initial state by assigning an object to `this.state` in the constructor of `RightsGroup`. This ensures that `getDerivedStateFromProps` arguments have a consistent shape.
```

Issue #774, opened 2019-05-29
Improve error handling in forms / avoid out-of-view error messages


kschiffer:
#### Summary
Submit error messages in our forms appear at the top of the form, which is potentially out-of-view in longer forms. This leads to users not getting proper error feedback from form submits. We have this problem e.g. in the device data forms (Add Device, Device General Settings)

#### Why do we need this?
Is bad UX

#### What is already there? What do you see now?
Frontend error handling; error messages appearing at the top of the form

#### What is missing? What do you want to see?
Making sure the error message can be seen by the user

#### Environment
Console in browser.

#### How do you propose to implement this?
We can apply two measures here:
1. Make sure that all mappable error message will be mapped to a form field (and not be set as generic error message at the top of the form)
2. Make sure that the viewport will scroll to a error message, or validation error, in case it is out of view.

#### Can you do this yourself and submit a Pull Request?
Yes.

bafonins:
Resolved in https://github.com/TheThingsNetwork/lorawan-stack/pull/931

PR #831, opened 2019-06-19
Fix out-of-view form errors

bafonins:
#### Summary

References #774 

#### Changes

- Add functionality to `<Field />` to focus on the underlying inputs that support focus
- Add functionality to scroll to the form error and focus on the first field
- Add functionality to focus on the first invalid field after submission

### DIFF
--- a/pkg/webui/components/form/index.js
+++ b/pkg/webui/components/form/index.js
@@ -13,16 +13,92 @@
 // limitations under the License.
 
 import React from 'react'
-import { Formik } from 'formik'
+import { Formik, getIn } from 'formik'
 import bind from 'autobind-decorator'
 
+import scrollIntoView from 'scroll-into-view-if-needed'
 import Notification from '../notification'
 import PropTypes from '../../lib/prop-types'
 import FormContext from './context'
 import FormField from './field'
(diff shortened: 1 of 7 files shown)


pkg/webui/components/form/index.js
```diff
@@ -13,16 +13,92 @@
...
+      if (err) {
+        field = this.fields[fieldName]
+        break
+      }
+    }
+
+    return field
+  }
+
+  _scroll (node, delay = 200) {
```
kschiffer:
```suggestion
  _scroll (node, duration = 1000) {
```

pkg/webui/components/form/index.js
```diff
@@ -13,16 +13,92 @@
...
+    if (node) {
+      scrollIntoView(node, {
+        scrollMode: 'if-needed',
+        behavior: 'smooth',
+        duration: delay,
```
kschiffer:
```suggestion
        duration,
```
Though scroll duration doesn't really seem to work in chrome (seems to be browser dependent)
https://github.com/stipsan/scroll-into-view-if-needed#ponyfill-smooth-scrolling

PR #844, opened 2019-06-21
Make JS/GS/NS/AS optional

### LINKED ISSUE #834: Flaky NS (Redis) test
#### Summary

There's one test that's consistently failing randomly 😛  

#### Steps to Reproduce

1. Push to a PR
2. Check travis output.
3. Either the `ARCH=amd64` or the `ARCH=386` fails randomly

#### What do you see now?

Travis

```
namespace=networkserver task_id=process_downlink test_name=TestFlow/Redis_registry/Redis_downlink_task_queue/Class_A/OTAA/EU868/1.0.2
--- FAIL: TestFlow (0.00s)
    --- FAIL: TestFlow/Redis_registry/Redis_downlink_task_queue (32.78s)
        --- FAIL: TestFlow/Redis_registry/Redis_downlink_task_queue/Class_A/OTAA/EU868/1.0.2 (32.78s)
            networkserver_flow_test.go:208: Cluster.GetPeer called
            networkserver_flow_test.go:208: ApplicationAccess.ListRights called
            networkserver_flow_test.go:214: Cluster.GetPeer called
            networkserver_flow_test.go:214: ApplicationAccess.ListRights called
            networkserver_flow_test.go:240: Device created
            networkserver_flow_test.go:322: Cluster.GetPeer called
            networkserver_util_test.go:458: Cluster.Auth called
            networkserver_flow_test.go:460: Cluster.GetPeer called
            networkserver_util_test.go:488: Cluster.Auth called
            networkserver_flow_test.go:508: Device successfully joined. DevAddr: C03E4049
            --- FAIL: TestFlow/Redis_registry/Redis_downlink_task_queue/Class_A/OTAA/EU868/1.0.2/uplink (32.75s)
                networkserver_flow_test.go:573: Timed out while waiting for deduplication window to close
            networkserver_flow_test.go:562: HandleUplink returned rpc error: code = Unavailable desc = transport is closing
FAIL
FAIL	go.thethings.network/lorawan-stack/pkg/networkserver	39.564s
```

No failure as for the other build config.

#### Environment

Travis, `ttn-lw-stack: v3.0.3`

Up to the NS expert @rvolosatovs 

Nopes

### LINKED ISSUE #842: CLI doesn't update JS device fields
#### Summary

CLI doesn't set JS fields on device update if `supports-join` isn't provided

#### Steps to Reproduce

1. Create a device with `resets_join_nonces` unset
2. Update the device with `resets_join_nonces` set

#### What do you see now?

Nothing happens

Make the set effective

We currently implicitly set `supports_join` when `root_keys` is part of the sets. We should probably do this for all JS specific fields

cc @rvolosatovs 

Yes

### LINKED ISSUE #843: CLI requires NS and AS for creating devices in JS
#### Summary

CLI requires NS and AS for creating devices in JS

#### Steps to Reproduce

1. Create a device with only JS fields, i.e. `$ ttn-lw-cli dev create test-app test-dev --join-eui 0102030405060708 --dev-eui 01020304050607ff --root-keys.app-key.key 01020304050607080102030405060708 --resets-join-nonces=true --net-id=000001 --network-server-address 000001`

#### What do you see now?

- Since `abp` is unset, `supports_join` is set, which is an NS field, triggering a NS hit for create/update
- On device create, [NS gets hit anyway](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/cmd/ttn-lw-cli/commands/end_devices_split.go#L221)
- But NS requires a device profile to be set, i.e. LoRaWAN MAC and PHY versions

I want to be able to create devices in JS only

Force CLI not to call JS/NS/AS. This is similar to the Console, where we disable functionality when the component isn't there.

Will review

johanstokking:
#### Summary

Closes #834 (not related but super annoying and small "fix")
Closes #842 
Closes #843 

#### Changes

- Make JS, GS, NS and AS optional through config. This is like Console
- Set device in JS when any JS field is set (not just `root_keys`)

#### Release Notes

- Added CLI support for enabling/disabling JS, GS, NS and AS through configuration

### DIFF
--- a/cmd/ttn-lw-cli/commands/end_devices.go
+++ b/cmd/ttn-lw-cli/commands/end_devices.go
@@ -141,6 +141,8 @@ func generateDevAddr(netID types.NetID) (types.DevAddr, error) {
 	return devAddr, nil
 }
 
+var errGatewayServerDisabled = errors.DefineFailedPrecondition("gateway_server_disabled", "Gateway Server is disabled")
+
 var (
 	endDevicesCommand = &cobra.Command{
 		Use:     "end-devices",
@@ -152,6 +154,10 @@ var (
 		Short:             "List available frequency plans for end devices",
 		PersistentPreRunE: preRun(),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if !config.GatewayServerEnabled {
+				return errGatewayServerDisabled
+			}
+
 			baseFrequency, _ := cmd.Flags().GetUint32("base-frequency")
 			ns, err := api.Dial(ctx, config.NetworkServerGRPCAddress)
 			if err != nil {
@@ -309,17 +315,21 @@ var (
 
 			setDefaults, _ := cmd.Flags().GetBool("defaults")
 			if setDefaults {
-				device.NetworkServerAddress = getHost(config.NetworkServerGRPCAddress)
-				device.ApplicationServerAddress = getHost(config.ApplicationServerGRPCAddress)
-				paths = append(paths,
-					"application_server_address",
-					"network_server_address",
-				)
+				if config.NetworkServerEnabled {
+					device.NetworkServerAddress = getHost(config.NetworkServerGRPCAddress)
+					paths = append(paths, "network_server_address")
+				}
+				if config.ApplicationServerEnabled {
+					device.ApplicationServerAddress = getHost(config.ApplicationServerGRPCAddress)
+					paths = append(paths, "application_server_address")
+				}
 			}
 
 			if abp, _ := cmd.Flags().GetBool("abp"); abp {
 				device.SupportsJoin = false
-				paths = append(paths, "supports_join")
+				if config.NetworkServerEnabled {
+					paths = append(paths, "supports_join")
+				}
 				if withSession, _ := cmd.Flags().GetBool("with-session"); withSession {
 					if device.ProvisionerID != "" {
 						return errEndDeviceKeysWithProvisioner
@@ -355,12 +365,16 @@ var (
 				}
 			} else {
 				device.SupportsJoin = true
-				paths = append(paths, "supports_join")
+				if config.NetworkServerEnabled {
+					paths = append(paths, "supports_join")
+				}
 				if setDefaults {
-					device.JoinServerAddress = getHost(config.JoinServerGRPCAddress)
-					paths = append(paths,
-						"join_server_address",
-					)
+					if config.JoinServerEnabled {
+						device.JoinServerAddress = getHost(config.JoinServerGRPCAddress)
+						paths = append(paths,
+							"join_server_address",
+						)
+					}
 				}
 				if withKeys, _ := cmd.Flags().GetBool("with-root-keys"); withKeys {
 					if device.ProvisionerID != "" {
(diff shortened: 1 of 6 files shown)


cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -152,8 +154,12 @@ var (
 		Short:             "List available frequency plans for end devices",
 		PersistentPreRunE: preRun(),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if !config.GatewayServerEnabled {
+				return errGatewayServerDisabled
+			}
+
 			baseFrequency, _ := cmd.Flags().GetUint32("base-frequency")
-			gs, err := api.Dial(ctx, config.NetworkServerGRPCAddress)
+			gs, err := api.Dial(ctx, config.GatewayServerGRPCAddress)
```
rvolosatovs: :eye:

htdvisser: This indeed doesn't seem quite right. Device frequency plans come from the network server, not the gateway server.

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -466,17 +480,13 @@ var (
...
-				paths = append(paths, "supports_join")
-			}
 			if err = util.SetFields(&device, setEndDeviceFlags); err != nil {
 				return err
 			}
 			device.Attributes = mergeAttributes(device.Attributes, cmd.Flags())
 			device.EndDeviceIdentifiers = *devID
 
-			isPaths, nsPaths, asPaths, jsPaths := splitEndDeviceSetPaths(device.SupportsJoin, paths...)
+			isPaths, nsPaths, asPaths, jsPaths := splitEndDeviceSetPaths(ttnpb.HasAnyField(paths, setEndDeviceToJS...), paths...)
```
htdvisser: I think this actually needs to be SupportsJoin

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -466,17 +480,13 @@ var (
...
 
 			if len(nsPaths) > 0 {
```
htdvisser: Here, and below we should check for `config.NetworkServerEnabled` etc.

cmd/ttn-lw-cli/commands/end_devices.go
johanstokking: Device frequency plans indeed, gateway frequency plans come from GS. Will fix that

cmd/ttn-lw-cli/commands/end_devices.go
johanstokking: We're here in update. If you pass `--resets_join_nonces=false`, it should go to JS.

Otherwise we require the user to pass `--supports-join` and that's not very user friendly.

cmd/ttn-lw-cli/commands/end_devices.go
johanstokking: Where exactly?

Which checks below?

Because `setEndDevice()` does this already and prints warnings if fields are being updated while the concerning components aren't enabled.

cmd/ttn-lw-cli/commands/end_devices.go
htdvisser: Then the first argument of splitEndDeviceSetPaths needs to be renamed

cmd/ttn-lw-cli/commands/end_devices.go
htdvisser: With `if len(nsPaths) > 0 {`, `if len(asPaths) > 0 {` and `if len(jsPaths) > 0 {`.
We don't need to ask the IS for the NS/AS/JS address if we're not going to use it.

cmd/ttn-lw-cli/commands/end_devices.go
johanstokking: https://github.com/TheThingsNetwork/lorawan-stack/pull/844/commits/2c8cf0658cb36193194362402098f889a8c5e26c

cmd/ttn-lw-cli/commands/end_devices.go
johanstokking: https://github.com/TheThingsNetwork/lorawan-stack/pull/844/commits/da60c91bea5e9e99353039715ac33e6a51f5dd5c

PR #846, opened 2019-06-24
Wrap keys from JS

### LINKED ISSUE #5: Encrypt session keys
**Summary:** 

Pass and store keys encrypted in the key envelopes.

**Why do we need this?**

For secure transfer and encryption at rest.

We have key envelopes everywhere that can contain wrapped keys with a key encryption key (KEK) label. We also have key vaults with wrapping and unwrapping functionality in the base component. Finally, we have configuration for key vaults.

We need to wrap the keys. The scope of this issue is the transfer of session keys from JS to NS and AS.

For this issue, we assume that KEKs are exchanged out-of-band. We may come up with bilateral KEK handshake using Diffie-Hellman or something later.

In any case, the server that wraps the key needs to know which KEK the client has. The server chooses the KEK based on the connection context.

Proposed order of preference;

1. If there is cluster authentication, use a cluster KEK
2. If there is network authentication, use a network KEK
3. If there is TLS client authentication (i.e. external AS and NS), use a mapping of client certificate to pre-shared KEK
4. If there is only TLS server authentication, do not wrap keys and send warnings
5. If there is neither cluster, network, TLS client nor server authentication, do not return keys at all

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/271 by @rvolosatovs

### LINKED ISSUE #58: Unwrap keys before returning it to clients
**Summary:** 

Unwrap keys before returning it to clients.

**Why do we need this?**

Because clients don't have the KEKs at-hand; components should decrypt it for them.

No unwrapping; the keys are returned as they are stored.

This is not a problem since they're not encrypted in the first place as long as #5 is still open.

Return key in the clear to the client.

When getting the keys from the registries, unwrap them and bogus-wrap them with an empty KEK label.

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/1483 by @johanstokking

johanstokking:
#### Summary

Closes #5 
Closes #58 

#### Changes

- Added KEK labeler to key vault. The in-memory implementation is quite simple but that's based on static config anyway
- Added wrapping keys in JS
- Some minor fixes

#### Notes for Reviewers

JS now uses the `network_server_address` and `application_server_address` for authentication (client certificate CN) as well as picking the KEK. If we want a different approach, please make a proposal. I think this is a good way forward.

- We cannot only use NetID for authentication and KEKs as there may be many clusters within the same NetID
- "AS ID" is basically `application_server_address`. It supports addresses and URL-like host names

### DIFF
--- a/pkg/applicationserver/grpc_deviceregistry.go
+++ b/pkg/applicationserver/grpc_deviceregistry.go
@@ -19,12 +19,13 @@ import (
 
 	pbtypes "github.com/gogo/protobuf/types"
 	"go.thethings.network/lorawan-stack/pkg/auth/rights"
+	"go.thethings.network/lorawan-stack/pkg/crypto/cryptoutil"
 	"go.thethings.network/lorawan-stack/pkg/errors"
 	"go.thethings.network/lorawan-stack/pkg/ttnpb"
 )
 
 type asEndDeviceRegistryServer struct {
-	registry DeviceRegistry
+	AS *ApplicationServer
 }
 
 // Get implements ttnpb.AsEndDeviceRegistryServer.
@@ -32,7 +33,25 @@ func (r asEndDeviceRegistryServer) Get(ctx context.Context, req *ttnpb.GetEndDev
 	if err := rights.RequireApplication(ctx, req.ApplicationIdentifiers, ttnpb.RIGHT_APPLICATION_DEVICES_READ); err != nil {
 		return nil, err
 	}
-	return r.registry.Get(ctx, req.EndDeviceIdentifiers, req.FieldMask.Paths)
+	dev, err := r.AS.deviceRegistry.Get(ctx, req.EndDeviceIdentifiers, req.FieldMask.Paths)
+	if err != nil {
+		return nil, err
+	}
+	if ttnpb.HasAnyField(req.FieldMask.Paths, "session.keys.app_s_key") && dev.GetSession().GetAppSKey() != nil {
+		key, err := cryptoutil.UnwrapAES128Key(*dev.Session.AppSKey, r.AS.KeyVault)
+		if err != nil {
+			return nil, err
+		}
+		dev.Session.AppSKey = &ttnpb.KeyEnvelope{Key: &key}
+	}
+	if ttnpb.HasAnyField(req.FieldMask.Paths, "pending_session.keys.app_s_key") && dev.GetPendingSession().GetAppSKey() != nil {
+		key, err := cryptoutil.UnwrapAES128Key(*dev.PendingSession.AppSKey, r.AS.KeyVault)
+		if err != nil {
+			return nil, err
+		}
+		dev.PendingSession.AppSKey = &ttnpb.KeyEnvelope{Key: &key}
+	}
+	return dev, nil
 }
 
 var (
@@ -49,7 +68,7 @@ func (r asEndDeviceRegistryServer) Set(ctx context.Context, req *ttnpb.SetEndDev
 	if err := rights.RequireApplication(ctx, req.EndDevice.ApplicationIdentifiers, ttnpb.RIGHT_APPLICATION_DEVICES_WRITE); err != nil {
 		return nil, err
 	}
-	return r.registry.Set(ctx, req.EndDevice.EndDeviceIdentifiers, req.FieldMask.Paths, func(dev *ttnpb.EndDevice) (*ttnpb.EndDevice, []string, error) {
+	return r.AS.deviceRegistry.Set(ctx, req.EndDevice.EndDeviceIdentifiers, req.FieldMask.Paths, func(dev *ttnpb.EndDevice) (*ttnpb.EndDevice, []string, error) {
 		sets := req.FieldMask.Paths
 		if dev != nil {
 			if err := ttnpb.ProhibitFields(req.FieldMask.Paths, "ids.dev_addr"); err != nil {
@@ -88,7 +107,7 @@ func (r asEndDeviceRegistryServer) Delete(ctx context.Context, ids *ttnpb.EndDev
 	if err := rights.RequireApplication(ctx, ids.ApplicationIdentifiers, ttnpb.RIGHT_APPLICATION_DEVICES_WRITE); err != nil {
 		return nil, err
 	}
-	_, err := r.registry.Set(ctx, *ids, nil, func(*ttnpb.EndDevice) (*ttnpb.EndDevice, []string, error) {
+	_, err := r.AS.deviceRegistry.Set(ctx, *ids, nil, func(*ttnpb.EndDevice) (*ttnpb.EndDevice, []string, error) {
 		return nil, nil, nil
 	})
 	if err != nil {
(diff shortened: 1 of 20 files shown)


pkg/crypto/cryptoutil/kek_labelers.go
```diff
@@ -0,0 +1,72 @@
...
+		host = addr
+	}
+	if h, _, err := net.SplitHostPort(host); err == nil {
+		host = h
+	}
+	return
+}
+
+// NsKEKLabel returns a KEK label in the form `ns:netID:host` from the given NetID and address. Empty parts are omitted.
+func (c ComponentPrefixKEKLabeler) NsKEKLabel(ctx context.Context, netID *types.NetID, addr string) string {
```
htdvisser: Shouldn't KEK labels include a sequence number to allow for rotation? Same for AS below.

johanstokking: Any `ComponentKEKLabeler` that the key vault provides can do that. The static/in-memory key vault simply uses a prefix, but I can imagine a more dynamic key vault using all kinds of rotation schemes. There’s no point in a sequence number when the key vault is static

pkg/applicationserver/grpc_deviceregistry.go
```diff
@@ -19,20 +19,49 @@ import (
...
 func (r asEndDeviceRegistryServer) Get(ctx context.Context, req *ttnpb.GetEndDeviceRequest) (*ttnpb.EndDevice, error) {
 	if err := rights.RequireApplication(ctx, req.ApplicationIdentifiers, ttnpb.RIGHT_APPLICATION_DEVICES_READ); err != nil {
 		return nil, err
 	}
-	return r.registry.Get(ctx, req.EndDeviceIdentifiers, req.FieldMask.Paths)
+	dev, err := r.AS.deviceRegistry.Get(ctx, req.EndDeviceIdentifiers, req.FieldMask.Paths)
+	if err != nil {
+		return nil, err
+	}
+	for _, s := range []struct {
```
rvolosatovs: IMO this is overly complex for what it does
```go
if ttnpb.HasAnyField(req.FieldMask.Paths, "session.keys.app_s_key") && dev.GetSession().GetAppSKey() != nil {
	key, err := cryptoutil.UnwrapAES128Key(*dev.Session.AppSKey, r.AS.KeyVault)
	if err != nil {
		return nil, err
	}
	dev.Session.AppSKey = &ttnpb.KeyEnvelope{Key: &key}
}

if ttnpb.HasAnyField(req.FieldMask.Paths, "pending_session.keys.app_s_key") && dev.GetPendingSession().GetAppSKey() != nil {
	key, err := cryptoutil.UnwrapAES128Key(*dev.PendingSession.AppSKey, r.AS.KeyVault)
	if err != nil {
		return nil, err
	}
	dev.PendingSession.AppSKey = &ttnpb.KeyEnvelope{Key: &key}
}
```

Is simpler, shorter and more efficient. Please consider simplifying this.

pkg/crypto/cryptoutil/kek_labelers.go
```diff
@@ -0,0 +1,72 @@
...
+	} else {
+		host = addr
+	}
+	if h, _, err := net.SplitHostPort(host); err == nil {
+		host = h
+	}
+	return
+}
+
+// NsKEKLabel returns a KEK label in the form `ns:netID:host` from the given NetID and address. Empty parts are omitted.
```
rvolosatovs: `ns:netID:host` is only true if separator is `":"` or `""` in `c`, right?

pkg/crypto/cryptoutil/kek_labelers.go
```diff
@@ -0,0 +1,72 @@
...
+	"go.thethings.network/lorawan-stack/pkg/types"
+)
+
+// ComponentPrefixKEKLabeler is a ComponentKEKLabeler that joins the component prefix, separator and host.
+type ComponentPrefixKEKLabeler struct {
+	Separator string
+}
+
+func hostFromAddr(addr string) (host string) {
+	if url, err := url.Parse(addr); err == nil && url.Host != "" {
```
rvolosatovs: Isn't this magnitudes simpler and clearer?
(assuming the named return variable is removed)
```go
host := addr
url, err := url.Parse(addr)
if err == nil && url.Host != "" {
	host = url.Host
}

h, _, err := net.SplitHostPort(host)
if err == nil {
	return h
}
return host
```

pkg/joinserver/joinserver.go
```diff
@@ -155,6 +156,20 @@ func validateCaller(dn pkix.Name, addr string) error {
 	return nil
 }
 
+// wrapKeyIfKEKExists wraps the given key with the KEK label.
+// If the configured key vault cannot find the KEK, the key is returned in the clear.
+func (js *JoinServer) wrapKeyIfKEKExists(key types.AES128Key, kekLabel string) (*ttnpb.KeyEnvelope, error) {
+	env, err := cryptoutil.WrapAES128Key(key, kekLabel, js.KeyVault)
+	if errors.IsNotFound(err) {
```
rvolosatovs: This may do a whole bunch of type assertions, maybe only do it if `err != nil` ?

pkg/joinserver/joinserver_internal_test.go
```diff
@@ -36,6 +37,14 @@ func KeyToBytes(key types.AES128Key) []byte { return key[:] }
 
 func KeyPtr(key types.AES128Key) *types.AES128Key { return &key }
 
+func MustWrapAES128Key(key types.AES128Key, kek []byte) []byte {
+	encryptedKey, err := crypto.WrapKey(key[:], kek)
+	if err != nil {
+		panic(err)
```
rvolosatovs: you can just use `test.Must` for this

pkg/networkserver/grpc_deviceregistry.go
```diff
@@ -34,7 +35,49 @@ func (ns *NetworkServer) Get(ctx context.Context, req *ttnpb.GetEndDeviceRequest
 			return nil, err
 		}
 	}
-	return ns.devices.GetByID(ctx, req.ApplicationIdentifiers, req.DeviceID, req.FieldMask.Paths)
+	dev, err := ns.devices.GetByID(ctx, req.ApplicationIdentifiers, req.DeviceID, req.FieldMask.Paths)
+	if err != nil {
+		return nil, err
+	}
+	for _, s := range []struct {
```
rvolosatovs: IDEM.

pkg/applicationserver/grpc_deviceregistry.go
johanstokking: Yeah this comes from copy-pasting from NS and we wanted to keep these gRPC device registry implementations in sync to a large extent, but I agree it's too extreme in this case.

johanstokking: This is obviously my responsibility but your suggestion was overly simple, https://github.com/TheThingsNetwork/lorawan-stack/issues/853

Issue #723, opened 2019-05-21
Introduce update actions for console entities


bafonins:
#### Summary

Entities are not updated after the successful edit form submission which leads to stale data being shown in the ui.
Affected entities:
- [x] Applications
- [x] Devices
- [x] Gateways

#### Steps to Reproduce

1. Go the Application General Settings page
2. Update the application
3. Go to the Application Overview page and back to the settings page
4. See no changes being reflected in the ui

#### What do you see now?

No changes after successful updates.

#### What do you want to see instead?

The latest data after edits.

#### Environment

browser (console)

#### How do you propose to implement this?

Introduce update actions for console entities

#### Can you do this yourself and submit a Pull Request?

yes

kschiffer:
Would we connect this action also to the middleware to perform the update with the API?

bafonins:
> Would we connect this action also to the middleware to perform the update with the API?

Would be nice, however this can be challenging when dealing with forms.

kschiffer:
Yeah. I'm also wondering whether it makes sense to differentiate between patching and updating, or whether we include patching also for the update action. See also https://github.com/TheThingsNetwork/lorawan-stack/pull/708#discussion_r286434743

kschiffer:
So we currently already tie the update action for devices to the respective API call. This works well in most cases. However, one problem I identified is that update actions that delete fields will not result in a proper update of the store. The reason is that the responses of such update actions have no indication for the reducer as to what field should be removed because it is simply missing. If we spread the result over the existing store data, the deleted field will still be there.

To solve this, I think we need a reference to the original request action, which contains the patch of the update request. Using this and the request result, we should be able to determine which fields need to be removed.

Question is whether there is a simpler solution, or whether we should deal with these cases manually, be doing the request and calling the success actions ourselves. What do you think?

PR #856, opened 2019-06-26
Introduce update action for applications

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/723

#### Changes

- Add tests for the applications reducer
- Simplify the Application General Settings page
- Fix issue when updating application in the store
- Add application update actions
- Add application delete actions
- Add functionality to the application root view to redirect to the applications list when `selectedApplication` is unset. In the future we can introduce a separate component that could wrap the `<Route />` and redirect automatically.
- Add delete functionality to the pagination reducer
- Add delete action creators to the pagination action creators

#### Notes for Reviewers

Please check the submit method in the general settings form. are you fine with such implementation?

I also changed the type of list request actions, from `GET_APPLICATIONS_LIST` to `GET_APPLICATION_LIST`. First, it doesnt really align with the api of the pagination higher order reducer. Second, the whole `SHARED_NAME`, `SHARED_NAME_SINGLE` introduces quite some mess and i think we should get rid of this. `GET_APPLICATION` and `GET_APPLICATION_LIST` gives enough context to distinguish which action fetches a single entity and which fetches the list.

### DIFF
--- a/pkg/webui/console/store/middleware/logics/applications.js
+++ b/pkg/webui/console/store/middleware/logics/applications.js
@@ -31,6 +31,17 @@ const getApplicationLogic = createRequestLogic({
   },
 })
 
+const updateApplicationLogic = createRequestLogic({
+  type: applications.UPDATE_APP,
+  async process ({ action }) {
+    const { id, patch } = action.payload
+
+    const result = await api.application.update(id, patch)
+
+    return { ...patch, ...result }
+  },
+})
+
 const getApplicationsLogic = createRequestLogic({
   type: applications.GET_APPS_LIST,
   latest: true,
@@ -149,6 +160,7 @@ const getWebhookFormatsLogic = createRequestLogic({
 
 export default [
   getApplicationLogic,
+  updateApplicationLogic,
   getApplicationsLogic,
   getApplicationsRightsLogic,
   getApplicationApiKeysLogic,
--- a/pkg/webui/console/store/middleware/logics/applications.js
+++ b/pkg/webui/console/store/middleware/logics/applications.js
@@ -42,6 +42,17 @@ const updateApplicationLogic = createRequestLogic({
   },
 })
 
+const deleteApplicationLogic = createRequestLogic({
+  type: applications.DELETE_APP,
+  async process ({ action }) {
+    const { id } = action.payload
+
+    await api.application.delete(id)
+
+    return { id }
+  },
+})
+
 const getApplicationsLogic = createRequestLogic({
   type: applications.GET_APPS_LIST,
   latest: true,
@@ -161,6 +172,7 @@ const getWebhookFormatsLogic = createRequestLogic({
 export default [
   getApplicationLogic,
   updateApplicationLogic,
+  deleteApplicationLogic,
   getApplicationsLogic,
   getApplicationsRightsLogic,
   getApplicationApiKeysLogic,
--- a/pkg/webui/console/store/reducers/pagination.js
+++ b/pkg/webui/console/store/reducers/pagination.js
@@ -15,6 +15,8 @@
 import {
   createPaginationByIdRequestActions,
   createPaginationRequestActions,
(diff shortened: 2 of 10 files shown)


pkg/webui/console/views/application-general-settings/index.js
```diff
@@ -77,14 +76,15 @@ export default class ApplicationGeneralSettings extends React.Component {
...
 
     await this.setState({ error: '' })
 
     const changed = diff(application, values)
+    const { ids: { application_id }} = application
+
     try {
-      const { ids: { application_id }} = application
-      await api.application.update(application_id, changed)
+      await updateApplication(application_id, changed)
```
kschiffer: Using the update action here [will cause issues when you remove a value](https://github.com/TheThingsNetwork/lorawan-stack/issues/723#issuecomment-505826575) that was previously filled (e.g. application name). The result of the API request is not able to act as a proper patch for the reducer, because the removed field is not present.
Meaning: the field will be properly updated, but the store value remains the same. 

I solved this [in my PR](https://github.com/TheThingsNetwork/lorawan-stack/pull/855) by using the `updateSuccess` action instead, after issuing the API request myself.

bafonins: this doesnt seem right. please see my implementation

pkg/webui/console/views/application/index.js
```diff
@@ -151,6 +152,14 @@ export default class Application extends React.Component {
     startStream(appId)
   }
 
+  componentDidUpdate (prevProps) {
+    const { application, redirectToList } = this.props
+
+    if (Boolean(prevProps.application) && !Boolean(application)) {
+      redirectToList()
```
kschiffer: The redirect results in an error for me, originating from the tabular component.
image

pkg/webui/console/store/middleware/logics/applications.js
```diff
@@ -31,6 +31,28 @@ const getApplicationLogic = createRequestLogic({
...
 })
 
+const updateApplicationLogic = createRequestLogic({
+  type: applications.UPDATE_APP,
+  async process ({ action }) {
+    const { id, patch } = action.payload
+
+    const result = await api.application.update(id, patch)
+
+    return { ...result, ...patch }
```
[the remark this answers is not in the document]
bafonins: any reasons for this?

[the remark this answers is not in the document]
kschiffer: From the request-response, we know 100% that it matches the new data. I could see cases where values might be in some way marshaled or otherwise modified, resulting in a discrepancy between store value and actual value.

bafonins:
Updated description

pkg/webui/console/store/reducers/pagination.js
```diff
@@ -33,6 +36,12 @@ export const createNamedPaginationReducer = function (reducerName, entityIdSelec
         totalCount: payload.totalCount,
         ids: payload.entities.map(entityIdSelector),
       }
+    case DELETE_PAGINATION_SUCCESS:
+      return {
+        ...state,
+        totalCount: state.totalCount - 1,
```
kschiffer:
```suggestion
        totalCount: Math.max(0, state.totalCount - 1),
```

Just an idea. Might catch some issues in rare cases, but I don't know if this is the place to do this.

pkg/webui/console/views/application-general-settings/index.js
```diff
@@ -52,14 +50,15 @@ const validationSchema = Yup.object().shape({
...
-@connect(function (state) {
-  return {
-    application: selectSelectedApplication(state),
-  }
-})
+@connect(state => ({
+  application: selectSelectedApplication(state),
+}),
+dispatch => ({
+  updateApplication: (id, patch) => dispatch(updateApplication(id, patch)),
```
kschiffer: You can use the shorthand here.

Issue #23, opened 2019-01-29
Add utility for testing events publishing from stack packages


johanstokking:
**Summary:** 

In #989 I added a little utility for testing that the NS publishes expected events under certain circumstances. I later removed it due to multiple nasty import cycles (as the events package also test with the test package). This cycle can probably be solved by doing all testing in `_test` packages.

**What is already there? What do you see now?**

See `pkg/networkserver/observability_util_test.go`

**What is missing? What do you want to see?**

It would be nice if we could put the "expect events" in the test package.

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/1008 by @htdvisser

PR #860, opened 2019-06-27
Event testing

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/23

#### Changes

- Allow partial application of `events.Definition` via `events.Definition.BindData` and resulting `events.DefinitionDataClosure`
- Add testing utilities for events


rvolosatovs:
No populators are added in this PR.
I added additional fields to `ttnpb.PopulatorConfig` to avoid the import on `crypto` in `ttnpb` (which should not have been there at the first place)
No functional changes in populators are made

pkg/events/default.go
```diff
@@ -14,24 +14,50 @@
...
-var DefaultPubSub = NewPubSub(DefaultBufferSize)
+var defaultPubSubMu = &sync.Mutex{}
+
+// defaultPubSub is the default event pubsub.
+var defaultPubSub = NewPubSub(DefaultBufferSize)
+
+// SetDefaultPubSub sets pubsub used by the package to ps and returns a function,
+// which can be used to undo the action.
+func SetDefaultPubSub(ps PubSub) func() {
+	defaultPubSubMu.Lock()
```
htdvisser: If production code calls `SetDefaultPubSub` without calling the undo func we'll likely have a deadlock. That should either not happen or be documented properly.

htdvisser: Also, why is this mutex only used for write and not for read? Are we trying to prevent concurrent access to `defaultPubSub` or are we trying to limit the number of times `SetDefaultPubSub` can be called?

rvolosatovs: We're trying to prevent `SetDefaultPubSub` from calling multiple times in production

rvolosatovs: Note, that in some use cases(e.g. testing) we, in fact, want to call `SetDefaultPubSub` multiple times and there we can use this to ensure that the last assignment is "done", and so it's safe to set the `defaultPubSub` to something else

htdvisser: It's not the responsibility of the events package to prevent implementations from replacing the default pubsub. Of course it's bad practice to do so after it's being used, but this is not the problem of the events package.

Even if it were the event package's responsibility, a deadlock is definitely the wrong way to tell implementations that they're doing something wrong.

rvolosatovs:
Moved synchronisation to `test` package

pkg/events/default.go
```diff
@@ -14,24 +14,60 @@
...
+	defaultPubSubSet bool
+)
+
+// SetDefaultPubSub sets pubsub used by the package to ps and returns a function,
+// which can be used to undo the action.
+// SetDefaultPubSub panics if called repeatedly before undo function is called.
+func SetDefaultPubSub(ps PubSub) func() {
+	defaultPubSubMu.Lock()
+	if defaultPubSubSet {
+		panic("SetDefaultPubSub called before previous usage finished")
```
htdvisser: I still think this is weird. The intention of `SetDefaultPubSub` is to replace the pubsub implementation. In production code there is absolutely no requirement of temporarily replacing it and blocking any other replacements until "finished". I understand that this can be useful for testing, but then this belongs in test code (or one of our test packages), and not in production code.

Why don't we move this `defaultPubSubSet` checking to some kind of `TemporarilyReplaceDefaultPubSub` in `pkg/util/test/events.go`?

htdvisser: Another (imo cleaner) option is having the `events.{Publish,Subscribe,Unsubscribe}` funcs take a `context.Context`. Then we can implement a `events.NewContextWithPubSub` and `events.PubSubFromContext` (with fallback to `DefaultPubSub`)

johanstokking:
@htdvisser do you approve the overall package API? If so, I can proceed with reviewing #872

pkg/events/default.go
```diff
@@ -14,24 +14,60 @@
...
+)
 
-// DefaultPubSub is the default event pubsub.
-var DefaultPubSub = NewPubSub(DefaultBufferSize)
+var (
+	defaultPubSubMu = &sync.Mutex{}
+	// defaultPubSub is the default event pubsub.
+	defaultPubSub = NewPubSub(DefaultBufferSize)
+	// defaultPubSubSet records whether SetDefaultPubSub has been called or not.
+	defaultPubSubSet bool
```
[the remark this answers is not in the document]
rvolosatovs: This is required in order to check if `SetDefaultPubSub` is set multiple times to panic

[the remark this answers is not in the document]
htdvisser: It's perfectly legal to replace it multiple times, so this has to go

rvolosatovs:
@htdvisser

pkg/events/default.go
```diff
@@ -14,24 +14,39 @@
 
 package events
 
-import "context"
+import (
```
htdvisser: gofumpt nitpick

PR #861, opened 2019-06-27
[Quickfix] Prevent writing of empty version values in CI

kschiffer:
#### Summary
The release builds for our master branches currently fail because of empty version strings being written to the `package.json`'s, which appears to happen when the commit has not been tagged. In these cases, we can simply abort the mage targets for writing the version values, because they didn't change.

#### Changes
- Add a check to see if the version value is empty and exit the target if so

#### Notes for Reviewers
I'm not really familiar with the release workflow in the CI, so this is my best bet of solving the issue. Please have a look if you can make more sense of it.

### DIFF
--- a/.mage/version.go
+++ b/.mage/version.go
@@ -70,6 +70,10 @@ func (Version) Files() error {
 		return err
 	}
 	version := strings.TrimPrefix(currentVersion, "v")
+	if version == "" {
+		fmt.Println("No version rewrite necessary")
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


htdvisser:
This isn't solving the actual problem, just hiding it. The real problem is https://github.com/travis-ci/travis-ci/issues/7422 which should be solved by adding 

```yml
git:
  depth: false
```

to `travis.yml`. I'll submit a pull request for that.

PR #858, opened 2019-06-27
Uplink MAC state handling fixes

rvolosatovs:
#### Summary

Closes #505 
Closes #795 
~References #42 (closes NS part)~
References #789 (closes uplink part)
Rerences https://github.com/TheThingsNetwork/lorawan-stack/issues/560
~Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/23~ (https://github.com/TheThingsNetwork/lorawan-stack/pull/860)

#### Changes

- Derive MAC state(apply MAC commands, account for MAC reset) of the device before computing MIC
- ~Add testing utilities for events~
- ~Reimplement NS uplink testing~
- ~Harmonize NS testing~
- Do not set `ReceivedAt` in AS uplink(that's a bug, right? @johanstokking - API states it's set by AS)
- ~Do not deduplicate uplinks, which NS cannot process~
- ~Close AS streams when NS closed~
- ~Close AS streams in a nicer way~
- Add NS downlink task after uplink at `Rx1-nsScheduleWindow`
- Avoid adding NS downlink task if Rx2 is missed
- Add NS downlink task even if AS did not (yet) process the uplink
- ~Allow partial application of `events.Definition` via `events.Definition.BindData` and resulting `events.DefinitionDataClosure`~
- Remove failing tests(will be reverted and fixed in #870)

#### Notes for Reviewers

I know the size of this PR is huge, but really most of the changes are NS tests.
If that's a problem - I'm open for suggestions on how to split this up.

PR #862, opened 2019-06-27
[Quickfix] Various fixes and improvements

rvolosatovs:
#### Summary

Various fixes and improvements

#### Changes

- Implement `types.NetID.Copy`
- Fix `ttnpb.RxDelay.Validate`

#### Notes for Reviewers

Part of #858

### DIFF
--- a/pkg/types/netid.go
+++ b/pkg/types/netid.go
@@ -112,6 +112,12 @@ func (id NetID) IDBits() uint {
 	panic(unmatchedNetID)
 }
 
+// Copy stores a copy of id in x and returns it.
+func (id NetID) Copy(x *NetID) *NetID {
+	copy(x[:], id[:])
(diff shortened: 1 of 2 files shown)


pkg/types/netid.go
```diff
@@ -112,6 +112,12 @@ func (id NetID) IDBits() uint {
 	panic(unmatchedNetID)
 }
 
+// Copy stores a copy of id in x and returns it.
+func (id NetID) Copy(x *NetID) *NetID {
```
johanstokking: Aren't we doing something wrong when we need things like these and develop our stack in Go?

rvolosatovs: Almost all types in `types` have this method already, so I'm just being consistent with the rest of the package here.

PR #863, opened 2019-06-27
Fix handling of AsNs streams

rvolosatovs:
#### Summary

Fixes handling of As-Ns streams

#### Changes

- Close AS streams when NS closed
- Close AS streams in a nicer way

#### Notes for Reviewers

Initially was part of #858

### DIFF
--- a/pkg/networkserver/grpc_asns.go
+++ b/pkg/networkserver/grpc_asns.go
@@ -34,7 +34,10 @@ type applicationUpStream struct {
 }
 
 func (s applicationUpStream) Close() error {
-	close(s.closeCh)
+	// First read succeeds when either closeCh is closed by LinkApplication or LinkApplication initializes closing sequence.
+	// Second read succeeds when closeCh is closed by LinkApplication.
+	<-s.closeCh
+	<-s.closeCh
 	return nil
 }
 
@@ -57,6 +60,9 @@ func (ns *NetworkServer) LinkApplication(link ttnpb.AsNs_LinkApplicationServer) 
 		AsNs_LinkApplicationServer: link,
 		closeCh:                    make(chan struct{}),
 	}
+	defer func() {
+		close(ws.closeCh)
+	}()
 
 	uid := unique.ID(ctx, ids)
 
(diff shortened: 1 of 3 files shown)


rvolosatovs:
Blocked by #858

rvolosatovs:
@johanstokking can you clarify what you mean exactly and how it would work?
Do you mean that `LinkApplication` would create a new context wrapped into `context.WithCancel` and `Close()` would call the `context.CancelFunc` returned?
How would `LinkApplication` determine if `io.EOF` should be returned or `ctx.Error()`?
In order for this approach to work, `LinkApplication` would need to maintain 2 completely separate contexts and `Close()` would need to be responsible for deleting the stream from the map. I don't see benefits of such approach, whereas the proposed solution is simpler and more efficient.

pkg/networkserver/grpc_asns.go
```diff
@@ -34,7 +34,10 @@ type applicationUpStream struct {
 }
 
 func (s applicationUpStream) Close() error {
-	close(s.closeCh)
+	// First read succeeds when either closeCh is closed by LinkApplication or LinkApplication initializes closing sequence.
+	// Second read succeeds when closeCh is closed by LinkApplication.
+	<-s.closeCh
+	<-s.closeCh
```
johanstokking: So `Close()` doesn't close but waits for a close

pkg/networkserver/grpc_asns.go
```diff
@@ -65,32 +71,36 @@ func (ns *NetworkServer) LinkApplication(link ttnpb.AsNs_LinkApplicationServer)
...
+	}()
+
+	select {
+	case <-ctx.Done():
+		err = ctx.Err()
+		logger.WithError(err).Debug("Context done - close link")
 		return err
-	case <-ws.closeCh:
-		return errDuplicateSubscription
+	case ws.closeCh <- struct{}{}:
```
johanstokking: You may have `ctx.Done()` so you don't guarantee writing to `closeCh`

rvolosatovs: Yes, that's expected

pkg/networkserver/grpc_asns.go
rvolosatovs: It triggers the "closing sequence" in `LinkApplication`

rvolosatovs:
Is there a problem with the proposed implementation?
It's really very simple - whenever `Close()` is called(possibly multiple times and/or concurrently) `LinkApplication` is signaled to close the stream gracefully - i.e. return `io.EOF` and delete the stream from the map.
If context is `Done()` before `Close()` is called, the `ctx.Err()` is returned and the stream is deleted from the map.

After the stream is closed(either via `Close()` or via `ctx.Done()`) all `Close()` calls will return immediately(since the channel is closed by `LinkApplication`).

pkg/networkserver/networkserver.go
```diff
@@ -346,3 +346,16 @@ func (ns *NetworkServer) RegisterHandlers(s *runtime.ServeMux, conn *grpc.Client
...
 }
+
+func (ns *NetworkServer) Close() {
+	ns.Component.Close()
+
+	logger := ns.Logger()
+	for _, cl := range ns.applicationServers {
+		logger.Debug("Close Application Server link")
+		if err := cl.Close(); err != nil {
+			logger.WithError(err).Warn("Failed to close AS link")
```
johanstokking: 1. Consistent component naming, `Application Server`
2. `else` for `Debug` log below

rvolosatovs:
@johanstokking I simply do not understand how to implement following semantics using 
`context.Context`:
- `Close()` could be called multiple times(possibly concurrently) safely
- `Close()` only returns once the stream is deleted from the map
- `Close()` returns instantly if stream is already closed
- `LinkApplication` is responsible for all the housekeeping, i.e. deleting/adding the strteam to the map

`context.WithCancel` can be used to cancel the context and trigger the "closing sequence" in `LinkApplication`, but there's no way to actually wait for `LinkApplication` in `Close()` to finish all the "housekeeping" if only one context is used.

pkg/networkserver/grpc_asns.go
```diff
@@ -65,32 +71,36 @@ func (ns *NetworkServer) LinkApplication(link ttnpb.AsNs_LinkApplicationServer)
...
+		logger.Debug("Close() is called, close link")
```
htdvisser: Inconsistent with debug log above, which uses `- close link`

htdvisser:
For the record, I think @rvolosatovs's points justify this "custom" cancelation mechanism. Especially the requirement _Close() only returns once the stream is deleted from the map_ is one that Go doesn't solve. Cancelation is easy, but waiting for cleanup after cancelation is currently impossible with context.Context.

johanstokking:
> `context.WithCancel` can be used to cancel the context and trigger the "closing sequence" in `LinkApplication`, but there's no way to actually wait for `LinkApplication` in `Close()` to finish all the "housekeeping" if only one context is used.

Do you need to wait for that though? Is it even desirable to wait for that? If so, why? 

Also, `Close()` is imperative mood, which is wrong (see earlier comment); it should be `Closed()` then, like `Done()` that `context.Context` provides.

pkg/networkserver/grpc_asns.go
johanstokking: Even though it's Debug I would not print code

PR #870, opened 2019-06-27
NS uplink testing

### LINKED ISSUE #42: Harmonize testing of RPC servers
**Summary:** 

Harmonize testing of RPC servers.

**Why do we need this?**

To be consistent across the codebase.

Two classes of inconsistencies;

1. Direct invocation of RPC exposed methods vs creating a client to call RPC services
2. Overriding an RPC server by a test sink vs spawning mock cluster peers

In general, I prefer a more "black box testing" approach; i.e. outside testing over inside hooks. This is still unit testing, but then the component from outside.

Regarding the two inconsistencies described above, I prefer;

1. Creating a client to call RPC servers, because this will include the interceptors for security, validation, etc. This may be or become functionally or non-functionally important to include in testing.
2. Spawning mock cluster peers, because this will include the asserting the outgoing context, again for security, and does not skip client RPC code that is relevant for testing

The occurrences of 1 and 2 are only in the NS and JS. @rvolosatovs if you agree with this approach, please list the tests that are relevant, and work with @M-Gregoire or yourself to harmonize this.

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/1361 by @johanstokking

rvolosatovs:
#### Summary

References #42 (closes NS part)

#### Changes

- Add testing utilities for events
- Reimplement NS uplink testing
- Harmonize NS testing
- Do not deduplicate uplinks, which NS cannot process

#### Notes for Reviewers

Requires #860


rvolosatovs:
Blocked by #858

htdvisser:
Removed my review assignment. Github requested me as code owner, but appartenly those "owned" files have since been removed from the Pull Request.

Issue #878, opened 2019-06-28
Fix form validation regexp


bafonins:
#### Summary

Currently, the console uses regexps to validate form fields that copied from the entity protos, see [regexp.js](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/webui/console/lib/regexp.js). In some cases it doesnt work as expected.

#### Steps to Reproduce

1. Go to the Add Gateway page
2. Fill in `some.host:1885` in the gateway server address field
3. See failed validation, while this is a valid server address

#### What do you see now?

<img width="643" alt="Screenshot 2019-06-28 at 16 26 04" src="https://user-images.githubusercontent.com/16374166/60349226-75774780-99c1-11e9-8ac8-0a3acc7b50c1.png">

#### What do you want to see instead?

No errors

#### Environment

browser (console)

#### How do you propose to implement this?

The problem is that we use regular expression literal which does not require to escape the `\` with another `\`, while the expressions copied from the protos do include those.
I propose to use the `RegExp` constructor for expressions copied from the protos to leave them unchanged and avoid inconsistencies between the frontend and the backend.

#### Can you do this yourself and submit a Pull Request?

Yes

PR #879, opened 2019-06-28
Fix address validation regexp in console

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/878

### DIFF
--- a/pkg/webui/console/lib/regexp.js
+++ b/pkg/webui/console/lib/regexp.js
@@ -15,7 +15,7 @@
 // alphanumeric, dashes, lowercase
 export const id = /^[a-z0-9]+(-[a-z0-9]+)*$/
 export const apiKey = /^NNSXS.[A-Z0-9]{39}.[A-Z0-9]{52}$/
-export const address = /^(?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*(?:[A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])(?::[0-9]{1,5})?$|^$/
+export const address = new RegExp('^(?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*(?:[A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])(?::[0-9]{1,5})?$|^$')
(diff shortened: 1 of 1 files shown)


pkg/webui/console/lib/regexp.js
```diff
@@ -15,7 +15,7 @@
 // alphanumeric, dashes, lowercase
 export const id = /^[a-z0-9]+(-[a-z0-9]+)*$/
 export const apiKey = /^NNSXS.[A-Z0-9]{39}.[A-Z0-9]{52}$/
-export const address = /^(?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*(?:[A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])(?::[0-9]{1,5})?$|^$/
+export const address = new RegExp('^(?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*(?:[A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])(?::[0-9]{1,5})?$|^$')
 export const latitude = /^[-+]?([1-8]?\d(\.\d+)?|90(\.0+)?)$/
-export const longitude = /^\s*[-+]?(180(\.0+)?|((1[0-7]\d)|([1-9]?\d))(\.\d+)?)$/
+export const longitude = /'^\s*[-+]?(180(\.0+)?|((1[0-7]\d)|([1-9]?\d))(\.\d+)?)$/
```
johanstokking: Is this ok?

bafonins: nope

kschiffer:
> Can't we just get validation regexes directly from the API description that we generate for the console? https://github.com/TheThingsNetwork/lorawan-stack/blob/master/sdk/js/generated/api.json has the exact rules for each field. If we can derive console validation rules from that file on build, we don't have to change things in multiple places if this ever changes.
> 
> The format of API keys could for example change in [TheThingsIndustries/lorawan-stack#1666](https://github.com/TheThingsIndustries/lorawan-stack/issues/1666), which would break all API key fields in the console, while this wouldn't be caught in any tests.

I agree that this would be a lot better, But let's look into this in a separate issue. I'll create one. For the time being this fix will serve us well.

bafonins:
> Can't we just get validation regexes directly from the API description that we generate for the console? 

I can imagine this not as trivial as i might seem.

PR #882, opened 2019-06-28
Avoid retrying non-retryable downlink errors

rvolosatovs:
#### Summary

References #631 

#### Changes

- Avoid retrying non-retryable downlink errors

#### Notes for Reviewers

@johanstokking in the issue you mentioned sending an uplink to AS about downlink being failed, but that's quite specific case, and will be addressed later:
- If attempt to schedule Class A downlink returns a non-retryable error, then NS shall not retry(i.e. add a new downlink slot) and a new downlink slot will only be added when an uplink is received by NS. Note, that the same, perfectly valid application downlink, would be scheduled in that new downlink slot, if present and applicable.
- If we're dealing with a class B/C device and absolute time is not set or application downlink is not available, the same happens as above, however the downlink slot may also be added as a consequence of application downlink queue being updated(in which case we should try again anyway, ideally, of course only when we get new downlink paths to try, but proposed implementation should be good enough for most use cases)
- If we're dealing with a class B/C device and Class B/C settings are set:
   - If absolute time is expired - the NS removes the downlink from the queue and notifies AS, which is already implemented
   - If absolute time is too far in the future - the NS adds a downlink slot later
   - If absolute time is available for scheduling, but non-retryable error is returned and `ClassBC` is set, it must be because of `ClassBC` parameters being incorrect. This is the only case, when the AS has to be notified

### DIFF
--- a/pkg/networkserver/downlink.go
+++ b/pkg/networkserver/downlink.go
@@ -566,12 +566,17 @@ func downlinkPathsFromRecentUplinks(ups ...*ttnpb.UplinkMessage) []downlinkPath 
 	return nil
 }
 
+type scheduledDownlink struct {
+	Message    *ttnpb.DownlinkMessage
+	TransmitAt time.Time
+}
+
 // scheduleDownlinkByPaths attempts to schedule payload b using parameters in req using paths.
 // scheduleDownlinkByPaths discards req.DownlinkPaths and mutates it arbitrarily.
 // scheduleDownlinkByPaths returns the scheduled downlink or error.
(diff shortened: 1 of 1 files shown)


rvolosatovs:
So the only case when the failure must be reported to AS is actually when downlink paths are set for the downlink.
An expired downlink being scheduled is very unlikely and if it does happen, the failure would be sent at next attempt to schedule(triggered by e.g. queue update or uplink)

rvolosatovs:
Testing will be added later, after all the NS PR's are merged, since there are a lot of changes related to testing.

rvolosatovs:
@johanstokking

PR #883, opened 2019-06-29
Application PubSub registry

### LINKED ISSUE #830: Use registry for AS pub/sub configuration per application
#### Summary

Use registry for AS pub/sub configuration per application

#### Why do we need this?

To differentiate pub/sub configuration(s) per application, i.e. whether they're enabled, what endpoints, etc

Since #673 we have pub/sub configuration on the AS instance level. This met the requirements, but doesn't allow for enough flexibility at runtime and does not provide enough application context.

Use a webhooks-like registry with configuration per application.

- Add a configuration proto message for the pub/sub configuration per application (like `ApplicationWebhook`), i.e. `ApplicationPubSub`. @htdvisser should we store the `go-cloud` url (pro: easy, con: we probably need some config help in CLI and Console and gives lock-in to `go-cloud`) or a one-of per supported pub/sub from where we build the url?
- Add a registry and Redis implementation for that (only get/set/delete)
- Add a service and rpcs to manage the pub/sub config
- Wire the subscription just like webhooks

It is basically very much like webhooks is w.r.t. configuration and wiring. 

Will review

adriansmares:
#### Summary

Closes #830

#### Changes

- PubSub integrations have been changed to work akin to how Webhooks work - they are stored inside a registry instead of being part of the application server configuration, and now exist on a per-application basis.
- Added the PubSub registry and gRPC support.
- Added CLI support for PubSub integrations.
- The NATS provider is no longer dependent on additional environment variables  - the client is manually created on a per-integration basis. This applies to other integrations which will be added in the future as well, since we are no longer using only the connection URLs.

#### Notes for Reviewers

I believe that @johanstokking should look into the AS implementation and @htdvisser could look into the CLI implementation in case I missed something there. There are no differences between the proposed implementation, both relating to the registry and the setup based on attributes. The huge number of new lines added is from the generated code.

The CLI syntax is very much similar to Webhooks:
```
ttn-lw-cli applications pubsubs set --application-id app1 \
  --pubsub-id ps1 \
  --format json \
  --nats \
  --nats-server-url nats://localhost \
  --base-topic "app1" \
  --downlink-push.topic down.push \
  --downlink-replace.topic down.replace
```

#### Release Notes

- Added the `go-cloud` integration to the Application Server. See `ttn-lw-cli applications pubsubs --help` for more details.


api/applicationserver_pubsub.proto
```diff
@@ -0,0 +1,138 @@
...
+      }
+    };
+  };
+
+  rpc Delete(ApplicationPubSubIdentifiers) returns (google.protobuf.Empty) {
+    option (google.api.http) = {
+      delete: "/as/pubsub/{application_ids.application_id}/{pubsub_id}",
+    };
+  };
+}
```
johanstokking: Line feed

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -17,119 +17,280 @@ package pubsub
...
+		target, err := ps.registry.Get(ctx, ids, ttnpb.ApplicationPubSubFieldPathsNested)
 		if err != nil {
-			return nil, err
+			if !errors.IsNotFound(err) {
+				log.FromContext(ctx).WithError(err).Error("Failed to get integration")
+			}
+			return nil
+		}
+
+		err = ps.integrate(ctx, target)
```
johanstokking: I would call this simply `start()`

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -17,119 +17,280 @@ package pubsub
...
+		switch {
+		case errors.IsFailedPrecondition(err),
+			errors.IsUnauthenticated(err),
+			errors.IsPermissionDenied(err),
+			errors.IsInvalidArgument(err):
+			log.FromContext(ctx).WithError(err).Warn("Failed to integrate")
```
johanstokking: And `Failed to start`; the context is (or should be) clear enough by namespace and log fields

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -17,119 +17,280 @@ package pubsub
...
+			logger.Debug("Publish upstream message")
+		}
 	}
-	for _, subURL := range subURLs {
-		subscription, err := pubsub.OpenSubscription(ctx, subURL)
+}
+
+func (i *integration) handleDown(ctx context.Context, op func(io.Server, context.Context, ttnpb.EndDeviceIdentifiers, []*ttnpb.ApplicationDownlink) error, subscription *pubsub.Subscription) {
+	logger := log.FromContext(ctx)
+	for ctx.Err() == nil {
```
johanstokking: Early exit

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -17,119 +17,280 @@ package pubsub
...
+		},
+	} {
+		if downlink.subscription == nil {
+			continue
+		}
+		go i.handleDown(log.NewContextWithField(ctx, "operation", downlink.name), downlink.op, downlink.subscription)
+	}
+}
+
+var errAlreadyIntegrated = errors.DefineAlreadyExists("already_integrated", "already integrated to `{application_uid} {pubsub_id}`")
```
johanstokking: - Split fields
- "Already configured"

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -17,119 +17,280 @@ package pubsub
...
+	if !ok {
+		return errFormatNotFound.WithAttributes("format", pb.Format)
+	}
+	i.format = format
+	go i.handleUp(ctx)
+	i.startHandleDown(ctx)
+	logger.Info("Integrated")
+	<-ctx.Done()
+	if err := ctx.Err(); errors.IsCanceled(err) {
+		logger.Info("Integration cancelled")
```
johanstokking: `canceled`; naming follows en-US

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -17,119 +17,280 @@ package pubsub
...
+		return err
+	}
+	format, ok := formats[pb.Format]
+	if !ok {
+		return errFormatNotFound.WithAttributes("format", pb.Format)
+	}
+	i.format = format
+	go i.handleUp(ctx)
+	i.startHandleDown(ctx)
+	logger.Info("Integrated")
```
johanstokking: `Started`

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -17,119 +17,280 @@ package pubsub
...
+	<-ctx.Done()
+	if err := ctx.Err(); errors.IsCanceled(err) {
+		logger.Info("Integration cancelled")
+		i.conn.Shutdown(ctx)
+	}
+	return
+}
 
-	return subs, nil
+func (ps *PubSub) cancelIntegration(ctx context.Context, ids ttnpb.ApplicationPubSubIdentifiers) error {
```
johanstokking: `stop()`

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -17,119 +17,280 @@ package pubsub
...
+	appUID := unique.ID(ctx, ids.ApplicationIdentifiers)
+	psUID := pubsubunique.ID(appUID, ids.PubSubID)
+	if val, ok := ps.integrations.Load(psUID); ok {
+		i := val.(*integration)
+		log.FromContext(ctx).WithFields(log.Fields(
+			"application_uid", appUID,
+			"pubsub_id", ids.PubSubID,
+		)).Debug("Integration cancelled")
```
johanstokking: Same

pkg/applicationserver/io/pubsub/redis/registry.go
```diff
@@ -0,0 +1,262 @@
...
+var errApplicationUID = errors.DefineCorruption("application_uid", "invalid application UID `{application_uid}`")
+
+// Range implements pubsub.Registry.
+func (r PubSubRegistry) Range(ctx context.Context, paths []string, f func(context.Context, ttnpb.ApplicationIdentifiers, *ttnpb.ApplicationPubSub) bool) error {
+	uids, err := r.Redis.SMembers(r.allKey(ctx)).Result()
+	if err != nil {
+		return err
+	}
+	for _, uid := range uids {
+		appUID, psID := pubsubunique.ToUIDs(uid)
```
johanstokking: This should be `ToPubsubID()` right?

pkg/applicationserver/io/pubsub/unique/unique.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"fmt"
+	"regexp"
+)
+
+var uniqueIDPattern = regexp.MustCompile("(.*)\\:(.*)")
+
+// ID generates an unique ID based on an application unique ID and a PubSub ID.
+func ID(appUID, id string) string {
```
johanstokking: Hmm we only use this in the Redis registry, right? Can we put this there instead?

This is going to become confusing with `pkg/unique` I'm afraid

api/applicationserver_pubsub.proto
```diff
@@ -0,0 +1,138 @@
...
+
+  // The provider for the PubSub.
+  enum Provider {
+      AWSSNSSQS = 0; // AWS SQS / SNS
+      AZURESB = 1;   // Azure Service Bus
+      GCPPUBSUB = 2; // GCP PubSub
+      KAFKA = 3;     // Apache Kafka
+      NATS = 4;      // NATS
+      RABBIT = 5;    // RabbitMQ
+  }
```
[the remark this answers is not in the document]
adriansmares: I followed https://github.com/TheThingsNetwork/lorawan-stack/issues/830#issuecomment-503461660 since it allows more flexibility regarding platform specific configuration. It is also the case that since we are missing the other implementations we might need new fields / remove fields later, which would violate API stability.

pkg/applicationserver/io/pubsub/unique/unique.go
adriansmares: Both Redis and the main package use this (the main package uses it for the `sync.Map`). Another name perhaps ?

pkg/applicationserver/applicationserver.go
```diff
@@ -156,12 +156,10 @@ func New(c *component.Component, conf *Config) (as *ApplicationServer, err error
...
 
-	if len(conf.PubSub.PublishURLs) > 0 {
-		pubsub, err := pubsub.Start(as.Context(), as, formatters.JSON, conf.PubSub.PublishURLs, conf.PubSub.SubscribeURLs)
-		if err != nil {
-			return nil, err
-		}
-		as.defaultSubscribers = append(as.defaultSubscribers, pubsub...)
+	if pubsub, err := conf.PubSub.NewPubSub(c, as, conf.PubSub.Registry); err != nil {
+		return nil, err
+	} else if pubsub != nil {
```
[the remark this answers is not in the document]
adriansmares: The `var` introduces a new line 😂.

pkg/applicationserver/io/pubsub/redis/registry.go
adriansmares: This returns both the PubSubID and the Application Unique ID.

pkg/applicationserver/io/pubsub/pubsub.go
adriansmares: I.e. break on error ? Since we cannot do `select` on `Receive` unfortunately.

api/applicationserver_pubsub.proto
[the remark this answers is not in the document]
johanstokking: Maybe I don't understand, but [that comment](https://github.com/TheThingsNetwork/lorawan-stack/issues/830#issuecomment-503461660) supports my argument instead of the opposite, right?

I mean that we have (sub) messages per platform, and a `oneof` which not only indicates which platform to use, but also per-platform specific fields.

pkg/applicationserver/io/pubsub/pubsub.go
johanstokking: Sorry bad comment of mine

pkg/applicationserver/io/pubsub/unique/unique.go
johanstokking: I see. Does it deserve its own package then?

Maybe `PubsubUID(context.Context, ttnpb.ApplicationIdentifiers, string) string` and `SplitPubsubUID() (string, string)`?

pkg/applicationserver/io/pubsub/redis/registry.go
johanstokking: See below

pkg/applicationserver/io/pubsub/unique/unique.go
adriansmares: It's a bit too small for a separate package, but I wanted to keep cyclic dependencies at bay. I'll move them to the main package, renamed....................

api/applicationserver_pubsub.proto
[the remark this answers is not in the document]
adriansmares: I'll convert it to a `oneof` then. We'll gain some type-safety in the process.

johanstokking:
@adriansmares looks ready for review but waiting for the request.

adriansmares:
Re-requested, but after the `oneoff` change the CLI still needs chages.

pkg/applicationserver/io/pubsub/provider/provider.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+// Provider represents a PubSub service provider.
+type Provider interface {
+	// OpenConnection opens the Connection of a given ttnpb.ApplicationPubSub.
+	OpenConnection(ctx context.Context, pb *ttnpb.ApplicationPubSub) (*Connection, error)
+}
+
+var (
+	errNotImplemented    = errors.DefineUnimplemented("provider_not_implemented", "provider `{provider_id}` is not implemented")
```
htdvisser: `provider_not_implemented` and `already_registered` are inconsistent

pkg/applicationserver/applicationserver.go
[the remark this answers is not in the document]
htdvisser: How about rewriting it to

```go
if as.pubsub, err = conf.PubSub.NewPubSub(c, as, conf.PubSub.Registry); err != nil {
	return nil, err
}
```

`as.pubsub` is already nil, so if `NewPubSub` returns a nil pubsub, nothing changes.

pkg/applicationserver/io/pubsub/grpc.go
```diff
@@ -0,0 +1,118 @@
+// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
```
htdvisser: Maybe rename this file to `grpc_webhooks.go` for consistency with package `pkg/applicationserver/io/web`?

pkg/ttnpb/applicationserver_pubsub.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ttnpb
+
+type ApplicationPubSub_Provider = isApplicationPubSub_Provider
```
htdvisser: Please explain (in an inline comment) why we're doing this and why it is acceptable

api/messages.proto
```diff
@@ -235,16 +235,6 @@ message ApplicationInvalidatedDownlinks {
   uint32 last_f_cnt_down = 2;
 }
 
-message DownlinkQueueOperation {
```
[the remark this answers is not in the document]
adriansmares: This was never in a release, only master.

api/applicationserver_pubsub.proto
```diff
@@ -0,0 +1,137 @@
...
+import "google/protobuf/timestamp.proto";
+import "lorawan-stack/api/identifiers.proto";
+
+package ttn.lorawan.v3;
+
+option go_package = "go.thethings.network/lorawan-stack/pkg/ttnpb";
+
+message ApplicationPubSubIdentifiers {
+  ApplicationIdentifiers application_ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (validate.rules).message.required = true];
+  string pubsub_id = 2 [(gogoproto.customname) = "PubSubID", (validate.rules).string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$" , max_len: 36}];
```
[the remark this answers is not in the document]
adriansmares: Settled for pub_sub/PubSub.

adriansmares:
Fixed the comments from @htdvisser and added the missing CLI flags.

pkg/applicationserver/io/pubsub/redis/registry.go
```diff
@@ -0,0 +1,263 @@
...
+
+func (r *PubSubRegistry) allKey(ctx context.Context) string {
+	return r.Redis.Key("all")
+}
+
+func (r *PubSubRegistry) appKey(uid string) string {
+	return r.Redis.Key("uid", uid)
+}
+
+func (r *PubSubRegistry) idKey(appUID, id string) string {
```
rvolosatovs:
```suggestion
func (r *PubSubRegistry) uidKey(appUID, id string) string {
```

nitpick; this is really UID in context of PubSubs. This key uniquely identifies a PubSub.

adriansmares: Also changed `makeUIDKeyFunc`.

api/applicationserver_pubsub.proto
```diff
@@ -0,0 +1,137 @@
...
+  // The NATS provider settings.
+  message NATSProvider {
+    // The server connection URL.
+    string server_url = 1 [(gogoproto.customname) = "ServerURL", (validate.rules).string.uri = true];
+  }
+  // The provider for the PubSub.
+  oneof provider {
+      option (validate.required) = true;
+
+      NATSProvider nats = 17 [(gogoproto.customname) = "NATS"];
```
johanstokking: Looks like there's an indentation issue here

pkg/applicationserver/io/pubsub/grpc_pubsub.go
```diff
@@ -0,0 +1,118 @@
...
+				"ids.application_ids",
+				"ids.pub_sub_id",
+			), nil
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+	if err := ps.stop(ctx, req.ApplicationPubSubIdentifiers); err != nil && !errors.IsNotFound(err) {
+		log.FromContext(ctx).WithError(err).Warn("Failed to cancel integration")
```
johanstokking: - `... to stop pubsub`
- add relevant fields (e.g. uids) to `logger`

pkg/applicationserver/io/pubsub/grpc_pubsub.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil, err
+	}
+	return &ttnpb.ApplicationPubSubs{
+		Pubsubs: pubsubs,
+	}, nil
+}
+
+// Set implements ttnpb.ApplicationPubSubRegistryServer.
+func (ps *PubSub) Set(ctx context.Context, req *ttnpb.SetApplicationPubSubRequest) (*ttnpb.ApplicationPubSub, error) {
+	if err := rights.RequireApplication(ctx, req.ApplicationIdentifiers, ttnpb.RIGHT_APPLICATION_TRAFFIC_READ); err != nil {
```
johanstokking: This this mean that with `READ` rights you can configure a pubsub that subscribes to a topic for downlink (requiring `WRITE` rights?)

pkg/applicationserver/io/pubsub/grpc_pubsub.go
```diff
@@ -0,0 +1,118 @@
...
+	res := &ttnpb.ApplicationPubSub{}
+	if err := res.SetFields(pubsub, req.FieldMask.Paths...); err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+// Delete implements ttnpb.ApplicationPubSubRegistryServer.
+func (ps *PubSub) Delete(ctx context.Context, ids *ttnpb.ApplicationPubSubIdentifiers) (*pbtypes.Empty, error) {
+	if err := rights.RequireApplication(ctx, ids.ApplicationIdentifiers, ttnpb.RIGHT_APPLICATION_TRAFFIC_READ); err != nil {
```
johanstokking: Idem

pkg/applicationserver/io/pubsub/provider/provider.go
```diff
@@ -0,0 +1,54 @@
...
+	errAlreadyRegistered = errors.DefineAlreadyExists("provider_already_registered", "provider `{provider_id}` already registered")
+
+	providers = map[reflect.Type]Provider{}
```
johanstokking: We typically use `map[string]...` for this. It's not wrong though

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -18,118 +18,279 @@ package pubsub
...
 		ctx:       ctx,
 		server:    server,
-		formatter: formatter,
+		registry:  registry,
 	}
+	ps.RegisterTask(ctx, "integrate_all", ps.integrateAll, component.TaskRestartOnFailure)
+	return ps, nil
+}
+
+func (ps *PubSub) integrateAll(ctx context.Context) error {
```
johanstokking: `startAll`

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -18,118 +18,279 @@ package pubsub
...
+func New(c *component.Component, server io.Server, registry Registry) (*PubSub, error) {
+	ctx := log.NewContextWithField(c.FillContext(c.Context()), "namespace", "applicationserver/io/pubsub")
+	ps := &PubSub{
+		Component: c,
 		ctx:       ctx,
 		server:    server,
-		formatter: formatter,
+		registry:  registry,
 	}
+	ps.RegisterTask(ctx, "integrate_all", ps.integrateAll, component.TaskRestartOnFailure)
```
johanstokking: `pubsubs_start_all`

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -18,118 +18,279 @@ package pubsub
...
+func (ps *PubSub) integrateAll(ctx context.Context) error {
+	return ps.registry.Range(ctx, []string{"ids"},
+		func(ctx context.Context, _ ttnpb.ApplicationIdentifiers, pb *ttnpb.ApplicationPubSub) bool {
+			ps.startIntegrationTask(ctx, pb.ApplicationPubSubIdentifiers)
+			return true
+		},
+	)
+}
+
+var integrationBackoff = []time.Duration{100 * time.Millisecond, 1 * time.Second, 10 * time.Second}
```
johanstokking: `startBackoff`

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -18,118 +18,279 @@ package pubsub
...
 
-	subs := []*io.Subscription{}
-	for _, pubURL := range pubURLs {
-		topic, err := pubsub.OpenTopic(ctx, pubURL)
+func (ps *PubSub) startIntegrationTask(ctx context.Context, ids ttnpb.ApplicationPubSubIdentifiers) {
```
johanstokking: `startTask`

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -18,118 +18,279 @@ package pubsub
...
+	ctx = log.NewContextWithFields(ctx, log.Fields(
+		"application_uid", unique.ID(ctx, ids.ApplicationIdentifiers),
+		"pub_sub_id", ids.PubSubID,
+	))
+	ps.StartTask(ctx, "integrate", func(ctx context.Context) error {
```
johanstokking: `pubsub`; we probably need the UID here, right? Don't we do that in AS linking too?

api/applicationserver_pubsub.proto
adriansmares: The newline ?

pkg/applicationserver/io/pubsub/grpc_pubsub.go
adriansmares: The UIDs are added above (line 74).

pkg/applicationserver/io/pubsub/pubsub.go
adriansmares: Not needed - you can use the `ApplicationPubSubIdentifiers` to get the actual PubSub inside the task.

adriansmares: AS linking needs application UIDs since the links are indexed on a per-application basis.

api/applicationserver_pubsub.proto
johanstokking: 4 chars instead of 2

pkg/applicationserver/io/pubsub/pubsub.go
johanstokking: I meant the task ID, but indeed this is not necessary

pkg/applicationserver/io/pubsub/grpc_pubsub.go
```diff
@@ -68,7 +69,7 @@ func (ps *PubSub) List(ctx context.Context, req *ttnpb.ListApplicationPubSubsReq
 
 // Set implements ttnpb.ApplicationPubSubRegistryServer.
 func (ps *PubSub) Set(ctx context.Context, req *ttnpb.SetApplicationPubSubRequest) (*ttnpb.ApplicationPubSub, error) {
-	if err := rights.RequireApplication(ctx, req.ApplicationIdentifiers, ttnpb.RIGHT_APPLICATION_TRAFFIC_READ); err != nil {
+	if err := rights.RequireApplication(ctx, req.ApplicationIdentifiers, ttnpb.RIGHT_APPLICATION_TRAFFIC_DOWN_WRITE); err != nil {
```
johanstokking: I think we need to assert both rights

adriansmares: In both cases ?

johanstokking: Yeah cause it implicitly ends up doing read/write, bypassing auth checks

@htdvisser what do you think?

htdvisser: I think that in both the Webhooks and Pubsub we should require `RIGHT_APPLICATION_SETTINGS_BASIC` for managing the integrations. Or perhaps we could introduce a new right for managing application integrations

johanstokking: Webhooks and pub/sub are strictly traffic related though.

How about a new right `TRAFFIC_ALL` that implies read/write?

htdvisser: Yes, but adding/removing webhook/pubsub configurations is a management action. If I give an external platform an API key for reading/writing traffic, it's strange that they would be able to read/update/delete my other webhook/pubsub configurations.

johanstokking: Right. In that case we need to assert `SETTINGS_BASIC` as well. We must still assert `TRAFFIC` rights, because you should not be able to read/write all traffic with just settings rights.

I'm also okay with a new right instead of asserting 4 rights or 1 + 1 implied traffic all right.

johanstokking:
Blocking #919

pkg/applicationserver/io/pubsub/grpc_pubsub.go
```diff
@@ -69,7 +69,12 @@ func (ps *PubSub) List(ctx context.Context, req *ttnpb.ListApplicationPubSubsReq
 
 // Set implements ttnpb.ApplicationPubSubRegistryServer.
 func (ps *PubSub) Set(ctx context.Context, req *ttnpb.SetApplicationPubSubRequest) (*ttnpb.ApplicationPubSub, error) {
-	if err := rights.RequireApplication(ctx, req.ApplicationIdentifiers, ttnpb.RIGHT_APPLICATION_TRAFFIC_DOWN_WRITE); err != nil {
+	if err := rights.RequireApplication(ctx, req.ApplicationIdentifiers,
+		ttnpb.RIGHT_APPLICATION_SETTINGS_BASIC,
+		ttnpb.RIGHT_APPLICATION_TRAFFIC_READ,
+		ttnpb.RIGHT_APPLICATION_TRAFFIC_UP_WRITE,
```
johanstokking: Writing uplink is not something that we'll do through pub/sub

adriansmares: Removed in 820d316319f8f80da2b92fe6577d72f7e06a554d . Had to rebase in order to get the CI to pass again since `protoc` got updated.

pkg/applicationserver/io/pubsub/redis/registry.go
```diff
@@ -0,0 +1,263 @@
...
+			if err != nil {
+				return err
+			}
+		}
+
+		var sets []string
+		pb, sets, err = f(pb)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
rvolosatovs: still not addressed

pkg/applicationserver/io/pubsub/redis/registry.go
```diff
@@ -0,0 +1,263 @@
...
+
+var (
+	errInvalidFieldmask   = errors.DefineInvalidArgument("invalid_fieldmask", "invalid fieldmask")
+	errInvalidIdentifiers = errors.DefineInvalidArgument("invalid_identifiers", "invalid identifiers")
+)
+
+// appendImplicitPubSubGetPaths appends implicit ttnpb.ApplicationPubSub get paths to paths.
+func appendImplicitPubSubGetPaths(paths ...string) []string {
+	return append(append(make([]string, 0, 4+len(paths)),
+		"created_at",
```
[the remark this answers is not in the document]
rvolosatovs: still not addressed

PR #884, opened 2019-07-01
Stop doing shallow git clones in Travis

htdvisser:
#### Summary

This Pull Request fixes the problem that git tags aren't present in Travis builds.

Replaces #861.

#### Changes

- Tell Travis to not do a shallow clone of the repository


kschiffer:
https://github.com/TheThingsNetwork/lorawan-stack/blob/master/package.json#L3

This seems to have broken version bumps in `package.json`, or am I missing something?

Issue #875, opened 2019-06-28
Minor layout, styling and text improvements


johanstokking:
#### Summary

Minor layout, style and text improvements for the Console.

#### Why do we need this?

To make it even more awesome than it already is.

#### What do you want to see?

- [x] Rename "Connected Devices" to "Devices". We can't claim they are "connected" (#888)
- [ ] Add "connected" indicator to gateway list by checking GS connection stats (#1155)
- [x] Increase page size from 5 to 10 (device list in application overview, table can be a bit more condensed) or 20 (list views of applications, devices and gateways) (#907)
- [ ] Use consistent Title Case, i.e. "Message _T_ype" in webhook config (#1273)
- [x] We need alphabetic sorting in lists by default (#1054 / #1107)
- [x] Move API key name to first column (#895)
- [x] Make terms like "TTN Stack" configurable (currently used in things like "TTN Stack Console" and "TTN Stack Account"). We have the site names for this, right?

#### Can you do this yourself and submit a Pull Request?

Can review

bafonins:
> Increase page size from 5 to 10 (device list in application overview, table can be a bit more condensed) or 20 (list views of applications, devices and gateways)

There is the `Devices` entry in the side navigation with the list of devices (25 devices per page).

> We need alphabetic sorting in lists by default

Shouldnt this be done by IS?

bafonins:
> Add "connected" indicator to gateway list by checking GS connection stats

and

>  Increase page size from 5 to 10 (device list in application overview, table can be a bit more condensed) or 20 (list views of applications, devices and gateways)

I can agree that showing the connection status for gateways is nice, however do we really want to make 5-20 calls to fetch the stats on every page load?

johanstokking:
> > Increase page size from 5 to 10 (device list in application overview, table can be a bit more condensed) or 20 (list views of applications, devices and gateways)
> 
> There is the `Devices` entry in the side navigation with the list of devices (25 devices per page).

OK, but not the applications and gateways, see e.g.

<img width="731" alt="Screenshot 2019-07-01 at 14 32 44" src="https://user-images.githubusercontent.com/13334001/60436934-78667800-9c0d-11e9-8bd8-687b02e81e45.png">

> > We need alphabetic sorting in lists by default
> 
> Shouldnt this be done by IS?

[...]

PR #895, opened 2019-07-01
Move API key name to first column

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/875

>  then maybe we should make the columns ID, name, rights.

Screenshot 2019-07-02 at 09 58 58

### DIFF
--- a/pkg/webui/console/containers/api-keys-table/api-keys-table.styl
+++ b/pkg/webui/console/containers/api-keys-table/api-keys-table.styl
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+.right-tag-group
+  display: flex
+  justify-content: flex-end
+
 .right-tag
   overflow: hidden
   white-space: nowrap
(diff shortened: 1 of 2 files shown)


johanstokking:
Yeah good point. If we keep the sorting by ID (makes sense, `name` is optional), then maybe we should make the columns ID, name, rights.

bafonins:
> then maybe we should make the columns ID, name, rights.

How is this better than the order before?

johanstokking:
> > then maybe we should make the columns ID, name, rights.
> 
> How is this better than the order before?

The name was burried on the right

bafonins:
updated description

kschiffer:
The ID value is very unhelpful IMO. I don't see any user scanning the table by a cryptic hash value. If we can, we should really sort by creation date or the name value. We might even consider making the name field mandatory, or setting default values like `API Key 1`, which I would recommend. We can then hide or at least abbreviate the ID field in the table view (that is, if there is no essential use of the ID value that I have missed).

PR #900, opened 2019-07-01
Use configured component addresses as default on device update

### LINKED ISSUE #899: CLI clears server addresses on device update
#### Summary

CLI clears server addresses on device update

#### Steps to Reproduce

1. Create OTAA device usual way, see Getting Started
2. Update device; change a JS field for example, i.e. `--resets-join-nonces=true`

#### What do you see now?

CLI performs the update but clears `join-server-address` in step 2. CLI command keep working only because of local config.

CLI shouldn't clear the address.

Propose to fill out the addresses on update, before applying the field mask from args, so args can override but fallback to config. If addresses mismatch, there's already a warning.

Yes

johanstokking:
#### Summary

Closes #899

#### Changes

- Use configured component address as default on device update
- Pretty unrelated; do not require MAC version when it's not necessary in CLI. When bulk create in JS, there's no need for MAC version, for example

#### Release Notes

- Fixed clearing component addresses on updating end devices through CLI

### DIFF
--- a/cmd/ttn-lw-cli/commands/end_devices.go
+++ b/cmd/ttn-lw-cli/commands/end_devices.go
@@ -492,12 +492,21 @@ var (
 			isPaths, nsPaths, asPaths, jsPaths := splitEndDeviceSetPaths(device.SupportsJoin, paths...)
 
 			if len(nsPaths) > 0 && config.NetworkServerEnabled {
+				if device.NetworkServerAddress == "" {
+					device.NetworkServerAddress = getHost(config.NetworkServerGRPCAddress)
+				}
 				isPaths = append(isPaths, "network_server_address")
 			}
 			if len(asPaths) > 0 && config.ApplicationServerEnabled {
+				if device.ApplicationServerAddress == "" {
+					device.ApplicationServerAddress = getHost(config.ApplicationServerGRPCAddress)
+				}
 				isPaths = append(isPaths, "application_server_address")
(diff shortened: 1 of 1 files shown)


cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -479,7 +479,11 @@ var (
 				logger.Warn("No fields selected, won't update anything")
 				return nil
 			}
-			var device ttnpb.EndDevice
+			device := ttnpb.EndDevice{
```
htdvisser: I don't think this is the right place for this. Setting the "fallback" addresses from config (if still zero in `device`) belongs together with the code that adds the field masks for the fields around L498 (`if len(nsPaths) > 0 && config.NetworkServerEnabled {`) 

Alternatively we can put it together with the section around L521 (`// EUIs can not be updated, ...`)

johanstokking: I agree it makes more sense there, but we would be overriding the value that comes from the passed arguments.

So afaik it has to be before `util.SetFields(&device, setEndDeviceFlags)`

htdvisser: > Setting the "fallback" addresses from config (if still zero in `device`)

With that "if still zero in `device`" I meant `if device.XXXAddress == ""` checks

PR #910, opened 2019-07-02
Use Backend Interfaces Result objects

### LINKED ISSUE #908: Invalid Result field type in Backend Interfaces
#### Summary

Invalid Result field type in Backend Interfaces

Reported by @brocaar

#### Steps to Reproduce

1. Send valid `JoinReq` message to Join Server
2. Receive `JoinAns` and observe `Result` field

#### What do you see now?

`"Result": "Success"`

Should be a Result Object:

Screenshot 2019-07-02 at 13 37 37

Change structs

Yes

johanstokking:
#### Summary

Closes #908 

#### Changes

- Use Result Object
- Return non-internal error cause in `Description`
- Add testing

### DIFF
--- a/pkg/component/interop_test.go
+++ b/pkg/component/interop_test.go
@@ -47,7 +47,9 @@ func (m mockInterop) JoinRequest(ctx context.Context, req *interop.JoinReq) (*in
 	}
 	return &interop.JoinAns{
 		JsNsMessageHeader: ansHeader,
-		Result:            interop.ResultSuccess,
+		Result: interop.Result{
+			ResultCode: interop.ResultSuccess,
+		},
 	}, nil
 }
 
@@ -58,7 +60,9 @@ func (m mockInterop) AppSKeyRequest(ctx context.Context, req *interop.AppSKeyReq
 	}
 	return &interop.AppSKeyAns{
 		JsAsMessageHeader: ansHeader,
-		Result:            interop.ResultSuccess,
+		Result: interop.Result{
+			ResultCode: interop.ResultSuccess,
+		},
 	}, nil
 }
 
@@ -69,7 +73,9 @@ func (m mockInterop) HomeNSRequest(ctx context.Context, req *interop.HomeNSReq) 
 	}
 	return &interop.HomeNSAns{
 		JsNsMessageHeader: ansHeader,
-		Result:            interop.ResultSuccess,
+		Result: interop.Result{
+			ResultCode: interop.ResultSuccess,
+		},
 	}, nil
 }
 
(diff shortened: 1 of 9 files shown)


pkg/component/interop_test.go
```diff
@@ -158,6 +164,6 @@ func TestInteropTLS(t *testing.T) {
 		if !a.So(json.NewDecoder(res.Body).Decode(&msg), should.BeNil) {
 			t.FailNow()
 		}
-		a.So(msg.Result, should.Equal, interop.ResultUnknownSender)
+		a.So(msg.Result.ResultCode, should.Equal, interop.ResultUnknownSender)
```
rvolosatovs: I'd prefer asserting `msg.Result` or even the whole `msg` via `should.Resemble` to ensure nothing is missed, but that's personal preference I guess.

pkg/interop/errors.go
```diff
@@ -48,26 +48,32 @@ var (
...
 // ErrorHandler is an echo.HTTPErrorHandler.
 func ErrorHandler(err error, c echo.Context) {
 	if httpErr, ok := err.(*echo.HTTPError); ok {
 		c.JSON(httpErr.Code, httpErr.Message)
 		return
 	}
 
-	result := ResultOther
-	statusCode := http.StatusInternalServerError
+	result, statusCode, description := ResultOther, http.StatusInternalServerError, ""
```
rvolosatovs: :frowning_face: 
This refactoring does not seem to help readability, please consider just declaring an additional `var description string` below the previous lines

pkg/interop/errors_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"testing"
+
+	echo "github.com/labstack/echo/v4"
+	"github.com/smartystreets/assertions"
+	"go.thethings.network/lorawan-stack/pkg/errors"
+	"go.thethings.network/lorawan-stack/pkg/util/test/assertions/should"
+)
+
+func TestErrorHandler(t *testing.T) {
+	errTest := errors.DefineInvalidArgument("test", "test error")
```
rvolosatovs: This will `panic` if another test in this package defines an error with the same ID.
Minor issue, but consider renaming this to something unique across package, e.g. `test-error-handler` ?

pkg/interop/interop_test.go
```diff
@@ -69,7 +69,7 @@ func TestServeHTTP(t *testing.T) {
 				}
 				var msg ErrorMessage
 				err := json.NewDecoder(res.Body).Decode(&msg)
-				return a.So(err, should.BeNil) && a.So(msg.Result, should.Equal, ResultUnknownSender)
+				return a.So(err, should.BeNil) && a.So(msg.Result.ResultCode, should.Equal, ResultUnknownSender)
```
rvolosatovs: IDEM

pkg/interop/errors.go
htdvisser: On one hand I agree that this isn't perfect. `""` signals that it's explicitly `""`, while declaring a `var description string` signals that we'll assign it later. However, I don't think this is important enough to fix. We all understand what's going on.

PR #912, opened 2019-07-02
Network Server uplink fixes and improvements

rvolosatovs:
#### Summary

Combination of
https://github.com/TheThingsNetwork/lorawan-stack/pull/872
https://github.com/TheThingsNetwork/lorawan-stack/pull/863
https://github.com/TheThingsNetwork/lorawan-stack/pull/858
https://github.com/TheThingsNetwork/lorawan-stack/pull/870
https://github.com/TheThingsNetwork/lorawan-stack/pull/860

#### Changes

See the PRs referenced above. 

#### Notes for Reviewers

Preserved merges, rebased on top of `master` and added 3 minor commits to this:
https://github.com/TheThingsNetwork/lorawan-stack/commit/64c6bb80d4574f51e91e3d737c5f08c7b52b9744
https://github.com/TheThingsNetwork/lorawan-stack/commit/46e98fc793fd80349e929c7b6f8659511705397d 
https://github.com/TheThingsNetwork/lorawan-stack/commit/d6f77b16f9e2239106c97ae90feced11feee57ab 
https://github.com/TheThingsNetwork/lorawan-stack/pull/912/commits/c829d6c1fd2ce2e32450fa8f4e2a8693a5b6cf36 @htdvisser 	
https://github.com/TheThingsNetwork/lorawan-stack/pull/912/commits/3a8ae54adb2c30149ff93d31ede03523b8ecb0c0 @johanstokking


rvolosatovs:
I originally preserved the merge commits on purpose, as mentioned in description.
The whole thing was tested on metal before event changes were made, but those only concern testing.

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -664,9 +657,30 @@ func TestGatewayServer(t *testing.T) {
...
+						defer cancel()
+
+						upEvents := map[string]events.Channel{}
+						for _, event := range []string{"gs.up.receive", "gs.down.tx.success", "gs.down.tx.fail", "gs.status.receive"} {
+							upEvents[event] = make(events.Channel, 5)
+						}
+						defer test.SetDefaultEventsPubSub(&test.MockEventPubSub{
+							PublishFunc: func(ev events.Event) {
+								switch name := ev.Name(); name {
+								case "gs.up.receive", "gs.down.tx.success", "gs.down.tx.fail", "gs.status.receive":
```
rvolosatovs: I tried to refactor this into using the assertions in `test` package, but there are too many unhandled events published in the test script, so I adapted it in such a way.

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -664,9 +657,30 @@ func TestGatewayServer(t *testing.T) {
 					t.Run(tc.Name, func(t *testing.T) {
 						a := assertions.New(t)
 
+						testCtx := test.ContextWithT(test.Context(), t)
+						testCtx, cancel := context.WithTimeout(testCtx, 5*timeout)
```
johanstokking: What is the benefit of this?

Now if operation 1 takes longer than `timeout`, it passes on to operation 2 which may not have enough time to finish, causing the wrong part of the test to fatal.

pkg/gatewayserver/gatewayserver_test.go
johanstokking: What would be the preferred way instead?

pkg/gatewayserver/gatewayserver_test.go
rvolosatovs: The `time.After(timeout)` was used in the `select` statement in order to prevent deadlocks, right? AFAIK there's no predefined time interval, within which an operation has to finish.
Given this, this approach achieves exactly the same goal as the old `time.After(timeout)`, however not it's done on per-testcase basis, not per-operation.
1. The `ctx.Done()` can be used not only by the test cases, but also by the functions, taking `ctx` (not relevant here, but is often the case in other tests)
2. If an operation 1 took a bit longer than `timeout`, the test case should not fail IMO, it should continue execution, since the `timeout` value is just an arbitrary value we chose to avoid deadlocks.

Please note that the context timeout is `5*timeout`, which seemed to approximate the worst-case timeout ceiling used in previous approach.

pkg/gatewayserver/gatewayserver_test.go
rvolosatovs: Catch and assert all events published. (e.g. `gs.gateway.{dis,}connect`, `gs.up.drop`, etc.)

pkg/gatewayserver/gatewayserver_test.go
rvolosatovs: To address case you describe exactly: if operation 1 took longer than `timeout`, but succeeded, then the `timeout` value is too low

johanstokking: > The `time.After(timeout)` was used in the `select` statement in order to prevent deadlocks, right?

No, it is used for testing timeouts of steps.

>AFAIK there's no predefined time interval, within which an operation has to finish.

It's not about the time interval, it is about giving each step a certain time to finish.

Again, what is the benefit of a timeout context? I get the context cancellation part for the test, but simply not the timeout.

> To address case you describe exactly: if operation 1 took longer than `timeout`, but succeeded, then the `timeout` value is too low

The problem is that if you have a timeout of 100 and you have 5 operations, that there's a timeout after 500, right? So if operation 1 takes 490, operation 2 is very unlikely to finish in the remaining 10, so it gives a "operation 2 timeout" error, while that is not the correct fatal.

rvolosatovs:
@johanstokking 	please see updated changes

PR #919, opened 2019-07-03
Update Go dependencies

htdvisser:
#### Summary

Just some Go dependency updates

#### Changes

- Update our Go modules
- Fix import path for OpenCensus Prometheus exporter

#### Notes for Reviewers

@rvolosatovs do we actually still need the `replace github.com/golang/lint => golang.org/x/lint` and `replace github.com/testcontainers/testcontainer-go => github.com/testcontainers/testcontainers-go`?


rvolosatovs:
The reason why we had these replacements was because of some dependencies using legacy paths.
I think the easiest is to just remove them and check if it works :man_shrugging:

htdvisser:
Removed them, still works. However, there were issues with `golang.org/x` modules, so I had to add those 🤦‍♂

Also, we now have test failures in

- [ ] pkg/applicationserver (NATS)
- [ ] pkg/applicationserver/io/mqtt
- [ ] pkg/gatewayserver/io/mqtt
- [ ] pkg/web

@johanstokking / @adriansmares can you look at the first three, then I'll look at the last one

adriansmares:
#883 fixes the NATS test by being compatible with the latest version of `go-cloud.dev`.

johanstokking:
Fixed MQTT tests, now compatible with latest tagged release of Paho

johanstokking:
Blocked by #883

johanstokking:
@adriansmares can you rebase and confirm that `pkg/applicationserver` passes tests again?

adriansmares:
Rebased on top of the latest master. Tests succed now, only the `web` package one is still not working.

htdvisser:
@rvolosatovs 	@johanstokking 	Travis is satisfied. Please approve.

go.mod
```diff
@@ -4,102 +4,110 @@ replace github.com/grpc-ecosystem/grpc-gateway => github.com/TheThingsIndustries
 
 replace github.com/robertkrimen/otto => github.com/TheThingsIndustries/otto v0.0.0-20181129100957-6ddbbb60554a
 
-replace github.com/labstack/echo/v4 => github.com/TheThingsIndustries/echo/v4 v4.0.1-0.20190409124425-ee570f243713
+// Fix versions of golang.org/x modules
```
rvolosatovs: It's quite clear that you `Fix versions of golang.org/x modules` here, can you add reasoning *why* you we need these replacements?

pkg/gatewayserver/io/mqtt/mqtt_test.go
```diff
@@ -93,8 +93,7 @@ func TestAuthentication(t *testing.T) {
 					client.Disconnect(uint(timeout / time.Millisecond))
 				}
 			} else {
-				a.So(ok, should.BeTrue)
```
rvolosatovs: Why did this change?

pkg/applicationserver/io/mqtt/mqtt_test.go
```diff
@@ -115,8 +115,7 @@ func TestAuthentication(t *testing.T) {
 					client.Disconnect(uint(timeout / time.Millisecond))
 				}
 			} else {
-				a.So(ok, should.BeTrue)
```
rvolosatovs: IDEM

pkg/gatewayserver/io/mqtt/mqtt_test.go
johanstokking: This is a breaking change in Paho.

Formerly, on timeout, `true` was returned.

Now, the return value indicates whether the action was successful (i.e. `false` on timeout). So in this statement, we assert `false` as we expect a timeout when the test case is supposed to fail.

See https://github.com/eclipse/paho.mqtt.golang/commit/ba971f185f1bce82b0e8d19c74da95e37c68234a#diff-ac704b2e102a5731d9d1b1707d3bc7a1R80

pkg/applicationserver/io/mqtt/mqtt_test.go
johanstokking: See above

PR #931, opened 2019-07-04
Scroll form errors into view

kschiffer:
#### Summary
This draft PR is a suggestion for a simple solution to scroll form errors into view.

#### Changes
- Add `scroll-into-view-if needed` package (somehow smooth scrolling works for me with this now)
- Add a data attribute to fields, holding whether they are valid or not
- Use this info to scroll the first item that has the error data attribute into view on form submits
- Additionally focus the input
- Scroll form notifications into view if necessary

#### Notes for Reviewers
It's a suggestion for implementation. It sort of leaves the react scope at times to achieve the effect, but I think it is fair here because it enables us having this essential functionality without a lot of logic necessary.

### DIFF
--- a/pkg/webui/components/form/index.js
+++ b/pkg/webui/components/form/index.js
@@ -24,9 +24,20 @@ import FormField from './field'
 import FormSubmit from './submit'
 
 class InnerForm extends React.PureComponent {
+  constructor (props) {
+    super(props)
+    this.notificationRef = React.createRef()
+  }
+
   componentDidUpdate (prevProps) {
     const { formError, isSubmitting, isValid } = this.props
-    const { isSubmitting: prevIsSubmitting } = prevProps
+    const { isSubmitting: prevIsSubmitting, formError: prevFormError } = prevProps
+
+    // Scroll form notification into view if needed
+    if (formError && !prevFormError) {
+      scrollIntoView(this.notificationRef.current, { behavior: 'smooth' })
+      this.notificationRef.current.focus({ preventScroll: true })
+    }
 
     // Scroll invalid fields into view if needed and focus them
     if (!prevIsSubmitting && isSubmitting && !isValid) {
@@ -51,8 +62,12 @@ class InnerForm extends React.PureComponent {
 
     return (
       <form className={className} onSubmit={handleSubmit}>
-        {formError && <Notification error={formError} small />}
-        {formInfo && <Notification info={formInfo} small /> }
+        {(formError || formInfo) && (
+          <div style={{ outline: 'none' }} ref={this.notificationRef} tabIndex="-1">
+            {formError && <Notification error={formError} small />}
+            {formInfo && <Notification info={formInfo} small /> }
+          </div>
+        )}
         <FormContext.Provider value={{
           ...rest,
           horizontal,
--- a/pkg/webui/components/form/index.js
+++ b/pkg/webui/components/form/index.js
@@ -15,6 +15,7 @@
 import React from 'react'
 import { Formik } from 'formik'
 import bind from 'autobind-decorator'
+import scrollIntoView from 'scroll-into-view-if-needed'
 
 import Notification from '../notification'
 import PropTypes from '../../lib/prop-types'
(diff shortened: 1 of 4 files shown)


kschiffer:
Following up on our quick discussion yesterday, I've removed the focus responsibility from the notification component again and work with a `React.createRef()` for notifications now. Additionally, I've added functionality to focus the first invalid field as well.

Please let me know what you think.

pkg/webui/components/form/index.js
```diff
@@ -23,6 +24,30 @@ import FormField from './field'
...
+    const { formError, isSubmitting, isValid } = this.props
+    const { isSubmitting: prevIsSubmitting, formError: prevFormError } = prevProps
+
+    // Scroll form notification into view if needed
+    if (formError && !prevFormError) {
+      scrollIntoView(this.focusGroup.current, { behavior: 'smooth' })
+    }
+
+    // Scroll invalid fields into view if needed and focus them
+    if (!prevIsSubmitting && !formError && isSubmitting && !isValid) {
```
bafonins: why not
```suggestion
    if (!prevIsSubmitting && isSubmitting && !isValid) {
```
imagine this:
1. the user submits the form and gets some error from the backend
2. the user changes data in the form
3. submits
4. gets frontend validation error

The `formError` is still rendered since the `onSubmit` function wasnt called, however I would still expect the invalid field being focused.

pkg/webui/components/form/index.js
```diff
@@ -23,6 +24,30 @@ import FormField from './field'
 import FormSubmit from './submit'
 
 class InnerForm extends React.PureComponent {
+  constructor (props) {
+    super(props)
+    this.focusGroup = React.createRef()
```
bafonins: `error`
`errorRef`
`formErrorRef`
`notification`
`notificationRef`
`formNotificationRef`

pkg/webui/components/form/index.js
```diff
@@ -23,6 +24,30 @@ import FormField from './field'
...
+  }
+
+  componentDidUpdate (prevProps) {
+    const { formError, isSubmitting, isValid } = this.props
+    const { isSubmitting: prevIsSubmitting, formError: prevFormError } = prevProps
+
+    // Scroll form notification into view if needed
+    if (formError && !prevFormError) {
+      scrollIntoView(this.focusGroup.current, { behavior: 'smooth' })
```
bafonins: can we focus on the first field as well?

kschiffer: I think that can lead to the false assumption that the first field is relevant for resolving the error. Also, for error's like "An unknown error occurred. Please try again later.", there is no remedy in changing the form values.

What we can do though, is laying focus in the notification container, so that another tab will put the user on the first input of the form again.

bafonins: > I think that can lead to the false assumption that the first field is relevant for resolving the error.

This is not about making assumptions about which field causes the backend to respond with an error. Scrolling to the top of the form without moving focus to the visible area is poor accessibility. For example, I always use `Tab`, `Tab+Shift` to navigate through forms. This implementation causes the viewport of my browser to jump back to the submit section and forces me to go through all the fields back to the top.

> What we can do though, is laying focus in the notification container

How do you plan to focus on `div`? Using `tabIndex` ?

kschiffer: I agree with you on moving the focus. I just don't agree with putting the focus on the input element, because IMO it would suggest to the user that a different input needs to be made, which we don't know. That's why I suggested putting it on the notification container.

> How do you plan to focus on `div`? Using `tabIndex`?

Yes. Putting a negative `tabIndex` on an element [will make it focusable via JS](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex).

See my implementation in https://github.com/TheThingsNetwork/lorawan-stack/pull/931/commits/bc8d9c460f55710240dd244cece02be98939c878 .

bafonins: i agree. do you want to leave the default outline?
Screenshot 2019-07-16 at 15 09 44

kschiffer: Ah, didn't experience that. Will make sure there will be no outline.

kschiffer: Removed the outline.

PR #947, opened 2019-07-05
Prefer channel RSSI over signal as RSSI and support channel index

### LINKED ISSUE #927: Incorrect handling of RSSIC/RSSIS
#### Summary

After investigating https://github.com/TheThingsNetwork/gateway-connector-bridge/issues/35 I discovered that we've been handling the RSSIC and RSSIS fields incorrectly. This results in incorrect RSSI values.

#### What do you see now?

1. https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/ttnpb/udp/translation.go#L135-L136
2. https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/gatewayserver/io/mqtt/format_protobufv2.go#L170-L174

The real "RSSI" value should be `RSSIC` or ChannelRSSI.

I think the cleanest way to fix this is to add a `SignalRSSI` field to `ttnpb.RxMetadata`, and always copy the value of `ChannelRSSI` to `RSSI` when converting.

Let's discuss the best way forward first and then work on fixing it

johanstokking:
#### Summary

Closes #927 

#### Changes

- Add `signal_rssi` to `RxMetadata`
- Use `channel_rssi` if present, fallback to `signal_rssi` when setting `rssi`
- Move `UplinkMessage.gateway_channel_index` to `RxMetadata[].channel_index`
- Add testing for `RSig` and minor fixes around that

#### Notes for Reviewers

- @htdvisser we're breaking API within components
- @rvolosatovs is the RSSI source any relevant for NS?
- @rvolosatovs is the gateway channel index any relevant for NS?


pkg/ttnpb/udp/packet_data.go
```diff
@@ -53,7 +53,8 @@ type RSig struct {
 	RSSIS  int16   `json:"rssis"`  // RSSI in dBm of the signal (signed integer, 1 DB precision) (Optional)
```
htdvisser: Maybe make this a pointer so that we can differentiate 0 from nil

api/metadata.proto
```diff
@@ -38,11 +40,14 @@ message RxMetadata {
   // Encrypted gateway's internal fine timestamp when the Rx finished (nanoseconds).
   bytes encrypted_fine_timestamp = 6;
   string encrypted_fine_timestamp_key_id = 7 [(gogoproto.customname) = "EncryptedFineTimestampKeyID"];
-  // Received signal strength (dBm).
+  // Received signal strength indicator (dBm).
+  // This value equals `channel_rssi` if present, otherwise `signal_rssi`.
```
htdvisser: No, it simply equals `channel_rssi`.

rvolosatovs:
@johanstokking NS does not use neither RSSI, nor gateway channel index

pkg/ttnpb/udp/translation.go
```diff
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"time"
 
+	pbtypes "github.com/gogo/protobuf/types"
```
htdvisser: I thought we only did this if we were also using our own types package

johanstokking: I typically do this for my future self and other devs to not rename imports later when using `pkg/types` and creating unnecessary diffs

rvolosatovs: I thought we always use `pbtypes`

PR #948, opened 2019-07-05
Device Overview style improvements and fixes

kschiffer:
#### Summary
This PR fixes a couple of styling issues and introduces improvements to the device data overview.

#### Changes
- Fix data sheet table layout (bad column widths)
- Modify data sheet to show a `No data available` message when a data entry has no items
- Fix top margin of the event widget in small screen widths
- Improve display logic of device data in the device overview (to show a general `No data available` message instead of individual entries with `n/a`)


johanstokking:
Can you show `JoinEUI` also before `DevEUI`? That is this view also right?

PR #952, opened 2019-07-07
Fail downlink based on GS errors, fix retry logic

### LINKED ISSUE #631: Fail downlink based on GS errors

### LINKED ISSUE #944: Only schedule ClassBC downlink for class B/C
#### Summary

Only schedule ClassBC downlink for class B/C

#### Why do we need this?

To be compliant with the behavior users expect

If zero `class_b_c` is set, the downlink may be scheduled as Class A

If any `class_b_c` is set, the downlink should only be scheduled as Class B/C in Rx2

Check if `classBC` is set instead of checking individual fields

yes

rvolosatovs:
#### Summary

Closes #631
Closes #944 

#### Changes

- Only retry Class C downlink slots, which can be retried
- Fail downlinks based on GS errors
- Fix unscheduled valid application downlink being removed from the queue, if it is updated


pkg/networkserver/downlink.go
```diff
@@ -573,17 +571,49 @@ type scheduledDownlink struct {
...
+	gtwErrs := make([]error, 0, len(errs))
+	allGatewayErrors := true
+	for _, gsErr := range errs {
+		ttnGSErr, ok := errors.From(gsErr)
+		if !ok {
+			allGatewayErrors = false
+			continue
+		}
+		for _, d := range ttnGSErr.Details() {
+			gtwErr, ok := d.(error)
```
johanstokking: In which case does this cast fail?

rvolosatovs: No clue, details are not type checked - they're just `interface{}` (https://github.com/TheThingsNetwork/lorawan-stack/blob/8102bd25586153ce41207714555402277ee84fbd/pkg/errors/details.go#L33-L56), so there are no guarantees on the type. Better safe than sorry.
@htdvisser any comment on how is this supposed to be used?

johanstokking: But isn't the details than a `ttnpb` type? Or (also) an `error`?

rvolosatovs: According to https://github.com/TheThingsNetwork/lorawan-stack/blob/8102bd25586153ce41207714555402277ee84fbd/pkg/errors/details.go#L47
it's
> Usually structs from ttnpb or google.golang.org/genproto/googleapis/rpc/errdetails.

so there's no guarantee

@htdvisser can we refactor these to use `error` instead of `interface{}`?

htdvisser: Error details wraps existing gRPC API. There is absolutely no guarantee that any of these actually implement the `error` interface. The details is just a list of proto messages. When we send errors over gRPC they contain a single `ttnpb.ErrorDetails` that carries ttn-lw-stack specific details of the error like the error namespace and name. This could be extended with gRPC (googleapis) error details such as field violation details, debug info, localized messages, retry info, etc. but I think we'll use our own protos instead.

Considering that `WithDetails` is currently only used in two places in master (and incorrectly by the looks of it) I think it may be better to completely remove the `Details` funcs from the `errors` package until we can actually use it correctly.

johanstokking: How would that work? These can come from any valid API implementation, right? I think we need proper type assertion.

My comment is mostly about directly type casting to `error`, which I don't think is going to work, but not sure.

htdvisser:
blocked on #967

rvolosatovs:
The build fails since the changes from #995 are not present, otherwise this is ready for review. Once #995 is merged, I will rebase to fix CI.

rvolosatovs:
unable to merge without @johanstokking approval. @htdvisser can you merge this?

PR #962, opened 2019-07-09
Fix metrics prefix and gRPC labels

htdvisser:
#### Summary

This is a quickfix for incorrect metrics prefix and gRPC labels (it used `server_addresss` for client addresses.

#### Changes

- Use `ttn_lw` metrics prefix instead of `ttn`
- Use `remote_address` for gRPC connection metrics labels

#### Notes for Reviewers

~We may want to get rid of the `remote_address` label for `server_conns_{opened,closed}_total` to reduce the number of series that we'll export.~

#### Release Notes

- Changed the prefix of Prometheus metrics from `ttn_` to `ttn_lw_`
- Removed the address label from Prometheus metric `grpc_server_conns_{opened,closed}_total`
- Renamed the label `server_address` of Prometheus metrics `grpc_client_conns_{opened,closed}_total` to `remote_address`

### DIFF
--- a/pkg/metrics/grpc.go
+++ b/pkg/metrics/grpc.go
@@ -19,7 +19,7 @@ import (
 	"net"
 	"runtime/pprof"
 
-	"github.com/grpc-ecosystem/go-grpc-prometheus"
+	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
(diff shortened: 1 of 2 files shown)


htdvisser:
@johanstokking please let me know if you agree with removing `remote_address` from `server_conns_{opened,closed}_total`, then I'll add that to this PR.

Issue #911, opened 2019-07-02
Flaky NS test


johanstokking:
#### Summary

Flaky NS test

#### Steps to Reproduce

Non-deterministic

#### What do you see now?

[flaky-ns-test.txt](https://github.com/TheThingsNetwork/lorawan-stack/files/3350742/flaky-ns-test.txt)

#### What do you want to see instead?

No errors

#### Environment

Travis CI

#### How do you propose to implement this?

No idea. Doesn't seem to be related to flow testing timing though.

#### Can you do this yourself and submit a Pull Request?

No

rvolosatovs:
@johanstokking the log you added seems to be trimmed at
```
  WARN grpc: addrConn.createTransport failed to connect to {in-process 0  <nil>}. Err :connection error: desc = "transport: Error while dialing context deadline exceeded". Reconnecting... namespace=grpc test_name=TestFlow/Redis_registry/Redis_downlink_task_queue/Class_A/OTAA/EU868/1.0.2
  WARN grpc: addrConn.createTransport failed to connect to {in-proces
```

Overall, it looks like the issue is again with the Travis being slow and exceeding the timeout.

johanstokking:
Yes it just ended there

johanstokking:
Seeing this more often than not on my PRs so please fix

rvolosatovs:
https://github.com/TheThingsNetwork/lorawan-stack/pull/912/commits/46537a32c6a07ef9f9734d6af811c958452c5176 should fix it once and for all, if not I'll investigate if there is something else causing this

johanstokking:
[log.txt](https://github.com/TheThingsNetwork/lorawan-stack/files/3529780/log.txt)

rvolosatovs:
Here we go again...
```
--- FAIL: TestFlow (0.00s)

    --- FAIL: TestFlow/Redis_registry/Redis_downlink_task_queue (2.34s)

        --- FAIL: TestFlow/Redis_registry/Redis_downlink_task_queue/Class_A/OTAA/EU868/1.0.2 (2.34s)

            networkserver_flow_test.go:153: Cluster.GetPeer called

            networkserver_flow_test.go:153: ApplicationAccess.ListRights called

            networkserver_flow_test.go:159: Cluster.GetPeer called

            networkserver_flow_test.go:159: ApplicationAccess.ListRights called

            networkserver_flow_test.go:185: Device created

[...]

PR #969, opened 2019-07-09
Fix flaky NS flow test

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/911 (again)

#### Changes

- Increase the deduplication/cooldown windows and time between duplicate uplink transmissions to ensure flakiness-free test experience


rvolosatovs:
A solution like this will only partially fix the issue.
The biggest issue here is that first call to `HandleUplink` blocks and has to happen in a goroutine *and* we need to the second (non-blocking) call to `HandleUplink` to happen after the first call reached the server and got to a specific point in execution(started deduplication)

johanstokking:
I looked into mocking time as well in GS tests and this is indeed less trivial than it looks, and may create a different scenario than what we want to test that leads to two classes of errors we don't want; failures that only happen with mock time, and failures that don't happen with mock time but with real time but don't get tested.

The testing issues that we're seeing are mostly to do with race conditions because in testing runs faster (i.e. very short timeouts) than the code executes in slow environments (i.e. CI).

Issue #951, opened 2019-07-06
One webhook overwrites another settings?


ecities:
#### Summary

I',m trying to create on the same instance of lorawan-stack:

- 1'st application + webhook for production environment (standard data exchange)
- 2'd application + webhook for firmware update of microcontroller device
- the same end-device belongs to both applications (the same name/eui - other app eui/keys)

After performing full initialization of one application it looks it overwrite access rights for previous webhook (it reports no access rights - status 403 on server console) 

#### Steps to Reproduce

Create ap2 for firmware upgrade and webhook for it:

```
ttn-lw-cli end-devices create ap2 dv1 \
  --dev-eui 00217E00000003FF \
  --app-eui 800000000000008C \
  --frequency-plan-id EU_863_870 \
  --root-keys.app-key.key 852BAEC23EAE7964AF27C325F4C23C9A \
  --lorawan-version 1.0.2 \
  --lorawan-phy-version 1.0.2-b
{
  "ids": {
    "device_id": "dv1",
    "application_ids": {
      "application_id": "ap2"
    },
    "dev_eui": "00217E00000003FF",
    "join_eui": "800000000000008C"
  },
  "created_at": "2019-06-26T07:49:08.966Z",
  "updated_at": "2019-06-26T07:49:09.197889667Z",
  "attributes": {
  },
  "network_server_address": "localhost",
  "application_server_address": "localhost",
  "join_server_address": "localhost",
  "lorawan_version": "1.0.2",
  "lorawan_phy_version": "1.0.2-b",
  "frequency_plan_id": "EU_863_870",
  "supports_join": true,
  "root_keys": {
    "app_key": {
      "key": "852BAEC23EAE7964AF27C325F4C23C9A"
    }
  }
}
```

```
ttn-lw-cli applications api-keys create \
  --name link \
  --application-id ap2 \
  --right-application-link

  INFO API key ID: LHUIKUKWZ62A2LXUYP4ZFC54423D3A3J36G33JY
  INFO API key value: NNSXS.LHUIKUKWZ62A2LXUYP4ZFC54423D3A3J36G33JY.NGYQQRDUJW36KJC3UP6PZYP2DNUYZTGISDJEXACL2TB5BB5EWQCA
  WARN The API key value will never be shown again
  WARN Make sure to copy it to a safe place
{
  "id": "LHUIKUKWZ62A2LXUYP4ZFC54423D3A3J36G33JY",
[...]

PR #973, opened 2019-07-10
Document context filling and clarify use

johanstokking:
#### Summary

Clean fix for #951 and ensure that it doesn't happen again

#### Changes

- Document that `FillContext()` should only be used for request contexts
- Practice what we preach in GS and AS testing
- Harmonize GS MQTT testing with AS
- Minor cleanups of unused application context filling in AS

### DIFF
--- a/pkg/applicationserver/io/mock/server.go
+++ b/pkg/applicationserver/io/mock/server.go
@@ -20,11 +20,13 @@ import (
 
 	"go.thethings.network/lorawan-stack/pkg/applicationserver/io"
 	"go.thethings.network/lorawan-stack/pkg/auth/rights"
+	"go.thethings.network/lorawan-stack/pkg/component"
 	"go.thethings.network/lorawan-stack/pkg/ttnpb"
 	"go.thethings.network/lorawan-stack/pkg/unique"
 )
 
 type server struct {
+	*component.Component
 	subscriptionsCh chan *io.Subscription
 	downlinkQueueMu sync.RWMutex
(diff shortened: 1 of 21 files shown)


pkg/applicationserver/io/mock/server.go
```diff
@@ -38,21 +40,17 @@ type Server interface {
 }
 
 // NewServer instantiates a new Server.
-func NewServer() Server {
+func NewServer(c *component.Component) Server {
```
adriansmares: Update the call sites for this in the GS/AS tests as well.

johanstokking:
OK this escalated completely, please see additional diffs. This testing aligns with AS.

References #42

pkg/gatewayserver/io/grpc/grpc.go
```diff
@@ -47,14 +47,14 @@ func (s *impl) LinkGateway(link ttnpb.GtwGs_LinkGatewayServer) (err error) {
...
+	if err != nil {
 		return
 	}
-	if err = rights.RequireGateway(ctx, ids, ttnpb.RIGHT_GATEWAY_LINK); err != nil {
+	if err = ids.ValidateContext(ctx); err != nil {
 		return
 	}
-	ctx, ids, err = s.server.FillGatewayContext(ctx, ids)
-	if err != nil {
+	if err = rights.RequireGateway(ctx, ids, ttnpb.RIGHT_GATEWAY_LINK); err != nil {
```
johanstokking: This actually resolves a never-tested panic; we were requiring gateway rights before filling the context (with the rights fetcher). With this PR we test this correctly.

PR #974, opened 2019-07-10
Fix error handling of event streams

### LINKED ISSUE #932: Console does not show error if events stream gets refused
#### Summary

During today's demo we found that the console doesn't show any error if the events stream gets refused (for the user not having sufficient rights).

#### Steps to Reproduce

1. Assume application `foo`
2. Give non-admin user `bar` `right_application_info` on application `foo`
3. As user `bar` look at the events widget on the overview page of application `foo`

#### What do you see now?

Nothing in the UI, but the development tools do show a 403 for the events call.

Ideally events flowing of course, which will be taken care of by #869, but if a user doesn't have sufficient rights to see events, I'd like to see an error in the UI.

Show an error if the event stream 403's

Nope, no idea where to start. Assigning to @bafonins, who can reassign if he things someone else should do this.

kschiffer:
#### Summary
Closes #932 and a general issue on event streams that caused errors for opening streams to be swallowed.

#### Changes
- Modify stream logic in JS SDK to await the initial request result of the stream endpoint enabling proper catching of errors
- Fix request logic middleware to call `done()` when the stream closes
- Pass stream errors to the `<EventSubscription />` via the error selector
- Add display logic for event stream errors
- Extend error proptype to accept `code` and `grpc_code` property
- Some related refactoring of `propTypes` and `defaultProps` into class statics

#### Notes for Reviewers
I think it's better to use async/await to the extent possible in `sdk/js/src/api/stream.js`, so we can await the initial request and react on errors, before returning the channel. The old implementation failed to catch/handle the error that was thrown in these cases. 

#### Release Notes
- Fixed an issue that resulted in some event errors not being shown in the console

### DIFF
--- a/pkg/webui/console/store/middleware/logics/events.js
+++ b/pkg/webui/console/store/middleware/logics/events.js
@@ -19,6 +19,7 @@ import {
   createStartEventsStreamActionType,
   createStopEventsStreamActionType,
   createStartEventsStreamFailureActionType,
+  createGetEventMessageFailureActionType,
   getEventMessageSuccess,
   getEventMessageFailure,
   startEventsStreamFailure,
@@ -43,6 +44,7 @@ const createEventsConnectLogics = function (
   const START_EVENTS = createStartEventsStreamActionType(reducerName)
   const START_EVENTS_FAILURE = createStartEventsStreamFailureActionType(reducerName)
   const STOP_EVENTS = createStopEventsStreamActionType(reducerName)
+  const GET_EVENT_MESSAGE_FAILURE = createGetEventMessageFailureActionType(reducerName)
   const startEventsSuccess = startEventsStreamSuccess(reducerName)
   const startEventsFailure = startEventsStreamFailure(reducerName)
   const stopEvents = stopEventsStream(reducerName)
@@ -55,7 +57,11 @@ const createEventsConnectLogics = function (
   return [
     createLogic({
       type: START_EVENTS,
+      cancelType: [ STOP_EVENTS, START_EVENTS_FAILURE, GET_EVENT_MESSAGE_FAILURE ],
       warnTimeout: 0,
+      processOptions: {
+        dispatchMultiple: true,
+      },
       validate ({ getState, action }, allow, reject) {
         const { id } = action
         if (!id) {
@@ -72,26 +78,26 @@ const createEventsConnectLogics = function (
 
         allow(action)
       },
-      async process ({ action }, dispatch, done) {
+      async process ({ getState, action }, dispatch) {
         const { id } = action
 
         try {
           channel = await onEventsStart([ id ])
+
           channel.on('start', () => dispatch(startEventsSuccess(id)))
           channel.on('event', message => dispatch(getEventSuccess(id, message)))
           channel.on('error', error => dispatch(getEventFailure(id, error)))
           channel.on('close', () => dispatch(stopEvents(id)))
         } catch (error) {
-          dispatch(startEventsFailure(error))
-          done()
+          dispatch(startEventsFailure(id, error))
         }
       },
     }),
     createLogic({
-      type: [ STOP_EVENTS, START_EVENTS_FAILURE ],
+      type: [ STOP_EVENTS, START_EVENTS_FAILURE, GET_EVENT_MESSAGE_FAILURE ],
       validate ({ getState, action }, allow, reject) {
         const { id } = action
-        if (!id || !channel) {
+        if (!id) {
           reject()
         }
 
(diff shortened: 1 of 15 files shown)


bafonins:
please assign me when it ready for review

pkg/webui/console/store/middleware/logics/events.js
```diff
@@ -68,31 +67,42 @@ const createEventsConnectLogics = function (
...
           channel.on('start', () => dispatch(startEventsSuccess(id)))
           channel.on('event', message => dispatch(getEventSuccess(id, message)))
-          channel.on('error', error => dispatch(getEventFailure(id, error)))
+          channel.on('error', function (error) {
+            const status = selectEntityEventsStatus(getState(), id)
+            if (status === CONNECTION_STATUS.CONNECTING) {
+              dispatch(startEventsFailure(id, error))
+              done()
+            } else {
+              dispatch(getEventFailure(id, error))
```
bafonins: if we stick to this implementation, should `done()` be called here as well?

kschiffer: Not sure about this. The way I see it, an error during retrieval of an event does not necessarily mean that the stream has closed altogether.

sdk/js/src/api/stream/stream.js
```diff
@@ -96,6 +97,10 @@ export default async function (payload, url) {
           notify(listeners[EVENTS.ERROR], error)
           listeners = null
         })
+    }).catch(function (error) {
```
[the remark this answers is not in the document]
kschiffer: If we don't add this, the error will be swallowed, because it occurs in a promise that runs after the subscription object has been returned.

What other implementation do you suggest?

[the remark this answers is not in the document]
kschiffer: I've tried several implementations here:
- Awaiting the initial request ⟶ makes us lose the start event (which I think is not an issue, is also similar to how [`nes`](https://github.com/hapijs/nes#client-1) works)
- Exposing an `open()` function to the returned subscription object, to be able to open the stream after the handlers have been set ⟶ messes up our SDK API
- Using current implementation with added catch block ⟶ makes `error` event ambiguous (event error, or stream start error)

pkg/webui/console/store/middleware/logics/events.js
bafonins: https://github.com/TheThingsNetwork/lorawan-stack/pull/974/files#diff-1ef843661c3cb6cc93d5f07c76d59dd8R97

kschiffer: Indeed. It should call `done()` then.

sdk/js/src/api/stream/stream.js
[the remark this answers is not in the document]
bafonins: 1. https://gist.github.com/bafonins/1a8d3f7840652fb1e5756c69d3a14df6 - something like this, can add some `async/await` to make it prettier. note, it doesnt really matter where you call the `open()` method (before/after adding listeners). dont really this approach.
2. https://gist.github.com/bafonins/2e60e9f28b5fd13b7c9aecf67348e74f - `resolve/reject` based on `fetch` result and propagate errors to the user 
3. add `.catch` to the `fetch` request and notify listeners on `events.error`, but then we have to clearly separate errors caused by reading chunked data and errors caused by failed `fetch`.

pkg/webui/console/store/middleware/logics/events.js
```diff
@@ -68,31 +67,42 @@ const createEventsConnectLogics = function (
...
 
         try {
           channel = await onEventsStart([ id ])
+
           channel.on('start', () => dispatch(startEventsSuccess(id)))
           channel.on('event', message => dispatch(getEventSuccess(id, message)))
-          channel.on('error', error => dispatch(getEventFailure(id, error)))
+          channel.on('error', function (error) {
+            const status = selectEntityEventsStatus(getState(), id)
+            if (status === CONNECTION_STATUS.CONNECTING) {
```
bafonins: see https://github.com/TheThingsNetwork/lorawan-stack/pull/974/files#r304372739, if you decide to go with option 3, then it would be nice to move this to selectors, something like `selectEventsStreamConnected`, etc.

sdk/js/src/api/stream/stream.js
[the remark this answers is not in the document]
kschiffer: Took inspiration from your approaches to implement it in a way that will allow us to await the request (like my first bullet point) without losing the `start` event.

This allows us to keep the original usage of the stream in the event logic.

kschiffer:
> Also, I noticed some prop-types related errors in the console if passing just regular Error object to the <ErrorMessage/> components

Yeah, this is a separate issue that I would like to solve in a different issue/PR.

PR #976, opened 2019-07-10
Add error code to ErrorDetails

### LINKED ISSUE #669: Add error code to ErrorDetails
#### Summary

Add `code` to `ErrorDetails`

#### Why do we need this?

To inform clients what the code is of errors and their causes, without knowing the TTN errors by namespace and name.

`ErrorDetails` with everything but the code.

The reason that the code isn't there, is that it's part of the `google.rpc.Status`, that's only returned on RPC and HTTP calls.

We don't have `google.rpc.Status` available everywhere (i.e. event error payload) and we don't have the error code of the causes.

Add `code` to `ErrorDetails` and fill out the field.

Can do and can review.

rvolosatovs:
#### Summary

Closes #669 

#### Changes

- Add `Code` to `ErrorDetails` as well as encoding and decoding logic


rvolosatovs:
Blocking #631

pkg/errors/conversion.go
```diff
@@ -103,4 +104,7 @@ func setErrorDetails(err *Error, details ErrorDetails) {
 	if cause := details.Cause(); cause != nil {
 		err.cause, _ = From(cause)
 	}
+	if err.code == 0 {
```
johanstokking: This doesn't seem consistent; checking if the target doesn't have a code set, while for other fields we check whether the source has the fields set

johanstokking:
```suggestion
	if code := details.Code(); code != 0 {
```

pkg/errors/conversion.go
```diff
@@ -103,4 +104,7 @@ func setErrorDetails(err *Error, details ErrorDetails) {
...
+		err.code = details.Code()
```
johanstokking:
```suggestion
		err.code = code
```

pkg/errors/conversion.go
rvolosatovs: That's what I thought as well initially, but that causes the `code` to be overwritten by whatever is in the details, which in turn causes panics and test failures all around.
I think we should just revert https://github.com/TheThingsNetwork/lorawan-stack/pull/976/commits/ffe7547b10695c83fe6f793d814a96c6089abe0f that @htdvisser 	added without my knowledge

htdvisser: I simply moved it from the other file

johanstokking: So what do we do with this?

rvolosatovs: Removed @htdvisser 's commit, please review

pkg/errors/conversion.go
```diff
@@ -49,6 +49,7 @@ func From(err error) (out *Error, ok bool) {
 		return &e, true
 	case ErrorDetails: // Received over an API.
 		var e Error
+		e.code = err.Code()
```
htdvisser: This is in the wrong place. Setting things from `ErrorDetails` into `Error` is done in `setErrorDetails`.

pkg/errors/grpc.go
```diff
@@ -50,6 +50,9 @@ func FromGRPCStatus(status *status.Status) Error {
 		err.details = detailIfaces
 	}
 	if details != nil {
+		if err.code == 0 {
```
htdvisser: This is in the wrong place. Setting things from `ErrorDetails` into `Error` is done in `setErrorDetails`.

rvolosatovs: This should be set only if not already present in `status`.
The behavior you propose it confusing (see https://github.com/TheThingsNetwork/lorawan-stack/pull/976#discussion_r302411445), since `setErrorDetails` then both sets if not already set and sets if present in the details
A possible solution is to just remove the `setErrorDetails` altogether and simply set fields directly
@johanstokking what solution do you propose?

rvolosatovs:
@htdvisser it was removed for a reason as well.
Please propose a solution, which is clear for us and our open-source developers

pkg/errors/grpc.go
htdvisser: The reason it's so confusing is that we're now creating an explicit code field while we previously derived the code from the thing that contained the error (the gRPC/HTTP status code).

The `setErrorDetails` func sets error fields from the error details, and that's exactly what we're doing here, so this code belongs in there.

I'm fine with making the logic the same as the existing behavior of `setErrorDetails`, so "if the ErrorDetails has a non-zero error code, set it in Error". That way the ErrorDetails becomes leading, with fallback to the old behavior.

htdvisser:
Proposal in https://github.com/TheThingsNetwork/lorawan-stack/pull/976#discussion_r302881619

rvolosatovs:
This is the same approach I followed initially in the draft PR.

After applying this diff:
```
diff --git a/pkg/errors/conversion.go b/pkg/errors/conversion.go
index 3873d1336..90be01c47 100644
--- a/pkg/errors/conversion.go
+++ b/pkg/errors/conversion.go
@@ -49,7 +49,6 @@ func From(err error) (out *Error, ok bool) {
                return &e, true
        case ErrorDetails: // Received over an API.
                var e Error
-               e.code = err.Code()
                setErrorDetails(&e, err)
                return &e, true
        case interface{ GRPCStatus() *status.Status }:
@@ -105,4 +104,7 @@ func setErrorDetails(err *Error, details ErrorDetails) {
        if cause := details.Cause(); cause != nil {
                err.cause, _ = From(cause)
        }
+       if code := details.Code(); code != 0 {
+               err.code = code
+       }
 }
diff --git a/pkg/errors/grpc.go b/pkg/errors/grpc.go
index d8a0fde0c..4a7e4e371 100644
--- a/pkg/errors/grpc.go
+++ b/pkg/errors/grpc.go
@@ -50,9 +50,6 @@ func FromGRPCStatus(status *status.Status) Error {
                err.details = detailIfaces
        }
        if details != nil {
-               if err.code == 0 {
-                       err.code = details.Code()
-               }
                setErrorDetails(&err, details)
        }
        return err
```

There are test failures in packages:
- `go.thethings.network/lorawan-stack/pkg/auth/rights`
- `go.thethings.network/lorawan-stack/pkg/oauth`

- `panic` in `go.thethings.network/lorawan-stack/pkg/basicstation/cups`:
```
     | panic: runtime error: invalid memory address or nil pointer dereference [recovered]
     |  panic: runtime error: invalid memory address or nil pointer dereference
     | [signal SIGSEGV: segmentation violation code=0x1 addr=0x70 pc=0x112be1a]
     | goroutine 82 [running]:
     | testing.tRunner.func1(0xc000011700)
     |  /nix/store/zvi4wy0kzh3iz7db7dl7p7dk1rqf7frj-go-1.12.6/share/go/src/testing/testing.go:830 +0x392
     | panic(0x1254000, 0x21872b0)
     |  /nix/store/zvi4wy0kzh3iz7db7dl7p7dk1rqf7frj-go-1.12.6/share/go/src/runtime/panic.go:522 +0x1b5
     | go.thethings.network/lorawan-stack/pkg/basicstation/cups.TestServer.func10(0xc001050740, 0xc00000ab40)
     |  /home/rvolosatovs/src/go.thethings.network/lorawan-stack/pkg/basicstation/cups/server_test.go:258 +0x2a
     | go.thethings.network/lorawan-stack/pkg/basicstation/cups.TestServer.func15(0xc000011700)
     |  /home/rvolosatovs/src/go.thethings.network/lorawan-stack/pkg/basicstation/cups/server_test.go:331 +0x809
     | testing.tRunner(0xc000011700, 0xc001050700)
     |  /nix/store/zvi4wy0kzh3iz7db7dl7p7dk1rqf7frj-go-1.12.6/share/go/src/testing/testing.go:865 +0xc0
     | created by testing.(*T).Run
     |  /nix/store/zvi4wy0kzh3iz7db7dl7p7dk1rqf7frj-go-1.12.6/share/go/src/testing/testing.go:916 +0x35a
```

This PR already took way more time than it should.
Can we discuss the next steps here?
Can the maintainers of mentioned packages assist with test fixing?
Can we merge this as-is and open a technical debt issue?

htdvisser:
Just looked into this. With the following changeset all tests pass for me: https://github.com/TheThingsNetwork/lorawan-stack/compare/feature/error-details-code-alt

rvolosatovs:
Applied the diff from https://github.com/TheThingsNetwork/lorawan-stack/pull/976#issuecomment-510840592 and cherry-picked https://github.com/TheThingsNetwork/lorawan-stack/commit/68ecde475f1bd605feadddbde20570ab60035405

Issue #975, opened 2019-07-10
Avoid unhandled promise rejections in the request logics


kschiffer:
#### Summary
With the way our request promise middleware works at the moment, it will produce unhandled promise rejections, when a request fails and the dispatch is not handled via a try/catch block.

#### Steps to Reproduce
1. Login to the console
2. Initiate an action that will fail (as in produce a `{BASE_ACTION}_FAILURE` action type)
3. Notice unhandled promise rejections in the logs

#### What do you see now?
Unhandled promise rejections for dispatches of request actions, when the dispatch is not wrapped in try/catch and the dispatch results in a failure action.

#### What do you want to see instead?
No unhandled promise rejections happening on failure actions

#### Environment
Console.

#### How do you propose to implement this?
The issue is caused [by this block](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/webui/console/store/middleware/logics/lib.js#L65-L68).
This can be fixed by resolving the promise on request failures as well. As far as the logic is concerned, this is ok because the logic still ran through as expected. We should reject the promise only when the dispatch or action creators itself threw an error.

~~Using this fix, failed request action will not cause the promise returned by the dispatch to reject, which solves the issue. In any case, determining whether an action resulted in success or failure is to be done by examining the error values in the store following the result action of the request.~~

EDIT: Not as straightforward, read my follow up below

#### Can you do this yourself and submit a Pull Request?
Yes.

cc @bafonins

kschiffer:
So after investigating this for a little, I can see the following approaches to fix this:

- Don't reject the promise at all, work with the resolved value instead
  - We would need to replace the `try/catch` blocks around dispatches (mostly in form submits)
  - We could resolve the promise with the result action of the request action, and then probe for the action's `error` prop, which we can assume to be present due to complying with Flux Standard Action
  - See #979 for this approach
- Remove the promise middleware altogether and move the api request logic from the store logic to the form submit handlers
  - We would then need to call the update success actions manually
- Remove the promise middleware and work with `fetching` and `error` store values
  - To enable something like toasts then would be quite fiddly because we'd need to react on the `fetching` value change
  - If we pass update action error messages from the store to the components, they would persist through unmounts. This means we would need to apply actions to clear the error values after unmounting
- Ignore unhandled promise rejections, or handle them via event handlers
  - Not desirable IMO

bafonins:
cant reproduce the issue

kschiffer:
[...]

PR #979, opened 2019-07-10
Fix unhandled promise rejections of request actions

kschiffer:
#### Summary
This is a draft for a possible resolution of #975 as basis for discussion.

#### Changes
- Change redux request logic wrapper to resolve promises also on request failures
- Resolve the request dispatch promise with the result action
- Replace try/catches in submit logics with if statements, checking for the error property of the returned action (Note: I only did some of these, to keep the diff small for now)

#### Notes for Reviewers
This issue turned out to be more complex than I had initially thought. In essence, we cannot reject the promise of our request dispatches, as that would lead to unhandled promise rejections whenever we have a dispatch that isn't wrapped in a try/catch. We definitely don't want to wrap all of our request action dispatches into try/catches, so we need to avoid rejecting the promise altogether.

However, if we do not reject the promise, we cannot work with try/catches anymore when e.g. dispatching an update action on a form submit. Instead, we need to check the resolved value of the promise. My solution works with resolving the promise with the result action of the request. Since we have a clear contract of how these look via FSA. We can safely assume the `error` prop to be present in the result action, whenever a request failed.

### DIFF
--- a/pkg/webui/components/location-form/index.js
+++ b/pkg/webui/components/location-form/index.js
@@ -60,16 +60,16 @@ class LocationForm extends Component {
 
     this.setState({ error: '' })
 
-    try {
-      await onSubmit(values)
+    const res = await onSubmit(values)
+    if (!res.error) {
       resetForm()
       toast({
         title: entityId,
         message: m.updateSuccess,
         type: toast.types.SUCCESS,
       })
-    } catch (error) {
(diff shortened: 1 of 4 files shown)


bafonins:
https://github.com/TheThingsNetwork/lorawan-stack/issues/975#issuecomment-514122045

PR #993, opened 2019-07-12
cli: Add error handling while not specifying usrID in organizations

### LINKED ISSUE #887: CLI panic on listing organizations without passing user ID
#### Summary

CLI panics when listing organizations without specifying a user ID.

#### Steps to Reproduce

1. Login
2. List organizations with `$ ttn-lw-cli organizations ls`

#### What do you see now?

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1c6f159]

goroutine 1 [running]:
go.thethings.network/lorawan-stack/cmd/ttn-lw-cli/commands.glob..func71(0x2b6a040, 0x2bbef90, 0x0, 0x0, 0x0, 0x0)
	/Users/johan/dev/src/lorawan-stack/cmd/ttn-lw-cli/commands/organizations.go:78 +0x1d9
github.com/spf13/cobra.(*Command).execute(0x2b6a040, 0x2bbef90, 0x0, 0x0, 0x2b6a040, 0x2bbef90)
	/Users/johan/go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:762 +0x465
github.com/spf13/cobra.(*Command).ExecuteC(0x2b6c640, 0x0, 0x0, 0x1c814cc)
	/Users/johan/go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:852 +0x2ec
github.com/spf13/cobra.(*Command).Execute(...)
	/Users/johan/go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:800
main.main()
	/Users/johan/dev/src/lorawan-stack/cmd/ttn-lw-cli/main.go:27 +0x32
```

No panic, but maybe a nice error that there's no user ID passed (see below)

#### Environment

Current `master`

`getUserID()` may return `nil` when there's no `user-id` flag set or if there's no single argument.

We either need to fallback to the "current user ID" if nothing is specified, or requiring the user to specify their own user ID on these calls.

If you're in doubt, ask @htdvisser as cli code owner

Will review

furtiman:
#### Summary

Closes #887 

#### Changes

- Added error return to `getUserID` to get rid of null pointer dereference panic when no `usedID`s are specified
- Corrected all the usage of this function in according places

#### Notes for Reviewers

This is the current response to `organizations ls` without usrID
```
➜ go run ./cmd/ttn-lw-cli organizations ls
error:cmd/ttn-lw-cli/commands:no_user_id (no user ID set, please specify one)
exit status 255
```

### DIFF
--- a/cmd/ttn-lw-cli/commands/organizations.go
+++ b/cmd/ttn-lw-cli/commands/organizations.go
@@ -74,8 +74,12 @@ var (
 				return err
 			}
 			limit, page, opt, getTotal := withPagination(cmd.Flags())
+			usrID, err := getUserID(cmd.Flags(), nil)
+			if err != nil {
+				return err
+			}
 			res, err := ttnpb.NewOrganizationRegistryClient(is).List(ctx, &ttnpb.ListOrganizationsRequest{
-				Collaborator: getUserID(cmd.Flags(), nil).OrganizationOrUserIdentifiers(),
+				Collaborator: usrID.OrganizationOrUserIdentifiers(),
 				FieldMask:    types.FieldMask{Paths: paths},
 				Limit:        limit,
 				Page:         page,
@@ -141,7 +145,11 @@ var (
 		Short:   "Create an organization",
 		RunE: asBulk(func(cmd *cobra.Command, args []string) (err error) {
 			orgID := getOrganizationID(cmd.Flags(), args)
-			collaborator := getUserID(cmd.Flags(), nil).OrganizationOrUserIdentifiers()
+			usrID, err := getUserID(cmd.Flags(), nil)
+			if err != nil {
+				return err
+			}
+			collaborator := usrID.OrganizationOrUserIdentifiers()
 			if collaborator == nil {
 				return errNoCollaborator
 			}
--- a/cmd/ttn-lw-cli/commands/organizations_access.go
+++ b/cmd/ttn-lw-cli/commands/organizations_access.go
@@ -85,7 +85,11 @@ var (
 			if orgID == nil {
 				return errNoOrganizationID
 			}
-			collaborator := getUserID(cmd.Flags(), nil).OrganizationOrUserIdentifiers()
+			usrID, err := getUserID(cmd.Flags(), nil)
+			if err != nil {
+				return err
+			}
+			collaborator := usrID.OrganizationOrUserIdentifiers()
 			if collaborator == nil {
 				return errNoCollaborator
 			}
@@ -121,7 +125,11 @@ var (
 			if orgID == nil {
 				return errNoOrganizationID
 			}
-			collaborator := getUserID(cmd.Flags(), nil).OrganizationOrUserIdentifiers()
+			usrID, err := getUserID(cmd.Flags(), nil)
+			if err != nil {
+				return err
+			}
+			collaborator := usrID.OrganizationOrUserIdentifiers()
 			if collaborator == nil {
 				return errNoCollaborator
 			}
--- a/cmd/ttn-lw-cli/commands/users.go
+++ b/cmd/ttn-lw-cli/commands/users.go
@@ -40,22 +40,26 @@ func userIDFlags() *pflag.FlagSet {
 	return flagSet
 }
 
-var errNoUserID = errors.DefineInvalidArgument("no_user_id", "no user ID set")
+var errNoUserID = errors.DefineInvalidArgument("no_user_id", "no user ID set, please specify one")
 
-func getUserID(flagSet *pflag.FlagSet, args []string) *ttnpb.UserIdentifiers {
+func getUserID(flagSet *pflag.FlagSet, args []string) (*ttnpb.UserIdentifiers, error) {
 	var userID string
+	var err error
 	if len(args) > 0 {
 		if len(args) > 1 {
 			logger.Warn("multiple IDs found in arguments, considering only the first")
 		}
 		userID = args[0]
 	} else {
-		userID, _ = flagSet.GetString("user-id")
+		userID, err = flagSet.GetString("user-id")
+		if err != nil {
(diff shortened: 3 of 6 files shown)


cmd/ttn-lw-cli/commands/users.go
```diff
@@ -40,22 +40,26 @@ func userIDFlags() *pflag.FlagSet {
...
 	var userID string
+	var err error
 	if len(args) > 0 {
 		if len(args) > 1 {
 			logger.Warn("multiple IDs found in arguments, considering only the first")
 		}
 		userID = args[0]
 	} else {
-		userID, _ = flagSet.GetString("user-id")
+		userID, err = flagSet.GetString("user-id")
```
htdvisser: No need to do this. It only returns `err` if the flag doesn't exist (which it always does) or if it can't be converted to string (which it already is).

cmd/ttn-lw-cli/commands/users.go
```diff
@@ -40,22 +40,26 @@ func userIDFlags() *pflag.FlagSet {
 	return flagSet
 }
 
-var errNoUserID = errors.DefineInvalidArgument("no_user_id", "no user ID set")
+var errNoUserID = errors.DefineInvalidArgument("no_user_id", "no user ID set, please specify one")
 
-func getUserID(flagSet *pflag.FlagSet, args []string) *ttnpb.UserIdentifiers {
+func getUserID(flagSet *pflag.FlagSet, args []string) (*ttnpb.UserIdentifiers, error) {
```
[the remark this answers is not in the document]
htdvisser: And if we decide to add an error here, we should do it in aoo other places too.

cmd/ttn-lw-cli/commands/organizations.go
```diff
@@ -74,8 +74,12 @@ var (
 				return err
 			}
 			limit, page, opt, getTotal := withPagination(cmd.Flags())
+			usrID, err := getUserID(cmd.Flags(), nil)
```
htdvisser: Ok, but a simple `nil` check would be sufficient, like in the other places. See also my comment in `getUserID` itself.

cmd/ttn-lw-cli/commands/organizations.go
```diff
@@ -141,7 +145,11 @@ var (
 		Short:   "Create an organization",
 		RunE: asBulk(func(cmd *cobra.Command, args []string) (err error) {
 			orgID := getOrganizationID(cmd.Flags(), args)
-			collaborator := getUserID(cmd.Flags(), nil).OrganizationOrUserIdentifiers()
+			usrID, err := getUserID(cmd.Flags(), nil)
+			if err != nil {
+				return err
+			}
+			collaborator := usrID.OrganizationOrUserIdentifiers()
 			if collaborator == nil {
```
htdvisser: Can remove this check and just use `usrID.OrganizationOrUserIdentifiers()` directly like you do above in list.

cmd/ttn-lw-cli/commands/organizations.go
```diff
@@ -141,7 +145,11 @@ var (
 		Short:   "Create an organization",
 		RunE: asBulk(func(cmd *cobra.Command, args []string) (err error) {
 			orgID := getOrganizationID(cmd.Flags(), args)
-			collaborator := getUserID(cmd.Flags(), nil).OrganizationOrUserIdentifiers()
+			usrID, err := getUserID(cmd.Flags(), nil)
```
htdvisser: Same

cmd/ttn-lw-cli/commands/organizations_access.go
```diff
@@ -85,7 +85,11 @@ var (
 			if orgID == nil {
 				return errNoOrganizationID
 			}
-			collaborator := getUserID(cmd.Flags(), nil).OrganizationOrUserIdentifiers()
+			usrID, err := getUserID(cmd.Flags(), nil)
```
htdvisser: Same nil check comment. Also below.

cmd/ttn-lw-cli/commands/organizations_access.go
```diff
@@ -85,7 +85,11 @@ var (
...
+			if err != nil {
+				return err
+			}
+			collaborator := usrID.OrganizationOrUserIdentifiers()
```
htdvisser: Can remove this check and just use `usrID.OrganizationOrUserIdentifiers()` directly. Also below.

furtiman:
@htdvisser 

> I think it's cleaner to refactor `OrganizationOrUserIdentifiers()` of `OrganizationIdentifiers` and `UserIdentifiers` to return `nil` if the IDs themselves are `nil`. That way we can keep most of the `if collaborator == nil {` checks and add them where needed.

I see your point. But in this case, the [`Collaborator`](https://github.com/TheThingsNetwork/lorawan-stack/pull/993/files#diff-ff2bc4cb32fa2568e9c777c8a8e36efeR82) field of the `OrganisationRegistryClient.List` request is set to `nil`.
This request is still sent, and then the error should be processed on the server side.

Isn't it better to validate it in-place before sending the request?

So far, the empty `Collaborator` field just doesn't match the required regex pattern upon request
```
--- error:pkg/errors:validation (invalid `user_id`: value does not match regex pattern "^[a-z0-9](?:[-]?[a-z0-9]){2,}$")
    name=UserIdentifiersValidationError
    field=user_id
    reason=value does not match regex pattern "^[a-z0-9](?:[-]?[a-z0-9]){2,}$"
exit status 255
```

htdvisser:
The collaborator in the `List` RPCs is allowed to be `nil`, in which case `List` returns the entities that the caller (instead of the given user or organization) has rights on.

htdvisser:
Replaced by #1010

PR #995, opened 2019-07-12
Fix error detail usage

### LINKED ISSUE #967: Incorrect use of errors.WithDetails
#### Summary

The GS uses `errors.WithDetails` incorrectly, resulting in non-functioning code in #952.

#### What do you see now?

In the GS: `errSchedule.WithDetails(details...)` and `errTxSchedule.WithDetails(errRxDetails...)` where the argument `[]interface{}` contains `error`s.

In the NS: 

```go
for _, d := range ttnGSErr.Details() {
	gtwErr, ok := d.(error)
	///...
```

where `d` never actually is `error`.

The GS shouldn't communicate these errors directly as "details". The details need to be proto messages.

- Change the signature of `WithDetails(details ...interface{}) Error` to `WithDetails(details ...proto.Message) Error`
- Some options:
	- Add a `releated ErrorDetails related_errors` to message `ErrorDetails`. This can be re-used if we need it in other places, but may be a bit confusing, since it can be unclear what is the "cause" and what is "related".
	- Add a generic `RelatedErrors` message to our API that contains `repeated ErrorDetails` and make the GS build such a message and add it as Details. Same problem as the previous.
	- (my preference) Add a specialized `DownlinkScheduleErrorDetails` message to our API and add that (or multiple) as Details. This then becomes a contract between GS and NS.

@rvolosatovs just asked me to create the issue. He'll take it.

rvolosatovs:
#### Summary

Closes #967 

#### Changes

- Add `ErrorDetails.Details`
- Add reserved field in `ErrorDetails` for code(see https://github.com/TheThingsNetwork/lorawan-stack/pull/976)
- Add `ScheduleDownlinkErrorDetails`
- Change type of error details to `proto.Message` 

#### Notes for Reviewers

`Details` field had to be added to the proto in order to not lose the details when `ttnpb.ErrorDetailsToProto` is called.
Note, that GS actually calls `WithDetails` on 2 different levels - per-Rx and per-Path. Took me a while to figure out why GS tests started failing after fixing the detail usage.
Initially this PR was based on #669 , but since that one is taking a while I ensured future rebase will go smooth and added TODOs


pkg/errors/details.go
```diff
@@ -32,31 +34,31 @@ func (e *Error) addDetails(details ...interface{}) {
...
 	e.addDetails(details...)
 	return e
 }
 
 // Details of the error. Usually structs from ttnpb or google.golang.org/genproto/googleapis/rpc/errdetails.
-func (e Error) Details() (details []interface{}) {
+func (e Error) Details() (details []proto.Message) {
 	if e.cause != nil {
-		details = append(details, Details(e.cause))
+		details = append(details, Details(e.cause)...)
```
rvolosatovs: Note the nasty bug fixed here by adding the type checking.

pkg/ttnpb/errors.go
```diff
@@ -50,8 +51,26 @@ func (e errorDetails) Cause() error {
...
+	if len(details) == 0 {
+		return nil
+	}
+
+	msgs := make([]proto.Message, 0, len(details))
+	for _, dAny := range details {
+		var msg types.DynamicAny
+		err := types.UnmarshalAny(dAny, &msg)
+		if err != nil {
+			panic(fmt.Sprintf("Failed to decode error details: %s", err))
```
rvolosatovs: I very much dislike the idea of `panic`ing here, but implemented it this way to be consistent with https://github.com/TheThingsNetwork/lorawan-stack/pull/995/files#diff-dc97abe08c40b3ea0c5eb2da325331a8R41

rvolosatovs:
GS tests fail, since code is lost when encoding to proto. Blocked by #976

pkg/ttnpb/errors.go
htdvisser: I agree. Please open a technical debt issue for that.

pkg/ttnpb/errors.go
```diff
@@ -62,14 +81,21 @@ func ErrorDetailsToProto(e errors.ErrorDetails) *ErrorDetails {
...
-			proto.Cause = ErrorDetailsToProto(ttnErr)
+			pb.Cause = ErrorDetailsToProto(ttnErr)
+		}
+	}
+	for _, d := range e.Details() {
+		dAny, err := types.MarshalAny(d)
+		if err != nil {
+			panic(fmt.Sprintf("Failed to encode error details: %s", err))
 		}
+		pb.Details = append(pb.Details, dAny)
```
htdvisser: This, together with 

https://github.com/TheThingsNetwork/lorawan-stack/blob/2c3d0873430c542c0c24c0bca7dede7edfb766ed/pkg/errors/grpc.go#L104-L108

and

https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/errors/grpc.go#L116

results in details being added twice to error responses. We shouldn't duplicate things like that. We should either communicate additional error detail protos with gRPC's error details or put them in our own ErrorDetails. Not both (unless we're going to separate them on package, so that `google.rpc.*` protos go into gRPC error details, and `ttn.lorawan.v3` protos go into TTN ErrorDetails).

pkg/ttnpb/errors.go
rvolosatovs: https://github.com/TheThingsNetwork/lorawan-stack/issues/1023

Issue #868, opened 2019-06-27
Restructure gateways redux store for pagination


bafonins:
#### Summary

Restructure gateway redux store similar to https://github.com/TheThingsNetwork/lorawan-stack/issues/825

#### Why do we need this?

Cleaner store structure, reuse pagination logic across the app.

#### What is already there? What do you see now?

* `gateway`/`gateways` reducers
* `gateway`/`gateways` action creators
* `gateway`/`gateways` selectors

#### What is missing? What do you want to see?

* All redux primitives for applications moved to `store/{selectors/reducers/actions/middleware}/gateways.js`
* Introduce reusable pagination redux primitives 

#### Environment

browser (console)

#### How do you propose to implement this?

The same way as in https://github.com/TheThingsNetwork/lorawan-stack/issues/825

#### Can you do this yourself and submit a Pull Request?

yes

Issue #933, opened 2019-07-04
Show sub-band channel utilization of gateways


htdvisser:
#### Summary

We would like to see channel utilization in the gateway connection stats.

Originally requested by @ajovertoom

#### Why do we need this?

It's helpful for users to see how much their gateway is used, how much each sub-band is used and how much each channel is used. 

#### What is already there? What do you see now?

We have the `uplink_count` and `downlink_count` in the API, and those fields are filled by the GS. The GS also keeps track of downlink utilization of sub-bands.

#### What is missing? What do you want to see?

To have the (downlink) utilization of a gateway's sub bands exposed through API.

#### How do you propose to implement this?

- [X] Expose utilization from scheduling package (@neoaggelos)
- [X] Add utilization to GatewayConnectionStats API (@neoaggelos)
- [X] Fill utilization fields into GatewayConnectionStats (@neoaggelos)
- [ ] Update view in Console (@bafonins or @kschiffer)

neoaggelos:
For this one, I would suggest we add a new SubBandUtilizations map[string]float32 field. For each subband, we can get the duty cycle from https://github.com/TheThingsNetwork/lorawan-stack/blob/13d2f9d6ccd7ffac56b144507e3a5ac65d464717/pkg/gatewayserver/scheduling/sub_band.go#L107.

johanstokking:
I would go with a `SubBandStatistics` struct which includes a float for duty-cycle, but potentially also other stats in the future, like SNR per DR, number of packets sent/received etc.

What would your proposal be for the `string` key? Sub-bands don't really have a name; they merely have a lower and upper frequency.

neoaggelos:
I was thinking `MinFrequency-MaxFrequency` (e.g. `863000000-865000000`) Since we are in MHz territory though, we could drop the extra zeros to make this more readable, e.g. `863-865`. Ι think that it looks OK with a single decimal too (e.g. `868.7-869.2`).

It makes more sense to go with a struct, in order to make adding more stats in the future.

johanstokking:
> to make this more readable, e.g. `863-865`. Ι think that it looks OK with a single decimal too (e.g. `868.7-869.2`).

This is not the job of the API for many reasons, including that the API doesn't know how the client will visualize this or what the decimal separator of the client is, see comments in #2513.

neoaggelos:
Reopening because we still need the Console part

johanstokking:
@neoaggelos please update issue comment

[...]

PR #901, opened 2019-07-02
Restructure gateways redux store

bafonins:
#### Summary

Referenced #868, #852

#### Changes

- Move gateway action creators to `actions/gateways.js`
- Move gateway selectors to `selectors/gateways.js`
- Move gateway logics to `middleware/logics/gateways.js`

#### Notes for Reviewers

Will create PR with the new gateways store stucture after this one to keep the diff smaller.

PR #998, opened 2019-07-15
Normalize gateways store

bafonins:
#### Summary

Follow-up for https://github.com/TheThingsNetwork/lorawan-stack/pull/901
This PR changes the structure of the gateways store, closes https://github.com/TheThingsNetwork/lorawan-stack/issues/868

#### Changes

- Move `store/reducers/gateway(s).js` to `store/reducers/gateways.js`
- Small adjustments to the `<GatewayStatistics />` component
- Adjust the `error` reducer

#### Notes for Reviewers

1. Removed check for `action.error` being `true` in the `error` reducer, since we cant assume that all failure actions are FSA compliant
2. Didnt add tests for the `statistics` reducer. I expect some changes when implementing https://github.com/TheThingsNetwork/lorawan-stack/issues/933. If is required I can add tests.


pkg/webui/console/store/middleware/logics/gateways.js
```diff
@@ -100,16 +112,33 @@ const startGatewayStatisticsLogic = createLogic({
...
-  validate ({ getState, action }, allow, reject) {
+  processOptions: {
+    dispatchMultiple: true,
+  },
+  transform ({ action }, next) {
+    const { meta = {}} = action
+
+    let transformed = action
+    if (!meta.timeout) {
+      transformed = { ...action, meta: { ...meta, timeout: 5000 }}
```
kschiffer: What's the reasoning behind decorating this here?
(I know this is only refactored, but I'm just wondering this)

kschiffer:
I've noticed another general issue that holds true for all our entities after the entity store refactor:

Try the following:
1. Create a gateway (not necessary if you already created one)
2. Create another gateway
3. Delete the gateway in general settings
4. Be sent back to the list of gateways
5. Try navigating to the other gateway that is still in the list
6. See console crashing

I could imagine this is being caused by the `selectedGateway` store value still holding the deleted gateway.

bafonins:
> Try the following:

    1. Create a gateway (not necessary if you already created one)
    2. Create another gateway
    3. Delete the gateway in general settings
    4. Be sent back to the list of gateways
    5. Try navigating to the other gateway that is still in the list
    6. See console crashing


good catch, here is what causes the issue: https://github.com/TheThingsNetwork/lorawan-stack/pull/998/files#diff-bf295d85376fb58b39a357590281d36eR99

> I've noticed another general issue that holds true for **all our entities after the entity store refactor**

cant reproduce it with applications

pkg/webui/console/store/middleware/logics/gateways.js
bafonins: to make sure the default polling timeout is set

pkg/webui/console/store/reducers/gateways.js
```diff
@@ -12,20 +12,113 @@
...
+    ...gateway,
+  }
+}
+
+const statistics = function (state = defaultState.statistics, { type, payload }) {
+  const { id } = payload
+  const stats = state[id] || {}
+
   switch (type) {
+  case UPDATE_GTW_STATS_SUCCESS:
```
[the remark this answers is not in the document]
bafonins: i agree, however this is a bit trickier and is not really about caching. the main reason for doing it this way is to properly map the update error to a specific gateway to display appropriate status.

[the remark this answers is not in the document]
kschiffer: Can you elaborate a little on this? Why does it have to be mapped, as opposed to having a single (unmapped) value?

pkg/webui/console/store/middleware/logics/gateways.js
kschiffer: Seems like a very verbose approach as opposed to just assuming this value as a default in the `process` hook, but I guess I am missing something here.

kschiffer:
> cant reproduce it with applications

False alert.

pkg/webui/console/store/middleware/logics/gateways.js
bafonins: you are right. it can be simplified

pkg/webui/console/store/reducers/gateways.js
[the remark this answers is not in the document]
bafonins: updated

bafonins:
Refactored the gtw statistics store to store only the latest stats for a single gateway

pkg/webui/console/store/middleware/logics/gateways.js
```diff
@@ -124,47 +143,44 @@ const startGatewayStatisticsLogic = createLogic({
...
-      meta.timeout
+      timeout
     )
 
     cancelled$.subscribe(() => clearInterval(interval))
   },
 })
 
-const updateGatewayStatisticsLogic = createRequestLogic({
+const updateGatewayStatisticsLogic = createLogic({
```
kschiffer: It should be possible to use `createRequestLogic` again then.

PR #1001, opened 2019-07-16
Add API for getting single collaborator and implement in IS

### LINKED ISSUE #721: Extend IS API to fetch single collaborator
#### Summary

Recently we added server-side pagination for collaborators without providing api to fetch a single collaborator.

#### Why do we need this?

Users of the api should be able to fetch/check a single collaborator without fetching all of them. For example, this is needed in the console.

API to fetch a paginated list of collaborators per entity.

API to fetch a single collaborator per entity (e.g. application/gateway/...)

#### Environment

IS

1. Extend `membership_store.go` with the `FindMember` method
2. Extend `{Application,Gateway,User,Organization}Access` rpcs with `Get{Entity}Collaborator` methods.
3. Extend respective protos with the new message:
```
message Get{Entity}CollaboratorRequest {
  {Entity}Identifiers {entity}_ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (validate.rules).message.required = true];
  OrganizationOrUserIdentifiers ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (validate.rules).message.required = true];
}
```

I can do the 2,3 steps, while the first step requires more insights into the is.

htdvisser:
#### Summary

This PR closes #721 by adding the requested API and implementing it in the Identity Server.

#### Changes

- Add API
- Implement in IS

#### Release Notes

- Added API for getting the rights of a single collaborator on (member of) an entity.


api/rights.proto
```diff
@@ -183,6 +183,10 @@ message Collaborator {
   repeated Right rights = 2 [(validate.rules).repeated.items.enum.defined_only = true];
 }
 
+message GetCollaboratorResponse {
+  repeated Right rights = 2;
```
johanstokking: Cleaner to start from `1`

htdvisser: Updated the protos with reserved fields and corresponding comments.

api/rights.proto
```diff
@@ -181,6 +181,21 @@ message APIKeys {
...
   repeated Right rights = 2 [(validate.rules).repeated.items.enum.defined_only = true];
+  reserved 3; // reserved for future EntityIdentifiers entity = 3;
+  reserved 4; // reserved for future State state = 4;
+  reserved 5; // reserved for future google.protobuf.Timestamp created_at = 5;
+  reserved 6; // reserved for future google.protobuf.Timestamp updated_at = 6;
+  // NOTE: Keep compatible with GetCollaboratorResponse.
+}
+
+message GetCollaboratorResponse {
+  reserved 1; // reserved for future OrganizationOrUserIdentifiers ids = 1;
```
bafonins: Why cant we return the complete collaborator object? including the ids?

htdvisser: Sure, we can just copy those from the request, but I'm not a big fan of copying the request into the response.

bafonins: I would prefer having the complete collaborator object being returned

bafonins:
Also, could you clarify the reason to distinguish between `organization` and `user` ?
```
$ curl -H "Authorization: Bearer $access_token" http://localhost:1885/api/v3/applications/123/collaborator/user/admin
{"rights":["RIGHT_ALL"]}                                                                   

$ curl -H "Authorization: Bearer $access_token" http://localhost:1885/api/v3/applications/123/collaborator/organization/admin
{"rights":["RIGHT_ALL"]}
```

api/rights.proto
htdvisser: Best I can do in this PR is add only the `OrganizationOrUserIdentifiers`, otherwise we also have to make API key stuff consistent, and I think that would be out of scope for this PR.

htdvisser:
The fact that http://localhost:1885/api/v3/applications/123/collaborator/user/admin and http://localhost:1885/api/v3/applications/123/collaborator/organization/admin returned the same was actually a bug. Fixed that in a5879792451373629590f0b46fa5bc657eed29c0

kschiffer:
> The fact that http://localhost:1885/api/v3/applications/123/collaborator/user/admin and http://localhost:1885/api/v3/applications/123/collaborator/organization/admin returned the same was actually a (non critical) bug. Fixed that in [a587979](https://github.com/TheThingsNetwork/lorawan-stack/commit/a5879792451373629590f0b46fa5bc657eed29c0).

Why is that? If the ids are unique among each other, why does the type need to be specified?

htdvisser:
Because that is how we designed our API

Issue #138, opened 2019-02-13
Cluster and Service discovery


htdvisser:
**Summary:** 

It would be really nice if we could do some kind of cluster/service discovery. 

**Why do we need this?**

1. It would be really helpful for users if they could get a list of public clusters to choose from.
2. Every cluster may have been deployed in a different way. There may be different domains, ports, TLS/non-TLS connections may or may not be available, etc.

**What is already there? What do you see now?**

- Clusters are currently not registered, but there are ideas/plans to register clusters in the Identity Server: #143
- We currently have default ports, and default configuration.

**What is missing? What do you want to see?**

Assuming that clusters will be registered in the Identity Server, we either need to put service information in there, or we need a different mechanism to discover service information.

**How do you propose to implement this?**

In https://github.com/TheThingsIndustries/lorawan-stack/issues/1131 I suggested to use DNS SRV records for service discovery.

A TTN cluster `eu-west.thethings.network` could be discovered through `SRV` records in DNS:

| **Field** | **Value** |
| --------- | --------- |
| _service._proto.name. | `_ttn-v3-{gs,ns,as,js}-{grpc,http,mqtt}._{tcp,tls}.eu-west.thethings.network.` |
| TTL | `<ttl>` |
| class | `IN` |
| priority | `<priority>` |
| weight | `<weight>` |
| port | `8884`/`1884`/`443`/`80`/`8883`/`1883`/... |
| target | domain |

```
_ttn-v3-gs-grpc._tls.eu-west.thethings.network. <ttl> IN <priority> <weight> 8884 eu-west.thethings.network.
_ttn-v3-ns-grpc._tls.eu-west.thethings.network. <ttl> IN <priority> <weight> 8884 eu-west.thethings.network.
_ttn-v3-as-grpc._tls.eu-west.thethings.network. <ttl> IN <priority> <weight> 8884 eu-west.thethings.network.
_ttn-v3-js-grpc._tls.eu-west.thethings.network. <ttl> IN <priority> <weight> 8884 eu-west.thethings.network.
_ttn-v3-gs-grpc._tcp.eu-west.thethings.network. <ttl> IN <priority> <weight> 1884 eu-west.thethings.network.
_ttn-v3-ns-grpc._tcp.eu-west.thethings.network. <ttl> IN <priority> <weight> 1884 eu-west.thethings.network.
[...]

PR #1004, opened 2019-07-16
Add client for JS interoperability

### LINKED ISSUE #4: Join server lookup and authorization
**Summary:** 

Join server lookup and authorization for external join servers.

To support this inbound, we need #717 

**Why do we need this?**

- A device may use a JS external to the cluster
- AS may need to fetch the `AppSKey` from the JS

Cluster-local JS with cluster authentication

Support for looking up an external JS and using TLS client authentication.

1. If there's a cluster-local JS, try that JS. If that works, great, we're done. If it can't handle the join request (for NS) or it can't find the AppSKey (for AS), it returns an error and go to 2. This way, the cluster-local JS overrides an external JS
2. The NS and AS lookup by DNS the JS and retry the operation. The DNS lookup is `<eui>.joineuis.lora-alliance.org`, but we may make that domain suffix configurable and support a list on which we attempt lookup in order. The `<eui>` is the reverse hex representation with dots on each nibble, i.e. `0.0.0.0.0.0.0.d.e.7.5.d.3.b.0.7.joineuis.lora-alliance.org`

Once we have an IP address, we need to figure out the protocol to use. Proposing a manual DNS lookup;

1. Lookup the SRV records of the record. If there are V3 services (#138), use gRPC. If the record is a CNAME, go to step 2. Otherwise, use Backend Interfaces
2. Try 1 again with the canonical name

### LINKED ISSUE #833: Support Semtech Join Server
#### Summary

Implement join flow through Semtech Join Server

Needs #4

#### Why do we need this?

Support Semtech modem and other devices provisioned on the Semtech Join Server

- Join flow through V3 Join Server (gRPC)
- Interop server and messages for Backend Interfaces
- Existing requirement for join server lookup and authorization (#4)

Implement Semtech Join Server, see [documentation](https://tracknet.bitbucket.io/js-doc/)

This entails;
- Join flow
- Autoprovisioning of devices
- Out of scope for this issue; change of ownership

There's support for Backend Interfaces, although with custom authentication, lookup and paths.

There's also support for websockets, but I only see use for that if we keep a pool of open websocket connections that we reuse to avoid TLS handshake for each activation.

For portability, I would go for the Backend Interfaces approach since we can also use the messages in `pkg/interop`

- Extend `pkg/interop` with a client for the join flow. It would be nice to have options for (custom) authentication, lookup and paths
- We probably need configuration of these options per JoinEUI prefix in NS. This is a level of out-of-band configuration that's definitely not ideal but since there's no standard imposed for authentication and lookup, we have to deal with this
- Incorporate this in Network Server, needs #4 

Will review

rvolosatovs:
#### Summary

Closes #4 
Closes #833 (it's just setting a custom fqdn and path)

#### Changes

- Add `interop.Client`, which handles JS requests via DNS lookup
- Add `interop.LoRaAllianceJoinServerFQDN`, which can be used to generate the FQDN for the DNS lookup using LoRa Alliance DNS service
- Fallback to interop client in NS and AS if cluster-local JS peer is not found

#### Notes for Reviewers

After this is merged I will create a (goldplating+performance) issue to allow e.g. caching of the connections via `interop.Client`

#### Release Notes

- Added support for Join-Server interoperability via Backend Interfaces specification protocol


pkg/interop/client.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+	}
+
+	buf := &bytes.Buffer{}
+	if err := json.NewEncoder(buf).Encode(&JoinReq{
+		NsJsMessageHeader: NsJsMessageHeader{
+			MessageHeader: MessageHeader{
+				ProtocolVersion: "1.0",
+				MessageType:     MessageTypeJoinReq,
+				// TODO: Add TransactionID if necessary
```
rvolosatovs: @johanstokking we don't really need the `TransactionID` and `SenderToken` here, right?
This RPC is answered immediately in the same call as far as I understand.
What about other JS RPCs?

pkg/networkserver/grpc_gsns.go
```diff
@@ -1027,17 +1028,25 @@ func (ns *NetworkServer) handleJoinRequest(ctx context.Context, up *ttnpb.Uplink
...
+	var resp *ttnpb.JoinResponse
+	if js := ns.GetPeer(ctx, ttnpb.PeerInfo_JOIN_SERVER, dev.EndDeviceIdentifiers); js != nil {
+		logger.Debug("Send join-request to cluster-local Join Server")
+		resp, err = ttnpb.NewNsJsClient(js.Conn()).HandleJoin(ctx, req, ns.WithClusterAuth())
+		if err != nil {
+			logger.WithError(err).Warn("Cluster-local Join Server failed to handle join-request")
+			return err
+		}
+	} else {
+		tlsConfig, err := ns.GetTLSConfig(ctx)
```
rvolosatovs: @htdvisser is this how this is meant to be used?

pkg/interop/client.go
```diff
@@ -0,0 +1,136 @@
...
+				// TODO: Add SenderToken if necessary
+			},
+			SenderID:   NetID(netID),
```
rvolosatovs: @johanstokking should the NS's `NetID` be filled in both `SenderdID` and `SenderNSID`?
What are these for? Is it for distinguishing sNS/fNS?

pkg/interop/client.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return nil, ErrJoinReq.WithCause(err)
+	}
+	defer res.Body.Close()
+
+	ans := &JoinAns{}
+	if err := json.NewDecoder(res.Body).Decode(ans); err != nil {
+		return nil, ErrJoinReq.WithCause(err)
+	}
+	if ans.Result.ResultCode != ResultSuccess {
```
rvolosatovs: @johanstokking do we want a more involved result code checking here?
Would be nice if we had primitives of converting these result codes into our errors.
@htdvisser should this be added to `pkg/errors`?

htdvisser: No, this is specific to the Backend Interfaces protocol, and therefore doesn't belong in `pkg/errors`.

pkg/interop/client.go
```diff
@@ -0,0 +1,136 @@
...
+
+const (
+	// loRaAllianceDomain is the domain of LoRa Alliance.
+	loRaAllianceDomain = "lora-alliance.org"
+
+	// loRaAllianceJoinEUIDomain is the LoRa Alliance domain used for JoinEUI resolution.
+	loRaAllianceJoinEUIDomain = "joineuis." + loRaAllianceDomain
+
+	// loRaAllianceNetIDDomain is the LoRa Alliance domain used for NetID resolution.
+	loRaAllianceNetIDDomain = "netids." + loRaAllianceDomain
```
johanstokking: Let's make these configurable on the cluster level with these values as default (no need for `const`)

pkg/interop/client.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+func LoRaAllianceJoinServerFQDN(eui types.EUI64) string {
```
johanstokking: So it's technically not `LoRaAlliance`; the default config is LoRa Alliance

pkg/interop/client.go
johanstokking: We don't need it. I don't see us supporting the asynchronous join flow in the foreseeable future.

pkg/interop/client.go
johanstokking: Yes exactly

pkg/interop/client.go
```diff
@@ -0,0 +1,136 @@
...
+	scheme := "http"
+	var cl http.Client
+	if tlsConfig != nil {
+		scheme = "https"
+		cl.Transport = &http.Transport{
+			TLSClientConfig: tlsConfig,
+		}
+	}
+
+	res, err := cl.Post(fmt.Sprintf("%s://%s", scheme, fqdn), "application/json", buf)
```
johanstokking: We need JS options through config here for path and port

pkg/interop/client.go
```diff
@@ -0,0 +1,136 @@
...
+
+	ans := &JoinAns{}
+	if err := json.NewDecoder(res.Body).Decode(ans); err != nil {
+		return nil, ErrJoinReq.WithCause(err)
+	}
+	if ans.Result.ResultCode != ResultSuccess {
+		return nil, ErrJoinReq
+	}
+
+	if ans.FNwkSIntKey != nil && ans.NwkSKey != nil && !reflect.DeepEqual(ans.FNwkSIntKey, ans.NwkSKey) {
```
johanstokking: Is this spec?

Can't you just require the concerning session keys based on the MAC version?

pkg/networkserver/grpc_gsns.go
johanstokking: TLS config is Join Server specific, please see https://github.com/thethingsnetwork/lorawan-stack/issues/4#issuecomment-504048663

pkg/interop/client.go
johanstokking: See https://github.com/thethingsnetwork/lorawan-stack/issues/4#issuecomment-504048663

pkg/interop/client.go
rvolosatovs: It's a FQDN in LoRa Alliance domain, right?
I don't think I understand what you mean here

pkg/interop/client.go
rvolosatovs: This is currently fully configurable, since e.g. `HandleJoinRequest` takes a `fqdn string`, which is turn can be generated by using `LoRaAllianceJoinServerFQDN`

pkg/interop/client.go
rvolosatovs: So if I refactor the `fqdn string` parameter to `url string` this is fixed, right?

rvolosatovs: So then also `LoRaAllianceJoinServerFQDN` turns into `LoRaAllianceJoinServerURL` and takes e.g. `types.EUI64{0x70, 0xb3, 0xd5, 0x7e, 0xd0, 0x00, 0x00, 0x00}` to `"https://0.0.0.0.0.0.0.d.e.7.5.d.3.b.0.7.joineuis.lora-alliance.org"` ?

rvolosatovs: Note, that nothing prevents you from passing `whatever.foo.bar:1234/asd` as the PRL

johanstokking:
Summary of offline discussion and a few additions:

- add Join Server lookup table by EUI prefix as well as DNS config to `pkg/config.Interop`. Note that if you use `map[string]interop.JoinServerConfig` afaik our `pkg/config` doesn't support that as all config needs to be yml/env/flags compatible. So consider `[]interop.JoinServerConfig` and put the EUI prefix in the struct
- add a `pkg/interop.Client` that does all the interop magic on a high level (i.e. handle join, get `AppSKey`, etc)
- support per Join Server entry:
   - protocol (now only Backend Interfaces 1.0 and 1.1)
   - host name override (fallback to DNS lookup)
   - path override (fallback to standards). These should be paths per component (JS, fNS, sNS, hNS, AS, which are by default `/`, `/sns`, `/sns`, `/hns`, `/` respectively, see 1.1 draft 3)
   - port override (default 443)
   - server CA override (fallback to system chain)
   - client cert and key (fallback to TLS server auth)
   - `Authorization` HTTP header value (fallback to none)
- configure the NS component with a `*interop.Client`
- keep interop stuff in NS to an absolute minimum as `pkg/interop` should implement basically everything. I.e. it should give NS just the join-accept and figure out protocols, sync or async calls, etc
- we don't implement async calls yet
- we don't implement any other protocols yet (i.e. no gRPC, no Semtech websockets)

htdvisser:
Copied from https://github.com/TheThingsNetwork/lorawan-stack/issues/138#issuecomment-507992042

> I thought we previously already concluded that the first phase for the Backend Interfaces Join flow would be a configuration file or repository. I came up with something like this:

```yml
- name: Name of the Join Configuration
  prefixes:
  - 0000000000000000/00
  - 0000000000000000/00
  # in case of DNS lookup:
  dns:
    resolver: 1.1.1.1
    records: CNAME
  # in case of static config:
  static:
    host: hostname.tld
    port: 1234
  protocol: ttn.lorawan-stack.v3 # or backend-interfaces-1.0, backend-interfaces-1.1, ...
  # in case of basic auth:
  basic_auth:
    username: username
    password: password
  # in case of token auth:
  bearer_token: XXX
  # in case of TLS:
  tls_config:
    ca_file: ...
    cert_file: ...
    key_file: ...
```

---

I think it would be a mistake to try cramming all this into the standard env/flag config.

rvolosatovs:
Added https://github.com/TheThingsNetwork/lorawan-stack/pull/1004/commits/f2c3d213bec2b314850aa8c50b378b69d292b9e5 @johanstokking please take a look and confirm it looks more like what you had in mind

@htdvisser we still need a `interop.ClientConfig`(see https://github.com/TheThingsNetwork/lorawan-stack/pull/1004/commits/f2c3d213bec2b314850aa8c50b378b69d292b9e5), right?
The "standard" config is loadable from a file, right? Of course, the intended use case for this config is to be loaded from a file, but I see no harm in allowing this to be set via flags/env as a side effect of using our "standard" config functionality and not rolling out something custom for this.

rvolosatovs:
@htdvisser please do a preliminary review.
Any idea on what can we do about the config?(see my Slack messages)

rvolosatovs:
I did not squash fixup commits to better show my thought process and make reverting things easier, please let me know if that's a problem

pkg/interop/server_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package interop
```
htdvisser: I'm never a big fan of unnecessary changes

pkg/networkserver/grpc_gsns.go
```diff
@@ -1027,17 +1027,20 @@ func (ns *NetworkServer) handleJoinRequest(ctx context.Context, up *ttnpb.Uplink
...
-	resp, err := ttnpb.NewNsJsClient(js.Conn()).HandleJoin(ctx, req, ns.WithClusterAuth())
-	if err != nil {
-		logger.WithError(err).Warn("Join Server failed to handle join-request")
-		return err
+	var resp *ttnpb.JoinResponse
+	if js := ns.GetPeer(ctx, ttnpb.PeerInfo_JOIN_SERVER, dev.EndDeviceIdentifiers); js != nil {
+		logger.Debug("Send join-request to cluster-local Join Server")
+		resp, err = ttnpb.NewNsJsClient(js.Conn()).HandleJoin(ctx, req, ns.WithClusterAuth())
+		if err != nil {
+			logger.WithError(err).Warn("Cluster-local Join Server failed to handle join-request")
```
htdvisser: Still not a big fan of this wording. There is a difference between "failing to handle" and "not accepting" a join request.

pkg/networkserver/grpc_gsns.go
```diff
@@ -1027,17 +1027,20 @@ func (ns *NetworkServer) handleJoinRequest(ctx context.Context, up *ttnpb.Uplink
...
+			return err
+		}
+	} else {
```
htdvisser: This shouldn't be an `else` right? It should check the `err` of the in-cluster HandleJoin, and do an interop join if it's an `errUnknownJoinEUI`

pkg/networkserver/grpc_gsns.go
```diff
@@ -1027,17 +1027,20 @@ func (ns *NetworkServer) handleJoinRequest(ctx context.Context, up *ttnpb.Uplink
...
+		resp, err = ns.interopClient.HandleJoinRequest(ctx, ns.netID, req)
+		if err != nil {
+			logger.WithError(err).Warn("Failed to handle join-request via DNS lookup")
```
htdvisser: Still not a big fan of this wording. There is a difference between "failing to handle" and "not accepting" a join request.

pkg/interop/client.go
```diff
@@ -0,0 +1,328 @@
...
+	// Inspired by https://github.com/golang/example/blob/46695d81d1fae905a270fb7db8a4d11a334562fe/stringutil/reverse.go#L20-L27
+	for i, j := 0, len(o)-1; i < len(o)/2; i, j = i+1, j-1 {
+		r[i+i], r[j+j] = o[j], o[i]
+		r[i+i+1] = '.'
+		r[j+j-1] = '.'
+	}
+	if domain == "" {
+		domain = LoRaAllianceJoinEUIDomain
+	}
+	return fmt.Sprintf("%s.%s", string(r), domain)
```
htdvisser: Something like

```go
fmt.Sprintf(
	"%02x.%02x.%02x.%02x.%02x.%02x.%02x.%02x.%s", 
	eui[7], eui[6], eui[5], eui[4], eui[3], eui[2], eui[1], eui[0],
	domain,
)
```

would also work, and then we'd actually understand what's going on

pkg/interop/messages.go
```diff
@@ -137,7 +137,7 @@ type JsNsMessageHeader struct {
 // AsJsMessageHeader contains the message header for AS to JS messages.
 type AsJsMessageHeader struct {
 	MessageHeader
-	SenderID Buffer
```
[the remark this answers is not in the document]
rvolosatovs: It's incorrect - this has to be an ASCII string according to the spec.
2019-08-01-10:40:34-screenshot

pkg/interop/messages.go
```diff
@@ -160,7 +160,7 @@ type JsAsMessageHeader struct {
 	MessageHeader
 	// SenderID is a JoinEUI.
 	SenderID   EUI64
-	ReceiverID Buffer
```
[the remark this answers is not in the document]
rvolosatovs: IDEM

pkg/interop/server.go
```diff
@@ -110,7 +110,7 @@ func New(ctx context.Context, config config.Interop) (*Server, error) {
 	)
 
 	senderClientCAs := make(map[string][]*x509.Certificate)
-	for senderID, filename := range config.SenderClientCAs {
```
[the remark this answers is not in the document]
rvolosatovs: We need to be consistent and we're renaming the function anyway, so I don't see a problem here

pkg/config/shared.go
```diff
@@ -107,12 +109,60 @@ type HTTP struct {
...
+	Protocol InteropJoinServerProtocol `name:"protocol" description:"Join Server protocol to use"`
+}
+
+// InteropClient represents the client-side interoperability through LoRaWAN Backend Interfaces configuration.
+type InteropClient struct {
+	JoinServers []RemoteJoinServer `name:"join-servers" description:"List of Join Servers configured by JoinEUI."`
+}
+
+type Interop struct {
+	Server InteropServer `name:"server"`
```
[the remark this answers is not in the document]
rvolosatovs: Yes, see https://github.com/TheThingsNetwork/lorawan-stack/pull/1004#issuecomment-516919467

pkg/interop/server_test.go
rvolosatovs: This was necessary during development due to import cycle.
This test is inconsistent with our guidelines, so it has to either be renamed or this change has to be made. Since I already made the change during development I don't see a reason to revert it just to perform the same change later in the future.
I can open a separate PR for this if you want :man_shrugging:

pkg/interop/client.go
rvolosatovs: Your suggestion produces an invalid result, but I made it work with:
```
	return fmt.Sprintf(
		"%01x.%01x.%01x.%01x.%01x.%01x.%01x.%01x.%01x.%01x.%01x.%01x.%01x.%01x.%01x.%01x.%s",
		eui[7]&0xf, eui[7]>>4,
		eui[6]&0xf, eui[6]>>4,
		eui[5]&0xf, eui[5]>>4,
		eui[4]&0xf, eui[4]>>4,
		eui[3]&0xf, eui[3]>>4,
		eui[2]&0xf, eui[2]>>4,
		eui[1]&0xf, eui[1]>>4,
		eui[0]&0xf, eui[0]>>4,
		domain,
	)
```

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -114,6 +118,37 @@ var (
 				return err
 			}
 
+			const (
+				interopClientConfRepoPath  = ""            // TODO: Define the location of interop client configuration repository.
```
rvolosatovs: @htdvisser I would like to add flags for these, but it's not obvious for me how to do that in current implementation.
What is the expected way to add CLI flags to binaries in the current implementation?

pkg/interop/client.go
htdvisser: Ah yeah of course, they were by nibble, not by byte. All the more reason to make this explicit. 👍

rvolosatovs:
For now the paths in config have to be absolute or relative to the running directory. I will make sure it works relative to the directory of the main config

rvolosatovs:
Afero is used to avoid implementing things like caching and for easier extending of this later to work with configs, which are not present on local OS filesystem(e.g. `git` or `HTTP` backed)

pkg/applicationserver/applicationserver.go
```diff
@@ -89,9 +93,25 @@ func New(c *component.Component, conf *Config) (as *ApplicationServer, err error
...
 		return nil, err
 	}
+
+	ctx := log.NewContextWithField(c.Context(), "namespace", "applicationserver")
+
+	var interopCl InteropClient
+	if conf.Interop.ConfigURI != "" {
+		tlsConf, err := c.GetTLSConfig(ctx)
+		if err == nil {
+			log.FromContext(ctx).WithError(err).Warn("Could not get fallback TLS config for interoperability")
```
johanstokking: With `err == nil` and no error handling?

pkg/interop/client_test.go
```diff
@@ -0,0 +1,669 @@
...
+  "MessageType": "AppSKeyAns",
+  "ReceiverToken": "01",
+  "SenderID": "70B3D57ED0000000",
+  "ReceiverID": "test-as",
+  "Result": {
+    "ResultCode": "Success"
+  },
+  "Lifetime": 0,
+  "AppSKey": {
+    "KEKLabel": "as:orne-test",
```
johanstokking: Let's not get this in `master`

pkg/interop/client_test.go
```diff
@@ -0,0 +1,669 @@
...
```
johanstokking: Idem

pkg/interop/server.go
```diff
@@ -95,7 +95,7 @@ type Server struct {
 }
 
 // New builds a new server.
```
johanstokking: Wrong comment

pkg/interop/client.go
```diff
@@ -0,0 +1,486 @@
...
+	// LoRaAllianceJoinEUIDomain is the LoRa Alliance domain used for JoinEUI resolution.
+	LoRaAllianceJoinEUIDomain = "joineuis." + loRaAllianceDomain
+
+	// LoRaAllianceNetIDDomain is the LoRa Alliance domain used for NetID resolution.
+	LoRaAllianceNetIDDomain = "netids." + loRaAllianceDomain
+
+	defaultHTTPSPort = 443
+)
+
+// JoinServerProtocol represents the protocol used for connection to Join-Server by interop client.
```
johanstokking: `Join Server`

pkg/interop/client.go
```diff
@@ -0,0 +1,486 @@
...
+	}
+}
+
+func (p *JoinServerProtocol) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var s string
+	if err := unmarshal(&s); err != nil {
+		return err
+	}
+	switch s {
+	case "LW1.1":
```
johanstokking: So this is basically "LoRaWAN Backend Interfaces 1.1"? `LW1.1` is quite confusing to me as it implies LoRaWAN 1.1

pkg/interop/client.go
```diff
@@ -0,0 +1,486 @@
...
+				prefix:           pre,
+			})
+		}
+	}
+	sort.Slice(jss, func(i, j int) bool {
+		pi, pj := jss[i].prefix, jss[j].prefix
+		if pi.Length != pj.Length {
+			return pi.Length > pj.Length
+		}
+		return pi.EUI64.MarshalNumber() > pj.EUI64.MarshalNumber()
```
johanstokking: I would do this one ascending and the bit prefix descending

pkg/applicationserver/applicationserver.go
rvolosatovs: It's fallback certificate, which is optional.

pkg/interop/client.go
rvolosatovs: Well, it's part of LoRaWAN, right? It just happens to be called `Backend Interfaces` spec, and I doubt there's going to be any other LoRaWAN interoperability spec, hence I chose this naming scheme.
Can you give a better suggestion? `LWBI1.1` ? To me that's even more confusing and IMO the "Backend Interfaces" part is redundant here.

pkg/interop/client.go
rvolosatovs: Why?

rvolosatovs:
@johanstokking standard library `io` is not used - everything is abstracted away by `afero`

htdvisser:
@johanstokking when we had the first idea for the frequency plans repository I already suggested to build it on top of the Afero API, and implement additional Afero (read only) backends for our custom storage options such as cloud blob and HTTP (with index files). I don't remember why we eventualaly chose not to do that.

I'm not against using Afero. On the contrary; I would actually be happy if we could rewrite the `fetch` package to use a combination of Afero and Go-Cloud.

rvolosatovs:
I added https://github.com/TheThingsNetwork/lorawan-stack/pull/1004/commits/e4092266de6e376bb38e06894be77115f9ecfe7f, since it cost me several hours wasted trying to make it work client-side, when it turned out that the configuration produced by this method is not actually usable by clients.

pkg/interop/client.go
johanstokking: It's not redundant; LoRaWAN, LoRaWAN Regional Parameters and LoRaWAN Backend Interfaces carry their own version number. Internally, LoRaWAN is specified as L2 to avoid confusion.

In V3, we clearly separated `MAC` and `PHY`, why confuse with `LW`? Why not just `BI`?

pkg/applicationserver/applicationserver.go
johanstokking: - You're checking `err == nil` while you use `WithError(err)` and print a warning. So that should be `err != nil`
- If `err != nil`, `tlsConf` is undefined. I don't care that the current implementation of `TLSConfig` happens to return `nil` on error

pkg/interop/client.go
```diff
@@ -0,0 +1,506 @@
...
+}
+
+func decodeYAML(r io.Reader, v interface{}) error {
+	dec := yaml.NewDecoder(r)
+	dec.SetStrict(true)
+	return dec.Decode(v)
+}
+
+func NewClient(ctx context.Context, conf ClientConfig, fallbackTLS *tls.Config) (*Client, error) {
+	u, err := url.Parse(conf.ConfigURI)
```
johanstokking: @htdvisser I meant usage of Afero here.

I'm fine with using Afero _in_ `pkg/fetch`, but now we get:

- Configuration of directory or URL through config and usage of `pkg/fetch` in `pkg/frequencyplans` and `pkg/devicerepository` and `pkg/applicationserver/io/web` (for templates)
- A URL  in `pkg/interop` which can start with `file` for files and anything else for URL

I don't like that inconsistency

pkg/interop/client.go
```diff
@@ -0,0 +1,506 @@
...
+type joinServerHTTPClient struct {
+	Client         http.Client
+	NewRequestFunc func(joinEUI types.EUI64, pld interface{}) (*http.Request, error)
+	Protocol       JoinServerProtocol
+}
+
+func (cl joinServerHTTPClient) exchange(joinEUI types.EUI64, req, res interface{}) error {
+	var httpReq *http.Request
+	var err error
+	if cl.NewRequestFunc == nil {
```
johanstokking: In which case is this `nil`?

pkg/interop/interop_util_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+
+	"go.thethings.network/lorawan-stack/pkg/util/test"
+)
+
+var (
+	WorkingDirectory = test.Must(os.Getwd()).(string)
```
johanstokking: Why do we need absolute paths?

pkg/interop/client.go
htdvisser: I agree

pkg/interop/interop_util_test.go
rvolosatovs: They work always, regardless of the working directory

pkg/interop/client.go
rvolosatovs: So what is the proposal?
The deadline for this is tomorrow *and* (as far as I'm concerned) it works.
Can we open a technical debt issue about this?

pkg/interop/client.go
rvolosatovs: None anymore

pkg/interop/interop_util_test.go
johanstokking: Isn't the point of package tests that they are self-containing? This is also the reason for `testdata` and ensuring that test functionality doesn't end up in binaries.

Also, if you `go test pkg/interop`, the working directory + `/testdata` doesn't work, does it?

pkg/interop/client.go
johanstokking: No, because fixing it breaks configuration and we cannot do that, see our compatibility commitment.

The proposal is to be consistent with what we have in;
- `pkg/frequencyplans`
- `pkg/devicerepository`
- `pkg/applicationserver/io/web`

I.e. config strings for folders vs URL, and use `pkg/fetch`.

pkg/interop/interop_util_test.go
rvolosatovs: Client tests write configuration to a temporary directory and use that directory as the repository. `testdata` is not present there.
Of course, we could copy the `testdata` contents to the temporary directory as part of setup, but IMO it's much simpler and cleaner to just use the absolute path to the certificates.
Note, that in the test cases operation with relative paths is tested as well.

rvolosatovs: The tests are self-containing and `go test ./pkg/interop` works as expected

pkg/interop/client.go
johanstokking: This is not addressed still

pkg/interop/config.yaml
```diff
@@ -0,0 +1,14 @@
+join-servers:
```
johanstokking: Shouldn't this be in `testdata`?

pkg/interop/config.yaml
```diff
@@ -0,0 +1,14 @@
...
+        - 70b3d57ed0001000/52
+
+   - file: foo/test-js-2.yaml
+     join-eui:
+        - 70b3d57ed0000000/40
+
+   - file: test-js-3.yaml
+     join-eui:
+        - 70b3d57ed0000000/39
+        - 70b3d83ed0000000/30
```
johanstokking: EOL

pkg/interop/client.go
```diff
@@ -0,0 +1,460 @@
...
+	}
+	confFileBytes, err := fetcher.File(InteropClientConfigurationName)
+	if err != nil {
+		return nil, err
+	}
+
+	var yamlConf struct {
+		JoinServers []struct {
+			File    string              `yaml:"file"`
+			JoinEUI []types.EUI64Prefix `yaml:"join-eui"`
```
johanstokking: `JoinEUIs` and `join-euis` is cleaner imo

pkg/interop/config.yaml
rvolosatovs: shouldn't be checked in at all

johanstokking: At least the file name is better than other files that you didn't check in 😉

rvolosatovs: Good point :rofl:

PR #1007, opened 2019-07-17
Redirect user when accessToken not found

### LINKED ISSUE #426: Redirect user back to login page when an invalid token is detected
**Summary:** 
The console should make sure the user is being sent back to the login screen once his token has expired or invalidated, as well as clearing all user data. This already happens on initial load, but since the console is a single page app, an invalid token can cause trouble at any point after the initial page load.

**Steps to Reproduce:** 
1. Log into the console
2. Open another console tab
3. Logout in one tab
4. In the other tab, try to perform any action that will trigger an API request (eg. list applications)
5. The requested action will result in an error

**What do you see now?**
The console stopping to work properly

The console redirecting the user back to the login page, once it notices that it can no longer obtain a valid token.

**Environment:**
Browser.

We currently use the `with-auth.js` wrapper component to check whether the store has a `user` object. If the object is not present, and the tree is wrapped around the `<withAuth />`-component, he/she will be redirected to the login page via react router. We could make use of this logic by clearing the user store, once the token cannot be retrieved anymore.
One issue here would be that the store will still hold all (potentially sensitive) data associated with the user.
Alternatively, we could simply trigger a hard page reload, which would also result in a redirect to the login page, since the console would not be able to get the user data.
Another question to solve is where to make the token check and where to trigger the corresponding redirect/reload logic.

**Can you do this yourself and submit a Pull Request?**
I can do this, once we agreed on a way to go.

cc @bafonins

pgalic96:
#### Summary

Closes #426 

#### Changes

- ~~Add check in `requestLogic` that makes an api request to the token and in case of an error dispatches a `logoutSuccess` action that clears the user store~~
- Add unauthenticated error handling in `createRequestLogic` wrapper
- Updated `isUnauthenticated` error such that it recognizes `401` error (@kschiffer )
- Add additional check in `events.js`where `createRequestLogic` is not used

### DIFF
--- a/pkg/webui/lib/errors/utils.js
+++ b/pkg/webui/lib/errors/utils.js
@@ -118,6 +118,7 @@ export const isPermissionDeniedError = error => (
  */
 export const isUnauthenticatedError = error => (
   grpcStatusCode(error) === 16
+  || (isBackend(error) && error.details[0].attributes.message === 'code=401, message=You are not logged in') // TODO: Remove this once #1014 is resolved
 )
 
 /**
--- a/pkg/webui/console/store/middleware/logics/events.js
+++ b/pkg/webui/console/store/middleware/logics/events.js
@@ -27,6 +27,8 @@ import {
   stopEventsStream,
 } from '../../actions/events'
 import { createEventsStatusSelector } from '../../selectors/events'
+import { isUnauthenticatedError } from '../../../../lib/errors/utils'
+import user from './user'
 
 /**
  * Creates `redux-logic` logic from processing entity events.
@@ -91,7 +93,11 @@ const createEventsConnectLogics = function (
           channel.on('error', error => dispatch(getEventFailure(id, error)))
           channel.on('close', () => dispatch(stopEvents(id)))
         } catch (error) {
-          dispatch(startEventsFailure(id, error))
+          if (isUnauthenticatedError(error)) {
+            dispatch(user.logoutSuccess())
+          } else {
+            dispatch(startEventsFailure(id, error))
+          }
         }
       },
     }),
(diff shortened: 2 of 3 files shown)


kschiffer:
This is not quite what I had in mind. You could simply in the logic middleware, check whether the `process()` throws an error that evaluates to `401 Unauthorized` and then dispatch the logout action. Your approach would redundantly query the token endpoint on every request.

The error returned from the backend in these cases looks like this:
```json
{  
  "code":2,
  "message":"error:pkg/errors/web:http (HTTP error: code=401, message=You are not logged in)",
  "details":[  
    {  
      "@type":"type.googleapis.com/ttn.lorawan.v3.ErrorDetails",
      "namespace":"pkg/errors/web",
      "name":"http",
      "message_format":"HTTP error: {message}",
      "attributes":{  
        "message":"code=401, message=You are not logged in"
      }
    }
  ]
}
```
It's a bit unfortunate that it has the code: 2 (Unknown), and also no `http_code` attribute.

kschiffer:
See my branch [fix/token-invalid-redirect](https://github.com/TheThingsNetwork/lorawan-stack/tree/fix/token-invalid-redirect) for a PoC.

pkg/webui/lib/errors/utils.js
```diff
@@ -117,7 +117,7 @@ export const isPermissionDeniedError = error => (
  * `false` otherwise.
  */
 export const isUnauthenticatedError = error => (
-  grpcStatusCode(error) === 16
+  grpcStatusCode(error) === 16 || (isBackend(error) && error.details[0].attributes.message === 'code=401, message=You are not logged in')
```
kschiffer:
```suggestion
  grpcStatusCode(error) === 16 
  || (isBackend(error) && error.details[0].attributes.message === 'code=401, message=You are not logged in')
```

PR #1009, opened 2019-07-17
console: Add redirect to login after successful registration

### LINKED ISSUE #930: Automatically redirect to Login page on successful account creation
#### Summary

On most platforms, the default action after creating a new user is to be redirected to the login page automatically.

#### Why do we need this?

Consistent UX when compared to most other online platforms.

- User has to manually click to login. 
- The form for creating a new user still exists with the original values filled in. I can click register again which will obviously give an error.

- Automatically redirect to the login page. Most consoles I've used do this iirc.
- Small wording change: `entity already exists` -> `user already exists` since we are creating users explicitly in this form.

#### Environment

Latest master.

I'll leave this to the experts.

I can test once done.. :)

mjamescompton:
#### Summary

Closes #930 

#### Changes

- After successful Registration dispatch a push redirect to take user to login page.

#### Notes for Reviewers

- I chose to add a dispatch a push rather than replace so that the visiting the registration page was in the browser history even though this meant importing extra functions.


mjamescompton:
Hey, just found an example of the way you are including 'mapDispatchToProps' 

```
@connect(
  null,
  (dispatch, ownProps) => ({
    onClear: () => dispatch(clearApplicationEventsStream(ownProps.appId)),
  }))
```

Would you prefer it in this style? @bafonins

bafonins:
Just noticed:
https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/webui/oauth/views/create-account/index.js#L86-L87 - should return `m.registrationPending`. could you fix this as well while you are on this issue?

kschiffer:
> I think, we should either move the notification to the login page […]

I think this is a good solution.

mjamescompton:
As the user will not see the button text change and info update whilst on the registration page as they are now redirected is it worth keeping it. Shall I remove everything related to `this.state.registered` and `this.state.info` in `pkg/webui/oauth/views/create-account/index.js` @kschiffer @bafonins

kschiffer:
Yeah, this is not necessary anymore then. You can remove it.

pkg/webui/oauth/views/login/index.js
```diff
@@ -85,6 +85,7 @@ export default class OAuth extends React.PureComponent {
       user_id: '',
       password: '',
     }
+    const info = this.props.location.state && this.props.location.state.info ? this.props.location.state.info : ''
```
bafonins: the overall idea is fine, but please refactor this piece

bafonins:
Please edit your commit messages from `console: ...` to `oauth: ...`

PR #1010, opened 2019-07-17
Give OrganizationOrUserIdentifiers() pointer receiver

### LINKED ISSUE #877: Component address comparison should not consider port
#### Summary

Component address comparison should not consider the port.

#### Steps to Reproduce

1. Create a gateway in the CLI. The `gateway_server_address` will be set to `localhost:1884` (i.e. gRPC)
2. Go to the gateway overview page  

#### What do you see now?

The link is not loaded, because `gateway_server_address` does not match the Console's config, i.e. `http://localhost:1885/api/v3`

I want the Console to get the gateway link regardless of port or even secure vs insecure.

I think we shouldn't take the port into account. The Console should just check the domain/IP only to check if the GS is the one of the Console cluster, and then use the Console's configured port (HTTP or HTTPS) to get the link stats.

@htdvisser please confirm

No

### LINKED ISSUE #887: CLI panic on listing organizations without passing user ID

htdvisser:
#### Summary

This Pull Request closes #887 with the changes I proposed in https://github.com/TheThingsNetwork/lorawan-stack/pull/993#pullrequestreview-261725926.

#### Changes

- Gave `OrganizationOrUserIdentifiers()` a pointer receiver
- Made them return `nil` if receiver `nil`
- Updated usage

#### Release Notes

- Fixed crash when running some `ttn-lw-cli organizations` commands without `--user-id` flag.

### DIFF
--- a/cmd/ttn-lw-cli/commands/organizations.go
+++ b/cmd/ttn-lw-cli/commands/organizations.go
@@ -75,7 +75,7 @@ var (
 			}
 			limit, page, opt, getTotal := withPagination(cmd.Flags())
 			res, err := ttnpb.NewOrganizationRegistryClient(is).List(ctx, &ttnpb.ListOrganizationsRequest{
-				Collaborator: getUserID(cmd.Flags(), nil).OrganizationOrUserIdentifiers(),
+				Collaborator: getUserID(cmd.Flags(), nil).GetOrganizationOrUserIdentifiers(),
 				FieldMask:    types.FieldMask{Paths: paths},
 				Limit:        limit,
 				Page:         page,
@@ -141,7 +141,7 @@ var (
 		Short:   "Create an organization",
 		RunE: asBulk(func(cmd *cobra.Command, args []string) (err error) {
 			orgID := getOrganizationID(cmd.Flags(), args)
-			collaborator := getUserID(cmd.Flags(), nil).OrganizationOrUserIdentifiers()
+			collaborator := getUserID(cmd.Flags(), nil).GetOrganizationOrUserIdentifiers()
 			if collaborator == nil {
 				return errNoCollaborator
 			}
--- a/cmd/ttn-lw-cli/commands/organizations_access.go
+++ b/cmd/ttn-lw-cli/commands/organizations_access.go
@@ -85,7 +85,7 @@ var (
 			if orgID == nil {
 				return errNoOrganizationID
 			}
-			collaborator := getUserID(cmd.Flags(), nil).OrganizationOrUserIdentifiers()
+			collaborator := getUserID(cmd.Flags(), nil).GetOrganizationOrUserIdentifiers()
 			if collaborator == nil {
 				return errNoCollaborator
 			}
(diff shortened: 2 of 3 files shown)


htdvisser:
As I wrote yesterday in https://github.com/TheThingsNetwork/lorawan-stack/pull/993#issuecomment-511771675, the `OrganisationOrUserIdentifiers` is allowed to be `nil`.

Therefore `nil` doesn't need any special treatment in the CLI.

furtiman:
Yes, but don’t we also want to indicate to the user _what_ the issue is as well?
I feel that returning an empty array, which in this scenario will always be empty, does not add much value over returning a nil pointer dereference. 

Please tell me if I’m wrong, I am just trying to get the sense of this issue correctly!

htdvisser:
A `nil` is actually the "normal flow". As you can see in https://github.com/TheThingsNetwork/lorawan-stack/blob/master/api/organization.proto#L49-L52

> By default we list all organizations the caller has rights on.

If you are sure that the current user has rights on organizations, but you still get an empty list returned from the list call, then there's something wrong in the Identity Server.

furtiman:
Alright, so it means that in fact it _does_ already take the current user (caller) by default.

In this case I misinterpreted the issue I guess
Thanks for the insights

htdvisser:
@johanstokking that works for me too

Issue #945, opened 2019-07-05
Tiles for supported webhook integations


johanstokking:
#### Summary

Tiles for supported webhook integrations.

#### Why do we need this?

To show recognizable logos for users of upstream platforms and make configuration easier.

#### What is already there? What do you see now?

Manual webhook configuration, i.e. URL, headers, format.

#### What is missing? What do you want to see?

Tiles that provide URL, headers and format.

#### How do you propose to implement this?

1. AS file based configuration of webhook templates and API - @adriansmares 
   - id
   - name, description, logo, info url, documentation url
   - format to use
   - url with path variables (for app ID, device ID, DevEUI)
   - field for user security token that can be used in various ways (e.g. `Authorization` header with optional prefix). Also the name of the user security token (i.e. "Subscription ID", "API Key")
2. AS rpcs to serve webhook templates (part of `pkg/applicationserver/io/web`) - @adriansmares 
3. Console support for fetching templates, rendering tiles and asking user security token - @kschiffer 

We should keep the template ID used in the application webhook so that we can render the logo of the webhook. Apart from that, for editing and deleting the webhook, I would suggest using the existing edit/delete functionality, so that the template is only used for creating the webhook.

#### Can you do this yourself and submit a Pull Request?

Can review

adriansmares:
As mentioned before, the Application Server part of the implementation has been finished in #1011.

The templates are served by the `ApplicationWebhookRegistry` service and follow the structure of the `ApplicationWebhookTemplate` proto. The main takeaways from the templates are the following:
- They represent standalone integrations such as TTN-Mapper, Collos etc.
- They provide the `base_url`, `headers`, `format` and paths for the webhook.
- In order to use a template, one must fill the values of its fields. After the fields have been filled, they are replaced in the `base_url`, `headers` and `format` and the webhook gets created normally.

Fields represents a custom field that needs to be filled by the user in order to use the template.
A field can be an API key, an username or password, or any custom platform specific field (such as region). They have to be replaced inside the webhook using RFC 6570 format.

Taking the following example template:
```yaml
applicationwebhooktemplateidentifiers:
  templateid: example
name: Example
description: Example template
logourl: example.svg
infourl: https://example.com/
documentationurl: https://www.thethingsnetwork.org/docs/applications/example/
baseurl: https://thethingsnetwork.example.com/v3{?env,create,access_token}
headers: {}
format: json
fields:
- id: env
  name: Region
  description: The platform region
[...]

Issue #263, opened 2019-03-08
Set up LoRaWAN Device Repository


htdvisser:
**Summary:** 

We should create a LoRaWAN device repository in TheThingsNetwork's Github organization.

**Why do we need this?**

The device respository will contain descriptions and definitions of well-known LoRaWAN devices, including the default MAC parameters the devices reset to, the payload formats, factory default frequencies (for ABP), and possibly other information used by the NS, AS (and JS).

**What is already there? What do you see now?**

We already have `pkg/devicerepository` implemented using `pkg/fetch`. The repository can be configured with the `--device-repository.url` or `--device-repository.directory` flags. The AS already uses it (if configured) for payload formatters.

**What is missing? What do you want to see?**

There is no default repository on our Github that we can configure.

**How do you propose to implement this?**

Create repository `TheThingsNetwork/lorawan-devices` and use that as default in the Stack.

**What can you do yourself and what do you need help with?**

I think it would be best if @johanstokking would set up the repository, add the README and maybe some examples.

johanstokking:
Note to self; https://github.com/Azure/IoTPlugandPlay/tree/master/DTDL

johanstokking:
Device repository is now setup; https://github.com/TheThingsNetwork/lorawan-devices.

Will file follow-up issues on implementing this in the Console for onboarding devices.

johanstokking:
Closed by https://github.com/TheThingsNetwork/lorawan-devices/pull/11

PR #1011, opened 2019-07-17
Webhooks template support and path variables

### LINKED ISSUE #1005: Custom variable support in Webhook URLs
#### Summary

Currently the final URL of a webhook is static, it does not support custom variables such as `JoinEUI`, `DevID`, `DevEUI` and `DevAddr`. The result should be that a base URL for webhook such as `http://example.com/app1/wh1` can become `http://example.com/app1/wh1/{{.DevEUI}}` and allow the Webhook endpoint to route the uplinks better. This should also be applicable inside the message specific URLs (uplink message, downlink queued etc.)

#### Why do we need this?

This will allow better routing of the Webhooks and also enable us to write certain integrations (such as OpenSensors) only as WebHooks.

Static URL support - `BaseURL` is joined with the topic based on the uplink type, but there are no custom routing fields.

The custom routing fields `{{.AppEUI}}`, `{{.JoinEUI}}`, `{{.DevID}}`, `{{.DevEUI}}`, `{{.DevAddr}}`.

#### Environment

Not applicable.

Replace custom variables in the URL after the base URL + config URL join occurs in `newRequest.`

Yes.

adriansmares:
#### Summary

Closes #1005 
References #945

#### Changes

- Add support for template variables inside the Webhook URLs. Currently supported variables are `appID`, `appEUI`, `joinEUI`, `devID`, `devEUI` and `devAddr`.
- Add Webhook templates support in `pkg/web`. The templates are defined as protos and are based on fields, which are service specific fields that differ from integration from integration.

#### Notes for Reviewers

This PR adds the initial groundwork for #945, by adding the structure of the webhook templates and their storage RPCs. The implementation is similar to the Frequency Plans Repository / Device Repository, using `pkg/fetch`. 

#### Release Notes

- Added support for URL templates inside the Webhook paths ! The currently supported fields are `appID`, `appEUI`, `joinEUI`, `devID`, `devEUI` and `devAddr`. They can be used using [RFC 6570](https://tools.ietf.org/html/rfc6570) syntax, such as `http://www.myapp.com/webhooks{/appID,devID}`, which would get translated to `http://www.myapp.com/webhooks/app1/dev1` for an uplink that is sent by the device `dev1` of application `app1`.


pkg/applicationserver/io/web/templates.go
```diff
@@ -0,0 +1,144 @@
...
+	ts.templateIDsMu.Lock()
+	defer ts.templateIDsMu.Unlock()
+	if ts.templateIDs != nil {
+		return *ts.templateIDs, ts.templateIDsErr
+	}
+	defer func() {
+		ts.templateIDs, ts.templateIDsErr = &ids, err
+	}()
+
+	data, err := ts.Fetcher.File("templates.json")
```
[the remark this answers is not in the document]
johanstokking: Probably also leave caching over to the fetcher. We don't have a caching wrapper now, but that's easy to add

pkg/applicationserver/io/web/templates.go
```diff
@@ -0,0 +1,144 @@
...
+	registeredI, ok := ts.templates.LoadOrStore(ids.TemplateID, &registeredTemplate{ready: make(chan struct{})})
+	registered := registeredI.(*registeredTemplate)
+	if ok {
+		<-registered.ready
+		return registered.t, registered.err
+	}
+	defer func() {
+		registered.t, registered.err = t, err
+		close(registered.ready)
+	}()
```
johanstokking: What is the above for?

pkg/applicationserver/io/web/webhooks.go
```diff
@@ -405,3 +410,24 @@ func (w *webhooks) handleDown(c echo.Context, op func(io.Server, context.Context
...
+		devEUI = up.DevEUI.String()
+	}
+	if up.DevAddr != nil {
+		devAddr = up.DevAddr.String()
+	}
+	googleapi.Expand(url, map[string]string{
+		"appID":   up.ApplicationID,
+		"appEUI":  joinEUI,
+		"joinEUI": joinEUI,
+		"devID":   up.DeviceID,
```
johanstokking: `deviceID`

pkg/applicationserver/io/web/webhooks.go
```diff
@@ -405,3 +410,24 @@ func (w *webhooks) handleDown(c echo.Context, op func(io.Server, context.Context
...
+	}
+	if up.DevEUI != nil {
+		devEUI = up.DevEUI.String()
+	}
+	if up.DevAddr != nil {
+		devAddr = up.DevAddr.String()
+	}
+	googleapi.Expand(url, map[string]string{
+		"appID":   up.ApplicationID,
+		"appEUI":  joinEUI,
```
johanstokking: Good one

pkg/applicationserver/io/web/webhooks.go
```diff
@@ -405,3 +410,24 @@ func (w *webhooks) handleDown(c echo.Context, op func(io.Server, context.Context
...
+		joinEUI = up.JoinEUI.String()
+	}
+	if up.DevEUI != nil {
+		devEUI = up.DevEUI.String()
+	}
+	if up.DevAddr != nil {
+		devAddr = up.DevAddr.String()
+	}
+	googleapi.Expand(url, map[string]string{
+		"appID":   up.ApplicationID,
```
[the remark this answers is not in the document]
johanstokking: Also, I mean

api/applicationserver_web.proto
```diff
@@ -31,6 +31,86 @@ message ApplicationWebhookIdentifiers {
...
+message ApplicationWebhookTemplate {
+  ApplicationWebhookTemplateIdentifiers ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (validate.rules).message.required = true];
+
+  string name = 2 [(validate.rules).string.max_len = 20];
+  string description = 3 [(validate.rules).string.max_len = 100];
+  string logo_url = 4 [(gogoproto.customname) = "LogoURL", (validate.rules).string.uri = true];
+  string info_url = 5 [(gogoproto.customname) = "InfoURL", (validate.rules).string.uri = true];
+  string documentation_url = 6 [(gogoproto.customname) = "DocumentationURL", (validate.rules).string.uri = true];
+
+  string format = 7 [(validate.rules).string.max_len = 10];
```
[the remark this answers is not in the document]
adriansmares: Yes - can be increased for both. Is 20 enough ?

api/applicationserver_web.proto
```diff
@@ -31,6 +31,86 @@ message ApplicationWebhookIdentifiers {
...
+
+  repeated ApplicationWebhookTemplateField fields = 8;
+
+  // The HTTP headers used by the template. Both the key and the value can contain TemplateFields.
```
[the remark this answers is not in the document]
adriansmares: The idea was to point to the fact that the headers can contain the fields, and need to be expanded before creating the actual Webhook.

pkg/applicationserver/io/web/templates.go
adriansmares: Synchronization - allows only one of the `get` calls to update the map. Will remove along with the caching.

api/applicationserver_web.proto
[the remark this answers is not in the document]
johanstokking: Yeah should be. It's driven by us anyway

api/applicationserver_web.proto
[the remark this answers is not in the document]
johanstokking: OK, that doesn't require `TitleCase`

pkg/applicationserver/io/web/templates.go
johanstokking: Yeah this is all the fetcher's responsibility

api/applicationserver_web.proto
```diff
@@ -31,6 +31,86 @@ message ApplicationWebhookIdentifiers {
...
 
+message ApplicationWebhookTemplateIdentifiers {
+  string template_id = 1 [(gogoproto.customname) = "TemplateID", (validate.rules).string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$" , max_len: 36}];
+}
+
+// ApplicationWebhookTemplateField represents a custom field that needs to be filled by the user in order to use the template.
+// A field can be an API key, an username or password, or any custom platform specific field (such as region).
+// The fields are meant to be replaced inside the URLs and headers when the webhook is created.
+message ApplicationWebhookTemplateField {
+  // The ID of the field.
```
htdvisser: We don't write comments for things that are obvious.

api/applicationserver_web.proto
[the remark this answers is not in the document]
htdvisser: 10 is consistent with the `format` in `ApplicationPubSub`. If we change it here, we should also change it there.

Max length is missing from `format` in `ApplicationWebhook`, maybe also make that consistent?

api/applicationserver_web.proto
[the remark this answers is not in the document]
htdvisser: Can we also try to keep the field order roughly consistent with `ApplicationWebhook`?

api/applicationserver_web.proto
```diff
@@ -31,6 +31,86 @@ message ApplicationWebhookIdentifiers {
...
+message GetApplicationWebhookTemplateRequest {
+  ApplicationWebhookTemplateIdentifiers ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (validate.rules).message.required = true];
+  google.protobuf.FieldMask field_mask = 2 [(gogoproto.nullable) = false];
+}
+
+message ListApplicationWebhookTemplatesRequest {
+  google.protobuf.FieldMask field_mask = 1 [(gogoproto.nullable) = false];
+}
+
+service ApplicationWebhookTemplateRegistry {
```
htdvisser: I don't see a reason why these RPCs shouldn't just be part of the `ApplicationWebhookRegistry` service

pkg/applicationserver/io/web/templates.go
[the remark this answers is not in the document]
htdvisser: It may be good to align the packages "get something from repository" to reduce work if/when we decide to refactor all of them.

pkg/applicationserver/io/web/templates.go
htdvisser: I don't think caching this is such a bad thing actually. Otherwise we'll be unmarshaling the same thing all the time.

go.mod
```diff
@@ -67,6 +67,7 @@ require (
 	github.com/jarcoal/httpmock v1.0.4
 	github.com/jaytaylor/html2text v0.0.0-20190408195923-01ec452cbe43
 	github.com/jinzhu/gorm v1.9.10
+	github.com/jtacoma/uritemplates v1.0.0
```
htdvisser: Unused?

pkg/applicationserver/io/web/templates.go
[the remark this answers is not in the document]
adriansmares: I'll move this to yaml, leave the caching as is for consistency with `pkg/frequencyplans` and `pkg/devicerepository`, and create a technical debt issue to refactor all 3 by adding a caching fetcher.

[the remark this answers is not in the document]
htdvisser: AFAIK the HTTP fetcher already has caching.

pkg/applicationserver/io/web/webhooks.go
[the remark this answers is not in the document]
adriansmares: Should we expand all of the names ? (so add `deviceEUI`, `applicationEUI`, `deviceAddress`) ?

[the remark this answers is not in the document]
htdvisser: `devEUI`, `joinEUI`, `appEUI` and `devAddr` are LoRaWAN terms that we usually don't expand.

api/applicationserver_web.proto
adriansmares: I wanted to keep them separate to keep them independent (so each could be disabled without affecting the other). Merged in a4ea0327f.

pkg/applicationserver/io/web/templates.go
johanstokking: If we do the caching, I rather see it being consistent with the (TTL) caching that we have in `pkg/frequencyplans`.

Note that `pkg/devicerepository` doesn't do any caching but is subject to change a bit anyway (#263).

Ideally we have a YAML decoder fetcher on top of the generic data fetcher that does caching, and that we can reuse across FPs, DR and this.

api/applicationserver_web.proto
```diff
@@ -31,6 +31,59 @@ message ApplicationWebhookIdentifiers {
...
+  string description = 3 [(validate.rules).string.max_len = 100];
+  string logo_url = 4 [(gogoproto.customname) = "LogoURL", (validate.rules).string.uri = true];
+  string info_url = 5 [(gogoproto.customname) = "InfoURL", (validate.rules).string.uri = true];
+  string documentation_url = 6 [(gogoproto.customname) = "DocumentationURL", (validate.rules).string.uri = true];
+
+  // The base URL of the template. Can contain template fields, in RFC 6570 format.
+  string base_url = 7 [(gogoproto.customname) = "BaseURL", (validate.rules).string.uri = true];
+
+  // The HTTP headers used by the template. Both the key and the value can contain template fields.
+  map<string,string> headers = 8;
```
htdvisser: Note to self: put validation on this one

api/applicationserver_web.proto
```diff
@@ -31,6 +31,59 @@ message ApplicationWebhookIdentifiers {
...
+
+  string format = 9 [(validate.rules).string.max_len = 20];
```
johanstokking: Shall we use the ID validator here?

api/applicationserver_web.proto
```diff
@@ -42,7 +95,12 @@ message ApplicationWebhook {
   map<string,string> headers = 5;
   // The format to use for the body.
   // Supported values depend on the Application Server configuration.
-  string format = 6;
+  string format = 6 [(validate.rules).string.max_len = 20];
```
johanstokking: Idem

adriansmares:
@rvolosatovs can you please check if the API changes are fine with you as well, such that we could merge this ?

johanstokking:
@htdvisser please review https://github.com/TheThingsNetwork/lorawan-stack/pull/1011/commits/3b7dfd4d388ff0fe51800a7ee5d6061c141ff5aa

pkg/applicationserver/io/web/templates.go
```diff
@@ -0,0 +1,223 @@
...
+	"go.thethings.network/lorawan-stack/pkg/errors"
+	"go.thethings.network/lorawan-stack/pkg/fetch"
+	"go.thethings.network/lorawan-stack/pkg/ttnpb"
+	ttnweb "go.thethings.network/lorawan-stack/pkg/web"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	yamlFetchErrorCache = 1 * time.Minute
+	directoryBaseURL    = "/as/webhook-templates/static"
```
johanstokking: @htdvisser we probably want the API prefix here, although it's static content? Or do we have a standard for this, i.e. a virtual folder in `public`?

pkg/applicationserver/io/web/templates.go
```diff
@@ -0,0 +1,223 @@
...
+		mountDirectory: mountDirectory,
+		baseURL:        baseURL,
+		templates:      make(map[string]queryResult),
+	}, nil
+}
+
+// RegisterRoutes implements ttnweb.Registerer.
+func (ts *TemplateStore) RegisterRoutes(server *ttnweb.Server) {
+	if ts.mountDirectory != "" {
+		server.Static(directoryBaseURL, http.Dir(ts.mountDirectory))
```
johanstokking: Note that there's no prefix here

htdvisser:
I don't quite get this. Why exactly do we need to mount the templates directory in the first place? We already have the API endpoints to list and get the templates, right?

johanstokking:
> I don't quite get this. Why exactly do we need to mount the templates directory in the first place? We already have the API endpoints to list and get the templates, right?

It's for serving the logos in `static`

PR #1016, opened 2019-07-18
Device templates

### LINKED ISSUE #949: Generic device format conversion
#### Summary

Generic device format conversion.

References https://github.com/TheThingsIndustries/lorawan-stack/issues/1672

#### Why do we need this?

To import devices from other sources in various formats.

JS `Provision()` rpc to turn provisioning data into a stream of `EndDevice` templates

The idea of turning a binary file in a template of `EndDevice` that can be created on the concerning components is good, but this endpoint is not specific to Join Server. There can also be a (standardized) device exchange format with NS and AS specific entries.

New gRPC service;
- Returns list of supported formats by the server configuration
- Converts a binary file into a stream of `EndDevice` that can either be created directly and/or be altered by the client (i.e. setting/overriding fields)

Yes

johanstokking:
#### Summary

Closes #949 
References https://github.com/TheThingsIndustries/lorawan-stack/issues/1672

#### Changes

- Add first class component Device Template Converter (DTC) that starts with anything with a device registry (IS, JS, NS, AS) or standalone
- Promote binary conversion to device templates from Join Server's provisioners to `pkg/devicetemplates`
- Add `DeviceTemplateConverter` service that converts binary input to device templates
- Extensive CLI support for working with templates
- Documentation

#### Notes for Reviewers

- Future work is making `EndDeviceTemplate` part of a device repository entry. That'll be part of #263 
- @Sypheos please review documentation

#### Release Notes

- Added Device Templates to create, convert and map templates and assign EUIs to create large amounts of devices
- Deprecated `JsEndDeviceRegistry.Provision()` rpc. Please use `EndDeviceTemplateConverter.Convert()` instead


api/joinserver.proto
```diff
@@ -172,9 +137,9 @@ service JsEndDeviceRegistry {
     };
   };
 
-  // Provision returns end devices that are provisioned using the given vendor-specific data.
-  // The devices are not set in the registry.
-  rpc Provision(ProvisionEndDevicesRequest) returns (stream EndDevice) {
+  // This method is deprecated; use EndDeviceTemplateConverter service instead.
+  // TODO: Remove (https://github.com/TheThingsNetwork/lorawan-stack/issues/999)
+  rpc Provision(EndDeviceIdentifiers) returns (stream EndDevice) {
```
adriansmares: Changing the signature still breaks the API guarantee, right ?

api/end_device_services.proto
```diff
@@ -61,3 +61,20 @@ service EndDeviceRegistry {
     };
   };
 }
+
+service EndDeviceTemplateConverter {
+  // Returns the configured formats to convert from.
+  rpc ListFormats(google.protobuf.Empty) returns (EndDeviceTemplateFormats) {
+    option (google.api.http) = {
+      get: "/edtc/formats"
```
adriansmares: Nitpick: Settle for either `etdc` or `dtc`, but avoid mixing.

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -46,43 +47,51 @@ var errUnknownComponent = errors.DefineInvalidArgument("unknown_component", "unk
...
-				JoinServer        bool
-				Console           bool
-				GCS               bool
+				IdentityServer             bool
+				GatewayServer              bool
+				NetworkServer              bool
+				ApplicationServer          bool
+				JoinServer                 bool
+				Console                    bool
+				GatewayConfigurationServer bool
```
adriansmares: Good catch.

pkg/devicetemplateconverter/devicetemplateconverter.go
```diff
@@ -0,0 +1,81 @@
...
+
+// DeviceTemplateConverter implements the Device Template Converter component.
+//
+// The Device Template Converter exposes the EndDeviceTemplateConverter service.
+type DeviceTemplateConverter struct {
+	*component.Component
+
+	converters map[string]devicetemplates.Converter
+
+	grpc struct {
```
adriansmares: What is the reason for this 1 field struct ? Will it be expanded in the future ?

pkg/devicetemplateconverter/devicetemplateconverter.go
```diff
@@ -0,0 +1,81 @@
...
+		endDeviceTemplateConverter *endDeviceTemplateConverterServer
+	}
+}
+
+var errNotFound = errors.DefineNotFound("converter", "converter `{id}` not found")
+
+// New returns a new *DeviceTemplateConverter.
+func New(c *component.Component, conf *Config) (*DeviceTemplateConverter, error) {
+	converters := make(map[string]devicetemplates.Converter, len(conf.Enabled))
```
adriansmares: Add a context with the component field set, i.e `log.NewContextWithField(c.Context(), "namespace", "dtc")`

api/joinserver.proto
johanstokking: As long as the proto is compatible, i.e. the field tags are of the same types, we're good.

Both `ProvisionEndDevicesRequest` and `EndDeviceIdentifiers` contains

```proto
ApplicationIdentifiers application_ids = 1
```

We need this one for the gRPC gateway, otherwise `google.protobuf.Empty` is cleanest.

I know, this is a hack and we'll remove it in V5

johanstokking: Yeah I made a mistake here; the field tags of `ProvisionEndDevicesRequest` and `EndDeviceIdentifiers` are not all compatible. Will revert partly.

johanstokking: https://github.com/TheThingsNetwork/lorawan-stack/pull/1016/commits/0accda90428bdefd9c456f637a327fb124512e87

api/end_device_services.proto
johanstokking: We consistently use `EndDevice` in our protos/API, including gRPC gateway paths, except within "applications" but those are always end devices.

Also, the `DeviceTemplateConverter` **component** may also expose the `GatewayTemplateConverter` **service** in the future; a gateway is also a device.

pkg/devicetemplateconverter/devicetemplateconverter.go
johanstokking: Yes, may be, also for consistency with other components.

pkg/devicetemplateconverter/devicetemplateconverter.go
johanstokking: https://github.com/TheThingsNetwork/lorawan-stack/pull/1016/commits/e3c02b7827de0539888162511dba908687096c9b

doc/content/concepts/devicetemplates/assigning-euis.md
```diff
@@ -0,0 +1,203 @@
+---
+title: "Assigning EUIs"
+description: ""
+weight: 6
+---
+
+You can use the `assign-euis` command to assign the `JoinEUI` and `DevEUI` to device templates. The `DevEUI` is assigned from the given start number. This personalizes the device templates.
```
adriansmares: Add full command for consistency.

doc/content/concepts/devicetemplates/assigning-euis.md
```diff
@@ -0,0 +1,203 @@
...
+
+You can use `assign-euis` in two different ways:
```
adriansmares: Same as above.

doc/content/concepts/devicetemplates/making.md
```diff
@@ -0,0 +1,73 @@
...
+  "created_at": "0001-01-01T00:00:00Z",
+  "updated_at": "0001-01-01T00:00:00Z",
+  "lorawan_version": "1.0.3",
+  "lorawan_phy_version": "1.0.3-a",
+  "frequency_plan_id": "US_902_928"
+}
+```
+</details>
+
+The `make` command **does not create** the end device. You can, however, easily pipe the output of `make` to create the device. But first, you need to personalize the devices by assigning EUIs since this is a generic device template, see [Assigning EUIs](../euis).
```
adriansmares: Same as above.

doc/content/concepts/devicetemplates/assigning-euis.md
johanstokking: https://github.com/TheThingsNetwork/lorawan-stack/pull/1016/commits/8fcc55545ece18e76b6c26611283ede09eab9afc

doc/content/concepts/devicetemplates/making.md
johanstokking: https://github.com/TheThingsNetwork/lorawan-stack/pull/1016/commits/d532236c620bc350e31ed6dac056eb4943425944

cmd/ttn-lw-cli/commands/end_device_templates.go
```diff
@@ -0,0 +1,436 @@
...
+				return err
+			}
+			var startDevEUI types.EUI64
+			if err := startDevEUI.UnmarshalText([]byte(startDevEUIHex)); err != nil {
+				return err
+			}
+			devEUIInt := binary.BigEndian.Uint64(startDevEUI[:])
+
+			count, _ := cmd.Flags().GetInt("count")
+			for {
```
htdvisser: Ok, so if you pipe 3 templates into this command, with `--count 100` you'd output 100 templates for each input template. Is there any practical use case for that?

cmd/ttn-lw-cli/commands/end_device_templates.go
```diff
@@ -0,0 +1,436 @@
...
+			}
+			res, err := ttnpb.NewEndDeviceTemplateConverterClient(dtc).ListFormats(ctx, ttnpb.Empty)
+			if err != nil {
+				return err
+			}
+			return io.Write(os.Stdout, config.OutputFormat, res)
+		},
+	}
+	endDeviceTemplatesConvertCommand = &cobra.Command{
+		Use:               "convert [format-id]",
```
htdvisser: `ttn-lw-cli end-device template convert` sound like we're converting a template to something. Perhaps rename this to something like `from-data`?

cmd/ttn-lw-cli/commands/end_device_templates.go
```diff
@@ -0,0 +1,436 @@
...
+			}
+		},
+	}
+	endDeviceTemplatesMapCommand = &cobra.Command{
+		Use:   "map [flags]",
+		Short: "Map end device templates",
+		Long: `Map end device templates
+
+This command matches the input template with the mapping file to create new
+templates. The mapping file contains end device templates in JSON format.
```
htdvisser: I would use "the same format as input". The inputDecoder will also be able to decode from yaml/toml/...

cmd/ttn-lw-cli/commands/end_device_templates.go
```diff
@@ -0,0 +1,436 @@
...
+				_, err := inputDecoder.Decode(&input)
+				if err != nil {
+					return err
+				}
+			} else {
+				data, err := getDataBytes("input", cmd.Flags())
+				if err != nil {
+					return err
+				}
+				if err := json.Unmarshal(data, &input); err != nil {
```
htdvisser: Need to use jsonpb here, or better:


```go
reader, err := getDataReader("input", cmd.Flags())
if err != nil {
	return err
}

switch config.InputFormat {
case "json":
	inputDecoder = io.NewJSONDecoder(reader)
default:
	return fmt.Errorf("unknown input format: %s", config.InputFormat)
}
```

We may want to move that second part to a `func getInputDecoder(stdio.Reader) io.Decoder` in `utils.go` for DRY reasons, and also use that func for the main `inputDecoder`, and below.

cmd/ttn-lw-cli/commands/end_device_templates.go
```diff
@@ -0,0 +1,436 @@
...
+					return err
+				}
+			}
+
+			var mapping []ttnpb.EndDeviceTemplate
+			reader, err := getDataReader("mapping", cmd.Flags())
+			if err != nil {
+				return err
+			}
+			mappingDecoder := io.NewJSONDecoder(reader)
```
htdvisser: meaning here

cmd/ttn-lw-cli/commands/end_device_templates.go
```diff
@@ -0,0 +1,436 @@
...
+				}
+			}
+		},
+	}
+	endDeviceTemplatesMapCommand = &cobra.Command{
+		Use:   "map [flags]",
+		Short: "Map end device templates",
+		Long: `Map end device templates
+
+This command matches the input template with the mapping file to create new
```
htdvisser: with `asBulk` this is "input templates" (plural)

.mage/git.go
```diff
@@ -110,6 +110,7 @@ var gitCommitPrefixes = []string{
 	"console",
 	"dev",
 	"doc",
+	"dtc",
```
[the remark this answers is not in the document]
johanstokking: The idea is that DTC can be used for end devices and potentially also for gateways. I don't think that's gonna be worth a dedicated component.

api/cluster.proto
```diff
@@ -39,6 +39,7 @@ message PeerInfo {
     APPLICATION_SERVER = 5;
     JOIN_SERVER = 6;
     CRYPTO_SERVER = 7;
+    DEVICE_TEMPLATE_CONVERTER = 8;
```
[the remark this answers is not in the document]
johanstokking: Yes, see above

api/end_device.proto
```diff
@@ -507,3 +508,25 @@ message SetEndDeviceRequest {
...
   google.protobuf.FieldMask field_mask = 2 [(gogoproto.nullable) = false];
 }
+
+message EndDeviceTemplate {
+  EndDevice end_device = 1 [(gogoproto.nullable) = false, (validate.rules).message.required = true];
+  google.protobuf.FieldMask field_mask = 2 [(gogoproto.nullable) = false];
+  string mapping_key = 3 [(validate.rules).string.max_len = 100];
+}
+
+message EndDeviceTemplateFormat {
```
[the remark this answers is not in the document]
johanstokking: We don't take this as input – should we still add these rules?

api/end_device.proto
```diff
@@ -507,3 +508,25 @@ message SetEndDeviceRequest {
...
+  string name = 1;
+  string description = 2;
+}
+
+message EndDeviceTemplateFormats {
+  map<string, EndDeviceTemplateFormat> formats = 1;
```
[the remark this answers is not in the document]
johanstokking: See above

api/end_device.proto
```diff
@@ -507,3 +508,25 @@ message SetEndDeviceRequest {
...
+}
+
+message ConvertEndDeviceTemplateRequest {
+  // ID of the format.
+  string format_id = 1 [(gogoproto.customname) = "FormatID", (validate.rules).string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$", max_len: 36}];
+  // Data to convert.
+  bytes data = 2;
```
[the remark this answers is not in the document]
johanstokking: I think we should go with the server's maximum configured message length. There's no reason to limit that within the server limits in this message type.

cmd/ttn-lw-cli/commands/end_device_templates.go
```diff
@@ -0,0 +1,436 @@
...
+			}
+
+			if mappingKey, _ := cmd.Flags().GetString("mapping-key"); mappingKey != "" {
+				res.MappingKey = mappingKey
+			}
+			if err := util.SetFields(&res.EndDevice, setEndDeviceFlags); err != nil {
+				return err
+			}
+			res.EndDevice.Attributes = mergeAttributes(res.EndDevice.Attributes, cmd.Flags())
+			res.FieldMask.Paths = ttnpb.BottomLevelFields(paths)
```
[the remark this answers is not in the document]
johanstokking: What is "this" and "it"? Using `ttnpb.BottomLevelFields`?

The goal here is that the templates are clean, don't contain duplicate fields and top level fields while bottom level fields are set. This function does exactly that.

Whether we clean the paths before calling a server is unrelated; users don't see that and we need to support all kinds of useless paths anyway from API clients.

cmd/ttn-lw-cli/commands/end_device_templates.go
johanstokking: Yes, you get each template times `--count`. That's a side effect.

The reason it's implemented this way is that you can use the same command for assigning EUIs to a number of templates, or to a single template.

Still, if you have EU and US templates and you want both times 100, you can do that too. Edge case, but supported, not the intention of this command.

cmd/ttn-lw-cli/commands/end_device_templates.go
```diff
@@ -0,0 +1,436 @@
...
+
+1. Assigning identifiers from a mapping file to device templates matching on
+   mapping key.
+2. Mapping a device profile (i.e. MAC and PHY versions, frequency plan and class
+   B/C support) from a mapping file to many end device templates.
+
+Use the create command to create a mapping file and (optionally) the assign-euis
+command to assign EUIs to map to end device templates.`,
+		PersistentPreRunE: preRun(),
+		RunE: asBulk(func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
johanstokking: It should be; it is many to many mapping.

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -746,18 +660,16 @@ func init() {
 	endDevicesUpdateCommand.Flags().AddFlagSet(setEndDeviceFlags)
 	endDevicesUpdateCommand.Flags().AddFlagSet(attributesFlags())
 	endDevicesCommand.AddCommand(endDevicesUpdateCommand)
-	endDevicesProvisionCommand.Flags().AddFlagSet(applicationIDFlags())
-	endDevicesProvisionCommand.Flags().AddFlagSet(dataFlags())
-	endDevicesProvisionCommand.Flags().String("provisioner-id", "", "provisioner service")
-	endDevicesProvisionCommand.Flags().String("join-eui", "", "(hex)")
-	endDevicesProvisionCommand.Flags().String("start-dev-eui", "", "starting DevEUI to provision (hex)")
-	endDevicesCommand.AddCommand(endDevicesProvisionCommand)
```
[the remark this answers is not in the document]
johanstokking: OK, reverted the whole thing

johanstokking:
@htdvisser changes here: https://github.com/TheThingsNetwork/lorawan-stack/pull/1016/files/2d311a1a9536eff2beac66bfa6b10103a8bee491..481ff2558d042f3455a2a92f6096edccd59065c0

cmd/ttn-lw-cli/commands/end_device_templates.go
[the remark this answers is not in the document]
johanstokking: Yeah will change to reading input and mapping once

api/joinserver.proto
```diff
@@ -172,8 +172,8 @@ service JsEndDeviceRegistry {
     };
   };
 
-  // Provision returns end devices that are provisioned using the given vendor-specific data.
-  // The devices are not set in the registry.
+  // This rpc is deprecated; use EndDeviceTemplateConverter service instead.
+  // TODO: Remove (https://github.com/TheThingsNetwork/lorawan-stack/issues/999)
   rpc Provision(ProvisionEndDevicesRequest) returns (stream EndDevice) {
     option (google.api.http) = {
```
htdvisser: Let's also add the `option deprecated = true;` option.

doc/content/concepts/devicetemplates/_index.md
```diff
@@ -0,0 +1,13 @@
+---
+title: "Device Templates"
+description: ""
+weight: 1
+---
+
+Device templates can be used to quickly create large amounts of end devices with common settings. Templates can be created from existing devices or converted from an input file. You can create end devices directly from templates. See [Creating Templates](./creating), [Converting Templates](./converting) and [Executing Templates](./executing).
```
htdvisser: Use the [`relref` shortcode](https://gohugo.io/content-management/cross-references/) for links, that gives us free ref checking. Also in other doc content.

doc/content/concepts/devicetemplates/_index.md
```diff
@@ -0,0 +1,13 @@
...
+
+Templates can also be be mapped with other templates to combine fields. For example, you can convert root key provisioning data to device templates and map that with a template containing the device profile from a device repository. See [Mapping Templates](./mapping).
+
+For device makers, the tooling supports assigning the LoRaWAN `JoinEUI` and `DevEUI` from a range automatically to create a template file that can be used in mapping or executing devices. See [Assigning EUIs](./assigning-euis).
+
+To work with device templates, use the command-line interface (CLI). See [Getting started with the command-line interface](../../gettingstarted/cli).
```
htdvisser: > Getting started with the command-line interface

We should think about guidelines for using title case

doc/content/concepts/devicetemplates/assigning-euis.md
```diff
@@ -0,0 +1,205 @@
+---
+title: "Assigning EUIs"
+description: ""
+weight: 6
+---
+
+You can use the `device templates assign-euis` command to assign the `JoinEUI` and `DevEUI` to device templates. The `DevEUI` is assigned from the given start number. This personalizes the device templates.
```
htdvisser: Shouldn't we make clear that this is a CLI command? either by prefixing `ttn-lw-cli` or by adding "of the CLI"

doc/content/concepts/devicetemplates/assigning-euis.md
```diff
@@ -0,0 +1,205 @@
...
+1. Pipe a number of device templates to assign EUIs to, for example the result of [Mapping Templates](../mapping) or [Converting Templates](../converting)
+2. Pipe a generic device template, for example the result of [Creating Templates](../creating), and pass the `--count` flag to generate a number of personalized device templates
+
+Once the device templates are personalized with EUIs, you can set additional fields (see [Creating Templates](../creating)) or execute templates (see [Executing Templates](../executing)).
+
+>Note: to make your device compliant with most LoRaWAN networks, use the `JoinEUI` of the LoRaWAN Join Server where the devices will be registered and use a `DevEUI` from an IEEE issued MAC address block that is assigned to your company. If you intend to operate your own LoRaWAN Join Server, use a `JoinEUI` from your IEEE assigned MAC address block as LoRa Alliance requires that for DNS lookup. [Learn how to apply for a IEEE issued MAC address block](https://standards.ieee.org/products-services/index.html).
+
+## Example
+
+This examples illustrates creating a device profile template, assigning 5 `DevEUI`s and creating them in your TTN Stack application.
```
htdvisser: > This examples

doc/content/concepts/devicetemplates/assigning-euis.md
```diff
@@ -0,0 +1,205 @@
...
+
+```bash
+$ ttn-lw-cli device template create \
+  --frequency-plan-id EU_863_870 \
+  --lorawan-version 1.0.3 \
+  --lorawan-phy-version 1.0.3-a \
+  --supports-join > profile.json
+```
+
+Second, assign the EUIs:
```
htdvisser: Mention that the first is the JoinEUI and the second is the start DevEUI.

doc/content/concepts/devicetemplates/assigning-euis.md
```diff
@@ -0,0 +1,205 @@
...
+
+```bash
+$ cat profile.json \
+  | ttn-lw-cli device template assign-euis 70b3d57ed0000000 70b3d57ed0000001 --count 5 > devices.json
+```
+
+<details><summary>Show output</summary>
```
htdvisser: Nice. Does this also work in Hugo? If so, then we should use it more. If not, we should make it work.

doc/content/concepts/devicetemplates/converting.md
```diff
@@ -0,0 +1,72 @@
+---
+title: "Converting Templates"
+description: ""
+weight: 3
+---
+
+You can convert data to device templates using formats supported by TTN Stack. Start with listing the supported formats:
```
htdvisser: > data

Be more clear about what kind of "data". People read from top to bottom, so maybe a little introduction about a typical process would help here.

doc/content/concepts/devicetemplates/converting.md
```diff
@@ -0,0 +1,72 @@
...
+  "formats": {
+    "microchip-atecc608a-mahtn-t": {
+      "name": "Microchip ATECC608A-MAHTN-T Manifest File",
+      "description": "JSON manifest file received through Microchip Purchasing \u0026 Client Services."
+    }
+  }
+}
+```
+
+Given input data, you can use the `device template from-data` command to get the device template using the specified formatter.
```
htdvisser: Same comment as before about `ttn-lw-cli` or "of the CLI"

doc/content/concepts/devicetemplates/assigning-euis.md
htdvisser: Use the official CLI command (`end-devices`), not the alias (`devices`). Also in your other examples.

doc/content/concepts/devicetemplates/creating.md
```diff
@@ -0,0 +1,116 @@
...
+title: "Creating Templates"
+description: ""
+weight: 2
+---
+
+You can create a template from scratch, set fields in an existing template or create a template an existing device.
+
+## Create from scratch
+
+Use the `device template create` command to create a template with some end device fields set:
```
htdvisser: Same comments as before

doc/content/concepts/devicetemplates/creating.md
```diff
@@ -0,0 +1,116 @@
...
+
+The `device template create` command may take an existing template as input. This allows you to set fields in an existing template.
+
+For example, create the base template and save to `base.json`:
+
+```bash
+$ ttn-lw-cli device template create \
+  --lorawan-version 1.0.3 \
+  --lorawan-phy-version 1.0.3-a \
+  --mapping-key test > base.json
```
htdvisser: I think having the mapping key in this particular example is a bit confusing. Maybe leave that for `mapping.md`?

doc/content/concepts/devicetemplates/creating.md
```diff
@@ -0,0 +1,116 @@
...
+    ]
+  },
+  "mapping_key": "test"
+}
+```
+</details>
+
+## Create from existing device
+
+You can use the `device template from-device` command to create a template from an existing device.
```
htdvisser: I'd explain the fields that are (not) taken here instead of below.

pkg/devicetemplateconverter/grpc.go
```diff
@@ -0,0 +1,68 @@
...
+// limitations under the License.
+
+package devicetemplateconverter
+
+import (
+	"bytes"
+	"context"
+
+	"go.thethings.network/lorawan-stack/pkg/errorcontext"
+
```
htdvisser: fmt

cmd/ttn-lw-cli/commands/end_device_templates.go
```diff
@@ -0,0 +1,451 @@
...
+
+			return io.Write(os.Stdout, config.OutputFormat, &res)
+		}),
+	}
+	endDeviceTemplatesFromDeviceCommand = &cobra.Command{
+		Use:   "from-device [flags]",
+		Short: "Create an end device template from an existing device",
+		Long: `Create an end device template from an existing device
+
+	This command takes end devices from stdin.`,
```
htdvisser: I think we should also start documenting what the flags do (they select fields in this case) and which fields are included/excluded by default.

cmd/ttn-lw-cli/commands/end_device_templates.go
```diff
@@ -0,0 +1,451 @@
...
+			return io.Write(os.Stdout, config.OutputFormat, &device)
+		}),
+	}
+	endDeviceTemplatesAssignEUIsCommand = &cobra.Command{
+		Use:     "assign-euis [join-eui] [start-dev-eui] [flags]",
+		Aliases: []string{"euis"},
+		Short:   "Assign JoinEUI and DevEUIs to end device templates",
+		Long: `Assign JoinEUI and DevEUIs to end device templates
+
+This command takes an end device template from stdin.`,
```
htdvisser: Let's document that it either takes multiple templates, or generates `--count` for a single template (even though it technically allows both mutliple and `--count`)

Sypheos:
I think it do exceed the context of **concept** a bit as it explain the whole process, include references and technical insight. Templates must have a few pages here to provide insight and not technical information. This part in the *concept* section should answer the following:
* What is it ?
* What is it for ?
* How does it work ?

The "How do I use it ?" is should be in a guide section.

See it as showcase and on-boarding section.

doc/content/concepts/devicetemplates/assigning-euis.md
johanstokking: Yeah that works. We should take care of template support.

Edge doesn't support it but it will come most probably when they switch to Chromium. See compatibility table here; https://www.w3schools.com/tags/tag_details.asp

doc/content/concepts/devicetemplates/_index.md
johanstokking: Removed capital G. This is not a page title

doc/content/concepts/devicetemplates/assigning-euis.md
Sypheos: Then it should be defined as a shorcode so we can easily use it in other sections of the doc.
https://gohugo.io/content-management/shortcodes/

johanstokking:
Let's be a little more constructive, pragmatic and concrete so we can move forward. This is the first PR that actually documents what is being added, and that is much better than nothing. This not the moment to come up with exhaustive guidelines for what goes where; that's going to be case by case. Some subsections will grow to their own, others may stay together. User experience is king.

> The "How do I use it ?" is should be in a guide section.

Where? What is the structure?

I do think that the workings _fall under_ the concept, but that the concept page contains the high level overview. I added that now:

Screenshot 2019-07-22 at 18 33 13

In my opinion, a "guide" is solving a particular problem and following some steps. That is clearly not what is being documented here; these are strictly examples of commands.

doc/layouts/shortcodes/cli-only.html
```diff
@@ -0,0 +1,4 @@
+<div class="panel panel-primary panel-header">
+  <h3>Command-line interface only</h3>
+  The instructions on this page use the command-line interface (CLI). <a href="{{- ref . "/gettingstarted/cli" -}}">Learn how to get started with the CLI</a>
```
htdvisser: Shouldn't we consistentl use ref or relref everywhere?

johanstokking: They have their purpose; `relref` within sections, `ref` cross sections. That allows for promoting/demoting sections by only updating entry links. I'm not in favor of `../../../../gettingstarted` (ok, bit exaggerated)

johanstokking:
Blocked by #1034

PR #1020, opened 2019-07-19
console: Include page-size constants for tables

### LINKED ISSUE #907: Increase page size of entity tables to 20
#### Summary
The entity tables currently have a page size of 5, which is a bit low. 20 is a better value that will fill a usual browser window (16:9) better.

#### Why do we need this?
Improve UX. Page sizes that are too small are hard to navigate through. Whether 20 is indeed the optimal default needs some progressive insight.

- Functionality to set the page size in our `<Pagination />` component.
- Various default value constants for page sizes, set to 5

A unified default value constant of 20 items per page for all entity tables. (Except for the device table in the application overview)

#### Environment
Console.

Trivial.

Yes.

mjamescompton:
#### Summary

Close #907 

#### Changes

- Include constants file `pkg/webui/console/constants/page-sizes.js` containing two constants `TABLE_SIZE_SMALL` = 5 and `TABLE_SIZE_LARGE` = 20
- Include and Applied constants to tables in the following files 
  - pkg/webui/console/views/application-api-keys-list/index.js
  - pkg/webui/console/views/application-collaborators-list/index.js
  - pkg/webui/console/views/application-overview/index.js
  - pkg/webui/console/views/applications-list/index.js
  - pkg/webui/console/views/device-list/index.js
  - pkg/webui/console/views/gateway-api-keys-list/index.js
  - pkg/webui/console/views/gateway-collaborators-list/index.js
  - pkg/webui/console/views/gateways-list/index.js

### DIFF
--- a/pkg/webui/console/constants/page-sizes.js
+++ b/pkg/webui/console/constants/page-sizes.js
@@ -1,0 +1,18 @@
+// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export default Object.freeze({
+  SMALL: 5,
+  REGULAR: 20,
+})
--- a/pkg/webui/console/views/application-api-keys-list/index.js
+++ b/pkg/webui/console/views/application-api-keys-list/index.js
@@ -21,7 +21,7 @@ import ApiKeysTable from '../../containers/api-keys-table'
 import { getApplicationApiKeysList } from '../../store/actions/applications'
 import sharedMessages from '../../../lib/shared-messages'
 
-const API_KEYS_TABLE_SIZE = 10
+import PAGE_SIZES from '../../constants/page-sizes'
 
 @bind
 export default class ApplicationApiKeys extends React.Component {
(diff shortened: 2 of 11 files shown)


pkg/webui/console/constants/page-sizes.js
```diff
@@ -0,0 +1,18 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export default Object.freeze({
+  TABLE_SIZE_SMALL: 5,
+  TABLE_SIZE_LARGE: 20,
```
kschiffer:
```suggestion
  TABLE_SIZE_REGULAR: 20,
```
We might come up with an actual `LARGE` value later.

pkg/webui/console/views/application-api-keys-list/index.js
```diff
@@ -21,7 +21,7 @@ import ApiKeysTable from '../../containers/api-keys-table'
 import { getApplicationApiKeysList } from '../../store/actions/applications'
 import sharedMessages from '../../../lib/shared-messages'
 
-const API_KEYS_TABLE_SIZE = 10
+import pageSizes from '../../constants/page-sizes'
```
kschiffer:
```suggestion
import PAGE_SIZES from '../../constants/page-sizes'
```

We usually use UPPER_CASE for constant object imports. Having that in mind, it actually makes more sense (and is more consistent) to rename the constant values to allow for a usage like `PAGE_SIZES.REGULAR`

pkg/webui/console/constants/page-sizes.js
kschiffer: See my other comment on renaming these to just `LARGE` and `REGULAR`.

mjamescompton:
> could you do the same to the application integrations table?

I have included the pageSize prop to the application integrations WebhooksTable but there is another issue making the pagination not work. Would you like me to look into it, or is it a separate issue? @bafonins

bafonins:
> Would you like me to look into it, or is it a separate issue? 

The problem is that the `page` and `limit` parameters are not sent with the request when fetching integrations. You can fix this in a separate `quickfix` pr.

Issue #722, opened 2019-05-21
Paginate collaborators in the console


bafonins:
#### Summary

Currently, the console omits the `page` and `limit` query parameters when fetching collaborators. This results in all collaborators being fetched.

#### Why do we need this?

We dont need all collaborators stored on the frontend.

#### What is already there? What do you see now?

* Pages to list and edit collaborators
* The collaborators service in the js sdk
* Middleware for application collaborators

#### What is missing? What do you want to see?

1. `params` arguments passed to the `getAll` method in the `collaborators` service
2. `page` and `limit` included when requesting a list of collaborators
3. extend js sdk to fetch a single collaborator (for the collaborator edit page), see https://github.com/TheThingsNetwork/lorawan-stack/issues/721

#### Environment

browser

#### How do you propose to implement this?

We first need the IS to support single collaborator fetching. Then it is the matter of adjusting the collaborators management similar to how we handle api-keys. See https://github.com/TheThingsNetwork/lorawan-stack/issues/664

#### Can you do this yourself and submit a Pull Request?

Yes, I can do this once https://github.com/TheThingsNetwork/lorawan-stack/issues/721 is implemented.

bafonins:
Note, that it will also break the console when navigating directly to the collaborator edit page of the entity that is not in the store. The upper bound for pagination is 1000 entities, see [L106](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/api/application.proto#L106)

Issue #1021, opened 2019-07-19
Fix persistent errors in entity sub views


bafonins:
#### Summary

Errors caused by API requests for sub entities (api keys, collaborators) persist causing the error view being shown when it shouldnt be the case.

#### Steps to Reproduce

1. Create application
2. Create application collaborator
3. Navigate to `/console/application/{app-id}/collaborators/collaborator-does-not-exist`
5. Go to the collaborators list
6. Select 

#### What do you see now?

Persistent error

#### What do you want to see instead?

The Application Collaborator Edit view without errors

 #### Environment

browser (console)

#### How do you propose to implement this?

Check for errors in `componentDidUpdate` instead of `render`. This would allow the view to dispatch the `_REQUEST` action and cleanup the errors state from previous errors.

#### Can you do this yourself and submit a Pull Request?

yes

Issue #852, opened 2019-06-26
Restructure redux store's file structure and use pagination workflow


kschiffer:
#### Summary
Complementary to #825, we should restructure the redux store for all other entities as well. This issue keeps track of that progress.

- [x] Applications ⟶ #825 
- [ ] Devices
- [x] Gateways ⟶ #868
- [ ] Webhooks
- [x] Collaborators
- [ ] Api keys

#### Why do we need this?
To clean up our primitives and file structure and to work with a clear contract on how store logic is structured.

#### What is already there? What do you see now?
Restructured application store

#### What is missing? What do you want to see?
- Related logic put into `store/{actions|middleware/logics|reducers|selectors}/{entity}.js`.
- No more scattering of related store logic in files with different names.
- Usage of new pagination workflow

#### Environment
Console.

#### How do you propose to implement this?
Complementary to #825.

#### Can you do this yourself and submit a Pull Request?
We can both pick these up while working on them.

PR #1022, opened 2019-07-19
Fix collaborators handling in console

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/722

#### Changes

- Add the `params` argument to the `getAll` method of the collaborators service
- Refactor app/gtw collaborators store similar to app/gtw api-keys store
- Change routes for editing collaborators `/console/{entity}/{id}/collaborators/{collaboratorId}` to `/console/{entity}/{id}/collaborators/user/{collaboratorId}` and `/console/{entity}/{id}/collaborators/organization/{collaboratorId}`

#### Notes for Reviewers

1. New routes to the edit view are necessary to distinguish between user and organization collaborators. If the user navigated directly to `/console/{entity}/{id}/collaborators/{collaboratorId}` there is no way to know the type of the collaborator and hence no clear way to decide which endpoint to call (after https://github.com/TheThingsNetwork/lorawan-stack/pull/1001 we cannot fetch organization collaborators via the `user` endpoint)
2. Error handling:
```
  componentDidUpdate (prevProps) {
    const { error } = this.props

    if (Boolean(error) && prevProps.error !== error) {
      throw error
    }
  }
```
See https://github.com/TheThingsNetwork/lorawan-stack/issues/1021
3. You can use [this scrpt](https://gist.github.com/bafonins/8dc9d62866132d136cd2489643520fc0) to populate the stack with test entities.

### DIFF
--- a/pkg/webui/console/store/reducers/index.js
+++ b/pkg/webui/console/store/reducers/index.js
@@ -31,6 +31,7 @@ import configuration from './configuration'
 import createNamedApiKeysReducer from './api-keys'
 import createNamedRightsReducer from './rights'
 import createNamedCollaboratorsReducer from './collaborators'
+import createNamedCollaboratorReducer from './collaborator'
 import createNamedEventsReducer from './events'
 import createNamedApiKeyReducer from './api-key'
 import link from './link'
@@ -64,6 +65,7 @@ export default combineReducers({
     gateways: createNamedRightsReducer(GATEWAY_SHARED_NAME),
   }),
   collaborators: combineReducers({
+    application: createNamedCollaboratorReducer(APPLICATION_SHARED_NAME),
     applications: createNamedCollaboratorsReducer(APPLICATION_SHARED_NAME),
     gateways: createNamedCollaboratorsReducer(GATEWAY_SHARED_NAME),
   }),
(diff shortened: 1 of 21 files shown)


pkg/webui/console/store/reducers/index.js
```diff
@@ -64,7 +65,9 @@ export default combineReducers({
     gateways: createNamedRightsReducer(GATEWAY_SHARED_NAME),
   }),
   collaborators: combineReducers({
+    application: createNamedCollaboratorReducer(APPLICATION_SHARED_NAME),
```
kschiffer: Why not using sth like `selectedCollaborator` here, like we do for app/gtw/dev?

pkg/webui/console/views/application-collaborator-add/index.js
```diff
@@ -77,6 +77,14 @@ export default class ApplicationCollaboratorAdd extends React.Component {
     getApplicationsRightsList()
   }
 
+  componentDidUpdate (prevProps) {
```
kschiffer: Since we will likely use this pattern a lot, it might make sense to HOC this and use it as a decorator. (Not in this PR obviously)

pkg/webui/console/views/application-collaborator-edit/index.js
```diff
@@ -101,6 +108,14 @@ export default class ApplicationCollaboratorEdit extends React.Component {
     loadData(appId)
```
kschiffer:
```suggestion
    loadData()
```

pkg/webui/console/store/reducers/index.js
bafonins: trying to make the diff smaller

kschiffer: Can you make a tech debt issue about this?

bafonins: https://github.com/TheThingsNetwork/lorawan-stack/issues/852

PR #1028, opened 2019-07-21
gs: Get fallback frequency plan ID from server context

### LINKED ISSUE #1027: Basic Station frontend does not use fallback FrequencyPlan ID
#### Summary

Basic Station frontend does not use fallback FrequencyPlan ID when it's configured.

#### Steps to Reproduce

1. Set env `TTN_LW_GS_BASIC_STATION_FALLBACK_FREQUENCY_PLAN_ID=EU_863_870`.
2. Connect an unregistered gateway

#### What do you see now?

- Gateway connection dropped.

- Gateway connected using the fallback frequency.

#### Environment

Latest master.

- Copy the fallback frequencyplan id from server context to request context.

yup.

KrishnaIyer:
#### Summary

Closes #1027 

#### Changes

- Copy the fallback frequency plan id from server context to request context.

### DIFF
--- a/pkg/gatewayserver/io/basicstationlns/basicstationlns.go
+++ b/pkg/gatewayserver/io/basicstationlns/basicstationlns.go
@@ -29,6 +29,7 @@ import (
 	"go.thethings.network/lorawan-stack/pkg/auth/rights"
 	"go.thethings.network/lorawan-stack/pkg/basicstation"
 	"go.thethings.network/lorawan-stack/pkg/errors"
+	"go.thethings.network/lorawan-stack/pkg/frequencyplans"
 	"go.thethings.network/lorawan-stack/pkg/gatewayserver/io"
 	"go.thethings.network/lorawan-stack/pkg/gatewayserver/io/basicstationlns/messages"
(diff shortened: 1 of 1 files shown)


pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -170,6 +171,12 @@ func (s *srv) handleTraffic(c echo.Context) error {
 		ctx = metadata.NewIncomingContext(s.ctx, md)
 	}
 
+	// If a fallback frequency is defined, inject it into the request context.
```
johanstokking: Wrong comment

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -170,6 +171,12 @@ func (s *srv) handleTraffic(c echo.Context) error {
...
+	fallback, ok := frequencyplans.FallbackIDFromContext(s.ctx)
+	if ok {
```
johanstokking: No need for change, but we put this stuff on a single line

PR #1031, opened 2019-07-22
Fix component throw persistent error

### LINKED ISSUE #1021: Fix persistent errors in entity sub views

bafonins:
#### Summary

Closes #1021 

#### Changes

- Add `withRequest` hoc
- Fix application api keys pages
- Fix gateway api keys pages
- Fix application integrations page
- Fix application collaborators pages
- Fix gateway collaborators pages

#### Notes for Reviewers

~Alternative solution could be:  https://gist.github.com/bafonins/e5738ac23bb0aff2ce1e144a4c3a536a, however I am not sure if we want to introduce this hoc.~

### DIFF
--- a/pkg/webui/console/views/application-api-key-add/index.js
+++ b/pkg/webui/console/views/application-api-key-add/index.js
@@ -18,13 +18,13 @@ import bind from 'autobind-decorator'
 import { connect } from 'react-redux'
 import { replace } from 'connected-react-router'
 
-import Spinner from '../../../components/spinner'
 import Breadcrumb from '../../../components/breadcrumbs/breadcrumb'
 import { withBreadcrumb } from '../../../components/breadcrumbs/context'
 import sharedMessages from '../../../lib/shared-messages'
 import Message from '../../../lib/components/message'
 import IntlHelmet from '../../../lib/components/intl-helmet'
 import { ApiKeyCreateForm } from '../../components/api-key-form'
+import withRequest from '../../../lib/components/with-request'
 
 import { getApplicationsRightsList } from '../../store/actions/applications'
 import {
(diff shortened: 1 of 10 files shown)


kschiffer:
> Alternative solution could be: https://gist.github.com/bafonins/e5738ac23bb0aff2ce1e144a4c3a536a, however I am not sure if we want to introduce this hoc.

I actually think the HOC solution is good. It saves us a lot of repetition and also ensures a streamlined UX for loading data and handling request errors. What are your concerns?

Also, what is the reasoning behind these lines:
https://gist.github.com/bafonins/e5738ac23bb0aff2ce1e144a4c3a536a#file-with-request-js-L24-L26

kschiffer:
Code looks good but what about
- Gateway/Application collaborator add/edit
- Application payload formatters

?

Should we use the new HOC also for our entity pages (`views/{application|gateway|device}.js`)?

bafonins:
> Gateway/Application collaborator add/edit

added

> Application payload formatters

what is the problem there?

> Should we use the new HOC also for our entity pages (views/{application|gateway|device}.js)?

We can

kschiffer:
> > Application payload formatters
> 
> what is the problem there?

The view is fetching the application link, could hence also use the HOC, at least for the fetching/spinner logic.

bafonins:
> The view is fetching the application link, could hence also use the HOC, at least for the fetching/spinner logic.

Does the view suffer from the problem described in the issue?
> Errors caused by API requests for sub entities (api keys, collaborators) persist causing the error view being shown when it shouldnt be the case.

I have no problems adding the hoc to other views that the same loading/error logic, but I would like to keep the diff as small as possible and solve only the problems listed in the issue

kschiffer:
> Does the view suffer from the problem described in the issue?

No, but this is how we will be piling up more and more technical debt and inconsistencies. So I would prefer to either having an issue to keep track of this or to solve it right away, while we're at it.

bafonins:
> So I would prefer to either having an issue to keep track of this or to solve it right away, while we're at it.

will create an issue

PR #1036, opened 2019-07-23
Fix HTTP Error code conversion

### LINKED ISSUE #1014: Error 401 returned from backend with wrong error code
#### Summary

When making api requests without access token and cokie, an error `401` unauthenticated is returned. However when trying to recognize the error using grpc or http codes, the error wasn't recognized. Upon inspection it was found that the grpc code returned is 2, which is unknown error.

#### Steps to Reproduce

1. Make request without access token or cookie (for example, log out in one console tab, and open Gateway list in another console tab)
2. Observe response from the server

#### What do you see now?

```
{  
  "code":2,
  "message":"error:pkg/errors/web:http (HTTP error: code=401, message=You are not logged in)",
  "details":[  
    {  
      "@type":"type.googleapis.com/ttn.lorawan.v3.ErrorDetails",
      "namespace":"pkg/errors/web",
      "name":"http",
      "message_format":"HTTP error: {message}",
      "attributes":{  
        "message":"code=401, message=You are not logged in"
      }
    }
  ]
}
```

The correct grpc code should be 16

Intercept this error somewhere in backend and set the correct error code

I can do it, but maybe its faster if @htdvisser looks at it (if time allows)

htdvisser:
#### Summary

This Pull Request fixes the conversion from Echo errors to TTN errors.

Additionally it turns the particular errors mentioned in #1014 into TTN errors.

Resolves #1014 	

#### Changes

- Fixed conversion from Echo errors to TTN errors
- Changed errors used in Console OAuth flow

#### Release Notes

- Fixed an issue where incorrect error codes were returned from the console's OAuth flow.

### DIFF
--- a/pkg/console/cookie_auth.go
+++ b/pkg/console/cookie_auth.go
@@ -45,27 +45,22 @@ type authCookie struct {
 	Expiry       time.Time
 }
 
-// setCookie sets the authCookie on the request.
-func (console *Console) setAuthCookie(c echo.Context, value authCookie) error {
-	return console.AuthCookie().Set(c, value)
-}
-
-// getCookie returns the authCookie from the echo context.
 func (console *Console) getAuthCookie(c echo.Context) (authCookie, error) {
 	value := authCookie{}
 	ok, err := console.AuthCookie().Get(c, &value)
 	if err != nil {
 		return authCookie{}, err
 	}
-
 	if !ok {
-		return authCookie{}, echo.NewHTTPError(http.StatusUnauthorized, "You are not logged in")
+		return authCookie{}, echo.NewHTTPError(http.StatusUnauthorized, "No auth cookie")
 	}
-
 	return value, nil
 }
 
-// removeCookie removes the authCookie.
+func (console *Console) setAuthCookie(c echo.Context, value authCookie) error {
+	return console.AuthCookie().Set(c, value)
+}
+
 func (console *Console) removeAuthCookie(c echo.Context) {
 	console.AuthCookie().Remove(c)
 }
--- a/pkg/webui/lib/errors/utils.js
+++ b/pkg/webui/lib/errors/utils.js
@@ -87,7 +87,7 @@ export const isNotFoundError = error => (
  * `false` otherwise.
  */
 export const isInternalError = error => (
-  grpcStatusCode(error) === 13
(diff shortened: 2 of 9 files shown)


htdvisser:
TODO: cleanup frontend/sdk code cc: @pgalic96

pgalic96:
https://github.com/TheThingsNetwork/lorawan-stack/blob/feature/971-reset-password-ui/pkg/webui/lib/errors/utils.js#L58-L62

pkg/webui/lib/errors/utils.js
```diff
@@ -118,9 +118,6 @@ export const isPermissionDeniedError = error => (
  */
 export const isUnauthenticatedError = error => (
   grpcStatusCode(error) === 16
```
pgalic96: Should we have this as well as in `404` error?
```suggestion
  grpcStatusCode(error) === 16 || httpStatusCode(error) === 401
```

htdvisser: We could do that, but then we should do the same for `isInternalError`, `isAlreadyExistsError` and `isPermissionDeniedError`.

pkg/console/cookie_auth.go
```diff
@@ -45,27 +45,22 @@ type authCookie struct {
...
 func (console *Console) getAuthCookie(c echo.Context) (authCookie, error) {
 	value := authCookie{}
 	ok, err := console.AuthCookie().Get(c, &value)
 	if err != nil {
 		return authCookie{}, err
 	}
-
 	if !ok {
-		return authCookie{}, echo.NewHTTPError(http.StatusUnauthorized, "You are not logged in")
+		return authCookie{}, echo.NewHTTPError(http.StatusBadRequest, "No auth cookie")
```
bafonins: Should this return `401` instead of `400`? 
This rollbacks changes made in https://github.com/TheThingsNetwork/lorawan-stack/pull/1007

htdvisser: Ok

pkg/webui/lib/errors/utils.js
htdvisser: Done.

pkg/console/cookie_auth.go
htdvisser: Done.

PR #1041, opened 2019-07-23
Add option to hide one page pagination

### LINKED ISSUE #728: Hide pagination component when there is only one page available
#### Summary
The console currently shows the pagination component even when the paginated list only has one page. In this case, the pagination is superfluous and should be hidden.

#### Why do we need this?
Less visual load ⟶ better UX.

Pagination mechanism that works well otherwise.

The pagination component hiding itself when there is only one page.

#### Environment
Console in browser.

Simply add a check to hide the component when there is only one page.

Yes, but might be worth for @bafonins to pick up.

kschiffer:
#### Summary
Closes #728 

#### Changes
- Add an option to the pagination component to hide the widget when there is only one page displayed (will be enabled by default)

### DIFF
--- a/pkg/webui/components/pagination/__snapshots__/index_test.js.snap
+++ b/pkg/webui/components/pagination/__snapshots__/index_test.js.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Pagination has only a single page should match snapshot 1`] = `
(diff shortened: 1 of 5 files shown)


kschiffer:
I see your point. But hiding pagination for one page is a generally applicable rule inherent to the pagination itself, which is why I added it there. Otherwise, we would have to re-define this logic wherever we use the component.

PR #1042, opened 2019-07-23
Reset and change password UI

### LINKED ISSUE #971: Password Reset from web UI
#### Summary

Users should be able to reset their password using the web UI.

Old issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/1123

#### Why do we need this?

Because not all users are comfortable with the API or CLI

We have the `CreateTemporaryPassword` RPC in the `UserRegistry` service. HTTP binding is `POST /api/v3/users/{user_ids.user_id}/temporary_password`. This needs to be called with the user ID and Email Address. A successful call results in an email being sent to the user with a temporary password that the can only use to reset their password with the `UpdatePassword` RPC in the `UserRegistry` service.

The following screens in the Web UI of the Identity Server:

- A "Forgot Password" screen where the user has to enter their username and email in order to receive a temporary password.
- A "Reset Password" screen where the user can enter their username, their (current or temporary) password and a new password (+confirmation). In the future this screen will also get a checkbox that will make the IS revoke all sessions and access tokens #267.

Assigning @pgalic96 since he was already assigned to the original issue. Please discuss with @kschiffer the design. I propose to implement it in a (new) "account" React app that will also have the screen I discussed in #265. Note that the user does not need to be logged in for the screens proposed in this issue. Also note that these must **not** be implemented in the Console, since the Console must never have access to the user's password.

pgalic96:
#### Summary

Closes #971

#### Changes

- Add `reset password` view for sending a reset password request
- Add `change password` view for resetting password after temporary token is obtained
- Add api requests for reset and change password

#### Notes for Reviewers

- Please note that the `change password` view is not intended to be accessed with a button, at least not for the user that is not logged in, we can reuse this view to make a change password view for the logged in user, so not in case of password reset
- when sending a reset password request, in the stack log you can obtain the email that is sent to the user with the temporary token 
- you can access the change password UI by accessing `/oauth/change-password/{user_id}/{temp_token}` (the email that contains the temporary token should also contain the full link)
- ideally in the future the password reset would happen by just inputting the e-mail address
- I reused the login component for these views, if you prefer other design (cc @kschiffer), let me know.

#### Release Notes

- Added web interface for requesting temporary password
- Added web interface for changing password


pkg/webui/oauth/api/index.js
```diff
@@ -26,6 +26,12 @@ export default {
     async register (userData) {
       return axios.post(`/api/v3/users`, userData)
     },
+    reset_password (user_id) {
+      return instance.post(`/api/v3/users/${user_id}/temporary_password`)
+    },
+    update_password (user_id, passwordData) {
+      return instance.put(`/api/v3/users/${user_id}/password`, passwordData)
```
bafonins: do you need the `csrf` token in the headers?

pkg/webui/oauth/api/index.js
```diff
@@ -26,6 +26,12 @@ export default {
     async register (userData) {
       return axios.post(`/api/v3/users`, userData)
     },
+    reset_password (user_id) {
+      return instance.post(`/api/v3/users/${user_id}/temporary_password`)
```
bafonins: do you need the `csrf` token in the headers?

pkg/webui/oauth/views/login/index.js
```diff
@@ -79,6 +80,13 @@ export default class OAuth extends React.PureComponent {
     }))
   }
 
+  navigateToResetPassword () {
+    const { dispatch, location } = this.props
+    dispatch(replace('/oauth/reset-password', {
```
bafonins: move to `mapDispatchToProps`

pkg/webui/oauth/views/reset-password/index.js
```diff
@@ -0,0 +1,131 @@
...
+      setSubmitting(false)
+    }
+  }
+
+  navigateToLogin () {
+    const { dispatch, location } = this.props
+    const state = location.state || {}
+    const back = state.back || '/oauth/login'
+
+    dispatch(replace(back))
```
bafonins: move to `mapDispatchToProps`

pkg/webui/oauth/views/change-password/index.js
```diff
@@ -0,0 +1,170 @@
...
+const m = defineMessages({
+  changePassword: 'Change Password',
+  inputNewPassword: 'Input a new password',
+  passwordChanged: 'Password successfully changed',
+  loginPage: 'Go to Login page'
+})
+
+const digit = /(?=.*[\d])/
+const uppercase = /(?=.*[A-Z])/
+const special = /(?=.*[!@#$%^&*])/
```
bafonins: move those to the shared file `oauth/lib/regexp.js`; the register page has the same definitions

pkg/webui/oauth/views/change-password/index.js
```diff
@@ -0,0 +1,170 @@
...
+import SubmitButton from '../../../components/submit-button'
+import Logo from '../../../components/logo'
+import IntlHelmet from '../../../lib/components/intl-helmet'
+import Message from '../../../lib/components/message'
+
+import style from '../login/login.styl'
+
+const m = defineMessages({
+  changePassword: 'Change Password',
+  inputNewPassword: 'Input a new password',
```
bafonins: not used anywhere

pkg/webui/oauth/views/change-password/index.js
```diff
@@ -0,0 +1,170 @@
...
+  passwordChanged: 'Password successfully changed',
+  loginPage: 'Go to Login page'
```
bafonins: the register page defines the same message

pkg/webui/oauth/views/reset-password/index.js
```diff
@@ -0,0 +1,131 @@
...
+  loginPage: 'Login Page',
+  passwordRequested: 'New password reset token has been sent to user email',
+  resetPassword: 'Reset Password',
+  requestTempPassword: 'Request a new password reset token',
+  stackAccount: 'TTN Stack Account',
+})
+
+const validationSchema = Yup.object().shape({
+  user_id: Yup.string()
+  .required(sharedMessages.validateRequired),
```
bafonins: please setup your code editor to format the code according to `.editorconfig` and `eslint` rules

pkg/webui/oauth/views/change-password/index.js
```diff
@@ -0,0 +1,170 @@
...
+    .required(sharedMessages.validateRequired),
+})
+
+const initialValues = {
+  new_password: '',
+  new_confirm: '',
+}
+
+@connect()
+@withRouter
```
bafonins: any reasons to use `withRouter` here?

pkg/webui/oauth/views/reset-password/index.js
```diff
@@ -0,0 +1,131 @@
...
+})
+
+@connect()
+@withRouter
```
bafonins: any reasons to use `withRouter` here?

pkg/webui/oauth/views/reset-password/index.js
```diff
@@ -0,0 +1,131 @@
...
+
+import Button from '../../../components/button'
+import Form from '../../../components/form'
+import Input from '../../../components/input'
+import SubmitButton from '../../../components/submit-button'
+import Logo from '../../../components/logo'
+import IntlHelmet from '../../../lib/components/intl-helmet'
+import Message from '../../../lib/components/message'
+
+import style from '../login/login.styl'
```
bafonins: define styles for this view, if there shared bits consider moving those to `mixins` or make a reusable component

pkg/webui/oauth/views/app/index.js
```diff
@@ -78,7 +80,9 @@ export default class OAuthApp extends React.PureComponent {
                     <Route path="/oauth/login" component={Login} />
                     <Route path="/oauth/authorize" component={Authorize} />
                     <Route path="/oauth/register" component={CreateAccount} />
+                    <Route path="/oauth/reset-password" component={ResetPassword} />
                     <Route path="/oauth/code" component={Code} />
+                    <Route path="/oauth/change-password/:user_id/:token" component={ChangePassword}/>
```
bafonins: Correct me if im wrong. Is the `token` param - the old password?

pgalic96: no, it's the new token obtained from the reset password and it's used as a parameter to change the password

pkg/webui/oauth/views/reset-password/index.js
```diff
@@ -0,0 +1,129 @@
...
+
+const m = defineMessages({
+  loginPage: 'Login Page',
+  passwordRequested: 'New password reset token has been sent to user email',
+  resetPassword: 'Reset Password',
```
kschiffer:
```suggestion
  resetPassword: 'Send',
```

pkg/webui/oauth/views/reset-password/index.js
```diff
@@ -0,0 +1,129 @@
...
+import SubmitButton from '../../../components/submit-button'
+import Logo from '../../../components/logo'
+import IntlHelmet from '../../../lib/components/intl-helmet'
+import Message from '../../../lib/components/message'
+
+import style from '../login/login.styl'
+
+const m = defineMessages({
+  loginPage: 'Login Page',
+  passwordRequested: 'New password reset token has been sent to user email',
```
kschiffer:
```suggestion
  passwordRequested: 'You will receive an email with reset instructions shortly.',
```

pkg/webui/oauth/views/change-password/index.js
```diff
@@ -0,0 +1,161 @@
...
+
+  navigateToLogin () {
+    const { replace } = this.props
+
+    replace('/oauth/login')
+  }
+
+  render() {
+    const { error, info } = this.state
+    const { match: {params } } = this.props
```
kschiffer:
```suggestion
    const { match: { params } } = this.props
```

pkg/webui/oauth/views/change-password/index.js
```diff
@@ -0,0 +1,161 @@
...
+  constructor (props) {
+    super(props)
+    this.state = {
+        error: '',
+        info: '',
+    }
+  }
+
+  async handleSubmit (values, { setSubmitting }) {
+    const { match: {params }, push } = this.props
```
kschiffer:
```suggestion
    const { match: { params }, push } = this.props
```

As @bafonins said, check your editor setup.

pkg/webui/oauth/views/change-password/index.js
```diff
@@ -0,0 +1,157 @@
...
+            <Message content={m.changePassword} component="h1" className={style.title} />
+            <Form
+              onSubmit={this.handleSubmit}
+              initialValues={initialValues}
+              error={error}
+              info={info}
+              validationSchema={validationSchema}
+              horizontal={false}
+            >
+              <Form.Field
```
htdvisser: Missing a form field for old password here (can be prefilled from query).

pkg/webui/oauth/views/change-password/index.js
```diff
@@ -0,0 +1,157 @@
...
+import style from '../create-account/create-account.styl'
+
+const m = defineMessages({
+  changePassword: 'Change Password',
+  passwordChanged: 'Password changed successfully',
+})
+
+const validationSchema = Yup.object().shape({
+  new_password: Yup.string()
+    .min(8)
```
htdvisser: Password rules are config, not static.

pkg/webui/oauth/views/change-password/index.js
```diff
@@ -0,0 +1,157 @@
...
+    .matches(digit, sharedMessages.validatePasswordDigit)
+    .matches(uppercase, sharedMessages.validatePasswordUppercase)
+    .matches(special, sharedMessages.validatePasswordSpecial)
+    .required(sharedMessages.validateRequired),
+  new_confirm: Yup.string()
+    .oneOf([ Yup.ref('new_password'), null ], sharedMessages.validatePasswordMatch)
+    .min(8)
```
htdvisser: Password rules are config, not static.

pkg/webui/oauth/views/app/index.js
```diff
@@ -78,7 +80,9 @@ export default class OAuthApp extends React.PureComponent {
                     <Route path="/oauth/login" component={Login} />
                     <Route path="/oauth/authorize" component={Authorize} />
                     <Route path="/oauth/register" component={CreateAccount} />
+                    <Route path="/oauth/reset-password" component={ResetPassword} />
                     <Route path="/oauth/code" component={Code} />
+                    <Route path="/oauth/change-password/:user_id/:token" component={ChangePassword} />
```
htdvisser: It's not `token`. It's `current_password` or in some cases `temporary_password`. Also I wouldn't put that in the path but in the query.

pkg/webui/oauth/views/change-password/index.js
```diff
@@ -0,0 +1,157 @@
...
+              />
+              <Form.Field
+                component={Input}
+                required
+                title={sharedMessages.confirmPassword}
+                name="new_confirm"
+                type="password"
+                autoComplete="new-password"
+              />
+              <Form.Submit
```
htdvisser: Please add a form field for `revoke_all_access` added in #1042

pkg/webui/oauth/views/create-account/index.js
```diff
@@ -59,15 +48,15 @@ const validationSchema = Yup.object().shape({
     .max(50, sharedMessages.validateTooLong),
   password: Yup.string()
     .min(8)
```
htdvisser: Password rules are config, not static.

pkg/webui/oauth/views/app/index.js
pgalic96: I think temporary password is a very confusing naming, since you cannot log in with it. The definition of a password is:
> a secret word or phrase that must be used to gain admission to a place.
IMO, this is a password reset token, only used when changing password. But if it's too much work to change the naming for the sake of clarity, I am fine with it as well.

pkg/webui/oauth/views/change-password/index.js
pgalic96: > What do you think about a flow where, after requesting the temporary password, the web UI would immediately switch to change-password?

This would not work if we intend to supply the `temporary_password` in URI, since the `temporary_password` would be available only in the e-mail. If we redirect to change-password view there would be no `temporary_password` supplied.

> Missing a form field for old password here (can be prefilled from query).
This should be there for the case when a logged in user wants to change the password.

pkg/webui/oauth/views/app/index.js
pgalic96: Also, not sure what you mean by query. Would we query user id that is in the input for the temporary password associated for them and then make an api request for changing password? Wouldn't that mean that anyone could reset and change a password for any user?

htdvisser: The term "temporary password" has been part of our API since October last year, it's a little bit too late to change it now.

What I mean with query is that we could put a link in the password reset email that prefills form fields of the "change password" form, for example:

http://localhost:1885/oauth/change-password?user_id=admin&current=supersecret

which would prefill the `user_id` and `current` fields with the given contents.

pkg/webui/oauth/views/change-password/index.js
htdvisser: > > Missing a form field for old password here (can be prefilled from query).
> This should be there for the case when a logged in user wants to change the password.

The temporary password is equivalent to the current password here, so why not treat them the same?

pkg/webui/oauth/views/app/index.js
kschiffer: I don't think we need to show the `user_id` and `temporary_password` fields prefilled in the reset form anyway. Both should be redundant for what the user is trying to do, he/she just cares about changing the password. This is also the flow I know from other apps.

Just two fields for changing the password should suffice, the request body can then be extended with the info from the query.

Implementing two slightly different views for the two scenarios of resetting after forgetting (2 fields) and simply changing the password (3 fields) is not expensive IMO.

@htdvisser would you be ok with that?

pkg/webui/oauth/views/change-password/index.js
kschiffer: I think it's better to fix this in a separate issue to maintain the scope here.
@pgalic96 can you follow up with an issue here?

pkg/webui/oauth/views/login/index.js
```diff
@@ -73,10 +77,17 @@ export default class OAuth extends React.PureComponent {
   }
 
   navigateToRegister () {
-    const { dispatch, location } = this.props
-    dispatch(replace('/oauth/register', {
+    const { replace, location } = this.props
+    replace('/oauth/register', {
```
kschiffer: Why are we not using `push` in these?
cc @bafonins

kschiffer:
> Could we perhaps do a little renaming for naming consistency?
> 
> `reset-password` to `forgot-password` (same as CLI)
> `change-password` to `update-password` (same as CLI and API call name)

:+1:

> @pgalic96 this pull request does not close the issue yet. It is currently only possible to update the password using the "temporary password over email" flow, not with the "current password" flow. So either implement both flows, or remove the "closes " from the PR description.

Right. I would vote for doing this in a separate PR as well.

> I don't think we should implement a separate page for something that results in the exact same API call, and something that is also one and the same action in the CLI.

The context is different;
1. The user is not logged in and wants to reset the password ⟶ he/she uses the forgot password flow and the reset password page, a one-off page, accessible without being logged in
2. The user is logged in and wants to change the password ⟶ he/she uses a change password view inside the (yet to be added) profile settings panel

We can implement a change password component that can then be used in both views.

* * *

Let me formulate the flow that I think is ideal and conventional (and possible with what we have):
**Forgot password**
1. From the login page, the user can click on "Forgot Password?"
2. The user would get info on the flow and prompted to enter his/her username (or even better: email)
3. The user gets a confirmation message on the screen that an email was sent
4. In the email, the user receives a concise explanation of the process and: 1) A tokenized link to the reset password page; 2) Just the token as bare value for usage in the CLI
5. The link leads to a one-off reset password page, consisting of just two fields for the new password and a submit button
6. Upon hitting submit, the user is redirected to the login page together with a message that the password was reset successfully

**Update password**
1. The user logs in
2. On the landing page after login, there is a section for profile settings (we can make a very bare minimum thing here for a start)
3. In the profile settings, there is an option to update the password
4. The update password page is a simple form with three fields (old pw, new pw, confirm pw)
5. After hitting "Change" submit button, a confirmation is shown on the screen

kschiffer:
Talking to @pgalic96 again about this, we concluded that for the sake of solving the issue, we will indeed use the same route/view for updating the password. The form will adapt based on whether the user is logged in then and whether the query is present. We can later flesh out the UX details for both flows.

pkg/webui/oauth/api/index.js
```diff
@@ -29,6 +29,12 @@ export default {
     async register (userData) {
       return axios.post(`/api/v3/users`, userData)
     },
+    async reset_password (user_id) {
```
bafonins:
```suggestion
    async resetPassword (user_id) {
```

pkg/webui/oauth/api/index.js
```diff
@@ -29,6 +29,12 @@ export default {
...
+      return axios.post(`/api/v3/users/${user_id}/temporary_password`)
+    },
+    async update_password (user_id, passwordData) {
```
bafonins:
```suggestion
    async updatePassword (user_id, passwordData) {
```

pkg/webui/oauth/views/forgot-password/index.js
```diff
@@ -0,0 +1,124 @@
...
+  resetPasswordDescription: 'Please enter your username to receive email with reset instructions',
+  requestTempPassword: 'Reset Password',
+})
+
+const validationSchema = Yup.object().shape({
+  user_id: Yup.string()
+    .required(sharedMessages.validateRequired),
+})
+
+@connect(null, { push })
```
bafonins: https://github.com/TheThingsNetwork/lorawan-stack/issues/1086

pkg/webui/oauth/views/forgot-password/index.js
```diff
@@ -0,0 +1,124 @@
...
+      this.setState({
+        error: error.response.data,
+        info: '',
+      })
+    } finally {
+      setSubmitting(false)
+    }
+  }
+
+  navigateToLogin () {
```
bafonins: you can move this completely to `mapDispatchToProps`

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,209 @@
...
+})
+
+const initialValues = {
+  password: '',
+  confirm: '',
+  password_changed: false,
+  revoke_all_access: false,
+}
+
+@connect(state => ({
```
bafonins: https://github.com/TheThingsNetwork/lorawan-stack/issues/1086

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,209 @@
...
+      this.setState({
+        error: error.response.data,
+        info: '',
+      })
+    } finally {
+      setSubmitting(false)
+    }
+  }
+
+  handleCancel () {
```
bafonins: you can move it to `mapDispatchToProps`

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  handleRevokeAllAccess (evt) {
+    this.setState({ revoke_all_access: evt.target.checked })
+  }
+
+  async handleSubmit (values, { setSubmitting }) {
+    const { user } = this.props
+    const userParams = queryString.parse(this.props.location.search)
+    const old_or_temp = values.old_password ? values.old_password : userParams.current
```
bafonins: please use snake and camel case consistently.
we usually keep using snake case when destructing models returned by the backend. the rest is camel case

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,209 @@
...
+    const user_id = Boolean(user) ? user.ids.user_id : userParams.user
+    try {
+      await api.users.update_password( user_id, {
+        user_ids: { 'user_id': user_id },
```
bafonins:
```suggestion
        user_ids: { user_id },
```
preferably
```suggestion
        user_ids: { 'user_id': userId },
```

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,209 @@
...
+          info: m.passwordChanged,
+          changed: true,
+        })
+      } else {
+        push('/login', {
+          info: m.passwordChanged,
+        })
+      }
+    } catch (error) {
+      console.log(error)
```
bafonins: remove

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,209 @@
...
+  password_changed: false,
+  revoke_all_access: false,
+}
+
+@connect(state => ({
+  fetching: state.user.fetching,
+  user: state.user.user,
+  }), {
+    push,
+    replace,
```
bafonins: indentation
https://github.com/TheThingsNetwork/lorawan-stack/pull/1042#discussion_r306339487

pkg/webui/oauth/views/forgot-password/index.js
```diff
@@ -0,0 +1,130 @@
...
+import PropTypes from '../../../lib/prop-types'
+
+import style from './forgot-password.styl'
+
+const m = defineMessages({
+  loginPage: 'Login Page',
+  forgotPassword: 'Forgot Password',
+  passwordRequested: 'You will receive an email with reset instructions shortly.',
+  resetPassword: 'Send',
+  resetPasswordDescription: 'Please enter your username to receive email with reset instructions',
```
kschiffer:
```suggestion
  resetPasswordDescription: 'Please enter your username to receive an email with reset instructions',
```

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,221 @@
...
+            >
+              {oldPasswordField}
+              <Form.Field
+                component={Input}
+                required
+                title={m.newPassword}
+                name="password"
+                type="password"
+                autoComplete="new-password"
+                autoFocus
```
kschiffer: Autofocus should not be set when the olf password field is rendered. In that case, the old password field should have the autofocus.

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,221 @@
...
+      info: '',
+      revoke_all_access: false,
+    }
+  }
+
+  handleRevokeAllAccess (evt) {
+    this.setState({ revoke_all_access: evt.target.checked })
+  }
+
+  async handleSubmit (values, { setSubmitting }) {
```
kschiffer:
```suggestion
  async handleSubmit (values, { resetForm }) {
```
See below.

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,221 @@
...
+        old: oldPassword,
+        revoke_all_access: values.revoke_all_access,
+      })
+      if (Boolean(user)) {
+        this.setState({
+          error: '',
+          info: m.passwordChanged,
+          changed: true,
+        })
+      } else {
```
kschiffer:
```suggestion
    resetForm(initialValues) 
      } else {
```
Best to clear the form after a successful update.

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,221 @@
...
+          info: m.passwordChanged,
+        })
+      }
+    } catch (error) {
+      this.setState({
+        error: error.response.data,
+        info: '',
+      })
+    } finally {
+      setSubmitting(false)
```
kschiffer: Move this to the error catch above and remove the `finally` block. See also above.

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,230 @@
...
+import IntlHelmet from '../../../lib/components/intl-helmet'
+import Message from '../../../lib/components/message'
+import style from '../create-account/create-account.styl'
+import Checkbox from '../../../components/checkbox'
+import Spinner from '../../../components/spinner'
+import PropTypes from '../../../lib/prop-types'
+
+const m = defineMessages({
+  newPassword: 'New Password',
+  oldPassword: 'Temporary or Old Password',
```
kschiffer:
```suggestion
  oldPassword: 'Temporary or Old Password',
```
Let's not mention the temporary password here. If the user obtained a temporary password, he/she was not logged in and will use the other update password flow. This might just generate confusion.

kschiffer:
Also in its current form, we should only merge this along with a change that will make the reset email contain a link to the update password page. Otherwise, the flow is disrupted.
@htdvisser, can you take care of that or assign somebody?

pkg/webui/oauth/views/forgot-password/index.js
```diff
@@ -0,0 +1,131 @@
...
+      this.setState({
+        error: error.response.data,
+        info: '',
+      })
+    } finally {
+      setSubmitting(false)
+    }
+  }
+
+  navigateToLogin () {
```
bafonins: you dont need this, you can pass `handleCancel` directly to the cancel button

pkg/webui/oauth/views/login/index.js
```diff
@@ -51,7 +52,10 @@ const validationSchema = Yup.object().shape({
 const appRoot = selectApplicationRootPath()
 
 @withRouter
-@connect()
+@connect(null, {
+  push,
```
bafonins: not used

bafonins: also, this breaks navigation to the register page, because you dont have `dispatch` in `props`. `L84`

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,230 @@
...
+          component={Input}
+          required
+          title={m.oldPassword}
+          name="old_password"
+          type="password"
+          autoComplete="old-password"
+          autoFocus
+        />
+      )
+    } else if (!userParams.user || !userParams.current) {
```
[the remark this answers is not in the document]
pgalic96: this doesn't make sense, this is a guard for accessing the update password view when not logged in and without query parameters. that's why it first checks if the user object is present, and if it isn't it checks for query parameters. If it would first check for query parameters then it would always return redirect even when the user is logged in since the query parameters are missing.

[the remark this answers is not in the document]
bafonins: https://gist.github.com/bafonins/13ef21b491073a65c6e23d8e911e26b0 ?

pkg/webui/oauth/views/forgot-password/index.js
```diff
@@ -0,0 +1,127 @@
...
+export default class ForgotPassword extends React.PureComponent {
+
+  static propTypes = {
+    handleCancel: PropTypes.func.isRequired,
+  }
+
+  constructor (props) {
+    super(props)
+
+    this.state = {
```
bafonins: you dont need constructor here, initialize `state` as a class property
```
export default class ForgotPassword extends React.PureComponent {
...
state = {...}
}

File: pkg/webui/oauth/views/forgot-password/index.js [hunk 44]
```diff
@@ -0,0 +1,127 @@
...
+        info: '',
+      })
+    } finally {
+      setSubmitting(false)
+    }
+  }
+
+  render () {
+    const { error, info, requested } = this.state
+    const initialUserId = { user_id: '' }
```
bafonins: `initialValues` and you can move it outside of the `render` method

pkg/webui/oauth/views/landing/index.js
```diff
@@ -30,15 +34,27 @@ export default class OAuth extends React.PureComponent {
   async handleLogout () {
     const { dispatch } = this.props
 
+    dispatch(initialize())
```
bafonins: doesnt look right

pkg/webui/oauth/views/landing/index.js
```diff
@@ -30,15 +34,27 @@ export default class OAuth extends React.PureComponent {
...
     dispatch(logout())
   }
 
+  async handleUpdatePassword () {
+    const { dispatch } = this.props
+
+    dispatch(initialize())
```
bafonins: doesnt look right

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,243 @@
...
+    user: PropTypes.object,
+    handleCancelUpdate: PropTypes.func.isRequired,
+    handleSessionRevoked: PropTypes.func.isRequired,
+    handlePasswordChanged: PropTypes.func.isRequired,
+  }
+
+  constructor (props) {
+    super(props)
+
+    this.state = {
```
bafonins: you dont need constructor here, initialize `state` as a class property
```
export default class ForgotPassword extends React.PureComponent {
...
state = {...}
}

File: pkg/webui/oauth/views/update-password/index.js [hunk 48]
```diff
@@ -0,0 +1,243 @@
...
+          title={m.oldPassword}
+          name="old_password"
+          type="password"
+          autoComplete="old-password"
+          autoFocus
+        />
+      )
+    }
+
+    var updatePasswordComponent = (
```
bafonins: dont use `var`s. 
once again: https://github.com/TheThingsNetwork/lorawan-stack/pull/1042#discussion_r306339487

```suggestion
    const updatePasswordComponent = (
```

pkg/webui/oauth/views/forgot-password/index.js
```diff
@@ -0,0 +1,127 @@
...
+                name="user_id"
+                component={Input}
+                autoFocus
+                required
+              />
+              <Form.Submit
+                component={SubmitButton}
+                message={m.send}
+              />
+              <Button naked secondary message={cancelButtonText} onClick={this.props.handleCancel} />
```
bafonins:
```suggestion
              <Button naked secondary message={cancelButtonText} onClick={handleCancel} />
```

pkg/webui/oauth/views/update-password/index.js
```diff
@@ -0,0 +1,243 @@
...
+                message={sharedMessages.changePassword}
+              />
+              <Button naked secondary message={sharedMessages.cancel} onClick={handleCancelUpdate} />
+            </Form>
+          </Col>
+        </Row>
+      </Container>
+    )
+
+    if (Boolean(user)) {
```
bafonins: doesnt look right

pkg/identityserver/user_registry.go
```diff
@@ -413,9 +413,10 @@ func (is *IdentityServer) updateUserPassword(ctx context.Context, req *ttnpb.Upd
 			return err
 		}
 		if valid {
-			if err := rights.RequireUser(ctx, req.UserIdentifiers, ttnpb.RIGHT_USER_ALL); err != nil {
-				return err
-			}
+			// TODO: Enable when we have 2FA
```
htdvisser: Please follow our guidelines for TODOs (including issue number)

pkg/identityserver/user_registry_test.go
```diff
@@ -119,50 +119,51 @@ func TestUserUpdateInvalidPassword(t *testing.T) {
 	})
 }
 
-func TestUsersPermissionDenied(t *testing.T) {
```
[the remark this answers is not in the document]
htdvisser: Apparently Github posted this on the red diff, it should be on the green line.

PR #1045, opened 2019-07-23
Do not require abp flag for multicast

johanstokking:
#### Summary

Do not require `--abp` when creating multicast devices

#### Changes

- Interpret `--multicast` as ABP implicitly


johanstokking:
Reason for release blocking is CLI support; multicast is new in the next minor. Plus it's already documented that you don't have to specify `--abp` when creating multicast devices.

Issue #1017, opened 2019-07-19
Flaky AS NATS test


johanstokking:
#### Summary

Flaky AS NATS test

#### Steps to Reproduce

Random

#### What do you see now?

```
  WARN No cookie hash key configured, generated a random one hash_key=CA90E1F03FA690E96351D8FB1163251BE7205552608AD070C6297727C322A7C100EF2DCD28DDF676B5B4B4529159585A748B4CC14F01F2685C4277897AF3E535 namespace=web test_name=TestApplicationServer
  WARN No cookie block key configured, generated a random one block_key=B2E6930B986F933271CB083C3912A2A2B056AE91196DAB0FA69E46DED7F24E39 namespace=web test_name=TestApplicationServer
  WARN No static assets found in any search path namespace=web search_paths=[] test_name=TestApplicationServer
  WARN No rights TTL configured                 test_name=TestApplicationServer
  WARN No cluster key configured, generated a random one key=58faddf0ecadac5f9f637c0e088435f5f8466f7fc1a0a89edabac396815ebd7c test_name=TestApplicationServer
  INFO Listening for connections                address=:9184 namespace=grpc protocol=gRPC test_name=TestApplicationServer
  INFO Listening for connections                address=:8099 namespace=web protocol=Web test_name=TestApplicationServer
  INFO Linked                                   application_uid=foo-app namespace=applicationserver network_server=ns test_name=TestApplicationServer
  INFO Subscribed                               application_uid=foo-app grpc_method=Subscribe grpc_service=ttn.lorawan.v3.AppAs namespace=applicationserver/io/grpc remote_addr=pipe request_id=01DG4K7VXZEKX3GKJ8D3AH9Q14 test_name=TestApplicationServer
  INFO Unsubscribed                             application_uid=foo-app grpc_method=Subscribe grpc_service=ttn.lorawan.v3.AppAs namespace=applicationserver/io/grpc remote_addr=pipe request_id=01DG4K7VXZEKX3GKJ8D3AH9Q14 test_name=TestApplicationServer
  WARN Finished streaming call                  duration=2.048457872s error=context canceled grpc_method=Subscribe grpc_service=ttn.lorawan.v3.AppAs namespace=grpc request_id=01DG4K7VXZEKX3GKJ8D3AH9Q14 test_name=TestApplicationServer
[...]

PR #1050, opened 2019-07-24
Fix flaky AS test in CI

adriansmares:
#### Summary

References #1017

#### Changes

- Increase timeout only for upstream messages inside tests

#### Notes for Reviewers

I'm not closing #1017 directly after merging this, since I do believe this heisenbug needs more time testing than just CI passing 10 times.


johanstokking:
CI is not actually running on draft PRs afaik

adriansmares:
I've restarted about 16 builds on Travis, and some locally, without the error. The flush seems to be working well.

Issue #683, opened 2019-05-14
Remove optional automatic setting of local values in JS SDK


kschiffer:
#### Summary
After discussing the matter in #654, it was concluded that values should not be set automatically at all in the JS SDK.

#### Why do we need this?
See #654 

#### What is already there? What do you see now?
Setting device values when the `setDefault` option was set to true, such as keys, addresses, etc.
Generating random byte values in the SDK for certain fields when they have not been provided and are necessary.

#### What is missing? What do you want to see?
Not having the options above.

#### Environment
JS SDK in node or browser.

#### How do you propose to implement this?
1. Remove automatic setting of device values in JS SDK
2. Add respective device values and their generation to the console to creation and update logic of devices

#### Can you do this yourself and submit a Pull Request?
Yes.

kschiffer:
Resolved via #1057

kschiffer:
We still generate values in the SDK for devices created in ABP mode:
https://github.com/TheThingsNetwork/lorawan-stack/blob/master/sdk/js/src/service/devices/index.js#L258-L278

PR #1057, opened 2019-07-25
Handling of server addresses in Console

htdvisser:
#### Summary

This is a quickfix for the console that changes the prefilled gateway server address of the gateway create view to only fill the hostname and omit the port. A similar change was made in the SDK.

#### Changes

- Changed GS/NS/AS/JS address defaults in console from `host` to `hostname`


sdk/js/src/service/gateways.js
```diff
@@ -74,11 +75,21 @@ class Gateways {
...
   // Create
 
-  async create (userId = this._defaultUserId, gateway) {
+  async create (userId = this._defaultUserId, gateway, { setDefaults = true }) {
+    let gtw = gateway
+    const Url = URL ? URL : window.URL
+
+    if (setDefaults) {
+      gtw = {
+        gateway_server_address: this._stackConfig.gs.enabled ? new Url(this._stackConfig.gs).hostname : '',
```
kschiffer:
```suggestion
        gateway_server_address: this._stackConfig.gs.enabled ? new Url(this._stackConfig.gs).hostname : '',
```

sdk/js/src/service/gateways.js
```diff
@@ -74,11 +75,21 @@ class Gateways {
 
   // Create
 
-  async create (userId = this._defaultUserId, gateway) {
+  async create (userId = this._defaultUserId, gateway, { setDefaults = true }) {
+    let gtw = gateway
+    const Url = URL ? URL : window.URL
+
+    if (setDefaults) {
```
kschiffer: We decided to not use the any set default flag anymore in the SDK:
https://github.com/TheThingsNetwork/lorawan-stack/issues/683
The server address should rather be passed explicitly to avoid *magic*.

(It also still needs to be removed from the device service)

htdvisser:
@kschiffer I fixed that, so I guess this PR now also closes #683.

Here's the updated device view:

Screenshot 2019-07-26 at 10 20 41

Maybe the field descriptions aren't really necessary, what do you think?

htdvisser:
Making this a release blocker, since the current console creates end devices and gateways with different defaults than the CLI. We want that to be the same to avoid having to fix inconsistent databases.

pkg/webui/console/components/device-data-form/messages.js
```diff
@@ -22,6 +22,9 @@ export default defineMessages({
   deviceIdDescription: 'Unique Identifier of your device; this cannot be changed afterwards',
   deviceNameDescription: 'Human friendly name of your device for display purposes',
   deviceDescDescription: 'Optional device description; can also be used to save notes about the device',
+  nsServerAddressDescription: 'The address of the device\'s Network Server',
```
kschiffer: I'm afraid this wording might be misleading, should we rather speak of `host` / `host name`? `address` sounds more like FQDN. (Same issue for gateway creation and linking form)

sdk/js/src/service/gateways.js
```diff
@@ -79,7 +80,6 @@ class Gateways {
       routeParams: { 'collaborator.user_ids.user_id': userId },
     },
     { gateway })
-
```
kschiffer: Unnecessary change.

pkg/webui/console/views/gateway-add/index.js
```diff
@@ -91,19 +91,12 @@ export default class GatewayAdd extends React.Component {
...
-      gsServerAddress = new URL(gs.base_url).host
-    }
-
     const initialValues = {
       ids: {
         gateway_id: undefined,
       },
       enforce_duty_cycle: true,
-      gateway_server_address: gsServerAddress,
+      gateway_server_address: env.config.gs.enabled ? new URL(env.config.gs.base_url).hostname : '',
```
kschiffer:
```suggestion
      gateway_server_address: config.gs.enabled
        ? new URL(config.gs.base_url).hostname 
        : '',
```
See comment above.

sdk/js/src/service/devices/index.js
```diff
@@ -252,18 +252,8 @@ class Devices {
     return this._responseTransform(response)
   }
 
-  async create (applicationId, device, { abp = false, setDefaults = true, withRootKeys = false } = {}) {
+  async create (applicationId, device, { abp = false, withRootKeys = false } = {}) {
     let dev = device
```
bafonins:
```suggestion
    const dev = device
```

sdk/js/src/service/gateways.js
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { URL } from 'url'
```
bafonins: not needed

pkg/webui/console/views/device-add/index.js
```diff
@@ -87,6 +89,13 @@ export default class DeviceAdd extends Component {
 
   render () {
     const { error } = this.state
+    const { env } = this.props
+
+    const initialValues = {
```
kschiffer: Please refactor like this
```js
    const { env: { config }} = this.props

    const initialValues = {
      network_server_address: config.ns.enabled
        ? new URL(config.ns.base_url).hostname 
        : '',
        
        …
    }
```

pkg/webui/console/components/device-data-form/index.js
```diff
@@ -179,6 +179,13 @@ class DeviceDataForm extends Component {
     const { update } = this.props
     return (
       <React.Fragment>
+        <Form.Field
+          title={sharedMessages.joinServerAddress}
+          description={m.jsServerAddressDescription}
+          placeholder={sharedMessages.addressPlaceholder}
```
kschiffer: Please update the placeholder as well, it's still saying `host:port`

sdk/js/src/service/devices/index.js
```diff
@@ -252,18 +252,8 @@ class Devices {
     return this._responseTransform(response)
   }
 
-  async create (applicationId, device, { abp = false, setDefaults = true, withRootKeys = false } = {}) {
```
kschiffer: Please remove the `URL` import as well.

pkg/webui/console/components/device-data-form/validation-schema.js
```diff
@@ -73,6 +73,12 @@ export default Yup.object().shape({
   lorawan_phy_version: Yup.string().required(sharedMessages.validateRequired),
   frequency_plan_id: Yup.string().required(sharedMessages.validateRequired),
   supports_class_c: Yup.boolean(),
+  network_server_address: Yup.string()
+    .matches(addressRegexp, sharedMessages.validateAddressFormat),
```
kschiffer: The validation message also needs to be adapted then, it is still saying to use `host:port`

pkg/webui/console/components/device-data-form/index.js
```diff
@@ -348,6 +354,18 @@ class DeviceDataForm extends Component {
           name="supports_class_c"
           component={Checkbox}
         />
+        <Form.Field
```
kschiffer: Do you think it makes sense to put these under another sub-heading like `Connections` or is it good as part of `General Settings`?

pkg/webui/console/components/device-data-form/messages.js
htdvisser: This was already removed as per @bafonins' suggestion

sdk/js/src/service/devices/index.js
htdvisser: Already done after @bafonins' review

pkg/webui/console/components/device-data-form/index.js
htdvisser: Not really. IMO everything stored on the Identity Server is "General".

Also, it's not actually connections (that's only for the GS)

kschiffer:
> @kschiffer I fixed that, so I guess this PR now also closes #683.

Indeed.

kschiffer:
> Maybe the field descriptions aren't really necessary, what do you think?

Agreed. At least the current ones are not really helpful. If you can think of something that comparably concise but increases clarity, go for it. Otherwise, I think we should remove this.

PR #1058, opened 2019-07-25
Revoke sessions and access tokens on password change

### LINKED ISSUE #267: Revoke all User Sessions and OAuth Access Tokens on password change
**Summary:** 

Users should get the possibility to revoke all User Sessions and OAuth Access Tokens on a password change.

**Why do we need this?**

In case a user login gets compromised we need to be able to revoke access.

We already have the password reset flow.

- A `revoke_all_access` boolean in `UpdateUserPasswordRequest` to indicate that all existing sessions and access tokens should be revoked.
- Code in the IS's UpdatePassword handler to revoke sessions and access tokens if the user wants that. 

1. Add the field to `UpdateUserPasswordRequest`
2. Add code to `updateUserPassword` in the Identity Server that deletes sessions and access tokens if the field is true. You can use `FindSessions`+`DeleteSession` of the `UserSessionStore` and `ListAccessTokens`+`DeleteAccessToken` of the `OAuthStore` for this.
3. Add a `--revoke_all_access` flag to `ttn-lw-cli update-password`

I can do this

furtiman:
#### Summary

Closes #267 

#### Changes

- Added a boolean field `RevokeAllAccess` to `UpdateUserPasswordRequest`
- Added is logic to revoke tokens and sessions if the field is true
- Added boolean cli flag to `user` `update-password`

#### Notes for Reviewers

I have tested the new flag, and if it is set upon password change, the console session logs out automatically, requesting a new login.

Changing the password without the flag lets you stay in the console after changing the password.


cmd/ttn-lw-cli/commands/users.go
```diff
@@ -339,6 +341,7 @@ func init() {
 	usersUpdatePasswordCommand.Flags().AddFlagSet(userIDFlags())
 	usersUpdatePasswordCommand.Flags().String("old", "", "")
 	usersUpdatePasswordCommand.Flags().String("new", "", "")
+	usersUpdatePasswordCommand.Flags().Bool("revoke_all_access", false, "revoke all sessions and access tokens after the password is updated")
```
rvolosatovs: Shouldn't this be `true` by default?

furtiman: As I understood the issue, 

> Users should get the possibility to revoke all User Sessions and OAuth Access Tokens on a password change.

I thought this should be optional? cc @htdvisser

htdvisser: I think it's fine to leave this optional. There is no reason to revoke existing tokens unless you suspect that your credentials were compromised.

api/user.proto
```diff
@@ -109,6 +109,7 @@ message UpdateUserPasswordRequest {
   UserIdentifiers user_ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (validate.rules).message.required = true];
   string new = 2;
   string old = 3;
+  bool revoke_all_access = 4;
```
htdvisser: Can you add a comment to explain what this does and when it's needed?

pkg/identityserver/user_registry.go
```diff
@@ -437,6 +437,36 @@ func (is *IdentityServer) updateUserPassword(ctx context.Context, req *ttnpb.Upd
 			usr.TemporaryPassword, usr.TemporaryPasswordCreatedAt, usr.TemporaryPasswordExpiresAt = "", nil, nil
 			updateMask = temporaryPasswordFieldMask
 		}
+		if req.RevokeAllAccess == true {
```
htdvisser: No need to compare a boolean to `true`

cmd/ttn-lw-cli/commands/users.go
johanstokking: I think this has to be `true`.

I have yet to see the first (corporate) platform where a password change does not lead to a signout everywhere. Even in consumer services it works this way; think Gmail, iCloud, etc.

johanstokking: And yes this also goes for OAuth client. If you change your Facebook password, you will go through the consent screens of third-party apps again.

htdvisser: I'm fine with either. I guess that most users will use the `--help` on this command and consider whether they need the flag anyway.

johanstokking: > I guess that most users will use the `--help` on this command and consider whether they need the flag anyway.

I guess not. Let's change this default to a more secure default

furtiman: so then from this logic the flag would be `--do-not-revoke-access` or something like this?

johanstokking: We'd just make the default `true`

BTW, this is a flag, why are we using snake case and not dashes?

htdvisser: The flag is still `--revoke-all-access`, but it's `true` by default. You can set it to `false` with `--revoke-all-access=false`

furtiman: > BTW, this is a flag, why are we using snake case and not dashes?

I guess I confused this with the way the error messages are defined.

I will open and take over the issue to fix this, both with the default value and snake case.

htdvisser: All flags are normalized by the CLI, so flags with `_` and `-` are equivalent. I think we're mixing them in more places.

furtiman: Shall change it anyway since I will bring other changes in the same line of code, for more consistency?

I see way more dashes in the cli flags definitions, so why not bring a bit more consistency

johanstokking: >I think we're mixing them in more places.

Which one? Looks like this is the first

> Shall change it anyway since I will bring other changes in the same line of code, for more consistency?

Yes please

htdvisser: > I see way more dashes in the cli flags definitions, so why not bring a bit more consistency

Be careful, some of those may be used in field masks that expect underscores.

PR #1063, opened 2019-07-26
Redirect logged in user when trying to access Create Account page

### LINKED ISSUE #1030: User can open create account page while logged in
#### Summary

If user is logged in, they can still open the create account page and create a new account, which results in a newly created account being the one that is logged in. This is not intended I believe and leads to confusion.

#### Steps to Reproduce

1. Log into `oauth` client
2. Go to url `/oauth/register`
3. Create account

#### What do you see now?

Create account form

Redirect to for example, `/oauth`.

#### Environment

Latest master

Add check if the user is already logged in and redirect

I can do it

cc @kschiffer @bafonins

pgalic96:
#### Summary

- closes #1030 

#### Changes

- Add check if the user object is present and if it is return `Redirect` component that returns to `/oauth`

### DIFF
--- a/pkg/webui/oauth/views/create-account/index.js
+++ b/pkg/webui/oauth/views/create-account/index.js
@@ -14,7 +14,7 @@
 
 import React from 'react'
 import { Container, Col, Row } from 'react-grid-system'
-import { withRouter } from 'react-router-dom'
+import { withRouter, Redirect } from 'react-router-dom'
 import bind from 'autobind-decorator'
 import { defineMessages } from 'react-intl'
 import { connect } from 'react-redux'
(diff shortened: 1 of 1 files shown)


pkg/webui/oauth/views/create-account/index.js
```diff
@@ -91,16 +93,20 @@ const getSuccessMessage = function (state) {
   }
 }
 
-@connect(null, {
+
```
bafonins: redundant new line

pkg/webui/oauth/views/create-account/index.js
```diff
@@ -138,6 +144,29 @@ export default class CreateAccount extends React.PureComponent {
...
+          <Message content={sharedMessages.loading} />
+        </Spinner>
+      )
+    }
+
+    if (Boolean(user)) {
+      return (
+        <Redirect
+          to={{
+            pathname: `${env.app_root}`,
```
bafonins:
```suggestion
            pathname: env.app_root,
```

Issue #252, opened 2019-03-06
Improve documentation of RPC services, methods, messages, fields and events


htdvisser:
**Summary:** 

We should improve the (generated) API documentation by adding more extensive comments to our protos.

See this PR for a gold standard API comments reference -> https://github.com/TheThingsNetwork/lorawan-stack/pull/3181/files#diff-aa11ad1a94ecc7c1db3ead25d3372bdd39c6a25dd998a2902da6e4c1cb3b8fdbR129

**Why do we need this?**

This will really improve the developer experience of SDK and application developers.

**What is already there? What do you see now?**

There's already a lot of documentation on fields, but less on services, methods and messages.

**What is missing? What do you want to see?**

- [x] Improved documentation on messages (for example, explaining what the effects of field masks are in UpdateXXXRequest messages)
- [ ] Improved documentation on methods (for example, explaining that As.SetLink does not return whether the actual linking is successful, perhaps explaining how many resources a method call consumes)
- [ ] Improved documentation on services (for example, being more clear about which components expose the services, which fields are accepted by the different EndDevice registries).
- [x] Improved documentation on events that are published by components (for example, explaining that AS reports `as.link.start`, `as.link.stop` and `as.link.fail`, GS reports similar events, etc)

**How do you propose to implement this?**

I suppose this can be a long-running issue, and we can gradually add more comments when we feel like it.

**What can you do yourself and what do you need help with?**

I think it would be best if I went through the Identity Server services, methods and the related messages+fields. @rvolosatovs and @johanstokking would do the same for the different EndDevice registries, and @johanstokking would work on the GS and AS services and methods for linking/link stats.

johanstokking:
Updated OP and title to include events to the scope of this issue.

htdvisser:
@benolayinka please see what's missing and assign "experts"/"component owners" to specific tasks.

benolayinka:
@htdvisser I don't know enough about the RPC stuff to know what is missing. Can this issue be closed, leaving Roman's part open? -> https://github.com/TheThingsNetwork/lorawan-stack/issues/3167

htdvisser:
This has nothing to do with "knowing enough about the RPC stuff", you need to go through the pages under https://thethingsstack.io/reference/api/ and list the services, methods, messages and fields that need improved documentation. If you can make those improvements yourself, that would be great, and otherwise ask code owners for their expertise.

[...]

Issue #206, opened 2019-02-27
Document device registry APIs for JS, NS and AS


pdubois22:
**Summary:** 
I am trying to implement the following CLI call:

dist/linux_amd64/ttn-lw-cli devices create caf dev3 --defaults=true \
--supports-join=true \
--join-eui=70B3D59520000000 --dev-eui=A81758FFFE031BF9 \
--root-keys.app-key.key=6E2DD93BA3DF1BE62D57C33D756594E4 \
--session.dev-addr=FC006A0C \
--session.keys.app-s-key.key=2412B7B219FCEE78BDFACB5944764116 \
--session.keys.f-nwk-s-int-key.key=57A976DF984AF3EC514B21E56C081510 \
--lorawan-version=1.0.2 --lorawan-phy-version=1.0.2-b \
--frequency-plan-id EU_863_870 \
--session.last-f-cnt-up=25860 \
--session.last-n-f-cnt-down=901

From what I understand, I need at least 3 http post:
one to the is:
/api/v3/applications/xxx/devices

one to the ns:
/api/v3/ns/applications/xxx/devices

one to the js:
/api/v3/js/applications/xxx/devices

and maybe one to the as ?
one to the ns:
/api/v3/as/applications/xxx/devices

Each call requires a subset of the info above and appropriate field_mask.paths. Correct ?

**Why do we need this?**
Simplify the API call or at least a clear documentation.

**What is already there? What do you see now?**
haven't seen much.

**What is missing? What do you want to see?**
a simple / single API to create / update devices.
If not possible because of the separation of components, at least some documentation about it.

**How do you propose to implement this?**
Ideally a single / simple call : http post or put.

**Environment:**
API call

**What can you do yourself and what do you need help with?**
I am not confortable enough with the components to write a go api, but I can try to write the doc when I sorted this out / or somebody has clarified the proper way to use the API.

johanstokking:
>If not possible because of the separation of components, at least some documentation about it.

We should document this better; a documentation system on the way.

Indeed, this stack is designed for distribution, with potentially different domains for JS, NS and AS. Only the client (API, CLI, Console) and ER (entity registry, exposed by IS) know what the JS, NS and AS endpoints are, but only the client has access in terms of security context as well as network access. For example, we can't call AS device registry from the community ER if that AS is hosted behind a firewall.

We are making this much easier for developers in the SDKs.

johanstokking:
[...]

PR #1068, opened 2019-07-29
Auto generate API reference in doc

Sypheos:
#### Summary
Generate API reference from protos into the documentation content tree.
ref #206 #252 
...

#### Changes
`mage proto:markdown` now generate markdown doc for services and message into `/doc/content/reference`
Added a reference index to the doc.
- ...

- ...

#### Notes for Reviewers
Draft PR.
Still deciding on the format. Choose to go for one Service and one Message page.
I'm noob when it comes to html and js so suggestion are welcomed.
...


htdvisser:
I think we can do better.

As I already told you before, please let me know what exactly we need, and I'll see how we can adapt protoc-gen-doc to achieve it.

Sypheos:
Without taking account of the current file nomenclature, I would go for something like that.
* Authentication `!`
  * Rights
  * Oauth
  * Clients
* Errors
* Identifiers
* Pagination (we have some right ?) `!`
* Core Entities (with related `_services`) `!`
  * User
  * Organisation
  * End Devices
  * Application
  * Gateways
  * Messages 
* Services `!`
... All services except `_services`. Each service describe the message defined in his file(s).  
* ... Left over files.

Pages with a `!` should be written by a human to provide some higher level explanation.

The problem here is to give each page a section and front matter when generating them.

Also making links to other pages will require to have the path of the page so when template used to generate the page should have access all API resources. I couldn't manage to generate only one page for one proto file while having the information from other files without writing a template per page and adding an unhealthy number of control structure.
So if it possible putting the Files list in a map instead of an array would be nice.

Issue #905, opened 2019-07-02
Fix nonrequired byte input validation in forms


bafonins:
#### Summary

Nonrequired byte inputs are not properly validated in device/gateway forms.

#### Steps to Reproduce

1. Go to the add gateway page
2. Fill in required fields
3. Focus on `Gateway EUI` field and input any character that does not match `/[0-9a-f]/i` or add any valid characters and clear the field.
4. Submit form
5. See the error while the field is empty

<img width="658" alt="Screenshot 2019-07-02 at 13 04 09" src="https://user-images.githubusercontent.com/16374166/60508089-f1c3a080-9cc9-11e9-98f9-e41147a10a04.png">

#### What do you see now?

Error on nonrequired, empty field

#### What do you want to see instead?

No errors

#### Environment

browser (console)

#### How do you propose to implement this?

Either extend the validation for those fields using `yup` or add functionality to distinguish between the empty string and `undefined`

#### Can you do this yourself and submit a Pull Request?

Yes

bafonins:
It seems that fixing only the `ByteInput` component is not enough. The main issue is the way we process nested values in forms. Once any deeply nested value gets updated and then cleared, it leaves the entry in form state, for example in the device form:

1. Fill in the `NwkKey` field
2. See the form state:
```
values: {
  ...,
  root_keys: {
    nwk_key: {
      key: "DDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD"
    }
  }
}
```
3. Empty the field:
```
values: {
  ...,
  root_keys: {
    nwk_key: {}
  }
}
```
In this case the `root_keys` entry would fail field mask validation

bafonins:
fixed

PR #1075, opened 2019-07-31
Fix validation rules for device form

bafonins:
#### Summary

References #905, see my comment https://github.com/TheThingsNetwork/lorawan-stack/issues/905#issuecomment-512876391

#### Changes

- Split update/create validation schemas
- Cast final form values before submitting

#### Notes for Reviewers

The main idea is to strip key values in the create form and skip their validation.
For the edit form we keep validation logic as it was before (the user has to specify the keys to pass validation)

### DIFF
--- a/pkg/webui/console/components/device-data-form/index.js
+++ b/pkg/webui/console/components/device-data-form/index.js
@@ -32,7 +32,7 @@ import errorMessages from '../../../lib/errors/error-messages'
 import { getDeviceId } from '../../../lib/selectors/id'
 import PropTypes from '../../../lib/prop-types'
 import m from './messages'
-import validationSchema from './validation-schema'
+import { createFormValidationSchema, updateFormValidationSchema } from './validation-schema'
 
 @bind
 class DeviceDataForm extends Component {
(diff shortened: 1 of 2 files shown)


bafonins:
> If I understand correctly, this should (also) take care of being able to remove byte values in the edit form, right?

I would say it is out of scope of this PR. Editing keys in js is tricky. Basically you have either change or delete all of them. Consider this:

```
// after creation
root_keys: {
  nwk_key: {
    key: "some valid key here"
  },
  app_key: {
    key: "some valid key here"
  }
}
```

If you want to delete one but keep the other one, options to submit are:
```
// after creation
root_keys: {
  nwk_key: {
    key: "" // with this js will generate the default key which is '00000....'
  },
  app_key: {
    key: "some valid key here"
  }
}
// or
root_keys: {
  app_key: {
    key: "some valid key here" // this will update only the app_key, but keep the previous nwk_key
  }
}
// this is will fail field mask validation
root_keys: {
  nwk_key: {},
  app_key: {
    key: "some valid key here"
  }
}
// this is will fail validation of the key
root_keys: {
  nwk_key: {
    key: null,
  },
  app_key: {
    key: "some valid key here"
  }
}
```
There was a brief discussion on this, but no results yet.

PR #1077, opened 2019-07-31
gs: Use Basic Station EUI as main identifier

### LINKED ISSUE #1076: Use Basic Station EUIs as main identifiers instead of IDs.
#### Summary

Use Basic Station EUIs as main identifiers instead of IDs.

#### Why do we need this?

Basic Station gateways are identified based on EUIs and not IDs and hence need to be handled similar to UDP.

Ref: [tti-issue](https://github.com/TheThingsIndustries/lorawan-stack/pull/1683)

The ID request URI is used as GatewayID for processing.

The ID request URI should be converted to an EUI in before processing.

#### Environment

`ttn-lw-stack v3.1.0`

Convert ID to EUI before processing.

Yes

KrishnaIyer:
#### Summary

Closes #1076 

#### Changes

- Convert ID from request to EUI before usage.
- Adapt Auth check.

### DIFF
--- a/pkg/gatewayserver/io/basicstationlns/basicstationlns.go
+++ b/pkg/gatewayserver/io/basicstationlns/basicstationlns.go
@@ -16,9 +16,11 @@ package basicstationlns
 
 import (
 	"context"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"math"
+	"regexp"
 	"strings"
 	"sync/atomic"
 	"time"
@@ -34,6 +36,7 @@ import (
 	"go.thethings.network/lorawan-stack/pkg/gatewayserver/io/basicstationlns/messages"
 	"go.thethings.network/lorawan-stack/pkg/log"
 	"go.thethings.network/lorawan-stack/pkg/ttnpb"
+	"go.thethings.network/lorawan-stack/pkg/types"
 	"go.thethings.network/lorawan-stack/pkg/unique"
 	"go.thethings.network/lorawan-stack/pkg/web"
 	"go.thethings.network/lorawan-stack/pkg/web/middleware"
@@ -46,6 +49,7 @@ var (
 		"listener",
 		"failed to serve Basic Station frontend listener",
 	)
+	errGatewayID = errors.DefineInvalidArgument("invalid_gateway_id", "invalid gateway id `{id}`")
 )
 
 type srv struct {
@@ -131,12 +135,14 @@ func (s *srv) handleDiscover(c echo.Context) error {
 	if c.IsTLS() {
 		scheme = "wss"
 	}
+
+	euiWithPrefix := fmt.Sprintf("eui-%s", ids.EUI.String())
 	res := messages.DiscoverResponse{
 		EUI: req.EUI,
 		Muxs: basicstation.EUI{
 			Prefix: "muxs",
 		},
-		URI: fmt.Sprintf("%s://%s%s", scheme, c.Request().Host, c.Echo().URI(s.handleTraffic, ids.GatewayID)),
+		URI: fmt.Sprintf("%s://%s%s", scheme, c.Request().Host, c.Echo().URI(s.handleTraffic, euiWithPrefix)),
 	}
 	data, err = json.Marshal(res)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -181,7 +189,19 @@ func (s *srv) handleTraffic(c echo.Context) error {
 		"remote_addr", c.Request().RemoteAddr,
 	))
 
-	ctx, ids, err := s.server.FillGatewayContext(ctx, ttnpb.GatewayIdentifiers{GatewayID: id})
+	// Convert the ID to EUI.
+	str := euiHexPattern.FindAllString(id, -1)
```
htdvisser: Why not `FindString`?

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -181,7 +189,19 @@ func (s *srv) handleTraffic(c echo.Context) error {
...
+	if len(str) != 2 || len(str[1]) != 16 {
+		return errGatewayID.WithAttributes("id", id)
+	}
+	hexValue, err := hex.DecodeString(str[1])
+	if err != nil {
+		return errGatewayID.WithAttributes("id", id)
+	}
+	eui := types.EUI64{}
```
htdvisser: I thought we preferred `var eui types.EUI64` in this situation? cc: @rvolosatovs?

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -152,25 +156,29 @@ func (s *srv) handleDiscover(c echo.Context) error {
 	return nil
 }
 
+var euiHexPattern = regexp.MustCompile("[a-f0-9A-F]+")
```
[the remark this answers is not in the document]
KrishnaIyer: the request is like `eui-58a0cbfffe800cbf`. The first `e` in `eui` is a also a match so the length is checked once extracted.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
KrishnaIyer: Same as above. `FindString` exits on the first `e`.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
KrishnaIyer: Ok updated. I'm neutral on this.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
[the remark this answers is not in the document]
htdvisser: Then shouldn't we change the regex to `^eui-([a-f0-9]{16})$`?

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
htdvisser: If you change the regex as suggested above, this should become `FindStringSubmatch`

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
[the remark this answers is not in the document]
KrishnaIyer: That would also match `eui-58a0cbfffe800cbf`. The winner is  matching the substring with `[a-f0-9A-F]{16}`. 
(Note to self: Don't write regexes at the end of the day)

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -152,25 +156,29 @@ func (s *srv) handleDiscover(c echo.Context) error {
...
 }
 
+var euiHexPattern = regexp.MustCompile("[a-f0-9A-F]{16}")
+
 func (s *srv) handleTraffic(c echo.Context) error {
 	var sessionID int32
 	id := c.Param("id")
 	auth := c.Request().Header.Get(echo.HeaderAuthorization)
 	ctx := c.Request().Context()
+	md := metadata.MD{}
```
rvolosatovs: If you apply https://github.com/TheThingsNetwork/lorawan-stack/pull/1077#discussion_r309272681 then it makes sense to do the same here.

KrishnaIyer:
@htdvisser: Can you please merge this? I can't do it without @johanstokking's approval.

PR #1078, opened 2019-07-31
Fix validation rules for gateway form

bafonins:
#### Summary

References #905

#### Changes

- Change validation rules for the gateway eui field

#### Notes for Reviewers

Approach in this pr is different than in https://github.com/TheThingsNetwork/lorawan-stack/pull/1075 because:
1. `null` is valid value for the gateway eui. We cant do that for device keys, because the js would return an error.
2. api allows users to change/remove eui, we have to reflect this in the form.


bafonins:
> Try the following

fixed

Issue #938, opened 2019-07-04
Missing headers from Webhook configuration


htdvisser:
#### Summary

We can't set headers (such as Authorization) when configuring webhooks

#### Steps to Reproduce

1. Go to Application > Integrations > Add

#### What do you see now?

![image](https://user-images.githubusercontent.com/181308/60672980-f5e0f100-9e76-11e9-8268-3ca92f84799d.png)

#### What do you want to see instead?

Some way to configure the request headers

#### How do you propose to implement this?

Leaving that to @kschiffer

johanstokking:
Even though important; let's not block `v3.1.0` for this. Console is really "community technology preview"

htdvisser:
Same for CLI

htdvisser:
Added to CLI in #1002.

adriansmares:
Blocked by #1072

PR #1087, opened 2019-08-01
Make allowed field masks part of the JS SDK

### LINKED ISSUE #1072: Wrong field mask for headers
#### Summary

The field mask used for requests that provide `headers` is broken, in a sense that it attempts to specify the keys of the field as part of the field mask, even though they are not supposed to be specified. I think this would also apply to `attributes` (device, application etc.).

#### Steps to Reproduce

1. Create a Webhook creation request. I've used the following object:
```
{
  "ids": {
    "application_ids": {},
    "webhook_id": "wh1"
  },
  "base_url": "http://example.com/webhooks",
  "format": "json",
  "headers": {
    "k1": "v1",
    "k2": "v2"
  },
  "uplink_message": null,
  "join_accept": null,
  "downlink_ack": null,
  "downlink_nack": null,
  "downlink_sent": null,
  "downlink_failed": null,
  "downlink_queued": null,
  "location_solved": null
}
```
2. Attempt to `api.application.webhooks.update(appId, webhookId, patch)` (I've encountered this inside `ApplicationIntegrationAdd.handleSubmit`).

#### What do you see now?

The request will look as follows:
```
{
	"webhook": {
		"ids": {
			"application_ids": {},
			"webhook_id": "wh2"
		},
		"base_url": "http://example.com/webhooks",
		"format": "json",
		"headers": {
			"k1": "v1",
			"k2": "v2"
		},
		"uplink_message": null,
		"join_accept": null,
		"downlink_ack": null,
		"downlink_nack": null,
		"downlink_sent": null,
		"downlink_failed": null,
		"downlink_queued": null,
		"location_solved": null
	},
	"field_mask": {
		"paths": ["ids", "ids.application_ids", "ids.application_ids.application_id", "ids.webhook_id", "base_url", "format", "headers", "headers.k1", "headers.k2", "uplink_message", "join_accept", "downlink_ack", "downlink_nack", "downlink_sent", "downlink_failed", "downlink_queued", "location_solved"]
	}
}
```
The important part being `"headers", "headers.k1", "headers.k2", `.

Only the `headers` field should be specified.

#### Environment

196e6e66591b1a6482c7c3375f213b5de310fe59

I think that fields such as `headers` and `attributes` shouldn't get parsed recursively when it comes to field masks. I think https://github.com/TheThingsNetwork/lorawan-stack/blob/196e6e66591b1a6482c7c3375f213b5de310fe59/sdk/js/src/util/marshaler.js#L113-L121 needs to skip these nodes. Thoughts on this @kschiffer ?

Yes.

adriansmares:
#### Summary

Closes #1072 
References #938

#### Changes

- Added a new mage target that exports `ttnpb.AllowedFieldMaskPathsForRPC` to a json file for the JS SDK.
- Changed the `api-definitions.json` generator to also include the allowed field masks next to the RPC methods.
- Added a way to filter field masks generated from object patches.

#### Notes for Reviewers

While I really liked @htdvisser solution from [the issue](https://github.com/TheThingsNetwork/lorawan-stack/issues/1072#issuecomment-517163102), unfortunately it does not scale really well with the automatic code generation - you would need to include the type of the message in every service definition in the JS SDK. 
This would have been fine, but our field masks are actually not based on the allowed field mask for the request objects (i.e. `SetApplicationWebhookRequest`) but rather the actual object that the request is mutating (i.e. `ApplicationWebhook`). There are also some cases (such as API keys) where you cannot actually query all of the fields that are present in the message. In order to avoid all of these edge cases, I consider that exporting the allowed field mask paths from the `ttnpb` package is more appropriate.


.mage/go.go
```diff
@@ -276,3 +276,8 @@ nextLine:
 func (g Go) Messages() error {
 	return execGo("run", "./cmd/internal/generate_i18n.go")
 }
+
+// AllowedFieldMaskPaths builds the allowed field masks file based on the ttnpb package.
+func (g Go) AllowedFieldMaskPaths() error {
```
htdvisser: I think this belongs under `jsSDK`. We're not making Go but JS code.

cmd/internal/generate_allowed_field_masks.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+
+	"go.thethings.network/lorawan-stack/pkg/ttnpb"
+)
+
+func main() {
+	messagesFile := "sdk/js/generated/allowed-field-masks.json"
```
htdvisser: It should be "allowed field mask paths" (also in filename of this cmd)

sdk/js/generated/allowed-field-masks.json
```diff
@@ -0,0 +1,1136 @@
+{
```
htdvisser: Do we actually need this file to be added to Git?

adriansmares: Since it is rarely edited, I believe we would be saving some build time by embedding it. But it can be removed and be made part of the JS SDK build pipeline.

.mage/js_sdk.go
```diff
@@ -123,20 +123,24 @@ func (k JsSDK) Clean() {
 }
 
 // CleanDeps removes all installed node packages (rm -rf node_modules).
-func (j JsSDK) CleanDeps() {
+func (k JsSDK) CleanDeps() {
```
htdvisser: Can also completely remove `k` if it's not needed

.mage/js_sdk.go
```diff
@@ -149,6 +153,10 @@ func (k JsSDK) Definitions() error {
 
 // DefinitionsClean removes the generated api-definition.json.
 func (k JsSDK) DefinitionsClean(context.Context) error {
+	err := sh.Rm(filepath.Join("sdk", "js", "generated", "allowed-field-mask-paths.json"))
+	if err != nil {
+		return err
```
htdvisser: Please make sure that we don't return on a "does not exist" and "forget" to delete api-definition.json

sdk/js/util/http-mapper.js
```diff
@@ -29,7 +30,11 @@ function map (files) {
       result[service.name] = {}
 
       for (const method of service.methods) {
-        result[service.name][method.name] = { file: file.name, http: []}
+        result[service.name][method.name] = {
+          file: file.name,
+          http: [],
+          allowed_field_masks: allowedFieldMaskPaths[`/${service.fullName}/${method.name}`],
```
kschiffer: I'd prefer lower camel case `allowedFieldMasks` here, which we use by default in the JS codebase (if not forced otherwise).

.mage/js_sdk.go
adriansmares: Tested now - it doesn't return error on not found and continues even if the file is missing.

PR #1088, opened 2019-08-01
Add webhook headers support to the console

### LINKED ISSUE #938: Missing headers from Webhook configuration

adriansmares:
#### Summary

Closes #938 

#### Changes

- Add headers as part of the creation / update form of the Webhooks


pkg/webui/components/key-value-map/entry.js
```diff
@@ -0,0 +1,90 @@
...
+
+import PropTypes from '../../lib/prop-types'
+import Input from '../input'
+import Button from '../button'
+
+@bind
+class Entry extends React.Component {
+
+  handleRemoveButtonClicked (event) {
+    this.props.onRemoveButtonClick(this.props.index, event)
```
kschiffer: Nitpick: we usually use destructured assignment for props.
```js
const { onRemoveButtonClick, index } = this.props
onRemoveButtonClick(index, event)
```

pkg/webui/components/key-value-map/entry.js
```diff
@@ -0,0 +1,90 @@
...
+    this.props.onChange(this.props.index, { key: newKey })
+  }
+
+  handleValueChanged (newValue) {
+    this.props.onChange(this.props.index, { value: newValue })
+  }
+
+  render () {
+    return (
+      <Row key={this.props.index}>
```
kschiffer: I think using the grid container is not appropriate here. You sould rather use a container with `display: flex` and `flex-direction: column` and then size the individual elements using percentage values for `width`.

kschiffer: If you want I can make a PR for this branch to take care of this.

pkg/webui/components/key-value-map/index.js
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  render () {
+    const {
+      className,
+      name,
+    } = this.props
+
+    return (
+      <Container className={classnames(className, this.props)} style={{ paddingLeft: '0px' }}>
```
kschiffer: See my other comment on avoiding the grid container. It should only be used for top-level layouting concerns.

pkg/webui/components/key-value-map/index.js
```diff
@@ -0,0 +1,90 @@
...
+      </Container>
+    )
+  }
+}
+
+KeyValueMap.propTypes = {
+  className: PropTypes.string,
+  name: PropTypes.string.isRequired,
+  value: PropTypes.oneOfType([ PropTypes.array, PropTypes.object ]),
+  keyPlaceholder: PropTypes.oneOfType([ PropTypes.object, PropTypes.string ]).isRequired,
```
kschiffer:
```suggestion
  keyPlaceholder: PropTypes.message.isRequired,
```

pkg/webui/components/key-value-map/index.js
```diff
@@ -0,0 +1,90 @@
...
+  valuePlaceholder: PropTypes.oneOfType([ PropTypes.object, PropTypes.string ]).isRequired,
```
kschiffer:
```suggestion
  valuePlaceholder: PropTypes.message.isRequired,
```

pkg/webui/components/key-value-map/index.js
```diff
@@ -0,0 +1,90 @@
...
+        </Row>
+      </Container>
+    )
+  }
+}
+
+KeyValueMap.propTypes = {
+  className: PropTypes.string,
+  name: PropTypes.string.isRequired,
+  value: PropTypes.oneOfType([ PropTypes.array, PropTypes.object ]),
```
kschiffer: Can it really be both an array of arrays and objects?

pkg/webui/components/key-value-map/index.js
```diff
@@ -0,0 +1,90 @@
...
+  keyPlaceholder: PropTypes.oneOfType([ PropTypes.object, PropTypes.string ]).isRequired,
+  valuePlaceholder: PropTypes.oneOfType([ PropTypes.object, PropTypes.string ]).isRequired,
+}
+
+KeyValueMap.defaultProps = {
+  value: [],
+}
+
+KeyValueMap.Entry = Entry
```
kschiffer: Since this is a very scoped child component, I don't think it needs to be exposed.

pkg/webui/components/key-value-map/index.js
```diff
@@ -0,0 +1,90 @@
...
+            keyPlaceholder={this.props.keyPlaceholder}
+            valuePlaceholder={this.props.valuePlaceholder}
+            index={index}
+            onRemoveButtonClick={this.removeEntry}
+            onChange={this.handleEntryChange}
+          />
+        ))}
+        <Row style={{ paddingTop: '4px' }}>
+          <Col>
+            <Button
```
kschiffer: Let's make the button more descriptive by adding an `Add header entry` message as well.

pkg/webui/components/key-value-map/entry.js
```diff
@@ -0,0 +1,90 @@
...
+            value={this.props.value.value}
+            required
+          />
+        </Col>
+        <Col sm={1}>
+          <Button
+            name={`${this.props.name}[${this.props.index}].remove`}
+            type="button"
+            onClick={this.handleRemoveButtonClicked}
+            icon="delete"
```
kschiffer:
```suggestion
            icon="delete"
            danger
```
Will change the button color to red.

pkg/webui/console/components/webhook-form/index.js
```diff
@@ -60,6 +49,34 @@ const m = defineMessages({
   deleteWebhook: 'Delete Webhook',
   modalWarning:
     'Are you sure you want to delete webhook "{webhookId}"? Deleting a webhook cannot be undone!',
+  headers: 'Headers',
+  headersKeyPlaceholder: 'Authorization',
+  headersValuePlaceholder: 'Token xyz',
```
bafonins: `xyz` ? please define something more appropriate

pkg/webui/console/components/webhook-form/index.js
```diff
@@ -60,6 +49,34 @@ const m = defineMessages({
...
+  headersAdd: 'Add header entry',
+  headersValidateRequired: 'All header entry values are required. Please remove empty entries.',
+})
+
+const headerCheck = headers =>
```
bafonins: please decompose this to multiple `yup` validations

pkg/webui/components/button/index.js
```diff
@@ -71,8 +73,22 @@ class Button extends React.PureComponent {
...
+      disabled,
+      name,
+      type,
+      value,
+      title: rawTitle,
+      intl,
+    } = this.props
+
+    let title = rawTitle
+    if (rawTitle && typeof rawTitle === 'object') {
```
bafonins:
```suggestion
    if (rawTitle && typeof rawTitle === 'object' && rawTitle.id && rawTitle.defaultMessage) {
```

pkg/webui/components/key-value-map/entry.js
```diff
@@ -0,0 +1,106 @@
...
+      name,
+      index,
+      keyPlaceholder,
+      valuePlaceholder,
+      value,
+      onBlur,
+    } = this.props
+
+    return (
+      <div className={style.entriesRow} key={index}>
```
bafonins:
```suggestion
      <div className={style.entriesRow}>
```

key not needed here

pkg/webui/components/key-value-map/entry.js
```diff
@@ -0,0 +1,106 @@
...
+          onBlur={onBlur}
+          value={value.value}
+        />
+        <Button
+          name={`${name}[${index}].remove`}
+          type="button"
+          onClick={this.handleRemoveButtonClicked}
+          icon="delete"
+          title={m.deleteEntry}
+          onBlur={onBlur}
```
bafonins: any reasons to react on `blur` events on the button?

pkg/webui/components/key-value-map/entry.js
```diff
@@ -0,0 +1,106 @@
...
+          className={style.input}
+          name={`${name}[${index}].value`}
+          placeholder={valuePlaceholder}
+          type="text"
+          onChange={this.handleValueChanged}
+          onBlur={onBlur}
+          value={value.value}
+        />
+        <Button
+          name={`${name}[${index}].remove`}
```
bafonins: do you use the name anywhere? you already pass a separate event handler for the button, why adding the name?

pkg/webui/components/key-value-map/index.js
```diff
@@ -0,0 +1,106 @@
...
+      index !== i ? kv : { ...kv, ...newValues }
+    )))
+  }
+
+  removeEntry (index) {
+    const { onChange, onBlur, value } = this.props
+    onChange(value.filter((_, i) => (i !== index)) || [])
+
+    // Trigger also the onBlur event to enforce revalidation
+    onBlur()
```
bafonins: seems more like a hack to me, `<KeyValueMap />` knows nothing about validation.
also, `onBlur` is not in the prop-types

pkg/webui/components/key-value-map/entry.js
bafonins: `onBlurz` is not in prop-types

pkg/webui/console/components/webhook-form/index.js
kschiffer: I decided against that as it would cause the validation error object for the field to be an array of objects as well.
That would mean that we would have to either 
- pass down the error into the component in order to display it correctly. With the way we are handling form fields now, the error is shown on field level, not input component level.
- apply some logic to map the nested validation error back to an overall field error

Let me know if you have an idea to do this better.

pkg/webui/components/key-value-map/index.js
kschiffer: Any idea to achieve this cleaner?

bafonins:
Will add several adjustments in a separate PR

PR #1093, opened 2019-08-02
Prevent browser auto-prefill of linking form

kschiffer:
#### Summary
This PR fixes an issue where the login credentials will be auto-prefilled by (at least) chrome:
image

Chrome appears to assume any form consisting of a text input followed by a password input to be a login form. Circumventing this behavior via workaround appears to be [very dirty and instable](https://gist.github.com/runspired/b9fdf1fa74fc9fb4554418dea35718fe). I hence decided to change the type of the Api Key field from `password` to `text`, which is also more correct semantically. 
Later, we should add a toggle to show/hide the value.

#### Changes
- Add a `code` prop to the `<Input />` component that will adjust the styling of the input to have monospace font
- Change the input type of the `api_key` field from `password` to `text`

### DIFF
--- a/pkg/webui/components/input/index.js
+++ b/pkg/webui/components/input/index.js
@@ -48,6 +48,7 @@ class Input extends React.Component {
     label: PropTypes.string,
     loading: PropTypes.bool,
     title: PropTypes.message,
+    code: PropTypes.bool,
   }
 
   static defaultProps = {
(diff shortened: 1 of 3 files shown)


pkg/webui/components/input/index.js
```diff
@@ -102,6 +102,7 @@ class Input extends React.Component {
       title,
       intl,
       horizontal,
+      code,
```
bafonins: please extend  prop-types

kschiffer:
> > Later, we should add a toggle to show/hide the value.
> 
> please file an issue

#1095

PR #1096, opened 2019-08-02
Use interop clients in components for JS lookups

rvolosatovs:
#### Summary

References #4 
Actual implementation for the `InteropClient`s required by components comes in #1004

#### Changes

- Use `InteropClient` interfaces in components for JS lookups
- Remove TODO about lookups in JS


pkg/applicationserver/applicationserver.go
```diff
@@ -401,21 +403,30 @@ func (as *ApplicationServer) DownlinkQueueList(ctx context.Context, ids ttnpb.En
...
+		res, err := ttnpb.NewAsJsClient(js.Conn()).GetAppSKey(ctx, req, as.WithClusterAuth())
+		if err == nil {
+			return res.AppSKey, nil
+		}
+		if !errors.IsNotFound(err) {
+			return ttnpb.KeyEnvelope{}, err
+		}
+	}
+	if as.interopClient != nil {
+		// TODO: Set AS-ID.
```
rvolosatovs: Not sure what to put here - @htdvisser @johanstokking ?

htdvisser: AFAIK the AS-ID is just the hostname of the AS, but let's wait for @johanstokking

rvolosatovs:
Will reopen when ready

pkg/applicationserver/applicationserver.go
johanstokking: I think this should come from config, I'm afraid.

Yes, TTN JS uses the host of `application_server_address` as AS-ID.

Other (non-TTN) JS issue AS IDs, and they may contain characters that are not valid in hosts, i.e. characters that do not pass the regex we have on our `*_address` fields. Still, TTN AS should identify itself with that AS ID, and since it can be something arbitrary from a JS, I think we should add this to AS config.

rvolosatovs:
Added everything in https://github.com/TheThingsNetwork/lorawan-stack/pull/1004, since the implementation became more interwined

PR #1098, opened 2019-08-02
dev: Generate allowed field mask paths before definitions

### LINKED ISSUE #1094: Generate allowed field mask paths on webui build
#### Summary

`allowed-field-mask-paths.json` should be generated on `js:build` or `jsSDK:build` targets. Or `mage jsSDK:allowedFieldMaskPaths jsSdk:definitions` should be added to the docume tation.

#### Why do we need this?

The js sdk/webui build should take care of all build deps

`js:build` `jsSDK:build` targets

`mage jsSDK:allowedFieldMaskPaths jsSdk:definitions` being run when building the sdk/webui

i can, but will leave to the author of https://github.com/TheThingsNetwork/lorawan-stack/pull/1087

adriansmares:
#### Summary

Closes #1094 

#### Changes

- Ensure that the `allowed-field-mask-paths.json` is generated before using it as a source in the `jsSDK:definitions` target.
- Make the generation smarter by using the corresponding ttnpb file as a source for `allowed-field-mask-paths.json`

### DIFF
--- a/.mage/js_sdk.go
+++ b/.mage/js_sdk.go
@@ -129,7 +129,7 @@ func (JsSDK) CleanDeps() {
 
 // Definitions extracts the api-definition.json from the proto generated api.json.
 func (k JsSDK) Definitions() error {
-	mg.Deps(Proto.JsSDK)
+	mg.Deps(Proto.JsSDK, JsSDK.AllowedFieldMaskPaths)
 	changed, err := target.Path(
 		filepath.Join("sdk", "js", "generated", "api-definition.json"),
 		filepath.Join("sdk", "js", "generated", "api.json"),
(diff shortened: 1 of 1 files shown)


.mage/js_sdk.go
```diff
@@ -187,5 +186,13 @@ func (k JsSDK) Link() error {
 
 // AllowedFieldMaskPaths builds the allowed field masks file based on the ttnpb package.
 func (k JsSDK) AllowedFieldMaskPaths() error {
+	changed, err := target.Dir(filepath.Join("sdk", "js", "generated", "allowed-field-mask-paths.json"),
```
htdvisser: Shouldn't this be `target.Path`?

PR #1108, opened 2019-08-05
Generate API ref in hugo format

Sypheos:
#### Summary
The markdown reference now generate markdown files who are added to the Hugo file tree.

#### Changes
* `mage protos:markdown` and `mage protos:markdownclean`  now source the file `.mage/proto_ref_tree.yaml` to identify the file to generate and clean.
* `.mage/proto_ref_tree.yaml` contains all information to generate the files in the hugo file tree as well as the font matter.
* API references are put under the `doc/content/reference/api/`.
* Add the hugo shortcode `reftab` and `refswitcher` to create content tabs.
* Remove the `api.md.tmpl` in favor of `applicatons.md.tmpl`

#### Notes for Reviewers
`mage docs:server` to visualize.

WIP, missing most of the reference. Experimenting with the layout and the file tree. Gathering feedback.
Link to the file tree: https://github.com/TheThingsNetwork/lorawan-stack/pull/1068#issuecomment-517234169
Still experimenting with how to organize the content and see which one fits more. Content is pretty irrelevant now. Please review on how the process on generating the docs, file tree and page layout. 

@htdvisser Can we pass the new `HugoPage` structure to `protoc-gen-doc` when it calls the template ? That way we can reduce the number of template and work to maintain that doc.
As of now we need one template per page, not really scalable but we are talking about ref doc so once the initial writing is done there shouldn't be much to do. Unless new services or sections appear.

**Todo before full PR**:
- [ ] Nest templates where possible.
- [ ] Generate the ref according to the file tree
- [ ] hide the HTTP methods by default
- [ ] Clean files
- [ ] Pass PR test


htdvisser:
I didn't change my mind since https://github.com/TheThingsNetwork/lorawan-stack/pull/1068. 

> I think we can do better.
> 
> As I already told you before, please let me know what exactly we need, and I'll see how we can adapt protoc-gen-doc to achieve it in a clean way.

We're not going to write templates like `api/application.md.tmpl` for each page we want to build, each containing an `if` statement to select the right proto file. Doing this will also encourage us to adapt the structure of our proto files to make them fit the documentation better, which we definitely shouldn't do.

Over the weekend I've been doing some research to see what's the cleanest way to generate flexible documentation from our protos in a clean way. Please give me some more time and refer to the existing API.md until that's ready.

Sypheos:
If we can pass additional elements (in `proto_ref_tree.yaml`) to protoc-gen-doc then it will only take a handful of templates while keeping the same proto.

I wasn't clear, in the current situation where we cannot pass additional information to the generator we need one template per page. If we can manage to pass a simple structure or configuration file to the generator we can easily do it with only one template per layout.
So can we add this kind of feature to the generator ? Only relying on the protoc file fo r information will inevitably lead us to modify it.

We could add the front matter in the proto file but I doubt we want it there.

Issue #820, opened 2019-06-17
Fill issued JoinEUIs and DevAddrs on Device create


kschiffer:
#### Summary
With #661 #758 being implemented, we can use the new RPCs to obtain the JoinEUI prefixes of the JS configuration and get a DevAddr from the NS. In the console this info should be used to prefill the value of the JoinEUI and DevAddr fields.

#### Why do we need this?
Reduce input errors and assist users when creating devices.

#### What is already there? What do you see now?
- The [`Js.GetJoinEUIPrefixes` RPC](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/api/api.md#service-js)
- The [`Ns.GenerateDevAddr` RPC](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/api/api.md#service-ns)
- The "Add new end device" form in the console

#### What is missing? What do you want to see?
Functionality to fetch the JoinEUIs and prefill the value in the form.

#### Environment
Console, js SDK

#### How do you propose to implement this?
A little button in the form field that can be clicked to fill an issued JoinEUI/DevAddr.

#### Can you do this yourself and submit a Pull Request?
Yes.

bafonins:
taking over this one, since working on improving the device form - https://github.com/TheThingsNetwork/lorawan-stack/issues/579

PR #1113, opened 2019-08-06
Add js service to js sdk

bafonins:
#### Summary

References #820 

#### Changes

- Add `Js` service in the JS SDK
- Add `api.js.joinEUIPRefixes.list` entry in the console api defintions

You can use [this script](https://gist.github.com/bafonins/d409e23958b313cd6769b008e60bff91) for testing

#### Release Notes

- Added the `Js` service to the JS SDK
- Added the `listJoinEUIPrefixes` method in the `Js` service

### DIFF
--- a/sdk/js/src/service/join-server.js
+++ b/sdk/js/src/service/join-server.js
@@ -21,9 +21,8 @@ class Js {
 
   async listJoinEUIPrefixes () {
     const result = await this._api.GetJoinEUIPrefixes()
-    const { prefixes } = Marshaler.payloadListResponse('prefixes', result)
 
-    return prefixes
+    return Marshaler.payloadListResponse('prefixes', result)
   }
 }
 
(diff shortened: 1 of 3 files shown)


sdk/js/src/service/join-server.js
```diff
@@ -0,0 +1,30 @@
...
+import Marshaler from '../util/marshaler'
+
+class Js {
+  constructor (service) {
+    this._api = service
+  }
+
+  async listJoinEUIPrefixes () {
+    const result = await this._api.GetJoinEUIPrefixes()
+    const { prefixes } = Marshaler.payloadListResponse('prefixes', result)
```
kschiffer: Any reason to no simply return the marshaled value, like we do in other services?
```suggestion
    return Marshaler.payloadListResponse('prefixes', result)
```

bafonins: we are not paginating prefixes, there is no reason to return the `totalCount` entry. even we decide to implement pagination for this, it will be an api change

kschiffer: True. We should refactor the configuration service then as well.

bafonins: Actually, it makes sense to include the `totalCount` entry if we do this for the configuration service

PR #1132, opened 2019-08-08
Prefill join EUI prefix in device form

### LINKED ISSUE #820: Fill issued JoinEUIs and DevAddrs on Device create

bafonins:
#### Summary

References #820 

1. device create form:

Screenshot 2019-08-08 at 15 35 07
Screenshot 2019-08-08 at 15 35 17
Screenshot 2019-08-08 at 15 35 28
Screenshot 2019-08-08 at 15 40 31
Screenshot 2019-08-08 at 15 41 03

2. device edit form:

Screenshot 2019-08-08 at 15 42 27

#### Changes

- Add `JoinEUIPrefixesInput`
- Add `js` redux primitives
- Extend `Input` and `Byte` with the `focus` method

#### Notes for Reviewers

@htdvisser please check if works as expected
@kschiffer you can extend the list of prefixes by extending the js config:
https://github.com/TheThingsNetwork/lorawan-stack/blob/master/cmd/internal/shared/joinserver/config.go#L24-L26
something like this:
```
JoinEUIPrefixes: []*types.EUI64Prefix{
  {EUI64: types.EUI64{0xff, 0x42, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff}, Length: 12},
  {EUI64: types.EUI64{0x10, 0x00, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff}, Length: 12},
  {EUI64: types.EUI64{0x42, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x00}, Length: 56},
},
```

### DIFF
--- a/pkg/webui/console/components/device-data-form/index.js
+++ b/pkg/webui/console/components/device-data-form/index.js
@@ -27,6 +27,7 @@ import SubmitBar from '../../../components/submit-bar'
 import ModalButton from '../../../components/button/modal-button'
 import FrequencyPlansSelect from '../../containers/freq-plans-select'
 import DevAddrInput from '../../containers/dev-addr-input'
+import JoinEUIPrefixesInput from '../../containers/join-eui-prefixes-input'
 
 import sharedMessages from '../../../lib/shared-messages'
 import errorMessages from '../../../lib/errors/error-messages'
@@ -184,16 +185,13 @@ class DeviceDataForm extends Component {
           name="join_server_address"
           component={Input}
         />
-        <Form.Field
+        <JoinEUIPrefixesInput
           title={sharedMessages.joinEUI}
           name="ids.join_eui"
-          type="byte"
-          min={8}
-          max={8}
           description={m.joinEUIDescription}
           required
           disabled={update}
-          component={Input}
+          showPrefixes={!update}
         />
         <Form.Field
           title={sharedMessages.devEUI}
--- a/pkg/webui/console/containers/join-eui-prefixes-input/compute-prefix.js
+++ b/pkg/webui/console/containers/join-eui-prefixes-input/compute-prefix.js
@@ -1,0 +1,29 @@
+// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const CHAR_BYTES = 4
+
+/**
+ * Computes the join EUI prefix given `joinEUI` and its `length`.
+ * @param {string} joinEUI - The join EUI.
+ * @param {number} length - The length of the prefix.
+ * @returns {string} - The join EUI prefix.
+ */
+function computePrefix(joinEUI, length = 0) {
+  // TODO: Support generating all prefixes that dont fulfil `length % CHAR_BYTES === 0`
+  // (https://github.com/TheThingsNetwork/lorawan-stack/issues/1263)
+  return joinEUI.slice(0, Math.ceil(length / CHAR_BYTES))
+}
+
+export default computePrefix
--- a/pkg/webui/console/containers/join-eui-prefixes-input/compute-prefix_test.js
+++ b/pkg/webui/console/containers/join-eui-prefixes-input/compute-prefix_test.js
@@ -1,0 +1,60 @@
+// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import computePrefix from './compute-prefix'
+
+describe('computePrefix', function() {
+  it('should compute empty prefix', function() {
+    const joinEUI = '0'.repeat(16)
+    const length = 0
+
+    expect(computePrefix(joinEUI, length)).toBe('')
+  })
+
+  it('should compute prefix of 1 byte', function() {
+    const joinEUI = '1'.repeat(16)
+    const length = 8
+
+    expect(computePrefix(joinEUI, length)).toBe('11')
+  })
+
+  it('should compute prefix of 1.5 bytes', function() {
+    const joinEUI = '1'.repeat(16)
+    const length = 12
+
+    expect(computePrefix(joinEUI, length)).toBe('111')
+  })
+
+  it('should compute prefix of 8 bytes', function() {
+    const joinEUI = '1'.repeat(16)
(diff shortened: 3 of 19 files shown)


pkg/webui/console/containers/join-eui-prefixes-input/join-eui-prefixes-input.js
```diff
@@ -0,0 +1,161 @@
...
+
+const emptyOption = { label: m.empty, value: '' }
+
+@bind
+class JoinEUIPrefixesInput extends React.PureComponent {
+
+  constructor (props) {
+    super(props)
+
+    this.inputRef = React.createRef()
```
kschiffer: Is the constructor necessary here?
```js
  inputRef = React.createRef()
  state = {
    …
  }
```
Saying this because you're doing it like this in `pkg/webui/components/input/byte.js` as well.

pkg/webui/console/containers/join-eui-prefixes-input/join-eui-prefixes-input.js
```diff
@@ -0,0 +1,161 @@
...
+    await this.setState({ prefix })
+    onChange(prefix)
+    if (this.inputRef.current) {
+      const instance = this.inputRef.current.getWrappedInstance()
+
+      instance.focus()
+    }
+  }
+
+  handleBlur (event) {
```
kschiffer: I don't see it used anywhere.

pkg/webui/console/store/reducers/join-server.js
```diff
@@ -0,0 +1,28 @@
...
+
+import { handleActions } from 'redux-actions'
+
+import { GET_JOIN_EUI_PREFIXES_SUCCESS } from '../actions/join-server'
+
+const defaultState = {
+  prefixes: [],
+}
+
+export default handleActions({
```
kschiffer: For the sake of consistency, I'd suggest not using another way of defining reducers via `handleActions()` here. I think doing this in a separate refactor would be better.

pkg/webui/console/components/device-data-form/index.js
```diff
@@ -187,16 +188,13 @@ class DeviceDataForm extends Component {
...
           title={sharedMessages.joinEUI}
           name="ids.join_eui"
-          type="byte"
-          min={8}
-          max={8}
           description={m.joinEUIDescription}
           required
           disabled={update}
-          component={Input}
+          select={!update}
```
kschiffer: Nitpick: just `select` is quite ambiguous. I'd prefer something like `showPrefixes` as prop name.

kschiffer:
@htdvisser Would it make more sense to select the first prefix by default (instead of no prefix)?

htdvisser:
@kschiffer the first prefix in the list isn't necessarily the preferred one and for LoRaWAN 1.0 you likely don't want to use any of the prefixes, so I think it's fine to not have an initial prefix in the field.

pkg/webui/console/containers/join-eui-prefixes-input/compute-prefix.js
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Computes the join EUI prefix given `joinEUI` and its `length`.
+ * @param {string} joinEUI - The join EUI.
+ * @param {number} length - The length of the prefix.
+ * @returns {string} - The join EUI prefix.
+ */
+function computePrefix (joinEUI, length = 0) {
+  if (length % CHAR_BYTES !== 0) {
+    const binaryPrefix = convertJoinEUI(joinEUI, HEX_BASE, BINARY_BASE).slice(0, length)
```
htdvisser: Will this work for large EUIs? I thought that converting a 64 bit uint (hex) to a JavaScript Number could be problematic?

pkg/webui/console/store/reducers/join-server.js
bafonins: why is this a problem?

pkg/webui/console/containers/join-eui-prefixes-input/compute-prefix_test.js
```diff
@@ -0,0 +1,52 @@
...
+    const length = 64
+
+    expect(computePrefix(joinEUI, length)).toBe(joinEUI)
+  })
+
+  it('should compute prefix of length 15', function () {
+    const joinEUI = '1'.repeat(16)
+    const length = 15
+
+    expect(computePrefix(joinEUI, length)).toBe('4444')
```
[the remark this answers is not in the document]
bafonins: right. for now it will return `111`

pkg/webui/console/containers/join-eui-prefixes-input/compute-prefix.js
bafonins: thats true, see my changes

pkg/webui/console/containers/join-eui-prefixes-input/compute-prefix.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Computes the join EUI prefix given `joinEUI` and its `length`.
+ * @param {string} joinEUI - The join EUI.
+ * @param {number} length - The length of the prefix.
+ * @returns {string} - The join EUI prefix.
+ */
+function computePrefix (joinEUI, length = 0) {
+  const diff = length % CHAR_BYTES
+
+  // TODO: Support generating all prefixes that dont fulfil `length % CHAR_BYTES === 0`
```
htdvisser: Refer to issue

pkg/webui/console/store/actions/join-server.js
```diff
@@ -0,0 +1,26 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { createRequestActions } from './lib'
+
+export const GET_JOIN_EUI_PREFIXES_BASE = 'GET_JOIN_EUI_PREFIXES'
+export const [{
+  request: GET_JOIN_EUI_PREFIXES,
+  success: GET_JOIN_EUI_PREFIXES_SUCCESS,
+  failure: GET_JOIN_EUI_PREFIXES_FAIURE,
```
kschiffer:
```suggestion
  failure: GET_JOIN_EUI_PREFIXES_FAILURE,
```

pkg/webui/console/containers/join-eui-prefixes-input/compute-prefix.js
```diff
@@ -0,0 +1,31 @@
...
+ * @param {string} joinEUI - The join EUI.
+ * @param {number} length - The length of the prefix.
+ * @returns {string} - The join EUI prefix.
+ */
+function computePrefix(joinEUI, length = 0) {
+  const diff = length % CHAR_BYTES
+
+  // TODO: Support generating all prefixes that dont fulfil `length % CHAR_BYTES === 0`
+  // (https://github.com/TheThingsNetwork/lorawan-stack/issues/1263)
+  return joinEUI.slice(0, (length - diff) / CHAR_BYTES)
```
htdvisser: We actually need to round this up instead of down. With this code, a prefix of `1111111111111111/15` gives `111`, allowing the user to choose something starting with `1114`, while only `1110` and `1111` are within `1111111111111111/15`.

htdvisser:
One small thing:

joineui mov

The placeholder dots turn "real" on field blur, so if I re-focus on the field I have to remove them.

bafonins:
@htdvisser Seems that all byte inputs have this problem. Ill create an issue.

PR #1145, opened 2019-08-12
Fix parsing of stream chunks in JS SDK

### LINKED ISSUE #1125: "Cannot show events" in the Console for user admin
#### Summary
When working with the Console to take a look into an Applications Event Stream (as admin) it first shows something for a few seconds and then enters "Cannot show events" error message, which then keeps on displaying even after browser refresh. The event stream does work in the CLI though.

#### Steps to Reproduce

1. Setup ttn-stack as of cli guide (with working device)
2. Setup the console client
3. Enter the console as admin and go to any application with a working device (check with CLI)
4. Open the "Data" view

[ttn-stack-config.txt](https://github.com/TheThingsNetwork/lorawan-stack/files/3477276/ttn-stack-config.txt)

#### What do you see now?
I can subscribe to the events streams of applications, gateways and devices assigned to the user-id admin when logged in as admin over the CLI.

I can see the event stream of gateways in the Console.

I cannot see the event streams of applications and devices in the Console, although being admin and collaborator with `ALL` access rights.

Seeing the event streams of applications and devices for the owner admin in the console, when logged in as admin and having `ALL` access rights.

#### Environment

docker-compose. stack container 3.1.0. Public IP and DNS, behind ACME TLS. Only TLS ports exposed (8xxx and 1700, plus 1885->80 and 8885->443).

```
ttn-lw-stack version
The Things Network Stack for LoRaWAN: ttn-lw-stack
Version:             3.1.0
Go version:          go1.12.7
OS/Arch:             linux/amd64
```

I have now idea right now, where the problem lies.

I don't think I can do this right now.

kschiffer:
#### Summary
Closes #1125

#### Changes
- Account for multi line stream chunks, with multiple separate json strings

#### Notes for Reviewers
Very likely the cause of #1125. I was able to reproduce and fix it with this diff.

#### Release Notes
- Fixed a bug that resulted in events streams crashing in the console.

### DIFF
--- a/sdk/js/src/api/stream/stream-node.js
+++ b/sdk/js/src/api/stream/stream-node.js
@@ -73,8 +73,11 @@ export default async function (payload, url) {
 
       stream.on('data', function (data) {
         const parsed = data.toString('utf8')
-        const result = JSON.parse(parsed).result
-        notify(listeners[EVENTS.EVENT], result)
+
+        for (const line of parsed.trim().split('\n')) {
+          const result = JSON.parse(line).result
+          notify(listeners[EVENTS.EVENT], result)
+        }
       })
       stream.on('end', function () {
         notify(listeners[EVENTS.CLOSE])
--- a/sdk/js/src/api/stream/stream.js
+++ b/sdk/js/src/api/stream/stream.js
@@ -87,8 +87,11 @@ export default async function (payload, url) {
       }
 
       const parsed = ArrayBufferToString(value)
-      const result = JSON.parse(parsed).result
-      notify(listeners[EVENTS.EVENT], result)
+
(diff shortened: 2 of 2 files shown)


bafonins:
event payload examples? steps to reproduce the error and verify that this indeed the solution?

kschiffer:
It is hard to reproduce this locally, without real data. I was able to track down this issue by registering my local built console to our demo environment. 
Then, on any connected gateway you have access to, see the issue occurring. You can modify the `stream.js` to output the raw parsed chunks before they are parsed to json. You'll see that occasionally there are multiple JSON object arriving, which causes trouble in the JSON parser.
After applying the fix, the events are parsed correctly and there are no more errors occurring.

PR #1151, opened 2019-08-13
Forward warning response header

htdvisser:
#### Summary

This is a quickfix that makes gRPC-Gateway forward the "warning" header.

It additionally fixes the CORS headers being set in the wrong place.

#### Changes

- Replaced unused "link" header with "warning"

#### Notes for Reviewers

These warning still need to be handled in the UI (cc: @kschiffer). See the `X-Warning` header in the response below:

```
% curl -v -H "Authorization: Bearer $access_token" http://localhost:1885/api/v3/auth_info
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 1885 (#0)
> GET /api/v3/auth_info HTTP/1.1
> Host: localhost:1885
> User-Agent: curl/7.54.0
> Accept: */*
> Authorization: Bearer MFRWG.DMQ64K36A4OAT2OTK2QBAXVP3JAK3OD4TGJK62Q.UW2PLPVNK2X6JQNE6OWVF5BLGIURP3WDOKT3DXNSYU2DBJE2OEPQ
> 
< HTTP/1.1 200 OK
< Access-Control-Allow-Origin: *
< Access-Control-Expose-Headers: Link, Date, Content-Length, X-Total-Count
< Content-Type: application/json
< Vary: Accept-Encoding
< Vary: Origin
< X-Content-Type-Options: nosniff
< X-Frame-Options: SAMEORIGIN
< X-Request-Id: 01DJ5N03KGBCTNRQEEX031Q3JC
< X-Warning: Restricted rights while account pending
< X-Xss-Protection: 1; mode=block
< Date: Tue, 13 Aug 2019 14:09:35 GMT
< Content-Length: 294
< 
* Connection #0 to host localhost left intact
{"oauth_access_token":{"user_ids":{"user_id":"newuser"},"client_ids":{"client_id":"cli"},"id":"DMQ64K36A4OAT2OTK2QBAXVP3JAK3OD4TGJK62Q","rights":["RIGHT_USER_DELETE","RIGHT_USER_SETTINGS_BASIC","RIGHT_USER_INFO"],"created_at":"2019-08-13T13:58:07.866Z","expires_at":"2019-08-13T14:58:07.866Z"}}% 
```

#### Release Notes

- Fixed propagation of warning headers in API responses

### DIFF
--- a/pkg/component/grpc.go
+++ b/pkg/component/grpc.go
@@ -55,7 +55,16 @@ func (c *Component) setupGRPC() (err error) {
 	for _, sub := range c.grpcSubsystems {
 		sub.RegisterHandlers(c.grpc.ServeMux, c.loopback)
 	}
-	c.web.RootGroup(ttnpb.HTTPAPIPrefix).Any("/*", echo.WrapHandler(http.StripPrefix(ttnpb.HTTPAPIPrefix, c.grpc)), middleware.CORS())
+	c.web.RootGroup(ttnpb.HTTPAPIPrefix).Any(
+		"/*",
+		echo.WrapHandler(http.StripPrefix(ttnpb.HTTPAPIPrefix, c.grpc)),
+		middleware.CORSWithConfig(middleware.CORSConfig{
+			Skipper:      middleware.DefaultSkipper,
+			AllowOrigins: []string{"*"},
+			AllowMethods: []string{http.MethodGet, http.MethodHead, http.MethodPut, http.MethodPatch, http.MethodPost, http.MethodDelete},
+			AllowHeaders: []string{"Date", "Content-Length", "X-Total-Count", "X-Warning"},
+		}),
+	)
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


kschiffer:
> X-Warning: Restricted rights while account pending

Any suggestion on how to handle these in the UI? We cannot use any i18n for these one-off messages. Can we use error identifiers or codes for these?

htdvisser:
It's just like an error, so instead of showing some red notification with the error in the top of the component that triggered it, it could be an orange one.

We could look into translating them, but not in this PR.

bafonins:
> It's just like an error, so instead of showing some red notification with the error in the top of the component that triggered it, it could be an orange one.

Thats something we discussed with @rish-c a while ago

> We could look into translating them, but not in this PR.

Could file an issue?

rish-c:
Not sure how we are planning account validation in V3 - it would be good to have a warning sign. Here are some examples on how it can be done - 

warning - 1

warning - 2

warning - 3

For now user validation is the first thing that comes to mind but I can imagine this can also be used to give other warnings (for instance in case someone hits fair usage on PCN, private network console - adding collaborator from a different domain etc.)

PR #1168, opened 2019-08-14
Identity Server email template fetching support

### LINKED ISSUE #148: Get the email template fetcher from the config
**Summary:** 

The identity server should use an email template fetcher in order to avoid recompiling the templates on each use.

**Why do we need this?**

In this way, the email templates are no longer recompiled on each used, and instead they can be reused by the template registry.

No fetcher is used right now, all of the templates are recompiled on each use.

The configuration required for the fetcher and the fetcher itself.

Create the required configuration and then add the fetcher, then change `getEmailTemplates` function to use it.

I can create this issue.

adriansmares:
#### Summary

Closes #148 
References https://github.com/TheThingsIndustries/lorawan-stack/issues/1648

#### Changes

- Add configuration support for email template fetchers
- Add email template registry caching inside the configuration

#### Notes for Reviewers

This implementation is compatible with https://github.com/TheThingsIndustries/lorawan-stack/issues/1648 . If holding the registry itself and the mutex inside the config is not fashionable, the cache can be moved to the IS itself, with the source as key, but won't support caching the in-memory templates. Suggestions are welcomed.

#### Release Notes

- Added support for external email templates in the Identity Server.

### DIFF
--- a/doc/content/concepts/email-templates.md
+++ b/doc/content/concepts/email-templates.md
@@ -1,0 +1,22 @@
+---
+title: "Email Templates"
+description: ""
+weight: 15
+---
+
+## What is it?
+
+Email templates define the contents of the emails that The Things Stack sends to its users. They allow network operators to override the default contents of the emails with their own custom contents.
+
+## Who is it for?
+
+Email templates are targeted at network operators that would like to customize the emails that The Things Stack sends to its users.
+
+### Typical use cases
+
+1. Adding additional styling to the default text-only emails that The Things Stack sends.
+2. Translating the emails The Things Stack sends to its users (applies deployment-wide).
+ 
+## How does it work?
+
+Email templates override the default emails that The Things Stack sends by providing custom template files written in Go's [html/template](https://golang.org/pkg/html/template/) format. The templates are retrieved when The Things Stack sends the first email of a certain type, and then are cached until it is restarted. See [Available Templates]({{< ref "/reference/email-templates/available.md" >}}) and [Overriding Default Templates]({{< ref "/reference/email-templates/overriding.md" >}}).
--- a/doc/content/reference/email-templates/available.md
+++ b/doc/content/reference/email-templates/available.md
@@ -1,0 +1,30 @@
+---
+title: "Available Templates"
+description: ""
+weight: 3
+---
+
+The following email templates are currently used by The Things Stack:
+
+Full Name | Identifier | Scope | Additional fields
+----------|------------|-------|------------------
+API Key changed | `api_key_changed` | Sent when the rights of an API Key have been changed. | `Identifiers` and `Rights`
+API Key created | `api_key_created` | Send when an API Key has been created. | `Identifiers` and `Rights`
+Collaborator changed | `collaborator_changed` | Sent when the rights of a collaborator have been changed. | `Collaborator`
+Password changed | `password_changed` | Sent when the the password of a user has been changed.
+Temporary password | `temporary_password` | Sent when a temporary password has been requested for an user. | `TemporaryPassword`
+Email validation | `validate` | Sent when a user is added as a collaborator of an entity, in order to validate their email. | `ID` and `Token`
+
+The following fields can be used inside all of the email templates:
+
+Field | Description | Notes
+------|-------------|------
+`User.ID` | The identifier of the user. | If the recipient is the contact of the target entity, the value will be empty.
+`User.Name` | The full name of the user. | If the recipient is the contact of the target entity, the value will be `user`.
+`User.Email` | The email of the user. | None.
+`Network.Name` | The full name of the network running on the stack. | None.
+`Network.IdentityServerURL` | The base URL of the Identity Server. | None.
+`Network.ConsoleURL` | The base URL of the Console. | None.
+`Entity.Type` | The type of the target entity, such as `application`, `client`, `gateway`, `organization` or `user`. | None.
+`Enitity.ID` | The identifier of the target entity. | None.
+`Contact.Type` | The contact type of the recipient, such as `technical`, `billing` or `abuse`. | Only used if the recipient is the contact of an entity.
--- a/doc/content/reference/email-templates/overriding.md
+++ b/doc/content/reference/email-templates/overriding.md
@@ -1,0 +1,67 @@
+---
+title: "Overriding Templates"
+description: ""
+weight: 6
+---
+
+> Note: We recommend getting familliar with the [html/template](https://golang.org/pkg/html/template/) template format first.
+
+## Template components
+
+An email template override has three components:
+
+1. The subject file, which contains the subject of the email and is named `<identifier>.subject.txt`. 
+2. The text contents file, which contains the contents of the email in text format and is named `<identifier>.txt`.
+3. The HTML contents file, which contains the contents of the email in HTML format and is named `<identifier>.html`.
+
+
+## Creating the overrides
+
+In order to override a template, one must provide all three files as part of the email templates repository and then provide them as part of the configuration.
+
+Let's consider that we want to override the email that a user receives once they register, the email validation email, which has the identifier `validate`. We need to create the following files:
+
+
+- `validate.subject.txt`
+```text
+Please confirm your email address for {{.Network.Name}}
+```
+- `validate.txt`
+```text
+Please confirm your email address for {{.Network.Name}}.
+Your email address will be used as contact for {{.Entity.Type}} "{{.Entity.ID}}". 
+
+Reference: {{.ID}}
+Confirmation Token: {{.Token}}
+```
+- `validate.html`
+```html
+Please confirm your email address for {{.Network.Name}}. <br> 
+Your email address will be used as contact for {{.Entity.Type}} "{{.Entity.ID}}". <br> <br> 
+
+Reference: {{.ID}} <br> 
+Confirmation Token: {{.Token}}
+```
+
+## Providing the overrides to the stack
+
+Once you have written your overrides, you can provide them to the stack either through an remote URL, or through the local file system.
+
+### Fetching from a remote URL
+
+In order to allow The Things Stack to access remote files, you must only provide the URL to the root folder that contains the files. Consider that you need to provide the The Things Stack access to a file called `validate.txt`, which you have uploaded on your host at `http://www.example.org/emails/validate.txt`. Then the URL that you provide to the email templates configuration is `http://www.example.org/emails/`.
+
+### Fetching from a local directory
+
+In order to allow the The Things Stack to access files which are hosted in your own file system, you must provide the path to the root folder that contains the files. Consider that you need to provide the The Things Stack access to a file called `validate.txt`, which is available on your file system in the `/srv/emails/validate.txt`. Then the path that you need to provide to the email templates configuration is `/srv/emails/`.
+
+### Example YAML configuration 
+
+```yaml
+is:
+  email:
+    templates:
+      # Specify only one source for the files.
+      directory: "/path/to/the/template/files"
+      url: "http://www.example.com/emails"
+```
--- a/doc/content/reference/email-templates/_index.md
+++ b/doc/content/reference/email-templates/_index.md
@@ -1,0 +1,5 @@
+---
+title: "Email Templates"
+description: ""
+weight: 2
+---
--- a/pkg/identityserver/email.go
+++ b/pkg/identityserver/email.go
@@ -22,18 +22,27 @@ import (
 	"go.thethings.network/lorawan-stack/pkg/email"
 	"go.thethings.network/lorawan-stack/pkg/email/sendgrid"
 	"go.thethings.network/lorawan-stack/pkg/email/smtp"
+	"go.thethings.network/lorawan-stack/pkg/fetch"
 	"go.thethings.network/lorawan-stack/pkg/identityserver/emails"
 	"go.thethings.network/lorawan-stack/pkg/identityserver/store"
 	"go.thethings.network/lorawan-stack/pkg/log"
 	"go.thethings.network/lorawan-stack/pkg/ttnpb"
 )
 
-var globalEmailTemplateRegistry = email.NewTemplateRegistry(nil)
-
-func (is *IdentityServer) getEmailTemplates(_ context.Context) *email.TemplateRegistry {
-	// TODO: Get email template fetcher from config (https://github.com/TheThingsNetwork/lorawan-stack/issues/148).
-	// Re-use existing registry for same fetcher (to avoid re-compiling templates).
-	return globalEmailTemplateRegistry
+func (is *IdentityServer) getEmailTemplates(ctx context.Context) *email.TemplateRegistry {
+	c := is.configFromContext(ctx).Email.Templates
+	var fetcher fetch.Interface
+	switch {
+	case c.Static != nil:
+		fetcher = fetch.NewMemFetcher(c.Static)
+	case c.Directory != "":
+		fetcher = fetch.FromFilesystem(c.Directory)
+	case c.URL != "":
+		fetcher = fetch.FromHTTP(c.URL, true)
+	default:
+		fetcher = nil
+	}
+	return email.NewTemplateRegistry(fetcher, c.Includes...)
 }
 
 // SendEmail sends an email.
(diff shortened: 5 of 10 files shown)


doc/content/concepts/email-templates.md
```diff
@@ -0,0 +1,22 @@
...
+weight: 15
+---
+
+## What is it?
+
+Email templates define the contents of the emails that The Things Stack sends to its users. They allow users to override the default contents of the emails with their own custom contents.
+
+## Who is it for?
+
+Email templates are targeted at network administrators that would like to customize the emails that the stack sends to its users.
```
htdvisser: Always write The Things Stack

htdvisser: Also below

doc/content/concepts/email-templates.md
```diff
@@ -0,0 +1,22 @@
...
+
+### Typical use cases
+
+1. Adding additional styling to the default text-only emails that the stack sends.
+2. Translating of the emails the stack sends to its users.
```
htdvisser: We don't have support for "user language", so that would be deployment-wide.

doc/content/concepts/email-templates.md
```diff
@@ -0,0 +1,22 @@
...
+ 
+## How does it work?
+
+Email templates override the default emails that the stack sends by providing custom template files written in Go's [html/template](https://golang.org/pkg/html/template/) format using [fetchers]({{< ref "/concepts/fetchers.md" >}}). The templates are retrieved by the stack when the stack sends the first email of a certain type, and then are cached until the stack is restarted. See [Available Templates]({{< ref "/reference/email-templates/available.md" >}}) and [Overriding Default Templates]({{< ref "/reference/email-templates/overriding.md" >}}).
```
htdvisser: "using fetchers" sounds a bit strange

doc/content/concepts/fetchers.md
```diff
@@ -0,0 +1,31 @@
...
+weight: 20
+---
+
+## What is it?
+
+Fetchers allow the stack to gain access to external files, such as [Email Templates]({{< ref "/concepts/email-templates.md" >}}), Frequency Plans, Device Data for the Device Repository, Webhook Templates and others.
+
+## Who is it for?
+
+The fetchers are used by users that want to customize certain parts of the stack. They are exposed indirectly through the stack configuration.
```
htdvisser: The term "users" is a bit confusing here.

doc/content/concepts/fetchers.md
```diff
@@ -0,0 +1,31 @@
+---
+title: "Fetchers"
+description: ""
+weight: 20
+---
+
+## What is it?
+
+Fetchers allow the stack to gain access to external files, such as [Email Templates]({{< ref "/concepts/email-templates.md" >}}), Frequency Plans, Device Data for the Device Repository, Webhook Templates and others.
```
htdvisser: The Things Stack

htdvisser: also below

doc/content/concepts/fetchers.md
```diff
@@ -0,0 +1,31 @@
...
+The fetchers are used by users that want to customize certain parts of the stack. They are exposed indirectly through the stack configuration.
+
+### Typical use cases
+
+1. Providing custom resources to the stack, such as custom [Email Templates]({{< ref "/concepts/email-templates.md" >}}) in order to change the behaviour of the stack.
+2. Allowing the stack to run in air-gapped environments, where internet connections are not available constantly and resources must be served locally.
+
+## How does it work?
+
+The fetchers provide the stack with a way to access files which can be hosted remotely (such as on a remote website), locally (by providing a local path to the files) or in memory (by providing the contents of the files directly as part of the configuration).
```
htdvisser: I'm not sure we fully support that "as part of the configuration"

doc/content/reference/email-templates/available.md
```diff
@@ -0,0 +1,30 @@
+---
+title: "Available Templates"
+description: ""
+weight: 3
+---
+
+The following email templates are currently used by The Things Stack:
```
htdvisser: Maybe add notes to `pkg/identityserver/emails` to update this doc when adding templates and when adding fields to the structs?

doc/content/reference/email-templates/available.md
```diff
@@ -0,0 +1,30 @@
...
+
+Full Name | Identifier | Scope | Additional fields
+----------|------------|-------|------------------
+API Key changed | api_key_changed | Sent when the rights of an API Key have been changed. | `Identifiers` and `Rights`
+API Key created | api_key_created | Send when an API Key has been created. | `Identifiers` and `Rights`
+Collaborator changed | collaborator_changed | Sent when the rights of a collaborator have been changed. | `Collaborator`
+Password changed | password_changed | Sent when the the password of an user has been changed.
```
htdvisser: a user

doc/content/reference/email-templates/available.md
```diff
@@ -0,0 +1,30 @@
...
+Temporary password | temporary_password | Sent when an temporary password has been requested for an user. | `TemporaryPassword`
```
htdvisser: a temporary password

doc/content/reference/email-templates/available.md
```diff
@@ -0,0 +1,30 @@
...
+Account validation | validate | Sent when a user is created in order to validate their email address. | `ID` and `Token`
```
htdvisser: This is not just for accounts, also for other (app,cli,gtw,org) contact info of EMAIL type

doc/content/reference/email-templates/available.md
```diff
@@ -0,0 +1,30 @@
...
+
+The following fields can be used inside all of the email templates:
```
htdvisser: They're not always all available, i.e. `User` is only set if we're sending an email to a User.

doc/content/reference/email-templates/available.md
```diff
@@ -0,0 +1,30 @@
...
+
+Field | Description
+------|------------
+`User.ID` | The identifier of the user.
+`User.Name` | The full name of the user.
+`User.Email` | The email of the user.
+`Network.Name` | The full name of the network running on the stack.
+`Network.IdentityServerURL` | The base URL of the Identity Server.
+`Network.ConsoleURL` | The base URL of the Console.
+`Entity.Type` | The type of the entity, such as `user`, `collaborator` etc.
```
htdvisser: Not `collaborator`. As far as I remember this can only be entity types like `application`, `client`, `gateway`, `organization`, `user`

doc/content/reference/email-templates/available.md
```diff
@@ -0,0 +1,30 @@
...
+`Enitity.ID` | The identifier of the entity.
+`Contact.Type` | The contact type of the user, such as `technical`, `billing` or `abuse`.
```
htdvisser: the recipient, not the user

doc/content/reference/email-templates/overriding.md
```diff
@@ -0,0 +1,41 @@
...
+----------|---------
+validate.subject.txt | Please confirm your email address for {{.Network.Name}}
+validate.txt | Please confirm your email address for {{.Network.Name}}. <br> Your email address will be used as contact for {{.Entity.Type}} "{{.Entity.ID}}". <br><br> Reference: {{.ID}} <br> Confirmation Token: {{.Token}}
+validate.html | Please confirm your email address for {{.Network.Name}}. &lt;br&gt;&lt;br&gt; <br> Your email address will be used as contact for {{.Entity.Type}} "{{.Entity.ID}}". &lt;br&gt;&lt;br&gt; <br> Reference: {{.ID}} &lt;br&gt; <br> Confirmation Token: {{.Token}}
+
+## Providing the overrides to the stack
+
+Once you have written your overrides, you can provide them to the stack either through an remote URL, or through the local file system.
+
+```yaml
```
htdvisser: I don't think we should do this here, otherwise we also need to document the corresponding flags/env.

doc/content/reference/email-templates/overriding.md
```diff
@@ -0,0 +1,41 @@
...
+3. The HTML contents file, which contains the contents of the email in HTML format and is named `<identifier>.html`.
+
+
+## Creating the overrides
+
+In order to override a template, one must provide all 3 files as part of the email templates repository and then provide them to a [fetcher]({{< ref "/concepts/fetchers.md" >}}).
+
+Let's consider that we want to override the email that a user receives once he registers, the validation email (`validate`). We need to create the following files:
+
+File name | Contents
```
htdvisser: I think code blocks would be better than a table here

doc/content/concepts/email-templates.md
johanstokking: I think the internal workings belong in Reference; Concepts are short and concise, focusing on how it works for the end user

doc/content/concepts/fetchers.md
johanstokking: We don't support memory through config; that's only for testing

doc/content/concepts/fetchers.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Fetching from a remote URL
+
+In order to allow the stack to access remote files, you must only provide the URL to the root folder that contains the files. Consider that you need to provide the stack access to a file called `temporary_password.txt`, which you have uploaded on your host at `http://www.example.org/emails/temporary_password.txt`. Then the URL that you provide to the email templates configuration is `http://www.example.org/emails/`.
```
johanstokking: This would belong in a Guide imo; it's quite specific to what's being fetched, i.e. in this case "the email templates configuration" which is not related to the concept Fetchers, explained here

doc/content/concepts/fetchers.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Fetching from a local directory
+
+In order to allow the stack to access files which are hosted in your own file system, you must provide the path to the root folder that contains the files. Consider that you need to provide the stack access to a file called `temporary_password.txt`, which is available on your file system in the `/srv/emails/temporary_password.txt`. Then the path that you need to provide to the email templates configuration is `/srv/emails/`.
```
johanstokking: Idem

doc/content/reference/email-templates/overriding.md
```diff
@@ -0,0 +1,41 @@
...
+1. The subject file, which contains the subject of the email and is named `<identifier>.subject.txt`. 
+2. The text contents file, which contains the contents of the email in text format and is named `<identifier>.txt`.
+3. The HTML contents file, which contains the contents of the email in HTML format and is named `<identifier>.html`.
+
+
+## Creating the overrides
+
+In order to override a template, one must provide all 3 files as part of the email templates repository and then provide them to a [fetcher]({{< ref "/concepts/fetchers.md" >}}).
+
+Let's consider that we want to override the email that a user receives once he registers, the validation email (`validate`). We need to create the following files:
```
johanstokking: No gender in users, check all occurrences.

Always they/their/them

johanstokking: Also what is `validate` here? Is there a list of identifiers? How does that work?

doc/content/reference/email-templates/available.md
adriansmares: Aren't `User.Name` and `User.Email` always available since they represent the recipient of the email ?

doc/content/reference/email-templates/overriding.md
adriansmares: Any other thoughts about how this should look ? I mentioned the `yaml` configuration since the getting started guide already mentions it for the CLI.

doc/content/reference/email-templates/overriding.md
adriansmares: Reiterated the explanation for `validate`.

doc/content/reference/email-templates/overriding.md
johanstokking: I do think we need some reference to configuration. When reading the docs I was curious to _how_ to configure this really.

doc/content/reference/email-templates/available.md
```diff
@@ -0,0 +1,30 @@
...
+title: "Available Templates"
+description: ""
+weight: 3
+---
+
+The following email templates are currently used by The Things Stack:
+
+Full Name | Identifier | Scope | Additional fields
+----------|------------|-------|------------------
+API Key changed | api_key_changed | Sent when the rights of an API Key have been changed. | `Identifiers` and `Rights`
```
johanstokking: Code tag for identifiers; `api_key_changed`

doc/content/reference/email-templates/overriding.md
```diff
@@ -0,0 +1,67 @@
...
+title: "Overriding Templates"
+description: ""
+weight: 6
+---
+
+> Note: We recommend getting familliar with the [html/template](https://golang.org/pkg/html/template/) template format first.
+
+## Template components
+
+An email template override has 3 components:
```
johanstokking: Nitpick; "three"

doc/content/reference/email-templates/overriding.md
```diff
@@ -0,0 +1,67 @@
...
+
+1. The subject file, which contains the subject of the email and is named `<identifier>.subject.txt`. 
+2. The text contents file, which contains the contents of the email in text format and is named `<identifier>.txt`.
+3. The HTML contents file, which contains the contents of the email in HTML format and is named `<identifier>.html`.
+
+
+## Creating the overrides
+
+In order to override a template, one must provide all 3 files as part of the email templates repository and then provide them as part of the configuration.
```
johanstokking: Idem

doc/content/reference/email-templates/overriding.md
```diff
@@ -0,0 +1,67 @@
...
+
+Let's consider that we want to override the email that a user receives once they register, the email validation email, which has the identifier `validate`. We need to create the following files:
+
+
+- validate.subject.txt
```
johanstokking: Backticks

doc/content/reference/email-templates/available.md
htdvisser: We also send emails to `ContactInfo` objects that don't have a `User`

doc/content/reference/email-templates/overriding.md
htdvisser: @johanstokking I agree, but doesn't the configuration reference belong in a **Configuration** section in the **Reference** section, instead of the **Email Templates**?

doc/content/concepts/email-templates.md
```diff
@@ -0,0 +1,22 @@
+---
+title: "Email Templates"
+description: ""
+weight: 15
+---
+
+## What is it?
+
+Email templates define the contents of the emails that The Things Stack sends to its users. They allow users to override the default contents of the emails with their own custom contents.
```
htdvisser: > They allow users to override

I think "users" should be "network operators" here.

@johanstokking is "network operators" the term we now use for the people that deploy the stack?

doc/content/reference/email-templates/overriding.md
johanstokking: What do we put there, information about configuration in general?

I think, at minimum, we should mention the concerning yaml and env keys in a reference. We should not redirect to generic configuration where people have to figure out the potentially relevant keys out of all the possible keys.

doc/content/concepts/email-templates.md
johanstokking: Yeah good catch. In this paragraph the mentioned "users" mean two different things. I think "network operators" is a better term here indeed.

doc/content/reference/email-templates/available.md
```diff
@@ -17,14 +17,14 @@ Email validation | `validate` | Sent when a user is added as a collaborator of a
...
-`Network.Name` | The full name of the network running on the stack.
-`Network.IdentityServerURL` | The base URL of the Identity Server.
-`Network.ConsoleURL` | The base URL of the Console.
-`Entity.Type` | The type of the entity, such as `application`, `client`, `gateway`, `organization` or `user`.
-`Enitity.ID` | The identifier of the entity.
-`Contact.Type` | The contact type of the recipient, such as `technical`, `billing` or `abuse`.
+Field | Description | Notes
+------|-------------|------
+`User.ID` | The identifier of the user. | If the recipient is the contact of the target entity, the ID will be empty.
+`User.Name` | The full name of the user. | If the recipient is the contact of the target entity, the Name will be `user`.
```
johanstokking: This is not very clear to me.

Should we have a matrix per field and template identifier which fields are available?

PR #1171, opened 2019-08-14
Always build the docs and the version template in CI

### LINKED ISSUE #1049: Check docs config diff and build pass
#### Summary

Check docs config diff and build pass

#### Why do we need this?

- To ensure in CI that a change in `config.toml.tmpl` is reflected in `config.toml`
- To ensure that there are no broken references or other mistakes

This is currently only being done on release while we want to catch mistakes as soon as possible.

- Check diff
- Check that doc builds successfully

#### Environment

Travis

- Suggesting to run `mage version:files` also in non-release builds to check diff
- Suggesting to run `mage docs:build` and check exit code

No

adriansmares:
#### Summary

Closes #1049

#### Changes

- Always run the versioning and docs building during the diff check

#### Notes for Reviewers

The `hugo` command errors automatically if the build fails with error code `255`, which will fail the CI build automatically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,7 @@ install:
 - |
   if [[ "$RUNTYPE" == "js" ]]; then
     $MAGE js:devDeps js:deps jsSDK:deps
+    $MAGE docs:deps
   fi
 - |
   if [[ "$RUNTYPE" == "release" ]]; then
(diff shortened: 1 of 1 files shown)


adriansmares:
Moved them both to the `js` build in order to avoid fetching the JS deps every time.

.travis.yml
```diff
@@ -72,6 +72,7 @@ install:
 - |
   if [[ "$RUNTYPE" == "js" ]]; then
     $MAGE js:devDeps js:deps jsSDK:deps
+    $MAGE docs:deps
```
htdvisser: We also need docs deps on `release`

.travis.yml
```diff
@@ -96,6 +97,8 @@ script:
 - if [[ "$RUNTYPE" == "js" ]]; then $MAGE js:translations; fi
 - if [[ "$RUNTYPE" == "js" ]]; then $MAGE js:test jsSDK:test; fi
 - if [[ "$RUNTYPE" == "js" ]]; then $MAGE js:lint; fi
+- if [[ "$RUNTYPE" == "js" ]]; then $MAGE version:files; fi
```
htdvisser: We also need version files on `release`

.travis.yml
adriansmares: We do - check line 84.

adriansmares:
I'm ok with delaying this until we can do it properly with the GitHub Actions. cc @johanstokking

johanstokking:
I think this fits with the model that we have today and it catches CI issues, so I'm fine merging it.

Cosmetically, however, it would be nice to keep the blocks together, removing the `||` but introducing explicit conditions.

Issue #1175, opened 2019-08-14
Remove @bind decorator from class components


bafonins:
#### Summary

We should remove `@bind` decorators from class components

#### Why do we need this?

https://github.com/andreypopp/autobind-decorator#discouraged-approaches

#### What is already there? What do you see now?

A lot of class components annotated with `@bind`

#### What is missing? What do you want to see?

Only certain methods that need access to `this` being decorated

#### Environment

browser (oauth, console)

#### How do you propose to implement this?

1. Only bind methods that need access to `this`
2. ~~Use arrow functions~~
```
class Component extends React.Component {
  ...
  handleClick = () => {}
  ...
```

The second approach might require some changes to `eslint` configuration

#### Can you do this yourself and submit a Pull Request?

Yes, but help is welcome.

kschiffer:
Mildly related:
https://github.com/TheThingsNetwork/lorawan-stack/blob/master/config/eslintrc.yaml#L24-L26

Decorating just before export statements is also discouraged since babel 7. We should take that into account as well if we do a refactor.

kschiffer:
> 1. Only bind methods that need access to `this`
> 2. Use arrow functions
```
class Component extends React.Component {
  ...
  handleClick = () => {}
  ...
```

Let's agree on using 1. only for consistency, since we already refactored some components to do this.

PR #1176, opened 2019-08-14
Add true hot module replacement for console/oauth apps

bafonins:
#### Summary

This PR adds HMR to the console and oauth apps

#### Changes

- Setup hmr for console and oauth
- Small fix to breadcrumbs context
- Change `lib/dev.js` to match the logic we use in the `webpack` configuration. Namely, any `NODE_ENV !== development` is regarded as the production env.

#### Notes for Reviewers

* Not every component is properly updated on changes. This is caused by `autobind-decorator` (see https://github.com/TheThingsNetwork/lorawan-stack/issues/1175). There are probably ways to make it work, but as far as I can tell it might require some changes to babel; this is something I dont really want to do in this PR.


bafonins:
> There is a lot of config file formatting in the diff; so can we merge #1181 before this?

sure

PR #1177, opened 2019-08-14
Replace remaining hardcoded `/api/v3` api prefixes

kschiffer:
#### Summary
This quickfix PR will remove a couple of hardcoded `/api/v3` prefixes for API URLs in the console and oauth app. These were not called through the JS SDK, which already handles API URLs correctly. 
Without this fix, the console and oauth apps would not work in environments where other api url prefixes are configured, or when the IS is served from another host.

#### Changes
- Add configurable IS API base route value to the oauth app of the IS
- Replace hardcoded bits with config passed via page data (in console and oauth app)

#### Notes for Reviewers
@htdvisser / @johanstokking, (and also @bafonins I guess) please review the IS/oauth backend part
@bafonins please review the console/oauth WebUI part

The oauth app currently queries the IS to register new users, which is why the base URL of the IS needs to be passed there. In the future, the responsibilities of the oauth app might change (with a new `account` app e.g., yet to be figured out), so then we might need to (re)move this config again.

### DIFF
--- a/pkg/oauth/server.go
+++ b/pkg/oauth/server.go
@@ -67,7 +67,8 @@ type UIConfig struct {
 
 // FrontendConfig is the configuration for the OAuth frontend.
 type FrontendConfig struct {
-	Language string `json:"language" name:"-"`
+	Language string          `json:"language" name:"-"`
+	IS       webui.APIConfig `json:"is" name:"is"`
 }
 
 // Config is the configuration for the OAuth server.
--- a/pkg/webui/oauth/api/index.js
+++ b/pkg/webui/oauth/api/index.js
@@ -15,9 +15,11 @@
 import axios from 'axios'
 
 import getCookieValue from '../../lib/cookie'
-import { selectApplicationRootPath } from '../../lib/selectors/env'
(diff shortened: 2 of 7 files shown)


pkg/oauth/server.go
```diff
@@ -65,9 +65,16 @@ type UIConfig struct {
 	FrontendConfig     `name:",squash"`
 }
 
+// APIConfig for upstream APIs.
+type APIConfig struct {
```
htdvisser: We may want to move `APIConfig` to `pkg/webui` instead of copy-pasting it from `console` to `oauth`.

pkg/webui/oauth/api/index.js
```diff
@@ -27,7 +29,7 @@ const instance = axios.create({
 export default {
   users: {
     async register(userData) {
-      return axios.post(`/api/v3/users`, userData)
+      return axios.post(`${isBaseUrl}/users`, userData)
```
bafonins: what about `resetPassword` and `updatePassword` ?

kschiffer: Ha. Missed them after the rebase.

PR #1186, opened 2019-08-15
Move generic oauth client logic to web package

### LINKED ISSUE #1184: Outsource oauth client logic to web package
#### Summary
We should outsource common (web based) oauth client logic to the web utility to avoid code repetition. With https://github.com/TheThingsIndustries/lorawan-stack/issues/1679 and https://github.com/TheThingsIndustries/lorawan-stack/issues/1673, we will implement another oauth client. Without moving the common oauth logic to a dedicated utility package, we would need to write redundant code, since the base setup of oauth client functionality is quite generic.

#### Why do we need this?
DRY principle, separation of concerns

Generic oauth client logic inside the console component.

Utility package with oauth client logic that we can use to simplify setup of clients.

#### Environment
Irrelevant.

I would suggest making this a part of our `pkg/web` package. We can move all routing, handler and config logic into a generic `OAuthClient` component, which can then be instantiated and added to our client components.

Yes.

kschiffer:
#### Summary
Closes #1184 

#### Changes
- Add `oauthclient` package to `pkg/web`
- Refactor console backend code to use the new package
- Update messages

#### Notes for Reviewers
I'm depending on this for https://github.com/TheThingsIndustries/lorawan-stack/issues/1679


pkg/web/oauthclient/callback.go
```diff
@@ -0,0 +1,68 @@
...
+	"net/http"
+
+	echo "github.com/labstack/echo/v4"
+)
+
+// HandleCallback is a handler that takes the auth code and exchanges it for the
+// access token.
+func (oc *OAuthClient) HandleCallback(c echo.Context) error {
+	if e := c.QueryParam("error"); e != "" {
+		return echo.NewHTTPError(http.StatusUnauthorized, c.QueryParam("error_description"))
```
htdvisser: Perhaps we should make this a registered error instead of an echo error. Same for the ones below.

pkg/web/oauthclient/callback.go
```diff
@@ -0,0 +1,68 @@
...
+	stateCookie, err := oc.getStateCookie(c)
+	if err != nil {
+		return err
+	}
+
+	if stateCookie.Secret != state {
+		return echo.NewHTTPError(http.StatusBadRequest, "Invalid state")
+	}
+
+	//Exchange token.
```
htdvisser: Space after `//`

pkg/web/oauthclient/oauthclient.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Config is the configuration for the OAuth client.
+type Config struct {
+	AuthorizeURL string `name:"authorize-url" description:"The OAuth Authorize URL"`
+	TokenURL     string `name:"token-url" description:"The OAuth Token Exchange URL"`
+
+	ClientID     string `name:"client-id" description:"The OAuth client ID"`
+	ClientSecret string `name:"client-secret" description:"The OAuth client secret" json:"-"`
+
+	StateCookieName string
```
htdvisser: Add `name:"-"` to `StateCookieName` and `AuthCookieName`. Those don't need to be in flag config.

pkg/web/oauthclient/logout.go
```diff
@@ -24,21 +24,22 @@ import (
...
 	creds := grpc.PerRPCCredentials(rpcmetadata.MD{
 		AuthType:      "Bearer",
 		AuthValue:     token.AccessToken,
-		AllowInsecure: console.AllowInsecureForCredentials(),
+		AllowInsecure: oc.c.AllowInsecureForCredentials(),
 	})
 
 	ctx := c.Request().Context()
-	if peer := console.GetPeer(ctx, ttnpb.PeerInfo_ACCESS, nil); peer != nil {
+	if peer := oc.c.GetPeer(ctx, ttnpb.PeerInfo_ACCESS, nil); peer != nil {
```
bafonins:
```suggestion
	if peer := oc.component.GetPeer(ctx, ttnpb.PeerInfo_ACCESS, nil); peer != nil {
```

pkg/web/oauthclient/callback.go
```diff
@@ -0,0 +1,68 @@
...
+	token, err := oc.oauth(c).Exchange(c.Request().Context(), code)
+	if err != nil {
+		return err
+	}
+
+	// Remove state cookie.
+	oc.removeStateCookie(c)
```
bafonins: method name here is descriptive enough, no need for comment.

PR #1188, opened 2019-08-15
Fix / improve frontend mage targets

kschiffer:
#### Summary
This quickfix PR will fix a couple of issues with our frontend mage targets:
- `mage js:deps` sometimes not running due to faulty mod-time checks
- `mage jsSdk:deps` sometimes not running due to faulty mod-time checks
- dependency updates not applied when running the `js:serve` command
- DLL bundle creation unnecessarily running in production environments

#### Changes
- Replace `target.Dir` with `target.Path`, as that seems to resolve issues with faulty modification time checks
- Add `js.Deps` as dependency of `js:BuildDll`
- Add env check to `js:BuildDll` to avoid unnecessary dll builds in production
- Fix some target descriptions

#### Notes for reviewers
@bafonins please confirm this causes no issues for you

### DIFF
--- a/.mage/js.go
+++ b/.mage/js.go
@@ -187,8 +187,9 @@ func (js Js) BuildMain() error {
 
 // BuildDll runs the webpack to build the DLL bundle
 func (js Js) BuildDll() error {
+	mg.Deps(js.Deps)
 	changed, err := target.Path("./public/libs.bundle.js", "./yarn.lock")
-	if os.IsNotExist(err) || (err == nil && changed) {
+	if changed || os.IsNotExist(err) {
 		if mg.Verbose() {
 			fmt.Println("Running Webpack for DLL…")
 		}
--- a/.mage/js.go
+++ b/.mage/js.go
@@ -185,7 +185,7 @@ func (js Js) BuildMain() error {
 	return webpack("--config", "config/webpack.config.babel.js")
 }
 
-// BuildDll runs the webpack to build the DLL bundle
+// BuildDll runs the webpack command to build the DLL bundle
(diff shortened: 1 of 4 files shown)


.mage/js.go
```diff
@@ -185,18 +185,25 @@ func (js Js) BuildMain() error {
...
-		if mg.Verbose() {
-			fmt.Println("Running Webpack for DLL…")
-		}
-		webpack, err := js.webpack()
-		if err != nil {
-			return err
+	mg.Deps(js.Deps)
+	if nodeEnv := os.Getenv("NODE_ENV"); nodeEnv == "development" {
+		changed, err := target.Path("./public/libs.bundle.js", "./yarn.lock")
+		if os.IsNotExist(err) || (err == nil && changed) {
```
htdvisser: Fine to just do `if changed || os.IsNotExist(err)` right?

.mage/js.go
```diff
@@ -185,18 +185,25 @@ func (js Js) BuildMain() error {
...
+			if mg.Verbose() {
+				fmt.Println("Running Webpack for DLL…")
```
htdvisser: Use `...` (3 chars), not `…` (1 char)

.mage/styl.go
```diff
@@ -36,7 +36,7 @@ func (styl Styl) stylint() (func(args ...string) (string, error), error) {
 	}, nil
 }
 
-// Lint runs eslint over frontend js files.
+// Lint runs styling over frontend styl files.
```
htdvisser: I think you meant `stylint` here

Issue #1191, opened 2019-08-16
User RequirePasswordUpdate field not enforced


htdvisser:
#### Summary

The `RequirePasswordUpdate` field of `User`s is not enforced.

#### What do you see now?

There is a field in the `User`, but nothing is done with it.

#### What do you want to see instead?

Not sure how exactly we should enforce it and what is a nice user experience.

cc: @johanstokking @bafonins @kschiffer 

#### Can you do this yourself and submit a Pull Request?

Let's discuss first and then come up with an implementation plan.

johanstokking:
We can do it in three places;

1. Login with entering password. I think we should enforce it at least at this point, i.e. after the user typed their current password and before redirecting.

2. Login with cookie. We may enforce it here, but I think this is more related to user session management. If the administrator thinks that revoking sessions is necessary, that should be separate step. In any case, setting the `RequirePasswordUpdate` flag is typically triggered after resetting the password, which optionally comes with revoking sessions anyway, ending up in scenario 1, so I would leave it there

3. On access token refresh. However, not allowing refreshing access tokens because the user requires a password update becomes too obtrusive and may not be nicely handled by OAuth clients. This is also related to user sessions, see 2

PR #1192, opened 2019-08-16
Allow admins to set temporary user password

### LINKED ISSUE #1147: Unable to update password as admin
#### Summary

Unable to update password as admin through CLI

#### Steps to Reproduce

1. `$ ttn-lw-cli user update <user> --password <secret>`
2. `$ ttn-lw-cli user update <user> --temporary-password <secret>`

#### What do you see now?

Forbidden path errors, like:

```
    forbidden_paths=[password]
    correlation_id=de1536c5b64a4059802a65986b2a4e4c
```

Successful password update

#### Environment

`v3.1.0`

Allow `password` and `temporary_password` for admins

Can review

htdvisser:
#### Summary

Closes #1147 if @johanstokking is okay with only allowing setting a temporary password (and not the actual password).

#### Changes

- Changed IS to allow admins to set a temporary password for users
	- Question: Should this trigger an email to the user?
- Some cleanup

#### Notes for Reviewers

I think it's bad practice for admins to be able to set real user passwords, at least until we can force them to change the password (#1191).

The API change doesn't affect external compatibility, it just removes the confusing zero `password_updated_at` from JSON.

#### Release Notes

- Allow admins to create temporary passwords for users.


pkg/identityserver/user_registry.go
```diff
@@ -310,6 +309,24 @@ func (is *IdentityServer) updateUser(ctx context.Context, req *ttnpb.UpdateUserR
...
 	}
 
+	if ttnpb.HasAnyField(req.FieldMask.Paths, "temporary_password") {
+		hashedTemporaryPassword, err := auth.Hash(ctx, req.User.TemporaryPassword)
+		if err != nil {
+			return nil, err
+		}
+		req.User.TemporaryPassword = hashedTemporaryPassword
+		now := time.Now()
+		if !ttnpb.HasAnyField(req.FieldMask.Paths, "temporary_password_created_at") {
```
rvolosatovs: Why allow setting this by the user?
IMO registries should fully manage the `created_at`/`updated_at` fields, hence set this to `now` always.(that's also the behavior of all other registries in `master` - `.*updated_at` and `.*created_at` fields are not allowed to be set)

pkg/identityserver/identityserver_test.go
```diff
@@ -312,3 +312,15 @@ func getIdentityServer(t *testing.T) (*IdentityServer, *grpc.ClientConn) {
 func testWithIdentityServer(t *testing.T, f func(is *IdentityServer, cc *grpc.ClientConn)) {
 	f(getIdentityServer(t))
 }
+
+func reverse(s string) string {
```
rvolosatovs: So the only use for this is to generate a different string from a constant in 5 test cases?
IMO, that's completely redundant - just define a second constant in test case and use that explicitly. That's easier to maintain, clearer and simpler to debug.

pkg/identityserver/user_registry.go
htdvisser: 1. It's an admin doing this, not a user.
2. Disallowing this now would break external API.

pkg/identityserver/identityserver_test.go
htdvisser: No, those are generated by the populators. The reason for this change is that the populators sometimes generated IDs that were already 36 long, and then adding "mod" would make the IDs invalid.

pkg/identityserver/user_registry.go
```diff
@@ -487,8 +505,9 @@ func (is *IdentityServer) updateUserPassword(ctx context.Context, req *ttnpb.Upd
 
 var errTemporaryPasswordStillValid = errors.DefineInvalidArgument("temporary_password_still_valid", "previous temporary password still valid")
 
+var temporaryPasswordValidity = time.Hour * 24
```
johanstokking: This should be (much) longer if there are two actors involved

htdvisser: What do you propose for (A) self-requesting a temporary password and (B) an admin setting one?

htdvisser:
> Should we revoke sessions, just like when the user changes their own password? If the human being doesn't have access to the account anymore, I wonder why we keep allowing machine clients access on behalf of the user.

When self-requesting a temporary password we shouldn't take action on existing tokens, because we haven't verified the identity of the user yet. If we'd do that, an adversary could DoS someones account simply by requesting a temporary password.

For an admin-set temporary password I think we should give the user the choice to revoke existing tokens when setting their new password.

pkg/identityserver/user_registry.go
johanstokking: A day for self-requesting is acceptable, I think even an hour (the old value) is fine too. But when you do it for somebody else considering days off, holidays, time zones, departments, agendas and all the other typical corporate inefficiencies, I would say 36 or 48 hours. Having a temporary password that has been emailed around is not super secure so quick action should be taken.

AWS uses a shorter default by the way;

>By default, temporary security credentials for an IAM user are valid for a maximum of 12 hours. But you can request a duration as short as 15 minutes or as long as 36 hours using the DurationSeconds parameter. For security reasons, a token for an AWS account root user is restricted to a duration of one hour.

johanstokking:
> For an admin-set temporary password I think we should give the user the choice to revoke existing tokens when setting their new password.

What if the user's accounts got compromised, how can the admin take action? Shouldn't it be a flag as well, that's enabled by default?

htdvisser:
That definitely doesn't belong in the `UserRegistry.Update` RPC, but we could create a separate RPC for it.

pkg/identityserver/user_registry.go
htdvisser: So for self-requesting an hour and for admin-set 36 hours?

johanstokking: Sounds good

Issue #1086, opened 2019-08-01
Add prop-types to all react components


bafonins:
#### Summary

A lot of react component in our codebase dont have prop-types.

#### Why do we need this?

Make sure that our components receive correct props, otherwise see errors in the dev tools console.
This will also help us to move to a stricter `eslint` config in the future.

#### What is already there? What do you see now?

A lot of component without full list of prop-types defined.

#### What is missing? What do you want to see?

All components having prop-types

#### Environment

browser (console,oauth)

#### How do you propose to implement this?

Simply define prop-types for each.

#### Can you do this yourself and submit a Pull Request?

yes. help is welcome

cc @kschiffer @pgalic96 @mjamescompton

kschiffer:
Additionally, we need to decide on a way to add these prop-types. We currently mix
```js
static propTypes = { … }
// and
Component.propTypes = { … }
```

a lot. Static prop-types can have the advantage that they take props passed by HOCs into account. On the other hand, static prop-types only work for class components.
Another thing I noticed is that jsdoc comments of static prop-types will not be displayed in storybook, yet that might be fixable (though I wasn't able to fix it quickly last time I tried).

kschiffer:
So let's make this a little more actionable.
I propose the following rules:
- All react components, regardless of view, containers, plain components; regardless of functional or class **must have** prop-types and default-props defined
  - The only exception is components that don't take any props at all
  - The `children` prop does not need to be defined in prop-types / default-props
  - Any prop that is not set as required, needs to have a default-prop and vice-versa
  - We use `PropTypes.shape({ … })` to define expected object shapes in detail
  - We use our `webui/lib/prop-types.js` library to define prop-types and also add recurring type-definitions there
  - We *will* use an eslint config that notifies about missing prop-types
- **For functional components:**
  - We assign prop-types and default props definition as object property `MyFunctionalComponent.propTypes = { … }` / `MyFunctionalComponent.defaultProps = { … }`
  - We do not use default arguments for the functions as replacement for `defaultProps` definitions
- **For class components:**
  - We use the static property of the class `static propTypes = { … }` / `static defaultProps = { … }`
  - We do not use the constructor function to assign these (unless necessary)

Please thumb-up if you agree with these or write a comment about what you'd like to change.
The codebase does obviously not yet adhere to these rules, but we should make sure to work towards this successively.

Additionally, we need an issue to resolve the description bug in storybook for `static` pro-types (see my other comment).

[...]

PR #1197, opened 2019-08-19
Add lint warnings for component prop-types

bafonins:
add warnings for component prop-types

#### Summary

References: https://github.com/TheThingsNetwork/lorawan-stack/issues/1086

#### Changes

- Add eslint rules for prop-types

#### Notes for Reviewers

With this PR we will be able to track issues with prop-types definitions and incrementally resolve those.

### DIFF
--- a/.mage/js.go
+++ b/.mage/js.go
@@ -324,7 +324,9 @@ func (js Js) Lint() error {
 	if err != nil {
 		return err
 	}
-	return eslint("./pkg/webui/**/*.js", "./config/**/*.js", "--no-ignore", "--color")
+	// TODO: Remove the `--quiet` flag after all component prop-types are defined
+	// (https://github.com/TheThingsNetwork/lorawan-stack/issues/1086)
+	return eslint("./pkg/webui/**/*.js", "./config/**/*.js", "--no-ignore", "--color", "--quiet")
 }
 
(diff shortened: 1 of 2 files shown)


kschiffer:
We should be aware that if we should reintroduce linting to our githooks (which we should do soon), we would see very lengthy warning output for every commit.

bafonins:
> Likewise, these will be output in the CI log for every build.

We can supress `eslint` warnings with the `--quiet` flag. Shall we this until https://github.com/TheThingsNetwork/lorawan-stack/issues/1086 is resolved?

kschiffer:
I think that would be good indeed.

bafonins:
@htdvisser please check `js.go`

.mage/js.go
```diff
@@ -324,7 +324,9 @@ func (js Js) Lint() error {
 	if err != nil {
 		return err
 	}
-	return eslint("./pkg/webui/**/*.js", "./config/**/*.js", "--no-ignore", "--color")
+	// TODO: Remove `--quiet` after https://github.com/TheThingsNetwork/lorawan-stack/issues/1086
```
rvolosatovs: Usually we make this of form `TODO: <Action> (<issue URL>)`

PR #1198, opened 2019-08-19
Support GS Upstream Interface

KrishnaIyer:
#### Summary

References: https://github.com/TheThingsIndustries/lorawan-stack/pull/1685#issuecomment-522545622

#### Changes

- Add Common Upstream interface
- Adapt the cluster NS.
- Move Downlink claiming to Upstream instead of frontends.
- Fix wrong context cancellation in io.Connection.

#### Release Notes

- Added GS Upstream interface.

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -370,7 +370,7 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids 
 		return nil, err
 	}
 
-	conn := io.NewConnection(ctx, cancelCtx, frontend.Protocol(), gtw, fp, scheduler)
+	conn := io.NewConnection(ctx, cancelCtx, frontend.Protocol(), frontend.SupportsStatusMessage(), gtw, fp, scheduler)
 	gs.connections.Store(uid, conn)
 	registerGatewayConnect(ctx, ids)
 	logger.Info("Connected")
@@ -379,7 +379,7 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids 
 	for _, handler := range gs.upstreamHandlers {
 		go func(handler upstream.Handler) {
 			logger := log.FromContext(ctx).WithField("handler", handler.GetName())
-			if err := handler.ConnectGateway(ctx, uid, conn); err != nil {
+			if err := handler.ConnectGateway(ctx, ids, conn); err != nil {
 				logger.WithError(err).Warn("Gateway connect errored on upstream")
 			}
 		}(handler)
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -100,7 +100,7 @@ var (
 		"failed to start frontend listener `{protocol}` on address `{address}`",
 	)
 	errNotConnected  = errors.DefineNotFound("not_connected", "gateway `{gateway_uid}` not connected")
-	errStartUpstream = errors.DefineFailedPrecondition("upstream", "upstream `{name}` failed to start")
+	errSetupUpstream = errors.DefineFailedPrecondition("upstream", "failed to setup upstream `{name}`")
 	errUpstreamType  = errors.DefineUnimplemented("upstream_type_not_implemented", "upstream `{name}` not implemented")
 )
 
@@ -221,7 +221,7 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 		if name == "" {
 			gs.upstreamHandlers["cluster"] = ns.NewHandler(ctx, "cluster", c, prefix)
 		} else {
-			str := strings.Split(name, ":")
+			str := strings.SplitN(name, ":", 2)
 			if len(str) != 2 {
 				continue
 			}
@@ -236,7 +236,7 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 
 	for _, handler := range gs.upstreamHandlers {
 		if err := handler.Setup(); err != nil {
-			return nil, errStartUpstream.WithCause(err).WithAttributes("name", handler.GetName())
+			return nil, errSetupUpstream.WithCause(err).WithAttributes("name", handler.GetName())
 		}
 	}
 
@@ -369,7 +369,7 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids 
 		return nil, err
 	}
 
-	conn, err := io.NewConnection(ctx, frontend.Protocol(), frontend.SupportsStatusMessage(), frontend.SupportsDownlinkClaim(), gtw, fp, gtw.EnforceDutyCycle)
+	conn, err := io.NewConnection(ctx, frontend, gtw, fp, gtw.EnforceDutyCycle)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -113,7 +113,7 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 		return nil, err
 	}
 	if len(forward) == 0 {
-		forward[""] = []types.DevAddrPrefix{{}}
+		forward["ttn.lorawan.v3.GsNs:cluster"] = []types.DevAddrPrefix{{}}
 	}
 	gs = &GatewayServer{
 		Component:                 c,
@@ -225,7 +225,7 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 		}
 		switch str[0] {
 		case "ttn.lorawan.v3.GsNs":
-			gs.upstreamHandlers[str[1]] = ns.NewHandler(ctx, str[1], *c, prefix)
+			gs.upstreamHandlers[str[1]] = ns.NewHandler(ctx, str[1], c, prefix)
 		default:
 			return nil, errUpstreamType.WithAttributes("name", name)
 		}
@@ -378,8 +378,9 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids 
 
 	for _, handler := range gs.upstreamHandlers {
 		go func(handler upstream.Handler) {
-			if err := handler.ConnectGateway(ctx, conn); err != nil {
-				logger.WithField("handler", handler.GetName()).Warn("Gateway connect errored on upstream")
+			logger := log.FromContext(ctx).WithField("handler", handler.GetName())
+			if err := handler.ConnectGateway(ctx, uid, conn); err != nil {
+				logger.WithError(err).Warn("Gateway connect errored on upstream")
 			}
 		}(handler)
 	}
@@ -430,12 +431,7 @@ func (gs *GatewayServer) handleUpstream(conn *io.Connection) {
 	logger := log.FromContext(ctx)
 	defer func() {
 		ids := conn.Gateway().GatewayIdentifiers
-		uid := unique.ID(ctx, ids)
 		gs.connections.Delete(unique.ID(ctx, ids))
-		gs.UnclaimDownlink(ctx, ids)
-		for _, handler := range gs.upstreamHandlers {
-			handler.DisconnectGateway(ctx, uid)
-		}
 		registerGatewayDisconnect(ctx, ids)
 		logger.Info("Disconnected")
 	}()
@@ -600,14 +596,4 @@ func (gs *GatewayServer) GetFrequencyPlan(ctx context.Context, ids ttnpb.Gateway
 		return nil, err
 	}
 	return gs.FrequencyPlans.GetByID(fpID)
-}
-
-// ClaimDownlink claims the downlink path for the given gateway.
-func (gs *GatewayServer) ClaimDownlink(ctx context.Context, ids ttnpb.GatewayIdentifiers) error {
-	return gs.ClaimIDs(ctx, ids)
-}
-
-// UnclaimDownlink releases the claim of the downlink path for the given gateway.
-func (gs *GatewayServer) UnclaimDownlink(ctx context.Context, ids ttnpb.GatewayIdentifiers) error {
-	return gs.UnclaimIDs(ctx, ids)
 }
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -99,9 +99,10 @@ var (
 		"listen_frontend",
 		"failed to start frontend listener `{protocol}` on address `{address}`",
 	)
-	errNotConnected  = errors.DefineNotFound("not_connected", "gateway `{gateway_uid}` not connected")
-	errSetupUpstream = errors.DefineFailedPrecondition("upstream", "failed to setup upstream `{hostname}`")
-	errUpstreamType  = errors.DefineUnimplemented("upstream_type_not_implemented", "upstream `{hostname}` not implemented")
+	errNotConnected        = errors.DefineNotFound("not_connected", "gateway `{gateway_uid}` not connected")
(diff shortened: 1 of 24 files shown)


pkg/gatewayserver/gatewayserver.go
```diff
@@ -206,6 +214,24 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 
 	hooks.RegisterUnaryHook("/ttn.lorawan.v3.NsGs", cluster.HookName, c.ClusterAuthUnaryHook())
 
+	for name, prefix := range gs.forward {
+		str := strings.Split(name, ":")
+		if len(str) != 2 {
+			continue
```
htdvisser: Can't break config within v3 major, so we need to handle legacy config here (a `name` of `""`)

pkg/gatewayserver/upstream/ns/ns.go
```diff
@@ -0,0 +1,91 @@
...
+	ctx             context.Context
+	name            string
+	c               *component.Component
+	devAddrPrefixes []types.DevAddrPrefix
+}
+
+var errNotFound = errors.DefineNotFound("network_server_not_found", "network server not found for ids `ids`")
+
+// NewHandler returns a new upstream handler.
+func NewHandler(ctx context.Context, name string, c *component.Component, devAddrPrefixes []types.DevAddrPrefix) *Handler {
```
htdvisser: The `name` argument can be a host according to the documentation of the corresponding field in `pkg/gatewayserver.Config`

pkg/gatewayserver/upstream/ns/ns.go
```diff
@@ -0,0 +1,91 @@
...
+
+// HandleUp implements upstream.Handler.
+func (h *Handler) HandleUp(ctx context.Context, gatewayUID string, ids ttnpb.EndDeviceIdentifiers, msg *ttnpb.GatewayUp) error {
+	ns := h.c.GetPeer(ctx, ttnpb.PeerInfo_NETWORK_SERVER, ids)
+	if ns == nil {
+		return errNotFound.WithAttributes("ids", ids)
+	}
+	client := ttnpb.NewGsNsClient(ns.Conn())
+	for _, up := range msg.UplinkMessages {
+		_, err := client.HandleUplink(ctx, up, h.c.WithClusterAuth())
```
htdvisser: Only use Cluster Auth if `h.name` equals `"cluster"`.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -102,13 +109,14 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 		return nil, err
 	}
 	if len(forward) == 0 {
-		forward[""] = []types.DevAddrPrefix{{}}
+		forward["ttn.lorawan.v3.GsNs:cluster"] = []types.DevAddrPrefix{{}}
```
johanstokking: For clarity, leave this to `""` and explicitly use this interface for empty. See https://github.com/TheThingsNetwork/lorawan-stack/pull/1198/files?file-filters%5B%5D=.go#r315555734

pkg/gatewayserver/gatewayserver.go
```diff
@@ -206,6 +214,24 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
...
+		switch str[0] {
+		case "ttn.lorawan.v3.GsNs":
+			gs.upstreamHandlers[str[1]] = ns.NewHandler(ctx, str[1], c, prefix)
+		default:
+			return nil, errUpstreamType.WithAttributes("name", name)
+		}
+	}
+	for _, handler := range gs.upstreamHandlers {
+		if err := handler.Setup(); err != nil {
+			return nil, errStartUpstream.WithCause(err).WithAttributes("name", handler.GetName())
```
johanstokking: Is it setup or start? Consistency is key for your future self and other developers

pkg/gatewayserver/gatewayserver.go
```diff
@@ -331,11 +358,21 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids
...
 	gs.connections.Store(uid, conn)
 	registerGatewayConnect(ctx, ids)
 	logger.Info("Connected")
 	go gs.handleUpstream(conn)
+
+	for _, handler := range gs.upstreamHandlers {
+		go func(handler upstream.Handler) {
+			logger := log.FromContext(ctx).WithField("handler", handler.GetName())
+			if err := handler.ConnectGateway(ctx, uid, conn); err != nil {
+				logger.WithError(err).Warn("Gateway connect errored on upstream")
```
johanstokking: `Failed to connect gateway on upstream`

pkg/gatewayserver/io/io.go
```diff
@@ -112,9 +107,12 @@ func (c *Connection) Disconnect(err error) {
 	c.cancelCtx(err)
 }
 
-// Protocol returns the protocol used for the connection, i.e. grpc, mqtt or udp.
+// Protocol returns the protocol used for the connection.
 func (c *Connection) Protocol() string { return c.protocol }
 
+// SupportsStatusMessage returns true if the protocol supports status messages.
+func (c *Connection) SupportsStatusMessage() bool { return c.protocol != "basicstation" }
```
johanstokking: Reiterating; no string typing. Move this to the frontend

pkg/gatewayserver/io/io.go
```diff
@@ -86,8 +82,7 @@ type Connection struct {
 }
 
 // NewConnection instantiates a new gateway connection.
-func NewConnection(ctx context.Context, protocol string, gateway *ttnpb.Gateway, fp *frequencyplans.FrequencyPlan, scheduler *scheduling.Scheduler) *Connection {
-	ctx, cancelCtx := errorcontext.New(ctx)
+func NewConnection(ctx context.Context, cancelCtx errorcontext.CancelFunc, protocol string, gateway *ttnpb.Gateway, fp *frequencyplans.FrequencyPlan, scheduler *scheduling.Scheduler) *Connection {
```
johanstokking: This looks like a dirty work-around for sharing context cancellation. Imo, the context lifetime ownership should not be shared. If `Connection` knows about cancellation (I think it should, like it was), it may expose `Context() context.Context` on which anyone can wait for done and an error. If there's an external actor that needs to disconnect (with an error), there's `Disconnect(error)` which `Connection` basically uses to `cancelCtx(err)`

cc @htdvisser please comment if you disagree

pkg/gatewayserver/gatewayserver.go
```diff
@@ -331,11 +358,21 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids
 		return nil, err
 	}
 
-	conn := io.NewConnection(ctx, frontend.Protocol(), gtw, fp, scheduler)
+	conn := io.NewConnection(ctx, cancelCtx, frontend.Protocol(), gtw, fp, scheduler)
```
johanstokking: So after this call, if you want to use the connection context, just take `conn.Context()`, or even re-assign the local `ctx`

pkg/gatewayserver/io/io.go
```diff
@@ -112,9 +107,12 @@ func (c *Connection) Disconnect(err error) {
 	c.cancelCtx(err)
 }
 
-// Protocol returns the protocol used for the connection, i.e. grpc, mqtt or udp.
+// Protocol returns the protocol used for the connection.
 func (c *Connection) Protocol() string { return c.protocol }
 
+// SupportsStatusMessage returns true if the protocol supports status messages.
```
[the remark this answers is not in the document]
KrishnaIyer: It's a replacement for this: https://github.com/TheThingsIndustries/lorawan-stack/blob/05325b5018d87b4bda4f134e75bdff61eab1069a/pkg/gatewayserver/upstream/v2/v2.go#L241

pkg/gatewayserver/gatewayserver.go
KrishnaIyer: `setup` was what we agreed on in the TTI draft PR.

pkg/gatewayserver/io/io.go
KrishnaIyer: Agreed. That was indeed a quick-fix to solve the leak in the PCN bridge. I've updated it now.

pkg/gatewayserver/upstream/ns/ns.go
KrishnaIyer: How about for non-cluster NSes? Or is that to-do for the future?

htdvisser: We're not going to send cluster auth to those
Still TODO yes

KrishnaIyer:
@johanstokking: For UDP, I've reintroduced the claim/unclaim functions on the GS and added `SupportsDownlinkClaim` on the frontend.
For UDP, the front-end claims/unclaims the downlink. For other, it's connection/disconnection based.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -206,6 +214,24 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 
 	hooks.RegisterUnaryHook("/ttn.lorawan.v3.NsGs", cluster.HookName, c.ClusterAuthUnaryHook())
 
+	for name, prefix := range gs.forward {
+		str := strings.Split(name, ":")
```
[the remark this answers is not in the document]
KrishnaIyer: Why? `SplitN` doesn't return an error so I can't enforce `type:name`.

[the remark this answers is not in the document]
htdvisser: As far as I remember, the original config was an address, which could be IPv6 or host:port. By splitting everything (instead of just splitting the `type` of `type:rest`) you'll lose that information.

KrishnaIyer:
Blocks https://github.com/TheThingsIndustries/lorawan-stack/pull/1739

pkg/gatewayserver/gatewayserver.go
johanstokking: OK, so `errSetupUpstream`

pkg/gatewayserver/gatewayserver.go
```diff
@@ -210,6 +217,29 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 
 	hooks.RegisterUnaryHook("/ttn.lorawan.v3.NsGs", cluster.HookName, c.ClusterAuthUnaryHook())
 
+	for name, prefix := range gs.forward {
+		if name == "" {
+			gs.upstreamHandlers["cluster"] = ns.NewHandler(ctx, "cluster", c, prefix)
+		} else {
+			str := strings.Split(name, ":")
```
johanstokking: Also `SplitN` here

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -160,7 +162,7 @@ func (s *srv) handleDiscover(c echo.Context) error {
 
 var euiHexPattern = regexp.MustCompile("^eui-([a-f0-9A-F]{16})$")
 
-func (s *srv) handleTraffic(c echo.Context) error {
+func (s *srv) handleTraffic(c echo.Context) (retErr error) {
```
johanstokking: `err`

pkg/gatewayserver/io/io.go
```diff
@@ -86,22 +92,28 @@ type Connection struct {
 }
 
 // NewConnection instantiates a new gateway connection.
-func NewConnection(ctx context.Context, protocol string, gateway *ttnpb.Gateway, fp *frequencyplans.FrequencyPlan, scheduler *scheduling.Scheduler) *Connection {
+func NewConnection(ctx context.Context, protocol string, supportsStatusMessage bool, supportsDownlinkClaim bool, gateway *ttnpb.Gateway, fp *frequencyplans.FrequencyPlan, enforceDutyCycle bool) (*Connection, error) {
```
johanstokking: Pass interface exposing these behavioral properties

pkg/gatewayserver/io/io.go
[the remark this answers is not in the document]
johanstokking: But isn't this provided by the frontend, not by the connection? Shouldn't we expose the frontend capabilities through the connection instead?

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
KrishnaIyer: If I use `err`, the current error will be referenced:
https://github.com/TheThingsNetwork/lorawan-stack/pull/1198#discussion_r315551032

pkg/gatewayserver/gatewayserver.go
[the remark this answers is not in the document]
KrishnaIyer: Ah I see. Didn't know that. Updated.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
htdvisser: Ehh no, that's not true. The problem in the other thread was that you bind the `err` of the defer to the current value at that point. 

```go
	defer func(error) {
		conn.Disconnect(err)
	}(err)
```

Instead you should have done

```go
	defer func() {
		conn.Disconnect(err)
	}()
```

pkg/gatewayserver/gatewayserver.go
```diff
@@ -210,6 +217,29 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 
 	hooks.RegisterUnaryHook("/ttn.lorawan.v3.NsGs", cluster.HookName, c.ClusterAuthUnaryHook())
 
+	for name, prefix := range gs.forward {
+		if name == "" {
+			gs.upstreamHandlers["cluster"] = ns.NewHandler(ctx, "cluster", c, prefix)
+		} else {
+			str := strings.SplitN(name, ":", 2)
+			if len(str) != 2 {
+				continue
```
htdvisser: I think we should just error on invalid config, right?

pkg/gatewayserver/upstream/ns/ns.go
htdvisser: My point was that it's not a name

pkg/gatewayserver/gatewayserver.go
htdvisser: Still open

pkg/gatewayserver/gatewayserver.go
```diff
@@ -210,6 +217,29 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 
 	hooks.RegisterUnaryHook("/ttn.lorawan.v3.NsGs", cluster.HookName, c.ClusterAuthUnaryHook())
 
+	for hostname, prefix := range gs.forward {
```
htdvisser: I think calling `ttn.lorawan.v3.GsNs:cluster` a hostname is a bit confusing

pkg/gatewayserver/gatewayserver.go
```diff
@@ -210,6 +217,29 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
...
+		} else {
+			str := strings.SplitN(hostname, ":", 2)
+			if len(str) != 2 {
+				continue
+			}
+			switch str[0] {
+			case "ttn.lorawan.v3.GsNs":
+				gs.upstreamHandlers[str[1]] = ns.NewHandler(ctx, str[1], c, prefix)
+			default:
+				return nil, errUpstreamType.WithAttributes("hostname", hostname)
```
htdvisser: `str[1]` is the hostname, `hostname` isn't actually a hostname (see above)

pkg/gatewayserver/gatewayserver.go
KrishnaIyer: Yeah ok. I guess that would make it consistent with other config handling.

KrishnaIyer:
Thanks @johanstokking, will merge once I test on metal.

johanstokking:
Please cleanup commit trail before merging

Issue #30, opened 2019-01-29
JavaScript SDK


johanstokking:
**Summary:** 
Since we would need abstractions on the stack API for the Console anyway, we decided to start building the JavaScript SDK alongside developing the Console. Here, it is planned to combine usage for the browser and Node.JS in a single SDK. The usage of the HTTP Bridge or gRPC directly can then be chosen as a config during instantiation of the SDK API object, depending on the capabilities of the given environment.
The SDK will implement a stack API abstraction that provides the same normalized interface for both configurations. For HTTP, a mapping from rpc name to HTTP request is generated by the build process which is used to generate function definitions. For gRPC the generated JS stubs are used.
The user will be able to use either the higher-level SDK API which provides simplified access to entities (like `ttn.Applications.getById(id)`) or call the underlying rpcs directly, resolved to HTTP or gRPC by the SDK.
Following our mono-repo idea, the SDK will be part of this repo and published isolated via npm, using the old (to be updated) SDK repo for issues.

**What is already there? What do you see now?**
- Groundwork
- (Rough) Draft of SDKs API / Architecture
- Generating HTTP API interface from a protobuf generated JSON mapping

**What is missing? What do you want to see?**
_Note: this is still incomplete / wip_
- [ ] Connecting to the stack API
  - [ ] Providing gRPC connection type
  - [x] Providing HTTP connection type
  - [ ] Normalized API that resolves to either connection type
- [x] Marshalling API specific things like `field/update_mask`s
- [ ] Handling authentication
  - [x] Using an access token
  - [ ] Using API Keys
- [x] Abstraction for handling Applications
  - [x] CRUD
  - [x] Handling collaborators
  - [x] Handling API Keys
- [ ] Abstraction for handling Devices (via Application interface)
  - [x] CRUD (See #131)
  - [x] Handling collaborators
  - [x] Streaming data
  - [ ] Handling downlink/uplink queues
- [x] Abstraction for handling Gateways
  - [x] CRUD
  - [ ] Streaming data
  - [ ] Handling API Keys
- [x] General Event Stream API
- [ ] CI
  - [ ] Making sure API changes result in rebuild and republish of SDK
  - [ ] Testing SDKs conformity with API
  - [ ] Switch from preliminary API mapping at runtime to generated JS files
- [ ] npm publishing workflow
  - [ ] Provide repository for issues by the community
  - [ ] Add npm publish as part of CI
  - [ ] Synchronize versioning with stack

[...]

PR #1202, opened 2019-08-20
Format and lint JS SDK code

bafonins:
#### Summary

This PR adds code linting and formatting to the js sdk.

#### Changes

- Add `prettier` 
- Add `eslint`
- Format js sdk code with prettier
- Add `jsSDK:lint` and `jsSDK:fmt` mage targets
- Include linting js sdk during ci

#### Notes for Reviewers

@htdvisser and @rvolosatovs  as always, only the `mage/ci` parts

### DIFF
--- a/.mage/js_sdk.go
+++ b/.mage/js_sdk.go
@@ -90,6 +90,21 @@ func (k JsSDK) Watch() error {
 	return yarn("run", "build:watch")
 }
 
+// Fmt formats all js files.
+func (k JsSDK) Fmt() error {
+	mg.Deps(JsSDK.Deps)
+
+	if mg.Verbose() {
+		fmt.Println("Running prettier on sdk .js files")
+	}
+	yarn, err := k.yarn()
+	if err != nil {
+		return err
+	}
+
+	return yarn("run", "fmt")
+}
+
 // Test runs jest unit tests.
 func (k JsSDK) Test() error {
 	if mg.Verbose() {
--- a/.mage/js_sdk.go
+++ b/.mage/js_sdk.go
@@ -105,6 +105,21 @@ func (k JsSDK) Fmt() error {
 	return yarn("run", "fmt")
 }
 
+// Lint runs eslint over sdk js files.
+func (k JsSDK) Lint() error {
+	mg.Deps(JsSDK.Deps)
+
+	if mg.Verbose() {
+		fmt.Println("Running eslint on sdk .js files")
+	}
+	yarn, err := k.yarn()
+	if err != nil {
+		return err
+	}
+
+	return yarn("run", "lint")
+}
+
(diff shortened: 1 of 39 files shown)


.mage/js_sdk.go
```diff
@@ -90,6 +90,36 @@ func (k JsSDK) Watch() error {
...
+
+	if mg.Verbose() {
+		fmt.Println("Running prettier on sdk .js files")
+	}
+	yarn, err := k.yarn()
+	if err != nil {
+		return err
+	}
+
+	return yarn("run", "fmt")
```
rvolosatovs: Why not define `func (k JsSDK) runYarn(string) error` and do
```go
return k.yarnRun(<cmd>) // <cmd> would be "fmt" here, for example
```

instead of the whole
```go
	yarn, err := k.yarn()
	if err != nil {
		return err
	}

	return yarn("run", <cmd>)
```
block in most of these methods?

.travis.yml
```diff
@@ -9,137 +9,137 @@ go: 1.12.x
...
-  - NODE_ENV=production
-  - YARN_CACHE_FOLDER=$HOME/.cache/yarn
-  - MAGE=$HOME/.cache/mage/mage
-  - TEST_SLOWDOWN=8
-  - TEST_REDIS=1
-  - PATH=/snap/bin:$PATH
-  - HUGO_BASE_URL=https://thethingsnetwork.github.io/lorawan-stack/
-  - DOCKER_IMAGE=thethingsnetwork/lorawan-stack
-  - DOCKER_IMAGE_DEV=thethingsnetwork/lorawan-stack-dev
+    - GOPROXY=https://proxy.golang.org
```
rvolosatovs: unrelated?

.travis.yml
```diff
@@ -9,137 +9,137 @@ go: 1.12.x
...
 services:
-- docker
+  - docker
 cache:
   directories:
-  - "$GOPATH/pkg/mod"
-  - "$HOME/.cache/go-build"
-  - "$HOME/.cache/mage"
-  - "$HOME/.cache/yarn"
+    - '$GOPATH/pkg/mod'
```
rvolosatovs: Why?

.mage/js_sdk.go
bafonins: not in this pr. we will revisit this part when publishing the sdk to the npm registry.

rvolosatovs: Can you add a TODO with an issue reference?

bafonins: no. your proposed improvement has nothing to with this specific pr. `Lint()` and `Fmt()` follow the structure that was implemented before.

see https://github.com/TheThingsNetwork/lorawan-stack/issues/30#issuecomment-522974595

Issue #939, opened 2019-07-04
[Concept] Collaborator Rights Handling


htdvisser:
#### Summary

This is an overview of how collaborator rights should be handled.

#### What do you want to see?

- [ ] IS: When an Application is created, the first collaborator should have `RIGHT_APPLICATION_ALL`
- [ ] IS: When a Client is created, the first collaborator should have `RIGHT_CLIENT_ALL`
- [ ] IS: When a Gateway is created, the first collaborator should have `RIGHT_GATEWAY_ALL`
- [ ] IS: When an Organization is created, the first collaborator should have `RIGHT_ALL` ⚠
- [ ] Console: Having universal rights does **not** mean a user is admin. Use the `is_admin` bool in the response of `/api/v3/auth_info` for that.
- [ ] Console: Non-admin users can assign/revoke rights on their entities to admin users
- [ ] Console: A user can see and manage collaborators (only) if they have `RIGHT_{ENTITY}_SETTINGS_COLLABORATORS` (the API key story is pretty much the same)
    - [ ] The current user can assign at most the rights that they have themselves to other users or organizations (see `/api/v3/{entity}s/{id}/rights`)
    - [ ] The current user can revoke the rights of any collaborator, regardless of their rights
    - [ ] If an existing collaborator has more rights than the current user (again based on `/api/v3/{entity}s/{id}/rights`), we can display a warning when editing that other collaborator's rights, but we should not prevent it.
    - [ ] The `RIGHT_ALL` pseudo-right implies `RIGHT_*_ALL`
    - [ ] The `RIGHT_{ENTITY}_ALL` pseudo-right implies `RIGHT_{ENTITY}_*`
    - [ ] When deselecting `RIGHT_ALL` or `RIGHT_{ENTITY}_ALL` you can expand+select the matching rights of the current user (again based on `/api/v3/{entity}s/{id}/rights`)

@kschiffer please let me know if there are other things that need to be specified.

kschiffer:
> If an existing collaborator has more rights than the current user (again based on `/api/v3/{entity}s/{id}/rights`), we can display a warning when editing that other collaborator's rights, but we should not prevent it.

One problem I have identified here is that if User A has rights that are outside of the scope of granted rights for User B, User B cannot modify the rights of User A properly because:
- Sending the out-of-scope rights together with the legitimately altered rights will result in a 403 Forbidden
[...]

PR #1203, opened 2019-08-20
Remove unnecessary constraints from collaborator management

kschiffer:
#### Summary
References #939.
This PR removes some constraints from the collaborator form in the console. These constraints have been wrongfully added by me, due to a lack of understanding of how our right intersection works. This has been cleared up in #939.

#### Changes
- Remove a constraint that forbids right modification of users with universal rights (and wrongfully referring to these rights as `admin rights`)
- Modify the logic for correctly handling collaborator rights that are out of the scope of the current user
  - Disable checkboxes for out-of-scope rights
  - Allow modification of remaining rights
  - Fix out-of-scope rights disappearing when unchecked
  - Disallow right modification for users with universal rights (while not having this right oneself), but still allow removal as collaborator, see #1364 
- While at it, fixing prop type warnings of the respective components (#1086) 

#### Notes for Reviewers
This PR will for now only remove unnecessary restrictions when modifying right, without changing the UI logic. I propose to continue on improving the UX here as discussed in https://github.com/TheThingsNetwork/lorawan-stack/issues/1204#issuecomment-528854149 in a separate PR.

#### Release Notes
- Removed unnecessary constraints for the right management of collaborators

cc @htdvisser


htdvisser:
I don't think we should use the term "scope" here. Instead of _"This entity possesses rights that are out of your scope of granted rights."_, can we do something like _"This {account_type} has rights that you don't have"_?

pkg/webui/components/checkbox/checkbox.js
```diff
@@ -131,36 +137,37 @@ class Checkbox extends React.PureComponent {
...
-  label: PropTypes.message,
-  name: PropTypes.string.isRequired,
-  value: PropTypes.oneOfType([PropTypes.bool, PropTypes.object]),
-  checked: PropTypes.bool,
   disabled: PropTypes.bool,
-  readOnly: PropTypes.bool,
-  autoFocus: PropTypes.bool,
-  onChange: PropTypes.func,
+  indeterminate: PropTypes.bool,
+  label: PropTypes.oneOfType([PropTypes.message, PropTypes.node]),
```
bafonins: use `children` for custom content

pkg/webui/console/components/rights-group/index.js
```diff
@@ -140,15 +146,21 @@ class RightsGroup extends React.Component {
         className={style.rightLabel}
         key={right}
         name={right}
-        label={{ id: `enum:${right}` }}
         disabled={outOfOwnScopeRights.includes(right)}
+        label={
+          <React.Fragment>
+            <Message content={{ id: `enum:${right}` }} />{' '}
+            {outOfOwnScopeRights.includes(right) && (
+              <Message className={style.revoke} content={m.revocableOnly} />
```
bafonins: I would prefer making this a custom checkbox component

pkg/webui/console/components/rights-group/index.js
```diff
@@ -50,41 +52,45 @@ const computeState = function(values, rights, universalRight) {
...
+    const { value, rights: grantableRights, universalRight: grantableUniversalRight } = props
+    const { outOfOwnScopeRights } = state
+    let universalRight = grantableUniversalRight
+    const allGrantableRights = [...grantableRights, ...grantableUniversalRight]
+    let derivedOutOfOwnScopeRights = outOfOwnScopeRights
+
+    // Initially identify given values that are out of the scope of the
+    // current user, to be able to persist them in the rights list
+    if (outOfOwnScopeRights.length === 0) {
+      derivedOutOfOwnScopeRights = value.filter(function(right) {
```
bafonins: this is hard to read. could you try refactoring it a bit?

pkg/webui/console/components/rights-group/index.js
kschiffer: I think to pass the additional content as children will simplify this enough.

kschiffer:
> The 'check all' checkbox label changes when uncheking and checking again:
> Screenshot 2019-08-22 at 08 48 24

How did you get to this state?

> Also, is it expected behaviour that the user with several rights can update rights of the creator of the app (having all possible rights and being an admin), but cannot change rights of the user that has just more rights?

See #1204. This will be changed as soon as the issue is resolved.

bafonins:
> How did you get to this state?

Create a new user with several rights and try to edit the user with all rights

bafonins:
blocked by https://github.com/TheThingsNetwork/lorawan-stack/pull/1345

kschiffer:
Updated OP. Changes resulting from #1345 are now integrated.
> This PR will for now only remove unnecessary restrictions when modifying right, without changing the UI mechanic. I propose to continue on improving the UX here as discussed in https://github.com/TheThingsNetwork/lorawan-stack/issues/1204#issuecomment-528854149 in a separate PR.

PR #1208, opened 2019-08-21
Update CORS config

### LINKED ISSUE #1207: Fix broken CORS
#### Summary
#1151 introduced a bug that broke CORS for the gRPC HTTP bridge. This means that browser XHR requests coming from external origins don't work anymore.

#### Steps to Reproduce
The easiest way to reproduce is by running the console in [a development setup](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.0.4/DEVELOPMENT.md#development). In development, we launch the console from `localhost:8080`, while requesting the API on `localhost:1885` (hence cors requirement)
1. In the console try to login
2. See the console looping you back to the login screen
3. In the dev console, go to the `Network` tab
4. See that the request to `localhost:1885/api/v3/auth_info` fails (stalls without response)

#### What do you see now?
XHR requests to the stack API from external origins failing

These requests to run through without issues (as it was the case before https://github.com/TheThingsNetwork/lorawan-stack/commit/589c216b36f64b7fea9dda84a09d6f6b9a375374)

#### Environment
Console in development mode

Likely an issue in here:
https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/component/grpc.go#L58-L66
But I don't have time to look at it in detail atm.

Rather @htdvisser or @bafonins

htdvisser:
#### Summary

This Pull Request closes #1207.

#### Changes

- Allow `Authorization` and `X-CSRF-Token` request headers
- Allow JavaScript code to read responses to requests sent with credentials
- Update list of response headers exposed to JavaScript
- Set validity of CORS preflight requests to 10 minutes

#### Notes for Reviewers

@kschiffer please confirm that this really fixes the issue for you

No release notes since this fixes an issue that wasn't present in a release yet.

### DIFF
--- a/pkg/component/grpc.go
+++ b/pkg/component/grpc.go
@@ -59,10 +59,10 @@ func (c *Component) setupGRPC() (err error) {
 		"/*",
 		echo.WrapHandler(http.StripPrefix(ttnpb.HTTPAPIPrefix, c.grpc)),
 		middleware.CORSWithConfig(middleware.CORSConfig{
-			Skipper:      middleware.DefaultSkipper,
-			AllowOrigins: []string{"*"},
-			AllowMethods: []string{http.MethodGet, http.MethodHead, http.MethodPut, http.MethodPatch, http.MethodPost, http.MethodDelete},
-			AllowHeaders: []string{"Date", "Content-Length", "X-Total-Count", "X-Warning"},
+			AllowHeaders:     []string{"Authorization", "Content-Type", "X-CSRF-Token"},
+			AllowCredentials: true,
+			ExposeHeaders:    []string{"Date", "Content-Length", "X-Request-Id", "X-Total-Count", "X-Warning"},
+			MaxAge:           600,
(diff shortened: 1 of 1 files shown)


htdvisser:
Ok, found it. The error is "Reason: missing token 'content-type' in CORS header".

htdvisser:
[MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers) writes "The CORS-safelisted request headers, Accept, Accept-Language, Content-Language, Content-Type are always allowed and don't need to be listed by this header necessarily." but apparently there are some exceptions, including one for `Content-Type` that only allows `application/x-www-form-urlencoded`, `multipart/form-data`, or `text/plain`, which means that `application/json` is not allowed. I'll explicitly allow it then.

htdvisser:
@kschiffer can you check the frontend code for any other headers that it might send?

Issue #654, opened 2019-05-07
Default value handling in console / sdk


kschiffer:
#### Summary
Following up the discussion starting here: https://github.com/TheThingsNetwork/lorawan-stack/pull/639#discussion_r281178795, we need to discuss handling default values in the client (console / sdk)

#### Why do we need this?
Consistency; better User and Developer Experience

#### What is already there? What do you see now?
- Default values handled by the backend, see eg here: #47 
- Default values being set in the CLI, eg: https://github.com/TheThingsNetwork/lorawan-stack/blob/master/cmd/ttn-lw-cli/commands/end_devices.go#L310-L318
- Default values being set by the JS SDK: https://github.com/TheThingsNetwork/lorawan-stack/blob/master/sdk/js/src/service/devices/index.js#L193-L200
- Default values being set by the console, to be displayed in fields: https://github.com/TheThingsNetwork/lorawan-stack/pull/639
- Generated values being communicated with empty fields and placeholders:
![image](https://user-images.githubusercontent.com/5710611/57289830-6b6d5280-70bc-11e9-8600-83183b8e376e.png)

#### What is missing? What do you want to see?
- A clear strategy on handling default values (should be set in SDK or console?)
- A clear strategy on communicating default values in the console (should be prefilled fields or empty with placeholder message?)

#### Environment
Various.

#### How do you propose to implement this?
Subject of discussion.

Copy of [my comment in the PR](https://github.com/TheThingsNetwork/lorawan-stack/pull/639#discussion_r281523080):
> This is about whether to set default values in the (JS) SDK and/or in the console itself (like server addresses here, which are not set by the backend, but we do have an idea of what a default should be in a specific context). Whereas I think we should do both, because (non console) SDK users will likely want to make use of that functionality as well. In the console we can determine these default values too and use them to prefill form values, to improve clarity.

> A special case would be randomly generated values which can be generated either by the client (SDK / console) or also by the backend (to ensure validity in eg. DevAddr). Here, in any case, I think showing the value up front to the user does not add any value and can be confusing rather.

#### Can you do this yourself and submit a Pull Request?
Is a shared responsibility.

cc @htdvisser

kschiffer:
Copying [comment](https://github.com/TheThingsNetwork/lorawan-stack/pull/639#discussion_r281543345) of @htdvisser:

[...]

PR #1214, opened 2019-08-22
Remove optional automatic setting of local values in JS SDK

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/1086
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/683

#### Changes

- Remove generated settings from the js sdk and move those to the console
- Extend/fix prop-types of `<DeviceDataForm />`, `<DeviceAdd />` and `<DeviceGeneralSettings />` components
- Add `device` and `env` prop-types
- Extend `yup` methods with `emptyOrLength`
- Fix `mac_settings` entry in the validation schema (it was `mac_state` before)
- Add missing entries to the validation schema
- Merge device update and add validation schemas


pkg/webui/lib/prop-types.js
```diff
@@ -61,4 +61,42 @@ PropTypes.event = PropTypes.shape({
   data: PropTypes.object,
 })
 
+PropTypes.env = PropTypes.shape({
```
kschiffer: Precision can be improved
```js
PropTypes.env = PropTypes.shape({
  appRoot: PropTypes.string.isRequired,
  assetsRoot: PropTypes.string.isRequired,
  siteName: PropTypes.string.isRequired,
  siteTitle: PropTypes.string.isRequired,
  siteSubTitle: PropTypes.string,
  pageData: PropTypes.shape({}),
  config: PropTypes.shape({
    language: PropTypes.string,
    is: PropTypes.shape({
      enabled: PropTypes.bool.isRequired,
      base_url: PropTypes.string.isRequired,
    }),
    as: PropTypes.shape({
      enabled: PropTypes.bool.isRequired,
      base_url: PropTypes.string.isRequired,
    }),
    ns: PropTypes.shape({
      enabled: PropTypes.bool.isRequired,
      base_url: PropTypes.string.isRequired,
    }),
    js: PropTypes.shape({
      enabled: PropTypes.bool.isRequired,
      base_url: PropTypes.string.isRequired,
    }),
    gs: PropTypes.shape({
      enabled: PropTypes.bool.isRequired,
      base_url: PropTypes.string.isRequired,
    }),
  }).isRequired,
})
```

sdk/js/src/service/devices/index.js
```diff
@@ -17,7 +17,6 @@
 import traverse from 'traverse'
 import Marshaler from '../../util/marshaler'
 import Device from '../../entity/device'
-import randomByteString from '../../util/random-bytes'
```
kschiffer: Should we remove the file as well then?

sdk/js/src/service/devices/index.js
```diff
@@ -275,74 +274,11 @@ class Devices {
...
-        },
-      }
-
       dev.supports_join = false
     } else {
       if ('provisioner_id' in dev && dev.provisioner_id !== '') {
         throw new Error('Setting a provisioner with end device keys is not allowed.')
       }
-      let root_keys = {}
-      if (withRootKeys) {
```
kschiffer: If we remove this, remove the argument as well.

sdk/js/src/service/devices/index.js
```diff
@@ -275,74 +274,11 @@ class Devices {
     const dev = device
 
     if (abp) {
-      const session = {
```
kschiffer: Removing this chunk will make the SDK behavior deviate from the one of the [CLI implementation](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/cmd/ttn-lw-cli/commands/end_devices.go#L332-L405).
The discussion in https://github.com/TheThingsNetwork/lorawan-stack/issues/654 did not make it entirely clear to me whether we want to keep this functionality in the SDK or not.

@htdvisser, can you have your say?

I just want to make sure we're doing the correct thing with removing this.

sdk/js/src/service/devices/index.js
bafonins: Screenshot 2019-08-28 at 14 18 12

pkg/webui/console/components/device-data-form/validation-schema.js
```diff
@@ -57,101 +41,116 @@ const baseSchemaShape = Yup.object({
...
     }),
-  }),
-  root_keys: Yup.object().shape({
-    nwk_key: Yup.object().shape({
-      key: Yup.string().length(16 * 2, m.validate32), // 16 Byte hex
+    root_keys: Yup.object().when('activation_mode', {
+      is: isOTAA,
+      then: schema =>
+        schema.shape({
+          nwk_key: Yup.object().shape({
```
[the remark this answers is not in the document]
bafonins: This is something to improve in the future. We can address it when doing in https://github.com/TheThingsNetwork/lorawan-stack/issues/579

sdk/js/src/service/devices/index.js
bafonins: The whole idea of this PR is to move generation to the client, in this case, the console.
The only difference I see that the console is not setting `root_key_id`. 
@htdvisser Should we append it to the payload? Something like `ttn-lw-console-generated`should suffice.

kschiffer: If you see the SDK as a standalone entity (and not only in conjunction with the console), I think it can make sense to consider matching its behavior with the CLI. 
I just want to be clear about whether we should think this way or not; both is fine with me.

htdvisser: The CLI is not an SDK, it's a client, just like the Console.

The purpose of any SDK is to provide a (ideally thin) language-specific interface on top of our HTTP and/or gRPC API. Some common business logic (like the decision to talk to the JS or not when creating an OTAA/ABP device) can be handled by SDKs. SDKs may also provide helpers for the client (such as generating random keys), but that indeed doesn't belong here.

bafonins: @htdvisser What about
> Should we append it to the payload? Something like ttn-lw-console-generatedshould suffice.

Issue #1166, opened 2019-08-14
Fix form submit value patch generation


kschiffer:
#### Summary
The generation of update patches is currently not working as expected. This causes errors e.g. in the device general settings form in some situations.

#### Steps to Reproduce
1. Create a device
2. Go to the general settings panel
3. Update the name of the device and submit
4. See the 'successful' message
5. Update the name of the device again and submit
6. See an error

#### What do you see now?
Faulty update patch values, causing errors in the device general settings form submit

#### What do you want to see instead?
Correct update patch values that cause no errors in the backend

#### Environment
Console.

#### How do you propose to implement this?
The issue originates from our `pkg/webui/lib/diff.js` not working reliably, especially for nested object values, but there are also other arbitrary issues. Faulty patches will then lead to errors in the backend, as e.g. invalid field masks for update operations are used. 
See the example:
<details><summary>Patch issue example</summary>

**Original**
```json
{  
  "ids":{  
    "device_id":"test-devdf",
    "application_ids":{  
      "application_id":"test-app"
    },
    "dev_eui":"ADF2D2EDEF23D2DA",
    "join_eui":"ADD2D2ED2D2D2D22"
  },
  "created_at":"2019-08-14T08:50:22.174Z",
  "updated_at":"2019-08-14T09:04:15.285Z",
  "frequency_plan_id":"EU_863_870",
  "supports_join":true,
  "lorawan_version":"1.0.0",
  "lorawan_phy_version":"1.0.0",
  "root_keys":{  
    "root_key_id":"ttn-lw-js-sdk-generated",
    "app_key":{  
      "key":"F45F7BEFA1DC9003A155E5D3997CA3EB"
    },
    "nwk_key":{  
      "key":"B5E34F75ABBDDA3857BF414EAE9EA683"
    }
  },
  "network_server_address":"localhost",
  "application_server_address":"localhost",
  "name":"aasd",
  "join_server_address":"localhost",
  "supports_class_c":false,
  "mac_state":{  

  },
  "resets_join_nonces":false
}
```
**Update**
```json
{  
  "ids":{  
    "device_id":"test-devdf",
    "application_ids":{  
      "application_id":"test-app"
    },
    "dev_eui":"ADF2D2EDEF23D2DA",
    "join_eui":"ADD2D2ED2D2D2D22"
  },
  "created_at":"2019-08-14T08:50:22.174Z",
  "updated_at":"2019-08-14T09:04:15.285Z",
[...]

PR #1227, opened 2019-08-22
Fix form submit value patch generation

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1166

#### Changes

- Remove `ids.dev_addr` from allowed field masks fort `set` requests
- Refactor `diff` to do deep object comparison 
- Add check for `boolean` values when merging devices from different sources
- Fix `mac_settings` in the device validation schema

#### Notes for Reviewers

The main issue in the device form was sending `ids.dev_addr` when updating device since the NS does not allow it and expects `dev_addr` to be in the `session` entry.

@rvolosatovs please check NS related changes

#### Release Notes

- Removed `ids.dev_addr` from allowed field masks for `/ttn.lorawan.v3.NsEndDeviceRegistry/Set`

### DIFF
--- a/pkg/ttnpb/field_mask_validation.go
+++ b/pkg/ttnpb/field_mask_validation.go
@@ -497,7 +497,6 @@ var AllowedFieldMaskPathsForRPC = map[string][]string{
 		"ids",
 		"ids.application_ids",
 		"ids.application_ids.application_id",
-		"ids.dev_addr",
 		"ids.dev_eui",
 		"ids.device_id",
 		"ids.join_eui",
--- a/pkg/webui/lib/diff.js
+++ b/pkg/webui/lib/diff.js
@@ -12,17 +12,27 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-const shallowDiff = function(initial, changed, ignorePaths = []) {
-  return Object.keys(changed).reduce(function(patch, field) {
-    const oldValue = initial[field]
-    const newValue = changed[field]
+import { observableDiff, applyChange } from 'deep-diff'
 
-    if (!ignorePaths.includes(field) && oldValue !== newValue) {
-      patch[field] = newValue
+/**
+ * Computes the structural differences (deep) between `original` and `updated` objects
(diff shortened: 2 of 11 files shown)


pkg/ttnpb/field_mask_validation.go
```diff
@@ -497,7 +497,6 @@ var AllowedFieldMaskPathsForRPC = map[string][]string{
 		"ids",
 		"ids.application_ids",
 		"ids.application_ids.application_id",
-		"ids.dev_addr",
```
rvolosatovs: @htdvisser  can you confirm that this is not passed by CLI anymore? If so, I'm good with it.
FTR, AS does not allow it to be set either.

htdvisser: The CLI uses `AllowedFieldMaskPathsForRPC` to determine what fields to set/get, so that should be fine.

johanstokking: Isn't this for setting ABP device addresses?

bafonins: > Isn't this for setting ABP device addresses?

The console sets it in `session.dev_addr`

johanstokking: Right

pkg/webui/lib/diff.js
```diff
@@ -12,17 +12,26 @@
...
-    if (!ignorePaths.includes(field) && oldValue !== newValue) {
-      patch[field] = newValue
+/**
+ * Computes the structural differences (deep) between `original` and `updated` objects
+ * by applying any differences (add/remove/update).
+ * @param {Object} original - The original object.
+ * @param {Object} updated - The updated version of the `original` object.
+ * @param {Array} exclude - A list of field names that should not be included in the
+ * final diff.
+ * @returns {Object} - A new object represing the structural differences between `original` and `updated`.
```
kschiffer:
```suggestion
 * @returns {Object} - A new object representing the structural differences between `original` and `updated`.
```

PR #1229, opened 2019-08-22
Add authorizing target applications for end device claiming

johanstokking:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/issues/1673

#### Changes

- Add rpcs for authorizing and unauthorizing apps


api/application.proto
```diff
@@ -43,6 +43,11 @@ message Applications {
   repeated Application applications = 1;
 }
 
+message ApplicationAPIKey {
```
htdvisser: Unused?

PR #1233, opened 2019-08-23
Rewrite Getting Started for online deployment

htdvisser:
#### Summary

In this Pull Request we change the Getting Started documentation in such a way that it produces an online deployment (on `your.thethingsstack.io`) instead of an offline one (on `localhost`). Other documentation is also updated accordingly.

#### Changes

- Rewrote Installation, Certificates, Configuration and Running The Things Stack
- Changed `localhost` to `your.thethingsstack.io` in other pages.

#### Notes for Reviewers

- Is `your.thethingsstack.io` an appropriate example host?
- I tested the entire walkthrough on https://your.thethingsstack.io
- I think we should address other (even recommended) config options (such as branding, alternative blob config, user registration settings) in the **Reference** section, and not in the Getting Started guide. Therefore I also think this is out of scope for this pull request.
- Let's try to have this in the next release (otherwise we need to wait with updating gh-pages until the next release)

#### Release Notes

- Improved getting started guide for a deployment of The Things Stack

### DIFF
--- a/doc/content/guides/getting-started/configuration.md
+++ b/doc/content/guides/getting-started/configuration.md
@@ -4,7 +4,7 @@ description: ""
 weight: 3
 ---
 
-The Things Stack can be started for development without passing any configuration. However, there are a number of things you need to configure for a production deployment. In this guide we'll set environment variables in the `docker-compose.yml` file to configure The Things Stack as a private deployment on `your.thethingsstack.io`.
+The Things Stack can be started for development without passing any configuration. However, there are a number of things you need to configure for a production deployment. In this guide we'll set environment variables in the `docker-compose.yml` file to configure The Things Stack as a private deployment on `thethings.example.com`.
 
 ## Databases
 
@@ -79,14 +79,14 @@ TTN_LW_IS_DATABASE_URI=postgres://root@cockroach:26257/ttn_lorawan?sslmode=disab
 TTN_LW_REDIS_ADDRESS=redis:6379
 ```
 
-Then we'll configure TLS with Let's Encrypt. Since we're deploying The Things Stack on `your.thethingsstack.io`, we configure it to only request certificates for that host, and to also use it as the default host.
+Then we'll configure TLS with Let's Encrypt. Since we're deploying The Things Stack on `thethings.example.com`, we configure it to only request certificates for that host, and to also use it as the default host.
 
 ```
 TTN_LW_TLS_ACME_ENABLE=true
 TTN_LW_TLS_ACME_DIR=/var/lib/acme
 TTN_LW_TLS_ACME_EMAIL=your@email.com
-TTN_LW_TLS_ACME_HOSTS=your.thethingsstack.io
-TTN_LW_TLS_ACME_DEFAULT_HOST=your.thethingsstack.io
+TTN_LW_TLS_ACME_HOSTS=thethings.example.com
+TTN_LW_TLS_ACME_DEFAULT_HOST=thethings.example.com
 ```
 
 Next, we'll configure the HTTP server with keys for encrypting and verifying cookies, and with passwords for endpoints that you may want to keep for internal use.
@@ -102,9 +102,9 @@ The Things Stack sends emails to users, so we need to configure how those are se
 
 ```
 TTN_LW_IS_EMAIL_SENDER_NAME=The Things Stack
-TTN_LW_IS_EMAIL_SENDER_ADDRESS=noreply@your.thethingsstack.io
-TTN_LW_IS_EMAIL_NETWORK_CONSOLE_URL=https://your.thethingsstack.io/console
-TTN_LW_IS_EMAIL_NETWORK_IDENTITY_SERVER_URL=https://your.thethingsstack.io/oauth
+TTN_LW_IS_EMAIL_SENDER_ADDRESS=noreply@thethings.example.com
+TTN_LW_IS_EMAIL_NETWORK_CONSOLE_URL=https://thethings.example.com/console
+TTN_LW_IS_EMAIL_NETWORK_IDENTITY_SERVER_URL=https://thethings.example.com/oauth
 ```
 
 You can either use Sendgrid or an SMTP server. If you don't set up an email provider, The Things Stack will print emails to the server log.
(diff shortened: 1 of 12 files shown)


johanstokking:
Great

> * Is `your.thethingsstack.io` an appropriate example host?

Hmm, then I think I would prefer `my.thethingsstack.io`, but maybe it's better to come with an explicit example domain, like `example.org`

doc/content/guides/getting-started/configuration.md
```diff
@@ -4,29 +4,158 @@ description: ""
...
+
+TTN_LW_CONSOLE_UI_CANONICAL_URL=https://your.thethingsstack.io/console
+TTN_LW_CONSOLE_UI_AS_BASE_URL=https://your.thethingsstack.io/api/v3
+TTN_LW_CONSOLE_UI_GS_BASE_URL=https://your.thethingsstack.io/api/v3
+TTN_LW_CONSOLE_UI_IS_BASE_URL=https://your.thethingsstack.io/api/v3
+TTN_LW_CONSOLE_UI_JS_BASE_URL=https://your.thethingsstack.io/api/v3
+TTN_LW_CONSOLE_UI_NS_BASE_URL=https://your.thethingsstack.io/api/v3
+
+TTN_LW_CONSOLE_OAUTH_CLIENT_ID=console
+TTN_LW_CONSOLE_OAUTH_CLIENT_SECRET=generate a good one (openssl rand -hex 32)
```
johanstokking: Indicate that this is hex and the length?

I can implicitly deduce that from the `openssl` command, but maybe something like `enter generated 32-bit hex secret ($ openssl rand -hex 32)`

doc/content/guides/getting-started/mqtt/_index.md
```diff
@@ -23,7 +23,7 @@ There are many MQTT clients available. Great clients are `mosquitto_pub` and `mo
 >Tip: when using `mosquitto_sub`, pass the `-d` flag to see the topics messages get published on. For example:
 >
 >```bash
-$ mosquitto_sub -h localhost -t '#' -u app1 -P 'NNSXS.VEEBURF3KR77ZR..' -d
+$ mosquitto_sub -h your.thethingsstack.io -t '#' -u app1 -P 'NNSXS.VEEBURF3KR77ZR..' -d
```
johanstokking: Not related to this diff; but maybe create the API key in the beginning of this document with the Console instead of CLI?

doc/content/guides/getting-started/webhooks/_index.md
```diff
@@ -55,7 +55,7 @@ The path are:
 For example:
 
 ```
-$ curl http://localhost:1885/api/v3/as/applications/app1/webhooks/wh1/devices/dev1/down/push \
+$ curl https://your.thethingsstack.io/api/v3/as/applications/app1/webhooks/wh1/devices/dev1/down/push \
   -X POST \
   -H 'Authorization: Bearer NNSXS.VEEBURF3KR77ZR..' \
```
johanstokking: Same for generating this API key

doc/content/guides/getting-started/configuration.md
```diff
@@ -4,29 +4,158 @@ description: ""
 weight: 3
 ---
 
-The Things Stack can be started for development without passing any configuration. However, there are a lot of things you can configure. In this guide we'll set some environment variables in `docker-compose.yml`. These environment variables will configure The Things Stack as a development server on `localhost`. For setting up up a public server or for requesting certificates from an ACME provider such as Let's Encrypt, take a closer look at the comments in `docker-compose.yml`.
+The Things Stack can be started for development without passing any configuration. However, there are a number of things you need to configure for a production deployment. In this guide we'll set environment variables in the `docker-compose.yml` file to configure The Things Stack as a private deployment on `your.thethingsstack.io`.
```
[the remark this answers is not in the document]
htdvisser: The `docker-compose.yml` in the repository can be used as a base, then it's just a matter of (un)commenting and editing lines. I still think it's good to quickly go through everything in the file.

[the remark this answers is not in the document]
htdvisser: I think it would be nice to eventually have a wizard (`ttn-lw-cli quickstart`) that would try to detect the environment (Docker, ...), ask some questions, write everything to a config file, initialize the database, set up the user and register the OAuth clients. That would be something for later though.

doc/content/guides/getting-started/configuration.md
htdvisser: Not hex, and doesn't matter. I'll reword.

doc/content/guides/getting-started/mqtt/_index.md
htdvisser: Indeed a bit out of scope for this PR, and then still I think it's fine to assume CLI when they're already using the terminal.

doc/content/guides/getting-started/configuration.md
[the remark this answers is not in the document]
johanstokking: > The `docker-compose.yml` in the repository can be used as a base, then it's just a matter of (un)commenting and editing lines. I still think it's good to quickly go through everything in the file.

Yes that's what I thought. What's missing then is the link to that file.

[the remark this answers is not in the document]
htdvisser: We can't really link to that file, since it may be release-specific. The documentation is bundled with the release archive, so if the reader downloaded an archive, they already have it. If they're looking at the online documentation, then we need to link to a specific version, which would require a shortcode to fill that version.

Of course we can pull the yml into the `doc` folder on build, or add a shortcode to fill a (perma)link to a specific version of the docker-compose file, but I'm not sure if we want to do all that in this PR.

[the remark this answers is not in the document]
johanstokking: I understand the technical reason why we cannot simply put a link here now, but this is a Getting Started and is supposed to make things easy. We'll be pointing directly from The Things Stack landing page to this Getting Started, so nobody downloaded a release yet. 

My concern is mostly with what's in **Installation** and where we build upon here, when configuring;

>Most releases contain an example `docker-compose.yml` file. You can also find this file in the Github repository of The Things Stack. In this guide we’ll use that example `docker-compose.yml` for our deployment.

There's not even a link to the GitHub repository here. Note that, for a lot of people, GitHub is quite intimidating, so figuring out that there is a `TheThingsNetwork` account, that there's a `lorawan-stack` repository in there, and then downloading the raw file of `docker-compose.yml`, is at best good natural selection for skilled people installing the stack, and at worst a way for people to get lost even before they started.

I would, in fact, suggest copying in `docker-compose.yml` with the exact purpose of a template for this Getting Started. We may even refer to documentation steps, which we cannot really do in the `docker-compose.yml` we keep in the root.

doc/content/guides/getting-started/mqtt/_index.md
johanstokking: They are using the terminal for Mosquitto, not (necessarily) for The Things Stack. I'm fine keeping it here, but then at least add the CLI only shortcode on top of this page

doc/content/guides/getting-started/configuration.md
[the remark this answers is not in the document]
htdvisser: Added direct link to the file in 3d5c0f42987a6c532f9c2a8f4959694cb6abdf03

doc/content/guides/getting-started/mqtt/_index.md
htdvisser: Then we should also add that shortcode to other pages that only exaplain how to do things with the CLI and cURL (webhooks, multicast, events), right? I really hoped that we could keep this PR properly scoped 😕

Issue #1226, opened 2019-08-22
client 'console' not found error when logging in


swattor:
#### Summary

When logging into the web console, the following error is produced:

Not Found
client `console` not found

The "take me back" button returns you to the /oauth URL indicating that the login successfully completed

...

#### Steps to Reproduce

1. Spin up stack
2. Configure
3. Attempt to login 

[ttn_config.txt](https://github.com/TheThingsNetwork/lorawan-stack/files/3530616/ttn_config.txt)

#### What do you see now?

...

#### What do you want to see instead?

...

#### Environment

...

#### How do you propose to implement this?

...

#### Can you do this yourself and submit a Pull Request?

...

johanstokking:
> 1. Spin up stack
> 2. Configure

Can you elaborate a bit on the second point?

Did you follow the Getting Started, especially the part where you need to [create an OAuth client for the Console](https://thethingsnetwork.github.io/lorawan-stack/v3.1.0/guides/getting-started/running-the-stack/)?

swattor:
Ah. 

See, the documentation is a bit confusing:

https://github.com/TheThingsNetwork/lorawan-stack/blob/master/doc/gettingstarted.md links to:
https://ttn.fyi/v3/getting-started which links to:
https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.0.4/doc/gettingstarted.md

Which doesn't have the bit about the console.

I've now run the create OAuth client for the console, and now I get 'Internal Server Error'!

swattor:
It's worth noting that I'm running this on a seperate machine (so accessing via http://remote-ip/ rather than http://localhost). I'm not sure whether that is relevant.

swattor:
Tried various things, including starting again. While following the instructions gets me further, I'm now stuck with an "Internal Server Error" when I login.

htdvisser:
Updated the redirect [ttn.fyi/v3/getting-started](https://ttn.fyi/v3/getting-started) to point to the latest getting started guide.

htdvisser:
@swattor Can you please expand the **Steps to Reproduce** section with the **exact** steps you took? Without detailed steps we can't reproduce the issue. The issue template also asks you to upload your configuration, terminal output / logs and screenshots (you can redact sensitive information). Please do that so that we can investigate this further.

htdvisser:
[...]

PR #1236, opened 2019-08-23
Make OAuth errors explicit

### LINKED ISSUE #1232: Stack returns Error 500 on authentication
#### Summary

I am setting up a private network using the TTN V3 Stack on a k8s local server.
The whole setup process goes fine, until the CLI authentication.
After entering login/pwd on the web UI i get a 500 Error page.
Attached, you will find the logs with debug level activated.

...

#### Steps to Reproduce

 1 - Get a minikube or a full k8s cluster
2. - Use the attached kubernetes files to setup the infrastructure 
[ttn-k8s.tar.gz](https://github.com/TheThingsNetwork/lorawan-stack/files/3534242/ttn-k8s.tar.gz)
3. - log with the CLI through the web, enter the credentials 

logs with debug activated doesn't really help but here they are.
[logs-from-stack-in-stack-dc4c979b5-4zhzq.txt](https://github.com/TheThingsNetwork/lorawan-stack/files/3534202/logs-from-stack-in-stack-dc4c979b5-4zhzq.txt)

Is there any way to get a more verbose logging ?
...

I want to get my oAuth Token, i get a 500 error
...

#### Environment

OS : Ubuntu 16
K8s : V1.15.2
...

...

...

htdvisser:
#### Summary

This pull request makes OAuth errors explicit by turning error IDs returned by the OAuth library into defined errors.

Closes #1232 and perhaps also #1226.

#### Changes

- Return defined errors from `/oauth/authorize` if possible
- Inspect errors from OAuth library and convert to defined errors

#### Release Notes

- Improved error handling in OAuth flow

### DIFF
--- a/pkg/oauth/oauth.go
+++ b/pkg/oauth/oauth.go
@@ -80,16 +80,28 @@ func (s *server) Authorize(authorizePage echo.HandlerFunc) echo.HandlerFunc {
 		ar.UserData = userData{UserIdentifiers: session.UserIdentifiers}
 		client := ttnpb.Client(ar.Client.(osinClient))
 		if !clientHasGrant(&client, ttnpb.GRANT_AUTHORIZATION_CODE) {
-			return errClientMissingGrant.WithAttributes("grant", "authorization_code")
+			resp.InternalError = errClientMissingGrant.WithAttributes("grant", "authorization_code")
+			resp.SetError(osin.E_INVALID_GRANT, resp.InternalError.Error())
+			oauth2.FinishAuthorizeRequest(resp, req, ar)
+			return s.output(c, resp)
 		}
 		switch client.State {
 		case ttnpb.STATE_REJECTED:
-			return errClientRejected
+			resp.InternalError = errClientRejected
+			resp.SetError(osin.E_INVALID_CLIENT, resp.InternalError.Error())
+			oauth2.FinishAuthorizeRequest(resp, req, ar)
+			return s.output(c, resp)
 		case ttnpb.STATE_SUSPENDED:
-			return errClientSuspended
+			resp.InternalError = errClientSuspended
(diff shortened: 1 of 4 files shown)


pkg/oauth/oauth.go
```diff
@@ -73,24 +80,16 @@ func (s *server) Authorize(authorizePage echo.HandlerFunc) echo.HandlerFunc {
 		ar.UserData = userData{UserIdentifiers: session.UserIdentifiers}
 		client := ttnpb.Client(ar.Client.(osinClient))
 		if !clientHasGrant(&client, ttnpb.GRANT_AUTHORIZATION_CODE) {
-			resp.SetError(osin.E_INVALID_CLIENT, "OAuth client does not have authorization code grant")
-			oauth2.FinishAuthorizeRequest(resp, req, ar)
-			return s.output(c, resp)
+			return errClientMissingGrant.WithAttributes("grant", "authorization_code")
```
johanstokking: How is `E_INVALID_CLIENT` propagated to the caller?

htdvisser: Previously these errors just resulted in a 500 error (the issue). Now we return an explicit error in the same way as all other errors. The error is rendered into the HTML page and then picked up by the webui code.

johanstokking: OK I see, but with this explicit call to `SetError()` with a code I was under the impression it's some application layer transport of error codes, i.e. as part of OAuth and not just HTTP error codes

htdvisser: Not for these internal errors. Some errors are indeed handled by the OAuth protocol (with an error param in the redirect), but if the OAuth client is invalid, we shouldn't redirect anyway.

htdvisser: Ok, it turns out that the OAuth spec in fact allows redirecting on these errors, I'll revert the deletes.

Issue #1225, opened 2019-08-22
Gateway activity update and wrong time rendering


johanstokking:
#### Summary

Gateway activity update is not updated and last active time rendering is wrong.

#### Steps to Reproduce

1. Go to one of the connected gateways in the demo environment, i.e. https://demo.thethings.industries/console/gateways/647fdafffe000162
2. See last activity time

#### What do you see now?

These issues are probably separate:

1. The time is shown as `-X`, where `X` is likely seconds. This seems regression; it worked before
2. Sometimes `X` stalls
3. Sometimes, when a gateway connects, events are shown correctly, but it keeps shown as **Disconnected**

<img width="1128" alt="Screenshot 2019-08-22 at 16 01 06" src="https://user-images.githubusercontent.com/13334001/63522323-6364e180-c4f8-11e9-8111-36301c309660.png">

#### What do you want to see instead?

1. The old behavior; i.e. `X seconds ago` 
2. Always keep incrementing `X`, or switch correctly to **Disconnected** state
3. Keep the **Connected** and **Disconnected** states updated

#### Environment

Latest `master` deployed currently at https://demo.thethings.industries

#### How do you propose to implement this?

Not sure

#### Can you do this yourself and submit a Pull Request?

Can review

bafonins:
The issue seems to appear only when using Safari.

KrishnaIyer:
@bafonins: I noticed this yesterday on Chrome as well.

bafonins:
> Sometimes, when a gateway connects, events are shown correctly, but it keeps shown as Disconnected

moved to https://github.com/TheThingsNetwork/lorawan-stack/issues/1242

PR #1237, opened 2019-08-23
Fix RelativeTimeFormat display and update

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/1225
References https://github.com/TheThingsNetwork/lorawan-stack/issues/1086

#### Changes

- Add polyfills for `Intl.PluralRules` and `Intl.RelativeTimeFormat`
- Fix relative time updates

#### Notes for Reviewers

All problems solved in this PR are caused by the new api of `react-intl v3`.
1. Relative time isnt always updating even on new values passed to `<RelativeTime />`. The reason for this is [relative.js L34](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/webui/lib/components/date-time/relative.js#L34) returning the same `delta` and hence not updating `<FormattedRelativeTime />`. This is now fixed by adding the `key` property that will force `<FormattedRelativeTime />` to update on new `dateTime`.
2. Broken relative time display in safari (see [the issue](https://github.com/TheThingsNetwork/lorawan-stack/issues/1225)). `react-intl v3` no longer comes with polyfills included and relies on native `Intl` API instead. According to https://caniuse.com/#search=Intl Safari does not support native localization (including IE11 and Edge).

@johanstokking Please verify that this solves the issue with displaying relative time.

#### Release Notes

- Fixed relative time display in the Console for  IE11, Edge and Safari

### DIFF
--- a/pkg/webui/lib/components/with-locale.js
+++ b/pkg/webui/lib/components/with-locale.js
@@ -20,10 +20,13 @@ import { connect } from 'react-redux'
 import { IntlProvider } from 'react-intl'
 import CancelablePromise from 'cancelable-promise'
 
+import Spinner from '../../components/spinner'
+
 import log, { error } from '../log'
 import { withEnv } from './env'
 
-const defaultLanguage = 'en'
+const defaultLocale = process.predefined.DEFAULT_MESSAGES_LOCALE // Note: defined by webpack define plugin
+const defaultLanguage = defaultLocale.split('-')[0] || 'en'
 const xx = 'xx'
 const dev = '../../dev'
 
@@ -47,6 +50,7 @@ export default class UserLocale extends React.PureComponent {
   state = {
     messages: process.predefined.DEFAULT_MESSAGES, // Note: defined by webpack define plugin
     xx: false,
+    loaded: false,
   }
 
   toggle() {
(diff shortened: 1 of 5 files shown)


bafonins:
Found issues with loading translated messages, will reopen once fixed and ready for review

pkg/webui/lib/components/with-locale.js
```diff
@@ -67,23 +71,29 @@ export default class UserLocale extends React.PureComponent {
   }
 
   check(prev, props) {
-    const current = (prev.user && prev.user.language) || prev.env.config.language || defaultLanguage
+    const current = (prev.user && prev.user.language) || prev.env.config.language
```
kschiffer: Since there is no way to change the locale atm, I think the comparison of previous and current props is redundant. Maybe we should just go with `prev.env.config.language` here for now, which is always present.

pkg/webui/lib/components/with-locale.js
```diff
@@ -92,7 +102,7 @@ export default class UserLocale extends React.PureComponent {
     }
   }
 
-  load(language) {
+  async load(language) {
     let locale = navigator.language || navigator.browserLanguage || 'en-US'
```
kschiffer:
```suggestion
    let locale = navigator.language || navigator.browserLanguage || defaultLanguage
```

bafonins: I think `defaultLocale` is more appropriate here

pkg/webui/lib/components/with-locale.js
bafonins: > Since there is no way to change the locale atm

But there will be an option later. Lets keep changes to a minimum.

PR #1238, opened 2019-08-23
Various interop-related fixes and improvements

rvolosatovs:
#### Summary

Various small fixes

#### Changes

- Add `fetch.WithBasePath` and use in `interop`
- Improve interop-related logging
- Fix `/` being added twice to configured path
- Set context in interop HTTP requests
- Set decoded payload in join-request(it's not used by cluster-local JS, but is used by the `interop` client)


rvolosatovs:
Test failure is due NS tests not accounting for https://github.com/TheThingsNetwork/lorawan-stack/pull/1238/commits/7ba094428fbc71d103c6fa01b321ae45d1e1eaa7
It's a very minor issue - I'll fix it on Monday

PR #1239, opened 2019-08-23
Broaden the IS password character limits, allow Unicode

### LINKED ISSUE #1223: "Special" password character set is non-extensive and opaque for the user
#### Summary

"Special" password character set is non-extensive and, worst of all, opaque for the user.

https://github.com/TheThingsNetwork/lorawan-stack/blob/7b6050dba6ca1bb546f90570c23402cb059ca0ba/pkg/identityserver/user_registry.go#L84

#### Steps to Reproduce

1. Go to any v3 deployment
2. Attempt to create an account with password 
```
mtMp`2mhbtytvsN?)`Zf(W/]
```

#### What do you see now?

`Should contain at least one special character` error and a major confusion

Successful creation of account or at least error message stating which characters stack expects to be present in the password as "special".

#### Environment

Any

Either the "special" password character set should be extended, or the "special" character set should be defined clearly, such that users know, which characters we consider "special".

yes

rvolosatovs:
#### Summary

Closes #1223 

#### Changes

- Broaden IS password character limits
- Support Unicode

#### Notes for Reviewers

It's a very small change, but I assigned both @johanstokking and @htdvisser to take look, since it's quite important that we agree on this.
Using unicode characters should not be a problem - it's 2019 already. This also gives a lot of freedom to our users, which I'm sure (at least) some will be happy about.

### DIFF
--- a/pkg/identityserver/user_registry.go
+++ b/pkg/identityserver/user_registry.go
@@ -20,6 +20,7 @@ import (
 	"runtime/trace"
 	"strings"
 	"time"
+	"unicode"
 
 	"github.com/gogo/protobuf/types"
 	"github.com/jinzhu/gorm"
@@ -77,11 +78,11 @@ func (is *IdentityServer) validatePasswordStrength(ctx context.Context, password
 	var uppercase, digits, special int
 	for _, r := range password {
 		switch {
-		case r >= 'A' && r <= 'Z':
+		case unicode.IsUpper(r):
 			uppercase++
-		case r >= '0' && r <= '9':
+		case unicode.IsDigit(r):
 			digits++
-		case r == '!' || r == '%' || r == '@' || r == '#' || r == '$' || r == '&' || r == '*':
+		case !unicode.IsLetter(r) && !unicode.IsNumber(r):
 			special++
 		}
 	}
--- a/pkg/identityserver/user_registry_test.go
+++ b/pkg/identityserver/user_registry_test.go
@@ -30,33 +30,41 @@ import (
 )
 
 func TestValidatePasswordStrength(t *testing.T) {
-	a := assertions.New(t)
+	for p, ok := range map[string]bool{
+		"āA0$": false, // Too short
+		strings.Repeat("āaaAAA➉23!@#aaaAAA12aaaAAA123!@#aaaAAA12aaaAAA123!@#aaaAAA12aaaAAA123!@#aaaAAA12aaaAAA123!@#aaaAAA12aaa", 10): false, // Too long.
+		"āaabbb➉23":    false, // No uppercase and special characters.
+		"āaabbb➉AA":    false, // No digits and special characters.
+		"āaabbb➉@#":    false, // No digits and uppercase characters.
+		"āaa123➉@#":    false, // No uppercase characters.
+		"āaaAAA➉@#":    false, // No digits.
+		"āaaAAA➉23":    false, // No special characters.
+		"āaaAAA123!@#": true,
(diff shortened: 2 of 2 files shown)


pkg/identityserver/user_registry.go
```diff
@@ -77,11 +78,11 @@ func (is *IdentityServer) validatePasswordStrength(ctx context.Context, password
...
 	for _, r := range password {
 		switch {
-		case r >= 'A' && r <= 'Z':
+		case unicode.IsUpper(r):
 			uppercase++
-		case r >= '0' && r <= '9':
+		case unicode.IsDigit(r):
 			digits++
-		case r == '!' || r == '%' || r == '@' || r == '#' || r == '$' || r == '&' || r == '*':
+		case !unicode.IsLetter(r) && !unicode.IsNumber(r):
```
rvolosatovs: This includes category `Z`, separators(e.g. ` `). IMO it makes sense to count space etc. as special, by looking at https://github.com/TheThingsNetwork/lorawan-stack/issues/1223#issuecomment-524226731 seems that @htdvisser as well.

rvolosatovs: Another possiblity is to be more strict and count `S`, `P`, `M`, `Z` here, for example, but IMO the current solution is best.

pkg/identityserver/user_registry.go
```diff
@@ -77,11 +78,11 @@ func (is *IdentityServer) validatePasswordStrength(ctx context.Context, password
 	var uppercase, digits, special int
 	for _, r := range password {
 		switch {
-		case r >= 'A' && r <= 'Z':
+		case unicode.IsUpper(r):
```
htdvisser: This is quite a significant change to our current definition of uppercase letters. Now it would also include letters like `Д` and `𝒬`, which IMO should be considered "special" in passwords.

pkg/identityserver/user_registry.go
```diff
@@ -77,11 +78,11 @@ func (is *IdentityServer) validatePasswordStrength(ctx context.Context, password
...
 			uppercase++
-		case r >= '0' && r <= '9':
+		case unicode.IsDigit(r):
```
htdvisser: Same here, I'm not sure if we should consider `٦` a "digit" or a "special"

pkg/identityserver/user_registry.go
htdvisser: We can also just catch all "other" `IsPrint` here.

johanstokking:
Yeah I approved this because it's ultimately the responsibility of the user to come up with a strong password. Administrators set these rules to enforce making passwords hard to guess, and imo if people use their local language chars the same rules apply and things do not become easier to guess.

For simplicity, I would agree with @htdvisser proposal except for explicitly disallowing non-printables; https://github.com/TheThingsNetwork/lorawan-stack/issues/1223#issuecomment-524315248. But it's essentially the same as this, except that we broaden applying our rules to non-latin chars.

pkg/identityserver/user_registry.go
rvolosatovs: Why so? Why would we not allow passwords in local languages?
I don't think we should force people to use latin alphabet in passwords, hence I think `Д`, `Ы`, `Ё` etc. are all perfectly valid uppercase characters.

pkg/identityserver/user_registry.go
rvolosatovs: If it's "digit enough" for Unicode, it should be "digit enough" for us

Issue #1243, opened 2019-08-26
Update password validation in oauth app


bafonins:
#### Summary

After [this PR](https://github.com/TheThingsNetwork/lorawan-stack/pull/1239) password validation in the oauth app does not match the validation rules in the IS.
Related https://github.com/TheThingsNetwork/lorawan-stack/issues/1067

#### Steps to Reproduce

1. Go to the Create Account page in the oauth app
2. Use `āaaAAA123` as a password
3. See client-side validation error

#### What do you see now?

<img width="399" alt="Screenshot 2019-08-26 at 12 28 06" src="https://user-images.githubusercontent.com/16374166/63684391-16855180-c7fd-11e9-8990-116b3b5c561a.png">
<img width="476" alt="Screenshot 2019-08-26 at 12 28 22" src="https://user-images.githubusercontent.com/16374166/63684392-171de800-c7fd-11e9-9ffd-c20c08e8c525.png">

#### What do you want to see instead?

No validation errors.

#### Environment

browser (oauth)

#### How do you propose to implement this?

Extend the password validation rules in the oauth app.
Consider also https://github.com/TheThingsNetwork/lorawan-stack/issues/1067

#### Can you do this yourself and submit a Pull Request?

Yes

Issue #1067, opened 2019-07-29
Make password rules configurable


pgalic96:
#### Summary

We should have a config file that would define the password rules so that they are not hardcoded in the front-end.

#### Why do we need this?

Currently password rules are defined in oauth front-end as `const`. If we change them such that they are config, we can change and extend them independently of the codebase.

#### What is already there? What do you see now?

Password rules are hard-coded in every file where password validation is used.

#### What is missing? What do you want to see?

Config file where all password rules would be defined.

#### Environment

Current master

#### How do you propose to implement this?

Have a config file and include it wherever we need validation.

#### Can you do this yourself and submit a Pull Request?

Yes, or someone else front-end related.

cc @kschiffer

johanstokking:
Please use the IS configured rules; https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/identityserver/identityserver.go#L54-L60

johanstokking:
@pgalic96 _always_ set a milestone

johanstokking:
cc @bafonins

bafonins:
@htdvisser Shall we merge [PasswordRequirements](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/identityserver/identityserver.go#L54-L60) with [OAuth Config](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/oauth/server.go#L76-L79) so the oauth server has access to those values and can embed them into `APP_DATA`?

bafonins:
Putting this one on hold until we decide how to properly propagate the is config to the oauth app

jpmeijers:
I was a little baffled by the default password complexity rule:
![image](https://user-images.githubusercontent.com/948278/74590451-3280e780-5017-11ea-8abc-fcfe95749b15.png)

I thought that the modern default complexity rules should be on length, not on special characters as those don't make a password stronger.

See the NIST guidelines, Appendix A:
https://pages.nist.gov/800-63-3/sp800-63b.html#appA

> Research has shown, however, that users respond in very predictable ways to the requirements imposed by composition rules [Policies]. For example, a user that might have chosen “password” as their password would be relatively likely to choose “Password1” if required to include an uppercase letter and a number, or “Password1!” if a symbol is also required.
> 
> Users also express frustration when attempts to create complex passwords are rejected by online services.

johanstokking:
[...]

PR #1244, opened 2019-08-26
Update password validation in oauth app

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1243

#### Changes

~- Adapt the `special` regex for password validation in the oauth app~
- Remove password complexity validation from the oauth app.

#### Notes for Reviewers

~@htdvisser Shall we address https://github.com/TheThingsNetwork/lorawan-stack/issues/1067 here as well? If so, shall we just inject [PasswordRequirements](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/identityserver/identityserver.go#L54-L60) into `window.PAGE_DATA` ?~


bafonins:
> It's actually a bit more complicated since #1239 
I see.

Then I think it makes sense to delegate this type of validation to the backend and validate on the client length, presence and whether passwords match.

> I don't know if we can put it in PAGE_DATA

Will check it

bafonins:
>> I think for now it's fine to remove it from the frontend and leave it to the backend. Of course in the future it could be nice to make it a bit more interactive with some kind of password strength indicator.

> Can you create an issue about this?

Will do

PR #1245, opened 2019-08-26
Implement LoRa Alliance TR005 QR code parsing

johanstokking:
#### Summary

Needed for https://github.com/TheThingsIndustries/lorawan-stack/issues/1673

#### Changes

- Implement basic QR code plumbing and interfaces
- Implement TR005 Draft 2

#### Notes for Reviewers

TR005 is expected to come out of draft during India AMM. We don't have a compatibility commitment on exposed struct names and `Parse()` returns QR-code-independent interfaces, but we could keep this implementation with a type alias if it doesn't change. Still using `Draft2` gives enough human friendly warnings in the future if the TR becomes stale.

TR005 Draft 2 is LoRa Alliance confidential so please search document in workspace.

### DIFF
--- a/pkg/qrcode/lora_alliance_tr005.go
+++ b/pkg/qrcode/lora_alliance_tr005.go
@@ -1,0 +1,134 @@
+// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package qrcode
+
+import (
+	"bytes"
+	"encoding/hex"
+	"fmt"
+	"strings"
+
+	"go.thethings.network/lorawan-stack/pkg/types"
+)
+
+// LoRaAllianceTR005Draft2 is the LoRa Alliance defined format in Technical Recommendation TR005 Draft 2.
+type LoRaAllianceTR005Draft2 struct {
+	JoinEUI,
+	DevEUI types.EUI64
+	VendorID             [2]byte
+	ModelID              [2]byte
+	DeviceValidationCode []byte
+	SerialNumber,
+	Proprietary string
+}
+
+// validTR005ExtensionChars defines the QR code alphanumeric character set except :, % and space.
+const validTR005ExtensionChars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ$*+-./"
+
+func (m LoRaAllianceTR005Draft2) validateExtensionChars(s string) error {
+	for _, r := range s {
+		if strings.IndexRune(validTR005ExtensionChars, r) == -1 {
+			return errCharacter.WithAttributes("r", r)
+		}
+	}
+	return nil
+}
+
+// Validate implements the Data interface.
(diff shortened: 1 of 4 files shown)


pkg/qrcode/qrcode.go
```diff
@@ -0,0 +1,48 @@
...
+type AuthenticatedEndDeviceIdentifiers interface {
+	AuthenticatedEndDeviceIdentifiers() (joinEUI, devEUI types.EUI64, authenticationCode []byte)
+}
+
+var errFormat = errors.DefineInvalidArgument("format", "invalid format")
+
+// Parse attempts to parse the given QR code data.
+func Parse(data []byte) (Data, error) {
+	for _, model := range [...]Data{
+		&LoRaAllianceTR005Draft2{},
```
johanstokking: We might change this to a registry of formats, but then the order becomes important.

TR005 is easily recognizable by the header, but many QR codes are not, so switching on the prefix here is not very future proof.

Suggesting to figure out a more decent solution when we have more QR codes to deal with.

pkg/qrcode/lora_alliance_tr005.go
```diff
@@ -0,0 +1,101 @@
...
+	Proprietary string
+}
+
+// MarshalText implements the TextMarshaler interface.
+func (m LoRaAllianceTR005Draft2) MarshalText() ([]byte, error) {
+	var ext string
+	if len(m.DeviceValidationCode) > 0 {
+		ext += fmt.Sprintf("%%V%X", m.DeviceValidationCode[:])
+	}
+	if m.SerialNumber != "" {
```
htdvisser: Do we need to validate the charset used in `SerialNumber` and `Proprietary`?

pkg/qrcode/lora_alliance_tr005.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	if ext != "" {
+		ext = ":" + ext
+	}
+	return []byte(fmt.Sprintf("URN:LW:DP:%X:%X:%X%X%s", m.JoinEUI[:], m.DevEUI[:], m.VendorID[:], m.ModelID[:], ext)), nil
+}
+
+// UnmarshalText implements the TextUnmarshaler interface.
+func (m *LoRaAllianceTR005Draft2) UnmarshalText(text []byte) error {
+	parts := strings.SplitN(string(text), ":", 7)
```
htdvisser: Could also use `bytes.SplitN` and `bytes.Equal` so that we don't have to convert back and forth all the time.

pkg/qrcode/lora_alliance_tr005.go
```diff
@@ -0,0 +1,101 @@
...
+	VendorID             [2]byte
+	ModelID              [2]byte
+	DeviceValidationCode []byte
+	SerialNumber,
+	Proprietary string
+}
+
+// MarshalText implements the TextMarshaler interface.
+func (m LoRaAllianceTR005Draft2) MarshalText() ([]byte, error) {
+	var ext string
```
htdvisser: Could consider using a `strings.Builder` here

pkg/qrcode/lora_alliance_tr005.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	if buf, err := hex.DecodeString(parts[5]); err == nil && len(buf) == 4 {
+		copy(m.VendorID[:], buf[:2])
+		copy(m.ModelID[:], buf[2:])
+	} else if len(buf) != 4 {
+		return errFormat
+	} else {
+		return err
+	}
+	if len(parts) == 7 {
```
htdvisser: I thought we preferred `return`ing if it's not `7`

pkg/qrcode/lora_alliance_tr005.go
johanstokking: Doesn't get more elegant and we're probably close to I/O or rendering here anyway

PR #1258, opened 2019-08-29
Don't try to parse EUI from ID in Basic Station frontend

htdvisser:
#### Summary

This Pull Request deletes some unnecessary code that attempted to parse a GatewayID into an EUI. `FillGatewayContext` already ensures that there is an EUI for gateways that have one.

This also fixes a bug where Basic Station gateways with a GatewayID that doesn't follow the v2 `eui-xxx` pattern couldn't send uplinks.

#### Changes

- Removed calls to `GetEUIfromUID`
- Removed `GetEUIfromUID` and tests

#### Release Notes

- Fixed a bug where uplinks from some Basic Station gateways resulted in the connection to break.

### DIFF
--- a/pkg/gatewayserver/io/basicstationlns/messages/upstream.go
+++ b/pkg/gatewayserver/io/basicstationlns/messages/upstream.go
@@ -33,7 +33,6 @@ var (
 	errJoinRequestMessage = errors.Define("join_request_message", "invalid join-request message received")
 	errUplinkDataFrame    = errors.Define("uplink_data_Frame", "invalid uplink data frame received")
 	errUplinkMessage      = errors.Define("uplink_message", "invalid uplink message received")
-	errGatewayID          = errors.Define("gateway_id", "invalid gateway ID `{id}`")
 )
 
 // UpInfo provides additional metadata on each upstream message.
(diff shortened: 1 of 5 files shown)


htdvisser:
All red diff 💪

pkg/gatewayserver/io/basicstationlns/messages/upstream.go
```diff
@@ -175,11 +175,6 @@ func (req *JoinRequest) ToUplinkMessage(ids ttnpb.GatewayIdentifiers, bandID str
 		rxTime = &val
 	}
 
-	ids.EUI, err = GetEUIfromUID(ids.GatewayID)
-	if err != nil {
-		return nil, errGatewayID.WithAttributes("id", ids.GatewayID).WithCause(err)
```
johanstokking: Not sure if we need this error still, but afaik we _always_ use unique IDs otherwise it doesn't say anything

htdvisser: You're right, the error is unused. I did a search for `errGatewayID` and found some references in `pkg/gatewayserver/io/basicstationlns` but apparently that error is redefined in that package. Will remove.

Issue #1250, opened 2019-08-26
Document interoperability via Semtech Join-Server


rvolosatovs:
#### Summary

Setting up interoperability for Semtech devices via Semtech Join Server should be documented

#### Why do we need this?

To make lives of our users easier

#### What is already there? What do you see now?

Nothing

#### What is missing? What do you want to see?

Step-by-step guide from registering NS(??) to successful join

#### Can you do this yourself and submit a Pull Request?

yes

rvolosatovs:
@johanstokking on Slack you mentioned:
> So it does include creating the NS/AS

However, our NS/AS would already be registered at Semtech.
Is the intended use case to use this guide for private deployments?

What about the KEKs? The users would need to be able to add KEKs to the NS/AS in order for devices to actually function. How can that be done? Is that already documented somewhere?

PR #1261, opened 2019-08-29
Allow multiple JS interop config paths

rvolosatovs:
#### Summary

Currently the JS interop config contains singular (optional) `path` option, e.g. for Semtech it's equal to `api/v1/rens/rens-1::3/lbi_joinreq`.

However, according the Backend Interfaces spec there's at least 1 other RPC possible that JS can expose - the AS-JS RPC for `AppSKey` delivery, which could possibly require a different path.
See https://thethingsindustries.slack.com/archives/C06UQSMPB/p1567068229008700 for more info.

#### Changes

- Allow different paths to be specified per-RPC for Join-Server
- (very) minor restructuring for greater flexibility
- Fix misplaced test-cases(`GetAppSKey` RPC probably will never fail with a `MICFailed`)

### DIFF
--- a/pkg/interop/client.go
+++ b/pkg/interop/client.go
@@ -93,6 +93,18 @@ func (p *JoinServerProtocol) UnmarshalYAML(unmarshal func(interface{}) error) er
 	}
 }
 
+type jsRPCPaths struct {
+	Join    string `yaml:"join"`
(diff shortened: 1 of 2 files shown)


rvolosatovs:
Blocking #1250

PR #1264, opened 2019-08-29
Increase connection waiting time in AS tests

adriansmares:
#### Summary

References #1017 

#### Changes

- Increase the waiting time for the connection to establish.

#### Notes for Reviewers

I think this should be one of the causes - some messages get lost since they are sent before the integration starts (since `Set` starts the task instead of blocking until it is done).

### DIFF
--- a/pkg/applicationserver/applicationserver_test.go
+++ b/pkg/applicationserver/applicationserver_test.go
@@ -841,7 +841,7 @@ hardware_versions:
 				}
 			}()
 			// Wait for connection to establish.
-			time.Sleep(Timeout)
+			time.Sleep(2 * Timeout)
 
 			t.Run("Upstream", func(t *testing.T) {
 				ns.reset()


htdvisser:
Can't we somehow add a check for that connection establishment? 

Maybe something like

```go
var connectionError error
for i := 0; i < someNumberOfRetries; i++ {
	connectionError = tryConnect()
	if connectionError == nil {
		break
	}
	time.Sleep(someWaitTime)
}
```

adriansmares:
The task we are waiting for is https://github.com/TheThingsNetwork/lorawan-stack/blob/45d517c9c9d1b240f46b6cefe238a1daffcdefa3/pkg/applicationserver/io/pubsub/pubsub.go#L78-L87
during which the AS connects to the external PubSub service - we cannot really poll the status of the task right now, but maybe we can add an status API to the registry.

adriansmares:
I'm going to try this in another way since just increasing timeouts just moves the issue to another point in the future - I'll add events (#1286) and then wait for the integration started event to become ready. This way we don't need dirty test hooks and we also support more of the events flow.

PR #1268, opened 2019-08-30
Remove trailing quote from Web UI configuration variables

kurtmc:
#### Summary

Tidy up `docker-compose.yml`. I encountered this issue when I tried to use these variables when trying to self host this system, the trailing quotes where not obvious enough for me to catch initially, so it took me a while to find this issue when getting setup.

#### Changes

- Remove unnecessary quote from commented out variables.


htdvisser:
Thanks!

Issue #1110, opened 2019-08-06
NS may schedule more MAC commands than can be answered


rvolosatovs:
#### Summary

1.1 spec states that NS may only schedule as much MAC requests, as can be answered in one frame by the device, however that is dropped in 1.1.1 and is not the case in 1.0.4

#### Why do we need this?

To be more efficient

#### What is already there? What do you see now?

Only scheduling as many MAC requests as can be answered

#### What is missing? What do you want to see?

Schedule as many MAC requests as fit in the frame

#### How do you propose to implement this?

Do not check if the MAC requests can be answered, only check the length. Possibly only do this for versions `1.0.x` and `1.1.1+`

#### Can you do this yourself and submit a Pull Request?

yes

Issue #1109, opened 2019-08-06
Schedule MAC answers as Class A downlink


rvolosatovs:
#### Summary

All MAC commands have to be answered in Class A downlinks, even for devices in class B/C modes.

#### Why do we need this?

To properly support class B/C devices

#### What is already there? What do you see now?

Scheduling Rx2 for class C "whenever", even if Class A MAC answers are present

#### What is missing? What do you want to see?

Schedule Rx2 as Class A downlink if class A MAC answers are present.
If (Class A) Rx2 is expired, the MAC answers should be dropped.

#### How do you propose to implement this?

Set both Rx1 and Rx2 in `TxRequest` in class C as in Class A if Class A MAC answers present.

#### Can you do this yourself and submit a Pull Request?

yes

PR #1287, opened 2019-09-02
Schedule MAC answers in Class A

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1109
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1110 (only account for this in 1.1)

#### Changes

- Only schedule MAC answers in Class A for all devices
- Only schedule `PingSlotReq` in Class A downlink slots
- Only skip Class A downlink slots for Class B/C devices if no MAC answers are to schedule and RX1 is already expired
- Removed necessity for `deepcopy` by moving the mutation of device to record the downlink out of `generateDownlink` 

#### Notes for Reviewers

The diff is quite unfortunate, but since class A downlink scheduling logic is now shared between devices of all classes, there's not much I can do to improve it.
`generateDownlink` goes through application downlink before MAC processing, such that we can access `dev.QueuedResponses` directly before it may get mutated. That also means we don't need to do redundant work of adding MAC commands to class A downlink buffer if we're going to abort it later to do B/C and that we actually may have more opportunities to send class A downlink(e.g. if absolute time downlink is expired and next one in queue is not Class B/C bound, we can immediately schedule the *next* downlink in Class A)

Tested with Tektelic + ff1705:
```
Initialized Arm Mbed LoRaWAN stack
[DBG ][LSTK]: Initiating OTAA
[DBG ][LSTK]: Sending Join Request ...
[DBG ][LMAC]: Frame prepared to send at port 0
[DBG ][LMAC]: TX: Channel=2, TX DR=5, RX1 DR=5
Joining network...
[DBG ][LSTK]: Transmission completed
[DBG ][LMAC]: RX1 slot open, Freq = 868500000
[DBG ][LSTK]: OTAA Connection OK!
Joined network
[DBG ][LMAC]: Changing device class to -> CLASS_C
[DBG ][LMAC]: RX2 slot open, Freq = 869525000
Switched to class C
[INFO][LMAC]: RTS = 0 bytes, PEND = 0, Port: 1
[DBG ][LMAC]: Frame prepared to send at port 1
[DBG ][LMAC]: TX: Channel=7, TX DR=5, RX1 DR=5
Sending 0 bytes
[DBG ][LSTK]: Transmission completed
[DBG ][LMAC]: RX2 slot open, Freq = 869525000
Message sent
[DBG ][LMAC]: RX1 slot open, Freq = 867900000
[DBG ][LMAC]: RX2 slot open, Freq = 869525000
Message sent
Message sent
[DBG ][LSTK]: Packet Received 2 bytes, Port=15
Message sent
Received message on port 15 (2 bytes, flags = 1): ab ba 
Message sent
[DBG ][LSTK]: Packet Received 4 bytes, Port=2
Message sent
Received message on port 2 (4 bytes, flags = 1): 2d 5a 6d 5d 
Synchronized clock to Mon Sep  2 18:06:37 2019
Message sent
```

#### Release Notes

- Improved compatibility with various Class C devices

### DIFF
--- a/pkg/networkserver/downlink.go
+++ b/pkg/networkserver/downlink.go
@@ -837,6 +837,157 @@ func (ns *NetworkServer) recordDownlink(dev *ttnpb.EndDevice, genDown *generated
 	return nextDownlinkAt
 }
 
+func (ns *NetworkServer) attemptClassADownlink(ctx context.Context, dev *ttnpb.EndDevice, phy band.Band, fp *frequencyplans.FrequencyPlan, maxUpLength uint16, queuedApplicationUplinks []*ttnpb.ApplicationUp, queuedEvents []events.Event) ([]string, []*ttnpb.ApplicationUp, []events.Event, time.Time, bool) {
+	var sets []string
+	logger := log.FromContext(ctx)
+	if len(dev.RecentUplinks) == 0 {
+		logger.Warn("Rx windows available, but no uplink present, skip class A downlink slot")
+		dev.MACState.QueuedResponses = nil
+		dev.MACState.RxWindowsAvailable = false
+		return append(sets,
+			"mac_state.queued_responses",
+			"mac_state.rx_windows_available",
+		), queuedApplicationUplinks, queuedEvents, time.Time{}, false
+	}
+
+	var rxDelay ttnpb.RxDelay
+	up := dev.RecentUplinks[len(dev.RecentUplinks)-1]
+	switch up.Payload.MHDR.MType {
+	case ttnpb.MType_CONFIRMED_UP, ttnpb.MType_UNCONFIRMED_UP:
+		rxDelay = dev.MACState.CurrentParameters.Rx1Delay
+		if rxDelay == ttnpb.RX_DELAY_0 {
+			rxDelay = ttnpb.RX_DELAY_1
+		}
+
+	case ttnpb.MType_REJOIN_REQUEST:
+		rxDelay = ttnpb.RxDelay(phy.JoinAcceptDelay1 / time.Second)
+
+	default:
+		logger.Warn("Last uplink is neither data uplink, nor rejoin-request, skip class A downlink slot")
+		dev.MACState.QueuedResponses = nil
+		dev.MACState.RxWindowsAvailable = false
+		return append(sets,
+			"mac_state.queued_responses",
+			"mac_state.rx_windows_available",
+		), queuedApplicationUplinks, queuedEvents, time.Time{}, false
+	}
+	ctx = events.ContextWithCorrelationID(ctx, up.CorrelationIDs...)
+
+	rx1, rx2, paths := downlinkPathsForClassA(rxDelay, dev.RecentUplinks...)
+	if !rx1 && !rx2 {
+		logger.Warn("Rx1 and Rx2 are expired, skip class A downlink slot")
+		dev.MACState.QueuedResponses = nil
+		dev.MACState.RxWindowsAvailable = false
+		return append(sets,
+			"mac_state.queued_responses",
+			"mac_state.rx_windows_available",
+		), queuedApplicationUplinks, queuedEvents, time.Time{}, false
+	}
+	if len(paths) == 0 {
+		logger.Warn("No downlink path available, skip class A downlink slot")
+		return sets, queuedApplicationUplinks, queuedEvents, time.Time{}, false
+	}
+	if dev.MACState.DeviceClass != ttnpb.CLASS_A && !rx1 && rx2 && len(dev.MACState.QueuedResponses) == 0 {
+		logger.Debug("Skip Rx2-only Class A downlink for non-class A device")
+		return sets, queuedApplicationUplinks, queuedEvents, time.Time{}, false
+	}
+
+	req, err := txRequestFromUplink(phy, dev.MACState, rx1, rx2, rxDelay, up)
+	if err != nil {
+		logger.WithError(err).Warn("Failed to generate Tx request from uplink, skip class A downlink slot")
+		return sets, queuedApplicationUplinks, queuedEvents, time.Time{}, false
+	}
+
+	var maxDR ttnpb.DataRateIndex
+	switch {
+	case rx1 && rx2:
+		maxDR = req.Rx1DataRateIndex
+		if req.Rx2DataRateIndex > maxDR {
+			maxDR = req.Rx2DataRateIndex
+		}
+
+	case rx1:
+		maxDR = req.Rx1DataRateIndex
+
+	case rx2:
+		maxDR = req.Rx2DataRateIndex
+	}
+
+	genDown, genState, err := ns.generateDownlink(ctx, dev, phy, ttnpb.CLASS_A,
+		phy.DataRates[maxDR].DefaultMaxSize.PayloadSize(fp.DwellTime.GetDownlinks()),
+		maxUpLength,
+	)
+	if genState.NeedsDownlinkQueueUpdate {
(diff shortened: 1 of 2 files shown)


pkg/networkserver/downlink.go
```diff
@@ -931,25 +965,32 @@ func (ns *NetworkServer) processDownlinkTask(ctx context.Context) error {
...
-					return dev, nil, nil
+				var maxUpLength uint16 = math.MaxUint16
+				if dev.MACState.LoRaWANVersion == ttnpb.MAC_V1_1 {
+					maxUpLength = maximumUplinkLength(fp, phy, dev.RecentUplinks...)
 				}
 
-				maxUpLength := maximumUplinkLength(fp, phy, dev.RecentUplinks...)
+				var sets []string
+			outer:
+				for dev.MACState.RxWindowsAvailable {
```
rvolosatovs: This is quite unconventional use of `for`, I use it here to get a way to go to exit early from the block.
Other options:
1. very deeply nested `if`
2. a very specialized named function with several pointer arguments/return values
3. anonymous function executed right on spot (`func(){ /* class A downlink handling */ }()`), which has a negative of confusing `return` statements
4. `goto` - works out-of-the-box, but AFAIK @johanstokking is anti-`goto`

2. 3. suffer from issue of inability to `return` directly from the block(stop the transaction), so additional error handling should be added outside the function

IMO this is a perfect use-case for `goto`

johanstokking: I think option 2 is the way to go. This method is already more than 500 lines long

I'm not strictly against this use of `for`, but I have increasing concerns with the maintainability of the NS code base

rvolosatovs: What are the concerns exactly?

johanstokking: Mostly the lengthy funcs. I think we do a good job given all the conditions, it's not meant offensive, but code tricks like this make it hard to read, let alone maintain, for somebody else. If a new person becomes code owner of NS at some time in the future, they have to figure out why there's a `for { }` here that's not meant as a loop, and make sure it never reaches an infinite loop.

rvolosatovs:
@johanstokking

pkg/networkserver/downlink.go
```diff
@@ -144,27 +143,119 @@ func (ns *NetworkServer) generateDownlink(ctx context.Context, dev *ttnpb.EndDev
...
+				if dev.MACState.DeviceClass != ttnpb.CLASS_A && len(dev.MACState.QueuedResponses) == 0 {
+					dev.QueuedApplicationDownlinks = dev.QueuedApplicationDownlinks[startIdx:]
+					st.NeedsDownlinkQueueUpdate = startIdx > 0
+					return nil, st, errNoDownlink
+				}
+				skipAppDown = true
+			}
+		}
+		break
+	}
```
johanstokking: Just for the record: the above switch statement is very nice and clear

pkg/networkserver/downlink.go
```diff
@@ -803,6 +809,227 @@ const gsScheduleWindow = 30 * time.Second
...
+
+func (res downlinkAttemptResult) AppendApplicationUplinks(ups ...*ttnpb.ApplicationUp) []*ttnpb.ApplicationUp {
+	if res.applicationUpAppender == nil {
+		return ups
+	}
+	return res.applicationUpAppender(ups, res.Scheduled)
+}
+
+func (ns *NetworkServer) attemptClassADownlink(ctx context.Context, dev *ttnpb.EndDevice, phy band.Band, fp *frequencyplans.FrequencyPlan, maxUpLength uint16) downlinkAttemptResult {
+	var sets []string
```
johanstokking: This one is not assigned to but used before L949, is that intentional?

pkg/networkserver/downlink.go
```diff
@@ -803,6 +809,227 @@ const gsScheduleWindow = 30 * time.Second
...
+		}
+		return downlinkAttemptResult{
+			SetPaths:              sets,
+			applicationUpAppender: genState.appendApplicationUplinks,
+		}
+	}
+
+	if rx1 && rx2 {
+		rx1 = len(genDown.Payload) <= int(phy.DataRates[req.Rx1DataRateIndex].DefaultMaxSize.PayloadSize(fp.DwellTime.GetDownlinks()))
+		rx2 = len(genDown.Payload) <= int(phy.DataRates[req.Rx2DataRateIndex].DefaultMaxSize.PayloadSize(fp.DwellTime.GetDownlinks()))
```
johanstokking: Changing the meaning of the variables here can cause confusion in this block

pkg/networkserver/downlink.go
```diff
@@ -803,6 +809,227 @@ const gsScheduleWindow = 30 * time.Second
...
+		if !rx1 && !rx2 {
+			logger.Error("Generated downlink payload size does not fit neither Rx1, nor Rx2, skip class A downlink slot")
+			return downlinkAttemptResult{
+				SetPaths:              sets,
+				applicationUpAppender: genState.appendApplicationUplinks,
+			}
```
johanstokking: This block is not tested; should we?

pkg/networkserver/downlink.go
```diff
@@ -803,6 +809,227 @@ const gsScheduleWindow = 30 * time.Second
...
+		req, err = txRequestFromUplink(phy, dev.MACState, rx1, rx2, rxDelay, up)
+		if err != nil {
+			logger.WithError(err).Warn("Failed to generate Tx request from uplink, skip class A downlink slot")
+			if genState.ApplicationDownlink != nil {
+				dev.QueuedApplicationDownlinks = append([]*ttnpb.ApplicationDownlink{genState.ApplicationDownlink}, dev.QueuedApplicationDownlinks...)
+			}
+			return downlinkAttemptResult{
+				SetPaths:              sets,
+				applicationUpAppender: genState.appendApplicationUplinks,
+			}
```
johanstokking: Idem

pkg/networkserver/downlink.go
```diff
@@ -1288,7 +1312,7 @@ func (ns *NetworkServer) processDownlinkTask(ctx context.Context) error {
 
 							if len(genState.ApplicationDownlink.GetClassBC().GetGateways()) > 0 &&
 								allErrors(nonRetryableFixedPathGatewayError, pathErrs...) {
-								logger.WithField("retry_at", nextDownlinkAt).Warn("Fixed paths invalid, fail downlink and retry downlink slot")
+								logger.WithField("retry_at", nextDownlinkAt).Warn("Fixed paths invalid, fail application downlink and retry attempt")
```
johanstokking: Why still retry here? The errors are all non-retryable right?

johanstokking: Also this block is not tested

pkg/networkserver/downlink.go
rvolosatovs: Yes.

pkg/networkserver/downlink.go
rvolosatovs: The meaning of the variables does not change - in both cases it's `is rx{1,2} possible?`

pkg/networkserver/downlink.go
rvolosatovs: The downlink causing the non-retryable errors is dropped(and AS is informed about it), but NS still tries to schedule a (different) downlink at a later time, if there is a downlink to schedule.

pkg/networkserver/downlink.go
johanstokking: OK

Issue #1294, opened 2019-09-03
Add organization overview page


bafonins:
#### Summary

The console should display extended information about organizations.

#### Why do we need this?

The users should be able to see the overview of an organization.

#### What is already there? What do you see now?

1. `<EventsSubscription />` component
2. `<Organizations />` component
3. Organizations service in the js sdk

#### What is missing? What do you want to see?

1. `<OrganizationOverview />` page
2. `<OrganizationData />` page (for convenience, since the widget requires a link to the full events components
3. `<OrganizationEvents />` component
4. Redux setup for fetching single organization and processing events
5. `<SideNavigation />` for organizations
6. `getById` and `openStream` methods in the organizations service in the js sdk

#### Environment

browser (console)

#### How do you propose to implement this?

Identical to application overview

#### Can you do this yourself and submit a Pull Request?

yes

PR #1295, opened 2019-09-03
Add organization overview/data pages

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1294

Screenshot 2019-09-03 at 13 42 05
Screenshot 2019-09-03 at 13 42 11

#### Changes

- Add required pages
- Change navigation to the organization overview on successful creation
- Add required redux primitives

### DIFF
--- a/pkg/webui/console/views/organization-data/organization-data.js
+++ b/pkg/webui/console/views/organization-data/organization-data.js
@@ -1,0 +1,67 @@
+// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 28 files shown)


pkg/webui/console/views/organization-data/organization-data.js
```diff
@@ -0,0 +1,67 @@
...
+    orgId: PropTypes.string.isRequired,
+  }
+
+  render() {
+    const { orgId } = this.props
+
+    return (
+      <Container>
+        <Row>
+          <Col lg={8} md={12}>
```
kschiffer:
```suggestion
          <Col sm={12}>
```
I know we might do it like that for the other data pages, but it really should be like this.

Issue #1200, opened 2019-08-19
Add feature flags to the console


kschiffer:
#### Summary
The console should provide certain functionality only if the current stack setup (enabled components, current rights) would allow supporting this functionality.

#### Why do we need this?
Based on the configuration and right setup, we already know which functionality would not work. It makes sense to not display this functionality in the console, as we would otherwise wrongfully suggest that it is still possible to use.

#### What is already there? What do you see now?
The SDK respecting the `stackConfig` (including what components are enabled) and avoids request to components that are known to be disabled.

#### What is missing? What do you want to see?
A way to easily disable components (and functionality in general) based on the known list of rights and disabled components.

#### Environment
Console.

#### How do you propose to implement this?
This appears like a very good use case for a higher-order component. I could envision a usage like this:
```js
@withRequirement(['as', 'is'], ['RIGHT_APPLICATION_LINK', 'RIGHT_APPLICATION_INFO'])
class MyComponent extends React.PureComponent {
…
}
```
Additionally, for a more fine-grained control inside rendering functions, we should add a wrapper component to check a set of requirements against the current context and configuration. It would then only render its children when the requirement passes:
```js
render () {
  …
  <Requires components=['as', 'is'] rights=['RIGHT_APPLICATION_LINK', 'RIGHT_APPLICATION_INFO']>
    <Link to='…'>Link</Link>
  </Requires>
  …
}
```

Both utilities would get the rights from the store (needs yet to be added there actually) and the component data from the global environment.

#### Can you do this yourself and submit a Pull Request?
Yes. 

Input welcome cc @bafonins

bafonins:
What about the `<SideNavigation />` entries?

kschiffer:
So I think we need to split this into three aspects:
1. Hiding links from the navigation (top, side and tab navigation)
2. Preventing access to views with disabled features (should be mostly a redirect to the parent view)
3. Hiding bits inside views that are not accessible (eg. the device table inside the application overview, when the user has no right to view devices)

All in all, I think it would be good to have a central module that determines whether a feature is available or not, which we can then refer to for all three cases. This enables us to change conditions for particular features easily.

```js
// pkg/webui/console/lib/feature-toggles.js

export mayViewApplicationDevices = () => 
  userHasRights([RIGHT_APPLICATION_DEVICES_READ])
  && stackComponentsEnabled(['is']) // I know, checking for 'is' here is unnecessary, it's just for illustration purposes
…
```

[...]

PR #1316, opened 2019-09-05
Add option to automatically link apps on creation

### LINKED ISSUE #1282: Automatic linking on application create
#### Summary

Automatic linking on application create

#### Why do we need this?

To avoid the hassle of creating an API key and setting the link

Manual creation of API key and setting the link

A checkbox on application create to link the application.

This is only supported if there's a AS in the cluster

On successful create in the Entity Registry and if the checkbox is checked, create a new API key with the `RIGHT_APPLICATION_LINK` right and probably a name to indicate it's created by the Console/CLI and set the link on AS.

Can review

kschiffer:
#### Summary
Closes #1282 
image

#### Changes
- Add fields for linking the app automatically
- Add logic to submit logic to create an api key and use it to link the app
- Refactor some messages

#### Notes for Reviewers
It would be good to also check whether the current user has the right to link applications, but I would like to do that via #1200. For now, it only checks whether the AS is in the cluster.

#### Release Notes
Added option to link applications automatically upon creation


johanstokking:
Hallelujah!

pkg/webui/console/views/application-add/index.js
```diff
@@ -39,12 +44,17 @@ const m = defineMessages({
...
+  linkFailure: 'There was a problem while linking the application',
+  linkFailureTitle: 'Application link failed',
 })
 
 const initialValues = {
   application_id: '',
   name: '',
   description: '',
+  link: true,
+  network_server_address: '',
```
bafonins: missing validation, see https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/webui/console/views/application-link/index.js#L78

pkg/webui/console/views/application-add/index.js
```diff
@@ -59,6 +69,7 @@ const validationSchema = Yup.object().shape({
   description: Yup.string(),
 })
 
+@withEnv
```
bafonins: we have env selectors. you can use `selectAsConfig` to pass `asEnabled` via `connect`

PR #1317, opened 2019-09-05
Fix callback regression from #1307

htdvisser:
#### Summary

This is a quickfix for a regression introduced by #1307 that caused `code` and `local-callback` redirects to lose the `/oauth` prefix.


johanstokking:
Works

Issue #789, opened 2019-06-03
Duplicate events


rvolosatovs:
#### Summary

NS publishes an event whenever a MAC command is added to a buffer(in `generateDownlink`), however that does not equate to NS actually scheduling the MAC command. Additionally, the downlink may be generated twice (e.g. in case Rx1 is not possible for Class C device)

#### Steps to Reproduce

1. Look at the source

#### What do you see now?

Duplicate/inapplicable events

#### What do you want to see instead?

Non-duplicate, applicable events

#### How do you propose to implement this?

Return a slice of events from `generateDownlink`

#### Can you do this yourself and submit a Pull Request?

yes

rvolosatovs:
Note that the issue also occurs when processing the uplink - in case the transaction fails and we retry(#33), the events would be duplicated

htdvisser:
> **How do you propose to implement this?**
> Return a slice of events from `generateDownlink`

This description of a proposed implementation needs a bit more information before I can decide whether I agree with the proposal or not. 

Maybe this was discussed with @johanstokking offline while I was on vacation, but I would still like to see some more details in this issue.

johanstokking:
FTR this wasn't discussed offline

rvolosatovs:
That's very simple really, MAC handlers/enqueuers shall not publish events, but rather return events, which NS can later decide to publish.

rvolosatovs:
https://github.com/rvolosatovs/lorawan-stack/blob/c10ca5fed8e317926b60cfb71e9f70c2414b3419/pkg/networkserver/downlink.go#L104

And so if NS downlink was successfully processed, NS does this:(`queuedEvents` contains the `Events` returned `generateDownlink` at that point)
https://github.com/rvolosatovs/lorawan-stack/blob/c10ca5fed8e317926b60cfb71e9f70c2414b3419/pkg/networkserver/downlink.go#L1198-L1204

PR #1329, opened 2019-09-06
Various MAC-related fixes and improvements

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1342
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1343

#### Changes

- Add `mac_settings.desired_max_duty_cycle`
- Add `mac_settings.desired_adr_ack_{delay,limit}`
- Trigger a downlink slot for devices in class C mode if desired MAC parameters are changed
- Allow changing desired MAC parameters(this triggers MAC commands in session)
- Fix `band.Band.ADRAck{Limit,Delay}` typing
- Fix `MACSettings` field validations
- Fix `newMACState` panic when desired Rx2 frequency is set
- Fix some small issues with class C downlink, which are caused by the custom behavior we added (cc @johanstokking)

#### Notes for Reviewers

These main purpose of these changes is to allow more fine-grained control over MAC layer and triggering of MAC commands in a session.
Class C downlink fixes are there, since I encountered them while testing this and @johanstokking was the intended reviewer initially. @htdvisser if that's too much for you, I can extract these to a different PR, but this should be doable if you just go commit-by-commit. I conveniently added them as last 2 commits.

Tested on metal

#### Release Notes

- Added support for updating `mac_state.desired_parameters`
- Added `mac_settings.desired_max_duty_cycle`
- Added `mac_settings.desired_adr_ack_delay_exponent`
- Added `mac_settings.desired_adr_ack_limit_exponent`
- Improved confirmed downlink operation


htdvisser:
Removing my review request. This PR is marked as draft and the PR template was not filled. Please mark it as "ready for review" and/or make sure the PR description has some content and/or let me know what I need to review.

pkg/networkserver/mac_adr_param_setup.go
```diff
@@ -54,16 +53,9 @@ func handleADRParamSetupAns(ctx context.Context, dev *ttnpb.EndDevice) ([]events
 	dev.MACState.PendingRequests, err = handleMACResponse(ttnpb.CID_ADR_PARAM_SETUP, func(cmd *ttnpb.MACCommand) error {
 		req := cmd.GetADRParamSetupReq()
 
-		dev.MACState.CurrentParameters.ADRAckDelay = lorawan.ADRAckDelayExponentToUint32(req.ADRAckDelayExponent)
-		dev.MACState.CurrentParameters.ADRAckLimit = lorawan.ADRAckLimitExponentToUint32(req.ADRAckLimitExponent)
+		dev.MACState.CurrentParameters.ADRAckLimit = req.ADRAckLimitExponent
+		dev.MACState.CurrentParameters.ADRAckDelay = req.ADRAckDelayExponent
 
-		if lorawan.Uint32ToADRAckDelayExponent(dev.MACState.DesiredParameters.ADRAckDelay) == req.ADRAckDelayExponent {
```
rvolosatovs: This was necessary for conversion before

rvolosatovs:
Blocking #789

api/end_device.proto
```diff
@@ -116,6 +114,11 @@ message MACParameters {
   // Whether downlink dwell time is set (400ms).
   // If this field is not set, then the value is either unknown or irrelevant(Network Server cannot modify it).
   google.protobuf.BoolValue downlink_dwell_time = 21;
+
+  // ADR: number of messages to wait before setting ADRAckReq.
+  ADRAckLimitExponent adr_ack_limit = 22 [(gogoproto.customname) = "ADRAckLimit", (validate.rules).enum.defined_only = true];
```
htdvisser: This breaks API compatibility for our JSON API (if they could read this field). It was a number, now it's a string.

api/end_device.proto
```diff
@@ -68,10 +68,8 @@ message MACParameters {
   uint32 adr_tx_power_index = 5 [(gogoproto.customname) = "ADRTxPowerIndex", (validate.rules).uint32.lte = 15];
   // ADR: number of retransmissions.
   uint32 adr_nb_trans = 6 [(gogoproto.customname) = "ADRNbTrans", (validate.rules).uint32.lte = 15];
-  // ADR: number of messages to wait before setting ADRAckReq.
-  uint32 adr_ack_limit = 7 [(gogoproto.customname) = "ADRAckLimit", (validate.rules).uint32 = {gte: 1, lte: 32768}];
-  // ADR: number of messages to wait after setting ADRAckReq and before changing TxPower or DataRate.
-  uint32 adr_ack_delay = 8 [(gogoproto.customname) = "ADRAckDelay", (validate.rules).uint32 = {gte: 1, lte: 32768}];
+  reserved 7; // Deprecated: adr_ack_limit
```
htdvisser: This breaks DB compatibility for already stored end devices, right?

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -458,43 +467,73 @@ func setField(rv reflect.Value, path []string, v reflect.Value) error {
 						field.Set(reflect.ValueOf(types.BytesValue{Value: buf}))
 					}
 				case ft.PkgPath() == "go.thethings.network/lorawan-stack/pkg/ttnpb":
-					switch ft.Name() {
+					switch typeName := ft.Name(); typeName {
 					case "MACSettings_DataRateIndexValue":
-						if enumValue, ok := proto.EnumValueMap("ttn.lorawan.v3.DataRateIndex")[v.String()]; ok {
+						if enumValue, ok := proto.EnumValueMap(unwrapMACSettingsEnumType(typeName))[v.String()]; ok {
```
htdvisser: FTR: I'm also fine to do everything through `UnmarshalText`

pkg/ttnpb/field_mask_validation.go
```diff
@@ -531,6 +543,21 @@ var AllowedFieldMaskPathsForRPC = map[string][]string{
 		"mac_settings.status_time_periodicity",
 		"mac_settings.supports_32_bit_f_cnt",
 		"mac_settings.use_adr",
+		"mac_state.desired_parameters.adr_nb_trans",
```
htdvisser: This means that `api/end_device.protoL57` (`// This is used internally by the Network Server and is read only.`) is no longer correct.

api/end_device.proto
rvolosatovs: It was not possible to set these via API and I'm quite sure console does not use these. @bafonins @kschiffer can you confirm?

api/end_device.proto
rvolosatovs: No. Why?
We do lose the old values, but all of these are default values from band, which, in fact, are all equal to the same value.
In order to keep the behavior the same a possibility is to wrap the ADRAck*Exponents in structs to distinguish between them being set and not and if not set, use the default values from band.

htdvisser: If we lose data, then I'd say this counts as breaking DB compatibility.

I don't think it's a big issue, because there was no way for users to set these fields to anything other than the defaults, but still.

api/end_device.proto
htdvisser: This isn't about the console using the fields or not, this is about breaking API for which we promised stability. If clients were able to GET these fields before, and now suddenly receive a string instead of a number, we might break those clients if they rely on a field being a specific type. This is the case for any hypothetical Go client that uses JSON encoding instead of Protobuf.

htdvisser: With our compatibility promise in mind, I think the right way to do this would be to add the new fields with a different name, like `adr_ack_limit_exponent`, while keeping the old ones and fill the value of `adr_ack_limit` with the corresponding number from the `_exponent`.

Since clients were only allowed to GET these fields, and not to SET them, we only have to implement the SET logic for the `_exponent` fields.

api/end_device.proto
```diff
@@ -116,6 +118,11 @@ message MACParameters {
   // Whether downlink dwell time is set (400ms).
   // If this field is not set, then the value is either unknown or irrelevant(Network Server cannot modify it).
   google.protobuf.BoolValue downlink_dwell_time = 21;
+
+  // ADR: number of messages to wait before setting ADRAckReq.
+  ADRAckLimitExponentValue adr_ack_limit_exponent = 22 [(gogoproto.customname) = "ADRAckLimit"];
```
htdvisser: Naming `adr_ack_limit_exponent` `ADRAckLimit` in Go is a bit confusing. Why can't we just call this `ADRAckLimitExponent`, and leave the old one the way it is (instead of renaming that to `DEPRECATED_ADRAckLimit`). Some "deprecated, use exponent instead" comments on those fields should be sufficient.

api/end_device.proto
```diff
@@ -69,9 +69,11 @@ message MACParameters {
   // ADR: number of retransmissions.
   uint32 adr_nb_trans = 6 [(gogoproto.customname) = "ADRNbTrans", (validate.rules).uint32.lte = 15];
   // ADR: number of messages to wait before setting ADRAckReq.
-  uint32 adr_ack_limit = 7 [(gogoproto.customname) = "ADRAckLimit", (validate.rules).uint32 = {gte: 1, lte: 32768}];
+  // This field is deprecated, use adr_ack_limit_exponent instead.
+  uint32 adr_ack_limit = 7 [(gogoproto.customname) = "DEPRECATED_ADRAckLimit"];
```
htdvisser: I don't like that we're now generating `DEPRECATED_` fields in our Go code. The comment and `(deprecated) = true` option should be sufficient to indicate that a field is deprecated.

api/end_device.proto
rvolosatovs: Because only the type of the field changed, it still represents the exact same this - the ADRAckLimit.
Also note that we have e.g. `mac_settings.desired_adr_ack_delay`, which is exponent type as well. The user does not need to use `mac_settings.desired_adr_ack_delay_exponent`, that's just confusing, however, by looking the docs the user will see that the ADRAckLimit is set *via* exponent.
Also note that we have multiple other fields, which use enum types instead of plain uints and we don't add any extra suffixes to indicate that.
The only reason the *_exponent* suffix was added to these 2 fields is to not break existing clients using JSON.

api/end_device.proto
rvolosatovs: What other suffix/prefix do you suggest?

htdvisser: My suggestion is to leave the name of this field as `ADRAckLimit`, add the `(deprecated) = true` option and name the new field `ADRAckLimitExponent`.

rvolosatovs: please see https://github.com/TheThingsNetwork/lorawan-stack/pull/1329#discussion_r326994634 why that's not a good solution.
What's your suggestion of naming deprecated fields when we have a name clash?

api/end_device.proto
htdvisser: 1. That is not true. The field now represents the exponent of the limit, not the limit itself.
2. This is something we should take as a lesson for future API design. If something is an index or exponent instead of the exact value, we should use a suffix. We can't change that now, so we'll have to move forward with this inconsistency.
3. There is absolutely no reason to make the generated name in Go different than the proto name.

api/end_device.proto
htdvisser: I think we shouldn't rename old fields at all. That's just something that has to wait until v4.

johanstokking: We can `reserve` deprecated fields if they're not used anymore at all. In this case, it seems to be used, and we cannot change rename.

rvolosatovs:
@htdvisser I'm only seeing these `ns` flags on `go run ./cmd/ttn-lw-stack --help`:
```
      --ns.cooldown-window duration                                    Time window starting right after deduplication window, during which, duplicate messages are discarded (default 1s)
      --ns.deduplication-window duration                               Time window during which, duplicate messages are collected for metadata (default 200ms)
      --ns.default-mac-settings.adr-margin float                       The default margin Network Server should add in ADR requests if not configured in device's MAC settings (default 15)
      --ns.default-mac-settings.class-b-timeout duration               Deadline for a device in class B mode to respond to requests from the Network Server if not configured in device's MAC settings (default 1m0s)
      --ns.default-mac-settings.class-c-timeout duration               Deadline for a device in class C mode to respond to requests from the Network Server if not configured in device's MAC settings (default 15s)
      --ns.default-mac-settings.desired-rx1-delay int32                Desired Rx1Delay value Network Server should use if not configured in device's MAC settings (default 5)
      --ns.default-mac-settings.status-count-periodicity uint          Number of uplink messages after which a DevStatusReq MACCommand shall be sent by Network Server if not configured in device's MAC settings (default 20)
      --ns.default-mac-settings.status-time-periodicity duration       The interval after which a DevStatusReq MACCommand shall be sent by Network Server if not configured in device's MAC settings (default 1h0m0s)
      --ns.dev-addr-prefixes strings                                   Device address prefixes of this Network Server
      --ns.downlink-priorities.join-accept string                      Priority for join-accept messages (lowest, low, below_normal, normal, above_normal, high, highest) (default "highest")
      --ns.downlink-priorities.mac-commands string                     Priority for messages carrying MAC commands (lowest, low, below_normal, normal, above_normal, high, highest) (default "highest")
      --ns.downlink-priorities.max-application-downlink string         Maximum priority for application downlink messages (lowest, low, below_normal, normal, above_normal, high, highest) (default "high")
      --ns.interop.directory string                                    Retrieve the interoperability client configuration from the filesystem
      --ns.interop.url string                                          Retrieve the interoperability client configuration from a web server
      --ns.net-id string                                               NetID of this Network Server (default "000000"
```

Even though this is the `networkserver.Config`:
https://github.com/TheThingsNetwork/lorawan-stack/pull/1329/files#diff-7563ac1c9d5923e766a6d6d80982389bR41-R49

Any idea why the e.g. `ns.default-mac-settings.desired-adr-ack-limit-exponent` and other options added in this PR are not recognized? Note, that the default values are not set for there if that's anyhow related.

api/end_device.proto
```diff
@@ -69,9 +69,11 @@ message MACParameters {
   // ADR: number of retransmissions.
   uint32 adr_nb_trans = 6 [(gogoproto.customname) = "ADRNbTrans", (validate.rules).uint32.lte = 15];
   // ADR: number of messages to wait before setting ADRAckReq.
-  uint32 adr_ack_limit = 7 [(gogoproto.customname) = "ADRAckLimit", (validate.rules).uint32 = {gte: 1, lte: 32768}];
+  // This field is deprecated, use adr_ack_limit_exponent instead.
+  uint32 adr_ack_limit = 7 [(gogoproto.customname) = "ADRAckLimit", deprecated = true];
```
htdvisser: Restore the validation rules, also below.

rvolosatovs: The intention is to not store this value in the registry, since it can be derived from the `_exponent` counterpart. Adding validation here breaks flows.

rvolosatovs:
I tried suggestion from https://github.com/TheThingsNetwork/lorawan-stack/pull/1329#pullrequestreview-292868899, but that did not do anything. By looking at the code it seems that the implementation should `panic` on unknown types, so I don't see how that could help anyway. Opened an issue about this in https://github.com/TheThingsNetwork/lorawan-stack/issues/1386

Issue #1326, opened 2019-09-05
Getting gateway connection stats bumps "last seen"


johanstokking:
#### Summary

Getting gateway connection stats bumps "last seen"

#### Steps to Reproduce

1. Go to gateway
2. Watch "last seen"

#### What do you see now?

Last seen gets bumped every 5 seconds, when the gateway connection stats get retrieved.

#### What do you want to see instead?

Per https://github.com/TheThingsNetwork/lorawan-stack/issues/1242#issuecomment-524808708:

- on `gs.up.receive`; increment up and bump last seen
- on `gs.down.send`; increment down and bump last seen
- on `gs.status.receive`; bump last seen

#### Environment

3.1.2

#### Can you do this yourself and submit a Pull Request?

Can review

Issue #1155, opened 2019-08-13
Add "connected" indicator to gateway list


kschiffer:
#### Summary
The gateway entity list in the console should fetch the stats information of each gateway and display the connection status, using our status indicator component.

See also https://github.com/TheThingsNetwork/lorawan-stack/issues/875

#### Why do we need this?
Enabling a quicker overview over connection statuses of gateways, right from the gateways page.

#### What is already there? What do you see now?
We already fetch the stat info from gateways on the gateway overview. There is also already a status indicator component to display the connection status.

#### What is missing? What do you want to see?
A column in the gateway list, containing the connection status.

#### Environment
Console.

#### How do you propose to implement this?
We would need to tap the status endpoints of each gateway individually. This could make issues in larger page sizes. The exact impact needs to be assessed when implementing this.
After fetching the info, it's trivial to display it inside the table.

#### Can you do this yourself and submit a Pull Request?
Yes, but help welcome @bafonins @mjamescompton @pgalic96. Feel free to take over.

PR #1334, opened 2019-09-09
Improve gateway connection status handling in console

bafonins:
#### Summary

This PR addds various improvements to the gateway connection status UI.
References: https://github.com/TheThingsNetwork/lorawan-stack/issues/1086 and https://github.com/TheThingsNetwork/lorawan-stack/issues/1175
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1326

Screenshot 2019-09-09 at 12 14 03

If no status events/stats received:
Screenshot 2019-09-24 at 14 43 33

#### Changes

- Add general event type selectors
- Update gateway stats on `gs.status.receive`, `gs.down.send`, `gs.up.receive`
- Increase gateway stats interval from 5s to **60s**
- Add uplink/downlink number formatting
- Rename `GatewayStatistics` to `GatewayConnection`. later we can split the component into `Statistics` and `Status` that can be used for https://github.com/TheThingsNetwork/lorawan-stack/issues/1155
- Add `gatewayStatus` reducer

#### Notes for Reviewers

@johanstokking According to the demo environment, some gateways receive uplinks every ~10s, imo it would be an overkill to poll for stats within the same interval. This is why I decided to increase it to 60s. Let me know what you think.

#### Notes for Reviewers

Tested with `simulate` from the cli, however it would be great to test changes with real gateway events.


johanstokking:
> According to the demo environment, some gateways receive uplinks every ~10s, imo it would be an overkill to poll for stats within the same interval. This is why I decided to increase it to 60s. Let me know what you think.

I think 60 seconds is fine. Most of the realtime things (receive uplink and status) should be driven by events anyway, not (primarily) by polling stats.

pkg/webui/console/store/middleware/logics/gateways.js
```diff
@@ -201,6 +206,10 @@ const updateGatewayStatisticsLogic = createRequestLogic({
 
     const stats = await api.gateway.stats(id)
 
+    if (!Boolean(stats.last_status_received_at)) {
+      stats.last_status_received_at = new Date()
```
kschiffer: Is this the correct assumption? Why would this prop be omitted from the response? In the proptypes it is also set as required.

bafonins: > Why would this prop be omitted from the response?

some gateways don’t send status at all. had discussion about this with @johanstokking

kschiffer: I see. Should we set `new Date()` only for heartbeat-triggered updates then?

bafonins: not sure what you mean

kschiffer: I mean that for gateways that don't send status, we would then still bump the last-seen in a 60s interval. Yet, we only safely know the gateway was seen when we received `gs.status.receive`, `gs.down.send`, `gs.up.receive`.

bafonins: @johanstokking

kschiffer:
It's hard for me to verify in the demo environment as all gateways I have available there don't send any `last_status_received_at` prop.

Also, in these cases, if the polling interval elapses we incorrectly bump the `last seen` seen value via `stats.last_status_received_at = new Date()`, which is not desired behavior as I understand it. 

Would it be an idea to fall back to `last_uplink_received_at`?

pkg/webui/console/store/middleware/logics/gateways.js
johanstokking: "Last seen" refers specifically and only to:

- `last_{status|uplink}_received_at`
- `gs.{status|up}.receive`

It should not be updated when fetching stats on the interval. If the gateway is connected but there are no status or uplinks received, it should not be implied that we've seen traffic (this diff). Still, the gateway shows connected.

Maybe "last seen" is the wrong term as it's not just that it's connected, but more "last message received"

johanstokking:
> [...] if the polling interval elapses we incorrectly bump the `last seen` seen value via `stats.last_status_received_at = new Date()`, which is not desired behavior as I understand it.
> 
> Would it be an idea to fall back to `last_uplink_received_at`?

Sounds right. Last seen is not related to polling or the connection being there; it's purely uplink and status receive (via stats timestamps or via events).

PR #1335, opened 2019-09-09
Fix naming of metrics in AS

htdvisser:
#### Summary

This is a quickfix for the metric names in the AS.

These are monotonic counters, so they indicate a total number of X. Prometheus has the convention of communicating that through the metric name.


htdvisser:
What would that clarification look like? I don't think "Follow the best practices of the libraries that we use" is going to make any difference.

Issue #1340, opened 2019-09-09
Support absolute paths in fetch package


rvolosatovs:
#### Summary

Fetch package should support absolute paths

#### Why do we need this?

There are use cases, where absolute paths are necessary to use, e.g. for usage of Docker secrets(which are available at `/run/secrets/<name>`). This is especially important for interop client configuration.

#### What is already there? What do you see now?

No absolute path support

#### What is missing? What do you want to see?

Support absolute paths without breaking configs

#### How do you propose to implement this?

Allow for fetchers to be rooted at a path, i.e. prepend the root path to requested path.
For a fetcher with a root path `r` and requested path `p`, the resulting path is:
```go
if r == "" {
   return p
}
return join(r, p) // specific to the fetcher implementation
```

Only add the basepath in basepath fetcher if requested path is not absolute.

#### Can you do this yourself and submit a Pull Request?

yes

johanstokking:
Why is this not closed by #1341 ?

PR #1341, opened 2019-09-09
Support absolute paths in fetch package

rvolosatovs:
#### Summary

Refs https://github.com/TheThingsNetwork/lorawan-stack/issues/1340

#### Changes

- Fix (some of) inconsistencies in `fetch` package
- Fix non-Unix-like OS support in `fetch` package
- Validate the `root` URL passed to `fetch.FromHTTP`
- Support absolute paths in `fetch` package
- `fetch.basePathFetcher.File` now only prepends base path if requested path is not absolute
- Fetchers, which support a `root` path to be specified now correctly either prepend the `root` path to relative paths or replace the root in absolute paths

#### Notes for Reviewers

Basically support for absolute OS paths is via `WithBasePath`
Existing deployments should not be affected, since the only "breaking" change is in the `WithBasePath` fetcher operation, which was only used by `interop` client config so far.

@bafonins @adriansmares no need for review

### DIFF
--- a/pkg/config/shared.go
+++ b/pkg/config/shared.go
@@ -200,8 +200,8 @@ type FrequencyPlansConfig struct {
 
 // Store returns a frequencyplan.Store fwith a fetcher based on the configuration.
 // The order of precedence is Static, Directory and URL.
-// If neither Static, Directory nor a URL is set, this method returns nil.
-func (c FrequencyPlansConfig) Store() *frequencyplans.Store {
+// If neither Static, Directory nor a URL is set, this method returns nil, nil.
+func (c FrequencyPlansConfig) Store() (*frequencyplans.Store, error) {
 	var fetcher fetch.Interface
 	switch {
 	case c.Static != nil:
@@ -209,11 +209,15 @@ func (c FrequencyPlansConfig) Store() *frequencyplans.Store {
 	case c.Directory != "":
 		fetcher = fetch.FromFilesystem(c.Directory)
 	case c.URL != "":
-		fetcher = fetch.FromHTTP(c.URL, true)
+		var err error
+		fetcher, err = fetch.FromHTTP(c.URL, true)
+		if err != nil {
+			return nil, err
+		}
 	default:
-		return nil
+		return nil, nil
 	}
-	return frequencyplans.NewStore(fetcher)
+	return frequencyplans.NewStore(fetcher), nil
 }
 
 // DeviceRepositoryConfig defines the source of the device repository.
@@ -225,8 +229,8 @@ type DeviceRepositoryConfig struct {
 
 // Client instantiates a new devicerepository.Client with a fetcher based on the configuration.
 // The order of precedence is Static, Directory and URL.
-// If neither Static, Directory nor a URL is set, this method returns nil.
-func (c DeviceRepositoryConfig) Client() *devicerepository.Client {
+// If neither Static, Directory nor a URL is set, this method returns nil, nil.
+func (c DeviceRepositoryConfig) Client() (*devicerepository.Client, error) {
 	var fetcher fetch.Interface
 	switch {
 	case c.Static != nil:
@@ -234,13 +238,17 @@ func (c DeviceRepositoryConfig) Client() *devicerepository.Client {
 	case c.Directory != "":
 		fetcher = fetch.FromFilesystem(c.Directory)
 	case c.URL != "":
-		fetcher = fetch.FromHTTP(c.URL, true)
+		var err error
+		fetcher, err = fetch.FromHTTP(c.URL, true)
+		if err != nil {
+			return nil, err
+		}
 	default:
-		return nil
+		return nil, nil
 	}
 	return &devicerepository.Client{
 		Fetcher: fetcher,
-	}
+	}, nil
 }
 
 // InteropClient represents the client-side interoperability through LoRaWAN Backend Interfaces configuration.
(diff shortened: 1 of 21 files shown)


htdvisser:
After reading the description, I'm a bit concerned with this. I don't think it should be possible for the fetch package to ignore the root path configured in the backends. If I configure the FS fetcher backend with a root path of `/path/to/fetchroot`, I don't want the fetch package to be able to read things in `/run/secrets`.

I think it's fine to have some default base path that can be ignored, but a root needs to stay a root.

rvolosatovs:
Maybe I was not clear in the description or you did not understand it, please look at the description and the issue again. It is indeed impossible(and it should not be) to read from `/run/secrets` if you have a fetcher configured with a root path of `/path/to/fetchroot`. The only fetcher which allows bypassing the root modification is the one returned by `WithBasePath`, which will only add the `basepath`, if path is not absolute.
Note that this is also the reason for explicit distinction(also in variable names) between *base* and *root* paths.

pkg/fetch/fetch.go
```diff
@@ -17,18 +17,70 @@
...
+	"fmt"
+	"net/url"
+	"path"
+	"path/filepath"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+// realPath replaces path root of p by r if r != "" and returns p otherwise.
```
rvolosatovs: The functions added here are used by the fetchers.
Please look through their implementations - that should fix your concerns in https://github.com/TheThingsNetwork/lorawan-stack/pull/1341#issuecomment-529835157
The algorithm implemented is defined in https://github.com/TheThingsNetwork/lorawan-stack/issues/1340

htdvisser:
Ok. I was confused by

> Fetchers, which support a `root` path to be specified now correctly either prepend the root path to relative paths or replace the root in absolute paths

I'll take a look at the code later today

rvolosatovs:
Indeed, so with a root of `/path/to/fetchroot` and requested path of `/run/secrets` the actual("real") path read is `/path/to/fetchroot/run/secrets`.
The root of requested path(`/`) is replaced by the root of the fetcher(`/path/to/fetchroot`)
Note, that `run/secrets` would result in the same (`/path/to/fetchroot/run/secrets`) path

rvolosatovs:
To clarify the purpose of this, the exact use case is as follows:
Assume two files:
`a.yaml`
`foo/b.yaml`

Assume root path is `r`(i.e., `a.yaml` is located at `r/a.yaml`)
In `a.yaml` relative paths should be relative to `r`
In `foo/b.yaml` relative paths should be relative to `r/foo`
Absolute paths, however in all cases shall be rooted at `r`, that is the root(`/` shall be replaced by `r`)

This also makes sense for e.g. HTTP fetcher.
Assume `r == "https://example.com/test"`
Relative path `p` in `a.yaml` will trigger a `GET` to `https://example.com/test/p`
Relative path `p` in `foo/b.yaml` will trigger a `GET` to `https://example.com/test/foo/p`
Absolute path `/p` in both cases will trigger a `GET` to `https://example.com/test/p`

pkg/config/shared.go
```diff
@@ -251,19 +259,19 @@ type InteropClient struct {
...
 // Fetcher returns fetch.Interface defined by conf.
-func (conf InteropClient) Fetcher() fetch.Interface {
+func (c InteropClient) Fetcher() (fetch.Interface, error) {
 	switch {
-	case conf.Directory != "":
-		return fetch.FromFilesystem(conf.Directory)
-	case conf.URL != "":
-		return fetch.FromHTTP(conf.URL, true)
+	case c.Directory != "":
+		return fetch.WithBasePath(fetch.FromFilesystem(), c.Directory), nil
```
htdvisser: My concern is here.

Assuming `c.Directory` is `/path/to/interop`, then calling `Fetch("/run", "secrets", "foo")` takes `basepath.goL33`, and the `fsFetcher` happily returns the foo secret, because `f.root` is empty.

It's way too easy to build insecure implementations with this.

rvolosatovs: 1. Creating a fetcher with empty root path today is allowed and is affected by this. The PR does not add anything new there.
2. This exact fetcher is used to fetch sensitive data. The concerns should not be *what* it fetches, but *where* it exposes it.
3. The possible security risk is if e.g. the fetcher would be exposed to the network and take arbitrary paths from the network and either expose the contents or other details(e.g. error - whether a file exists or not), however we do not do that and if we did, any issues would be mitigated by simply setting the root path.

Overall, I don't see anything to be concerned about here. The user configures the path to read here and runs the process. There's no third party anyhow affecting the paths read and/or being able to access the contents.
For extra safety the process can be run in an isolated environment, where no unnecessary sensitive data is available. In any case, nothing like this belongs here.

htdvisser: 1. Yes, but then this is done with explicit config so people know what they're getting into.
2. Sure, but I still think that should be limited to the configured root boundary. If a directory is configured as root, that shouldn't just be ignored.
3. _"any issues would be mitigated by simply setting the root path"_  
  that is exactly what I'm asking.

Bumping this over to @johanstokking.

rvolosatovs: So the purpose of all of this is to be able to fetch arbitrary absolute paths.
That is not possible with root of fetcher being anything else than `/` or ``.
We need to be able to fetch arbitrary absolute paths - we discussed this on Slack as well. It seems that you're against that.

What other solution did you have in mind? Can you give any kind of other suggestion, which can be applied here?

pkg/fetch/basepath.go
```diff
@@ -14,15 +14,28 @@
...
 	basePath []string
 }
 
 func (f basePathFetcher) File(pathElements ...string) ([]byte, error) {
+	if len(pathElements) == 0 {
+		return nil, errFilenameNotSpecified
+	}
+
+	// NOTE: filepath.IsAbs returns true for paths starting with '/' on all supported operating systems.
+	if filepath.IsAbs(pathElements[0]) {
```
rvolosatovs: See testcases below for the use case

rvolosatovs:
2019-09-18-21:46:13-screenshot
@johanstokking

Issue #1313, opened 2019-09-04
CUPS update-info returning 403


mattdean-digicatapult:
#### Summary
Submitted as a result of a slack conversation. I've been trying to configure A Kerlink Wirnet Station using `basicstation` to connect to a v3 stack. When the station tries to call `update-info` it errors with 403.

#### Steps to Reproduce

1. Add a new gateway
2. Add a gateway api-key with rights `RIGHT_GATEWAY_ALL`
3. Update the gateway setting `gateway_server_address`, and the attributes `cups-credentials`, `lns-credentials` and `cups-uri`. Not setting these either leads to a different cause of the 403 on the backend or a 500 error.
4. Configure `basicstation` on the gateway setting the authorization header in `cups-boot.key`, the stack uri in `cups-boot.uri` and the stack cert in `cups-boot.trust`
5. Start `basicstation`

#### What do you see now?
The backend logs for the request (these repeat as the gateway retries):
```
INFO Finding gateway... gateway_eui=024B0BFFFE030465 method=POST namespace=web remote_addr=10.15.18.77:13480 request_id=01DKYFYC486KNB1N5QG2KAZT5A url=/update-info
DEBUG Run database query duration=3ms grpc_method=GetIdentifiersForEUI grpc_service=ttn.lorawan.v3.GatewayRegistry namespace=db query=SELECT id, created_at, updated_at, gateway_id, gateway_eui FROM "gateways" WHERE "gateways"."deleted_at" IS NULL AND ((gateway_eui = $1)) ORDER BY "gateways"."id" ASC LIMIT 1 request_id=01DKYFYC49S6RVY3TYXY5R4TSZ rows=1 source=gateway_store.go:117 values=[024B0BFFFE030465]
INFO Finished unary call duration=5.640867ms grpc_method=GetIdentifiersForEUI grpc_service=ttn.lorawan.v3.GatewayRegistry namespace=grpc request_id=01DKYFYC49S6RVY3TYXY5R4TSZ
DEBUG Getting gateway with request credentials gateway_eui=024B0BFFFE030465 gateway_id=gtw-0465 method=POST namespace=web remote_addr=10.15.18.77:13480 request_id=01DKYFYC486KNB1N5QG2KAZT5A url=/update-info
[...]

PR #1344, opened 2019-09-10
gcs: Fix CUPS update-info returning 403

mattdean-digicatapult:
#### Summary
Demonstrates the plausible cause of #1313. It seems the internal RPC calls don't actually ever use the credentials passed to the `update-info` call event when `RegisterUnknown` is not desired. The code structure makes this look like a mistake but I could be wrong.

#### Changes
Changed the cups `server.go` to use the provided credentials for internal RPC calls in the case where `RegisterUnknown.APIKey` is not set.

#### Notes for Reviewers
This is intentionally a bit of a hack just to demonstrate what might be the issue. Reviewers should feel free to discard this if not useful. Equally if I am on the right track I can spruce things up a bit and make it ready for review.

### DIFF
--- a/pkg/basicstation/cups/server.go
+++ b/pkg/basicstation/cups/server.go
@@ -19,6 +19,7 @@ import (
 	"crypto"
 	"crypto/tls"
 	"crypto/x509"
+	"fmt"
 	"net"
 	"net/url"
 	"strings"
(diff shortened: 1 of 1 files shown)


mattdean-digicatapult:
Thanks @KrishnaIyer, I'd forgotten about that. The conversation has moved on a little bit on this so you should probably catch up on the original issue #1313

Issue #857, opened 2019-06-26
Collaborators can revoke all rights from other collaborators, regardless of their own rights


kschiffer:
#### Summary
A collaborator that has only the `RIGHT_{entity}_SETTINGS_COLLABORATORS` right can revoke all rights from any other collaborator of the same entity. This should, to my understanding, be limited to only the rights that the collaborator has himself. 
It is already not possible to enable a right for another collaborator that is not granted to oneself. However, revoking such rights is still possible.

#### Steps to Reproduce
1. Create a user with just the `RIGHT_{entity}_SETTINGS_COLLABORATORS` right
2. Create another user with `RIGHT_{entity}_ALL` within the same entity
3. As the first user, call the `SetCollaborator` rpc for the second user with an empty array as set of rights
4. The collaborator which had all rights is removed as a collaborator (by a user with fewer rights)

#### What do you see now?
It might be that it is indeed intended for a user with `RIGHT_{entity}_SETTINGS_COLLABORATORS` to be able to modify any right of another collaborator. However, at the moment out-of-own-scope rights can only be removed and not set, which is a weird inconsistency.

#### What do you want to see instead?
Either: restriction to modify only rights that the collaborator has himself
Or: possibility to fully modify all rights, meaning enabling and disabling them

#### How do you propose to implement this?
I think I need some input as to what the actual strategy is for these cases.

#### Can you do this yourself and submit a Pull Request?
Needs clarification.

htdvisser:
This is how we designed it. The `RIGHT_{entity}_SETTINGS_COLLABORATORS` gives someone the right to view and edit collaborators of an entity, with one constraint: it is not possible to assign someone else more rights than you have yourself.

I can't think of any realistic scenario where a collaborator would have the right to manage collaborators, but no other rights, so I don't see this edge case being a problem.

I also think it's too late to overhaul the rights even if we wanted to.

johanstokking:
>This is how we designed it.

Indeed.

I don't see this as an issue.

PR #1345, opened 2019-09-10
Fix rights modification of collaborators with more rights than…

### LINKED ISSUE #1204: Fix rights modification of collaborators with more rights than current user
#### Summary
It is currently not possible to properly manage the rights of a collaborator that has more rights than oneself. The problem here is that if User A has rights that are outside of the scope of granted rights for User B, User B cannot modify the rights of User A properly because:
- Sending the out-of-scope rights together with the legitimately altered rights will result in a 403 Forbidden
- Not sending the out-of-scope together with the legitimately altered rights will not result in a 403 but will revoke these rights, even when that might not be desired

#### Steps to Reproduce

- Call the `SetCollaborator` RPC on an entity that has more rights than the current user
- See a 403 if you pass the out-of-scope rights in the `rights` prop (legitimately attempting to leave them enabled)
- See a 200 if you do not pass them, but the rights are now revoked

#### What do you see now?
Inability to set rights of collaborators correctly if they have more rights than oneself.

Being able to selectively revoke rights that the current user does not have without having to revoke them altogether. Likewise, being able to leave these rights enabled while modifying other rights.

#### Environment
Irrelevant.

No. Assigned @htdvisser but feel free to reassign to whoever is more appropriate.

See also #939 and https://github.com/TheThingsNetwork/lorawan-stack/issues/939#issuecomment-522966655

bafonins:
#### Summary

Closes #1204 

#### Changes

- Change implementation of `SetApplicationCollaboratorRequest`
- Change implementation of `SetCollaboratorCollaboratorRequest`
- Change implementation of `SetGatewayCollaboratorRequest`
- Add tests for each entity to test new handling of collaborator rights

#### Notes for Reviewers 

~~@htdvisser I left the clients access rpc without changes, since it seems to be inherently different from applications, gateways and collaborators.~~

@johanstokking @rvolosatovs no need for review
@kschiffer please verify that this PR solves the issue

#### Release Notes

- Changed the way collaborator rights are handled in the `SetCollaborator` rpc for Applications, Gateways and Organizations. Collaborators can revoke or add rights to other collaborators as long as they have these rights.

### DIFF
--- a/api/api.md
+++ b/api/api.md
@@ -630,7 +630,7 @@ where the user or organization is collaborator on.
 | `GetAPIKey` | [`GetApplicationAPIKeyRequest`](#ttn.lorawan.v3.GetApplicationAPIKeyRequest) | [`APIKey`](#ttn.lorawan.v3.APIKey) |  |
 | `UpdateAPIKey` | [`UpdateApplicationAPIKeyRequest`](#ttn.lorawan.v3.UpdateApplicationAPIKeyRequest) | [`APIKey`](#ttn.lorawan.v3.APIKey) | Update the rights of an existing application API key. To generate an API key, the CreateAPIKey should be used. To delete an API key, update it with zero rights. |
 | `GetCollaborator` | [`GetApplicationCollaboratorRequest`](#ttn.lorawan.v3.GetApplicationCollaboratorRequest) | [`GetCollaboratorResponse`](#ttn.lorawan.v3.GetCollaboratorResponse) | Get the rights of a collaborator (member) of the application. Pseudo-rights in the response (such as the "_ALL" right) are not expanded. |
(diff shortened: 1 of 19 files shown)


kschiffer:
This fixes #1204 indeed. However, it was formerly possible to revoke (universal) rights of a collaborator even when one would not have them oneself. This was briefly problematized [here](https://github.com/TheThingsNetwork/lorawan-stack/issues/857) and concluded as intended behavior. 
Should we care to keep this behavior? cc @htdvisser

htdvisser:
There is no such thing as assigning/revoking "universal" rights.

The `*_SETTINGS_API_KEYS` and `*_SETTINGS_COLLABORATORS` rights mean that a user can create and delete API keys and collaborators, and can add/remove rights from these. It is never possible to give an API key or collaborator more rights than you have yourself. For consistency we now also don't allow taking away individual rights that you don't have yourself (but you can still completely delete an API key or collaborator).

PR #1355, opened 2019-09-13
Give GCS a ClusterRole

### LINKED ISSUE #1278: Define Gateway Configuration Server ClusterRole
#### Summary

We need to define the GCS as a ClusterRole enum value.

#### Why do we need this?

For consistency

https://github.com/TheThingsNetwork/lorawan-stack/blob/master/api/enums.proto#L50

The GCS is currently announcing itself as GS.

The GCS should have its own ClusterRole.

Add it to the enum, and change throughout codebase where needed.

Yes

htdvisser:
#### Summary

This Pull Request gives the GCS its own ClusterRole.

Closes #1278

#### Changes

- Added GCS ClusterRole to API
- Made GCS announce that ClusterRole to the cluster

#### Notes for Reviewers

Since CLI isn't talking to GCS, I didn't think it would make sense to add it there.

#### Release Notes

- Upgraded Gateway Configuration Server to a first-class cluster role.


pkg/gatewayconfigurationserver/gatewayconfigurationserver_test.go
```diff
@@ -46,6 +46,19 @@ var (
...
+	ctx := log.NewContext(test.Context(), test.GetLogger(t))
+
+	conf := &component.Config{}
+	c := component.MustNew(test.GetLogger(t), conf)
+
+	test.Must(New(c, testConfig))
+	test.Must(c.Start(), nil)
+	defer c.Close()
+
+	mustHavePeer(ctx, c, ttnpb.ClusterRole_GATEWAY_CONFIGURATION_SERVER)
```
rvolosatovs: This test function contains no assertions and only panics on errors.
Would be nicer/cleaner/easier to debug if there were actual assertions here.

htdvisser: This was just copied from `TestDeviceTemplateConverter`, and I guess it's done this way in more places, so if this is something you think is important, please file a technical debt issue.

Issue #1328, opened 2019-09-06
Support LoRa Cloud Device Management


johanstokking:
#### Summary

Support LoRa Cloud Device Management (and plumbing for application packages now that we need it)

#### Why do we need this?

To natively support the Semtech modems

#### What is already there? What do you see now?

No support, only some ideas about supporting application-layer services (called application packages) based on FPort. These ideas have not been written down, so I'll do that in this issue.

##### Supporting application packages

The Application Server should get a pluggable and dynamic table of "application packages" per `FPort`. These packages provide functionality on the end device and on the server.

>Example application packages are time synchronization, remote multicast setup, data fragmentation, etc.

Application packages are:
1. MAC layer independent. However, they need to be aware of the maximum (downlink) payload size
   - @rvolosatovs please comment on how to communicate this between NS-AS. My requirement here is that AS is not going to compute that based on DR and PHY; NS should do that. My proposal is to carry this in `ApplicationUp` to keep things streaming
2. Region independent
3. Versioned (as in, each version is essentially a separate implementation)
4. For application traffic, independent of other packages
5. Stateful. They need one (abstract) way to store state. I would suggest a common registry for application packages to store binary data (marshaled from end device state protos per package)
6. Be able to publish messages, triggering the AS to publish a new type of `ApplicationUp.Up` message with "the payload" (whatever that may be) of an application package. We need to keep this very generic, so a proto `Struct` would suffice
7. Be able to schedule downlink messages
8. In the future, be able to expose gRPC services

See Application Layer Package Design Guidelines (TR003) [1]

Packages are registered on an `FPort`; there can only be one package implementation per port. This registration is dynamic and can fluctuate per device during its lifetime by user configuration (through device registry) or through other packages.

There is a special package ("Multi-package access protocol" [2]) that is itself fixed on `FPort` 225, but can discover packages and register them for the concerning end device, i.e. configure "remote multicast setup V2" on `FPort` 223.

So, each `EndDevice` gets a mapping of `FPort` to some versioned package identifier, that is stored in Application Server.

In the future, there's a default mapping per end device coming from the Device Repository. References #263 

- [x] Plumbing for application packages with registration by (versioned) identifier, interface, storage in API, CLI support for management

Resources:
[1] Application Layer Package Design Guidelines (TR003). LoRa Alliance confidential, see linked issue
[2] Multi-package access protocol. LoRa Alliance confidential, see linked issue

##### LoRa Cloud Device Management support

[...]

PR #1358, opened 2019-09-13
Allow application frontends to send upstream traffic

adriansmares:
#### Summary

References #1328

#### Changes

- Add `link.sendUp`, `ApplicationServer.SendUp` and added `SendUp` to `io.Server`

#### Notes for Reviewers

Implemented as per https://github.com/TheThingsNetwork/lorawan-stack/issues/1328#issuecomment-528786379

### DIFF
--- a/pkg/applicationserver/io/io.go
+++ b/pkg/applicationserver/io/io.go
@@ -25,13 +25,15 @@ import (
 
 const bufferSize = 32
 
-// Server represents the Application Server to gateway frontends.
+// Server represents the Application Server to application frontends.
 type Server interface {
 	// GetBaseConfig returns the component configuration.
 	GetBaseConfig(ctx context.Context) config.ServiceBase
 	// FillContext fills the given context.
 	// This method should only be used for request contexts.
 	FillContext(ctx context.Context) context.Context
+	// SendUp sends upstream traffic to the Application Server.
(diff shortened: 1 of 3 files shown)


pkg/applicationserver/io/io.go
```diff
@@ -25,13 +25,16 @@ import (
...
 
-// Server represents the Application Server to gateway frontends.
+// Server represents the Application Server to application frontends.
 type Server interface {
 	// GetBaseConfig returns the component configuration.
 	GetBaseConfig(ctx context.Context) config.ServiceBase
 	// FillContext fills the given context.
 	// This method should only be used for request contexts.
 	FillContext(ctx context.Context) context.Context
+	// SendUp sends a synthetic uplink to the Application Server, as if it came from the Network Server.
```
johanstokking: It's not uplink, it's upstream traffic. The NS part in this comment can be removed.

AS is entitled to independently of NS generate upstream traffic. Uplink is indeed from devices through NS.

pkg/applicationserver/linking.go
```diff
@@ -95,6 +95,8 @@ type link struct {
 	connReady chan struct{}
 	callOpts  []grpc.CallOption
 
+	*ApplicationServer
```
johanstokking: Please do this through an behavioral interface to avoid tight coupling

Issue #1310, opened 2019-09-04
Rewrite confusing hex numbers in binary format


rvolosatovs:
#### Summary

Go 1.13 adds support for binary numbers, which can be used to improve clarity of LoRaWAN-related functionality.

#### Why do we need this?

To improve clarity and readability

#### What is already there? What do you see now?

Usage of hexadecimal constants

#### What is missing? What do you want to see?

Binary numbers where applicable (e.g. `MHDR`, `FCtrl`, MAC command payloads etc.)

#### How do you propose to implement this?

Replace hex numbers by binary where applicable

#### Can you do this yourself and submit a Pull Request?

yes

PR #1360, opened 2019-09-13
Fix downlink length computation in Network Server

rvolosatovs:
#### Summary

In https://github.com/TheThingsNetwork/lorawan-stack/pull/1287 application downlink validation was moved before MAC buffer adding to downlink, which is incorrect, since NS shall first compute the MAC buffer in order to determine if application downlink fits or not.

Refs #1310 

#### Changes

- Compute MAC buffer before validating application downlink
- Do not fail downlink, which does not fit due to MAC commands.
- Only update `queued_application_downlinks` if actually changed
- Use binary number syntax in NS where applicable
- Account for FHDR length when computing `FRMPayload` length limit (can't believe this went unnoticed for so long)

#### Notes for reviewers

https://github.com/TheThingsNetwork/lorawan-stack/pull/1360/commits/f2569005ba1b6274732b29cf9d21d08067ddeb27 is just moving stuff around

### DIFF
--- a/pkg/networkserver/downlink.go
+++ b/pkg/networkserver/downlink.go
@@ -137,11 +137,11 @@ func (ns *NetworkServer) generateDownlink(ctx context.Context, dev *ttnpb.EndDev
 	))
 	logger := log.FromContext(ctx)
 
-	// NOTE: len(MHDR) + len(MIC) = 1 + 4 = 5
-	if maxDownLen < 5 || maxUpLen < 5 {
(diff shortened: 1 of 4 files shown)


rvolosatovs:
2019-09-16-11:44:19-screenshot

rvolosatovs:
Added testing and some related fixes/improvements

rvolosatovs:
@johanstokking

Issue #1281, opened 2019-09-02
Manage application PubSub integrations in Console


adriansmares:
#### Summary

Manage application PubSub integrations in Console. References #935.

#### Why do we need this?

PubSub integrations are a good way to work with application data. They provide more scalability than vanilla MQTT and allow users to use their favorite PubSub provider to handle the data.

#### What is already there? What do you see now?

AS supports PubSub integrations natively using 2 providers: NATS and MQTT. API for managing these integrations is available and CLI support is already provided. Also they are very much like webhooks (except for the service specific settings).

#### What is missing? What do you want to see?

Console support.

#### Environment

3674744, but this is not relevant.

#### How do you propose to implement this?

Per application, be able to manage the PubSub integrations. This is managed through the CRUD API exposed in [ApplicationPubSubRegistry](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/api/api.md#service-applicationpubsubregistry).

Per integration, to be able to manage the PubSub (see [ApplicationPubSub](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/api/api.md#message-applicationpubsub)). The following fields are relevant:
- PubSub ID, a user defined ID that uses the same rules for all our other IDs. (same as webhooks)
- Format, get options through the get-formats endpoint, see API (same as webhooks).
- The base path, which is the base part of the topic (equivalent to the webhooks base URL, but provider specific - I think it should be provided as a string only).
- The provider, which controls the endpoint of the integration:
    - NATS provider, see API first, then read my comment [here](https://github.com/TheThingsNetwork/lorawan-stack/issues/935#issuecomment-508747292) which explains one way to show the NATS settings to the users.
     - MQTT provider, see API first, for which every field should be provided. The QoS fields should be a dropdown of `0`, `1`, and `2` and their descriptions should appear while holding the mouse over. The TLS settings should be hidden behind a checkbox, and they require 3 files to be provided as binary (the CA, the public key and the private key). 
- The paths, which again are only strings and follow the same structure as webhooks: 
    - Uplink messages
    - Join-accept messages
    - Downlink queued
    - Downlink sent
    - Downlink failed
    - Downlink acknowledged
    - Downlink not acknowledged
    - Location solved

#### Can you do this yourself and submit a Pull Request?

Can review.

Issue #935, opened 2019-07-04
Restructure Integrations


htdvisser:
#### Summary

The integrations should be restructured in the sidebar

#### What is already there? What do you see now?

<img width="258" alt="Screenshot 2019-07-04 at 16 06 56" src="https://user-images.githubusercontent.com/181308/60672496-c7aee180-9e75-11e9-8a9a-3f450c90ecd0.png">

#### What is missing? What do you want to see?

- Integrations
  - MQTT overview page with basic connection info as suggested by @wienke (probably requires #138 or configuration)
  - Webhooks (what we have now)
  - Pubsub (to be built into #883)

adriansmares:
#883 has been merged.

PR #1361, opened 2019-09-17
Refactor data fetching selectors

kschiffer:
#### Summary
This PR will reduce code redundancy of selector components that fetch their own options. This will help us implementing these sort of components quicker in the future.

#### Changes
- Introduce the `CreateFetchSelect` function that will take store selectors and fetch action and return a Select/Field component taking care of fetching, formatting and displaying the options provided from the store.
- Refactor the frequency plan and webhook format selector container components to use the new logic

#### Notes for Reviewers
I did this while implementing #935 and #1281 which will necessitate building another selector component that fetches the list of PubSub formats. So this is an extraction from there, to reduce the diffs.


pkg/webui/console/containers/freq-plans-select/index.js
```diff
@@ -32,78 +26,22 @@ import {
...
-
-FrequencyPlansSelect.defaultProps = {
-  title: sharedMessages.frequencyPlan,
-  autoFocus: false,
-  menuPlacement: 'auto',
-  required: false,
-}
-
-export default FrequencyPlansSelect
+const CreateFrequencyPlansSelector = ns =>
```
bafonins:
```suggestion
const CreateFrequencyPlansSelect = source =>
```

pkg/webui/console/containers/fetch-select/index.js
```diff
@@ -0,0 +1,97 @@
...
+const formatOptions = formats =>
+  Object.keys(formats).map(key => ({ value: key, label: formats[key] }))
+
+export default function({
+  optionsSelector,
+  errorSelector,
+  fetchingSelector,
+  fetchOptions,
+  warningMessage,
+  titleMessage,
```
bafonins: With this implementation, there is no way to enforce `titleMessage` to be present, except for checking it in the ui. Would be nice to add prop-types validation here to see errors.

pkg/webui/console/containers/fetch-select/index.js
```diff
@@ -0,0 +1,97 @@
...
+  )
+  class FetchSelect extends React.PureComponent {
+    static propTypes = {
+      autoFocus: PropTypes.bool,
+      error: PropTypes.error,
+      fetchOptions: PropTypes.func.isRequired,
+      fetching: PropTypes.bool,
+      menuPlacement: PropTypes.oneOf(['top', 'bottom', 'auto']),
+      name: PropTypes.string.isRequired,
+      options: PropTypes.oneOfType([PropTypes.shape({}), PropTypes.arrayOf(PropTypes.shape({}))]),
```
bafonins: I believe thats the shape of options:
```suggestion
      options: PropTypes.arrayOf(PropTypes.shape({value: PropTypes.string, label: PropTypes.message})),
```

pkg/webui/console/containers/fetch-select/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+    componentDidMount() {
+      const { fetchOptions } = this.props
+      fetchOptions()
+    }
+
+    getOptions() {
+      const { options } = this.props
+
+      return optionsFormatter(options)
```
bafonins: Any reason for doing this? The `options` prop is passed from the parent, so why not enforcing the parent component do transform `options` into the shape needed by the select component?

pkg/webui/console/components/gateway-data-form/index.js
```diff
@@ -105,13 +105,7 @@ class GatewayDataForm extends React.Component {
...
         />
         <Message component="h4" content={sharedMessages.lorawanOptions} />
-        <FrequencyPlansSelect
-          horizontal
-          source="gs"
-          name="frequency_plan_id"
-          menuPlacement="top"
-          required
-        />
+        <GsFrequencyPlansSelect horizontal name="frequency_plan_id" menuPlacement="top" required />
```
bafonins: you can remove `horizontal`
```suggestion
        <GsFrequencyPlansSelect name="frequency_plan_id" menuPlacement="top" required />
```

pkg/webui/console/containers/fetch-select/index.js
kschiffer: Not sure if I get your concern. The options are coming from the store in an arbitrary format. The formatter is needed to bring it into a form that the select component understands, and is passed in from the creator function.

pkg/webui/console/containers/fetch-select/index.js
kschiffer: I don't know if the prop-types pattern should be applied in this particular case because the component is built from a generator function, where I simply have to assume the arguments have been passed in correctly (like we do in other places too).
What can already help is to set the `title` prop of the field component as required.
Let me know if you have any other idea in mind.

pkg/webui/console/containers/fetch-select/index.js
kschiffer: Ok, thanks for clarifying. I pass the options now in the correct format directly. Also fixed the prop-type accordingly.

PR #1366, opened 2019-09-18
Support unmarshaling of SX1301 Conf Messages

### LINKED ISSUE #808: Support Unmarshalling of SX1301 Conf
#### Summary

Support Unmarshalling of SX1301 Conf

#### Why do we need this?

To un-marshal SX1301 Configuration when used stand-alone or as part of packet-forwarder/router config.

Not Implemented `errUnmarshalNotImplemented `

Implement it.

#### Environment

`ttn-lw-stack v3.03`

- Multi-stage un-marshaling due to the sub-messages.

Yup

KrishnaIyer:
#### Summary

Closes #808 

#### Changes

- Implement unmarshaling.
- Add corresponding test to Basic Station.

#### Notes for Reviewers

- As discussed in the issue, this is only to support tests.

#### Release Notes

- Added unmarshaling support to SX1301Conf.

### DIFF
--- a/pkg/pfconfig/shared/shared.go
+++ b/pkg/pfconfig/shared/shared.go
@@ -146,6 +146,8 @@ func (c *SX1301Config) UnmarshalJSON(msg []byte) error {
 			var index int
 			if _, err := fmt.Sscanf(key, "chan_multiSF_%d", &index); err == nil {
 				chanMap[index] = channel
+			} else {
+				return err
 			}
 		case strings.HasPrefix(key, "tx_lut_"):
 			var txLut TxLUTConfig
@@ -155,6 +157,8 @@ func (c *SX1301Config) UnmarshalJSON(msg []byte) error {
 			var index int
 			if _, err := fmt.Sscanf(key, "tx_lut_%d", &index); err == nil {
 				txLutMap[index] = txLut
+			} else {
+				return err
 			}
 		case strings.HasPrefix(key, "radio_"):
 			var radio RFConfig
@@ -164,6 +168,8 @@ func (c *SX1301Config) UnmarshalJSON(msg []byte) error {
 			var index int
 			if _, err := fmt.Sscanf(key, "radio_%d", &index); err == nil {
 				radioMap[index] = radio
+			} else {
+				return err
 			}
 		}
 	}
--- a/pkg/pfconfig/shared/shared.go
+++ b/pkg/pfconfig/shared/shared.go
@@ -19,6 +19,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"regexp"
+	"strconv"
 	"time"
 
 	"go.thethings.network/lorawan-stack/pkg/band"
@@ -27,7 +29,7 @@ import (
 )
 
 // SX1301Config contains the configuration for the SX1301 concentrator.
-// It marshals to compliant JSON, but does currently **not** unmarshal from it.
+// It marshals to and unmarshals from compliant JSON.
 type SX1301Config struct {
 	LoRaWANPublic       bool
 	ClockSource         uint8
@@ -40,10 +42,7 @@ type SX1301Config struct {
 	TxLUTConfigs        []TxLUTConfig
 }
 
-var errUnmarshalNotImplemented = errors.DefineUnimplemented(
-	"unmarshal_not_implemented",
-	"unmarshaling SX1301 config is not implemented",
-)
+var errInvalidKey = errors.DefineInvalidArgument("invalid_key", "key `{key}` invalid")
 
 type kv struct {
 	key   string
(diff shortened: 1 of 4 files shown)


pkg/pfconfig/shared/shared.go
```diff
@@ -108,10 +107,90 @@ func (c SX1301Config) MarshalJSON() ([]byte, error) {
 	return json.Marshal(m)
 }
 
-// UnmarshalJSON implements json.Unmarshaler. It currently just errors because
-// unmarshaling is not supported.
-func (c *SX1301Config) UnmarshalJSON(_ []byte) error {
-	return errUnmarshalNotImplemented
+var (
+	radioKey = regexp.MustCompile("^radio_([0|1])$")
```
htdvisser: I don't think you really meant `([0|1])` here. Likely `([01])` or `(0|1)`.

pkg/pfconfig/shared/shared.go
```diff
@@ -108,10 +107,90 @@ func (c SX1301Config) MarshalJSON() ([]byte, error) {
...
+	chanKey  = regexp.MustCompile("^chan_multiSF_([0-9]{1,2})$")
+)
+
+// UnmarshalJSON implements json.Unmarshaler.
+func (c *SX1301Config) UnmarshalJSON(msg []byte) error {
+	var root map[string]json.RawMessage
+	if err := json.Unmarshal(msg, &root); err != nil {
+		return err
+	}
+	radioMap, txLutMap, chanMap := make(map[int]RFConfig), make(map[int]TxLUTConfig), make(map[int]IFConfig)
```
htdvisser: Nit: don't really like this style

pkg/pfconfig/shared/shared.go
```diff
@@ -108,10 +107,90 @@ func (c SX1301Config) MarshalJSON() ([]byte, error) {
...
+	for key, value := range root {
+		if key == "lorawan_public" {
```
htdvisser: Better to use a `switch` statement here, no need to do all comparisons for every key.

```go
switch {
	case key == "lorawan_public":
	// ...
}
```

pkg/pfconfig/shared/shared.go
```diff
@@ -108,10 +107,90 @@ func (c SX1301Config) MarshalJSON() ([]byte, error) {
...
+			if err := json.Unmarshal(value, &c.LoRaStandardChannel); err != nil {
+				return err
+			}
+		}
+		if key == "chan_FSK" {
+			if err := json.Unmarshal(value, &c.FSKChannel); err != nil {
+				return err
+			}
+		}
+		if match := radioKey.FindStringSubmatch(key); len(match) == 2 {
```
htdvisser: Nit: Using a `strings.HasPrefix` and `fmt.Scanf` is probably more efficient than matching 3 regexes in order to get to the most common one (the IF config). Could also choose to re-order.

pkg/pfconfig/shared/shared.go
```diff
@@ -108,10 +107,90 @@ func (c SX1301Config) MarshalJSON() ([]byte, error) {
...
+			if err := json.Unmarshal(value, &channel); err != nil {
+				return err
+			}
+			if index, err := strconv.Atoi(match[1]); err == nil {
+				chanMap[index] = channel
+			}
+		}
+	}
+
+	c.Radios, c.TxLUTConfigs, c.Channels = make([]RFConfig, len(radioMap)), make([]TxLUTConfig, len(txLutMap)), make([]IFConfig, len(chanMap))
```
htdvisser: Nit: don't really like this style

pkg/pfconfig/shared/shared.go
KrishnaIyer: Damn this is super cool. I don't need those regexes. Also the only reason for using the `if` was the regex match.
Thanks!

pkg/pfconfig/shared/shared.go
```diff
@@ -108,10 +105,80 @@ func (c SX1301Config) MarshalJSON() ([]byte, error) {
...
+			if err := json.Unmarshal(value, &c.FSKChannel); err != nil {
+				return err
+			}
+		case strings.HasPrefix(key, "chan_multiSF_"):
+			var channel IFConfig
+			if err := json.Unmarshal(value, &channel); err != nil {
+				return err
+			}
+			var index int
+			if _, err := fmt.Sscanf(key, "chan_multiSF_%d", &index); err == nil {
```
htdvisser: We could be a bit more strict and return this `err` if it's non-nil

KrishnaIyer: ok done. Re-requesting review since it won't let me merge. 🙏

KrishnaIyer:
@johanstokking: Seems like you're the code owner for this and requires your approval.

PR #1371, opened 2019-09-20
Send Basic Station version info upstream

### LINKED ISSUE #1370: Forward Basic Station version info upstream
#### Summary

Forward fields from the Basic Station `Version` message upstream.

#### Why do we need this?

To monitor versions of gateways.

`Version` message locally handled.

Forward `model` and `station` fields of the version message.

#### Environment

`ttn-lw-stack v3.1.3`, Basic Station.

Send a Status message upstream upon every `version` request.

Yup

KrishnaIyer:
#### Summary

Closes #1370 

#### Changes

- Send a status message with version info on every connect.
- Remove `SupportsStatusMessage` since this PR makes it obsolete.

#### Notes for Reviewers

- Targeted the PR to `feature/gs-upstreams` (#1198)

#### Release Notes

- Added  Basic Station version information to status message.

### DIFF
--- a/pkg/gatewayserver/io/basicstationlns/basicstationlns.go
+++ b/pkg/gatewayserver/io/basicstationlns/basicstationlns.go
@@ -25,6 +25,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	pbtypes "github.com/gogo/protobuf/types"
 	"github.com/gorilla/websocket"
 	echo "github.com/labstack/echo/v4"
 	echomiddleware "github.com/labstack/echo/v4/middleware"
@@ -61,7 +62,6 @@ type srv struct {
 }
 
 func (*srv) Protocol() string            { return "basicstation" }
-func (*srv) SupportsStatusMessage() bool { return false }
 func (*srv) SupportsDownlinkClaim() bool { return false }
 
 // New creates the Basic Station front end.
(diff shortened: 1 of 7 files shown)


pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -324,6 +323,15 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 					logger.WithError(err).Warn("Failed to send router configuration")
 					return err
 				}
+				stat := &ttnpb.GatewayStatus{
+					Time: receivedAt,
+					Versions: map[string]string{
+						"platform": fmt.Sprintf("%s:%s", version.Model, version.Station),
```
htdvisser: This wasn't really the intention of the `versions` map (see API definition).

I couldn't find any documentation on the exact spec of these fields in the LNS documentation, but the CUPS documentation says:

> The `station`, `model`, and `package` fields describe the current version of Station, the gateway model, and the version of the package. The package version may contain information about the state of the underlying system and its configuration.

So `station` and `package` are version for sure, but `model` isn't. I suppose it would therefore make sense to take the `station` and `package` fields (and I guess also `firmware`) from the `version` message, and put those in the `Versions` map. We could put the `model` and `features` into the `Advanced` struct.

KrishnaIyer:
```
  // - for example:
  //     firmware: "2.0.4"
  //     forwarder: "v2-3.3.1"
  //     fpga: "48"
  //     dsp: "27"
  //     hal: "v2-3.5.0"
```

The API doc only provides examples and isn't explicit about the keys so I chose my own. 
I've made the changes.

PR #1372, opened 2019-09-20
Expose MQTT public address and username

### LINKED ISSUE #1363: Expose AS endpoints through rpc
#### Summary
We need to expose the endpoints of the AS via RPC to be able to provide them to the user. 

#### Why do we need this?
E.g. to display MQTT connection information to the user, especially in the console. The endpoint info is configurable, so it cannot be derived.

The configuration service of the API.

A way to get endpoint information from the API.

Should not be a big issue, but I am not an AS expert.

@johanstokking / @adriansmares 

Currently blocking #935.

adriansmares:
#### Summary

Closes #1363

#### Changes

- Added a service that exposes the MQTT public listening addresses and computes the username of a given application.

#### Notes for Reviewers

@rvolosatovs no need to review. @htdvisser please check if the API looks fine.
@kschiffer you can test this by using the API described [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/feature/1363-mqtt-address/api/api.md#service-asmqttconfiguration) and setting `--as.mqtt.public` or `--as.mqtt.public-tls`.

I've moved `MQTTConfig` from `applicationserver` to `applicationserver/io/mqtt` in order to avoid an import cycle and to keep the MQTT related functionality inside that package.


api/applicationserver.proto
```diff
@@ -151,3 +151,32 @@ service AsEndDeviceRegistry {
   };
 }
 
+// MQTTConfig represents the configuration of the Application Server MQTT frontend.
+message MQTTConfig {
```
htdvisser: I think it's better to (1) include `ApplicationServer` in here or (2) remove the comment about "Application Server". We'll likely add a similar endpoint to the Gateway Server too.

api/applicationserver.proto
```diff
@@ -151,3 +151,32 @@ service AsEndDeviceRegistry {
...
+  // The public listen address of the frontend.
+  string public = 1 [(validate.rules).string.uri = true];
```
htdvisser: No need to validate response messages, but ok.

api/applicationserver.proto
```diff
@@ -151,3 +151,32 @@ service AsEndDeviceRegistry {
...
+  // The public listen address of the TLS frontend.
+  string public_tls = 2 [(gogoproto.customname) = "PublicTLS", (validate.rules).string.uri = true];
+}
+
+// MQTTUsername represents the username that an application can use to connect to the MQTT frontend.
+message MQTTUsername {
```
htdvisser: Same here, we'll probably re-use this on the gateway side.

api/applicationserver.proto
```diff
@@ -151,3 +151,32 @@ service AsEndDeviceRegistry {
...
+  string username = 1 [(validate.rules).string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$" , max_len: 36}];
+}
+
+// AsMQTTConfiguration provides the MQTT configuration of the Application Server.
+service AsMQTTConfiguration {
+  // GetConfig returns the configuration for a given application.
+  rpc GetConfig(ApplicationIdentifiers) returns (MQTTConfig) {
```
htdvisser: Should we perhaps merge `GetConfig` and `GetUsername` into one?

api/applicationserver.proto
```diff
@@ -151,3 +151,32 @@ service AsEndDeviceRegistry {
...
+  string public_tls = 2 [(gogoproto.customname) = "PublicTLS", (validate.rules).string.uri = true];
+}
+
+// MQTTUsername represents the username that an application can use to connect to the MQTT frontend.
+message MQTTUsername {
+  string username = 1 [(validate.rules).string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$" , max_len: 36}];
+}
+
+// AsMQTTConfiguration provides the MQTT configuration of the Application Server.
+service AsMQTTConfiguration {
```
htdvisser: These RPCs would also fit quite well in `AppAs` I think

cmd/internal/shared/applicationserver/config.go
```diff
@@ -18,12 +18,13 @@ import (
...
+	"go.thethings.network/lorawan-stack/pkg/applicationserver/io/mqtt"
 )
 
 // DefaultApplicationServerConfig is the default configuration for the Application Server.
 var DefaultApplicationServerConfig = applicationserver.Config{
 	LinkMode: "all",
-	MQTT: applicationserver.MQTTConfig{
+	MQTT: mqtt.Config{
 		Listen:    ":1883",
 		ListenTLS: ":8883",
```
htdvisser: I think we should also set default public addresses in here, like we do with the public addresses in the email and canonical URLs in the webui configurations.

api/applicationserver.proto
rvolosatovs: I disagree:
1. It's extremely useful when e.g. storing messages to registries, since we use fieldmasks. Missing a `public` fieldpath here would instantly fail the transaction and would be fixed, however otherwise it could easily go unnoticed and cause hard-to-debug issues/panics in production.
2. It's extra context/documentation.

api/applicationserver.proto
johanstokking: Yes, this is reusable and not AS specific

api/applicationserver.proto
johanstokking: I don't think it hurts either.

Consider `public_address` here for the name

api/applicationserver.proto
```diff
@@ -151,3 +151,32 @@ service AsEndDeviceRegistry {
   };
 }
 
+// MQTTConfig represents the configuration of the Application Server MQTT frontend.
+message MQTTConfig {
+  // The public listen address of the frontend.
+  string public = 1 [(validate.rules).string.uri = true];
+  // The public listen address of the TLS frontend.
+  string public_tls = 2 [(gogoproto.customname) = "PublicTLS", (validate.rules).string.uri = true];
```
johanstokking: Idem

api/applicationserver.proto
johanstokking: 1. Indeed make this generic
2. Add a `message MQTTConnectionInfo` or something with the addresses and user name

api/applicationserver.proto
```diff
@@ -151,3 +151,32 @@ service AsEndDeviceRegistry {
...
+}
+
+// MQTTUsername represents the username that an application can use to connect to the MQTT frontend.
+message MQTTUsername {
+  string username = 1 [(validate.rules).string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$" , max_len: 36}];
```
johanstokking: MQTT user names do not follow identifier constraints; we support UIDs

pkg/applicationserver/io/grpc/grpc.go
```diff
@@ -20,20 +20,27 @@ import (
...
 type impl struct {
-	server io.Server
+	server   io.Server
+	provider MQTTConfigProvider
 }
 
 // New returns a new gRPC frontend.
-func New(server io.Server) ttnpb.AppAsServer {
-	return &impl{server}
+func New(server io.Server, provider MQTTConfigProvider) ttnpb.AppAsServer {
```
htdvisser: I'm afraid that adding args to the constructor isn't the best way to go as we'll likely keep adding more options. 

How about using variadic options?

```go
type Option interface {
	apply(*impl)
}

type optionFunc func(*impl)

func (f optionFunc) apply(i *impl) { f(i) }

func WithMQTTConfigProvider(provider MQTTConfigProvider) Option {
	return optionFunc(func(i *impl) {
		i.mqttConfigProvider = provider
	})
}

func New(server io.Server, opts ...Option) ttnpb.GtwGsServer {
	i := &impl{server}
	for _, opt := range opts {
		opt.apply(i)
	}
	return i
}
```

pkg/component/grpc_test.go
```diff
@@ -71,6 +71,10 @@ func (as *asImplementation) DownlinkQueueList(context.Context, *ttnpb.EndDeviceI
 	return nil, errors.New("not implemented")
 }
 
+func (as *asImplementation) GetMQTTConnectionInfo(context.Context, *ttnpb.ApplicationIdentifiers) (*ttnpb.MQTTConnectionInfo, error) {
```
htdvisser: We should probably start using the `ttnpb.UnimplementedXXXServer`s instead of writing all this ourselves (here, `asImplementation` could embed a `ttnpb.UnimplementedAppAsServer`). Not really necessary to do it in this PR, but good to change it at some point.

api/mqtt.proto
```diff
@@ -0,0 +1,43 @@
...
+
+// MQTTUsername represents the username used for MQTT authentication.
+message MQTTUsername {
+  string username = 1;
+}
+
+// MQTTConnectionInfo provides both MQTTConfiguration and MQTTUsername.
+message MQTTConnectionInfo {
+  MQTTConfiguration config = 1 [(gogoproto.embed) = true];
+  MQTTUsername username = 2 [(gogoproto.embed) = true];
```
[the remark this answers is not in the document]
adriansmares: Artefact from the old representation. Will remove.

docker-compose.yml
```diff
@@ -113,6 +113,14 @@ services:
       # - TTN_LW_IS_EMAIL_SMTP_USERNAME=enter username
       # - TTN_LW_IS_EMAIL_SMTP_PASSWORD=enter password
 
+      # If Application Server enabled, defaults for "thethings.example.com":
+      - TTN_LW_AS_MQTT_PUBLIC=mqtt://thethings.example.com:1883
```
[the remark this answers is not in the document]
johanstokking: Is `mqtt://` a scheme that (most) MQTT clients understand? I haven't seen it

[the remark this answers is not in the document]
johanstokking: Let's make this `..._PUBLIC_ADDRESS` also, just like we do with `_URL`

api/mqtt.proto
```diff
@@ -0,0 +1,38 @@
...
+
+// MQTTConfiguration represents the configuration of an MQTT frontend.
+message MQTTConfiguration {
+  // The public listen address of the frontend.
+  string public_address = 1 [(validate.rules).string.uri = true];
+  // The public listen address of the TLS frontend.
+  string public_tls_address = 2 [(gogoproto.customname) = "PublicTLSAddress", (validate.rules).string.uri = true];
+}
+
+// MQTTConnectionInfo provides both MQTTConfiguration and MQTTUsername.
```
johanstokking: `MQTTUsername` is not a thing anymore

cmd/internal/shared/gatewayserver/config.go
```diff
@@ -31,9 +35,11 @@ var DefaultGatewayServerConfig = gatewayserver.Config{
 			":1700": "",
 		},
 	},
-	MQTT: gatewayserver.MQTTConfig{
+	MQTT: config.MQTT{
 		Listen:    ":1882",
 		ListenTLS: ":8882",
+		Public:    fmt.Sprintf("mqtt://%s:1883", shared.DefaultPublicHost),
+		PublicTLS: fmt.Sprintf("mqtts://%s:8883", shared.DefaultPublicHost),
```
johanstokking: @htdvisser should we use `Address` suffix, like we do with `URL` in other config?

pkg/config/shared.go
```diff
@@ -292,3 +293,16 @@ type ServiceBase struct {
 	Rights           Rights                 `name:"rights"`
 	KeyVault         KeyVault               `name:"key-vault"`
 }
+
+// MQTT contains the listen and public addresses of an MQTT frontend.
+type MQTT struct {
+	Listen    string `name:"listen" description:"Address for the MQTT frontend to listen on"`
+	ListenTLS string `name:"listen-tls" description:"Address for the MQTTS frontend to listen on"`
+	Public    string `name:"public" description:"Public address of the MQTT frontend"`
+	PublicTLS string `name:"public-tls" description:"Public address of the MQTTs frontend"`
```
johanstokking: So basically `Public{TLS}Address` and `public-{tls-}address`

cmd/internal/shared/gatewayserver/config.go
adriansmares: This is consistent with `Listen`, so we either add `Address` to both or leave it as is.

johanstokking: Yes, that's true. OK, fine then!

api/gatewayserver.proto
```diff
@@ -49,6 +50,18 @@ service GtwGs {
...
   // GetConcentratorConfig associated to the gateway.
   rpc GetConcentratorConfig(google.protobuf.Empty) returns (ConcentratorConfig);
+  // GetMQTTConnectionInfo returns the MQTT server address and the username for the gateway.
+  rpc GetMQTTConnectionInfo(GatewayIdentifiers) returns (MQTTConnectionInfo) {
+    option (google.api.http) = {
+      get: "/gs/gateways/{gateway_id}/mqtt-connection-info"
+    };
+  };
+  // GetMQTTv2ConnectionInfo returns the MQTT server address and the username for the gateway.
+  rpc GetMQTTv2ConnectionInfo(GatewayIdentifiers) returns (MQTTConnectionInfo) {
```
htdvisser: "MQTT v2" may be a bit confusing, as it sounds like we're referring to an MQTT version. Not sure how to improve that...

api/mqtt.proto
```diff
@@ -0,0 +1,34 @@
...
+import "lorawan-stack/api/identifiers.proto";
+
+package ttn.lorawan.v3;
+
+option go_package = "go.thethings.network/lorawan-stack/pkg/ttnpb";
+
+// MQTTConnectionInfo provides the connection information of an MQTT frontend.
+message MQTTConnectionInfo {
+  // The public listen address of the frontend.
+  string public_address = 1 [(validate.rules).string.uri = true];
```
htdvisser: This is no longer a `(validate.rules).string.uri` right?

api/gatewayserver.proto
johanstokking: Not sure either; in GS config we also call this `mqtt-v2`, so it's consistent with that.

I'd say `GetMQTTV2ConnectionInfo()` though

Also for the comments, don't do not have to follow Go comment conventions. We used to do that, but we don't do that anymore because it doesn't look nice in generated docs.

adriansmares: It follows the same convention as the GS configuration. I'll reword the comment.

api/mqtt.proto
johanstokking: This one and the one below should indeed use the address validation

johanstokking:
@adriansmares this is ready to merge

Issue #960, opened 2019-07-09
Improve Test Coverage


htdvisser:
#### Summary

...

#### Steps to Reproduce

See Coveralls for the latest `master` branch

#### What do you see now?

The following top-50 of missed lines (excluding pkg/util/test). I already added the names of developers I think are best able to work on these.

File | Missed | Who
-- | -- | --
pkg/applicationserver/applicationserver.go | 121 | @adriansmares 
pkg/applicationserver/io/web/webhooks.go | 65 | @adriansmares 
pkg/band/band.go | 48 | ?
pkg/basicstation/cups/messages.go | 55 | @KrishnaIyer
pkg/basicstation/cups/server.go | 63 | @KrishnaIyer
pkg/basicstation/cups/update_info.go | 79 | @KrishnaIyer
pkg/blob/bucket.go | 54 | @htdvisser
pkg/component/component.go | 54 | ?
pkg/config/config.go | 69 | @htdvisser
pkg/config/tls.go | 44 | @htdvisser
pkg/crypto/join_messages.go | 54 | @rvolosatovs
pkg/email/registry.go | 68 | @htdvisser
pkg/email/smtp/smtp.go | 68 | @htdvisser
pkg/encoding/lorawan/lorawan.go | 252 | @rvolosatovs
pkg/encoding/lorawan/mac.go | 115 | @rvolosatovs
pkg/encoding/lorawan/messages.go | 142 | @rvolosatovs
pkg/errors/grpc.go | 60 | @htdvisser
pkg/frequencyplans/frequencyplans.go | 100 | ?
pkg/gatewayserver/gatewayserver.go | 80 | @KrishnaIyer 
pkg/gatewayserver/io/io.go | 44 | @KrishnaIyer 
pkg/gatewayserver/io/udp/udp.go | 70 | @KrishnaIyer 
pkg/identityserver/application_registry.go | 61 | @htdvisser
pkg/identityserver/client_registry.go | 63 | @htdvisser
pkg/identityserver/entity_access.go | 95 | @htdvisser
pkg/identityserver/gateway_registry.go | 63 | @htdvisser
pkg/identityserver/organization_registry.go | 50 | @htdvisser
pkg/identityserver/profile_picture.go | 48 | @htdvisser
pkg/identityserver/registry_search.go | 54 | @htdvisser
pkg/identityserver/store/store.go | 65 | @htdvisser
pkg/identityserver/user_registry.go | 151 | @htdvisser
[...]

PR #1373, opened 2019-09-20
Network Server testing

rvolosatovs:
#### Summary

References #960 
The US915 flow test will be added later and it will be largely based on this one, so decided to first bring the existing flow test in order.

#### Changes

- Do ~all~ *not do any* observability/metrics stuff in goroutines ~(no reason to block execution of critical functionality)~
- Use `networkserver.StartTest` in flow tests
- Add more event testing utilities
- Ensure NS tests does not panic on missed function calls
- Remove custom `ns.Close()` logic and rely exclusively on context

### DIFF
--- a/pkg/networkserver/grpc_asns.go
+++ b/pkg/networkserver/grpc_asns.go
@@ -82,15 +82,17 @@ func (ns *NetworkServer) LinkApplication(link ttnpb.AsNs_LinkApplicationServer) 
 	logger.Debug("Linked application")
 
 	events.Publish(evtBeginApplicationLink(ctx, ids, nil))
-	defer events.Publish(evtEndApplicationLink(ctx, ids, err))
 
 	select {
 	case <-ctx.Done():
 		err = ctx.Err()
 		logger.WithError(err).Debug("Context done - close stream")
+		events.Publish(evtEndApplicationLink(ctx, ids, err))
 		return err
+
 	case ws.closeCh <- struct{}{}:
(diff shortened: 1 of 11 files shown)


pkg/networkserver/grpc_asns.go
```diff
@@ -81,16 +81,18 @@ func (ns *NetworkServer) LinkApplication(link ttnpb.AsNs_LinkApplicationServer)
 
 	logger.Debug("Linked application")
 
-	events.Publish(evtBeginApplicationLink(ctx, ids, nil))
-	defer events.Publish(evtEndApplicationLink(ctx, ids, err))
+	go events.Publish(evtBeginApplicationLink(ctx, ids, nil))
```
johanstokking: @htdvisser is there a reason (not) to publish events in goroutines?

If there's a good reason to do that, should `Publish()` be non-blocking (if it's not)?

I don't like adding subtleties like this in code that are for no obvious reason.

htdvisser: There is no reason to publish events from goroutines. The default `events.PubSub` publishes to a buffered channel (default size 64) and `Publish` doesn't block unless that channel is full.

adriansmares: Also not doing it with `go events.Publish(...)` ensures that the order of the events is maintained (Otherwise publishing an End event could occur before publishing the Start event - the channel maintains this order).

PR #1381, opened 2019-09-25
Add pubsub service to JS SDK

kschiffer:
#### Summary
Add the PubSub service to JavaScript SDK.

References #1281 

#### Changes
- Add functionality to remap field masks from patches to SDK
- Add pubsub service to the SDK
- Add pubsub to the console api tree

#### Notes for Reviewers
PubSubs use a [oneof structure](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/api/applicationserver_pubsub.proto#L71-L77) to store the provider information. This results in the field masks not being directly mappable to the patch. E.g.`provider.nats` is expected for the nats property, which is actually in the root of the entity.  This is why I added functionality to the Marshaler to set field mask paths that should be remapped.

You can use [this script](https://gist.github.com/kschiffer/c97fd138d45a00dfb1a4307543a0e60e) to test.

#### Release Notes
- Add PubSub service to JavaScript SDK

### DIFF
--- a/pkg/webui/console/api/index.js
+++ b/pkg/webui/console/api/index.js
@@ -120,6 +120,14 @@ export default {
       delete: ttnClient.Applications.Webhooks.deleteById.bind(ttnClient.Applications.Webhooks),
       getFormats: ttnClient.Applications.Webhooks.getFormats.bind(ttnClient.Applications.Webhooks),
     },
+    pubsubs: {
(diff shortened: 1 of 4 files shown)


kschiffer:
Sorry. The `getAll` call was faulty. I've [updated the gist](https://gist.github.com/kschiffer/c97fd138d45a00dfb1a4307543a0e60e).

PR #1384, opened 2019-09-25
Add enforced validation to fields after change

kschiffer:
#### Summary
This PR adds a way to enforce validation of a field after it was changed.

#### Changes
- Add `enforceValidation` argument to `handleChange` handler of fields
- Add enforced validation to file input
- Add enforced validation to key-value-map (remove handler)

#### Notes for Reviewers
Added this as for some fields validation on blur is not really applicable (file input, key value map), meaning that sometimes fields are not validated when they are changed, although one would expect this to happen.
I think adding the option to the change handler is a nice and easy way to solve this. Let me know if you think otherwise.

### DIFF
--- a/pkg/webui/components/file-input/index.js
+++ b/pkg/webui/components/file-input/index.js
@@ -77,7 +77,7 @@ export default class FileInput extends Component {
     const { result: content } = event.target
 
     const data = dataTransform(content)
-    onChange(data)
+    onChange(data, true)
   }
 
   @bind
@@ -103,7 +103,7 @@ export default class FileInput extends Component {
     const { onChange } = this.props
 
     this.setState({ filename: '', error: undefined })
-    onChange(undefined)
+    onChange(undefined, true)
   }
 
   get statusMessage() {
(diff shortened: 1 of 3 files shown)


kschiffer:
I'd like to tie this option to the input component really.
What do you think of adding a static prop like `static enforceValidateOnChange = true` to respective input components, that the field component will then check for on change events?

bafonins:
> I'd like to tie this option to the input component really.

Why? 

> What do you think of adding a static prop like `static enforceValidateOnChange = true`

I dont like this solution

kschiffer:
> Why?

1. Because it is a behavior preference tied to the input component
2. Solving this at `Field` level will force us to either
  - Add a respective prop whenever we use certain input component
  - Check for certain input component types by name that have this preference
3. Our input components are already closely tied to the form fields

Hence, I don't see an issue with defining and passing field related preferences inside input components.

Please either close this PR for your solution or tell me what exact solution you're after. I need this for #1385.

PR #1391, opened 2019-09-26
Only publish MAC downlink events on success

### LINKED ISSUE #789: Duplicate events

rvolosatovs:
#### Summary

Closes #789 

#### Changes

- Only publish MAC downlink events on successful scheduling
- Fix invalid CIDs used for `RejoinParamSetupReq` and `PingSlotReq` MAC commands
- Wrap multiple return values of MAC enqueuers into a struct

#### Notes for Reviewers

Tested on metal

### DIFF
--- a/pkg/networkserver/mac_adr_param_setup.go
+++ b/pkg/networkserver/mac_adr_param_setup.go
@@ -19,6 +19,7 @@ import (
 
 	"go.thethings.network/lorawan-stack/pkg/events"
 	"go.thethings.network/lorawan-stack/pkg/frequencyplans"
+	"go.thethings.network/lorawan-stack/pkg/log"
 	"go.thethings.network/lorawan-stack/pkg/ttnpb"
 )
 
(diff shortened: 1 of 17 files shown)


pkg/networkserver/mac_adr_param_setup.go
```diff
@@ -60,23 +64,37 @@ func enqueueADRParamSetupReq(ctx context.Context, dev *ttnpb.EndDevice, maxDownL
...
+			"ack_delay_exponent", req.ADRAckDelayExponent,
+		)).Debug("Enqueued ADRParamSetupReq")
+		return []*ttnpb.MACCommand{
+				req.MACCommand(),
+			},
+			1,
+			[]events.DefinitionDataClosure{
+				evtEnqueueADRParamSetupRequest.BindData(req),
+			},
+			true
```
johanstokking: This looks a bit weird on my screen

rvolosatovs: Putting it all on 1 line is even worse IMO.
I tried different layouts, but in the end I couldn't find a way to make it look nice - so I went for clarity. Please open a technical debt issue if you think it's necessary to do something about this.

PR #1397, opened 2019-09-30
Add axios interceptor for catching errors outside SDK

### LINKED ISSUE #1396: Console Logout not working sometimes
#### Summary

Sometimes when auth cookie is not present in client the console logout doesn't work and returns `401 unauthorized`

#### Steps to Reproduce

1. Log in the console
2. Delete console auth cookie in the developer console
3. Click logout button

#### What do you see now?

`401 unauthorized` when clicking logout

Proper logout, redirect to login page

#### Environment

latest master

 add check for unauthorized error when making `logout` api request

yes

cc @kschiffer

pgalic96:
#### Summary

Closes #1396 

#### Changes

- Add axios interceptor to return errors in the same format as SDK

#### Release Notes

- Fixed bug with logout sometimes not working in the console

### DIFF
--- a/pkg/webui/console/api/index.js
+++ b/pkg/webui/console/api/index.js
@@ -43,6 +43,18 @@ const instance = axios.create({
   headers: { 'X-CSRF-Token': csrf },
 })
 
+instance.interceptors.response.use(
+  response => {
+    return response
+  },
+  error => {
+    if ('response' in error && error.response && 'data' in error.response) {
+      return error.response.data
+    }
+    return error
+  },
+)
+
 export default {
   console: {
     token() {


kschiffer:
Please solve this from the origin, which is the fact that our logout requests are plain axios request. So it won't return the error response message by default, but a whole axios specific error object, which our error library cannot work with.
[This is how we handle request errors in the SDK](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/sdk/js/src/api/http.js#L101-L107). Later on, we should look into integrating the  login/logout flow into the SDK as well.

PR #1402, opened 2019-10-01
Add support button to console

kschiffer:
#### Summary
Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/1734
References #1086 

image

#### Changes
- Add optional `support-link` config to console
- Display a support button in the footer, when the config is set
- Update env selectors, prop-types, lib

#### Release Notes

- Added `support-link` URI config to the Console to show a "Get Support" button

### DIFF
--- a/pkg/console/console.go
+++ b/pkg/console/console.go
@@ -35,12 +35,13 @@ type UIConfig struct {
 
 // FrontendConfig is the configuration for the Console frontend.
 type FrontendConfig struct {
-	Language string          `json:"language" name:"-"`
-	IS       webui.APIConfig `json:"is" name:"is"`
-	GS       webui.APIConfig `json:"gs" name:"gs"`
-	NS       webui.APIConfig `json:"ns" name:"ns"`
-	AS       webui.APIConfig `json:"as" name:"as"`
-	JS       webui.APIConfig `json:"js" name:"js"`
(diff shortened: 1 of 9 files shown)


pkg/console/console.go
```diff
@@ -35,12 +35,13 @@ type UIConfig struct {
...
 // FrontendConfig is the configuration for the Console frontend.
 type FrontendConfig struct {
-	Language string          `json:"language" name:"-"`
-	IS       webui.APIConfig `json:"is" name:"is"`
-	GS       webui.APIConfig `json:"gs" name:"gs"`
-	NS       webui.APIConfig `json:"ns" name:"ns"`
-	AS       webui.APIConfig `json:"as" name:"as"`
-	JS       webui.APIConfig `json:"js" name:"js"`
+	Language    string          `json:"language" name:"-"`
+	SupportLink string          `json:"support_link" name:"support-link" description:"The URL pointing to a support website, shown as a button in the footer"`
```
johanstokking: Technically it can be a `mailto:` link as well, and we probably see people using that

johanstokking: So it's still `SupportLink`, but not not URL but URI (which doesn't have to be a website)

htdvisser:
New config requires release note

PR #1403, opened 2019-10-01
Support empty root keys from Crypto Server

### LINKED ISSUE #1401: Join Server fails when Crypto Server does not expose root keys
#### Summary

Join Server fails when Crypto Server does not expose root keys

#### Steps to Reproduce

1. Build CS implementation that returns an error when root keys are not exposed (TTI CS currently returns `PERMISSION_DENIED`)
2. Wire JS to that CS
3. Get device from JS, passing `root_keys` in field mask

#### What do you see now?

JS passes the permission denied error back to the caller (i.e. Console shows Forbidden and does not show device at all)

The API defines behavior;

https://github.com/TheThingsNetwork/lorawan-stack/blob/de84884651e6f7f835a1cb2defc8e726d91d072d/api/joinserver.proto#L107

https://github.com/TheThingsNetwork/lorawan-stack/blob/de84884651e6f7f835a1cb2defc8e726d91d072d/api/joinserver.proto#L114

JS does not check the error and `UNIMPLEMENTED` is not the right status code for this.

No error from CS propagated to caller; if CS doesn't expose root keys, the caller should just get empty root keys

- CS should return `FAILED_PRECONDITION`
- Change API to document this API behavior
- JS should check this error type and return empty root keys to the caller
- Caller should support getting empty root keys from JS if requested

Yes

johanstokking:
#### Summary

Closes #1401 

#### Changes

- Document expected error code when root keys are not exposed
- Support `nil` root keys from CS in JS

#### Release Notes

- Added support for Crypto Servers that do not expose device root keys

### DIFF
--- a/pkg/crypto/cryptoservices/cryptoservices.go
+++ b/pkg/crypto/cryptoservices/cryptoservices.go
@@ -35,13 +35,17 @@ type Network interface {
 	EncryptJoinAccept(ctx context.Context, dev *ttnpb.EndDevice, version ttnpb.MACVersion, payload []byte) ([]byte, error)
 	EncryptRejoinAccept(ctx context.Context, dev *ttnpb.EndDevice, version ttnpb.MACVersion, payload []byte) ([]byte, error)
 	DeriveNwkSKeys(ctx context.Context, dev *ttnpb.EndDevice, version ttnpb.MACVersion, jn types.JoinNonce, dn types.DevNonce, nid types.NetID) (NwkSKeys, error)
-	GetNwkKey(ctx context.Context, dev *ttnpb.EndDevice) (types.AES128Key, error)
+	// GetNwkKey returns the NwkKey of the given end device.
+	// If the implementation does not expose root keys, this method returns nil, nil.
+	GetNwkKey(ctx context.Context, dev *ttnpb.EndDevice) (*types.AES128Key, error)
 }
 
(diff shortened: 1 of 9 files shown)


johanstokking:
Blocking https://github.com/TheThingsIndustries/lorawan-stack/pull/1750

pkg/crypto/cryptoservices/cryptoservices.go
```diff
@@ -35,13 +35,13 @@ type Network interface {
 	EncryptJoinAccept(ctx context.Context, dev *ttnpb.EndDevice, version ttnpb.MACVersion, payload []byte) ([]byte, error)
 	EncryptRejoinAccept(ctx context.Context, dev *ttnpb.EndDevice, version ttnpb.MACVersion, payload []byte) ([]byte, error)
 	DeriveNwkSKeys(ctx context.Context, dev *ttnpb.EndDevice, version ttnpb.MACVersion, jn types.JoinNonce, dn types.DevNonce, nid types.NetID) (NwkSKeys, error)
-	GetNwkKey(ctx context.Context, dev *ttnpb.EndDevice) (types.AES128Key, error)
+	GetNwkKey(ctx context.Context, dev *ttnpb.EndDevice) (*types.AES128Key, error)
```
htdvisser: Add comment that the returned key may still be `nil` even if the returned error is `nil`. Also below.

Issue #1240, opened 2019-08-26
Use service discovery for CLI configuration


johanstokking:
#### Summary

Add endpoint in The Things Stack for serving the `ttn-lw-cli` config file

#### Why do we need this?

To download the concerning configuration file, most importantly including the host names of the gRPC services and potentially CA to use

#### What is already there? What do you see now?

Manual configuration

#### What is missing? What do you want to see?

A public endpoint to download the `ttn-lw-cli` configuration.

Bonus would be;
- Add a command in the CLI that takes the host name and downloads the `yml` file
- By default saves it to `ttn-lw-cli.yml` in the working dir (confirm overwrite file etc). We can have flags also like `--local` or `--global` to specify where the file should go
- Uses `ttn-lw-cli.yml` from the working dir (before checking `$HOME/.ttn-lw-cli.yml`) if present

That way, users can simply:

```
$ ttn-lw-cli use join.thethings.industries --local
$ ttn-lw-cli login
```

#### How do you propose to implement this?

For serving the configuration, simply a public endpoint to `GET` the configuration. I guess that YAML is all we need, but we may consider other formats through `Accept`

#### Can you do this yourself and submit a Pull Request?

Can review

htdvisser:
We'll simplify this to generate a config in the CLI (without calling the server) for a standard deployment on `host`.

johanstokking:
Insight from #1408;

- With `--local` (default), put the dotfile in the working dir (`os.Getwd()`)
- With `--user`, put the dotfile in the user config dir (`os.UserConfigDir()`)
- Extend `pkg/rpcmiddleware/discover` to return the endpoints based on the user provided hostname. Probably introduce a new struct there with the discovery result. If #138 isn't closed yet, return sane defaults (i.e. all components enabled and use current default port fallback)

neoaggelos:
With #2150 merged, is there anything else that we would like to have for this, or can we close the issue?

johanstokking:
In #2150 I suggest to use `pkg/rpcmiddleware/discover` instead of hardcoding the default ports in the CLI.

Please extract https://github.com/TheThingsNetwork/lorawan-stack/blob/feature/pba/pkg/rpcmiddleware/discover/discover.go#L60-L63 to an exposed function and use that to get the address for TLS/non-TLS based on user input instead.

johanstokking:
Changing labels as it's not broken but can be nicer. Since it works, lowering priority.

PR #1408, opened 2019-10-01
Fix reading config from files in order

johanstokking:
#### Summary

References #1240

#### Changes

- Fixed a bug introduced [long ago](https://github.com/TheThingsNetwork/lorawan-stack/commit/64b25ff2c035d899f51785b087bfadfadadb1797) that we never evaluated multiple locations. Or it was intended this way, but then I don't get the rest of the logic (joining paths and reading the files in order)

#### Notes for Reviewers

This is a preparation for #1240 in the sense that the `use` command can write the `.ttn-lw-cli.yml` to `$PWD`, which then gets picked up because of this.

#### Release Notes

- Fixed reading configuration from current, home and XDG directories


pkg/config/path.go
```diff
@@ -43,21 +43,22 @@ func WithConfigFileFlag(flag string) Option {
...
+		// check PWD
 		if pwd := os.Getenv("PWD"); pwd != "" {
-			envPaths = []string{path.Join("$PWD", dotfile)}
-			paths = []string{path.Join(pwd, dotfile)}
+			envPaths = append(envPaths, path.Join("$PWD", dotfile))
+			paths = append(paths, path.Join(pwd, dotfile))
 		}
 
 		// check HOME
 		if home := os.Getenv("HOME"); home != "" {
```
htdvisser: While we're at it, shouldn't we also rewrite this to use [`os.UserHomeDir`](https://golang.org/pkg/os/#UserHomeDir) and [`os.UserConfigDir`](https://golang.org/pkg/os/#UserConfigDir) so that it also works on Windows?

htdvisser: Not sure what to do with `$PWD` above on Windows, but I'm pretty sure that doesn't work either.

johanstokking: > While we're at it, shouldn't we also [insert improvement here]

... is a general feeling we can apply to `pkg/config`

In any case, let's see what we can do here

PR #1409, opened 2019-10-01
Fix stack config issue in console

kschiffer:
#### Summary
This quickfix PR will change the json structure of the frontend config that is passed to the console frontend. This resolves a bug that caused the overview page to crash since #1402.

#### Changes
- Change the config structure of the console config setup (config flags remain the same)
  - Move the stack config from the root into `stack_config` prop (json)
- Update env selectors, lib and references accordingly

### DIFF
--- a/pkg/webui/console/views/overview/index.js
+++ b/pkg/webui/console/views/overview/index.js
@@ -113,7 +113,9 @@ export default class Overview extends React.Component {
   }
 
   render() {
-    const { config } = this.props.env
+    const {
+      config: { stack: stackConfig },
+    } = this.props.env
     const { fetching, applicationCount, gatewayCount, userId } = this.props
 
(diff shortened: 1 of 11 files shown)


pkg/webui/console/views/overview/index.js
```diff
@@ -179,12 +181,13 @@ export default class Overview extends React.Component {
...
             <div className={style.componentCards}>
-              {Object.keys(config).map(function(componentKey) {
+              {Object.keys(stackConfig).map(function(componentKey) {
                 if (componentKey === 'language') {
                   return null
                 }
-                const component = config[componentKey]
+                const component = stackConfig[componentKey]
                 const name = componentMap[componentKey]
+                console.log(stackConfig)
```
pgalic96: Unnecessary log

johanstokking:
@kschiffer let me know when I should merge

Issue #1404, opened 2019-10-01
Device templates from data and bulk create


johanstokking:
#### Summary

Device templates from data and bulk create

#### Why do we need this?

To import devices from a file, i.e. migration or claiming by manifests.

#### What is already there? What do you see now?

In Console, individual device creation.

In CLI we have device template conversion support, template execution (to turn a template into an end device) which can be piped to device create.

#### What is missing? What do you want to see?

Console support for;
1. Listing formats for conversion (rpc call)
2. Convert a file with the chosen format (rpc call)
3. Template execution (split template to end device and field mask)
4. Bulk create with progress bar (rpc calls)

#### How do you propose to implement this?

1. Listing formats for conversion

CLI equivalent:
```
$ ttn-lw-cli device template list-formats
```

API call: https://github.com/TheThingsNetwork/lorawan-stack/blob/f8c5f452744ff56ae1e2c67ae6972ec76b3795f0/api/end_device_services.proto#L69

Show a dropdown sorted by `name` and pass the `id` in the next call

2. Convert a file with the chosen format

CLI equivalent:
```
$ ttn-lw-cli device template from-data --local-file $filename --format-id $formatid > ui.json
```

API call: https://github.com/TheThingsNetwork/lorawan-stack/blob/f8c5f452744ff56ae1e2c67ae6972ec76b3795f0/api/end_device_services.proto#L76

This returns an array of device templates. A device template is basically a device with a field mask (i.e. which fields are known): https://github.com/TheThingsNetwork/lorawan-stack/blob/f8c5f452744ff56ae1e2c67ae6972ec76b3795f0/api/end_device.proto#L543-L547

There can be many (in the order of tens of thousands). For now, I wouldn't bother showing them, only the count.

3. Execute the template

CLI equivalent:
```
$ cat ui.json | ttn-lw-cli device template execute > devices.json
```

This doesn't do anything besides making an `EndDevice` with the values and the field mask in the template. You can probably skip this step and go straight to:

4. Bulk create

CLI equivalent:
```
$ cat devices.json | ttn-lw-cli device create --application-id <app-id> [--with-claim-authentication-code]
```

This is basically creating all the devices. Depending on Console cluster config, this has to go to IS, JS, NS and AS. When a create fails, rollback that device (like we do now), and stop.

Note that flag here; if this checkbox is checked (for all devices), generate a random 4-byte hex string (per device) and put that in `claim_authentication_code.value` (and add that to the field mask).

[...]

PR #1411, opened 2019-10-02
Add file extensions to device template formats

johanstokking:
#### Summary

Add file extensions to device template formats

This is needed for #1404 to filter files in the frontend already. Server can't (and shouldn't) enforce it; it's just a hint to clients

#### Changes

- Add optional file extensions to `EndDeviceTemplateFormat`
- Indicate that Microchip manifests are `.json`


api/end_device.proto
```diff
@@ -549,6 +549,7 @@ message EndDeviceTemplate {
 message EndDeviceTemplateFormat {
   string name = 1 [(validate.rules).string.max_len = 100];
   string description = 2 [(validate.rules).string.max_len = 200];
+  repeated string file_extensions = 3 [(validate.rules).repeated = {max_items: 100, unique: true, items: {string: {pattern: "^(?:\\.[a-z0-9]{1,16}){1,2}$"}}}];
```
johanstokking: Made this up myself; not sure what conventional is. This allows for quite long extensions (i.e. `.cloudformation`) and double extensions (i.e. `.tar.gz`) which I guess enough

PR #1413, opened 2019-10-02
QR Code Generator service

### LINKED ISSUE #1410: QR code generation service
#### Summary

QR code generation service

#### Why do we need this?

To generate QR codes in one place instead of moving the responsibility to the clients.

See https://github.com/TheThingsNetwork/lorawan-stack/issues/1407#issuecomment-537034928

#1392 is introducing an offline QR code generation in CLI. This is offline, but imports QR code formats and rendering in the CLI, which may run out of sync with the server version.

We have a `pkg/qrcode` with the necessary plumbing

First class service for QR code generation and potentially parsing

It returns a list of formats for entities (i.e. `EndDevice` and `Gateway`), the required field mask and has an rpc for generating the QR code. It should be able to render it (png in requested size), return a bitmap (`[][]bool`) and/or text value.

First class component, new gRPC service, start with `EndDevice` and generation of QR codes, but allow for `Gateway` and parsing QR codes too

Yes

johanstokking:
#### Summary

Closes #1410 

#### Changes

- Add service to list QR code formats and generate them
- Use the service in the CLI

#### Notes for Reviewers

@rvolosatovs no review required

#### Release Notes

- Added service to list QR code formats and generate QR codes in PNG format


api/qrcodegenerator.proto
```diff
@@ -0,0 +1,108 @@
...
+
+  // Returns the supported formats.
+  rpc ListFormats(google.protobuf.Empty) returns (QRCodeFormats) {
+    option (google.api.http) = {
+      get: "/qrcodes/end-devices/formats"
+    };
+  };
+
+  // Generates the text value of the QR code.
+  rpc GenerateText(GenerateEndDeviceQRCodeTextRequest) returns (GenerateQRCodeTextResponse) {
```
htdvisser: Since `GenerateImage` also includes `text`, isn't it better to merge `GenerateText` and `GenerateImage` into one RPC and have a request message like

```proto
message GenerateEndDeviceQRCodeRequest {
  string format_id = 1 [(gogoproto.customname) = "FormatID", (validate.rules).string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$", max_len: 36}];
  EndDevice end_device = 2 [(gogoproto.nullable) = false, (validate.rules).message.required = true];

  message ImageConfig {
    (maybe) image_type = 1; // bmp, png, ...
    uint32 image_size = 2 [(validate.rules).uint32.gte = 10, (validate.rules).uint32.lte = 1000];
  }

  ImageConfig image_config = 3; // if set, include the QR code rendered as an image
}
```

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -830,79 +830,142 @@ values will be stored in the Join Server.`,
 			return io.Write(os.Stdout, config.OutputFormat, ids)
 		},
 	}
+	endDevicesListQRCodeFormatsCommand = &cobra.Command{
+		Use:     "list-qr-formats",
+		Aliases: []string{"lsqrfmts", "lsqrfmt"},
```
htdvisser: For consistency with `endDeviceTemplatesListFormats`, this would be `[]string{"ls-qr-formats", "listqrformats", "lsqrformats"},`. 

You may also want to add `fmts`, `fmt` to the other one.

Also consider `endDevicesListFrequencyPlans` and `gatewaysListFrequencyPlans`

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -973,11 +1036,11 @@ func init() {
 	endDevicesClaimCommand.Flags().String("target-application-server-id", "", "")
 	endDevicesClaimCommand.Flags().Bool("invalidate-authentication-code", true, "invalidate the claim authentication code to block subsequent claiming attempts")
 	endDevicesCommand.AddCommand(endDevicesClaimCommand)
+	endDevicesCommand.AddCommand(endDevicesListQRCodeFormatsCommand)
 	endDevicesGenerateQRCommand.Flags().AddFlagSet(endDeviceIDFlags())
-	endDevicesGenerateQRCommand.Flags().String("format", "tr005draft2", "format (tr005draft2)")
-	endDevicesGenerateQRCommand.Flags().Int("size", 300, "")
-	endDevicesGenerateQRCommand.Flags().String("file", "", "file to write QR code in PNG format")
-	endDevicesGenerateQRCommand.Flags().Bool("print", false, "print QR code as text")
+	endDevicesGenerateQRCommand.Flags().String("format-id", "tr005draft2", "")
```
htdvisser: If we set this as a default, we'll be stuck with it until v3, otherwise we would break CLI compatibility

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -973,11 +1036,11 @@ func init() {
...
+	endDevicesGenerateQRCommand.Flags().Uint32("size", 300, "")
```
htdvisser: This deserves a comment, because this is about the pixel size of the PNG, not the module size of the QR.

pkg/qrcode/qrcode.go
```diff
@@ -59,3 +60,41 @@ func Parse(data []byte) (Data, error) {
...
+
+// EndDeviceFormat is a end device QR code format.
+type EndDeviceFormat interface {
+	Format() *ttnpb.QRCodeFormat
+	New() EndDeviceData
+}
+
+var (
+	endDeviceFormats   = map[string]EndDeviceFormat{}
+	endDeviceFormatsMu sync.RWMutex
```
adriansmares: Are there any plans for dynamic registration of formats ? If they are only registered at `init()` I don't see why the mutex should be used.

api/picture.proto
```diff
@@ -0,0 +1,38 @@
...
+message Picture {
+  message Embedded {
+    // MIME type of the picture.
+    string mime_type = 1;
+    // Picture data. A data URI can be constructed as follows:
+    // `data:<mime_type>;base64,<data>`.
+    bytes data = 2;
+  }
+
+  // Embedded picture, always maximum 128px in size.
```
htdvisser: This comment is no longer valid for QR codes

api/qrcodegenerator.proto
```diff
@@ -0,0 +1,92 @@
...
+  string text = 1;
+  // QR code in PNG format, if requested.
+  Picture image = 2;
+}
+
+service EndDeviceQRCodeGenerator {
+  // Return the QR code format.
+  rpc GetFormat(GetQRCodeFormatRequest) returns (QRCodeFormat) {
+    option (google.api.http) = {
+      get: "/qrcodes/end-devices/formats/{format_id}"
```
htdvisser: Should the URL path be `qr-codes` instead of `qrcodes`?

pkg/qrcode/qrcode.go
johanstokking: We don't really need the mutex indeed. It's doesn't hurt though. I didn't want to expose the `map[string]EndDeviceFormat{}` so it could be changed by anyone. And while I was at it, I protected it.

PR #1415, opened 2019-10-03
Add restart button on event stream error

### LINKED ISSUE #1337: Allow restarting events stream on error
#### Summary

Currently, there is no way no restart the events stream other than refreshing the page.

#### Why do we need this?

1. Refreshing the page would remove all persisted events (on the client)
2. The users should be able to recover errors

1. Redux setup for events
2. Error handling in the `<Events />` and `<EventsWidget />` component
3. Error notification when the events stream fails

UI to restart the events stream

#### Environment

console (browser) 3.1.2

Add a button to the error notification that request the as to open a new events stream. Similar to gateway connection (?)
Screenshot 2019-09-09 at 16 37 44

yes

pgalic96:
#### Summary

Closes #1337 

#### Changes

- Add `onRestart` function prop to `EventSubscription` component which dispatches `startStream` event
- Set error to `undefined` on start stream success

### DIFF
--- a/pkg/webui/console/containers/events-subscription/events-subscription.js
+++ b/pkg/webui/console/containers/events-subscription/events-subscription.js
@@ -21,13 +21,29 @@ const { Widget } = Events
 
 class EventsSubscription extends React.Component {
   render() {
-    const { id, widget, events, onClear, toAllUrl, error } = this.props
+    const { id, widget, events, onClear, toAllUrl, onRestart, error } = this.props
 
     if (widget) {
-      return <Widget emitterId={id} events={events} toAllUrl={toAllUrl} error={error} />
+      return (
+        <Widget
+          emitterId={id}
+          events={events}
+          toAllUrl={toAllUrl}
+          onRestart={onRestart}
+          error={error}
+        />
+      )
     }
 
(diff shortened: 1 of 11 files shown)


pkg/webui/console/containers/events-subscription/events-subscription.js
```diff
@@ -36,13 +52,15 @@ EventsSubscription.propTypes = {
   events: PropTypes.events,
   id: PropTypes.string.isRequired,
   onClear: PropTypes.func,
+  onRestart: PropTypes.func,
```
kschiffer: I think this can be required.

pkg/webui/components/notification/notification.styl
```diff
@@ -63,3 +63,19 @@ left-border($color)
...
+    flex-grow: 2
+    justify-content: flex-end
+
+    &-button
+      width: 10%
+
+    &-message
+      width: 50%
+      color: $tc-subtle-gray
+      font-size: $fs.s
```
kschiffer: I think you can do all of this a lot simpler if you wrapped the content of the notification around a div and then display the text and the button next to each other as inline. Regard the button as part of the content.

PR #1419, opened 2019-10-03
Test coverage improvements

rvolosatovs:
#### Summary

References #960 

#### Changes

- Improve test coverage in `pkg/encoding/lorawan`
- Improve test coverage in `pkg/crypto`, fix test organization. (looks like `gofumports` performed some changes as well)

#### Notes for Reviewers

Did not compute MICs by hand, but merely used the existing code to do that. If I had 1.1 device lying around, I would use that to generate the MIC, but that's not the case.

### DIFF
--- a/pkg/encoding/lorawan/lorawan_test.go
+++ b/pkg/encoding/lorawan/lorawan_test.go
@@ -15,6 +15,7 @@
 package lorawan_test
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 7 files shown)


pkg/encoding/lorawan/lorawan_test.go
```diff
@@ -818,3 +819,675 @@ func TestLoRaWANEncodingRawJoinAcceptPayload(t *testing.T) {
 		})
 	}
 }
+
+func TestDeviceEIRPToFloat32(t *testing.T) {
+	for _, tc := range []struct {
+		Enum  ttnpb.DeviceEIRP
+		Float float32
+	}{
+		{
```
htdvisser: It may be nice to format these on one line each so that they look a little more like a table

pkg/encoding/lorawan/lorawan_test.go
```diff
@@ -818,3 +819,675 @@ func TestLoRaWANEncodingRawJoinAcceptPayload(t *testing.T) {
...
+			Float: 38,
+			Enum:  ttnpb.DEVICE_EIRP_36,
+		},
+		{
+			Float: 37,
+			Enum:  ttnpb.DEVICE_EIRP_36,
+		},
+		{
+			Float: 36,
```
htdvisser: This case is already covered in `TestDeviceEIRPToFloat32`, some more below.

pkg/encoding/lorawan/lorawan_test.go
```diff
@@ -818,3 +819,675 @@ func TestLoRaWANEncodingRawJoinAcceptPayload(t *testing.T) {
...
+	for _, tc := range []struct {
+		Uint uint32
+		Enum ttnpb.ADRAckLimitExponent
+	}{
+		{
+			Uint: 32769,
+			Enum: ttnpb.ADR_ACK_LIMIT_32768,
+		},
+		{
+			Uint: 32768,
```
htdvisser: This case is already covered in `TestADRAckLimitExponentToFloat32`, more below.

PR #1426, opened 2019-10-07
Get nested error clause message and attributes

### LINKED ISSUE #1319: Improve error rendering in Web UI
#### Summary

We need to improve how errors are rendered in the web UI.

#### Why do we need this?

Right now they're not informative enough.

Creating an entity (in this case an organization) with an invalid ID returns the following error stack in the CLI:

```
error:pkg/errors:validation (invalid `organization`: embedded message failed validation)
    name=CreateOrganizationRequestValidationError
    reason=embedded message failed validation
    field=organization
    correlation_id=ff913a0396cc4b109c869cf4d172c62d
--- error:pkg/errors:validation (invalid `ids`: embedded message failed validation)
    field=ids
    name=OrganizationValidationError
    reason=embedded message failed validation
--- error:pkg/errors:validation (invalid `organization_id`: value does not match regex pattern "^[a-z0-9](?:[-]?[a-z0-9]){2,}$")
    reason=value does not match regex pattern "^[a-z0-9](?:[-]?[a-z0-9]){2,}$"
    field=organization_id
    name=OrganizationIdentifiersValidationError
```

but the console only shows

Screenshot 2019-09-05 at 11 44 45

I'd like to see what caused the error.

Look at the `cause` field in the response:

```json
{
  "code": 3,
  "message": "error:pkg/errors:validation (invalid `organization`: embedded message failed validation)",
  "details": [
    {
      "namespace": "pkg/errors",
      "name": "validation",
      "message_format": "invalid `{field}`: {reason}",
      "attributes": {
        "field": "organization",
        "name": "CreateOrganizationRequestValidationError",
        "reason": "embedded message failed validation"
      },
      "correlation_id": "eefb90988c284758956cc3585218ea7f",
      "cause": {
        "namespace": "pkg/errors",
        "name": "validation",
        "message_format": "invalid `{field}`: {reason}",
        "attributes": {
          "field": "ids",
          "name": "OrganizationValidationError",
          "reason": "embedded message failed validation"
        },
        "cause": {
          "namespace": "pkg/errors",
          "name": "validation",
          "message_format": "invalid `{field}`: {reason}",
          "attributes": {
            "field": "organization_id",
            "name": "OrganizationIdentifiersValidationError",
            "reason": "value does not match regex pattern \"^[a-z0-9](?:[-]?[a-z0-9]){2,}$\""
          },
          "code": 3
        },
        "code": 3
      },
      "code": 3
    }
  ]
}
```

pgalic96:
#### Summary

Closes #1319 

#### Changes

- Add functions in error utils to render the correct error message in the last nested clause

#### Notes for Reviewers

@kschiffer not sure if this could be done by just using arrow functions?


pkg/webui/lib/errors/utils.js
```diff
@@ -124,12 +124,45 @@ export const getBackendErrorId = error => error.message.split(' ')[0]
...
+  }
+  return error.details[0].message_format || error.message.replace(/^.*\s/, '')
+}
+
+/**
+ * Returns the error cause message of the error.
+ * @param {Object} cause - The backend error cause.
+ * @returns {string} The cause message.
+ */
+export function getBackendErrorCause(cause) {
```
kschiffer: I think it would be better if the argument was always the root error object, otherwise, the usage might become confusing.

pkg/webui/lib/errors/utils.js
```diff
@@ -124,12 +124,45 @@ export const getBackendErrorId = error => error.message.split(' ')[0]
...
+  }
+  return error.details[0].attributes
+}
+
+/**
+ * Returns the error cause attributes of the error.
+ * @param {Object} cause - The backend error cause.
+ * @returns {string} The cause attributes or undefined.
+ */
+export function getBackendErrorCauseAttributes(cause) {
```
kschiffer: See above.

htdvisser:
Please be aware that the entire chain may be relevant, not only the root error.

kschiffer:
What root error? With this PR, we will display the error's cause message if possible, instead of the root error message. This should improve the relevance of the message, though it's certainly still not an optimal solution.

htdvisser:
With the root of an error I mean the root cause.

The error I picked in #1319 has three errors:

1. error:pkg/errors:validation (invalid `organization`: embedded message failed
2. error:pkg/errors:validation (invalid `ids`: embedded message failed validation)
3. error:pkg/errors:validation (invalid `organization_id`: value does not match regex pattern "^[a-z0-9](?:[-]?[a-z0-9]){2,}$")

In this specific case, the root cause (number 3, the deepest in the cause chain) may be the most relevant, but that isn't always the case. The code in this PR always returns the deepest cause in the chain, and would ignore everything in between.

Consider the following hypothetical error chain:

1. could not update application
2. database transaction failed
3. application not found
4. database error: row count is zero

With this PR merged, the console would report the database error, and not the more useful "application not found" error.

kschiffer:
Indeed. See my thoughts on this in the [issue](https://github.com/TheThingsNetwork/lorawan-stack/issues/1319#issuecomment-539425875):

> I think it makes sense to extend our error notification to have a collapsible "Error Details" section, which contains the raw error message returned from the backend. We could then display a user-friendly message (e.g. "There was a problem while creating the End Device", or provide just the cause message, see #1426) and provide developer-directed details on demand.

htdvisser:
I'm fine with that. So should we change this PR to display the entire error chain in such a collapsible "Error Details"? This would be in addition to the current view with only the outermost error (if `cause` is not null/undefined).

kschiffer:
Yes, I think that would be best.

@pgalic96 So additionally to the changes here we need some adaptations to the notification component. I'd suggest introducing a new `ErrorNotification` component that will take care of:
- Rendering an appropriate error message title
- Rendering the error cause (if present) into the notification body
- Rendering a collapsible "Show Error Details" section
- Render backend error messages as markdown (or at least convert ` `` `'s into `<pre/>`'s)

It should look like this:
image

pgalic96:
@kschiffer so you mean that we should remove the logic for rendering error messages from `Notification` and handle it all in a new component `ErrorNotification`? Because we could also conditionally render this additional information inside the `Notification` component.

kschiffer:
Yeah, that's also possible. If you can manage to do that in a lean way then that's fine too. Otherwise, I think it's good to avoid making these sorts of _swiss-army-knife_-type components, as they can be hard to maintain. I was thinking of doing it via [component specialization](https://reactjs.org/docs/composition-vs-inheritance.html#specialization):
- Refactor the `Notification` component to become more generic
  - Remove conditional error message rendering logic
  - Extend the component to hold collapsible detail info (add `detailMessage` ("Show Details"), `details` props)
- Introduce a new and specific `ErrorNotification` component, which will basically only take the error object and computes the necessary props from it to render the `Notification` component in the way needed

PR #1429, opened 2019-10-08
Support fetching configuration from blobs

### LINKED ISSUE #1272: Refactor fetch package to support blobs
#### Summary

Refactor `pkg/fetch` to support `pkg/blob`

#### Why do we need this?

To fetch from blobs

- Fetching from local disk and URLs
- Fetching from bucket (there's a `pkg/fetch.FromBucket` but misses a context)
- Blobs package to work with cloud buckets and multiple providers

Be able to use blobs with the generic fetcher, respecting service configuration and context

- Use the generic `pkg/fetch` package but use `pkg/blobs`
- This requires refactoring the fetcher interface to support passing context for reading files
- Implement configuration for frequency plans, webhook templates and device repository
- Note that the `pkg/config.FrequencyPlansConfig.Store()` doesn't have access to `Blob` config, which it should, somehow, probably by promoting this `Store()` method to `ServiceBase`

Can review

rvolosatovs:
#### Summary

Closes #1272 

#### Changes

- Support fetching configuration from blobs
- Remove `blob.Config` to avoid import cycles. Moving this logic to `config` is also more consistent with what we do for everything else

#### Notes for Reviewers

Instead of explicitly passing `config.BlobConfig` it could be stored in the context, however we should keep in mind that `config.BlobConfig` may(and in most cases does) contain sensitive data, which may get exposed, hence I decided to err on the side of caution and simplicity and just pass it explicitly.

@adriansmares @bafonins no need for review

#### Release Notes

- Frequency plans, device repository and interoperability configuration can now be stored in AWS S3 buckets or GCP blobs.

### DIFF
--- a/pkg/applicationserver/applicationserver.go
+++ b/pkg/applicationserver/applicationserver.go
@@ -38,6 +38,7 @@ import (
 	"go.thethings.network/lorawan-stack/pkg/config"
 	"go.thethings.network/lorawan-stack/pkg/crypto"
 	"go.thethings.network/lorawan-stack/pkg/crypto/cryptoutil"
+	"go.thethings.network/lorawan-stack/pkg/devicerepository"
 	"go.thethings.network/lorawan-stack/pkg/errors"
 	"go.thethings.network/lorawan-stack/pkg/events"
 	"go.thethings.network/lorawan-stack/pkg/interop"
@@ -117,7 +118,7 @@ func New(c *component.Component, conf *Config) (as *ApplicationServer, err error
 		}
 	}
 
-	drCl, err := baseConf.DeviceRepository.Client(ctx, baseConf.Blob)
+	drFetcher, err := baseConf.DeviceRepository.Fetcher(ctx, baseConf.Blob)
 	if err != nil {
 		return nil, err
 	}
@@ -130,7 +131,9 @@ func New(c *component.Component, conf *Config) (as *ApplicationServer, err error
 		linkRegistry:   conf.Links,
 		deviceRegistry: conf.Devices,
 		formatter: payloadFormatter{
-			repository: drCl,
+			repository: &devicerepository.Client{
+				Fetcher: drFetcher,
+			},
 			upFormatters: map[ttnpb.PayloadFormatter]messageprocessors.PayloadDecoder{
 				ttnpb.PayloadFormatter_FORMATTER_JAVASCRIPT: javascript.New(),
 				ttnpb.PayloadFormatter_FORMATTER_CAYENNELPP: cayennelpp.New(),
--- a/pkg/applicationserver/applicationserver.go
+++ b/pkg/applicationserver/applicationserver.go
@@ -105,15 +105,13 @@ func New(c *component.Component, conf *Config) (as *ApplicationServer, err error
 
 	var interopCl InteropClient
 	if !conf.Interop.IsZero() {
-		var fallbackTLS *tls.Config
-		cTLS, err := c.GetTLSClientConfig(ctx)
-		if err != nil {
-			log.FromContext(ctx).WithError(err).Warn("Could not get fallback TLS config for interoperability")
-		} else {
-			fallbackTLS = cTLS
+		interopConf := conf.Interop.InteropClient
+		interopConf.GetFallbackTLSConfig = func(ctx context.Context) (*tls.Config, error) {
+			return c.GetTLSClientConfig(ctx)
 		}
+		interopConf.BlobConfig = c.GetBaseConfig(ctx).Blob
 
-		interopCl, err = interop.NewClient(ctx, conf.Interop.InteropClient, baseConf.Blob, fallbackTLS)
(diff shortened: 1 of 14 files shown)


johanstokking:
> Instead of explicitly passing `config.BlobConfig` it could be stored in the context

Shouldn't the clients (interop, device repository, frequency plans, etc) just take a fetcher interface? If they are configured for using blobs, the blob fetcher gets passed, otherwise a file fetcher based on the configured folder?

rvolosatovs:
@johanstokking because `Store` methods construct the fetcher and in order to do that the blob config is necessary.
I can add the `Fetcher(blobConf BlobConfig) fetch.Interface` methods if you want and let callers construct fetchers using that and then pass it to the `Store` method, but I don't see any use cases for this in our codebase

rvolosatovs:
See https://github.com/TheThingsNetwork/lorawan-stack/pull/1429/commits/9ad18396e1ec45bf02b1acada4ca4bb62a194885

rvolosatovs:
It's not necessary in current implementation

rvolosatovs:
2019-10-08-18:28:40-screenshot
@htdvisser

rvolosatovs:
Rebased since there was a merge conflict

pkg/applicationserver/applicationserver.go
```diff
@@ -101,23 +102,23 @@ func New(c *component.Component, conf *Config) (as *ApplicationServer, err error
...
 	var interopCl InteropClient
 	if !conf.Interop.IsZero() {
-		var fallbackTLS *tls.Config
-		cTLS, err := c.GetTLSClientConfig(ctx)
-		if err != nil {
-			log.FromContext(ctx).WithError(err).Warn("Could not get fallback TLS config for interoperability")
-		} else {
-			fallbackTLS = cTLS
+		interopConf := conf.Interop.InteropClient
+		interopConf.GetFallbackTLSConfig = func(ctx context.Context) (*tls.Config, error) {
```
htdvisser: Can't you just write

```go
interopConf.GetFallbackTLSConfig = c.GetTLSClientConfig
```

pkg/config/shared.go
```diff
@@ -17,14 +17,20 @@ package config
 import (
 	"context"
 	"crypto/tls"
+	"io/ioutil"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	ttnblob "go.thethings.network/lorawan-stack/pkg/blob"
```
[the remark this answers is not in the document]
htdvisser: I see a 👍 but what is the action here?

pkg/config/shared.go
```diff
@@ -179,99 +188,164 @@ type BlobConfigAWS struct {
...
 
-// Store returns a frequencyplan.Store fwith a fetcher based on the configuration.
-// The order of precedence is Static, Directory and URL.
-// If neither Static, Directory nor a URL is set, this method returns nil, nil.
-func (c FrequencyPlansConfig) Store() (*frequencyplans.Store, error) {
-	var fetcher fetch.Interface
+// Fetcher returns a fetch.Interface based on the configuration.
+// The order of precedence is Static, Directory, URL and Blob.
+// If neither Static, Directory, URL nor a Blob is set, this method returns nil, nil.
+func (c FrequencyPlansConfig) Fetcher(ctx context.Context, blobConf BlobConfig) (fetch.Interface, error) {
```
htdvisser: I'm okay with merging this now, since we don't have a compatibility promise on our internal Go packages, but I still think it's weird to pass a blob config when in most cases (the default) we're not using blob.

pkg/fetch/bucket.go
```diff
@@ -41,7 +42,7 @@ func (f *bucketFetcher) File(pathElements ...string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	content, err := f.bucket.ReadAll(context.TODO(), rp)
+	content, err := f.bucket.ReadAll(f.context, rp)
```
htdvisser: I think it would be better if `ctx` was the first arg of the `File` func, but out of scope for this PR I guess.

pkg/applicationserver/applicationserver.go
rvolosatovs: Before rebase `GetTLSConfig` was used, which had a different signature

pkg/config/shared.go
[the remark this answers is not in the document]
rvolosatovs: I think the whole `shared.go` is a terrible idea, however we need to avoid import cycles and be consistent with what we already have.
Refactoring and changing this is out of scope of this PR.

pkg/config/shared.go
rvolosatovs: It's not pretty, but we need something working now.
I think that the whole `pkg/config` approach to solving our problem is incorrect and these kind of ugly workarounds are necessary due to that.

PR #1431, opened 2019-10-08
TLS configuration sources and key vault support

johanstokking:
#### Summary

TLS configuration sources

#### Changes

- Move reading certificate from file, watching it etc from `pkg/config` to `pkg/component` where it's solely used and to move logic away from `pkg/config`
- Put TLS certificate sources in one place and check that there's no ambiguity. So for example if both a local file and ACME are configured, an error is returned because we don't know what the user wanted
- Support loading certificates from the key vault
- `RootCA` and `InsecureSkipVerify` are now applied regardless of the certificate source. Before, if ACME was configured, these settings were not used, and I think that was wrong. ACME is for inbound; these two settings are for outbound TLS. Typically, when using ACME, people are using commercial certs and system root chain, but we cannot assume that is always the case
- Fix lint warning on component cancelation

#### Release Notes

- Fixed not respecting `RootCA` and `InsecureSkipVerify` TLS settings when ACME was configured for requesting TLS certificates

### DIFF
--- a/pkg/basicstation/cups/server.go
+++ b/pkg/basicstation/cups/server.go
@@ -75,7 +75,7 @@ func (conf ServerConfig) NewServer(c *component.Component, customOpts ...Option)
 			}),
 		)
 	}
-	if tlsConfig, err := c.GetTLSConfig(c.Context()); err == nil {
+	if tlsConfig, err := c.GetTLSServerConfig(c.Context()); err == nil {
 		opts = append(opts, WithTLSConfig(tlsConfig))
 	}
 	s := NewServer(c, append(opts, customOpts...)...)
--- a/pkg/component/tls.go
+++ b/pkg/component/tls.go
@@ -63,8 +63,8 @@ var (
 	errTLSKeyVaultID      = errors.DefineFailedPrecondition("tls_key_vault_id", "invalid TLS key vault ID")
 )
 
-// GetTLSConfig gets the component's TLS config and applies the given options.
-func (c *Component) GetTLSConfig(ctx context.Context, opts ...TLSConfigOption) (*tls.Config, error) {
+// GetTLSServerConfig gets the component's server TLS config and applies the given options.
+func (c *Component) GetTLSServerConfig(ctx context.Context, opts ...TLSConfigOption) (*tls.Config, error) {
 	var (
 		logger = log.FromContext(ctx)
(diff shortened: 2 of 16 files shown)


pkg/component/tls.go
```diff
@@ -50,31 +57,129 @@ func WithNextProtos(protos ...string) TLSConfigOption {
...
+	}
+	return res, nil
+}
+
+// GetTLSClientConfig gets the component's client TLS config and applies the given options.
+func (c *Component) GetTLSClientConfig(ctx context.Context, opts ...TLSConfigOption) (*tls.Config, error) {
+	conf := c.GetBaseConfig(ctx).TLS
+	res := &tls.Config{}
+	if conf.RootCA != "" {
+		pem, err := ioutil.ReadFile(conf.RootCA)
```
htdvisser: Should we perhaps do this only once? Otherwise we're going to do a lot of reads on that file.

pkg/component/tls.go
```diff
@@ -50,31 +57,129 @@ func WithNextProtos(protos ...string) TLSConfigOption {
...
+		}
+		res = &tls.Config{
+			GetCertificate: fn,
+		}
+	}
+	if res == nil {
+		return nil, errEmptyTLSConfig
+	}
+	res.MinVersion = tls.VersionTLS12
+	res.NextProtos = []string{"h2", "http/1.1"}
```
htdvisser: I think this should at some point be moved to the listeners, because the MQTT listeners for example should't offer http2 or http1.1 in ALPN

pkg/basicstation/cups/server.go
```diff
@@ -75,7 +75,7 @@ func (conf ServerConfig) NewServer(c *component.Component, customOpts ...Option)
 			}),
 		)
 	}
-	if tlsConfig, err := c.GetTLSConfig(c.Context()); err == nil {
+	if tlsConfig, err := c.GetTLSServerConfig(c.Context()); err == nil {
```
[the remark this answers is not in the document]
johanstokking: Sure?

pkg/component/tls.go
johanstokking: We might, but this PR is not changing current behavior which reads it on each use. Note that we're in a contextual configuration so we'd have to wire in the file fetcher. Let's not bother for now, but I'll file a `performance` issue.

Issue #1018, opened 2019-07-19
Show not found view on non-existing routes


bafonins:
#### Summary

We should inform the user when visiting non-existing routes of application/gateway sub views.
Currently, we handle only the root routes, i.e. `/console/{route}`

#### Steps to Reproduce

1. Go to `/console/applications/{existing-id}/does-not-exist`
2. See empty page

#### What do you see now?

<img width="994" alt="Screenshot 2019-07-19 at 11 47 30" src="https://user-images.githubusercontent.com/16374166/61526621-07033380-aa1b-11e9-9f0f-b5c66003d353.png">

#### What do you want to see instead?

Something like this:
<img width="578" alt="Screenshot 2019-07-19 at 11 41 49" src="https://user-images.githubusercontent.com/16374166/61526571-e9ce6500-aa1a-11e9-8d8e-28a83b46a557.png">

#### Environment

browser (console)

#### How do you propose to implement this?

Affected routes:

- [ ] /console/applications/{id}/does-not-exist
- [ ] /console/applications/{id}/payload-formatters/does-not-exist
- [ ] /console/applications/{id}/collaborators/does-not-exist
- [ ] /console/applications/{app-id}/devices/{dev-id}/does-not-exist
- [ ] /console/applications/{app-id}/devices/{dev-id}/payload-formatters/does-not-exist
- [ ] /console/applications/{app-id}/organizations/{dev-id}/does-not-exist
- [ ] /console/applications/{app-id}/gateways/{gtw-id}/does-not-exist
- [ ] /console/gateways/{id}/does-not-exist
- [ ] /console/gateways/{id}/collaborators/does-not-exist

TBA

#### Can you do this yourself and submit a Pull Request?

yes

bafonins:
Not sure exactly how to integrate it into error handling logic we have. I see two options:

Just throw an error on unknown route:
```
lib/components/not-found-route

export default () => {
  throw { statusCode: 404 }
}
```

```diff
// views/application
export default class Application extends React.Component {
  componentWillUnmount() {
    const { appId, stopStream } = this.props

    stopStream(appId)
  }

  render() {
    const { match, application, appId, env } = this.props

    return (
      <BreadcrumbView>
        <IntlHelmet titleTemplate={`%s - ${application.name || appId} - ${env.siteName}`} />
        <Switch>
          <Route exact path={`${match.path}`} component={ApplicationOverview} />
[...]

PR #1437, opened 2019-10-09
Handle non-existing routes in console

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1018

#### Changes

- Add `<NotFoundRoute />` to application/device/gateway/organization sub views

#### Notes for Reviewers

See discussion https://github.com/TheThingsNetwork/lorawan-stack/issues/1018

#### Release Notes

- Added handling of non-existing routes in console


bafonins:
> One other thing I noticed is that we sometimes inappropriately use non-exact routes (e.g here), which leads to sub-paths of these being matched as well, instead of 404ing. Should we fix this here or in a separate PR?

I guess we can fix this in this pr

PR #1443, opened 2019-10-09
KEK labeler escaping

johanstokking:
#### Summary

KEK labeler escaping

#### Changes

- Add support to escape in automatically generated KEK labels
- Minor housekeeping

### DIFF
--- a/pkg/config/shared.go
+++ b/pkg/config/shared.go
@@ -168,7 +168,7 @@ func (v KeyVault) KeyVault() (crypto.KeyVault, error) {
 		kv.ReplaceOldNew = []string{":", "_"}
 		return kv, nil
 	default:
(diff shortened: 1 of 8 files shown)


pkg/config/shared.go
```diff
@@ -163,9 +163,12 @@ type KeyVault struct {
 func (v KeyVault) KeyVault() (crypto.KeyVault, error) {
 	switch {
 	case v.Static != nil:
-		return cryptoutil.NewMemKeyVault(v.Static), nil
+		kv := cryptoutil.NewMemKeyVault(v.Static)
+		kv.Separator = ":"
```
johanstokking: This is needed for the configuration compatibility commitment; default separator is `/` to not collide with IPv6 colons in hosts

Issue #491, opened 2019-04-08
Treat unapproved users as unauthenticated on other entities


htdvisser:
**Summary:** 

We should prevent unapproved users from seeing information that is "public" in the network such as entity presence (does this gateway exist?), listing available frequency plans, listing available webhook formats, listing of OAuth client collaborators, payload encoding/decoding, etc.

**Why do we need this?**

The idea was that these RPCs are available to all users of the network, but right now we unintentionally allow users with pending accounts (in networks where admin approval is required) to call these RPCs, and in some places don't even check for sufficient rights.

**What is already there? What do you see now?**

Being part of the network is currently defined as "successfully authenticated with an API key or Access Token". This does not consider whether the user actually has an ACCEPTED state (in networks where approval is required) or with a valid email address (in networks where that is required).

Some low-confidentiality or (currently) unexposed RPC implementations don't even check anything (ListFrequencyPlans, UplinkMessageProcessor/DownlinkMessageProcessor, ApplicationWebhookRegistry.GetFormats)

**What is missing? What do you want to see?**

All RPCs should at least assert that the caller is a "member" of the network, where "membership" also considers the rules that define whether a user is "approved".

**How do you propose to implement this?**

Change `RequireAuthenticated` in pkg/identityserver/entity_access.go to also check that the user is in the "approved" state if `authInfo.GetEntityIdentifiers()` is UserIdentifiers.

**Can you do this yourself and submit a Pull Request?**

Yes

PR #1447, opened 2019-10-10
Treat unapproved users as unauthenticated

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/491

#### Changes

- Modify `RequireAuthenticated` to check if the user has `STATE_APPROVED`

#### Release Notes

- Changed authentication logic for users to treat unapproved users (ones without `STATE_APPROVED`) as unauthenticated on other entities

### DIFF
--- a/pkg/identityserver/entity_access.go
+++ b/pkg/identityserver/entity_access.go
@@ -39,6 +39,9 @@ var (
 	errInvalidAuthorization     = errors.DefineUnauthenticated("invalid_authorization", "invalid authorization")
 	errTokenNotFound            = errors.DefineUnauthenticated("token_not_found", "access token not found")
 	errTokenExpired             = errors.DefineUnauthenticated("token_expired", "access token expired")
+	errUserRejected             = errors.DefineUnauthenticated("user_rejected", "user account was rejected")
+	errUserRequested            = errors.DefineUnauthenticated("user_requested", "user account approval is pending")
+	errUserSuspended            = errors.DefineUnauthenticated("user_suspended", "user account was suspended")
 	errOAuthClientRejected      = errors.DefinePermissionDenied("oauth_client_rejected", "OAuth client was rejected")
 	errOAuthClientSuspended     = errors.DefinePermissionDenied("oauth_client_suspended", "OAuth client was suspended")
 )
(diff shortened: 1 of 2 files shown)


pkg/identityserver/entity_access.go
```diff
@@ -245,6 +246,25 @@ func (is *IdentityServer) RequireAuthenticated(ctx context.Context) error {
...
+	if userID := authInfo.GetEntityIdentifiers().GetUserIDs(); userID != nil {
+		err = is.withDatabase(ctx, func(db *gorm.DB) (err error) {
+			user, err := store.GetUserStore(db).GetUser(ctx, userID, &types.FieldMask{Paths: []string{
+				"state",
+			}})
+			if err != nil {
+				return err
+			}
+			if user.State != ttnpb.STATE_APPROVED {
+				return errUnnaprovedUser
```
htdvisser: I don't think we should use the term "unapproved" for all cases. It would be better to do a switch similar to the one in https://github.com/TheThingsNetwork/lorawan-stack/blob/4af802d1fb06e54e21de16fc7f34b6c829f5afe8/pkg/identityserver/entity_access.go#L216-L235

bafonins: Does this mean that users with `STATE_FLAGGED` should also pass the authentication check similar to `STATE_APPROVED` ?

pkg/identityserver/entity_access.go
```diff
@@ -39,6 +39,9 @@ var (
 	errInvalidAuthorization     = errors.DefineUnauthenticated("invalid_authorization", "invalid authorization")
 	errTokenNotFound            = errors.DefineUnauthenticated("token_not_found", "access token not found")
 	errTokenExpired             = errors.DefineUnauthenticated("token_expired", "access token expired")
+	errRejectedUser             = errors.DefineUnauthenticated("user_rejected", "user account approval is rejected")
+	errRequestedUser            = errors.DefineUnauthenticated("user_requested", "user account approval is pending")
+	errSuspendedUser            = errors.DefineUnauthenticated("user_suspended", "user account is suspended")
```
htdvisser: Please keep consistent with the OAuth errors below:

- `errUserXxx` instead of `errXxxUser`
- remove "approval" on the rejected error
- "is" -> "was" on rejected and suspended errors

htdvisser:
Please rebase on top of `master` and move the **Release Notes** section of this PR to `CHANGELOG.md`.

Issue #1450, opened 2019-10-10
Require configuration sources and providers


johanstokking:
#### Summary

Require configuration sources and providers.

#### Why do we need this?

To not determine them on set values which can become ambiguous

#### What is already there? What do you see now?

Fallback to set values

#### What is missing? What do you want to see?

Require them set in;
- Interop config file
- Frequency plans source
- Device repository source
- Key vault source
- TLS certificate source

#### How do you propose to implement this?

- Add defaults to the default configuration (or leave empty)
- Remove fallback in `pkg/config/shared.go` for the above config types

#### Can you do this yourself and submit a Pull Request?

Yes

PR #1451, opened 2019-10-10
Use explicit config sources and providers with fallback

johanstokking:
#### Summary

Use explicit config sources and providers. There's fallback to not break existing configuration.

References https://github.com/TheThingsNetwork/lorawan-stack/issues/1450

#### Changes

- Add configuration sources and providers setting
- Fallback to what we've been doing so far

#### Release Notes

- Supported specifying configuration sources for TLS, frequency plans, device repository, key vault and interop configuration explicitly

### DIFF
--- a/cmd/internal/shared/config.go
+++ b/cmd/internal/shared/config.go
@@ -107,7 +107,9 @@ var DefaultRightsConfig = config.Rights{
 }
 
 // DefaultKeyVaultConfig is the default config for key vaults.
-var DefaultKeyVaultConfig = config.KeyVault{}
+var DefaultKeyVaultConfig = config.KeyVault{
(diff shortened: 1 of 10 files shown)


cmd/internal/shared/config.go
```diff
@@ -107,7 +107,9 @@ var DefaultRightsConfig = config.Rights{
 }
 
 // DefaultKeyVaultConfig is the default config for key vaults.
-var DefaultKeyVaultConfig = config.KeyVault{}
+var DefaultKeyVaultConfig = config.KeyVault{
+	Provider: "static",
```
johanstokking: This is the only one we can do, because before this version there's only `static`

PR #1467, opened 2019-10-14
Publish builds to S3

htdvisser:
#### Summary

This PR adds a configuration to Goreleaser that will upload builds to S3.

#### Notes for Reviewers

- ACL setting is still private, will make public-read when this works well.
- Still need to set up credentials in Travis

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -162,6 +162,14 @@ brews:
       bin.install "ttn-lw-cli"
       libexec.install doc
 
+s3:
+  - bucket: {{ or (index .Env "AWS_BUCKET") "the-things-stack-releases" }}
+    ids:
+    - stack
+    - cli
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -162,6 +162,14 @@ brews:
       bin.install "ttn-lw-cli"
       libexec.install doc
 
+s3:
+  - bucket: {{ or (index .Env "AWS_BUCKET") "the-things-stack-releases" }}
+    ids:
+    - stack
+    - cli
+    region: {{ or (index .Env "AWS_REGION") "eu-west-1" }}
+    folder: "{{ .Version }}"
```
KrishnaIyer: Does it create the folder if it doesn't exist?

htdvisser: There is not actually such thing as a folder in s3. It's just a key/filename that contains slashes.

PR #1266, opened 2019-08-29
Add Semtech Join Server interoperability guide

rvolosatovs:
#### Summary

Closes #1250 

#### Changes

- Document interoperability repository
- Add an example of setting up interoperability with Semtech Join Server

#### Notes for Reviewers

~Still waiting on feedback from Semtech on the JS to use and the API - once that's done I will extend the guide with registering the NS and possibly change some of the configurations.
@johanstokking please take a look and confirm that this looks similar to what you expected
FTR, the authorization token is not the "actual" token issued to us by Semtech, but it's based on it.~

PR #1481, opened 2019-10-16
Configuration Reference starting point

htdvisser:
#### Summary

This Pull Request adds a starting point for the configuration reference section.

#### Changes

- Rewrote `doc/config.md` and moved it to `reference/configuration`
- Added documentation for generic server options.

#### Notes for Reviewers

I wanted the reference to be a bit more than just a copy-paste of the `--help` output, so I decided to split it into sections for each "level". Ordering isn't following anything specific. I wrote some introduction for each section and in some cases I also split on "sub-levels" and wrote some text in between, especially for the "selector" configs (source/backend/...).

My process for the other components (IS, NS, AS, JS, ...) would be:

1. Copy the output of `ttn-lw-stack -h | grep -- --is.` (for `is`)
2. Remove the flag types. Users can see those when running `ttn-lw-stack -h` or see examples with `ttn-lw-stack config` (with `--yml`/`--env`).
3. Remove the defaults. Same as 2.
4. Make the options a bullet list with `-`, add backticks around the option name, and a colon after it.
5. Add h2's between sub-sections (email, oauth, profile-picture, user-registration)
6. Write introduction for each section
7. See where we need extra split with notes
8. Re-order things if more "natural"


htdvisser:
@TheThingsNetwork/stack-developers it would be nice to get some feedback on this

furtiman:
It looks pretty good to me, did not get that much into the contents, I only feel that there might be too much of the word `Options`
Every section is `... Options`, and inside it has more `... Options`

I think there might be a way to generalise it somehow, but it's a small nitpick that really hit my sight

htdvisser:
Yes, that's a bit of a difficult one, we have the same issue in the API section. On one hand we don't want to stutter, but we also want the page `<title>` to contain "Options", otherwise search engines may think it's describing the component itself.

doc/content/reference/configuration/the-things-stack.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Global Options
+
+Under normal circumstances, only `info`, `warn` and `error` logs are printed to the console. For development, you may also want to see `debug` logs.
+
+- `log.level`: The minimum level log messages must have to be shown (default "info")
+
+## TLS Options
+
+The Things Stack serves several endpoints using TLS. TLS certificates can come from different sources.
```
rvolosatovs:
```suggestion
The Things Stack serves several endpoints using TLS. TLS configurations can come from different sources.
```

doc/content/reference/configuration/the-things-stack.md
```diff
@@ -0,0 +1,174 @@
...
+
+- `tls.source`: Source of the TLS certificate (file, acme, key-vault)
+
+When using the `file` source, the location of the certificate and key need to be configured.
+
+- `tls.certificate`: Location of TLS certificate
+- `tls.key`: Location of TLS private key
+
+When using ACME, certificates will be requested from Let's Encrypt (the default `endpoint`) and stored in the given `dir`.
```
rvolosatovs:
```suggestion
If `acme` is specified as `tls.source`, certificates will be requested from Let's Encrypt (the default `tls.acme.endpoint`) and stored in the given `tls.acme.dir`.
```

doc/content/reference/configuration/the-things-stack.md
```diff
@@ -0,0 +1,174 @@
...
+
+- `tls.acme.endpoint`: ACME endpoint
+- `tls.acme.dir`: Location of ACME storage directory
+- `tls.acme.email`: Email address to register with the ACME account
+- `tls.acme.hosts`: Hosts to enable automatic certificates for
+- `tls.acme.default-host`: Default host to assume for clients without SNI
+
+When using the key vault,
```
rvolosatovs: IDEM

doc/content/reference/configuration/the-things-stack.md
```diff
@@ -0,0 +1,174 @@
...
+
+- `log.level`: The minimum level log messages must have to be shown (default "info")
+
+## TLS Options
+
+The Things Stack serves several endpoints using TLS. TLS certificates can come from different sources.
+
+- `tls.source`: Source of the TLS certificate (file, acme, key-vault)
+
+When using the `file` source, the location of the certificate and key need to be configured.
```
rvolosatovs:
```suggestion
If `file` is specified as `tls.source`, the location of the certificate and key need to be configured.
```

doc/content/reference/configuration/the-things-stack.md
```diff
@@ -0,0 +1,174 @@
...
+- `http.redirect-to-tls`: Redirect HTTP requests to HTTPS
+
+The HTTP server serves static files for the web UI. If these files are not in the standard location, you may need to change the search path.
+
+- `http.static.mount`: Path on the server where static assets will be served
+- `http.static.search-path`: List of paths for finding the directory to serve static assets from
+
+## Interop Options
+
+The Things Stack supports interoperability with the LoRaWAN Backend Interfaces specification. The following options are used to configure the server for this.
```
rvolosatovs:
```suggestion
The Things Stack supports interoperability according to LoRaWAN Backend Interfaces specification. The following options are used to configure the server for this.
```

rvolosatovs: IMO this section is a bit confusing, since interoperability is also configured using `{as,ns}.interop.*`. I think we should specify explicitly here that this only concerns the server-side of things and clients are configured per-component

doc/content/reference/configuration/the-things-stack.md
```diff
@@ -0,0 +1,174 @@
...
+
+- `http.redirect-to-host`: Redirect all requests to one host
+- `http.redirect-to-tls`: Redirect HTTP requests to HTTPS
+
+The HTTP server serves static files for the web UI. If these files are not in the standard location, you may need to change the search path.
+
+- `http.static.mount`: Path on the server where static assets will be served
+- `http.static.search-path`: List of paths for finding the directory to serve static assets from
+
+## Interop Options
```
rvolosatovs:
```suggestion
## Interoperability Options
```

doc/content/reference/configuration/the-things-stack.md
rvolosatovs: Possibly with a reference to https://github.com/TheThingsNetwork/lorawan-stack/pull/1266

htdvisser: Let's merge this as-is and then update it in #1266

doc/content/reference/configuration/_index.md
```diff
@@ -0,0 +1,59 @@
...
+
+```yaml
+console:
+  ui:
+    canonical-url: "https://thethings.example.com/console"
+    # other console UI options
+  # other console options
+```
+
+You can specify the location of the YAML configuration file with the command-line flag `-c` or `--config`. If this flag is not present, The Things Stack will look for config files in the following locations:
```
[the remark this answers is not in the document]
htdvisser: That's below

doc/content/reference/configuration/_index.md
```diff
@@ -0,0 +1,59 @@
...
+
+The Things Stack first looks at command-line flags. This looks as follows:
+
+```bash
+$ ttn-lw-stack start console --console.ui.canonical-url "https://thethings.example.com/console"
+```
+
+### Environment variables
+
+Environment variables for configuration options are very similar to the command-line flags, except that they are in uppercase, and all separators (`.` or `-`) are replaced by underscores (`_`). Environment variables are also prefixed with `TTN_LW_`. In many cases you'll want to use a `.env` file that is loaded using the [`dotenv` command of direnv](https://direnv.net/man/direnv-stdlib.1.html) or the [`env_file` option of Docker Compose](https://docs.docker.com/compose/compose-file/#env_file). You can also `export` each environment variable, or run `export $(grep -v '^#' .env | xargs)` to export all variables in the `.env` file.
```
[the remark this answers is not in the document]
rvolosatovs: My 2. point still stands though. `direnv` is not supposed to be used in production scenarios, so IMO it should only be added as a note/remark

doc/content/reference/configuration/the-things-stack.md
htdvisser: Only certificates come from different sources, the other configuration (min TLS version, ciphers, etc.) is the same for all.

KrishnaIyer:
Personally this looks good to me 👍

PR #1482, opened 2019-10-16
Optimize downlink task timing and conditions

### LINKED ISSUE #936: Optimize downlink task starting times
#### Summary

Network Server should smarter choosing the starting time of downlink tasks

#### Why do we need this?

To ensure we don't miss downlink slots and not waste resources generating downlinks, when we know it's impossible to schedule

Apart from uplink-triggered downlink tasks, they're added at `time.Now()`

For Class B/C:
- Add downlink task after downlink timing-related MAC settings are changed(e.g. `status_count_periodicity`) via registry
- Compute precisely when should the downlink task be added if possible

When `lorawan_phy_version`, `frequency_plan_id`, `mac_state`, `mac_settings`, `last_dev_status_received_at` and `recent_uplinks` are available, NS can determine precisely when should the downlink task be added, if it's time-bound.

The downlink tasks adding after downlink queue update should be changed.

In fact, for all classes NS should only add a downlink task after an uplink, if there's application downlink that can be scheduled or MAC commands to schedule.
After the app queue or the downlink-related device settings are updated, NS shall look at the proto to determine if a downlink can be scheduled and depending on that, add a downlink task.

yes

rvolosatovs:
#### Summary

Closes #936 

#### Changes

- Instead of just blindly adding downlink tasks whenever something *may* trigger a downlink, calculate if and when a downlink is required and schedule accordingly
- Ensure confirmed class C downlinks are allowed before timeout expires if an uplink is received
- Time downlink task relative to RX2 instead of RX1 if the latter is expired
- Use (approximate) transmission time for determining whether or not `DevStatusReq` should be scheduled instead of using generation time(e.g. if a class A downlink is generated 30 sec before RX1 starts and `DevStatusReq` is not ready to be sent yet,  NS can send it regardless, if `DevStatusReq` will be ready at RX1.)
- Downlink queue pushes/replaces and device registry *may* trigger a class A downlink task if there exists an unanswered uplink, for which either RX1 or RX2 is not expired.

#### Notes for Reviewers

Note that for class A devices a downlink task will not be added after an uplink if there's no MAC and no application downlink to send.

Tested on metal in EU868 OTAA flow using ff1705 in classes A and C

>NOTE: This PR contains many changed LoC, big part of this is `TestNeeds*` tests(I was trying to decrease the coverage drop).
Please keep in mind that main goals of this PR are:
> 1. Limit unnecessary registry transactions(e.g. processing a downlink task when no downlink available)
> 2. Make sure downlinks are scheduled efficiently

### DIFF
--- a/pkg/networkserver/downlink.go
+++ b/pkg/networkserver/downlink.go
@@ -115,6 +115,14 @@ func nextConfirmedClassCDownlinkAt(dev *ttnpb.EndDevice, defaults ttnpb.MACSetti
 	if dev.GetMACState().GetLastConfirmedDownlinkAt() == nil {
 		return time.Time{}
 	}
+	if dev.GetMACState().GetRxWindowsAvailable() {
+		return time.Time{}
+	}
+	if len(dev.RecentUplinks) > 0 {
+		if dev.RecentUplinks[len(dev.RecentUplinks)-1].ReceivedAt.After(*dev.MACState.LastConfirmedDownlinkAt) {
+			return time.Time{}
+		}
+	}
 	return dev.MACState.LastConfirmedDownlinkAt.Add(deviceClassCTimeout(dev, defaults))
 }
 
--- a/pkg/networkserver/downlink.go
+++ b/pkg/networkserver/downlink.go
@@ -129,7 +129,7 @@ func nextConfirmedClassCDownlinkAt(dev *ttnpb.EndDevice, defaults ttnpb.MACSetti
 // For example, a sequence of 'NewChannel' MAC commands could be generated for a
 // device operating in a region where a fixed channel plan is defined in case
 // dev.MACState.CurrentParameters.Channels is not equal to dev.MACState.DesiredParameters.Channels.
-func (ns *NetworkServer) generateDownlink(ctx context.Context, dev *ttnpb.EndDevice, phy band.Band, class ttnpb.Class, maxDownLen, maxUpLen uint16) (*generatedDownlink, generateDownlinkState, error) {
+func (ns *NetworkServer) generateDownlink(ctx context.Context, dev *ttnpb.EndDevice, phy band.Band, class ttnpb.Class, scheduleAt time.Time, maxDownLen, maxUpLen uint16) (*generatedDownlink, generateDownlinkState, error) {
 	if dev.MACState == nil {
 		return nil, generateDownlinkState{}, errUnknownMACState
 	}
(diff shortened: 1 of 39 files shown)


pkg/networkserver/downlink.go
```diff
@@ -1203,20 +1186,24 @@ func (ns *NetworkServer) processDownlinkTask(ctx context.Context) error {
...
-						nextDownlinkAt = a.NextDownlinkAt
-					}
 					if a.Scheduled {
+						t, hasNext := nextDataDownlinkAt(ctx, dev, phy, ns.defaultMACSettings)
+						if hasNext {
+							earliestAt := a.DownAt.Add(dev.MACState.CurrentParameters.Rx1Delay.Duration())
+							if t.Before(earliestAt) {
+								nextDownlinkAt = earliestAt
+							} else {
+								nextDownlinkAt = t
```
johanstokking: This is intended as `nextDownlinkAt = max(t, earliestAt)` right?

pkg/networkserver/grpc_deviceregistry.go
```diff
@@ -169,24 +169,35 @@ func (ns *NetworkServer) Set(ctx context.Context, req *ttnpb.SetEndDeviceRequest
...
-				"queued_application_downlinks",
-			)
+		if p == "lorawan_phy_version" ||
+			p == "frequency_plan_id" ||
+			p == "mac_settings" ||
+			p == "mac_state" ||
+			p == "session" ||
+			strings.HasPrefix(p, "mac_settings.") ||
+			strings.HasPrefix(p, "mac_state.") ||
+			strings.HasPrefix(p, "session.") {
```
johanstokking: Use `ttnpb.HasAnyField()`

pkg/networkserver/grpc_deviceregistry.go
```diff
@@ -340,13 +345,28 @@ func (ns *NetworkServer) Set(ctx context.Context, req *ttnpb.SetEndDeviceRequest
...
+
+	if !needsDownlinkCheck {
+		return dev, nil
+	}
+
+	var downAt time.Time
+	_, phy, err := getDeviceBandVersion(dev, ns.FrequencyPlans)
+	if err != nil {
+		log.FromContext(ctx).WithError(err).Warn("Failed to determine device band")
+		downAt = timeNow().UTC()
```
johanstokking: Can this get us in an endless loop?

pkg/networkserver/downlink.go
rvolosatovs: yes

pkg/networkserver/grpc_deviceregistry.go
rvolosatovs: What this does is just passes on the responsibility to figure this out to the downlink processor.
In case the `processDownlinkTask` fails to determine the band as well, a retry is performed after `downlinkRetryInterval`(currently set to 1 second).
So effectively, in case it's not possible to determine the frequency plan, the NS will retry every second until the frequency plan can be determined.

johanstokking: Yeah that's what I meant with an endless loop. I think we should avoid that, and instead restart the downlink tasks when the frequency plan is set (like you do now)

rvolosatovs: That means that if e.g. github is unreachable(or throwing unicorns) and frequency plan could not be fetched we would stop scheduling downlinks unless something triggers it.
I think that's not the right solution and can have bad consequences. Consider, for example, class C multicast device with a bunch of downlinks queued. Scheduling would only resume if a new downlink is pushed or something downlink-related is changed in the device proto via registry.
What we could do, though, would be increasing the retry interval.
Also note that it's not possible to create a device without frequency plan and we validate that it's a valid frequency plan on creation, so the value should not be the problem if errors happen here.

johanstokking: OK, I see about this. I think we should have a backoff though. Frequency plans are subject to change and indeed availability may be limited when using an external source.

rvolosatovs: I'll create an issue about it - this is out of scope of this PR

rvolosatovs: https://github.com/TheThingsNetwork/lorawan-stack/issues/1503

PR #1483, opened 2019-10-17
Documentation Consistency and Guidelinelines proposal

htdvisser:
#### Summary

This PR makes our Markdown content more consistent. I also have a proposal for guidelines to discuss:

#### Proposal

- The title of a doc page is rendered by Hugo as a h1, don't add an extra one.
- A doc page starts with an introduction, and then the first heading. Make use of `<!--more-->` to define the summary that Hugo uses.
- Since the title is h1, everything in the content is h2-...
- Paragraphs typically consist of at least two sentences.
- Use an empty line between all blocks (headings, paragraphs, lists, ...).
- Prefer text over bullet lists or enumerations. For bullets, use `-`, for enumerations `1.`.
- Write guides as a goal-oriented journey.
- Unless already clear from context, use a clearer term than "user", especially if there are multiple kinds (network operator, gateway owner, application developer, ...).
- The user does not have a gender (use they/them/their).
- Taking screenshots as follows:
  - In Chrome:
    1. Activate Developer Tools
    2. Toggle Device Toolbar (`⌘⇧M`)
    3. In the Device Toolbar: select "Laptop with HiDPI screen" (add it if not already there)
    4. In the Device Toolbar: `⠇` -> Capture Screenshot
  - In Firefox:
    1. Enter Responsive Design Mode
    2. In the Device Toolbar: select "Laptop with HiDPI screen" (add it if not already there)
    3. In the Device Toolbar: Take a screenshot of the viewport
- Use `**Strong**` when referring to buttons in the Console
- Use fenced code blocks with a language:
  - `bash` for lists of environment variables: `SOME_ENV="value"`.
  - `bash` for CLI examples. Prefix commands with `$ `. Wrap strings with `""` (except when JSON, which uses double).
  - Wrap large CLI output with `<details><summary>Show CLI output</summary> ... output here ... </details>`.
  - `yaml` (not `yml`) for YAML. Wrap strings with single quotes `''` (because of frequent Go templates that use double).


rvolosatovs:
Can you add these guidelines as a file somewhere?(I suppose it belongs in one of `CONTRIBUTING.md` or `DEVELOPMENT.md`)

htdvisser:
Yes, as soon as we agree on the guidelines, I'll add it to the other code style guidelines.

doc/content/guides/getting-started/configuration.md
```diff
@@ -44,103 +44,104 @@ Before we go into the details of The Things Stack's configuration, we'll take a
...
+```
+
+Next, we'll have a look at the configuration options for your private deployment. We'll set these options in the `.env` file that is referenced by the `env_file` option of the `stack` service in `docker-compose.yml`.
 
 First we'll make sure that The Things Stack uses the correct databases.
 
-```
-TTN_LW_IS_DATABASE_URI=postgres://root@cockroach:26257/ttn_lorawan?sslmode=disable
-TTN_LW_REDIS_ADDRESS=redis:6379
+```bash
```
KrishnaIyer: This is still in yaml right? Or do you mean this is `env` stuff that can be injected in many ways?
Same with the other occurences

doc/themes/the-things-stack/assets/css/theme.scss
```diff
@@ -180,3 +180,5 @@ html.has-navbar-fixed-top {
     min-height: calc(100vh - #{$navbar-height});
   }
 }
+
+@import "highlight.scss";
```
KrishnaIyer: Is this and the corresponding file related to the PR? It's completely fine to keep it. Just checking.

doc/content/guides/getting-started/configuration.md
htdvisser: It can be injected in multiple ways indeed.

doc/themes/the-things-stack/assets/css/theme.scss
htdvisser: Good catch. That commit was supposed to be in a future PR.

johanstokking:
Proposing also:

- No shortening, i.e. write "it is" instead of "it's"
- Always use literals for LoRaWAN terms, like `DevEUI`, `JoinEUI`, and have plural as `JoinEUI`s

johanstokking:
And:

- Explicitly call this product The Things Stack. Not "the stack" etc

PR #1486, opened 2019-10-17
Add cache config to service base

KrishnaIyer:
#### Summary

Add a Cache to the `ServiceBase`. This enables the configuration of a separate redis for the IS cache.

#### Changes

- Add a cache (currently Redis) to `ServiceBase`.
- Use this in the IS Cache setting.

#### Notes for Reviewers

- Either this file has not been gofumpted or my tooling has issues.
- This kinda breaks config right?

### DIFF
--- a/cmd/ttn-lw-stack/commands/start.go
+++ b/cmd/ttn-lw-stack/commands/start.go
@@ -45,231 +45,229 @@ import (
 	"go.thethings.network/lorawan-stack/pkg/web"
 )
 
-var (
-	errUnknownComponent = errors.DefineInvalidArgument("unknown_component", "unknown component `{component}`")
-)
+var errUnknownComponent = errors.DefineInvalidArgument("unknown_component", "unknown component `{component}`")
 
-var (
-	startCommand = &cobra.Command{
-		Use:   "start [is|gs|ns|as|js|console|gcs|dtc|all]... [flags]",
-		Short: "Start The Things Stack",
-		RunE: func(cmd *cobra.Command, args []string) error {
-			var start struct {
-				IdentityServer             bool
-				GatewayServer              bool
-				NetworkServer              bool
-				ApplicationServer          bool
-				JoinServer                 bool
-				Console                    bool
-				GatewayConfigurationServer bool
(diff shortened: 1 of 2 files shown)


pkg/config/shared.go
```diff
@@ -139,6 +139,12 @@ type CloudEvents struct {
 	SubscribeURL string `name:"subscribe-url" description:"URL for the subscription to receiving events"`
 }
 
+// Cache represents configuration for a caching system.
+type Cache struct {
+	Service string `name:"type" description:"Service used for caching (redis)"`
```
htdvisser: `name` should probably be `service`

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -45,231 +45,229 @@ import (
...
-
-			if start.IdentityServer || startDefault {
-				logger.Info("Setting up Identity Server")
-				is, err := identityserver.New(c, &config.IS)
-				if err != nil {
-					return shared.ErrInitializeIdentityServer.WithCause(err)
-				}
+			if config.Cache.Service == "redis" {
 				is.SetRedisCache(redis.New(&redis.Config{
 					Redis:     config.Redis,
```
htdvisser: Shouldn't this then be `config.Cache.Redis` instead of `config.Redis`?

KrishnaIyer: Indeed. Stupid me copied the wrong code while rebasing.

pkg/config/shared.go
```diff
@@ -400,6 +406,7 @@ func (c InteropClient) Fetcher(ctx context.Context, blobConf BlobConfig) (fetch.
 type ServiceBase struct {
 	Base             `name:",squash"`
 	Cluster          Cluster                `name:"cluster"`
+	Cache            Cache                  `name:"Cache"`
```
htdvisser: @KrishnaIyer I overlooked this during review. The `name` should be lowercase.

KrishnaIyer: Indeed... :(

PR #1498, opened 2019-10-22
Ignore local build binaries

KrishnaIyer:
#### Summary

Ignore local binaries in the code base 

#### Changes

- Ignore `ttn-lw-stack`, `ttn-lw-cli`  in the local repo similar to the `public` folder.

#### Notes for Reviewers

- Corresponding TTI PR to follow

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,3 +110,7 @@ key.pem
 
 # Release files
 /snapcraft.login
+
+# Default builds
+ttn-lw-stack
+ttn-lw-cli


johanstokking:
We have no tooling that creates these files so I don't think we should be adding these

KrishnaIyer:
That's true but just using `go build ./cmd/ttn-lw-{stack/cli}` to build binaries for is very common practice for local testing.

johanstokking:
> That's true but just using `go build ./cmd/ttn-lw-{stack/cli}` to build binaries for is very common practice for local testing.

Prefer `go install` and `go run` for that anyway, or specifying `-o`. These are not default builds caused by our tooling. Consider also using `.git/info/exclude` that stays local.

PR #1502, opened 2019-10-23
Allow read-only fields in update mask if they don't change

### LINKED ISSUE #1493: Allow read-only field masks for update operations if the value does not change
#### Summary
When querying the RPCs through the http bridge, using read-only field masks in write operations will error out, even when the value remains the same. It would be better to ignore these field masks then rather than returning a validation error.
The console would benefit from such behavior, as we sometimes want to *upsert* entities, using the same request. This is currently not possible, as the `ids` field mask cannot be used when the entry already exists, so we need to use two different requests to achieve upserting.

#### Steps to Reproduce

Try running the following cURL (given that the `test-app` application exists and the `test-device` does not)
```shell
curl -X POST \
  http://localhost:1885/api/v3/js/applications/test-app/devices/test-device \
  -H 'Accept: application/json, text/plain, */*' \
  -H 'Accept-Encoding: gzip, deflate' \
  -H 'Authorization: Bearer $access_token' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Length: 784' \
  -H 'Content-Type: application/json;charset=UTF-8' \
  -H 'Host: localhost:1885' \
  -H 'cache-control: no-cache' \
  -d '{
  "end_device":{
    "root_keys":{
      "app_key":{
        "key":"665A771492D6E16FD0E271AA1D9E6A06"
      },
      "nwk_key":{
        "key":"F5B9273CBF3227041FA66ADB20569BF8"
      }
    },
    "lorawan_version":"1.0.0",
    "resets_join_nonces":false,
    "supports_class_c":false,
    "frequency_plan_id":"EU_863_870",
    "lorawan_phy_version":"1.0.0",
    "ids":{
      "device_id":"test-device",
      "application_ids":{
        "application_id":"test-app"
      },
      "join_eui":"70B3D57ED0000000",
      "dev_eui":"0004A310001FF9E2"
    },
    "supports_join":true
  },
  "field_mask":{
    "paths":[
      "root_keys.app_key.key",
      "root_keys.nwk_key.key",
      "resets_join_nonces",
      "ids.device_id",
      "ids.application_ids.application_id"
    ]
  }
}'
```
It will run through the first time, but error our the second time, because the `ids` field masks won't pass the validator. The requests targets the Join Server, but this holds true for all end device registries.

#### What do you see now?
Errors when using field masks of read-only props, even when the values would not change.

Such field masks being ignored without error.

#### Environment
TTS v3.2.2

🤷‍♂, hopefully not a big deal

@johanstokking / @rvolosatovs

johanstokking:
#### Summary

Closes #1493 

#### Changes

- Replace merciless `ProhibitFields()` by checking if there's a change

### DIFF
--- a/pkg/applicationserver/io/packages/redis/registry.go
+++ b/pkg/applicationserver/io/packages/redis/registry.go
@@ -31,6 +31,7 @@ import (
 var (
 	errInvalidFieldmask   = errors.DefineInvalidArgument("invalid_fieldmask", "invalid fieldmask")
 	errInvalidIdentifiers = errors.DefineInvalidArgument("invalid_identifiers", "invalid identifiers")
+	errReadOnlyField      = errors.DefineInvalidArgument("read_only_field", "read-only field `{field}`")
 )
 
 // appendImplicitAssociationGetPaths appends implicit ttnpb.ApplicationPackageAssociation get paths to paths.
@@ -215,12 +216,14 @@ func (r ApplicationPackagesRegistry) Set(ctx context.Context, ids ttnpb.Applicat
 					return errInvalidIdentifiers
 				}
 			} else {
-				if err := ttnpb.ProhibitFields(sets,
-					"ids.end_device_ids.application_ids",
-					"ids.end_device_ids.device_id",
-					"ids.f_port",
-				); err != nil {
-					return errInvalidFieldmask.WithCause(err)
+				if ttnpb.HasAnyField(sets, "ids.end_device_ids.application_ids.application_id") && pb.ApplicationID != stored.ApplicationID {
+					return errReadOnlyField.WithAttributes("field", "ids.end_device_ids.application_ids.application_id")
+				}
+				if ttnpb.HasAnyField(sets, "ids.end_device_ids.device_id") && pb.DeviceID != stored.DeviceID {
+					return errReadOnlyField.WithAttributes("field", "ids.end_device_ids.device_id")
+				}
+				if ttnpb.HasAnyField(sets, "ids.f_port") && pb.FPort != stored.FPort {
+					return errReadOnlyField.WithAttributes("field", "ids.f_port")
 				}
 				if err := cmd.ScanProto(updated); err != nil {
 					return err
--- a/pkg/applicationserver/io/pubsub/redis/registry.go
+++ b/pkg/applicationserver/io/pubsub/redis/registry.go
@@ -30,6 +30,7 @@ import (
 var (
 	errInvalidFieldmask   = errors.DefineInvalidArgument("invalid_fieldmask", "invalid fieldmask")
 	errInvalidIdentifiers = errors.DefineInvalidArgument("invalid_identifiers", "invalid identifiers")
+	errReadOnlyField      = errors.DefineInvalidArgument("read_only_field", "read-only field `{field}`")
 )
 
 // appendImplicitPubSubGetPaths appends implicit ttnpb.ApplicationPubSub get paths to paths.
@@ -224,11 +225,11 @@ func (r PubSubRegistry) Set(ctx context.Context, ids ttnpb.ApplicationPubSubIden
 					return errInvalidIdentifiers
 				}
 			} else {
-				if err := ttnpb.ProhibitFields(sets,
-					"ids.application_ids",
-					"ids.pub_sub_id",
-				); err != nil {
-					return errInvalidFieldmask.WithCause(err)
+				if ttnpb.HasAnyField(sets, "ids.application_ids.application_id") && pb.ApplicationID != stored.ApplicationID {
+					return errReadOnlyField.WithAttributes("field", "ids.application_ids.application_id")
+				}
+				if ttnpb.HasAnyField(sets, "ids.pub_sub_id") && pb.PubSubID != stored.PubSubID {
+					return errReadOnlyField.WithAttributes("field", "ids.pub_sub_id")
 				}
 				if err := cmd.ScanProto(updated); err != nil {
 					return err
--- a/pkg/applicationserver/redis/registry.go
+++ b/pkg/applicationserver/redis/registry.go
@@ -31,6 +31,7 @@ var (
 	errInvalidFieldmask     = errors.DefineInvalidArgument("invalid_fieldmask", "invalid fieldmask")
 	errInvalidIdentifiers   = errors.DefineInvalidArgument("invalid_identifiers", "invalid identifiers")
 	errDuplicateIdentifiers = errors.DefineAlreadyExists("duplicate_identifiers", "duplicate identifiers")
+	errReadOnlyField        = errors.DefineInvalidArgument("read_only_field", "read-only field `{field}`")
 )
 
 // appendImplicitDeviceGetPaths appends implicit ttnpb.EndDevice get paths to paths.
@@ -75,6 +76,13 @@ func (r *DeviceRegistry) Get(ctx context.Context, ids ttnpb.EndDeviceIdentifiers
 		return nil, err
 	}
 	return applyDeviceFieldMask(nil, pb, appendImplicitDeviceGetPaths(paths...)...)
+}
+
+func equalEUI64(x, y *types.EUI64) bool {
+	if x == nil || y == nil {
+		return x == y
+	}
+	return x.Equal(*y)
 }
 
 // Set creates, updates or deletes the end device by its identifiers.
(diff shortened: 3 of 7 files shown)


pkg/applicationserver/io/packages/redis/registry.go
```diff
@@ -215,16 +216,15 @@ func (r ApplicationPackagesRegistry) Set(ctx context.Context, ids ttnpb.Applicat
...
-					"ids.end_device_ids.application_ids",
-					"ids.end_device_ids.device_id",
-					"ids.f_port",
-				); err != nil {
-					return errInvalidFieldmask.WithCause(err)
-				}
 				if err := cmd.ScanProto(updated); err != nil {
 					return err
 				}
+				if ttnpb.HasAnyField(sets, "ids.end_device_ids") && !pb.EndDeviceIdentifiers.Equal(updated.EndDeviceIdentifiers) {
```
rvolosatovs: This incorrect `ttnpb.HasAnyField(sets, "ids.end_device_ids")` returns `false` if e.g. `"ids.end_device_ids.application_ids"` is specified.

Try for yourself:
```
diff --git a/pkg/ttnpb/fieldmask_utils_test.go b/pkg/ttnpb/fieldmask_utils_test.go
index 887f61a5c..b448d670e 100644
--- a/pkg/ttnpb/fieldmask_utils_test.go
+++ b/pkg/ttnpb/fieldmask_utils_test.go
@@ -94,6 +94,7 @@ func TestHasAnyField(t *testing.T) {
 		"a",
 		"b.c",
 		"d",
+		"e.f.g",
 	}
 	a.So(ttnpb.HasAnyField(requested, "x", "a"), should.BeTrue)
 	a.So(ttnpb.HasAnyField(requested, "x.y", "b"), should.BeFalse)
@@ -101,6 +102,7 @@ func TestHasAnyField(t *testing.T) {
 	a.So(ttnpb.HasAnyField(requested, "x", "b.c.d"), should.BeTrue)
 	a.So(ttnpb.HasAnyField(requested, "d"), should.BeTrue)
 	a.So(ttnpb.HasAnyField(requested, "d.e", "b"), should.BeTrue)
+	a.So(ttnpb.HasAnyField(requested, "e.f"), should.BeTrue)
 }
 
 func TestFlattenPaths(t *testing.T) {
```

rvolosatovs: `pb.EndDeviceIdentifiers.Equal` uses reflection *and* we don't care about anything else besides appID/devID.
Why not just use these fields explicitly both in paths and in comparison?

pkg/applicationserver/io/pubsub/redis/registry.go
```diff
@@ -224,15 +225,15 @@ func (r PubSubRegistry) Set(ctx context.Context, ids ttnpb.ApplicationPubSubIden
...
-				if err := ttnpb.ProhibitFields(sets,
-					"ids.application_ids",
-					"ids.pub_sub_id",
-				); err != nil {
-					return errInvalidFieldmask.WithCause(err)
-				}
 				if err := cmd.ScanProto(updated); err != nil {
 					return err
 				}
+				if ttnpb.HasAnyField(sets, "ids.application_ids") && !pb.ApplicationIdentifiers.Equal(updated.ApplicationIdentifiers) {
```
rvolosatovs: Can't you just `!=` here?

pkg/applicationserver/redis/registry.go
```diff
@@ -174,18 +175,21 @@ func (r *DeviceRegistry) Set(ctx context.Context, ids ttnpb.EndDeviceIdentifiers
...
 				if err := cmd.ScanProto(updated); err != nil {
 					return err
 				}
+				if ttnpb.HasAnyField(sets, "ids.application_ids") && !pb.ApplicationIdentifiers.Equal(updated.ApplicationIdentifiers) {
+					return errReadOnlyField.WithAttributes("field", "ids.application_ids")
+				}
+				if ttnpb.HasAnyField(sets, "ids.device_id") && pb.DeviceID != updated.DeviceID {
+					return errReadOnlyField.WithAttributes("field", "ids.device_id")
+				}
+				if ttnpb.HasAnyField(sets, "ids.join_eui") && (pb.JoinEUI == nil || !updated.JoinEUI.Equal(*pb.JoinEUI)) {
```
rvolosatovs: `JoinEUI` can be `nil`, if it's not stored. `pb.JoinEUI == nil` condition here is incorrect

pkg/joinserver/redis/registry.go
```diff
@@ -214,19 +215,27 @@ func (r *DeviceRegistry) set(tx *redis.Tx, uid string, gets []string, f func(pb
...
 			if err := cmd.ScanProto(updated); err != nil {
 				return nil, err
 			}
+			if ttnpb.HasAnyField(sets, "ids.application_ids") && !pb.ApplicationIdentifiers.Equal(updated.ApplicationIdentifiers) {
+				return nil, errReadOnlyField.WithAttributes("field", "ids.application_ids")
+			}
+			if ttnpb.HasAnyField(sets, "ids.device_id") && pb.DeviceID != updated.DeviceID {
+				return nil, errReadOnlyField.WithAttributes("field", "ids.device_id")
+			}
+			if ttnpb.HasAnyField(sets, "ids.join_eui") && (pb.JoinEUI == nil || !updated.JoinEUI.Equal(*pb.JoinEUI)) {
```
rvolosatovs: Assuming you make the change above it's better to be consistent with AS registry and just check if these are equal without the `nil` check

pkg/networkserver/redis/registry.go
```diff
@@ -242,17 +243,21 @@ func (r *DeviceRegistry) SetByID(ctx context.Context, appID ttnpb.ApplicationIde
...
 				if err := cmd.ScanProto(updated); err != nil {
 					return err
 				}
+				if ttnpb.HasAnyField(sets, "ids.application_ids") && !pb.ApplicationIdentifiers.Equal(updated.ApplicationIdentifiers) {
+					return errReadOnlyField.WithAttributes("field", "ids.application_ids")
+				}
+				if ttnpb.HasAnyField(sets, "ids.device_id") && pb.DeviceID != updated.DeviceID {
+					return errReadOnlyField.WithAttributes("field", "ids.device_id")
+				}
+				if ttnpb.HasAnyField(sets, "ids.join_eui") && (pb.JoinEUI == nil || !updated.JoinEUI.Equal(*pb.JoinEUI)) {
```
rvolosatovs: IDEM

pkg/networkserver/redis/registry.go
```diff
@@ -242,17 +243,21 @@ func (r *DeviceRegistry) SetByID(ctx context.Context, appID ttnpb.ApplicationIde
...
-					"ids.dev_eui",
-					"ids.device_id",
-					"ids.join_eui",
-				); err != nil {
-					return errInvalidFieldmask.WithCause(err)
-				}
 				if err := cmd.ScanProto(updated); err != nil {
 					return err
 				}
+				if ttnpb.HasAnyField(sets, "ids.application_ids") && !pb.ApplicationIdentifiers.Equal(updated.ApplicationIdentifiers) {
```
rvolosatovs: IDEM

pkg/applicationserver/io/packages/redis/registry.go
```diff
@@ -215,16 +216,18 @@ func (r ApplicationPackagesRegistry) Set(ctx context.Context, ids ttnpb.Applicat
...
-					"ids.end_device_ids.application_ids",
-					"ids.end_device_ids.device_id",
-					"ids.f_port",
-				); err != nil {
-					return errInvalidFieldmask.WithCause(err)
-				}
 				if err := cmd.ScanProto(updated); err != nil {
 					return err
 				}
+				if ttnpb.HasAnyField(sets, "ids.end_device_ids.application_ids.application_id") && pb.ApplicationIdentifiers.ApplicationID != updated.ApplicationID {
```
rvolosatovs: `updated` -> `stored`

pkg/applicationserver/io/packages/redis/registry.go
```diff
@@ -215,16 +216,18 @@ func (r ApplicationPackagesRegistry) Set(ctx context.Context, ids ttnpb.Applicat
...
+					return errReadOnlyField.WithAttributes("field", "ids.end_device_ids.application_ids.application_id")
+				}
+				if ttnpb.HasAnyField(sets, "ids.end_device_ids.device_id") && pb.EndDeviceIdentifiers.DeviceID != updated.EndDeviceIdentifiers.DeviceID {
```
rvolosatovs: IDEM

pkg/applicationserver/io/packages/redis/registry.go
rvolosatovs: - `pb` is the proto being set(input)
- `stored` is the proto currently in the registry
- `updated` is the result of applying `pb` with specified fieldmask to `stored`

PR #1511, opened 2019-10-24
Support disabling channels in Network Server

### LINKED ISSUE #1499: Network Server should disable channels using NewChannelReq
#### Summary

`NewChannelReq` supports sending `Freq == 0` to disable a channel

#### Steps to Reproduce

1. Look at the implementation

#### What do you see now?

`NewChannelReq` implementation does not supports sending `Freq == 0` to disable a channel

Support sending `Freq == 0` to disable a channel

If length of desired channels is not equal to length of current channels we should send a `NewChannelReq`. If length of desired channels is higher than length of current channels, we need to disable `len(dev.MACState.CurrentParameters.Channels)-len(dev.MACState.DesiredParameters.Channels)` channels

Note that `LinkADR` implementaton should be alligned with this

yes

rvolosatovs:
#### Summary

Closes #1499 

#### Changes

- Support disabling channels via `NewChannelReq`
- Account for channels possibly being `nil` in Network Server
- Fix `DLChannelReq` scheduling conditions

### DIFF
--- a/pkg/networkserver/downlink.go
+++ b/pkg/networkserver/downlink.go
@@ -780,8 +780,7 @@ func txRequestFromUplink(phy band.Band, macState *ttnpb.MACState, rx1, rx2 bool,
 			return nil, err
 		}
 		if uint(rx1ChIdx) >= uint(len(macState.CurrentParameters.Channels)) ||
-			macState.CurrentParameters.Channels[int(rx1ChIdx)] == nil ||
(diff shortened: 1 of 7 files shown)


htdvisser:
Shouldn't this have an entry in `CHANGELOG.md`?

rvolosatovs:
Users should not care about details of this.
We can add something like `Various MAC-layer operation improvements` or `Improved NewChannel and DLChannel operation` if you think it's necessary.

PR #1517, opened 2019-10-29
Add connection indicator to gateways table

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1155
Screenshot 2019-10-29 at 13 37 16

#### Changes

- Extend gateways logic to fetch stats
- Extend gateways table to display the status indicator


pkg/webui/console/store/middleware/logics/gateways.js
```diff
@@ -78,8 +79,24 @@ const getGatewaysLogic = createRequestLogic({
         )
       : await api.gateways.list({ page, limit }, selectors)
 
+    const entities = await Promise.all(
```
kschiffer: Can you make the decoration with the status value optional via a flag in `action.meta`?
I think there will be situations where we want to avoid the extra request.

PR #1522, opened 2019-10-30
Support LoRa Alliance TR005 draft 3 QR code format

johanstokking:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/1757

#### Changes

- Support TR005 draft 3

### DIFF
--- a/pkg/qrcode/lora_alliance_tr005d3.go
+++ b/pkg/qrcode/lora_alliance_tr005d3.go
@@ -1,0 +1,175 @@
+// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package qrcode
+
+import (
+	"bytes"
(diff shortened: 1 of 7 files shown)


pkg/qrcode/lora_alliance_tr005d3.go
```diff
@@ -0,0 +1,175 @@
...
+	for _, r := range s {
+		if strings.IndexRune(validTR005Draft3ExtensionChars, r) == -1 {
+			return errCharacter.WithAttributes("r", r)
+		}
+	}
+	return nil
+}
+
+// Validate implements the Data interface.
+func (m LoRaAllianceTR005Draft3) Validate() error {
```
rvolosatovs: FTR, this is not used by generated validators. It may make sense to implement `ValidateFields(...string) error` as well. `Validate() error` is then just `func() error { return ValidateFields(); }`

pkg/qrcode/lora_alliance_tr005d3_test.go
```diff
@@ -0,0 +1,167 @@
...
+					return assertions.New(t).So(errors.IsInvalidArgument(err), should.BeTrue)
+				},
+			},
+		} {
+			t.Run(tc.Name, func(t *testing.T) {
+				a := assertions.New(t)
+
+				var data LoRaAllianceTR005Draft3
+				err := data.UnmarshalText(tc.Data)
+				if tc.ErrorAssertion != nil && a.So(tc.ErrorAssertion(t, err), should.BeTrue) {
```
rvolosatovs:
```suggestion
				if tc.ErrorAssertion != nil {
					 a.So(tc.ErrorAssertion(t, err), should.BeTrue)
					 return
				}
```

pkg/qrcode/lora_alliance_tr005d3.go
johanstokking: There are no fields here

Issue #783, opened 2019-05-31
Component documentation


Sypheos:
**Umbrella issue**

#### Summary
Services and components need a high level documentation to enable knowledge transfert and make it easier to maintain, improve components.

#### Why do we need this?
Without documentation maintaining and collaboration is nearly impossible. If we want to enable the community to contribute to and use the stack we need an high level documentation about each component and service about the stack. 

#### What is already there? What do you see now?
#395 All documentation is under `/docs/content`

...

#### What is missing? What do you want to see?
A `/docs/content/components` with high level description on all the services and components in the stack following this plan.

1. description / For what purpose
2. It's place in the architecture and/or role
3. What does it do.
4. Link annex resources (like API reference)
...

#### Environment
Hugo

...

#### How do you propose to implement this?
Each Codeowner write the documentation for the component he his the owner of.
Under `/docs/content/components/<component>.md` 
...

#### Can you do this yourself and submit a Pull Request?

See https://github.com/TheThingsNetwork/lorawan-stack/issues/783#issuecomment-552809589

htdvisser:
I think this kind of (code) documentation belongs in godoc, and not in Hugo. We could consider pulling package docs/comments from go files into markdown, but IMO the godoc should be leading for documenting our go packages.

Sypheos:
Not aiming to make a technical doc but more of an introduction, description of the components, how they work with the other services and what they provide with some example.

Maybe shorten the scope to the subsystems.

johanstokking:
There's a difference between component documentation and code documentation. This issue is about component documentation, see also the issue title and points 1-2.

Code documentation should be godoc, which addresses a different audience, may have code examples and usage guidelines, etc. Since the stack is not intended to be used as a library, we don't have to spend time on that now.

I do think, however, that the list of packages is confusing; there shouldn't be anything like that in our documentation. But I guess this is simply taken from `CODEOWNERS` to figure out who's most knowledgeable about the topic.

htdvisser:
Ok yeah, the list of packages (and especially the quoted part below) confused me.

> Security
> /pkg/auth @johanstokking @htdvisser
> /pkg/crypto @johanstokking @htdvisser
> 
> Shared components
> /pkg/component @johanstokking @htdvisser
> /pkg/types @johanstokking @htdvisser @rvolosatovs

Can we agree on the following list of tasks then?

[...]

PR #1529, opened 2019-10-31
Add high level IS component documentation

htdvisser:
#### Summary

This Pull Request adds high level documentation for the Identity Server to the Components section.

Refs #783


doc/content/components/identity-server.md
```diff
@@ -4,8 +4,80 @@ description: ""
...
+
+### Applications
+
+Applications are used to organize registrations and traffic of multiple end devices in once place. An application typically corresponds to a collection of end devices that are in the same deployment, or of the same type.
+
+For more details, see the the [application API reference]({{< ref "/reference/api/application.md" >}}).
+
+### End Devices
+
+The end device registry in the Identity Server stores only metadata of end devices, allowing clients such as the Console and CLI to list end devices in an application. It typically stores metadata about the brand, model, hardware, firmware and location of the end device. It also stores addresses of the Network Server, Application Server and Join Server, so that clients such as the Console and CLI know where to fetch other properties of the end device.
```
bafonins: `It also stores addresses of the Network Server, Application Server and Join Server, so that clients such as the Console and CLI know where to fetch other properties of the end device.`
Nothing wrong here, however, this is not how the console fetches ns/js/as related fields for devices. The console uses addresses from the stack configuration. Should this be addressed @kschiffer ?

doc/content/components/identity-server.md
```diff
@@ -4,8 +4,80 @@ description: ""
...
+
+For more details, see the the [end device API reference]({{< ref "/reference/api/end_device.md" >}}).
+
+### Organizations
+
+Organizations are used to create groups of multiple users, and easily assign rights to the entire group of users.
```
bafonins: Maybe it is worth mentioning explicitly that organizations cannot be added as collaborators on other organizations.

doc/content/components/identity-server.md
kschiffer: Indeed, the console will not attempt to fetch from arbitrary URIs based on the `*_address` properties in the entity, which is as far as I understand also not the desired behavior.
Is there any misunderstanding here @htdvisser ?

htdvisser: Ok, I reworded this so that it doesn't imply anything about fetching, just knowing where the data is stored.

Issue #1530, opened 2019-10-31
NS and AS Device Registries do not check for _KEYS rights


htdvisser:
#### Summary

The device registry gRPC services in the Network Server and Application Server do not check for the `RIGHT_APPLICATION_DEVICES_READ_KEYS` and `RIGHT_APPLICATION_DEVICES_WRITE_KEYS` rights.

#### What do you see now?

Only the Join Server checks: https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/joinserver/grpc_deviceregistry.go#L55-L56

#### What do you want to see instead?

The Network Server and Application Server need to check as well.

#### How do you propose to implement this?

Field mask path magic and defer statements?

PR #1550, opened 2019-11-05
Check _KEYS rights, various key-related registry fixes

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1530

#### Changes

- Check _KEYS rights in NS and AS registries
- Fix JS registry key-related logic and field path checks.
- Filter the returned protos to only include the requested and implicit fields. (Some registry operations may require getting more fields than requested)
- Fix JS tests

#### Checklist


rvolosatovs:
IMO JS changes are closely related and, hence, belong here. If you think otherwise - please let me know and I will extract them into a different PR.

pkg/joinserver/grpc_deviceregistry.go
```diff
@@ -46,20 +46,49 @@ type jsEndDeviceRegistryServer struct {
 	JS *JoinServer
 }
 
+func fieldmaskHasRootKeys(paths ...string) bool {
+	return ttnpb.HasAnyField(paths,
+		"root_keys.app_key.key",
+		"root_keys.nwk_key.key",
+		"root_keys.root_key_id",
```
htdvisser: Maybe we should allow reading the key ID, which may be useful for debugging

pkg/applicationserver/grpc_deviceregistry.go
```diff
@@ -84,6 +95,16 @@ func (r asEndDeviceRegistryServer) Set(ctx context.Context, req *ttnpb.SetEndDev
 	if err := rights.RequireApplication(ctx, req.EndDevice.ApplicationIdentifiers, ttnpb.RIGHT_APPLICATION_DEVICES_WRITE); err != nil {
 		return nil, err
 	}
+	if ttnpb.HasAnyField(req.FieldMask.Paths,
+		"pending_session.keys.app_s_key.key",
+		"pending_session.keys.session_key_id",
```
htdvisser: Maybe we should allow reading the key ID, which may be useful for debugging

rvolosatovs:
Do we only care for CLI compatibility here or also compatibility of direct calls to the registry?
As I mentioned in OP, we can simply set-up an alias, but the question is whether it belongs in the client(e.g. CLI) or the registries themselves?

johanstokking:
> Do we only care for CLI compatibility here or also compatibility of direct calls to the registry?

See https://github.com/TheThingsNetwork/lorawan-stack#commitments-and-releases point 3.

I don't see a reason to break either. Why not just blank the encrypted key and label?

pkg/ttnpb/fieldmask_utils.go
```diff
@@ -182,10 +183,98 @@ outer:
 	return excluded
 }
 
+// EnsureFields appends all ensurePaths which are not already present in paths to paths and returns the result.
+func EnsureFields(paths []string, ensurePaths ...string) []string {
```
rvolosatovs: Any better naming suggestion?
Is there an "upsert" analogue for appending?

htdvisser: If we consider a FieldMask as a set type, Add would be fine too

pkg/applicationserver/grpc_deviceregistry.go
```diff
@@ -92,7 +120,9 @@ func (r asEndDeviceRegistryServer) Set(ctx context.Context, req *ttnpb.SetEndDev
 			evt = evtCreateEndDevice(ctx, req.EndDevice.EndDeviceIdentifiers, nil)
 		} else {
 			evt = evtUpdateEndDevice(ctx, req.EndDevice.EndDeviceIdentifiers, req.FieldMask.Paths)
-			if err := ttnpb.ProhibitFields(req.FieldMask.Paths, "ids.dev_addr"); err != nil {
+			if err := ttnpb.ProhibitFields(req.FieldMask.Paths,
+				"ids.dev_addr",
+			); err != nil {
```
johanstokking: These diffs are not really necessary imo

pkg/ttnpb/field_mask_validation.go
```diff
@@ -141,10 +141,9 @@ var AllowedFieldMaskPathsForRPC = map[string][]string{
 		"ids.device_id",
 		"ids.join_eui",
 		"session.dev_addr",
-		"session.keys.app_s_key",
```
johanstokking: Breaks API

pkg/ttnpb/field_mask_validation.go
```diff
@@ -211,9 +210,7 @@ var AllowedFieldMaskPathsForRPC = map[string][]string{
 		"provisioning_data",
 		"resets_join_nonces",
 		"root_keys",
-		"root_keys.app_key",
```
johanstokking: Idem and onward

pkg/ttnpb/field_mask_validation.go
rvolosatovs: It does not break any promises in https://github.com/TheThingsNetwork/lorawan-stack#commitments-and-releases as far as I am concerned

pkg/applicationserver/grpc_deviceregistry.go
rvolosatovs: This is for consistency with other registries

pkg/ttnpb/field_mask_validation.go
rvolosatovs: It may break console, in case they use `session.*_key` directly, in which case we need to fix that and bump minor

johanstokking: Yes it does, because in the past we allowed `session.keys.app_s_key` and now we don't anymore. That simply breaks clients.

rvolosatovs: In https://github.com/TheThingsNetwork/lorawan-stack/pull/1550#issuecomment-550311221 you explicitly pointed out point 3 of "commitments and releases", which reads "We will not break the public command-line interface and configuration within the major version. This means that you can safely build scripts and migrate configurations".
This PR is compliant with this and 3 other commitments.

So do you propose to allow setting `*_key.{encrypted_key,kek_label}` fields? That is the only way to not break clients, who use the registry API directly.

rvolosatovs:
Blocking #1591

CHANGELOG.md
```diff
@@ -49,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix device creation rollback potentially deleting existing device with same ID.
 - Fix missing transport credentials when using external NS linking.
+- Fix device key unwrapping
```
johanstokking: \+ `.`

pkg/ttnpb/fieldmask_utils.go
```diff
@@ -182,10 +182,64 @@ outer:
 	return excluded
 }
 
+// AddFields appends all addPaths which are not already present in paths to paths and returns the result.
+func AddFields(paths []string, addPaths ...string) []string {
```
johanstokking: Please add tests

PR #1555, opened 2019-11-06
Support AWS instance credentials in blob config

htdvisser:
#### Summary

This is a quickfix to make the blob config support AWS instance credentials.

Previously it expected the Access Key ID and Secret Access Key to be configured explicitly, now we only set those if present, and let the AWS SDK do its own behavior otherwise.

#### Changes

- Replace custom AWS credentials provider

#### Checklist

### DIFF
--- a/pkg/config/shared.go
+++ b/pkg/config/shared.go
@@ -197,21 +197,6 @@ type BlobConfigAWS struct {
 	SessionToken    string `name:"session-token" description:"Session token"`
 }
 
-type blobConfigAWSCredentials BlobConfigAWS
-
-func (c blobConfigAWSCredentials) Retrieve() (credentials.Value, error) {
-	if c.AccessKeyID == "" || c.SecretAccessKey == "" {
-		return credentials.Value{}, errMissingBlobConfig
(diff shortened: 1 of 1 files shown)


pkg/config/shared.go
```diff
@@ -232,11 +217,19 @@ func (c BlobConfig) Bucket(ctx context.Context, bucket string) (*blob.Bucket, er
 	case "local":
 		return ttnblob.Local(ctx, bucket, c.Local.Directory)
 	case "aws":
-		return ttnblob.AWS(ctx, bucket, &aws.Config{
-			Endpoint:    &c.AWS.Endpoint,
-			Region:      &c.AWS.Region,
-			Credentials: credentials.NewCredentials(blobConfigAWSCredentials(c.AWS)),
-		})
+		config := aws.NewConfig()
```
rvolosatovs:
```suggestion
		conf := aws.NewConfig()
```
> https://github.com/TheThingsNetwork/lorawan-stack/blob/master/DEVELOPMENT.md#single-word-entities

PR #1558, opened 2019-11-06
Allow configuring interop server fetch source

### LINKED ISSUE #1553: Load trusted sender ID CAs from blob
#### Summary

Load trusted sender ID CAs from blob

#### Why do we need this?

To manage configuration on the server side also via blobs, just like clients.

Currently, we have sender client CA config via;

https://github.com/TheThingsNetwork/lorawan-stack/blob/cf3c5115718f3df88a66f431171bd406bd6e94b0/pkg/config/shared.go#L112

I'd like to have the option to load this config from a yaml in a file/blob/http whatever, using the fetcher, like the client configuration. We need to stay compatible with this config; it may act as static configuration.

Add a `source` setting to the interop server config as well as fields for defining the sources.

Can review

rvolosatovs:
#### Summary

Closes #1553 

#### Changes

- Allow configuring interop server fetch source

#### Checklist


pkg/config/shared.go
```diff
@@ -402,6 +396,49 @@ func (c InteropClient) Fetcher(ctx context.Context, blobConf BlobConfig) (fetch.
...
+	BlobConfig BlobConfig `name:"-"`
+}
+
+// Fetcher returns fetch.Interface defined by conf.
+// If no configuration source is set, this method returns nil, nil.
+func (c InteropServer) Fetcher(ctx context.Context, blobConf BlobConfig) (fetch.Interface, error) {
+	// TODO: Remove detection mechanism (https://github.com/TheThingsNetwork/lorawan-stack/issues/1450)
+	if c.ConfigSource == "" {
+		switch {
+		case c.Directory != "":
```
johanstokking: I don't think this is entirely backward compatible; if we have no `ConfigSource` but we have `len(c.SenderClientCAs) > 0`, we should revert to existing behavior and use `directory` config source with directory being empty or working dir

johanstokking:
Looping @htdvisser in for admin merging as I won't be available in time to do that

pkg/config/shared.go
rvolosatovs: Indeed, I copy-pasted from client implementation and did not account for this

rvolosatovs:
That's not what I understood from the issue - will update tomorrow!

rvolosatovs:
Added https://github.com/TheThingsNetwork/lorawan-stack/pull/1558/commits/ee88e3f0135d5b0d612a224aeeb24045d057146e to make @htdvisser happy.
IMO the approach with explicit flag in `cmd/ttn-lw-stack` is way cleaner. So @htdvisser if you agree, I will revert it.

htdvisser:
> Added ee88e3f to make @htdvisser happy.

You don't have to do this to make me happy. You have to do this because you're working in an existing codebase, and this is how we do things in that existing codebase.

pkg/config/shared.go
htdvisser: I don't think we need to be backwards compatible here. This code works on the newly added `sender-client-ca` option. Since it didn't exist yet in the past, there is no need for compatibility, right? And I don't think we need to be compatible with how other (similar) configs do this. 

Fine to leave this as it is, just saying I don't think it's necessary.

rvolosatovs:
Cannot merge without @johanstokking's approval. @htdvisser - please merge this ASAP

PR #1569, opened 2019-11-07
Add status tooltip

### LINKED ISSUE #1562: Add tooltips to status component
#### Summary
Our status component should show tooltips on hover.

#### Why do we need this?
We sometimes use the status dot without label. To improve the accessibility, we should make sure to provide a `title` prop with the written title of the status.

Status component.

Title prop set for status component.

#### Environment
Browser.

Trivial.

@pgalic96

pgalic96:
#### Summary

Closes #1562 

#### Changes

- add `title` prop to status component
#### Notes for Reviewers

- @kschiffer is this what you meant, we could also infer title from the value of status (for example, if `bad` then title is `Disconnected`) but not sure if the colors mean the same everywhere.

#### Checklist


pkg/webui/components/status/index.js
```diff
@@ -42,7 +42,7 @@ const Status = function({ className, status, label, pulse, labelValues, children
   return (
     <span className={classnames(className, style.container)}>
       {statusLabel}
-      <span className={classnames(cls)} />
+      <span className={classnames(cls)} title={status} />
```
kschiffer: Keep in mind that the title needs to be translated, so you'd need to also create a message for each status and evaluate that first using `@injectIntl`. Let's introduce a title prop as well, in case one wants to control the title specifically. You could then do: `title={translatedTitle || translatedLabel || translatedStatus}`.

pgalic96: Makes sense!

pkg/webui/components/status/index.js
```diff
@@ -14,13 +14,20 @@
...
 import classnames from 'classnames'
-
+import { defineMessages, injectIntl } from 'react-intl'
 import Message from '../../lib/components/message'
 import PropTypes from '../../lib/prop-types'
 
 import style from './status.styl'
 
-const Status = function({ className, status, label, pulse, labelValues, children }) {
+defineMessages({
```
kschiffer:
```suggestion
const m = defineMessages({
```

pkg/webui/components/status/index.js
```diff
@@ -39,10 +46,26 @@ const Status = function({ className, status, label, pulse, labelValues, children
...
+      id: title.id,
+    })
+  } else if (label) {
+    translatedTitle = intl.formatMessage({
+      id: label.id,
+    })
+  } else {
+    translatedTitle = intl.formatMessage({
+      id: `components.status.index.${status}`,
+    })
```
kschiffer:
```js
translatedTitle = intl.formatMessage(m[status])
```

PR #1570, opened 2019-11-07
Fix blob config sharing

rvolosatovs:
#### Summary

Minor configuration cleanup

#### Changes

- Fix blob config sharing


rvolosatovs:
Cannot merge without @johanstokking's approval - @htdvisser can you do that?

PR #1588, opened 2019-11-14
Touch devices in JS, NS and AS on set

johanstokking:
#### Summary

Touch devices in NS and AS on set if field mask is empty

#### Changes

- Hit NS and AS set even if there are no field masks defined

#### Notes for Reviewers

This is draft and discussion, although it's a small and pretty harmless change.

The rationale behind this is that routing component device registries don't make a different between create and update. So when the device does exist in ER and JS, but not in NS and AS (because of cluster configurations), we can neither use the `ttn-lw-cli end-device`'s `create` nor `update` commands to get the devices created in NS and AS. That only works when field masks are specified, which in case of AS doesn't make sense.

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/end_devices.go
+++ b/cmd/ttn-lw-cli/commands/end_devices.go
@@ -467,7 +467,7 @@ var (
 
 			device.SetFields(isRes, append(isPaths, "created_at", "updated_at")...)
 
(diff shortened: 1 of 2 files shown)


htdvisser:
I'm not sure if this is a good idea. It adds unnecessary calls when a user just updates the description of a device in the ER. This results in unnecessary delay in the command, and it generates unnecessary load on the servers.

I think a better approach would be to make this behavior explicit using a flag.

PR #1594, opened 2019-11-15
Fix infinite loop in Redis

rvolosatovs:
#### Summary

Fixes infinite loop in NS Redis introduced by #1524 

#### Changes

- Fix `XReadGroup` call - split `XReadGroup` for pending messages and `XReadGroup` for new messages in 2 calls, since `XReadGroup` for pending messages does not block

#### Notes for Reviewers

I misread https://redis.io/commands/xreadgroup
After reading this:
> Any other ID, that is, 0 or any other valid ID or incomplete ID (just the millisecond time part), will have the effect of returning entries that are pending for the consumer sending the command with IDs greater than the one provided. So basically if the ID is not >, then the command will just let the client access its pending entries: messages delivered to it, but not yet acknowledged. Note that in this case, both BLOCK and NOACK are ignored.

I assumed the `XReadGroup` would still block if there are multiple streams being read and there were no pending messages, but that turned out to not be the case. (https://github.com/antirez/redis/issues/6587)

#### Checklist

### DIFF
--- a/pkg/networkserver/redis/application_uplink_queue.go
+++ b/pkg/networkserver/redis/application_uplink_queue.go
@@ -129,28 +129,18 @@ func (q *ApplicationUplinkQueue) Subscribe(ctx context.Context, appID ttnpb.Appl
 			logger.WithError(err).Error("Failed to delete Redis key")
 		}
 	}()
-	for {
-		_, err := q.Redis.XGroupCreateMkStream(upStream, q.Group, "0").Result()
-		if err != nil && !ttnredis.IsConsumerGroupExistsErr(err) {
-			return ttnredis.ConvertError(err)
-		}
 
-		var timeout time.Duration
-		if hasDL {
-			timeout = time.Until(dl)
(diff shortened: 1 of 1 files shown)


rvolosatovs:
@htdvisser that is not the fix - the fix is separately processing the pending messages and then blocking for new ones, since Redis does not allow doing both in 1 call.
Creation was not supposed to be in the loop as well.

htdvisser:
Okay, can you clarify that in the **Changes** section of the PR?

And maybe smart to add some comments to the code to warn future developers about this.

rvolosatovs:
I think current implementation is actually "more correct" and should be clear enough as is. In any case this behavior is(and should be) documented at https://redis.io/commands/xreadgroup, no need to document Redis commands here.

htdvisser:
I'm not saying we should document Redis commands. I'm just proposing to warn future developers about this "surprise". If you don't think that's necessary, fine.

PR #1604, opened 2019-11-18
Optimize Redis application uplink queue

### LINKED ISSUE #1598: AS links capped at 10 per CPU
#### Summary

#1524 introduced blocking `XREADGROUP` per each AS-NS link, but did not account for the fact that default limit of connections in Redis library we're using is capped at `10*runtime.NumCPU()`

#### Steps to Reproduce

1. Start stack
2. Link more than `10*runtime.NumCPU()` applications and not generate any uplinks

#### What do you see now?

`redis: connection pool timeout`

Successful execution

#### Environment

`v3.2.5`

Not sure, polling is one possible fix, but seems dirty.
We should probably shard the uplinks.

yes

rvolosatovs:
#### Summary

Closes #1598 

#### Changes

- Use a channel to block for new uplinks instead of TCP connection.

#### Notes for Reviewers

This implementation only works when there's 1 NS in the cluster and/or they're properly sharded, i.e. only NS holding link to application A should be generating uplinks for application A.

#### Checklist

### DIFF
--- a/pkg/networkserver/redis/application_uplink_queue.go
+++ b/pkg/networkserver/redis/application_uplink_queue.go
@@ -16,13 +16,11 @@ package redis
 
 import (
 	"context"
-	"crypto/rand"
-	"time"
+	"fmt"
+	"sync"
 
 	"github.com/go-redis/redis"
-	ulid "github.com/oklog/ulid/v2"
 	"go.thethings.network/lorawan-stack/pkg/errors"
-	"go.thethings.network/lorawan-stack/pkg/log"
 	ttnredis "go.thethings.network/lorawan-stack/pkg/redis"
 	"go.thethings.network/lorawan-stack/pkg/ttnpb"
 	"go.thethings.network/lorawan-stack/pkg/unique"
@@ -33,6 +31,8 @@ type ApplicationUplinkQueue struct {
 	MaxLen int64
 	Group  string
 	ID     string
+
+	subscriptions sync.Map
 }
 
 // NewApplicationUplinkQueue returns new application uplink queue.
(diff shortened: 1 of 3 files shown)


pkg/networkserver/redis/application_uplink_queue.go
```diff
@@ -82,119 +89,62 @@ var (
...
-		}
-	}()
-
-	_, err = q.Redis.XGroupCreateMkStream(upStream, q.Group, "0").Result()
-	if err != nil && !ttnredis.IsConsumerGroupExistsErr(err) {
-		return ttnredis.ConvertError(err)
+	upCh := make(chan *ttnpb.ApplicationUp, 1)
+	_, ok := q.subscriptions.LoadOrStore(uid, upCh)
+	if ok {
+		panic(fmt.Sprintf("duplicate subscription for application %s", uid))
```
htdvisser: This should error, not panic. `ApplicationUplinkQueue.Subscribe` is called by the implementation of the `LinkApplication` RPC, which will be called by an (external) Application Server.

Situation: When an Application Server restarts because of a crash, the old application link may not have been canceled yet (the TCP connection may not have been closed cleanly) when the new instance calls LinkApplication, that causes a `panic` 💥

rvolosatovs: The previous link is always closed before a new one is opened - https://github.com/TheThingsNetwork/lorawan-stack/blob/d6724d1cfb463b4096207e829cbbeb2e6daafe33/pkg/networkserver/grpc_asns.go#L85-L91
This is only reachable when there's a bug in NS implementation

PR #1605, opened 2019-11-18
Provide received_at in appropriate uplink messages

### LINKED ISSUE #1439: Application upstream message `received_at` never set
#### Summary

AS never sets `received_at` on the `ApplicationUp` message.

#### Steps to Reproduce

1. Subscribe to AS messages, i.e. MQTT or webhooks
2. See any upstream message, for example uplink

#### What do you see now?

No `received_at` set

For uplink-triggered messages (uplink, join-accept, downlink ack/nack), the `received_at` set to the `UplinkMessage` received from NS. In other scenarios (downlink queued/sent/failed, location solved, package-triggered messages, etc), don't set the value.

Can review

adriansmares:
#### Summary

Closes #1439

#### Changes

- Pass the `received_at` from the `UplinkMessage` to the `ApplicationUp` that is sent to the ApplicationServer.

#### Notes for Reviewers

The initial issue mentioned that the AS should fill in the timestamp, but it was actually the NS which didn't fill it.

#### Checklist


CHANGELOG.md
```diff
@@ -17,7 +17,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+<<<<<<< HEAD
```
rvolosatovs: :eyes:

adriansmares: Forgot to save before commiting.

adriansmares:
> Why would NS set `ReceivedAt` in the application uplink?
> NS receives `UplinkMessage` and sets `ReceivedAt`, when it's received
> AS receives `ApplicationUp` and sets `ReceivedAt`, when it's received

My understanding from https://github.com/TheThingsNetwork/lorawan-stack/issues/1439#issue-504707230 is that the timestamp should actually be the one from the NS (`the received_at set to the UplinkMessage received from NS`). @johanstokking can you clarify this ?

rvolosatovs:
Please see https://github.com/TheThingsNetwork/lorawan-stack/issues/1439#issuecomment-555030465 in the issue you linked

pkg/networkserver/networkserver_flow_test.go
```diff
@@ -413,6 +414,7 @@ func handleClassAOTAAEU868FlowTest1_0_2(ctx context.Context, conn *grpc.ClientCo
 					Key: &appSKey,
 				},
 				SessionKeyID: []byte("session-key-id"),
+				ReceivedAt:   asUp.GetJoinAccept().GetReceivedAt(),
```
adriansmares: @rvolosatovs do you see a way of achieving this without simulating the wall clock ? (we could do that in this test as well, if you consider it wouldn't be too artificial).

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- The Network Server now provides the timestamp at which it received `JoinAccept` and `ApplicationUplink` messages.
```
htdvisser: Empty line between h3 and ul

api/messages.proto
```diff
@@ -166,6 +166,7 @@ message ApplicationUplink {
   google.protobuf.Struct decoded_payload = 5;
   repeated RxMetadata rx_metadata = 6 [(validate.rules).repeated.min_items = 1];
   TxSettings settings = 7 [(gogoproto.nullable) = false, (validate.rules).message.required = true];
+  google.protobuf.Timestamp received_at = 8 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
```
htdvisser: Maybe add a comment to explain _by_ what it is received at this time.

api/messages.proto
```diff
@@ -184,6 +185,7 @@ message ApplicationJoinAccept {
   // Indicates whether the security context refers to the pending session, i.e. when this join-accept is an answer to a
   // rejoin-request.
   bool pending_session = 4;
+  google.protobuf.Timestamp received_at = 8 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
```
htdvisser: Maybe add a comment to explain _by_ what it is received at this time.

pkg/networkserver/networkserver_flow_test.go
rvolosatovs: This looks good to me - I don't we should mock time in flow tests.

CHANGELOG.md
rvolosatovs:
```suggestion
- The Network Server now provides the timestamp at which it received join-accept or data uplink messages
```

or

```suggestion
- The Network Server now provides the timestamp at which it received `JoinAccept` and `UplinkMessage` messages.
```

In any case:
a. NS does not receive `ApplicationUplink`
b. I don't think we should use proto message names here - we are writing this for users, they should need to look through our source code to understand what a particular change means

htdvisser:
This PR has API changes that are visible externally, so it requires a minor version bump.

johanstokking:
> This PR has API changes that are visible externally, so it requires a minor version bump.

I agree it requires a minor bump, but not for this reason. We should be safely _adding_ externally visible fields. The reason for a minor bump is that we're changing API between components (NS and AS) for this to work.

PR #1608, opened 2019-11-19
Set AU_915_928 duty-cycle to 1%

johanstokking:
#### Summary

Set AU_915_928 duty-cycle to 1%

cc @ama9910

#### Changes

- Change from 100% to 1%

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix duty-cycle limitation in `AU_915_928` band
+
 ### Security
 
 ## [3.2.6] - 2019-11-18
--- a/pkg/band/au_915_928.go
+++ b/pkg/band/au_915_928.go
@@ -77,7 +77,7 @@ func init() {
 			{
 				MinFrequency: 902000000,
 				MaxFrequency: 928000000,
-				DutyCycle:    1,
+				DutyCycle:    0.01,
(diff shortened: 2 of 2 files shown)


johanstokking:
@ama9910 can you provide the source link (or document name + year + page)?

Note to self; file issue to reduce the many `DevStatusReq`

ama9910:
@htdvisser @johanstokking I can't find this documented anywhere so have contacted ACMA. ETA 29-Nov-2019.

MrAPierce:
I cant find it either. it will be interesting to see what the ACMA say. every ACMA document that I have read states that the frequency hopping is limiting factor, even going through Australian certification of devices, hopping was there main concern for the ISM band here.

I don't believe Australia is bound by the 1% duty cycle like other countries. @ama9910 let us know what you find out 👍

johanstokking:
Blocked by confirmation from ACMA

ama9910:
Well, it looks like I was mistaken. The ACMA have confirmed there is no duty cycle requirement for Australia! Pity. As TTN users we are careful with the band-plan (30 secs per day per device). Non-TTN users, who I thought were limited to 1%, are actually unlimited.

johanstokking:
Actually this is good news to me. Uplink should indeed be limited in the light of overall network capacity – we will add feedback in The Things Stack – but 1% for the entire band would be very restrictive, significantly more restrictive than EU 863-867 MHz. Firmware updates would not be practically possible. Gateways should not be too busy transmitting, but with many deployments having devices covered by 2, 3 gateways or more, and class C and B picking up, having downlink capacity available is good news.

Issue #1552, opened 2019-11-05
Compute nsScheduleWindow at runtime


rvolosatovs:
#### Summary

`nsScheduleWindow` can be computed at runtime

#### Why do we need this?

To allow more time for AS to update the queue on uplinks, for example, and generally improve performance and throughput.

#### What is already there? What do you see now?

`nsScheduleWindow` hard-coded to a particular value(since #1551 it will be equal to 100ms)

#### What is missing? What do you want to see?

`nsScheduleWindow` computed at runtime.

#### How do you propose to implement this?

Each time processing a downlink task results in scheduling attempt on GS, NS can compute the difference between scheduled starting time of the task and the time response from GS is received. Use moving average(e.g. EMA) of the difference of these 2 timestamps as the `nsScheduleWindow` value.

#### Can you do this yourself and submit a Pull Request?

yes

rvolosatovs:
Using 99th percentile is definitely better approach here

johanstokking:
Is this still relevant since #2418 ?

rvolosatovs:
> Is this still relevant since #2418 ?

Yes

johanstokking:
> Each time processing a downlink task results in scheduling attempt on GS, NS can compute the difference between scheduled starting time of the task and the time response from GS is received.

Few things to consider here;
1. Not all downlink traffic goes to a cluster-local GS. We can only do this for downlink to a cluster-local GS target
2. The response time of GS doesn't say anything
3. The reported `delay` in the GS response is partly based on gateway round-trip times and partly on duty-cycle constraints in non-time-bound downlink (class C)
4. What you need here is the round-trip time from NS to gateways that are in the cluster. This needs reliable round-trip time observation by GS and the API needs to accommodate feeding this information to NS

rvolosatovs:
> starting time of the task and the time response from GS is received.

This is incorrect, `nsScheduleDelay` is difference between the time downlink task is scheduled for and the time downlink task was actually executed at.
If task is scheduled for time `T`, but it executes at `T+X`, `nsScheduleDelay == X`

PR #1611, opened 2019-11-19
Support class B

### LINKED ISSUE #19: Class B support
**Summary:** 

Class B devices should be supported

**Why do we need this?**

To support Class B devices and make end customers happy

No class B support

Class B support

According to the spec

Notes:
- In class C, there is a timeout within which the device can respond to downlink traffic (i.e. MAC answer, downlink ack). During this time, NS may not send other downlink. This is not specified for class B; i.e. there is no requirement for the device to answer downlink traffic in ping slots before the next ping slot, and there is no requirement for the NS to hold traffic while the device has not responded. When implementing this, we may want to consider confirmed downlink (so that the device responds per spec, before the next ping slot), like we do with class C. Background here: https://github.com/TheThingsNetwork/lorawan-stack/pull/412#discussion_r270765410

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/833 by @rvolosatovs

rvolosatovs:
#### Summary

Closes #19 

#### Changes

- Support class B ping slots in NS
- Add `crypto.ComputePingOffset`
- Improve logging
- Mark Rx windows closed after all (i.e. failed as well) class A downlink attempts
- Refactor downlink timing (references #1552) - this is necessary to prevent NS from infinitely looping over ping slots and to actually be able to determine the ping slot to attempt
- Always use Rx2 if available for Class B/C
- Add device class switch events
- Add `should.ResembleFields`, which only compares selected field paths
- Account for MAC requests being able to be scheduled at `AbsoluteTime` of an application downlink or when an unconfirmed class B/C application downlink is scheduled (NS needs to wait until it can schedule another confirmed downlink before scheduling any MAC request in class B/C)

#### Checklist


pkg/crypto/ping_slot.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+const (
+	minPingPeriod = 1 << 5
+	maxPingPeriod = 1 << 12
+)
+
+var errInvalidPingPeriod = errors.DefineInvalidArgument("ping_period", fmt.Sprintf("ping period must be a power of 2 between %d and %d, got {value}", minPingPeriod, maxPingPeriod), "value")
+
+func ComputePingOffset(beaconTime uint32, devAddr types.DevAddr, pingPeriod uint16) (uint16, error) {
```
rvolosatovs: @johanstokking please confirm that this is compliant with the spec.
Based it on 1.0.3:
2019-11-19-17:00:36-screenshot

pkg/networkserver/utils.go
```diff
@@ -151,11 +154,60 @@ func nextClassADataDownlinkSlot(dev *ttnpb.EndDevice) (time.Time, bool) {
 	}
 }
 
-func nextConfirmedClassCDownlinkAt(dev *ttnpb.EndDevice, defaults ttnpb.MACSettings) time.Time {
-	if dev.GetMACState().GetLastConfirmedDownlinkAt() == nil {
-		return time.Time{}
+const (
```
rvolosatovs: @johanstokking please confirm that I got the spec right here.
Also the const durations - e.g. the `tBeaconDelay`:
1.0.3 spec:
2019-11-19-17:03:08-screenshot

`mSec` stands for `microsecond` here, right?

johanstokking:
@rvolosatovs please ask for review when things work, unless you want me to look at something specifically in the meantime

pkg/networkserver/downlink.go
```diff
@@ -249,21 +236,14 @@ func (ns *NetworkServer) generateDownlink(ctx context.Context, dev *ttnpb.EndDev
 			if err != nil {
 				return nil, generateDownlinkState{}, errEncodeMAC.WithCause(err)
 			}
-			if mType == ttnpb.MType_UNCONFIRMED_DOWN &&
```
rvolosatovs: `ExpectAnswer` is redundant, since NS can just use `PendingRequests` slice for this purpose.
Also, adapt for class B, move this closer to where it's actually used and out of the loop.

rvolosatovs:
Seems that downlink timing alignments broke the timing calculation for scheduling, so class B downlinks are not being scheduled on physical devices. It's a small issue which I will fix on Monday.

pkg/crypto/ping_slot.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	pingOffsetCipher = c
+}
+
+const (
+	minPingPeriod = 1 << 5
+	maxPingPeriod = 1 << 12
+)
+
+var errInvalidPingPeriod = errors.DefineInvalidArgument("ping_period", fmt.Sprintf("ping period must be a power of 2 between %d and %d, got {value}", minPingPeriod, maxPingPeriod), "value")
```
johanstokking: - no need to make `value` public as it is part of the message format
- backticks for attributes

pkg/networkserver/downlink.go
```diff
@@ -423,12 +406,19 @@ func (ns *NetworkServer) generateDownlink(ctx context.Context, dev *ttnpb.EndDev
 	default:
 		return nil, genState, errNoDownlink
 	}
+	if mType == ttnpb.MType_UNCONFIRMED_DOWN && len(dev.MACState.PendingRequests) > 0 &&
+		(dev.MACState.DeviceClass == ttnpb.CLASS_B ||
```
johanstokking: Do we always want an immediate answer in class B, also in 1.1+?

pkg/networkserver/downlink.go
```diff
@@ -849,14 +846,26 @@ loop:
 	return phy.DataRates[maxUpDRIdx].DefaultMaxSize.PayloadSize(fp.DwellTime.GetUplinks())
 }
 
+// networkInitiatedDownlinkInterval is the minimum time.Duration passed before a network-initiated(e.g. Class B or C) downlink following an arbitrary downlink.
+const networkInitiatedDownlinkInterval = 5 * time.Second
```
johanstokking: What is this 5 seconds based on? Is this a trade-off?

pkg/networkserver/downlink.go
```diff
@@ -849,14 +846,26 @@ loop:
...
+
 // downlinkRetryInterval is the time interval, which defines the interval between downlink task retries.
 const downlinkRetryInterval = 2 * time.Second
 
-// gsScheduleWindow is the time interval, which is sufficient for GS to ensure downlink is scheduled.
-const gsScheduleWindow = 30 * time.Second
+// gsScheduleWindow returns minimum time.Duration between downlink being sent to GS for transmission and it being transmitted to the device by the gateway.
```
johanstokking: This is longer in peering scenarios.

What is this used for exactly?

What is the consequence of a lower and a higher value?

"returns minimum time" implies that it doesn't get faster than this, but GS can make a gateway transmit within 500 ms

pkg/networkserver/utils.go
```diff
@@ -193,54 +280,98 @@ func nextConfirmedClassCDownlinkAt(dev *ttnpb.EndDevice, defaults ttnpb.MACSetti
...
 		}
 
-		statusAt, ok := deviceNeedsDevStatusReqAt(dev, defaults)
-		if !ok {
-			if absTime.IsZero() {
-				return time.Time{}, false
+	case ttnpb.CLASS_C:
+		earliestConfirmedAt = nextConfirmedClassCDownlinkAt(dev, defaults)
+	default:
+		panic(fmt.Sprintf("Unmatched device class: %v", dev.MACState.DeviceClass))
```
johanstokking: lowercase

pkg/networkserver/utils.go
```diff
@@ -193,54 +280,98 @@ func nextConfirmedClassCDownlinkAt(dev *ttnpb.EndDevice, defaults ttnpb.MACSetti
...
+	if !ok {
+		if absTime.IsZero() {
+			return time.Time{}, false
+		}
+		return absTime, true
+	}
+	if !absTime.IsZero() && absTime.Before(t) {
+		return absTime, true
+	}
+	return t, true
```
johanstokking: Not covered by test, should it?

pkg/networkserver/downlink.go
rvolosatovs: Arbitrary value - this is also the default RxDelay we use and Join-request delay, so if a device would send an uplink we'd know about it by that time instead of trying to schedule another class B/C downlink.
What this value means is that if NS has `n` unconfirmed downlinks to schedule, it will schedule first one at `T`, second one at `T+networkInitiatedDownlinkInterval`, third one at `T+2*networkInitiatedDownlinkInterval` .. `n` at `T+(n-1)*networkInitiatedDownlinkInterval`.

rvolosatovs: It should probably be configurable on NS level at level at least, maybe even device

pkg/networkserver/downlink.go
rvolosatovs: So this is the absolute minimum time GS needs to be able to schedule a downlink.
If downlink is to be scheduled at `T`, this is lowest `d`, where downlink sent to GS at `T-d` will be scheduled at 99% of the cases.

rvolosatovs: NS uses this value to select a ping slot to attempt, for example

pkg/networkserver/downlink.go
johanstokking: Does this mean that in case of FOTA for example, many class C unconfirmed downlink messages, there's always 5 seconds in between?

pkg/networkserver/downlink.go
johanstokking: Discussed offline; conclusion is that we're budgeting the RX1 delay and give 1 second and 50% of the remaining to getting back to GS. We cannot make NS aware of latencies of individual downlink paths due to protocol restrictions, complexity and predictability.

pkg/networkserver/downlink.go
rvolosatovs: Yes, unless there's an uplink.
Do you think it should be lower?

johanstokking: Yeah I think it should be lower indeed. Why don't you use the GS's delay on the schedule response?

rvolosatovs: NS does that, but it still needs to have some kind of time interval to wait after the transmission by GS. To account for e.g. possible uplink.

rvolosatovs: Lowered it to 1 second

pkg/networkserver/downlink.go
rvolosatovs: AFAIK no LoRaWAN spec version released so far contains anything similar to the clause in 1.1 class C specifying the timeout between confirmed downlink transmissions, so I'd say yes.

pkg/networkserver/downlink.go
johanstokking: If it's unconfirmed downlink, I wouldn't bother too much about that.

Indeed we shouldn't send confirmed downlink too fast.

pkg/networkserver/downlink.go
```diff
@@ -112,6 +98,45 @@ func (s generateDownlinkState) appendApplicationUplinks(ups []*ttnpb.Application
 	}
 }
 
+// deviceTransmissionDelay returns the estimated one-way transmission delay between
+// frame being trasmitted by the device and received by the Network Server and other way round.
```
johanstokking: typo

pkg/networkserver/downlink.go
```diff
@@ -112,6 +98,45 @@ func (s generateDownlinkState) appendApplicationUplinks(ups []*ttnpb.Application
...
+func deviceTransmissionDelay(dev *ttnpb.EndDevice, defaults ttnpb.MACSettings, phy band.Band) time.Duration {
+	if dev.MACSettings != nil && dev.MACSettings.DesiredRx1Delay != nil {
+		return dev.MACSettings.DesiredRx1Delay.Value.Duration() / 2
+	} else if dev.PendingMACState != nil {
+		return dev.PendingMACState.DesiredParameters.Rx1Delay.Duration() / 2
+	} else if dev.MACState != nil {
+		return dev.MACState.DesiredParameters.Rx1Delay.Duration() / 2
```
johanstokking: Why does desired take precedence?

For style; afaik this gives lint warnings (else-if after return); consider switch cases

pkg/networkserver/downlink.go
```diff
@@ -1327,17 +1377,24 @@ func (ns *NetworkServer) processDownlinkTask(ctx context.Context) error {
...
+				}
+				switch {
+				case req.Class == ttnpb.CLASS_B && downAt.IsZero():
+					logger.Error("Class B downlink with no absolute time generated")
+					return dev, sets, nil
+
+				case genState.ApplicationDownlink.GetClassBC().GetAbsoluteTime() != nil:
+					req.AbsoluteTime = genState.ApplicationDownlink.GetClassBC().GetAbsoluteTime()
+
+				case req.Class == ttnpb.CLASS_B, req.Class == ttnpb.CLASS_C && downAt.After(timeNow().Add(infrastructureDelay)):
```
johanstokking: Isn't this `infrastructureDelay/2`?

pkg/networkserver/downlink.go
rvolosatovs: Because *desired* will always contain the most up-to-date information.

johanstokking: But if my RX1 delay is 1 second, and I'm changing it to 15 seconds, will class A downlink ever make it? If so, how?

rvolosatovs: Changed this to always using the active Rx1Delay except for join-requests

Issue #1565, opened 2019-11-06
Set location settings via map


kschiffer:
#### Summary
Location settings for devices and gateway antennas should be possible to be done via setting markers on the map.

#### Why do we need this?
Improved UX; the current approach via forms is not very pleasant.

#### What is already there? What do you see now?
Setting location settings via the location form.

#### What is missing? What do you want to see?
A way to set locations via the map itself (dropping markers)

#### Environment
Browser.

#### How do you propose to implement this?
This is a matter of wiring leaflet features.

#### Can you do this yourself and submit a Pull Request?
@mjamescompton but please reassign to @pgalic96 if you don't have time to work on this soon.

kschiffer:
Duplicate of #1380

Issue #1095, opened 2019-08-02
Introduce input component with toggle to hide the value


kschiffer:
#### Summary
We need an input component that can securely hold sensitive information by being able to hide its value. It should be possible to toggle this hiding feature on and off.

#### Why do we need this?
`password` type inputs have problems that can result in the browser trying to auto-fill the login password into them even when it is not desired (See: #1093). Also semantically, it is not correct to treat these values as passwords.
Still, we want to secure sensitive data from onlookers by hiding the value by default.

#### What is already there? What do you see now?
We already have the safe inspector which has a similar interface but is read-only.

#### What is missing? What do you want to see?
A respective text input component. 

#### Environment
Console in browser.

#### How do you propose to implement this?
Extending our input component with a show/hide toggle, similarly to how we extended the byte input.

#### Can you do this yourself and submit a Pull Request?
Yes, but help welcome.

kschiffer:
Closed via #4754

PR #573, opened 2019-04-25
Add Device view

kschiffer:
**Summary:**
Closes #332

This PR adds an "Add Devices" view to the console, with which devices can be created.
image

**ABP Bit:**
image

**Changes:**
- Add/refactor routing for add device view
- Add textarea option to `<Field />`
- Small tweaks/fixes for `<Form />`, `<Field />` `<Input type="byte" />`
- Extend API tree for device creation
- Add "Add Device" view
  - Add "Add Device" form with conditional fields
  - Validation scheme
  - Submit logic
- Various styling tweaks
- Add a generic fallback error message, in case a proper one cannot be obtained

**Notes for Reviewers:**
Make sure to rebuild the SDK because of #569 for the submit to work properly.

**Release Notes: _(optional)_**
- Add a view for adding devices to an application

cc @htdvisser @johanstokking

PR #1616, opened 2019-11-20
Refactor device general settings page

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/579
This PR changes how the General Settings page for devices is structured as well as fixes several issues 

<details>
  <summary>screenshots</summary>
  
Screenshot 2019-11-20 at 17 59 04
Screenshot 2019-11-20 at 18 03 07
</details>

#### Changes

- Add the `<Collapse />` component
- Add `hidden` prop to the `<Field />` component
- Take into account `multicast` devices and display/hide fields depending on the actiovation mode
- Take into account `lorawan_version` and change field title/description as well as hide/show fields
- Make separate forms for end device fields per stack component
- Fix proper `root_keys` updates when switching between external and local js modes

#### Notes for Reviewers

you might want to use [this script](https://gist.github.com/htdvisser/2d146748d01e56feea480b322bb0c400) to create various devices for testing different scenarios.

@rvolosatovs please verify that the network server section shows correct fields with different device types
@johanstokking please verify all join server related things + overall consistency
@htdvisser overall consistency
@kschiffer ui suggestions, also note that this is just a poc and I will remove inconsistencies in imports/form setups in the actual PR

Note, there are features that are not implemented, but I am aware of them and plan to add them later in the actual PR. 
1. Handling 404 responses per component. Currently, the forms are meant only for editing existing devices.
2. Deletion per component is missing

#### Checklist


rvolosatovs:
2019-11-21-19:57:41-screenshot
This looks quite broken to me. That's Firefox 70.0.1

rvolosatovs:
Additionally, I don't see any MAC-related fields available neither on device creation, nor update:
2019-11-21-20:04:19-screenshot

bafonins:
@rvolosatovs this PR has nothing to do with the create form. I suggest we add mac related fields later.

Please verify that this implementation of the network server section behaves as expected and displays correct field titles/descriptions for different device types.

rvolosatovs:
The summary of original issue you linked in OP reads:
> The add device form (added in #573) is currently not composing fields based on constraints (except for ABP / OTAA selection). For example, it could hide the fields that are only relevant for specific LoRaWAN Versions, or rename fields accordingly (eg. `NwkSKey ` vs. `FNwkSIntKey`).

All the discussion in that issue concerns the device creation as well.

I would expect the PR referencing such an issue be related to creation of devices.

With that out of the way, I don't see any new fields available to be set from https://github.com/TheThingsNetwork/lorawan-stack/issues/579#issuecomment-553347858 (
2019-11-22-13:46:40-screenshot
)

Additionally, I don't think that tab-per-component is the right approach. Some of the fields may need to shared between components and I don't think it makes sense to allow users to create inconsistencies this way. Moreover, I highly doubt it's simple to use for our users - they don't need to know which field belongs in which component - they should be able to "just" change a particular setting they need.
For example, if I have an ABP device and I want to change it's `DevAddr`, I should be able to go the device settings, select `DevAddr` field and change it - with proposed implementation the user would need to find *all* tabs, in which the `DevAddr` is actually configured in and then make sure to change it in *all* of these tabs. In case of `DevAddr` that must be Network Server and Application Server - I know that, because I work on v3 stack and I've read the LoRaWAN spec - I doubt that would be as clear to a user, who wants to just make their device measure room temperature and send it to some external integration service.

Note, that having "LoRaWAN field" tab, for example, as was discussed in the original issue seems very reasonable to me and, in fact, preferable, because that immediately gives the user at least the idea of what kind of settings they will find there.

rvolosatovs:
I attempted to start the console to give you more detailed answer on https://github.com/TheThingsNetwork/lorawan-stack/pull/1616#issuecomment-557466192, however I'm unable to access applications in the console on e2a83fa15
```
ttn-lw-cli app list       
 DEBUG Using access token (valid until 5:30PM) 
 DEBUG ccResolverWrapper: sending update to cc: {[{localhost:8884 0  <nil>}] <nil>}
 DEBUG pickfirstBalancer: HandleSubConnStateChange: 0xc000235340, CONNECTING
 DEBUG pickfirstBalancer: HandleSubConnStateChange: 0xc000235340, READY
 DEBUG Finished unary call                      duration=44.970985ms grpc_method=List grpc_service=ttn.lorawan.v3.ApplicationRegistry namespace=grpc
 DEBUG Total results: 1                        
[{
  "ids": {
    "application_id": "test-app"
  },
  "created_at": "2019-11-22T15:31:00.524Z",
  "updated_at": "2019-11-22T15:31:00.524Z"
}]
```

At the same time in the console:
2019-11-22-16:42:15-screenshot

When trying to add an application:
2019-11-22-16:27:26-screenshot
This is what I see in the browser console:
2019-11-22-16:40:49-screenshot

rvolosatovs:
To answer https://github.com/TheThingsNetwork/lorawan-stack/pull/1616#issuecomment-557466192 with more details:
The currently present fields mostly look fine(albeit they are not enough and this kind of design would probably not work once you add the rest of the fields)

For 1.1, the 1.1-specific key comments should not include the "(only for LoRaWAN 1.1+)"
2019-11-22-17:03:27-screenshot

Again, no idea if that belongs here, but several times I have encountered device page loading infinitely:
2019-11-22-17:02:57-screenshot

johanstokking:
> This looks quite broken to me. That's Firefox 70.0.1

Same in Brave 1.0.1. @kschiffer @bafonins this is unrelated to this PR, but the **NwkSKey** label seems not to get rendered.

pkg/webui/console/views/device-general-settings/join-server-form/index.js
```diff
@@ -0,0 +1,263 @@
...
+        />
+      )}
+      <Form.Field
+        title={sharedMessages.macVersion}
+        name="_lorawan_version"
+        component={Input}
+        type="hidden"
+        hidden
+        disabled
+      />
```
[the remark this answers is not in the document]
bafonins: not really. the lorawan verison is needed in the validation schema for the js form

pkg/webui/console/views/device-general-settings/network-server-form/index.js
```diff
@@ -0,0 +1,287 @@
...
+
+const lorawanPhyVersions = [
+  { value: '1.0.0', label: 'PHY V1.0' },
+  { value: '1.0.1', label: 'PHY V1.0.1' },
+  { value: '1.0.2-a', label: 'PHY V1.0.2 REV A' },
+  { value: '1.0.2-b', label: 'PHY V1.0.2 REV B' },
+  { value: '1.0.3-a', label: 'PHY V1.0.3 REV A' },
+  { value: '1.1.0-a', label: 'PHY V1.1 REV A' },
+  { value: '1.1.0-b', label: 'PHY V1.1 REV B' },
+]
```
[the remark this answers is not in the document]
bafonins: Cant find those. Can you share a link?

pkg/webui/console/views/device-general-settings/join-server-form/index.js
```diff
@@ -0,0 +1,263 @@
...
+      }
+    case actionTypes.RESETS_JOIN_NONCES:
+      return {
+        ...state,
+        resetsJoinNonces: !state.resetsJoinNonces,
+      }
+    default:
+      return state
+  }
+}
```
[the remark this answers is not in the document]
bafonins: What would you suggest?

bafonins:
to @johanstokking 
> Reorder the components: IS, NS, AS, JS. The reason to put JS last is that keys depend on NS settings, so it's more chronologically

done

> Are we getting Console settings, besides delete? I would call that the Danger Zone, like GitHub

@kschiffer any suggestions here?

> Can you add the NetID to the Join Server settings? 6 bytes (fixed length) hex, optional

done

> Maybe expand IS settings by default

done

> We need a way to "touch" devices, like #1588.

This is related to the "create" functionality in the general settings page. I mentioned in the PR that will do that later since it required quite some work in js sdk

> AS settings say "Application Server does not store any OTAA specific device fields", but we have payload formatters right?

We have payload formatters in a separate section in the side navigation

to @rvolosatovs 

> Additionally, I don't think that tab-per-component is the right approach. 

I agree with this and you can actually see some fields being duplicated across component sections, e.g. `join_server_address` in is/js sections or `dev_address` in ns/as sections. The reason for this is that we want to make device handling easier (both for the user and for us developers) as well as adding #1588.

Would be nice to hear from @htdvisser as well

pkg/webui/console/views/device-general-settings/identity-server-form/index.js
```diff
@@ -0,0 +1,188 @@
...
+      }
+    },
+    [initialValues, onSubmit],
+  )
+
+  const isOTAA = isDeviceOTAA(device)
+  const externalJs = hasExternalJs(device)
+  const isNewLorawanVersion = parseLorawanMacVersion(device.lorawan_version) >= 110
+  const hasJoinEUI = Boolean(device.ids.join_eui)
+  const hasDevEUI = Boolean(device.ids.dev_eui)
```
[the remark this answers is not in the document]
bafonins: Why? These helper functions are not used anywhere else and were added just for the sake of readability

pkg/webui/console/views/device-general-settings/messages.js
```diff
@@ -0,0 +1,38 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { defineMessages } from 'react-intl'
+
+const messages = defineMessages({
+  isTitle: 'Identity Server Settings',
+  isDescription: 'Edit end device fields related to the Identity Server',
```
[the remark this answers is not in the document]
bafonins: See my latest comment. 

Would be nice to come up with section titles and descriptions, cc @johanstokking @htdvisser

pkg/webui/console/views/device-general-settings/identity-server-form/index.js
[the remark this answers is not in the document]
kschiffer: Was just a suggestion. I could see us using these sort of checks more often in the future.

pkg/webui/console/views/device-general-settings/network-server-form/index.js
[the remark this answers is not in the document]
kschiffer: Sorry, I was mistaken. Mixed it up with the frequency plans.

htdvisser:
Part A of my review: Sections

- Strictly speaking the contents of the Location tab belongs under IS Settings
- Strictly speaking the contents of the Payload Formatters tab belongs under AS settings
- IS Settings are "General Settings" or perhaps more "Metadata", since it's not actually settings.
- JS Settings should not be loaded (be disabled) if `join_server_address` does not match console config.
- NS Settings could be "Network Layer"
- NS Settings should not be loaded (be disabled) if `network_server_address` does not match console config.
- AS Settings could be "Application Layer"
- AS Settings should not be loaded (be disabled) if `application_server_address` does not match console config.
- I don't like having the "Delete" in a "Console Settings" section. Can we just pull that out of the sections?
- Not sure what to do with the Delete button with external NS/AS/JS

htdvisser:
Part B of my review: Identity Server Settings

- I don't think we should allow changing the NS/AS/JS addresses after creation
- When an address field is disabled, the placeholder shouldn't say "host"

And (out of scope for this PR) be aware of the fact that end devices can have multiple locations (from different sources).

htdvisser:
Part C of my review: Network Server Settings

- Keys should be masked
- Users may want to regenerate keys to random values
- Missing MAC layer fields, but I'm also fine with doing this iteratively
- OTAA and multicast devices also have DevAddrs (MulticastAddr?)
	- In fact, the DevAddr on the Network Server is leading
- OTAA and multicast devices also have NwkSKeys

htdvisser:
Part D of my review: Application Server Settings

- Keys should be masked
- OTAA and multicast devices also have AppSKeys

htdvisser:
Part E of my review: Join Server Settings

- Keys should be masked
- Missing fields (claiming, provisioning), but I'm also fine with doing this iteratively

pkg/webui/console/views/device-general-settings/network-server-form/index.js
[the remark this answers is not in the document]
htdvisser: This deserves an issue

rvolosatovs:
@htdvisser regarding Network Server settings(or, as you suggested Network Layer settings) - things like OTAA keys and DevAddr are immutable. I believe only mutable fields belong here?
If purpose of this is to not only show mutable "settings", but also immutable general information about devices, then I'll make another review of this.

pkg/webui/console/views/device-general-settings/join-server-form/index.js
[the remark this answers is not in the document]
kschiffer: Using plain `useState` rather, which saves quite some lines:
https://gist.github.com/kschiffer/c3c333af7ea8d77034b5aa20ceacf402

Maybe I overlooked something that speaks for the reducer approach. It's a suggestion; use at your discretion.

kschiffer:
> > This looks quite broken to me. That's Firefox 70.0.1
> 
> Same in Brave 1.0.1.

I don't get what exactly looks broken. As @bafonins already said, this PR is refactoring the general settings page only, as a first step. The device creation form is still the old "all-in-one" form. Maybe a misunderstanding, but just to be sure.

htdvisser:
@rvolosatovs Most of the fields that we currently have here are immutable, and I think it's just fine to also have those fields here. We just need to make sure the inputs are disabled.

johanstokking:
> I don't get what exactly looks broken

See https://github.com/TheThingsNetwork/lorawan-stack/pull/1616#issuecomment-558147038; **NwkSKey** label is not rendered. See the empty space between DevAddr and FNwkSIntKey

bafonins:
1. Updates screenshots
2. Moved `join_server_address` and `external_js` fields to the is form to allow the users to switch between the external Js option
3. Added `application_server_id`, `net_id`, `application_server_kek_label` and `network_server_kek_label` fields to the js form cc @johanstokking 
4. Some minor adjustments

@htdvisser Implemented some your suggestions that are relevant to the structure of the sections. Will implement the rest once we aprove the current structure of the sections/forms

kschiffer:
I think it works pretty well after fiddling around with it. 

Some things I noticed:
1. Changing "Resets Join Nonces" makes the view crash upon saving (happened to me with otaa device)
2. Save actions still miss a toast confirmation
3. `as server id`/`as server kek label` is no good input placeholder. You can remove these placeholders and add this to the field descriptions: `Defaults to the server id` / `Defaults to server kek label` (if I get it correctly). Generally, we should aim to use placeholder primarily to display example values.
4. NwkSKey label is still not rendered in the create form (must have something to do with changed messages)
5. The styling looks a lot better now. I'll follow up with some small change suggestions in the code.
6. Consider rebasing to latest master to avoid other reviewers getting affected by unrelated `redux-offline` issues

@johanstokking / @htdvisser please also think about what fields should have descriptions too or better descriptions than what we currently have. Same for the descriptions of the form sections.

pkg/webui/console/views/device-general-settings/collapse/collapse.styl
```diff
@@ -0,0 +1,35 @@
...
+  border-normal('bottom')
+
+.header
+  display: block
+  position: relative
+  padding-right: $ls.xl
+
+  .title
+    margin-bottom: $cs.xs
+
```
kschiffer:
```
.section:not(:last-child)
  border-normal('bottom')

.header
  display: block
  position: relative
  padding-right: $ls.xl
  padding-top: $ls.s
  padding-bottom: $ls.s

  .title
    margin-top:0
    margin-bottom: $cs.xs

  .description
    one-liner()
```

pkg/webui/console/views/device-general-settings/index.js
```diff
@@ -98,24 +123,137 @@ export default class DeviceGeneralSettings extends React.Component {
...
+      nsDescription = m.nsDescriptionMissing
+    } else if (!sameNsAddress) {
+      nsDescription = m.notInCluster
+    }
 
     return (
       <Container>
         <IntlHelmet title={sharedMessages.generalSettings} />
         <Row>
           <Col lg={8} md={12}>
```
kschiffer:
```suggestion
          <Col lg={8} md={12} className={style.container}>
```
and then please add this:
```css
.container
  margin-top: $ls.s * -1

**Comment by htdvisser** [comment]
In addition to my previous comments:

- The "External Join Server" checkbox is not relevant for ABP devices
- I don't think we should be able to change the DevAddr in the AS settings. It's just informational there.
- "AppEUI identifies the Join Server" is not correct (OTAA 1.0.2)
- "The DevEUI is the unique identifier for this device on the network": remove the "on the network" part
- The "Join Server Address" may still be present in the IS when using an external join server.
- The ordering of NS/AS/JS addresses in the "IS Settings" section is inconsistent. with the sections.
- The NS/AS/JS addresses in the "IS Settings" section aren't handled consistently (external JS vs external NS+AS). 
- It thinks my device is ABP/multicast if it's registered on external NS+AS

**Comment by johanstokking** [comment]
We're getting there.

Adding to @htdvisser's points above.

Illustration:

Screenshot 2019-12-03 at 17 41 05

- "Net ID" -> "Home NetID" (no space)
- I changed the device from 1.0.3 to 1.1; NwkKey appears but as "unexposed"
   - `nwk_key == nil` means it's not set. This is the case here, and the user should be able to set it
   - `nwk_key.key == nil` then the key is not exposed
- The placeholders in Application Server ID, Application Server KEK Label and Network Server KEK Label should not be "as server {id|label}" (not sure what it should be, but something a bit nicer)
- KEK in capitals. It's an abbreviation for key encryption key
- The label under AS ID and AS KEK Label are reversed
- The label under NS KEK Label is wrong; there's no NS-ID

**Comment by johanstokking** [comment]
- AppKey label: "The root key to derive the **application session key** to secure communication between the device and the application". In 1.0.x, the strong test should be **session keys** (i.e. all session keys). In 1.1.x, this text is correct
- If you switch from 1.0.x to 1.1, the NwkKey correctly appears. However, it's empty (that's also okay), but it says that the keys are automatically generated. The **Save changes** button is however disabled. What should we do here? Enable **Save changes** in this case?
- @kschiffer what do you think about "Danger zone" instead of "Console settings" to delete the device?

File: pkg/webui/console/views/device-general-settings/join-server-form/index.js [hunk 8]
```diff
@@ -0,0 +1,241 @@
...
+
+import { parseLorawanMacVersion, hasExternalJs } from '../utils'
+import validationSchema from './validation-schema'
+
+const getComponentBaseUrl = config => {
+  try {
+    const { base_url } = config
+
+    return new URL(base_url).hostname
+  } catch (e) {}
```
kschiffer: Emtpy catch is a bit smelly, I'd say.

bafonins:
@johanstokking addressed the issues you mentioned. please have another look.

bafonins:
> Strictly speaking the contents of the Location tab belongs under IS Settings
Strictly speaking the contents of the Payload Formatters tab belongs under AS settings

@kschiffer shall we move the payload formatters and locations forms to the general settings?

kschiffer:
> @kschiffer what do you think about "Danger zone" instead of "Console settings" to delete the device?

I would say it's not necessary since we already have an expressive confirmation modal for all delete operations. If we rename the titles/descriptions of the tabs to be more user-friendly, we could rename the IS settings to `General Settings` and simply put the delete button in there. 

It depends on how strictly we want to communicate the component separation to the user. I'd say that these technical specifics should not be too relevant for the user and could also be potentially confusing, but it depends on how TTS-backend knowledgeable we want to regard the users of the console.

kschiffer:
> > Strictly speaking the contents of the Location tab belongs under IS Settings
> > Strictly speaking the contents of the Payload Formatters tab belongs under AS settings
> 
> @kschiffer shall we move the payload formatters and locations forms to the general settings?

I'm against that for payload formatters because 
- it's a complex setting going beyond a simple key/value field form. I wouldn't want to cram that into the AS form.
- it's likely frequently accessed, so it's better UX to have it more easily accessible
- people are used to this being a "top level setting" from the v2 console

For location settings, I'm less strict but would also lean against since we will be adding setting of location via map (#1565) soon, making this setting more complex as well. I would rather see this as a separate tab in the settings panel if anything. See also my other comment about how to communicate the tabs.

kschiffer:
So to be more concrete, I suggest to label the sections as follows (as suggested by @htdvisser):

`Identity Server Settings` -> `Basic Information` (this includes the delete button as well)
Desc.: `Metadata and host information of your end device`

`Network Server Settings` -> `Network Layer Settings`
Desc.: `LoRaWAN related options such as MAC/PHY Versions, Class C Support and frequency plans`

`Application Server Settings` -> `Application Layer Settings`
Desc.: `Application related options such as device Address and App Session Key`

`Join Server Settings` -> `Join Settings` (?)
Desc.: `Join related options such as root keys, reset options and KEK labels`

`General Settings` -> remove

@htdvisser @johanstokking please confirm or suggest changes.

johanstokking:
I think it's clear to use the component names, but I'd agree with nicer group names for UX.

> Metadata and host information of your end device

I'd suggest "Description, cluster information and metadata."

> LoRaWAN related options such as MAC/PHY Versions, Class C Support and frequency plans

I'd suggest "LoRaWAN network-layer settings, behavior and session"

> Application related options such as device Address and App Session Key

Address part is wrong

I'd suggest "Application-layer settings, behavior and session"

> Join related options such as root keys, reset options and KEK labels

I'd suggest "Root keys and network settings for device activation"

johanstokking:
@bafonins now getting:

>Join Server Settings
>
>Activation mode unknown because Network Server is not available

In a IS+JS+Console cluster. In this case, Console should assume highest LoRaWAN version for backward-compatibility, i.e. 1.1.

bafonins:
@johanstokking fixed

kschiffer:
> I think it's clear to use the component names, but I'd agree with nicer group names for UX.
> 
> […]

Perfect, I've updated the [original comment](https://github.com/TheThingsNetwork/lorawan-stack/pull/1616#issuecomment-563255050).

johanstokking:
I created the devices with the script.

- [ ] In `lw-102-abp`, in NS settings, when generating a new `DevAddr` and saving changes, the AS fields become invalid. However; 1) there should still be an AS session, and 2) the device address should also be written to AS
- [ ] In `lw-otaa-ext-nsas`, in JS settings, when checking join nonces reset and save changes, I'm getting "invalid fieldmask" error. There are calls being made to `/ns/` while this shouldn't happen (they're external)
- [ ] In `lw-otaa-ext-nsas` (with AS and NS disabled), in JS settings, the AS-ID and AS/NS KEK labels are disabled. They should be editable here

htdvisser:
Structure:

- [ ] I thought we discussed moving the "Delete End Device" from "Console Settings" to "Basic Settings"

Security:

- [ ] Root Keys should only be sent to the JS, not to the NS/AS
- [ ] Keys should be masked
- [ ] Keys can't be set to random values (generate button like DevAddr)

Missing information:

- [ ] Join Server Address field value is missing for `lw-otaa-ext-js` (it shows `External` instead of `externalhost`)
- [ ] I still think it's a shame that we don't show the DevAddr and Session Keys for OTAA devices (we even show those in the Overview tab!)

Performance:

- [ ] Why do I see 4 GETs on "Save Changes"?
- [ ] Why do I see a PUT to the AS when I change an NS field?

Goldplating:

- [ ] It feels like we're repeating "Settings" too often. We're in the "General Settings" tab and see a lot of "Settings" sections

Future work:

- [ ] Still need to figure out what to do with the delete button when using external NS/AS/JS
- [ ] Still need to figure out where "Location" tab belongs
- [ ] Still need to figure out where "Payload Formatters" tab belongs

htdvisser:
Use case:

```
ttn-lw-cli dev create \
  --application-id admin-app \
  --device-id lw-otaa-ext-nsasjs \
  --name "LoRaWAN 1.1.0 OTAA Device with Defaults on external NS/AS/JS" \
  --lorawan-version 1.1.0 \
  --lorawan-phy-version 1.1.0-b \
  --frequency-plan-id EU_863_870_TTN \
  --join-eui 0102030405060708 \
  --dev-eui 0000110000000004 \
  --root-keys.root-key-id BEEF \
  --root-keys.nwk-key.key 01020304050607080102030405060708 \
  --root-keys.app-key.key 01020304050607080102030405060708 \
  --defaults \
  --network-server-enabled=false \
  --network-server-address=externalhost \
  --application-server-enabled=false \
  --application-server-address=externalhost \
  --join-server-enabled=false \
  --join-server-address=externalhost
```

This would be for a multi-cluster deployment with a single IS. We're on the `currenthost` Console viewing a device registered on the NS/AS/JS of `externalhost`.

bafonins:
@htdvisser 

> Keys should be masked

Lets do it in a separate PR. There is an issue for this already: https://github.com/TheThingsNetwork/lorawan-stack/issues/1095

> Keys can't be set to random values (generate button like DevAddr)

The user can always regenerate the keys by emptying the key field:
Screenshot 2019-12-16 at 12 04 36

> It feels like we're repeating "Settings" too often. We're in the "General Settings" tab and see a lot of "Settings" sections

@kschiffer what do you think?

> Still need to figure out where "Location" tab belongs
> Still need to figure out where "Payload Formatters" tab belongs

https://github.com/TheThingsNetwork/lorawan-stack/pull/1616#issuecomment-563247884

kschiffer:
> > It feels like we're repeating "Settings" too often. We're in the "General Settings" tab and see a lot of "Settings" sections
> 
> @kschiffer what do you think?

I think indeed we can just remove all `Settings` from the tab titles. Leaving just `Basic`, `Network Layer`, `Application Layer`, etc…

pkg/webui/console/components/device-data-form/index.js
```diff
@@ -177,13 +177,13 @@ class DeviceDataForm extends Component {
           required
         />
         <Form.Field
-          title={sharedMessages.fwdNtwkKey}
+          title={sharedMessages.nwkSKey}
```
htdvisser: It looks like we're losing information here

pkg/webui/console/components/device-data-form/messages.js
```diff
@@ -29,26 +38,33 @@ export default defineMessages({
...
+  fNwkSIntKeyDescription: 'Forwarding Network Session Integrity Key',
+  homeNetID: 'Home NetID',
+  joinEUIDescription: 'JoinEUI identifies the Join Server',
   leaveBlankPlaceholder: 'Leave blank to generate automatically',
-  ntwkSEncKeyDescription: 'Network Session Encryption Key (only for LoRaWAN 1.1+)',
+  multicast: 'Multicast',
+  nsServerKekLabel: 'Network Server KEK Label',
+  nsServerKekLabelDescription:
+    'The KEK label of the Network Server to use for wrapping the network session key',
+  ntwkSEncKeyDescription: 'Network Session Encryption Key',
```
htdvisser: Can we use the official nwk instead of ntwk? Also in other places

sdk/js/src/service/devices/index.js
```diff
@@ -143,6 +143,31 @@ class Devices {
...
+      } catch (e) {}
+
+      try {
+        const asHost = new URL(this._stackConfig.as).hostname
+
+        if (application_server_address !== asHost) {
+          delete requestTree.as
+        }
+      } catch (e) {}
+    }
```
kschiffer: This can be put into one try/catch. I'm still not very fond of empty catch blocks. At least a comment as to why the error can be ignored would be good.

pkg/webui/components/form/field/field.styl
```diff
@@ -39,6 +39,10 @@
   flex-wrap: wrap
   width: 100%
 
+  &.hidden
```
kschiffer: `display: none`?

pkg/webui/console/components/device-data-form/index.js
bafonins: `<DeviceDataForm />` is not related here. The only place we use it is in https://github.com/TheThingsNetwork/lorawan-stack/blob/feature/579-device-general-settings/pkg/webui/console/views/device-general-settings/network-server-form/index.js#L258

Is this correct?

pkg/webui/console/components/device-data-form/messages.js
```diff
@@ -29,26 +38,33 @@ export default defineMessages({
...
   nwkKeyDescription:
     'The root key to derive network session keys to secure communication between the device and the network',
+  nwkSKeyDescription: 'Network Session Key',
   otaa: 'Over The Air Activation (OTAA)',
   resetsFCnt: 'Resets Frame Counters',
   resetsJoinNonces: 'Resets Join Nonces',
   resetWarning: 'Reseting is insecure and makes your device susceptible for replay attacks',
-  sNtwkSIKeyDescription: 'Serving Network Session Integrity Key (only for LoRaWAN 1.1+)',
+  sNtwkSIKeyDescription: 'Serving Network Session Integrity Key',
```
kschiffer: Same here.
```suggestion
  sNwkSIKeyDescription: 'Serving Network Session Integrity Key',
```

pkg/webui/console/components/device-data-form/messages.js
bafonins: See my comment above

pkg/webui/console/views/device-general-settings/identity-server-form/index.js
```diff
@@ -0,0 +1,288 @@
...
+import m from '../../../components/device-data-form/messages'
+import PropTypes from '../../../../lib/prop-types'
+import sharedMessages from '../../../../lib/shared-messages'
+import { parseLorawanMacVersion, hasExternalJs, isDeviceOTAA } from '../utils'
+import validationSchema from './validation-schema'
+
+const messages = defineMessages({
+  deleteDevice: 'Delete End Device',
+  deleteWarning:
+    'Are you sure you want to delete "{deviceId}"? Deleting an end device cannot be undone!',
```
```suggestion
    'Are you sure you want to delete "{deviceId}"? This action cannot be undone and it will not be possible to reuse the end device ID!',
```

pkg/webui/console/views/device-general-settings/identity-server-form/validation-schema.js
```diff
@@ -0,0 +1,92 @@
...
+
+import randomByteString from '../../../lib/random-bytes'
+import sharedMessages from '../../../../lib/shared-messages'
+import { id as deviceIdRegexp, address as addressRegexp } from '../../../lib/regexp'
+import m from '../../../components/device-data-form/messages'
+
+import { parseLorawanMacVersion } from '../utils'
+
+const random16BytesString = () => randomByteString(32)
+const toUndefined = value => (!Boolean(value) ? undefined : value)
```
kschiffer: We use this frequently, so I'd say it would be better to outsource into a module.

rvolosatovs:
2019-12-20-00:08:06-screenshot
Application Layer description should just be (almost) the same as NS it has nothing to do with OTAA/ABP disctinction:
"LoRaWAN application-layer settings, behavior and session"

pkg/webui/console/views/device-general-settings/join-server-form/index.js
```diff
@@ -0,0 +1,191 @@
...
+        />
+      )}
+      <Form.Field
+        title={m.homeNetID}
+        name="net_id"
+        type="byte"
+        min={3}
+        max={3}
+        component={Input}
+        disabled={externalJs}
```
kschiffer: Can we add a description here too?

pkg/webui/console/views/device-general-settings/network-server-form/index.js
```diff
@@ -0,0 +1,293 @@
...
+  )
+
+  return (
+    <Form
+      validationSchema={validationSchema}
+      initialValues={initialValues}
+      onSubmit={onFormSubmit}
+      error={error}
+      formikRef={formRef}
+      enableReinitialize
```
kschiffer: Please add a field description for MAC/PHY/FP as well.
@htdvisser can you help out with wording?

pkg/webui/console/views/device-general-settings/application-server-form/index.js
```diff
@@ -0,0 +1,129 @@
...
+      initialValues={initialValues}
+      onSubmit={onFormSubmit}
+      error={error}
+      enableReinitialize
+    >
+      <DevAddrInput
+        title={sharedMessages.devAddr}
+        name="session.dev_addr"
+        placeholder={m.leaveBlankPlaceholder}
+        description={m.deviceAddrDescription}
```
kschiffer: Message does not exist.

PR #1636, opened 2019-11-22
Added support for Redis Sentinel

### LINKED ISSUE #1635: Add support for Redis Sentinel
#### Summary

Allow The Things Stack to use Redis Sentinel as database.

#### Why do we need this?

Redis Sentinel provides high availability for Redis. The Things Stack greatly benefits from this, as it uses Redis as database for most of its storage needs. 

The Things Stack can only connect to a single Redis instance.

Support for Redis Sentinel.

#### Environment

All

The Things Stack uses the `go-redis` library. `go-redis` transparently supports Redis Sentinel connections using `redis.NewFailoverClient()`.

Can implement. May need help with review + tests.

neoaggelos:
#### Summary

Implements support for Redis Sentinel. Closes #1635 

#### Changes

- Added new options `redis.failover.enable`, `redis.failover.addreses`, `redis.failover.master-name`
- When user uses the Redis Sentinel options, then a `redis.Client` is created using `redis.NewFailoverClient()` instead of `redis.NewClient()`.
- Updated `config.Redis.IsZero()`
- Updated relevant documentation

#### Notes for Reviewers

- `redis.NewFailoverClient()` returns a `redis.Client`, just like `redis.NewClient()`. This means that The Things Stack can trasparently use either Redis or Redis Sentinel without further changes.
- It may be desired to rename the configuration options related to Redis Sentinel.
- Default value for `redis.enable-failover` is set to false, so that the change is backwards compatible and does not break existing installations.
- Some kind of test should be added.
- Redis Sentinel comes with some log messags printed out by default, mixed with log messages from the stack. This should probably be fixed as well. 
- meta: Is the `all` tag correct?

#### Checklist

### DIFF
--- a/doc/content/reference/configuration/the-things-stack.md
+++ b/doc/content/reference/configuration/the-things-stack.md
@@ -105,10 +105,21 @@ The `blob` source loads from the given path in a bucket. This requires the globa
 
 Redis is the main data store for the [Network Server]({{< relref "network-server.md" >}}), [Application Server]({{< relref "application-server.md" >}}) and [Join Server]({{< relref "join-server.md" >}}). Redis is also used by the [Identity Server]({{< relref "identity-server.md" >}}) for caching and can be used by the [events system]({{< ref "#events-options" >}}) for exchanging events between components.
 
-- `redis.address`: Address of the Redis server
+Redis configuration options:
+
 - `redis.password`: Password of the Redis server
 - `redis.database`: Redis database to use
 - `redis.namespace`: Namespace for Redis keys
+
+If connecting to a single Redis instance:
+
+- `redis.address`: Address of the Redis server
+
+Or you can enable failover using [Redis Sentinel](https://redis.io/topics/sentinel):
+
+- `redis.failover.enable`: Set to `true`
+- `redis.failover.addresses`: List of addresses of the Redis Sentinel instances (required)
+- `redis.failover.master-name`: Redis Sentinel master name (required)
 
 ## Blob Options
 
@@ -170,7 +181,7 @@ The `blob` source loads from the given path in a bucket. This requires the globa
 
 ## Cluster Options
 
-The `cluster` options configure how The Things Stack communicates with other components in the cluster. These options do not need to be set when running a single instance of The Things Stack. The most important options are the ones to configure the addresses of the other components in the cluster. 
+The `cluster` options configure how The Things Stack communicates with other components in the cluster. These options do not need to be set when running a single instance of The Things Stack. The most important options are the ones to configure the addresses of the other components in the cluster.
 
 - `cluster.identity-server`: Address for the Identity Server
 - `cluster.gateway-server`: Address for the Gateway Server
--- a/pkg/config/shared.go
+++ b/pkg/config/shared.go
@@ -116,16 +116,28 @@ type HTTP struct {
 	Health          Health           `name:"health"`
 }
 
+// RedisFailover represents Redis failover configuration.
+type RedisFailover struct {
+	Enable     bool     `name:"enable" description:"Enable failover using Redis Sentinel"`
+	Addresses  []string `name:"addresses" description:"Redis Sentinel server addresses"`
+	MasterName string   `name:"master-name" description:"Redis Sentinel master name"`
+}
+
 // Redis represents Redis configuration.
 type Redis struct {
-	Address   string   `name:"address" description:"Address of the Redis server"`
-	Password  string   `name:"password" description:"Password of the Redis server"`
-	Database  int      `name:"database" description:"Redis database to use"`
-	Namespace []string `name:"namespace" description:"Namespace for Redis keys"`
+	Address   string        `name:"address" description:"Address of the Redis server"`
+	Password  string        `name:"password" description:"Password of the Redis server"`
+	Database  int           `name:"database" description:"Redis database to use"`
+	Namespace []string      `name:"namespace" description:"Namespace for Redis keys"`
+	Failover  RedisFailover `name:"failover" description:"Redis failover configuration"`
 }
 
 // IsZero returns whether the Redis configuration is empty.
-func (r Redis) IsZero() bool { return r.Address == "" && r.Database == 0 && len(r.Namespace) == 0 }
+func (r Redis) IsZero() bool {
+	return r.Database == 0 && len(r.Namespace) == 0 &&
+		(r.Failover.Enable && r.Failover.MasterName == "" && len(r.Failover.Addresses) == 0 ||
+			!r.Failover.Enable && r.Address == "")
+}
 
 // CloudEvents represents configuration for the cloud events backend.
 type CloudEvents struct {
--- a/pkg/events/redis/redis.go
+++ b/pkg/events/redis/redis.go
@@ -22,17 +22,14 @@ import (
 	"github.com/go-redis/redis"
 	"go.thethings.network/lorawan-stack/pkg/config"
 	"go.thethings.network/lorawan-stack/pkg/events"
+	ttnredis "go.thethings.network/lorawan-stack/pkg/redis"
 )
 
 // WrapPubSub wraps an existing PubSub and publishes all events received from Redis to that PubSub.
 func WrapPubSub(wrapped events.PubSub, conf config.Redis) (ps *PubSub) {
 	ps = &PubSub{
-		PubSub: wrapped,
-		client: redis.NewClient(&redis.Options{
-			Addr:     conf.Address,
-			Password: conf.Password,
-			DB:       conf.Database,
-		}),
+		PubSub:       wrapped,
+		client:       ttnredis.New(&ttnredis.Config{Redis: conf}).Client,
 		eventChannel: strings.Join(append(conf.Namespace, "events"), ":"),
 		closeWait:    make(chan struct{}),
 	}
--- a/pkg/redis/redis.go
+++ b/pkg/redis/redis.go
@@ -19,6 +19,8 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"io/ioutil"
+	"log"
 	"strconv"
 	"strings"
 	"time"
@@ -78,15 +80,29 @@ type Config struct {
 	Namespace []string
 }
 
+// newRedisClient returns a Redis client, which connects using correct client type.
+func newRedisClient(conf config.Redis) *redis.Client {
+	if conf.Failover.Enable {
+		redis.SetLogger(log.New(ioutil.Discard, "", 0))
+		return redis.NewFailoverClient(&redis.FailoverOptions{
+			MasterName:    conf.Failover.MasterName,
+			SentinelAddrs: conf.Failover.Addresses,
+			Password:      conf.Password,
+			DB:            conf.Database,
+		})
+	}
+	return redis.NewClient(&redis.Options{
+		Addr:     conf.Address,
+		Password: conf.Password,
+		DB:       conf.Database,
+	})
+}
+
 // New returns a new initialized Redis store.
 func New(conf *Config) *Client {
 	return &Client{
 		namespace: Key(append(conf.Redis.Namespace, conf.Namespace...)...),
-		Client: redis.NewClient(&redis.Options{
-			Addr:     conf.Address,
-			Password: conf.Password,
-			DB:       conf.Database,
-		}),
+		Client:    newRedisClient(conf.Redis),
 	}
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add support for Redis Sentinel (see `redis.failover.enable`, `redis.failover.master-name`, `redis.failover.addresses` options).
+
 ### Changed
 
 ### Deprecated
@@ -33,7 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fix an issue causing unexpected behavior surrounding login, logout and token management in the Console.
-- Fix an issue causing the application link page of the Console to load infinitely. 
+- Fix an issue causing the application link page of the Console to load infinitely.
 
 ## [3.2.6] - 2019-11-18
 


pkg/config/shared.go
```diff
@@ -118,14 +118,20 @@ type HTTP struct {
...
 type Redis struct {
-	Address   string   `name:"address" description:"Address of the Redis server"`
-	Password  string   `name:"password" description:"Password of the Redis server"`
-	Database  int      `name:"database" description:"Redis database to use"`
-	Namespace []string `name:"namespace" description:"Namespace for Redis keys"`
+	Address            string   `name:"address" description:"Address of the Redis server"`
+	Password           string   `name:"password" description:"Password of the Redis server"`
+	Database           int      `name:"database" description:"Redis database to use"`
+	Namespace          []string `name:"namespace" description:"Namespace for Redis keys"`
+	Sentinel           bool     `name:"sentinel" description:"Connect to Redis Sentinel"`
```
johanstokking:
```suggestion
	EnableFailover     bool     `name:"enable-failover" description:"Enable failover using Redis Sentinel"`
```

pkg/config/shared.go
```diff
@@ -118,14 +118,20 @@ type HTTP struct {
...
+	SentinelAddresses  []string `name:"sentinel-addresses" description:"Redis Sentinel server addresses (when sentinel=true)"`
```
johanstokking: `FailoverAddresses`

pkg/config/shared.go
```diff
@@ -118,14 +118,20 @@ type HTTP struct {
...
+	SentinelMasterName string   `name:"sentinel-master-name" description:"Redis Sentinel master name (when sentinel=true)"`
```
johanstokking: `FailoverMasterName`

pkg/config/shared.go
```diff
@@ -118,14 +118,20 @@ type HTTP struct {
...
 }
 
 // IsZero returns whether the Redis configuration is empty.
-func (r Redis) IsZero() bool { return r.Address == "" && r.Database == 0 && len(r.Namespace) == 0 }
+func (r Redis) IsZero() bool {
+	return (r.Sentinel == true && r.SentinelMasterName == "" && len(r.SentinelAddresses) == 0) ||
```
johanstokking: Use go idiomatic boolean checks (i.e. no `== true` or `== false`)

johanstokking:
> * Some kind of test should be added.

@htdvisser @rvolosatovs should we make this part of CI?

> * Redis Sentinel comes with some log messags printed out by default, mixed with log messages from the stack. This should probably be fixed as well.

I suppose you refer to Docker Compose logs? Or is the Redis client printing logs directly?

neoaggelos:
> I suppose you refer to Docker Compose logs? Or is the Redis client printing logs directly?

No, this is the go-redis module logging stuff to stderr directly. Example:

```
 DEBUG Creating listener                        address=:8887
 DEBUG Start serving                            address=[::]:1887
 DEBUG Start serving                            address=[::]:8887
 DEBUG Loaded TLS certificate                  
  INFO Setting up Network Server               
redis: 2019/11/23 11:12:54 sentinel.go:379: sentinel: discovered new sentinel="bf962e747d4c658689a02f7747eff779b4db08e3" for master="mymaster"
redis: 2019/11/23 11:12:54 sentinel.go:379: sentinel: discovered new sentinel="5e7bb518d84c1638223c57f4c1b873ae180f874e" for master="mymaster"
redis: 2019/11/23 11:12:54 sentinel.go:379: sentinel: discovered new sentinel="fd196c1a327a9db12849d19e3f4a0958dc27237b" for master="mymaster"
redis: 2019/11/23 11:12:54 sentinel.go:379: sentinel: discovered new sentinel="fd196c1a327a9db12849d19e3f4a0958dc27237c" for master="mymaster"
redis: 2019/11/23 11:12:54 sentinel.go:332: sentinel: new master="mymaster" addr="10.121.205.250:6379"
  INFO Setting up Application Server           
 DEBUG Creating listener                        address=:1883
 DEBUG Creating listener                        address=:8883
 DEBUG Start serving                            address=[::]:1883
 DEBUG Loaded TLS certificate                  
  INFO Setting up Join Server                  
```

The easy solution is to `2> /dev/null`, but I think extra care need be taken.

The go-redis module has a `redis.SetLogger()` that can be used for this, but it requires a `log.Logger`, so passing the custom TTN logger cannot work. What do you suggest? I think that we could either add on option to silence those logs (e.g. `--redis.disable-failover-logs`), or try to somehow integrate with the TTN logger (this could need changes in `go-redis` itself).

pkg/config/shared.go
neoaggelos: I have updated the code as per the suggestion

johanstokking:
Thanks for clarifying!

> The go-redis module has a `redis.SetLogger()` that can be used for this, but it requires a `log.Logger`, so passing the custom TTN logger cannot work

So we basically need to implement the indirection. That seems to be useful anyway to do in `pkg/log`.

cmd/internal/shared/config.go
```diff
@@ -76,9 +76,12 @@ var DefaultGRPCConfig = config.GRPC{
...
 // DefaultRedisConfig is the default config for Redis.
 var DefaultRedisConfig = config.Redis{
-	Address:   "localhost:6379",
-	Database:  0,
-	Namespace: []string{"ttn", "v3"},
+	Address:            "localhost:6379",
+	Database:           0,
+	Namespace:          []string{"ttn", "v3"},
+	EnableFailover:     false,
+	FailoverMasterName: "mymaster",
```
johanstokking: I think it's cleaner if we require the user to specify something rather than `mymaster`. Our aim is to keep default settings sane for local use.

cmd/internal/shared/config.go
```diff
@@ -76,9 +76,12 @@ var DefaultGRPCConfig = config.GRPC{
...
+	FailoverAddresses:  []string{},
```
johanstokking: nit; `nil` is a bit cleaner

doc/content/reference/configuration/the-things-stack.md
```diff
@@ -105,11 +105,22 @@ The `blob` source loads from the given path in a bucket. This requires the globa
...
 Redis is the main data store for the [Network Server]({{< relref "network-server.md" >}}), [Application Server]({{< relref "application-server.md" >}}) and [Join Server]({{< relref "join-server.md" >}}). Redis is also used by the [Identity Server]({{< relref "identity-server.md" >}}) for caching and can be used by the [events system]({{< ref "#events-options" >}}) for exchanging events between components.
 
+Redis configuration options:
+
 - `redis.address`: Address of the Redis server
 - `redis.password`: Password of the Redis server
 - `redis.database`: Redis database to use
 - `redis.namespace`: Namespace for Redis keys
 
+Or you can enable failover using Redis Sentinel:
```
johanstokking: nit; add link to Redis Sentinel docs

pkg/redis/redis.go
```diff
@@ -78,15 +78,29 @@ type Config struct {
 	Namespace []string
 }
 
+// NewRedisClient connects with the correct client type to Redis
+func NewRedisClient(conf config.Redis) *redis.Client {
+	if conf.EnableFailover {
+		return redis.NewFailoverClient(&redis.FailoverOptions{
+			MasterName:    conf.FailoverMasterName,
```
johanstokking: But there's no error checking if this is empty, right?

Shouldn't we fail if failover is set to true but a master name is not defined? That is now silently considered `IsZero()`

johanstokking:
Please also update release notes

cmd/internal/shared/config.go
neoaggelos: I had added `mymaster` as default because it is the typical default master name for Redis Sentinel. I agree with your suggestion, removed the default value

cmd/internal/shared/config.go
neoaggelos: Removed default empty list value as well

pkg/redis/redis.go
neoaggelos: If the user enables failover but has invalid (or missing) arguments, then the client creation fails:

```
$ go run ./cmd/ttn-lw-stack start --redis.enable-failover=true  
  INFO Setting up core component               
(.........)        
  INFO Setting up Network Server               
error:cmd/internal/shared:initialize_network_server (could not initialize Network Server)
--- error:pkg/redis:store (store error)
--- redis: all sentinels are unreachable
exit status 255
```

This is consistent with how the stack fails to initiate when Redis is unavailable:

```
$ go run ./cmd/ttn-lw-stack start --redis.address=localhost:16379
  INFO Setting up core component    
(.........) 
  INFO Setting up Network Server               
error:cmd/internal/shared:initialize_network_server (could not initialize Network Server)
--- error:pkg/redis:store (store error)
--- dial tcp 127.0.0.1:16379: connect: connection refused
exit status 255
```

Is this the desired behaviour?

neoaggelos:
I have updated the Changelog, and I also disabled the logs from the go-redis module.

pkg/redis/redis.go
johanstokking: Yes

pkg/config/shared.go
```diff
@@ -118,14 +118,20 @@ type HTTP struct {
...
 type Redis struct {
-	Address   string   `name:"address" description:"Address of the Redis server"`
-	Password  string   `name:"password" description:"Password of the Redis server"`
-	Database  int      `name:"database" description:"Redis database to use"`
-	Namespace []string `name:"namespace" description:"Namespace for Redis keys"`
+	Address            string   `name:"address" description:"Address of the Redis server"`
+	Password           string   `name:"password" description:"Password of the Redis server"`
+	Database           int      `name:"database" description:"Redis database to use"`
+	Namespace          []string `name:"namespace" description:"Namespace for Redis keys"`
+	EnableFailover     bool     `name:"enable-failover" description:"Enable failover using Redis Sentinel (default false)"`
```
htdvisser: For consistency with the other options, it would be nice to use

`failover.enable`
`failover.addresses`
`failover.master-name`

in the `name` struct tags.

pkg/events/redis/redis.go
```diff
@@ -22,17 +22,15 @@ import (
 	"github.com/go-redis/redis"
 	"go.thethings.network/lorawan-stack/pkg/config"
 	"go.thethings.network/lorawan-stack/pkg/events"
+
```
htdvisser: You can remove this empty line

pkg/config/shared.go
rvolosatovs: In fact, wrapping the failover config into a separate struct and having it as field on `Redis` is a better approach here IMO.
Also I don't think we specify defaults in the description anywhere else?
```suggestion
	Failover struct{
	   Enable     bool `name:"enable" description:"Enable failover using Redis Sentinel"`
	   Addresses  []string `name:"addresses" description:"Redis Sentinel server addresses"`
	   MasterName string   `name:"master-name" description:"Redis Sentinel master name"`
	} `name:"failover"`
```
(or `type RedisFailover struct`)

pkg/config/shared.go
```diff
@@ -118,14 +118,20 @@ type HTTP struct {
...
+	FailoverAddresses  []string `name:"failover-addresses" description:"Redis Sentinel server addresses (when enable-failover=true)"`
+	FailoverMasterName string   `name:"failover-master-name" description:"Redis Sentinel master name (when enable-failover=true)"`
 }
 
 // IsZero returns whether the Redis configuration is empty.
-func (r Redis) IsZero() bool { return r.Address == "" && r.Database == 0 && len(r.Namespace) == 0 }
+func (r Redis) IsZero() bool {
+	return (r.EnableFailover && r.FailoverMasterName == "" && len(r.FailoverAddresses) == 0) ||
```
rvolosatovs: This is not correct - `r.Namespace` and `r.Database` are only checked with `!r.EnableFailover` in proposed implementation, but they have to be checked always

neoaggelos: Yeah, totally missed that. Should be OK now

pkg/config/shared.go
neoaggelos: Added options in a new `type RedisFailover struct`. The options are now `--redis.failover.enable`, etc, like @rvolosatovs suggested.

neoaggelos:
Hello,

> Also I don't think we specify defaults in the description anywhere else?

Default values are only shown for truthy-values, so `(default false)` was hidden. Removed it all the same.

> In fact, wrapping the failover config into a separate struct and having it as field on Redis is a better approach here IMO.

Added a `type RedisFailover struct`.

I also rebased against the new changes in `master`.

pkg/config/shared.go
```diff
@@ -116,16 +116,28 @@ type HTTP struct {
...
+	Database  int           `name:"database" description:"Redis database to use"`
+	Namespace []string      `name:"namespace" description:"Namespace for Redis keys"`
+	Failover  RedisFailover `name:"failover" description:"Redis failover configuration"`
 }
 
 // IsZero returns whether the Redis configuration is empty.
-func (r Redis) IsZero() bool { return r.Address == "" && r.Database == 0 && len(r.Namespace) == 0 }
+func (r Redis) IsZero() bool {
+	return r.Database == 0 && len(r.Namespace) == 0 &&
+		((r.Failover.Enable && r.Failover.MasterName == "" && len(r.Failover.Addresses) == 0) ||
```
rvolosatovs:
```suggestion
		(r.Failover.Enable && r.Failover.MasterName == "" && len(r.Failover.Addresses) == 0) ||
					(!r.Failover.Enable && r.Address == "")
```

cmd/internal/shared/config.go
```diff
@@ -79,6 +79,9 @@ var DefaultRedisConfig = config.Redis{
 	Address:   "localhost:6379",
 	Database:  0,
 	Namespace: []string{"ttn", "v3"},
+	Failover: config.RedisFailover{
```
rvolosatovs: nit; You can also not set `Failover` for the same result.

htdvisser:
Travis is sometimes a bit flaky, no need to worry about that.

@neoaggelos it would be nice if you could do another (interactive) rebase to reword your commit messages so that they are in the imperative mood, so "Fix" instead of "Fixes"/"Fixed" (see also https://github.com/TheThingsNetwork/lorawan-stack/blob/master/CONTRIBUTING.md#commit-messages).

@rvolosatovs will do a final review, and then get this merged.

neoaggelos:
Applied changes requested and squashed all changes in a single commit.

rvolosatovs:
Applied small diff https://github.com/TheThingsNetwork/lorawan-stack/compare/3e83fb856e5602201fc694bb88ea886dd5068772..8f6155d4dc2435601100000d692242dde2506370 and merging

PR #1647, opened 2019-11-27
End Device Claiming API Reference

htdvisser:
#### Summary

This PR adds a reference page for the End Device Claiming APIs.

Related to #206 

#### Changes

- Add reference for End Device Claiming API

Loosly related changes:

- Update proto reference template to include `/api/v3`
- Regenerate API data for doc content (have to do that by hand until https://github.com/TheThingsIndustries/docker-protobuf/issues/33)


htdvisser:
Here's what it looks like:

image

PR #1650, opened 2019-11-27
Remove v prefix from development versions

htdvisser:
#### Summary

This quickfix removes the `v` from `version.TTN` in development builds. In production builds it already didn't have a `v`.

### DIFF
--- a/pkg/version/ttn.go
+++ b/pkg/version/ttn.go
@@ -3,4 +3,4 @@
 package version
 
 // TTN Version
-var TTN = "v3.3.1-dev"
+var TTN = "3.3.1-dev"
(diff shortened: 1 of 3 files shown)


pkg/version/ttn.go
```diff
@@ -3,4 +3,4 @@
 package version
 
 // TTN Version
-var TTN = "v3.3.1-dev"
+var TTN = "3.3.1-dev"
```
rvolosatovs: Not related to this PR, but worth a quickfix as well I guess - shouldn't we rename this variable? E.g. `Stack` or `TTS`?

htdvisser: Not in this PR, but yes. I created an issue for you: https://github.com/TheThingsNetwork/lorawan-stack/issues/1656

Issue #1654, opened 2019-11-27
End Device Search


vsbandaru:
#### Summary
Search functionality in Console by DevEUI, DevID or Name, more fields could be added. 

#### Why do we need this?
Looking up device status to debug/monitor gets harder as the number of devices deployed reaches hundreds or thousands. Most organizations would follow a naming convention for the Device ID/Name and having the Search feature eases the management. 

#### What is already there? What do you see now?
No Search functionality available currently.

#### What is missing? What do you want to see?
A Search UI on the Applications page with the following fields

- Device ID
- Device Name
- Dev EUI
- and more if suggested by other users.

#### Environment
Console

#### How do you propose to implement this?
Implement a Search Header Section on **Applications** page.

#### Can you do this yourself and submit a Pull Request?
No

kschiffer:
Related: https://github.com/TheThingsNetwork/lorawan-stack/issues/73

htdvisser:
API definition for EndDeviceRegistrySearch: https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.3.1/api/search_services.proto#L84-L115

That API allows you to (substring) search by

- ID
- Friendly Name
- Description
- Attributes
- EUI
- DevAddr (*)

A simple version of this API can be implemented on the Identity Server. This implementation would be able to search by all above fields, except DevAddr (the DevAddr is owned by the Network Server, not the Identity Server). I can work on that.

This would be simpler than resolving #73, because we can already scope the search to the Application ID that we know the caller has rights on.

Let's separate the issues (I'll remove EndDevice search form #73) and focus on this one (#1654) first.

kschiffer:
Closing this, since it intersects with #1803. Please continue discussion there if need be.
The issue has a PR and should be resolved soon.

Issue #73, opened 2019-01-29
Entity Search


johanstokking:
**Summary:** 

The Identity Server should provide entity search functionality for Applications, Clients, Gateways, Organizations, Users and End Devices.

**Why do we need this?**

Applications, Clients, Gateways, Organizations, Users mostly for admins, EndDevices for applications with lots of devices.

**What is already there? What do you see now?**

Applications, Clients, Gateways, Organizations, Users search (restricted to admins)

**What is missing? What do you want to see?**

- Wildcards (such as `%`) in search queries aren't properly escaped
- No EndDevice search (#1654)
- Deleted entities aren't shown (they can't be restored yet either)

**How do you propose to implement this?**

I think query escaping is pretty simple. EndDevice search can mostly be copied from entity search. Searching for deleted entities requires API additions (because we can't indicate that an entity is deleted).

Most difficult part here is that search will match all entities (also ones the caller doesn't have rights for), and filtering needs to be done after the results are returned by the database, so this could become pretty expensive.

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/1396 by @htdvisser

PR #1665, opened 2019-11-29
Entity Registry Search

htdvisser:
#### Summary

This PR makes entity search available to non-admin users by considering membership data in the query. It also adds support for searching end devices.

Refs #73 and #1654 

#### Changes

- I started by modifying the membership query so that we can re-use it for search (in addition to regular listing)
- Then I refactored the search queries so that they build on top of the membership query
- Added a small change to leverages Postgres full text search (instead of simple LIKE) when running with Postgres (Cockroach doesn't support FTS)
- Added pagination to search (in both IS and CLI)
- Added search for end devices to IS and CLI.

#### Checklist


cmd/ttn-lw-cli/commands/clients.go
```diff
@@ -94,17 +94,18 @@ var (
...
 			is, err := api.Dial(ctx, config.IdentityServerGRPCAddress)
 			if err != nil {
 				return err
 			}
-			res, err := ttnpb.NewEntityRegistrySearchClient(is).SearchClients(ctx, req)
+			res, err := ttnpb.NewEntityRegistrySearchClient(is).SearchClients(ctx, req, opt)
 			if err != nil {
 				return err
 			}
+			getTotal()
```
rvolosatovs: This looks very weird - maybe add a comment explaining why return value of this function(which returns a `uint64`) is discarded.

Also/or
```suggestion
_ = getTotal()
```
For additional clarity? Explicit `_` shows makes it obvious this is intentional and not a bug

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -223,6 +223,35 @@ var (
...
+
+			is, err := api.Dial(ctx, config.IdentityServerGRPCAddress)
+			if err != nil {
+				return err
+			}
+			res, err := ttnpb.NewEndDeviceRegistrySearchClient(is).SearchEndDevices(ctx, req, opt)
+			if err != nil {
+				return err
+			}
+			getTotal()
```
rvolosatovs: IDEM

cmd/ttn-lw-cli/commands/clients.go
htdvisser: This is also how we do it in other places (all list commands), let's please not change all of that in this PR.

rvolosatovs: What does this do exactly?

htdvisser: https://github.com/TheThingsNetwork/lorawan-stack/blob/master/cmd/ttn-lw-cli/commands/pagination.go#L37-L49

PR #1674, opened 2019-12-02
Display login error for unapproved users

### LINKED ISSUE #1660: Properly inform console users about unapproved accounts
#### Summary
The console needs to properly react to unapproved users by showing them a notification.

#### Why do we need this?
Currently, the user will just be redirected to the login page after trying to log in.

Login page.

A notification about the user account being unapproved (together with some info on how it can be approved)

#### Environment
v.3.3.1 console

During the login process on the console, after being redirected back from the oauth provider, a request to fetch the user will be made which will get a response like this:
```js
{"code":16,"message":"error:pkg/identityserver:user_requested (user account approval is pending)","details":[{"@type":"type.googleapis.com/ttn.lorawan.v3.ErrorDetails","namespace":"pkg/identityserver","name":"user_requested","message_format":"user account approval is pending","correlation_id":"1a5599e1b8804fd3ae8ab5b21adf108b","code":16}]}
```

We could fetch that message and make sure to redirect to the login page with a state object containing a respective info message. This message can then be displayed inside a notification on the login page.

@pgalic

pgalic96:
#### Summary

This draft PR closes #1660 

#### Changes

- Expand reducer to have a login failure option

#### Notes for Reviewers

- @kschiffer this is a first solution, it's quite hacky since I haven't found the reason why the error gets mapped to response instead of getting caught. This wasn't the problem with other requests and the interceptor didn't interfere before. I understand it is more desired to find another approach but I think this is an isolated issue so it might be good enough.

#### Checklist


pkg/webui/console/store/middleware/logics/init.js
```diff
@@ -42,7 +42,11 @@ const consoleAppLogic = createRequestLogic({
       try {
         const userId = info.data.oauth_access_token.user_ids.user_id
         const result = await api.users.get(userId)
-        dispatch(user.getUserMeSuccess(result.data))
+        if (result.code === 7) {
```
htdvisser: Shouldn't you be looking at HTTP response codes?

pgalic96: Yep, indeed.

Issue #1645, opened 2019-11-26
Expired downlinks prevent new ones from being pushed


rvolosatovs:
#### Summary

If expired downlink is present in the NS queue, new dowlinks cannot be pushed

#### Steps to Reproduce

1. Schedule a class B/C downlink with absolute time
2. Wait until the time expires and attempt to schedule a new downlink

#### What do you see now?

```
cli device downlink push test-app test-dev --f_port 0x15 --frm_payload 0xabba  --class-b-c.gateways "test,gateway"
 DEBUG Using access token (valid until 2:40PM) 
 DEBUG ccResolverWrapper: sending update to cc: {[{localhost:8884 0  <nil>}] <nil>}
 DEBUG pickfirstBalancer: HandleSubConnStateChange: 0xc00068d340, CONNECTING
 DEBUG pickfirstBalancer: HandleSubConnStateChange: 0xc00068d340, READY
  WARN Finished unary call                      duration=53.909951ms error=rpc error: code = FailedPrecondition desc = error:pkg/networkserver:downlink_expired (queued downlink is expired) error_correlation_id=91371c6266f34cfb9bbb535d67c6f52d error_name=downlink_expired error_namespace=pkg/networkserver grpc_code=FailedPrecondition grpc_method=DownlinkQueuePush grpc_service=ttn.lorawan.v3.AppAs namespace=grpc
error:pkg/networkserver:downlink_expired (queued downlink is expired)
    correlation_id=91371c6266f34cfb9bbb535d67c6f52d
exit status 255
```

#### What do you want to see instead?

Success

#### How do you propose to implement this?

Instead of validating the whole queue and returning errors if something is invalid, NS shall check the existing queue and uplinks being pushed separately - if anything is invalid in the existing queue(e.g. expired downlink) - corresponding uplink has to be sent to AS and the queue has to be fixed, possibly updating the downlink task as well.
In case downlink being pushed is invalid - the operation simply returns an error.
The existing queue is checked regardless of the downlinks being pushed.

#### Can you do this yourself and submit a Pull Request?

yes

PR #1680, opened 2019-12-03
Fixes and improvements for application downlink validation in Network Server

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1645

#### Changes

- Various fixes and improvements for the application downlink validation
   - Validate per-session
   - Do not validate existing downlinks - assume those are already validated
   - Be smarter in validation
- Record `Recent{Up,Down}links` per-MAC state
- Introduce `lastUplink` for clarity
- Allow Class B/C downlinks to be queued even for devices, which are not currently in respective class - class may change in a lifetime of a MAC state(e.g. by setting `ClassB` bit in uplink). Class B/C downlinks for Class A devices do not stall the queue, but are skipped. By the way, should we add some kind of a limit here?
- Allow downlinks to be enqueued for 2 different sessions - current one and pending one.
- Treat push of no downlinks as noop

#### Notes for Reviewers

You may notice that there are many changes in this PR. That is unfortunate, but necessary - partial merging of these features would result in bugs and we don't want that. All of these changes have to be applied atomically to prevent bugs. Please review commit-by-commit if this is too much to take all at once(also, diffs are not always very useful due to whitespace changes)
One of the reasons why we need `MACState.RecentUplinks` is that, for example, currently NS contains multiple `len(dev.RecentUplinks)` checks, however these checks make no sense, since `RecentUplinks` contain not only data uplinks, but also, e.g. join-requests and the uplinks contained may belong to different sessions. So e.g. `len(dev.RecentUplinks)` cannot be used to determine if a device sent an uplink in a session or not.

#### Checklist


pkg/networkserver/downlink.go
```diff
@@ -1111,7 +1140,7 @@ func (ns *NetworkServer) processDownlinkTask(ctx context.Context) error {
 
 					rxDelay := ttnpb.RxDelay(phy.JoinAcceptDelay1 / time.Second)
 
-					rx1, rx2, paths := downlinkPathsForClassA(rxDelay, dev.RecentUplinks...)
+					rx1, rx2, paths := downlinkPathsForClassA(rxDelay, up)
```
rvolosatovs: Only the join-accept should be used here

pkg/networkserver/downlink.go
```diff
@@ -1184,7 +1213,7 @@ func (ns *NetworkServer) processDownlinkTask(ctx context.Context) error {
 
 				var maxUpLength uint16 = math.MaxUint16
 				if dev.MACState.LoRaWANVersion == ttnpb.MAC_V1_1 {
-					maxUpLength = maximumUplinkLength(fp, phy, dev.RecentUplinks...)
+					maxUpLength = maximumUplinkLength(fp, phy, dev.MACState.RecentUplinks...)
```
rvolosatovs: It makes no sense to look at uplinks from various sessions, only the uplinks from current session should be looked at

rvolosatovs:
Existing `Recent{Up,Down}links` can be just discarded - at most 1 uplink and 1 downlink in existing OTAA sessions per-device is required to make everything work as it should. Most of existing checks make no sense anyway.

PR #1722, opened 2019-12-12
Fix incorrect clear of time field for downlink messages

virtualguy:
#### Summary
The NS was incorrectly dropping the time when scheduling a downlink and therefore not requesting the gateway to use gps/absolute time using the tmms field (for UDP forwarders)

#### Changes
Fix absolute time for class b/c downlinks

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/io.go
+++ b/pkg/gatewayserver/io/io.go
@@ -339,6 +339,7 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
 			settings.Downlink.InvertPolarization = true
 		}
 		var f func(context.Context, int, ttnpb.TxSettings, scheduling.RTTs, ttnpb.TxSchedulePriority) (scheduling.Emission, error)
+		settings.Time = nil
 		switch request.Class {
 		case ttnpb.CLASS_A:
 			f = c.scheduler.ScheduleAt
@@ -362,7 +363,6 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
 			rxErrs = append(rxErrs, errRxWindowSchedule.WithCause(err).WithAttributes("window", i+1))
 			continue
 		}
-		settings.Time = nil
 		settings.Timestamp = uint32(time.Duration(em.Starts()) / time.Microsecond)
 		msg.Settings = &ttnpb.DownlinkMessage_Scheduled{
 			Scheduled: &settings,


pkg/gatewayserver/io/io.go
```diff
@@ -339,6 +339,7 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
 			settings.Downlink.InvertPolarization = true
 		}
 		var f func(context.Context, int, ttnpb.TxSettings, scheduling.RTTs, ttnpb.TxSchedulePriority) (scheduling.Emission, error)
+		settings.Time = nil
```
KrishnaIyer: I think it's better to move this out of this function. So maybe after `L340`.

johanstokking: This is ineffective

pkg/gatewayserver/io/io.go
```diff
@@ -362,7 +363,6 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
 			rxErrs = append(rxErrs, errRxWindowSchedule.WithCause(err).WithAttributes("window", i+1))
 			continue
 		}
-		settings.Time = nil
```
johanstokking: GS scheduler does the translation from absolute time to timestamp, instead of letting the gateway do that. The reason for this is that GS does conflict detection and mitigation, without having to make roundtrips to the gateway.

GS synchronizes GPS time on uplink messages, so it knows which timestamp belongs to which GPS absolute time. If there's no GPS time, it uses RTT (round-trip times) to calculate the timestamp:

https://github.com/TheThingsNetwork/lorawan-stack/blob/72cdab470c1fdfc8e2c0b9f274a255784b931ab1/pkg/gatewayserver/scheduling/scheduler.go#L186-L193

We basically don't use `tmms` in The Things Stack.

virtualguy: In the testing we have done, the gateway transmits the message almost immediately after the http downlink is sent. Is there a minimum time we need to schedule the message in advance (we are currently scheduling approximately 5 seconds in advance). The gateway is on a fast ethernet connection and push time is around 30ms and the ec2 instance the NS is running on is in accurate NTP sync so it should be calculating correctly.

virtualguy: Should I close this and open a bug report?

KrishnaIyer: Yes please. We'd really appreciate Stack and Gateway logs as that will help us debug the issue.

johanstokking: @virtualguy yes please file an issue.

Are you specifying absolute time in the downlink message at all? If you file the issue, please include the full (redacted) payload that you send to the webhook handler.

PR #1725, opened 2019-12-12
Improve user id validation in oauth

### LINKED ISSUE #1705: Validate `user_id` in OAuth views
#### Summary

We need to do better frontend validation of the user ID field in OAuth views.

#### Why do we need this?

We're seeing that users are making too many mistakes with user IDs in the OAuth views

First of all, the validations are inconsistent. We sometimes see

```js
user_id: Yup.string().required(sharedMessages.validateRequired)
```

while in other places we also see a min/max length.

The user_id should be validated the same way everywhere:

- Required
- Min length 3
- Max length 36
- Regex `^[a-z0-9](?:[-]?[a-z0-9]){2,}$`

> NOTE: maybe out of scope for this issue, but it might be good to check all ID fields in the console as well. All IDs have the same rules.

Write once, copy-paste many.

pgalic96:
#### Summary

Closes #1705 

#### Changes

- Add user regex
- Improve user validation and add it to all `oauth` views

#### Checklist

### DIFF
--- a/pkg/webui/lib/regexp.js
+++ b/pkg/webui/lib/regexp.js
@@ -12,6 +12,5 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/* eslint-disable import/prefer-default-export */
-
 export const url = /[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)?/gi
+export const id = /^[a-z0-9](?:[-]?[a-z0-9]){2,}$/
--- a/pkg/webui/lib/shared-messages.js
+++ b/pkg/webui/lib/shared-messages.js
@@ -227,6 +227,7 @@ export default defineMessages({
   validateTooLong: 'Too long',
   validateTooShort: 'Too short',
   validateUrl: 'Invalid URL format',
+  validateIdFormat: 'IDs can only contain lowercase letters, numbers and dashes (-)',
   webhookBaseUrl: 'Base URL',
   webhookFormat: 'Webhook Format',
   webhookId: 'Webhook ID',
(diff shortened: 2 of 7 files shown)


pkg/webui/lib/regexp.js
```diff
@@ -15,3 +15,4 @@
 /* eslint-disable import/prefer-default-export */
 
 export const url = /[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)?/gi
+export const user = /^[a-z0-9](?:[-]?[a-z0-9]){2,}$/
```
htdvisser: I'd just call this `id` as it's the same for all IDs.

pkg/webui/lib/shared-messages.js
```diff
@@ -215,6 +215,7 @@ export default defineMessages({
   validateTooLong: 'Too long',
   validateTooShort: 'Too short',
   validateUrl: 'Invalid URL format',
+  validateUserFormat: 'Username can contain only dash (-) as the special char',
```
htdvisser: (any) ID can be 3-36 characters (already covered by your other validations) and can only contain lowercase letters, numbers and dashes (-)

pgalic96: alright. there are some cases not covered though, but in my opinion its too specific to be mentioned. these cases are if the username is for example `p-g` or any id that has dash in the end `tti-` for example.

htdvisser: I agree. If someone tries to use `----` as a username, they can probably already imagine that we wouldn't accept it.

pkg/webui/lib/regexp.js
```diff
@@ -15,3 +15,4 @@
 /* eslint-disable import/prefer-default-export */
```
kschiffer:
```suggestion

```

pkg/webui/lib/shared-messages.js
```diff
@@ -215,6 +215,7 @@ export default defineMessages({
   validateTooLong: 'Too long',
   validateTooShort: 'Too short',
   validateUrl: 'Invalid URL format',
+  validateIdFormat: 'ID can only contain lowercase letters, numbers and dashes (-)',
```
```suggestion
  validateIdFormat: 'IDs can only contain lowercase letters, numbers and dashes (-)',
```

PR #1726, opened 2019-12-12
Documentation clean up

SebastienCHEVALIER:
#### Summary
Changes from https://github.com/TheThingsIndustries/lorawan-stack/pull/1845 belonging to TTS.

#### Changes
~~Replace to tts shortcode in running-the-stack.md~~
Remove summary on email-templates.md and put why is it section as first paragraph

### DIFF
--- a/doc/content/concepts/email-templates.md
+++ b/doc/content/concepts/email-templates.md
@@ -2,12 +2,11 @@
 title: "Email Templates"
 description: ""
 weight: 15
-summary: Email templates define the contents of the emails that The Things Stack sends to its users. They allow network operators to override the default contents of the emails with their own custom contents.
 ---
 
-## What is it?
-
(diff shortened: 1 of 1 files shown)


doc/content/concepts/email-templates.md
```diff
@@ -2,7 +2,6 @@
 title: "Email Templates"
 description: ""
 weight: 15
-summary: Email templates define the contents of the emails that The Things Stack sends to its users. They allow network operators to override the default contents of the emails with their own custom contents.
```
johanstokking: This is not redundant. Can you replace `The Things Stack` by the shortcode here?

htdvisser: No, you can't. I think it would be better to remove the `What is it` header and add a `<!--more-->` below the first paragraph.

johanstokking: I'm also fine with that. As long as it renders nicely on a tile

Issue #1631, opened 2019-11-22
Add Kerlink Wirnet connection guide


rvolosatovs:
#### Summary

Document connecting Kerlink Wirnet

#### Why do we need this?

To get people to use our network

#### What is already there? What do you see now?

Nothing

#### What is missing? What do you want to see?

A clear guide on how to connect Kerlink Wirnet to a specific v3 stack instance

#### How do you propose to implement this?

- Create a script, which would generate DOTA configured to connect to specified host and use specified FP, which users can install on their gateways. E.g. `generate-dota.sh <my stack instance> <my FP>`
- Document how to use the script to configure the gateway
- Document how to update the gateway to latest firmware
- Provide the latest firmware - I suppose that goes to https://github.com/TheThingsNetwork/kerlink-station-firmware ? (cc @htdvisser) - if so, let's make that repo look a bit nicer?
   1. Let's use releases page for actual packages.
   2. Let's add the script, which would generate the required DOTA there by using the latest firmware release in the repo.
   3. @htdvisser I will need the access to the repo as well(preferably admin)

#### Can you do this yourself and submit a Pull Request?

yes

rvolosatovs:
Blocked by https://github.com/TheThingsNetwork/lorawan-stack/pull/1619

Let's use the format there as an example

rvolosatovs:
cc @johanstokking

htdvisser:
@rvolosatovs you're now maintainer on https://github.com/TheThingsNetwork/kerlink-station-firmware

I don't think we should provide Kerlink firmware or packages like the Kerlink Packet Forwarder or Semtech Packet Forwarder, but we can definitely improve the repo.

htdvisser:
Assigning @johanstokking and myself for the discussion

rvolosatovs:
> I don't think we should provide Kerlink firmware or packages like the Kerlink Packet Forwarder or Semtech Packet Forwarder, but we can definitely improve the repo.

Why not? We already do that in https://github.com/TheThingsNetwork/kerlink-station-firmware, right?
All I propose is to update the versions and do it in a more systematic way.

johanstokking:
I think we should have sane defaults, including binaries, and a script to build them and set host names. Also, it would be nice to have pre-built packages for common deployments (PCN regions, CH regions).

rvolosatovs:
Blocked until I manage to actually get class B to work

htdvisser:
[...]

PR #1716, opened 2019-12-10
Document connecting Kerlink Wirnet Station to TTS

rvolosatovs:
#### Summary

Closes #1631 

#### Changes

- Document connecting Kerlink Wirnet Station to TTS

#### Notes for Reviewers

~Leaving this as draft until https://github.com/TheThingsNetwork/kerlink-station-firmware/pull/7 is merged~

Do we need to document this in the `CHANGELOG`? I guess not, but @htdvisser may have a different opinion.

Based this on existing TTKG guide. ~`Registration` section is just copy-paste~

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, database and configuration, according to the stability commitments in `README.md`.
- [x] Testing: The changes are covered with unit tests. The changes are tested manually as well.
- [x] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #1747, opened 2019-12-18
Add initial CPF support to GCS

rvolosatovs:
#### Summary

References #1631 https://github.com/TheThingsNetwork/kerlink-station-firmware/pull/9

#### Changes

- Add support for (some of) CPF configuration to GCS - enough to provision the gateway to forward traffic to the configured stack instance
- Improve testing(e.g. actually test the output on endpoints)

#### Notes for Reviewers

I will file issue(s) for added support for CPF in GCS before "undrafting" this

#### Checklist

### DIFF
--- a/pkg/gatewayconfigurationserver/gatewayconfigurationserver.go
+++ b/pkg/gatewayconfigurationserver/gatewayconfigurationserver.go
@@ -16,6 +16,7 @@ package gatewayconfigurationserver
 
 import (
 	"context"
+	"encoding"
 	"net/http"
 
 	"github.com/gogo/protobuf/types"
@@ -24,6 +25,7 @@ import (
 	bscups "go.thethings.network/lorawan-stack/pkg/basicstation/cups"
 	"go.thethings.network/lorawan-stack/pkg/component"
 	"go.thethings.network/lorawan-stack/pkg/gatewayconfigurationserver/gcsv2"
+	"go.thethings.network/lorawan-stack/pkg/pfconfig/cpf"
 	"go.thethings.network/lorawan-stack/pkg/pfconfig/semtechudp"
 	"go.thethings.network/lorawan-stack/pkg/ttnpb"
 	"go.thethings.network/lorawan-stack/pkg/web"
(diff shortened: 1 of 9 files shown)


pkg/pfconfig/cpf/cpf.go
```diff
@@ -0,0 +1,128 @@
...
+	fp, err := fps.GetByID(gtw.FrequencyPlanID)
+	if err != nil {
+		return nil, err
+	}
+	sx1301Conf, err := shared.BuildSX1301Config(fp)
+	if err != nil {
+		return nil, err
+	}
+	var gatewayConf LoradGatewayConf
+	if len(gtw.Antennas) > 0 {
```
rvolosatovs: Not sure if this makes sense - it seems that `SX1301` only allows for 1 antenna gain to be configured, what should happen when there are multiple?
Same question for location.

rvolosatovs:
Blocking https://github.com/TheThingsNetwork/lorawan-stack/issues/1631 https://github.com/TheThingsNetwork/lorawan-stack/pull/1716 (transitively)

pkg/gatewayconfigurationserver/gatewayconfigurationserver_test.go
```diff
@@ -127,20 +136,89 @@ func TestWeb(t *testing.T) {
...
+					switch res.StatusCode {
+					case http.StatusOK:
+						if !a.So(res.Header.Get("Content-Type"), should.Equal, "application/json; charset=UTF-8") {
+							t.FailNow()
+						}
+						b, err := ioutil.ReadAll(res.Body)
+						if err != nil {
+							t.Fatalf("Failed to read response body: %s", err)
+						}
+						a.So(string(b), should.Equal, string(test.Must(json.MarshalIndent(test.Must(semtechudp.Build(is.res.Get, fps)).(*semtechudp.Config), "", "\t")).([]byte))+"\n")
```
KrishnaIyer: Please split this to make it easier to read and update

pkg/gatewayconfigurationserver/gatewayconfigurationserver_test.go
```diff
@@ -127,20 +136,89 @@ func TestWeb(t *testing.T) {
...
+					switch res.StatusCode {
+					case http.StatusOK:
+						if !a.So(res.Header.Get("Content-Type"), should.Equal, "application/json; charset=UTF-8") {
+							t.FailNow()
+						}
+						b, err := ioutil.ReadAll(res.Body)
+						if err != nil {
+							t.Fatalf("Failed to read response body: %s", err)
+						}
+						a.So(string(b), should.Equal, string(test.Must(json.MarshalIndent(test.Must(cpf.BuildLorad(is.res.Get, fps)).(*cpf.LoradConfig), "", "\t")).([]byte))+"\n")
```
KrishnaIyer: IDEM

pkg/gatewayconfigurationserver/gatewayconfigurationserver_test.go
```diff
@@ -127,20 +136,89 @@ func TestWeb(t *testing.T) {
...
+					switch res.StatusCode {
+					case http.StatusOK:
+						if !a.So(res.Header.Get("Content-Type"), should.Equal, "application/toml") {
+							t.FailNow()
+						}
+						b, err := ioutil.ReadAll(res.Body)
+						if err != nil {
+							t.Fatalf("Failed to read response body: %s", err)
+						}
+						a.So(string(b), should.Equal, string(test.Must(test.Must(cpf.BuildLorafwd(is.res.Get)).(*cpf.LorafwdConfig).MarshalText()).([]byte)))
```
KrishnaIyer: Also here

pkg/gatewayconfigurationserver/gatewayconfigurationserver.go
```diff
@@ -47,6 +49,56 @@ type GatewayConfigurationServer struct {
 	config *Config
 }
 
+func (gcs *GatewayConfigurationServer) makeGatewayConfigHandler(f func(context.Context, echo.Context, *ttnpb.Gateway) error) func(echo.Context) error {
```
KrishnaIyer: Nit: Maybe you can drop the word `Gateway` in the function names to make them shorter. Within the context of this service, it's clear that it's gateway config.

pkg/gatewayconfigurationserver/gatewayconfigurationserver.go
```diff
@@ -47,6 +49,56 @@ type GatewayConfigurationServer struct {
...
+	return gcs.makeGatewayConfigHandler(func(ctx context.Context, c echo.Context, gtw *ttnpb.Gateway) error {
+		msg, err := f(ctx, gtw)
+		if err != nil {
+			return err
+		}
+		return c.JSONPretty(http.StatusOK, msg, "\t")
+	})
+}
+
+func (gcs *GatewayConfigurationServer) makeTextMarshalerGatewayConfigHandler(contentType string, f func(context.Context, *ttnpb.Gateway) (encoding.TextMarshaler, error)) func(echo.Context) error {
```
KrishnaIyer: Nit: "TextMarshalling" is a golang term right? I think it would be nice to use the output file format in the funtion name. For example, `makeJSON`, `makeTOML`, `makeYAML` etc..

pkg/pfconfig/cpf/cpf.go
KrishnaIyer: IIRC, with the v1 and v1.5 ref design, there's usually only one antenna per concentrator and each concentrator gets one SX1301_Conf object.
@htdvisser can confirm.

pkg/pfconfig/shared/shared.go
```diff
@@ -365,3 +367,31 @@ func BuildSX1301Config(frequencyPlan *frequencyplans.FrequencyPlan) (*SX1301Conf
 
 	return conf, nil
 }
+
+// DefaultGatewayServerPort is the default port used for connecting to Gateway Server.
```
KrishnaIyer: `DefaultGatewayServerUDPPort`

pkg/gatewayconfigurationserver/gatewayconfigurationserver.go
rvolosatovs: It's named this way, because this method uses the `MarshalText` *method* on an `encoding.TextMarshaler`.
E.g. `makeJSONHandler` takes an `interface{}`

KrishnaIyer: I still think that it's better to name a function after what it creates rather than what method it calls internally. But again this is nitpick.

rvolosatovs: This method does not know what it creates - that's the point.
The content-type is a parameter - it may create TOML, JSON, YAML or anything else.

PR #1776, opened 2019-12-23
Display disabled APIs correctly

### LINKED ISSUE #1614: Disabled APIs displayed incorrectly
#### Summary

The console displays disabled servers incorrectly

#### Steps to Reproduce

1. Open any console on a JS-only deployment

#### What do you see now?

Screenshot 2019-11-20 at 17 04 42

- Color of disabled servers shouldn't be red, but perhaps gray.
- Descenders shouldn't fall off the letters.
- Address should be hidden for disabled servers.
- The console shouldn't crash when `base_url` for disabled components is omitted by the backend (Warning: Failed prop type: The prop `env.config.stack.gs.base_url` is marked as required in `ConsoleApp`, but its value is `undefined`.)

Note: The backend shouldn't include a `base_url` for disabled APIs. I tried to change that by adding the following in `pkg/webui/types.go`. Please include that code in the PR for this issue.

```go
// MarshalJSON implements json.Marshaler.
func (c APIConfig) MarshalJSON() ([]byte, error) {
	out := struct {
		Enabled bool   `json:"enabled"`
		BaseURL string `json:"base_url,omitempty"`
	}{
		Enabled: c.Enabled,
		BaseURL: c.BaseURL,
	}
	if !out.Enabled {
		out.BaseURL = ""
	}
	return json.Marshal(out)
}
```

#### Environment

v3.2.6 JS-only deployment

pgalic96:
#### Summary

Closes #1614 

#### Changes

- Restructure component card such that the component urls have enough space to be expanded
- Add a check if component base url is provided (cc @htdvisser )
- Change disabled component status to `unknown` (grey)
- Fix prop types accordingly

#### Notes for Reviewers

@kschiffer I made changes to the component card since Hylke wanted it to be displayed fully. I organized it in a list one under each other to add some structure since if we expand it based on the length of the url it can get quite messy. Maybe the better way would be to expand the component urls on hover.

#### Checklist

### DIFF
--- a/pkg/webui/console/views/overview/index.js
+++ b/pkg/webui/console/views/overview/index.js
@@ -55,7 +55,7 @@ const m = defineMessages({
   gotoApplications: 'Go to applications',
   gotoGateways: 'Go to gateways',
   welcome: 'Welcome to the Console!',
-  welcomeBack: 'Welcome back, {userId} 👋!',
+  welcomeBack: 'Welcome back, {userId}! 👋',
   getStarted: 'Get started right away by creating an application or registering a gateway.',
   continueWorking: 'Walk right through to your applications and/or gateways.',
   componentStatus: 'Component Status',
--- a/pkg/webui/console/views/overview/index.js
+++ b/pkg/webui/console/views/overview/index.js
@@ -251,7 +251,7 @@ export default class Overview extends React.Component {
                 }
                 const component = stackConfig[componentKey]
                 const name = componentMap[componentKey]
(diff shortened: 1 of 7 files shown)


htdvisser:
Screenshot 2019-12-23 at 13 33 42

My complaint was not that the end onf the domain name was missing (which is also good to fix), but that the descenders of letters are cut off, in the above screenshot, the `j` and `g`.

pgalic96:
alright, thats fixed as well I think.

PR #1817, opened 2020-01-07
Subscribe to multiple events sources if necessary (JS SDK)

### LINKED ISSUE #1719: Subscribe to multiple events sources in Console
#### Summary

Subscribe to multiple event sources in the Console.

#### Steps to Reproduce

1. Deploy IS in cluster A, and Console and other components in cluster B
2. Console connects to cluster A for events, as only the events source of IS is used. Events from B are now subscribed to

#### What do you see now?

No events from components that are not in the IS cluster.

Events from the cluster where the Console is in.

#### Environment

3.3.2

Depending on where you open the event stream, subscribe to the addresses that are relevant. For example;

* Application-domain: IS, JS, NS, AS, DTC, DCS
* Gateway-domain: IS, GS

@htdvisser noted that there may be a limitation in the number of concurrent requests. In practice, we won't need more than 2 different sources: the IS and the cluster of the Console.

This may be a three-stage rocket:

1. Keep one subscription, but use the local cluster and not IS. This makes at least application and gateway events work, but breaks the entity events
2. Use two subscriptions; the IS and the local cluster (closes this issue)
3. Use an event server (optional, needs https://github.com/TheThingsIndustries/lorawan-stack/issues/1804)

Can test

kschiffer:
#### Summary
Closes #1719 

#### Changes
- Add utility to combine streams to JS SDK
- Refactor streaming function in entity services to subscribe to multiple sources (if necessary)

#### Notes for Reviewers
The way this works now is that we can define what components are relevant per stream (eg. IS, GS for gateway streams) and the SDK will determine whether it would need to connect to multiple hosts to do so. It will then open a stream for every distinct host but keep the subscription interface as it is. See #1719 for more info.
The change is solely SDK scoped, no changes needed in the console to resolve the issue.

@johanstokking which components are relevant for the organization stream? I suppose it's only IS, but I'm not sure. For the device stream, I took the same as for applications: IS, AS, JS, NS, DTC. Please check whether this solves the issue.

In theory, this could mean that the SDK will open up to 6 different connections, which would block all other connections of the browser. However, it's very unlikely that all stack components are on different hosts. Still, it might be necessary to add logic to limit the number of connections.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Respect stack components on different hosts when connected to event sources in the Console.
+
 ### Security
 
 ## [3.4.2] (2020-01-08)
--- a/sdk/js/src/service/applications.js
+++ b/sdk/js/src/service/applications.js
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 import Marshaler from '../util/marshaler'
+import { getComponentsWithDistinctBaseUrls } from '../util/stack-components'
+import combineStreams from '../util/combine-streams'
 import Devices from '../service/devices'
 import Application from '../entity/application'
 import ApiKeys from './api-keys'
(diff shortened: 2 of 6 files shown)


kschiffer:
@johanstokking can you verify that this resolves the issue?

johanstokking:
> @johanstokking which components are relevant for the organization stream? I suppose it's only IS, but I'm not sure

I would say only IS indeed. @htdvisser thoughts?

>  For the device stream, I took the same as for applications: IS, AS, JS, NS, DTC. Please check whether this solves the issue.

Yes, that's good.

> However, it's very unlikely that all stack components are on different hosts. Still, it might be necessary to add logic to limit the number of connections.

This won't practically be the case.

htdvisser:
Indeed only the IS publishes organization events.

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Respect stack components on different hosts when connecting to event sources in the Console.
```
johanstokking: nit; changelog is in past tense

htdvisser: No it's not. See https://keepachangelog.com/en/1.0.0/

kschiffer: Hahah now I pushed a change already.

Issue #1824, opened 2020-01-08
Support Multiple Frequency Plans in Scheduler


KrishnaIyer:
#### Summary

Support Multiple Frequency Plans in Scheduler (refs #761)

#### Why do we need this?

Enable downlinks on multiple concentrators.

#### What is already there? What do you see now?

Branch `feature/761-multi-fplan-gtw` contains the rest of the items discussed in #761.

#### What is missing? What do you want to see?

Use the supported `FrequencyPlans` struct in the scheduler and use the `frequency_plan_id` as part of the `TxRequest` to choose the correct downlink parameters.

#### Environment

ttn-lw-stack v3.4.1

#### How do you propose to implement this?

Better left to the expert.

#### Can you do this yourself and submit a Pull Request?

Can review and test if necessary.

johanstokking:
Can you see if you can pick this up? Would be great as a separate PR and not part of #1826.

KrishnaIyer:
Ok. I'll take this.

KrishnaIyer:
To summarise our discussions here:
- For now the FPs have to be from the same band.

As for combining frequency plans in the scheduler
- The timeOffAir parameter should be the same (same value or unset) across the FPs?
- As for sub bands, it would be a **union** of the frequencies right? The sub-bands encapsulate the difference in duty cycling so creating sub-bands based on the union of the FPs should work right?

johanstokking:
> * The timeOffAir parameter should be the same (same value or unset) across the FPs?

Yes. If they don't match we can safely error.

> * As for sub bands, it would be a **union** of the frequencies right? The sub-bands encapsulate the difference in duty cycling so creating sub-bands based on the union of the FPs should work right?

That would be best. There are three scenarios then;

1. There are no FR sub-bands defined, which is great
2. The same sub-band for all FPs is configured, this is also good
3. The sub-bands of the given frequency plans do not overlap, this is also fine
4. The sub-bands of the given frequency plans are not the same and overlap. This is not good

KrishnaIyer:
closed via #1859

PR #1826, opened 2020-01-08
Support multiple frequency plans on gateways

### LINKED ISSUE #761: Support Generation of router_config for 16/64 Channel Basic Station gateways
#### Summary

Support gateway config for gateways with >8 channels.

#### Why do we need this?

This is to support configuration of 16 ch gateways Ex: TTOG US Version and potentially higher channels.

- `pfconfig` supports only one `SX1301_conf` object.
- `lns` generates only one `SX1301_conf`.

- Support combining channels from frequency plans from different frequency plans and pass that to the Basic Station.

#### Environment

All

- [x] API: introduce a repeated string frequency_plan_ids to Gateway. Falls in scope of this issue
- [x] ER: handle new field as described in #1353 (comment) (i.e. staying backward compatible with clients who only know frequency_plan_id, potentially having them accidentally reset multiple plans). Falls in scope of this issue
GS
- [x] Support multiple frequency plans in Basic Station LNS. Falls in scope of this issue
  - [x] promote basicstation `router_config` to `pkg/pfconfig`
  - [x] make `sx1301` a slice.
  - [x] Generate one `SX1301_conf` object for each configured frequency plan.
  - [x] Update the `hwspec` field of the `router_config` for basicstation.
- [x] Support multiple frequency plans in the scheduler. This needs a separate issue, scoped to same band frequency plans, referencing #1394 to support mixed bands
- [x] GCS: add support to return a UDP packet forwarder global_conf.json with multiple concentrator support. This needs a separate issue: #1837 

Yes.

cc: @htdvisser

KrishnaIyer:
#### Summary

Closes #761 

#### Changes

- Add `frequency_plan_ids` to proto and corresponding changes to IS (by @htdvisser) 
- Add `frequency_plan_id` to `TxRequest` and use that in `ScheduleDown`.
- Move `router_config` code to `pkg/pfconfig` and support generation with multiple frequency plans. 
- Adapt code and tests.

#### Notes for Reviewers

@htdvisser : I haven't modified your commits. Please check if [](https://github.com/TheThingsNetwork/lorawan-stack/commit/696796acfae2f55252e15459cc953db22e810fab) is necessary as it was in your branch.
@rvolosatovs: Since the `router_config` stuff is very specific to the Basic Station spec, it's better if @johanstokking looks at it.

This code doesn't compile since the schedular isn't ready to accept multiple frequency plans. Though All tests pass when I hacked this part. Please target the PR for #1824 to this branch.

Once that is merged, I'll close this draft and open a PR.

#### Checklist


pkg/pfconfig/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,181 @@
...
+
+// GetRouterConfig returns the routerconfig message to be sent to the gateway.
+// Currently as per the basic station docs, all frequency plans have to be from the same region (band) https://doc.sm.tc/station/tcproto.html#router-config-message.
+func GetRouterConfig(fps []*frequencyplans.FrequencyPlan, isProd bool, dlTime time.Time) (RouterConfig, error) {
+	for _, fp := range fps {
+		if err := fp.Validate(); err != nil {
+			return RouterConfig{}, errFrequencyPlan
+		}
+	}
+	cfg := RouterConfig{}
```
rvolosatovs:
```suggestion
	conf := RouterConfig{}
```
https://github.com/TheThingsNetwork/lorawan-stack/blob/master/DEVELOPMENT.md#single-word-entities

pkg/pfconfig/basicstationlns/basicstationlns.go
```diff
@@ -0,0 +1,181 @@
...
+
+	MuxTime float64 `json:"MuxTime"`
+}
+
+// MarshalJSON implements json.Marshaler.
+func (cfg RouterConfig) MarshalJSON() ([]byte, error) {
+	type Alias RouterConfig
+	return json.Marshal(struct {
+		Type string `json:"msgtype"`
+		Alias
```
rvolosatovs: +1 point for creativity
-2 points for clarity

Not necessary to change, but isn't there a simpler way to do this?

pkg/gatewayserver/gatewayserver.go
```diff
@@ -381,18 +381,24 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids
...
 			DownlinkPathConstraint: ttnpb.DOWNLINK_PATH_CONSTRAINT_NONE,
 		}
 	} else if err != nil {
 		return nil, err
 	}
-	fp, err := gs.FrequencyPlans.GetByID(gtw.FrequencyPlanID)
-	if err != nil {
-		return nil, err
+
+	var fps []*frequencyplans.FrequencyPlan
```
johanstokking: You can `make` the slice here with known capacity

pkg/gatewayserver/gatewayserver.go
```diff
@@ -612,21 +618,29 @@ func (gs *GatewayServer) GetFrequencyPlan(ctx context.Context, ids ttnpb.Gateway
...
+		fpID, ok := frequencyplans.FallbackIDFromContext(ctx)
 		if !ok {
 			return nil, err
 		}
+		fpIDs = append(fpIDs, fpID)
 	} else {
 		return nil, err
 	}
-	return gs.FrequencyPlans.GetByID(fpID)
+	var fps []*frequencyplans.FrequencyPlan
```
johanstokking: Idem

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -375,7 +376,8 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 					logger.WithError(err).Debug("Failed to unmarshal join-request message")
 					return nil
 				}
-				up, err := jreq.ToUplinkMessage(ids, fp.BandID, receivedAt)
+				// According to spec, all frequency plans must be from the same band, so choosing the first one.
```
johanstokking: We can and must enforce this everywhere, at IS level. cc @htdvisser

pkg/gatewayserver/io/io.go
```diff
@@ -263,7 +265,19 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
...
 	}
-	phy, err := band.GetByID(c.fp.BandID)
+	var fpIndex int
+	var gtwFP *frequencyplans.FrequencyPlan
+	for fpIndex, gtwFP = range c.fps {
+		if gtwFP.BandID == request.FrequencyPlanID {
+			break
+		}
+	}
+	if fpIndex == len(c.fps) {
```
johanstokking: Unclear and too much heuristic of how indexer of for-range works. Just assign `gtwFP` in the loop and check here if it's assigned

pkg/gatewayserver/io/io.go
```diff
@@ -245,6 +245,8 @@ func (c *Connection) SendDown(msg *ttnpb.DownlinkMessage) error {
 	return nil
 }
 
+var errUnconfiguredFrequencyPlan = errors.DefineInvalidArgument("unconfigured_frequency_plan", "frequency plan `{name}` is not configured for this gateway")
```
johanstokking: `errFrequencyPlanNotConfigured`

pkg/identityserver/application_registry.go
```diff
@@ -169,16 +169,18 @@ func (is *IdentityServer) listApplications(ctx context.Context, req *ttnpb.ListA
...
-				apps.Applications[i] = app.PublicSafe()
-			}
-		}
 		return nil
 	})
 	if err != nil {
 		return nil, err
 	}
+
+	for i, app := range apps.Applications {
```
johanstokking: How is this change related?

pkg/pfconfig/basicstationlns/basicstationlns.go
KrishnaIyer: Hmm. I don't see one. Do you have a suggestion?

pkg/identityserver/gateway_registry.go
```diff
@@ -57,6 +60,28 @@ func (is *IdentityServer) createGateway(ctx context.Context, req *ttnpb.CreateGa
...
+			}
+			fp2, err := is.FrequencyPlans.GetByID(req.FrequencyPlanIDs[i+1])
+			if err != nil {
+				return nil, err
+			}
+			if fp1.BandID != fp2.BandID {
+				return nil, errFrequencyPlans
+			}
+		}
+	}
```
htdvisser: I'm not sure the Identity Server should do this kind of validation. Doing that would prevent us from having (test) clusters that support experimental frequency plans.

If we do want to do this, I would write it as:

```go
if len(req.FrequencyPlanIDs) > 0 {
	fp0, err := is.FrequencyPlans.GetByID(req.FrequencyPlanIDs[0])
	if err != nil {
		return nil, err
	}
	for i := 1; i < len(req.FrequencyPlanIDs); i++ {
		fpn, err := is.FrequencyPlans.GetByID(req.FrequencyPlanIDs[i])
		if err != nil {
			return nil, err
		}
		if fpn.BandID != fp0.BandID {
			return nil, errFrequencyPlans
		}
	}
}
```

instead of fetching two frequency plans in every iteration of the loop

pkg/identityserver/application_registry.go
htdvisser: Consistency. If we move the same logic out of the transaction for Gateways, we should also do it in the other places.

pkg/identityserver/gateway_registry.go
KrishnaIyer: @johanstokking ?

johanstokking: I don't think we should be enforcing this in ER indeed; it doesn't have any authority in frequency plans.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
johanstokking: So this ain't true anymore

pkg/identityserver/gateway_registry.go
KrishnaIyer: Ok? Then did I mis understand your comment here:https://github.com/TheThingsNetwork/lorawan-stack/pull/1826#discussion_r364665915?

johanstokking:
@KrishnaIyer what is the scope for review now? This PR is draft and CI is still failing. Let me know what to review and what the plan is with merging this

KrishnaIyer:
- I've fixed the failing tests (I think). 
- I've moved the check for Bands from IS to GS (while connection).
- Apart from this, can you check if the scheduler needs anything?  As we discussed, since we enforce that all frequency plans have the same band, the scheduler currently receives the first FP.
- Also, this change would require the NS to send the `FrequencyPlanID` as part of the `TxRequest`. Ping @rvolosatovs

rvolosatovs:
> Also, this change would require the NS to send the FrequencyPlanID as part of the TxRequest. Ping @rvolosatovs

@KrishnaIyer please see https://github.com/TheThingsNetwork/lorawan-stack/pull/1826/commits/77b81cd95468d16fa4d6c196a73ac4b4ccdc9763

KrishnaIyer:
> You cannot simply pass the first frequency plan to the scheduler, because some settings are taken from the frequency plan and not from the underlying band.

Yes of course not. That would be too easy. This is only to make the tests pass (There seems to be a flake in the IS).

This PR is not meant to have the changes for the scheduler. That's why [this issue](https://github.com/TheThingsNetwork/lorawan-stack/issues/1824) exists.

KrishnaIyer:
Opening an actual PR since #1824 is done

rvolosatovs:
> Opening an actual PR since #1824 is done

@KrishnaIyer you can just click "ready to review" button to convert a draft PR into an "actual PR"

This(#1826) is the "actual PR", since it contains all the discussion etc.
We're already losing a lot of context by relying heavily on GitHub and not `git` directly(e.g. long commit messages utilizing the body and not just the header) and this kind of stuff makes it even worse. Please consider closing #1864 and reopening this instead and rebasing.

KrishnaIyer:
That's a good point @rvolosatovs. I didn't know you could convert a draft PR to an actual one. In this case unfortunately, Github doesn't allow re-opening of Draft PRs. Even @htdvisser tried in admin mode.  So I've reopened #1864 instead. Sorry for the confusion.

rvolosatovs:
You cannot reopen this PR, because the commit hash does not match
You need to:
1. force-push `feature/761-multi-fplan-gtw` to 6833a20
2. reopen the PR
3. force-push `feature/761-multi-fplan-gtw` to correct commit (currently eaefe77 in #1864)

pkg/gatewayserver/io/io.go
```diff
@@ -265,7 +269,22 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
...
+		if err != nil {
+			return 0, err
+		}
+		if gtwFP.BandID == reqFP.BandID {
+			fp = gtwFP
+			break
+		}
+	}
+	if fp == nil {
+		return 0, errFrequencyPlanNotConfigured
```
johanstokking: Missing attribute

pkg/gatewayserver/io/io.go
```diff
@@ -265,7 +269,22 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
 	if err != nil {
 		return 0, err
 	}
-	phy, err := band.GetByID(c.fp.BandID)
+	var fp *frequencyplans.FrequencyPlan
+	for _, gtwFP := range c.fps {
+		reqFP, err := c.fpStore.GetByID(request.FrequencyPlanID)
```
johanstokking: We should not be getting the frequency plans from the store in the hot path; we have the frequency plans already loaded in the connection's gateway right?

pkg/gatewayserver/io/io.go
```diff
@@ -265,7 +269,22 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
...
+		if err != nil {
+			return 0, err
+		}
+		if gtwFP.BandID == reqFP.BandID {
+			fp = gtwFP
```
johanstokking: Also what does this mean; the frequency plan matches if the band matches?

pkg/gatewayserver/io/io.go
johanstokking: Also, we don't break API here; if there's no frequency plan in the `TxRequest`, we do what we did before

Please keep (some) test cases with the current behavior to test that

pkg/gatewayserver/scheduling/scheduler.go
```diff
@@ -65,9 +65,14 @@ type RTTs interface {
 	Stats() (min, max, median time.Duration, count int)
 }
 
+var (
+	errFrequencyPlansTimeOffAir     = errors.DefineInvalidArgument("frequency_plans_timeoffair", "frequency plans must have the same time off air value")
```
johanstokking: nit; `time_off_air`

pkg/pfconfig/basicstationlns/basicstationlns.go
johanstokking: This is fine

pkg/gatewayserver/io/io.go
KrishnaIyer: No indeed. The frequency plan ID should match. Fixed.

pkg/gatewayserver/io/io.go
KrishnaIyer: That's a good one. Updated.

pkg/gatewayserver/io/io.go
```diff
@@ -272,7 +274,22 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
...
+		fp = c.fps[fpID]
+		if fp == nil {
+			return 0, errFrequencyPlanNotConfigured.WithAttributes("id", request.FrequencyPlanID)
+		}
+	} else {
+		// Backwards compatibility. It's assumed in this case, that there's only one Frequency Plan configured.
+		for _, v := range c.fps {
+			fp = v
+			break
+		}
```
johanstokking: Why?

pkg/gatewayserver/scheduling/scheduler.go
```diff
@@ -83,41 +88,75 @@ func NewScheduler(ctx context.Context, fp *frequencyplans.FrequencyPlan, enforce
...
+						MinFrequency: subBand.MinFrequency,
+						MaxFrequency: subBand.MaxFrequency,
+						DutyCycle:    subBand.DutyCycle,
+					}
+					sb := NewSubBand(ctx, params, s.clock, nil)
+					var isIdentical bool
+					for _, subBand := range s.subBands {
+						if subBand.IsIdentical(sb) {
+							isIdentical = true
+							break
```
johanstokking: `continue` and skip check on L126

pkg/gatewayserver/scheduling/scheduler.go
```diff
@@ -83,41 +88,75 @@ func NewScheduler(ctx context.Context, fp *frequencyplans.FrequencyPlan, enforce
...
```
johanstokking: Idem

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -375,6 +376,12 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 					logger.WithError(err).Debug("Failed to unmarshal join-request message")
 					return nil
 				}
+				// According to spec, all frequency plans must be from the same band, so choosing the first one.
+				var fp *frequencyplans.FrequencyPlan
+				for _, v := range fps {
+					fp = v
+					break
+				}
```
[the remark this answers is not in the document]
KrishnaIyer: because `fps` is a `map[string]*frequencyplans.FrequencyPlan`

pkg/gatewayserver/scheduling/scheduler.go
KrishnaIyer: Hmm why? I also want to check if there's an overlap which is on `L126`

pkg/gatewayserver/io/io.go
KrishnaIyer: https://thethingsindustries.slack.com/archives/C06UQSMPB/p1579597038008000?thread_ts=1579594672.002700&cid=C06UQSMPB

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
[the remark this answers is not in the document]
johanstokking: 1. Use a slice
2. In this case, you don't care about the frequency plan, but about the band (L385). If the spec says all frequency plans are from the same band, why not introduce a field `BandID` and keep away from this loop?

KrishnaIyer:
Ok @johanstokking, I've explicitly used `BandID` where applicable.

PR #1834, opened 2020-01-09
Add configurable ScheduleAnytimeDelay per gateway

### LINKED ISSUE #1812: Configure ScheduleTimeLong per gateway
#### Summary

Configure `ScheduleTimeLong` per gateway

#### Why do we need this?

To be able to adjust the time that GS schedules class C messages in advance per gateway.

This is necessary as we should not be increasing `ScheduleTimeLong` to arbitrary high values (see i.e. https://github.com/TheThingsNetwork/lorawan-stack/pull/1811), because that adds delays to class C scheduling. Those delays are not desired as end devices should not be in class C for too long to save battery.

Fixed `ScheduleTimeLong` value for all gateways.

We already have some values per-gateway, like `schedule_downlink_late`, `enforce_duty_cycle` and `downlink_path_constraint`.

Have a configurable `schedule_anytime_delay` per gateway that can be configured. This is useful for gateways that have a known high latency backhaul, like 3G and satellite.

Introduce `schedule_anytime_delay` field to `Gateway` entity, store in ER (IS) and fetch in GS, like the aforementioned fields. When non-nil/zero, use value from `Gateway` entity, otherwise use `ScheduleTimeLong`.

Can review, can assist, let me know

neoaggelos:
#### Summary

Closes #1812

#### Changes

- [X] Add a new field `schedule_anytime_delay` in Gateway
- [X] Use that value instead of `ScheduleTimeLong` in the scheduler when non-nil
- [x] Fix existing tests
- [x] Add tests for the new functionality
- [ ] Configure `schedule_anytime_delay` via the Console

#### Notes for Reviewers

- Will require a db migration when updating from previous versions

#### Checklist


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Support for adjusting the time that the Gateway Server schedules class C messages in advance per gateway.
+  - This requires a database migration (`ttn-lw-stack is-db migrate`) because of the added columns.
```
rvolosatovs:
```suggestion
- This requires a database migration (`ttn-lw-stack is-db migrate`) because of the added columns.
```

pkg/gatewayserver/scheduling/scheduler.go
```diff
@@ -66,19 +67,32 @@ type RTTs interface {
...
+	logger := log.FromContext(ctx)
 	if timeSource == nil {
 		timeSource = SystemTimeSource
 	}
+
+	scheduleAnytimeDelay := time.Duration(scheduleAnytimeDelayMs) * time.Millisecond
+	if scheduleAnytimeDelay <= ScheduleTimeShort {
+		scheduleAnytimeDelay = ScheduleTimeLong
+	} else if scheduleAnytimeDelayMs > 0 {
+		logger.
```
rvolosatovs:
```suggestion
		logger.WithField(
			"minimum", ScheduleTimeShort,
		).Warnf("Failed to set scheduleAnytimeDelay to %vms", scheduleAnytimeDelayMs)
```

CHANGELOG.md
johanstokking: Probably just join the lines

api/gateway.proto
```diff
@@ -115,6 +115,8 @@ message Gateway {
   // duty cycle only in controlled research and development environments.
   bool enforce_duty_cycle = 17;
   DownlinkPathConstraint downlink_path_constraint = 18 [(validate.rules).enum.defined_only = true];
+  // Adjust the time that GS schedules class C messages in advance. This is useful for gateways that have a known high latency backhaul, like 3G and satellite.
+  uint32 schedule_anytime_delay = 19;
```
johanstokking: Make this a nice duration, see other occurrences

pkg/gatewayserver/gatewayserver.go
```diff
@@ -365,6 +365,7 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids
 				"frequency_plan_id",
 				"schedule_downlink_late",
 				"enforce_duty_cycle",
+				"schedule_anytime_delay",
```
johanstokking: Order this list

pkg/gatewayserver/scheduling/scheduler.go
```diff
@@ -66,19 +67,32 @@ type RTTs interface {
...
 // NewScheduler instantiates a new Scheduler for the given frequency plan.
 // If no time source is specified, the system time is used.
-func NewScheduler(ctx context.Context, fp *frequencyplans.FrequencyPlan, enforceDutyCycle bool, timeSource TimeSource) (*Scheduler, error) {
+func NewScheduler(ctx context.Context, fp *frequencyplans.FrequencyPlan, enforceDutyCycle bool, scheduleAnytimeDelayMs uint32, timeSource TimeSource) (*Scheduler, error) {
+	logger := log.FromContext(ctx)
 	if timeSource == nil {
 		timeSource = SystemTimeSource
 	}
+
+	scheduleAnytimeDelay := time.Duration(scheduleAnytimeDelayMs) * time.Millisecond
```
johanstokking: It should already be a duration

pkg/gatewayserver/scheduling/scheduler.go
```diff
@@ -66,19 +67,32 @@ type RTTs interface {
...
+	if scheduleAnytimeDelay <= ScheduleTimeShort {
+		scheduleAnytimeDelay = ScheduleTimeLong
+	} else if scheduleAnytimeDelayMs > 0 {
+		logger.
+			WithField("minimum", ScheduleTimeShort.String()).
+			Warnf("Failed to set scheduleAnytimeDelay to %vms", scheduleAnytimeDelayMs)
```
johanstokking: We don't use variables in log messages; we have fields for that

pkg/gatewayserver/scheduling/scheduler.go
```diff
@@ -66,19 +67,32 @@ type RTTs interface {
...
-func NewScheduler(ctx context.Context, fp *frequencyplans.FrequencyPlan, enforceDutyCycle bool, timeSource TimeSource) (*Scheduler, error) {
+func NewScheduler(ctx context.Context, fp *frequencyplans.FrequencyPlan, enforceDutyCycle bool, scheduleAnytimeDelayMs uint32, timeSource TimeSource) (*Scheduler, error) {
+	logger := log.FromContext(ctx)
 	if timeSource == nil {
 		timeSource = SystemTimeSource
 	}
+
+	scheduleAnytimeDelay := time.Duration(scheduleAnytimeDelayMs) * time.Millisecond
+	if scheduleAnytimeDelay <= ScheduleTimeShort {
+		scheduleAnytimeDelay = ScheduleTimeLong
```
johanstokking: - I think the sane minimum is `ScheduleTimeShort`
- If the value is zero/nil, default to `ScheduleTimeLong`

pkg/gatewayserver/scheduling/scheduler.go
```diff
@@ -66,19 +67,32 @@ type RTTs interface {
...
+	} else if scheduleAnytimeDelayMs > 0 {
+		logger.
+			WithField("minimum", ScheduleTimeShort.String()).
+			Warnf("Failed to set scheduleAnytimeDelay to %vms", scheduleAnytimeDelayMs)
+	}
+	logger.Debugf("Set schedule_anytime_delay to %v", scheduleAnytimeDelay.String())
```
johanstokking: Idem; and probably this log entry is not necessary

pkg/identityserver/store/gateway.go
```diff
@@ -52,8 +52,9 @@ type Gateway struct {
 	StatusPublic   bool `gorm:"not null"`
 	LocationPublic bool `gorm:"not null"`
 
-	ScheduleDownlinkLate   bool `gorm:"not null"`
-	EnforceDutyCycle       bool `gorm:"not null"`
+	ScheduleDownlinkLate   bool   `gorm:"not null"`
+	EnforceDutyCycle       bool   `gorm:"not null"`
+	ScheduleAnytimeDelay   uint32 `gorm:"default:0"`
```
johanstokking: @htdvisser what do we do with durations?

CHANGELOG.md
neoaggelos: There is an entry in the 3.3.0 version changelist with this format, which is why I used it. Can change if needed, but perhaps consistency would be better?

pkg/identityserver/store/gateway.go
neoaggelos: As a first solution I have added the duration as an `int64` on the database, doing conversions on load/save.

gorm does not support time.Duration

pkg/gatewayserver/scheduling/scheduler.go
neoaggelos: Should be OK now

CHANGELOG.md
johanstokking: OK fine with me too

pkg/gatewayserver/scheduling/scheduler.go
```diff
@@ -66,19 +67,32 @@ type RTTs interface {
...
 	if timeSource == nil {
 		timeSource = SystemTimeSource
 	}
+
+	if scheduleAnytimeDelay == nil || *scheduleAnytimeDelay == 0 {
+		scheduleAnytimeDelay = &ScheduleTimeLong
+	} else if *scheduleAnytimeDelay < ScheduleTimeShort {
+		logger.WithFields(log.Fields(
+			"schedule_anytime_delay_minimum", ScheduleTimeShort.String(),
+			"schedule_anytime_delay_requested", scheduleAnytimeDelay.String(),
```
johanstokking: - Context is clear; you can use `minimum` and `requested`
- Let the logger handle the stringing, but you can dereference (`*scheduleAnytimeDelay`)

pkg/gatewayserver/scheduling/scheduler.go
```diff
@@ -66,19 +67,32 @@ type RTTs interface {
...
+		)).Warn("schedule_anytime_delay too small, default to minimum value")
```
johanstokking: - Logs are not for developers; `Requested scheduling delay is too small ...`
- Warnings are reserved for infrastructural things that operators have to look at; you can use Debug or at most Info here

pkg/identityserver/store/gateway.go
```diff
@@ -88,7 +90,11 @@ var gatewayPBSetters = map[string]func(*ttnpb.Gateway, *Gateway){
 	statusPublicField:         func(pb *ttnpb.Gateway, gtw *Gateway) { pb.StatusPublic = gtw.StatusPublic },
 	locationPublicField:       func(pb *ttnpb.Gateway, gtw *Gateway) { pb.LocationPublic = gtw.LocationPublic },
 	scheduleDownlinkLateField: func(pb *ttnpb.Gateway, gtw *Gateway) { pb.ScheduleDownlinkLate = gtw.ScheduleDownlinkLate },
-	enforceDutyCycleField:     func(pb *ttnpb.Gateway, gtw *Gateway) { pb.EnforceDutyCycle = gtw.EnforceDutyCycle },
+	scheduleAnytimeDelayField: func(pb *ttnpb.Gateway, gtw *Gateway) {
+		d := time.Duration(gtw.ScheduleAnytimeDelay)
+		pb.ScheduleAnytimeDelay = &d
+	},
+	enforceDutyCycleField: func(pb *ttnpb.Gateway, gtw *Gateway) { pb.EnforceDutyCycle = gtw.EnforceDutyCycle },
```
johanstokking: @htdvisser can you quickly look at this?

pkg/identityserver/store/gateway.go
```diff
@@ -115,17 +121,24 @@ var gatewayModelSetters = map[string]func(*Gateway, *ttnpb.Gateway){
...
+	statusPublicField:         func(gtw *Gateway, pb *ttnpb.Gateway) { gtw.StatusPublic = pb.StatusPublic },
+	locationPublicField:       func(gtw *Gateway, pb *ttnpb.Gateway) { gtw.LocationPublic = pb.LocationPublic },
+	scheduleDownlinkLateField: func(gtw *Gateway, pb *ttnpb.Gateway) { gtw.ScheduleDownlinkLate = pb.ScheduleDownlinkLate },
+	scheduleAnytimeDelayField: func(gtw *Gateway, pb *ttnpb.Gateway) {
+		if pb.ScheduleAnytimeDelay == nil {
+			gtw.ScheduleAnytimeDelay = 0
+		} else {
+			gtw.ScheduleAnytimeDelay = int64(*pb.ScheduleAnytimeDelay)
+		}
+	},
```
johanstokking: @htdvisser idem

pkg/identityserver/store/gateway.go
```diff
@@ -52,8 +53,9 @@ type Gateway struct {
 	StatusPublic   bool `gorm:"not null"`
 	LocationPublic bool `gorm:"not null"`
 
-	ScheduleDownlinkLate   bool `gorm:"not null"`
-	EnforceDutyCycle       bool `gorm:"not null"`
+	ScheduleDownlinkLate   bool  `gorm:"not null"`
+	EnforceDutyCycle       bool  `gorm:"not null"`
+	ScheduleAnytimeDelay   int64 `gorm:"default:0"`
```
htdvisser: - This can be unsigned, it should never be lower than 0
- This can be `not null` as well, we don't need to differentiate betwen `0` and `nil`

pkg/identityserver/store/gateway_store_test.go
```diff
@@ -64,6 +66,7 @@ func TestGatewayStore(t *testing.T) {
 			}
 			a.So(created.CreatedAt, should.HappenAfter, time.Now().Add(-1*time.Hour))
 			a.So(created.UpdatedAt, should.HappenAfter, time.Now().Add(-1*time.Hour))
+			a.So(*created.ScheduleAnytimeDelay, should.Equal, time.Second)
```
htdvisser: Let's also test that updating it to `nil` works as expected.

johanstokking:
Please have CI passed for final review

pkg/gatewayserver/scheduling/scheduler.go
neoaggelos: Changed to Info. I think it's important to let the operator/user that their value is not being used.

pkg/identityserver/store/gateway.go
neoaggelos: intially added as `int64` because `time.Duration` is an int64. Can change

htdvisser: how about the `not null`?

neoaggelos:
> how about the `not null`?

@htdvisser I am unsure whether having `not null` would delete existing rows from the database during migrate. Perhaps this should not be a problem along with a default value.

pkg/identityserver/store/gateway.go
neoaggelos: I am unsure whether adding `not null` here would delete existing rows from the database during migrate. Perhaps this should not be a problem along with a default value.

htdvisser: If you put `not null;default0` a migration should create the column and set the value in all rows to 0. If you feel more comfortable leaving it null, that's fine with me too.

neoaggelos: added not null

Issue #1120, opened 2019-08-07
Reusing EUIs of Deleted Gateways


adamsondelacruz:
#### Summary
When you delete a gateway and then add a new one with the same EUI, if fails because it complains that the gateway already exists.

#### Steps to Reproduce
1.  Remove an existing gateway
2. Add it using the CLI or 
3. Add it using the API

#### What do you see now?

```
{
    "code": 6,
    "message": "error:pkg/identityserver/store:already_exists (entity already exists)",
    "details": [
        {
            "@type": "type.googleapis.com/ttn.lorawan.v3.ErrorDetails",
            "namespace": "pkg/identityserver/store",
            "name": "already_exists",
            "message_format": "entity already exists",
            "attributes": {
                "field": "gateway_eui",
                "value": "'3135313749005303'"
            },
            "correlation_id": "c0dc6bb73d714702bd8d0be57e83f369"
        }
    ]
}
```

#### What do you want to see instead?

Successfullly create the new gateway

#### Environment
amazon linux using TTN CLI and Postman using API Call

#### How do you propose to implement this?
When user try to insert a deleted gateway, update the deleted flag to false

#### Can you do this yourself and submit a Pull Request?

...

htdvisser:
Related to https://github.com/TheThingsNetwork/lorawan-stack/issues/1121

htdvisser:
I updated this issue a bit to have a slightly different scope than #1121, which we can make more about restoring deleted entities in general (instead of only applications).

In the current issue, let's focus more on adding new gateways with the same EUI as a deleted gateway.

htdvisser:
Related to https://github.com/TheThingsNetwork/lorawan-stack/issues/604 (releasing IDs/EUIs)

johanstokking:
@htdvisser even having https://github.com/TheThingsNetwork/lorawan-stack/issues/1703 (now in Next Up) is not entirely covering the common use case of being able to create a gateway with an EUI that has been used before. Looping in an administrator to purge an entity is not a nice user experience, plus purging an entity is more destructive than simply releasing the EUI.

Afaik we don't have security issues with releasing the EUI when deleting the gateway. It is optional already. It would, obviously, make recovery partial, i.e. the EUI is gone, but I think that is acceptable and also aligns with how device EUIs are released on deletion immediately.

Can we close this issue with simply releasing the EUI on gateway delete? If not, why not?

htdvisser:
[...]

PR #1843, opened 2020-01-13
Release EUIs of deleted gateways

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1120

#### Changes

- Add `afterDelete` hook for gateways to reset their eui.
- Update gateway deletion modal text.

#### Checklist

### DIFF
--- a/pkg/webui/console/views/gateway-general-settings/index.js
+++ b/pkg/webui/console/views/gateway-general-settings/index.js
@@ -43,7 +43,7 @@ const m = defineMessages({
   updateSuccess: 'Successfully updated gateway',
   deleteGateway: 'Delete Gateway',
   modalWarning:
-    'Are you sure you want to delete "{gtwName}"? This action cannot be undone and it will not be possible to reuse the gateway ID and EUI!',
+    'Are you sure you want to delete "{gtwName}"? This action cannot be undone and it will not be possible to reuse the gateway ID!',
 })
 
 @connect(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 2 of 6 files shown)


pkg/webui/console/views/gateway-general-settings/index.js
```diff
@@ -42,8 +42,10 @@ import { mayEditBasicGatewayInformation, mayDeleteGateway } from '../../lib/feat
 const m = defineMessages({
   updateSuccess: 'Successfully updated gateway',
   deleteGateway: 'Delete Gateway',
-  modalWarning:
-    'Are you sure you want to delete "{gtwName}"? This action cannot be undone and it will not be possible to reuse the gateway ID and EUI!',
+  modalWarning: `
+Are you sure you want to delete "{gtwName}"? This action cannot be undone and it will not be possible to reuse the gateway ID!
+Note, deleting this gateway will release its EUI and will make it available for registering a new one with the same gateway EUI.
```
kschiffer: Do we need to be this explicit?

htdvisser: I'd say no. Let's make it consistent with the message for other entities.

t0mZ4:
Hi

I still see an issue when I use the CLI to delete a gateway, and then I cannot re-add the gateway with the same id.

```

error:pkg/identityserver/store:id_taken (ID already taken)
    correlation_id=fc206b725e0543d691dafa5a4d8d2674

```

any way to workaround this!?

htdvisser:
The EUI was released when you deleted the gateway, but the ID is still reserved.

This is for security reasons, as historical data (in external integrations) is linked to the ID, and if someone else would register a gateway with the same ID after you delete it, they would have access to your historical data, which probably isn't what you want.

With #3467 it's possible for admin users to release the IDs, but of course that could cause some security issues like the one I described above.

Issue #1815, opened 2020-01-07
Update gateway location observed by Gateway Server


ecities:
#### Summary

Hi. I've got questions with comes up during development.

Kerlink Wirenet +with GPS and GSM sends its position (logs in spf) however it is not automatically updated in the console (map or position) v3.3.2. 
Can it be automatically updated?
Can GPS position be added to the JSON status of the device (for integration)?

#### Why do we need this?

This is good way to test range the whole system (many "static" LoRaWAN controllers and single "mobile" LoRaWAN gateway which send GPS position).
...

#### What is already there? What do you see now?

...

#### What is missing? What do you want to see?

...

#### Environment

...

#### How do you propose to implement this?

...

#### Can you do this yourself and submit a Pull Request?

...

johanstokking:
Thanks, good suggestion.

The gateway status is observed by the Gateway Server, while the location is stored in Identity Server. To keep things simple, we might want to allow the Gateway Server to update the gateway's `location` with location source set to `SOURCE_GPS`. We would need to figure out what rights GS needs to do that. There's already an API key but intended only for linking the gateway.

@htdvisser thoughts?

htdvisser:
This issue has two components:

1. The GS automatically updating a gateway (antenna) location in the IS:
  - I think we should only do this for mobile gateways (we don't want to update the locations of thousands of fixed gateways with an inaccurate GPS location every 30 seconds). We also have another issue (#545) that might benefit from a "mobile_gateway" flag in the registration.
  - I think it's not too crazy to allow the LINK to update the gateway status and location.
2. Propagating the gateway (antenna) location to devices.
  - It seems the GS doesn't actually inject the `location` of the `RxMetadata` (which the field comment says it should), let's start with that.

neoaggelos:
For 1:

- We should add a new gateway field called `auto_update_location` which tells the stack that this gateway reports its location and we want to use that.
- We would need one more upstream handler (I believe using the existing `ns` one is not correct, since the NS is not relevant here) with a nil `HandleUplink` and a `HandleStatus` that listens for gateway status messages and updates the gateway antenna location accordingly (using SOURCE_GPS as location source).
- In order to scale better, this handler should only be installed for gateways with `auto_update_location` on. We could also add a second `update_location_interval` option (with type duration), and debounce location updates in a separate goroutine.
[...]

PR #1849, opened 2020-01-13
gs: Include antenna location in forwarded message RxMetadata

neoaggelos:
#### Summary

Required for #1815 

#### Changes

- Gateway Server properly adds antenna location on the `RxMetadata` field of uplink messages forwarded to the Network Server

#### Notes for Reviewers

- Implementing this as a separate small fix.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -362,11 +362,11 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids 
 		GatewayIdentifiers: ids,
 		FieldMask: pbtypes.FieldMask{
 			Paths: []string{
-				"location_public",
+				"antennas",
 				"downlink_path_constraint",
 				"enforce_duty_cycle",
-				"antennas",
 				"frequency_plan_id",
+				"location_public",
 				"schedule_anytime_delay",
 				"schedule_downlink_late",
 			},
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -362,6 +362,7 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids 
 		GatewayIdentifiers: ids,
 		FieldMask: pbtypes.FieldMask{
 			Paths: []string{
+				"location_public",
 				"downlink_path_constraint",
 				"enforce_duty_cycle",
 				"antennas",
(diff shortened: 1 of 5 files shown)


pkg/gatewayserver/io/io.go
```diff
@@ -159,6 +159,11 @@ func (c *Connection) HandleUp(up *ttnpb.UplinkMessage) error {
 		}
 		md.UplinkToken = buf
 		md.DownlinkPathConstraint = c.gateway.DownlinkPathConstraint
+
+		if len(c.gateway.Antennas) > int(md.AntennaIndex) {
+			location := c.gateway.Antennas[md.AntennaIndex].Location
+			md.Location = &location
```
johanstokking: Ineffective variable

pkg/gatewayserver/io/io.go
```diff
@@ -159,6 +159,11 @@ func (c *Connection) HandleUp(up *ttnpb.UplinkMessage) error {
 		}
 		md.UplinkToken = buf
 		md.DownlinkPathConstraint = c.gateway.DownlinkPathConstraint
+
+		if len(c.gateway.Antennas) > int(md.AntennaIndex) {
```
johanstokking: Where does this come from? 

Should we be getting `antennnas` here?

https://github.com/TheThingsNetwork/lorawan-stack/blob/96ebc355c10a456224f1957bf494c10f1fef38f3/pkg/gatewayserver/gatewayserver.go#L361-L371

Also, make sure that `antennas` is returned when GS uses cluster authentication (i.e. in an unauthenticated gateway context, when the gateway uses UDP for example).

pkg/gatewayserver/io/io.go
neoaggelos: Purpose is to pass a copy of the antenna location value. Is it safe to just pass a reference?

pkg/gatewayserver/io/io.go
neoaggelos: Should be OK now

johanstokking:
The open issue is that when gateways link with an API key that has `INFO` rights, GS gets the location even when `location_public` is false, and GS will make it part of the `RxMetadata`. This is an edge case; gateways should not be linking with rights other than `LINK`, but we should account for this. Because users.

So GS also has to get the `location_public` setting, and if that is false, strip the location.

neoaggelos:
GS now honors the `location_public` flag. We still need to add a few tests for this before it can be merged.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -362,8 +362,10 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids
 		GatewayIdentifiers: ids,
 		FieldMask: pbtypes.FieldMask{
 			Paths: []string{
+				"location_public",
 				"downlink_path_constraint",
 				"enforce_duty_cycle",
+				"antennas",
```
johanstokking: Keep this list sorted ascending

PR #1853, opened 2020-01-13
Support switching to external JS from CLI

rvolosatovs:
#### Summary

Support switching to external JS from CLI. Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1854

#### Changes

- Add --external-join-server to device update

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Support for releasing gateway EUI after deletion.
-- Added support in the Application Server for the `X-Downlink-Apikey`, `X-Downlink-Push` and `X-Downlink-Replace` webhook headers. They allow webhook integrations to determine which endpoints to use for downlink queue operations.
-- Added the `as.webhooks.downlinks.public-address` and `as.webhooks.downlinks.public-tls-address` configuration options to the Application Server.
+- Support in the Application Server for the `X-Downlink-Apikey`, `X-Downlink-Push` and `X-Downlink-Replace` webhook headers. They allow webhook integrations to determine which endpoints to use for downlink queue operations.
+- `as.webhooks.downlinks.public-address` and `as.webhooks.downlinks.public-tls-address` configuration options to the Application Server.
 - Support for adjusting the time that the Gateway Server schedules class C messages in advance per gateway.
   - This requires a database migration (`ttn-lw-stack is-db migrate`) because of the added columns.
 - `end-devices use-external-join-server` CLI subcommand, which disassociates and deletes the device from Join Server.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added the `as.webhooks.downlinks.public-address` and `as.webhooks.downlinks.public-tls-address` configuration options to the Application Server.
 - Support for adjusting the time that the Gateway Server schedules class C messages in advance per gateway.
   - This requires a database migration (`ttn-lw-stack is-db migrate`) because of the added columns.
+- `end-devices use-external-join-server` CLI subcommand, which disassociates and deletes the device from Join Server.
 
 ### Changed
 
--- a/cmd/ttn-lw-cli/commands/end_devices.go
+++ b/cmd/ttn-lw-cli/commands/end_devices.go
@@ -160,7 +160,10 @@ func generateDevAddr(netID types.NetID) (types.DevAddr, error) {
 	return devAddr, nil
 }
 
-var errNetworkServerDisabled = errors.DefineFailedPrecondition("network_server_disabled", "Network Server is disabled")
+var (
+	errJoinServerDisabled    = errors.DefineFailedPrecondition("join_server_disabled", "Join Server is disabled")
+	errNetworkServerDisabled = errors.DefineFailedPrecondition("network_server_disabled", "Network Server is disabled")
+)
 
 var (
 	endDevicesCommand = &cobra.Command{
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for releasing gateway EUI after deletion.
 - Added support in the Application Server for the `X-Downlink-Apikey`, `X-Downlink-Push` and `X-Downlink-Replace` webhook headers. They allow webhook integrations to determine which endpoints to use for downlink queue operations.
 - Added the `as.webhooks.downlinks.public-address` and `as.webhooks.downlinks.public-tls-address` configuration options to the Application Server. 
+- `--external-join-server` flag to `end-device update` CLI subcommand, which disassociates and deletes the device from cluster-local Join Server.
```
johanstokking: Added ...

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -536,8 +537,12 @@ var (
 			if err != nil {
 				return err
 			}
+			extJS, err := cmd.Flags().GetBool("external-join-server")
+			if err != nil {
```
johanstokking: Afaik we don't need errors check here, just `extJS, _ := ...`

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -618,6 +623,25 @@ var (
...
+				if err != nil {
+					return err
+				}
+				_, err = ttnpb.NewJsEndDeviceRegistryClient(js).Delete(ctx, &device.EndDeviceIdentifiers)
+				if errors.IsNotFound(err) {
+					logger.WithError(err).Error("Could not delete end device from Join Server")
+				} else if err != nil {
+					return err
+				}
+				config.JoinServerEnabled = false
```
johanstokking: What is this for?

CHANGELOG.md
htdvisser: And use the "official" form of the command instead of the alias (`end-devices update`; see `--help`)

cmd/ttn-lw-cli/commands/end_devices.go
htdvisser: We're not always consistent with this. AFAIK the only way this can error is if the flag isn't defined, or if it's defined as a string (or any other type) and try to read it as a bool

cmd/ttn-lw-cli/commands/end_devices.go
htdvisser: I think this entire block could use some comments to tell other developers what's going on

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -1083,6 +1107,7 @@ func init() {
 	endDevicesUpdateCommand.Flags().AddFlagSet(setEndDeviceFlags)
 	endDevicesUpdateCommand.Flags().AddFlagSet(attributesFlags())
 	endDevicesUpdateCommand.Flags().Bool("touch", false, "set in all registries even if no fields are specified")
+	endDevicesUpdateCommand.Flags().Bool("external-join-server", false, "disassociate and delete the device from cluster-local Join Server")
```
htdvisser: The concept "cluster-local" does not exist in the CLI. If the CLI is configured to point to TTI's "global" join server, this flag would delete it from there.

cmd/ttn-lw-cli/commands/end_devices.go
johanstokking: In either case, we don't have to check errors here. If people pass a `--external-join-server="the world is flat"`, they won't enable the external JS.

CHANGELOG.md
rvolosatovs: At https://keepachangelog.com/en/1.0.0/ these are not preceeded by `Added` and that was also the format we used so far

johanstokking:
Okay so here's the difference between update and delete from a user's perspective:

1. I want to `change` my device to use an external Join Server.
2. I want to `delete` the cluster Join Server registration of my device so that the Network Server doesn't try the cluster Join Server and attempts to go to the external Join Server instead.

I very much prefer option 1 UX wise. This also aligns with the Console. I agree that CLI has no concept of "cluster JS" other than what's defined in the configuration (which could be the external JS), so we can do better on explaining what's happening.

cmd/ttn-lw-cli/commands/end_devices.go
rvolosatovs: IMO the user should know that the flag value is dropped instead of silently ignored. E.g. if user passes `--external-join-server=tru` (typo) and that is silently ignored, how does user know that his command actually failed?

cmd/ttn-lw-cli/commands/end_devices.go
rvolosatovs: That's to avoid setting fields on JS

htdvisser:
How about adding a `move` command to the CLI then?

cmd/ttn-lw-cli/commands/end_devices.go
johanstokking: OK fine

johanstokking:
I'm also fine with that!

rvolosatovs:
`move` implies having a source and destination.
In this case this just disassociates device from the cluster JS, doesn't actually move data anywhere.

johanstokking:
That's also true; there's no action being done on the external JS.

Maybe;

```bash
$ ttn-lw-cli end-devices use-external-js <app-id> <device-id>
```

Aliased `use-external-js` and `extjs`?

CHANGELOG.md
```diff
@@ -10,10 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Support for releasing gateway EUI after deletion.
-- Added support in the Application Server for the `X-Downlink-Apikey`, `X-Downlink-Push` and `X-Downlink-Replace` webhook headers. They allow webhook integrations to determine which endpoints to use for downlink queue operations.
-- Added the `as.webhooks.downlinks.public-address` and `as.webhooks.downlinks.public-tls-address` configuration options to the Application Server.
+- Support in the Application Server for the `X-Downlink-Apikey`, `X-Downlink-Push` and `X-Downlink-Replace` webhook headers. They allow webhook integrations to determine which endpoints to use for downlink queue operations.
```
rvolosatovs: Fixed these to avoid confusion

johanstokking:
References https://github.com/TheThingsNetwork/lorawan-stack/issues/1857

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -1025,6 +1028,43 @@ To generate a QR code for multiple end devices:
...
+				return err
+			}
+			if !config.JoinServerEnabled {
+				return errJoinServerDisabled
+			}
+			js, err := api.Dial(ctx, config.JoinServerGRPCAddress)
+			if err != nil {
+				return err
+			}
+			_, err = ttnpb.NewJsEndDeviceRegistryClient(js).Delete(ctx, devID)
```
johanstokking: @htdvisser do we want to do this when the JS address matches what is currently in IS, or do we not care about this?

htdvisser: We also check the registered addresses against the configured addresses in other commands, so I think we should also do that here.

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -1110,6 +1150,7 @@ func init() {
 	endDevicesGenerateQRCommand.Flags().Uint32("size", 300, "size of the image in pixels")
 	endDevicesGenerateQRCommand.Flags().String("folder", "", "folder to write the QR code image to")
 	endDevicesCommand.AddCommand(endDevicesGenerateQRCommand)
+	endDevicesCommand.AddCommand(endDevicesExternalJSCommand)
```
htdvisser: Don't forget `endDevicesCommand.AddFlagSet(endDeviceIDFlags())`

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -1025,6 +1028,43 @@ To generate a QR code for multiple end devices:
 			return nil
 		}),
 	}
+	endDevicesExternalJSCommand = &cobra.Command{
+		Use:     "external-join-server [application-id] [device-id]",
```
htdvisser: We want our CLI commands to be actions, and `external-join-server` is not an action. Maybe `use-external-join-server`? Aliases can be `use-external-js`, `use-ext-js`.

PR #1873, opened 2020-01-17
Webhook templates documentation

### LINKED ISSUE #1782: Document webhook templates formatting and usage
#### Summary

We need to document the webhook templates format and the process of creating such templates.

#### Why do we need this?

Such that integrators can create their own templates and contribute them to the Webhook Templates repository (or use them on their own).

[This](https://github.com/TheThingsNetwork/lorawan-stack/issues/945#issuecomment-520896156) comment which touches the surface of the template format.

A reference section about the templates themselves and how to build one.

Create the required section in `doc`.

Yes. cc @johanstokking

adriansmares:
#### Summary

Closes #1782 

#### Changes

- Add documentation relating to the webhook templates and how the values are replaced inside the templates in order to obtain the final webhook integration.

#### Notes for Reviewers

The repository itself should be mentioned once we have set it up. I couldn't find a good location for a full template example, but I did provide an explanation for all of the fields. Any thoughts on this ? cc @johanstokking 

#### Checklist

### DIFF
--- a/doc/content/reference/webhook-templates/format.md
+++ b/doc/content/reference/webhook-templates/format.md
@@ -6,18 +6,20 @@ weight: 2
 
 Webhook templates are described using the [YAML](https://yaml.org/) language. Their format is very closely related to that of a normal webhook integration, but with additional fields added.
 
+<!--more-->
+
 ## Service Description
 
 All of the webhook templates must contain the following fields which describe the service provided by the template to the user.
 
-- `applicationwebhooktemplateidentifiers.templateid`: The unique identifier of the template.
+- `template-id`: The unique identifier of the template.
 - `name`: The (human readable) name of the service.
 - `description`: The description of the service.
-- `logourl`: The URL of the logo of the service.
-- `infourl`: The URL of the main page of the service.
-- `documentationurl`: The URL of the documentation of the service. 
+- `logo-url`: The URL of the logo of the service.
+- `info-url`: The URL of the main page of the service.
+- `documentation-url`: The URL of the documentation of the service. 
 
-> Note: The difference between `documentationurl` and `infourl` is that `infourl` should lead to the home page of the service (i.e. `https://www.thethingsnetwork.org/`), while `documentationurl` should lead directly to the location of the documentation (i.e. `https://www.thethingsnetwork.org/docs/applications/example/`).
+> Note: The difference between `documentation-url` and `info-url` is that `info-url` should lead to the home page of the service (i.e. `https://www.thethingsnetwork.org/`), while `documentation-url` should lead directly to the location of the documentation (i.e. `https://www.thethingsnetwork.org/docs/applications/example/`).
 
 ## Template Fields
 
@@ -27,7 +29,7 @@ Templates can contain fields which will be filled by the user on instantiation. 
 - `name`: The (human readable) name of the field.
 - `description`: The description of the field.
 - `secret`: Controls if the contents of the field should be hidden. To be used in the case of secrets such as passwords, tokens or API keys.
-- `defaultvalue`: The value which should be pre filled for the user initially.
+- `default-value`: The value which should be pre filled for the user initially.
 
 For more information on the instantiation process, see [Instantiation]({{< ref "/reference/webhook-templates/instantiation.md" >}}).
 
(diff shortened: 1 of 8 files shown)


doc/content/reference/webhook-templates/format.md
```diff
@@ -0,0 +1,54 @@
...
+weight: 2
+---
+
+Webhook templates are described using the [YAML](https://yaml.org/) language. Their format is very closely related to that of a normal webhook integration, but with additional fields added.
+
+## Service Description
+
+All of the webhook templates must contain the following fields which describe the service provided by the template to the user.
+
+- `applicationwebhooktemplateidentifiers.templateid`: The unique identifier of the template.
```
johanstokking: I think we have to do two things;
- Get rid of this long thing
- Introduce dashes in keys

Would this break configuration? Did we release this stuff at all?

adriansmares: I think that the long part (`applicationwebhooktemplateidentifiers`) is there for the RPC validation.
Regarding the dashes - maybe it's an option in the YAML de/serializer ? I'll check.

johanstokking: How so RPC validation?

adriansmares: Nevermind - I thought we have (only) the identifiers as part of the `GET` request. We can simply use the validated string instead. Will add.

doc/content/reference/webhook-templates/format.md
```diff
@@ -0,0 +1,51 @@
...
+- `secret`: Controls if the contents of the field should be hidden. To be used in the case of secrets such as passwords, tokens or API keys.
+- `default-value`: The value which should be pre filled for the user initially.
+
+For more information on the instantiation process, see [Instantiation]({{< ref "/reference/webhook-templates/instantiation.md" >}}).
+
+## Endpoint
+
+The endpoint of the webhook can be configured using the following fields:
+
+- `format`: The format which the endpoint expects. Currently `json` and `grpc` are supported.
```
johanstokking: I think `pb` or something like that

doc/content/reference/webhook-templates/format.md
```diff
@@ -0,0 +1,51 @@
...
+- `uplink-message-path`: The path to which uplink messages will be sent. Can contain template fields.
+- `join-accept-path`: The path to which join accept messages will be sent. Can contain template fields.
+- `downlink-ack-path`: The path to which downlink acknowledgements will be sent. Can contain template fields.
+- `downlink-nack-path`: The path to which downlink not-acknowledged messages will be sent. Can contain template fields.
+- `downlink-sent-path`: The path to which downlink sent will be sent. Can contain template fields.
+- `downlink-failed-path`: The path to which downlink failures will be sent. Can contain template fields.
+- `downlink-queued-path`: The path to which downlink queued status will be sent. Can contain template fields.
+- `location-solved-path`: The path to which the location of the device will be sent when resolved. Can contain template fields.
+
+> Note: Not all of the messages types must be handled by the service. By omitting the field the message type will be disabled and the related messages will not be passed to the endpoint.
```
johanstokking: It would be very helpful to have a full example here

doc/content/reference/webhook-templates/format.md
```diff
@@ -0,0 +1,51 @@
+---
+title: "Format"
+description: ""
+weight: 2
+---
+
+Webhook templates are described using the [YAML](https://yaml.org/) language. Their format is very closely related to that of a normal webhook integration, but with additional fields added.
```
johanstokking: Probably add the more-tag here, also in other places

doc/content/reference/webhook-templates/instantiation.md
```diff
@@ -0,0 +1,46 @@
...
+- id: token
+  name: Authentication token
+  description: The token used for authentication
+  secret: true
+  default-value:
+headers:
+- Authorization: Bearer {token}
+```
+
+The user has filled in the value of `token` with `Zpdc7jWMvYzVTeNQ`. The resulting webhook will contain a header named `Authorization` with the value `Bearer Zpdc7jWMvYzVTeNQ`.
```
johanstokking: If, the user, then ...

doc/content/reference/webhook-templates/format.md
johanstokking: OK, so that comes in `instantiation`. Can we somehow combine that?

pkg/applicationserver/io/web/templates.go
```diff
@@ -208,3 +223,86 @@ func applyWebhookTemplateFieldMask(dst, src *ttnpb.ApplicationWebhookTemplate, p
...
+	Description          string                 `yaml:"description"`
+	LogoURL              string                 `yaml:"logo-url"`
+	InfoURL              string                 `yaml:"info-url"`
+	DocumentationURL     string                 `yaml:"documentation-url"`
+	BaseURL              string                 `yaml:"base-url"`
+	Headers              map[string]string      `yaml:"headers,omitempty"`
+	Format               string                 `yaml:"format"`
+	Fields               []webhookTemplateField `yaml:"fields,omitempty"`
+	CreateDownlinkAPIKey bool                   `yaml:"create-downlink-api-key"`
+	UplinkMessagePath    *string                `yaml:"uplink-message-path,omitempty"`
```
johanstokking: We might want to group these under a `struct {...} ``yaml:"paths"`` ` for structure

pkg/applicationserver/io/web/templates.go
```diff
@@ -273,6 +277,15 @@ func (webhookTemplate) pathToMessage(s *string) *ttnpb.ApplicationWebhookTemplat
 	}
 }
 
+func (webhookTemplate) realFormat(s string) string {
+	switch s {
+	case "pb":
+		return "grpc"
```
johanstokking: Wait, what is happening here?

adriansmares: Well the naming of the format in the AS frontends it's always `pb` for the binary protobuffs one. If you want it, for some reason, to be exposed in the templates as `pb`, we can rename it while loading the webhook.

johanstokking: I don't understand. Where is that exactly?

I see [this](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/applicationserver/io/web/format_protobuf.go#L20) and this is what I meant

adriansmares: I mixed the frontend name name with the format. Fixed.

PR #1875, opened 2020-01-17
Automatically reconnect frontends on link failure

### LINKED ISSUE #1831: Messages stop being sent to connected mqtt clients when re-linking application
#### Summary
When a existing mqtt or grpc client is connected and the application link is updated (to the same value) the existing client stops receiving messages for all topics. The client does continue to stay connected and gets keep alive messages (pingresp). On restarting the mqtt client the traffic returns to normal

#### Steps to Reproduce
1. Create an application as per the getting started guide
2. Create api key and use to connect via mqtt
3. Validate that uplink messages arrive as expected
4. Re link the application using the same api key as when it was created
5. Observe that uplinks arrive on the web console but not to the mqtt client

The same behaviour happens when using 
ttn-lw-cli applications subscribe

#### What do you see now?
Uplink traffic stops when the application link is re set

Traffic to momentarily stop while the application link is 'unlinked' and the restore for all connected and subscribed clients when the link is restored

#### Environment
Stack and cli version: 3.4.2

adriansmares:
#### Summary

Closes #1831 

#### Changes

- Added a new `io.Server` that attempts to reconnect to the upstream server (Application Server) if the subscription fails.
- Subscriptions now automatically disconnect on link failure (before this PR they would just get stuck).

#### Notes for Reviewers

None.

#### Checklist

### DIFF
--- a/pkg/applicationserver/io/middleware.go
+++ b/pkg/applicationserver/io/middleware.go
@@ -1,4 +1,4 @@
-// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


pkg/applicationserver/io/middleware.go
```diff
@@ -0,0 +1,154 @@
...
+	if err != nil {
+		return nil, err
+	}
+	go func() {
+		logger := log.FromContext(ctx)
+	nextUp:
+		for {
+			select {
+			case up := <-upstreamSub.Up():
+				{
```
johanstokking: nit; no braces here

pkg/applicationserver/io/middleware.go
```diff
@@ -0,0 +1,154 @@
...
+					err := downstreamSub.SendUp(up.Context, up.ApplicationUp)
+					if err != nil {
+						logger.WithError(err).Error("Failed to send the uplink downstream")
```
johanstokking: Probably Warn; Error is reserved for when shit hits the fan infrastructurally

pkg/applicationserver/io/middleware.go
```diff
@@ -0,0 +1,154 @@
...
+			case <-upstreamSub.Context().Done():
+				{
+					err := upstreamSub.Context().Err()
+					if errors.IsCanceled(err) {
+						for _, backoff := range rs.backoff {
+							upstreamSub, err = rs.upstream.Subscribe(ctx, protocol, ids)
+							if err == nil {
+								continue nextUp
+							}
+							time.Sleep(random.Jitter(backoff, rs.jitter))
```
johanstokking: Use `time.After` and check context cancelation in parallel

pkg/applicationserver/io/middleware.go
```diff
@@ -0,0 +1,154 @@
...
+				{
+					err := downstreamSub.SendUp(up.Context, up.ApplicationUp)
+					if err != nil {
+						logger.WithError(err).Error("Failed to send the uplink downstream")
+					}
+				}
+			case <-upstreamSub.Context().Done():
+				{
+					err := upstreamSub.Context().Err()
+					if errors.IsCanceled(err) {
```
johanstokking: How do we detect a graceful cancel?

pkg/applicationserver/io/middleware_test.go
```diff
@@ -0,0 +1,124 @@
+// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
```
neoaggelos: New file, this could be 2020

pkg/applicationserver/io/middleware.go
```diff
@@ -0,0 +1,177 @@
...
+	"time"
+
+	"go.thethings.network/lorawan-stack/pkg/config"
+	"go.thethings.network/lorawan-stack/pkg/errors"
+	"go.thethings.network/lorawan-stack/pkg/log"
+	"go.thethings.network/lorawan-stack/pkg/random"
+	"go.thethings.network/lorawan-stack/pkg/ttnpb"
+)
+
+var defaultBackoff = []time.Duration{100 * time.Millisecond, 1 * time.Second, 10 * time.Second}
```
neoaggelos: Unless it is considered too obvious, perhaps having the ability to configure the `backoff` times or adding a comment here would help.

pkg/applicationserver/io/middleware.go
johanstokking: Not addressed

pkg/applicationserver/io/middleware.go
```diff
@@ -0,0 +1,177 @@
...
+			case <-ctx.Done():
+				{
+					err := ctx.Err()
+					downstreamSub.Disconnect(err)
+					upstreamSub.Disconnect(err)
+					logger.WithError(err).Debug("Parent context canceled")
+					return
+				}
+			case <-upstreamSub.Context().Done():
+				{
```
johanstokking: Idem

pkg/applicationserver/io/middleware.go
```diff
@@ -0,0 +1,177 @@
...
+							}
+							logger.WithError(err).WithField("delay", delay).Debug("Resubscription failed")
+						}
+					}
+					downstreamSub.Disconnect(err)
+					logger.WithError(err).Debug("Upstream resubscription attempts failed. Downstream subscription canceled")
+					return
+				}
+			case <-downstreamSub.Context().Done():
+				{
```
johanstokking: Idem

pkg/applicationserver/linking.go
```diff
@@ -265,7 +269,12 @@ func (as *ApplicationServer) getLink(ctx context.Context, ids ttnpb.ApplicationI
 	val, ok := as.links.Load(uid)
 	if !ok {
 		if val, ok := as.linkErrors.Load(uid); ok {
-			if err := val.(error); !errors.IsCanceled(err) {
+			err := val.(error)
+			switch err.Error() {
+			case io.ErrLinkDeleted.Error():
+			case io.ErrLinkReset.Error():
```
johanstokking: Use error codes checking here

It's also not safe to use our errors for comparison; in fact it's better to not expose them at all

PR #1882, opened 2020-01-20
Mage git:diff and hooks

### LINKED ISSUE #1866: Use mage instead of make
#### Summary

Do not depend on `make` for setting up a development environment. @rvolosatovs Did some initial work with #1861, this is a follow-up.

#### What do you see now?

Mage targets instead, e.g.

```bash
$ ./mage dev:dbStart
```

and similar targets

#### Environment

all

Yes

neoaggelos:
#### Summary

Closes #1866

#### Changes

- Add `./mage git:diff` to replace `make git.diff`
- Update git hooks to call mage directly (instead of calling make which then calls mage)

#### Notes for Reviewers

- Requires uninstalling and then installing git hooks again `./mage git:uninstallhooks git:installhooks`
- Verify that all CI pipelines have been updated properly.
- With this, make is only used to build `./mage` itself.

#### Checklist


.mage/git.go
```diff
@@ -42,7 +43,11 @@ func (Git) installHook(name string) (err error) {
 			err = closeErr
 		}
 	}()
-	if _, err = fmt.Fprintf(f, "ARGS=\"$@\" make git.%s\n", name); err != nil {
+	mage, mageErr := filepath.Abs(os.Args[0])
+	if mageErr != nil {
+		mage = "./mage"
+	}
+	if _, err = fmt.Fprintf(f, "HOOK=\"%s\" ARGS=\"$@\" \"%s\" git:runHook\n", name, mage); err != nil {
```
rvolosatovs: I'd assume that `./mage` should already be there, however maybe a better solution is to use `go run github.com/magefile/mage` instead of `./mage` or anything else here?

.mage/git.go
```diff
@@ -219,3 +224,20 @@ func (g Git) RunHook() error {
 		return fmt.Errorf("Unknown hook %s", hook)
 	}
 }
+
+// Diff returns error if `git diff` is not empty
+func (g Git) Diff() error {
+	if mg.Verbose() {
+		fmt.Println("Checking git diff")
+	}
+	fmt.Println("Test")
```
rvolosatovs: Is this on purpose? :thinking:

.mage/git.go
```diff
@@ -219,3 +224,20 @@ func (g Git) RunHook() error {
...
+	output, err := sh.Output("git", "diff")
+	if err != nil {
+		return err
+	}
+
```
rvolosatovs: Remove newline

.mage/git.go
neoaggelos: Going with `go run github.com/magefile/mage`, as it seems like a cleaner solution

.mage/git.go
neoaggelos: definitely

.mage/git.go
```diff
@@ -42,7 +43,7 @@ func (Git) installHook(name string) (err error) {
 			err = closeErr
 		}
 	}()
-	if _, err = fmt.Fprintf(f, "ARGS=\"$@\" make git.%s\n", name); err != nil {
+	if _, err = fmt.Fprintf(f, "HOOK=\"%s\" ARGS=\"$@\" \"%s\" run github.com/magefile/mage git:runHook\n", name, mg.GoCmd()); err != nil {
```
htdvisser: I think we should just keep doing this with Make.
If you really think this is better, then at least call `./mage`. I don't want to `go run` (compile) Mage on every git commit.

.mage/mage.make
```diff
@@ -23,16 +23,4 @@ init: $(MAGE)
 	@$(MAGE) init
 	@echo "Run \"./mage -l\" for a list of build targets"
 
-.PHONY: git.pre-commit
-git.pre-commit: $(MAGE) # NOTE: DO NOT CHANGE - will break previously installed git hooks.
```
htdvisser: Undo. We can just keep doing this with Make.
Also, as the notes say, this will break existing development environments.

PR #1889, opened 2020-01-22
Various GS-related timing fixes

rvolosatovs:
#### Summary

Refs #1611 

#### Changes

- Support nanosecond precision in `gpstime` (a.k.a use `time.Duration`)
- Actually set `Settings.Time` from gateway RX packets.
- Use milisecond precision in `tmms` as specified in https://github.com/Lora-net/packet_forwarder/blob/d0226eae6e7b6bbaec6117d0d2372bf17819c438/PROTOCOL.TXT#L134
- Only set `Timestamp`(`tmst`) in downlink on GS if `Time` is not present

#### Notes for Reviewers

These are necessary for #1611
GS did not use `tmms` field at all up to now as far as I understand, so no reason to update changelog?

@johanstokking is the main reviewer
@KrishnaIyer please take a quick look at GS as well

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/io.go
+++ b/pkg/gatewayserver/io/io.go
@@ -372,8 +372,9 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
 			rxErrs = append(rxErrs, errRxWindowSchedule.WithCause(err).WithAttributes("window", i+1))
 			continue
 		}
-		settings.Time = nil
-		settings.Timestamp = uint32(time.Duration(em.Starts()) / time.Microsecond)
+		if settings.Time == nil {
+			settings.Timestamp = uint32(time.Duration(em.Starts()) / time.Microsecond)
+		}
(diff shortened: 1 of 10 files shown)


pkg/gatewayserver/io/io.go
```diff
@@ -372,8 +372,9 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
 			rxErrs = append(rxErrs, errRxWindowSchedule.WithCause(err).WithAttributes("window", i+1))
 			continue
 		}
-		settings.Time = nil
-		settings.Timestamp = uint32(time.Duration(em.Starts()) / time.Microsecond)
+		if settings.Time == nil {
```
johanstokking: This only works for GPS time synced gateways, hence the original code, which always used fallback to concentrator timestamp.

It's good to use GPS time when we know it's available, but still we must otherwise rely on timestamps, i.e. for absolute time scheduling using RTTs.

So add a check here to see if there's a gateway absolute time sync available

PR #1904, opened 2020-01-23
cli: Fix ttn-lw-cli failing when not in a TTY

### LINKED ISSUE #1841: Create gateway fails unless run from a TTY
## Summary
When automating the gateway creation process the 'ttn-lw-cli gateway create' command silently fails if not run from a TTY. 

Note that this appears to be the case for gateway creation. Other cli commands (gateways list etc) seems to work correctly when run without a tty.

#### Steps to Reproduce
1. Run the following collection of scripts to simulate running the command without an attached TTY

```
#!/bin/bash
true | ttn-lw-cli gateways create test-gw-$RANDOM --user-id admin
```

#### What do you see now?
The ttn-lw-cli thinks input is coming via a pipe (as there is no stdin attached) and then the inputDecoder throws an error on EOF. It then exits with a return code of 0 (success) even though the command failed to take action on the variables passed in by argument

Gateway should be created successfully

#### Environment
Version:             3.4.2
Build date:          2020-01-08T16:30:08Z
Git commit:          2714422a2
Go version:          go1.13.5
OS/Arch:             linux/amd64

Gratefully catch stdin being disconnected and not expect data for the inputDecoder

neoaggelos:
#### Summary

Closes #1841

#### Changes

- Accept inputDecoder failing with EOF. Return EOF unless any other errors occur

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Displaying total device count in application overview section when using device search in the Console
 - FQDN used for Backend Interfaces interoperability requests.
 - Exposing device sensitive fields to unrelated stack components in the Console.
+- CLI trying to read input while none available.
 
 ### Security
 
--- a/cmd/ttn-lw-cli/internal/io/io.go
+++ b/cmd/ttn-lw-cli/internal/io/io.go
@@ -97,7 +97,7 @@ func Write(w io.Writer, format string, data interface{}) (err error) {
 func IsPipe(r io.Reader) bool {
 	if f, ok := r.(*os.File); ok {
 		if stat, err := f.Stat(); err == nil {
-			return (stat.Mode() & os.ModeCharDevice) == 0
+			return (stat.Mode()&os.ModeCharDevice) == 0 && stat.Size() > 0
 		}
 	}
 	return false


cmd/ttn-lw-cli/commands/gateways.go
```diff
@@ -259,7 +262,12 @@ var (
 				return err
 			}
 
-			return io.Write(os.Stdout, config.OutputFormat, res)
+			if e := io.Write(os.Stdout, config.OutputFormat, res); e != nil {
```
rvolosatovs:
```suggestion
			if err := io.Write(os.Stdout, config.OutputFormat, res); err != nil {
```

cmd/ttn-lw-cli/internal/io/io.go
```diff
@@ -124,9 +125,23 @@ func NewJSONDecoder(r io.Reader) Decoder {
...
+		return true
+	}
+	return false
+}
+
 func (r *jsonDecoder) Decode(data interface{}) (paths []string, err error) {
 	t, err := r.rd.ReadByte()
-	if err != nil {
+	if err == io.EOF {
+		return []string{}, nil
```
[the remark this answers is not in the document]
neoaggelos: Perhaps it is just better to have `inputDecoder` be nil when the underlying reader (`io.Stdin` here) returns EOF, so we would not need the IsOK() check afterwards

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updating `supports_class_c` field in the Device General Settings Page in the Console.
 - Updating MQTT pubsub configuration in the Console
 - Handling multiple consequent updates of MQTT pubsub/webhook integrations in the Console.
+- `ttn-lw-cli` failing when input is not a TTY.
```
htdvisser: I would reword this to "CLI trying to read input while none available"

PR #1916, opened 2020-01-27
Improve MQTT error handling in Application Server

htdvisser:
#### Summary

This pull request makes some improvements to how MQTT timeouts and errors are handled in AS pubsub.

#### Changes

- Define errors
- Fix how MQTT timeouts and errors are handled

### DIFF
--- a/pkg/applicationserver/io/pubsub/provider/mqtt/driver.go
+++ b/pkg/applicationserver/io/pubsub/provider/mqtt/driver.go
@@ -58,6 +58,8 @@ func openDriverTopic(client mqtt.Client, topicName string, timeout time.Duration
 	return dt, nil
 }
 
+var errPublishFailed = errors.Define("publish_failed", "publish to MQTT topic failed")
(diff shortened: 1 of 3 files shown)


pkg/applicationserver/io/pubsub/provider/mqtt/driver.go
```diff
@@ -58,6 +58,8 @@ func openDriverTopic(client mqtt.Client, topicName string, timeout time.Duration
 	return dt, nil
 }
 
+var errPublishFailed = errors.Define("publish_failed", "publish to MQTT topic failed")
```
adriansmares: Is `Unknown` the best we can do here ? Would 'FailedPrecondition' be more appropriate ?

htdvisser: `FailedPrecondition` would indicate that a retry is pointless, and we don't actually know.

PR #1917, opened 2020-01-27
Refactor upstream handling, prepare PBA offloading and goroutine optimization

johanstokking:
#### Summary

Refactor upstream handling, prepare PBA offloading and goroutine optimization

#### Changes

- Refactored upstream handling; no more switching on gRPC service name with a concrete implementation. This makes things simpler, while not breaking configuration. Also did some necessary decoupling and gateway status handling fix
- Added `GsPba` service and `GatewayUplinkMessage`. The latter is now used in GS because we need the band in upstreams
- Small fix for moving the gc goroutine in each sub-band to the scheduler. This is a 6x goroutine save in EU, for example

#### Checklist


pkg/gatewayserver/scheduling/scheduler.go
```diff
@@ -193,6 +194,28 @@ func (s *Scheduler) findSubBand(frequency uint64) (*SubBand, error) {
 	return nil, errSubBandNotFound.WithAttributes("frequency", frequency)
 }
 
+func (s *Scheduler) gc(ctx context.Context) error {
```
KrishnaIyer: nice!

pkg/gatewayserver/upstream/ns/ns.go
```diff
@@ -18,45 +18,44 @@ package ns
...
 
-	"go.thethings.network/lorawan-stack/pkg/component"
 	"go.thethings.network/lorawan-stack/pkg/errors"
 	"go.thethings.network/lorawan-stack/pkg/gatewayserver/io"
 	"go.thethings.network/lorawan-stack/pkg/ttnpb"
 	"go.thethings.network/lorawan-stack/pkg/types"
+	"google.golang.org/grpc"
 )
 
+// Cluster represents the interface the cluster.
```
htdvisser: Maybe reword this comment

pkg/gatewayserver/io/io.go
```diff
@@ -282,16 +327,23 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
...
 		if fp == nil {
-			return 0, errFrequencyPlanNotConfigured.WithAttributes("id", request.FrequencyPlanID)
+			// The requested frequency plan is not configured for the gateway. Load the plan and enforce that it's in the same band.
+			fp, err = c.fps.GetByID(fpID)
+			if err != nil {
+				return 0, errFrequencyPlanNotConfigured.WithCause(err).WithAttributes("id", request.FrequencyPlanID)
+			}
+			if fp.BandID != c.bandID {
+				return 0, errFrequencyPlansNotFromSameBand
+			}
```
johanstokking: @htdvisser @KrishnaIyer FYI we're loading the frequency plan in the hot path if the NS's frequency plan doesn't match that of the gateway.

We have these circumstances with `EU_863_870` vs `EU_863_870_TTN` where we do need to check band. We might need to tune frequency plan store caching (it's already enabled, but implement background refresh so they're immediately available from memory).

pkg/gatewayserver/upstream/upstream.go
```diff
@@ -24,16 +24,14 @@ import (
...
 	GetDevAddrPrefixes() []types.DevAddrPrefix
 	// Setup performs all the preparation necessary to connect the handler to a particular upstream host.
-	Setup() error
+	Setup(context.Context) error
 	// ConnectGateway informs the upstream handler that a particular gateway is connected to the front end.
 	ConnectGateway(context.Context, ttnpb.GatewayIdentifiers, *io.Connection) error
-	// HandleUp handles ttnpb.UplinkMessage.
-	HandleUplink(context.Context, ttnpb.GatewayIdentifiers, ttnpb.EndDeviceIdentifiers, *ttnpb.UplinkMessage) error
+	// HandleUp handles ttnpb.GatewayUplinkMessage.
+	HandleUplink(context.Context, ttnpb.EndDeviceIdentifiers, *ttnpb.GatewayUplinkMessage) error
```
KrishnaIyer: This needs to be reverted since v2 Offloading needs the gateway identifiers/

PR #1933, opened 2020-01-28
Generate local session key ID if interop JS didn't

### LINKED ISSUE #1928: Bad payload decryption after a new join when using external JS
#### Summary
When a device joins with external JS for the 2nd time or more the payload is not decrypted correctly, it seems that session keys used by the LNS are wrong

#### Steps to Reproduce
1. Create a device using[ Semtech's join server](https://www.loracloud.com)
2. Join network
3. Send a payload => In MQTT uplink, payload is ok
4. Reset the device
5. Join the netwok
6. Send a payload => In MQTT uplink, payload is not ok

Reproduced in self hosted v3.5.0 and TTI v3.5.1-rc1

Device payload decrypted correctly each time

No

johanstokking:
#### Summary

Fix #1928

#### Changes

- According to BI 1.0, either `AppSKey`, or `SessionKeyID`, or both are sent in `JoinAns`
- If `SessionKeyID` is not sent, we still need to have one for AS to detect session changes. We cannot reliably use `DevAddr` for that (as it may return to an old one while the AS wasn't up)
- We are now, in the interop client, generating a `SessionKeyID`. This is prepended with an identifier
- AS uses this identifier to not try to get the `AppSKey` from the interop JS

#### Checklist

### DIFF
--- a/pkg/interop/client.go
+++ b/pkg/interop/client.go
@@ -17,6 +17,7 @@ package interop
 import (
 	"bytes"
 	"context"
+	"crypto/rand"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
@@ -29,6 +30,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/oklog/ulid/v2"
 	"go.thethings.network/lorawan-stack/pkg/config"
 	"go.thethings.network/lorawan-stack/pkg/encoding/lorawan"
 	"go.thethings.network/lorawan-stack/pkg/errors"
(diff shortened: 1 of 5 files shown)


pkg/interop/client.go
```diff
@@ -285,10 +293,22 @@ func (cl joinServerHTTPClient) HandleJoinRequest(ctx context.Context, netID type
 	if req.SelectedMACVersion.Compare(ttnpb.MAC_V1_1) <= 0 {
 		fNwkSIntKey = interopAns.NwkSKey
 	}
+
+	sessionKeyID := []byte(interopAns.SessionKeyID)
+	if len(sessionKeyID) == 0 {
+		log.FromContext(ctx).Debug("Interop join-accept does not contain session key ID, use random ID")
```
rvolosatovs:
```suggestion
		log.FromContext(ctx).Debug("Interop join-accept does not contain session key ID, generate an ID")
```

PR #1944, opened 2020-02-02
per lorawan spec, the channel mask should be LSB

### LINKED ISSUE #1943: ChannelMask in LinkADRReq should be LSB
#### Summary
per lorawan spec, the channel mask is LSB:
0xFF00 (LSB) = 0b00000000 11111111 (MSB) = ChMask = channels 1 thru 8

Unless I miss something, with the current code, channels 1-8 are encoded as 0x00FF (as channels
9-16)

LinkADRReq are therefore rejected with pld.ChannelMaskAck = false, and
all other ADR settings (DR for example) are discarded.

#### Steps to Reproduce
decode a link ADR mac request

lora-packet-decode --appkey A925F106BB491162C61AC270EBF8C1CA --nwkkey 8A145D9AFD6D0A9016A30DF319EAB438 --base64 YAxoAPyAHAAA+qrXxbYGPnWe
decoding from Base64:  YAxoAPyAHAAA+qrXxbYGPnWe

Decoded packet
--------------
Message Type = Data
            PHYPayload = 600C6800FC801C0000FAAAD7C5B6063E759E
          ( PHYPayload = MHDR[1] | MACPayload[..] | MIC[4] )
                  MHDR = 60
            MACPayload = 0C6800FC801C0000FAAAD7C5B6
                   MIC = 063E759E (OK)
          ( MACPayload = FHDR | FPort | FRMPayload )
                  FHDR = 0C6800FC801C00
                 FPort = 00
            FRMPayload = FAAAD7C5B6
            ==> WRONG : Plaintext = 0340**00FF**01 ('.@...')
                ( FHDR = DevAddr[4] | FCtrl[1] | FCnt[2] | FOpts[0..15] )
               DevAddr = FC00680C (Big Endian)
                 FCtrl = 80
                  FCnt = 001C (Big Endian)
                 FOpts = 
          Message Type = Unconfirmed Data Down
             Direction = down
                  FCnt = 28
             FCtrl.ACK = false
             FCtrl.ADR = true

#### What do you see now?
channelmask encoded as 00FF instead of FF00 for channels 1-8

...

channelmask encoded as FF00 for channels 1-8

...

#### Environment
lorawan-stack master

...

fix pkg/encoding/lorawan/mac.go

yes

...

pdubois22:
0xFF00 (LSB) = 0b00000000 11111111 (MSB) = ChMask = channels 1 thru 8

with the current code, channels 1-8 are encoded as 0x00FF (channels
9-16)

LinkADRReq are therefore rejected with pld.ChannelMaskAck = false, and
all other ADR settings (DR) are discarded.

#### Summary
fix #1943

...

#### Changes
pkg/encoding/lorawan/mac.go

#### Notes for Reviewers

...

#### Checklist

I have tested a LinkDRReq / LinkADRAns cycle, but I was not able to test the Unmarshalling portion.

### DIFF
--- a/pkg/encoding/lorawan/mac.go
+++ b/pkg/encoding/lorawan/mac.go
@@ -192,7 +192,7 @@ var DefaultMACCommands = MACCommandSpec{
 			b = append(b, byte((pld.DataRateIndex&0xf)<<4)^byte(pld.TxPowerIndex&0xf))
 			chMask := make([]byte, 2)
 			for i, v := range pld.ChannelMask {
-				chMask[(15-i)/8] = chMask[(15-i)/8] ^ boolToByte(v)<<(i%8)
+				chMask[i/8] = chMask[i/8] ^ boolToByte(v)<<(i%8)
 			}
 			b = append(b, chMask...)
 			b = append(b, byte((pld.ChannelMaskControl&0x7)<<4)^byte(pld.NbTrans&0xf))
@@ -201,7 +201,7 @@ var DefaultMACCommands = MACCommandSpec{
 		UnmarshalDownlink: newMACUnmarshaler(ttnpb.CID_LINK_ADR, "LinkADRReq", 4, func(phy band.Band, b []byte, cmd *ttnpb.MACCommand) error {
 			var chMask [16]bool
 			for i := 0; i < 16; i++ {
-				chMask[i] = (b[1+(15-i)/8]>>(i%8))&1 == 1
+				chMask[i] = (b[i/8]>>(i%8))&1 == 1
 			}
 			cmd.Payload = &ttnpb.MACCommand_LinkADRReq_{
 				LinkADRReq: &ttnpb.MACCommand_LinkADRReq{


rvolosatovs:
`ChMask` is a 2-byte field. According to 1.1 spec
> The channel mask (ChMask) encodes the channels usable for uplink access as follows with bit 0 corresponding to the encodes the LSB

2020-02-04-14:03:02-screenshot

So that's what I understand from the spec. : 
- `0x000000000000000*` - channel 1
- `0x00000000000000*0` - channel 2
- `0x0000000000000*00` - channel 3
...
- `0x*000000000000000` - channel 16

What you are suggesting is that the ordering actually starts with MSB?

We actually just changed this 2 weeks ago in https://github.com/TheThingsNetwork/lorawan-stack/pull/1879 The issue was exactly that the channel mask we were sending starting with MSB(like what you suggest) in US915, which is why we changed that.

I will check with https://github.com/Lora-net/LoRaMac-node and https://github.com/ARMmbed/mbed-os, to see what they expect here.

What device are you using?

pdubois22:
my undesrtanding is channel 1 is
0b0000000000000001

which is 0x0100 in LSB

channels 1-8 are 
0b0000000011111111

which is 0xFF00 in LSB

my test devices are:
Adeunis Fiel test device / EU band 
Elsys ELT2 / EU band

both started to react positively after I made the change.

rvolosatovs:
You are right. The byte order is little endian, like everything else in LoRaWAN, while bit order is actually big endian.
Fixed the decoding, tests and did some refactoring for CFList as well. https://github.com/TheThingsNetwork/lorawan-stack/pull/1946

Issue #1778, opened 2019-12-23
Javascript SDK exposes sensitive device fields


bafonins:
#### Summary

The JS SDK should not expose sensitive device fields to irrelevant components.

Related to https://github.com/TheThingsNetwork/lorawan-stack/issues/579

#### Steps to Reproduce

1. Create a new otaa device using the console
2. See `root_keys` sent to NS/AS, while this should be sent only to the JS

#### What do you see now?

Sensitive device information is exposed to irrelevant components

#### What do you want to see instead?

Stack component should only receive fields that the component is responsible for.

#### Environment

v3.3.2

#### How do you propose to implement this?

Filter device payload in the js sdk and create the request tree accordingly

#### Can you do this yourself and submit a Pull Request?

yes, @kschiffer @pgalic96 feel free to reassign

PR #1949, opened 2020-02-05
Fix device fields leak in JS SDK

bafonins:
#### Summary

This PR fixes issue in the devices service in the js sdk related to sending *all* device fields to all stack components.

References https://github.com/TheThingsNetwork/lorawan-stack/issues/579
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1778

#### Changes

- Add the `pathsFromPatch` method and related tests
- Adjust the `makeRequests` function from the devices service to filter payload for each request according to the list of allowed field masks

#### Notes for Reviewers

The logic of `splitPayload` resembles solution used in `splitPaths` very much, since both methods traverse the device entity map.

Im currently working on refactoring device handling in js sdk and will most probably change this bit later as it seems to be more of a dirty fix. However, exposing keys and other sensitive data should be fixed. I couldnt find another way of fixing the issue without doing a major refactoring mainly because of the structure of the device service. Let me know if you can find another solution.

#### Checklist

### DIFF
--- a/sdk/js/src/service/devices/split.js
+++ b/sdk/js/src/service/devices/split.js
@@ -108,7 +108,7 @@ export async function makeRequests(
   operation,
   requestTree,
   params,
-  payload,
+  payload = {},
   ignoreNotFound = false,
 ) {
   const isCreate = operation === 'create'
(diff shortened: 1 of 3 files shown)


sdk/js/src/service/devices/split.js
```diff
@@ -138,6 +138,50 @@ export async function makeRequests(
...
+            map(comp, path)
+          }
+        } else {
+          map(definition, path)
+        }
+      }
+    }
+
+    return Marshaler.payload(result.value, 'end_device')
+  }
```
kschiffer: Not sure if I miss something here, but couldn't you just compose the filtered payload using the paths array inside the `requestTree`? Like so:
```js
  function splitPayload(payload, paths) {
    if (Object.keys(payload).length === 0) return {}

    const { end_device } = payload

    const result = traverse({})
    const endDevice = traverse(end_device)

    for (const path of paths) {
      result.set(path, endDevice.get(path))
    }

    return Marshaler.payload(result.value, 'end_device')
  }
```

and then use it like e.g.:
```js
splitPayload(payload, requestTree.is)
```

PR #1955, opened 2020-02-05
Use Gateway ID from registry for Basic Station Auth

KrishnaIyer:
#### Summary

This fixes the situation where a Basic Station gateway is registered with an ID other than `eui-xxx` but the traffic endpoint is always `/eui-xxx`.

#### Changes

- Move auth check to after retrieving gateway context.
- Use retrieved Gateway ID instead of the `id` from the request.

#### Notes for Reviewers

Not yet tested. Will test in staging env before merging.

#### Checklist


pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -171,26 +171,6 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 	id := c.Param("id")
```
johanstokking: So this is always a EUI? Please name it that way then

PR #1958, opened 2020-02-06
Fix redundant header fetching on device search

### LINKED ISSUE #1910: Extend store logic for search results
#### Summary
We need to change the way we handle search results in our redux store. Currently, we treat search request just like regular list requests.

#### Why do we need this?
- The `totalCount` prop we're using will be affected by search requests, resulting in the UI showing wrong values for the total number of entities
- Search requests will trigger a `fetching` toggle for the `GET_*_LIST` actions, resulting in some elements rendering a loading spinner when they don't actually have to

- Current state of the store which includes the new pagination logic

A distinction of `LIST_*` and `SEARCH_*` requests

#### Environment
Console @ v3.5.0

Cannot think this through in detail right now, but maybe adding a `filteredIds` to the pagination store, which will hold the currently filtered entities.

Yes but @bafonins feel free to take over

pgalic96:
#### Summary

Relevant issue #1910 (not closing the issue itself, but the discussion in the issue, also I think separating search and list requests is not necessary)

#### Changes

- Create new redux request `getApplicationDeviceCount` and new store variable `applicationDeviceCount`
- Use `applicationDeviceCount` in application overview header instead of `totalCount` obtained via `getDeviceList` request
- also removed `selectDeviceFetching` selector from the application overview page since it's not relevant for the device count anymore

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updating `supports_class_c` field in the Device General Settings Page in the Console.
 - Updating MQTT pubsub configuration in the Console
 - Handling multiple consequent updates of MQTT pubsub/webhook integrations in the Console.
+- Displaying total device count in application overview section when using device search in the Console
 
 ### Security
 
--- a/pkg/webui/console/store/actions/applications.js
+++ b/pkg/webui/console/store/actions/applications.js
@@ -42,6 +42,20 @@ export const [
   { request: getApplication, success: getApplicationSuccess, failure: getApplicationFailure },
 ] = createRequestActions(GET_APP_BASE, id => ({ id }), (id, selector) => ({ selector }))
 
+export const GET_APP_DEV_COUNT_BASE = 'GET_APPLICATION_DEVICE_COUNT'
+export const [
+  {
+    request: GET_APP_DEV_COUNT,
+    success: GET_APP_DEV_COUNT_SUCCESS,
+    failure: GET_APP_DEV_COUNT_FAILURE,
+  },
+  {
+    request: getApplicationDeviceCount,
+    success: getApplicationDeviceCountSuccess,
+    failure: getApplicationDeviceCountFailure,
+  },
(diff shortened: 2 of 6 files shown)


pkg/webui/console/views/application/index.js
```diff
@@ -79,6 +80,7 @@ import PropTypes from '../../../lib/prop-types'
     loadData: id => {
       dispatch(getApplication(id, 'name,description'))
       dispatch(getApplicationsRightsList(id))
+      dispatch(getApplicationDeviceCount(id))
```
bafonins: This call is only relevant if the user goes to the overview page. In case the user has no rights to view the devices the request will result in a redundant request with `403` code. We can detect this beforehand.

You can move this call to the application overview page, something like this:
```diff
@ pkg/webui/console/views/application-overview/connect.js:28 @ import {
  selectCollaboratorsTotalCount,
  selectCollaboratorsFetching,
} from '../../store/selectors/collaborators'
+ import { getApplicationDeviceCount } from '../../store/actions/applications'
import { selectApiKeysTotalCount, selectApiKeysFetching } from '../../store/selectors/api-keys'
import { getCollaboratorsList } from '../../store/actions/collaborators'
import { getApiKeysList } from '../../store/actions/api-keys'
@ pkg/webui/console/views/application-overview/connect.js:85 @ const mapDispatchToProps = dispatch => ({
    mayViewApplicationCollaborators,
    mayViewApplicationApiKeys,
    mayViewApplicationLink,
+   mayViewDevices,
    appId,
  ) {
    if (mayViewApplicationCollaborators) dispatch(getCollaboratorsList('application', appId))
    if (mayViewApplicationApiKeys) dispatch(getApiKeysList('application', appId))
    if (mayViewApplicationLink) dispatch(getApplicationLink(appId))
+    if (mayViewDevices) dispatch(getApplicationDeviceCount(appId))
  },
})

@ pkg/webui/console/views/application-overview/connect.js:104 @ const mergeProps = (stateProps, dispatchProps, ownProps) => ({
      stateProps.mayViewApplicationCollaborators,
      stateProps.mayViewApplicationApiKeys,
      stateProps.mayViewApplicationLink,
+     stateProps.mayViewDevices,
      stateProps.appId,
    ),
})
```

Issue #1956, opened 2020-02-06
Error updating mqtt pubsub configuration in the Console


bafonins:
#### Summary

Cannot update fields from the configuration section of the pubsub form.

#### Steps to Reproduce

1. Create a new mqtt pubsub
2. Go to the edit page
3. Edit any field from the `MQTT Configuration` section except for the `Server URL` field
4. See failed update request

#### What do you see now?

Failed updated request:
```
{
   "code":3,
   "message":"error:pkg/errors:validation (invalid `mqtt`: embedded message failed validation)",
   "details":[
      {
         "@type":"type.googleapis.com/ttn.lorawan.v3.ErrorDetails",
         "namespace":"pkg/errors",
         "name":"validation",
         "message_format":"invalid `{field}`: {reason}",
         "attributes":{
            "field":"mqtt",
            "name":"ApplicationPubSubValidationError",
            "reason":"embedded message failed validation"
         },
         "correlation_id":"c12a7a748ace41c1bc3a38b41ed5e55a",
         "cause":{
            "namespace":"pkg/errors",
            "name":"validation",
            "message_format":"invalid `{field}`: {reason}",
            "attributes":{
               "field":"server_url",
               "name":"ApplicationPubSub_MQTTProviderValidationError",
               "reason":"value must be absolute"
            },
            "code":3
         },
         "code":3
      }
   ]
}
```

#### What do you want to see instead?

Successful update of the pubsub or validation error in the form

#### Environment

3.5.1

#### How do you propose to implement this?

After quick debug it seems that the backend expects the `mqtt.server_url` field to be included when the `mqtt` field is sent.

#### Can you do this yourself and submit a Pull Request?

yes, but feel free to reassign, @kschiffer @pgalic96 

cc @adriansmares is this expected behaviour?

adriansmares:
This is not expected behavior.

The following patch should enable partial updates of the provider, but please also do a sanity check for the modified whitelisting logic (before the proposed change bellow, additional paths would be added).

```
diff --git a/pkg/webui/console/views/application-integrations-pubsub-edit/index.js b/pkg/webui/console/views/application-integrations-pubsub-edit/index.js
index b906c22b6..d83614c90 100644
--- a/pkg/webui/console/views/application-integrations-pubsub-edit/index.js
+++ b/pkg/webui/console/views/application-integrations-pubsub-edit/index.js
@@ -48,8 +48,7 @@ const m = defineMessages({
[...]

PR #1936, opened 2020-01-28
Fix gateway/device location form issues

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/1175
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1934 and https://github.com/TheThingsNetwork/lorawan-stack/issues/1935

@pgalic96 Could you verify that there are no regressions related to https://github.com/TheThingsNetwork/lorawan-stack/issues/1574 ?

#### Changes

- Update `validationSchema` of the gatway/device location forms to make them numbers
- Handle cases when coordinates are set to 0
- Extend redux primitives for devices
- Move `<LocationForm />` to the console components folder
- Move location validation logic to the `<LocationForm />` component

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, database and configuration, according to the stability commitments in `README.md`.
- [x] Testing: The changes are covered with unit tests. The changes are tested manually as well.
- [x] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

Issue #1935, opened 2020-01-28
Handle zero coordinates in location form


bafonins:
#### Summary

The `<LocationForm />` component does not handle zero values properly.

#### Steps to Reproduce

1. Go to a page with `<LocationForm />`
2. Fill in the fields with `0`'s and submit
3. See successfull submission and no location set

--

1. Go to a page with `<LocationForm />`
2. Fill any field with `0` and submit
3. See successfull submission and no value for that field present in the form

#### What do you see now?

`<LocationForm />` not handling zero values

#### What do you want to see instead?

If the user sets certain coordinates to `0`'s and the backend accepts this, we should display this. Especially when the `location` entry gets created.

#### Environment

v3.5.0

#### How do you propose to implement this?

The issue is caused by the grpc gateway swallowing nullable values, however it returns the `antennas`/`locations` fields that we can use to determine location values.

#### Can you do this yourself and submit a Pull Request?

yes

PR #1961, opened 2020-02-07
Add defaults emitter utility

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/1935
Blocks https://github.com/TheThingsNetwork/lorawan-stack/pull/1936

#### Changes

- Add lodash to the js sdk
- Add factory function to create custom `defaultsEmitter` function
- Refactor defaults handling for gateways
- Refactor defaults handling for users
- Add defaults handling for devices

#### Notes for Reviewers

Unify handling default/nullable/falsy values swallowed by the grpc-gateway. Currently, we have different approaches scattered across the sdk, e.g. [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/sdk/js/src/service/users.js#L22-L30) or [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/sdk/js/src/service/pubsubs.js#L24-L49).
Didnt refactor [pubsubs](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/sdk/js/src/service/pubsubs.js#L24-L49), because of https://github.com/TheThingsNetwork/lorawan-stack/issues/1956

#### Checklist

### DIFF
--- a/sdk/js/src/util/create-defaults-emitter/index.js
+++ b/sdk/js/src/util/create-defaults-emitter/index.js
@@ -1,0 +1,64 @@
+// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { set, get, toPath } from 'lodash'
+
+/**
+ * @example
+ *
+ * const userDefaultsEmitter = createDefaultsEmitterFromFieldMask((fmKey, value) => {
+ *  if (fmKey === 'state' && (!Boolean(value) || value === null)) {
+ *    return 'STATE_REQUESTED'
+ *  }
+ * })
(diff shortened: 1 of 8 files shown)


sdk/js/src/util/create-defaults-emitter/index.js
```diff
@@ -0,0 +1,64 @@
...
+ * - `sourceObject` is the object to add default values to
+ * - `fieldMask` the field mask array
+ * The defaults emitter function returns the updated version copy of the `sourceObject`.
+ */
+const createDefaultsEmitterFromFieldMask = formatter => {
+  return function(object, fieldMask) {
+    const result = { ...object }
+
+    for (const fmKey of fieldMask) {
+      const path = toPath(fmKey)
```
pgalic96: This call is not needed since the fieldmask path format is supported by lodash therefore you can just use `get(object, fmKey)`

sdk/js/src/util/create-defaults-emitter/index.js
```diff
@@ -0,0 +1,64 @@
...
+      const value = get(object, path)
+
+      const updated = formatter(fmKey, value)
+      if (typeof updated !== 'undefined') {
+        set(result, path, updated)
+      }
```
pgalic96: This call is not needed since the fieldmask path format is supported by lodash therefore you can just use `fmKey` directly instead of `path`

bafonins:
After thinking for a while this does not seem like a good solution. It is definitely better than how we handle setting default values atm, however, proposed approach adds quite some overhead in creating/cloning new objects. Besides that, the api of `defaultsEmitter` is not clear and can be used in both mutable and immutable ways which are confusing and can cause issues.

PR #1980, opened 2020-02-12
doc: Use YAML config file for getting started guide

### LINKED ISSUE #1979: Use YAML configuration file for Getting Started guide and development environment
#### Summary

Use a YAML configuration file for the getting started guide. The current example with environment file is error-prone (e.g. quotes on option values are not handled properly)

Supersedes #1945

#### Why do we need this ?

The current example with environment file is error-prone (e.g. quotes on option values are not handled properly)

Getting Started guide currently suggests using an env file. Current docs are also wrong, because they show quoted values in configuration options.

A YAML config file. 

Update existing documentation

yes

neoaggelos:
#### Summary

Closes #1979

#### Changes

- In the getting started guide, create a stack configuration YAML file instead of an env file
- Add example `ttn-lw-stack.yml` config file

#### Notes for Reviewers

- This is what new users will probably use when trying to setup the stack for the first time, so it should not be intimidating.
- The `docker-compose.yml` for the repository is still setting environment variables, but this is needed for setting up the dev environment properly.
- Maybe the comments could use some small fixes.

#### Checklist

- [] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.

### DIFF
--- a/doc/content/guides/getting-started/configuration.md
+++ b/doc/content/guides/getting-started/configuration.md
@@ -48,14 +48,16 @@ The `ports` section exposes {{% tts %}}'s ports to the world. Port `80` and `443
 stack:
   image: 'thethingsnetwork/lorawan-stack:<the tag>'
   entrypoint: 'ttn-lw-stack'
-  command: 'start'
+  command: 'start all -c /config/ttn-lw-stack.yml'
   restart: 'unless-stopped'
   depends_on:
     - 'cockroach'
     - 'redis'
   volumes:
-    - './acme:/var/lib/acme'
     - './data/blob:/srv/ttn-lorawan/public/blob'
+    - './config/stack:/config:ro'
+    # If using Let's Encrypt
+    - './acme:/var/lib/acme'
   ports:
     - '80:1885'
     - '443:8885'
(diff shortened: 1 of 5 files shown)


doc/content/guides/getting-started/configuration.md
```diff
@@ -56,6 +56,7 @@ stack:
   volumes:
     - './acme:/var/lib/acme'
     - './data/blob:/srv/ttn-lorawan/public/blob'
+    - './ttn-lw-stack.yml:/.ttn-lw-stack.yml:ro'
```
htdvisser: Doing this might lead to some confusion, since Docker won't pick up on updates when making changes to the file. Maybe better to put it in a dir

doc/content/guides/getting-started/configuration.md
```diff
@@ -70,82 +71,100 @@ stack:
...
-
-```bash
-TTN_LW_IS_EMAIL_SENDER_NAME="{{% tts %}}"
-TTN_LW_IS_EMAIL_SENDER_ADDRESS="noreply@thethings.example.com"
-TTN_LW_IS_EMAIL_NETWORK_CONSOLE_URL="https://thethings.example.com/console"
-TTN_LW_IS_EMAIL_NETWORK_IDENTITY_SERVER_URL="https://thethings.example.com/oauth"
-```
-
-You can either use Sendgrid or an SMTP server. If you don't set up an email provider, {{% tts %}} will print emails to the server log.
+- The databases used by the stack. We will set these to the CockroachDB and Redis
```
htdvisser: We prefer paragraphs over bullet lists in our documentation.

doc/content/guides/getting-started/configuration.md
```diff
@@ -70,82 +71,100 @@ stack:
...
-TTN_LW_CONSOLE_UI_NS_BASE_URL="https://thethings.example.com/api/v3"
-TTN_LW_CONSOLE_UI_AS_BASE_URL="https://thethings.example.com/api/v3"
-TTN_LW_CONSOLE_UI_JS_BASE_URL="https://thethings.example.com/api/v3"
-TTN_LW_CONSOLE_UI_EDTC_BASE_URL="https://thethings.example.com/api/v3"
-TTN_LW_CONSOLE_UI_QRG_BASE_URL="https://thethings.example.com/api/v3"
-
-TTN_LW_CONSOLE_OAUTH_CLIENT_ID="console"
-TTN_LW_CONSOLE_OAUTH_CLIENT_SECRET=... # choose or generate a secret
+```yaml
+---
```
htdvisser: ?

doc/content/guides/getting-started/configuration.md
```diff
@@ -70,82 +71,100 @@ stack:
...
+    # smtp:
+    #   address:  '...'             # enter SMTP server address
+    #   username: '...'             # enter SMTP server username
+    #   password: '...'             # enter SMTP server password
+
+  # Web UI configuration
+  oauth:
+    ui:
+      canonical-url: 'https://thethings.example.com/oauth'
+      is: {base-url: 'https://thethings.example.com/api/v3'}
```
htdvisser: Just put `base-url` on the next line

ttn-lw-stack.yml
```diff
@@ -0,0 +1,72 @@
+---
```
htdvisser: ?

htdvisser: And maybe add a note to keep this in sync with `doc/content/guides/getting-started/configuration.md`

neoaggelos:
> Can you also align this with what's in `docker-compose.yml`?

I also think having a `docker-compose.yml` consistent with the Getting Started guide (which is currently not the case) is important, but this would have the implication of breaking the mage dev environment (which depends on overriding a few environment variables to work). As such, I believe that would fit better in a separate PR.

doc/content/guides/getting-started/configuration.md
neoaggelos: YAML syntax for document begin, not required, removed

doc/content/guides/getting-started/configuration.md
neoaggelos: Updated to use a dir-mount. Moved stack config to `config/stack/ttn-lw-stack.yml`

htdvisser:
Let's do it right, and make sure that the `docker-compose` file is consistent with the documentation. This lets people use the `docker-compose.yml` as an example, and forces us to make sure it works correctly (dogfooding).

neoaggelos:
The config file show in the docs is slightly different from the `config/stack/ttn-lw-stack.yml` used by docker-compose. Some configuration options have been removed in order to not make it scary for new-comers.

Also, the dev environment now uses the `ttn_lorawan` database as well. We can avoid that by adding a second `ttn-lw-stack-dev.yml` file if it is important.

htdvisser:
Both the config file and the contents of the documentation serve exactly the same purpose. If you consider anything to be "advanced" or "scary", it doesn't belong in there, and only belongs in the configuration reference pages.

neoaggelos:
> Both the config file and the contents of the documentation serve exactly the same purpose. If you consider anything to be "advanced" or "scary", it doesn't belong in there, and only belongs in the configuration reference pages.

In that case, the interop configuration should probably go. It is incomplete and definitely not relevant for newcomers to the stack. Will add a relevant link to the documentation instead.

docker-compose.yml
```diff
@@ -74,97 +69,10 @@ services:
...
-      # - TTN_LW_IS_EMAIL_PROVIDER=smtp
-      # - TTN_LW_IS_EMAIL_SMTP_ADDRESS=enter mail server address
-      # - TTN_LW_IS_EMAIL_SMTP_USERNAME=enter username
-      # - TTN_LW_IS_EMAIL_SMTP_PASSWORD=enter password
-
-      # If Application Server enabled, defaults for "thethings.example.com":
-      # - TTN_LW_AS_MQTT_PUBLIC_ADDRESS=thethings.example.com:1883
-      # - TTN_LW_AS_MQTT_PUBLIC_TLS_ADDRESS=thethings.example.com:8883
-
-      # If Application Server enabled, Application Server interoperability configuration:
```
rvolosatovs: Why is this removed and where is this documentation moved?

htdvisser: https://github.com/TheThingsNetwork/lorawan-stack/pull/1980#issuecomment-587091137

neoaggelos:
Fixed a broken link that made the CI fail, rebased and squashed commits. Will wait for CI and then merge

PR #1985, opened 2020-02-12
Add optional setting of pubsub credentials

### LINKED ISSUE #1947: Remove required constrains of PubSub credential fields
#### Summary
It should not be required to enter credentials for PubSubs, since it is not actually required by the backend to provide them when key authentication is used (via the `Secure` checkbox)

#### Steps to Reproduce
1. Go to Add PubSub view
2. Try to create a key authenticated PubSub without Username/Password
3. See the form validation preventing this

#### What do you see now?
Not being able to add a PubSub without username/password, when the `Secure` checkbox has been checked.

~~The required prop of the field being set/unset based on whether the `Secure` checkbox has been checked. We should also move the `Secure` checkbox (and its expanded options) to the top then.~~
Credentials are actually always optional, regardless of TLS.

#### Environment
Console @ v3.5.1

- Render the `required` props conditionally
- Tune the validation schema accordingly

@pgalic96

pgalic96:
#### Summary

Closes #1947 

#### Changes

- Add conditional requirement based on the value of `secure` field in Pubsub integration
- Update generation of server url for nats

#### Notes for Reviewers

@bafonins pls check if the nats server url is generated correctly, and if mqtt pubsub integration is working correctly since I had some issues with creating one regardless of these changes

#### Checklist


adriansmares:
I believe the above mentioned issue comes from the non-uniform ways the username and password are setup in the provider.
- The NATS provider is rather simple - it uses only the `ServerURL`, which follows basically the format of `nats://user:password@server:port`. I believe the swap between the username and the password occurs while parsing the badly formed `ServerURL`. My suggestion here would be to require both the username and password to be provided if the NATS connection is marked as secure.
- The MQTT provider is a bit more refined and has the username and password fields separately. Nonetheless, I would expect the same behavior - either both of the fields are provided, or none are provided.

kschiffer:
> My suggestion here would be to require both the username and password to be provided if the NATS connection is marked as secure.

> * The MQTT provider is a bit more refined and has the username and password fields separately. Nonetheless, I would expect the same behavior - either both of the fields are provided, or none are provided.

So let's do a checkbox then `Use credentials`, which is activated by default and disables the fields when deactivated.

Otherwise LGTM.

pgalic96:
Updated the PR to include `Use Credentials` checkbox as suggested by @kschiffer

pkg/webui/console/components/pubsub-form/index.js
```diff
@@ -69,23 +74,33 @@ export default class PubsubForm extends Component {
...
+    const initialUseCredentialsMqtt = initialIsMqtt
+      ? Boolean(initialPubsubValue.mqtt.username && initialPubsubValue.mqtt.password)
+      : true
+
+    const initialIsNats = update && 'nats' in initialPubsubValue
+    const { password, username, secure } = initialIsNats
+      ? mapNatsServerUrlToFormValue(initialPubsubValue.nats.server_url)
+      : { password: undefined, username: undefined, secure: false }
+    const initialNatsSecure = initialIsNats ? secure : false
+    const initialUseCredentialsNats = initialNatsSecure ? Boolean(password && username) : true
```
bafonins: see my review comment

```suggestion
    const initialUseCredentialsNats = initialNatsSecure ? Boolean(password || username) : true
```

pkg/webui/console/components/pubsub-form/index.js
```diff
@@ -69,23 +74,33 @@ export default class PubsubForm extends Component {
 
     const initialIsMqtt = update && 'mqtt' in initialPubsubValue
     const initialMqttSecure = initialIsMqtt ? initialPubsubValue.mqtt.use_tls : false
+    const initialUseCredentialsMqtt = initialIsMqtt
+      ? Boolean(initialPubsubValue.mqtt.username && initialPubsubValue.mqtt.password)
```
bafonins: see my review comment

```suggestion
      ? Boolean(initialPubsubValue.mqtt.username || initialPubsubValue.mqtt.password)
```

pgalic96:
> However, since credentials are not enforced by the backend it can happen that only username or password is specified via cli/api/sdk and then the edit form is in inconsistent state (the `use_credentials` is not checked while the `password` or `username` field is provided). @adriansmares such behavior seems incorrect to me, anything we can do with this?
 So it should be possible in the console as well to specify only username or only password?

bafonins:
No, I dont think so. Just take into account that this can happen (with the current backend validation) and address https://github.com/TheThingsNetwork/lorawan-stack/pull/1985#discussion_r382631818, https://github.com/TheThingsNetwork/lorawan-stack/pull/1985#discussion_r382632048

cc @adriansmares

pgalic96:
Alright. Regarding this:

> Also can we reuse the `use_credentials` checkbox between providers?

This would conflict with the idea of having `Use Credentials` checkbox checked by default. For example say we have a pubsub with nats provider which doesn't have `Use Credentials` checked. When changing the provider, all the settings are reset to default, so `Use Credentials` would be expected to be set to default value. However this would be disabled since its value is reused. Also, when changing back the existing values would be used except for this checkbox. Let me know what you think.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Handling multiple consequent updates of MQTT pubsub/webhook integrations in the Console.
 - Displaying total device count in application overview section when using device search in the Console
 - FQDN used for Backend Interfaces interoperability requests.
+- Username and password not required in pubsub integration if tls is enabled
```
```suggestion
- Changing username and password to be not required in pubsub integration
```

kschiffer:
> This would conflict with the idea of having `Use Credentials` checkbox checked by default. For example say we have a pubsub with nats provider which doesn't have `Use Credentials` checked. When changing the provider, all the settings are reset to default, so `Use Credentials` would be expected to be set to default value. However this would be disabled since its value is reused. Also, when changing back the existing values would be used except for this checkbox. Let me know what you think.

You can use a ref to the formik element to get the `setValues` method of it:
```js
const formRef = React.useRef(null)
…
const { setValues, state } = formRef.current
// then set the value when the provider is changed:
setValues({
  ...state.values,
  _use_credentials: true, // you'd need some logic here to determine a consistent value
 })
```
Don't know though if it's worth the hassle.

pgalic96:
According to @htdvisser comment here https://github.com/TheThingsNetwork/lorawan-stack/issues/2045#issuecomment-590719772 I think we need to allow setting only username for the mqtt. Not sure about nats? cc @bafonins 

Also I think reusing the `use_credentials` is not worth the hassle, but if you deem it important, I will work on fixing it.

pgalic96:
Fixed the PR to include optional setting of username, also fixed an issue where unsetting `use_credentials` would still make an update request with the previously set username and password

pkg/webui/console/components/pubsub-form/index.js
```diff
@@ -69,23 +74,34 @@ export default class PubsubForm extends Component {
...
+      ? mapNatsServerUrlToFormValue(initialPubsubValue.nats.server_url)
+      : { password: undefined, username: undefined, secure: false }
+    const initialNatsSecure = initialIsNats ? secure : false
+    const initialUseCredentialsNats = initialNatsSecure ? Boolean(password || username) : true
 
     this.state = {
       error: '',
       isMqtt: initialIsMqtt,
       mqttSecure: initialMqttSecure,
+      natsSecure: initialNatsSecure,
```
bafonins: What is the purpose of this entry? You never read it from the state

pkg/webui/console/components/pubsub-form/index.js
```diff
@@ -112,6 +128,14 @@ export default class PubsubForm extends Component {
     this.setState({ isMqtt: false })
   }
 
+  handleNatsSecureChange(event) {
+    this.setState({ natsSecure: event.target.checked })
```
bafonins: What is the purpose of this entry? You never read it from the state

pgalic96:
> 1. `mqtt.use_credentials` should not be sent to the backend

Fixed, not using this in the mapping function to determine what to send to server.
> 
> 1. Try this when editing mqtt pubsub:
> 
> * Check the `use_credentials` checkbox
> * Provide username and password, submit
> * Check the `use_credentials` checkbox again and submit
> * Check the `use_credentials` checkbox again, submit and see error
This (and the first issue) is fixed by setting values to null instead of stripping them, this issue was caused by casting the schema, but now I removed the schema casting, and the correct values are sent to backend and there is no need to cast the schema.
> 1. Why is only username required with mqtt provider?

Did you read my comment from yesterday? We also discussed it yesterday at the meeting



> According to @htdvisser comment here [#2045 (comment)](https://github.com/TheThingsNetwork/lorawan-stack/issues/2045#issuecomment-590719772) I think we need to allow setting only username for the mqtt. Not sure about nats? cc @bafonins
> 
> Also I think reusing the `use_credentials` is not worth the hassle, but if you deem it important, I will work on fixing it.

pgalic96:
@bafonins rebased and squashed the fixup comments

Issue #1730, opened 2019-12-13
Basic Station Integration: Race condition in re-connection handling causes permanent failure of uplink forwarding


beitler:
#### Summary

The Basic Station protocol is based on TCP. Occasionally it may happen, that the client drops this connection without executing a clean TCP connection close sequence. This may occur if link/net layer connectivity suddenly disappears and causes the TCP layer to reset and retry (e.g. a gateway is switching from ethernet backhaul to 3G backhaul because ethernet went away; or gateway is seeing unexpected power cycle and boots up quickly - a common unplug/plugin scenario for TTIG). If Basic Station establishes a new connection within a certain time after the last uplink for the old connection was forwarded, the LNS stops processing uplinks from this gateway permanently. It looks like this timeout is about 60 seconds. Given the symptoms and the fact that this does not happen on the v3 stack, it looks like this issue is related to #1729.

This issue has also been discussed in the [TTN forums](https://www.thethingsnetwork.org/forum/t/ttig-problems-no-location-data-wrong-date-time-wrong-channel-and-stability-issues/29652/64)

#### Steps to Reproduce
 0. To simulate unclean TCP connection termination, introduce an iptables rule which blocks TCP FIN packets: `iptables -A OUTPUT -d 52.169.76.203 --protocol tcp --tcp-flags FIN FIN -j DROP`
 1. Start Basic on the machine where the iptables rule is active. Make sure you are in an environment of regular uplinks (every 10 seconds or so). Observe `https://console.thethingsnetwork.org/gateways/<GATEWAYID>/traffic` for incoming traffic.
 2. After a few uplinks are forwarded, stop the station process with CTRL+C (the server will see an unclean TCP termination, because of the missing FIN packet). Let us define the time `T` as the time where the last uplink message was forwarded before the station process was killed.
 3. Shortly after, start the station process again (Basic Station will connect and forward uplinks).
 4. At time `T + 60 s`, the error condition kicks in.

#### What do you see now?

The error condition: The gateway console `https://console.thethingsnetwork.org/gateways/<GATEWAYID>/traffic` will stop showing uplinks while the connection between Basic Station and the Server as kept alive (TCP keep alive messages are exchanged) and Basic Station continues to receive uplinks. A TCP/IP packet capture shows that the uplinks are actually transferred over the websocket and the TCP packets are acknowledged by the server - i.e. the server definitely receives the uplink messages but does not show them in the gateway console.

#### What do you want to see instead?

[...]

Issue #1729, opened 2019-12-13
Basic Station Integration: Server-side connection termination causing unreliable downlink scheduling


beitler:
#### Summary

The Basic Station protocol server integration on the community network`*` terminates the TCP connection if 60 seconds of inactivity is detected. In an event of such a server-side disconnect, Basic Station continues running the radio and tries to re-connect with an exponential back-off. The whole re-connection procedure can potentially take multiple seconds, during which uplink messages could be received and queued internally. After the connection is established again, the internal uplink buffer is pushed to the server and the radio is re-started with the configuration contained in the router_config message. Downlinks, which are scheduled in response to any of those uplinks will be rejected by Basic Station because the xtime session is not in sync (this is the whole purpose of the xtime session indicator). Therefore, the periodic disconnects are causing less reliable downlinks. In addition to that, if an uplink with a new xtime session is forwarded before the downlink in response to an uplink from the old xtime session is scheduled, we are running into issue #1707 which blocks the gateway for any future downlinks.

The v3 stack does not seem to implement the periodic server-side disconnects due to inactivity. So, it is unclear, why it is needed on the community network. Maybe the underlying problem which introduced the need for those disconnects can be met in another way?

`*` I'm posting the issue here because it looks to me that the actual code that drives the Basic Station integration on the community network is not available on github, but I assume it is a derivative of what is done in pkg/gatewayserver/*.

#### Steps to Reproduce
 1. Start Basic Station and make sure no uplinks are forwarded within 60 seconds
 2. After the server side disconnects start sending uplink packets
 3. Wait for Basic Station to re connect.

#### What do you see now?

 - Downlinks in response to uplinks received before the new connection was established, are rejected.
 - Downlinks in response to uplinks received after the new connection was established but before the downlink was scheduled by the server, will trigger issue #1707 and block the gateway for future downlinks.

#### What do you want to see instead?

I want to see a stable connection without server-side connection terminations even when no uplinks are received. Re-connects should be rare events because they introduce a switch from one xtime session to another, which makes it impossible to schedule downlinks across xtime sessions.

#### Environment

Basic Station (latest version), TTN community network.

(On the v3 stack those disconnects do not happen).

#### How do you propose to implement this?

[...]

PR #1989, opened 2020-02-13
Disconnect existing gateway connections

### LINKED ISSUE #1987: New connections overwrite existing connections
#### Summary

New connections overwrite existing connections

#### Steps to Reproduce

1. Connect gateway over UDP
2. Connect another gateway with the same EUI
3. Disconnect any of the gateways
4. The gateway by UID gets removed from a map which is used for downlink, so downlink is broken

#### What do you see now?

Gateway Server maps the connections by UID and removes entries on gateway disconnects, while there may be another gateway connected. This can be due to an unclean disconnect or anything else. References https://github.com/TheThingsNetwork/lorawan-stack/issues/1729 and probably #1730 (and probably https://github.com/TheThingsIndustries/lorawan-stack/issues/1735). 

Actively disconnect existing connections when a gateway connects.

https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/gatewayserver/gatewayserver.go#L415

Here we probably want a loop that does a `LoadOrStore()`, and as long as there's a connection loaded, disconnect it and wait for that connection. You can close a signaling channel here: https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/gatewayserver/gatewayserver.go#L476

That waiting is necessary to clean up all upstream handling of the old connection and ensure that the `Delete()` from the map is the concerning connection, and not already the new connection, still breaking downlink.

Can review, can do

johanstokking:
#### Summary

Closes #1987

#### Changes

- Disconnect existing gateway connections before handling a new connection. This cleans up all existing Basic Station, MQTT and gRPC connections and is tested. UDP is not necessary since it maps by EUI and has its own connection expiry mechanism

#### Notes for Reviewers

n/a

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -68,7 +68,7 @@ type GatewayServer struct {
 
 	upstreamHandlers map[string]upstream.Handler
 
-	connections sync.Map
+	connections sync.Map // string to connectionEntry
 }
 
 func (gs *GatewayServer) getRegistry(ctx context.Context, ids *ttnpb.GatewayIdentifiers) (ttnpb.GatewayRegistryClient, error) {
@@ -345,6 +345,13 @@ var (
 	)
 )
 
+var errNewConnection = errors.DefineAborted("new_connection", "new connection from same gateway")
+
+type connectionEntry struct {
+	*io.Connection
+	upstreamDone chan struct{}
+}
+
 // Connect connects a gateway by its identifiers to the Gateway Server, and returns a io.Connection for traffic and
 // control.
 func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids ttnpb.GatewayIdentifiers) (*io.Connection, error) {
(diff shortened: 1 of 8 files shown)


johanstokking:
References #1729 #1730 https://github.com/TheThingsIndustries/lorawan-stack/issues/1735

pkg/gatewayserver/gatewayserver.go
```diff
@@ -412,10 +419,23 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids
 	if err != nil {
 		return nil, err
 	}
-	gs.connections.Store(uid, conn)
+	connEntry := connectionEntry{
+		Connection:   conn,
+		upstreamDone: make(chan struct{}),
+	}
+	for existing, exists := gs.connections.LoadOrStore(uid, connEntry); exists; existing, exists = gs.connections.LoadOrStore(uid, connEntry) {
```
KrishnaIyer: I see a potential deadlock here. 
- There's Conn A.
- Conn B now replaces Conn A and blocks until A is closed.
- Now Conn C will replace Conn B and force it to close.
- But if Conn B continues the loop before Conn C closes it  via `L431`, then Conn B again replaces Conn C and try to close it.

It's kinda rare but theoretically possible. How about disallowing a second (or multiple) connection and also closing the existing connection?
Then when the gateway retries we can have a clean connect from the beginning?

johanstokking: >But if Conn B continues the loop before Conn C closes it 

This doesn't happen. `LoadOrStore()` is atomic; if it gets loaded, that existing connection gets closed and we can wait for that. If it doesn't get loaded, the connection is by definition stored, the loop breaks and the upstream handler gets started. Even if the upstream handler is not up yet, connection C can safely wait for the channel close, as connection B will definitely start that goroutine.

johanstokking:
@neoaggelos indeed the front-ends stop receiving traffic when the connection gets closed.

As far as there is buffered traffic: closing the connection context stops the upstream handlers as soon as their current message in flight is done. Buffered messages from the old connection are discarded and they’ll get garbage collected because everything goes out of scope.

In fact, the for-loop only continues when all upstream handlers are done so nothing happens in parallel. It may be that the new connection needs to wait for one uplink message in flight to be published to Packet Broker, for example, but that’s the worst that can happen, and also that context gets cancelled.

Or do you mean something else?

johanstokking:
Ah I see what you mean. Yes, a read on a socket returns an error when the connection gets closed.

PR #1991, opened 2020-02-14
Fix changing pubsub provider in Console

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/1956 and fixes the pubsub form to switch betwee nats/mqtt provider.
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2061

#### Changes

- Properly handle changing provider in the pubsub form
- Fix resetting the form values when switching provider

#### Notes for Reviewers

The problem is in the provider field mask. Basically, when we change the provider from `nats` to `mqtt` and vice versa, the backend expects the `provider` field mask (id doesnt matter if we include all nested field masks, e.g. `provider.mqtt` , `provider.mqtt.username`, …), however in case of just updating provider fields it is not needed and the backend accepts only the nested field masks. If we add `provider` when, say, updating just `username` the backend requires `server_url`  to be present as well.

The simplest solution is to sent the whole pubsub payload on update requests. In this case the `provider` field mask is always present in the payload as well as all required fields.

#### Checklist

### DIFF
--- a/pkg/webui/console/store/reducers/pubsubs.js
+++ b/pkg/webui/console/store/reducers/pubsubs.js
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { merge } from 'lodash'
+import { mergeWith } from 'lodash'
 
 import {
   GET_PUBSUB,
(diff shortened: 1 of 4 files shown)


pkg/webui/console/store/reducers/pubsubs.js
```diff
@@ -61,7 +61,25 @@ const pubsubs = function(state = defaultState, { type, payload }) {
         ...state,
         entities: {
           ...state.entities,
-          [pubsubId]: merge({}, state.entities[pubsubId], payload),
+          [pubsubId]: mergeWith(
```
kschiffer: Why don't we just use the whole return value as new value without merging, since the backend will return the whole updated object anyway?

bafonins:
@adriansmares found an issue in the js sdk related to this commit https://github.com/TheThingsNetwork/lorawan-stack/pull/1991/commits/873af74e537946b3b55d9a65c1822ad6325b7c41
Always adding the `provider` field mask if any of `provider.{mqtt|nats}` field changed can result in reseting the rest of the fields.

Technically this doesnt break the console, however this makes field mask generation for the pubsubs service broken, you can use [thsi script](https://gist.github.com/bafonins/6676f7daf1446f3c0071e2c6922bd9de) to check this.

Will continue on this

bafonins:
Will continue after https://github.com/TheThingsNetwork/lorawan-stack/pull/1985 to introduce less changes

bafonins:
Merging because of regression in https://github.com/TheThingsNetwork/lorawan-stack/issues/2061
Will fix issue mentioned in https://github.com/TheThingsNetwork/lorawan-stack/pull/1991#issuecomment-587072856 when working on https://github.com/TheThingsNetwork/lorawan-stack/issues/2068

Issue #1981, opened 2020-02-12
Gateway Server tests


neoaggelos:
#### Summary

Re-factor the gateway server tests.

#### Why do we need this?

Currently the tests are written in a very hard to read manner: There are 4 mock frontends defined as structs inline with the actual test cases. This runs a full test for gateway server, but makes it really hard to test simple scenarios, or test with different gateway settings.

This was something that came up during #1849, and will probably be an issue with writing proper tests for #1815 as well.

#### What is already there? What do you see now?

Hard to extend tests

#### What is missing? What do you want to see?

- Move mock frontends code in a separate file (or at least separate them from the test cases themselves
- Make it easier to add different test scenarios

#### Environment

- dev 3.5.2

#### How do you propose to implement this?

Re-factor the gateway server tests

#### Can you do this yourself and submit a Pull Request?

Can try

neoaggelos:
This issue was created when the GS tests were slow and contained lots of flakes. The situation is already much better and we haven't had problems in a long time.

Closing this, we can return to it if ever GS tests become slow/flaky again.

PR #1999, opened 2020-02-16
Update gateway location from status messages

### LINKED ISSUE #1815: Update gateway location observed by Gateway Server

neoaggelos:
#### Summary

Closes #1815 

#### Changes

- Add new gateway field `update_location_from_status` (and accompanying option `--gs.update-gateway-location-debounce-time`)
- On new connections, spawn a go-routine if `update_location_from_status` is true for the gateway. When a new status message arrives that contains antenna location, update gateway antenna location on the GS and the IS. Otherwise, do nothing more.
- Location updates are debounced in case status messages start flooding, debounce time is configurable.

#### Notes for Reviewers

- Added TODOs for handling multiple gateway antennas. Not implemented because it would make this PR very large. Also, handling multiple antennas has not been fully implemented in a few other places.
- Tests are starting to become a little messy. Diff is messed up because of indentation, see my comments for the important changes.
- Status messages typically arrive every ~30 seconds, so debounce is not really needed under normal circumstances.
- Please review authentication between GS and IS, it probably needs changing.

#### Extra

- Configuring the new gateway options is left for a separate PR after this one is merged, for clarity reasons.
- Requiring/not requiring auth per gateway for updating antenna locations on the IS is also left for a future PR for clarity reasons. Adding this would also require a minor version bump if I am not mistaken.

#### Checklist


pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -122,1083 +122,1124 @@ func TestGatewayServer(t *testing.T) {
 	time.Sleep(timeout) // Wait for tasks to be started.
 
 	for _, publicLocation := range []bool{true, false} {
-		is.add(ctx, ttnpb.GatewayIdentifiers{
-			GatewayID: registeredGatewayID,
-			EUI:       &registeredGatewayEUI,
-		}, registeredGatewayKey, publicLocation)
+		for _, updateLocationFromStatus := range []bool{true, false} {
```
neoaggelos: Tests: change 1

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -122,1083 +122,1124 @@ func TestGatewayServer(t *testing.T) {
...
 
-				wg := &sync.WaitGroup{}
-				wg.Add(1)
-				go func() {
-					defer wg.Done()
-					err := ptc.Link(ctx, t, ids, registeredGatewayKey, upCh, downCh)
-					if !errors.IsCanceled(err) {
-						t.Fatalf("Expected context canceled, but have %v", err)
+					// Expected location for RxMetadata
+					location := &ttnpb.Location{
```
neoaggelos: Tests: change 2

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -122,1083 +122,1124 @@ func TestGatewayServer(t *testing.T) {
...
 								}
-							}
-							if expected := tc.Up.TxAcknowledgment; expected != nil {
-								select {
-								case <-upEvents["gs.down.tx.success"]:
-								case evt := <-upEvents["gs.down.tx.fail"]:
-									received, ok := evt.Data().(ttnpb.TxAcknowledgment_Result)
-									if !ok {
-										t.Fatal("No acknowledgment attached to the downlink emission fail event")
+
```
neoaggelos: Tests: change 3

pkg/gatewayserver/gatewayserver_util_test.go
```diff
@@ -97,6 +98,17 @@ func (is *mockIS) Get(ctx context.Context, req *ttnpb.GetGatewayRequest) (*ttnpb
 	return gtw, nil
 }
 
+func (is *mockIS) Update(ctx context.Context, req *ttnpb.UpdateGatewayRequest) (*ttnpb.Gateway, error) {
```
neoaggelos: Tests: change 4.

Maybe we should not ignore the FieldMasks and blindly update antennas.

johanstokking:
> * Status messages typically arrive every ~30 seconds, so debounce is not really needed under normal circumstances.

With 10K gateways already connected to PCN, that is 20K per minute, i.e. 333 per second. We really shouldn't be updating a relational database that often, with information that is not that volatile in most cases. So that debounce is absolutely necessary.

> * On new connections, spawn a go-routine if `update_location_from_status` is true for the gateway.

Didn't look at the code yet, but we should avoid spawning long-running goroutines per gateway connection, especially if they're not doing hot path stuff. So either combine a goroutine that reads gateway status updates from a channel or work with short-running goroutines.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -636,6 +643,83 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
+		"gateway_uid", uid,
+	))
+
+	var err error
+	var callOpt grpc.CallOption
+	callOpt, err = rpcmetadata.WithForwardedAuth(ctx, gs.AllowInsecureForCredentials())
+	if errors.IsUnauthenticated(err) {
+		callOpt = gs.WithClusterAuth()
+	} else if err != nil {
+		logger.WithError(err).Errorf("Not authenticated for location updates")
```
johanstokking: Errors are intended for operations to look at, which is not bad user config. Also no need for the `f`. Same below

pkg/gatewayserver/io/io.go
```diff
@@ -233,12 +235,27 @@ func (c *Connection) HandleStatus(status *ttnpb.GatewayStatus) error {
 	case c.statusCh <- status:
 		c.lastStatus.Store(status)
 		atomic.StoreInt64(&c.lastStatusTime, time.Now().UnixNano())
+
+		antennaLocations := status.GetAntennaLocations()
+		if len(antennaLocations) > 0 && c.gateway.UpdateLocationFromStatus {
+			// TODO: handle multiple antenna locations
```
johanstokking: TODOs always need a referenced issue

pkg/gatewayserver/gatewayserver_test.go
johanstokking: This escalates.

Can you instead write test cases with flags, like

```go
for _, := range []struct { Flag1, Flag2 bool } { { ... }, { ... } } {
   ...
}
```

johanstokking: For location and updating location from status, and probably other things in the future.

We should also start planning to refactor this testing, and have the frontends test theirselves against a test bench, instead of putting the frontends as test cases, but that's not urgent.

neoaggelos: Yeah, it actually feels very bad in the current state. I have opened an issue for updating this already https://github.com/TheThingsNetwork/lorawan-stack/issues/1981

Not too sure if I understand the proposed change, will try and ask if something goes bad

api/gateway.proto
```diff
@@ -125,8 +125,12 @@ message Gateway {
   DownlinkPathConstraint downlink_path_constraint = 18 [(validate.rules).enum.defined_only = true];
   // Adjust the time that GS schedules class C messages in advance. This is useful for gateways that have a known high latency backhaul, like 3G and satellite.
   google.protobuf.Duration schedule_anytime_delay = 19 [(gogoproto.stdduration) = true, (gogoproto.nullable) = true];
+  // update the location of this gateway from status messages
+  bool update_location_from_status = 21;
+  // debounce time for gateway location updates from status messages.
+  google.protobuf.Duration update_location_from_status_debounce_time = 22 [(gogoproto.stdduration) = true, (gogoproto.nullable) = true];
```
htdvisser: I'm not sure how smart it is to let users configure this. The entire idea of debouncing is that we protect our database from malicious users/gateways, so if those users can just disable that protection by setting it to one nanosecond, we can just as well not debounce at all.

neoaggelos:
> With 10K gateways already connected to PCN, that is 20K per minute, i.e. 333 per second. We really shouldn't be updating a relational database that often, with information that is not that volatile in most cases. So that debounce is absolutely necessary.

Definitely keeping debounce in. Also see my comment at https://github.com/TheThingsNetwork/lorawan-stack/issues/1815#issuecomment-586881811
 
> Didn't look at the code yet, but we should avoid spawning long-running goroutines per gateway connection, especially if they're not doing hot path stuff. So either combine a goroutine that reads gateway status updates from a channel or work with short-running goroutines.

Is there any way to perform multiple updates (for many gateways) at once? If we had that, then we let the Gateway Server perform batch updates and get much better performance (but would need some logic for handling different permissions/errors/etc)

> I'm not sure how smart it is to let users configure this. The entire idea of debouncing is that we protect our database from malicious users/gateways, so if those users can just disable that protection by setting it to one nanosecond, we can just as well not debounce at all.

I see your point. But some gateways may be OK with `1h` debounce times, whereas others could require seconds. Maybe add an option for minimum allowed debounce time? Or drop the ability for setting different values for gateway altogether?

johanstokking:
> But some gateways may be OK with `1h` debounce times, whereas others could require seconds. Maybe add an option for minimum allowed debounce time? Or drop the ability for setting different values for gateway altogether?

Let's not over-engineer this and stick with a minimum setting in GS config.

> Is there any way to perform multiple updates (for many gateways) at once?

No, indeed, the per-message auth is the first roadblock.

BTW I'm also good with combining this with the goroutine that (debounces) update connection stats, which is also per-gateway I believe.

pkg/ttnpb/udp/translation.go
```diff
@@ -298,8 +298,18 @@ func FromGatewayUp(up *ttnpb.GatewayUp) (rxs []*RxPacket, stat *Stat, ack *TxPac
...
+			stat = &Stat{
+				Time: ExpandedTime(up.GatewayStatus.Time),
+				Long: &up.GatewayStatus.AntennaLocations[0].Longitude,
+				Lati: &up.GatewayStatus.AntennaLocations[0].Latitude,
+				Alti: &up.GatewayStatus.AntennaLocations[0].Altitude,
+			}
+		} else {
+			stat = &Stat{
+				Time: ExpandedTime(up.GatewayStatus.Time),
+			}
```
johanstokking: Style-wist; DRY, so 

```go
stat = &Stat{
	Time: ExpandedTime(up.GatewayStatus.Time),
}
if len(up.GatewayStatus.AntennaLocations) > 0 {
	loc := up.GatewayStatus.AntennaLocations[0]
	stat.Long = loc.Longitude
	...
}
```

pkg/gatewayserver/gatewayserver_test.go
neoaggelos: Tests have been re-factored to a new test case.

pkg/gatewayserver/gatewayserver.go
neoaggelos: Removed log message

api/gateway.proto
neoaggelos: Changed to a global GS option, defaulting to 1 hour per gateway.

neoaggelos:
> Let's not over-engineer this and stick with a minimum setting in GS config.

Changed to an option on the GS for all gateways, defaulting to 1 hour. Still think we could do better. In 10K gateways, very few would actually need to keep being updated.

> BTW I'm also good with combining this with the goroutine that (debounces) update connection stats, which is also per-gateway I believe.

That would probably cause some serious merging trouble down the line.

johanstokking:
> Still think we could do better. In 10K gateways, very few would actually need to keep being updated.

We can always do better. Please file a backlog issue marked technical debt for that.

api/gateway.proto
```diff
@@ -125,8 +125,11 @@ message Gateway {
   DownlinkPathConstraint downlink_path_constraint = 18 [(validate.rules).enum.defined_only = true];
   // Adjust the time that GS schedules class C messages in advance. This is useful for gateways that have a known high latency backhaul, like 3G and satellite.
   google.protobuf.Duration schedule_anytime_delay = 19 [(gogoproto.stdduration) = true, (gogoproto.nullable) = true];
+  // update the location of this gateway from status messages
+  bool update_location_from_status = 21;
+  // debounce time for gateway location updates from status messages.
 
-  // next: 21
+  // next: 23
```
htdvisser: Next is 22

api/gateway.proto
```diff
@@ -125,8 +125,11 @@ message Gateway {
   DownlinkPathConstraint downlink_path_constraint = 18 [(validate.rules).enum.defined_only = true];
   // Adjust the time that GS schedules class C messages in advance. This is useful for gateways that have a known high latency backhaul, like 3G and satellite.
   google.protobuf.Duration schedule_anytime_delay = 19 [(gogoproto.stdduration) = true, (gogoproto.nullable) = true];
+  // update the location of this gateway from status messages
+  bool update_location_from_status = 21;
+  // debounce time for gateway location updates from status messages.
```
htdvisser: This comment applies to a removed field

pkg/identityserver/store/gateway.go
```diff
@@ -60,6 +60,9 @@ type Gateway struct {
 	ScheduleAnytimeDelay   int64 `gorm:"default:0 not null"`
 	DownlinkPathConstraint int
 
+	UpdateLocationFromStatus             bool  `gorm:"not null"`
+	UpdateLocationFromStatusDebounceTime int64 `gorm:"default:0 not null"`
```
htdvisser: This field was removed

neoaggelos:
Rebased in order to get the CI fix from master

johanstokking:
@neoaggelos please keep an eye on the commit messages;

```
1661a2b86 gs: Small fixes

5b902ac9c gs: Small fix on GS test

6c589a1f0 gs: Small code update
```

This is not clear enough, regardless of the body in the last commit. Please use fixup commits and/or clarify the commit message. Same for commits like these: https://github.com/TheThingsNetwork/lorawan-stack/pull/1999/commits/9ffc8df8b83cf92fa196ad65fb98d1bc40fdfa03

cmd/internal/shared/gatewayserver/config.go
```diff
@@ -25,7 +26,8 @@ import (
 
 // DefaultGatewayServerConfig is the default configuration for the GatewayServer.
 var DefaultGatewayServerConfig = gatewayserver.Config{
-	RequireRegisteredGateways: false,
+	RequireRegisteredGateways:         false,
+	UpdateGatewayLocationDebounceTime: 60 * time.Minute,
```
johanstokking: `time.Hour`

doc/content/reference/configuration/gateway-server.md
```diff
@@ -13,6 +13,7 @@ The Gateway Server forwards traffic to upstream hosts based on the `gs.forward`
 ## Security Options
 
 - `gs.require-registered-gateways`: Require the gateways to be registered in the Identity Server
+- `gs.update-gateway-location-debounce-time`: Debounce time between location updates for a single gateway.
```
johanstokking: This is not a security option

johanstokking: Also use same text as config description

pkg/gatewayserver/config.go
```diff
@@ -36,7 +38,8 @@ type BasicStationConfig struct {
 
 // Config represents the Gateway Server configuration.
 type Config struct {
-	RequireRegisteredGateways bool `name:"require-registered-gateways" description:"Require the gateways to be registered in the Identity Server"`
+	RequireRegisteredGateways         bool          `name:"require-registered-gateways" description:"Require the gateways to be registered in the Identity Server"`
+	UpdateGatewayLocationDebounceTime time.Duration `name:"update-gateway-location-debounce-time" description:"Debounce time for gateway location updates from status messages"`
```
johanstokking: This one

pkg/gatewayserver/gatewayserver.go
```diff
@@ -438,6 +440,10 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids
 	logger.Info("Connected")
 	go gs.handleUpstream(connEntry)
 
+	if gtw.UpdateLocationFromStatus {
+		go gs.handleLocationUpdates(connEntry)
```
johanstokking: Did you consider combining this goroutine with that of updating the connection stats?

pkg/gatewayserver/gatewayserver.go
```diff
@@ -638,6 +644,76 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
+				},
+				FieldMask: pbtypes.FieldMask{
+					Paths: []string{
+						"antennas",
+					},
+				},
+			}, callOpt)
+
+			if err != nil {
+				logger.WithError(err).Errorf("Antenna location update failed")
```
johanstokking: - Errors are for findings bugs, use `Warn`
- No need for the formatted version
- `Failed to ...`

pkg/gatewayserver/gatewayserver.go
```diff
@@ -638,6 +644,76 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
+
+			antennas := conn.Gateway().Antennas
+			if location != nil && len(antennas) > 0 {
+				// TODO: Handle multiple antenna locations (https://github.com/TheThingsNetwork/lorawan-stack/issues/2006).
+				location.Source = ttnpb.SOURCE_GPS
+				antennas[0].Location = *location
+
+				conn.UpdateAntennas(antennas)
+			} else {
+				logger.Warnf("No antenna, not updating location")
```
johanstokking: This is not a warning

pkg/gatewayserver/gatewayserver.go
```diff
@@ -638,6 +644,76 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
+				location = <-conn.Location()
+			}
+
+			antennas := conn.Gateway().Antennas
+			if location != nil && len(antennas) > 0 {
+				// TODO: Handle multiple antenna locations (https://github.com/TheThingsNetwork/lorawan-stack/issues/2006).
+				location.Source = ttnpb.SOURCE_GPS
+				antennas[0].Location = *location
+
+				conn.UpdateAntennas(antennas)
```
johanstokking: Why do we do this?

pkg/gatewayserver/gatewayserver.go
neoaggelos: These two goroutines may be totally unrelated (one updates Redis, other updates GS + IS). Location updates may also be disabled per-gateway, whereas stats updates are always going to happen. Finally, it would be really hard to maintain two separate versions, merging every time one changes could be a serious cause of troubles in the long run.

neoaggelos: Still, eager to test out how that would work. Should we add an issue to the backlog for that?

pkg/gatewayserver/gatewayserver.go
neoaggelos: Leftover from a few tests, removed

neoaggelos:
Rebasing with master to update conflicts in the changelog. Also, CI seems stuck since yesterday. Will squash commits after reviewing is done.

pkg/gatewayserver/gatewayserver.go
johanstokking: >These two goroutines may be totally unrelated (one updates Redis, other updates GS + IS).

They are related in the sense that there's a non-hot-path goroutine per gateway connection, that updates persistent state about a gateway. Also, these tasks don't bite each other.

>Still, eager to test out how that would work. Should we add an issue to the backlog for that?

Let's focus on adding functionality. I'm just trying to avoid spinning up a goroutine per gateway connection for each new feature.

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Class B support.
 - WebSocket Ping-Pong support for Basic Station frontend in the Gateway Server.
 - LoRaWAN 1.0.4 support.
+- Update gateway antenna location from incoming status message (see `update_location_from_status` gateway field and `--gs.update-gateway-location-debounce-time` option).
+  - This requires a database migration (`ttn-lw-stack is-db migrate`) because of the added columns.
```
johanstokking: Please rebase and move this to the Unreleased section

doc/content/reference/configuration/gateway-server.md
```diff
@@ -10,6 +10,12 @@ The Gateway Server forwards traffic to upstream hosts based on the `gs.forward`
 
 - `gs.forward`: Forward the DevAddr prefixes to the specified hosts. This parameter accepts a string in the format `name=devaddrprefixes`
 
+## Location Update options
```
johanstokking: Title case for titles; Options

pkg/gatewayserver/gatewayserver.go
```diff
@@ -638,6 +644,72 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
 	}
 }
 
+func (gs *GatewayServer) handleLocationUpdates(conn connectionEntry) {
+	ctx := conn.Context()
+	uid := unique.ID(ctx, conn.Gateway())
+
+	logger := log.FromContext(ctx).WithFields(log.Fields(
+		"protocol", conn.Frontend().Protocol(),
+		"gateway_uid", uid,
```
johanstokking: Aren't these in the connection's context already?

pkg/gatewayserver/gatewayserver.go
```diff
@@ -638,6 +644,72 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
+		case <-ctx.Done():
+			return
+		case location = <-conn.Location():
+			// Drop older updates
+			for len(conn.Location()) > 0 {
+				location = <-conn.Location()
+			}
+
+			antennas := conn.Gateway().Antennas
+			if location != nil && len(antennas) > 0 {
```
johanstokking: How can `location` be `nil`?

pkg/gatewayserver/io/io.go
```diff
@@ -233,6 +235,16 @@ func (c *Connection) HandleStatus(status *ttnpb.GatewayStatus) error {
...
 		atomic.StoreInt64(&c.lastStatusTime, time.Now().UnixNano())
+
+		antennaLocations := status.GetAntennaLocations()
+		if len(antennaLocations) > 0 && c.gateway.UpdateLocationFromStatus {
+			// TODO: Handle multiple antenna locations (https://github.com/TheThingsNetwork/lorawan-stack/issues/2006).
+			select {
+			case c.locCh <- antennaLocations[0]:
+			default:
+				log.FromContext(c.ctx).Warnf("Drop gateway location")
+			}
```
johanstokking: Why are you treating locations like traffic, i.e. storing them in a buffer, but then draining the buffer to get the latest location?

We are already storing state in `c .lastStatus`. Why not send a signal over a channel that the last status updated? That can be a buffered channel of size 1. You don't need to warn here if that channel is full; the signal is ready in the channel, and the receiver can read `c.lastStatus` and get the latest location from there. That also decouples functionality; now there's location stuff here _and_ in the goroutine.

neoaggelos: Updated to use a signal channel instead.

neoaggelos:
Also rebased and squashed commits

pkg/gatewayserver/gatewayserver.go
neoaggelos: Agreed, if we are in no hurry for this particular feature, let's do this. Waiting for #2035 to be merged, then I will rebase and merge the two goroutines.

A simple solution would be to have a single `gatewayStatusChanged` channel, which would replace the `locationChanged` and the `statsChanged` channels.

johanstokking: I think having separate channels is fine, you can select on both. You just need to figure out something smart for the debounce. An `After` also spins up a goroutine so that ain't gonna help.

Don't spend too much time on this. Sleeping goroutines is not a big problem.

PR #2000, opened 2020-02-17
Support multiple resulting messages from one format

### LINKED ISSUE #1148: Support multiple resulting messages from one format
#### Summary

One `ttnpb.ApplicationUp` should be able to result into multiple front-end messages, instead of just one.

#### Why do we need this?

Currently, one `ttnpb.ApplicationUp` is marshalled into only one byte slice that results into one front-end message (one HTTP call, one MQTT message etc.). However, if the payload of the uplink contains multiple measurements (think of multiple sensors attached to one device), it would be helpful to allow certain formats to create multiple messages for each measurement.

https://github.com/TheThingsNetwork/lorawan-stack/blob/9112dd7bcd3f1f03190055542908b427b27acd1c/pkg/applicationserver/io/formatters/formatter.go#L23
The format currently returns only one message to be sent to the front-ends. 

The result should be changed from `[]byte` to `[][]byte`.

#### Environment

93ea01b4b5af2672da5883b570be825a54b8afc1

Change the `Format` interface. After that, each front-end should be changed to send all of the messages instead of just one:
- Webhooks
https://github.com/TheThingsNetwork/lorawan-stack/blob/9112dd7bcd3f1f03190055542908b427b27acd1c/pkg/applicationserver/io/web/webhooks.go#L325-L375
The result should be a slice of requests (so `[]http.Request` instead of `http.Request`), and the result of `format.FromUp` should be used to create these requests.
- PubSub
https://github.com/TheThingsNetwork/lorawan-stack/blob/93ea01b4b5af2672da5883b570be825a54b8afc1/pkg/applicationserver/io/pubsub/pubsub.go#L146-L158
Do `topic.Send` for each result of `i.format.FromUp`.
- MQTT
https://github.com/TheThingsNetwork/lorawan-stack/blob/93ea01b4b5af2672da5883b570be825a54b8afc1/pkg/applicationserver/io/mqtt/mqtt.go#L152-L163
Same as above.
- JSON format
https://github.com/TheThingsNetwork/lorawan-stack/blob/93ea01b4b5af2672da5883b570be825a54b8afc1/pkg/applicationserver/io/formatters/json.go#L34
Wrap the result into an `[][]byte`, since the default JSON format returns only one message.
- Protobuf format
https://github.com/TheThingsNetwork/lorawan-stack/blob/93ea01b4b5af2672da5883b570be825a54b8afc1/pkg/applicationserver/io/formatters/protobuf.go#L22
Same as above.

Yes, but I will leave this to the community.

mihirr93:
#### Summary

Closes #1148

#### Changes

- Changing the return types of both the interface and the implementations such that `FromUp` can return multiple payloads (a slice of slices of byte).
- Wrapping the return in existing Protobuf and JSON formats, so that there is no change in the behaviour and as such continues to return only 1 payload.
- Updating the call sites of `FromUp` and make them loop through the list of payloads and send each of them.

#### Notes for Reviewers

None.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Do not use `personal-files` plugin for snap package.
+
 ### Deprecated
 
 ### Removed
@@ -36,6 +38,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Exposing device sensitive fields to unrelated stack components in the Console.
 - CLI trying to read input while none available.
 - Reconnections of gateways whose previous connection was not cleaned up properly. New connections from the same gateway now actively disconnects existing connections.
+- `ttn-lw-stack` and `ttn-lw-cli` file permission errors when installed using snap.
+  - You may need to run `sudo snap connect ttn-lw-stack:personal-files`
 
 ## [3.5.2] (2020-02-06)
 
@@ -57,8 +61,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Crashes on Gateway Server start when traffic flow started while The Things Stack was still starting.
 - Not detecting session change in Application Server when interop Join Server did not provide a `SessionKeyID`.
-- `ttn-lw-stack` and `ttn-lw-cli` file permission errors when installed using snap.
-  - You may need to run `sudo snap connect ttn-lw-stack:personal-files`
 
 ## [3.5.0] (2020-01-24)
 
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -115,16 +115,11 @@ snapcrafts:
     grade: stable
     confinement: strict
     publish: true
-    plugs:
-      personal-files:
-        read:
-        - $HOME/.ttn-lw-stack.yml
-        - $HOME/.ttn-lw-cli.yml
     apps:
       ttn-lw-stack:
-        plugs: [ home, network, network-bind, personal-files ]
+        plugs: [ home, network, network-bind ]
       ttn-lw-cli:
-        plugs: [ home, network, network-bind, personal-files ]
+        plugs: [ home, network, network-bind ]
 
 brews:
   - name: ttn-lw-stack
--- a/doc/content/guides/getting-started/cli/_index.md
+++ b/doc/content/guides/getting-started/cli/_index.md
@@ -24,7 +24,6 @@ $ brew install TheThingsNetwork/lorawan-stack/ttn-lw-stack
 ```bash
 $ sudo snap install ttn-lw-stack
 $ sudo snap alias ttn-lw-stack.ttn-lw-cli ttn-lw-cli
-$ sudo snap connect ttn-lw-stack:personal-files      # for $HOME/.ttn-lw-cli.yml
 ```
 
 ### Binaries
@@ -38,6 +37,8 @@ The command-line needs to be configured to connect to your deployment on `thethi
 1. Environment: `export TTN_LW_CONFIG=/path/to/ttn-lw-cli.yml`
 2. Command-line flag: `-c /path/to/ttn-lw-cli.yml`
 3. Save as `.ttn-lw-cli.yml` in `$XDG_CONFIG_HOME`, your home directory, or the working directory.
+
+> NOTE: When using the snap packages, `~/.ttn-lw-cli.yml` will fail with permission errors. Choose a different path.
 
 ```yaml
 oauth-server-address: 'https://thethings.example.com/oauth'


pkg/applicationserver/io/formatters/json_test.go
```diff
@@ -81,10 +81,12 @@ func TestJSONUpstream(t *testing.T) {
...
 			buf, err := formatter.FromUp(tc.Message)
-			if !a.So(err, should.BeNil) {
-				t.FailNow()
+			for _, v := range buf {
+				if !a.So(err, should.BeNil) {
+					t.FailNow()
+				}
+				a.So(string(v), should.Equal, tc.Result)
 			}
-			a.So(string(buf), should.Equal, tc.Result)
```
adriansmares: Instead of ranging, assert that the length of the result is 1, and compare with `tc.Result` only the element at position `0`.

pkg/applicationserver/io/formatters/protobuf.go
```diff
@@ -18,8 +18,9 @@ import "go.thethings.network/lorawan-stack/pkg/ttnpb"
 
 type protobuf struct{}
 
-func (protobuf) FromUp(msg *ttnpb.ApplicationUp) ([]byte, error) {
-	return msg.Marshal()
+func (protobuf) FromUp(msg *ttnpb.ApplicationUp) ([][]byte, error) {
+	m, e := msg.Marshal()
```
adriansmares: Same comment as in the JSON formatter.

pkg/applicationserver/io/formatters/protobuf_test.go
```diff
@@ -83,10 +83,12 @@ func TestProtobufUpstream(t *testing.T) {
...
 			expected, err := tc.Message.Marshal()
-			if !a.So(err, should.BeNil) {
-				t.FailNow()
+			for _, v := range actual {
+				if !a.So(err, should.BeNil) {
+					t.FailNow()
+				}
+				a.So(v, should.Resemble, expected)
 			}
-			a.So(actual, should.Resemble, expected)
```
adriansmares: Same as in the JSON test.

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -161,12 +161,14 @@ func (c *connection) setup(ctx context.Context) error {
 					continue
 				}
 				logger.Debug("Publish upstream message")
-				c.session.Publish(&packet.PublishPacket{
-					TopicName:  topic.Join(topicParts),
-					TopicParts: topicParts,
-					QoS:        qosUpstream,
-					Message:    buf,
-				})
+				for _, v := range buf {
```
adriansmares: Nit: Prefer `msg` and `msgs` instead of `v` and `buf`.

pkg/applicationserver/io/pubsub/pubsub.go
```diff
@@ -149,12 +149,14 @@ func (i *integration) handleUp(ctx context.Context) {
 				logger.WithError(err).Warn("Failed to marshal upstream message")
 				continue
 			}
-			err = topic.Send(ctx, &pubsub.Message{
-				Body: buf,
-			})
-			if err != nil {
-				logger.WithError(err).Warn("Failed to publish upstream message")
-				continue
+			for _, v := range buf {
```
adriansmares: Same nit as in the MQTT frontend.

pkg/applicationserver/io/web/webhooks.go
```diff
@@ -337,24 +337,26 @@ func (w *webhooks) handleUp(ctx context.Context, msg *ttnpb.ApplicationUp) error
...
-				logger.WithError(err).Warn("Failed to create request")
-				return
-			}
-			if req == nil {
-				return
-			}
-			logger.WithField("url", req.URL).Debug("Process message")
-			if err := w.target.Process(req); err != nil {
-				logger.WithError(err).Warn("Failed to process message")
+			for i := range req {
```
adriansmares: Prefer ranged loop with `req` and `reqs` over `i` and `req`.

pkg/applicationserver/io/web/webhooks.go
```diff
@@ -337,24 +337,26 @@ func (w *webhooks) handleUp(ctx context.Context, msg *ttnpb.ApplicationUp) error
...
+				if err != nil {
```
adriansmares: This should be outside of the loop.

pkg/applicationserver/io/web/webhooks.go
```diff
@@ -337,24 +337,26 @@ func (w *webhooks) handleUp(ctx context.Context, msg *ttnpb.ApplicationUp) error
...
+					logger.WithError(err).Warn("Failed to create request")
+					return
+				}
+				if req[i] == nil {
```
adriansmares: This is no longer required, since the looping over an empty slice is no-op.

pkg/applicationserver/io/web/webhooks.go
```diff
@@ -337,24 +337,26 @@ func (w *webhooks) handleUp(ctx context.Context, msg *ttnpb.ApplicationUp) error
...
+				}
 			}
 		}()
 	}
 	wg.Wait()
 	return nil
 }
 
-func (w *webhooks) newRequest(ctx context.Context, msg *ttnpb.ApplicationUp, hook *ttnpb.ApplicationWebhook) (*http.Request, error) {
+func (w *webhooks) newRequest(ctx context.Context, msg *ttnpb.ApplicationUp, hook *ttnpb.ApplicationWebhook) ([]*http.Request, error) {
```
adriansmares: Change to `newRequests` (since it now returns a slice).

pkg/applicationserver/io/web/webhooks.go
```diff
@@ -394,21 +396,25 @@ func (w *webhooks) newRequest(ctx context.Context, msg *ttnpb.ApplicationUp, hoo
...
-		req.Header.Set(key, value)
-	}
-	if hook.DownlinkAPIKey != "" {
-		req.Header.Set(downlinkKeyHeader, hook.DownlinkAPIKey)
-		req.Header.Set(downlinkPushHeader, w.createDownlinkURL(ctx, hook.ApplicationWebhookIdentifiers, msg.EndDeviceIdentifiers, "push"))
-		req.Header.Set(downlinkReplaceHeader, w.createDownlinkURL(ctx, hook.ApplicationWebhookIdentifiers, msg.EndDeviceIdentifiers, "replace"))
+	var requests []*http.Request
+	for i, v := range buf {
+		req, err := http.NewRequest(http.MethodPost, url.String(), bytes.NewReader(v))
+		requests[i] = req
```
adriansmares: This will panic since the slice is empty. Use `requests = append(requests, req)` instead.
Also do the appending at the end of the loop (otherwise you will append values for which errors might exist).

pkg/webui/locales/en.json
```diff
@@ -1,7 +1,5 @@
 {
   "components.data-sheet.index.noData": "No data available",
-  "components.error-notification.story.problem": "We got a problem here!",
```
adriansmares: This file and the one bellow in `pkg/webui` shouldn't get changed.

pkg/applicationserver/io/formatters/json_test.go
mihirr93: Should I use, 
```
if len(buf) == 1{      
}
```
for the length assertion? or is there any alternate approach?

adriansmares: `a.So(buf, should.HaveLength, 1) ` and `a.So(string(buf[0]), should.Equal, tc.Result)`

pkg/applicationserver/io/web/webhooks.go
```diff
@@ -336,25 +336,24 @@ func (w *webhooks) handleUp(ctx context.Context, msg *ttnpb.ApplicationUp) error
...
 				logger.WithError(err).Warn("Failed to create request")
 				return
 			}
-			if req == nil {
-				return
-			}
-			logger.WithField("url", req.URL).Debug("Process message")
-			if err := w.target.Process(req); err != nil {
-				logger.WithError(err).Warn("Failed to process message")
+			for req := range reqs {
```
adriansmares: Prefer `_, req := range reqs` here (same as PubSub).

mihirr93: As we are only using index here, it gives the following error: 
```
should omit 2nd value from range; this loop is equivalent to `for req := range ...`
```

adriansmares: Use `req` as value bellow, as opposed to `reqs[req]`.

mihirr93: Corrected. Looks like this now:
```
for _, req := range reqs {
				logger.WithField("url", req.URL).Debug("Process message")
				if err := w.target.Process(req); err != nil {
					logger.WithError(err).Warn("Failed to process message")
				}
			}
```

johanstokking:
Don't forget the changelog also

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Display of error payloads in console event log.
 - Zero coordinate handling in location form in the Console.
+- Support for multiple resulting messages from one format
```
johanstokking: This should be added to the Unreleased part, as it is not released yet, and not part of 3.5.3. Also add a trailing dot.

PR #2010, opened 2020-02-17
Ignore default config files that cannot be read

### LINKED ISSUE #1997: Fix snap CLI permission errors without using personal-files plugin
#### Summary

`personal-files` snap plugin causes issues with reviewing new releases.

#### What do you see now?

We use `personal-files` to allow CLI to access `~/.ttn-lw-cli.yml` but this causes issues with snap releases.

A different way to avoid permission errors, without using the `personal-files` plugin

#### Environment

3.5.3

...

yes

neoaggelos:
#### Summary

Closes #1997 

#### Changes

- CLI (+ the stack) does not exit with permission errors if any of the default config files cannot be read

#### Notes for Reviewers

- Possible edge case is that if all default paths cannot be read then no error is displayed and the stack simply falls back to the default (localhost) config, without any indication to the user. This could be misleading or unclear

#### Checklist

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -259,9 +259,9 @@ func (m *Manager) inCLIFlags(path string) bool {
 func (m *Manager) ReadInConfig() error {
 	files := m.viper.GetStringSlice(m.configFlag)
 	for _, file := range files {
-		// ignore default config files that do not exist
+		// ignore default config files that are missing or do not have read permissions
 		if m.isDefault(file) && !m.inCLIFlags(file) {
-			if _, err := os.Stat(file); os.IsNotExist(err) {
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -268,7 +268,7 @@ func (m *Manager) ReadInConfig() error {
 
 		m.viper.SetConfigFile(file)
 		err := m.viper.MergeInConfig()
-		if err != nil {
+		if err != nil && m.inCLIFlags(file) {
```
htdvisser: I think it would be better to a check for `os.IsPermission` on L264 (and update the comment above).

If files can't be read for another reason (for instance invalid yaml), we should error here, also if those files are in an implicit/default location

Issue #1422, opened 2019-10-04
Introduce new account app (to replace oauth app)


kschiffer:
#### Summary
We currently have a lot of issues revolving around our OAuth flow, account switching, token management, respective branding, etc. This is an attempt to make all of these more actionable under a new `account` app that will take care of all of these concerns.

#### Why do we need this?
- Missing account switching functionality ⟶ #488
- Missing branding and context for the user ⟶ #730
- Missing session management ⟶ #1217
- User confusion around OAuth authentication flow ⟶ #265 

#### What is already there? What do you see now?
- OAuth app with limited UI

#### What is missing? What do you want to see?
An app that acts as OAuth provider and fully featured and branded configuration UI around all account and authentication concerns

#### Environment
Web

#### How do you propose to implement this?
I think a big problem with the current implementation is the lacking clarity and context for the user. It is hard to recognize the OAuth app as an independent authentication entity. Reasons are:
- lack of branding and context
- very reduced and uninformative UI
- console acting as if it handles authentication itself ("Login" message instead of "Login with your {OAuth provider service name} account", skipping of authorization)
To improve clarity, we need to make the app more visible, distinguishable and purposeful:
- Add configurable branding
- Flesh out functionality
  - Account switching
  - Session management
  - Authorizations management
  - Profile settings (change/forgot password, email, profile picture)

I've created three wireframes that give an idea of how I envision the account app:
<details><summary>See wireframes</summary>

![account_login](https://user-images.githubusercontent.com/5710611/66202268-9e386880-e6a5-11e9-9c53-48fe07486778.png)
![account_overview](https://user-images.githubusercontent.com/5710611/66202274-a09ac280-e6a5-11e9-9c22-03d8885e69ba.png)
![account_profile_settings](https://user-images.githubusercontent.com/5710611/66202278-a2fd1c80-e6a5-11e9-9875-c72a597c1b27.png)
(notice the account switcher integrated in the user dropdown in the top right)
</details>

#### Can you do this yourself and submit a Pull Request?
Yes. @johanstokking @htdvisser, please let me know whether you think its good to continue.

johanstokking:
References https://github.com/TheThingsIndustries/lorawan-stack/issues/1659

kschiffer:
Regarding the whole login/logout confusion issue between console and oauth provider, here's my plan:

Instead of having two separate login screens for console and account app, we should tie the experience closer together, meaning:
- We will remove the console login screen `/console/login`
[...]

PR #2012, opened 2020-02-18
Implement new login/logout flow for the console

kschiffer:
#### Summary
This PR implements the new login/logout flow for the console. This means the console will not have its own login page anymore, but rather automatically redirect to the OAuth login. Respectively, the console logout will now comprise both logging out from the console client, as well as the OAuth provider. This also means that one will always be logged into the console automatically if one is already logged in to the OAuth provider.

References #1422 

#### Changes
- Add a new `oauth-base-url` config to the console, which will be passed to the frontend
- Use the base url to call the logout endpoint of the OAuth provider from within the console client
- Refactor the login view to only render errors that occurred during the login process, or a loading spinner during the redirect
- Enhance the logout handler inside the `<Header />`-component to handle unsuccessful logouts

#### Notes for Reviewers
Main reviewers: @bafonins @htdvisser 
I was experimenting for a while with different implementations and I think this one is thorough enough. Some problems that I see and would like to have some input for:
- There is a chance now for incomplete logouts (eg. when the console logout succeeded and the oauth logout failed) that will result in unsafe app states. The only good way to handle this situation I could think of is to notify the user about the failed attempt and to refresh the page altogether after a short amount of time, so we'll have a proper reinitialization.
- I've read through the [OpenID connect](https://medium.com/@robert.broeckelmann/openid-connect-logout-eccc73df758f) guide that @htdvisser provided and saw that client logouts in OIDC work similar to the login, with a state parameter and an actual redirect to the logout endpoint (instead of AJAX calls), which will redirect back to a callback endpoint on the client side. The advantage I see here is that we can handle errors within the domain that they occur. I'd be OK with implementing it this way but would need some help on the backend for this.

I thought about merging this into an overall branch for #1422 but I think this implementation can already be merged into master and will solve some of the confusion around the console login.

#### Checklist


pkg/webui/console/api/index.js
```diff
@@ -77,6 +82,11 @@ export default {
       return instance.post(`${appRoot}/api/auth/logout`)
     },
   },
+  oauth: {
+    logout() {
+      return instance.post(`${oAuthBaseUrl}/api/auth/logout`)
```
htdvisser: I'm not sure our CORS headers would allow this (and no, we shouldn't change those)

pkg/oauth/server.go
```diff
@@ -261,6 +260,7 @@ func (s *server) RegisterRoutes(server *web.Server) {
 	group.GET("/*", webui.Template.Handler, middleware.CSRF())
 
 	// No CSRF here:
+	group.POST("/api/auth/logout", s.Logout, s.requireLogin)
```
htdvisser: I don't think it's a good idea to remove CSRF protection for this route

kschiffer: Ok then I see two ways to do it:
- Add the OP logout request to the backend logout handler of the console and using only that endpoint in the browser
- Implement a redirect/callback approach, similar to the one described in the article

Which do you prefer?

htdvisser: I'd go for option 2 then. With 1 the Console would need to know about the session ID that the user has on the OAuth server, and that would be a bit smelly.

kschiffer: Ok, can someone help me on the backend side with this? I already have a lot of stuff to do on the frontend as far as the Account app goes. cc @bafonins

pkg/webui/console/store/middleware/logics/user.js
```diff
@@ -15,14 +15,27 @@
...
         await api.console.logout()
+
+        try {
+          // Then logout of the OAuth Provider
+          await api.oauth.logout()
+        } catch (err) {
+          if (!isUnauthenticatedError(err)) {
+            // Unauthenticated errors indicate that the user is already logged
+            // out from the OAuth provider. We can hence ignore this error.
+            throw err
```
bafonins: The comments lets me think that throwing should occur outside of the if block

pkg/webui/console/store/middleware/logics/user.js
```diff
@@ -15,14 +15,27 @@
...
   createRequestLogic({
     type: user.LOGOUT,
     async process() {
       try {
+        // Logout of the Console first
         await api.console.logout()
+
+        try {
+          // Then logout of the OAuth Provider
+          await api.oauth.logout()
```
bafonins: I dont think we should trigger this from the console

pkg/oauth/server.go
bafonins: will try to look into this, however I also have quite some tasks for this and next milestones

pkg/webui/console/store/middleware/logics/user.js
kschiffer: What do you suggest then?

kschiffer:
Closing this for the time being until we are clearer about how to move forward with this issue.

PR #1976, opened 2020-02-11
Add support for LoRaWAN 1.0.4

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/1880

#### Changes

- Require `DevNonce` to be an incrementing counter
- Require `DevEUI` to be set for ABP devices
- Drop uplinks exceeding data rate payload length limits
- Fix terminology used in matching
- Determine retransmissions by `PHYPayload`, not `MHDR+MACPayload`
- Ignore PingSlotInfoReq from device devices not in class A mode
- Avoid setting FPending on class C downlinks - has no meaning in all specs except for 1.0.4, explicitly prohibited in 1.0.4
- Use no-change TXPower/DRIdx in LinkADR if possible and necessary in 1.0.4+

#### Notes to reviewers

@bafonins feel free to review the changes in `pkg/webui`, but they're trivial enough that probably do not need review (https://github.com/TheThingsNetwork/lorawan-stack/pull/1976/commits/c00d5beb95bce43c68dbd8e44c6e1c47a9eb93d3).

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, database and configuration, according to the stability commitments in `README.md`.
- [x] Testing: The changes are covered with unit tests. The changes are tested manually as well.
- [x] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #2015, opened 2020-02-19
Fix LinkADR ChMask handling

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2016

#### Changes

- Simplify `ChMask` parsing and generation logic
- Always compute minimal, most generic `ChMask` sequence
- Fix `ChMaskCntl==5` endianness
- Add US915 OTAA 1.0.3 flow test
- Fix `MaxTxPowerIndex` in various bands/version combinations
- Assume `MaxTxPowerIndex` is the default `TxPowerIndex` as per LoRaWAN spec.

#### Notes for Reviewers

@adriansmares may want to take a look since you did https://github.com/TheThingsIndustries/lorawan-stack/pull/1401

#### Checklist

### DIFF
--- a/pkg/band/band.go
+++ b/pkg/band/band.go
@@ -16,6 +16,7 @@
 package band
 
 import (
+	"fmt"
 	"math"
 	"strings"
 	"time"
@@ -182,10 +183,13 @@ type Band struct {
 	Rx1DataRate func(ttnpb.DataRateIndex, uint32, bool) (ttnpb.DataRateIndex, error)
 
 	// GenerateChMasks generates a mapping ChMaskCntl -> ChMask.
-	// Length of chs must be equal to the maximum number of channels defined for the particular band.
-	// Meaning of chs is as follows: for i in range 0..len(chs) if chs[i] == true,
+	// Length of desiredChs must be equal to length of currentChs.
+	// Meaning of desiredChs is as follows: for i in range 0..len(desiredChs) if desiredChs[i] == true,
 	// then channel with index i should be enabled, otherwise it should be disabled.
-	GenerateChMasks func(chs []bool) ([]ChMaskCntlPair, error)
+	// Meaning of currentChs is as follows: for i in range 0..len(currentChs) if currentChs[i] == true,
+	// then channel with index i is enabled, otherwise it is disabled.
+	// In case desiredChs equals currentChs, GenerateChMasks returns a singleton, which repesents a noop.
+	GenerateChMasks func(currentChs, desiredChs []bool) ([]ChMaskCntlPair, error)
 	// ParseChMask computes the channels that have to be masked given ChMask mask and ChMaskCntl cntl.
 	ParseChMask func(mask [16]bool, cntl uint8) (map[uint8]bool, error)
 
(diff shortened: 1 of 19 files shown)


rvolosatovs:
Blocking #1976

pkg/band/band.go
```diff
@@ -300,248 +304,368 @@ var usAuBeaconFrequencies = func() (freqs [8]uint64) {
...
-		}
+		return parseChMask(0, mask[:]...), nil
 	case 6:
-		for i := uint8(0); i < 16; i++ {
-			chans[i] = true
-		}
-	default:
-		return nil, errUnsupportedChMaskCntl.WithAttributes("chmaskcntl", cntl)
+		return parseChMask(0,
+			true, true, true, true, true, true, true, true,
```
adriansmares: Can we introduce a small `repeat(count, value)` function ? I find it error prone to write the mask by hand. You can even create the result in `init` and reuse it from there, just don't put 16 `true`s one after another.

pkg/band/band.go
```diff
@@ -300,248 +304,368 @@ var usAuBeaconFrequencies = func() (freqs [8]uint64) {
...
+			mask[2], mask[2], mask[2], mask[2], mask[2], mask[2], mask[2], mask[2],
+			mask[3], mask[3], mask[3], mask[3], mask[3], mask[3], mask[3], mask[3],
+			mask[4], mask[4], mask[4], mask[4], mask[4], mask[4], mask[4], mask[4],
+			mask[5], mask[5], mask[5], mask[5], mask[5], mask[5], mask[5], mask[5],
+			mask[6], mask[6], mask[6], mask[6], mask[6], mask[6], mask[6], mask[6],
+			mask[7], mask[7], mask[7], mask[7], mask[7], mask[7], mask[7], mask[7],
+		), nil
+	case 6:
+		return parseChMask(0,
+			false, false, false, false, false, false, false, false,
```
adriansmares: Same as the comment before regarding generating this mask.

pkg/band/band.go
```diff
@@ -300,248 +304,368 @@ var usAuBeaconFrequencies = func() (freqs [8]uint64) {
...
+			false, false, false, false, false, false, false, false,
+			false, false, false, false, false, false, false, false,
+			false, false, false, false, false, false, false, false,
+			false, false, false, false, false, false, false, false,
+			mask[0], mask[1], mask[2], mask[3], mask[4], mask[5], mask[6], mask[7],
+		), nil
+	case 7:
+		return parseChMask(0,
+			true, true, true, true, true, true, true, true,
```
adriansmares: Same here.

pkg/band/band.go
```diff
@@ -300,248 +304,368 @@ var usAuBeaconFrequencies = func() (freqs [8]uint64) {
...
-		for i := uint8(0); i < 64; i++ {
-			chans[i] = cntl == 6
-		}
-		for i := uint8(64); i < 72; i++ {
-			chans[i] = mask[i-64]
-		}
-	default:
-		return nil, errUnsupportedChMaskCntl.WithAttributes("chmaskcntl", cntl)
+		return parseChMask(0,
+			mask[0], mask[0], mask[0], mask[0], mask[0], mask[0], mask[0], mask[0],
```
adriansmares: Abstract this away using `repeat`.

pkg/band/band.go
```diff
@@ -300,248 +304,368 @@ var usAuBeaconFrequencies = func() (freqs [8]uint64) {
...
-		}
+		return parseChMask(cntl*16, mask[:]...), nil
 	case 6:
-		for i := uint8(0); i < 96; i++ {
-			chans[i] = true
-		}
-	default:
-		return nil, errUnsupportedChMaskCntl.WithAttributes("chmaskcntl", cntl)
+		return parseChMask(0,
+			true, true, true, true, true, true, true, true,
```
adriansmares: Same here.

pkg/band/band.go
```diff
@@ -300,248 +304,368 @@ var usAuBeaconFrequencies = func() (freqs [8]uint64) {
...
 		}
+	}
+	if len(pairs) <= 2 {
+		return pairs, nil
+	}
+	// Count amount of pairs required assuming either ChMaskCntl==6 or ChMaskCntl==7 is sent first.
+	// The minimum amount of pairs required in such case will be 2, hence only attempt this if amount
+	// of generated pairs so far is higher than 2.
+	cntl6Pairs, err := generateChMaskMatrix(make([]ChMaskCntlPair, 0, 4), []bool{
+		true, true, true, true, true, true, true, true,
```
adriansmares: Same here.

pkg/band/band.go
```diff
@@ -300,248 +304,368 @@ var usAuBeaconFrequencies = func() (freqs [8]uint64) {
...
+		true, true, true, true, true, true, true, true,
+		true, true, true, true, true, true, true, true,
+		true, true, true, true, true, true, true, true,
+	}, desiredChs[0:64])
+	if err != nil {
+		return nil, err
+	}
+	cntl7Pairs, err := generateChMaskMatrix(make([]ChMaskCntlPair, 0, 4), []bool{
+		false, false, false, false, false, false, false, false,
```
adriansmares: Same here.

pkg/band/band.go
```diff
@@ -300,248 +304,368 @@ var usAuBeaconFrequencies = func() (freqs [8]uint64) {
...
-			}
+		if n := trueCount(fsbs[:]...); n == 0 || n == 8 {
+			// Since there are either no enabled FSBs, or no disabled FSBs we won't be able to compute a
+			// more efficient result that one using ChMaskCntl==6 or ChMaskCntl==7.
+			return pairs, nil
 		}
-		if !equalChMasks(majorityMask[:], mask[64:72]) {
-			outliers = append(outliers, 4)
+		cntl5Pairs, err := generateChMaskMatrix(make([]ChMaskCntlPair, 0, 5), []bool{
+			fsbs[0], fsbs[0], fsbs[0], fsbs[0], fsbs[0], fsbs[0], fsbs[0], fsbs[0],
```
adriansmares: Same here.

pkg/band/band.go
```diff
@@ -300,248 +304,368 @@ var usAuBeaconFrequencies = func() (freqs [8]uint64) {
...
+		return nil, err
+	}
+	if len(pairs) <= 2 {
+		return pairs, nil
+	}
+	// Count amount of pairs required assuming ChMaskCntl==6 is sent first.
+	// The minimum amount of pairs required in such case will be 2, hence only attempt this if amount
+	// of generated pairs so far is higher than 2.
+	cntl6Pairs, err := generateChMaskMatrix(make([]ChMaskCntlPair, 0, 6), []bool{
+		true, true, true, true, true, true, true, true,
```
adriansmares: Same here.

pkg/band/band.go
rvolosatovs: I took this approach, because it's simpler and it's obvious what's going on. I'd argue that constructing these by using utility functions is actually more error-prone, since then you have to merge stuff.

pkg/band/band.go
rvolosatovs: No. This would require 7 chained appends, which would only obfuscate behavior.

pkg/band/band.go
rvolosatovs: Same answer as the answer before regarding generating this mask.

rvolosatovs:
Regarding repeating `bool`s:
I agree that in some cases it would be cleaner to use such utility, however it's not true for all the cases and IMO consistency is key here, so we should either do it everywhere or not do it at all. Such utility usage would require chaining appends in many cases, which is anything but readable.
I'm open to adding comments to index rows.
Each row contains 8 values - 1 FSB.

adriansmares:
I can agree that multiple appends are definitely obfuscating the code. I was living under the impression that append would allow expansion akin to `append(original, exp1..., exp2...)`  which would look 'better' on multiple lines.

I went back to the matrices that I've marked in the original review and checked the indices - they are fine.

I'd like an US915 checkup before merging this, but other than that I'm fine with the current status.

rvolosatovs:
> I'd like an US915 checkup before merging this

What do you mean by "US915 checkup" ?

pkg/networkserver/utils.go
```diff
@@ -501,7 +501,7 @@ func newMACState(dev *ttnpb.EndDevice, fps *frequencyplans.Store, defaults ttnpb
 	macState.CurrentParameters.ADRDataRateIndex = ttnpb.DATA_RATE_0
 	macState.DesiredParameters.ADRDataRateIndex = macState.CurrentParameters.ADRDataRateIndex
 
-	macState.CurrentParameters.ADRTxPowerIndex = 0
+	macState.CurrentParameters.ADRTxPowerIndex = uint32(phy.MaxTxPowerIndex)
```
htdvisser: > Assume `MaxTxPowerIndex` is the default `TxPowerIndex` as per LoRaWAN spec.

Where does the spec say this? As far as I remember, "vanilla" end devices start at `TxPowerIndex` 0 (the `TxPowerIndex` that corresponds to the maximum `TxPower`).

rvolosatovs: I was referring to this paragraph in 1.0.4 spec: [1.0.4](https://imgur.com/jJeN5oJ).
Maximum transmission power is `TxPowerIndex==0` however indeed.

PR #2023, opened 2020-02-19
Use default LNS port for BasicStation CUPS Update Info request

KrishnaIyer:
#### Summary

Use default LNS port for BasicStation CUPS Update Info request.

In most of our "standard" deployments, most users choose the pre-filled GatewayServerAddress when creating a gateway. This does not have any port info so Basic Station gateways complain. 

#### Changes

- Use port 8887 in case no port is provided.

#### Notes for Reviewers

Users can just append the port in the console/cli but we should support a default port as we already support a default protocol (`wss`).
Users should only have to edit the port if they choose to customize.

#### Checklist

### DIFF
--- a/pkg/basicstation/cups/server.go
+++ b/pkg/basicstation/cups/server.go
@@ -255,7 +255,7 @@ var errNoTrust = errors.DefineInternal("no_trust", "no trusted certificate found
 // It supports the typical format "host:port" (port being optional).
 // It allows schemes "http://host:port" to be present.
 // If schemes http/https/ws/wss are used, the port is inferred if not present.
-func parseAddress(address string) (scheme, host, port string, err error) {
+func parseAddress(defaultScheme, address string) (scheme, host, port string, err error) {
 	if address == "" {
 		return
 	}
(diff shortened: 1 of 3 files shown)


adriansmares:
I think it would be better to add a new `Option` that allows the Gateway Server to propagate the listener port directly to the CUPS frontend.

KrishnaIyer:
Yeah I thought about that but when the GS is behind a proxy, it doesn't know what the front-facing port is. We already solved that by setting the `GatewayServerAddress` to `mydomain:port` on the Gateway object.

The reason why I added this is;
* User creates a Gateway on Console/CLI using all defaults.
* User creates an API key.
* User connects gateway.
* Gateway shows error saying that the port is missing.

So this is just a sane default value that works on 90% of deployments. For custom ones, the user can always set the port manually.

pkg/basicstation/cups/server.go
```diff
@@ -293,13 +300,10 @@ func (s *Server) getTrust(address string) (*x509.Certificate, error) {
 		}
 		return nil, errNoTrust
 	}
-	_, host, port, err := parseAddress(address)
+	_, host, port, err := parseAddress("https:", address)
```
htdvisser: I think you meant `https` without the colon

KrishnaIyer: Yup. Good catch.

PR #2025, opened 2020-02-20
Fix payload size limits used by Network Server

rvolosatovs:
#### Summary

Fix payload size limits used by Network Server

#### Changes

- Account for 1 byte of CID in MAC answers when generating downlink
- Log the maximum downlink length and application downlink length on length limit errors
- Refactor data rates in PHY to a `map[ttnpb.DataRateIndex]band.DataRate`
- Validate queued application downlink length according to PHY and frequency plan limitations
- Be explicit about payload length limits in PHY being MAC payload lengths and treat them as such in NS.
- De-obfuscate data rate definitions.
- Fix `band.Band` structs not being referred to as `phy`

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/simulate.go
+++ b/cmd/ttn-lw-cli/commands/simulate.go
@@ -83,12 +83,16 @@ func (m *simulateMetadataParams) setDefaults() error {
 		}
 	}
 	if m.Bandwidth == 0 || m.SpreadingFactor == 0 {
-		drIdx := int(m.DataRateIndex)
-		if drIdx < int(phy.UplinkChannels[0].MinDataRate) || drIdx > int(phy.UplinkChannels[0].MaxDataRate) {
-			drIdx = int(phy.UplinkChannels[0].MaxDataRate)
+		drIdx := ttnpb.DataRateIndex(m.DataRateIndex)
+		if drIdx < phy.UplinkChannels[0].MinDataRate || drIdx > phy.UplinkChannels[0].MaxDataRate {
+			drIdx = phy.UplinkChannels[0].MaxDataRate
 		}
-		dr := phy.DataRates[drIdx].Rate.GetLoRa()
-		m.SpreadingFactor, m.Bandwidth = dr.SpreadingFactor, dr.Bandwidth
+		dr, ok := phy.DataRates[drIdx]
+		if !ok {
+			return errInvalidDataRateIndex
+		}
+		lora := dr.Rate.GetLoRa()
+		m.SpreadingFactor, m.Bandwidth = lora.SpreadingFactor, lora.Bandwidth
 	} else if m.DataRateIndex == 0 {
 		for i, dr := range phy.DataRates {
 			if dr.Rate.GetLoRa().SpreadingFactor == m.SpreadingFactor && dr.Rate.GetLoRa().Bandwidth == m.Bandwidth {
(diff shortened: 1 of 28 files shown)


cmd/ttn-lw-cli/commands/simulate.go
```diff
@@ -83,12 +84,16 @@ func (m *simulateMetadataParams) setDefaults() error {
...
-			drIdx = int(phy.UplinkChannels[0].MaxDataRate)
+		drIdx := ttnpb.DataRateIndex(m.DataRateIndex)
+		if drIdx < phy.UplinkChannels[0].MinDataRate || drIdx > phy.UplinkChannels[0].MaxDataRate {
+			drIdx = phy.UplinkChannels[0].MaxDataRate
 		}
-		dr := phy.DataRates[drIdx].Rate.GetLoRa()
-		m.SpreadingFactor, m.Bandwidth = dr.SpreadingFactor, dr.Bandwidth
+		dr, ok := phy.DataRates[drIdx]
+		if !ok {
+			panic(fmt.Sprintf("data rate '%d' is not defined by PHY", drIdx))
```
htdvisser: Let's not panic on bad user input.

pkg/frequencyplans/frequencyplans.go
```diff
@@ -198,16 +199,20 @@ func (lsc *LoRaStandardChannel) Clone() *LoRaStandardChannel {
...
 // ToConcentratorConfig returns the LoRa standard channel configuration in the protobuf format.
-func (lsc *LoRaStandardChannel) ToConcentratorConfig(band band.Band) *ttnpb.ConcentratorConfig_LoRaStandardChannel {
+func (lsc *LoRaStandardChannel) ToConcentratorConfig(phy band.Band) *ttnpb.ConcentratorConfig_LoRaStandardChannel {
 	if lsc == nil {
 		return nil
 	}
-	dr := band.DataRates[lsc.DataRate].Rate.GetLoRa()
+	dr, ok := phy.DataRates[ttnpb.DataRateIndex(lsc.DataRate)]
+	if !ok {
+		panic(fmt.Sprintf("data rate '%d' does not exist", lsc.DataRate))
```
htdvisser: Prefer returning err instead of panicking

pkg/gatewayserver/io/basicstationlns/messages/util.go
```diff
@@ -56,35 +57,35 @@ func getFCtrlAsUint(fCtrl ttnpb.FCtrl) uint {
 }
 
 func getDataRateFromIndex(bandID string, index int) (ttnpb.DataRate, bool, error) {
-	band, err := band.GetByID(bandID)
+	phy, err := band.GetByID(bandID)
 	if err != nil {
 		return ttnpb.DataRate{}, false, errDataRateIndex.WithCause(err)
 	}
-	if index >= len(band.DataRates) {
+	if index < 0 || index > math.MaxInt32 {
```
htdvisser: Maybe use `len(phy.DataRates)` instead of `math.MaxInt32`. Most people wouldn't understand why you're using `math.MaxInt32`.

rvolosatovs: I can add a comment.
`len(phy.DataRates)` does not work, because length of `n` does not mean that data rate index `n-1` is valid.

pkg/frequencyplans/frequencyplans.go
rvolosatovs: The original code simply used the data rate directly without checking anything, so I assumed that this is expected to always exist and panicking is arguably better than silently using undefined value.
I will try to adapt this to use errors.

PR #2032, opened 2020-02-21
Migrate to new Sentry SDK

### LINKED ISSUE #2026: Migrate from raven-go to sentry-go
#### Summary

We should migrate our `raven-go` to `sentry-go`.

See https://github.com/getsentry/sentry-go/blob/master/MIGRATION.md

Sure, but if anyone else wants to take this, let me know!

htdvisser:
#### Summary

This PR migrates to the new Sentry SDK and improves how errors are reported to Sentry.

Closes #2026


go.mod
```diff
@@ -89,7 +87,7 @@ require (
 	github.com/kamilsk/retry/v4 v4.4.1 // indirect
 	github.com/kr/pretty v0.2.0
 	// Do not upgrade Echo beyond v4.1.2 - see https://github.com/TheThingsNetwork/lorawan-stack/issues/977 .
```
adriansmares: Can remove the warning.

htdvisser:
Echo isn't actually upgraded (we have a `replace` directive).
The change is caused by the Sentry SDK which has the newer version of Echo in their `go.mod`. In any case, I removed the comment, which would otherwise just be confusing

Issue #2039, opened 2020-02-24
Users w/o RIGHT_APPLICATION_DEVICES_{READ|WRITE}_KEYS right cannot successfully create/edit end devices


bafonins:
#### Summary

The users without the `RIGHT_APPLICATION_DEVICES_WRITE` right cant successfully create/update end devices in the Console

#### Steps to Reproduce

1. Create a new user and add to an application as a collaborator without the `RIGHT_APPLICATION_DEVICES_WRITE_KEYS` right on that application
2. Login as the new user and try registering a new end device (otaa/abp doesnt really matter here)
3. See the end device not being created in all stack components, form error (similar applies to the edit procedure)

#### What do you see now?

Partial end device creation, no UI is hidden based on user rights

#### What do you want to see instead?

No errors, if the user has the `RIGHT_APPLICATION_DEVICES_WRITE` right it should be possible to create/edit end devices, but not the keys if the `RIGHT_APPLICATION_DEVICES_WRITE_KEYS` right is missing

#### Environment

v3.5.3

#### How do you propose to implement this?

Main problems here:

1. UI is not hidden from the user without appropriate rights
2. The js sdk does not rollback failed device creation https://github.com/TheThingsNetwork/lorawan-stack/pull/2041
3. The console does not consider user rigths when generating diffs/keys through the validation schema

#### Can you do this yourself and submit a Pull Request?

yes

bafonins:
Regarding creating devices without the `RIGHT_APPLICATION_DEVICES_WRITE_KEYS` right. The problem here is that it is not possible to create an ABP device (more specifically the entry in NS) without session keys, NS will refuse such request; the same applies to multicast devices. On the other hand, it is possible to create OTAA devices without root keys (consider the case when `root_keys` are provisioned on an external join server, while the cluster JS still has the end device stored).

That being said, shall we allow creating devices for users that do not have the right to edit device keys?

We can:
1. Require both `RIGHT_APPLICATION_DEVICES_WRITE_KEYS` and `RIGHT_APPLICATION_DEVICES_WRITE` to create devices via the Console.
2. Imply `RIGHT_APPLICATION_DEVICES_WRITE_KEYS` if the user has `RIGHT_APPLICATION_DEVICES_WRITE`. This, however, breaks the idea that every right is **explicit** in IS.
3. Disable creating ABP/Multicast devices via the Console and allow creating only OTAA devices.

cc @htdvisser @johanstokking

[...]

PR #2048, opened 2020-02-25
Fix device general settings for users w/o RIGHT_APPLICATION_DEVICES_{READ|WRITE}_KEYS rights

bafonins:
#### Summary

This PR adjusts the device general settings page to check for `RIGHT_APPLICATION_DEVICES_{READ|WRITE}_KEYS` rights to adjust the UI and behavior accordingly avoiding unnecessary faulty requests, incosistent device state, etc.

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2039

<details>
ABP device view for the user with only `RIGHT_APPLICATION_DEVICES_WRITE_KEYS`
Screenshot 2020-03-04 at 12 59 46

OTAA device view for the user with only `RIGHT_APPLICATION_DEVICES_WRITE_KEYS`
Screenshot 2020-03-04 at 13 00 03

MULTICAST device view for the user with only `RIGHT_APPLICATION_DEVICES_WRITE_KEYS`
Screenshot 2020-03-04 at 13 01 44

ABP device view for the user with only `RIGHT_APPLICATION_DEVICES_READ_KEYS`
Screenshot 2020-03-04 at 13 04 02

OTAA device view for the user with only `RIGHT_APPLICATION_DEVICES_READ_KEYS `
Screenshot 2020-03-04 at 13 04 17

MULTICAST device view for the user with only `RIGHT_APPLICATION_DEVICES_WRITE_KEYS`
Screenshot 2020-03-04 at 13 04 35

</details>

#### Changes

- Do not generate `root_keys` or `session` keys if the user has no appropriate rigths
- Hide key fields if the user has no appropriate rights
- Extend application feature checks

#### Notes for Reviewers

Quick summary:
1. Show notification if the user has only the write right to inform that the keys might be there.
2. Disable key inputs if the user has no write rights
3. Force the users to explicitly generate root/session keys instead of doing this automatically on empty fields
5. Do not fetch `root_keys` and `session` fields if the user has no rights to read device keys to avoid `403` errors, since the user should able to see other fields from js/ns/as.

You can use [this script](https://gist.github.com/htdvisser/2d146748d01e56feea480b322bb0c400) to generate a bunch of devices for tests.

#### Checklist

### DIFF
--- a/pkg/webui/console/views/device-general-settings/application-server-form/index.js
+++ b/pkg/webui/console/views/device-general-settings/application-server-form/index.js
@@ -29,16 +29,12 @@ import PropTypes from '../../../../lib/prop-types'
 import sharedMessages from '../../../../lib/shared-messages'
 
 const random16BytesString = () => randomByteString(32)
-const toUndefined = value => (!Boolean(value) ? undefined : value)
 
 const validationSchema = Yup.object().shape({
   session: Yup.object().shape({
     keys: Yup.object().shape({
       app_s_key: Yup.object().shape({
-        key: Yup.string()
-          .emptyOrLength(16 * 2, m.validate32) // 16 Byte hex
-          .transform(toUndefined)
-          .default(random16BytesString),
+        key: Yup.string().emptyOrLength(16 * 2, m.validate32), // 16 Byte hex
       }),
     }),
   }),
@@ -53,7 +49,9 @@ const ApplicationServerForm = React.memo(props => {
     const { session = {} } = device
     const {
       keys = {
-        app_s_key: {},
+        app_s_key: {
+          key: '',
+        },
       },
     } = session
 
@@ -104,9 +102,10 @@ const ApplicationServerForm = React.memo(props => {
         type="byte"
         min={16}
         max={16}
-        placeholder={m.leaveBlankPlaceholder}
         description={m.appSKeyDescription}
-        component={Input}
+        component={Input.Generate}
+        mayGenerateValue={mayEditKeys}
+        onGenerateValue={random16BytesString}
       />
       <SubmitBar>
         <Form.Submit component={SubmitButton} message={sharedMessages.saveChanges} />
--- a/pkg/webui/console/views/device-general-settings/application-server-form/index.js
+++ b/pkg/webui/console/views/device-general-settings/application-server-form/index.js
@@ -19,14 +19,17 @@ import SubmitButton from '../../../../components/submit-button'
 import SubmitBar from '../../../../components/submit-bar'
 import Input from '../../../../components/input'
 import Form from '../../../../components/form'
+import Notification from '../../../../components/notification'
 
 import diff from '../../../../lib/diff'
 import m from '../../../components/device-data-form/messages'
+import messages from '../messages'
 import randomByteString from '../../../lib/random-bytes'
 import PropTypes from '../../../../lib/prop-types'
 import sharedMessages from '../../../../lib/shared-messages'
 
-const random16BytesString = () => randomByteString(32) const toUndefined = value => (!Boolean(value) ? undefined : value)
+const random16BytesString = () => randomByteString(32)
+const toUndefined = value => (!Boolean(value) ? undefined : value)
 
 const validationSchema = Yup.object().shape({
   session: Yup.object().shape({
@@ -42,12 +45,12 @@ const validationSchema = Yup.object().shape({
 })
 
 const ApplicationServerForm = React.memo(props => {
-  const { device, onSubmit, onSubmitSuccess } = props
+  const { device, onSubmit, onSubmitSuccess, mayEditKeys, mayReadKeys } = props
 
   const [error, setError] = React.useState('')
 
   const initialValues = React.useMemo(() => {
-    const { session = {}, } = device
+    const { session = {} } = device
     const {
       keys = {
         app_s_key: {},
@@ -81,6 +84,10 @@ const ApplicationServerForm = React.memo(props => {
     [initialValues, onSubmit, onSubmitSuccess],
   )
 
+  // Notify the user that the session keys might be there, but since there are no rights
+  // to read the keys we cannot display them.
+  const showResetNotification = !mayReadKeys && mayEditKeys && !Boolean(device.session)
+
   return (
     <Form
       validationSchema={validationSchema}
@@ -88,7 +95,9 @@ const ApplicationServerForm = React.memo(props => {
       onSubmit={onFormSubmit}
       error={error}
       enableReinitialize
+      disabled={!mayEditKeys}
     >
+      {showResetNotification && <Notification content={messages.keysResetWarning} info small />}
       <Form.Field
         title={sharedMessages.appSKey}
         name="session.keys.app_s_key.key"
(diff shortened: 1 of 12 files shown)


pkg/webui/console/views/device-general-settings/index.js
```diff
@@ -187,9 +200,12 @@ export default class DeviceGeneralSettings extends React.Component {
     // 2. Disable the section if the device is OTAA and joined since no fields are stored in the AS.
     // 3. Disable the section if NS is not in cluster, since activation mode is unknown.
     // 4. Disable the seciton if `application_server_address` is not equal to the cluster AS address.
+    // 5. Disable the section if the user has no appropriate rights to edit keys.
     const sameAsAddress = getComponentBaseUrl(asConfig) === device.application_server_address
     const isJoined = isDeviceJoined(device)
-    const asDisabled = !asEnabled || (isOTAA && !isJoined) || !nsEnabled || !sameAsAddress
+    const showKeys = mayReadKeys && mayEditKeys
```
htdvisser: - If a user is allowed to read keys, but not write them, we can show the keys as read-only
- If a user is allowed to write keys, but not read them, we can still let them reset the keys

pkg/webui/console/views/device-general-settings/index.js
```diff
@@ -187,9 +200,12 @@ export default class DeviceGeneralSettings extends React.Component {
...
+    const asDisabled =
```
htdvisser: I think it's confusing if we disable the entire AS section only because someone doesn't have access to keys.

pkg/webui/console/views/device-general-settings/index.js
```diff
@@ -207,7 +223,7 @@ export default class DeviceGeneralSettings extends React.Component {
     // 3. Disable the seciton if `join_server_address` is not equal to the cluster JS address.
     // 4. Disable the seciton if an external JS is used.
     const sameJsAddress = getComponentBaseUrl(jsConfig) === device.join_server_address
-    const externalJs = hasExternalJs(device)
+    const externalJs = hasExternalJs(device) && mayReadKeys
     const jsDisabled = !jsEnabled || !isOTAA || !sameJsAddress || externalJs
```
htdvisser: I think it's confusing if we disable the entire JS section only because someone isn't allowed to read keys.

pkg/webui/console/views/device-general-settings/index.js
bafonins: > If a user is allowed to read keys, but not write them, we can show the keys as read-only

The users can see the keys in the device overview page if they have the read right. the only reason we show the `session.dev_addr` field and disable it in the AS section is for cases when NS not in cluster. however, now it seems redundant

> If a user is allowed to write keys, but not read them, we can still let them reset the keys

Shall we inform the user that the key might be there it just not visible because of insufficient rights? Asking because imo an empty field can make the user think that the key (`session.keys.app_s_key.key`) is not set, while him/her just not having the rights to view the key. Note, that `app_s_key` is not required by AS

pkg/webui/console/views/device-general-settings/index.js
bafonins: Well, there are no fields other than the keys that require the `RIGHT_APPLICATION_DEVICES_WRITE_KEYS` right.

htdvisser: Not right now, but in the future I can imagine we'll add other AS fields, perhaps the `skip_payload_crypto` field

pkg/webui/console/views/device-general-settings/index.js
bafonins: > ...we could render an empty input together with a warning that values cannot be viewed and new values potentially overwrite existing ones.

I dont think we should not do this. Such a combination of rights (no read, but write) is applicable mostly to API keys, not users. Hence I dont think we should invest time and make the implementation more complex to support such edge case. I would like to hear from @htdvisser and @johanstokking regarding the notification.

The main purpose of this PR is to avoid errors when submitting without the rights as well as still fetching js/ns fields without the keys if the user is not allowed to fetch them.

Other implementation specific questions if we decide to add the notification and handle the case when the user is able to edit but not read the keys:
1. Do we still generate the keys automatically if the user submits the form with empty key fields?
2. What is the message we want to show in the notification? Asking because any wording I can come up with is very stack/implementation specific and I guess we dont want to bother the users with this.

johanstokking: > Hence I dont think we should invest time and make the implementation more complex to support such edge case.

I agree.

> I would like to hear from @htdvisser and @johanstokking regarding the notification.

>What is the message we want to show in the notification? 

Which notification exactly?

> 1. Do we still generate the keys automatically if the user submits the form with empty key fields?

No, this should be an explicit action/checkbox

htdvisser: > Such a combination of rights (no read, but write) is applicable mostly to API keys, not users

I wouldn't assume that. Maybe I want to allow an entire org to create devices with keys, but don't read them after create.

> Do we still generate the keys automatically if the user submits the form with empty key fields?

I've always been against this. I think we just need a 🔁 button to generate keys.

> What is the message we want to show in the notification?

Perhaps something like "You can not read this value, but you can overwrite it if you want"

pkg/webui/console/views/device-general-settings/messages.js
```diff
@@ -34,6 +34,7 @@ const messages = defineMessages({
   activationModeUnknown: 'Activation mode unknown because Network Server is not available',
   notInCluster: 'Not registered in this cluster',
   updateSuccess: 'The end device has been updated successfully',
+  keysResetWarning: 'You can not read the end device keys, but you can overwrite it if you want',
```
kschiffer: The tone is too casual. I suggest
```suggestion
  keysResetWarning: "You're not authorized to view the end device keys. However, overwriting is allowed.",
```

pkg/webui/console/views/device-general-settings/application-server-form/index.js
```diff
@@ -87,31 +82,30 @@ const ApplicationServerForm = React.memo(props => {
     [initialValues, onSubmit, onSubmitSuccess],
   )
 
+  // Motify the user that the session keys might be there, but since there are no rights
```
kschiffer:
```suggestion
  // Notify the user that the session keys might be there, but since there are no rights
```

pkg/webui/console/views/device-general-settings/join-server-form/index.js
```diff
@@ -88,20 +100,24 @@ const JoinServerForm = React.memo(props => {
...
 
-  let nwkKeyPlaceholder = m.leaveBlankPlaceholder
+  let nwkKeyPlaceholder
   if (externalJs) {
     nwkKeyPlaceholder = sharedMessages.provisionedOnExternalJoinServer
   } else if (nwkKeyHidden) {
     nwkKeyPlaceholder = m.unexposed
   }
 
+  // Motify the user that the root keys might be there, but since there are no rights
```
kschiffer:
```suggestion
  // Notify the user that the root keys might be there, but since there are no rights
```

pkg/webui/console/views/device-general-settings/network-server-form/index.js
```diff
@@ -186,6 +177,10 @@ const NetworkServerForm = React.memo(props => {
     [initialValues],
   )
 
+  // Motify the user that the session keys might be there, but since there are no rights
```
```suggestion
  // Notify the user that the session keys might be there, but since there are no rights
```

PR #2052, opened 2020-02-25
Per-session application downlink queue in Network Server

rvolosatovs:
#### Summary

Closes NS part of https://github.com/TheThingsNetwork/lorawan-stack/issues/1954

#### Changes

- Add `{,pending_}session.queued_application_downlinks`
- Use `session.queued_application_downlinks` in NS for scheduling
- Match downlinks to correct session on push/replace
- List downlinks in following order: `<active session downlinks>:<pending session downlinks>`
- Do not drop the session queue on Join-request

#### Notes for Reviewers

Will mark as ready to review once the dependencies (#2053 #2054 #2055 ) are merged. I will meanwhile do some integration testing as well.
@adriansmares you can already cherry-pick https://github.com/TheThingsNetwork/lorawan-stack/commit/e55290a9afb22271def9583379b2fa3b165f8795 https://github.com/TheThingsNetwork/lorawan-stack/commit/642ae82210ec02b590f3bcc3697b1f70e20e5a80 and proceed with AS implementation

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, database and configuration, according to the stability commitments in `README.md`.
- [x] Testing: The changes are covered with unit tests. The changes are tested manually as well.
- [x] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

Issue #1954, opened 2020-02-05
Store downlink queue per session


johanstokking:
#### Summary

Store downlink queue per session

#### Why do we need this?

To avoid hitting AS to recalculate the downlink queue when the end device confirms the session. Having a downlink queue of both the current session and the pending session in NS allows for downlink items to be available immediately.

#### What is already there? What do you see now?

Current flow is that on device activation, the new session is stored as pending session in AS. Newly scheduled downlink items are, however, encrypted with the current session. When the end device sends the first uplink message in the session, AS recalculates the downlink queue, but this is not fast enough to make the RX window.

#### What is missing? What do you want to see?

Store the downlink queue in the end device session. This changes the protos, but it should not change the API; the downlink queue operations towards AS should remain the same.

#### How do you propose to implement this?

Add a downlink queue to `Session`, so it's available in the end device's `session` and `pending_session`. You can use the existing downlink queue in `EndDevice` as the current session's downlink queue.

AS should change behavior; it should schedule downlink items for both the current and the pending session, and NS should store the items in the concerning session.

Whenever a device confirms the session to NS, NS and AS switch the session, but AS does not longer need to recalculate the downlink queue.

#### Can you do this yourself and submit a Pull Request?

Can review. Please coordinate. I would suggest that @rvolosatovs starts on the NS side, and @adriansmares basically removes the recalculation on session confirmation in AS.

johanstokking:
Blocking #1971

rvolosatovs:
I'd prefer merging #1611 before working on this. There are quite some downlink-related changes in #1611, which are very likely to cause a bunch of merge conflicts.

PR #2053, opened 2020-02-25
Refactor handling of deprecated fields in Network Server

rvolosatovs:
#### Summary

Refactor handling of deprecated fields in Network Server.
Refs #1954 

#### Changes

- Instead of manually specifying deprecated fields in RPCs, wrap the internal registry in a layer handling these
- Fix ineffective values being returned to the user

#### Notes for Reviewers

Unfortunately, so far handling of deprecated fields was basically non-existent, as in the old values are just ignored by NS - the values are `SET`, but they have no effect on NS operation. Kept it that way for backwards-compatibility.
The `GET` logic is also weird and confusing, since values returned do not reflect the actual state of the device.
For example usage of this see https://github.com/TheThingsNetwork/lorawan-stack/pull/2052/commits/4e74f74b999da5d2cd3eec767993514fa49f5184#diff-41cf204f41bb63a20074d957c04b6d89R243-R293 from #2052
We may want to extract some of this into a shared package later.

Functionality here provides NS ability to seamlessly migrate from deprecated fields to newer format.
On SET, the migration is performed.
There's always at most 1 value set in registry (either `Old` or `New` field)

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - ADR in 72-channel regions.
 - Payload length limits used by Network Server being too low.
 - CLI ignores default config files that cannot be read.
-- Fix device creation rollback potentially deleting existing device with same ID.
+- Device creation rollback potentially deleting existing device with same ID.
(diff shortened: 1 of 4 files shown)


rvolosatovs:
Blocking #2052

PR #2063, opened 2020-02-26
Remove inconsistencies in webhook path handling

### LINKED ISSUE #1995: Webhook configured URL trailing slash dropped
#### Summary

When you configure a Webhook via the console, and the URL ends in a / the / is dropped when the actual call is made.

#### Steps to Reproduce

Webhook is configured as in this example:

image

#### What do you see now?

In the docker-compose logs:
```
stack_1      | 2020-02-16T07:33:49.240509848Z  DEBUG Process message                          application_uid=jpm-crickets hook=ttnmapper namespace=applicationserver network_server=ace3b78e8707 url=http://private.ttnmapper.org/integrations/ttn/v3/uplink_message
```

Because the POST is done to an incorrect URL, the server will redirect the call to the correct one that does have a trailing slash. The original call was a POST, but according to the HTTP specs the retry to the correct URL should be a GET. The problem is that the GET does not contain the JSON payload. The server therefore sees an empty request body.

In the logs I want to see:
`url=http://private.ttnmapper.org/integrations/ttn/v3/uplink_message/`

#### Environment

The Things Stack v3.5.3 Docker image

Webhooks should use the specified URL as is, without stripping the trailing slash.

No

adriansmares:
#### Summary

Closes #1995 

#### Changes

- Changed the way the webhook paths are joined from `path.Join` to `url.ResolveReference`.

#### Notes for Reviewers

First of all, there is an additional bug which currently exist in the latest master which is produced by the current `path.Join` usage: the path URL is being escaped even if it contains a query string. See [this](https://play.golang.org/p/wE-9kZC0oP1) example which can be replicated in master with a base URL of `http://localhost/test` and an uplink message path of `bar?something=else`.

Edited: I've solved the issues that were here originally. The proposed changes are backwards compatible.

Using path URLs such as `?variable=something` is also possible now, without URL encoding issues.

#### Checklist

### DIFF
--- a/pkg/applicationserver/io/web/webhooks_test.go
+++ b/pkg/applicationserver/io/web/webhooks_test.go
@@ -54,285 +54,309 @@ func TestWebhooks(t *testing.T) {
 		ApplicationIdentifiers: registeredApplicationID,
 		WebhookID:              registeredWebhookID,
 	}
-	_, err := registry.Set(ctx, ids, nil, func(_ *ttnpb.ApplicationWebhook) (*ttnpb.ApplicationWebhook, []string, error) {
-		return &ttnpb.ApplicationWebhook{
-				ApplicationWebhookIdentifiers: ids,
-				BaseURL:                       "https://myapp.com/api/ttn/v3{/appID,devID}",
-				Headers: map[string]string{
-					"Authorization": "key secret",
-				},
-				DownlinkAPIKey: "foo.secret",
-				Format:         "json",
-				UplinkMessage: &ttnpb.ApplicationWebhook_Message{
-					Path: "up",
-				},
-				JoinAccept: &ttnpb.ApplicationWebhook_Message{
-					Path: "join",
-				},
-				DownlinkAck: &ttnpb.ApplicationWebhook_Message{
-					Path: "down/ack",
-				},
-				DownlinkNack: &ttnpb.ApplicationWebhook_Message{
-					Path: "down/nack",
-				},
-				DownlinkSent: &ttnpb.ApplicationWebhook_Message{
-					Path: "down/sent",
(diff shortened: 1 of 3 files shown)


johanstokking:
> * A webhook with a base URL of `http://localhost/bar` and a message path of `/foo` currently results in a path of `http://localhost/bar/foo`. After this PR, it results in `http://localhost/foo`, since the reference URL is `/foo` which means that the path is relative to the root. _This is breaking_

We can't do this.

pkg/applicationserver/io/web/webhooks_test.go
```diff
@@ -54,285 +54,291 @@ func TestWebhooks(t *testing.T) {
...
-							EndDeviceIdentifiers: registeredDeviceID,
-							Up: &ttnpb.ApplicationUp_UplinkMessage{
-								UplinkMessage: &ttnpb.ApplicationUplink{
-									SessionKeyID: []byte{0x11},
-									FPort:        42,
-									FCnt:         42,
-									FRMPayload:   []byte{0x1, 0x2, 0x3},
-								},
+	for _, prefix := range []string{"", "/"} {
+		for _, suffix := range []string{"", "/"} {
```
johanstokking: Please put this in a test case

pkg/applicationserver/io/web/webhooks_test.go
```diff
@@ -54,285 +54,291 @@ func TestWebhooks(t *testing.T) {
...
+			t.Run(fmt.Sprintf("Prefix'%s'/Suffix'%s'", prefix, suffix), func(t *testing.T) {
```
johanstokking: nit; `%q` is already quoted

Issue #1134, opened 2019-08-08
Make NS/AS KEK labels, AS-ID, NetID and claim authentication code editable


johanstokking:
#### Summary

Make NS/AS KEKs, AS-ID, home NetID and claim authentication code editable in the Console

#### Why do we need this?

User experience

#### What is already there? What do you see now?

No input fields

#### What is missing? What do you want to see?

Input fields in JS section for, AS address + ID + KEK label and NS address + KEK label, home NetID and claim authentication code.

This is only available in cluster-JS.

The fields are;

- `application_server_kek_label` (free-form <= 2 KB length)
- `application_server_id` (free-form, <= 100 length)
- `network_server_kek_label` (free-form, <= 2 KB length)
- `net_id` (6 digit hex, optional)
- `claim_authentication_code{.{valid_from|valid_to|value}}`

#### Can you do this yourself and submit a Pull Request?

No

@kschiffer please reassign if desired.

htdvisser:
Blocked by #579

johanstokking:
Mostly blocked by https://github.com/TheThingsNetwork/lorawan-stack/pull/1616

johanstokking:
Added home NetID and claim authentication code fields

johanstokking:
This is not entirely in progress by #1616; claim authentication code is not part of that (yet)

bafonins:
Regarding `claim_authentication_code` should the users be able to set the date and time, or only the date? @johanstokking

johanstokking:
@bafonins I think the date is fine, but internally it's a timestamp, so including time. If the user specifies 1-1-2021, it will be 1-1-2021 0:00:00, effectively 31-12-2020, so take care of that.

PR #2060, opened 2020-02-26
Fix disabled and narrow tabs styling

bafonins:
#### Summary

This PR fixes `disabled` and `narrow` styling of the `<Tabs />` component

#### Changes

- Do not render `<NavLink />` if tab is disabled, otherwise it is still possible to navigate
- Fix `narrow` styles. Styles were not applied to the tab components, as well as the values seemed incorrect.

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, database and configuration, according to the stability commitments in `README.md`.
- [x] Testing: The changes are covered with unit tests. The changes are tested manually as well.
- [x] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #2072, opened 2020-02-27
Add device authentication code page

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/1134

<details>

Screenshot 2020-02-28 at 13 41 07
Screenshot 2020-02-28 at 13 41 34
Screenshot 2020-02-28 at 13 41 46

</details>

#### Changes

- Add the `<DeviceClaimAuthenticationCode />` component
- Include emitting default value for `claim_authentication_code` field in the js sdk

#### Notes for Reviewers

@johanstokking Please verify that works as expected

Note: The users can only set date for `valid_from` and `valid_to` fields as mentioned in https://github.com/TheThingsNetwork/lorawan-stack/issues/1134#issuecomment-590930986. Changing this to the datetime input isnt difficult.

Also, it is still possible to access the page for abp/external js devices because of the issue in the `<Tabs />` component not disabling the link tabs, this is fixed in https://github.com/TheThingsNetwork/lorawan-stack/pull/2060 and will work properly after merging.

I will also try to find an appropriate abstraction for checking whether the end device has cluster js/as/ns addresses rather than just comparing those in the render function. Will add this in a separate PR with disabling the payload formatters tab when as is disabled or not in cluster. 

#### Checklist


pkg/webui/console/views/device-authentication-code/index.js
```diff
@@ -0,0 +1,212 @@
...
+  validateToDate: 'Invalid code expiration date',
+  deleteFailure: 'There was a problem deleting the authentication code',
+  deleteSuccess: 'The authentication code has been deleted successfully',
+  updateSuccess: 'The authentication code has been updated successfully',
+})
+
+const validationSchema = Yup.object({
+  claim_authentication_code: Yup.object({
+    value: Yup.string()
+      .matches(/^[A-Z0-9]{1,32}$/, sharedMessages.validateFormat)
```
kschiffer: The validation message should be clearer, eg. telling that only uppercase letters and numbers are allowed.

pkg/webui/console/views/device-authentication-code/index.js
```diff
@@ -0,0 +1,212 @@
...
+      .required(sharedMessages.validateRequired),
+    valid_from: Yup.date().required(sharedMessages.validateRequired),
+    valid_to: Yup.date()
```
kschiffer: Should we allow to-dates in the past?

bafonins: @johanstokking

bafonins:
> Any reason this is not part of the join server settings?

`claim_authentication_code` entry feels like a standalone settings, similar to location, payload formatters. But ill leave this to @johanstokking 

Also, making the authentication code form as a separate section makes the implementation much easier,.

pkg/webui/console/views/device-authentication-code/index.js
johanstokking: I don't think we should care too much. The backend doesn't check this, for example.

johanstokking:
I'm pretty neutral on whether it should be in claiming or a separate section.

I do, however, suggest to call this section Claiming, and it's a Claim Authentication Code.

We may come up with more authentication codes in the future, for other purposes.

bafonins:
Updates screenshots

johanstokking:
I only see the clear code just now; that should be `*****`.

pkg/webui/console/views/device-claim-authentication-code/index.js
```diff
@@ -0,0 +1,213 @@
...
+            error={error}
+            initialValues={initialValues}
+            validationSchema={validationSchema}
+            onSubmit={handleSubmit}
+          >
+            <Form.Field
+              title={sharedMessages.claimAuthCode}
+              name="claim_authentication_code.value"
+              component={Input}
+              required
```
kschiffer:
```suggestion
              type="password"
              required
```

pkg/webui/console/views/device/index.js
```diff
@@ -151,6 +165,12 @@ export default class Device extends React.Component {
         link: payloadFormattersLink,
         exact: false,
       },
+      {
+        title: sharedMessages.claiming,
+        name: 'claim-auth-code',
+        link: `${basePath}/claim-auth-code`,
+        disabled: !hasJs,
```
kschiffer: (I'd actually prefer to add a `mayViewOrSetClaimAuthenticationCode` feature check for this. Though the feature checks are not configured to include external conditions (`supports_join`), so it doesn't seem as straightforward. Okay to leave it like this then, unless you have another idea here.)

bafonins: From the PR:
> I will also try to find an appropriate abstraction for checking whether the end device has cluster js/as/ns addresses rather than just comparing those in the render function. Will add this in a separate PR with disabling the payload formatters tab when as is disabled or not in cluster.

PR #2080, opened 2020-02-29
UDP firewall flood checking

### LINKED ISSUE #2079: Message flood checking on UDP frontend
#### Summary

Allow UDP firewall to filter excessive incoming packets.

Discussion started in #1406, created here as a separate issue.

#### Why do we need this?

In case of gateway errors, too many messages may flood to the GS. We need to be able to protect GS from such scenarios. Also, this protects us from a handful of gateways overloading the GS because of very high traffic.

If no packet handlers are available when a new packet arrives, the packet is dropped.

That may still overload the GS in case of message flooding.

#### Environment

3.6.0 TTS

@johanstokking in #1406 suggested having a circular buffer with the last N packet timestamps. If a new packet arrives and the timestamp of the oldest message in the buffer is not too old (~a few ms), then drop the packet. 

yes

neoaggelos:
#### Summary

Closes #2079

#### Changes

- Add a circular buffer of timestamps on UDP firewall per gateway
- Drop messages that exceed a specified rate limit.

#### Notes for Reviewers

- Naming is all over the place, please suggest better names where possible
- Current limits are hard-coded in the code. Should probably be GS configuration options?
- Relevant tests have been added.
- Circular buffer implementation is based on `./pkg/gatewayserver/rtts.go`. In this case, maybe mutexes are not required at all?

#### Checklist

### DIFF
--- a/doc/content/reference/configuration/gateway-server.md
+++ b/doc/content/reference/configuration/gateway-server.md
@@ -73,3 +73,11 @@ Using the `packet-buffer` and `packet-handlers` options, the throughput of UDP p
 Specify options for gateway connection statistics:
 
 - `gs.update-connection-stats-debounce-time`: Time before repeated refresh of the gateway connection stats
+
+## UDP Rate Limiting Options
+
+The Gateway Server supports rate limiting traffic for gateways that are using the Semtech UDP protocol.
+
+- `gs.udp.rate-limiting.enable`: Enable rate limiting for gateways
+- `gs.udp.rate-limiting.messages`: Number of past messages to check timestamp for
+- `gs.udp.rate-limiting.threshold`: Filter packet if timestamp is not newer than the older timestamps of the previous messages by this threshold
--- a/pkg/gatewayserver/io/udp/firewall_ratelimit.go
+++ b/pkg/gatewayserver/io/udp/firewall_ratelimit.go
@@ -1,0 +1,75 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package udp
+
+import (
+	"sync"
+	"time"
+
+	"go.thethings.network/lorawan-stack/pkg/errors"
+	encoding "go.thethings.network/lorawan-stack/pkg/ttnpb/udp"
+)
+
+type rateLimitingFirewall struct {
+	f Firewall
+
+	m sync.Map // string to timestamps
+
+	messages  int
+	threshold time.Duration
+}
+
+// NewRateLimitingFirewall returns a Firewall with rate limiting capabilities.
+func NewRateLimitingFirewall(firewall Firewall, messages int, threshold time.Duration) Firewall {
+	return &rateLimitingFirewall{
+		f:         firewall,
+		messages:  messages,
+		threshold: threshold,
+	}
+}
+
+var (
+	errRateExceeded = errors.DefineResourceExhausted("rate_exceeded", "gateway traffic exceeded allowed rate")
+)
+
+func (f *rateLimitingFirewall) Filter(packet encoding.Packet) error {
+	if packet.GatewayEUI == nil {
+		return errNoEUI
+	}
+	if packet.GatewayAddr == nil {
+		return errNoAddress
+	}
+	now := time.Now().UTC()
+	eui := *packet.GatewayEUI
+	val, ok := f.m.Load(eui)
+	var ts *timestamps
+	if ok {
+		ts = val.(*timestamps)
+	} else {
+		ts = newTimestamps(f.messages)
+		f.m.Store(eui, ts)
+	}
+
+	oldestTimestamp := ts.Append(now)
+	if !oldestTimestamp.IsZero() && now.Sub(oldestTimestamp) < f.threshold {
+		return errRateExceeded
+	}
+
+	// Continue filtering
+	if f.f != nil {
+		return f.f.Filter(packet)
+	}
+	return nil
+}
--- a/pkg/gatewayserver/io/udp/timestamps.go
+++ b/pkg/gatewayserver/io/udp/timestamps.go
@@ -1,0 +1,43 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package udp
+
+import (
+	"sync"
+	"time"
+)
+
+type timestamps struct {
+	position int
+	mu       sync.RWMutex
+	items    []time.Time
+}
+
+func newTimestamps(count int) *timestamps {
+	return &timestamps{
+		items: make([]time.Time, count),
+	}
+}
+
+// Append adds new timestamp and returns oldest.
+func (i *timestamps) Append(val time.Time) time.Time {
+	i.mu.Lock()
+	defer i.mu.Unlock()
+	result := i.items[i.position]
+	i.items[i.position] = val
+	i.position = (i.position + 1) % len(i.items)
+
+	return result
+}
--- a/pkg/gatewayserver/io/udp/udp.go
+++ b/pkg/gatewayserver/io/udp/udp.go
@@ -32,6 +32,14 @@ import (
 	"go.thethings.network/lorawan-stack/pkg/unique"
 )
 
+// RateLimitingConfig contains configuration settings for the rate limiting
+// capabilities of the UDP gateway frontend firewall.
+type RateLimitingConfig struct {
+	Enable    bool          `name:"enable" description:"Enable rate limiting for gateways"`
+	Messages  int           `name:"messages" description:"Number of past messages to check timestamp for"`
+	Threshold time.Duration `name:"threshold" description:"Filter packet if timestamp is not newer than the older timestamps of the previous messages by this threshold"`
+}
+
 // Config contains configuration settings for the UDP gateway frontend.
 // Use DefaultConfig for recommended settings.
 type Config struct {
@@ -51,6 +59,8 @@ type Config struct {
 	ScheduleLateTime time.Duration `name:"schedule-late-time" description:"Time in advance to send downlink to the gateway when scheduling late"`
 	// AddrChangeBlock defines the time to block traffic when the address changes.
 	AddrChangeBlock time.Duration `name:"addr-change-block" description:"Time to block traffic when a gateway's address changes"`
+	// RateLimitingConfig is the configuration for the rate limiting firewall capabilities.
+	RateLimiting RateLimitingConfig `name:"rate-limiting"`
 }
 
 // DefaultConfig contains the default configuration.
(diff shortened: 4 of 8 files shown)


pkg/gatewayserver/io/udp/firewall.go
```diff
@@ -25,25 +25,50 @@ import (
...
 )
 
+var (
+	// Configure how many recent messages to use for rate limiting.
+	defaultRateLimitingMessages = 10
+
+	// Configure age threshold for oldest message when rate limiting. If oldest
+	// message is not older than this duration, then the rate is considered to
+	// have been exceeded and the packet is dropped.
+	defaultRateLimitingAgeThreshold = 10 * time.Millisecond
```
johanstokking: Comments start with the symbol name

johanstokking: And a `const` is not configuration

neoaggelos:
Changes since review:

- Refactored into a `rateLimitingFirewall`, which can wrap an existing `Firewall` while also providing the rate limiting capabilities.
- Rate limiting now takes precedence over the wrapped firewall, which is much better (during the flood, we avoid doing more stuff than needed).
- Added configuration options on the GS (`--gs.udp.rate-limiting`).
- Updated circular buffer implementation to use a fixed size array and a running pointer.
- Updated documentation and changelog

Will squash commits after reviewing is done.

pkg/gatewayserver/io/udp/firewall_ratelimit.go
```diff
@@ -0,0 +1,94 @@
...
+	encoding "go.thethings.network/lorawan-stack/pkg/ttnpb/udp"
+)
+
+var (
+	// defaultMessages is the default number of messages to use for rate limiting.
+	defaultMessages = 10
+
+	// defaultThreshold is the duration after which a new packet will be accepted.
+	defaultThreshold = 10 * time.Millisecond
+)
```
neoaggelos: These are probably not needed

pkg/gatewayserver/io/udp/firewall_ratelimit.go
```diff
@@ -0,0 +1,93 @@
...
+		return errNoAddress
+	}
+	now := time.Now().UTC()
+	eui := *packet.GatewayEUI
+	val, ok := f.m.Load(eui)
+	var timestamps *incoming
+	if ok {
+		timestamps = val.(*incoming)
+		oldestTimestamp := timestamps.Oldest()
+		if !oldestTimestamp.IsZero() && now.Sub(oldestTimestamp) < f.threshold {
```
neoaggelos: The `IsZero()` check is probably not needed, since if that is true, then the second condition will most likely fail

pkg/gatewayserver/io/udp/udp.go
```diff
@@ -61,6 +71,11 @@ var DefaultConfig = Config{
 	ConnectionExpires:   1 * time.Minute,  // Expire connection after missing typically 2 status messages.
 	ScheduleLateTime:    800 * time.Millisecond,
 	AddrChangeBlock:     1 * time.Minute, // Release address when the connection expires.
+	RateLimiting: RateLimitingConfig{
+		Enable:    true,
```
neoaggelos: We might not want to enable this by default

pkg/gatewayserver/io/udp/udp.go
```diff
@@ -32,6 +32,14 @@ import (
 	"go.thethings.network/lorawan-stack/pkg/unique"
 )
 
+// RateLimitingConfig contains configuration settings for the rate limiting
+// capabilities of the UDP gateway frontend firewall.
+type RateLimitingConfig struct {
+	Enable    bool          `name:"enable" description:"Enable rate limiting for gateways"`
+	Messages  int           `name:"messages" description:"Number of past messages to check timestamp for"`
+	Threshold time.Duration `name:"threshold" description:"Filter packet if timestamp is not newer than the older timestamps of the previous messages by this threshold"`
```
neoaggelos: Not too good at naming new things, these may need to change

doc/content/reference/configuration/gateway-server.md
```diff
@@ -62,3 +62,10 @@ Using the `packet-buffer` and `packet-handlers` options, the throughput of UDP p
 - `gs.udp.packet-buffer`: Buffer size of unhandled packets
 - `gs.udp.packet-handlers`: Number of concurrent packet handlers
 
+## UDP Rate Limiting Options
+
+The Gateway Server supports rate limiting traffic for gateways that are using the Semtech UDP protocol.
+
+- `gs.udp.rate-limiting.enable`: Enable rate limiting for gateways.
+- `gs.udp.rate-limiting.messages`: Number of past messages to check timestamp for
```
johanstokking: Trailing dot

pkg/gatewayserver/io/udp/firewall_ratelimit.go
johanstokking: Indeed, they can be removed. GS has default config already

pkg/gatewayserver/io/udp/firewall_ratelimit.go
```diff
@@ -0,0 +1,93 @@
...
+	defaultMessages = 10
+
+	// defaultThreshold is the duration after which a new packet will be accepted.
+	defaultThreshold = 10 * time.Millisecond
+)
+
+type rateLimitingFirewall struct {
+	f Firewall
+
+	m sync.Map // eui to incoming
```
johanstokking: `eui` is not a type

pkg/gatewayserver/io/udp/firewall_ratelimit.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// NewRateLimitingFirewall returns a Firewall with rate limiting capabilities.
+func NewRateLimitingFirewall(firewall Firewall, messages int, threshold time.Duration) Firewall {
+	if messages == 0 {
+		messages = defaultMessages
+	}
+	if threshold == 0 {
+		threshold = defaultThreshold
+	}
```
johanstokking: These checks are useful in libraries with a public API

If there's sanity checking for bad config, it should return an error here. But I don't think that is all necessary and that you can trust the values

pkg/gatewayserver/io/udp/firewall_ratelimit.go
```diff
@@ -0,0 +1,93 @@
...
+		if !oldestTimestamp.IsZero() && now.Sub(oldestTimestamp) < f.threshold {
+			return errRateExceeded
+		}
+	} else {
+		timestamps = newIncoming(f.messages)
+	}
+
+	// Update firewall
+	timestamps.Append(now)
+	f.m.Store(eui, timestamps)
```
johanstokking: There's no need to update the map. You probably just want to store a pointer to the timestamps

pkg/gatewayserver/io/udp/firewall_test.go
```diff
@@ -148,4 +149,57 @@ func TestMemoryFirewall(t *testing.T) {
 			time.Sleep(tc.WaitAfter)
 		})
 	}
+
+	t.Run("FloodCheck", func(t *testing.T) {
+		ctx := test.Context()
+
+		for _, tc := range []struct {
+			rateLimiting bool
+			verify       func(error) bool
```
johanstokking: `errorAssertion`

pkg/gatewayserver/io/udp/firewall_test.go
```diff
@@ -148,4 +149,57 @@ func TestMemoryFirewall(t *testing.T) {
...
+		}{
+			{
+				rateLimiting: true,
+				verify:       func(err error) bool { return errors.IsResourceExhausted(err) },
```
johanstokking: `errorAssertion: errors.IsResourceExhausted`

pkg/gatewayserver/io/udp/firewall_test.go
```diff
@@ -148,4 +149,57 @@ func TestMemoryFirewall(t *testing.T) {
...
+			},
+			{
+				rateLimiting: false,
+				verify:       func(err error) bool { return err == nil },
+			},
+		} {
+			t.Run(fmt.Sprintf("RateLimiting=%v", tc.rateLimiting), func(t *testing.T) {
```
johanstokking: High coupling; I would prefer keeping the tests for rate limiting separate

pkg/gatewayserver/io/udp/firewall_ratelimit.go
```diff
@@ -0,0 +1,93 @@
...
+		oldestTimestamp := timestamps.Oldest()
+		if !oldestTimestamp.IsZero() && now.Sub(oldestTimestamp) < f.threshold {
+			return errRateExceeded
+		}
+	} else {
+		timestamps = newIncoming(f.messages)
+	}
+
+	// Update firewall
+	timestamps.Append(now)
```
johanstokking: `Oldest()` is a read, `Append()` a write. Both are independently locked, but it's nicer to have one atomic operation that appends and returns the delta with the oldest value.

pkg/gatewayserver/io/udp/incoming.go
```diff
@@ -0,0 +1,49 @@
...
+// limitations under the License.
+
+package udp
+
+import (
+	"sync"
+	"time"
+)
+
+type incoming struct {
```
johanstokking: Isn't this just `timestamps`?

pkg/gatewayserver/io/udp/incoming.go
```diff
@@ -0,0 +1,49 @@
...
+	count   int
```
johanstokking: What is the purpose of this var?

pkg/gatewayserver/io/udp/udp.go
johanstokking: Name is good

pkg/gatewayserver/io/udp/udp.go
johanstokking: We actually do

pkg/gatewayserver/io/udp/firewall_test.go
neoaggelos: Will move into separate `firewall_ratelimiting_test.go`

pkg/gatewayserver/io/udp/incoming.go
neoaggelos: Used to reset counter. We can drop this if `len(items)` is O(1).

pkg/gatewayserver/io/udp/firewall_ratelimit.go
neoaggelos: Updated `Append()` to also return oldest value.

pkg/gatewayserver/io/udp/incoming.go
neoaggelos: Renamed to timestamps

pkg/gatewayserver/io/udp/firewall_ratelimit.go
neoaggelos: Removed, GS default config should suffice

pkg/gatewayserver/io/udp/firewall_ratelimit.go
neoaggelos: Updated comment

pkg/gatewayserver/io/udp/firewall_ratelimit.go
neoaggelos: Updated to only store the reference once (when creating the new object)

neoaggelos:
We still need to make sure that stale entries on this table are removed at regular intervals. One way would be to have another `gc()` here like with the `memoryFirewall` and empty the map every X minutes.

A slightly better approach could be to extend the `Firewall` interface with a `Clean(ttnpb.GatewayIdentifiers)` function. This is out of scope for the existing PR.

@johanstokking thoughts?

pkg/gatewayserver/io/udp/incoming.go
johanstokking: It is

pkg/gatewayserver/io/udp/timestamps.go
```diff
@@ -0,0 +1,45 @@
...
+
+package udp
+
+import (
+	"sync"
+	"time"
+)
+
+type timestamps struct {
+	count   int
```
johanstokking: Remove, see https://blog.golang.org/go-slices-usage-and-internals

pkg/gatewayserver/io/udp/timestamps.go
```diff
@@ -0,0 +1,45 @@
...
+	current int
```
johanstokking: This is not used as `current`, it's used as the (new) position, i.e. `pos` or `position`

pkg/gatewayserver/io/udp/timestamps.go
```diff
@@ -0,0 +1,45 @@
...
+		count: count,
+		items: make([]time.Time, count),
+	}
+}
+
+// Append adds new timestamp and returns oldest.
+func (i *timestamps) Append(val time.Time) time.Time {
+	i.mu.Lock()
+	defer i.mu.Unlock()
+	result := i.items[i.current]
```
johanstokking: See here; the oldest is "current", that can't be right

PR #2083, opened 2020-03-03
Skip forward entry if prefixes are empty

johanstokking:
#### Summary

Skip forward entry if prefixes are empty

Needed for https://github.com/TheThingsIndustries/lorawan-stack-aws/pull/110

#### Changes

- Skip upstream if the prefixes are not defined

#### Notes for Reviewers

This doesn't break configuration. It used to work before, but an upstream would be configured and connected, which is now skipped if there are no prefixes to forward traffic for.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Network and Application Servers now maintain application downlink queue per-session.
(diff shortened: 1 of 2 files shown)


johanstokking:
> This is to allow empty values to upstream name in config?

Yes; to basically declare:

```yaml
forward:
  packetbroker: ''
```

Which doesn't set up the Packet Broker upstream at all. It's an optimization and allows for easier configuration generation.

PR #2084, opened 2020-03-03
Move config structs to appropriate packages

htdvisser:
#### Summary

This PR moves the Redis and Cluster config structs to their respective packages in order to avoid import cycles.

#### Notes for Reviewers

...

#### Checklist

### DIFF
--- a/pkg/redis/redis.go
+++ b/pkg/redis/redis.go
@@ -27,7 +27,6 @@ import (
 
 	"github.com/go-redis/redis"
 	"github.com/gogo/protobuf/proto"
-	"go.thethings.network/lorawan-stack/pkg/config"
 )
 
 const (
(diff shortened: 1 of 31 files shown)


pkg/redis/redis.go
```diff
@@ -76,12 +75,38 @@ type Client struct {
...
+	Database      int            `name:"database" description:"Redis database to use"`
+	RootNamespace []string       `name:"namespace" description:"Namespace for Redis keys"`
+	PoolSize      int            `name:"pool-size" description:"The maximum number of database connections"`
+	Failover      FailoverConfig `name:"failover" description:"Redis failover configuration"`
+	namespace     []string
+}
+
+func (c Config) WithNamespace(namespace ...string) *Config {
+	deriv := c
+	deriv.namespace = namespace
```
rvolosatovs: This should be `append` IMO. Otherwise this method should be called `SetNamespace`.

johanstokking: I don't think so. That might result in accidentally concatenating namespaces.

Namespace is one value. That is also why it is singular `namespace` and not `namespaces`. It is passed as a string slice but that is to leave the delimiter up to the store.

PR #2088, opened 2020-03-03
Fix panic when unregistered Basic Station gateways connect

KrishnaIyer:
#### Summary

Fix panic when unregistered Basic Station gateways connect

Original error message from Sentry
```
http_recovered: Internal Server Error
  File "recover.go", line 38, in go.thethings.network/lorawan-stack/pkg/web/middleware.Recover.func1.1.1
  File "context.go", line 40, in go.thethings.network/lorawan-stack/pkg/log.FromContext
  File "basicstationlns.go", line 464, in go.thethings.network/lorawan-stack/pkg/gatewayserver/io/basicstationlns.writeDiscoverError
  File "basicstationlns.go", line 141, in go.thethings.network/lorawan-stack/pkg/gatewayserver/io/basicstationlns.(*srv).handleDiscover
  File "echo.go", line 500, in (*Echo).add.func1
  File "recover.go", line 44, in go.thethings.network/lorawan-stack/pkg/web/middleware.Recover.func1.1
  File "log.go", line 44, in go.thethings.network/lorawan-stack/pkg/web/middleware.Log.func1.1
  File "body_limit.go", line 87, in BodyLimitWithConfig.func1.1
  File "id.go", line 47, in go.thethings.network/lorawan-stack/pkg/web/middleware.ID.func1.1
  File "echo.go", line 611, in (*Echo).ServeHTTP
  File "server.go", line 2802, in serverHandler.ServeHTTP
  File "server.go", line 1890, in (*conn).serve
DEVICE
```

#### Changes

- `FillGatewayContext` returns a `nil` context for unregistered gateways. Recreate the logger when calling `writeDiscoverError`.

#### Notes for Reviewers

...

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/basicstationlns/basicstationlns.go
+++ b/pkg/gatewayserver/io/basicstationlns/basicstationlns.go
@@ -135,12 +135,14 @@ func (s *srv) handleDiscover(c echo.Context) error {
 	ids := ttnpb.GatewayIdentifiers{
 		EUI: &req.EUI.EUI64,
 	}
-	ctx, ids, err = s.server.FillGatewayContext(ctx, ids)
+	filledCtx, ids, err := s.server.FillGatewayContext(ctx, ids)
 	if err != nil {
 		logger.WithError(err).Warn("Failed to fetch gateway")
 		writeDiscoverError(ctx, ws, fmt.Sprintf("Failed to fetch gateway: %s", err.Error()))
 		return err
 	}
+
+	ctx = filledCtx
 
 	scheme := "ws"
 	if c.IsTLS() || s.useTrafficTLSAddress {


KrishnaIyer:
Well that wouldn't be idiomatic since we should ignore other return values in case of errors right?

htdvisser:
That is not strictly required. Take for example the [`io.Reader`](https://golang.org/pkg/io/#Reader) interface, were you also need to look at the other return value (`n`) before considering `err`.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -137,6 +137,7 @@ func (s *srv) handleDiscover(c echo.Context) error {
 	}
 	ctx, ids, err = s.server.FillGatewayContext(ctx, ids)
 	if err != nil {
+		ctx = log.NewContext(c.Request().Context(), logger)
 		logger.WithError(err).Warn("Failed to fetch gateway")
 		writeDiscoverError(ctx, ws, fmt.Sprintf("Failed to fetch gateway: %s", err.Error()))
```
johanstokking: I think we shouldn't rely on the returned `ctx` if there's an error, especially if the method is called `FillGatewayContext`.

I would instead declare `filledCtx` only assign `ctx` when there's no error. In the error handling, use the original `ctx`

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -135,9 +135,8 @@ func (s *srv) handleDiscover(c echo.Context) error {
 	ids := ttnpb.GatewayIdentifiers{
 		EUI: &req.EUI.EUI64,
 	}
-	ctx, ids, err = s.server.FillGatewayContext(ctx, ids)
+	_, ids, err = s.server.FillGatewayContext(ctx, ids)
```
johanstokking: We should use the context, that's the point of filling it. We just shouldn't use it when this errors

KrishnaIyer: Yeah but the context is not actually needed here. `/router-info` only needs to check if the gateway is registered (or required to be registered). The actual returned context is not needed for this function.

johanstokking: You don't know that at this point. We should be able to decorate the gateway context by GS, and we should not have specific frontends make assumptions about the need for that

KrishnaIyer: I don't get what you mean. Where in this bit of code do you see the need for using this returned context: https://github.com/TheThingsNetwork/lorawan-stack/blob/d2bd53d6d8b84700084d66b0375b591645542a85/pkg/gatewayserver/io/basicstationlns/basicstationlns.go#L139-L170?

johanstokking: Taking the discussion offline

PR #2107, opened 2020-03-09
Fix external JS check for end devices in Console

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2039
This PR fixes the way we check for devices registered on an external join server. Previously we checked for the root keys, however checking for addresses is more robust way to determine if the device is not in cluster JS. This fixes issues for users with only `RIGHT_APPLICATION_DEVICES_WRITE_KEYS` right, so they can still view JS related fields except for keys.

#### Changes

- Check for addresses instead of root keys
- Use `host-from-url` utility in the device general settings view

#### Notes for Reviewers

The main issue here is that otaa devices:
1. Can be registered on an external join server, which means that the cluster js (if exists) doesnt have the entry for the device (even if it has, we ignore it in the Console). With this PR we handle this case
2. Can have root keys provisioned by an external join server, which means that the cluster can have the entry for the device without the keys. Unexposed root keys are handler in the join server form of the device general settings separately.

#### Checklist

### DIFF
--- a/pkg/webui/console/views/device-general-settings/utils.js
+++ b/pkg/webui/console/views/device-general-settings/utils.js
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import { selectJsConfig } from '../../../lib/selectors/env'
+import getHostnameFromUrl from '../../../lib/host-from-url'
 
 const lwRegexp = /^[1-9].[0-9].[0-9]$/
 const lwCache = {}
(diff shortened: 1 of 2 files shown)


pkg/webui/console/views/device-general-settings/utils.js
```diff
@@ -75,20 +76,17 @@ export const isDeviceABP = device =>
 export const isDeviceMulticast = device => Boolean(device) && Boolean(device.multicast)
 
 /**
- * Returns whether device keys are provisioned on external join server. Note: It is possible
- * to create end devices with root keys being provisioned by an external Join Server.
- * In this case both `root_keys.nwk_key` and `root_keys.app_key` must be missing in the payload.
+ * Returns whether an end device is not provisioned on the cluster join server.
```
pgalic96:
```suggestion
 * Returns whether an end device is provisioned on an external join server.
```
It is a bit confusing the way you worded it.

PR #2108, opened 2020-03-09
gs: Only update statistics for new traffic

neoaggelos:
#### Summary

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/1949

#### Changes

- `GatewayConnectionStatsRegistry.Set()` only updates the relevant stats and does not overwrite the other values.
- Also added a forgotten `notifyStatsChanged()` called when sending downlinks

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -653,19 +653,24 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
 	}
 }
 
-// UpdateConnectionStats updates the stats for a single gateway connection.
-func (gs *GatewayServer) UpdateConnectionStats(ctx context.Context, conn *io.Connection) error {
-	return gs.statsRegistry.Set(ctx, conn.Gateway().GatewayIdentifiers, conn.Stats())
-}
-
 func (gs *GatewayServer) updateConnStats(conn connectionEntry) {
 	ctx := conn.Context()
 	logger := log.FromContext(ctx)
 
+	// Initial dummy update, so that gateway appears connected
+	if err := gs.UpdateConnectionStats(conn.Connection, false, false, true); err != nil {
+		logger.WithError(err).Error("Failed to initialize connection stats")
+	}
+
+	var (
+		lastUpdateUp, lastUpdateDown, lastUpdateStatus time.Time
+		newUp, newDown, newStatus                      bool
+	)
+
 	defer func() {
 		logger.Debug("Delete connection stats")
-		err := gs.statsRegistry.Set(ctx, conn.Gateway().GatewayIdentifiers, nil)
-		if err != nil {
+
+		if err := gs.ClearConnectionStats(conn.Connection, newUp, newDown, true); err != nil {
 			logger.WithError(err).Error("Failed to delete connection stats")
 		}
 	}()
@@ -674,17 +679,31 @@ func (gs *GatewayServer) updateConnStats(conn connectionEntry) {
 		case <-ctx.Done():
 			return
 		case <-conn.StatsChanged():
-			err := gs.UpdateConnectionStats(ctx, conn.Connection)
-			if err != nil {
-				logger.WithError(err).Error("Failed to update connection stats")
-			}
+		}
 
-			timeout := time.After(gs.updateConnectionStatsDebounceTime)
-			select {
-			case <-ctx.Done():
-				return
-			case <-timeout:
-			}
+		stats := conn.Stats()
+		if stats.LastUplinkReceivedAt != nil && stats.LastUplinkReceivedAt.After(lastUpdateUp) {
+			newUp = true
+			lastUpdateUp = *stats.LastUplinkReceivedAt
+		}
+		if stats.LastDownlinkReceivedAt != nil && stats.LastDownlinkReceivedAt.After(lastUpdateDown) {
+			newDown = true
+			lastUpdateDown = *stats.LastDownlinkReceivedAt
+		}
+		if stats.LastStatusReceivedAt != nil && stats.LastStatusReceivedAt.After(lastUpdateStatus) {
+			newStatus = true
+			lastUpdateStatus = *stats.LastStatusReceivedAt
+		}
+
+		if err := gs.UpdateConnectionStats(conn.Connection, newUp, newDown, newStatus); err != nil {
+			logger.WithError(err).Error("Failed to update connection stats")
+		}
+
+		timeout := time.After(gs.updateConnectionStatsDebounceTime)
+		select {
+		case <-ctx.Done():
+			return
+		case <-timeout:
 		}
 	}
 }
@@ -831,4 +850,14 @@ func recoverHandler(ctx context.Context) error {
 		return err
 	}
 	return nil
+}
+
+// UpdateConnectionStats updates the connection stats for a gateway
+func (gs *GatewayServer) UpdateConnectionStats(conn *io.Connection, up, down, status bool) error {
+	return gs.statsRegistry.Set(conn.Context(), conn.Gateway().GatewayIdentifiers, conn.Stats(), up, down, status)
+}
+
+// ClearConnectionStats clears the connection stats for a gateway
+func (gs *GatewayServer) ClearConnectionStats(conn *io.Connection, up, down, status bool) error {
+	return gs.statsRegistry.Set(conn.Context(), conn.Gateway().GatewayIdentifiers, nil, up, down, status)
 }
--- a/pkg/gatewayserver/gatewayserver_test.go
+++ b/pkg/gatewayserver/gatewayserver_test.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"os"
 	"sync"
 	"testing"
 	"time"
@@ -86,9 +87,7 @@ func TestGatewayServer(t *testing.T) {
 		},
 	})
 	c.FrequencyPlans = frequencyplans.NewStore(test.FrequencyPlansFetcher)
-	statsRedisClient, statsFlush := test.NewRedis(t, "gatewayserver_test")
-	defer statsFlush()
-	defer statsRedisClient.Close()
+
 	config := &gatewayserver.Config{
 		RequireRegisteredGateways:         false,
 		UpdateGatewayLocationDebounceTime: 0,
@@ -113,10 +112,17 @@ func TestGatewayServer(t *testing.T) {
 			WSPingInterval: wsPingInterval,
 		},
 		UpdateConnectionStatsDebounceTime: 0,
-		Stats: &gsredis.GatewayConnectionStatsRegistry{
-			Redis: statsRedisClient,
-		},
 	}
+	if os.Getenv("TEST_REDIS") == "1" {
+		statsRedisClient, statsFlush := test.NewRedis(t, "gatewayserver_test")
+		defer statsFlush()
+		defer statsRedisClient.Close()
+		statsRegistry := &gsredis.GatewayConnectionStatsRegistry{
+			Redis: statsRedisClient,
+		}
+		config.Stats = statsRegistry
+	}
+
 	gs, err := gatewayserver.New(c, config)
 	if !a.So(err, should.BeNil) {
 		t.FailNow()
@@ -1058,9 +1064,14 @@ func TestGatewayServer(t *testing.T) {
 								time.Sleep(timeout)
 							}
 
+							time.Sleep(2 * timeout)
+
 							conn, ok := gs.GetConnection(ctx, ids)
 							a.So(ok, should.BeTrue)
-							gs.UpdateConnectionStats(ctx, conn)
+							a.So(conn.Stats(), should.NotBeNil)
+							if config.Stats != nil {
+								a.So(gs.UpdateConnectionStats(conn, true, true, true), should.BeNil)
+							}
 
 							stats, err := statsClient.GetGatewayConnectionStats(statsCtx, &ids)
 							if !a.So(err, should.BeNil) {
@@ -1324,9 +1335,14 @@ func TestGatewayServer(t *testing.T) {
 								t.Fatal("Expected downlink timeout")
 							}
 
+							time.Sleep(2 * timeout)
+
 							conn, ok := gs.GetConnection(ctx, ids)
 							a.So(ok, should.BeTrue)
-							gs.UpdateConnectionStats(ctx, conn)
+							a.So(conn.Stats(), should.NotBeNil)
+							if config.Stats != nil {
+								a.So(gs.UpdateConnectionStats(conn, true, true, true), should.BeNil)
+							}
 
 							stats, err := statsClient.GetGatewayConnectionStats(statsCtx, &ids)
 							if !a.So(err, should.BeNil) {
--- a/pkg/gatewayserver/io/io.go
+++ b/pkg/gatewayserver/io/io.go
@@ -313,6 +313,8 @@ func (c *Connection) SendDown(msg *ttnpb.DownlinkMessage) error {
 	case c.downCh <- msg:
 		atomic.AddUint64(&c.downlinks, 1)
 		atomic.StoreInt64(&c.lastDownlinkTime, time.Now().UnixNano())
+
+		c.notifyStatsChanged()
 	default:
 		return errBufferFull.New()
 	}
--- a/pkg/gatewayserver/redis/registry.go
+++ b/pkg/gatewayserver/redis/registry.go
@@ -16,7 +16,10 @@ package redis
 
 import (
 	"context"
+	"runtime/trace"
 
+	"github.com/go-redis/redis"
+	"go.thethings.network/lorawan-stack/pkg/errors"
 	ttnredis "go.thethings.network/lorawan-stack/pkg/redis"
 	"go.thethings.network/lorawan-stack/pkg/ttnpb"
 	"go.thethings.network/lorawan-stack/pkg/unique"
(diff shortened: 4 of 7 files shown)


neoaggelos:
> This state and memory becomes dirty, but I understand the need
> 
> Can you instead write over the notification channel an enum with whether up/down/status has been updated?

Writing what has changed over the notification channel would not really work. Since the notification channel has a fixed size, only the first enum sent would be read (the other messages sent to the channel will be dropped), which is not correct. 

> For storing, you can do two things;
> 
> 1. Load the proto, update it, and write it back. This is more TTS-like
> 2. Use Redis hash maps for up/down/status, which requires three reads to get the full state

Going with 1 for the time being.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -675,6 +690,9 @@ func (gs *GatewayServer) updateConnStats(conn connectionEntry) {
 		case <-ctx.Done():
 			return
 		case <-conn.StatsChanged():
+			if conn.Connection != nil {
+				logger.Error("Connection does not exist")
```
johanstokking: Doesn't look very logical to me

pkg/gatewayserver/io/io.go
```diff
@@ -90,6 +90,15 @@ type Connection struct {
 
 	statsChangedCh chan struct{}
 	locCh          chan struct{}
+
+	newTraffic Traffic
```
johanstokking: I really don't think we should be keeping state like this

johanstokking:
> Writing what has changed over the notification channel would not really work. Since the notification channel has a fixed size, only the first enum sent would be read (the other messages sent to the channel will be dropped), which is not correct.

Add channels for each message change in that case.

pkg/gatewayserver/gatewayserver.go
neoaggelos: Leftover from testing, deleted

pkg/gatewayserver/io/io.go
neoaggelos: Changed to channels per traffic type.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -654,38 +654,51 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
+// UpdateAllConnectionStats updates the stats for a single gateway connection.
+func (gs *GatewayServer) UpdateAllConnectionStats(ctx context.Context, conn *io.Connection) error {
+	return gs.statsRegistry.Set(ctx, conn.Gateway().GatewayIdentifiers, conn.Stats(), allTraffic)
 }
 
 func (gs *GatewayServer) updateConnStats(conn connectionEntry) {
 	ctx := conn.Context()
 	logger := log.FromContext(ctx)
 
+	if gs.statsRegistry == nil {
```
johanstokking: This function will (or should not) be called in this case

pkg/gatewayserver/io/io.go
```diff
@@ -88,8 +88,17 @@ type Connection struct {
...
-	statsChangedCh chan struct{}
-	locCh          chan struct{}
+	upChangedCh     chan struct{}
+	downChangedCh   chan struct{}
+	statusChangedCh chan struct{}
+	locCh           chan struct{}
+}
+
+// Traffic denotes the kind of traffic.
+type Traffic struct {
```
johanstokking: Now I wonder why this is necessary still, if you basically keep local state in the func

pkg/gatewayserver/gatewayserver.go
```diff
@@ -654,38 +654,51 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
+			traffic.Status = true
+		}
 
-			timeout := time.After(gs.updateConnectionStatsDebounceTime)
-			select {
-			case <-ctx.Done():
-				return
-			case <-timeout:
-			}
+		err := gs.statsRegistry.Set(ctx, conn.Gateway().GatewayIdentifiers, conn.Stats(), conn.GetTraffic(traffic))
```
johanstokking: You know from the local vars above which fields are set, right? You read those channels already, why read them again in `GetTraffic()`?

pkg/gatewayserver/redis/registry.go
```diff
@@ -32,14 +36,66 @@ func (r *GatewayConnectionStatsRegistry) key(uid string) string {
...
+				}
+				if newTraffic.Down {
+					stored.LastDownlinkReceivedAt = stats.LastDownlinkReceivedAt
+					stored.DownlinkCount = stats.DownlinkCount
+					stored.RoundTripTimes = stats.RoundTripTimes
+				}
+				if newTraffic.Status {
+					stored.LastStatus = stats.LastStatus
+					stored.LastStatusReceivedAt = stats.LastStatusReceivedAt
+				}
```
johanstokking: Is there a reason _not_ to keep these in separate keys, while using the same proto?

pkg/gatewayserver/gatewayserver.go
neoaggelos: This is just to make sure that we we don't continue with a `nil` registry for whatever reason, can remove

pkg/gatewayserver/io/io.go
neoaggelos: Removed, updating based on timestamps as discussed.

pkg/gatewayserver/gatewayserver.go
neoaggelos: As discussed, this was to drain updates of all message types reliably, removed since we use one channel only.

pkg/gatewayserver/redis/registry.go
neoaggelos: If we store in separate keys (and the stats are updated based on timestamp), then we have two options:

- For every stats update, first read all three keys from Redis (for up, down and status respectively), check if you have a newer timestamp and then update which are needed. (3 reads, <= 3 writes)
- Keep the timestamp of the last update as state, and blind update whichever stats have a more recent timestamp. This is not as safe as the first option (a blind write may be incorrect in this case), requires extra state for each `io.Connection`, but only does <= 3 writes.

Currently, storing on a single key needs 1 read, 1 write, but has more collisions with concurrent updates (e.g. from multiple server instances updating different stats).

pkg/gatewayserver/gatewayserver.go
johanstokking: That’s an implementation error for which we can panic

pkg/gatewayserver/redis/registry.go
johanstokking: The thing is that uplink, downlink and status streams are statefully routed to instances, but separately, so that uplink, downlink and status may end up on different instances. You don’t have to account for multiple instances handling uplink, only for multiple instances handling uplink vs downlink.

You can therefore safely keep timestamps in memory for uplink, downlink and status separately.

pkg/gatewayserver/gatewayserver.go
johanstokking: Remove check; let the stack panic on nil pointer reference. It's really an implementation error and we cannot begin to add these kinds of assertions everywhere.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -654,18 +654,17 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
 	logger := log.FromContext(ctx)
 
+	if gs.statsRegistry == nil {
+		panic("No stats registry available")
+	}
+
 	defer func() {
 		logger.Debug("Delete connection stats")
-		err := gs.statsRegistry.Set(ctx, conn.Gateway().GatewayIdentifiers, nil)
+		err := conn.ClearStats(gs.statsRegistry)
```
johanstokking: Are the statistics cleared even if upstream/downstream/status can be connected still to another instance?

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -113,9 +116,7 @@ func TestGatewayServer(t *testing.T) {
 			WSPingInterval: wsPingInterval,
 		},
 		UpdateConnectionStatsDebounceTime: 0,
-		Stats: &gsredis.GatewayConnectionStatsRegistry{
-			Redis: statsRedisClient,
-		},
+		Stats:                             statsRegistry,
```
johanstokking: Out of curiosity; why this diff?

pkg/gatewayserver/gatewayserver.go
```diff
@@ -675,17 +674,17 @@ func (gs *GatewayServer) updateConnStats(conn connectionEntry) {
...
-			}
+		}
 
-			timeout := time.After(gs.updateConnectionStatsDebounceTime)
-			select {
-			case <-ctx.Done():
-				return
-			case <-timeout:
-			}
+		if err := conn.UpdateStats(gs.statsRegistry); err != nil {
```
johanstokking: Can you keep the timestamps here locally in this function, and pass to `UpdateStats()` what exactly needs to be updated?

Better yet, can we keep the `io` layer separate from persistency? Is there a reason to add storage to `io.Connection`?

pkg/gatewayserver/gatewayserver.go
neoaggelos: Initially did not account for this, I have updated the code to only clear the specified statistics

pkg/gatewayserver/gatewayserver_test.go
neoaggelos: In a previous iteration `statsRegistry` was used below for some debugging. Was left because it looks like the code is a bit more clear.

pkg/gatewayserver/gatewayserver.go
neoaggelos: Moved logic into the go-routine instead.

johanstokking:
Please request review when ready

neoaggelos:
> Please request review when ready

Will do, currently testing and updating the unit tests.

pkg/gatewayserver/io/io.go
```diff
@@ -92,6 +92,22 @@ type Connection struct {
...
+
+// LastUpdate holds the timestamps for the latest update for each message type.
+type LastUpdate struct {
+	Up, Down, Status time.Time
+}
+
+// Traffic describes the type of traffic.
+type Traffic struct {
+	Up, Down, Status bool
+}
```
johanstokking: OK, now I would turn this into flags

pkg/gatewayserver/io/io.go
```diff
@@ -92,6 +92,22 @@ type Connection struct {
...
+// GatewayConnectionStatsRegistry is the interface of the GatewayConnectionStats as seen by the Connection.
+type GatewayConnectionStatsRegistry interface {
+	// Set updates or clears the connection stats
+	Set(ctx context.Context, ids ttnpb.GatewayIdentifiers, stats *ttnpb.GatewayConnectionStats, update Traffic) error
+}
+
+// LastUpdate holds the timestamps for the latest update for each message type.
+type LastUpdate struct {
+	Up, Down, Status time.Time
+}
```
johanstokking: And remove this one; these can be just local vars in the method

pkg/gatewayserver/redis/registry.go
```diff
@@ -27,28 +31,127 @@ type GatewayConnectionStatsRegistry struct {
...
-		stats = nil
+		return nil, ttnredis.ConvertError(err)
 	}
-	return stats, err
+
+	if retrieved[0] == nil && retrieved[1] == nil && retrieved[2] == nil {
+		return nil, errNotFound
+	}
+
+	// uplink stats
```
johanstokking: nit; comments are sentences starting with capital and ending with dot

pkg/gatewayserver/redis/registry.go
```diff
@@ -27,28 +31,127 @@ type GatewayConnectionStatsRegistry struct {
...
+	if retrieved[0] != nil {
+		if ttnredis.UnmarshalProto(retrieved[0].(string), stats) == nil {
+			if retrieved[2] == nil {
+				result.ConnectedAt = stats.ConnectedAt
+				result.Protocol = stats.Protocol
+			}
```
johanstokking: This doesn't seem necessary

pkg/gatewayserver/redis/registry.go
```diff
@@ -27,28 +31,127 @@ type GatewayConnectionStatsRegistry struct {
...
 	}
-	return stats, err
+
+	if retrieved[0] == nil && retrieved[1] == nil && retrieved[2] == nil {
+		return nil, errNotFound
+	}
+
+	// uplink stats
+	if retrieved[0] != nil {
+		if ttnredis.UnmarshalProto(retrieved[0].(string), stats) == nil {
```
johanstokking: Use proper error handling. Also, if this fails, that's not a `errNotFound` but a corrupt store

pkg/gatewayserver/gatewayserver.go
```diff
@@ -837,11 +836,11 @@ func (gs *GatewayServer) GetMQTTConfig(ctx context.Context) (*config.MQTT, error
...
-func (gs *GatewayServer) UpdateConnectionStats(conn *io.Connection, traffic io.Traffic) error {
-	return gs.statsRegistry.Set(conn.Context(), conn.Gateway().GatewayIdentifiers, conn.Stats(), traffic)
+func (gs *GatewayServer) UpdateConnectionStats(conn *io.Connection, up bool, down bool, status bool) error {
+	return gs.statsRegistry.Set(conn.Context(), conn.Gateway().GatewayIdentifiers, conn.Stats(), up, down, status)
 }
 
 // ClearConnectionStats clears the connection stats for a gateway
-func (gs *GatewayServer) ClearConnectionStats(conn *io.Connection, traffic io.Traffic) error {
-	return gs.statsRegistry.Set(conn.Context(), conn.Gateway().GatewayIdentifiers, nil, traffic)
+func (gs *GatewayServer) ClearConnectionStats(conn *io.Connection, up bool, down bool, status bool) error {
```
johanstokking: nit; `up, down, status bool`

pkg/gatewayserver/gatewayserver.go
```diff
@@ -659,18 +659,17 @@ func (gs *GatewayServer) updateConnStats(conn connectionEntry) {
 	logger := log.FromContext(ctx)
 
 	var (
-		lastUpdate io.LastUpdate
-		traffic    io.Traffic
+		lastUpdateUp, lastUpdateDown, lastUpdateStatus time.Time
+		trafficUp, trafficDown, trafficStatus          bool
```
johanstokking: nit; `newUp`, `newDown`, `newStatus`

pkg/gatewayserver/gatewayserver.go
```diff
@@ -659,18 +659,17 @@ func (gs *GatewayServer) updateConnStats(conn connectionEntry) {
...
 	defer func() {
 		logger.Debug("Delete connection stats")
 
 		// Unconditionally clear gateway status when disconnecting.
 		// Required for frontends without support for Status (e.g. BasicStation)
-		traffic.Status = true
-
-		err := gs.ClearConnectionStats(conn.Connection, traffic)
+		trafficStatus = true
+		err := gs.ClearConnectionStats(conn.Connection, trafficUp, trafficDown, trafficStatus)
```
johanstokking: nit; `gs.ClearConnectionStats(conn.Connection, trafficUp, trafficDown, true)`

pkg/gatewayserver/redis/registry.go
```diff
@@ -32,36 +31,37 @@ type GatewayConnectionStatsRegistry struct {
...
 				}
 			}
-			if update.Down {
+			if updateDown {
 				if err := p.Del(r.key(down, uid)).Err(); err != nil {
 					return err
 				}
 			}
-			if update.Status || (!update.Up && !update.Down) {
+			if updateStatus || (!updateUp && !updateDown) {
```
johanstokking: Hmm, what is the use case for this?

pkg/gatewayserver/redis/registry.go
```diff
@@ -112,45 +112,43 @@ func (r *GatewayConnectionStatsRegistry) Get(ctx context.Context, ids ttnpb.Gate
...
-			result.RoundTripTimes = stats.RoundTripTimes
-		} else {
-			return nil, errNotFound
+		if err = ttnredis.UnmarshalProto(retrieved[0].(string), stats); err != nil {
+			return nil, errInvalidStats.WithAttributes("type", "uplink").WithCause(err)
+		}
+
+		if retrieved[2] == nil {
+			result.ConnectedAt = stats.ConnectedAt
+			result.Protocol = stats.Protocol
```
johanstokking: I meant that you get to this case below, so why set these fields here?

pkg/gatewayserver/registry.go
```diff
@@ -26,5 +25,5 @@ type GatewayConnectionStatsRegistry interface {
 	// Get returns connection stats for a gateway.
 	Get(ctx context.Context, ids ttnpb.GatewayIdentifiers) (*ttnpb.GatewayConnectionStats, error)
 	// Set sets or clears the connection stats for a gateway.
-	Set(ctx context.Context, ids ttnpb.GatewayIdentifiers, stats *ttnpb.GatewayConnectionStats, update io.Traffic) error
+	Set(ctx context.Context, ids ttnpb.GatewayIdentifiers, stats *ttnpb.GatewayConnectionStats, up bool, down bool, status bool) error
```
johanstokking: `up, down, status bool`

pkg/gatewayserver/redis/registry.go
neoaggelos: This looks like something trivial but it is not actually too trivial if we want to be precise. The `ConnectedAt` and `Protocol` fields are considered to be part of the gateway status. However, basicstation does not support statuses, so these would never be returned in the response. The code above is a not too clean hack for this.

If we want to be very precise, we should probably have `UpConnectedAt` and `DownConnectedAt` fields, but that breaks many things.

pkg/gatewayserver/redis/registry.go
neoaggelos: Gateway will not be shown as connected before any actual traffic occurs.

pkg/gatewayserver/registry.go
neoaggelos: Thought that it wouldn't play well with godoc, will update

pkg/gatewayserver/redis/registry.go
johanstokking: So then it's a responsibility of the frontend to generate status messages. It's definitely not the responsibility of the persistency layer to implement work-arounds for specific frontends.

pkg/gatewayserver/redis/registry.go
johanstokking: Indeed. If the gateway is connected, it's online

pkg/gatewayserver/redis/registry.go
neoaggelos: This is now handled on new gateway connections.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -654,18 +654,24 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
+		lastUpdateUp, lastUpdateDown, lastUpdateStatus time.Time
+		newUp, newDown, newStatus                      bool
+	)
+
 	defer func() {
 		logger.Debug("Delete connection stats")
-		err := gs.statsRegistry.Set(ctx, conn.Gateway().GatewayIdentifiers, nil)
+
+		// Unconditionally clear gateway status when disconnecting.
+		// Required for frontends without support for Status (e.g. BasicStation)
```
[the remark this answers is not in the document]
neoaggelos: Comment is no longer relevant, removed

pkg/gatewayserver/redis/registry.go
```diff
@@ -48,44 +48,35 @@ func (r *GatewayConnectionStatsRegistry) Set(ctx context.Context, ids ttnpb.Gate
...
-				if kind.update {
-					if err := p.Del(kind.key).Err(); err != nil {
-						return err
-					}
-				}
-			}
-			return nil
-		})
+		setter = func(p redis.Pipeliner, key string, stats *ttnpb.GatewayConnectionStats) error {
+			return p.Del(key).Err()
```
johanstokking: In pipeliner, you only get the errors after executing the pipeline

pkg/gatewayserver/redis/registry.go
```diff
@@ -48,44 +48,35 @@ func (r *GatewayConnectionStatsRegistry) Set(ctx context.Context, ids ttnpb.Gate
...
+		for _, kind := range []struct {
+			key    string
+			update bool
+		}{
+			{r.key(up, uid), updateUp},
+			{r.key(down, uid), updateDown},
+			{r.key(status, uid), updateStatus},
+		} {
+			if kind.update {
+				if err := setter(p, kind.key, stats); err != nil {
```
johanstokking: Why not check `stats != nil` here?

pkg/gatewayserver/gatewayserver.go
```diff
@@ -654,19 +654,24 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
+		lastUpdateUp, lastUpdateDown, lastUpdateStatus time.Time
+		newUp, newDown, newStatus                      bool
+	)
+
 	defer func() {
 		logger.Debug("Delete connection stats")
-		err := gs.statsRegistry.Set(ctx, conn.Gateway().GatewayIdentifiers, nil)
-		if err != nil {
+
+		if err := gs.ClearConnectionStats(conn.Connection, newUp, newDown, true); err != nil {
```
johanstokking: What I still don't entirely understand, is that status is deleted regardless of whether this connection ever handled a status message? Shouldn't that be only if there has been a status message at all here?

neoaggelos: The `status` is used to keep the `ConnectedAt` and `Protocol` fields as well. If we do not delete this on disconnect, then the gateway will still show up as connected.

johanstokking: OK well I guess that is the way to go then

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -1041,9 +1047,14 @@ func TestGatewayServer(t *testing.T) {
...
 							}
 
+							time.Sleep(2 * timeout)
+
 							conn, ok := gs.GetConnection(ctx, ids)
 							a.So(ok, should.BeTrue)
-							gs.UpdateConnectionStats(ctx, conn)
+							a.So(conn.Stats(), should.NotBeNil)
+							if config.Stats != nil {
+								gs.UpdateConnectionStats(conn, true, true, true)
```
johanstokking: Error checking

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -1307,9 +1318,14 @@ func TestGatewayServer(t *testing.T) {
...
```
johanstokking: Idem

Issue #2102, opened 2020-03-09
Application Server discards written keys


bafonins:
#### Summary

NS/AS/JS handle update keys request inconsistently for users that have only `RIGHT_APPLICATION_DEVICES_WRITE_KEYS`, but no `RIGHT_APPLICATION_DEVICES_READ_KEYS`. Namely, NS and JS returns the updated keys, while AS ommits them in the response

#### Steps to Reproduce

<details>
<summary>Details</summary>

Update keys in JS with the following payload:
```
{
   "end_device":{
      "ids":{
         "device_id":"lw-100-otaa",
         "application_ids":{
            "application_id":"admin-app"
         },
         "join_eui":"0102030405060708",
         "dev_eui":"0000100000000000"
      },
      "root_keys":{
         "app_key":{
            "key":"F535A60D4BC3C522C5E404C052C04FC2"
         }
      }
   },
   "field_mask":{
      "paths":[
         "root_keys.app_key.key",
         "ids.device_id",
         "ids.application_ids.application_id"
      ]
   }
}
```
See response (note the `app_key` included in the response payload): 
```
{
   "ids":{
      "device_id":"lw-100-otaa",
      "application_ids":{
         "application_id":"admin-app"
      },
      "dev_eui":"0000100000000000",
      "join_eui":"0102030405060708"
   },
   "created_at":"2020-03-09T09:25:17.289544Z",
   "updated_at":"2020-03-09T10:23:38.148097Z",
   "root_keys":{
      "app_key":{
         "key":"F535A60D4BC3C522C5E404C052C04FC2"
      }
   }
}
```

Update keys in NS with the following payload:
```
{
   "end_device":{
      "session":{
         "keys":{
            "f_nwk_s_int_key":{
               "key":"9EE8A50FED138DDBFB1F95FAA037E764"
            }
         }
      },
      "ids":{
         "device_id":"lw-100-abp",
         "application_ids":{
            "application_id":"admin-app"
         }
      }
   },
   "field_mask":{
      "paths":[
         "session.keys.f_nwk_s_int_key.key",
         "ids.device_id",
         "ids.application_ids.application_id"
      ]
   }
}
```
See response (note the `f_nwk_s_int_key` included in the response payload as well nullable `dev_addr` in `session` and proper address in `ids`): 
```
{
   "ids":{
      "device_id":"lw-100-abp",
      "application_ids":{
         "application_id":"admin-app"
      },
[...]

PR #2117, opened 2020-03-10
Require key read rights on key set

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2102

#### Changes

- Check for key read rights when setting keys in registries

#### Checklist

### DIFF
--- a/pkg/applicationserver/grpc_deviceregistry.go
+++ b/pkg/applicationserver/grpc_deviceregistry.go
@@ -123,6 +123,13 @@ func (r asEndDeviceRegistryServer) Set(ctx context.Context, req *ttnpb.SetEndDev
 		return nil, err
 	}
 	if ttnpb.HasAnyField(req.FieldMask.Paths,
+		"session.keys.session_key_id",
+	) {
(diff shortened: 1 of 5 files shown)


bafonins:
The idea of https://github.com/TheThingsNetwork/lorawan-stack/issues/2102 is not to require both the read and write rights to set the keys, but to make the api responses consistent. Currently, if the user has **only** the write right, only AS hides the key from the response payload:
```
   "session":{
      "keys":{
         "app_s_key":{

         }
      }
```
While JS and NS return the actual key values in the response payload (see examples in the issue).

With this implementation, the users with write/no-read rights cannot set the end device keys. imo, this is incorrect and we should allow the users to set the keys as long as they have the `RIGHT_APPLICATION_DEVICES_WRITE_KEYS` right.

rvolosatovs:
1. AS, NS and JS registries all function exactly the same in this regard, so what you're seeing is either specific to your usage of the API in the console or a bug.
2. All `SET` operations return the values actually being `SET`. You cannot read any data from the registry via `SET`. Hiding the keys being written from the caller makes no sense, because the caller already knows them, they are supplied in the request plain text and returned plain text to the caller, just like any other field.

rvolosatovs:
See https://github.com/TheThingsNetwork/lorawan-stack/issues/2102#issuecomment-597276293

PR #2121, opened 2020-03-10
Move config from Dockerfile to YAML and docker-compose

### LINKED ISSUE #2120: Cannot change some configuration values when using Docker
#### Summary

The Dockerfile defines some stack config variables as environment variables, e.g. `TTN_LW_REDIS_ADDRESS`. These are passed to the stack whatever the rest of the configuration is. Since environment variables have higher precedence over config files, this means that we cannot change these values

#### Steps to Reproduce

Example for the database URI:

1. Change the database URI on ttn-lw-stack.yml
2. Run the stack
3. Database URI is still `postgres://root@cockroach...` because the `TTN_LW_IS_DATABASE_URI` environment variable from the Dockerfile takes precedence.

#### What do you see now?

The Dockerfile defines environment variables.

This was useful when we did all configuration with env files. This is no longer the case.

#### Environment

3.6.0 docker

Move environment and health checks away from the Dockerfile and into the stack and the docker-compose config files.

yes

neoaggelos:
#### Summary

Closes #2120

#### Changes

- ~Move healthcheck config to docker-compose.yml~
- ~Drop environment variables from Dockerfile~
- Configure databases and blob directory with environment variables when deploying with Docker, because Dockerfile overrides them anyway

#### Checklist

### DIFF
--- a/doc/content/guides/getting-started/configuration.md
+++ b/doc/content/guides/getting-started/configuration.md
@@ -44,11 +44,13 @@ Before we go into the details of {{% tts %}}'s configuration, we'll take a look 
 
 The `ports` section exposes {{% tts %}}'s ports to the world. Port `80` and `443` are mapped to the internal HTTP and HTTPS ports. The other ports have a direct mapping. If you don't need support for gateways and applications that don't support TLS, you can remove ports starting with `188`.
 
+In the `environment` section, we configure the databases used by the stack. We will set these to the CockroachDB and Redis instances that are defined in the `docker-compose.yml` above.
+
 ```yaml
 stack:
   image: 'thethingsnetwork/lorawan-stack:<the tag>'
-  entrypoint: 'ttn-lw-stack'
-  command: 'start all -c /config/ttn-lw-stack.yml'
+  entrypoint: 'ttn-lw-stack -c /config/ttn-lw-stack.yml'
+  command: 'start'
   restart: 'unless-stopped'
   depends_on:
     - 'cockroach'
@@ -72,6 +74,10 @@ stack:
     - '1887:1887'
     - '8887:8887'
     - '1700:1700/udp'
+  environment:
+    TTN_LW_BLOB_LOCAL_DIRECTORY: '/srv/ttn-lorawan/public/blob'
+    TTN_LW_REDIS_ADDRESS: 'redis:6379'
+    TTN_LW_IS_DATABASE_URL: 'postgres://root@cockroach:26257/ttn_lorawan?sslmode=disable'
 
   # If using (self) signed certificates
   # secrets:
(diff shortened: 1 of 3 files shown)


KrishnaIyer:
@neoaggelos : This will break existing deployments and we can't actually do this.

neoaggelos:
> @neoaggelos : This will break existing deployments and we can't actually do this.

Alright, this is understandable.

For the short-term, I suppose we can just require that when using Docker, the database URLs are set via environment variables. I'm changing this PR to do that instead.

For the long term, would it be actually worth to have a config option that specifies whether the environment variables are taking precedence over the rest of the configuration? That means we can keep the environment variables and not break anything, but also enable people to use the config file  .

doc/content/guides/getting-started/configuration.md
```diff
@@ -44,11 +44,13 @@ Before we go into the details of {{% tts %}}'s configuration, we'll take a look
 
 The `ports` section exposes {{% tts %}}'s ports to the world. Port `80` and `443` are mapped to the internal HTTP and HTTPS ports. The other ports have a direct mapping. If you don't need support for gateways and applications that don't support TLS, you can remove ports starting with `188`.
 
+In the `environment` section, we configure the databases used by the stack. We will set these to the CockroachDB and Redis instances that are defined in the `docker-compose.yml` above.
```
KrishnaIyer: `CockroachDB` or `Postgres`

docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
   stack:
     image: thethingsnetwork/lorawan-stack
     entrypoint: ttn-lw-stack -c /config/ttn-lw-stack.yml
-    command: start
+    command: start all
```
KrishnaIyer: Why? `start` already starts all the basic services right?

doc/content/guides/getting-started/configuration.md
neoaggelos: I can add this as a note, but we do not actually mention PostgreSQL at the getting started guide at all.

docker-compose.yml
neoaggelos: Reverted

doc/content/guides/getting-started/configuration.md
neoaggelos: Here would probably be a more appropriate place for this

https://github.com/TheThingsNetwork/lorawan-stack/blob/76baee22cb65b6e6f2cd3da781a0c09c797d6d9d/doc/content/guides/getting-started/configuration.md#L10-L12

docker-compose.yml
neoaggelos: Also removed from the getting started guide for consistency

docker-compose.yml
```diff
@@ -17,7 +17,7 @@ services:
   #   image: postgres
   #   restart: unless-stopped
   #   environment:
-  #     - POSTGRES_PASSWORD=
+  #     - POSTGRES_PASSWORD=root
   #     - POSTGRES_USER=root
   #     - POSTGRES_DB=${DEV_DATABASE_NAME:-ttn_lorawan}
```
htdvisser: Can remove `DEV_DATABASE_NAME` here if `stack` is hardcoded to `ttn_lorawan`.

PR #2122, opened 2020-03-11
Client initiated logout and new client auth flow

kschiffer:
#### Summary
References #1422, #1086 
Follow up of #2012 

This PR will add client initiated logout to the oauth server. It will enable logout propagation from the webui clients to the oauth provider, basically tying the authentication of our official clients (console, account app) to the authentication of the auth provider. 

#### Changes
- Add `logout_redirect_uris` to client schema and update `dev:init` mage target to set this value for the console
- Add default redirect URI config to the auth server (defaults to console for now)
- Refactor oauthclient util for the new auth flow
- Refactor oauth server logout endpoint and routing for the new auth flow
- Remove unnecessary authenticated views from the oauth app
- Remove login view from console and automatically redirect to oauth login instead
- Remove landing view, since there are now no unauthenticated views for the console anymore
- Add a utility to create frontend defined error objects
- Refactor `withAuth` util component to also handle insufficient rights issues
- Refactor init logic in the store; moving insufficient rights check to the `withAuth` utility
- Improve styling and logic of error view
- Minor related styling fixes

#### Notes for Reviewers
The implementation is inspired by OpenID connect's client initiated logout functionality.

- We will use the OAuth server frontend now solely to perform authentication and remove the (very limited) authenticated view that we had before. All necessary user management will be done by the account app client, using an access token.
- The oauth server exposes a logout endpoint that takes an access token and optional post logout redirect URI as argument. The oauth provider will then terminate all sessions associated with the user that the access token belongs to. Later on, we might want to associate the session that was used to generate the access token with it and terminate only this session.
  - CSRF protection on the oauth logout endpoint is not necessary since the endpoint expects the access token to be present
  - The `post_logout_redirect_uri` needs to be previously registered in the client (similar to the login redirect URI) and defines, where the logout endpoint will redirect the user after successful logout. Right now, it will just redirect back to the console root, but it could be used for specific post logout views
  - In case of standalone (no redirect) logins into the oauth server, we will define a default redirect (currently the console, but the idea is to use the account app, once it is merged)
    - Alternatively, we could also disable logins that are not client initiated
- Logging in and out of the oauth provider will be intiated via our official clients (console, account app))
  - Console and account app will automatically initiate the auth flow when there is no access token present, or the access token is invalid, without any intermediate client login view
  - To log out, the clients will clear their current access token from the local storage and query their (CSRF protected) logout endpoint, which will remove the auth cookie and return the defined logout endpoint of the oauth server using the access token and redirect URI. The client can then redirect to this URI to also terminate the session within the oauth server (which is what we will do by default)

In order to test the setup in dev environment, it is necessary to set two new settings accordingly:
```
export TTN_LW_IS_OAUTH_DEFAULT_REDIRECT_URI="http://localhost:8080/console"
export TTN_LW_CONSOLE_OAUTH_LOGOUT_URL="http://localhost:8080/oauth/logout"
```
Also the console client needs to be reinitiated: `mage dev:dbStop dev:dbErase dev:dbStart dev:initStack`

Very sorry for the text-wall but the auth specifics are quite intricate, so I wanted to provide a clear summary.

#### Checklist


api/client.proto
```diff
@@ -56,22 +56,26 @@ message Client {
   // If the authorization request does not pass a redirect URI, the first one
   // from this list is taken.
   repeated string redirect_uris = 9 [(gogoproto.customname) = "RedirectURIs"];
+  // The allowed logout redirect URIs against which client initiated logout
+  // requests are checked. If the authorization request does not pass a redirect
+  // URI, the first one from this list is taken.
+  repeated string logout_redirect_uris = 10 [(gogoproto.customname) = "LogoutRedirectURIs"];
```
htdvisser: Changing the numbers of existing fields breaks API. Use a unused number for new fields.

cmd/ttn-lw-stack/commands/is_db_create_oauth_client.go
```diff
@@ -93,6 +97,7 @@ var (
 				"name",
 				"secret",
 				"redirect_uris",
+				"logout-redirect-uris",
```
htdvisser: Field masks use underscores

pkg/oauth/user.go
```diff
@@ -207,16 +211,53 @@ func (s *server) Login(c echo.Context) error {
...
+		return err
+	}
+	if err = s.store.DeleteAccessToken(ctx, id); err != nil {
 		return err
 	}
 	s.removeAuthCookie(c)
-	return c.NoContent(http.StatusNoContent)
+	events.Publish(evtUserLogout(ctx, at.UserIDs, nil))
+	for _, session := range sessions {
+		if err = s.store.DeleteSession(ctx, &at.UserIDs, session.SessionID); err != nil {
```
htdvisser: I don't think this is what a user wants. If I logout using my phone, I don't want to also be logged out from my laptop.

htdvisser:
Re-requesting myself for a more in-depth review later.

pkg/oauth/user.go
kschiffer: Right, I went with the easy way for now. To be able to do this, we'd need to know the session that was used when obtaining the access token. To do this, we could either:
- pass the session id to the console together with the access token
- store the session id in the access token store when creating it

pkg/webui/console/store/reducers/user.js
```diff
@@ -51,11 +52,10 @@ const user = function(state = defaultState, { type, payload }) {
         ...state,
         rights: payload,
       }
-    case LOGOUT_SUCCESS:
+    case LOGOUT_FAILURE:
```
bafonins: With this change, the user won't be logged out from the console automatically on an unauthenticated error. The are still places where `logoutSuccess` is dispatched, see [L78](https://github.com/TheThingsNetwork/lorawan-stack/blob/feature/1422-global-logout/pkg/webui/console/store/middleware/logics/lib.js#L78) and [L95](https://github.com/TheThingsNetwork/lorawan-stack/blob/feature/1422-global-logout/pkg/webui/console/store/middleware/logics/events.js#L95)

pkg/webui/lib/components/with-auth.js
```diff
@@ -13,57 +13,124 @@
...
+
+    let err
+
+    if (error) {
+      err = error
+    } else if (
+      // Check whether the user has at least basic rights, without which it
+      // makes no sense to access the console
+      Boolean(user) &&
+      !isAdmin &&
```
bafonins: the admin account can be suspended/rejected as well. the only allowed operation for a suspended admin account is to list and delete other accounts. Do we want to allow that? 

Also, while checking this I also found out that you are not logged out of the console when you delete your self. Most probably this is related to https://github.com/TheThingsNetwork/lorawan-stack/pull/2122/files#r391668222

Also, see https://github.com/TheThingsNetwork/lorawan-stack/issues/2144

johanstokking:
What is the status here?

kschiffer:
Was superseded by #2146. I'll delete the branch.

PR #2127, opened 2020-03-11
Include pending downlink queue in invalidation

rvolosatovs:
#### Summary

References #1954 

#### Changes

- Include the pending session queue in NS queue invalidation

#### Notes for Reviewers

Reported by @adriansmares 
Testing will be covered in #2110 due to uplink test refactor.

#### Checklist

### DIFF
--- a/pkg/networkserver/grpc_gsns.go
+++ b/pkg/networkserver/grpc_gsns.go
@@ -1151,6 +1151,7 @@ func (ns *NetworkServer) handleJoinRequest(ctx context.Context, up *ttnpb.Uplink
 		[]string{
 			"frequency_plan_id",
 			"lorawan_phy_version",
+			"pending_session.queued_application_downlinks",
 			"recent_uplinks",
 			"session.queued_application_downlinks",
 		},
@@ -1159,7 +1160,7 @@ func (ns *NetworkServer) handleJoinRequest(ctx context.Context, up *ttnpb.Uplink
 				logger.Warn("Device deleted during join-request handling, drop")
 				return nil, nil, errOutdatedData
 			}
-			invalidatedQueue = stored.GetSession().GetQueuedApplicationDownlinks()
+			invalidatedQueue = append(stored.GetSession().GetQueuedApplicationDownlinks(), stored.GetPendingSession().GetQueuedApplicationDownlinks()...)
 			stored.PendingMACState = macState
 			stored.RecentUplinks = appendRecentUplink(stored.RecentUplinks, up, recentUplinkCount)
 			return stored, []string{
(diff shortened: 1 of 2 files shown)


pkg/networkserver/grpc_gsns.go
```diff
@@ -1159,7 +1160,7 @@ func (ns *NetworkServer) handleJoinRequest(ctx context.Context, up *ttnpb.Uplink
 				logger.Warn("Device deleted during join-request handling, drop")
 				return nil, nil, errOutdatedData
 			}
-			invalidatedQueue = stored.GetSession().GetQueuedApplicationDownlinks()
+			invalidatedQueue = append(stored.GetSession().GetQueuedApplicationDownlinks(), stored.GetPendingSession().GetQueuedApplicationDownlinks()...)
```
johanstokking: We should be sure not to change `stored.Session.QueuedApplicationDownlinks` after, because that will change the contents of `invalidatedQueue`

Why not initialize a new slice here?

rvolosatovs: 1. `stored.Session.QueuedApplicationDownlinks` must not be changed during a join-request (and neither should the pending session counterpart)
2. Your statement is not true https://play.golang.org/p/lpK2kCKCTgc

johanstokking: 1. "must be changed"; indeed, `append()` changes it, see 2. Or do you mean that it will not be changed? My point is that after this line, `stored.Session.QueuedApplicationDownlinks` is dirty
2. Your example is a reallocation which involves copy. If you stay in the capacity, the original memory is used. See https://play.golang.org/p/jUQNTHRAo92 and https://blog.golang.org/go-slices-usage-and-internals (section **Growing slices**)

rvolosatovs: `stored.Session.QueuedApplicationDownlinks` and `stored.PendingSession.QueuedApplicationDownlink` *slices* are read-only and never mutated in `handleJoinRequest`.
1. (irrelevant here, but good to note) `len(stored.Session.QueuedApplicationDownlinks) == cap(stored.Session.QueuedApplicationDownlinks)`
2. If `stored.PendingSession.QueuedApplicationDownlink` is empty, `invalidatedQueue` is the same as `stored.Session.QueuedApplicationDownlinks` and reference the same array under the hood. We never modify that array in this function, so that's fine.
3. If `stored.PendingSession.QueuedApplicationDownlink` is not empty, `stored.Session.QueuedApplicationDownlinks` is copied and copy of `stored.PendingSession.QueuedApplicationDownlink` is appended to a new array under the hood, due to (1.)
4. Assuming (1.) would not be `true` and `cap(stored.Session.QueuedApplicationDownlinks) > len(stored.Session.QueuedApplicationDownlinks)` and `len(stored.PendingSession.QueuedApplicationDownlinks) > 0` and `len(stored.PendingSession.QueuedApplicationDownlinks)+len(stored.Session.QueuedApplicationDownlinks) <= cap(stored.Session.QueuedApplicationDownlinks)`, `append` would modify the array referenced by `stored.Session.QueuedApplicationDownlinks` under the hood, but would not modify neither the slice itself, nor the contents of it, because, again,  we never modify that array in this function, so that's fine.

PR #2131, opened 2020-03-11
util: Remove hardware specific values from BS router_config

### LINKED ISSUE #2130: Basic Station Integration: router_config should not include hardware-specific configuration
#### Summary

In the Basic Station protocol, the LNS sends a configuration structure router_config down to the gateway. Upon reception, Station merges the received configuration with the local station.conf file where the server-side router_config has precedence. TTN servers include configuration inside the LNS-side router_config structure which is highly hardware specific:
`clksrc`
`radio_x.rssi_offset`
`radio_x.tx_enable`

This breaks the configuration on platforms, which don't comply with these assumptions.

#### Steps to Reproduce

1. Connect a Basic Station gateway.
2. Send the `version` message.
3. Check the returned JSON.

#### What do you see now?

Example JSON:
```
{"msgtype":"router_config","NetID":null,"JoinEui":null,"region":"EU863","hwspec":"sx1301/1","freq_range":[863000000,870000000],"DRs":[[12,125,0],[11,125,0],[10,125,0],[9,125,0],[8,125,0],[7,125,0],[7,250,0],[0,0,0],[0,0,0],[0,0,0],[0,0,0],[0,0,0],[0,0,0],[0,0,0],[0,0,0],[0,0,0]],"sx1301_conf":[{"lorawan_public":true,"clksrc":1,"antenna_gain":0,"radio_0":{"enable":true,"freq":867500000,"rssi_offset":-166,"tx_enable":true},"radio_1":{"enable":true,"freq":868500000,"rssi_offset":-166,"tx_enable":false},"chan_multiSF_0":{"enable":true,"radio":1,"if":-400000},"chan_multiSF_1":{"enable":true,"radio":1,"if":-200000},"chan_multiSF_2":{"enable":true,"radio":1,"if":0},"chan_multiSF_3":{"enable":true,"radio":0,"if":-400000},"chan_multiSF_4":{"enable":true,"radio":0,"if":-200000},"chan_multiSF_5":{"enable":true,"radio":0,"if":0},"chan_multiSF_6":{"enable":true,"radio":0,"if":200000},"chan_multiSF_7":{"enable":true,"radio":0,"if":400000},"chan_Lora_std":{"enable":true,"radio":1,"if":-200000,"bandwidth":250000,"spread_factor":7},"chan_FSK":{"enable":true,"radio":1,"if":300000,"bandwidth":125000,"datarate":50000}}],"nocca":true,"nodc":true,"nodwell":true,"MuxTime":1586337434.8006055}
```

- [x]  `clksrc`, `radio_x.rssi_offset`, `radio_x.tx_enable` should not be present in the returned config.
- [ ] `antenna_gain`should only be returned when explicitly set by the user. Moved here; https://github.com/TheThingsNetwork/lorawan-stack/issues/4191

#### Environment

`v3.7.0`

This required modification of the frequency plans repository.

@KrishnaIyer will take this.

KrishnaIyer:
#### Summary

Closed #2130 

#### Changes

- Add `omitempty` to necessary fields and set type-default values to them.
- Update tests.

#### Notes for Reviewers

Pending testing on metal.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,8 +26,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
-- Hardware specific fields from Basic Station router config.
-
 ### Fixed
 
 - Telemetry and events for gateway statuses.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- Hardware specific fields from Basic Station router config.
+
 ### Fixed
 
 - Telemetry and events for gateway statuses.
--- a/pkg/pfconfig/basicstationlns/basicstationlns.go
+++ b/pkg/pfconfig/basicstationlns/basicstationlns.go
@@ -117,12 +117,12 @@ func GetRouterConfig(bandID string, fps map[string]*frequencyplans.FrequencyPlan
 		sx1301Conf.Radios[0].TxFreqMin = 0
 		sx1301Conf.Radios[0].TxFreqMax = 0
 		// Remove hardware specific values that are handled by the gateway itself.
-		sx1301Conf.ClockSource = 0
+		sx1301Conf.ClockSource = nil
 		sx1301Conf.TxLUTConfigs = nil
 		for i := range sx1301Conf.Radios {
 			sx1301Conf.Radios[i].Type = ""
-			sx1301Conf.Radios[i].TxEnable = false
-			sx1301Conf.Radios[i].RSSIOffset = 0
+			sx1301Conf.Radios[i].TxEnable = nil
+			sx1301Conf.Radios[i].RSSIOffset = nil
 		}
 		if err != nil {
 			return RouterConfig{}, err
--- a/pkg/pfconfig/basicstationlns/basicstationlns.go
+++ b/pkg/pfconfig/basicstationlns/basicstationlns.go
@@ -116,10 +116,13 @@ func GetRouterConfig(bandID string, fps map[string]*frequencyplans.FrequencyPlan
 		// These fields are not defined in the v1.5 ref design https://doc.sm.tc/station/gw_v1.5.html#rfconf-object and would cause a parsing error.
 		sx1301Conf.Radios[0].TxFreqMin = 0
 		sx1301Conf.Radios[0].TxFreqMax = 0
-		// Remove hardware specific values that are not necessary.
+		// Remove hardware specific values that are handled by the gateway itself.
+		sx1301Conf.ClockSource = 0
 		sx1301Conf.TxLUTConfigs = nil
 		for i := range sx1301Conf.Radios {
 			sx1301Conf.Radios[i].Type = ""
+			sx1301Conf.Radios[i].TxEnable = false
+			sx1301Conf.Radios[i].RSSIOffset = 0
 		}
 		if err != nil {
 			return RouterConfig{}, err
--- a/pkg/pfconfig/shared/shared.go
+++ b/pkg/pfconfig/shared/shared.go
@@ -86,7 +86,7 @@ func (m orderedMap) MarshalJSON() ([]byte, error) {
 func (c SX1301Config) MarshalJSON() ([]byte, error) {
 	var m orderedMap
 	m.add("lorawan_public", c.LoRaWANPublic)
-	m.add("clksrc", c.ClockSource)
+	m.add("clksrc,omitempty", c.ClockSource)
 	m.add("antenna_gain", c.AntennaGain)
 	if c.LBTConfig != nil {
 		m.add("lbt_cfg", *c.LBTConfig)
--- a/pkg/pfconfig/shared/shared.go
+++ b/pkg/pfconfig/shared/shared.go
@@ -33,7 +33,7 @@ import (
 // SX1301Config contains the configuration for the SX1301 concentrator.
 type SX1301Config struct {
 	LoRaWANPublic       bool
-	ClockSource         uint8
+	ClockSource         uint8 `json:"omitempty"`
 	AntennaGain         float32
 	LBTConfig           *LBTConfig
 	Radios              []RFConfig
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- Hardware specific fields from Basic Station router config.
```
htdvisser: I think this section is only for removed features, not for bug fixes

pkg/pfconfig/shared/shared.go
```diff
@@ -33,7 +33,7 @@ import (
 // SX1301Config contains the configuration for the SX1301 concentrator.
 type SX1301Config struct {
 	LoRaWANPublic       bool
-	ClockSource         uint8
+	ClockSource         uint8 `json:"omitempty"`
```
htdvisser: 1. Your `json` tag now renames this field to `omitempty`
2. This has no effect, because `SX1301Config` has a `MarshalJSON` implementation. That's where you need to change this.
3. You need to make this a `*uint8` so that we can make a distinction between 0 and "omit"

pkg/pfconfig/shared/shared.go
```diff
@@ -209,8 +209,8 @@ type RFConfig struct {
 	Enable      bool    `json:"enable"`
 	Type        string  `json:"type,omitempty"`
 	Frequency   uint64  `json:"freq"`
-	RSSIOffset  float32 `json:"rssi_offset"`
-	TxEnable    bool    `json:"tx_enable"`
+	RSSIOffset  float32 `json:"rssi_offset,omitempty"`
+	TxEnable    bool    `json:"tx_enable,omitempty"`
```
htdvisser: You need to make these pointers so that we can make a distinction between the zero value and "omit"

pkg/util/pointers/pointers.go
```diff
@@ -0,0 +1,32 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pointers
+
+import "time"
+
+// Bool returns a boolean pointer for the given value.
+func Bool(v bool) *bool { return &v }
```
johanstokking: Let's please not do this

pkg/pfconfig/basicstationlns/basicstationlns.go
```diff
@@ -116,10 +116,13 @@ func GetRouterConfig(bandID string, fps map[string]*frequencyplans.FrequencyPlan
 		// These fields are not defined in the v1.5 ref design https://doc.sm.tc/station/gw_v1.5.html#rfconf-object and would cause a parsing error.
 		sx1301Conf.Radios[0].TxFreqMin = 0
 		sx1301Conf.Radios[0].TxFreqMax = 0
-		// Remove hardware specific values that are not necessary.
+		// Remove hardware specific values that are handled by the gateway itself.
+		sx1301Conf.ClockSource = nil
```
johanstokking: Isn't the underlying problem that we have radio specific fields in the frequency plan?

pkg/util/pointers/pointers.go
rvolosatovs: If this is a one-time thing you could just oneline this. E.g. https://github.com/TheThingsNetwork/lorawan-stack/blob/9be9736f91cc0ea954e7773d675c738dc87ce4fd/cmd/internal/shared/networkserver/config.go#L36

Otherwise just have a non-exported `boolPtr` next to the usage.

KrishnaIyer: Why not? Then I need to define the same `boolPtr ` and functions for each type in 4 different files.

pkg/pfconfig/basicstationlns/basicstationlns.go
KrishnaIyer: Well these fields are valid `SX1301_conf` fields. Issue is that basic station specifically already has a config file on each hardware and hardware specific fields are already present there.

pkg/pfconfig/shared/shared.go
```diff
@@ -285,7 +286,7 @@ func BuildSX1301Config(frequencyPlan *frequencyplans.FrequencyPlan) (*SX1301Conf
 	conf := new(SX1301Config)
 
 	conf.LoRaWANPublic = true
-	conf.ClockSource = frequencyPlan.ClockSource
+	conf.ClockSource = pointers.Uint8(frequencyPlan.ClockSource)
```
rvolosatovs:
```suggestion
	conf.ClockSource = &frequencyPlan.ClockSource
```

pkg/pfconfig/shared/shared_test.go
```diff
@@ -88,25 +89,24 @@ func TestSX1301Conf(t *testing.T) {
 			},
 			SX1301Config{
 				LoRaWANPublic: true,
-				ClockSource:   1,
+				ClockSource:   pointers.Uint8(1),
```
rvolosatovs:
```suggestion
				ClockSource:   func(v uint8) *uint8 { return &v }(1),
```

pkg/pfconfig/shared/shared_test.go
```diff
@@ -88,25 +89,24 @@ func TestSX1301Conf(t *testing.T) {
...
 				AntennaGain:   0,
 				Radios: []RFConfig{
 					{
 						Enable:     true,
 						Type:       "SX1257",
 						Frequency:  867500000,
-						TxEnable:   true,
+						TxEnable:   pointers.Bool(true),
```
rvolosatovs:
```suggestion
						TxEnable:    func(v bool) *bool { return &v }(true),
```

pkg/pfconfig/shared/shared_test.go
```diff
@@ -88,25 +89,24 @@ func TestSX1301Conf(t *testing.T) {
...
 						TxFreqMin:  863000000,
 						TxFreqMax:  870000000,
-						RSSIOffset: -166,
+						RSSIOffset: pointers.Float32(-166),
```
rvolosatovs: This is the only place you actually may want to have a helper function, since there are 2 usages.
However, why not just define `const defaultRSSIOffset = -166` and use `&defaultRSSIOffset` here and below?

pkg/pfconfig/basicstationlns/basicstationlns.go
johanstokking: But isn't that true for other gateways too?

KrishnaIyer: Most gateways use what's provided in the `global_conf.json` right? Some have overrides in `local_conf.json` but not all of them afaik.

johanstokking: What I understand from #2130 is that we send information too specific about a gateway to a gateway. That makes me think that we have fields in our frequency plans that don't belong there. So removing those from BSLNS as you do here is a fix, but it's a fix for the symptoms. We may need to remove these fields from the frequency plans, `pfconfig` and GCS.

pkg/util/pointers/pointers.go
johanstokking: Yes, that is inherent to Go

KrishnaIyer:
Closing this in favor of https://github.com/TheThingsIndustries/lorawan-stack/issues/2018

PR #2148, opened 2020-03-13
Client initiated logout / logout propagation to oauth server

### LINKED ISSUE #1702: Redirect to OAuth Server Logout after Logout from Client
#### Summary

It would be good to redirect to the logout page of the OAuth server after logging out from an OAuth client (console).

#### Why do we need this?

That way the user can also logout from the OAuth server, and later when we have #1422, also revoke access tokens of other logged in OAuth clients.

For example in the console: after you click logout, you're redirected to the console login page, where you can start a new OAuth flow, but since you're still logged in with the OAuth server, you're redirected back immediately.

A redirect to the OAuth server logout page after logout from OAuth clients (such as the console)

#### Environment

v3.3.2

Perhaps extra config for `pkg/web/oauthclient` and a redirect in `HandleLogout`. Not sure if the frontend would pick that up?

kschiffer:
#### Summary
References #1422 
Closes #1702 

This PR will enable logout propagation from our official clients, meaning that the console logout will also result in a deletion of the session in the oauth server that was used to create the access token.

#### Changes
- Add a `LogoutRedirectURI` field to the clients, used to check the validity of the requested redirect during the client-initiated logout
- Update mage targets to set logout redirect URIs
- Add a `LogoutURL` config to oauthclient util, used to trigger a logout of the oauth provider
- Refactor oauthclient and oauth server for the new flow

#### Notes for Reviewers

In order to test the setup in dev environment, it is necessary to set two new settings accordingly:
```
export TTN_LW_IS_OAUTH_DEFAULT_REDIRECT_URI="http://localhost:8080/console"
export TTN_LW_CONSOLE_OAUTH_LOGOUT_URL="http://localhost:8080/oauth/logout"
```
Also the console client needs to be reinitiated: `mage dev:dbStop dev:dbErase dev:dbStart dev:initStack`

The flow works like this:
<details><summary>click to enlarge</summary>

1. The user clicks on a logout button inside the clients web ui
2. The web app will delete the access token
3. The client backend will destroy the auth cookie and return an `op_logout_uri`, containing the access token ID, as well as a desired `post_logout_redirect_uri`
4. (Based on whether logout propagation is desired, ) the web app will redirect the browser to the logout URI
5. The oauth server will retrieve the associated session from the passed `access_token_id` query parameter, delete the associated session of the access token and the access token itself
6. The oauth server will check whether the `post_logout_redirect_uri` has been previously registered and redirect back to this URI (if no `post_logout_redirect_uri` was passed, the first `LogoutRedirectURI` of the client will be used
7. The client can use the redirect URI to show a dedicated post logout view (but the console will simply redirect back to the root path)
8. (Since the access token was deleted, the console will redirect to the login screen, from which the auth flow can be started again)
</details>

I've kept the frontend changes in this PR to a minimum, so we can focus on the backend changes. I will follow up with a PR that does a proper refactor of the webui after this PR has been merged. This will include removing the login view, refactoring `withAuth` util component, removing unnecessary store logic and removing the landing view of the oauth app.

As discussed with @htdvisser, we also plan to to federated logout (meaning logging out all currently authenticated SSO clients associated with the respective oauth session). 
This would necessitate adding an `sso` flag to the client model, which will be used to determine what other clients need to be logged out. Should I also integrate this into this PR or use another scoped one?

#### Checklist


pkg/oauth/user.go
```diff
@@ -207,16 +211,49 @@ func (s *server) Login(c echo.Context) error {
 	return c.NoContent(http.StatusNoContent)
 }
 
+var errInvalidLogoutRedirectURI = errors.DefinePermissionDenied("invalid_logout_redirect_uri", "the redirect URI did not match the one(s) defined in the client")
+var errMissingAccessTokenIDParam = errors.DefinePermissionDenied("missing_param_access_token_id", "Access token ID was not provided")
```
htdvisser: 1. Formatting: you can also put these in a `var` block, and use some newlines to make it more readable:
	```go
	var (
        errInvalidLogoutRedirectURI = errors.DefinePermissionDenied(
			"invalid_logout_redirect_uri", 
			"the redirect URI did not match the one(s) defined in the client",
		)
		errMissingAccessTokenIDParam = // ...
	)
	```
2. Lowercase "Access" in description of the second

pkg/oauth/server.go
```diff
@@ -245,20 +254,22 @@ func (s *server) RegisterRoutes(server *web.Server) {
...
 	api := group.Group("/api", middleware.CSRF())
 	api.POST("/auth/login", s.Login)
-	api.POST("/auth/logout", s.Logout, s.requireLogin)
 	api.GET("/me", s.CurrentUser, s.requireLogin)
 
 	page := group.Group("", middleware.CSRFWithConfig(middleware.CSRFConfig{
 		TokenLookup: "form:csrf",
 	}))
 	page.GET("/login", webui.Template.Handler, s.redirectToNext)
+	page.GET("/logout", s.Logout, s.requireLogin)
```
htdvisser: We may need to replace `requireLogin` with something else, otherwise logout+redirect will fail if the user is already logged out from the OAuth provider.

pkg/oauth/user.go
```diff
@@ -207,16 +211,49 @@ func (s *server) Login(c echo.Context) error {
...
+
 func (s *server) Logout(c echo.Context) error {
 	ctx := c.Request().Context()
-	session, err := s.getSession(c)
```
htdvisser: How about the currenlty active session on the OAuth server? Do we just ignore that now?

pkg/oauth/user.go
```diff
@@ -207,16 +211,49 @@ func (s *server) Login(c echo.Context) error {
...
+	accessTokenID := c.QueryParam("access_token_id")
+	if accessTokenID == "" {
+		return errMissingAccessTokenIDParam
```
htdvisser: So `access_token_id` is required now? How would the user logout from their current (cookie) session on the OAuth server?

pkg/oauth/user.go
```diff
@@ -207,16 +211,49 @@ func (s *server) Login(c echo.Context) error {
...
+	}
+	at, err := s.store.GetAccessToken(ctx, accessTokenID)
 	if err != nil {
 		return err
```
htdvisser: The access token may already be deleted by the OAuth client at this point, right? Then we would get a NotFound error here, while the end result (logging out) would still be successful.

pkg/oauth/user.go
```diff
@@ -207,16 +211,49 @@ func (s *server) Login(c echo.Context) error {
...
 	}
-	events.Publish(evtUserLogout(ctx, session.UserIdentifiers, nil))
-	if err = s.store.DeleteSession(ctx, &session.UserIdentifiers, session.SessionID); err != nil {
+	if err = s.store.DeleteAccessToken(ctx, accessTokenID); err != nil {
 		return err
 	}
 	s.removeAuthCookie(c)
-	return c.NoContent(http.StatusNoContent)
+	events.Publish(evtUserLogout(ctx, at.UserIDs, nil))
+	if err = s.store.DeleteSession(ctx, &at.UserIDs, at.UserSessionID); err != nil {
```
htdvisser: Again, how about the current (cookie) session?

pkg/oauth/user.go
```diff
@@ -207,16 +211,49 @@ func (s *server) Login(c echo.Context) error {
...
+	}
+	var redirectURI = ""
+	redirectParam := c.QueryParam("post_logout_redirect_uri")
+	if redirectParam == "" {
+		redirectURI = client.LogoutRedirectURIs[0]
+		if redirectURI == "" {
+			redirectURI = s.config.UI.MountPath()
+		}
+	} else {
+		redirectURI, err = osin.ValidateUriList(strings.Join(client.LogoutRedirectURIs, ","), redirectParam, ",")
```
htdvisser: Better to use `osin.ValidateUri` instead of joining, and then letting `osin.ValidateUriList` split them again.

pkg/oauth/user.go
```diff
@@ -207,16 +211,49 @@ func (s *server) Login(c echo.Context) error {
...
+		if err != nil {
+			return errInvalidLogoutRedirectURI
```
htdvisser: Add `.WithCause(err)`

pkg/web/oauthclient/logout.go
```diff
@@ -16,45 +16,45 @@ package oauthclient
...
-					UserIDs:   tokenInfo.UserIDs,
-					ClientIDs: tokenInfo.ClientIDs,
-					ID:        tokenInfo.ID,
-				}, creds)
-				if err != nil {
-					log.FromContext(ctx).WithError(err).Error("Could not invalidate access token")
-				}
-			}
+	oc.removeAuthCookie(c)
+	u, err := url.Parse(oc.config.LogoutURL)
```
htdvisser: If there is no `LogoutURL` config, we should keep using the old behavior (the red lines above)

pkg/web/oauthclient/logout.go
```diff
@@ -16,45 +16,45 @@ package oauthclient
...
-				_, err := ttnpb.NewOAuthAuthorizationRegistryClient(cc).DeleteToken(ctx, &ttnpb.OAuthAccessTokenIdentifiers{
-					UserIDs:   tokenInfo.UserIDs,
-					ClientIDs: tokenInfo.ClientIDs,
-					ID:        tokenInfo.ID,
-				}, creds)
-				if err != nil {
-					log.FromContext(ctx).WithError(err).Error("Could not invalidate access token")
-				}
-			}
+	oc.removeAuthCookie(c)
```
htdvisser: Move this right before `c.JSON(http.StatusOK, ...`, otherwise we might prematurely remove the cookie on err.

pkg/web/oauthclient/logout.go
```diff
@@ -16,45 +16,45 @@ package oauthclient
...
+		return err
+	}
+	logoutURL := oc.config.LogoutURL
+	redirectURL := strings.TrimSuffix(oc.config.RootURL, "/")
+	if oauthRootURL, err := url.Parse(oc.config.RootURL); err == nil {
+		rootURL := (&url.URL{Scheme: oauthRootURL.Scheme, Host: oauthRootURL.Host}).String()
+		if strings.HasPrefix(logoutURL, rootURL) {
+			redirectURL = strings.TrimPrefix(redirectURL, rootURL)
 		}
 	}
```
htdvisser: This way of handling URLs is a bit messy, I'm sure we can implement it in a cleaner way.

pkg/oauth/user.go
kschiffer: The idea is that logging out will only work over our official SSO clients. Since the access token and session are connected, we don't need an isolated oauth server logout anymore.
Even if the access token is lost, our clients will automatically obtain a new access token that is connected to the current session in the OAuth server and can thus be used to logout (thus deleting the connected session on the OAuth server).

pkg/oauth/user.go
kschiffer: See my other comment. The access token will be the source for referencing the relevant session now and the auth flows will be initiated by our official (SSO) clients. The oauth server will only handle authentication from now on, without any authenticated views.

pkg/oauth/server.go
```diff
@@ -79,8 +79,9 @@ type FrontendConfig struct {
 
 // Config is the configuration for the OAuth server.
 type Config struct {
-	Mount string   `name:"mount" description:"Path on the server where the OAuth server will be served"`
-	UI    UIConfig `name:"ui"`
+	DefaultRedirectURI string   `name:"default-redirect-uri" description:"The default URI the OAuth server will redirect to after login succeeded"`
```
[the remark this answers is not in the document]
kschiffer: I think then it's best to disable standalone (not client-initiated) logins for the time being. Any other idea?

pkg/web/oauthclient/logout.go
kschiffer: Maybe it would be better to remove the whole relative URL conversion altogether and add another `--logout-redirect-uri https://localhost:8080/console` to our mage init targets. Same for login redirect uris. This whole conversion seems a bit weird. What do you think? Otherwise, I don't really know how to make this cleaner.

pkg/oauth/server_test.go
```diff
@@ -396,16 +397,18 @@ func TestOAuthFlow(t *testing.T) {
...
 				s.res.session = mockSession
-				s.res.user = mockUser
+				s.res.client = mockClient
+				s.res.accessToken = mockAccessToken
 			},
-			Method:       "POST",
-			Path:         "/oauth/api/auth/logout",
-			ExpectedCode: http.StatusNoContent,
+			Method:       "GET",
+			Path:         "/oauth/logout?access_token_id=access-token-id&post_logout_redirect_uri=http://uri/logout-callback",
```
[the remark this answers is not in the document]
kschiffer: This made me realize that we should only proceed with the logout after the redirect URI has been verified. The current implementation would log the user out despite mismatching redirect URI.

pkg/oauth/server.go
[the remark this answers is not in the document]
htdvisser: I'd prefer to keep this PR scoped to one thing, and do other stuff later.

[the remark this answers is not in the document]
kschiffer: Ok, I've reintroduced the old login handler which delets the active session, so we can show the current authenticated view of the oauth server again. Also updated the tests accordingly. We will have two logout handlers then in the interim.

kschiffer:
Blocking #1422

pkg/oauth/user.go
```diff
@@ -207,6 +210,65 @@ func (s *server) Login(c echo.Context) error {
...
+		if redirectParam == "" {
+			redirectURI = client.LogoutRedirectURIs[0]
+		} else {
+			for _, uri := range client.LogoutRedirectURIs {
+				redirectURI, err = osin.ValidateUri(uri, redirectParam)
+			}
+			if err != nil {
+				return errInvalidLogoutRedirectURI.WithCause(err)
+			}
+		}
```
htdvisser, line 252: Move this to the end. For existing clients in existing deployments we won't have a registered logout redirect URI, but users still need to be able to logout from those.

pkg/webui/console/store/reducers/user.js
```diff
@@ -51,11 +51,10 @@ const user = function(state = defaultState, { type, payload }) {
         ...state,
         rights: payload,
       }
-    case LOGOUT_SUCCESS:
+    case LOGOUT_FAILURE:
       return {
         ...state,
-        user: undefined,
-        authInfo: undefined,
+        error: payload,
```
bafonins, line 57: `error` is not selected anywhere. no error is shown on failed logout

pkg/webui/console/store/middleware/logics/init.js
```diff
@@ -49,7 +49,6 @@ const consoleAppLogic = createRequestLogic({
       dispatch(user.getUserRightsSuccess(rights))
     } catch (error) {
       dispatch(user.getUserRightsFailure())
-      dispatch(user.logout())
```
bafonins, line 52: mb not in this pr, but i guess we dont want to just ignore  `getUserRightsFailure` action. otherwise the user cant login without any error ui

kschiffer: Right. This will be caught by refactoring the `withAuth` component to check for these errors in the next PR. For now, I'll just rethrow the error so it will be caught by the error boundary.

kschiffer:
> One thing I noticed:
> 
> 1. Log in
> 2. While in the console, clear cache/local and session storage/cookies
> 3. Initiate an api request
> 4. Be redirected to the login
> 5. Login again
> 6. Try logging out
> 7. Get 403 forbidden for the `/logout` call

This was actually a bug related to the echo csrf middleware we're using, which spammed the browser with token cookies for all the routes that it was called on and hence mixing up tokens during lookup. This is now fixed.

pkg/webui/console/store/reducers/user.js
kschiffer: Indeed. I've included a very basic passing of the error now as interim solution. This will be fleshed out properly [with the next PR](#2318).

pkg/oauth/server.go
```diff
@@ -245,20 +245,23 @@ func (s *server) RegisterRoutes(server *web.Server) {
 
 	api := group.Group("/api", middleware.CSRF())
 	api.POST("/auth/login", s.Login)
-	api.POST("/auth/logout", s.Logout, s.requireLogin)
+	api.POST("/auth/logout", s.Logout)
```
htdvisser, line 248: Won't this cause trouble in `s.Logout` for users that are not logged in?

pkg/oauth/user.go
```diff
@@ -207,6 +210,64 @@ func (s *server) Login(c echo.Context) error {
...
+func (s *server) ClientLogout(c echo.Context) error {
+	ctx := c.Request().Context()
+	accessTokenID := c.QueryParam("access_token_id")
+	var redirectURI string
+	if accessTokenID == "" {
+		return errMissingAccessTokenIDParam
+	}
+	at, err := s.store.GetAccessToken(ctx, accessTokenID)
+	if err != nil {
+		if !errors.IsNotFound(err) {
```
htdvisser, line 233: This can be combined to `if err != nil && !errors.IsNotFound(err) {`

pkg/oauth/user.go
```diff
@@ -207,6 +210,64 @@ func (s *server) Login(c echo.Context) error {
...
+	}
+	if at != nil {
+		client, err := s.store.GetClient(ctx, &at.ClientIDs, &types.FieldMask{Paths: []string{"logout_redirect_uris"}})
+		if err != nil {
+			return err
+		}
+		if err = s.store.DeleteAccessToken(ctx, accessTokenID); err != nil {
+			return err
+		}
+		s.removeAuthCookie(c)
```
htdvisser, line 245: We don't know that the auth cookie corresponds with the session we're deleting. If it doesn't, we may leave a session active while the user thinks it's revoked.

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ns.down.data.schedule.attempt` and `ns.down.join.schedule.attempt` events, which are triggered when Network Server attempts to schedule a respective downlink on Gateway Server.
 - `ns.down.data.schedule.success` and `ns.down.join.schedule.success` events, which are triggered when Network Server successfully schedules a respective downlink on Gateway Server.
 - `ns.down.data.schedule.fail` and `ns.down.join.schedule.fail` events, which are triggered when Network Server fails to schedule a respective downlink on Gateway Server.
+- Console logout is now propagated to the OAuth provider.
+  - This requires a database migration (`ttn-lw-stack is-db migrate`) because of the added columns.
+  - To set the `logout-redirect-uris` for existing clients, the CLI client can be used, e.g.: `ttn-lw-cli clients update console --redirect-uris "https://localhost:1885/console" --redirect-uris "http://localhost:1885/console"`.
```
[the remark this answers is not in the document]
kschiffer: Indeed wrong example. This should just be `logout-redirect-uris` instead of `redirect-uris`.

pkg/oauth/server.go
kschiffer: Indeed. This is a leftover from before https://github.com/TheThingsNetwork/lorawan-stack/pull/2148#discussion_r394793389

pkg/oauth/user.go
kschiffer: I can't think of any realistic scenario where that would be the case. The user has obtained the access token using this auth cookie (session). Since our official clients tie them together, there should be no corruptions or mixups. So I think this is unelectable. In any case we still have the old logout handler, so clients can decide for themselves.

kschiffer: As discussed, the logout will now
- delete the access token
- delete the session that was associated with the access token
- delete the session associated with the current auth cookie of the user (if it differs from the sessions stored in the access token)

pkg/oauth/user.go
```diff
@@ -207,6 +210,76 @@ func (s *server) Login(c echo.Context) error {
...
+		client, err := s.store.GetClient(ctx, &at.ClientIDs, &types.FieldMask{Paths: []string{"logout_redirect_uris"}})
+		if err != nil {
+			return err
+		}
+		if err = s.store.DeleteAccessToken(ctx, accessTokenID); err != nil {
+			return err
+		}
+		events.Publish(evtUserLogout(ctx, at.UserIDs, nil))
+		if err = s.store.DeleteSession(ctx, &at.UserIDs, at.UserSessionID); err != nil {
+			if !errors.IsNotFound(err) {
```
htdvisser, line 245: Can combine `err != nil && !errors.IsNotFound(err)`

PR #2150, opened 2020-03-13
Generate CLI config file

neoaggelos:
#### Summary

References #1240

#### Changes

- Adds a new `ttn-lw-cli use` command that can generate and write configuration file for the CLI, using sane defaults for a v3 cluster.

#### Notes for Reviewers

- This is on the CLI side only. We could add similar functionality to the v3 stack, but this is for a different PR, which is why I am not closing the initial issue.
- Still working on the documentation.

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/config.go
+++ b/cmd/ttn-lw-cli/commands/config.go
@@ -15,39 +15,40 @@
 package commands
 
 import (
+	"fmt"
+
 	"go.thethings.network/lorawan-stack/cmd/internal/commands"
 	conf "go.thethings.network/lorawan-stack/pkg/config"
 	"go.thethings.network/lorawan-stack/pkg/log"
 )
 
 var (
-	clusterHost        = "localhost"
-	clusterGRPCAddress = clusterHost + ":8884"
-	clusterHTTPAddress = "https://" + clusterHost + ":8885"
+	defaultClusterHost = "localhost"
+	defaultInsecure    = false
 )
 
 // Config for the ttn-lw-cli binary.
 type Config struct {
 	conf.Base                          `name:",squash"`
-	CredentialsID                      string `name:"credentials-id" description:"Credentials ID (if using multiple configurations)"`
-	InputFormat                        string `name:"input-format" description:"Input format"`
-	OutputFormat                       string `name:"output-format" description:"Output format"`
-	AllowUnknownHosts                  bool   `name:"allow-unknown-hosts" description:"Allow sending credentials to unknown hosts"`
-	OAuthServerAddress                 string `name:"oauth-server-address" description:"OAuth Server address"`
-	IdentityServerGRPCAddress          string `name:"identity-server-grpc-address" description:"Identity Server address"`
-	GatewayServerEnabled               bool   `name:"gateway-server-enabled" description:"Gateway Server enabled"`
-	GatewayServerGRPCAddress           string `name:"gateway-server-grpc-address" description:"Gateway Server address"`
-	NetworkServerEnabled               bool   `name:"network-server-enabled" description:"Network Server enabled"`
-	NetworkServerGRPCAddress           string `name:"network-server-grpc-address" description:"Network Server address"`
-	ApplicationServerEnabled           bool   `name:"application-server-enabled" description:"Application Server enabled"`
-	ApplicationServerGRPCAddress       string `name:"application-server-grpc-address" description:"Application Server address"`
-	JoinServerEnabled                  bool   `name:"join-server-enabled" description:"Join Server enabled"`
-	JoinServerGRPCAddress              string `name:"join-server-grpc-address" description:"Join Server address"`
-	DeviceTemplateConverterGRPCAddress string `name:"device-template-converter-grpc-address" description:"Device Template Converter address"`
-	DeviceClaimingServerGRPCAddress    string `name:"device-claiming-server-grpc-address" description:"Device Claiming Server address"`
-	QRCodeGeneratorGRPCAddress         string `name:"qr-code-generator-grpc-address" description:"QR Code Generator address"`
-	Insecure                           bool   `name:"insecure" description:"Connect without TLS"`
-	CA                                 string `name:"ca" description:"CA certificate file"`
+	CredentialsID                      string `name:"credentials-id" yaml:"credentials-id" description:"Credentials ID (if using multiple configurations)"`
+	InputFormat                        string `name:"input-format" yaml:"input-format" description:"Input format"`
+	OutputFormat                       string `name:"output-format" yaml:"output-format" description:"Output format"`
+	AllowUnknownHosts                  bool   `name:"allow-unknown-hosts" yaml:"allow-unknown-hosts" description:"Allow sending credentials to unknown hosts"`
+	OAuthServerAddress                 string `name:"oauth-server-address" yaml:"oauth-server-address" description:"OAuth Server address"`
+	IdentityServerGRPCAddress          string `name:"identity-server-grpc-address" yaml:"identity-server-grpc-address" description:"Identity Server address"`
+	GatewayServerEnabled               bool   `name:"gateway-server-enabled" yaml:"gateway-server-enabled" description:"Gateway Server enabled"`
+	GatewayServerGRPCAddress           string `name:"gateway-server-grpc-address" yaml:"gateway-server-grpc-address" description:"Gateway Server address"`
+	NetworkServerEnabled               bool   `name:"network-server-enabled" yaml:"network-server-enabled" description:"Network Server enabled"`
+	NetworkServerGRPCAddress           string `name:"network-server-grpc-address" yaml:"network-server-grpc-address" description:"Network Server address"`
+	ApplicationServerEnabled           bool   `name:"application-server-enabled" yaml:"application-server-enabled" description:"Application Server enabled"`
+	ApplicationServerGRPCAddress       string `name:"application-server-grpc-address" yaml:"application-server-grpc-address" description:"Application Server address"`
+	JoinServerEnabled                  bool   `name:"join-server-enabled" yaml:"join-server-enabled" description:"Join Server enabled"`
+	JoinServerGRPCAddress              string `name:"join-server-grpc-address" yaml:"join-server-grpc-address" description:"Join Server address"`
+	DeviceTemplateConverterGRPCAddress string `name:"device-template-converter-grpc-address" yaml:"device-template-converter-grpc-address" description:"Device Template Converter address"`
+	DeviceClaimingServerGRPCAddress    string `name:"device-claiming-server-grpc-address" yaml:"device-claiming-server-grpc-address" description:"Device Claiming Server address"`
+	QRCodeGeneratorGRPCAddress         string `name:"qr-code-generator-grpc-address" yaml:"qr-code-generator-grpc-address" description:"QR Code Generator address"`
+	Insecure                           bool   `name:"insecure" yaml:"insecure" description:"Connect without TLS"`
+	CA                                 string `name:"ca" yaml:"ca" description:"CA certificate file"`
 }
 
 func (c Config) getHosts() []string {
@@ -71,29 +72,42 @@ func (c Config) getHosts() []string {
 	return getHosts(hosts...)
 }
 
-// DefaultConfig contains the default config for the ttn-lw-cli binary.
-var DefaultConfig = Config{
-	Base: conf.Base{
-		Log: conf.Log{
-			Level: log.InfoLevel,
+// MakeDefaultConfig builds the default config for the ttn-lw-cli binary for a given host.
+func MakeDefaultConfig(clusterHost string, insecure bool) Config {
+	clusterGRPCAddress := fmt.Sprintf("%s:8884", clusterHost)
+	clusterHTTPAddress := fmt.Sprintf("https://%s:8885", clusterHost)
+	if insecure {
+		clusterGRPCAddress = fmt.Sprintf("%s:1884", clusterHost)
+		clusterHTTPAddress = fmt.Sprintf("http://%s:1885", clusterHost)
+	}
+
+	return Config{
+		Base: conf.Base{
+			Log: conf.Log{
+				Level: log.InfoLevel,
+			},
 		},
-	},
-	InputFormat:                        "json",
-	OutputFormat:                       "json",
-	OAuthServerAddress:                 clusterHTTPAddress + "/oauth",
-	IdentityServerGRPCAddress:          clusterGRPCAddress,
-	GatewayServerEnabled:               true,
-	GatewayServerGRPCAddress:           clusterGRPCAddress,
-	NetworkServerEnabled:               true,
-	NetworkServerGRPCAddress:           clusterGRPCAddress,
-	ApplicationServerEnabled:           true,
-	ApplicationServerGRPCAddress:       clusterGRPCAddress,
-	JoinServerEnabled:                  true,
-	JoinServerGRPCAddress:              clusterGRPCAddress,
-	DeviceTemplateConverterGRPCAddress: clusterGRPCAddress,
-	DeviceClaimingServerGRPCAddress:    clusterGRPCAddress,
-	QRCodeGeneratorGRPCAddress:         clusterGRPCAddress,
+		InputFormat:                        "json",
+		OutputFormat:                       "json",
+		OAuthServerAddress:                 clusterHTTPAddress + "/oauth",
+		IdentityServerGRPCAddress:          clusterGRPCAddress,
+		GatewayServerEnabled:               true,
+		GatewayServerGRPCAddress:           clusterGRPCAddress,
+		NetworkServerEnabled:               true,
+		NetworkServerGRPCAddress:           clusterGRPCAddress,
+		ApplicationServerEnabled:           true,
+		ApplicationServerGRPCAddress:       clusterGRPCAddress,
+		JoinServerEnabled:                  true,
+		JoinServerGRPCAddress:              clusterGRPCAddress,
+		DeviceTemplateConverterGRPCAddress: clusterGRPCAddress,
+		DeviceClaimingServerGRPCAddress:    clusterGRPCAddress,
+		QRCodeGeneratorGRPCAddress:         clusterGRPCAddress,
+		Insecure:                           insecure,
+	}
 }
+
+// DefaultConfig contains the default config for the ttn-lw-cli binary.
+var DefaultConfig = MakeDefaultConfig(defaultClusterHost, defaultInsecure)
 
 var configCommand = commands.Config(mgr)
 
--- a/cmd/ttn-lw-cli/commands/use.go
+++ b/cmd/ttn-lw-cli/commands/use.go
@@ -1,0 +1,137 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import (
+	"bytes"
+	"crypto/md5"
+	"crypto/tls"
+	"encoding/hex"
+	"encoding/pem"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+	"go.thethings.network/lorawan-stack/pkg/errors"
+	"gopkg.in/yaml.v2"
+)
+
+const configFileName = ".ttn-lw-cli.yml"
+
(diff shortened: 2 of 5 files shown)


cmd/ttn-lw-cli/commands/config.go
```diff
@@ -71,30 +72,51 @@ func (c Config) getHosts() []string {
 	return getHosts(hosts...)
 }
 
-// DefaultConfig contains the default config for the ttn-lw-cli binary.
-var DefaultConfig = Config{
-	Base: conf.Base{
-		Log: conf.Log{
-			Level: log.InfoLevel,
+// BuildDefaultConfig builds the default config for the ttn-lw-cli binary for a given host.
+func BuildDefaultConfig(clusterHost string, insecure bool) Config {
```
rvolosatovs: Perhaps `MakeDefaultConfig` is more consistent with rest of codebase

cmd/ttn-lw-cli/commands/config.go
```diff
@@ -71,30 +72,51 @@ func (c Config) getHosts() []string {
...
+	httpPort := map[bool]int{
+		true:  1885,
+		false: 8885,
+	}
+	httpProto := map[bool]string{
+		true:  "http",
+		false: "https",
+	}
+	clusterGRPCAddress := fmt.Sprintf("%s:%d", clusterHost, grpcPort[insecure])
+	clusterHTTPAddress := fmt.Sprintf("%s://%s:%d", httpProto[insecure], clusterHost, httpPort[insecure])
```
rvolosatovs: While I appreciate the creativity, please simply do

```suggestion
	clusterGRPCAddress := fmt.Sprintf("%s:8884", clusterHost)
	clusterHTTPAddress := fmt.Sprintf("https://%s:8885", clusterHost)
	if insecure {
		clusterGRPCAddress = fmt.Sprintf("%s:1884", clusterHost)
		clusterHTTPAddress = fmt.Sprintf("http://%s:1885", clusterHost)
	}
```

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,87 @@
...
+			cfg := BuildDefaultConfig(host, insecure)
+
+			// Add CA certificate
+			if !insecure && ca != "" {
+				if abs, err := filepath.Abs(ca); err == nil {
+					cfg.CA = abs
+				}
+			}
+
+			bytestr, err := yaml.Marshal(cfg)
```
rvolosatovs:
```suggestion
			b, err := yaml.Marshal(cfg)
```

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,87 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			insecure, _ := cmd.Flags().GetBool("insecure")
+			ca, _ := cmd.Flags().GetString("ca")
+			host, _ := cmd.Flags().GetString("host")
+
+			user, _ := cmd.Flags().GetBool("user")
+			overwrite, _ := cmd.Flags().GetBool("overwrite")
+
+			// Build configuration
+			cfg := BuildDefaultConfig(host, insecure)
```
rvolosatovs:
```suggestion
			conf := BuildDefaultConfig(host, insecure)
```

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,87 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
+)
+
+var (
+	configFile = ".ttn-lw-cli.yml"
```
rvolosatovs:
```suggestion
const configFileName = ".ttn-lw-cli.yml"

var (
```

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+
+			fname := configFile
+			if user {
+				dir, err := os.UserConfigDir()
+				if err != nil {
+					return err
+				}
+				fname = filepath.Join(dir, configFile)
+			}
```
rvolosatovs:
```suggestion
			configFile := configFileName
			if user {
				dir, err := os.UserConfigDir()
				if err != nil {
					return err
				}
				configFile = filepath.Join(dir, configFileName)
			}
```

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,87 @@
...
+
+			_, err = os.Stat(fname)
+			if err == nil && !overwrite {
+				fmt.Printf("%s exists. Not overwriting.\n", fname)
```
rvolosatovs: Don't we have some kind of logger for this @htdvisser ?

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,87 @@
...
+				os.Exit(-1)
+			}
+
+			err = ioutil.WriteFile(fname, bytestr, 0644)
+			if err == nil {
+				fmt.Printf("Config file for %s written in %s\n", host, fname)
+			}
+			return err
```
rvolosatovs:
```suggestion
			if err = ioutil.WriteFile(fname, bytestr, 0644); err != nil {
				fmt.Printf("Failed to write config file: %s\n", err)
				os.Exit(-1)
			}
			fmt.Printf("Config file for %s written in %s\n", host, fname)
			return nil
```

cmd/ttn-lw-cli/commands/use.go
neoaggelos: Switched to using the logger. Needed to fix the pre-run hook to not validate authentication.

neoaggelos:
Also rebased with latest master

cmd/ttn-lw-cli/commands/config.go
```diff
@@ -71,30 +72,43 @@ func (c Config) getHosts() []string {
...
-	NetworkServerEnabled:               true,
-	NetworkServerGRPCAddress:           clusterGRPCAddress,
-	ApplicationServerEnabled:           true,
-	ApplicationServerGRPCAddress:       clusterGRPCAddress,
-	JoinServerEnabled:                  true,
-	JoinServerGRPCAddress:              clusterGRPCAddress,
-	DeviceTemplateConverterGRPCAddress: clusterGRPCAddress,
-	DeviceClaimingServerGRPCAddress:    clusterGRPCAddress,
-	QRCodeGeneratorGRPCAddress:         clusterGRPCAddress,
+		InputFormat:                        "json",
```
htdvisser: We also want to set the `CredentialsID` so that we can be logged in to multiple deployments at the same time.

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,87 @@
...
+
+var (
+	configFileName = ".ttn-lw-cli.yml"
+	useCommand     = &cobra.Command{
+		Use:               "use",
+		Short:             "Use",
+		PersistentPreRunE: preRun(),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			insecure, _ := cmd.Flags().GetBool("insecure")
+			ca, _ := cmd.Flags().GetString("ca")
```
htdvisser: 1. No need to do any `ca` stuff if `insecure == true`
2. We could also consider adding another flag to load a custom CA by connecting to the deployment:

```go
conn, err := tls.Dial("tcp", host+":8884", &tls.Config{InsecureSkipVerify: true})
if err != nil {
	return err
}
defer conn.Close()
f, err := os.OpenFile(filepath.Join(dir, "ca.pem"))
defer func() {
	if closeErr := f.Close(); err == nil && closeErr != nil {
		err = closeErr
	}
}()
for _, cert := range conn.ConnectionState().PeerCertificates {
	if !cert.IsCA {
		continue
	}
	pem.Encode(f, &pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw})
}
```

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,87 @@
...
+			host, _ := cmd.Flags().GetString("host")
```
htdvisser: Shouldn't this be an arg instead of a flag?

neoaggelos: Changed to be a flag.

cmd/ttn-lw-cli/commands/use.go
neoaggelos: No CA stuff is done if `insecure == true`. Does this not suffer by the same security issues discussed in #2156?

cmd/ttn-lw-cli/commands/config.go
neoaggelos: The `credentials-id` is kept empty, as is currently the case. Added a `--credentials-id` flag that adds that id to the resulting config file.

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,133 @@
...
+			ca, _ := cmd.Flags().GetBool("ca")
+			user, _ := cmd.Flags().GetBool("user")
+			overwrite, _ := cmd.Flags().GetBool("overwrite")
+			credentialsID, _ := cmd.Flags().GetString("credentials-id")
+
+			switch len(args) {
+			case 1:
+			default:
+				return errNoHost
+			}
```
rvolosatovs:
```suggestion
			if len(args) != 1 {
				return errNoHost
			}
```

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,133 @@
...
+
+			host := args[0]
+
+			// Build configuration
+			conf := MakeDefaultConfig(host, insecure)
+
+			// Credentials ID
```
rvolosatovs: Redundant comment

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,133 @@
...
+
+			switch len(args) {
+			case 1:
+			default:
+				return errNoHost
+			}
+
+			host := args[0]
+
+			// Build configuration
```
rvolosatovs: Redundant comment

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,133 @@
...
+
+			configFile, err := destPath(configFileName, user)
+			if err != nil {
+				return err
+			}
+
+			_, err = os.Stat(configFile)
+			if err == nil && !overwrite {
+				logger.Errorf("%s exists. Use --overwrite", configFile)
+				os.Exit(-1)
```
rvolosatovs: Why is this inconsistent with other error handling? Most of the time this function simply returns, but here you do `os.Exit(-1)` explicitly, is there a reason for that?

neoaggelos: Added appropriate errors

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,129 @@
...
+				defer func() {
+					if closeErr := f.Close(); err == nil && closeErr != nil {
+						err = closeErr
+					}
+				}()
+				for _, cert := range conn.ConnectionState().PeerCertificates {
+					if !cert.IsCA {
+						continue
+					}
+					pem.Encode(f, &pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw})
```
rvolosatovs: Do not discard errors, also perhaps a better approach would be to use a `bytes.Buffer`, to which you'd write all certificates first and then use `ioutil.WriteFile`(https://golang.org/pkg/io/ioutil/#WriteFile) to write the file with correct permissions. Also note that current implementation approach fails if `caFile` already exists.
Additionally we probably want to ask the user before overwriting, so perhaps check for `--overwrite` here as well?

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,129 @@
...
+				if err != nil {
+					return err
+				}
+				defer conn.Close()
+				caFile, err := destPath("ca.pem", user)
+				if err != nil {
+					return err
+				}
+
+				f, err := os.Create(caFile)
```
rvolosatovs: Do not discard errors

cmd/ttn-lw-cli/commands/use.go
neoaggelos: Improved code to handle ca.pem like ttn-lw-cli.yml (and check for overwrites). Also switched to `ioutil.WriteFile`

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,140 @@
...
+				}
+				defer conn.Close()
+
+				var b bytes.Buffer
+				w := bufio.NewWriter(&b)
+				for _, cert := range conn.ConnectionState().PeerCertificates {
+					if !cert.IsCA {
+						continue
+					}
+					pem.Encode(w, &pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw})
```
rvolosatovs: Do not discard errors

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,140 @@
...
+				}
+				if err = w.Flush(); err != nil {
+					return err
+				}
+				if err = ioutil.WriteFile(caFile, b.Bytes(), 0644); err != nil {
+					return errFailWrite.WithCause(err).WithAttributes("file", caFile)
+				}
```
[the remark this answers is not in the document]
rvolosatovs: Because we want to only either write the complete file or not do anything at all. If we start writing via `pem.Encode` in a loop for each certificate, we may end up with only partially written file. IMO using the buffer is the best approach here.

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,140 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+	"go.thethings.network/lorawan-stack/pkg/errors"
+	"gopkg.in/yaml.v2"
+)
+
+func destPath(base string, user bool, overwrite bool) (string, error) {
```
rvolosatovs: This function is very specific to this command, I would just define it inside `useCommand.RunE` itself

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,140 @@
...
+					return err
+				}
+
+				conn, err := tls.Dial("tcp", conf.NetworkServerGRPCAddress, &tls.Config{InsecureSkipVerify: true})
+				if err != nil {
+					return err
+				}
+				defer conn.Close()
+
+				var b bytes.Buffer
```
rvolosatovs:
```suggestion
				var buf bytes.Buffer
```
`b` is usually used for `[]byte`

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,140 @@
...
+				w := bufio.NewWriter(&b)
```
rvolosatovs: `bytes.Buffer` is already `io.Writer`, so no need for this.

cmd/ttn-lw-cli/commands/use.go
neoaggelos: So, one thing I do not really like about this is that we have to make sure to sanitize `host` before trying to connect to port 8884, which is generally not an easy problem in itself.

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,140 @@
...
+			host := args[0]
+			conf := MakeDefaultConfig(host, insecure)
+			if credentialsID != "" {
+				conf.CredentialsID = credentialsID
+			}
+
+			// Get CA certificate from server
+			if !insecure && ca {
+				logger.Infof("Will retrieve certificate from %s", conf.NetworkServerGRPCAddress)
+				caFile, err := destPath("ca.pem", user, overwrite)
```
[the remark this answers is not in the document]
neoaggelos: Though I agree, we cannot really use the host name without sanitizing the input first. Going with using a hash, even though it may hurt readability.

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,140 @@
...
+			logger.Infof("Config file for %s written in %s", host, configFile)
+			return nil
+		},
+	}
+)
+
+func init() {
+	useCommand.Flags().Bool("insecure", defaultInsecure, "Connect without TLS")
+	useCommand.Flags().String("host", defaultClusterHost, "Server host name")
+	useCommand.Flags().Bool("ca", false, "Certificate file to use")
```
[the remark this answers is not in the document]
neoaggelos: Updated description

cmd/ttn-lw-cli/commands/use.go
neoaggelos: Printing a warning if `pem.Encode` fails

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,147 @@
...
+			logger.Infof("Config file for %s written in %s", host, configFile)
+			return nil
+		},
+	}
+)
+
+func init() {
+	useCommand.Flags().Bool("insecure", defaultInsecure, "Connect without TLS")
+	useCommand.Flags().String("host", defaultClusterHost, "Server host name")
+	useCommand.Flags().Bool("ca", false, "Connect to server and retrieve CA")
```
htdvisser: Can we rename this flag to `--fetch-ca` or something? I think `--ca` is confusing.

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,147 @@
...
+				return errNoHost
+			}
+			insecure, _ := cmd.Flags().GetBool("insecure")
+			ca, _ := cmd.Flags().GetBool("ca")
+			user, _ := cmd.Flags().GetBool("user")
+			overwrite, _ := cmd.Flags().GetBool("overwrite")
+
+			hash := func(in string) string {
+				h := md5.New()
+				io.Copy(h, strings.NewReader(in))
```
rvolosatovs:
```suggestion
				io.WriteString(h, in)
```

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,147 @@
...
+						return "", err
+					}
+					fileName = filepath.Join(dir, base)
+				}
+
+				_, err := os.Stat(fileName)
+				if os.IsNotExist(err) || overwrite {
+					return fileName, nil
+				}
+
```
rvolosatovs: nit; remove newline

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,147 @@
...
+				if user {
+					dir, err := os.UserConfigDir()
+					if err != nil {
+						return "", err
+					}
+					fileName = filepath.Join(dir, base)
+				}
+
+				_, err := os.Stat(fileName)
+				if os.IsNotExist(err) || overwrite {
```
rvolosatovs: nit; prefer inverting this `if` to have indented error flow and successful flow on same indentation level

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,147 @@
...
+					return err
+				}
+
+				conn, err := tls.Dial("tcp", conf.NetworkServerGRPCAddress, &tls.Config{InsecureSkipVerify: true})
+				if err != nil {
+					return err
+				}
+				defer conn.Close()
+
+				var buf bytes.Buffer
```
rvolosatovs: nit; to avoid doing `&buf` below in a loop, you could also just do
```suggestion
				buf := &bytes.Buffer{}
```

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -0,0 +1,144 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return errNoHost
+			}
+			insecure, _ := cmd.Flags().GetBool("insecure")
+			fetchCA, _ := cmd.Flags().GetBool("fetch-ca")
+			user, _ := cmd.Flags().GetBool("user")
+			overwrite, _ := cmd.Flags().GetBool("overwrite")
+
+			hash := func(in string) string {
```
rvolosatovs: This function is not necessary(only one usage), see below

cmd/ttn-lw-cli/commands/use.go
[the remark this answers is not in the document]
neoaggelos: Changed to the proposed `--fetch-ca`

doc/content/guides/getting-started/cli/_index.md
```diff
@@ -40,6 +40,22 @@ The command-line needs to be configured to connect to your deployment on `thethi
...
 
+### Generate configuration file
+
+For a standard deployment on `thethings.example.com`, all you need is:
+
+```bash
+$ ttn-lw-cli use thethings.example.com [--fetch-ca] [--user] [--overwrite]
+```
+
+This will generate and save the required CLI config file. Use the `--fetch-ca` flag to also connect to the server and retrieve the CA required for establishing secure communication. By default, the file is saved on the current directory, use the `--user` to save it under the user config directory.
```
htdvisser: Note that `--fetch-ca` is only necessary when using a CA that isn't already trusted by the system.

PR #2156, opened 2020-03-13
cli: Allow including the CA in the ttn-lw-cli config file

neoaggelos:
#### Summary

References #1240

#### Changes

- Enable setting the stack CA inline in the CLI config file, so that we can have the CLI configuration bundled in a single file (and in the future have the stack server it under a public GET URL).

#### Notes for Reviewers

The naming here might become a bit mixed. We can't really change `--ca` το `--ca-file` because that would break compatibility. What we can do is deprecate `--ca`, then add new `--cert-file` and `--cert` options. 

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/root.go
+++ b/cmd/ttn-lw-cli/commands/root.go
@@ -113,11 +113,21 @@ func preRun(tasks ...func() error) func(cmd *cobra.Command, args []string) error
 		if config.Insecure {
 			api.SetInsecure(true)
 		}
-		if config.CA != "" {
-			pemBytes, err := ioutil.ReadFile(config.CA)
+
+		useCA := true
+		var pemBytes []byte
+		if config.CAFile != "" {
+			pemBytes, err = ioutil.ReadFile(config.CAFile)
 			if err != nil {
 				return err
 			}
+		} else if config.CA != "" {
+			pemBytes = []byte(config.CA)
+		} else {
+			useCA = false
+		}
+
+		if useCA {
 			rootCAs := http.DefaultTransport.(*http.Transport).TLSClientConfig.RootCAs
 			if rootCAs == nil {
 				if rootCAs, err = x509.SystemCertPool(); err != nil {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - This requires a database migration (`ttn-lw-stack is-db migrate`) because of the added columns.
 - Gateway Server rate limiting support for the UDP frontend, see (`--gs.udp.rate-limiting` options).
 - Uplink deduplication via Redis in Network Server.
+- Setting the CA certificates inline in the CLI configuration file (see `--ca-cert` option).
 
 ### Changed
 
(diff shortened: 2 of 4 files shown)


neoaggelos:
> But why do we need this exactly?

This allows us to include the certificate inside the `.ttn-lw-cli.yml` directly, like so:

```
gateway-server-grpc-address: server.name:8884
<grpc-addresses-of-other-components-here>
cacert: |
  -----BEGIN CERTIFICATE-----
  ....
  -----END CERTIFICATE-----
```

Which can be very helpful, because stack deployments that need a custom CA can generate a single config file for the CLI containing the addresses of all the components along with the TLS certificate that needs to be used for secure communication.

cmd/ttn-lw-cli/commands/root.go
```diff
@@ -113,11 +113,21 @@ func preRun(tasks ...func() error) func(cmd *cobra.Command, args []string) error
...
-			pemBytes, err := ioutil.ReadFile(config.CA)
+
+		useCA := true
+		var pemBytes []byte
+		if config.CAFile != "" {
+			pemBytes, err = ioutil.ReadFile(config.CAFile)
 			if err != nil {
 				return err
 			}
+		} else if config.CA != "" {
```
rvolosatovs: IMO `CA` should take precedence. Ideally the flag specified on the command-line should take precedence, in fact, but not sure how to achieve that with cobra @htdvisser ?

neoaggelos: Coming back to it, I think we should not really allow using both. It might be better to just exit with a useful error message.

johanstokking:
Actually I wonder if we need and want this feature, and if this opens the door for security issues to sneak in.

I personally find the operating system responsible for managing CAs when using CLI, or the browser.

It is nice to make the configuration files fully self-contained, but a verifiable chain of trust that it outside the scope of the application is there for a reason. For example, if one were to sneak in a malicious CA of a man-in-the-middle, all security is gone. And it is really hard to see this.

The idea behind https://github.com/TheThingsNetwork/lorawan-stack/issues/1240 is mostly user experience to generate a configuration file. I mentioned "potentially CA to use", but I would not make it so easy to put this in a configuration file.

Other notes:

- For most of the deployments, the server cert should be trusted already. So this would only be useful for development and local networks. For local networks, a proper environment (OS, browser) would trust the local network's CA already
- Configuration files can easily be passed around by people, from system to system
- This only works for CLI, while we actively push using the Console and will do that more and more in documentation

neoaggelos:
> Actually I wonder if we need and want this feature, and if this opens the door for security issues to sneak in.
> 
> I personally find the operating system responsible for managing CAs when using CLI, or the browser.
> 
> It is nice to make the configuration files fully self-contained, but a verifiable chain of trust that it outside the scope of the application is there for a reason. For example, if one were to sneak in a malicious CA of a man-in-the-middle, all security is gone. And it is really hard to see this.
> 
> The idea behind #1240 is mostly user experience to generate a configuration file. I mentioned "potentially CA to use", but I would not make it so easy to put this in a configuration file. Instead, I would write it to a file if a flag is set, and inform the user to trust the CA in the system root (or pass with `--ca`). Leave users responsible for security.
> 
> Other notes:
> 
> * For most of the deployments, the server cert should be trusted already. So this would only be useful for development and local networks. For local networks, a proper environment (OS, browser) would trust the local network's CA already
> * Configuration files can easily be passed around by people, from system to system
> * This only works for CLI, while we actively push using the Console and will do that more and more in documentation

Mostly agree with the points being made. If we find that this is not as useful as it is a possibly security risk, then we might as well just not add it.

However, security-wise, I think it is not any less safer than the existing option of choosing the certificate file. If an attacker is able to tinker with `ttn-lw-cli.yml` and change the included certificate, then he should be just as able to change the certificate path and sneak in his malicious CA.

Also, the existing behaviour (`ca: /path/to/cert.pem`) remains untouched.

PR #2163, opened 2020-03-16
Pending session downlink queue support in the Application Server

### LINKED ISSUE #1954: Store downlink queue per session

### LINKED ISSUE #1971: Application Server fails to recalculate downlink queue on joins
#### Summary

Application Server fails to recalculate downlink queue on joins.
Refs #1954 

#### Steps to Reproduce

1. Enqueue a bunch of downlinks to a device, such that queue accumulates
2. Make the device join
3. Observe queue being dropped

#### What do you see now?

```
DEBUG Recalculating downlink queue to restore downlink queue on join application_uid=test-app count=16 dev_eui=DEADBEEF01020304 device_uid=test-app.test-dev-b join_eui=01020304DEADBEEF namespace=applicationserver network_server=cobalt session_key_id=01702F766D45A98E561A3B319E7AF104
 DEBUG Recalculate downlink queue               application_uid=test-app count=16 device_uid=test-app.test-dev-b namespace=applicationserver network_server=cobalt new_session_key_id=01702F6D0BE16D224EF716753BA21D70 next_a_f_cnt_down=1
  WARN Failed to replace application downlink queue device_uid=test-app.test-dev-b error=error:pkg/networkserver:f_cnt_too_low (FCnt is too low) grpc_method=DownlinkQueueReplace grpc_service=ttn.lorawan.v3.AsNs namespace=networkserver request_id=01E0QQCVGFGZNCWJ96JH7P2B03
  WARN Recalculate downlink queue failed; clearing the downlink queue application_uid=test-app count=16 device_uid=test-app.test-dev-b error=rpc error: code = InvalidArgument desc = error:pkg/networkserver:f_cnt_too_low (FCnt is too low) namespace=applicationserver network_server=cobalt new_session_key_id=01702F6D0BE16D224EF716753BA21D70 next_a_f_cnt_down=1
 DEBUG Replaced application downlink queue      device_uid=test-app.test-dev-b grpc_method=DownlinkQueueReplace grpc_service=ttn.lorawan.v3.AsNs namespace=networkserver queue_length=0 request_id=01E0QQCVH9AH0PFW1TA2QECYGV
  WARN Failed to recalculate downlink queue, items lost application_uid=test-app count=16 dev_eui=DEADBEEF01020304 device_uid=test-app.test-dev-b error=rpc error: code = InvalidArgument desc = error:pkg/networkserver:f_cnt_too_low (FCnt is too low) join_eui=01020304DEADBEEF namespace=applicationserver network_server=cobalt session_key_id=01702F766D45A98E561A3B319E7AF104
```

That's the join-accept:
```json
{
  "name": "as.up.join.forward",
  "time": "2020-02-10T14:17:35.553635062Z",
  "identifiers": [
    {
      "device_ids": {
        "device_id": "test-dev-b",
        "application_ids": {
          "application_id": "test-app"
        },
        "dev_eui": "DEADBEEF01020304",
        "join_eui": "01020304DEADBEEF",
        "dev_addr": "00B717F9"
      }
    }
  ],
  "data": {
    "@type": "type.googleapis.com/ttn.lorawan.v3.ApplicationUp",
    "end_device_ids": {
      "device_id": "test-dev-b",
      "application_ids": {
        "application_id": "test-app"
      },
      "dev_eui": "DEADBEEF01020304",
      "join_eui": "01020304DEADBEEF",
      "dev_addr": "00B717F9"
    },
    "correlation_ids": [
      "as:up:01E0QQCVGF2CEWJV0RMKPVN3D8",
      "gs:conn:01E0QQB2AZD5F88R8N2HDZWGGN",
      "gs:uplink:01E0QQCVA1PGX54T3STWKMXVRA",
      "ns:uplink:01E0QQCVA2Q17MRFE4VRJ9HRSQ",
      "rpc:/ttn.lorawan.v3.GsNs/HandleUplink:01E0QQCVA1K321HGYGD0B9R9QT"
    ],
    "received_at": "2020-02-10T14:17:35.503187641Z",
    "join_accept": {
      "session_key_id": "AXAvdm1FqY5WGjsxnnrxBA==",
      "received_at": "2020-02-10T14:17:35.304388899Z"
    }
  },
  "correlation_ids": [
    "as:up:01E0QQCVGF2CEWJV0RMKPVN3D8",
    "gs:conn:01E0QQB2AZD5F88R8N2HDZWGGN",
    "gs:uplink:01E0QQCVA1PGX54T3STWKMXVRA",
    "ns:uplink:01E0QQCVA2Q17MRFE4VRJ9HRSQ",
    "rpc:/ttn.lorawan.v3.GsNs/HandleUplink:01E0QQCVA1K321HGYGD0B9R9QT"
  ],
  "origin": "cobalt",
  "visibility": {
    "rights": [
    ]
  }
}
```

Queue successfully recalculated

#### Environment

3.5.2

It seems that AS uses wrong (old) session key ID when recalculating the queue after join and so NS rejects the queue replacement.

@johanstokking

adriansmares:
#### Summary

Closes #1954 
Closes #1971

#### Changes

- Downlink queue operations are now done on both sessions (pending + current).
- Downlink queue listing returns only the queue of the current session.
- On join-accept, the pending downlink queue is recomputed, but the current downlink queue is unchanged.
- Full downlink queue invalidations (`DownlinkQueueInvalidated` event) and individual downlink message nacks result in the computation of the both queues, if the pending session is present. If there is no pending session, only the current session queue is updated.
- On session restore a full replace is still required, and may result in a downlink queue invalidation. The replace occurs only for the current session (since there is no pending session after session restore).

#### Notes for Reviewers

@rvolosatovs I've moved the `Partition*` utilities to `pkg/ttnpb` from `pkg/networkserver` since both components use them. Is this location correct or should I make a new package in `pkg/util` ?

~I think the only part missing from #1954 is the switch between the pending and the current session in the NS on the first uplink. (cc @rvolosatovs)~

#### Checklist

### DIFF
--- a/pkg/applicationserver/applicationserver.go
+++ b/pkg/applicationserver/applicationserver.go
@@ -338,19 +338,26 @@ func (as *ApplicationServer) downlinkQueueOp(ctx context.Context, ids ttnpb.EndD
 			var encryptedItems []*ttnpb.ApplicationDownlink
 			for _, session := range sessions {
 				for _, item := range items {
-					encryptedItem := *item
-					encryptedItem.SessionKeyID = session.SessionKeyID
-					encryptedItem.FCnt = session.LastAFCntDown + 1
+					encryptedItem := &ttnpb.ApplicationDownlink{
+						SessionKeyID:   session.SessionKeyID,
+						FPort:          item.FPort,
+						FCnt:           session.LastAFCntDown + 1,
+						FRMPayload:     item.FRMPayload,
+						DecodedPayload: item.DecodedPayload,
+						Confirmed:      item.Confirmed,
+						ClassBC:        item.ClassBC,
+						Priority:       item.Priority,
+						CorrelationIDs: item.CorrelationIDs,
+					}
 					if !dev.SkipPayloadCrypto {
-						if err := as.encodeAndEncrypt(ctx, dev, session, &encryptedItem, link.DefaultFormatters); err != nil {
+						if err := as.encodeAndEncrypt(ctx, dev, session, encryptedItem, link.DefaultFormatters); err != nil {
 							logger.WithError(err).Warn("Encoding and encryption of downlink message failed; dropping item")
 							return nil, nil, err
 						}
 					}
 					encryptedItem.DecodedPayload = nil
-					encryptedItem.CorrelationIDs = item.CorrelationIDs
 					session.LastAFCntDown = encryptedItem.FCnt
-					encryptedItems = append(encryptedItems, &encryptedItem)
+					encryptedItems = append(encryptedItems, encryptedItem)
 				}
 			}
 			client := ttnpb.NewAsNsClient(link.conn)
@@ -461,10 +468,11 @@ func (as *ApplicationServer) DownlinkQueueList(ctx context.Context, ids ttnpb.En
 		return nil, err
 	}
 	for _, item := range queue {
-		item.FRMPayload, err = crypto.DecryptDownlink(appSKey, session.DevAddr, item.FCnt, item.FRMPayload, false)
+		b, err := crypto.DecryptDownlink(appSKey, session.DevAddr, item.FCnt, item.FRMPayload, false)
 		if err != nil {
 			return nil, err
 		}
+		item.FRMPayload = b
 	}
 	return queue, nil
 }
@@ -569,11 +577,11 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
 			mask = append(mask, "pending_session")
 			if len(joinAccept.InvalidatedDownlinks) > 0 {
 				// The Network Server does not reset the downlink queues as the new security session is established.
-				// The old downlink queues are passed as part of the join-accept and the Application Server should
+				// The old downlink queue is passed as part of the join-accept and the Application Server should
 				// recalculate the pending downlink queue.
 				logger := logger.WithField("count", len(joinAccept.InvalidatedDownlinks))
 				logger.Debug("Recalculating downlink queue to restore downlink queue on join")
-				if err := as.recomputePendingQueue(ctx, dev, link, previousSession, joinAccept.InvalidatedDownlinks); err != nil {
+				if err := as.recalculatePendingDownlinkQueue(ctx, dev, link, previousSession, joinAccept.InvalidatedDownlinks); err != nil {
 					logger.WithError(err).Warn("Failed to recalculate downlink queue; items lost")
 				}
 			}
@@ -586,13 +594,29 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
 	return nil
 }
 
-// downlinkQueuesUpdater creates the pending and current session downlink queues.
-// The updater generally mutates the end device's sessions.
-type downlinkQueuesUpdater func(context.Context, *ttnpb.EndDevice) ([]*ttnpb.ApplicationDownlink, []*ttnpb.ApplicationDownlink, error)
+// clearDownlinkQueue clears the queue downlink queue of the provided device and publishes the appropriate events.
+func (as *ApplicationServer) clearDownlinkQueue(ctx context.Context, ids ttnpb.EndDeviceIdentifiers, link *link) error {
+	logger := log.FromContext(ctx)
+	client := ttnpb.NewAsNsClient(link.conn)
+	req := &ttnpb.DownlinkQueueRequest{
+		EndDeviceIdentifiers: ids,
+	}
+	_, err := client.DownlinkQueueReplace(ctx, req, link.callOpts...)
+	if err != nil {
+		logger.WithError(err).Warn("Failed to clear the downlink queue; any queued items in the Network Server are invalid")
+		events.Publish(evtInvalidQueueDataDown(ctx, ids, err))
+	} else {
+		events.Publish(evtLostQueueDataDown(ctx, ids, err))
+	}
+	return err
+}
 
-// changeDownlinkQueue updates the downlink queue of the end device with the given downlink queues updater.
-// This method rollbacks any changes to the sessions and resets the queue if the replacement fails.
-func (as *ApplicationServer) changeDownlinkQueue(ctx context.Context, dev *ttnpb.EndDevice, link *link, f downlinkQueuesUpdater) (err error) {
+// downlinkQueueTransaction represents a transaction to be run on the downlink queues of the provided device.
+type downlinkQueueTransaction func(context.Context, *ttnpb.EndDevice) error
+
+// runDownlinkQueueTransaction runs the provided downlink queue transaction on the device. If the transaction
+// fails, the LastAFCntDown session fields are restored to their previous values and the downlink queue is reset.
+func (as *ApplicationServer) runDownlinkQueueTransaction(ctx context.Context, dev *ttnpb.EndDevice, link *link, t downlinkQueueTransaction) (err error) {
 	logger := log.FromContext(ctx)
 	pendingSession := dev.PendingSession
 	if pendingSession != nil {
@@ -616,52 +640,34 @@ func (as *ApplicationServer) changeDownlinkQueue(ctx context.Context, dev *ttnpb
 				session.LastAFCntDown = oldLastAFCntDown
 			}
 			logger.WithError(err).Warn("Failed to recalculate downlink queue; clearing the downlink queue")
-			client := ttnpb.NewAsNsClient(link.conn)
-			req := &ttnpb.DownlinkQueueRequest{
-				EndDeviceIdentifiers: dev.EndDeviceIdentifiers,
-			}
-			if _, err := client.DownlinkQueueReplace(ctx, req, link.callOpts...); err != nil {
-				logger.WithError(err).Warn("Failed to clear the downlink queue; any queued items in the Network Server are invalid")
-				events.Publish(evtInvalidQueueDataDown(ctx, dev.EndDeviceIdentifiers, err))
-			} else {
-				events.Publish(evtLostQueueDataDown(ctx, dev.EndDeviceIdentifiers, err))
-			}
+			as.clearDownlinkQueue(ctx, dev.EndDeviceIdentifiers, link)
 		}
 	}()
 	if dev.SkipPayloadCrypto {
 		return errPayloadCryptoDisabled.New()
 	}
-	newCurrentQueue, newPendingQueue, err := f(ctx, dev)
-	if err != nil {
-		return err
-	}
-	downlinks := append(newCurrentQueue[:0:0], newCurrentQueue...)
-	downlinks = append(downlinks, newPendingQueue...)
-	client := ttnpb.NewAsNsClient(link.conn)
-	req := &ttnpb.DownlinkQueueRequest{
-		EndDeviceIdentifiers: dev.EndDeviceIdentifiers,
-		Downlinks:            downlinks,
-	}
-	_, err = client.DownlinkQueueReplace(ctx, req, link.callOpts...)
-	return err
+	return t(ctx, dev)
 }
 
 var errUnknownSession = errors.DefineNotFound("unknown_session", "unknown session")
 
-// recomputePendingQueue partitions the invalidated downlinks based on their session key ID. Downlinks which are part of the
-// current session are not modified, while the downlinks from the previous pending session are migrated to the new pending
-// session.
-// This method requires the given end device's pending session to be set. This method mutates the end device's pending session LastAFCntDown.
-// This method does not change the contents of the given invalid downlink queue.
-func (as *ApplicationServer) recomputePendingQueue(ctx context.Context, dev *ttnpb.EndDevice, link *link, previousPendingSession *ttnpb.Session, invalidatedDownlinks []*ttnpb.ApplicationDownlink) error {
-	return as.changeDownlinkQueue(ctx, dev, link, func(ctx context.Context, dev *ttnpb.EndDevice) ([]*ttnpb.ApplicationDownlink, []*ttnpb.ApplicationDownlink, error) {
-		var previousPendingQueue, currentQueue []*ttnpb.ApplicationDownlink
+// recalculatePendingDownlinkQueue computes the downlink queue of the device's pending session, by decrypting the
+// invalidated queue using the device session and moving them to the pending session.
+// The previous pending session is used for cases in which the device has not been activated, and as such does not
+// have a device session. In such cases, if there are downlinks are present in the pending queue, they are decrypted
+// using the previous pending session and encrypted using new pending session.
+// This method mutates the LastAFCntDown of pending session. Downlinks which cannot be decrypted are dropped.
+// This method uses the downlink queue transaction mechanism, so any errors that occur during recomputation will
+// result in an downlink queue reset attempt.
+func (as *ApplicationServer) recalculatePendingDownlinkQueue(ctx context.Context, dev *ttnpb.EndDevice, link *link, previousPendingSession *ttnpb.Session, invalidatedDownlinks []*ttnpb.ApplicationDownlink) error {
+	return as.runDownlinkQueueTransaction(ctx, dev, link, func(ctx context.Context, dev *ttnpb.EndDevice) error {
+		var previousPendingQueue, previousQueue []*ttnpb.ApplicationDownlink
 		unmatched := invalidatedDownlinks
 		if previousPendingSession != nil {
 			previousPendingQueue, unmatched = ttnpb.PartitionDownlinksBySessionKeyIDEquality(previousPendingSession.SessionKeyID, unmatched...)
 		}
 		if dev.Session != nil {
-			currentQueue, unmatched = ttnpb.PartitionDownlinksBySessionKeyIDEquality(dev.Session.SessionKeyID, unmatched...)
+			previousQueue, unmatched = ttnpb.PartitionDownlinksBySessionKeyIDEquality(dev.Session.SessionKeyID, unmatched...)
 		}
 		logger := log.FromContext(ctx)
 		for _, item := range unmatched {
@@ -678,10 +684,10 @@ func (as *ApplicationServer) recomputePendingQueue(ctx context.Context, dev *ttn
 		switch {
 		case previousPendingSession == nil && dev.Session == nil:
 			// We cannot decode any of the downlinks if both sessions are missing.
-			return nil, nil, errNoDeviceSession.New()
+			return errNoDeviceSession.New()
 		case dev.Session != nil:
 			// In the presence of a current session we copy the items from the current session to the pending session.
-			sourceQueue = currentQueue
+			sourceQueue = previousQueue
 			sourceSession = dev.Session
 			logger.Debug("Initializing pending downlink queue from the current session")
 		case previousPendingSession != nil:
@@ -694,18 +700,30 @@ func (as *ApplicationServer) recomputePendingQueue(ctx context.Context, dev *ttn
 
 		newPendingQueue, err := as.migrateDownlinkQueue(ctx, dev.EndDeviceIdentifiers, sourceQueue, sourceSession, dev.PendingSession, 1)
 		if err != nil {
-			return nil, nil, err
+			return err
 		}
-		return currentQueue, newPendingQueue, nil
+
+		client := ttnpb.NewAsNsClient(link.conn)
+		_, err = client.DownlinkQueueReplace(ctx, &ttnpb.DownlinkQueueRequest{
+			EndDeviceIdentifiers: dev.EndDeviceIdentifiers,
+			Downlinks:            append(previousQueue, newPendingQueue...),
+		}, link.callOpts...)
+		return err
 	})
 }
 
-// replaceDownlinkQueue replaces the current and pending queues of the end device with the provided downlinks.
-// This method requires the given end device's session to be set. This method mutates the end device's session LastAFCntDown.
-// This method mutates the end device's pending session LastAFCntDown if the pending session exists and replacePending is true.
-// This method does not change the contents of the given downlinks.
-func (as *ApplicationServer) replaceDownlinkQueue(ctx context.Context, dev *ttnpb.EndDevice, link *link, previousSession *ttnpb.Session, downlinks []*ttnpb.ApplicationDownlink, nextAFCntDown uint32, replacePending bool) error {
-	return as.changeDownlinkQueue(ctx, dev, link, func(ctx context.Context, dev *ttnpb.EndDevice) (newCurrentQueue []*ttnpb.ApplicationDownlink, newPendingQueue []*ttnpb.ApplicationDownlink, err error) {
+// recalculateDownlinkQueue computes downlink queue of the end device's current session by decrypting the provided
+// downlinks using the previous session, and then re-encrypting them using current session starting from the
+// provided AFCntDown.
+// The previous session is provided for cases in which the downlink queue has to be migrated to a new device
+// session. This can occur on device re-activation, when a device sends an uplink with a previously used session key
+// ID for which the session key is available.
+// This method mutates the LastAFCntDown of end device's session. Downlinks which cannot be decrypted are dropped.
+// The pending downlink queue of the end device is discarded.
+// This method uses the downlink queue transaction mechanism, so any errors that occur during recomputation will
+// result in an downlink queue reset attempt.
+func (as *ApplicationServer) recalculateDownlinkQueue(ctx context.Context, dev *ttnpb.EndDevice, link *link, previousSession *ttnpb.Session, downlinks []*ttnpb.ApplicationDownlink, nextAFCntDown uint32) error {
+	return as.runDownlinkQueueTransaction(ctx, dev, link, func(ctx context.Context, dev *ttnpb.EndDevice) (err error) {
 		downlinks, unmatched := ttnpb.PartitionDownlinksBySessionKeyIDEquality(previousSession.SessionKeyID, downlinks...)
 		for _, item := range unmatched {
 			log.FromContext(ctx).WithFields(log.Fields(
@@ -715,23 +733,25 @@ func (as *ApplicationServer) replaceDownlinkQueue(ctx context.Context, dev *ttnp
 			)).Warn("Downlink message with unknown session key ID found; dropping item")
 			registerDropDownlink(ctx, dev.EndDeviceIdentifiers, item, errUnknownSession)
 		}
-		if replacePending && dev.PendingSession != nil {
-			newPendingQueue, err = as.migrateDownlinkQueue(ctx, dev.EndDeviceIdentifiers, downlinks, previousSession, dev.PendingSession, 1)
-			if err != nil {
-				return nil, nil, err
-			}
-		}
-		newCurrentQueue, err = as.migrateDownlinkQueue(ctx, dev.EndDeviceIdentifiers, downlinks, previousSession, dev.Session, nextAFCntDown)
+		var newQueue []*ttnpb.ApplicationDownlink
+		newQueue, err = as.migrateDownlinkQueue(ctx, dev.EndDeviceIdentifiers, downlinks, previousSession, dev.Session, nextAFCntDown)
 		if err != nil {
-			return nil, nil, err
+			return err
 		}
-		return newCurrentQueue, newPendingQueue, nil
+
+		client := ttnpb.NewAsNsClient(link.conn)
+		req := &ttnpb.DownlinkQueueRequest{
+			EndDeviceIdentifiers: dev.EndDeviceIdentifiers,
+			Downlinks:            newQueue,
+		}
+		_, err = client.DownlinkQueueReplace(ctx, req, link.callOpts...)
+		return err
 	})
 }
 
-// migrateDownlinkQueue constructs a new downlink queue by decrypting the elements of the old queue and encrypting them using the new session.
-// If re-encrypting a message fails, the message is skipped.
-// This method mutates the new session's LastAFCntDown if any downlinks have been migrated.
+// migrateDownlinkQueue constructs a new downlink queue by decrypting the items of the old queue using the
+// old session and encrypting them using the new session.
+// This method mutates the LastAFCntDown of the new session. Downlinks which cannot be decrypted are dropped.
 // This method does not change the contents of the old downlink queue.
 func (as *ApplicationServer) migrateDownlinkQueue(ctx context.Context, ids ttnpb.EndDeviceIdentifiers, oldQueue []*ttnpb.ApplicationDownlink, oldSession *ttnpb.Session, newSession *ttnpb.Session, nextAFCntDown uint32) ([]*ttnpb.ApplicationDownlink, error) {
 	if oldSession.AppSKey == nil || newSession.AppSKey == nil {
@@ -755,9 +775,15 @@ func (as *ApplicationServer) migrateDownlinkQueue(ctx context.Context, ids ttnpb
 			"f_cnt", oldItem.FCnt,
 			"session_key_id", oldItem.SessionKeyID,
 		))
-		frmPayload, err := crypto.DecryptDownlink(oldAppSKey, oldSession.DevAddr, oldItem.FCnt, oldItem.FRMPayload)
+		frmPayload, err := crypto.DecryptDownlink(oldAppSKey, oldSession.DevAddr, oldItem.FCnt, oldItem.FRMPayload, false)
 		if err != nil {
 			logger.WithError(err).Warn("Failed to decrypt downlink message; dropping item")
+			registerDropDownlink(ctx, ids, oldItem, err)
+			continue
+		}
+		newFRMPayload, err := crypto.EncryptDownlink(newAppSKey, newSession.DevAddr, newSession.LastAFCntDown+1, frmPayload, false)
+		if err != nil {
+			logger.WithError(err).Warn("Failed to encrypt downlink message; dropping item")
 			registerDropDownlink(ctx, ids, oldItem, err)
 			continue
 		}
@@ -765,16 +791,11 @@ func (as *ApplicationServer) migrateDownlinkQueue(ctx context.Context, ids ttnpb
 			SessionKeyID:   newSession.SessionKeyID,
 			FPort:          oldItem.FPort,
 			FCnt:           newSession.LastAFCntDown + 1,
+			FRMPayload:     newFRMPayload,
 			Confirmed:      oldItem.Confirmed,
 			ClassBC:        oldItem.ClassBC,
 			Priority:       oldItem.Priority,
 			CorrelationIDs: oldItem.CorrelationIDs,
-		}
-		newItem.FRMPayload, err = crypto.EncryptDownlink(newAppSKey, newSession.DevAddr, newItem.FCnt, frmPayload)
-		if err != nil {
-			logger.WithError(err).Warn("Failed to encrypt downlink message; dropping item")
-			registerDropDownlink(ctx, ids, oldItem, err)
-			continue
 		}
 		newQueue = append(newQueue, newItem)
 		newSession.LastAFCntDown = newItem.FCnt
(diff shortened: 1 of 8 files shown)


pkg/applicationserver/applicationserver.go
```diff
@@ -327,33 +327,37 @@ func (as *ApplicationServer) downlinkQueueOp(ctx context.Context, ids ttnpb.EndD
 			if dev == nil {
 				return nil, nil, errDeviceNotFound.WithAttributes("device_uid", unique.ID(ctx, ids))
 			}
-			session := dev.Session
-			if session != nil {
+			var sessions []*ttnpb.Session
```
johanstokking: nit; `sessions := make([]*ttnpb.Session, 0, 2)`

pkg/applicationserver/applicationserver.go
```diff
@@ -567,16 +573,14 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
 				StartedAt: time.Now().UTC(),
 			}
 			mask = append(mask, "pending_session")
-			if !joinAccept.PendingSession && len(joinAccept.InvalidatedDownlinks) > 0 {
-				// The Network Server reset the downlink queue as the new security session invalidated it.
-				// The invalidated downlink queue is passed as part of the join-accept and the Application Server should recalculate it.
-				// This changes the LastAFCntDown in the session, so it should be run as part of the transaction.
+			if len(joinAccept.InvalidatedDownlinks) > 0 {
+				// The Network Server does not reset the downlink queues as the new security session is established.
```
johanstokking: Is this true @rvolosatovs ?

pkg/applicationserver/applicationserver.go
```diff
@@ -588,6 +592,202 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
 	return nil
 }
 
+// downlinkQueuesUpdater creates the pending and current session downlink queues.
+// The updater generally mutates the end device's sessions.
+type downlinkQueuesUpdater func(context.Context, *ttnpb.EndDevice) ([]*ttnpb.ApplicationDownlink, []*ttnpb.ApplicationDownlink, error)
+
+// changeDownlinkQueue updates the downlink queue of the end device with the given downlink queues updater.
```
johanstokking: Change vs update

It's not clear to me what really happens here from this naming and description

pkg/applicationserver/applicationserver.go
```diff
@@ -588,6 +592,202 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
...
+// This method rollbacks any changes to the sessions and resets the queue if the replacement fails.
+func (as *ApplicationServer) changeDownlinkQueue(ctx context.Context, dev *ttnpb.EndDevice, link *link, f downlinkQueuesUpdater) (err error) {
+	logger := log.FromContext(ctx)
+	pendingSession := dev.PendingSession
+	if pendingSession != nil {
+		logger = logger.WithField("pending_session_key_id", pendingSession.SessionKeyID)
+	}
+	currentSession := dev.Session
+	if currentSession != nil {
+		logger = logger.WithField("current_session_key_id", currentSession.SessionKeyID)
```
johanstokking: Hmm not sure if we need to call this current. Do we do this elsewhere in NS or AS?

Afaik we have "pending session" and "session". If we need to clarify anything about the latter, it would be that it's the session confirmed by the end device.

pkg/applicationserver/applicationserver.go
```diff
@@ -588,6 +592,202 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
...
+		EndDeviceIdentifiers: dev.EndDeviceIdentifiers,
+		Downlinks:            downlinks,
+	}
+	_, err = client.DownlinkQueueReplace(ctx, req, link.callOpts...)
+	return err
+}
+
+var errUnknownSession = errors.DefineNotFound("unknown_session", "unknown session")
+
+// recomputePendingQueue partitions the invalidated downlinks based on their session key ID. Downlinks which are part of the
```
johanstokking: > recomputePendingQueue partitions the invalidated downlinks based on their session key ID.

Please update this godoc to mention in one line what this method does

pkg/applicationserver/applicationserver.go
```diff
@@ -588,6 +592,202 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
...
+			// We cannot decode any of the downlinks if both sessions are missing.
+			return nil, nil, errNoDeviceSession
+		case dev.Session != nil:
+			// In the presence of a current session we copy the items from the current session to the pending session.
+			sourceQueue = currentQueue
+			sourceSession = dev.Session
+			logger.Debug("Initializing pending downlink queue from the current session")
+		case previousPendingSession != nil:
+			// In the presence of a previous session and the absence of a current session we migrate the items from the
+			// previous pending session to the new pending session.
```
johanstokking: Shouldn't we take the `session`'s queue and put that in the new pending queue?

adriansmares: Given the structure of the case, we entered the branch of `previousPendingSession != nil`, the branch of `dev.Session != nil` must be false, and as such there is no current session (or queue). If there is a current session, we indeed copy them to the queue of the pending session.

adriansmares:
Regarding the general comment about simplicity - the presence of two queues itself is not simple, at all. I do agree that the naming itself is not optimal, and I'll probably have to think of some new names.

I do have to say that the current changes (which introduce new verbs) are there because there is no longer a simple `recalculate from this to that` operation, due to the added complexity of the secondary queue.

Regarding the cases:

1. `Enqueueing downlink in both queues`: When an frontend (be it webhooks, MQTT etc.) submits a downlink queue operation, the AS encrypts it with both sessions (both `dev.Session` and `dev.PendingSession`) if they are available. 
 For both operations (`push` and `replace`) it means that we send both sets of downlinks in the same request, since the NS will automatically partition these two and do the changes to the required queue. 

2. `Re-encrypting invalidated downlink items on join-accept. I guess only the pending queue is involved here` . It depends:

 - If there is a `dev.Session`, we indeed just copy the downlinks from the `dev.Session` to the `dev.PendingSession`. We still have to submit both queues as part of the `replace`, since otherwise the NS will interpret this as emptying the `dev.Session` queue. 
  This case also applies if there is `dev.PendingSession` available, since I believe we should always consider the `dev.Session` queue as being the single source of truth, if it is available.

 - If there is no `dev.Session`, we check if we previously had a `dev.PendingSession`, and migrate the items from the previous pending session to the new pending session.

 - If both sessions are missing both we still have items which have been invalidated this is an error (and should not happen, ever).

3. `Confirm the pending session on first uplink, i.e. replace session by pending_session and delete pending_session`. I believe the following cases are possible:
- The `dev.Session` session key ID does not match, but the `dev.PendingSession`'s session key ID matches: the AS simply switches the sessions and deletes the pending session. No queue operations from AS to NS, since the NS should have already been able to switch the queues on it's own.

- The `dev.PendingSession` key does not match, but this session has already been confirmed before: The AS will delete the `dev.PendingSession` and start a new session. This requires a replace since this 3rd session is not (and I would say could not have been) stored in the NS.

4. The whole downlink queue has been invalidated (resulting in a new queue), or a message has been `nack`ed (resulting in a new queue which has the nacked item as a front item):
 We have to recompute the `dev.Session` queue (so decrypt and re-encrypt it again). If there is a `dev.PendingSession`, we must also encrypt this new queue using the `dev.PendingSession`, because otherwise the NS will empty the queue of the `dev.PendingSession`.

rvolosatovs:
> @rvolosatovs I've moved the Partition* utilities to pkg/ttnpb from pkg/networkserver since both components use them. Is this location correct or should I make a new package in pkg/util ?

Not a fan of it, but I guess we don't have a better option.

> I think the only part missing from #1954 is the switch between the pending and the current session in the NS on the first uplink. (cc @rvolosatovs)

Not sure what you mean. NS had been switching from pending session to current, as had AS, for a long time before this issue was created.

Answer to https://github.com/TheThingsNetwork/lorawan-stack/pull/2163/files#r393640144
> Is this true @rvolosatovs ?

Yes, https://github.com/TheThingsNetwork/lorawan-stack/blob/2868421fb964755c8302e92fa62067c7e49e05e9/pkg/networkserver/grpc_gsns.go#L1212
https://github.com/TheThingsNetwork/lorawan-stack/pull/2127

However, now looking at this PR and having a better understanding of what AS does here I don't think NS should do that, because `session.queued_application_downlinks` always contains the most recent version of the queue and `pending_session.queued_application_downlinks` is merely a (possibly out-of-date) duplicate of it maintained by AS.

To me it all seems very simple:

Consider:
- current queue `== session.queued_application_downlinks`
- pending queue `== pending_session.queued_application_downlinks`

1. When NS receives a join-accept, it sends the current queue and new key to AS
	- AS decrypts the received queue, reencrypts using new key and pushes the result to NS
2. When NS schedules a downlink with FPort==0 in <1.1, it sends the current queue to AS
	- AS decrypts the received queue, and:
		- reencrypts the queue using the pending session key(if present)
		- increases the frame counters for the queue and reencrypts using current session key
		- replaces the queue (NOTE: if AS could only replace the current queue, then it would not need to recalculate the pending queue and would only need to replace the current queue)
3. When a downlink is pushed to AS, it pushes a downlink to NS per-session
4. When downlinks are replaced in AS, it replaces downlinks per-session
5. When uplink is sent and 2 sessions are present in AS, no downlink queue update takes place, simply one of the sessions is removed.
6. `List` lists only the current queue

I think we should allow AS to explicitly specify the queue it wants to update/list in RPCs.

pkg/applicationserver/applicationserver.go
```diff
@@ -695,27 +895,29 @@ func (as *ApplicationServer) handleDownlinkQueueInvalidated(ctx context.Context,
...
 		_, err := as.deviceRegistry.Set(ctx, ids,
 			[]string{
+				"pending_session",
 				"session",
 				"skip_payload_crypto",
 			},
 			func(dev *ttnpb.EndDevice) (*ttnpb.EndDevice, []string, error) {
-				if err := as.recalculateDownlinkQueue(ctx, dev, nil, queue, msg.FCnt+1, link); err != nil {
+				queue := append([]*ttnpb.ApplicationDownlink{msg}, res.Downlinks...)
+				if err := as.replaceDownlinkQueue(ctx, dev, link, dev.Session, queue, msg.FCnt+1, true); err != nil {
```
rvolosatovs: I'd expect to see
```go
queue, err := as.recalculateDownlinkQueue(ctx, dev.Session, msg.FCnt+1, queue...)
```
With a following `client.DownlinkQueueReplace` using the returned queue.
Let's allow AS to specify which queue to replace https://github.com/TheThingsNetwork/lorawan-stack/pull/2163#issuecomment-600103266

pkg/applicationserver/applicationserver.go
```diff
@@ -438,35 +438,41 @@ func (as *ApplicationServer) DownlinkQueueList(ctx context.Context, ids ttnpb.En
...
-				session = s
-				break
-			}
-		}
-		if session == nil {
-			return nil, errNoDeviceSession
-		}
-		if session.AppSKey == nil {
-			return nil, errNoAppSKey
+	// Avoid hitting the device registry if no downlinks are to be decrypted.
```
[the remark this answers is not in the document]
adriansmares: Valid concern, will revert.

pkg/applicationserver/applicationserver.go
adriansmares: All of the queue operations in the AS have a rollback mechanism - if a queue `replace` fails, the AS attempts to replace the queue in the NS with an empty queue. This rollback boilerplate is part of the `replaceDownlinkQueue`.

The naming itself is not descriptive, and I'll update it.

rvolosatovs:
Blocked by #2177

pkg/applicationserver/applicationserver.go
rvolosatovs: So if the replace fails, AS tries to delete the queue?

adriansmares: Yes. This behavior has been here for a long time:
https://github.com/TheThingsNetwork/lorawan-stack/blob/8d8634e3e3f34e466b53eb3f9dcd9af85e96456a/pkg/applicationserver/applicationserver.go#L779-L780
https://github.com/TheThingsNetwork/lorawan-stack/blob/8d8634e3e3f34e466b53eb3f9dcd9af85e96456a/pkg/applicationserver/applicationserver.go#L639-L650

pkg/applicationserver/applicationserver.go
```diff
@@ -588,6 +592,202 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
...
+		// If something fails, clear the downlink queue as an empty downlink queue is better than a downlink queue
+		// with items that are encrypted with the wrong AppSKey.
+		if err != nil {
+			if pendingSession != nil {
+				pendingSession.LastAFCntDown = oldPendingLastAFCntDown
+			}
+			if currentSession != nil {
+				currentSession.LastAFCntDown = oldCurrentLastAFCntDown
+			}
+			logger.WithError(err).Warn("Recalculate downlink queue failed; clearing the downlink queue")
```
rvolosatovs:
```suggestion
			logger.WithError(err).Warn("Failed to recalculate downlink queue; clear the downlink queue")
```

pkg/applicationserver/applicationserver.go
```diff
@@ -588,6 +592,202 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
...
+	newQueue := make([]*ttnpb.ApplicationDownlink, 0, len(oldQueue))
+	for _, oldItem := range oldQueue {
+		logger := logger.WithFields(log.Fields(
+			"f_port", oldItem.FPort,
+			"f_cnt", oldItem.FCnt,
+			"session_key_id", oldItem.SessionKeyID,
+		))
+		frmPayload, err := crypto.DecryptDownlink(oldAppSKey, oldSession.DevAddr, oldItem.FCnt, oldItem.FRMPayload)
+		if err != nil {
+			logger.WithError(err).Warn("Drop downlink message; failed to decrypt")
```
rvolosatovs: Use either `<what happened>;<action>` form or `<action>;<what happened>`, but don't mix both in different logging statements

pkg/applicationserver/applicationserver.go
```diff
@@ -626,31 +826,31 @@ func (as *ApplicationServer) handleUplink(ctx context.Context, ids ttnpb.EndDevi
...
-					log.WithError(err).Warn("Failed to list downlink queue for recalculation; clearing the downlink queue")
-					req := &ttnpb.DownlinkQueueRequest{
-						EndDeviceIdentifiers: ids,
-					}
-					_, err = client.DownlinkQueueReplace(ctx, req, link.callOpts...)
+					// At this point, the application downlink queue in the Network Server is invalid; recalculation is necessary.
+					// Next AFCntDown 1 is assumed. If this is a LoRaWAN 1.0.x end device and the Network Server sent MAC layer
+					// downlink already, the Network Server will trigger the DownlinkQueueInvalidated event. Therefore, this
+					// recalculation may result in another recalculation.
+					client := ttnpb.NewAsNsClient(link.conn)
```
rvolosatovs: AS does not need to do anything here, because NS simply discards the pending session, and the current queue is already there.

pkg/applicationserver/applicationserver.go
```diff
@@ -588,6 +592,202 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
...
+		}
+		return currentQueue, newPendingQueue, nil
+	})
+}
+
+// replaceDownlinkQueue replaces the current and pending queues of the end device with the provided downlinks.
+// This method requires the given end device's session to be set. This method mutates the end device's session LastAFCntDown.
+// This method mutates the end device's pending session LastAFCntDown if the pending session exists and replacePending is true.
+// This method does not change the contents of the given downlinks.
+func (as *ApplicationServer) replaceDownlinkQueue(ctx context.Context, dev *ttnpb.EndDevice, link *link, previousSession *ttnpb.Session, downlinks []*ttnpb.ApplicationDownlink, nextAFCntDown uint32, replacePending bool) error {
```
rvolosatovs: I think what this method should do, is only recalculation of the active queue with a higher FCnt value. It should never modify the pending queue, because that is never necessary. You don't need the `previousSession` either (it's always `dev.Session`)
Also, name it `recalculateDownlinkQueue` like it was before.

pkg/applicationserver/applicationserver.go
```diff
@@ -588,6 +592,202 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
...
+var errUnknownSession = errors.DefineNotFound("unknown_session", "unknown session")
+
+// recomputePendingQueue partitions the invalidated downlinks based on their session key ID. Downlinks which are part of the
+// current session are not modified, while the downlinks from the previous pending session are migrated to the new pending
+// session.
+// This method requires the given end device's pending session to be set. This method mutates the end device's pending session LastAFCntDown.
+// This method does not change the contents of the given invalid downlink queue.
+func (as *ApplicationServer) recomputePendingQueue(ctx context.Context, dev *ttnpb.EndDevice, link *link, previousPendingSession *ttnpb.Session, invalidatedDownlinks []*ttnpb.ApplicationDownlink) error {
+	return as.changeDownlinkQueue(ctx, dev, link, func(ctx context.Context, dev *ttnpb.EndDevice) ([]*ttnpb.ApplicationDownlink, []*ttnpb.ApplicationDownlink, error) {
+		var previousPendingQueue, currentQueue []*ttnpb.ApplicationDownlink
```
rvolosatovs: This should simply do `migrated, err := as.migrateDownlinkQueue(ctx, dev.EndDeviceIdentifiers, invalidatedDownlinks, dev.Session, dev.PendingSession, 1)` and `return invalidatedQueue, migrated, nil`

There's no need for this method, in fact, just put this at `L582`

pkg/applicationserver/applicationserver.go
adriansmares: I disagree. Unless I'm missing something very obvious, here the following conditions hold:
1. `dev.Session == nil || !bytes.Equal(dev.Session.SessionKeyID, uplink.SessionKeyID)` (line 807).
2. `!(dev.PendingSession != nil && bytes.Equal(dev.PendingSession.SessionKeyID, uplink.SessionKeyID))` (line 808).

Due to 1, the downlink queue in NS for `dev.Session` cannot be valid, since it uses a different session key.

rvolosatovs: Assume device has current queue `CQ` and pending queue `PQ` (either of them may be `nil`)
1. Device sends a join-request, NS receives a join-accept from JS
2. AS receives `CQ` if not `nil` and `PQ` otherwise, reencrypts using the new key. Let's call the resulting queue `PQ'`. AS replaces the device queue, which is now `CQ ++ PQ'`
3. NS may arbitrary mutate `CQ` by scheduling application downlinks (not implemented currently). `PQ'` does not get mutated. Let's call the new, possibly mutated, current queue `CQ'`
4. Device sends an uplink using either the key used for `CQ'` or `PQ'`
5. NS discards the queue for which the key does not match and takes either `CQ'` or `PQ'` as the current queue according to that.
5. AS discards one of the sessions. There's nothing else to do, because both `CQ'` and `PQ'` are already valid.

rvolosatovs: This use case, in fact, is exactly the reason for the issue we're closing here (https://github.com/TheThingsNetwork/lorawan-stack/issues/1954).

> #### Why do we need this?
> To avoid hitting AS to recalculate the downlink queue when the end device > confirms the session. Having a downlink queue of both the current session > and the pending session in NS allows for downlink items to be available immediately.

rvolosatovs:
Also note that pending queue can only be modified if either application pushes downlinks or replaces downlinks, in all other cases it's immutable (so it never needs to be recalculated and no `ApplicationUp` message can trigger a change in pending queue)

pkg/applicationserver/applicationserver.go
```diff
@@ -583,6 +588,208 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
...
+			ClassBC:        oldItem.ClassBC,
+			Priority:       oldItem.Priority,
+			CorrelationIDs: oldItem.CorrelationIDs,
+		}
+		newItem.FRMPayload, err = crypto.EncryptDownlink(newAppSKey, newSession.DevAddr, newItem.FCnt, frmPayload)
+		if err != nil {
+			logger.WithError(err).Warn("Failed to encrypt downlink message; dropping item")
+			registerDropDownlink(ctx, ids, oldItem, err)
+			continue
+		}
```
rvolosatovs:
```suggestion
		newFRMPayload, err := crypto.EncryptDownlink(newAppSKey, newSession.DevAddr, newItem.FCnt, frmPayload)
		if err != nil {
			logger.WithError(err).Warn("Failed to encrypt downlink message; dropping item")
			registerDropDownlink(ctx, ids, oldItem, err)
			continue
		}
		newItem := &ttnpb.ApplicationDownlink{
			SessionKeyID:   newSession.SessionKeyID,
			FPort:          oldItem.FPort,
			FRMPayload: 	newFRMPayload,
			FCnt:           newSession.LastAFCntDown + 1,
			Confirmed:      oldItem.Confirmed,
			ClassBC:        oldItem.ClassBC,
			Priority:       oldItem.Priority,
			CorrelationIDs: oldItem.CorrelationIDs,
		}
		
```

pkg/applicationserver/applicationserver.go
```diff
@@ -600,7 +807,6 @@ func (as *ApplicationServer) handleUplink(ctx context.Context, ids ttnpb.EndDevi
 				return nil, nil, errDeviceNotFound.WithAttributes("device_uid", unique.ID(ctx, ids))
 			}
 			if dev.Session == nil || !bytes.Equal(dev.Session.SessionKeyID, uplink.SessionKeyID) {
```
rvolosatovs: The nesting here is very difficult to understand, consider refactoring (removing unnecessary nesting) and/or adding comments

pkg/applicationserver/applicationserver.go
```diff
@@ -621,31 +828,31 @@ func (as *ApplicationServer) handleUplink(ctx context.Context, ids ttnpb.EndDevi
 					dev.DevAddr = ids.DevAddr
 					mask = append(mask, "session", "ids.dev_addr")
 					logger.Debug("Restored session")
```
rvolosatovs: This is not necessarily restoration, right? If AS did not receive the join-accept, the uplink may be for a totally new session, which is not initialized

pkg/applicationserver/applicationserver.go
```diff
@@ -583,6 +588,208 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
 	return nil
 }
 
+// downlinkQueueTransaction represents a transaction to be run on the downlink queues of the provided device.
+type downlinkQueueTransaction func(context.Context, *ttnpb.EndDevice) error
+
+// runDownlinkQueueTransaction runs the provided downlink queue transaction on the device. If the transaction
+// fails, the LastAFCntDown session values are restored to their previous values and the downlink queue is reset.
```
rvolosatovs: Something's off here

pkg/applicationserver/applicationserver.go
```diff
@@ -621,31 +828,31 @@ func (as *ApplicationServer) handleUplink(ctx context.Context, ids ttnpb.EndDevi
...
+						req := &ttnpb.DownlinkQueueRequest{
+							EndDeviceIdentifiers: ids,
+						}
+						_, err = client.DownlinkQueueReplace(ctx, req, link.callOpts...)
+						if err != nil {
+							logger.WithError(err).Warn("Failed to clear the downlink queue; any queued items in the Network Server are invalid")
+							events.Publish(evtInvalidQueueDataDown(ctx, ids, err))
+						} else {
+							events.Publish(evtLostQueueDataDown(ctx, ids, err))
+						}
```
rvolosatovs: Consider extracting this into a function/method and use both here and in `runDownlinkQueueTransaction`

pkg/applicationserver/applicationserver.go
```diff
@@ -621,31 +828,31 @@ func (as *ApplicationServer) handleUplink(ctx context.Context, ids ttnpb.EndDevi
...
+					} else if err := as.recalculateDownlinkQueue(ctx, dev, link, previousSession, res.Downlinks, 1); err != nil {
+						logger.WithError(err).Warn("Failed to recalculate downlink queue; items lost")
 					}
```
rvolosatovs:
```suggestion
					}
					if err := as.recalculateDownlinkQueue(ctx, dev, link, previousSession, res.Downlinks, 1); err != nil {
						logger.WithError(err).Warn("Failed to recalculate downlink queue; items lost")
					}
```

pkg/applicationserver/applicationserver.go
```diff
@@ -610,6 +816,7 @@ func (as *ApplicationServer) handleUplink(ctx context.Context, ids ttnpb.EndDevi
 					if err != nil {
 						return nil, nil, errFetchAppSKey.WithCause(err)
 					}
+					previousSession := dev.Session
```
rvolosatovs: This will cause a `panic` in `recalculateDownlinkQueue` if `nil`
https://github.com/TheThingsNetwork/lorawan-stack/pull/2163/files#diff-2719373b9eca80ef1abcf09309389399R715 https://github.com/TheThingsNetwork/lorawan-stack/pull/2163/files#diff-2719373b9eca80ef1abcf09309389399R725

pkg/applicationserver/applicationserver.go
```diff
@@ -621,31 +828,31 @@ func (as *ApplicationServer) handleUplink(ctx context.Context, ids ttnpb.EndDevi
...
+						logger.WithError(err).Warn("Failed to recalculate downlink queue; items lost")
 					}
-				} else if err := as.recalculateDownlinkQueue(ctx, dev, previousSession, res.Downlinks, 1, link); err != nil {
-					log.WithError(err).Warn("Failed to recalculate downlink queue")
 				}
+				dev.PendingSession = nil
+				mask = append(mask, "pending_session")
 			} else if dev.Session.AppSKey == nil {
 				return nil, nil, errNoAppSKey.New()
 			}
```
rvolosatovs: At this point `dev.Session` must be non-`nil` and `AppSKey` being `nil` must cause an error in pending session switch as well (https://github.com/TheThingsNetwork/lorawan-stack/pull/2163/files#diff-2719373b9eca80ef1abcf09309389399R810)
```suggestion
			}
			if dev.Session.AppSKey == nil {
				return nil, nil, errNoAppSKey.New()
			}
```

pkg/applicationserver/applicationserver.go
```diff
@@ -621,31 +828,31 @@ func (as *ApplicationServer) handleUplink(ctx context.Context, ids ttnpb.EndDevi
...
+							EndDeviceIdentifiers: ids,
+						}
+						_, err = client.DownlinkQueueReplace(ctx, req, link.callOpts...)
+						if err != nil {
+							logger.WithError(err).Warn("Failed to clear the downlink queue; any queued items in the Network Server are invalid")
+							events.Publish(evtInvalidQueueDataDown(ctx, ids, err))
+						} else {
+							events.Publish(evtLostQueueDataDown(ctx, ids, err))
+						}
+					} else if err := as.recalculateDownlinkQueue(ctx, dev, link, previousSession, res.Downlinks, 1); err != nil {
```
rvolosatovs: This will not account for known `dev.Session` downlinks at https://github.com/TheThingsNetwork/lorawan-stack/pull/2163/files#diff-2719373b9eca80ef1abcf09309389399R722 and incorrectly observe drops.

pkg/applicationserver/applicationserver.go
```diff
@@ -677,7 +884,7 @@ func (as *ApplicationServer) handleDownlinkQueueInvalidated(ctx context.Context,
 			if dev == nil {
 				return nil, nil, errDeviceNotFound.WithAttributes("device_uid", unique.ID(ctx, ids))
 			}
-			if err := as.recalculateDownlinkQueue(ctx, dev, nil, invalid.Downlinks, invalid.LastFCntDown+1, link); err != nil {
+			if err := as.recalculateDownlinkQueue(ctx, dev, link, dev.Session, invalid.Downlinks, invalid.LastFCntDown+1); err != nil {
```
rvolosatovs: This causes a `panic` if `dev.Session` is `nil`

pkg/applicationserver/applicationserver.go
```diff
@@ -690,27 +897,29 @@ func (as *ApplicationServer) handleDownlinkQueueInvalidated(ctx context.Context,
...
 		_, err := as.deviceRegistry.Set(ctx, ids,
 			[]string{
+				"pending_session",
 				"session",
 				"skip_payload_crypto",
 			},
 			func(dev *ttnpb.EndDevice) (*ttnpb.EndDevice, []string, error) {
-				if err := as.recalculateDownlinkQueue(ctx, dev, nil, queue, msg.FCnt+1, link); err != nil {
+				queue := append([]*ttnpb.ApplicationDownlink{msg}, res.Downlinks...)
+				if err := as.recalculateDownlinkQueue(ctx, dev, link, dev.Session, queue, msg.FCnt+1); err != nil {
```
rvolosatovs: Same as above, this neglects the downlinks for `dev.PendingSession` and will register them as dropped (as well as publish an event)

pkg/applicationserver/applicationserver.go
```diff
@@ -583,6 +588,208 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
...
+			logger.Debug("Initializing pending downlink queue from the current session")
+		case previousPendingSession != nil:
+			// In the presence of a previous session and the absence of a current session we migrate the items from the
+			// previous pending session to the new pending session.
+			sourceQueue = previousPendingQueue
+			sourceSession = previousPendingSession
+			logger.Debug("Initializing pending downlink queue from the previous pending session")
+		}
+
+		newPendingQueue, err := as.migrateDownlinkQueue(ctx, dev.EndDeviceIdentifiers, sourceQueue, sourceSession, dev.PendingSession, 1)
```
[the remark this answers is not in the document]
adriansmares: The queue `FCnt` cannot start from 0. It would mean that the `LastAFCntDown` is -1, which cannot the case since `LastAFCntDown` is `uint32`.

pkg/applicationserver/applicationserver.go
adriansmares: I cannot really find a good name here - `Restored previous session` or `Restored external session` seem off here.

pkg/applicationserver/applicationserver.go
adriansmares: How can a device with no active session have its queue invalidated ?

pkg/applicationserver/applicationserver.go
adriansmares: `res.Downlinks` is invalid if the error of the previous `if` is not `nil`.

pkg/applicationserver/applicationserver.go
rvolosatovs: > @rvolosatovs so is NS now discarding the current queue on join-accept, or not? The items in there are invalid for sure, so I would expect so, hence the original comment here.

No, NS discards the queue as part of session switch when the uplink confirming the new session is received. It does send the "current" queue to AS as part of join-accept.

pkg/applicationserver/applicationserver.go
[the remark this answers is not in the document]
rvolosatovs: Isn't `FCnt` starting from `0` a part of LoRaWAN certification? I know the AS part is not specified yet, but I can very much imagine that would be the requirement in the future(isn't it for NS?). Besides, there's no functional reason why `FCnt` would not be able to start from `0`, it's purely implementation-specific. You can look at NS implementation for inspiration perhaps? https://github.com/TheThingsNetwork/lorawan-stack/commit/aed9da0cfcfddd1eb43493614d8b003be282c8da As far as I understand AS does not store "recent downlinks", so it might be trickier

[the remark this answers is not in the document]
johanstokking: > As far as I understand AS does not store "recent downlinks", so it might be trickier

Indeed, let's leave it for now

> I know the AS part is not specified yet, but I can very much imagine that would be the requirement in the future(isn't it for NS?)

AFCntDown is up to AS indeed, and that is not specified to start from 0.

pkg/applicationserver/applicationserver.go
```diff
@@ -632,21 +863,16 @@ func (as *ApplicationServer) handleUplink(ctx context.Context, ids ttnpb.EndDevi
...
-					req := &ttnpb.DownlinkQueueRequest{
-						EndDeviceIdentifiers: ids,
-					}
-					_, err = client.DownlinkQueueReplace(ctx, req, link.callOpts...)
-					if err != nil {
-						log.WithError(err).Warn("Failed to clear the downlink queue; any queued items in the Network Server are invalid")
-						events.Publish(evtInvalidQueueDataDown(ctx, ids, err))
-					} else {
-						events.Publish(evtLostQueueDataDown(ctx, ids, err))
+					logger.WithError(err).Warn("Failed to list downlink queue for recalculation; clearing the downlink queue")
```
johanstokking, line 866: nit; imperative clear

pkg/applicationserver/applicationserver.go
```diff
@@ -632,21 +863,16 @@ func (as *ApplicationServer) handleUplink(ctx context.Context, ids ttnpb.EndDevi
...
+					as.resetInvalidDownlinkQueue(ctx, ids, link)
+				} else {
+					previousQueue, _ := ttnpb.PartitionDownlinksBySessionKeyIDEquality(previousSession.SessionKeyID, res.Downlinks...)
+					if err := as.recalculateDownlinkQueue(ctx, dev, link, previousSession, previousQueue, 1); err != nil {
+						logger.WithError(err).Warn("Failed to recalculate downlink queue; items lost")
```
johanstokking, line 871: Should there be a lost event here?

pkg/applicationserver/applicationserver.go
johanstokking: So what do we do with this?

pkg/applicationserver/applicationserver.go
johanstokking: Is this addressed?

pkg/applicationserver/applicationserver.go
adriansmares: I've added a `nil` check above (before re-requesting the review you just gave).

pkg/applicationserver/applicationserver.go
adriansmares: Yes, the `queue` is partitioned now before adding the `nacked` element to the front.

pkg/applicationserver/applicationserver.go
adriansmares: `recalculateDownlinkQueue` calls `runDownlinkQueueTransaction` which will call `resetInvalidDownlinkQueue` on failure thus issuing the events.

johanstokking:
Will this close https://github.com/TheThingsNetwork/lorawan-stack/issues/1971 too?

adriansmares:
I finally got around testing this, and successfully performed the following tests:

- Schedule a downlink before the device joins for the first time - the downlink is received on the confirmation uplink.
- Have a device join, send 4-5 uplinks, and then schedule a downlink - the downlinks is received on the next uplink, as expected.
- Schedule a downlink, have the device rejoin - it receives the downlink on the confirmation uplink, as expected.

I've also ran the steps in https://github.com/TheThingsNetwork/lorawan-stack/issues/1971#issuecomment-584211868 and can confirm that the downlink queue invalidation issue is no longer there (and the second scheduling of a downlink does not fail anymore).

I've also tested it with the LoRaCloud DAS integration (just for the sake of realism) and it worked fine (both with one session and with rejoins).

PR #2190, opened 2020-03-18
Fixes for GS flaky tests

neoaggelos:
#### Summary

References #1332

#### Changes

- Fix flaky tests in mqtt_test.go
- Fix flaky test for context exceeded

#### Notes for Reviewers

- Nothing for the marshal errors, these are part of #2108

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver_test.go
+++ b/pkg/gatewayserver/gatewayserver_test.go
@@ -556,17 +556,53 @@ func TestGatewayServer(t *testing.T) {
 					},
 				} {
 					t.Run(ctc.Name, func(t *testing.T) {
-						ctx, cancel := context.WithDeadline(ctx, time.Now().Add(timeout))
+						ctx, cancel := context.WithCancel(ctx)
 						upCh := make(chan *ttnpb.GatewayUp)
 						downCh := make(chan *ttnpb.GatewayDown)
-						err := ptc.Link(ctx, t, ctc.ID, ctc.Key, upCh, downCh)
-						cancel()
-						if errors.IsDeadlineExceeded(err) {
-							if !ptc.TimeoutOnInvalidAuth && !ptc.ValidAuth(ctx, ctc.ID, ctc.Key) {
-								t.Fatal("Expected link error due to invalid auth")
+
+						upEvents := map[string]events.Channel{}
+						for _, event := range []string{"gs.gateway.connect"} {
+							upEvents[event] = make(events.Channel, 5)
+						}
+						defer test.SetDefaultEventsPubSub(&test.MockEventPubSub{
+							PublishFunc: func(ev events.Event) {
+								switch name := ev.Name(); name {
+								case "gs.gateway.connect":
+									go func() {
+										upEvents[name] <- ev
+									}()
+								default:
+									t.Logf("%s event published", name)
+								}
+							},
+						})()
+
+						connectedWithInvalidAuth := make(chan struct{}, 1)
+						expectedProperLink := make(chan struct{}, 1)
+						go func() {
+							select {
+							case <-upEvents["gs.gateway.connect"]:
+								if !ptc.ValidAuth(ctx, ctc.ID, ctc.Key) {
+									connectedWithInvalidAuth <- struct{}{}
+								}
+							case <-time.After(timeout):
+								if ptc.ValidAuth(ctx, ctc.ID, ctc.Key) {
+									expectedProperLink <- struct{}{}
+								}
 							}
-						} else if ptc.ValidAuth(ctx, ctc.ID, ctc.Key) {
-							t.Fatalf("Expected deadline exceeded with valid auth, but have %v", err)
+							time.Sleep(test.Delay)
+							cancel()
+						}()
+						err := ptc.Link(ctx, t, ctc.ID, ctc.Key, upCh, downCh)
+						if !errors.IsCanceled(err) && ptc.ValidAuth(ctx, ctc.ID, ctc.Key) {
+							t.Fatalf("Expect canceled context but have %v", err)
+						}
+						select {
+						case <-connectedWithInvalidAuth:
+							t.Fatal("Expected link error due to invalid auth")
+						case <-expectedProperLink:
+							t.Fatal("Expected proper link")
+						default:
 						}
 					})
 				}
@@ -599,7 +635,7 @@ func TestGatewayServer(t *testing.T) {
 				case <-time.After(timeout):
 				}
 
-				ctx2, cancel2 := context.WithDeadline(ctx, time.Now().Add(timeout))
+				ctx2, cancel2 := context.WithDeadline(ctx, time.Now().Add(2*timeout))
 				upCh := make(chan *ttnpb.GatewayUp)
 				downCh := make(chan *ttnpb.GatewayDown)
 				err := ptc.Link(ctx2, t, id, registeredGatewayKey, upCh, downCh)
@@ -610,13 +646,13 @@ func TestGatewayServer(t *testing.T) {
 				select {
 				case <-ctx1.Done():
 					t.Logf("First connection failed when second connected with %v", ctx1.Err())
-				case <-time.After(timeout):
+				case <-time.After(2 * timeout):
 					t.Fatalf("Expected link failure on first connection when second connected")
 				}
 			})
 
 			// Wait for gateway disconnection to be processed.
-			time.Sleep(timeout)
+			time.Sleep(3 * timeout)
 
 			t.Run(fmt.Sprintf("Traffic/%v", ptc.Protocol), func(t *testing.T) {
 				a := assertions.New(t)
@@ -968,13 +1004,13 @@ func TestGatewayServer(t *testing.T) {
 							a := assertions.New(t)
 
 							upEvents := map[string]events.Channel{}
-							for _, event := range []string{"gs.up.receive", "gs.down.tx.success", "gs.down.tx.fail", "gs.status.receive"} {
+							for _, event := range []string{"gs.up.receive", "gs.down.tx.success", "gs.down.tx.fail", "gs.status.receive", "gs.status.forward"} {
 								upEvents[event] = make(events.Channel, 5)
 							}
 							defer test.SetDefaultEventsPubSub(&test.MockEventPubSub{
 								PublishFunc: func(ev events.Event) {
 									switch name := ev.Name(); name {
-									case "gs.up.receive", "gs.down.tx.success", "gs.down.tx.fail", "gs.status.receive":
+									case "gs.up.receive", "gs.down.tx.success", "gs.down.tx.fail", "gs.status.receive", "gs.status.forward":
 										go func() {
 											upEvents[name] <- ev
 										}()
@@ -1054,8 +1090,11 @@ func TestGatewayServer(t *testing.T) {
 									t.Fatal("Expected gateway status event timeout")
 								}
 
-								// Wait for gateway status to be processed; no event available here.
-								time.Sleep(timeout)
+								select {
+								case <-upEvents["gs.status.forward"]:
+								case <-time.After(timeout):
+									t.Fatal("Expected gateway status forward event timeout")
+								}
 							}
 
 							conn, ok := gs.GetConnection(ctx, ids)
--- a/pkg/gatewayserver/io/mqtt/mqtt_test.go
+++ b/pkg/gatewayserver/io/mqtt/mqtt_test.go
@@ -45,7 +45,7 @@ var (
 	registeredGatewayUID = unique.ID(test.Context(), registeredGatewayID)
 	registeredGatewayKey = "test-key"
 
-	timeout = 10 * test.Delay
+	timeout = 20 * test.Delay
 )
 
 func TestAuthentication(t *testing.T) {


pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -1311,6 +1323,16 @@ func TestGatewayServer(t *testing.T) {
...
+							for i := 0; i < 5; i++ {
+								stats, err := statsClient.GetGatewayConnectionStats(statsCtx, &ids)
+								if !a.So(err, should.BeNil) {
+									t.FailNow()
+								}
+								if stats.DownlinkCount == uint64(downlinkCount) {
+									break
+								}
+								time.Sleep(timeout)
+							}
```
johanstokking: Same

pkg/gatewayserver/io/mqtt/mqtt_test.go
```diff
@@ -112,6 +112,8 @@ func TestAuthentication(t *testing.T) {
 			client := mqtt.NewClient(clientOpts)
 			token := client.Connect()
 			if tc.OK {
+				for i := 0; i < 3 && token.WaitTimeout(timeout); i++ {
+				}
```
johanstokking: This looks wrong; if you need up to 3 times the wait timeout, why not wait for `3 * timeout` ?

pkg/gatewayserver/io/mqtt/mqtt_test.go
```diff
@@ -165,6 +167,9 @@ func TestTraffic(t *testing.T) {
 	clientOpts.SetPassword(registeredGatewayKey)
 	client := mqtt.NewClient(clientOpts)
 	token := client.Connect()
+
+	for i := 0; i < 3 && token.WaitTimeout(timeout); i++ {
+	}
```
johanstokking: Idem

pkg/gatewayserver/io/mqtt/mqtt_test.go
```diff
@@ -248,6 +253,8 @@ func TestTraffic(t *testing.T) {
 				buf, err := tc.Message.Marshal()
 				a.So(err, should.BeNil)
 				token := client.Publish(tc.Topic, 1, false, buf)
+				for i := 0; i < 3 && token.WaitTimeout(timeout); i++ {
+				}
```
johanstokking: Idem

pkg/gatewayserver/redis/registry.go
```diff
@@ -38,6 +39,7 @@ func (r *GatewayConnectionStatsRegistry) Set(ctx context.Context, ids ttnpb.Gate
 		return r.Redis.Del(r.key(uid)).Err()
 	}
 
+	fmt.Println(">>>>>>>>>>>", stats)
```
johanstokking: Remove and check diff before requesting review

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -560,6 +560,7 @@ func TestGatewayServer(t *testing.T) {
 						upCh := make(chan *ttnpb.GatewayUp)
 						downCh := make(chan *ttnpb.GatewayDown)
 						err := ptc.Link(ctx, t, ctc.ID, ctc.Key, upCh, downCh)
+						time.Sleep(2 * test.Delay)
```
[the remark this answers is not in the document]
neoaggelos: Hmm, that was not the best. Changed to try again a few times when getting context cancelled.

pkg/gatewayserver/io/mqtt/mqtt_test.go
neoaggelos: Went with this because you don't always have to wait for 3 times the timeout. That would make the tests take an unnecessarily long time under normal circumstances. Thus, we only wait for as along there is a reason to. If the test takes a bit longer, then it will not fail as it currently does.

pkg/gatewayserver/io/mqtt/mqtt_test.go
neoaggelos: See comment

pkg/gatewayserver/io/mqtt/mqtt_test.go
neoaggelos: See comment

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -1045,6 +1047,16 @@ func TestGatewayServer(t *testing.T) {
...
+							for i := 0; i < 5; i++ {
+								stats, err := statsClient.GetGatewayConnectionStats(statsCtx, &ids)
+								if !a.So(err, should.BeNil) {
+									t.FailNow()
+								}
+								if stats.UplinkCount == uint64(uplinkCount) {
+									break
+								}
+								time.Sleep(timeout)
+							}
```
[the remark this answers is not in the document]
neoaggelos: Moved assertions in the loop.

pkg/gatewayserver/gatewayserver_test.go
neoaggelos: Moved assertions in the loop

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -556,11 +556,18 @@ func TestGatewayServer(t *testing.T) {
...
-						err := ptc.Link(ctx, t, ctc.ID, ctc.Key, upCh, downCh)
-						cancel()
+						var err error
+						for i := 0; i < 5; i++ {
+							ctx, cancel := context.WithDeadline(ctx, time.Now().Add(timeout))
+							upCh := make(chan *ttnpb.GatewayUp)
+							downCh := make(chan *ttnpb.GatewayDown)
+							err = ptc.Link(ctx, t, ctc.ID, ctc.Key, upCh, downCh)
+							cancel()
+							if err.Error() != "error:unknown:unknown (context deadline exceeded)" {
```
johanstokking: Nah, use proper error checking, like `errors.IsDeadlineExceeded(err)`, as is also being used below

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -556,11 +556,18 @@ func TestGatewayServer(t *testing.T) {
 					},
 				} {
 					t.Run(ctc.Name, func(t *testing.T) {
-						ctx, cancel := context.WithDeadline(ctx, time.Now().Add(timeout))
-						upCh := make(chan *ttnpb.GatewayUp)
-						downCh := make(chan *ttnpb.GatewayDown)
-						err := ptc.Link(ctx, t, ctc.ID, ctc.Key, upCh, downCh)
-						cancel()
+						var err error
+						for i := 0; i < 5; i++ {
```
johanstokking: I don't fully understand this. Are we trying 5 times with the same timeout to link? And hope it works one of those 5 times?

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -599,11 +606,19 @@ func TestGatewayServer(t *testing.T) {
...
+					upCh := make(chan *ttnpb.GatewayUp)
+					downCh := make(chan *ttnpb.GatewayDown)
+					err = ptc.Link(ctx2, t, id, registeredGatewayKey, upCh, downCh)
+					cancel2()
+
+					if err.Error() != "error:unknown:unknown (context deadline exceeded)" {
+						break
+					}
+					time.Sleep(test.Delay)
+				}
```
johanstokking: This seems like retrying the same thing.

Should the deadline timeout be longer?

pkg/gatewayserver/io/mqtt/mqtt_test.go
johanstokking: How is this different then?

The `WaitTimeout()` does not actually wait for `timeout`, it waits as long as it's necessary until the wait times out

pkg/gatewayserver/gatewayserver_test.go
neoaggelos: What I understand is that the tests checks what happens when linking with/without auth. In case the auth is valid, then `ptc.Link` will return once the context deadline has been exceeded. This is OK, because it verifies that the auth check has actually been checked.

`error:unknown:unknown (context deadline exceeded)` is different: It shows that the deadline exceeded before even getting to the part of validating the gateway auth. In that sense, it is wrong to consider it a proper deadline exceeded error (because at that point the test has not finished all its checks yet).

At least that's what I gather to be the original intent of the test here. And this is why the test is repeated in case the test repeats if the context expires earlier. Merely increasing the timeout is also not a solution, because it can still lead to failures.

pkg/gatewayserver/gatewayserver_test.go
neoaggelos: Yes, see comment below

pkg/gatewayserver/gatewayserver_test.go
neoaggelos: See comment above. Merely increasing the timeout does not solve this (I also got CI failures even with `2*timeout`)

pkg/gatewayserver/io/mqtt/mqtt_test.go
neoaggelos: Unless I have misinterpreted the way `WaitTimeout` works:

The loop repeats as long as `WaitTimeout()` returns false (which means that no traffic has arrived yet. For the test we expect it to return `true`. Initially, the test will fail if that does not happen within the initial `timeout` duration. With this loop, we retry a few times (in `timeout` intervals) before we conclude that the test has failed. If `WaitTimeout()` returns `true` fast, then we also finish fast. If it is late, then we give it more time to succeed before failing.

Also doing the same below.

pkg/gatewayserver/gatewayserver_test.go
johanstokking: Indeed, if `Link()` is successful, it doesn't return and it gets canceled via timeout, and we check if that is the error that occurred.

But what happens here exactly then? Is the context deadline exceeded before it can actually test anything?

If that is the case, shouldn't we give more time to the linking until the context times out, i.e. shouldn't the `WithTimeout()` value be higher?

What error is `err:unknown:unknown (context deadline exceeded)`? That seems like a `context.DeadlineExceeded` as cause, in an unknown error of ours.

pkg/gatewayserver/io/mqtt/mqtt_test.go
neoaggelos: Yeah, upon further inspection, I had misinterpreted how `WaitTimeout` worked. Reverted this and doubled the timeout duration instead. I think we will keep getting CI failures even after that, though.

pkg/gatewayserver/gatewayserver_test.go
neoaggelos: Based on testing, the error is returned by `GetConcentratorConfig` in https://github.com/TheThingsNetwork/lorawan-stack/blob/80083f2e26de5565ab4d3e6fa2c4067113d37c5e/pkg/gatewayserver/gatewayserver_test.go#L171

> But what happens here exactly then? Is the context deadline exceeded before it can actually test anything?

Yes

> If that is the case, shouldn't we give more time to the linking until the context times out, i.e. shouldn't the WithTimeout() value be higher?

Also yes, but this will severely slow down the tests. Also, Travis can be very slow at times and will fail even with a 2*timeout. We can also do that instead, but we will keep getting CI failures, just less frequently

johanstokking: So what is your proposal?

pkg/gatewayserver/gatewayserver_test.go
johanstokking: This is not acceptable

pkg/gatewayserver/gatewayserver_test.go
johanstokking: Then we have to increase it even more

pkg/gatewayserver/gatewayserver_test.go
neoaggelos: The simple solution I can think of is retry the test if it fails because the context expired early (the current change). 

A more elaborate approach could be to have an event on authentication success and wait for that instead (so that the test is not based on waiting for an arbitrary amount of time instead)

pkg/gatewayserver/gatewayserver_test.go
neoaggelos: We repeat only if the error is that the context expired early, for a max of 5 times. Maybe it is better to generate a proper event and check for that on the test.

pkg/gatewayserver/gatewayserver_test.go
johanstokking: > The simple solution I can think of is retry the test if it fails because the context expired early (the current change)

This is not a solution. This only reducing the probability of the flaky test.

Why not simply wait longer for the link to establish, i.e. `timeout*N` for the deadline?

pkg/gatewayserver/gatewayserver_test.go
johanstokking: See above

An event is also a good solution

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -62,7 +62,7 @@ var (
 	unregisteredGatewayID  = "eui-bbff000000000000"
 	unregisteredGatewayEUI = types.EUI64{0xBB, 0xFF, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}
 
-	timeout        = (1 << 5) * test.Delay
+	timeout        = (1 << 7) * test.Delay
```
johanstokking: Multiply the `timeout` selectively, not GS wide

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -1045,17 +1045,23 @@ func TestGatewayServer(t *testing.T) {
 							a.So(ok, should.BeTrue)
 							gs.UpdateConnectionStats(ctx, conn)
 
-							stats, err := statsClient.GetGatewayConnectionStats(statsCtx, &ids)
-							if !a.So(err, should.BeNil) {
-								t.FailNow()
-							}
-							a.So(stats.UplinkCount, should.Equal, uplinkCount)
+							for i := 0; i < 5; i++ {
```
johanstokking: Remove retry of lakes

pkg/gatewayserver/io/mqtt/mqtt_test.go
```diff
@@ -45,7 +45,7 @@ var (
 	registeredGatewayUID = unique.ID(test.Context(), registeredGatewayID)
 	registeredGatewayKey = "test-key"
 
-	timeout = 10 * test.Delay
+	timeout = 30 * test.Delay
```
johanstokking: Is this really necessary?

pkg/gatewayserver/gatewayserver_test.go
neoaggelos: Rolled back

pkg/gatewayserver/gatewayserver_test.go
neoaggelos: Changed test to use the `gs.gateway.connect` event. Can you make sure all cases are checked as before (also cc @KrishnaIyer)

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -556,17 +556,44 @@ func TestGatewayServer(t *testing.T) {
...
+									t.Logf("%s event published", name)
+								}
+							},
+						})()
+
+						connected := false
+						go func() {
+							select {
+							case <-upEvents["gs.gateway.connect"]:
+								connected = true
```
johanstokking: Hmm why not select this channel read after `Link()`, to avoid a dirty write to `connected`?

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -556,17 +556,44 @@ func TestGatewayServer(t *testing.T) {
...
-						} else if ptc.ValidAuth(ctx, ctc.ID, ctc.Key) {
-							t.Fatalf("Expected deadline exceeded with valid auth, but have %v", err)
+							time.Sleep(test.Delay)
+							cancel()
+						}()
+						err := ptc.Link(ctx, t, ctc.ID, ctc.Key, upCh, downCh)
+						if connected && !ptc.ValidAuth(ctx, ctc.ID, ctc.Key) {
+							t.Fatal("Expected link error due to invalid auth")
+						} else if connected && !errors.IsCanceled(err) {
+							t.Fatalf("Expect cancelled context but have %v", err)
```
johanstokking: `canceled`

pkg/gatewayserver/gatewayserver_test.go
neoaggelos: We expect `Link` to return after the context has been cancelled, with in turn must happen after reading the `gs.gateway.connect` event. Dropped the `connected` variable though, and switched to using signal channels for errors.

Decided to use channels there because calling `t.Fatal()` inside goroutines leads to panics when running the tests.

PR #2196, opened 2020-03-19
Improve error stacks

htdvisser:
#### Summary

This PR introduces a `New` function on `errors.Definition` that builds and returns an `errors.Error`, which then also contains a stack trace. In many other places, we use `(errors.Definition).WithAttributes` or `(errors.Definition).WithCause`, which does the same. This PR makes sure we also have stack traces for other errors.

#### Notes for Reviewers

This is a relatively simple pull request, but it touches a lot of files, sorry for that.

### DIFF
--- a/pkg/applicationserver/applicationserver.go
+++ b/pkg/applicationserver/applicationserver.go
@@ -449,10 +449,10 @@ func (as *ApplicationServer) DownlinkQueueList(ctx context.Context, ids ttnpb.En
 			}
 		}
 		if session == nil {
-			return nil, errNoDeviceSession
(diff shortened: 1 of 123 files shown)


rvolosatovs:
Also, instead of introducing `.New()` we could refactor the definition to be a function, so instead of `errFoo.New()`, we'd be forced to do `errFoo()`

htdvisser:
I actually did that at first, but then found out that we have quite a few situations where we do `errSomeError.WithCause(errSomeCause)`. Requiring `New` on that `errSomeCause` doesn't add much value, so I decided not to do that.

PR #2208, opened 2020-03-21
Fix too restrictive MQTT client validation

### LINKED ISSUE #1737: PubSub MQTT input field validation is too strict
#### Summary
When configuring an MQTT PubSub Integration valid entries for client-id or username are blocked by UI validation. Using http://www.dioty.co/mqtt provides an email address for username. This was not allowed. client-id or username containing uppercase letters are also not allowed.

#### Steps to Reproduce

1. Select an Application
2. Add a new PubSub Integration
3. Select MQTT as type
4. Enter "aaaBBB" as client-id or username

#### What do you see now?

image

Less restrictive validation should be used.

#### Environment

v3.3.2

Review validation requirements.

No

kschiffer:
#### Summary
Closes #1737 

#### Changes
- Add new regex to check that strings contain no spaces
- Update pubsub form validation schema

#### Notes for Reviewers
Had trouble finding good info on what exactly are validation rules for mqtt client ids and usernames, but it seems to be quite permissive. So I went with disallowing spaces only.

#### Checklist


pkg/webui/console/components/pubsub-form/validation-schema.js
```diff
@@ -71,14 +72,14 @@ export default Yup.object().shape({
...
-        .matches(idRegexp, sharedMessages.validateIdFormat)
+        .matches(noSpacesRegexp, sharedMessages.validateNoSpaces)
         .min(2, sharedMessages.validateTooShort)
         .max(23, sharedMessages.validateTooLong)
         .required(sharedMessages.validateRequired),
       username: Yup.string().when('use_credentials', {
         is: true,
         then: Yup.string()
-          .matches(idRegexp, sharedMessages.validateIdFormat)
+          .matches(noSpacesRegexp, sharedMessages.validateNoSpaces)
```
bafonins: For both `client_id` and `username` there are no such restrictions enforces by the backend. However, according to the spec:

> The ClientId MUST be a UTF-8 encoded string ...

> The Server MUST allow ClientIds which are between 1 and 23 UTF-8 encoded bytes in length, and that contain only the characters
> "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"

Note, no special characters.

cc @adriansmares, @htdvisser  I guess we should make the backend validation more strict to comply with the spec.

htdvisser: Also from the spec:

> The Server MAY allow ClientID’s that contain more than 23 encoded bytes. The Server MAY allow ClientID’s that contain characters not included in the list given above.

PR #2214, opened 2020-03-21
Feature/cli all fields mask

### LINKED ISSUE #2213: CLI flag to return all field masks of API entities
#### Summary

Fetching information for an object (application, gateway, end device, user, etc) can be done with the CLI:

```
$ ttn-lw-cli gtw get GTW_ID
```

Normally, this does not fetch all fields of the object. For example, in order to get the frequency plan for a gateway, one needs to explicitly specify the relevant field mask:

```
$ ttn-lw-cli gtw get GTW_ID --frequency-plan-id
```

This may be misleading, or difficult to use for newcomers to the stack. Also, this gets very long and error prone when one wants to retrieve multiple properties.

It would be good to have a new `--all` flag for list, get and search commands, which would return all of the allowed field mask paths for an object.

#### Why do we need this?

Helpful for debugging, helpful for newcomers, shorter command lines

User has to specify all field masks he wants

User can pass an `--all` flag to easily get all field mask paths:

E.g. `ttn-lw-cli gtw get GTW_ID --all`

#### Environment

3.6.2

Add new `--all` flag. When that is present, select all available field mask paths for each object.

yes

neoaggelos:
#### Summary

Closes #2213

#### Changes

- Adds `--all` flag for all get, list, search CLI commands

#### Notes for Reviewers

- Will ignore forbidden field masks with a warning message.
- Flag is turned off by default, preserving existing behaviour

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - CLI can now dump JSON encoded `grpc_payload` field for unary requests (see `--dump-requests` flag).
 - Template ID column in the webhook table in the Console.
+- Select all field mask paths in CLI get, list and search commands (see `--all` option).
 
 ### Changed
 
--- a/cmd/ttn-lw-cli/commands/applications.go
+++ b/cmd/ttn-lw-cli/commands/applications.go
@@ -30,6 +30,8 @@ import (
 var (
 	selectApplicationFlags = util.FieldMaskFlags(&ttnpb.Application{})
 	setApplicationFlags    = util.FieldFlags(&ttnpb.Application{})
+
+	selectAllApplicationFlags = util.SelectAllFlagSet("application")
 )
 
 func applicationIDFlags() *pflag.FlagSet {
@@ -68,6 +70,7 @@ var (
 		Short:   "List applications",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			paths := util.SelectFieldMask(cmd.Flags(), selectApplicationFlags)
+			paths = ttnpb.AllowedFields(paths, ttnpb.AllowedFieldMaskPathsForRPC["/ttn.lorawan.v3.ApplicationRegistry/List"])
 
 			is, err := api.Dial(ctx, config.IdentityServerGRPCAddress)
 			if err != nil {
@@ -94,6 +97,7 @@ var (
 		Short: "Search for applications",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			paths := util.SelectFieldMask(cmd.Flags(), selectApplicationFlags)
+			paths = ttnpb.AllowedFields(paths, ttnpb.AllowedFieldMaskPathsForRPC["/ttn.lorawan.v3.EntityRegistrySearch/SearchApplications"])
 
 			req, opt, getTotal := getSearchEntitiesRequest(cmd.Flags())
 			req.FieldMask.Paths = paths
@@ -121,6 +125,7 @@ var (
 				return errNoApplicationID
 			}
 			paths := util.SelectFieldMask(cmd.Flags(), selectApplicationFlags)
+			paths = ttnpb.AllowedFields(paths, ttnpb.AllowedFieldMaskPathsForRPC["/ttn.lorawan.v3.ApplicationRegistry/Get"])
 
 			is, err := api.Dial(ctx, config.IdentityServerGRPCAddress)
 			if err != nil {
@@ -252,12 +257,15 @@ func init() {
 	applicationsListCommand.Flags().AddFlagSet(selectApplicationFlags)
 	applicationsListCommand.Flags().AddFlagSet(paginationFlags())
 	applicationsListCommand.Flags().AddFlagSet(orderFlags())
+	applicationsListCommand.Flags().AddFlagSet(selectAllApplicationFlags)
 	applicationsCommand.AddCommand(applicationsListCommand)
 	applicationsSearchCommand.Flags().AddFlagSet(searchFlags())
 	applicationsSearchCommand.Flags().AddFlagSet(selectApplicationFlags)
+	applicationsSearchCommand.Flags().AddFlagSet(selectAllApplicationFlags)
 	applicationsCommand.AddCommand(applicationsSearchCommand)
 	applicationsGetCommand.Flags().AddFlagSet(applicationIDFlags())
 	applicationsGetCommand.Flags().AddFlagSet(selectApplicationFlags)
+	applicationsGetCommand.Flags().AddFlagSet(selectAllApplicationFlags)
 	applicationsCommand.AddCommand(applicationsGetCommand)
 	applicationsCreateCommand.Flags().AddFlagSet(applicationIDFlags())
 	applicationsCreateCommand.Flags().AddFlagSet(collaboratorFlags())
--- a/cmd/ttn-lw-cli/commands/clients.go
+++ b/cmd/ttn-lw-cli/commands/clients.go
@@ -30,6 +30,8 @@ import (
 var (
 	selectClientFlags = util.FieldMaskFlags(&ttnpb.Client{})
 	setClientFlags    = util.FieldFlags(&ttnpb.Client{})
+
+	selectAllClientFlags = util.SelectAllFlagSet("client")
 )
 
 func clientIDFlags() *pflag.FlagSet {
@@ -68,6 +70,7 @@ var (
 		Short:   "List clients",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			paths := util.SelectFieldMask(cmd.Flags(), selectClientFlags)
+			paths = ttnpb.AllowedFields(paths, ttnpb.AllowedFieldMaskPathsForRPC["/ttn.lorawan.v3.ClientRegistry/List"])
 
 			is, err := api.Dial(ctx, config.IdentityServerGRPCAddress)
 			if err != nil {
@@ -94,6 +97,7 @@ var (
 		Short: "Search for clients",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			paths := util.SelectFieldMask(cmd.Flags(), selectClientFlags)
+			paths = ttnpb.AllowedFields(paths, ttnpb.AllowedFieldMaskPathsForRPC["/ttn.lorawan.v3.EntityRegistrySearch/SearchClients"])
 
 			req, opt, getTotal := getSearchEntitiesRequest(cmd.Flags())
 			req.FieldMask.Paths = paths
@@ -121,6 +125,7 @@ var (
 				return errNoClientID
 			}
 			paths := util.SelectFieldMask(cmd.Flags(), selectClientFlags)
+			paths = ttnpb.AllowedFields(paths, ttnpb.AllowedFieldMaskPathsForRPC["/ttn.lorawan.v3.ClientRegistry/Get"])
 
 			is, err := api.Dial(ctx, config.IdentityServerGRPCAddress)
 			if err != nil {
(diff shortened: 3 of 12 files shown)


cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -86,6 +101,22 @@ func UpdateFieldMask(cmdFlags *pflag.FlagSet, fieldMaskFlags ...*pflag.FlagSet)
 	return
 }
 
+func DropForbiddenFieldMaskPaths(paths []string, forbiddenPaths []string) []string {
```
johanstokking: Use `ttnpb.ExcludeFields()`

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -61,15 +61,30 @@ func NormalizeFlags(f *pflag.FlagSet, name string) pflag.NormalizedName {
 }
 
 func SelectFieldMask(cmdFlags *pflag.FlagSet, fieldMaskFlags ...*pflag.FlagSet) (paths []string) {
+	all := false
 	cmdFlags.Visit(func(flag *pflag.Flag) {
 		flagName := toUnderscore.Replace(flag.Name)
+		if flagName == "all" {
```
johanstokking: I guess we can check `--all` like any other flag?

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- CLI get, list and search commands now accept `--all` field mask.
```
johanstokking: Explain the purpose

cmd/ttn-lw-cli/commands/applications.go
```diff
@@ -252,12 +252,15 @@ func init() {
 	applicationsListCommand.Flags().AddFlagSet(selectApplicationFlags)
 	applicationsListCommand.Flags().AddFlagSet(paginationFlags())
 	applicationsListCommand.Flags().AddFlagSet(orderFlags())
+	applicationsListCommand.Flags().Bool("all", false, "Show all application fields")
```
johanstokking: Consider making `all` part of `selectXXXFlags`

cmd/ttn-lw-cli/commands/clients.go
```diff
@@ -30,6 +30,8 @@ import (
 var (
 	selectClientFlags = util.FieldMaskFlags(&ttnpb.Client{})
 	setClientFlags    = util.FieldFlags(&ttnpb.Client{})
+
+	forbiddenSelectClientFlags = []string{"secret"}
```
johanstokking: It's not forbidden; 1) it's just paths that aren't fetched by default when `-all` is passed and 2) forbidden or not would be a server decision.

cmd/ttn-lw-cli/commands/clients.go
```diff
@@ -68,6 +70,10 @@ var (
 		Short:   "List clients",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			paths := util.SelectFieldMask(cmd.Flags(), selectClientFlags)
+			dropped := util.DropForbiddenFieldMaskPaths(paths, forbiddenSelectClientFlags)
+			for _, path := range dropped {
+				logger.Warnf("Ignoring forbidden field mask `%s`", path)
+			}
```
johanstokking: So what happens with `dropped`?

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -61,15 +61,30 @@ func NormalizeFlags(f *pflag.FlagSet, name string) pflag.NormalizedName {
 }
 
 func SelectFieldMask(cmdFlags *pflag.FlagSet, fieldMaskFlags ...*pflag.FlagSet) (paths []string) {
+	all := false
```
htdvisser: Can just `all, _ := cmdFlags.GetBool("all")`

cmd/ttn-lw-cli/commands/clients.go
htdvisser: It may indeed be good to filter the paths using `ttnpb.AllowedFieldMaskPathsForRPC["/ttn.lorawan.v3.ClientRegistry/Get"]`. We could add such functionality to `util.FieldMaskFlags`.

cmd/ttn-lw-cli/internal/util/flags.go
htdvisser: Also:

- Mutating `paths` isn't what you'd expect from the function signature
- This func misses documentation

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -61,15 +61,30 @@ func NormalizeFlags(f *pflag.FlagSet, name string) pflag.NormalizedName {
...
+		}
 		for _, fieldMaskFlags := range fieldMaskFlags {
 			if b, err := fieldMaskFlags.GetBool(flagName); err == nil && b {
 				paths = append(paths, flagName)
 				return
 			}
 		}
 	})
+
+	if all {
```
htdvisser: Why do we need a separate loop here? Can't we just add an "or all" to the `if` in the loop above?

cmd/ttn-lw-cli/commands/clients.go
neoaggelos: The `dropped` paths had already been dropped from the `paths`.

No longer relevant, switched to using `ttnpb.ExcludeFields` instead.

cmd/ttn-lw-cli/commands/applications.go
neoaggelos: `util.SelectFieldMask` uses `selectXXXFlags` to filter arguments. If we make `all` part of it, then we also have to update `util.SelectFieldMask` to discard it in the result value. I think this would cause more harm than good.

cmd/ttn-lw-cli/commands/clients.go
neoaggelos: Maybe this could fit better inside a GRPC unary interceptor? Get the field mask paths of the request, and remove any that do not exist in `ttnpb.AllowedFieldMaskPathsForRPC[method_name]`

cmd/ttn-lw-cli/internal/util/flags.go
neoaggelos: Improved handling of `--all` flag, using the same loop

cmd/ttn-lw-cli/internal/util/flags.go
neoaggelos: Switched to the existing `ttnpb.ExcludeFields()` function

CHANGELOG.md
neoaggelos: Updated the changelog text

cmd/ttn-lw-cli/commands/clients.go
neoaggelos: In any case, the problem we are trying to solve with this is the following:

```
$ ttn-lw-cli clients list --all
    forbidden_paths=[secret]
    correlation_id=85b624ee54fe4b2da6647f65fd4e16ba
exit status 255
```

The `all` flag selects some of the forbidden field mask paths by default. The current solution is to hard code the paths that need to be removed from the field mask paths, and it is similar to what we do for the `AllowedFieldMaskPathsForRPC` as well:

https://github.com/TheThingsNetwork/lorawan-stack/blob/6e25d130324f2486842019a6aaafc8c3e825e626/pkg/ttnpb/field_mask_validation.go#L94

neoaggelos: Another alternative could be to allow an `all` or `*` field mask on the server side, and pass that from the CLI (that way anyone that uses the API can benefit from this)

cmd/ttn-lw-cli/commands/applications.go
johanstokking: In that case, add an `allFlag`. My comment was most to avoid repetition than making it part of `selectXXXFlags()`

cmd/ttn-lw-cli/commands/clients.go
johanstokking: > The current solution is to hard code the paths that need to be removed from the field mask paths, and it is similar to what we do for the `AllowedFieldMaskPathsForRPC` as well

OK I see, but can't you use this function then? Just before you make the call, you strip the paths. A unary client interceptor is not a bad idea, but it also feels like a work-around. I'll leave this up to @htdvisser 

> Another alternative could be to allow an `all` or `*` field mask on the server side, and pass that from the CLI (that way anyone that uses the API can benefit from this)

Nah, this is really client-side. The question is where to do the filtering; in the "application" layer (i.e. commands), or in the service layer (i.e. API client).

neoaggelos: I don't think this should be done in the "Application" layer, mostly because this would require patching every single API call across multiple files.

Using `AllowedFieldMaskPathsForRPC` is not really an option either, because we don't know the method name at that point (and hard-coding it is similarly a bad idea)

neoaggelos: So, if we are to use a GRPC interceptor, we would need to something like the following:

```

// filterFieldMaskPaths is an interceptor that filters field mask paths that are not allowed
func filterFieldMaskPaths(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
	switch req.(type) {
	case *ttnpb.GetGatewayRequest:
		req.(*ttnpb.GetGatewayRequest).FieldMask.Paths = ttnpb.AllowedFieldMaskPathsForRPC[method]
	case *ttnpb.ListGatewaysRequest:
		req.(*ttnpb.ListGatewaysRequest).FieldMask.Paths = ttnpb.AllowedFieldMaskPathsForRPC[method]
        // more here...
	default:
	}

	return invoker(ctx, method, req, reply, cc, opts...)
}
```

htdvisser: I don't think a client-side gRPC interceptor is appropriate here. We're already constructing the field masks in each command, so it's more appropriate to also do the filtering in the commands. The `applications list` command knows that it's calling the `ApplicationRegistry.List` RPC, so it can easily filter the fields using the allowed fields for that RPC. This is very similar to what we already do in `end_devices_split.go`.

---

If we do choose to go with the client interceptor, we should make it generic, instead of type checking the request for known messages.

```go
func Intercept(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
	if v, ok := req.(interface {
		GetFieldMask() pbtypes.FieldMask
	}); ok {
		fm := v.GetFieldMask()
		if allowedPaths, ok := ttnpb.AllowedFieldMaskPathsForRPC[method]; ok {
			fm.Paths = ttnpb.AllowedFields(fm.Paths, allowedPaths)
		}
	}
	return invoker(ctx, method, req, reply, cc)
}
```

neoaggelos: For the record, this is what I initially thought to do as well. But it does not work, because `fm` is a copy of the request FieldMask. I didn't find any way to mutate the field mask of a request without explicitly using its type (hence the example above).

neoaggelos: For the commands that are affected (`clients` and `users`), the flags are now filtered using `ttnpb.AllowedFieldMaskPathsForRPC`

neoaggelos:
Added filters for all cli commands.

johanstokking:
Personally I find this too much service layer specifics in the CLI commands

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -281,6 +284,10 @@ var (
 			if len(jsPaths) > 0 {
 				isPaths = append(isPaths, "join_server_address")
 			}
+			isPaths = ttnpb.AllowedFields(isPaths, ttnpb.AllowedFieldMaskPathsForRPC["/ttn.lorawan.v3.EndDeviceRegistry/Get"])
+			jsPaths = ttnpb.AllowedFields(isPaths, ttnpb.AllowedFieldMaskPathsForRPC["/ttn.lorawan.v3.JsEndDeviceRegistry/Get"])
+			nsPaths = ttnpb.AllowedFields(isPaths, ttnpb.AllowedFieldMaskPathsForRPC["/ttn.lorawan.v3.NsEndDeviceRegistry/Get"])
+			asPaths = ttnpb.AllowedFields(isPaths, ttnpb.AllowedFieldMaskPathsForRPC["/ttn.lorawan.v3.AsEndDeviceRegistry/Get"])
```
htdvisser, line 290: This is already done by `splitEndDeviceGetPaths` above.

cmd/ttn-lw-cli/commands/clients.go
```diff
@@ -68,7 +70,7 @@ var (
 		Short:   "List clients",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			paths := util.SelectFieldMask(cmd.Flags(), selectClientFlags)
-
+			paths = ttnpb.AllowedFields(paths, ttnpb.AllowedFieldMaskPathsForRPC["/ttn.lorawan.v3.ClientRegistry/List"])
```
htdvisser, line 73: In all other places we have an empty line below.

PR #2219, opened 2020-03-23
Add defer/recover to UDP and MQTT frontends

KrishnaIyer:
#### Summary

Add defer/recover to UDP and MQTT frontends

#### Changes

- Add recovery to all possible paths

#### Notes for Reviewers

...

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/mqtt/mqtt.go
+++ b/pkg/gatewayserver/io/mqtt/mqtt.go
@@ -46,8 +46,7 @@ type srv struct {
 	lis    mqttnet.Listener
 }
 
-// ErrMQTTFrontendRecovered is returned when a panic is caught from the MQTT frontend.
-var ErrMQTTFrontendRecovered = errors.DefineInternal("mqtt_frontend_recovered", "Internal Server Error")
+var errMQTTFrontendRecovered = errors.DefineInternal("mqtt_frontend_recovered", "internal server error")
 
 // Serve serves the MQTT frontend.
 func Serve(ctx context.Context, server io.Server, listener net.Listener, format Format, protocol string) error {
@@ -325,10 +324,10 @@ func recoverMQTTFrontend(ctx context.Context) {
 		os.Stderr.Write(debug.Stack())
 		var err error
 		if pErr, ok := p.(error); ok {
-			err = ErrMQTTFrontendRecovered.WithCause(pErr)
+			err = errMQTTFrontendRecovered.WithCause(pErr)
 		} else {
-			err = ErrMQTTFrontendRecovered.WithAttributes("panic", p)
+			err = errMQTTFrontendRecovered.WithAttributes("panic", p)
 		}
-		log.FromContext(ctx).WithError(err).Error("MQTT Frontend failed")
+		log.FromContext(ctx).WithError(err).Error("MQTT frontend failed")
 	}
 }
--- a/pkg/gatewayserver/io/mqtt/mqtt.go
+++ b/pkg/gatewayserver/io/mqtt/mqtt.go
@@ -95,9 +95,14 @@ type connection struct {
 func (*connection) Protocol() string            { return "mqtt" }
 func (*connection) SupportsDownlinkClaim() bool { return false }
 
-func (c *connection) setup(ctx context.Context) error {
+func (c *connection) setup(ctx context.Context) (err error) {
 	ctx = auth.NewContextWithInterface(ctx, c)
-	defer recoverMQTTFrontend(ctx)
+	defer func() {
+		retrievedErr := recoverMQTTFrontend(ctx)
+		if retrievedErr != nil {
+			err = retrievedErr
+		}
+	}()
 	c.session = session.New(ctx, c.mqtt, c.deliver)
 	if err := c.session.ReadConnect(); err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


pkg/gatewayserver/io/mqtt/mqtt.go
```diff
@@ -44,6 +46,9 @@ type srv struct {
 	lis    mqttnet.Listener
 }
 
+// ErrMQTTFrontendRecovered is returned when a panic is caught from the MQTT frontend.
+var ErrMQTTFrontendRecovered = errors.DefineInternal("mqtt_frontend_recovered", "Internal Server Error")
```
johanstokking: - we don't expose errors
- is the error that it's recovered, or that there's an internal server error?
- no title case in error messages

pkg/gatewayserver/io/mqtt/mqtt.go
```diff
@@ -273,6 +282,7 @@ func (c *connection) CanWrite(info *auth.Info, topicParts ...string) bool {
 
 func (c *connection) deliver(pkt *packet.PublishPacket) {
 	logger := log.FromContext(c.io.Context()).WithField("topic", pkt.TopicName)
+	defer recoverMQTTFrontend(c.io.Context())
```
johanstokking: We're already in a goroutine that's being recovered here I guess

pkg/gatewayserver/io/mqtt/mqtt.go
```diff
@@ -308,3 +318,17 @@ func (c *connection) deliver(pkt *packet.PublishPacket) {
...
+	if p := recover(); p != nil {
+		fmt.Fprintln(os.Stderr, p)
+		os.Stderr.Write(debug.Stack())
+		var err error
+		if pErr, ok := p.(error); ok {
+			err = ErrMQTTFrontendRecovered.WithCause(pErr)
+		} else {
+			err = ErrMQTTFrontendRecovered.WithAttributes("panic", p)
+		}
+		log.FromContext(ctx).WithError(err).Error("MQTT Frontend failed")
```
johanstokking: `frontend`

pkg/gatewayserver/io/mqtt/mqtt.go
KrishnaIyer: basically adapted from our [echo middleware](https://github.com/TheThingsNetwork/lorawan-stack/blob/c83dcc8f6aa77a886768723a4963169f509ce46e/pkg/web/middleware/recover.go#L30).

KrishnaIyer: > we don't expose errors

I'm not actually returning this anywhere right? It's only printing to `stderr` and logs it.

pkg/gatewayserver/io/mqtt/mqtt.go
KrishnaIyer: This function seems to be a callback to `mystique` and so I added the recovery here as well.

pkg/gatewayserver/io/mqtt/mqtt.go
johanstokking: OK for consistency, although I don't really like it there

It's package exposed, i.e. `Err...` instead of `err...`

pkg/gatewayserver/io/mqtt/mqtt.go
johanstokking: Yes, but is it a callback on our goroutine or does Mystique spin up a goroutine?

KrishnaIyer: No clue @htdvisser ?

johanstokking: See https://github.com/TheThingsIndustries/mystique

htdvisser: Mystique doesn't start a lot of goroutines. As far as I remember, just two per connection (read+write). This would mean that the delivery would be done by the goroutine that calls `session.ReadPacket()`.

pkg/gatewayserver/io/mqtt/mqtt.go
```diff
@@ -93,6 +97,7 @@ func (*connection) SupportsDownlinkClaim() bool { return false }
 
 func (c *connection) setup(ctx context.Context) error {
 	ctx = auth.NewContextWithInterface(ctx, c)
+	defer recoverMQTTFrontend(ctx)
```
htdvisser: Shouldn't this propagate the panic to the returned error?

pkg/gatewayserver/io/udp/udp.go
```diff
@@ -122,6 +127,7 @@ func Serve(ctx context.Context, server io.Server, conn *net.UDPConn, config Conf
 }
 
 func (s *srv) read() error {
+	defer recoverUDPFrontend(s.ctx)
```
htdvisser: Shouldn't this propagate the panic to the returned error?

pkg/gatewayserver/io/mqtt/mqtt.go
```diff
@@ -95,9 +95,11 @@ type connection struct {
 func (*connection) Protocol() string            { return "mqtt" }
 func (*connection) SupportsDownlinkClaim() bool { return false }
 
-func (c *connection) setup(ctx context.Context) error {
+func (c *connection) setup(ctx context.Context) (err error) {
 	ctx = auth.NewContextWithInterface(ctx, c)
-	defer recoverMQTTFrontend(ctx)
+	defer func() {
+		err = recoverMQTTFrontend(ctx)
```
johanstokking: Aren't we now setting `err` to `nil` when there's no panic?

KrishnaIyer: Ah s*** yeah good catch.

pkg/gatewayserver/io/mqtt/mqtt.go
```diff
@@ -308,3 +321,19 @@ func (c *connection) deliver(pkt *packet.PublishPacket) {
 		logger.Debug("Publish to invalid topic")
 	}
 }
+
+func recoverMQTTFrontend(ctx context.Context) (err error) {
```
htdvisser:
```suggestion
func recoverMQTTFrontend(ctx context.Context) error {
```

PR #2225, opened 2020-03-24
Packet Broker Agent

johanstokking:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/1919

#### Changes

- Open source the current proprietary Packet Broker Agent
- Add downlink support
- Add test coverage and small fixes

#### Notes for Reviewers

Review from `pba: Upgrade to V2 API`

Still needs:

- [x] Add configuration documentation
- [x] End-to-end testing locally
- [x] Import tagged `go.packetbroker.org/api/v3`

#### Checklist


pkg/networkserver/downlink.go
```diff
@@ -696,6 +703,58 @@ func nonRetryableFixedPathGatewayError(err error) bool {
...
+type packetBrokerDownlinkTarget struct {
+	peer cluster.Peer
+}
+
+func (t *packetBrokerDownlinkTarget) Equal(target downlinkTarget) bool {
+	_, ok := target.(*packetBrokerDownlinkTarget)
+	return ok
+}
+
+const packetBrokerScheduleDelay = 5 * time.Second
```
rvolosatovs: shouldn't this be based of `infrastructureDelay`?

cmd/ttn-lw-stack/commands/config.go
```diff
@@ -60,6 +64,8 @@ var DefaultConfig = Config{
 	JS:          shared_joinserver.DefaultJoinServerConfig,
 	Console:     shared_console.DefaultConsoleConfig,
 	GCS:         shared_gatewayconfigurationserver.DefaultGatewayConfigurationServerConfig,
+	DTC:         shared_devicetemplateconverter.DefaultDeviceTemplateConverterConfig,
```
rvolosatovs: Let's maintain the same order here as in `Config` above.
Is `QRG` missing as well?

pkg/networkserver/downlink.go
```diff
@@ -587,29 +588,35 @@ func downlinkPathsFromMetadata(mds ...*ttnpb.RxMetadata) []downlinkPath {
...
+				body = append(body, path)
+			}
+		case *ttnpb.RxMetadata_PacketBroker:
 			tail = append(tail, path)
 		}
 	}
-	return append(head, tail...)
+	res := append(head, body...)
+	res = append(res, tail...)
+	return res
```
rvolosatovs: Since we're already allocating capacities for `head`, `body` and `tail`, seems a waste not to do this:
```suggestion
	res := make([]downlinkPath, 0, len(head) + len(body) + len(tail))
	res = append(res, head...)
	res = append(res, body...)
	res = append(res, tail...)
	return res
```
clearer as well IMO

pkg/networkserver/downlink.go
```diff
@@ -707,31 +766,43 @@ func (ns *NetworkServer) scheduleDownlinkByPaths(ctx context.Context, req *ttnpb
...
+			}
+			target = &gatewayServerDownlinkTarget{peer}
+		} else {
+			logger := logger.WithField("target", "packet_broker_agent")
+			peer, err := ns.GetPeer(ctx, ttnpb.ClusterRole_PACKET_BROKER_AGENT, nil)
+			if err != nil {
+				logger.WithError(err).Warn("Failed to get Packet Broker Agent peer")
+				continue
+			}
+			target = &packetBrokerDownlinkTarget{peer}
```
rvolosatovs:
```suggestion
			target = &packetBrokerDownlinkTarget{peer: peer}
```

pkg/networkserver/downlink.go
```diff
@@ -707,31 +766,43 @@ func (ns *NetworkServer) scheduleDownlinkByPaths(ctx context.Context, req *ttnpb
...
+			logger := logger.WithFields(log.Fields(
+				"target", "gateway_server",
+				"gateway_uid", unique.ID(ctx, path.GatewayIdentifiers),
+			))
+			peer, err := ns.GetPeer(ctx, ttnpb.ClusterRole_GATEWAY_SERVER, *path.GatewayIdentifiers)
+			if err != nil {
+				logger.WithError(err).Warn("Failed to get Gateway Server peer")
+				continue
+			}
+			target = &gatewayServerDownlinkTarget{peer}
```
rvolosatovs:
```suggestion
			target = &gatewayServerDownlinkTarget{peer: peer}
```

pkg/networkserver/mac_link_check.go
```diff
@@ -46,7 +47,12 @@ func handleLinkCheckReq(ctx context.Context, dev *ttnpb.EndDevice, msg *ttnpb.Up
 	gtws := make(map[string]struct{}, len(msg.RxMetadata))
 	maxSNR := msg.RxMetadata[0].SNR
 	for _, md := range msg.RxMetadata {
-		gtws[unique.ID(ctx, md.GatewayIdentifiers)] = struct{}{}
+		switch s := md.GetSource().(type) {
+		case *ttnpb.RxMetadata_GatewayIDs:
+			gtws[unique.ID(ctx, s.GatewayIDs)] = struct{}{}
+		case *ttnpb.RxMetadata_PacketBroker:
+			gtws[fmt.Sprintf("%s@%s/%s", s.PacketBroker.ForwarderID, s.PacketBroker.ForwarderNetID, s.PacketBroker.ForwarderTenantID)] = struct{}{}
```
rvolosatovs: Is this the only occurence when this is helpful? Perhaps we want to extract this into a separate function?

pkg/packetbrokeragent/agent.go
```diff
@@ -0,0 +1,741 @@
...
+		if len(devAddrPrefixes) == 0 {
+			devAddr, err := types.NewDevAddr(conf.NetID, nil)
+			if err != nil {
+				return nil, errNetID.WithAttributes("net_id", conf.NetID).WithCause(err)
+			}
+			devAddrPrefix := types.DevAddrPrefix{
+				DevAddr: devAddr,
+				Length:  uint8(conf.NetID.IDBits()),
+			}
+			devAddrPrefixes = append(devAddrPrefixes, devAddrPrefix)
```
rvolosatovs: nit; can just inline `devAddrPrefix`

pkg/packetbrokeragent/agent.go
```diff
@@ -0,0 +1,741 @@
...
+			enc jose.ContentEncryption
+			alg jose.KeyAlgorithm
+		)
+		switch len(a.forwarderConfig.TokenKey) {
+		case 16:
+			enc, alg = jose.A128GCM, jose.A128GCMKW
+		case 32:
+			enc, alg = jose.A256GCM, jose.A256GCMKW
+		default:
+			return nil, errTokenKey.New()
```
rvolosatovs: Include the length here for debugging?

pkg/networkserver/downlink.go
johanstokking: Yeah good one. There would be a multiplier but makes sense to start with `infrastructureDelay`.

pkg/networkserver/downlink.go
johanstokking: So that does require an additional allocation, while everything is now fit in the `head` memory space

pkg/networkserver/mac_link_check.go
johanstokking: It is the only occurrence I think.

There was something similar in observability, but that wasn't consistent with the logging of duplication count, so I made that consistent by taking the `RxMetadata` length. I think that looking at that length makes sense for deduplication, and looking really at the unique gateways for coverage is here only.

CHANGELOG.md
```diff
@@ -30,17 +32,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-
-### Fixed
-
-- Fix organization collaborator view not being accessible in the Console.
-- Error display on Data pages in the Console.
-- Fix too restrictive MQTT client validation in PubSub form in the Console.
-- Fix faulty display of device event stream data for end devices with the same ID in different applications.
-- Trailing slashes handling in webhook paths.
-- Limited throughput in upstream handlers in Gateway Server when one gateway's upstream handler is busy.
-
```
htdvisser: This doesn't seem right

api/packetbrokeragent.proto
```diff
@@ -25,3 +25,12 @@ option go_package = "go.thethings.network/lorawan-stack/pkg/ttnpb";
 service GsPba {
   rpc PublishUplink(GatewayUplinkMessage) returns (google.protobuf.Empty);
 }
+
+// The NsPba service connects a Network Server to a Packet Broker Agent.
+service NsPba {
+  // PublishDownlink instructs the Packet Broker Agent to publish a downlink message to Packet Broker Router.
+  // The Packet Broker Agent returns on the first successful message state change from Packet Broker Router;
+  // it does not wait for Packet Broker Router to check the routing policy, nor does it wait for the Home Network to
+  // confirm downlink message reception, downlink path availability or confirmation from the gateway.
```
htdvisser: I still think this can return as soon as the PBA "accepts" or "takes ownership of" the message.

pkg/networkserver/observability.go
```diff
@@ -239,11 +238,7 @@ func registerReceiveUplink(ctx context.Context, msg *ttnpb.UplinkMessage) {
 }
 
 func registerMergeMetadata(ctx context.Context, msg *ttnpb.UplinkMessage) {
-	gtws := make(map[string]struct{}, len(msg.RxMetadata))
-	for _, md := range msg.RxMetadata {
-		gtws[unique.ID(ctx, md.GatewayIdentifiers)] = struct{}{}
-	}
-	nsMetrics.uplinkGateways.WithLabelValues(ctx).Observe(float64(len(gtws)))
+	nsMetrics.uplinkGateways.WithLabelValues(ctx).Observe(float64(len(msg.RxMetadata)))
```
htdvisser: With this change, the metric can no longer be called `uplinkGateways`. At least for non-packetbroker metadatas it now indicates the number of antennas, not the number of gateways that received the uplink.

pkg/gatewayserver/upstream/packetbroker/packetbroker.go
```diff
@@ -0,0 +1,85 @@
...
+
+// HandleUplink implements upstream.Handler.
+func (h *Handler) HandleUplink(ctx context.Context, _ ttnpb.GatewayIdentifiers, ids ttnpb.EndDeviceIdentifiers, msg *ttnpb.GatewayUplinkMessage) error {
+	pbaConn, err := h.cluster.GetPeerConn(ctx, ttnpb.ClusterRole_PACKET_BROKER_AGENT, ids)
+	if err != nil {
+		return errPacketBrokerAgentNotFound.WithCause(err)
+	}
+	ctx, cancel := context.WithTimeout(ctx, publishUplinkTimeout)
+	_, err = ttnpb.NewGsPbaClient(pbaConn).PublishUplink(ctx, msg, h.cluster.WithClusterAuth())
+	cancel()
```
htdvisser: Just use `defer cancel()`. Since Go 1.14 there is no reason to avoid `defer` anymore, and it's better to consistently cancel contexts in a `defer`.

cmd/ttn-lw-stack/commands/config.go
```diff
@@ -48,6 +51,7 @@ type Config struct {
 	GCS              gatewayconfigurationserver.Config `name:"gcs"`
 	DTC              devicetemplateconverter.Config    `name:"dtc"`
 	QRG              qrcodegenerator.Config            `name:"qrg"`
+	PBA              packetbrokeragent.Config          `name:"pba"`
```
htdvisser: Can we also add a page to the configuration reference in the documentation?

api/metadata.proto
```diff
@@ -31,7 +31,13 @@ option go_package = "go.thethings.network/lorawan-stack/pkg/ttnpb";
 message RxMetadata {
   option (gogoproto.populate) = false;
 
-  GatewayIdentifiers gateway_ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (validate.rules).message.required = true];
+  oneof source {
+    option (validate.required) = true;
+
+    GatewayIdentifiers gateway_ids = 1 [(gogoproto.customname) = "GatewayIDs"];
+    PacketBrokerMetadata packet_broker = 18;
+  }
```
htdvisser: This violates our compatibility commitment. `RxMetadata` is included in `ApplicationUplink`, so it's external API. Previously the `gateway_ids` was indicated as non-nullable and required, now we may omit it, which may break applications that enforce the old validation rule (for packetbroker uplinks).

To be fully compliant, we can simply add the `PacketBrokerMetadata`, but also still include the `GatewayIdentifiers` as we do now (with the `packetbroker` ID, as we already did in The Things Enterprise Stack).

cmd/internal/shared/packetbrokeragent/config.go
```diff
@@ -0,0 +1,40 @@
...
+	"go.thethings.network/lorawan-stack/pkg/packetbrokeragent"
+)
+
+// DefaultPacketBrokerAgentConfig is the default configuration for the Packet Broker Agent.
+var DefaultPacketBrokerAgentConfig = packetbrokeragent.Config{
+	HomeNetwork: packetbrokeragent.HomeNetworkConfig{
+		WorkerPool: packetbrokeragent.WorkerPoolConfig{
+			MaximumWorkerCount: (1 << 12),
+			IdleTimeout:        (1 << 7) * time.Millisecond,
+			BusyTimeout:        (1 << 6) * time.Millisecond,
```
htdvisser: Is there a reason for writing it like this instead of just using human-friendly decimal numbers?
IMO `64` is much clearer than `(1 << 6)`.

pkg/gatewayserver/upstream/packetbroker/packetbroker.go
johanstokking: OK for consistence

cmd/ttn-lw-stack/commands/config.go
johanstokking: Yes, see original comment, that will be part of this PR

api/metadata.proto
johanstokking: Ah yeah the JSONPB ...

cmd/internal/shared/packetbrokeragent/config.go
johanstokking: No particular reason. There's actually little value in exposing this as config. Will remove timing settings

pkg/networkserver/mac_link_check.go
johanstokking: FTR; I extracted this logic because we also use it in metrics.

pkg/packetbrokeragent/agent.go
```diff
@@ -0,0 +1,723 @@
...
+	}
+
+	ids, msg, err := fromPBDownlink(ctx, down.Message, receivedAt, a.forwarderConfig)
+	if err != nil {
+		logger.WithError(err).Warn("Failed to convert incoming uplink message")
+		return err
+	}
+
+	req := msg.GetRequest()
+	logger = logger.WithFields(log.Fields(
```
rvolosatovs, line 436: Let's reuse https://github.com/TheThingsNetwork/lorawan-stack/blob/8debfce391dc65c0e65fe649220389865daefd21/pkg/networkserver/downlink.go#L793-L819 here.
Maybe extract it into an `internal` package?

pkg/cluster/cluster.go
```diff
@@ -294,9 +294,22 @@ func (c *cluster) GetPeers(ctx context.Context, role ttnpb.ClusterRole) ([]Peer,
...
+		}
+	default:
+	}
+	return role
+}
+
 var errPeerUnavailable = errors.DefineUnavailable("peer_unavailable", "{cluster_role} cluster peer unavailable")
 
 func (c *cluster) GetPeer(ctx context.Context, role ttnpb.ClusterRole, ids ttnpb.Identifiers) (Peer, error) {
+	role = overridePeerRole(ctx, role, ids)
```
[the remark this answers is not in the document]
johanstokking: Yes this is still necessary as fallback

doc/content/guides/enable-peering/_index.md
```diff
@@ -0,0 +1,89 @@
...
+```yaml
+# Edit the Gateway Server configuration in your configuration file:
+
+# Gateway Server configuration
+gs:
+  forward:
+  # Forward all traffic to the Network Server in the cluster
+  - 'cluster=00000000/0'
+  # Forward all traffic also to Packet Broker
+  - 'packetbroker=00000000/0'
```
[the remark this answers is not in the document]
johanstokking: > I can imagine that one of the first questions would be how to only offload traffic that is not intended for the cluster (so traffic that is _not_ `26000000/7`).

We don't have support for negating prefixes in the GS forward table. I can imagine we will support this indeed, but let's wait for requests for that.

pkg/packetbrokeragent/agent.go
johanstokking: Imo not worth it

pkg/networkserver/downlink.go
```diff
@@ -767,25 +836,22 @@ func (ns *NetworkServer) scheduleDownlinkByPaths(ctx context.Context, req *sched
...
+		logger.WithFields(log.Fields(
+			"transmission_delay", delay,
+			"transmit_at", transmitAt,
+		)).Debug("Scheduled downlink")
 		return &scheduledDownlink{
 			Message:    down,
-			TransmitAt: timeNow().Add(res.Delay),
-		}, append(queuedEvents, req.SuccessEvent(ctx, req.EndDeviceIdentifiers, res)), nil
+			TransmitAt: transmitAt,
+		}, append(queuedEvents, req.SuccessEvent(ctx, req.EndDeviceIdentifiers, nil)), nil
```
rvolosatovs, line 854: Successful GS scheduling still should trigger a `SuccessEvent` containing the response.

pkg/packetbrokeragent/agent.go
rvolosatovs: Can you at least make it consistent then?

johanstokking: https://github.com/TheThingsNetwork/lorawan-stack/pull/2225/commits/bd0d78f02c7ba01bf395bda84eeba30ca7d91a14

pkg/networkserver/downlink.go
johanstokking: https://github.com/TheThingsNetwork/lorawan-stack/pull/2225/commits/9e1b191e549295ed2c17450f142d64baa3f7afcf

PR #2239, opened 2020-03-25
Unblock from busy upstream handler

johanstokking:
#### Summary

Unblock from busy upstream handler

#### Changes

- Drop message immediately when the upstream handlers are all busy
- Fix payload detection of `TX_ACK`

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -487,8 +487,6 @@ var (
 	maxUpstreamHandlers = int32(1 << 5)
 	// upstreamHandlerIdleTimeout is the duration after which an idle upstream handler stops to save resources.
 	upstreamHandlerIdleTimeout = (1 << 7) * time.Millisecond
-	// upstreamHandlerBusyTimeout is the duration after traffic gets dropped if all upstream handlers are busy.
-	upstreamHandlerBusyTimeout = (1 << 6) * time.Millisecond
 )
 
 type upstreamHost struct {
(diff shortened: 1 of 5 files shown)


pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -995,10 +995,27 @@ func TestGatewayServer(t *testing.T) {
...
+							for _, t := range tc.Forwards {
+								notSeen[t] = struct{}{}
+							}
+							for len(notSeen) > 0 {
 								select {
 								case msg := <-ns.Up():
-									expected := tc.Up.UplinkMessages[msgIdx]
+									var expected *ttnpb.UplinkMessage
+									for _, up := range tc.Up.UplinkMessages {
+										if ts := up.Settings.Timestamp; ts == msg.Settings.Timestamp {
```
johanstokking: Messages can come in out-of-order now

pkg/gatewayserver/gatewayserver_internal_test.go
```diff
@@ -14,10 +14,4 @@
 
 package gatewayserver
 
-var (
-	ErrSchedule = errSchedule
-)
-
-func init() {
-	maxUpstreamHandlers = 1
```
johanstokking: Because we use the same pool behavior as in production

pkg/gatewayserver/gatewayserver.go
```diff
@@ -640,17 +636,17 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
 					atomic.AddInt32(&host.handlers, 1)
 					host.handleWg.Add(1)
 					go handleFn(host)
+					go func(host *upstreamHost) {
+						host.handleCh <- item
```
johanstokking: `host` is a loop iterator so need local copy

johanstokking:
> Should the PBA worker pools have the same behavior ? I can do the fix there as well.

No because there we don't have multiple destinations there. The problem was that we're looping over the destinations (`hosts`) and we had a wait on the channel send. If one host was busy, it would limit the overall throughput as the loop per message would go slower.

PR #2240, opened 2020-03-25
Update gateway location handling fixes

### LINKED ISSUE #2008: Allow gateway location to be updated from status messages once per connection
#### Summary

With #1999 all gateways with `update_location_from_status` have their location updated (debounced) everytime a new status message arrives  This creates unnecessary load on the stack

#### Why do we need this?

Even with a 1 hour debounce time, with 10K+ gateways this is ~ 3 updates on the database per second. Most of those updates are also useless, since the majority of the gateways does not move.

All gateways are updated even when their location does not change, spawning of many go-routines that do not do anything useful.

Ability to opt out of unnecessary location updates per gateway
#### Environment

3.5.3 dev

An option (configurable per gateway) to only update the gateway location once per connection. Afterwards, we can cleanup all relevant goroutines. This would greatly lower the amount of updates required on the database.

Yes

neoaggelos:
#### Summary

Closes #2008 

#### Changes

- Set gateway location even if there were no locations previously set
- Do not perform any updates if location has not changed

#### Notes for Reviewers

...

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -747,6 +747,9 @@ func (gs *GatewayServer) handleLocationUpdates(conn connectionEntry) {
 					// Add an antenna if none are present
 					conn.Gateway().Antennas = []ttnpb.GatewayAntenna{{}}
 				}
+				if sameLocation(conn.Gateway().Antennas[0].Location, *status.AntennaLocations[0]) {
+					break
+				}
 				status.AntennaLocations[0].Source = ttnpb.SOURCE_GPS
 				conn.Gateway().Antennas[0].Location = *status.AntennaLocations[0]
 				_, err := registry.Update(ctx, &ttnpb.UpdateGatewayRequest{
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -18,6 +18,7 @@ package gatewayserver
 import (
 	"context"
 	"fmt"
+	"math"
 	"net"
 	"net/http"
 	"os"
@@ -708,6 +709,16 @@ func (gs *GatewayServer) updateConnStats(conn connectionEntry) {
 	}
 }
 
+const (
+	allowedLocationDelta = 0.00001
+)
+
+func sameLocation(a, b ttnpb.Location) bool {
+	return a.Altitude == b.Altitude && a.Accuracy == b.Accuracy &&
+		math.Abs(a.Latitude-b.Latitude) <= allowedLocationDelta &&
+		math.Abs(a.Longitude-b.Longitude) <= allowedLocationDelta
+}
+
 func (gs *GatewayServer) handleLocationUpdates(conn connectionEntry) {
 	ctx := conn.Context()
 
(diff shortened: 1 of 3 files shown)


pkg/gatewayserver/gatewayserver.go
```diff
@@ -717,11 +717,22 @@ func (gs *GatewayServer) handleLocationUpdates(conn connectionEntry) {
 			status, _, ok := conn.StatusStats()
 			locations := status.AntennaLocations
 			antennas := conn.Gateway().Antennas
-			if ok && len(locations) > 0 && len(antennas) > 0 {
-				// TODO: Handle multiple antenna locations (https://github.com/TheThingsNetwork/lorawan-stack/issues/2006).
-				locations[0].Source = ttnpb.SOURCE_GPS
-				antennas[0].Location = *locations[0]
+			if len(conn.Gateway().Antennas) == 0 {
```
johanstokking: nit; `len(antennas)`

pkg/gatewayserver/gatewayserver.go
```diff
@@ -717,11 +717,22 @@ func (gs *GatewayServer) handleLocationUpdates(conn connectionEntry) {
...
+				conn.Gateway().Antennas = []ttnpb.GatewayAntenna{
+					{
+						Location: ttnpb.Location{},
```
johanstokking: Are we now always setting an empty location?

neoaggelos: Changed to only adding an empty antenna if the status message has a location set.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -713,11 +713,19 @@ func (gs *GatewayServer) handleLocationUpdates(conn connectionEntry) {
...
+				if len(antennas) == 0 {
+					// Add an antenna if none are present
+					conn.Gateway().Antennas = []ttnpb.GatewayAntenna{{}}
+					antennas = conn.Gateway().Antennas
+				}
+				locations[0].Source = ttnpb.SOURCE_REGISTRY
+				if antennas[0].Location.Equal(*locations[0]) {
+					break
+				}
+				antennas[0].Location = *locations[0]
```
johanstokking: So now the location is set to registry, but these are GPS values, right?

neoaggelos: This is actually because the location from the registry never stores location source. https://github.com/TheThingsNetwork/lorawan-stack/blob/6a873f552411bb41ecffb91e46041dbc6982b46c/pkg/identityserver/store/types.go#L124-L131

So using SOURCE_GPS as a source not only has no effect, it will also cause the equality check below to fail because of the difference in location sources.

Fixing that requires a database migration (adding a `Source` to `GatewayAntenna` ORM model. Creating a pull request for that.

neoaggelos: Reverted change to ttnpb.SOURCE_GPS.

johanstokking: Why do we depend on what IS stores exactly? Isn't this just about checking if the location doesn't change at the GS level, before it updates it in IS?

neoaggelos: The change is checked at the GS level. But location equality check also includes the location source, and that is always hard-coded to `SOURCE_REGISTRY` for gateway locations from the IS at the moment. This only really effects the first status message received (because the initial location on GS is the one retrieved from IS). Discussion is in #2245, not sure if we really care about it.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -709,6 +709,11 @@ func (gs *GatewayServer) updateConnStats(conn connectionEntry) {
 	}
 }
 
+func sameLocation(A, B ttnpb.Location) bool {
```
johanstokking: lowerCamelCase for arguments (or vars for that matter)

pkg/gatewayserver/gatewayserver.go
```diff
@@ -709,6 +709,11 @@ func (gs *GatewayServer) updateConnStats(conn connectionEntry) {
...
+	return A.Latitude == B.Latitude && A.Longitude == B.Longitude &&
```
johanstokking: These are floating points; we probably need a very small delta that is allowed

pkg/gatewayserver/gatewayserver.go
```diff
@@ -733,11 +738,19 @@ func (gs *GatewayServer) handleLocationUpdates(conn connectionEntry) {
 			status, _, ok := conn.StatusStats()
 			locations := status.AntennaLocations
 			antennas := conn.Gateway().Antennas
```
johanstokking: I don't think we really need `locations` and `antennas` anymore

neoaggelos: Removed them, using `status.AntennaLocations` and `conn.Gateway().Antennas` directly.

PR #2245, opened 2020-03-27
is: Include antenna location source in store

neoaggelos:
#### Summary

Identity server now stores antenna location as well.

#### Changes

- Add `Source` field in `GatewayAntenna`.

#### Notes for Reviewers

- Field was not added in `Location` type, because it breaks the `EndDevice` struct (which has a `Location` and a separate `Source` field).

- Testing:
  - Unit tests are passing
  - Tested updating gateway location. Without change, whatever is used as source is ignored. With change, the correct location source is returned.

#### Checklist

### DIFF
--- a/pkg/identityserver/store/gateway_antenna.go
+++ b/pkg/identityserver/store/gateway_antenna.go
@@ -29,6 +29,8 @@ type GatewayAntenna struct {
 	Gain float32
 
 	Location
+
+	Source int32 `gorm:"default:3 not null"`
 }
 
 func init() {
@@ -36,6 +38,10 @@ func init() {
 }
 
 func (a GatewayAntenna) toPB() ttnpb.GatewayAntenna {
+	source := ttnpb.LocationSource(a.Source)
+	if source == ttnpb.SOURCE_UNKNOWN {
+		source = ttnpb.SOURCE_REGISTRY
+	}
 	return ttnpb.GatewayAntenna{
 		Gain: a.Gain,
 		Location: ttnpb.Location{
@@ -43,7 +49,7 @@ func (a GatewayAntenna) toPB() ttnpb.GatewayAntenna {
 			Longitude: a.Longitude,
 			Altitude:  a.Altitude,
 			Accuracy:  a.Accuracy,
-			Source:    ttnpb.SOURCE_REGISTRY,
+			Source:    source,
 		},
 		Attributes: attributes(a.Attributes).toMap(),
 	}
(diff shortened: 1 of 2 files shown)


pkg/identityserver/store/gateway_antenna.go
```diff
@@ -29,21 +29,27 @@ type GatewayAntenna struct {
 	Gain float32
 
 	Location
+
+	Source int32 `gorm:"default:3 not null"`
```
neoaggelos: Using default value 3, which is `SOURCE_REGISTRY`, to preserve info for existing gateways.

The `Source` field itself probably belongs inside the `Location` type, but that breaks `EndDeviceLocation`:

https://github.com/TheThingsNetwork/lorawan-stack/blob/f62b09c8c06e46f8b26d9e9f757212a0f8f7cc9c/pkg/identityserver/store/end_device_location.go#L20-L29

pkg/identityserver/store/gateway_antenna.go
```diff
@@ -29,21 +29,27 @@ type GatewayAntenna struct {
...
 }
 
 func init() {
 	registerModel(&GatewayAntenna{})
 }
 
 func (a GatewayAntenna) toPB() ttnpb.GatewayAntenna {
+	source := ttnpb.LocationSource(a.Source)
+	if source == ttnpb.SOURCE_UNKNOWN {
```
neoaggelos: This makes sense to me, but I'm not sure if it is something we would want.

neoaggelos:
Blocks #2240

neoaggelos:
Since a migration is required for this, it would be good to have it before 3.7.0 is released.

neoaggelos:
> I don't think we should do this. The location stored in the gateway registry has source "REGISTRY" by definition. The feature described in #2008 simply updates the location in the registry.

Okay, I can agree to that. But why is that different than the end devices? End devices do store location sources in the registry.

In general, I think the question is: if we store a location in the registry, then should its source become `SOURCE_REGISTRY` regardless of the original source?

johanstokking:
> Okay, I can agree to that. But why is that different than the end devices? End devices do store location sources in the registry.

Because there are multiple ways for devices to locate themselves. In fact, we might want to support multiple location sources of devices side-by-side to compare localization performance results. This is not applicable to gateways.

> In general, I think the question is: if we store a location in the registry, then should its source become `SOURCE_REGISTRY` regardless of the original source?

For gateways, yes, always.

neoaggelos:
Okay, then. Closing this PR and adjusting #2240 accordingly

PR #2246, opened 2020-03-27
Add `--dump-requests` flag to the CLI

### LINKED ISSUE #816: Add flag to CLI to dump requests
#### Summary

It would be nice to have a flag in the CLI that dumps request payloads for RPCs.

#### Why do we need this?

This could help with debugging potential CLI problems, or could help developers that want to call the API endpoints directly from their own scripts

The typical behavior is for the CLI to make a request and print the response JSON to stdout.

A flag to print the request JSON to stderr.

#### Environment

ttn-lw-cli <= v3.0.3

This could perhaps be implemented as a gRPC interceptor. If a `--dump-requests` flag is provided, the interceptor would print which RPC is being called, and the JSON-encoded request payload.

Yes. In any case I will look into the possibility of doing this in an interceptor. When that's easily doable, I can reassign the issue to someone else.

neoaggelos:
#### Summary

Closes #816

#### Changes

- Introduces a new `DefaultDialOptionsWithDump()`. Passes a `dump` flag to the `rpclog.UnaryClientInterceptor`. When that is true, the interceptor adds a `grpc_request` log field with the request payload encoded as JSON

#### Notes for Reviewers

The client uses the `DefaultDialOptions()`, which already includes a whole stack of middleware interceptors, so doing this directly on CLI-only code is not an option (as it would drop the other interceptors).

The new `--dump-requests` flag is turned off by default, maintaining backwards compatibility.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- CLI can now dump JSON encoded `grpc_payload` field for unary requests (see `--dump-requests` flag).
+
 ### Changed
 
 ### Deprecated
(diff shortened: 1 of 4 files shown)


johanstokking:
I was not involved in the original issue – leaving this to the issue author.

I scanned through the code and it doesn't look wrong, but mostly wonder if we need this.

neoaggelos:
Updated to use `grpc.WithChainUnaryInterceptor()`. A "downside" is that payload is now printed in a separate log message. Example:

```
 DEBUG Request payload                          grpc_method=List grpc_payload={"field_mask":{},"limit":50,"page":1} grpc_service=ttn.lorawan.v3.GatewayRegistry namespace=grpc
 DEBUG Finished unary call                      duration=44.733135ms grpc_method=List grpc_service=ttn.lorawan.v3.GatewayRegistry namespace=grpc
```

PR #2248, opened 2020-03-27
cli: Local config takes precedence over global

### LINKED ISSUE #2247: CLI reads all available config files
#### Summary

The cli searchs for config files on a number of different locations (local, home dir, config dir). Even though this order suggests an hierarchy where the local files are given precedence, this is not the case. The cli will attempt to read all available files by default, overriding config in a not apparent manner.

#### Steps to Reproduce

1. Create a config file in user directory (`~/.ttn-lw-cli.yaml`) for host `other.example.com`
2. Create a config file in local directory (`./.ttn-lw-cli.yaml`) for host `this.example.com`
3. Run cli

#### What do you see now?

The CLI will try to connect to `other.example.com`, even though there is a local config file with a different host.

This makes having a local config file pointless, since the one from the user directory will always be read instead.

The CLI connecting to `this.example.com`, i.e. the host that is defined on the local config file.

An obvious work-around for this is to use `ttn-lw-cli -c .ttn-lw-cli.yaml`, but this is not obvious and may cause confusion to new users.

#### Environment

3.7.0

When loading config files, stop after successfully loading one, insteading of looping for the rest of the config files.

Yes

neoaggelos:
#### Summary

Closes #2247

#### Changes

- Read config files in reverse order (global -> local), so that configuration in local files takes precedence.

#### Notes for Reviewers

~Before reviewing or merging this, please make sure #2247 is actually classified as a bug and not desired behaviour.~

Changed PR after discussion in #2247

Also updating tests.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - End device location display bug when deleting the location entry in the Console.
 - GS could panic when gateway connection stats were updated while updating the registry.
+- Local CLI and stack config files now properly override global config.
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


KrishnaIyer:
Won't this break compatibility?

johanstokking:
Let's discuss in #2247 first before discussing a particular solution.

Issue #2249, opened 2020-03-27
Usage of lodash.merge in device reducer can cause faulty store updates


kschiffer:
#### Summary
Usage of `lodash.merge` can lead to faulty updates to the store.

#### Steps to Reproduce
1. Go to end device location settings
2. Set a location
3. Remove the location using the `Delete location entry` button
4. See the location still being displayed as before, even though the operation succeeded

#### What do you see now?
Faulty updates of the end device store

#### What do you want to see instead?
End device store being updated correctly

#### Environment
v3.7.0

#### How do you propose to implement this?
While working on gateway location form, I noticed that the end device location is not updated correctly in the end device store when the location is deleted. The reason being that we use `lodash.merge` to merge the current location entry with the updated end device coming from the update response. 
```js
// end device structure
{
   …
   location: {
     source: "SOURCE_REGISTRY"
     longitude: 1
     latitude: 1
     altitude: 1 
  }
}

// updated end device structure
{
  …
  location: {}
}
```
If we perform a deep merge, then the end device will not update the location value correctly. In other reducers we simply spread the patch over the current object, which will overwrite all top level properties of the object. In that case, the value would be applied correctly.

I don't know how dependant we are on the `lodash.merge` functionality for end device updates, especially with all the complexity surrounding end device settings. Hence, I did not replace the function.

@bafonins let me know whether you think we can replace it or what else you suggest here.

#### Can you do this yourself and submit a Pull Request?
Assigning @bafonins who should have the best overview of the issue.

PR #2252, opened 2020-03-27
Fix end device location reset in Console

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2249

#### Changes

- Always use `locations` entry from the response payload

#### Notes for Reviewers

> If we perform a deep merge, then the end device will not update the location value correctly. In other reducers we simply spread the patch over the current object, which will overwrite all top level properties of the object. In that case, the value would be applied correctly.

The problem here is not end device specific. It is about how we perform partial updates of deeply nested fields on entities.
In case of handling end devices, the problem is the object has many deeply nested fields that we allow updating in the general settings page. Think of `root_keys` with structure
```
	root_keys: {
		app_key: {
			key: ...
		}
	}
```
Now, in case when the user updates an end device from lw 1.0.0+ to 1.1.0 the `nwk_key` can be provided. We send it as `root_keys.nwk_key.key` and hence get the `root_keys.nwk_key.key` entry back. In order to preserve the `app_key` we need to perform a deep merge. With `session` it becomes even trickier.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- End device location display bug when deleting the location entry in the Console.
+
 ### Security
 
 ## [3.7.0] (2020-03-19)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Resetting end device location.
```
```suggestion
- End device location display bug when deleting the location entry.
```
The reset was working, it was just not displayed correctly.

PR #2253, opened 2020-03-27
Always enable device import button in console

kschiffer:
#### Summary
This quickfix PR will make the device import button always be enabled. Additionally, a warning is shown on the import view when no device templates have been found and consequently the device import cannot be performed. This should mitigate the confusion that is occasionally arising as to why the device import button is sometimes disabled.

#### Changes
- Make import button always enabled
- Check whether device templates are configured and show a warning in the import view if this is not the case
- Various related small styling fixes

#### Checklist


kschiffer:
> While you are on it, can you also translate the stack components names, e.g. Identity Server and the rest. Currently, those are hardcoded (https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/webui/constants/components.js)

I don't think we want to translate those. To my understanding, they're more like brand names belonging to TTS. cc @johanstokking

johanstokking:
Those won’t change indeed

PR #2260, opened 2020-03-29
Fix panic in GS stats update

neoaggelos:
#### Summary

References #1332 

Marshalling the connection stats object could panic if stats are updated at the same time. Reason was that the gateway status returned was a pointer. Change `io.Connection.StatusStats()` to create a deep copy of the `GatewayStatus` instead.

#### Changes

- `io.Connection.StatusStats()` to create a deep copy of the `GatewayStatus` instead of returning pointer.

#### Notes for Reviewers

The `GatewayStatus.Advanced` field is a `pbtypes.Struct` object. Not sure how to deepcopy that one except for marshalling/unmarshalling. That could pose the same risk as the one we are trying to solve (marshal will panic if the Advanced field changes).

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/io.go
+++ b/pkg/gatewayserver/io/io.go
@@ -20,6 +20,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/mohae/deepcopy"
 	"go.thethings.network/lorawan-stack/pkg/band"
 	"go.thethings.network/lorawan-stack/pkg/config"
 	"go.thethings.network/lorawan-stack/pkg/errorcontext"
@@ -238,7 +239,7 @@ func (c *Connection) HandleStatus(status *ttnpb.GatewayStatus) error {
 	case <-c.ctx.Done():
 		return c.ctx.Err()
 	case c.statusCh <- status:
-		c.lastStatus.Store(status)
+		c.lastStatus.Store(deepcopy.Copy(status))
 		atomic.StoreInt64(&c.lastStatusTime, time.Now().UnixNano())
 		c.notifyStatsChanged()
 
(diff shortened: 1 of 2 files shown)


pkg/gatewayserver/io/io.go
```diff
@@ -542,9 +545,46 @@ func (c *Connection) ConnectTime() time.Time { return time.Unix(0, c.connectTime
 
 // StatusStats returns the status statistics.
 func (c *Connection) StatusStats() (last *ttnpb.GatewayStatus, t time.Time, ok bool) {
```
johanstokking: The return value of `StatusStats()` should not be changed by any callers.

I think the solution is not copying the status on get, but making sure that an immutable instance is set.

neoaggelos: Does that not mean that anyone who updates the gateway status (and by extension, any of the connection stats) should make sure (now and in the future) should be careful to make the change immutable? How do we enforce this?

Also, what would be a good way to set an immutable instance of the gateway status?

neoaggelos: Updated code to store a copy of the gateway stats instead.

neoaggelos:
Okay, so, it appears that with the change of storing a copy of the gateway stats CI is failing again.

neoaggelos:
> * How about deepcopy of status itself, instead of a custom clone that we need to maintain?
> * How about making sure that the writer to the channel, i.e. those calling `HandleStatus()` are constructing a new object at all times?

Hmm, I had tried that at first inside `StatusStats()` but it didn't work. Could have been an unrelated change. Updated to simply deep copy gateway status

PR #2264, opened 2020-03-30
Add frontend error reporting

### LINKED ISSUE #2017: Add Sentry Error Reporting to Frontend
#### Summary

Report frontend errors to Sentry.

#### Why do we need this?

To improve our ability to debug our frontend.

The backend can report errors to Sentry.

The frontend cant'

#### Environment

Account app and Console app in the browser

- ~Blocked on https://github.com/TheThingsIndustries/infrastructure/issues/731~
- Propagate the `sentry.dsn` option to the frontend
- Implement with https://docs.sentry.io/platforms/javascript/

pgalic96:
#### Summary

Closes #2017 

#### Changes

- Add Sentry `dsn` to console config
- Add `@sentry/browser` to yarn
- Initialize sentry and add it to error boundary view

#### Notes for Reviewers

@htdvisser mentioned that we should test this on the dev environment before pushing it to production, not sure what is the process for deploying this?

Also, there is quite a lot of different options in Sentry that I haven't configured, such as user feedback dialogue in the Error View. Furthermore, sentry won't catch XHR errors right now, but seeing that we use some of these errors as part of original flow (401, 403 to logout user, for example), we should not add this functionality.

#### Checklist

### DIFF
--- a/cmd/ttn-lw-stack/commands/start.go
+++ b/cmd/ttn-lw-stack/commands/start.go
@@ -129,6 +129,9 @@ var startCommand = &cobra.Command{
 
 		if start.IdentityServer || startDefault {
 			logger.Info("Setting up Identity Server")
+			if config.IS.OAuth.UI.TemplateData.SentryDSN == "" {
+				config.IS.OAuth.UI.TemplateData.SentryDSN = config.Sentry.DSN
+			}
 			is, err := identityserver.New(c, &config.IS)
 			if err != nil {
 				return shared.ErrInitializeIdentityServer.WithCause(err)
@@ -228,6 +231,9 @@ var startCommand = &cobra.Command{
 
 		if start.Console || startDefault {
 			logger.Info("Setting up Console")
+			if config.Console.UI.TemplateData.SentryDSN == "" {
+				config.Console.UI.TemplateData.SentryDSN = config.Sentry.DSN
+			}
 			console, err := console.New(c, config.Console)
 			if err != nil {
 				return shared.ErrInitializeConsole.WithCause(err)
--- a/pkg/webui/lib/components/error-view.js
+++ b/pkg/webui/lib/components/error-view.js
@@ -14,6 +14,7 @@
 
 import React from 'react'
 import { withRouter } from 'react-router-dom'
+import * as Sentry from '@sentry/browser'
 
 import PropTypes from '../../lib/prop-types'
 
@@ -36,10 +37,14 @@ class ErrorView extends React.Component {
     this.unlisten()
   }
 
-  componentDidCatch(error) {
-    this.setState({
-      hasCaught: true,
-      error,
+  componentDidCatch(error, errorInfo) {
+    Sentry.withScope(scope => {
+      scope.setExtras(errorInfo)
+      Sentry.captureException(error)
+      this.setState({
+        hasCaught: true,
+        error,
+      })
     })
 
     // Clear the error when the route changes (e.g. user clicking a link)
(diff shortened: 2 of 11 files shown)


pkg/console/console.go
```diff
@@ -49,6 +50,7 @@ type FrontendConfig struct {
 	Language    string `json:"language" name:"-"`
 	SupportLink string `json:"support_link" name:"support-link" description:"The URI that the support button will point to"`
 	StackConfig `json:"stack_config" name:",squash"`
+	SentryLink  config.Sentry `json:"sentry_link" name:"sentry-link" description:"The Sentry URI used for error reporting"`
```
htdvisser: 1. This is shared between frontend apps, right? Then it belongs in `webui.TemplateData`.
2. Just call it `SentryDSN` instead of including the `config.Sentry` type
3. Please keep the description consistent.

htdvisser:
Regarding testing: I meant that we should make sure to test everything locally before we deploy it to production and realize that we're getting hundreds of errors per hour.

pkg/webui/lib/components/error-view.js
```diff
@@ -14,7 +14,7 @@
 
 import React from 'react'
 import { withRouter } from 'react-router-dom'
-
+import * as Sentry from '@sentry/browser'
```
kschiffer, line 17:
```suggestion
import * as Sentry from '@sentry/browser'

```

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -228,6 +228,9 @@ var startCommand = &cobra.Command{
 
 		if start.Console || startDefault {
 			logger.Info("Setting up Console")
+			if config.Console.UI.TemplateData.SentryDSN == "" {
```
kschiffer, line 231: I don't think this if-statement is necessary since the `SentryDSN` value cannot be set explicitly.

kschiffer: The config should also be added to the oauth app.

kschiffer: @htdvisser Do we want to enable sentry for the web UIs by default, when a global sentry DSN config is set? Implemented like that, there would be no way to opt-out.

pgalic96:
@kschiffer @bafonins I added the dsn to oauth as well, please check it with the actual dsn from our sentry. I have tested and it seems there is no danger of bombing the sentry instance since only the right errors are caught, but this should be reviewed.

cmd/ttn-lw-stack/commands/start.go
htdvisser: I think that is fine. If a deployment wants to enable error reporting, that includes both frontend and backend errors. They may choose to use a different DSN for frontend, which is possible with the current implementation.

pkg/webui/lib/selectors/env.js
```diff
@@ -28,6 +28,8 @@ export const selectApplicationSiteTitle = () => configSelector().SITE_TITLE
 
 export const selectApplicationSiteSubTitle = () => configSelector().SITE_SUB_TITLE
 
+export const selectSentryDsnConfig = () => selectApplicationConfig().SENTRY_DSN
```
kschiffer, line 31:
```suggestion
export const selectSentryDsnConfig = () => configSelector().SENTRY_DSN
```

pkg/webui/template.go
```diff
@@ -92,7 +93,8 @@ const appHTML = `
       window.APP_CONFIG={{.AppConfig}};
       window.SITE_NAME={{.SiteName}};
       window.SITE_TITLE={{.Title}};
-      window.SITE_SUB_TITLE={{.SubTitle}};
+			window.SITE_SUB_TITLE={{.SubTitle}};
+			window.SENTRY_DSN={{.SentryDSN}}
```
kschiffer, line 97:
```suggestion
			window.SENTRY_DSN={{.SentryDSN}};
```

pgalic96:
As I mentioned, I don't think we need to catch any XHR errors (at least not for now). The internal errors are reported by backend already.

PR #2274, opened 2020-03-31
Fix content flashing in console overview

### LINKED ISSUE #2218: Console overview page content flashing
#### Summary

The overview page changes its title/choosers content when the page loaded:
overview-flashing

#### Steps to Reproduce

1. Go to the overview page and refresh
2. See titles/choosers change

When throttling network requests it becomes even worse

#### What do you see now?

Content changing while the page is available

No content changes

#### Environment

v3.6.2	

@kschiffer @pgalic96

pgalic96:
#### Summary

Closes #2218

#### Changes

- Pass `_BASE` actions instead of general actions to fetching selector

#### Checklist

### DIFF
--- a/pkg/webui/console/store/selectors/fetching.js
+++ b/pkg/webui/console/store/selectors/fetching.js
@@ -21,8 +21,8 @@ const selectFetchingEntry = (state, id) => selectFetchingStore(state)[id] || fal
 /**
  * @example
  * const selectFetching = createFetchingSelector([
- * 'GET_ENTITY_LIST',
- * 'SEARCH_ENTITY_LIST'
+ * GET_ENTITY_LIST_BASE,
+ * SEARCH_ENTITY_LIST_BASE
  * ])
  * const selectEntityFetching = (state) => selectFetching(state)
  *
--- a/pkg/webui/console/views/overview/index.js
+++ b/pkg/webui/console/views/overview/index.js
@@ -29,9 +29,9 @@ import Spinner from '../../../components/spinner'
 import Link from '../../../components/link'
 import Animation from '../../../lib/components/animation'
 import { selectApplicationsTotalCount } from '../../store/selectors/applications'
-import { getApplicationsList, GET_APPS_LIST } from '../../store/actions/applications'
+import { getApplicationsList, GET_APPS_LIST_BASE } from '../../store/actions/applications'
 import { selectGatewaysTotalCount } from '../../store/selectors/gateways'
-import { getGatewaysList, GET_GTWS_LIST } from '../../store/actions/gateways'
+import { getGatewaysList, GET_GTWS_LIST_BASE } from '../../store/actions/gateways'
 import { createFetchingSelector } from '../../store/selectors/fetching'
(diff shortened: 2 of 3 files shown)


pkg/webui/console/views/overview/index.js
```diff
@@ -77,7 +77,7 @@ const componentMap = {
     return {
       applicationCount: selectApplicationsTotalCount(state),
       gatewayCount: selectGatewaysTotalCount(state),
-      fetching: createFetchingSelector([GET_APPS_LIST, GET_GTWS_LIST])(state),
+      fetching: createFetchingSelector([GET_APPS_LIST_BASE, GET_GTWS_LIST_BASE])(state),
```
bafonins: Please move it outside of `connect` so the function is not created any time the store gets updated.

```js
const overviewFetchingSelector = createFetchingSelector([GET_APPS_LIST_BASE, GET_GTWS_LIST_BASE])
```

and then simply use it

pkg/webui/console/store/selectors/fetching.js
```diff
@@ -21,8 +21,8 @@ const selectFetchingEntry = (state, id) => selectFetchingStore(state)[id] || fal
 /**
  * @example
  * const selectFetching = createFetchingSelector([
- * 'GET_ENTITY_LIST',
- * 'SEARCH_ENTITY_LIST'
+ * 'GET_ENTITY_LIST_BASE',
+ * 'SEARCH_ENTITY_LIST_BASE'
```
kschiffer, line 25: The old example is actually correct. What you mean is:
```suggestion
 * GET_ENTITY_LIST_BASE,
 * SEARCH_ENTITY_LIST_BASE
```

pgalic96:
@kschiffer I added webfontloader as per your instructions, I don't observe text flashing anymore but please double check this.

pkg/webui/oauth.js
```diff
@@ -25,6 +26,12 @@ import Init from './lib/components/init'
 
 import createStore from './oauth/store'
 
+WebFont.load({
```
kschiffer, line 29: While this fixes the issue of FUOT, it is not what we want, since the webfontloader will load the font from Google. Since we need to support offline mode, we need to continue fetching the font as a local asset.

kschiffer: After researching about this for a little more, I found this:
https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display

If that works then it should be the way to go.

pgalic96:
@kschiffer `fallback` seems as a prefered option which somewhat reduces FOUT but also does not limit accessibility. I have tested it and don't observe both FOUT and FOIT.

pkg/webui/styles/fonts.styl
```diff
@@ -66,5 +70,6 @@
   font-family: 'Material Icons'
   font-style: normal
   font-weight: 400
+  font-display: fallback
```
kschiffer, line 73:
```suggestion
  font-display: block
```
For icons, I think this is the better option. It's better not showing anything than the unstyled, lengthy icon ids.

PR #2292, opened 2020-04-03
Add webhook template functionality to the console

### LINKED ISSUE #945: Tiles for supported webhook integations

kschiffer:
#### Summary
Closes #945
References #1175 

This PR adds webhook template functionality to the console. Webhooks can then be created using webhook templates that have been set up in the stack config.

<details>

image
image
image
</details>

#### Changes
- Add `url-templates` package to assemble the base URLs
- Extend JS SDK with webhook template RPCs
- Add redux primitives and store logic for webhook templates
- Add webhook template form which takes the template and assembles the respective templated webhook form
- Update template form to show information about associated template
- Add webhook template info component
- Add template chooser view
- Extend and adapt webhook related routes and views

#### Notes for Reviewers
To set up the template(s), refer to @adriansmares [instructions](https://github.com/TheThingsNetwork/lorawan-stack/issues/945#issuecomment-584258732)
I've used the following template setup:
<details>
<summary>example.yml</summary>

```yml
template-id: example
name: Example Template
description: A simple webhook template
logo-url: https://upload.wikimedia.org/wikipedia/commons/3/30/Vector-based_example.svg
info-url: https://www.example.com
documentation-url: https://www.iana.org/domains/reserved
fields:
  - id: region
    name: Server region
    description: On which region should the service endpoint land on
    secret: false
    default-value: eu
format: json
headers:
  x-test: test {region}
create-downlink-api-key: true
base-url: https://www.example.com{/region}
paths:
  uplink-message: /uplink
```
</details>

I have implemented it so that the flow with choosing the template is only applied when there is at least one template configured. Otherwise, there will be no change in the UX.

#### Checklist


johanstokking:
Awesome. @adriansmares can you thoroughly review this?

@bafonins also the frontend technicalities.

adriansmares:
I've just finished testing this one out and I love it. Good job @kschiffer ! UX for webhooks just goes through the roof when you no longer see 1000 options but instead only the relevant ones for the integration you're trying to add.

I believe we can go further with this: at the moment it does not leverage the fact that you can save the values of the fields as part of the webhook using the `template_fields` mapping in the webhook (see [webhook definition](https://github.com/TheThingsNetwork/lorawan-stack/blob/master/api/api.md#ttn.lorawan.v3.ApplicationWebhook)). If during creation the fields are saved, then editing the webhook can reuse the same form as creation - the user will be fully shielded from the full options of a webhook.

The drawback of this is that the user cannot tweak the insides of a webhook created by a template, so one possible solution is to have a button `Convert to custom webhook` which basically sets the `template_id` and `template_fields` to `null` and allows the user to use the webhook as if it were a classical webhook.

pkg/webui/console/components/webhook-template-form/index.js
```diff
@@ -0,0 +1,170 @@
...
+
+    const webhook = {
+      ids: {
+        webhook_id: values.webhook_id,
+      },
+      template_ids: template.ids,
+      format: template.format,
+      headers,
+      base_url: urlTemplate.parse(template.base_url).expand(fields),
+      uplink_message: template.uplink_message,
```
adriansmares, line 70: These should be get expanded as well.

kschiffer:
> I believe we can go further with this: at the moment it does not leverage the fact that you can save the values of the fields as part of the webhook using the template_fields mapping in the webhook (see webhook definition). If during creation the fields are saved, then editing the webhook can reuse the same form as creation - the user will be fully shielded from the full options of a webhook.

Cool, let's do that. @johanstokking do you agree?

> The drawback of this is that the user cannot tweak the insides of a webhook created by a template, so one possible solution is to have a button Convert to custom webhook which basically sets the template_ids and template_fields to null and allows the user to use the webhook as if it were a classical webhook.

Good idea, will implement it this way.

johanstokking:
Yes, but I would rather push this to Next Up and consider other priorities first at the moment. But in general, yes, I do support this completely.

PR #2299, opened 2020-04-05
Improve cli config file generation

### LINKED ISSUE #1240: Use service discovery for CLI configuration

neoaggelos:
#### Summary

Closes #1240

#### Changes

- Use service discovery for the gRPC service ports
- Add `--grpc-port` and `--oauth-server-address` options in use command.

#### Notes for Reviewers

...

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ns.down.data.schedule.attempt` and `ns.down.join.schedule.attempt` events, which are triggered when Network Server attempts to schedule a respective downlink on Gateway Server.
 - `ns.down.data.schedule.success` and `ns.down.join.schedule.success` events, which are triggered when Network Server successfully schedules a respective downlink on Gateway Server.
 - `ns.down.data.schedule.fail` and `ns.down.join.schedule.fail` events, which are triggered when Network Server fails to schedule a respective downlink on Gateway Server.
+- Specify gRPC port and OAuth server address when generating a CLI config file with `ttn-lw-cli use` (see `--grpc-port` and `--oauth-server-address` options).
 
 ### Changed
 
--- a/cmd/ttn-lw-cli/commands/config.go
+++ b/cmd/ttn-lw-cli/commands/config.go
@@ -15,16 +15,18 @@
 package commands
 
 import (
-	"fmt"
-
 	"go.thethings.network/lorawan-stack/cmd/internal/commands"
 	conf "go.thethings.network/lorawan-stack/pkg/config"
 	"go.thethings.network/lorawan-stack/pkg/log"
+	"go.thethings.network/lorawan-stack/pkg/rpcmiddleware/discover"
 )
 
 var (
-	defaultClusterHost = "localhost"
-	defaultInsecure    = false
+	defaultInsecure                  = false
+	defaultClusterHost               = "localhost"
+	defaultGRPCAddress, _            = discover.DefaultPort(defaultClusterHost, discover.DefaultPorts[!defaultInsecure])
+	defaultOAuthServerBaseAddress, _ = discover.DefaultURL(defaultClusterHost, discover.DefaultHTTPPorts[!defaultInsecure], !defaultInsecure)
+	defaultOAuthServerAddress        = defaultOAuthServerBaseAddress + "/oauth"
 )
 
 // Config for the ttn-lw-cli binary.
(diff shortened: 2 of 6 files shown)


cmd/ttn-lw-cli/commands/config.go
```diff
@@ -20,11 +20,21 @@ import (
...
+	"go.thethings.network/lorawan-stack/pkg/rpcmiddleware/discover"
 )
 
+func httpAddress(host string, tls bool) string {
+	if tls {
+		return fmt.Sprintf("https://%s:8885", host)
+	}
+	return fmt.Sprintf("http://%s:1885", host)
+}
+
```
neoaggelos, line 32: This should probably be extracted in a separate package, just like `discover.WithDefaultPort`

htdvisser: Why don't we just use the same logic for HTTP as we do with GRPC? So actually using `discover.WithDefaultPort`?

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- You can specify server gRPC port and OAuth server address when generating a CLI config file with `ttn-lw-cli use` (see `--grpc-port` and `--oauth-server-address` options).
```
johanstokking, line 12: Changelog doesn't talk to the user

cmd/ttn-lw-cli/commands/config.go
johanstokking: This should definitely be in the discovery package

pkg/rpcmiddleware/discover/discover.go
```diff
@@ -28,7 +28,8 @@ import (
 
 var errAddress = errors.DefineInvalidArgument("address", "invalid address")
 
-func defaultPort(target string, port int) (string, error) {
+// WithDefaultPort appends port if target does not already have one.
+func WithDefaultPort(target string, port int) (string, error) {
```
johanstokking, line 32: `With` is reserved for options or funcs that change state

cmd/ttn-lw-cli/commands/config.go
neoaggelos: For HTTP, we also need the protocol (http/https). Added a `discover.DefaultURL` and using that.

CHANGELOG.md
neoaggelos: Updated text

pkg/rpcmiddleware/discover/discover.go
neoaggelos: Renamed to the original `DefaultPort`

pkg/rpcmiddleware/discover/discover.go
```diff
@@ -49,15 +50,37 @@ func defaultPort(target string, port int) (string, error) {
...
+var DefaultPorts = map[bool]int{
 	false: 1884,
 	true:  8884,
 }
 
+// DefaultHTTPPorts is a map of the default HTTP ports, with/without TLS.
+var DefaultHTTPPorts = map[bool]int{
+	false: 1885,
+	true:  8885,
+}
```
htdvisser, line 63: Not completely sure about these defaults. On local deployments we indeed use 1885/8885, but for online deployments all our guides use 80/443.

johanstokking: Right. This should indeed be 80/443.

pkg/rpcmiddleware/discover/discover.go
```diff
@@ -49,15 +50,37 @@ func defaultPort(target string, port int) (string, error) {
...
+
+// HTTPProto is a map of the HTTP protocols, with/without TLS.
+var HTTPProto = map[bool]string{
```
johanstokking, line 66: This is generally called a scheme, not a proto(col)

neoaggelos:
> How is this tested?

The discovery package is tested. There are no automated tests for most of the CLI functionality currently. It would be a good thing to have.

johanstokking:
Sure I see the unit tests. But did you test this manually, locally?

neoaggelos:
Just for reference, the changes have been tested locally:

- `grpc-port` can override the default port numbers (8884 or 1884)
- `oauth-server-address` can override the default oauth server address
- `insecure` is checked in order to decide default ports (1xxx vs 8xxx) and http scheme

Issue #2250, opened 2020-03-27
Support unsetting values in CLI


rvolosatovs:
#### Summary

CLI should support unsetting `nil`-able values

#### Why do we need this?

To be able to unset values

#### What is already there? What do you see now?

Only set of `nil`-able values
```
 --mac-settings.desired-rx1-delay string                           RX_DELAY_5|RX_DELAY_9|RX_DELAY_15|RX_DELAY_3|RX_DELAY_11|RX_DELAY_14|RX_DELAY_10|RX_DELAY_6|RX_DELAY_7|RX_DELAY_8|RX_DELAY_12|RX_DELAY_0|RX_DELAY_1|RX_DELAY_2|RX_DELAY_4|RX_DELAY_13
```

#### What is missing? What do you want to see?

Support for unsetting `nil`-able values
```
 --mac-settings.desired-rx1-delay string                           RX_DELAY_5|RX_DELAY_9|RX_DELAY_15|RX_DELAY_3|RX_DELAY_11|RX_DELAY_14|RX_DELAY_10|RX_DELAY_6|RX_DELAY_7|RX_DELAY_8|RX_DELAY_12|RX_DELAY_0|RX_DELAY_1|RX_DELAY_2|RX_DELAY_4|RX_DELAY_13|null
```

#### How do you propose to implement this?

Allow `null` as a value for nillable fields

#### Can you do this yourself and submit a Pull Request?

yes

neoaggelos:
Closed with #2301

Issue #2258, opened 2020-03-29
Inconsistent mac_settings enum values in NS responses


bafonins:
#### Summary

References: https://github.com/TheThingsNetwork/lorawan-stack/issues/2047

Values returned by NS for enums within the `mac_settings` field are inconsistent. Some for some NS returns raw values and for others a string representation is returned.

#### Steps to Reproduce

1. Set `mac_settings.rx1_delay.value`

```
// request
{
   "end_device":{
      "mac_settings":{
         "rx1_delay":{
            "value":5 // or 'RX_DELAY_5'
         }
      }
   },
   "field_mask":{
      "paths":[
         "mac_settings.rx1_delay.value",
      ]
   }
}

// response
{
   ...
   "mac_settings":{
      "rx1_delay":{
         "value":5
      },
   }
}

// Note, regardless of what you send as the value for mac_settings.rx1_delay.value you always get the raw enum value.
```

2. Set `mac_settings.ping_slot_periodicity.value`
```
// request

{
   "end_device":{
      "mac_settings":{
         "ping_slot_periodicity":{
            "value": 1 // or 'PING_EVERY_2S'
         }
      }
   },
   "field_mask":{
      "paths":[
         "mac_settings.ping_slot_periodicity.value"
      ]
   }
}

// response

{
   "mac_settings":{
      "ping_slot_periodicity":{
         "value":"PING_EVERY_2S"
      }
   }
}

// Again, regardless of what you send you always get the string representation of the value.

```

#### What do you see now?

Inconsistent values for `mac_settings` enum fields

#### What do you want to see instead?

**All** values should be returned in the same format (raw/string)

#### Environment

v3.7.0-rc2

#### How do you propose to implement this?

I guess first we have to agree on the format. Then it is the matter of adjusting https://github.com/TheThingsNetwork/lorawan-stack/blob/master/api/lorawan.proto and https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/ttnpb/lorawan.go

#### Can you do this yourself and submit a Pull Request?

Yes, but need input from @rvolosatovs @johanstokking @htdvisser

rvolosatovs:
What input is required from us (@rvolosatovs @johanstokking @htdvisser)? This this issue missing a `discussion` label?

[...]

PR #2301, opened 2020-04-05
Improve CLI handling of updating values to null

neoaggelos:
#### Summary

References #2250 

#### Changes

- Appends `null` as one of the optional values for null-able enums
- When value is `null`, do not set the proto field, effectively cleaning it up
- Fix a small error: For custom enum values (e.g. `RxDelayValue`) passing a random string would count as a zero or nil value (e.g. `--mac-settings.rx1-delay=random` would work). This has also been fixed.

#### Notes for Reviewers

Marking as a draft pull request for three reasons:

- Verify that the enum fields that can be set to `null` are properly recognised (please check the output of `ttn-lw-cli dev update --help` and make sure that there is no `null` value proposed for fields that it shouldn't.
- Currently, I have hard-coded field names that can not/should not be set to null (lorawan + mac versions). This could probably be done in a better way.  
- Doing the same for floating point values is a lot harder, as they are declared as floating point flags, so passing in null is not allowed. Example affected field: `--mac-settings.adr-margin`. An obvious solution would be to change these to string flags as well and parse the floating point value or null. Not sure if this would be required 

#### Testing

- I have tested that one can actually set and unset values for a number of fields.
- Probably needs more testing.

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/end_devices.go
+++ b/cmd/ttn-lw-cli/commands/end_devices.go
@@ -86,6 +86,7 @@ func forwardDeprecatedDeviceFlags(flagSet *pflag.FlagSet) {
 }
 
 var (
+	errConflictingPaths             = errors.DefineInvalidArgument("conflicting_paths", "conflicting set and unset field mask paths")
 	errEndDeviceEUIUpdate           = errors.DefineInvalidArgument("end_device_eui_update", "end device EUIs can not be updated")
 	errEndDeviceKeysWithProvisioner = errors.DefineInvalidArgument("end_device_keys_provisioner", "end device ABP or OTAA keys cannot be set when there is a provisioner")
 	errInconsistentEndDeviceEUI     = errors.DefineInvalidArgument("inconsistent_end_device_eui", "given end device EUIs do not match registered EUIs")
@@ -511,7 +512,7 @@ var (
 
 			device.SetFields(isRes, append(isPaths, "created_at", "updated_at")...)
 
-			res, err := setEndDevice(&device, nil, nsPaths, asPaths, jsPaths, []string{}, true, false)
+			res, err := setEndDevice(&device, nil, nsPaths, asPaths, jsPaths, nil, true, false)
 			if err != nil {
 				logger.WithError(err).Error("Could not create end device, rolling back...")
 				if err := deleteEndDevice(context.Background(), &device.EndDeviceIdentifiers); err != nil {
@@ -549,6 +550,10 @@ var (
 			if len(paths)+len(unsetPaths) == 0 {
 				logger.Warn("No fields selected, won't update anything")
 				return nil
+			}
+			if remainingPaths := ttnpb.ExcludeFields(paths, unsetPaths...); len(remainingPaths) != len(paths) {
+				overlapPaths := ttnpb.ExcludeFields(paths, remainingPaths...)
+				return errConflictingPaths.WithAttributes("field_mask_paths", overlapPaths)
 			}
 			var device ttnpb.EndDevice
 			if ttnpb.HasAnyField(paths, setEndDeviceToJS...) {
--- a/cmd/ttn-lw-cli/commands/end_devices.go
+++ b/cmd/ttn-lw-cli/commands/end_devices.go
@@ -511,7 +511,7 @@ var (
 
 			device.SetFields(isRes, append(isPaths, "created_at", "updated_at")...)
 
-			res, err := setEndDevice(&device, nil, nsPaths, asPaths, jsPaths, true, false)
+			res, err := setEndDevice(&device, nil, nsPaths, asPaths, jsPaths, []string{}, true, false)
 			if err != nil {
 				logger.WithError(err).Error("Could not create end device, rolling back...")
 				if err := deleteEndDevice(context.Background(), &device.EndDeviceIdentifiers); err != nil {
@@ -543,7 +543,10 @@ var (
 				return err
 			}
 			paths := util.UpdateFieldMask(cmd.Flags(), setEndDeviceFlags, attributesFlags(), endDevicePictureFlags)
-			if len(paths) == 0 {
+			rawUnsetPaths, _ := cmd.Flags().GetStringSlice("unset")
+			unsetPaths := util.NormalizePaths(rawUnsetPaths)
+
+			if len(paths)+len(unsetPaths) == 0 {
 				logger.Warn("No fields selected, won't update anything")
 				return nil
 			}
@@ -557,8 +560,8 @@ var (
 			device.Attributes = mergeAttributes(device.Attributes, cmd.Flags())
 			device.EndDeviceIdentifiers = *devID
 
+			paths = append(paths, unsetPaths...)
 			isPaths, nsPaths, asPaths, jsPaths := splitEndDeviceSetPaths(device.SupportsJoin, paths...)
-
 			if len(nsPaths) > 0 && config.NetworkServerEnabled {
 				if device.NetworkServerAddress == "" {
 					device.NetworkServerAddress = getHost(config.NetworkServerGRPCAddress)
@@ -593,7 +596,7 @@ var (
 			logger.WithField("paths", isPaths).Debug("Get end device from Identity Server")
 			existingDevice, err := ttnpb.NewEndDeviceRegistryClient(is).Get(ctx, &ttnpb.GetEndDeviceRequest{
 				EndDeviceIdentifiers: *devID,
-				FieldMask:            pbtypes.FieldMask{Paths: isPaths},
+				FieldMask:            pbtypes.FieldMask{Paths: ttnpb.ExcludeFields(isPaths, unsetPaths...)},
 			})
 			if err != nil {
 				return err
@@ -625,7 +628,7 @@ var (
 			}
 
 			touch, _ := cmd.Flags().GetBool("touch")
-			res, err := setEndDevice(&device, isPaths, nsPaths, asPaths, jsPaths, false, touch)
+			res, err := setEndDevice(&device, isPaths, nsPaths, asPaths, jsPaths, unsetPaths, false, touch)
 			if err != nil {
 				return err
 			}
@@ -1138,6 +1141,7 @@ func init() {
 	endDevicesUpdateCommand.Flags().AddFlagSet(attributesFlags())
 	endDevicesUpdateCommand.Flags().Bool("touch", false, "set in all registries even if no fields are specified")
 	endDevicesUpdateCommand.Flags().AddFlagSet(endDevicePictureFlags)
+	endDevicesUpdateCommand.Flags().AddFlagSet(util.UnsetFlagSet())
 	endDevicesCommand.AddCommand(endDevicesUpdateCommand)
 	endDevicesProvisionCommand.Flags().AddFlagSet(applicationIDFlags())
 	endDevicesProvisionCommand.Flags().AddFlagSet(dataFlags("", ""))
--- a/cmd/ttn-lw-cli/internal/util/flags.go
+++ b/cmd/ttn-lw-cli/internal/util/flags.go
@@ -56,6 +56,15 @@ var (
 	toUnderscore = strings.NewReplacer("-", "_")
 )
 
+// NormalizePaths converts arguments to field mask paths, replacing '-' with '_'
+func NormalizePaths(paths []string) []string {
+	normalized := make([]string, len(paths))
+	for i, path := range paths {
+		normalized[i] = toUnderscore.Replace(path)
+	}
+	return normalized
+}
+
 func NormalizeFlags(f *pflag.FlagSet, name string) pflag.NormalizedName {
 	return pflag.NormalizedName(toDash.Replace(name))
 }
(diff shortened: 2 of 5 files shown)


cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -603,7 +624,7 @@ func setField(rv reflect.Value, path []string, v reflect.Value) error {
 							break
 						}
 						var enum ttnpb.ADRAckLimitExponent
-						if err := enum.UnmarshalText([]byte(v.String())); err != nil {
+						if err := enum.UnmarshalText([]byte(v.String())); err != nil && enum != 0 {
```
rvolosatovs, line 627: IDEM

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -592,7 +613,7 @@ func setField(rv reflect.Value, path []string, v reflect.Value) error {
 							break
 						}
 						var enum ttnpb.ADRAckDelayExponent
-						if err := enum.UnmarshalText([]byte(v.String())); err != nil {
+						if err := enum.UnmarshalText([]byte(v.String())); err != nil && enum != 0 {
```
rvolosatovs, line 616: IDEM

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -570,7 +591,7 @@ func setField(rv reflect.Value, path []string, v reflect.Value) error {
 							break
 						}
 						var enum ttnpb.AggregatedDutyCycle
-						if err := enum.UnmarshalText([]byte(v.String())); err != nil {
+						if err := enum.UnmarshalText([]byte(v.String())); err != nil && enum != 0 {
```
rvolosatovs, line 594: IDEM

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -559,7 +580,7 @@ func setField(rv reflect.Value, path []string, v reflect.Value) error {
 							break
 						}
 						var enum ttnpb.PingSlotPeriod
-						if err := enum.UnmarshalText([]byte(v.String())); err != nil {
+						if err := enum.UnmarshalText([]byte(v.String())); err != nil && enum != 0 {
```
rvolosatovs, line 583: IDEM

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -581,7 +602,7 @@ func setField(rv reflect.Value, path []string, v reflect.Value) error {
 							break
 						}
 						var enum ttnpb.RxDelay
-						if err := enum.UnmarshalText([]byte(v.String())); err != nil {
+						if err := enum.UnmarshalText([]byte(v.String())); err != nil && enum != 0 {
```
rvolosatovs, line 605: Technically this can also be set to `0`, however we probably don't want to allow that.
This does not belong here though, and, if set to `0` CLI should ideally silently convert it to `1`, but still set. I think the simplest solution here is to just allow `0` for now and later think about conversion.

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -466,14 +478,23 @@ func SetFields(dst interface{}, flags *pflag.FlagSet, prefix ...string) error {
...
+	textUnmarshalerType = reflect.TypeOf((*encoding.TextUnmarshaler)(nil)).Elem()
+)
 
 func setField(rv reflect.Value, path []string, v reflect.Value) error {
 	rt := rv.Type()
 	vt := v.Type()
 	props := proto.GetProperties(rt)
 	for _, prop := range props.Prop {
 		if prop.OrigName == path[0] {
+			if len(path) == 1 && v.String() == "null" {
```
rvolosatovs, line 492: This should take the `field` type into account as well. Move this statement inside the `if` on `L499` and perhaps set the field to `nil` value explicitly?

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -466,14 +478,23 @@ func SetFields(dst interface{}, flags *pflag.FlagSet, prefix ...string) error {
...
+				if !prop.Required {
```
rvolosatovs, line 493: This should not be necessary

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -548,7 +569,7 @@ func setField(rv reflect.Value, path []string, v reflect.Value) error {
 							break
 						}
 						var enum ttnpb.DataRateIndex
-						if err := enum.UnmarshalText([]byte(v.String())); err != nil {
+						if err := enum.UnmarshalText([]byte(v.String())); err != nil && enum != 0 {
```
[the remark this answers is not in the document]
neoaggelos: Long comment ahead. Relevant part of the PR description:

>  For custom enum values (e.g. RxDelayValue) passing a random string would count as a zero or nil value (e.g. --mac-settings.rx1-delay=random would work). This has also been fixed.

One of the relevant fields for that is `mac-settings.ping-slot-data-rate-index`. The accepted values for that field is:

```
 --mac-settings.ping-slot-data-rate-index string                   DATA_RATE_1|DATA_RATE_3|DATA_RATE_4|DATA_RATE_5|DATA_RATE_6|DATA_RATE_13|DATA_RATE_14|DATA_RATE_15|DATA_RATE_12|DATA_RATE_0|DATA_RATE_2|DATA_RATE_7|DATA_RATE_8|DATA_RATE_9|DATA_RATE_10|DATA_RATE_11
```

So, in order to set to 0, one has to set `--mac-settings.ping-slot-data-rate-index=DATA_RATE_0`

Now, **Without** the highlighted change you get the following behaviour:

```
# DATA_RATE_0 works ok, sets value to 0
$ ttn-lw-cli dev update app1 dev1 --mac-settings.ping-slot-data-rate-index=DATA_RATE_0 

# 0 prints an error, as expected according to the help text
$ ttn-lw-cli dev update app1 dev1 --mac-settings.ping-slot-data-rate-index=0
error:cmd/ttn-lw-cli/internal/util:flag_value (invalid flag value)
    mac-settings.ping-slot-data-rate-index=invalid value "0" for DataRateIndexValue
exit status 255

# any random string, instead of printing an error, will set the value to 0 instead
$ ttn-lw-cli dev update app1 dev1 --mac-settings.ping-slot-data-rate-index=aasdfasdfasfasdf
```

This is because `UnmarshalText` will return a valid enum value equal to 0 instead of throwing an error. https://github.com/TheThingsNetwork/lorawan-stack/blob/85767bc21f60409b24572fd523cba21a820eada4/cmd/ttn-lw-cli/internal/util/flags.go#L551

**With** the change, what we get is:

```
# DATA_RATE_0 works ok, sets value to 0
$ ttn-lw-cli dev update app1 dev1 --mac-settings.ping-slot-data-rate-index=DATA_RATE_0 

# 0 prints an error, as expected according to the help text
$ ttn-lw-cli dev update app1 dev1 --mac-settings.ping-slot-data-rate-index=0
error:cmd/ttn-lw-cli/internal/util:flag_value (invalid flag value)
    mac-settings.ping-slot-data-rate-index=invalid value "0" for DataRateIndexValue
exit status 255

# any random string also prints an error
$ ttn-lw-cli dev update app1 dev1 --mac-settings.ping-slot-data-rate-index=aasdfasdfasfasdf
```

Same things apply for the rest of the enum values, resolving the other conversations.

What is interesting, we also have:

```
      --mac-state.desired-parameters.rx2-data-rate-index string         0|1|10|11|12|13|14|15|2|3|4|5|6|7|8|9
```

In general, for the desired parameters, the numerical values are expected, instead of the `DATA_RATE_x` strings. This looks to be a difference because of `DataRateIndexValue` vs `DataRateIndex` types. Yet, the cli accepts both:

```
$ ttn-lw-cli dev update app1 dev1 --mac-state.desired-parameters.rx2-data-rate-index=1
$ ttn-lw-cli dev update app1 dev1 --mac-state.desired-parameters.rx2-data-rate-index=DATA_RATE_1
```

Considering the above, this is why I went with the proposed change. I am open to suggestions.

cmd/ttn-lw-cli/internal/util/flags.go
neoaggelos: I agree with your comment, but I also do not think this is relevant to this PR. Resolving conversation just like the others.

cmd/ttn-lw-cli/internal/util/flags.go
neoaggelos: Would we get anything from setting the field to nil instead of not setting it at all? Preferred the latter.

cmd/ttn-lw-cli/internal/util/flags.go
neoaggelos: Indeed no fields are currently set as required (at least for end device updates that I checked). Don't know about the rest, plus I tried to make this safer for the future.

cmd/ttn-lw-cli/internal/util/flags.go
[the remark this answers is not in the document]
neoaggelos: #2258 also seems to relevant.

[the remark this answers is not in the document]
rvolosatovs: 1. `UnmarshalText` should return an error on invalid input, otherwise it's a bug and needs to be fixed. Giving a (very) quick look at `pkg/ttnpb/lorawan.go` I did not see anything wrong in this regard, but maybe I'm wrong.
2. We do need to support `0` as valid enum values for some enums. The only one which should not support `0` that I can think of now is `Rx1Delay`, but that, again, belongs in validation not actual unmarshaling logic.
3. Regarding numerical and string syntax, we *should* support both.

cmd/ttn-lw-cli/internal/util/flags.go
rvolosatovs: Let's add this when we need it

cmd/ttn-lw-cli/internal/util/flags.go
rvolosatovs: Maybe @htdvisser can comment on correctness here(e.g. if the original value may be non-zero, e.g a default value, I could imagine `--otaa` flag on create could set some fields, which the user may want to override), but it does seem like we always write for each other field type, so good to do at least for consistency's sake.

cmd/ttn-lw-cli/internal/util/flags.go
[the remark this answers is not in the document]
neoaggelos: 1. Okay, so no bugs in `UnmarshalText`. Just noticed that the error actually lies in the `err != nil` condition, does it not? :dancer:
2. Agreed, fixing the condition, cause this was the actual issue, not allowing zero values.
3. Both are supported, it's just that the values are missing in the help text.

neoaggelos:
Marked as ready for review.

rvolosatovs:
IMO `--some-field null` is better for consistency, but `--unset-some-field` is also fine by me. All nullable values must be `null` by default, so on creation none of this is necessary anyway.

neoaggelos:
Also agree with the `--unset-<field>` (or maybe `--unset <field>`). Chose not to go that way at first to avoid possible confusion that could be caused by something like `--field VAL --unset field`

rvolosatovs:
> Also agree with the `--unset-<field>` (or maybe `--unset <field>`). Chose not to go that way at first to avoid possible confusion that could be caused by something like `--field VAL --unset field`

True, another reason to go for `--field null`, otherwise we have to add additional error handling for this case.

htdvisser:
But then won't we have issues with the flags package? Because a `Uint64Value` is a `pflag.Uint64`, and that won't accept a `null` value, right?

cmd/ttn-lw-cli/internal/util/flags.go
neoaggelos: Added an explicit `--unset` option, this discussion is no longer relevant.

neoaggelos:
After trying multiple options, it seems that `--unset field1,field2` is the cleanest method to implement, as we only have to deal with field mask flags, and we can completely ignore the flag data types.

For now, implemented for the `dev update` command only, and testing to make sure that we don't break any stuff by accident. We should extend this to the other CLI update flags as well.

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -511,7 +511,7 @@ var (
 
 			device.SetFields(isRes, append(isPaths, "created_at", "updated_at")...)
 
-			res, err := setEndDevice(&device, nil, nsPaths, asPaths, jsPaths, true, false)
+			res, err := setEndDevice(&device, nil, nsPaths, asPaths, jsPaths, []string{}, true, false)
```
rvolosatovs, line 514:
```suggestion
			res, err := setEndDevice(&device, nil, nsPaths, asPaths, jsPaths, nil, true, false)
```

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -543,7 +543,10 @@ var (
 				return err
 			}
 			paths := util.UpdateFieldMask(cmd.Flags(), setEndDeviceFlags, attributesFlags(), endDevicePictureFlags)
-			if len(paths) == 0 {
+			rawUnsetPaths, _ := cmd.Flags().GetStringSlice("unset")
+			unsetPaths := util.NormalizePaths(rawUnsetPaths)
+
+			if len(paths)+len(unsetPaths) == 0 {
```
rvolosatovs, line 549: Let's check that there's no overlap between the two

neoaggelos: https://github.com/TheThingsNetwork/lorawan-stack/pull/2301/files#diff-06f9dcb39ac587641facf395bb27e8fbR554-R557

Issue #2171, opened 2020-03-17
LoRaWAN 1.1 uplink issues


sidpat:
#### Summary
TTN enterprise cloud hosted: Uplinks not working anymore.

#### Steps to Reproduce
1.  Create an OTAA v1.1 end device on the console
2. Send confirmed uplink from end device (also doesnt work for unconfirmed uplinks)

#### What do you see now?
The console shows the following 2 messages in the gateway logs after an uplink,
dropped uplink:
```
{
  "namespace": "pkg/gatewayserver",
  "name": "host_handle",
  "message_format": "host `{host}` failed to handle message",
  "attributes": {
    "host": "cluster"
  },
  "cause": {
    "namespace": "pkg/networkserver",
    "name": "device_not_found",
    "message_format": "device not found",
    "correlation_id": "3304c25c69a744c8ae1a519f9c8d7ad3",
    "code": 5
  },
  "code": 5
}
```

dropped uplink:
```
{
  "namespace": "pkg/encoding/lorawan",
  "name": "unknown_field",
  "message_format": "unknown `{lorawan_field}`",
  "attributes": {
    "lorawan_field": "MType",
    "value": "CONFIRMED_DOWN"
  },
  "code": 3
}
```

#### What do you want to see instead?
successful uplinks

#### Environment
TTN enterprise cloud-hosted (things were working 5 days ago, probably the new update broke something)

#### How do you propose to implement this?

...

#### Can you do this yourself and submit a Pull Request?

...

rvolosatovs:
Thanks for reporting, `RekeyInd` handling was broken recently, indeed. This is fixed in https://github.com/TheThingsNetwork/lorawan-stack/pull/2186 and will land in `3.6.2`, which will be released and deployed very soon.

sidpat:
Hi,

The issue still persists. Now upon uplink, I am still getting the below error message on the gateway data page,

```
{
  "namespace": "pkg/gatewayserver",
  "name": "host_handle",
  "message_format": "host `{host}` failed to handle message",
  "attributes": {
    "host": "cluster"
  },
  "cause": {
    "namespace": "pkg/networkserver",
    "name": "device_not_found",
    "message_format": "device not found",
    "correlation_id": "63140b75dc984db2a9c77761c39d1eb3",
    "code": 5
  },
  "code": 5
}
```

sidpat:
[...]

PR #2221, opened 2020-03-24
Network Server event handling and monitoring

rvolosatovs:
#### Summary

Fix NS uplink event handling and monitoring, event testing assertions
Closes #2238

#### Changes

- Refactor NS event handling
	- On uplink, trigger event on:
		- receive
		- successful processing
		- join-request sending to JS
		- join response receival from JS
	- On downlink, trigger event on:
		- sending TxRequest to GS
		- failure of TxRequest to GS
		- success of TxRequest to GS

- Only register and trigger AS "forwarding" events when actually sending to AS
- Significantly refactor flow test events
- Add 1.1 Class C OTAA flow test
- Improve logging:
	- On uplink, log:
		- frequency
		- data rate parameters (either bit rate or SF/BW)
		- devAddr match count

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, database and configuration, according to the stability commitments in `README.md`.
- [x] Testing: The changes are covered with unit tests. The changes are tested manually as well.
- [x] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #2307, opened 2020-04-06
Fix LoRaWAN 1.1 support

rvolosatovs:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2171

#### Changes

- Fix 1.1 uplink MIC calculation naming (aligned it with the spec directly) - block called `b0` currently is actually `b1` according to the spec, that also caused the confusion in making me think that the implementation is incorrect, let's fix this to avoid this happening in the future.
- Fix `DeviceModeInd` being marked as "initiated by NS"
- Add myself to `pkg/crypto` `CODEOWNERS`

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - End device claim display bug when claim dates not set.
+- DeviceModeInd handling for LoRaWAN 1.1 devices.
 
 ### Security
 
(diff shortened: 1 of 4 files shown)


rvolosatovs:
Tested this with FF1705 running mbed-os with patch from https://github.com/ARMmbed/mbed-os/pull/12762 applied on top of `feature-lorawan-1-1`

rvolosatovs:
The `InitiatedByDevice` fields in `pkg/encoding/lorawan/mac.go` is not something we can or should unit test, since that's just static data. This will be tested in Network Server indirectly though in #2221 

Regarding `pkg/crypto` I agree, but I'd expect the original implementation to be already covered by existing unit tests?

PR #2325, opened 2020-04-08
Set device and gateway location on map

### LINKED ISSUE #1380: Set Devices and Gateway location with map
#### Summary

The location form for devices and gateways should include a map, allowing auto filling of location by clicking on map.

#### Why do we need this?

Simplify User Experience 

A form with text inputs.

#### Environment

Browser

Yes

pgalic96:
#### Summary

Closes #1380

#### Changes

- Add `react-leaflet` which allows us to set `onClick` events and propagate values correctly without doing vanilla javascript magic to update markers
- Add map to location form and propagate values from map to fields and reverse

#### Notes for Reviewers

- There is one issue, since the map is re-rendered every time a marker is set on the map, the zoom value gets reset. Suggestions how to avoid this re-render while still reset the markers are welcome cc @kschiffer @bafonins 

#### Checklist

### DIFF
--- a/pkg/webui/components/map/index.js
+++ b/pkg/webui/components/map/index.js
@@ -1,4 +1,4 @@
-// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -13,15 +13,19 @@
 // limitations under the License.
 
 import React from 'react'
+
 import PropTypes from 'prop-types'
+import { Map, Marker, TileLayer } from 'react-leaflet'
 import classnames from 'classnames'
 import Leaflet from 'leaflet'
+import bind from 'autobind-decorator'
 
 import MarkerIcon from '@assets/auxiliary-icons/location_pin.svg'
 
 import style from './map.styl'
 
 // Reset default marker icon.
+
 delete Leaflet.Icon.Default.prototype._getIconUrl
 Leaflet.Icon.Default.mergeOptions({
   iconRetinaUrl: MarkerIcon,
@@ -35,79 +39,112 @@ Leaflet.Icon.Default.mergeOptions({
   shadowUrl: require('leaflet/dist/images/marker-shadow.png'),
 })
 
-export default class Map extends React.Component {
+const tileLayer = (
+  <TileLayer
+    url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
+    attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
+  />
+)
+const defaultMinZoom = 7
+
+export default class LocationMap extends React.Component {
   static propTypes = {
-    // Id is a string used to give the map a unique ID.
-    id: PropTypes.string.isRequired,
-    // LeafletConfig is an object which can contain any number of properties
+    className: PropTypes.string,
+    clickable: PropTypes.bool,
+    // `LeafletConfig` is an object which can contain any number of properties
     // defined by the leaflet plugin and is used to overwrite the default
     // configuration of leaflet.
-    leafletConfig: PropTypes.shape({}),
-    // Markers is an array of objects containing a specific properties.
+    leafletConfig: PropTypes.shape({
+      zoom: PropTypes.number,
+    }),
+    mapCenter: PropTypes.arrayOf(PropTypes.number),
+    mapRef: PropTypes.string,
+    // `markers` is an array of objects containing a specific properties.
     markers: PropTypes.arrayOf(
-      // Position is a object containing two properties latitude and longitude
-      // which are both numbers.
+      // `position` is a object containing two properties latitude and longitude which are both numbers.
       PropTypes.shape({
-        position: PropTypes.objectOf(PropTypes.number),
+        position: PropTypes.shape({
+          longitude: PropTypes.number,
+          latitude: PropTypes.number,
+        }),
       }),
-    ).isRequired,
-    // Widget is a boolean used to add a class name to the map container div for
-    // styling.
+    ),
+    onClick: PropTypes.func,
+    // `widget` is a boolean used to add a class name to the map container div for styling.
     widget: PropTypes.bool,
   }
 
   static defaultProps = {
     leafletConfig: {},
+    className: undefined,
     widget: false,
+    markers: [],
+    onClick: () => null,
+    mapCenter: undefined,
+    clickable: false,
+    mapRef: 'map',
   }
 
-  getMapCenter(markers) {
-    // This will calculate zoom and map center long/lang based on all markers
-    // provided. Currently it just returns the first marker.
-    // TODO: Add multiple markers
-    // https://github.com/TheThingsNetwork/lorawan-stack/issues/1241
-    return markers[0]
-  }
+  constructor(props) {
+    super(props)
 
-  createMap(config, id) {
-    this.map = Leaflet.map(id, {
-      ...config,
-    })
-  }
-
-  createMarkers(markers) {
-    markers.map(marker =>
-      Leaflet.marker([marker.position.latitude, marker.position.longitude]).addTo(this.map),
-    )
-  }
-
-  componentDidMount() {
-    const { id, markers } = this.props
-
-    const { position } = markers.length >= 1 ? this.getMapCenter(markers) : markers[0]
-
-    const config = {
-      layers: [
-        Leaflet.tileLayer('http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
-          attribution: '&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors',
-        }),
-      ],
-      center: [position.latitude, position.longitude],
-      zoom: 11,
-      minZoom: 1,
-      ...this.props.leafletConfig,
+    this.state = {
+      zoomLevel: props.leafletConfig.zoom,
     }
+  }
 
-    this.createMap(config, id)
-    this.createMarkers(markers, id)
+  // Fix the issue where tiles sometimes partially load.
+  @bind
+  componentDidMount() {
+    const map = this.refs.map.leafletElement
+    setTimeout(() => {
+      map.invalidateSize()
+    }, 250)
+  }
+
+  @bind
+  onZoomEvent(event) {
+    this.setState({ zoomLevel: event.target._zoom })
   }
 
   render() {
-    const { id, widget } = this.props
-
+    const {
+      className,
+      mapCenter,
+      clickable,
+      widget,
+      markers,
+      leafletConfig,
+      mapRef,
+      onClick,
+    } = this.props
+    const { zoomLevel } = this.state
     return (
-      <div className={style.container}>
-        <div className={classnames(style.map, { [style.widget]: widget })} id={id} />
+      <div className={classnames(style.container, className)}>
+        <Map
+          ref={mapRef}
+          className={classnames(style.map, {
+            [style.widget]: widget,
+            [style.click]: clickable,
+          })}
+          minZoom={defaultMinZoom}
+          onZoomend={this.onZoomEvent}
+          onClick={onClick}
+          {...leafletConfig}
+          zoom={zoomLevel}
+          center={mapCenter}
+        >
+          {tileLayer}
+          {markers.map(
+            marker =>
+              marker && (
+                <Marker
+                  key={`marker-${marker.position.latitude}-${marker.position.altitude}`}
+                  position={[marker.position.latitude, marker.position.longitude]}
+                />
+              ),
+          )}
+        </Map>
       </div>
     )
   }
--- a/pkg/webui/components/map/map.styl
+++ b/pkg/webui/components/map/map.styl
@@ -1,4 +1,4 @@
-// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -19,6 +19,9 @@
 
 .widget
   height: 25rem
+
+.click
+  cursor: pointer !important
 
 .container
   :global {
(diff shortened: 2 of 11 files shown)


pkg/webui/components/map/index.js
```diff
@@ -34,75 +33,62 @@ Leaflet.Icon.Default.mergeOptions({
...
+    const { widget, markers, onClick, leafletConfig } = this.props
+    const position =
+      markers.length >= 1
+        ? [markers[0].position.latitude, markers[0].position.longitude]
+        : [52.7, 4.9]
+    const map = (
+      <Map
+        className={classnames(style.map, { [style.widget]: widget })}
+        center={position}
+        zoom={7}
```
bafonins, line 72: can you try saving the zoom value as well? i think it might solve the issue with resetting zoom value after setting a marker

pkg/webui/console/components/location-form/index.js
```diff
@@ -129,6 +132,23 @@ class LocationForm extends Component {
     }
   }
 
+  @bind
+  handleClick(event) {
+    this.setState({ latitude: event.latlng.lat, longitude: event.latlng.lng })
+    this.form.current.setFieldValue('latitude', event.latlng.lat)
+    this.form.current.setFieldValue('longitude', event.latlng.lng)
```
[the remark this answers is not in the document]
bafonins: Also, you can use `setValues` here

pkg/webui/components/map/index.js
```diff
@@ -34,75 +33,62 @@ Leaflet.Icon.Default.mergeOptions({
...
-    return (
-      <div className={style.container}>
-        <div className={classnames(style.map, { [style.widget]: widget })} id={id} />
-      </div>
+    const { widget, markers, onClick, leafletConfig } = this.props
+    const position =
+      markers.length >= 1
+        ? [markers[0].position.latitude, markers[0].position.longitude]
+        : [52.7, 4.9]
+    const map = (
```
bafonins, line 68: You dont really need a local variable for this. why not wrapping it directly in the div container?

bafonins:
Could you update `leaflet` to 1.6.0 ?

pkg/webui/components/map/index.js
```diff
@@ -34,75 +33,62 @@ Leaflet.Icon.Default.mergeOptions({
...
+      <Map
+        className={classnames(style.map, { [style.widget]: widget })}
+        center={position}
+        zoom={7}
+        minZoom={7}
+        onClick={onClick}
+        {...leafletConfig}
+      >
+        <TileLayer
+          url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
```
kschiffer, line 78: This should also be a constant.

pkg/webui/console/components/location-form/index.js
```diff
@@ -155,10 +175,22 @@ class LocationForm extends Component {
...
+        ? [
+            {
+              position: {
+                latitude: Number(latitude),
+                longitude: Number(longitude),
+              },
+            },
+          ]
+        : undefined
+
```
kschiffer, line 193: I think we should look into avoiding multiline ternaries.
```suggestion
    let marker = undefined
    if (latitude && longitude) {
      marker =  [{
        position: {
          latitude: Number(latitude),
          longitude: Number(longitude),
        },
      }]
```

pgalic96: this needs to be fixed in prettier config, otherwise travis will complain

kschiffer: The styling is arbitrary, I was talking about replacing multiline ternaries with if statements.

pgalic96: aah, sorry noticed the wrong thing

pkg/webui/components/map/index.js
pgalic96: fixed by storing the local state of the variable. had to move it under `{...leafletConfig}` in component to avoid overwriting in case the zoom value is set in a parent component.

pgalic96:
@kschiffer 
Fixed stuff according to your feedback.

> Clicking on the map now leads to a weird slide. Seems like an issue with the dimensions of the map vs the dimensions of the container but 🤷‍♂️)

This is due to recentering the map based on the marker position. I fixed this, but as you can see, this is not so convenient when setting the location with latitude and longitude fields, as map does not automatically recenter to that field. I can try fixing this, but I think we should not overengineer this logic, for the sake of a very small UX improvement. Your decision, I don't think that this sliding is annoying tbh.

pkg/webui/components/map/index.js
```diff
@@ -34,74 +35,111 @@ Leaflet.Icon.Default.mergeOptions({
...
+          : this.getCurrentLocation(),
+    }
   }
 
-  createMarkers(markers) {
-    markers.map(marker =>
-      Leaflet.marker([marker.position.latitude, marker.position.longitude]).addTo(this.map),
-    )
+  getCurrentLocation() {
+    if ('geolocation' in navigator) {
```
bafonins, line 88: Can you create a react hook to access data from the geolocation api?

pkg/webui/components/map/index.js
```diff
@@ -34,74 +35,111 @@ Leaflet.Icon.Default.mergeOptions({
...
+      navigator.geolocation.getCurrentPosition(function(position) {
+        return [position.coords.latitude, position.coords.longitude]
```
bafonins, line 90: you can also read the altitude value

kschiffer:
> @kschiffer
> Fixed stuff according to your feedback.
> 
> > Clicking on the map now leads to a weird slide. Seems like an issue with the dimensions of the map vs the dimensions of the container but 🤷‍♂️)
> 
> This is due to recentering the map based on the marker position. I fixed this, but as you can see, this is not so convenient when setting the location with latitude and longitude fields, as map does not automatically recenter to that field. I can try fixing this, but I think we should not overengineer this logic, for the sake of a very small UX improvement. Your decision, I don't think that this sliding is annoying tbh. My suggestion is to revert this change and let it recenter each time the marker is set.

You should be able to hook into the `onBlur` event of the inputs to trigger the update in the map.

I get asked now for permission to use my location, but the map does not update after that and stays in Amsterdam. See my code comment about that. 
Also, the cursor in the map should be the default one. Otherwise, it's hard to target the precise position.

pgalic96:
so, I resolved most of the issues that we discussed:
- introduced new prop in location form `mapCenter` that gets updated only when the field input changes. this is connected to `mapCenter` state variable in the map itself via `getDerivedStateFromProps` and its used for centering the map. So, the map gets recentered only when the field change because the `mapCenter` doesn't get updated on map click.

- added new prop `clickable` to map which allows us to avoid `onClick` handler if the map is not clickable, and also changed the cursor to pointer (maybe a bit more user friendly than the default one?)

- regarding the geolocation, the location is updated asynchronously by updating the `mapCenter` state variable. I added the overlay component, but the leaflet css file has hardcoded `z-values` which conflict hard with overlay z-values so this component is not ideal for this case. any ideas how to approach this issue?

cc @kschiffer @bafonins

pkg/webui/components/map/map.styl
```diff
@@ -20,6 +20,9 @@
 .widget
   height: 25rem
 
+.click
+  cursor: pointer !important
```
kschiffer, line 24: Any way to get this to work without `!important`? (Though don't invest more than 5 minutes in that)

pkg/webui/components/map/index.js
```diff
@@ -34,74 +36,143 @@ Leaflet.Icon.Default.mergeOptions({
...
-    id: PropTypes.string.isRequired,
+    className: PropTypes.string,
+    clickable: PropTypes.bool,
     // LeafletConfig is an object which can contain any number of properties defined by the leaflet plugin and is used to overwrite the default configuration of leaflet.
-    leafletConfig: PropTypes.shape({}),
+    leafletConfig: PropTypes.shape({
+      zoom: PropTypes.number,
+    }),
+    mapCenter: PropTypes.arrayOf(PropTypes.number),
     // Markers is an array of objects containing a specific properties
```
kschiffer, line 57:
```suggestion
    // Markers is an array of objects containing a specific properties.
```

pkg/webui/components/map/index.js
```diff
@@ -34,74 +36,143 @@ Leaflet.Icon.Default.mergeOptions({
   shadowUrl: require('leaflet/dist/images/marker-shadow.png'),
 })
 
-export default class Map extends React.Component {
+const defaultLocation = [52.3, 4.8]
```
kschiffer, line 39: Let's use a neutral default location `[0,0]`

pkg/webui/components/map/index.js
```diff
@@ -34,74 +36,143 @@ Leaflet.Icon.Default.mergeOptions({
...
-    this.createMap(config, id)
-    this.createMarkers(markers, id)
+  @bind
+  handleClick(event) {
+    const { onClick, clickable } = this.props
+    if (!clickable) {
+      return
+    }
+    const map = this.refs.map.leafletElement
+    this.setState({ mapCenter: map.getCenter() })
```
kschiffer, line 136: I think the map component itself should stay neutral. If we make too many assumptions about the map behavior we will have a hard time using the map for other purposes too. You can already see that as your implementation now needs the `clickable` prop. The map should be first and foremost concerned with rendering a map based on its props.

So instead of keeping track of the map center in the state in the map component itself, that should be done by a parent component (in this case the `<LocationForm />`)

pkg/webui/components/map/index.js
```diff
@@ -34,74 +36,143 @@ Leaflet.Icon.Default.mergeOptions({
...
+  constructor(props) {
+    super(props)
 
-  createMap(config, id) {
-    this.map = Leaflet.map(id, {
-      ...config,
-    })
+    this.state = {
+      zoomLevel: props.leafletConfig.zoom,
+      mapCenter: defaultLocation,
```
kschiffer, line 87: What about the location from the marker? Currently the map widget on the overview just stays at the default location, instead of centering on the (first) marker.

pgalic96:
@kschiffer Moved the whole centering logic to location form. Indeed this is a much leaner approach.

pkg/webui/components/map/index.js
```diff
@@ -34,74 +35,110 @@ Leaflet.Icon.Default.mergeOptions({
...
+)
+const defaultMinZoom = 7
+
+export default class LocationMap extends React.Component {
   static propTypes = {
-    // Id is a string used to give the map a unique ID.
-    id: PropTypes.string.isRequired,
+    className: PropTypes.string,
+    clickable: PropTypes.bool,
     // LeafletConfig is an object which can contain any number of properties defined by the leaflet plugin and is used to overwrite the default configuration of leaflet.
```
kschiffer, line 50:
```suggestion
    // `LeafletConfig` is an object which can contain any number of properties defined by the leaflet plugin and is used to overwrite the default configuration of leaflet.
```

pkg/webui/components/map/index.js
```diff
@@ -34,74 +35,110 @@ Leaflet.Icon.Default.mergeOptions({
...
-    leafletConfig: PropTypes.shape({}),
-    // Markers is an array of objects containing a specific properties
+    leafletConfig: PropTypes.shape({
+      zoom: PropTypes.number,
+    }),
+    mapCenter: PropTypes.arrayOf(PropTypes.number),
+    mapRef: PropTypes.string,
+    // Markers is an array of objects containing a specific properties.
     markers: PropTypes.arrayOf(
       // Position is a object containing two properties latitude and longitude which are both numbers.
```
kschiffer, line 58:
```suggestion
      // `position` is an object containing two properties latitude and longitude which are both numbers.
```

pkg/webui/components/map/index.js
```diff
@@ -34,74 +35,110 @@ Leaflet.Icon.Default.mergeOptions({
...
+    clickable: PropTypes.bool,
     // LeafletConfig is an object which can contain any number of properties defined by the leaflet plugin and is used to overwrite the default configuration of leaflet.
-    leafletConfig: PropTypes.shape({}),
-    // Markers is an array of objects containing a specific properties
+    leafletConfig: PropTypes.shape({
+      zoom: PropTypes.number,
+    }),
+    mapCenter: PropTypes.arrayOf(PropTypes.number),
+    mapRef: PropTypes.string,
+    // Markers is an array of objects containing a specific properties.
```
kschiffer, line 56:
```suggestion
    // `markers` is an array of objects containing a specific properties.
```

pkg/webui/components/map/index.js
```diff
@@ -34,74 +35,110 @@ Leaflet.Icon.Default.mergeOptions({
...
-        position: PropTypes.objectOf(PropTypes.number),
+        position: PropTypes.shape({
+          longitude: PropTypes.number,
+          latitude: PropTypes.number,
+        }),
       }),
-    ).isRequired,
+    ),
+    onClick: PropTypes.func,
     // Widget is a boolean used to add a class name to the map container div for styling.
```
kschiffer, line 67:
```suggestion
    // `widget` is a boolean used to add a class name to the map container div for styling.
```

pkg/webui/components/map/widget/index.js
```diff
@@ -42,10 +42,19 @@ export default class MapWidget extends React.Component {
 
     const leafletConfig = {
       zoomControl: false,
+      zoom: 7,
+      minZoom: 1,
     }
+    const mapCenter = [markers[0].position.latitude, markers[0].position.longitude]
```
kschiffer, line 48:
```suggestion
    const mapCenter = Boolean(markers) && markers.length !== 0 ? [markers.length && markers[0].position.latitude, markers[0].position.longitude] : defaultLocation
```
This will crash the console if there is no location set. Please fix and double check that there is no issues with different location configurations.

pkg/webui/components/overlay/index.js
```diff
@@ -22,15 +22,15 @@ import Message from '../../lib/components/message'
...
     <div
-      className={classnames(style.overlay, {
+      className={classnames(overlayClassName, style.overlay, {
         [style.overlayVisible]: visible,
       })}
     />
     {visible && loading && (
-      <Spinner center>
+      <Spinner center className={spinnerClassName}>
         <Message content={sharedMessages.fetching} />
```
kschiffer, line 34: The message of the spinner should be passable (while defaulting to `sharedMessages.fetching`). In case of the location map it should say `Loading location…`

PR #2335, opened 2020-04-10
Drop FPort 0 uplinks with non-empty FOpts

rvolosatovs:
#### Summary

Drop FPort 0 uplinks with non-empty FOpts

#### Changes

- Drop FPort 0 uplinks with non-empty FOpts

#### Checklist


johanstokking:
Why? Missing issue with why we need this.

rvolosatovs:
According to the spec `FPort==0` frames carrying `FOpts` are invalid, currently we do not validate that and we should.
Regarding what it gives us, is that if there's, for example, a device, which sends `FPort==0` uplink with non-empty `FOpts` and the user tells us that something "does not work", we can clearly show that it's an issue of the malfunctioning device.

rvolosatovs:
> According to the spec `FPort==0` frames carrying `FOpts` are invalid, currently we do not validate that and we should.
> Regarding what it gives us, is that if there's, for example, a device, which sends `FPort==0` uplink with non-empty `FOpts` and the user tells us that something "does not work", we can clearly show that it's an issue of the malfunctioning device.

FTR, this is not true - the spec actually allows such frames.

PR #2336, opened 2020-04-10
Cleanup devices service in js sdk

bafonins:
#### Summary

This PR adjusts the `bulkCreate` method in the devices service to use `create` method instead of `_setDevice`

#### Changes

- Remove `_setDevice`
- Adjust the `create` method to accept `mask` to override field mask
- Check for selected components in the Console, instead of in the js sdk

#### Notes for Reviewers

Tested creating device via the Console, tested device import for microchip TNGLORA and eui-csv manifests.

Removed check:
```js
    if (supports_join && 'provisioner_id' in device && device.provisioner_id !== '') {
      throw new Error('Setting a provisioner with end device keys is not allowed.')
    }
```
as it contradicts the microchip TNGLORA template, @kschiffer do you recall the reason of adding it in the first place?

#### Checklist


kschiffer:
> @kschiffer do you recall the reason of adding it in the first place?

Can't recall. If I added this then it must have been as a result of a discussion with @johanstokking

johanstokking:
> Can't recall. If I added this then it must have been as a result of a discussion with @johanstokking

I don't recall the need for this check either.

Issue #2230, opened 2020-03-25
Console feature gaps v2/v3


kschiffer:
#### Summary
We need to organize our work towards closing the current feature gap of the v2 and v3 console. This umbrella issue keeps track of the progress.

#### Why do we need this?
- We should not work on completely new features until the v3 console can provide all features of v2 console.
- Organizing our process / roadmap

#### What is already there? What do you see now?
We already provide most of the features that users know from v2, but there are still some gaps.

##### General
- [ ] Profile settings (Name, Description, Email, Profile Picture, Delete own account) ⟶ #1422
- [ ] Providing Japanese and Chinese language version of the console ⟶ #2435
- [ ] Filters for event streams ⟶ #2231
- [x] Using maps to set location values ⟶ #1380

##### Applications
- [ ] Providing payload format testing (dry-running byte payloads through the payload formatted) ⟶ #2232
- [x] Providing template tiles for webhook integrations ⟶ #945 
- [x] Storage integration (TTI issue) [#1830](https://github.com/TheThingsIndustries/lorawan-stack/issues/)

##### End Devices
- [ ] Showing end device statuses in end device list ⟶ #2436
- [x] Showing the Device EUIs in the end device list ⟶ #2437
- [x] Showing end device last seen and frame counts in device overview ⟶ #2167 
- [ ] Resetting frame counts for end devices ⟶ #2434
- [x] Scheduling downlinks ⟶ #1948
- [x] Simulating uplinks ⟶ #2234
- [x] Setting frame counter widths ⟶ #2438

##### Gateways
- [ ] Setting and displaying gateway manufacturer info (Brand, Model, Antenna(s), etc) ⟶ #2441
- [x] Setting automatic update behavior ⟶ #2439
- [ ] Setting Indoor/Outdoor location flag ⟶ #2440
- [ ] Show "Last seen" when a gateway is offline

@htdvisser @johanstokking please review the list. There might be things in there that we actually don't want to add to v3, or in different form.

johanstokking:
I'm very happy with the overview issue. And I think it's good to have the issues filed. We can then prioritize them and put them in the concerning milestones.

kschiffer:
I filed all issues now. @laurensslats any specific priority you want us to respect?

benolayinka:
@kschiffer how about creating users in the console? This is still CLI only

kschiffer:
[...]

PR #2338, opened 2020-04-10
Schedule downlinks in the Console

### LINKED ISSUE #1948: Support Downlinks via the Console
#### Summary

Support Downlinks via the Console

#### Why do we need this?

To be on par with V2, also non-technical individuals can't be expected to setup the CLI (or use HTTP/MQTT hooks) only to send Downlinks.

Downlinks only via API (HTTP/MQTT) or CLI.

Support Downlink operations via the Console

#### Environment

`ttn-lw-stack v.3.5.1`

🤷‍♂ 

I can help deploy and test.

kschiffer:
#### Summary
Closes #1948 
References #2230 

image

#### Changes
- Add downlink queue service to JS SDK
- Add downlink form component
- Add device messages view
- Add byte utilities module
- Enable byte inputs without max value
- Make tabs hideable

#### Notes for Reviewers
The device needs to be activated for the push/replace functions to work. The easiest way to test this is therefore with ABP activated devices.

Byte input for byte values with unlimited length is not optimal yet. I'll follow up with a refactor that will fix this and add more useful functionality as well, but did not want to make this PR depend on that.

#### Checklist


pkg/webui/console/components/downlink-form/index.js
```diff
@@ -0,0 +1,137 @@
...
+  confirmed: 'Confirmed',
+  scheduleDownlink: 'Schedule Downlink',
+  downlinkSuccess: 'Downlink was scheduled successfully',
+  payloadDescription: 'The desired payload bytes of the downlink message',
+})
+
+const validationSchema = Yup.object({
+  _mode: Yup.string()
+    .oneOf(['replace', 'push'])
+    .isRequired(),
```
bafonins, line 48:
```suggestion
    .required(sharedMessages.validateRequired),
```

the whole app is broken

pkg/webui/console/components/downlink-form/index.js
```diff
@@ -0,0 +1,137 @@
...
+  frm_payload: Yup.string(),
+})
+
+const DownlinkForm = ({ appId, devId }) => {
+  const [error, setError] = useState('')
+  const handleSubmit = useCallback(
+    async ({ _mode, ...values }, { setSubmitting, resetForm }) => {
+      try {
+        values.frm_payload = hexToBase64(values.frm_payload)
+        await api.downlinkQueue[_mode](appId, devId, [values])
```
bafonins, line 63: If you type `01` into the `f_port` field the backend returns an error, while this should be a valid input. This happens because its sends `"01"` instead of just `1`. Please cast form values with `validationSchema.cast(...)`.

pkg/webui/console/components/downlink-form/index.js
kschiffer: Regarding the missing validation message: It's not possible to not provide a value in a radio group so 🤷‍♂️.

kschiffer:
> Looks good, however it is currently impossible to send an empty payload in the replace mode on fport 1

Oh yeah indeed. I will make a separate PR that will make the form submit button always enabled by default, since that is better UX.

PR #2345, opened 2020-04-11
doc: Fix broken format with <details> sections

### LINKED ISSUE #2344: Documentation formatting for <details> field is broken
#### Summary

Example: https://thethingsstack.io/v3.7.0/reference/end-device-templates/creating/

Pre-formatted output is broken, and also hides rest of document 

#### Why do we need this ?

Have proper documentation for our product/

Broken formatting in multiple places, for example:

image

Proper formatting, spacing, for example:

image

Fixable by adding new-lines below the `</summary>` tags, and a few `<br>`s

yes

neoaggelos:
#### Summary

Closes #2344

#### Changes

- Adds proper newlines and line breaks to fix formatting

#### Notes for Reviewers

- Tested locally.
- Does this belong in the changelog?

#### Checklist

### DIFF
--- a/doc/content/reference/end-device-templates/executing.md
+++ b/doc/content/reference/end-device-templates/executing.md
@@ -22,7 +22,7 @@ You can execute an end device as follows:
 $ cat example.json | ttn-lw-cli end-device template execute
 ```
 
-<details><summary>Show output</summary>
+Output:
 
 ```json
 {
@@ -38,7 +38,6 @@ $ cat example.json | ttn-lw-cli end-device template execute
   "frequency_plan_id": "US_902_928"
 }
 ```
-</details>
 
 The `end-device template execute` command **does not create** the end device. You can, however, easily pipe the output of `end-device template execute` to create the device. But first, you need to personalize the devices by assigning EUIs since this is a generic device template, see [Assigning EUIs]({{< relref "assigning-euis.md" >}}).
 
@@ -49,7 +48,7 @@ $ cat example.json \
   | ttn-lw-cli device create --application-id test-app
 ```
 
-<details><summary>Show output</summary>
+Output:
 
 ```json
 {
@@ -74,4 +73,3 @@ $ cat example.json \
   "supports_join": true
 }
 ```
-</details>
--- a/doc/content/reference/end-device-templates/executing.md
+++ b/doc/content/reference/end-device-templates/executing.md
@@ -23,6 +23,7 @@ $ cat example.json | ttn-lw-cli end-device template execute
 ```
 
 <details><summary>Show output</summary>
+
 ```json
 {
   "ids": {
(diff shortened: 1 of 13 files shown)


doc/content/reference/end-device-templates/executing.md
```diff
@@ -38,6 +39,7 @@ $ cat example.json | ttn-lw-cli end-device template execute
 }
 ```
 </details>
+<br>
```
neoaggelos, line 42: `<br>` line breaks have been added at the end of all `details` sections, because otherwise there is very little whitespace between the collapsible section and the text below.

johanstokking: I think we should fix this with styling, not with declarations in the content

neoaggelos: Changed to use a class instead of `<br>` tags. We cannot add this without a class because details sections before headers look ugly.

johanstokking:
@benolayinka are you proposing not using details at all, hide everything behind details, or make some clear guidelines of when (not) to use them?

doc/themes/the-things-stack/assets/css/theme.scss
```diff
@@ -217,4 +217,8 @@ html.has-navbar-fixed-top {
   }
 }
 
+details.with-padding {
+  padding-bottom: 1em;
+}
```
johanstokking, line 222: OK so what is now the rule for using this?

Is it when there is _no_ heading after the details?

Shouldn't we always add padding after the details instead?

The problem I have with this, is that content writers now have to think about styling. That goes against the concept of using Markdown as a documentation tool; separate designers from domain experts.

benolayinka:
> @benolayinka are you proposing not using details at all, hide everything behind details, or make some clear guidelines of when (not) to use them?

Proposing to not use details at all
- Code snippets are a critical part of the instruction flow
- Details require instruction writers to think about design
- The documentation I've seen looks good and is easy to follow without hidden sections

johanstokking:
OK, @neoaggelos can you consider removing the details?

Also please point @benolayinka at some examples where there is a lot of output, which might be too much not to put in details.

doc/themes/the-things-stack/assets/css/theme.scss
neoaggelos: Switched to using `details + p`  and `details + blockquote` selectors instead. This removes the need from content writers to think about styling, and covers our case.

We can also simply add padding after every `<details>` section, but that pushes the following headers too far down and makes the doc look kinda bad.

neoaggelos:
> Proposing to not use details at all
> 
> * Code snippets are a critical part of the instruction flow
> * Details require instruction writers to think about design
> * The documentation I've seen looks good and is easy to follow without hidden sections

I also agree with this. For most cases, details sections can actually go away. The places where they actually cover a lot of content, and are useful towards not distracting the user from the document flow that I can see are:

https://github.com/TheThingsNetwork/lorawan-stack/blob/1aea203c60d9226425014729e1f89049cafe95fa/doc/content/reference/end-device-templates/assigning-euis.md#L41-L201

And

https://github.com/TheThingsNetwork/lorawan-stack/blob/1aea203c60d9226425014729e1f89049cafe95fa/doc/content/reference/application-server-data/data-formats/_index.md

I am removing details sections from other places, where the content is very short.

johanstokking:
OK, this is good to go with @benolayinka blessing

benolayinka:
> OK, this is good to go with @benolayinka blessing

yep, thanks for removing the details sections @neoaggelos

PR #2349, opened 2020-04-11
Ensure webhook template store is not nil before accessing

### LINKED ISSUE #2348: AS returns 500 if no webhook templates provided
#### Summary

AS returns 500 when no webhook templates provided via the `--as.webhooks.templates.directory` flag.

#### Steps to Reproduce

1. Skip the `--as.webhooks.templates.directory` when initializing the lorawan-stack
2. Run:
```
$ curl -H "Authorization: Bearer $access_token" "http://localhost:8080/api/v3/as/webhook-templates?field_mask=base_url" | jq
{
  "code": 13,
  "message": "error:pkg/rpcserver:rpc_recovered (Internal Server Error)",
  "details": [
    {
      "namespace": "pkg/rpcserver",
      "name": "rpc_recovered",
      "message_format": "Internal Server Error",
      "correlation_id": "65b11d1b3bae4ff7bbf8f6c24ebe499a",
      "code": 13
    }
  ]
}
```

#### What do you see now?

500 Internal Server Error

log:
```
 DEBUG Finished unary call                      duration=517.941µs error=rpc error: code = InvalidArgument desc = error:pkg/errors:validation (invalid `ids`: embedded message failed validation) error_correlation_id=066dabbfb5cc40068f49f1ae2a8cc6ca error_name=validation error_namespace=pkg/errors grpc_code=InvalidArgument grpc_method=GetTemplate grpc_service=ttn.lorawan.v3.ApplicationWebhookRegistry namespace=grpc request_id=01E5N4JK0TT5E921X10ZC4RACB
runtime error: invalid memory address or nil pointer dereference
```

200 and an empty list as a response payload

#### Environment

v3.7.0

@adriansmares

neoaggelos:
#### Summary

Closes #2348

#### Changes

- Check if TemplateStore is nil before accessing.

#### Checklist

### DIFF
--- a/pkg/applicationserver/io/web/templates.go
+++ b/pkg/applicationserver/io/web/templates.go
@@ -93,8 +93,15 @@ func (ts *TemplateStore) prependBaseURL(template *ttnpb.ApplicationWebhookTempla
 	return nil
 }
 
+var (
+	errTemplateNotFound = errors.DefineNotFound("template_not_found", "template `{template_id}` not found")
+)
+
 // GetTemplate returns the template with the given identifiers.
 func (ts *TemplateStore) GetTemplate(ctx context.Context, req *ttnpb.GetApplicationWebhookTemplateRequest) (*ttnpb.ApplicationWebhookTemplate, error) {
+	if ts == nil {
+		return nil, errTemplateNotFound.WithAttributes("template_id", req.TemplateID)
+	}
 	template, err := ts.getTemplate(req.ApplicationWebhookTemplateIdentifiers)
 	if err != nil {
 		return nil, err
@@ -112,6 +119,9 @@ func (ts *TemplateStore) GetTemplate(ctx context.Context, req *ttnpb.GetApplicat
 
 // ListTemplates returns the available templates.
 func (ts *TemplateStore) ListTemplates(ctx context.Context, req *ttnpb.ListApplicationWebhookTemplatesRequest) (*ttnpb.ApplicationWebhookTemplates, error) {
+	if ts == nil {
+		return &ttnpb.ApplicationWebhookTemplates{}, nil
+	}
 	ids, err := ts.getAllTemplateIDs()
 	if err != nil {
 		return nil, err
--- a/pkg/applicationserver/io/web/templates.go
+++ b/pkg/applicationserver/io/web/templates.go
@@ -58,20 +58,37 @@ func (c TemplatesConfig) NewTemplateStore() (TemplateStore, error) {
 		var err error
 		fetcher, err = fetch.FromHTTP(c.URL, true)
 		if err != nil {
-			return nil, err
+			return &noopTemplateStore{}, err
 		}
 	default:
-		return nil, nil
+		return &noopTemplateStore{}, nil
 	}
 	baseURL, err := url.Parse(c.LogoBaseURL)
 	if err != nil {
-		return nil, err
+		return &noopTemplateStore{}, err
 	}
 	return &templateStore{
 		fetcher:   fetcher,
 		baseURL:   baseURL,
 		templates: make(map[string]queryResult),
 	}, nil
+}
+
+type noopTemplateStore struct {
+}
+
+var (
+	errTemplateNotFound = errors.DefineNotFound("template_not_found", "template `{template_id}` not found")
+)
+
+// GetTemplate implements TemplateStore.
+func (ts *noopTemplateStore) GetTemplate(ctx context.Context, req *ttnpb.GetApplicationWebhookTemplateRequest) (*ttnpb.ApplicationWebhookTemplate, error) {
+	return nil, errTemplateNotFound.New().WithAttributes("template_id", req.TemplateID)
+}
+
+// ListTemplates implements TemplateStore.
+func (ts *noopTemplateStore) ListTemplates(ctx context.Context, req *ttnpb.ListApplicationWebhookTemplatesRequest) (*ttnpb.ApplicationWebhookTemplates, error) {
+	return &ttnpb.ApplicationWebhookTemplates{}, nil
 }
 
 // templateStore implements TemplateStore using an underlying fetcher.
@@ -100,10 +117,6 @@ func (ts *templateStore) prependBaseURL(template *ttnpb.ApplicationWebhookTempla
 	template.LogoURL = ts.baseURL.ResolveReference(logoURL).String()
 	return nil
 }
-
-var (
-	errTemplateNotFound = errors.DefineNotFound("template_not_found", "template `{template_id}` not found")
-)
 
 // GetTemplate implements the TemplateStore interface.
 func (ts *templateStore) GetTemplate(ctx context.Context, req *ttnpb.GetApplicationWebhookTemplateRequest) (*ttnpb.ApplicationWebhookTemplate, error) {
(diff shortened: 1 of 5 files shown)


neoaggelos:
`TemplateStore` is nil when configuration is empty because of https://github.com/TheThingsNetwork/lorawan-stack/blob/7459858273e997a65431d3d8cbc2d40bd8210d5b/pkg/applicationserver/io/web/templates.go#L70

An alternative way would be to make sure TemplateStore is not nil instead.

pkg/applicationserver/io/web/templates.go
```diff
@@ -165,6 +166,9 @@ func (ts *TemplateStore) allTemplateIDs() (ids []string, err error) {
 }
 
 func (ts *TemplateStore) getAllTemplateIDs() ([]string, error) {
+	if ts == nil {
```
johanstokking, line 169: This feels weird to me; the caller should not call this method if the store is nil.

pkg/applicationserver/io/web/templates.go
```diff
@@ -196,6 +200,9 @@ func (ts *TemplateStore) template(ids ttnpb.ApplicationWebhookTemplateIdentifier
 }
 
 func (ts *TemplateStore) getTemplate(ids ttnpb.ApplicationWebhookTemplateIdentifiers) (t *ttnpb.ApplicationWebhookTemplate, err error) {
+	if ts == nil {
```
johanstokking, line 203: Idem. This is the caller's responsibility imo

neoaggelos:
We might need to do something in https://github.com/TheThingsNetwork/lorawan-stack/blob/0da92e3acedf7e8003b2550c4f3b4d5b7872dacb/pkg/applicationserver/applicationserver.go#L234 as well.

@adriansmares should this condition check for `as.webhookTemplates != nil` as well?

In that case, if the template store is `nil`, then the following error is returned:

```
{
  "code": 12,
  "message": "error:unknown:unknown (unknown service ttn.lorawan.v3.ApplicationWebhookRegistry)",
  "details": [
    {
      "message_format": "unknown service ttn.lorawan.v3.ApplicationWebhookRegistry",
      "code": 12
    }
  ]
}
```

like this one instead error 500 or 404.

Leaving it up to you whether we want to do this.

pkg/applicationserver/io/web/templates.go
neoaggelos: The nil check is now done on the method implementation. It feels weird there as well, since that one is also under `TemplateStore`. Resolving this conversation, also see my comment below.

johanstokking: Yes I see this is done on the method implementation, and that is not how it's supposed to be.

This would only work if we register a nil-service implementation. That would be wrong to do.

We would either register a "not found" service implementation, simply returning 404s on getters, or not register the service at all, leading to Unimplemented errors (probably 500s).

Up to @adriansmares to decide

adriansmares:
My take on this would be to do the following things:
1. Decouple the template store implementation from the interface - make `TemplateStore` an interface and make the current `TemplateStore` unexported (lowercase initial letter).
2. Create an no-op `TemplateStore` implementation that always returns an empty list for `List` and a not found error for `Get`.
3. Use the no-op implementation for the case when there is no underlying fetcher available. 

We cannot avoid registering the the service since the webhook templates methods are part of the webhooks service.

pkg/applicationserver/io/web/templates.go
```diff
@@ -38,22 +38,16 @@ type TemplatesConfig struct {
...
+	GetTemplate(ctx context.Context, req *ttnpb.GetApplicationWebhookTemplateRequest) (*ttnpb.ApplicationWebhookTemplate, error)
+	// ListTemplates returns the available templates.
+	ListTemplates(ctx context.Context, req *ttnpb.ListApplicationWebhookTemplatesRequest) (*ttnpb.ApplicationWebhookTemplates, error)
 }
 
-// NewTemplateStore returns a new *web.TemplateStore based on the configuration.
-// If no stores are provided, this method returns nil.
-func (c TemplatesConfig) NewTemplateStore() (*TemplateStore, error) {
+// NewTemplateStore returns a new *web.templateStore based on the configuration.
+// If no stores are provided, this method returns a *web.noopTemplateStore.
```
johanstokking, line 49: Don't mention unexposed types in exposed godoc

pkg/applicationserver/io/web/templates.go
```diff
@@ -64,24 +58,55 @@ func (c TemplatesConfig) NewTemplateStore() (*TemplateStore, error) {
...
 		baseURL:   baseURL,
 		templates: make(map[string]queryResult),
 	}, nil
 }
 
+type noopTemplateStore struct {
+}
+
+var (
+	errNoStore = errors.DefineUnavailable("err_no_store", "no template store")
```
johanstokking, line 81: `Unavailable` is retryable for clients. This is not retryable

My suggestion would be to:
- Return an empty list on `ListTemplates()` (and not return an error)
- Always return a `NotFound` (404) error on `GetTemplate()`

pkg/applicationserver/io/web/templates.go
```diff
@@ -93,8 +118,12 @@ func (ts *TemplateStore) prependBaseURL(template *ttnpb.ApplicationWebhookTempla
 	return nil
 }
 
-// GetTemplate returns the template with the given identifiers.
-func (ts *TemplateStore) GetTemplate(ctx context.Context, req *ttnpb.GetApplicationWebhookTemplateRequest) (*ttnpb.ApplicationWebhookTemplate, error) {
+var (
+	errTemplateNotFound = errors.DefineNotFound("template_not_found", "template `{template_id}` not found")
```
johanstokking, line 122: Maybe it's not in this diff, but where is this used?

pkg/applicationserver/io/web/templates.go
```diff
@@ -64,24 +58,55 @@ func (c TemplatesConfig) NewTemplateStore() (*TemplateStore, error) {
 		var err error
 		fetcher, err = fetch.FromHTTP(c.URL, true)
 		if err != nil {
-			return nil, err
+			return &noopTemplateStore{}, err
```
adriansmares, line 61: I'd rather leave `nil` here. In all cases if `err != nil` the store is unused.

pkg/applicationserver/io/web/templates.go
```diff
@@ -64,24 +58,55 @@ func (c TemplatesConfig) NewTemplateStore() (*TemplateStore, error) {
...
 		}
 	default:
-		return nil, nil
+		return &noopTemplateStore{}, nil
 	}
 	baseURL, err := url.Parse(c.LogoBaseURL)
 	if err != nil {
-		return nil, err
+		return &noopTemplateStore{}, err
```
adriansmares, line 68: Same as before, I'd rather have the `nil`.

pkg/applicationserver/io/web/templates.go
```diff
@@ -67,21 +60,52 @@ func (c TemplatesConfig) NewTemplateStore() (*TemplateStore, error) {
...
+type noopTemplateStore struct {
+}
+
+var (
+	errTemplateNotFound = errors.DefineNotFound("template_not_found", "template `{template_id}` not found")
+)
+
+// GetTemplate implements TemplateStore.
+func (ts *noopTemplateStore) GetTemplate(ctx context.Context, req *ttnpb.GetApplicationWebhookTemplateRequest) (*ttnpb.ApplicationWebhookTemplate, error) {
+	return nil, errTemplateNotFound.New().WithAttributes("template_id", req.TemplateID)
```
johanstokking, line 85: nit; `New()` is not necessary if you use `WithCause()` or `WithAttributes()`

PR #2350, opened 2020-04-11
console: Use manifest in public dir for asset hashes

### LINKED ISSUE #1704: Use Manifest in Public Dir for Frontend Asset Hashes
#### Summary

We should write the frontend manifest to a (JSON) file in the public dir, and make The Things Stack read that instead of compiling the manifest into the binary.

#### Why do we need this?

This would allow us to compile the backend and frontend in parallel.

The manifest is written to `pkg/webui/manifest.go` and compiled into the binary. This means that we need to compile the frontend **before** we can compile the backend.

We may be able to speed things up when we can compile them in parallel.

- Write the manifest to a (JSON) file in the `public` dir instead of `pkg/webui/manifest.go`.
- Make `pkg/web` read that file on start and register the hashes with `pkg/webui`

neoaggelos:
#### Summary

Closes #1704

#### Changes

- Load asset hashes from a YAML manifest.
- Babel produces the yaml manifest instead of Go code that goes into the binary.

#### Notes for Reviewers

Chose YAML instead of the proposed JSON format, because it is easier to write the template.

#### Checklist


pkg/web/web.go
```diff
@@ -168,16 +172,35 @@ func New(ctx context.Context, opts ...Option) (*Server, error) {
...
 			break
 		}
 	}
-	if staticDir != "" {
-		logger.WithFields(log.Fields("path", staticDir, "mount", options.staticMount)).Debug("Serving static assets")
+	if staticPath != "" {
+		staticDir := http.Dir(staticPath)
+		logger := logger.WithFields(log.Fields("path", staticDir, "mount", options.staticMount))
+		logger.Debug("Serving static assets")
 		s.Static(options.staticMount, staticDir, middleware.Immutable)
```
htdvisser, line 186: unrelated to this PR, but let's remove this line while we're editing this file

pkg/web/web.go
```diff
@@ -168,16 +172,35 @@ func New(ctx context.Context, opts ...Option) (*Server, error) {
...
+			staticPath = path
 			break
 		}
 	}
-	if staticDir != "" {
-		logger.WithFields(log.Fields("path", staticDir, "mount", options.staticMount)).Debug("Serving static assets")
+	if staticPath != "" {
+		staticDir := http.Dir(staticPath)
+		logger := logger.WithFields(log.Fields("path", staticDir, "mount", options.staticMount))
+		logger.Debug("Serving static assets")
```
htdvisser, line 185: Let's put this debug line in the end.

pkg/web/web.go
```diff
@@ -168,16 +172,35 @@ func New(ctx context.Context, opts ...Option) (*Server, error) {
...
+		// register hashed filenames
+		manifest, err := ioutil.ReadFile(filepath.Join(staticPath, "manifest.yaml"))
+		if err != nil {
+			logger.WithError(err).Warn("Failed to load manifest.yaml")
+			return s, nil
+		}
+		hashedFiles := make(map[string]string)
+		err = yaml.Unmarshal(manifest, &hashedFiles)
+		if err != nil {
+			logger.WithError(err).Warn("Corrupted manifest.yaml")
```
htdvisser, line 197: Shouldn't this error just be (wrapped and) returned?

neoaggelos: Not sure if that is what we want. Will that break development builds by chance?

If we are sure that it does not, I am fine with wrapping and returning the error.

htdvisser: I'd say that if the file exists, it must be valid yaml.

neoaggelos: Hmm, yeah if the file exists then it should be valid. Updated the code.

pkg/web/web.go
neoaggelos: Removed immutable middleware

PR #2361, opened 2020-04-15
cli: Use SNAP_USER_COMMON for config files if available

### LINKED ISSUE #2355: Snap generating config file fails because of missing directory.
#### Summary

`ttn-lw-cli use <host> --user`  fails when CLI is installed using snap.

#### Steps to Reproduce

1. `sudo snap install ttn-lw-stack`
2. `ttn-lw-stack.ttn-lw-cli use localhost --user`

#### What do you see now?

```
 $ ttn-lw-stack.ttn-lw-cli use localhost --user
error:cmd/ttn-lw-cli/commands:fail_write (failed to write `HOME/snap/ttn-lw-stack/181/.config/.ttn-lw-cli.yml`)
    file=HOME/snap/ttn-lw-stack/181/.config/.ttn-lw-cli.yml
--- open HOME/snap/ttn-lw-stack/181/.config/.ttn-lw-cli.yml: no such file or directory
```

This happens because there is no directory `.config` directory inside the snap folder. If you create the directory by hand then it works.

The command to succeed (somehow)

#### Environment

3.7.0

Two ideas:

- Create the destination directory if it does not exist already (not in favor of this one)
- When running in snap, add an extra default config file path `$SNAP_USER_COMMON` (which would be `HOME/snap/ttn-lw-stack/common`). Also, write configuration file there instead of the user config dir, which changes in every revision.

yes, but let's discuss what's best first @htdvisser, @rvolosatovs

neoaggelos:
#### Summary

Closes #2355

#### Changes

- Write configuration files in `SNAP_USER_COMMON` directory. This is so that config is using the common snap directory instead of the revision specific one.
- Read config from `SNAP_USER_COMMON` as well, if that is available.  

#### Notes for Reviewers

We should probably be doing something similar with the cache as well.

#### Checklist

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -118,11 +118,20 @@ snapcrafts:
     grade: stable
     confinement: strict
     publish: true
+    extra_files:
+      - source: config/snap/ttn-lw-stack.wrapper
+        destination: ttn-lw-stack.wrapper
+        mode: 0755
+      - source: config/snap/ttn-lw-cli.wrapper
+        destination: ttn-lw-cli.wrapper
+        mode: 0755
     apps:
       ttn-lw-stack:
         plugs: [ home, network, network-bind ]
+        command: ttn-lw-stack.wrapper
       ttn-lw-cli:
         plugs: [ home, network, network-bind ]
+        command: ttn-lw-cli.wrapper
 
 brews:
   - name: ttn-lw-stack
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - JSON uplink message doc edited for clarity.
+- The CLI snap version uses the `$SNAP_USER_COMMON` directory for config by default, so that it is preserved between revisions.
 
 ### Deprecated
 
--- a/config/snap/ttn-lw-cli.wrapper
+++ b/config/snap/ttn-lw-cli.wrapper
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+if [ -z "$XDG_CONFIG_HOME" ]; then
+    /usr/bin/env XDG_CONFIG_HOME="$SNAP_USER_COMMON" "$SNAP/ttn-lw-cli" "$@"
+else
+    "$SNAP/ttn-lw-cli" "$@"
+fi
--- a/config/snap/ttn-lw-stack.wrapper
+++ b/config/snap/ttn-lw-stack.wrapper
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+if [ -z "$XDG_CONFIG_HOME" ]; then
+    /usr/bin/env XDG_CONFIG_HOME="$SNAP_USER_COMMON" "$SNAP/ttn-lw-stack" "$@"
+else
+    "$SNAP/ttn-lw-stack" "$@"
+fi


rvolosatovs:
FTR, see also what other open-source projects do regarding this:
https://github.com/syncthing/syncthing/pull/3730
https://github.com/snapcrafters/eclipse/pull/9
https://github.com/rust-lang/rustup/pull/1898/files#diff-c16e29ded21c9ee98b67ecd74b9fa0cdR108

Approach with `XDG_CONFIG_HOME` is the most generic and cleanest though IMO. Unless it's possible to bind it to runtime-value of `SNAP_USER_COMMON` via config, we should just use `env` to set `XDG_CONFIG_HOME` when launching binaries in the snaps.

neoaggelos:
> FTR, see also what other open-source projects do regarding this:
> [syncthing/syncthing#3730](https://github.com/syncthing/syncthing/pull/3730)
> [snapcrafters/eclipse#9](https://github.com/snapcrafters/eclipse/pull/9)
> https://github.com/rust-lang/rustup/pull/1898/files#diff-c16e29ded21c9ee98b67ecd74b9fa0cdR108
> 
> Approach with `XDG_CONFIG_HOME` is the most generic and cleanest though IMO. Unless it's possible to bind it to runtime-value of `SNAP_USER_COMMON` via config, we should just use `env` to set `XDG_CONFIG_HOME` when launching binaries in the snaps.

Unfortunately, it looks like `goreleaser` does not support setting the `command` field for apps:

https://github.com/goreleaser/goreleaser/blob/master/pkg/config/config.go#L299-L304

neoaggelos:
That said, I agree that having `env XDG_CONFIG_HOME=$SNAP_USER_COMMON ttn-lw-cli` is probably the cleanest solution.

rvolosatovs:
> > FTR, see also what other open-source projects do regarding this:
> > [syncthing/syncthing#3730](https://github.com/syncthing/syncthing/pull/3730)
> > [snapcrafters/eclipse#9](https://github.com/snapcrafters/eclipse/pull/9)
> > [#1898 (files)](https://github.com/rust-lang/rustup/pull/1898/files#diff-c16e29ded21c9ee98b67ecd74b9fa0cdR108)
> > Approach with `XDG_CONFIG_HOME` is the most generic and cleanest though IMO. Unless it's possible to bind it to runtime-value of `SNAP_USER_COMMON` via config, we should just use `env` to set `XDG_CONFIG_HOME` when launching binaries in the snaps.
> 
> Unfortunately, it looks like `goreleaser` does not support setting the `command` field for apps:
> 
> [goreleaser/goreleaser:pkg/config/config.go@`master`#L299-L304](https://github.com/goreleaser/goreleaser/blob/master/pkg/config/config.go#L299-L304) [](/goreleaser/goreleaser/blob/HEAD@{2020-04-17T00:11:33Z}/pkg/config/config.go#L299-L304)

Can you open an issue upstream about this? I don't think we should introduce this kind of hacks to work around issues of our releasing software - it should be other way round. Can we somehow hook into the release process here? Can we use some kind of wrapper for this purpose? Can we just manually update the snap?

neoaggelos:
https://github.com/goreleaser/goreleaser/issues/1452

I would prefer to avoid custom-made solutions that might break randomly in the future.

neoaggelos:
Alright, after a lot of testing and trying to manually build the snap with the custom command, I get this interesting feat:

```
$ sudo snap install --dangerous local.snap
error: cannot read snap file: invalid definition of application "ttn-lw-stack": app description
       field 'command' contains illegal env XDG_CONFIG_DIR="$SNAP_USER_COMMON"
       "$SNAP/ttn-lw-stack" (legal: '^[A-Za-z0-9/. _#:$-]*$')
```

It seems that `=` is not a valid character anymore. This is `snapcraft` and not `goreleaser`, so not much we can do.

Tried a different way, setting `XDG_CONFIG_HOME` during stack initialization if needed. Not too clean and I would not be satisfied with that, but it is better than before.

rvolosatovs:
I still think this does not belong inside our binary, especially overwriting the environment.
I would not expect it to be much work to add env/command support to goreleaser, is it?
Otherwise, as I said earlier, we could find a way to hook into the pipeline and modify the snap somehow.
I'm fine also blocking this on https://github.com/goreleaser/goreleaser/issues/1452
@htdvisser what do you think we should do here?

htdvisser:
I agree with @rvolosatovs that we shouldn't be setting environment variables ourselves to work around limitations of the user environment (not having XDG), Go (not supporting SNAP_USER_COMMON) or goreleaser (not supporting environ in snap).

I think the cleanest workaround is configuring the environment in the Snap, so I also think it's best to block this on Goreleaser support for that.

neoaggelos:
Converted to draft until https://github.com/goreleaser/goreleaser/pull/1553

neoaggelos:
Updated configuration to use a wrapper script to set the environment properly. Keeping still as a draft until the goreleaser PR is released.

neoaggelos:
After updating goreleaser to v0.138.0, which is required for the changes made in this PR, running `go mod tidy` introduces a huge diff and version changes on other dependencies. Some are minor versions even. @htdvisser what should I do here?

htdvisser:
Please discuss with @rvolosatovs. He's the reviewer on this PR.

rvolosatovs:
@neoaggelos see https://github.com/TheThingsNetwork/lorawan-stack/pull/2749

.goreleaser.yml
```diff
@@ -118,11 +118,20 @@ snapcrafts:
     grade: stable
     confinement: strict
     publish: true
+    extra_files:
+      - source: config/snap/ttn-lw-stack.wrapper
```
rvolosatovs, line 122: Is it a convention in snapcraft ecosystem to add `.wrapper` prefix to these?

neoaggelos: Yes, and most of the time they are automatically created when using snapcraft directly.

PR #2372, opened 2020-04-17
Set DevEUI if available for ABP devices

johanstokking:
#### Summary

Set DevEUI if available for ABP devices

#### Changes

- Set DevEUI even if `supports_join` is false, i.e. ABP. Since LoRaWAN 1.0.4, DevEUIs are mandatory for all devices

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/end_devices.go
+++ b/cmd/ttn-lw-cli/commands/end_devices.go
@@ -18,12 +18,13 @@ import (
 	"bufio"
 	"context"
 	"encoding/hex"
-	stdio "io"
 	"io/ioutil"
 	"mime"
 	"os"
 	"path"
 	"strings"
+
+	stdio "io"
 
 	pbtypes "github.com/gogo/protobuf/types"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -18,13 +18,14 @@ import (
...
 	"context"
 	"encoding/hex"
-	stdio "io"
 	"io/ioutil"
 	"mime"
 	"os"
 	"path"
 	"strings"
 
+	stdio "io"
```
johanstokking, line 27: Not sure if this is gofumports or Go 1.14 but let's not care about this

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -475,13 +476,11 @@ var (
...
 					device.ApplicationID = devID.ApplicationID
 				}
-				if device.SupportsJoin {
-					if devID.JoinEUI != nil {
-						device.JoinEUI = devID.JoinEUI
-					}
-					if devID.DevEUI != nil {
-						device.DevEUI = devID.DevEUI
-					}
+				if device.SupportsJoin && devID.JoinEUI != nil {
```
rvolosatovs, line 479: IMO we should not check for `SupportsJoin` here at all. Just set whatever user has set.

FTR, 1.0.4 spec reads:
2020-04-17-14:28:35-screenshot
The way understand it is that JoinEUI, if fact, *may* be set for ABP.

johanstokking: JoinEUI has no function with ABP

Issue #2323, opened 2020-04-08
Inconsistent responses on end device deletion requests


bafonins:
#### Summary

IS and JS behave differently from NS and AS on repeated end device deletion.

#### Steps to Reproduce

1. Create a new end device
```
// delete existing entry from is:
$ curl -X DELETE -H "Authorization: Bearer $access_token" http://localhost:1885/api/v3/applications/test-app/devices/test-dev2 | jq

{}

// try again:
$ curl -X DELETE -H "Authorization: Bearer $access_token" http://localhost:1885/api/v3/applications/test-app/devices/test-dev2 | jq

{
  "code": 5,
  "message": "error:pkg/identityserver/store:end_device_not_found (end device `test-app:test-dev2` not found)",
  "details": [
    {
      "@type": "type.googleapis.com/ttn.lorawan.v3.ErrorDetails",
      "namespace": "pkg/identityserver/store",
      "name": "end_device_not_found",
      "message_format": "end device `{application_id}:{device_id}` not found",
      "attributes": {
        "application_id": "test-app",
        "device_id": "test-dev2"
      },
      "correlation_id": "1899b283a1f847e0b5ddd328fdf9f713",
      "code": 5
    }
  ]
}

// delete existing entry from js:
$ curl -X DELETE -H "Authorization: Bearer $access_token" http://localhost:1885/api/v3/js/applications/test-app/devices/test-dev2 | jq

{}

// try again:
$ curl -X DELETE -H "Authorization: Bearer $access_token" http://localhost:1885/api/v3/js/applications/test-app/devices/test-dev2 | jq

{
  "code": 5,
  "message": "error:pkg/joinserver:device_not_found (device not found)",
  "details": [
    {
      "@type": "type.googleapis.com/ttn.lorawan.v3.ErrorDetails",
      "namespace": "pkg/joinserver",
      "name": "device_not_found",
      "message_format": "device not found",
      "correlation_id": "73d3a76c6b04481395ac0aba5a04531c",
      "code": 5
    }
  ]
}

// delete existing entry from ns:
$ curl -X DELETE -H "Authorization: Bearer $access_token" http://localhost:1885/api/v3/ns/applications/test-app/devices/test-dev2 | jq

{}

// try again:
$ curl -X DELETE -H "Authorization: Bearer $access_token" http://localhost:1885/api/v3/ns/applications/test-app/devices/test-dev2 | jq

{}

// delete existing entry from as:
[...]

PR #2374, opened 2020-04-20
Remove todo from devices service in JS SDK

bafonins:
#### Summary

This PR removes todo from the `_deleteDevice` method in the devices service since https://github.com/TheThingsNetwork/lorawan-stack/issues/2323

#### Changes

- Remove todo 
- Remove redux related imports from the view to a separate `connect.js` file

#### Notes for Reviewers

Previously I though that we should throw an error on 404 responses, however now it seems redundant as the only option to delete a device via the sdk is by using the `deleteById` method which by contract deletes the device from all components. In case of ABP device, entry in the join server wouldnt exist and hence 404 returned by js whould be thrown. Similar logic applies when deleting devices that were created via the API (or console with previously disabled components), for example IS+JS combination. If we really want to throw an error on 404 for as,ns and js we should either expose methods like `deleteNS`, `deleteAS`, etc, or allow the users to specify the list of components to query (`deleteById (..., components = ['is', ...]`). In this case the user of the sdk is responsible for checking the device existence and other prerequisites to handle 404 cases. However I dont think that adding component name abbreviations to public API is a good idea. 

I think for now the contract of throwing 404 only if request to IS fails works fine.

#### Checklist


sdk/js/src/service/devices/index.js
```diff
@@ -136,9 +136,8 @@ class Devices {
       ),
     )
 
-    // Check for errors and filter out 404 errors, since we cannot consistently return
-    // not found errors.
-    // TODO: Check for 404 errors, see https://github.com/TheThingsNetwork/lorawan-stack/issues/2323
+    // Check for errors and filter out 404 errors. We do not 404 responses
```
kschiffer, line 139:
```suggestion
    // Check for errors and filter out 404 errors. We do not regard 404 responses
```

PR #2382, opened 2020-04-21
Starting point for migrating away from Echo

htdvisser:
#### Summary

This is the starting point for migrating away from Echo.

After this PR is ready and merged, I'll create issues for replacing smaller parts.

#### Changes

- Implement a number of middlewares that are or can be used in `pkg/web` or other packages with a web server (such as interop and basicstation).
	- Notable new middleware is the ProxyHeaders middleware that makes The Things Stack only trust `X-Forwarded-*` headers for configured clients (by default, clients on the local network). For other clients, it deletes those headers, as they may be unreliable.
- Wrap the Echo server with Gorilla.Mux

#### Notes for Reviewers

What's left for this PR:

- [x] Unit tests for `pkg/webmiddleware`
- [ ] Regression testing

#### Checklist


pkg/webmiddleware/proxy_headers.go
```diff
@@ -0,0 +1,122 @@
...
+	"regexp"
+	"strings"
+)
+
+const (
+	headerForwarded = "Forwarded"
+
+	headerXForwardedFor   = "X-Forwarded-For"
+	headerXForwardedHost  = "X-Forwarded-Host"
+	headerXForwardedProto = "X-Forwarded-Proto" // We don't support non-standard headers such as Front-End-Https, X-Forwarded-Ssl, X-Url-Scheme.
```
johanstokking, line 30: X-headers are also non-standard I believe, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

pkg/webmiddleware/proxy_headers.go
```diff
@@ -0,0 +1,122 @@
...
+	headerXRealIP         = "X-Real-IP"
+
+	schemeHTTP  = "http"
+	schemeHTTPS = "https"
+)
+
+var (
+	proxyHeaders        = []string{headerForwarded, headerXForwardedFor, headerXForwardedHost, headerXForwardedProto, headerXRealIP}
```
johanstokking, line 38: Consider adding `X-Forwarded-Client-Cert` for mTLS

pkg/webmiddleware/proxy_headers.go
htdvisser: Those are indeed not standardized, but they are de-facto standard

cmd/internal/shared/config.go
```diff
@@ -48,8 +48,9 @@ var DefaultClusterConfig = cluster.Config{}
 
 // DefaultHTTPConfig is the default HTTP config.
 var DefaultHTTPConfig = config.HTTP{
-	Listen:    ":1885",
-	ListenTLS: ":8885",
+	Listen:         ":1885",
+	ListenTLS:      ":8885",
+	TrustedProxies: []string{"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"},
```
rvolosatovs, line 53: Shouldn't this be explicitly specified by the user? Or at least the user should be able to overwrite these. Currently, it seems that the only option is append.

pkg/webmiddleware/proxy_headers.go
htdvisser: Ok. I added the headers injected by tls-terminating reverse proxies Envoy/HAProxy and Traefik. In `nginx` the header is configurable.

Similar to what we (don't) do with the Forwarded/XFF header, we'll leave it to handler implementations to detect the header and parse it.

cmd/internal/shared/config.go
htdvisser: The goal of the default configuration is to have some defaults that work well for most deployments. These three CIDRs match the private address space specified in RFC1918. Setting this config explicitly in config files, environment or flags will overwrite it, not append.

kschiffer:
What's the reasoning behind going away from echo?

htdvisser:
@kschiffer because HTTP is an important part of our project. A patch release of Echo (v4.1.2 to v4.1.3) broke our routing, and the issue we created about it didn't get any attention. We are no longer confident that we can rely on a critical dependency like this.

PR #2385, opened 2020-04-21
doc: Add upgrade instructions

### LINKED ISSUE #2373: Document update procedure (Docker Compose)
#### Summary

I'm running The Things Stack V3.6.0 and want to update to V3.7.0. It's running using the provided docker-compose file with customisation as documented in the installation guide (https://thethingsstack.io/v3.7.0/guides/getting-started/installation/).

#### Why do we need this ?

It's unclear what procedure should be followed to to an update, as rumour has it that a database migrations is necessary. Therefore it is not a simple change of the docker image version in the compose file.

See https://github.com/TheThingsNetwork/lorawan-stack/issues/2284

The documentation explains very well how to do a new clean installation of the stack.

I would like to see a step by step guide how to perform a version upgrade.

As another guide on https://thethingsstack.io/v3.7.0/guides/getting-started/installation/

I'm not sure how to do the upgrade yet. And perhaps there are a couple of steps to follow to prevent data loss or corruption.

benolayinka:
#### Summary

Fixes #2373 

#### Changes

- Add `Upgrading` section to docs

#### Notes for Reviewers

Is it okay to include output of terminal in bash instructions? It can be helpful but we don't typically do it.

#### Checklist

### DIFF
--- a/doc/content/getting-started/upgrading.md
+++ b/doc/content/getting-started/upgrading.md
@@ -1,9 +1,12 @@
 ---
 title: "Upgrading"
 description: ""
+weight: -1
 ---
 
 Follow these instructions to upgrade an existing {{% tts %}} instance.
+
+<!--more-->
 
 Replace the tag in your **docker-compose.yml** file with the newer version of {{% tts %}}:
 
--- a/doc/content/guides/getting-started/upgrading.md
+++ b/doc/content/guides/getting-started/upgrading.md
@@ -1,0 +1,62 @@
+---
+title: "Upgrading"
+description: ""
+---
+
+Follow these instructions to upgrade an existing {{% tts %}} instance.
+
+Replace the tag in your **docker-compose.yml** file with the newer version of {{% tts %}}:
+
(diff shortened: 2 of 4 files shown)


doc/content/guides/getting-started/upgrading.md
```diff
@@ -0,0 +1,62 @@
+---
+title: "Upgrading"
+description: ""
```
johanstokking, line 3: I think we do need a weight here for order. The getting started is one of those where it makes sense

doc/content/getting-started/upgrading.md
```diff
@@ -0,0 +1,65 @@
+---
+title: "Upgrading"
+description: ""
+weight: -1
```
johanstokking, line 4: Isn't this putting the upgrading section before anything else?

benolayinka: Installation also has weight -1, so it's alphabetical between those and then everything else after

Issue #2244, opened 2020-03-26
Network Server schedules downlink tasks when not necessary


rvolosatovs:
#### Summary

Network Server schedules downlink tasks when not necessary

#### Steps to Reproduce

1. Run the stack
2. Create devices
3. Observe downlink tasks added (sometimes, depends on device config)

#### What do you see now?

Downlink tasks added when not necessary

#### What do you want to see instead?

Downlink tasks added only when necessary

#### How do you propose to implement this?

Improve `nextDataDownlinkAt` testing, find the issue and fix it 

#### Can you do this yourself and submit a Pull Request?

yes

PR #2418, opened 2020-04-24
Downlink timing

### LINKED ISSUE #1637: DevStatusReq is sent too often
#### Summary

NS sends `DevStatusReq` too often, draining gateway downlink capacity.

cc @ama9910

#### Steps to Reproduce

1. Activate device
2. Observe gateway downlink traffic over time

#### What do you see now?

`DevStatusReq` every 20 frames or every hour

Introduced here: https://github.com/TheThingsNetwork/lorawan-stack/commit/d3e6c340fdcb679b94ba60ac85f76b77978237a5#diff-fa23c3c73eb9a1b751ae8903f80ac916R30-R33

Unless otherwise specified via per-device settings or per-Network Server config, I'd suggest increasing this significantly, like every 24 hours and _from_ 20 frames.

That from N frames would mean not forcing `DevStatusReq` every Nth frame, but piggybacking on any downlink that's being scheduled.

1. Change default settings
2. Change how the count is interpreted

Can review

### LINKED ISSUE #2078: Class A downlink timing
#### Summary

This issue describes the technical decision we took for class A downlink timing, which has impact on class B and timed class C downlink as well.

TODO: summary

#### Situation

End devices open the RX1 window for class A downlink `RX1Delay` time [1..15] seconds after transmitting an uplink message. This time can be used for the network to make a roundtrip:

```
ED -> Infr (up) -> Net (up) [-> App (up) -> App (down)] -> Net (down) -> Infr (down) -> ED
```

- Infr(astructure) (up/down): `GW <-> GS [<-> PBA <-> PB <-> PBA]`. Packet Broker part is optional if the device is in the cluster where the gateway is connected
- Net(work) (up/down): `NS`, downlink via downlink task
- App(lication) (up/down): `AS <-> Integration` via rpcs, streams and downlink queue operations. This step is optional as not all class A downlink are data messages

The delays between all these steps are unknown.

Since #1611, NS will never make RX1 if `RX1Delay` is 1 second.

With class B downlink and class C downlink with absolute time, only the downstream part can be considered.

#### Scope

All timed downlink: class A downlink in RX1 and RX2, as well as class B downlink and class C downlink with absolute time. 

#### Criteria

1. Easy of migration from V2
2. Full LoRaWAN support
3. Low battery consumption
4. Distribute available time over the unknowns
5. Easy management

#### Alternatives

0. No change

---

1. Timing settings for NS instance, devices and/or downlink paths

This allows for fine-tuning timing per NS instance, per device and/or per downlink path. It allows for configuring NS instances and devices for specific scenarios, and extends APIs for GS and intermediate infrastructure (Packet Broker) to add meta data that help NS to drive timing decisions

---

2. Simple time division over the unknowns with best effort for making RX1

This is my original idea; divide `RX1Delay` over three components: infrastructure, network and application:

```
rx1Delay = infrastructureDelay + networkDelay + applicationDelay
infrastructureDelay = 1s
networkDelay = (rx1Delay - infrastructureDelay) / 2
applicationDelay = networkDelay = (rx1Delay - infrastructureDelay) / 2
```

This results in even division of time to the network-layer and the application-layer with increasing `RX1Delay`

The delays are two-sided for a class A RX1 roundtrip: in class B and timed class C, divide all delays in 2.

`infrastructureDelay` includes any latencies between the gateway and NS, as well as NS deduplication and downlink scheduling. So, in the ideal world where the latency between the gateway and NS is 0 ms, the deduplication window is 150 ms, the downlink scheduling follows 0 ms after the uplink, the downlink message arrives 850 ms early on the gateway.

`networkDelay` is the time NS schedules the downlink task in advance. This gives time for the application-layer to have scheduled a downlink message, before anything gets scheduled. `networkDelay` includes all time NS needs, including downlink task scheduling and deduplication. If `networkDelay` is 0 (i.e. `RX1Delay` is 1), NS yields all time to the infrastructure and schedules the downlink task immediately.

`applicationDelay` is the time the AS (and upstream) has to respond to a data uplink message.

#### Consideration

The consideration is each alternative per criteria:

| Criteria | Alternative 0 | Alternative 1 | Alternative 2 |
--- | --- | --- | ---
Easy of migration from V2 | no pass | pass | pass
Full LoRaWAN support | no pass; always breaks RX1 with RX1Delay of 1 second | pass | pass
Low battery consumption | rely on RX2 is costly and may require retries when the gateway's RX2 is exhausted | pass | pass
Distribute available time over the unknowns | pass | pass | pass, although we may need to make the 50/50% split configurable in the future
Easy management | pass | no pass; hard to maintain, hard to measure delays with non-synced system clocks, hard to document the trade-offs, hard to change the values for high volumes of devices | pass

#### Decision

Alternative 2, which is already pretty much implemented. This issue serves therefore more as documentation.

It should be pretty close to what we have. I don't know exactly what the current implementation is, but it seems to be taking NS's `ReceivedAt` into account in the downlink task execution:

> @rvolosatovs: NS will not attempt Rx1 after `ReceivedAt+(Rx1Delay-infrastructureDelay)/2`

If `RX1Delay` is 1, RX1 will never be tried. That's not in the spirit of alternative 2.

Since upstream and downstream are decoupled, indeed, downstream makes an independent decision whether it is still worth trying RX1. This should only be based on NS's `ReceivedAt` and the `RX1Delay`.

Since the infrastructure delay is a sunk cost, NS should yield it to the infrastructure, and not try to make assumptions. It doesn't, can't and shouldn't know how `ReceivedAt` relates to the end device's transmission time. Therefore, NS can only safely consider downlink in RX1 "too late" when the RX1 delay has passed since `ReceivedAt`, regardless of the infrastructure, network and application delays described above. Those only apply to downlink task scheduling.

Since the delays are two-sided, for scheduling class B and timed class C downlink tasks, consider half or the full infrastructure delay.

The reason `TxRequest` contains an RX1 _and_ RX2 option, is for the GS – closest to the gateway – to make the final decision. This is already implemented in the GS scheduler. NS shouldn't start bothering about this too because that's infeasible because of the many unknown latencies.

rvolosatovs:
#### Summary

Closes #2078 
Closes #1637 
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2422
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2244

#### Changes

- Network Server will attempt RX1 for devices with `Rx1Delay` of 1 second, if possible. In terms of implementation, it means that class A downlink tasks are scheduled ASAP after an uplink, if required.
- Network Server will not attempt to schedule MAC-only frames in ping slots or RXC windows.
- Network Server will only attempt to schedule in a ping slot or RXC window after RX2 has passed.
- Network Server will schedule all time-bound network-initiated downlinks at most RX1 delay ahead of time.

In summary with technical details, that's the new behavior per-downlink opportunity:
- Absolute time downlink - schedule at most Rx1Delay ahead, as long as the time is not before `now()`.
- Class A downlink - schedule ASAP, as long as RX2 is not before `now()` (RX1 is either attempted or not on similar condition)
- Class B ping slot - schedule at most Rx1Delay ahead. Do not attempt ping slots, which are sooner than Rx1Delay/2.
- Class C RXC - schedule when it becomes available (if confirmed downlink is to be sent)

More low-level:
- Network Server is now much smarter when determining whether a downlink is required or not and takes into account e.g. the paths and ADRAckReq/confirmed uplink (at all times).
- `processDownlinkTask` now only tries one downlink slot at a time (based on what `nextDataDownlinkSlot` says). It improves correctness, performance, fail-safety and simplifies testing.

#### Notes for reviewers

~@johanstokking I mark this PR as closing #1637 due to this change:~
~>  Avoid scheduling class B/C downlink tasks for MAC-only frames~

~That means that NS may still send an `FPort==0` downlink with single `DevStatusReq` at RX1/RX2, but that's the only possibility and IMO correct behavior - the device must open the RX1/RX2 after uplink, so we might as well use these. That also better represents the behavior the user would expect. Note, that since https://github.com/TheThingsNetwork/lorawan-stack/pull/2237 was merged, gateway downlink capacity should not be a problem. Please let me know if you agree/disagree.~

#### Checklist


johanstokking:
> That means that NS may still send an `FPort==0` downlink with single `DevStatusReq` at RX1/RX2, but that's the only possibility and IMO correct behavior - the device must open the RX1/RX2 after uplink, so we might as well use these

Yes I agree

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed login page of the Console (now redirects straight to the OAuth login).
 - Network Server now records `LinkADRReq` rejections and will not retry rejected values.
 - Improved `NewChannelReq`, `DLChannelReq` and `LinkADRReq` efficiency.
+- Improved downlink scheduling efficiency.
```
johanstokking, line 25: Please elaborate this a bit

rvolosatovs:
> What is left for final state of this PR, besides CI pass?

At this point the only thing left is fixing the tests.
Before that it was the need to introduce different handling of downlink slots based on type (i.e. class A/absolute time/continuous available at time `t`)

PR #2432, opened 2020-04-29
Make submit buttons enabled by default

### LINKED ISSUE #2389: Avoid showing form submit buttons as disabled initially
#### Summary
We currently show the submit buttons of our forms as disabled by default. This is [bad UX](https://ux.stackexchange.com/a/76306/80506) and it also introduces some side effects, such as [not being able to submit a form as it is](https://github.com/TheThingsNetwork/lorawan-stack/pull/2338#pullrequestreview-397233753) (if the validation rules allow it).

#### What do you see now?
Submit button being disabled until any input has been made

Submit button being enabled by default and showing relevant validation messages after submitting, if necessary.

#### Environment
console v.3.7.1

Trivial

Yes.

kschiffer:
#### Summary
Closes #2389 

Also fixes a validation issue in the webhook template form I noticed alongside.

#### Changes
- Remove `dirty` and `isValid` check from submit button
- Fix missing entries in `initialValues` of the webhook template form
- Fix missing `yup.passValues` in webhook template form

#### Notes for Reviewers
It's still possible to render the submit button as disabled via the `disabled` prop, or when the form is disabled as a whole. But the default should be enabled now always.

#### Checklist

### DIFF
--- a/pkg/webui/components/submit-button/index.js
+++ b/pkg/webui/components/submit-button/index.js
@@ -20,7 +20,6 @@ import PropTypes from '@ttn-lw/lib/prop-types'
 
 class SubmitButton extends React.PureComponent {
   static propTypes = {
-    dirty: PropTypes.bool.isRequired,
     disabled: PropTypes.bool,
     icon: PropTypes.string,
     isSubmitting: PropTypes.bool.isRequired,
@@ -33,9 +32,8 @@ class SubmitButton extends React.PureComponent {
     icon: undefined,
   }
   render() {
-    const { message, icon, isSubmitting, isValidating, disabled, dirty } = this.props
+    const { message, icon, disabled, isSubmitting, isValidating } = this.props
 
-    const buttonDisabled = disabled || isSubmitting || !dirty
     const buttonLoading = isSubmitting || isValidating
 
     return (
@@ -43,7 +41,7 @@ class SubmitButton extends React.PureComponent {
         type="submit"
         icon={icon}
         message={message}
-        disabled={buttonDisabled}
+        disabled={disabled}
         busy={buttonLoading}
       />
     )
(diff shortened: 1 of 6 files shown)


pkg/webui/components/submit-button/index.js
```diff
@@ -33,17 +32,16 @@ class SubmitButton extends React.PureComponent {
     icon: undefined,
   }
   render() {
-    const { message, icon, isSubmitting, isValidating, disabled, dirty } = this.props
+    const { message, icon, disabled, isSubmitting, isValidating } = this.props
 
-    const buttonDisabled = disabled || isSubmitting || !dirty
```
bafonins, line 38:
```js
const buttonDisabled = disabled || buttonLoading
```

We do want to disable the submit button when the form is submitting/validating and the button is in the loading state, otherwise the user can still send requests which can result in weird behavior, especially with slow internet connection. Also, I dont see any reasons why should we allow the submit button to be clickable when the form is submitting.

kschiffer: The `<Button />` is de-facto disabled, when it is set to `busy`:
https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/pkg/webui/components/button/index.js#L73-L75

bafonins: The `<Button />` is still clickable and since it has `type="submit"` the button submits the form data to the server on each click. Please try it yourself.

Throttle your network connection in the dev tools and try submitting the form, say application create` several times while the `<Button />` is in the `busy` state.

kschiffer: Fixed this. The busy state will now also make the button disabled via the `disabled` prop and also disregard any click events that might still come through. I've also fixed a related issue in the oauth login form, where the button would be clickable again right before the redirect.

PR #2450, opened 2020-04-29
Improve byte input UX

### LINKED ISSUE #2404: Improve UX for ByteInputs without fixed value length
#### Summary
Our `ByteInput` component is not really working well for getting byte values without length restriction.

#### Why do we need this?
We need to get byte values without length restriction in the schedule downlink secrtion.

ByteInput component enabling inputs without length restriction.

A better behavior of such input: don't show the masked input dots.

#### Environment
v3.7.1

I think it's best to refactor the byte input into its own component. We currently have it as variation of the normal `<Input />` component, which makes it hard to maintain and extend and it also arbitrarily mixes prop types between the two.

@pgalic96

pgalic96:
#### Summary

Closes #2404 

#### Changes

- Fix the issue with showing placeholder chars when input is unbounded
- Fix the issue with pasting values when input is unbounded

#### Notes for Reviewers

cc @kschiffer . I have looked into moving `ByteInput` to a separate component, but I am not sure this is the best idea. It would include a lot of replication since there is still a lot of similarities with the `Input` component. I have managed to fix this issue by combination of `showMask` and `guide` flags in the input. I think then most of the issues with this input are solved.

However, if you still prefer moving the `ByteInput` to a separate component, we can discuss this.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+<<<<<<< HEAD
 - OAuth authorization page crashing.
+=======
+- Byte input in scheduling downlink view
+>>>>>>> doc: Update CHANGELOG
 
 ### Security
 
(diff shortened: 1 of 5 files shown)


pgalic96:
> The byte input should not allow incomplete inputs, e.g. 0FF, should always be a round number of characters

Fixed! As per our discussion it's better to move the refactoring byte input to a new goldplating issue.

PR #2458, opened 2020-04-29
CLI create user config dir if missing

### LINKED ISSUE #2355: Snap generating config file fails because of missing directory.

neoaggelos:
#### Summary

Closes #2355 

#### Changes

- Create user config directory if it does not exist, so that the config file generation does not fail.

#### Notes for Reviewers

Do we need a flag for disable this? e.g. `--no-make-dirs`

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/use.go
+++ b/cmd/ttn-lw-cli/commands/use.go
@@ -78,6 +78,9 @@ var (
 					if err != nil {
 						return "", err
 					}
+					if err = os.MkdirAll(dir, 0755); err != nil {
+						return "", err
+					}
 					fileName = filepath.Join(dir, base)
 				}
 				_, err := os.Stat(fileName)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Upgrading guide in docs
 - Unset end device fields using the CLI (see `--unset` option)
 - Join EUI and Dev EUI columns to the end device table in the Console.
+- CLI creates user configuration directory if it does not exist when generating configuration file.
 
 ### Changed
 


cmd/ttn-lw-cli/commands/use.go
```diff
@@ -78,6 +78,12 @@ var (
 					if err != nil {
 						return "", err
 					}
+					if _, err := os.Stat(dir); os.IsNotExist(err) {
+						if err = os.MkdirAll(dir, 0775); err != nil {
```
htdvisser, line 82: Isn't 775 a bit too open? 755 would be sufficient, right?

cmd/ttn-lw-cli/commands/use.go
```diff
@@ -78,6 +78,12 @@ var (
...
+							return "", err
+						}
+						logger.Infof("Created missing directory %s", dir)
+					}
```
rvolosatovs, line 86: Let's just keep it simple:
```suggestion
					if err = os.MkdirAll(dir, 0775); err != nil {
						return "", err
					}
```

From https://golang.org/pkg/os/#MkdirAll
> If path is already a directory, MkdirAll does nothing and returns nil.

Issue #2310, opened 2020-04-06
Allow device payload formatters to be unspecified


adriansmares:
#### Summary

Allow device payload formatters to be unspecified (null).

#### Why do we need this?

In order to allow users to delegate the payload formatters to the application default formatter.

#### What is already there? What do you see now?

For both devices and applications, `None` can be selected. But this `None` is not always what a user would want. When the device its first created, the device payload formatter field is `null`. This means that the AS will skip it and try to look up the default payload formatter, which can be found in the link.  If the user now configures in the console a payload formatter for the device (doesn't matter which one), he can no longer go back to this state in which the payload formatter of the device is the default one of the application. 

I believe that AS should also allow per-device overrides, so if a device is individually set to `None` this shouldn't mean that the application wide payload formatter should be used.

This is important since if you are using something like a JavaScript payload formatter for an application with multiple devices, you're out of luck if you accidentally set the payload formatter for one of them (you'll have to duplicate the code until this issue is fixed).

#### What is missing? What do you want to see?

An option such as `Application` or `Default` that would correspond to a null device payload formatter. We would probably need either a better name or a small holdover bubble that describes that this option uses the application-wide payload formatter.

#### How do you propose to implement this?

Add the above described payload formatter option.

#### Can you do this yourself and submit a Pull Request?

I can review. Feel free to redistribute this issue.

PR #2462, opened 2020-04-30
Allow resetting end device payload formatters in the Console

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2310
Screenshot 2020-04-30 at 16 54 46

#### Changes

- Allow settings empty object when merging device responses
- Add the `default` option to device payload formatters

#### Checklist


pkg/webui/console/components/payload-formatters-form/index.js
```diff
@@ -187,6 +187,7 @@ class PayloadFormattersForm extends React.Component {
             component={Radio.Group}
             onChange={this.onTypeChange}
           >
+            {allowReset && <Radio label={sharedMessages.default} value={TYPES.DEFAULT} />}
```
kschiffer, line 190: I think we should use another label, like `Use application payload formatter`, otherwise it's still ambiguous.
Then we can also remove the info notification and update the application level info notification to `Application level payload formatters are the default for all end devices. End device level payload formatters can be used to overwrite this default individually.`.

pkg/webui/console/containers/device-payload-formatters/downlink.js
```diff
@@ -65,22 +65,37 @@ class DevicePayloadFormatters extends React.PureComponent {
...
+
   async onSubmit(values) {
-    const { appId, devId, formatters, updateDevice } = this.props
+    const { appId, devId, formatters: initialFormatters, updateDevice } = this.props
+
+    if (values.type === PAYLOAD_FORMATTER_TYPES.DEFAULT) {
+      return updateDevice(appId, devId, {
+        formatters: null,
+      })
+    }
```
kschiffer, line 83: This is not right. That would also erase the other (in this case uplink) payload formatter.

pkg/webui/console/containers/device-payload-formatters/uplink.js
```diff
@@ -65,23 +65,36 @@ class DevicePayloadFormatters extends React.PureComponent {
...
+
   async onSubmit(values) {
-    const { appId, devId, formatters, updateDevice } = this.props
+    const { appId, devId, formatters: initialFormatters, updateDevice } = this.props
+
+    if (values.type === PAYLOAD_FORMATTER_TYPES.DEFAULT) {
+      return updateDevice(appId, devId, {
+        formatters: null,
+      })
+    }
```
kschiffer, line 83: Idem.

pkg/webui/console/containers/device-payload-formatters/downlink.js
bafonins: This is intentional. You cannot set a particular formatter to null.

pkg/webui/console/containers/device-payload-formatters/uplink.js
bafonins: This is intentional. You cannot set a particular formatter to null.

pkg/webui/console/components/payload-formatters-form/index.js
bafonins: Changed `Default` to `Use application payload formatter`. For the rest we can do that in a separate PR. I created an issue to improve the payload formatters pages in general, see https://github.com/TheThingsNetwork/lorawan-stack/issues/2463

pkg/webui/console/containers/device-payload-formatters/downlink.js
kschiffer: I see. Hmm, then I'm afraid we need to apply a different UI logic to make this constraint clear for the user. Essentially, instead of using an option in the radio button group, we need a global switch. Something like this:
image

And maybe, in case of using the default, you can also add a link to the application payload formatters.

bafonins: I agree that such ui is definitely better than the current one, however in order to implement this, quite many changes to our code base are required. Such refactoring can cause regression to functionality that already works well.

Currently, the issue is that the users cannot reset device level payload formatters via the Console. This PR adds this option.

I suggest we discuss possible improvements to the payload formatters pages (both app and device level) in https://github.com/TheThingsNetwork/lorawan-stack/issues/2463 and merge this fix without major refactoring so that in the next release the users have this functionality in place.

kschiffer: I see your point, but we're now solving one problem while deliberately introducing another. If we merge it like this, we're knowingly putting users at risk of losing their payload formatter settings. We should at least add some information that setting the formatter to default will erase the formatters altogether.

kschiffer: A simple field warning could do the trick here.

johanstokking:
What is the status here?

PR #2466, opened 2020-04-30
Disable ADR for ISM2400

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2043

#### Changes

- Disable ADR for ISM2400

#### Checklist

### DIFF
--- a/pkg/networkserver/utils.go
+++ b/pkg/networkserver/utils.go
@@ -62,7 +62,10 @@ func timePtr(t time.Time) *time.Time {
 	return &t
 }
 
-func deviceUseADR(dev *ttnpb.EndDevice, defaults ttnpb.MACSettings) bool {
+func deviceUseADR(dev *ttnpb.EndDevice, defaults ttnpb.MACSettings, phy band.Band) bool {
+	if !phy.EnableADR {
+		return false
+	}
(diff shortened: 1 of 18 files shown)


pkg/networkserver/utils.go
```diff
@@ -62,7 +62,10 @@ func timePtr(t time.Time) *time.Time {
 	return &t
 }
 
-func deviceUseADR(dev *ttnpb.EndDevice, defaults ttnpb.MACSettings) bool {
+func deviceUseADR(dev *ttnpb.EndDevice, defaults ttnpb.MACSettings, phy band.Band) bool {
+	if phy.ID == band.ISM_2400 {
+		return false
+	}
```
johanstokking, line 68: Please make this a band field; this shouldn't be hidden in NS utilities

rvolosatovs: I was thinking to do the same originally, but it's not that ISM_2400 does not support ADR, it's that Network Server does not know how to do it for this band. So by adding a `EnableADR: false` to `ISM_2400` we actually pollute `band` package with Network Server-specific information.

johanstokking: No it's the 2.4 GHz band has too much variation in noise that ADR doesn't make sense

PR #2479, opened 2020-05-05
Decode LoRaWAN Payload in Gateway Server before pushing message

### LINKED ISSUE #2442: Decode PHYPayload before forwarding uplink message
#### Summary

References #1962 

Decode PHYPayload before forwarding uplink message

#### Why do we need this?

For event subscribers to see LoRaWAN identifiers, i.e. EUIs in join-requests and DevAddr in uplink messages.

Basic Station has the `(*UplinkMessage).Payload` set, because it reconstructs the `RawPayload`. This is obviously not the case on other frontends (UDP, MQTT), so `Payload` is unset.

Make sure that `Payload` is set.

- Check if `Payload` is set. If it's set, fine, done
- _Try_ to decode the LoRaWAN message. That may fail. If it does, log a debug level message and still forward the message. The reason for this is that GS is not responsible for LoRaWAN handling; the `Payload` is informative only; NS will always try to decode the payload at all times

Can review

neoaggelos:
#### Summary

Closes #2442

#### Notes for Reviewers

I get a feeling that we do not get a lot out of having both `GetUplinkMessageIdentifiersFromMessage` and `GetUplinkMessageIdentifiers`.

Targetting `develop`, just because we are changing the current behaviour (we are only adding things though)

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -607,6 +607,14 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
 		case msg := <-conn.Up():
 			ctx = events.ContextWithCorrelationID(ctx, fmt.Sprintf("gs:uplink:%s", events.NewCorrelationID()))
 			msg.CorrelationIDs = append(msg.CorrelationIDs, events.CorrelationIDsFromContext(ctx)...)
+			if msg.Payload == nil {
+				pld := &ttnpb.Message{}
+				if err := lorawan.UnmarshalMessage(msg.RawPayload, pld); err != nil {
+					log.FromContext(ctx).WithError(err).Debug("Failed to decode message payload")
+				} else {
+					msg.Payload = pld
+				}
+			}
 			val = msg
 			registerReceiveUplink(ctx, gtw, msg.UplinkMessage, protocol)
 		case msg := <-conn.Status():
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Gateway Server events for uplink messages now contain end device identifiers.
+
 ### Changed
 
 ### Deprecated


johanstokking:
> Targetting `develop`, just because we are changing the current behaviour (we are only adding things though)

This can be to `master` as it's not affecting our compatibility commitment

pkg/encoding/lorawan/messages.go
```diff
@@ -639,3 +639,30 @@ func GetUplinkMessageIdentifiers(phyPayload []byte) (ttnpb.EndDeviceIdentifiers,
...
+
+// GetUplinkMessageIdentifiersFromMessage retrieves the EndDeviceIdentifers from a LoRaWAN message (except DeviceID).
+func GetUplinkMessageIdentifiersFromMessage(msg *ttnpb.Message) (ttnpb.EndDeviceIdentifiers, error) {
+	var ids ttnpb.EndDeviceIdentifiers
+	if msg == nil {
+		return ttnpb.EndDeviceIdentifiers{}, errInvalidPayload.New()
+	}
+	switch msg.MHDR.MType {
+	case ttnpb.MType_CONFIRMED_UP, ttnpb.MType_UNCONFIRMED_UP:
+		payload := msg.GetMACPayload()
```
johanstokking, line 651: Is it guaranteed that `payload != nil` if the `MType` is set?

johanstokking: I would suggest something like;

```go
if macPayload := msg.GetMACPayload(); payload != nil {
   ...
} else if ...; ... != nil {
   ...
}
```

pkg/encoding/lorawan/messages.go
```diff
@@ -639,3 +639,30 @@ func GetUplinkMessageIdentifiers(phyPayload []byte) (ttnpb.EndDeviceIdentifiers,
...
+		ids.DevAddr = &types.DevAddr{}
+		(&(payload.DevAddr)).Copy(ids.DevAddr)
```
johanstokking, line 653: What is this `(&(payload.DevAddr))` for?

neoaggelos: Copy is only available for (*types.DevAddr). Same below. This is not too clean. Any suggested alternatives? I suppose keeping the old behavior here (retrieve ids from the RawPayload) would be easier and a bit cleaner?

neoaggelos: To be honest, this solves many troubles with zero vs nil identifiers as well.

pkg/encoding/lorawan/messages.go
neoaggelos: Changed to only access the payload if it is not nil. Is the mtype check useful?

johanstokking: No

pkg/encoding/lorawan/messages.go
johanstokking: Probably yes, leave the old behavior as-is

neoaggelos: Reverted

pkg/gatewayserver/gatewayserver.go
```diff
@@ -607,6 +607,13 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
 		case msg := <-conn.Up():
 			ctx = events.ContextWithCorrelationID(ctx, fmt.Sprintf("gs:uplink:%s", events.NewCorrelationID()))
 			msg.CorrelationIDs = append(msg.CorrelationIDs, events.CorrelationIDsFromContext(ctx)...)
+			if msg.Payload == nil {
+				msg.Payload = &ttnpb.Message{}
+				if err := lorawan.UnmarshalMessage(msg.RawPayload, msg.Payload); err != nil {
+					msg.Payload = nil
+					log.FromContext(ctx).Debug("Failed to decode message payload")
+				}
```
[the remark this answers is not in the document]
johanstokking: Mostly missing the `WithError()`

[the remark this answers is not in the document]
neoaggelos: Updated, also rebased with new develop

PR #2493, opened 2020-05-07
Make go.thethings.network/lorawan-stack/v3/pkg/ttnpb a module

johanstokking:
#### Summary

Make `go.thethings.network/lorawan-stack/v3/pkg/ttnpb` a Go module

#### Changes

- Add `go.{mod,sum}`

#### Notes for Reviewers

This is draft because we need to decide:

1. Do we want this?
2. How do we keep the replacements in sync with the main `go.mod`?
3. Does this even work correctly?

#### Checklist

>TODO


rvolosatovs:
I don't think we need the replacements for this one at all, in fact, the `go.{mod,sum}` files should be quite minimal for this module.

pkg/ttnpb/go.mod
```diff
@@ -0,0 +1,46 @@
+module go.thethings.network/lorawan-stack/v3/pkg/ttnpb
+
+go 1.14
+
+replace go.thethings.network/lorawan-stack/v3 => ../../
```
rvolosatovs, line 5: This seems to be the only replacement we need here.
Not sure if this works as expected when imported by a third party though. I guess it should.

rvolosatovs:
> Do we want this?

The only benefit of this approach over just using the parent module would be minimizing the dependency size for third-party importers. Then again if `go mod` is smart it would not even need to download packages it does not need.
In my opinion, we do not need this now and I would prefer to do this when we need it, i.e. when the import size downstream is a problem. This switch would also not break clients, since `go mod` would just handle all of that automatically on update. I don't think the added maintenance cost is justified at the moment.

johanstokking:
> Also, note that perhaps a cleaner solution could be introducing a module for our `./cmd`, where goreleaser would be a dependency, similar to how go team manages `gopls` (https://github.com/golang/tools/tree/master/gopls). I believe this way we would drastically reduce the amount of replacements needed downstream.

I'll try this.

But which replacements wouldn't be needed anymore?

johanstokking:
Replaced by #2495

Issue #977, opened 2019-07-10
Valid web paths return 404


adriansmares:
#### Summary

#919 introduced a regression related to web paths - it seems that one of the dependency updates broke down the route handling. 
@kschiffer observed this while attempting to `GET` an application link (`GET http://localhost:1885/api/v3/as/applications/app1/link?field_mask=api_key,network_server_address`)

#### Steps to Reproduce

1. Create an application `app1`
2. Set the application link for `app1` through the CLI.
3. Attempt to `GET http://localhost:1885/api/v3/as/applications/app1/link?field_mask=api_key,network_server_address`

#### What do you see now?

```
"error:pkg/errors/web:http (HTTP error: code=404, message=Not Found)"
```

#### What do you want to see instead?

```
{
  "code": 7,
  "message": "error:pkg/auth/rights:no_application_rights (no rights for application `app1`)",
  "details": [
    {
      "@type": "type.googleapis.com/ttn.lorawan.v3.ErrorDetails",
      "namespace": "pkg/auth/rights",
      "name": "no_application_rights",
      "message_format": "no rights for application `{uid}`",
      "attributes": {
        "uid": "app1"
      },
      "correlation_id": "63d6d3b76c4a444ba939fc6b7e5da015"
    }
  ]
}
```

#### Environment

On d6f3b8df290d6741bdd1a81da321698cb41f9fab (the commit that merged #919) the resulting output is the `404`. On 6fe2c82241ad2cfc7a610c8b5ef9071d4a3fcef1 (the `master` before the merge) the result is the `no_application_rights`.

#### How do you propose to implement this?

No idea yet. But I believe it is related to https://github.com/TheThingsNetwork/lorawan-stack/pull/919/commits/36cb90327ede646f913d8286a5e1338f649b3a14 , since that test shouldn't have became broken during the update.

#### Can you do this yourself and submit a Pull Request?

I'll continue to investigate which package introduced this regression and report it here.

htdvisser:
The tests were actually wrong before that commit. Creating an Echo group is and should be a no-op until you actually register routes/handlers to it.

adriansmares:
I'm not really sure it is a real no-op:
https://github.com/labstack/echo/blob/master/echo.go#L531-L535
https://github.com/labstack/echo/blob/master/group.go#L21-L30
If the group contains has a default middleware, a default route to the not found handler is created when you actually create the group.

adriansmares:
[...]

PR #2495, opened 2020-05-07
Make go.thethings.network/lorawan-stack/v3/cmd a module

johanstokking:
#### Summary

Replaces #2493 

Make `go.thethings.network/lorawan-stack/v3/cmd` a module

#### Changes

- Move `tools.go` into new `tools` mod
- Make `cmd` a module
- Build development tooling to `.bin` instead of `go run` which would use the repo's module

#### Notes for Reviewers

1. [ ] Check need for the replacements in `/go.mod` and `/cmd/go.mod`
2. [x] Make sure that we can use the tools in `/cmd/tools.go`. I'm also open to a `/tools` module
3. [x] Check if we need to merge some changes from github.com/robertkrimen/otto in github.com/TheThingsIndustries/otto 
4. [ ] Figure out how we can update `.bin` binaries if modules update in `tools/go.mod`; maybe always install on `make init`?

#### Checklist


johanstokking:
Blocked by #2496

johanstokking:
In order to use versioned tools, I would suggest;

- put `tools.go` in `cmd`, or move to `tools/main.go` and init new module
- install the tools via `go install` to a shared repo-local bin folder
- use the binaries from the bin folder instead of the `go run` construct in mage (via `execGo`)

Example Makefile;

```Makefile
GO = go
GOBIN = $(PWD)/.bin
export GOBIN

.PHONY: deps.dev
deps.dev: .venv
	@pushd cmd && cat tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % $(GO) install % && popd
```

And then `./bin/unconvert` etc

.mage/mage.make
```diff
@@ -12,11 +12,16 @@
...
 
+MAKE ?= make
 MAGE ?= ./mage
+GOBIN = $(PWD)/.bin
+export GOBIN
 
-$(MAGE): magefile.go $(wildcard .mage/*.go)
-	GO111MODULE=on go install github.com/magefile/mage
-	GO111MODULE=on go run github.com/magefile/mage -compile $(MAGE)
+$(GOBIN):
```
johanstokking, line 20: Should we turn this into `$(GOBIN): tools/go.mod` for it to track that file for changes to the tools and versions?

tools/main.go
```diff
@@ -14,7 +14,7 @@
 
 // +build tools
 
-package tools
+package main
```
rvolosatovs, line 17: Why?

rvolosatovs:
The benefit of using `go run` to run commands is the automatic installation/versioning with no `Makefile` etc. maintenance cost. What's the benefit of using compiled binaries? How do you propose to keep them up-to-date?

johanstokking:
> The benefit of using `go run` to run commands is the automatic installation/versioning with no `Makefile` etc. maintenance cost.

Yes, that definitely has benefits. But I do want to keep those modules (and replacements) out of the project root's `go.mod`. Two possible steps forward:

1. `go run` from the `tools` dir and account for that in the paths
2. Watch `tools/go.mod` in the Makefile so that changes trigger reinstallation

rvolosatovs:
I'd vote for `1`. We could move all the mage stuff there as well, in fact, I believe. That way we'd have everything tooling-related with all the dependency hell isolated in `./tools` and a clean top-level module at root. I believe we don't need a `./cmd` module either this way.

johanstokking:
OK, please try if that works.

johanstokking:
> That way we'd have everything tooling-related with all the dependency hell isolated in `./tools` and a clean top-level module at root

Looking at replacements we have in this changeset;

- `gopkg.in/DATA-DOG/go-sqlmock.v1`: that's because of `github.com/heptiolabs/healthcheck` required by `pkg/component`
- `github.com/labstack/echo/v4`: we're on an old version because of https://github.com/TheThingsNetwork/lorawan-stack/issues/977. With #2382 we're closer to getting rid of Echo. @htdvisser can you file the issues to migrate away completely, as announced in that PR?
- `github.com/Azure/go-autorest` is quite challenging, see https://github.com/google/go-cloud/issues/1990 (this is our dependency) and https://github.com/kubernetes/client-go/issues/628
- `github.com/robertkrimen/otto` replaced by our fork: I keep pinging these guys (https://github.com/robertkrimen/otto/pull/283) but no luck there
- `github.com/grpc-ecosystem/grpc-gateway`: is this tooling?

PR #2511, opened 2020-05-11
Fix OAuth token exchange/refresh issues when using TLS

kschiffer:
#### Summary
This quickfix PR fixes issues during the token exchange/refresh of the OAuth flow, caused by the TLS config not being passed to backend-side requests made by the `oauth2` package. This caused the `token exchange refused` error (mentioned in #2353)  in setups using a root CA certificate, since the request could not check the server certificate against the root CA.

#### Changes
- Extend `oauthclient` config to include the TLS client config
- Attaching the decorated HTTP client to the context used by the oauth2 package
- Update console config to pass the TLS config of the component

#### Notes for Reviewers
This is how I reproduced the issue:
1. Configure the stack to use TLS with a root CA
2. Temporarily distrust this root CA using mac keychain (if not already so)
3. Start the stack
4. Trust the certificate using mac keychain (to make the browser trust the webUI, while still leaving the certificate untrusted in the already started stack)
5. Login to the console
6. See the `token exchange refused` error

If you restart the stack then, without untrusting the root CA again, the login will work, since the certificate is now trusted by your root, getting picked up during the stack restart.

After applying this PR, it should work either way, since the root CA is now being checked as well.

Main reviewer: @htdvisser 

#### Checklist

### DIFF
--- a/pkg/web/oauthclient/oauthclient.go
+++ b/pkg/web/oauthclient/oauthclient.go
@@ -17,6 +17,7 @@ package oauthclient
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"net/url"
 	"strings"
 
@@ -49,6 +50,18 @@ func (oc *OAuthClient) getMountPath() string {
 	}
 
 	return path
+}
+
+func (oc *OAuthClient) withTLSClientConfig(ctx context.Context) (context.Context, error) {
+	config, err := oc.component.GetTLSClientConfig(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return context.WithValue(ctx, oauth2.HTTPClient, &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: config,
+		},
+	}), nil
 }
 
 // New returns a new OAuth client instance.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - OAuth authorization page crashing.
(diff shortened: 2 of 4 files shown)


pkg/web/oauthclient/oauthclient.go
```diff
@@ -51,6 +52,16 @@ func (oc *OAuthClient) getMountPath() string {
 	return path
 }
 
+func (oc *OAuthClient) attachTLSClient(c echo.Context) context.Context {
+	transport := &http.Transport{
+		TLSClientConfig: oc.config.TLSClientConfig,
+	}
```
johanstokking, line 58: I would personally do this here;

```go
func (oc *OAuthClient) withTLSClientConfig(ctx context.Context) (context.Context, error) {
  config, err := oc.component.GetTLSClientConfig(ctx)
  if err != nil {
    return nil, err
  }
  return context.WithValue(ctx, oauth2.HTTPClient, &http.Client{
    Transport: &http.Transport{
      TLSClientConfig: config,
    }
  }), nil
}
```

and use as;

```go
ctx, err := oc.withTLSClientConfig(c.Request().Context())
if err != nil {
  ...
}
..., err := oc.oauth(c).TokenSource(ctx, token).Token()
```

Because;
- this decouples echo context from configuration
- it uses the request context for getting TLS client config, which may become useful in the future
- it feels more idiomatic to derive contexts with the with-pattern

kschiffer: Oh indeed! I completely overlooked that the `OAuthClient` had a reference to the component.

johanstokking:
Please update changelog also

PR #2540, opened 2020-05-14
Application packages default associations CLI support

adriansmares:
#### Summary

References #2412 

#### Changes

- Add CLI support for default associations.

#### Notes for Reviewers

Documentation is part of #2539 .

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md`. The target branch is set to `master` if the changes are fully compatible with existing API, database, configuration and CLI.
- [x] Testing: The changes are covered with unit tests. The changes are tested manually as well.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #2539, opened 2020-05-14
Support application wide application packages associations

adriansmares:
#### Summary

Closes #2412
Closes #2473
Closes #2474

#### Changes

- Add application packages default associations. These are the application-wide equivalent of normal application packages associations. They allow users to setup an application package for the whole application instead of an per-device basis. 
- Application packages may now handle _unrelated_ traffic (traffic on a different FPort than the one they've been associated with). This allows application packages to schedule downlinks even when their FPort hasn't been used.
- The LoRaCloud DAS application package now merges application-specific and device-specific state. This means that a normal association (device-level) may be used to override settings imposed by the default association (application-level).
- Add support for custom LoRaCloud DAS server URLs. Can be configured using the package data (and override on a per-device level, should that ever be needed). Old configurations are backwards-compatible.
- LoRaCloud DAS is now informed of new device sessions on `JoinAccept`.
- Some additional improvements to the structure of the API used by LoRaCloud DAS (there have been some changes since the original implementation was introduced).

#### Notes for Reviewers

I've did my best w.r.t naming but I realize that having both `associations` and `default-associations` may be confusing. Any suggestions are more than welcome.

@rvolosatovs, @KrishnaIyer please check the API changes mainly, and do a sanity check only if time allows you to do so.

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md`. The target branch is set to `master` if the changes are fully compatible with existing API, database, configuration and CLI.
- [x] Testing: The changes are covered with unit tests. The changes are tested manually as well.
- [x] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #2541, opened 2020-05-14
Application Server service data uplink messages support

### LINKED ISSUE #2413: LoRa Cloud DAS application package answers should be provided to users
#### Summary
LoRaCloud DAS is returning useful values to the user like device state, streaming or file data. LoRaWAN-stack should provide a way to forward those data to the user, ideally embedded with the uplink payload.

#### Why do we need this?

The LoRaCloud DAS application package is only half useful if users do not have DAS answer's provided by LoRaWAN-stack, they will have to poll the DAS for streaming or file upload content

No

adriansmares:
#### Summary

Closes #2413

#### Changes

- Add support for a new `ApplicationUplink` message: `ServiceData`. It allows services such as application packages to provide a constant stream of data to its users.
- Provide LoRaCloud DAS answers as service data.

#### Notes for Reviewers

The conversion to `types.Struct` is rather inefficient, using serialization/de serialization operations. I've attempted to use the functions in the `pkg/gogogproto` but they don't seem to handle structs from other packages nicely. I suggest that we create an issue with label `scalability` and revisit this later unless someone can propose a better solution now. 

A follow up issue for the Console support for this new message field will follow later.

@Oliv4945 you may checkout this branch and test it (it contains all of the changes of #2539 and #2540 as well).

#### Checklist


pkg/applicationserver/io/packages/loradms/v1/api/uplinks_test.go
```diff
@@ -108,3 +108,7 @@ func toUint32(x uint32) *uint32 {
 func toFloat64(x float64) *float64 {
 	return &x
 }
+
+func toHex(x objects.Hex) *objects.Hex {
```
johanstokking, line 112: see previous

pkg/applicationserver/io/packages/loradms/v1/package.go
```diff
@@ -123,7 +130,30 @@ func (p *DeviceManagementPackage) sendUplink(ctx context.Context, up *ttnpb.Appl
...
 	}
 
-	downlink := response.Result.Downlink
+	result := response.Result
+	resultStruct, err := toStruct(&result)
+	if err != nil {
+		return err
+	}
+
+	ctx = events.ContextWithCorrelationID(ctx, append(up.CorrelationIDs, fmt.Sprintf("as:integration:loradms:%s", events.NewCorrelationID()))...)
```
johanstokking, line 139: `as:packages` maybe

pkg/applicationserver/io/packages/loradms/v1/api/uplinks_test.go
```diff
@@ -109,6 +109,6 @@ func toFloat64(x float64) *float64 {
 	return &x
 }
 
-func toHex(x objects.Hex) *objects.Hex {
+func toHexPtr(x objects.Hex) *objects.Hex {
```
johanstokking, line 112: nit; it would be `hexPtr()`

You don't need to change this, but it would be consistent with the codebase

johanstokking: `$ grep -R -E 'func (?:[a-zA-Z0-9]+)Ptr' pkg`

adriansmares: No, I'll change it for consistency. I've misread the suggestion in the other PR as 'add Ptr', that's on me.

PR #2547, opened 2020-05-14
Github Action to add labels to Pull Requests

htdvisser:
#### Summary

While reviewing #2539 and #2541 I noticed that the pull requests weren't labeled with the compat labels for API and DB changes. I understand it's easy to forget, so this pull request adds configuration for Github's pull request labeler action.

#### Notes for Reviewers

Unfortunately it won't work for pull requests from forks, but considering that most of our pull requests come from branches of this repo, I still think it adds value.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,71 @@
+c/identity server:
+- pkg/identityserver/*
+- pkg/identityserver/**/*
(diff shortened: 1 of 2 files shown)


KrishnaIyer:
This is awesome stuff.

PR #2549, opened 2020-05-15
Allow unsetting NwkKey

johanstokking:
#### Summary

Allow unsetting NwkKey

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2110

#### Changes

- Allow setting empty `nwk_key` in Join Server. This partly reverts ddd8f3f

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Clearing end device events list in the Console.
 - Some views not being accessible in the OAuth app (e.g. update password).
 - `LinkADRReq` scheduling.
+- Unsetting NwkKey in Join Server.
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


rvolosatovs:
I approved it since this fixes the issue, but I don't really understand the use case. If NwkKey was incorrectly set at the first place, why not just delete the device?

johanstokking:
> I approved it since this fixes the issue, but I don't really understand the use case. If NwkKey was incorrectly set at the first place, why not just delete the device?
> "setting empty" is something that should never happen. In that case `nwk_key` should just not be present in the fieldmask on creation.

NwkKey is optional; unset is a valid value.

The field mask says what the client knows about the data. It knows that the NwkKey is empty. In LoRaWAN 1.0.x, NwkKey is allowed to be empty, and JS doesn't know the MAC version so it should accept an empty value.

PR #2564, opened 2020-05-18
Enable authorization via session cookie

### LINKED ISSUE #2528: Change CSRF middleware to gorilla/csrf
#### Summary

In #2382 we started migrating away from Echo, but there are still some parts that need to be changed. This issue focuses on replacing the CSRF middleware.

#### Why do we need this?

1. Because we are dropping Echo (see also https://github.com/TheThingsNetwork/lorawan-stack/pull/2382#issuecomment-621032344)
2. Because the current implementation has some shortcomings (i.e. it doesn't use `SameSite` cookies).

CSRF is used in `pkg/console` and `pkg/oauth`.

We need to replace the existing CORS middleware with `gorilla/cors`.

My proposal would be to make `pkg/web.(*Server)` return a CORS MiddlewareFunc based on the HTTP config (including the cookie keys). This middleware can then be used in `RegisterRoutes`.

In `pkg/console` and `pkg/oauth` we can use `echo.WrapMiddleware()` until we've migrated them away from Echo.

kschiffer:
#### Summary
This PR add authorization for the API via a session cookie and secures actions authorized by the session cookie with a CSRF token.

References #1422 
Closes #2528

#### Changes
- Add a `SessionToken` as auth type
- Extend the `UserSession` model to hold token information
- Add `CookieAuth` middleware that passes the session token stored in the session cookie to the `Authentication` header, as well as adding CSRF enabling in such case
- Extend entity access logic to allow the session token as authorization
- Various related refactors, especially relating to the recent decision to move away from echo
  - Use gorilla/csrf as CSRF library
  - Update / partially port the usage of cookie middleware for vanilla

#### Notes for Reviewers
~~Before I spend time updating and extending unit tests, I would like some input on the implementation.~~ There could be more work done to bring this more in line with the echo/vanilla refactor, but I would much rather do that to not spill the scope much more. There was already quite some plumbing involved in order to share the CSRF configuration between the global web scope and the scope of the OAuth component.

##### Security apsects
- This implementation will give precedence to any other authentication header already present in the request
- When the session token is used for authorization
  - CORS headers are set to not allow requests from other origins.
  - CSRF token needs to be present and valid
- The session cookie currently implies all possible rights but we could also limit the rights to the ones needed for actions of the account app
- We currently do not set expiry dates for sessions; we should probably do that but I'm not sure what a good duration would be

I tested this by extracting the cookies and CSRF headers, obtained after authenticating in the OAuth app (via dev tools) to run API requests in cURL (or postman, which is more convenient).

#### Checklist


pkg/webmiddleware/cookie_auth.go
```diff
@@ -0,0 +1,57 @@
...
+				if err != nil {
+					next.ServeHTTP(w, r)
+					return
+				}
+				if authCookie.TokenKey != "" {
+					key := auth.JoinToken(auth.SessionToken, authCookie.TokenID, authCookie.TokenKey)
+					r.Header.Add("Authorization", "Bearer "+key)
+					if w.Header().Get("Access-Control-Allow-Origin") != "" {
+						w.Header().Del("Access-Control-Allow-Origin")
+					}
```
htdvisser, line 48: There are more CORS headers than just this one. Also, you can just call `Del` without wrapping it in an `if`.

pkg/webmiddleware/cookie_auth.go
```diff
@@ -0,0 +1,57 @@
...
+	"net/http"
+
+	"github.com/gorilla/csrf"
+	"go.thethings.network/lorawan-stack/v3/pkg/auth"
+)
+
+// CookieAuth extracts the auth cookie and forwards it to the Authorization
+// header as SessionToken while securing the authorization via CSRF token.
+func CookieAuth(cookieKey string, csrfAuthKey []byte) MiddlewareFunc {
+	CSRFMiddleware := CSRF(csrfAuthKey, csrf.CookieName("_csrf"))
```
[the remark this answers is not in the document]
htdvisser: You're also missing quite a few other options to actually make the cookie secure. How about Path, Domain, Expires, MaxAge, Secure, HttpOnly, SameSite? Are those all set correctly by `gorilla/csrf`?

pkg/web/middleware/csrf.go
```diff
@@ -0,0 +1,50 @@
...
+				return next(c)
+			}
+		}
+		return func(c echo.Context) error {
+			return echo.WrapMiddleware(
+				webmiddleware.CSRF(
+					[]byte(authKey),
+					csrf.CookieName(cookieName),
+					csrf.Secure(c.Request().TLS != nil),
+					csrf.Path(path),
```
htdvisser, line 44: These options are not consistent with other places where we use CSRF

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -118,6 +119,14 @@ var startCommand = &cobra.Command{
 
 		var componentOptions []component.Option
 
+		if len(config.ServiceBase.HTTP.CSRF.AuthKey) == 0 || web.IsZeros(config.ServiceBase.HTTP.CSRF.AuthKey) {
+			if !web.IsZeros(config.ServiceBase.HTTP.Cookie.HashKey) && len(config.ServiceBase.HTTP.Cookie.HashKey) == 32 {
+				config.ServiceBase.HTTP.CSRF.AuthKey = config.ServiceBase.HTTP.Cookie.HashKey
+			} else {
+				config.ServiceBase.HTTP.CSRF.AuthKey = random.Bytes(32)
```
htdvisser, line 126: This is double. You're doing the same in `pkg/web/web.go`

pkg/web/web.go
```diff
@@ -259,7 +274,8 @@ func New(ctx context.Context, opts ...Option) (*Server, error) {
 	return s, nil
 }
 
-func isZeros(buf []byte) bool {
+// IsZeros checks whether a byte array consists only of zeros.
+func IsZeros(buf []byte) bool {
```
htdvisser, line 278: Not sure we should export such a small func only to use it in `cmd/ttn-lw-stack/commands/start.go`

pkg/config/shared.go
```diff
@@ -99,6 +106,7 @@ type HTTP struct {
 	RedirectToHTTPS bool             `name:"redirect-to-tls" description:"Redirect HTTP requests to HTTPS"`
 	Static          HTTPStaticConfig `name:"static"`
 	Cookie          Cookie           `name:"cookie"`
+	CSRF            CSRF             `name:"csrf"`
```
htdvisser, line 109: Can we really not just re-use the existing cookie conf for CSRF cookies?

pkg/identityserver/entity_access.go
```diff
@@ -197,6 +197,42 @@ func (is *IdentityServer) authInfo(ctx context.Context) (info *ttnpb.AuthInfoRes
...
+			}
+			user, err = store.GetUserStore(db).GetUser(ctx, &session.UserIDs, userFieldMask)
+			if err != nil {
+				if errors.IsNotFound(err) {
+					return errTokenNotFound.WithCause(err)
+				}
+				return err
+			}
+
+			userRights = ttnpb.RightsFrom(ttnpb.RIGHT_ALL).Implied()
```
htdvisser, line 233: This could use a comment because it looks dangerous

htdvisser:
Requesting @adriansmares as a second pair of eyes. I've always been afraid of cookies, so don't want to be the only one to review this.

api/user.proto
```diff
@@ -170,11 +170,13 @@ message UserSessionIdentifiers {
 }
 
 message UserSession {
-  UserIdentifiers user_ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (validate.rules).message.required = true];
-  string session_id = 2 [(gogoproto.customname) = "SessionID", (validate.rules).string.max_len = 64];
+  UserIdentifiers user_ids = 1 [(gogoproto.customname) = "UserIDs", (gogoproto.nullable) = false, (validate.rules).message.required = true];
```
adriansmares, line 173: Nit: what's the reason for the custom name here/the removal of embed ? It makes the change set longer than it could have been.

pkg/config/shared.go
```diff
@@ -66,6 +66,13 @@ type Cookie struct {
 	BlockKey []byte `name:"block-key" description:"Key for cookie contents encryption (16, 24 or 32 bytes)"`
 }
 
+// CSRF represents cross-site request forgery protection configuration.
+// The authentication key should be 32-bytes long and persist across application restarts.
+// Changing the auth key will invalidate csrf tokens and break your CSRF validation.
```
adriansmares, line 71: `CSRF tokens`

pkg/identityserver/store/user_session.go
```diff
@@ -24,8 +24,11 @@ import (
 type UserSession struct {
 	Model
 
-	User   *User
-	UserID string `gorm:"type:UUID;index:user_session_user_index;not null"`
+	TokenID string `gorm:"type:VARCHAR;unique_index:api_key_id_index"`
```
adriansmares, line 27: The index name is wrong.

pkg/oauth/user.go
```diff
@@ -192,16 +187,32 @@ func (s *server) Login(c echo.Context) error {
...
 	userIDs := ttnpb.UserIdentifiers{UserID: req.UserID}
+	tokenID, err := auth.GenerateID(ctx)
+	if err != nil {
+		return err
+	}
+	tokenKey, err := auth.GenerateKey(ctx)
+	if err != nil {
+		return err
+	}
+	hashedKey, err := auth.Hash(ctx, tokenKey)
```
adriansmares, line 198: What is the added benefit of also hashing this ? The bytes returned by `GenerateKey` are just random bytes (it's not a derived key by any means).

pkg/webmiddleware/cookie.go
```diff
@@ -0,0 +1,45 @@
...
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			ctx := r.Context()
+
+			next.ServeHTTP(w, r.WithContext(context.WithValue(ctx, ctxKey, s)))
+		})
+	}
+}
+
+// GetEncoder retrieves the secure cookie encoder from the context.
+func GetEncoder(ctx context.Context) (encoder *securecookie.SecureCookie, err error) {
```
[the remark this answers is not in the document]
adriansmares: `GetSecureCookie` since you're returning the actual implementation, not some interface. Or maybe `GetCookieCodec`.

pkg/webmiddleware/cookie_auth.go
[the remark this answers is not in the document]
kschiffer: They are quite restrictive by default but I added some more options.

pkg/config/shared.go
kschiffer: Yeah, I guess this is unnecessary configurability. Removed.

pkg/oauth/user.go
kschiffer: We're doing the same thing for API keys and access tokens as well. I guess the rationale is that the actual keys do not get exposed in case of a DB leak.

api/user.proto
kschiffer: I wanted to bring the structure closer to the one of access tokens. But yeah, I guess in the end it's an unnecessary change, so I reset it to the original structure.

kschiffer:
I've made various updates according to your reviews. Most notably, I've improved the structuring of the middlewares in a way that improves the separation of concerns and makes them properly isolated. I've also removed the superfluous CSRF config and use the cookie hashes now by default, which should be sufficient.

Please have another look while I'll work on updating the unit tests. Thanks for the helpful reviews!

kschiffer:
I've finalized everything and added/updated the unit tests as well now.

johanstokking:
@kschiffer many reviewers here; anything specific I should look at?

api/user.proto
```diff
@@ -175,6 +175,8 @@ message UserSession {
   google.protobuf.Timestamp created_at = 3 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
   google.protobuf.Timestamp updated_at = 4 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
   google.protobuf.Timestamp expires_at = 5 [(gogoproto.stdtime) = true];
+  // The session secret is used to compose an authorization key and is never returned.
+  string session_secret = 7;
```
johanstokking, line 179: Why not `6`?

Also, is there no validation here at all? Do we ever receive this value from a client?

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -118,6 +119,20 @@ var startCommand = &cobra.Command{
...
+		if len(cookieHashKey) == 0 || isZeros(cookieHashKey) {
+			cookieHashKey = random.Bytes(64)
+			config.ServiceBase.HTTP.Cookie.HashKey = cookieHashKey
+			logger.WithField("hash_key", cookieHashKey).Warn("No cookie hash key configured, generated a random one")
+		}
+
+		if len(cookieBlockKey) == 0 || isZeros(cookieBlockKey) {
+			cookieBlockKey = random.Bytes(32)
+			config.ServiceBase.HTTP.Cookie.BlockKey = cookieBlockKey
+			logger.WithField("block_key", cookieBlockKey).Warn("No cookie block key configured, generated a random one")
```
johanstokking, line 133: I wonder why we print secrets in the log. Is there a reason to add the actual value as field? It's quite hard for sysops to protect logs

kschiffer:
> @kschiffer many reviewers here; anything specific I should look at?

Code-owned files is sufficient.

api/user.proto
kschiffer: Indeed. Leftover from when I added a token id as well.

cmd/ttn-lw-stack/commands/start.go
kschiffer: Good point. Removed.

pkg/webui/oauth/views/authorize/index.js
```diff
@@ -73,6 +72,7 @@ export default class Authorize extends PureComponent {
     const {
       env: {
         pageData: { client, user, error },
+        csrfToken,
```
bafonins, line 75: Could you add to the prop-types definitions of `env` ?

pkg/webui/console/api/index.js
```diff
@@ -76,24 +80,31 @@ export default {
...
+          if (freshCsrf) {
+            return axios.post(`${appRoot}/api/auth/logout`, undefined, headers(freshCsrf))
+          }
         }
-      }
 
-      return instance.post(`${appRoot}/api/auth/logout`, undefined, {
-        headers: { 'X-CSRF-Token': csrf },
-      })
+        return error
```
bafonins, line 106: should this `throw` ? otherwise the user middleware processes this as a valid response and tries to extract `op_logout_uri`

pkg/identityserver/store/user_session_store.go
```diff
@@ -97,6 +98,30 @@ func (s *userSessionStore) GetSession(ctx context.Context, userIDs *ttnpb.UserId
...
+	defer trace.StartRegion(ctx, "get user session by session ID").End()
+	query := s.query(ctx, UserSession{}).Where(UserSession{Model: Model{ID: sessionID}})
+	var sessionModel UserSession
+	if err := query.Find(&sessionModel).Error; err != nil {
+		if gorm.IsRecordNotFoundError(err) {
+			return nil, errSessionNotFound.WithAttributes("session_id", sessionID)
+		}
+		return nil, err
+	}
+	query = s.query(ctx, Account{}).Where(Account{AccountID: sessionModel.UserID})
```
htdvisser, line 111: The index is by `AccountID` and `AccountType`, so please query by both.
```suggestion
	query = s.query(ctx, Account{}).Where(Account{
		AccountID: sessionModel.UserID,
		AccountType: "user",
	})
```

pkg/web/middleware/csrf.go
```diff
@@ -0,0 +1,53 @@
...
+			}
+		}
+		return func(c echo.Context) error {
+			return echo.WrapMiddleware(
+				webmiddleware.CSRF(
+					nil,
+					authKey,
+					csrf.CookieName(lookupName),
+					csrf.FieldName(lookupName),
+					csrf.Secure(c.Request().TLS != nil),
```
htdvisser, line 45: This may give incorrect behavior if we're behind a TLS-terminating reverse proxy. Our `RequestURL` middleware (`pkg/webmiddleware/request_url.go`) and `ProxyHeaders` middleware (`pkg/webmiddleware/proxy_headers.go`) set `(*http.Request).URL.Scheme` to `"https"` for this purpose.

pkg/webmiddleware/cookie.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"context"
+	"net/http"
+
+	"github.com/gorilla/securecookie"
+)
+
+type ctxKeyType struct{}
+
+var ctxKey ctxKeyType
```
htdvisser, line 26: These context keys are not for all of `pkg/webmiddleware`, so better to rename them to `secureCookieCtxKey` or something.

pkg/webmiddleware/cookie_auth.go
```diff
@@ -0,0 +1,50 @@
...
+package webmiddleware
+
+import (
+	"net/http"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/auth"
+)
+
+// CookieAuth extracts the auth cookie and forwards it to the Authorization
+// header.
```
htdvisser, line 24: Nit: Avoid unnatural line splitting

pkg/webmiddleware/cookie_auth.go
```diff
@@ -0,0 +1,50 @@
...
+func CookieAuth(cookieKey string) MiddlewareFunc {
```
htdvisser, line 25: Nit: I think it's just called `name`, not `key`

pkg/webmiddleware/cookie.go
```diff
@@ -0,0 +1,45 @@
...
+		})
+	}
+}
+
+// GetSecureCookie retrieves the secure cookie encoder from the context.
+func GetSecureCookie(ctx context.Context) (secureCookie *securecookie.SecureCookie, err error) {
+	secureCookie = ctx.Value(ctxKey).(*securecookie.SecureCookie)
+
+	return secureCookie, nil
+}
```
htdvisser, line 45: - The returned error is always `nil`
- This panics when calling `GetSecureCookie` on a context that doesn't have one.

pkg/webmiddleware/cookie_auth.go
```diff
@@ -0,0 +1,50 @@
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if r.Header.Get("Authorization") == "" {
```
htdvisser, line 28: Better to `next.ServeHTTP(w, r)` and `return` if `r.Header.Get("Authorization") != ""`

pkg/webmiddleware/cookie_auth.go
```diff
@@ -0,0 +1,50 @@
...
+				cookieValue, err := r.Cookie(cookieKey)
+				if err != nil {
+					next.ServeHTTP(w, r)
+					return
+				}
+				encoder, _ := GetSecureCookie(r.Context())
```
htdvisser, line 34: - Smell: not checking errors
- Nit: Technically this is a `codec`, not just an `encoder`

pkg/webmiddleware/cookie_auth.go
```diff
@@ -0,0 +1,50 @@
...
+				authCookie := &auth.CookieShape{}
+				err = encoder.Decode(cookieKey, cookieValue.Value, authCookie)
+				if err != nil {
+					next.ServeHTTP(w, r)
+					return
+				}
+				if authCookie.SessionSecret != "" {
+					key := auth.JoinToken(auth.SessionToken, authCookie.SessionID, authCookie.SessionSecret)
+					r.Header.Add("Authorization", "Bearer "+key)
```
htdvisser, line 43: Better to use use `r.Header.Set`

pkg/webmiddleware/cors.go
```diff
@@ -49,7 +55,20 @@ func (c CORSConfig) options() []handlers.CORSOption {
 	return options
 }
 
-// CORS returns a middleware that handles Cross-Origin Resource Sharing.
-func CORS(config CORSConfig) MiddlewareFunc {
-	return MiddlewareFunc(handlers.CORS(config.options()...))
+// CORS returns a middleware that handles Cross-Origin Resource Sharing. The
+// skip parameter can be used to skip the decoration of CORS headers, based
+// on the request interface.
+func CORS(skip skipFunction, config CORSConfig) MiddlewareFunc {
```
htdvisser, line 61: Can't we make the skip func part of the config?

pkg/webmiddleware/cors.go
```diff
@@ -49,7 +55,20 @@ func (c CORSConfig) options() []handlers.CORSOption {
...
+	return func(next http.Handler) http.Handler {
+		if skip != nil {
```
htdvisser, line 63: In Go we typically prefer the early exit:

```go
if skip == nil {
	return handlers.CORS(config.options()...)(next)
}
```

pkg/webmiddleware/cors.go
htdvisser: Actually, I have a better idea. Will comment later.

pkg/webmiddleware/csrf.go
```diff
@@ -0,0 +1,46 @@
...
+
+package webmiddleware
+
+import (
+	"net/http"
+
+	"github.com/gorilla/csrf"
+)
+
+type skipCheckFunction func(*http.Request) bool
```
htdvisser, line 23: I don't like this skipping logic in the middleware. Will comment with my suggestion later.

pkg/webmiddleware/csrf.go
```diff
@@ -0,0 +1,46 @@
...
+		if skipCheck != nil {
+			return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				req := r
+				if skipCheck(r) {
+					r = csrf.UnsafeSkipCheck(r)
+				}
+				checkHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					next.ServeHTTP(w, req)
+				})
+				csrf.Protect(authKey, append(opts, csrf.Secure(r.TLS != nil))...)(checkHandler).ServeHTTP(w, r)
```
htdvisser, line 39: See my other comment about the TLS field behind a TLS-terminating proxy. Also below.

tools/mage/dev.go
```diff
@@ -69,7 +69,7 @@ func (Dev) Misspell() error {
 var (
 	devDatabases          = []string{"cockroach", "redis"}
 	devDataDir            = ".env/data"
-	devDatabaseName       = "ttn_lorawan"
+	devDatabaseName       = "ttn_lorawan_dev"
```
htdvisser, line 72: cc: @rvolosatovs

htdvisser:
Follow-up on my comment about the "skipper".

I think it's better to not make skipping middleware for specific requests part of the middleware itself. If we do this, we'd have to duplicate that code for every middleware, and we shouldn't do that.

Instead, I propose to add a "conditional" middleware that conditionally executes or skips another middleware:

```go
// Conditional is a middleware that only executes middleware if the condition
// returns true for the request. If the condition returns false, the middleware
// is skipped, and request handling moves on to the next handler in the chain.
func Conditional(middleware MiddlewareFunc, condition func(r *http.Request) bool) MiddlewareFunc {
	return func(next http.Handler) http.Handler {
		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			handler := next
			if condition(r) {
				handler = middleware(next)
			}
			handler.ServeHTTP(w, r)
		})
	}
}
```

This could be used as follows:

```go
apiRouter.Use(
	mux.MiddlewareFunc(webmiddleware.Conditional(
		webmiddleware.CORS(webmiddleware.CORSConfig{
			// ...
		}),
		func(r *http.Request) bool {
			auth := r.Header.Get("Authorization")
			return auth == "" || strings.HasPrefix(strings.ToLower(auth), "bearer ")
		},
	)),
)
```

pkg/web/middleware/csrf.go
kschiffer: Hmm. If the Cookie is set to secure by default, it will not be included in non-TLS requests (meaning it would break the CSRF). Any idea what could be done about that, other than not using secure cookies?

tools/mage/dev.go
kschiffer: This is actually not related to this PR, so, I'll remove it. Still, this should be the database name, since the default value for the DB URI is \`postgresql://root@localhost:26257/ttn\_lorawan\_dev?sslmode=disable\`

pkg/web/middleware/csrf.go
htdvisser: My suggestion was to replace `c.Request().TLS != nil` with `c.Request().URL.Scheme == "https"` because that's what we set for (forwarded) HTTPS requests.

pkg/webmiddleware/csrf.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/gorilla/csrf"
+)
+
+// CSRF returns a middleware that enables CSRF protection via a sync token. The
+// skipCheck parameter can be used to skip CSRF protection based on the request
+// interface.
+func CSRF(authKey []byte, opts ...csrf.Option) MiddlewareFunc {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			csrf.Protect(authKey, append([]csrf.Option{csrf.Secure(r.TLS != nil)}, opts...)...)(next).ServeHTTP(w, r)
```
htdvisser, line 29: I see you made the change in `pkg/web/middleware/csrf.go`, but also here please.

kschiffer: Sorry about that.

pkg/webui/console/api/index.js
bafonins: seems that it should not, but could you explain why?

kschiffer: Sorry, I had a commit fixing this, but seems like it got missing somehow. Fixed now.

kschiffer:
> Will get back to this tomorrow to review the rest and test more use cases

@bafonins let me know if this is already done or if I can merge

kschiffer:
@pgalic96 / @bafonins please. I can't merge otherwise.

PR #2565, opened 2020-05-18
Various event-related fixes and improvements

rvolosatovs:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/pull/2096
Close https://github.com/TheThingsNetwork/lorawan-stack/issues/2569

#### Changes

- Fix `rightsutil.EventIsVisible`
- Avoid mutating proto when decoding
- Remove event stream start/end events, because they do not seem necessary, please let me know if and why we would need them. When we have the historical data as well, these won't make any sense at all IMO.
- Introduce `test.AllTrue` and use in NS

#### Checklist

### DIFF
--- a/pkg/auth/rights/rightsutil/rightsutil.go
+++ b/pkg/auth/rights/rightsutil/rightsutil.go
@@ -22,16 +22,13 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 )
 
-// EventIsVisible returns whether evt is visible given rights in the context.
-func EventIsVisible(ctx context.Context, evt events.Event) (bool, error) {
-	visibility := evt.Visibility().Union(ttnpb.RightsFrom(
-		ttnpb.RIGHT_APPLICATION_ALL,
-		ttnpb.RIGHT_CLIENT_ALL,
-		ttnpb.RIGHT_GATEWAY_ALL,
-		ttnpb.RIGHT_ORGANIZATION_ALL,
-		ttnpb.RIGHT_USER_ALL,
-	))
-	for _, entityIDs := range evt.Identifiers() {
+// EventIsVisible returns whether ev is visible given rights in the context.
+func EventIsVisible(ctx context.Context, ev events.Event) (bool, error) {
+	visibility := ev.Visibility()
+	if len(visibility.Rights) == 0 {
+		return true, nil
+	}
+	for _, entityIDs := range ev.Identifiers() {
 		switch ids := entityIDs.Identifiers().(type) {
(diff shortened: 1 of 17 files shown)


rvolosatovs:
Blocking https://github.com/TheThingsIndustries/lorawan-stack/pull/2096

pkg/events/grpc/grpc.go
```diff
@@ -80,18 +81,14 @@ type EventsServer struct {
...
-)
+var errNoIdentifiers = errors.DefineInvalidArgument("no_identifiers", "no identifiers")
 
 // Stream implements the EventsServer interface.
 func (srv *EventsServer) Stream(req *ttnpb.StreamEventsRequest, stream ttnpb.Events_StreamServer) error {
-	ctx := stream.Context()
-
 	if len(req.Identifiers) == 0 {
-		return nil
+		return errNoIdentifiers
```
johanstokking, line 89: This is an API behavioral change

rvolosatovs: Doesn't this fall under these two:
> 5. We will not break the API between components and events within minor versions. So at least the same minor versions of components are compatible with each other.
> 6. We reserve the right to fix bugs in API, configuration and storage in patches and minor updates. This may break components, gateways and applications that rely on buggy behavior.

IMO silently failing here is a bug

johanstokking: I don't see a bug report that is referenced

rvolosatovs: https://github.com/TheThingsNetwork/lorawan-stack/issues/2569

pkg/auth/rights/rightsutil/rightsutil.go
```diff
@@ -22,16 +22,13 @@ import (
...
-		ttnpb.RIGHT_APPLICATION_ALL,
-		ttnpb.RIGHT_CLIENT_ALL,
-		ttnpb.RIGHT_GATEWAY_ALL,
-		ttnpb.RIGHT_ORGANIZATION_ALL,
-		ttnpb.RIGHT_USER_ALL,
-	))
-	for _, entityIDs := range evt.Identifiers() {
+// EventIsVisible returns whether ev is visible given rights in the context.
+func EventIsVisible(ctx context.Context, ev events.Event) (bool, error) {
+	visibility := ev.Visibility().Implied()
```
htdvisser, line 27: This is not correct in this context. 

Maybe `Implied()` is a bit confusing. If a user has the "admin right", that implies that the user has all rights. If an event is visible to callers with the "admin right", that doesn't mean it's also visible to the implicit rights (i.e. for a caller with only the "info right").

PR #2580, opened 2020-05-21
Add authentication method, ID and IP event metadata when available

### LINKED ISSUE #2227: Add Auth Method + ID and Remote IP to CRUD event data
#### Summary

We should add the Auth Method and API Key / Access Token ID to CRUD events. It would also be good to add the remote IP.

#### Why do we need this?

For audit logging

We should extract both the auth (rpcmetadata) and remote address (peer) from the context when creating an event.

- Add fields to the Event definition (api)
- In `events.New`, set those fields.
- Can use `peer.FromContext` to extract the remote IP. Propbably also need to consider `x-forwarded-for` headers.
- Can use `rpcmetadata.FromIncomingContext` to extract the `AuthType` (usually `Bearer`) and `AuthValue`. Use `auth.SplitToken` on the latter to extract the `tokenType` and `tokenID`. The `tokenKey` should **NEVER** be added to events.

neoaggelos:
#### Summary

Closes #2227

#### Changes

- Add Auth Method, Token ID, Remote IP as event metadata when available
- Adds `rpcmetadata.XForwardedFor` field

#### Testing

Updated unit tests, tested in development environment with dummy API create + delete events.

##### Regressions

...

#### Notes for Reviewers

- Cannot use `auth.SplitToken()` since it creates an import cycle in the tests.
- We might want to do something different for `x-forwarded-for`.

#### Checklist


pkg/events/events.go
```diff
@@ -164,6 +195,31 @@ func New(ctx context.Context, name string, identifiers CombinedIdentifiers, data
...
+		// Cannot use auth.SplitToken() here because of import cycle
+		parts := strings.Split(md.AuthValue, ".")
+		if len(parts) == 3 {
+			authentication.TokenType = parts[0]
+			authentication.TokenID = parts[1]
+		}
+	}
+	if md.XForwardedFor != "" {
+		authentication.RemoteIP = md.XForwardedFor
+	}
```
htdvisser, line 219: The value of the XFF header isn't equal to RemoteIP. We may have to split it and take the correct one if there are multiple.

pkg/events/events.go
```diff
@@ -164,6 +195,31 @@ func New(ctx context.Context, name string, identifiers CombinedIdentifiers, data
 	if identifiers != nil {
 		evt.innerEvent.Identifiers = identifiers.CombinedIdentifiers().GetEntityIdentifiers()
 	}
+	authentication := &ttnpb.Event_Authentication{}
+	if p, ok := peer.FromContext(ctx); ok && p.Addr.String() != "pipe" {
+		host, _, err := net.SplitHostPort(p.Addr.String())
+		if err == nil {
```
htdvisser, line 201: Nit: can shorten this to `if host, _, err := ...; err == nil`

pkg/events/events.go
```diff
@@ -164,6 +195,31 @@ func New(ctx context.Context, name string, identifiers CombinedIdentifiers, data
...
+			authentication.RemoteIP = host
+		}
+	}
+	md := rpcmetadata.FromIncomingContext(ctx)
+	if md.AuthType != "" {
+		authentication.Type = md.AuthType
+	}
+	if md.AuthValue != "" {
+		// Cannot use auth.SplitToken() here because of import cycle
```
htdvisser, line 210: Then we should probably fix that differently. `pkg/auth` itself isn't importing `pkg/events`. Perhaps the cycle is because `password_test.go` uses `package auth` instead of `package auth_test`, which should be an easy fix.

pkg/events/events.go
```diff
@@ -242,3 +298,12 @@ func UnmarshalJSON(data []byte) (Event, error) {
 	}
 	return e, nil
 }
+
+// DropAuthenticationMetadata returns the evetn without authentication metadata.
```
htdvisser, line 302: race condition between your left hand and your right hand: evetn

pkg/events/events.go
rvolosatovs: Disagree; prefer short `if` only for single-value returns.

htdvisser: I don't think we should start discussions for nit comments. I probably shouldn't have pointed this out in the first place, and instead have let @neoaggelos write in whatever style he likes best.

FTR: My own rule of thumb is to combine it if I have max 1 value (`_` doesn't count), an optional err/ok and if the combination fits on the line.

pkg/events/events.go
neoaggelos: Actually it is `auth_test` that imports `assertions/should`, which in turn imports `pkg/events`. We cannot avoid that.

pkg/events/events.go
neoaggelos: How is "the correct one" defined in this context?

neoaggelos: Updated to only keep the first part of the X-forwarded-for header (client IP)

neoaggelos:
> Please revert [ca2422c](https://github.com/TheThingsNetwork/lorawan-stack/commit/ca2422c0899a9be59ccccb17da93a439af4d8518) and instead correctly adjust
> 
> https://github.com/TheThingsNetwork/lorawan-stack/blob/2f72186c65a495f4de8984f60d4c80484119a772/pkg/util/test/events.go#L113-L129

In general I agree, but a test keeps failing because it uses `should.HaveSameElementsFunc` instead. https://github.com/TheThingsNetwork/lorawan-stack/blob/ca2422c0899a9be59ccccb17da93a439af4d8518/pkg/networkserver/networkserver_flow_test.go#L538

A quick search shows `should.HaveSameElementsFunc` is only used for events, but its name does not imply that. Do you have a particular suggestion for this?

rvolosatovs:
> > Please revert [ca2422c](https://github.com/TheThingsNetwork/lorawan-stack/commit/ca2422c0899a9be59ccccb17da93a439af4d8518) and instead correctly adjust
> > https://github.com/TheThingsNetwork/lorawan-stack/blob/2f72186c65a495f4de8984f60d4c80484119a772/pkg/util/test/events.go#L113-L129
> 
> In general I agree, but a test keeps failing because it uses `should.HaveSameElementsFunc` instead.
> 
> https://github.com/TheThingsNetwork/lorawan-stack/blob/ca2422c0899a9be59ccccb17da93a439af4d8518/pkg/networkserver/networkserver_flow_test.go#L538
> 
> A quick search shows `should.HaveSameElementsFunc` is only used for events, but its name does not imply that. @rvolosatovs Do you have a particular suggestion for this, maybe we should add a new version of it that ignores event authentication fields?

It does not fail because of `should.HaveSameElementsFunc`, but because of `flowTestEventEqual`, which should also be adapted to the changes in this PR

rvolosatovs:
Also, note that
> Cannot use `auth.SplitToken()` since it creates an import cycle in the tests.

Is a bug, there must not be import cycles in tests, if there are, it means that some tests are incorrectly placed in the same package as the implementation. Tests for a package `x` should be placed in `x_test` package if possible, that way import cycles are avoided.

neoaggelos:
Reverted commit, updated equality checks for events to ignore authentication metadata.

Fixed the package name in `password_test.go`, no import cycles now.

Finally, changed the `token_type` field to use a proper description instead of a cryptic base32 encoded string.

pkg/auth/password_test.go
```diff
@@ -12,12 +12,13 @@
...
 // limitations under the License.
 
-package auth
+package auth_test
 
 import (
 	"testing"
 
 	"github.com/smartystreets/assertions"
+	"go.thethings.network/lorawan-stack/v3/pkg/auth"
```
rvolosatovs, line 21:
```suggestion
	. "go.thethings.network/lorawan-stack/v3/pkg/auth"
```

pkg/auth/pbkdf2/algorithm_test.go
```diff
@@ -12,49 +12,50 @@
...
 // limitations under the License.
 
-package pbkdf2
+package pbkdf2_test
 
 import (
 	"testing"
 
 	. "github.com/smartystreets/assertions"
+	"go.thethings.network/lorawan-stack/v3/pkg/auth/pbkdf2"
```
rvolosatovs, line 21:
```suggestion
	. "go.thethings.network/lorawan-stack/v3/pkg/auth/pbkdf2"
```

pkg/auth/pbkdf2/pbkdf2_test.go
```diff
@@ -12,27 +12,28 @@
...
 // limitations under the License.
 
-package pbkdf2
+package pbkdf2_test
 
 import (
 	"testing"
 
 	. "github.com/smartystreets/assertions"
+	"go.thethings.network/lorawan-stack/v3/pkg/auth/pbkdf2"
```
rvolosatovs, line 21:
```suggestion
	. "go.thethings.network/lorawan-stack/v3/pkg/auth/pbkdf2"
```

pkg/events/events.go
```diff
@@ -130,6 +138,30 @@ func (e event) CorrelationIDs() []string                { return e.innerEvent.Co
 func (e event) Origin() string                          { return e.innerEvent.Origin }
 func (e event) Caller() string                          { return e.caller }
 func (e event) Visibility() *ttnpb.Rights               { return e.innerEvent.Visibility }
+func (e event) AuthType() string {
+	if e.innerEvent.Authentication != nil {
```
rvolosatovs, line 142: nit; prefer indenting error flow and, hence, flipping this `if`(same below). See https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

pkg/events/events.go
```diff
@@ -164,6 +196,29 @@ func New(ctx context.Context, name string, identifiers CombinedIdentifiers, data
...
+		if tokenType, tokenID, _, err := auth.SplitToken(md.AuthValue); err == nil {
+			authentication.TokenType = tokenType.String()
+			authentication.TokenID = tokenID
+		}
+	}
+	if md.XForwardedFor != "" {
+		xff := strings.Split(md.XForwardedFor, ",")
+		authentication.RemoteIP = strings.Trim(xff[0], " ")
+	}
+	if !authentication.Equal(&ttnpb.Event_Authentication{}) {
```
rvolosatovs, line 219: Do we really need to allocate a new object on the heap to check if this is empty (note that you're actually constructing it here, so you definitely can determine if it will be empty without doing this)?
Sure, this will not be the performance bottleneck in production, but quite a bad code smell IMO.

pkg/networkserver/networkserver_flow_test.go
```diff
@@ -48,9 +49,11 @@ func hasProperStringSubset(x, y []string) bool {
 }
 
 func flowTestEventEqual(x, y events.Event) bool {
+	fmt.Println(">>>", x, y)
```
rvolosatovs, line 52: :eyes: 
```suggestion
```

pkg/networkserver/networkserver_flow_test.go
```diff
@@ -48,9 +49,11 @@ func hasProperStringSubset(x, y []string) bool {
...
 	if test.EventEqual(x, y) {
 		return true
 	}
+	fmt.Println("###", x, y)
```
rvolosatovs, line 56:
```suggestion
```

pkg/networkserver/networkserver_flow_test.go
```diff
@@ -673,6 +678,7 @@ func makeClassCOTAAFlowTest(macVersion ttnpb.MACVersion, phyVersion ttnpb.PHYVer
 		defer func() {
 			closeLink()
 			if !a.So(test.AssertEventPubSubPublishRequest(ctx, env.Events, func(ev events.Event) bool {
+				// ev = events.DropAuthenticationMetadata(ev)
```
rvolosatovs, line 681:
```suggestion
```

pkg/networkserver/networkserver_flow_test.go
```diff
@@ -846,6 +852,9 @@ func makeClassCOTAAFlowTest(macVersion ttnpb.MACVersion, phyVersion ttnpb.PHYVer
 			return
 		}
 		a.So(test.AssertEventPubSubPublishRequests(ctx, env.Events, 2+len(expectedEvs), func(evs ...events.Event) bool {
+			// for idx, e := range evs {
```
rvolosatovs, line 855: Remove

neoaggelos:
Changed to target develop because of `compat/api` label, even though this is just additions to the API

PR #2582, opened 2020-05-23
all: Convert some standard errors into well known errors

### LINKED ISSUE #2448: Convert common/well-known errors to defined errors
#### Summary

We should convert some common and well-known errors to defined errors.

#### Why do we need this?

To give more insight into what is going on when an error is returned.

If during OAuth token exchange a DNS error, timeout or other network error occurs, an internal server error is returned without details.

Some details about the network error. Maybe DNS couldn't be resolved? Maybe the TCP connection timed out? Maybe the TLS expired? ...

We could make `errors.From` recognize some common errors and return rich errors with attributes (built from definitions).

Yes, but if anyone else would like to do this, just comment and re-assign.

neoaggelos:
#### Summary

Closes #2448

#### Changes

- Well defined errors for context canceled and deadline exceeded
- `errors.From` can parse most common net.Errors

#### Testing

unit tests

##### Regressions

...

#### Checklist

### DIFF
--- a/pkg/errors/conversion.go
+++ b/pkg/errors/conversion.go
@@ -15,8 +15,22 @@
 package errors
 
 import (
+	"context"
+	"net"
+
 	"github.com/golang/protobuf/proto"
 	"google.golang.org/grpc/status"
+)
+
+var (
+	errContextDeadlineExceeded = DefineDeadlineExceeded("context_deadline_exceeded", "context deadline exceeded")
+	errContextCanceled         = DefineCanceled("context_canceled", "context canceled")
+
+	errNetInvalidAddr    = DefineInvalidArgument("net_invalid_addr", "{message}", "message", "temporary", "timeout")
+	errNetAddr           = DefineUnavailable("net_addr", "{message}", "message", "temporary", "timeout")
+	errNetDNS            = DefineUnavailable("net_dns", "{message}", "message", "temporary", "timeout", "not_found")
+	errNetUnknownNetwork = DefineNotFound("net_unknown_network", "{message}", "message", "temporary", "timeout")
+	errNetOperation      = DefineUnavailable("net_operation", "{message}", "message", "op", "net", "source", "address", "timeout", "temporary")
 )
 
 // From returns an *Error if it can be derived from the given input.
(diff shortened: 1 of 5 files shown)


pkg/errors/errors_test.go
```diff
@@ -74,3 +77,152 @@ func TestFields(t *testing.T) {
 		"error_cause_cause": "back",
 	})
 }
+
+func TestContextCancelled(t *testing.T) {
```
johanstokking, line 81: Canceled

pkg/errors/conversion.go
```diff
@@ -65,6 +86,38 @@ func From(err error) (out *Error, ok bool) {
 			e = e.WithCause(cause)
 		}
 		return &e, true
+	case *net.DNSError:
+		e := build(errNetDNS, 0).WithAttributes(
+			append([]interface{}{"not_found", err.IsNotFound}, netErrorDetails(err)...)...,
```
johanstokking, line 91: You should be able to chain the call with `WithAttributes()`

pkg/errors/conversion.go
```diff
@@ -15,10 +15,24 @@
...
 
+var (
+	errContextDeadlineExceeded = DefineDeadlineExceeded("context_deadline_exceeded", "context deadline exceeded")
+	errContextCanceled         = DefineCanceled("context_canceled", "context canceled")
+
+	errNetInvalidAddr    = DefineInvalidArgument("net_invalid_addr", "invalid address error: `{message}`", "message", "temporary", "timeout")
+	errNetAddr           = DefineUnavailable("net_addr", "address error: `{message}`", "message", "temporary", "timeout")
+	errNetDNS            = DefineUnavailable("net_dns", "DNS error: `{message}`", "message", "temporary", "timeout", "not_found")
+	errNetUnknownNetwork = DefineNotFound("net_unknown_network", "unknown network: `{message}`", "message", "temporary", "timeout")
+	errNetOperation      = DefineUnavailable("net_operation", "network operation error: `{message}`", "message", "op", "net", "source", "address", "timeout", "temporary")
```
johanstokking, line 33: - We need the `error:` text in the error message
- You don't need to use backticks for the error message
- We can also consider just using the error message as-is, and use the error name to know what kind of error it is

neoaggelos:
> Why is this targeting `develop`?

This is possibly changing errors shown in the console or returned by the API. Is this not a breaking change?

johanstokking:
> This is possibly changing errors shown in the console or returned by the API. Is this not a breaking change?

No, that is not a breaking change. See the compatibility commitment; it does not explicitly state errors

htdvisser:
@neoaggelos Travis is satisfied. You can merge this.

Issue #2415, opened 2020-04-24
doc: Hidden nav item css breakpoints


benolayinka:
#### Summary

After doc restructuring #2368 , nav items don't fit on small screens 
![image](https://user-images.githubusercontent.com/6963436/80193409-08e8cb80-8619-11ea-84f3-2dacf566df50.png)

#### Steps to Reproduce

1. Pull the master branch and run the docs server.
2. Resize your window to approx 1050px wide

#### What do you see now?

![image](https://user-images.githubusercontent.com/6963436/80193409-08e8cb80-8619-11ea-84f3-2dacf566df50.png)

#### What do you want to see instead?

Maybe burger menu? 

<img width="1532" alt="Screenshot 2020-04-24 at 10 52 35" src="https://user-images.githubusercontent.com/6963436/80193890-dd1a1580-8619-11ea-83e4-c5c6d59de255.png">

#### Environment

Chrome 80

#### How do you propose to implement this?

Css breakpoints?

#### Can you do this yourself and submit a Pull Request?

Not cleanly. @kschiffer @htdvisser @pgalic96 ideas?

htdvisser:
Changing the breakpoint is easy (see below), but I don't think it improves the user experience. Maybe we should reconsider whether we really want the "Github", "Forum" and "Get The Things Stack" links in the top bar, or if those can actually be removed. The "Github" and "Forum" links are also in the footer, and we also have a huge "hero" on the homepage with a "Get The Things Stack" link.

This is how you change the breakpoint (default is `960px + (2 * $gap)`):

```diff
diff --git a/doc/themes/the-things-stack/assets/css/theme.scss b/doc/themes/the-things-stack/assets/css/theme.scss
index 5b146d66f..c5a984c5f 100644
--- a/doc/themes/the-things-stack/assets/css/theme.scss
+++ b/doc/themes/the-things-stack/assets/css/theme.scss
@@ -56,6 +56,8 @@ $table-cell-border: 1px solid $white-ter;
 $table-cell-border-width: 1px 1px 1px;
 $table-cell-padding: 1.2em 0.75em;
 
+$desktop: 1024px + (2 * $gap);
+
 @import "bulma/sass/utilities/functions.sass";
 @import "bulma/sass/utilities/derived-variables.sass";
 @import "bulma/sass/utilities/animations.sass";

```

benolayinka:
[...]

PR #2588, opened 2020-05-25
Add search to docs

### LINKED ISSUE #782: Search content in the documentation
#### Summary
Search function is common and essential function for any documentation.
...

#### Why do we need this?
The documentation must be searchable to help finding document related to a topic and find problem solving solution faster.

...

Nothing
...

A search bar proposing result based on the input
...

#### Environment
Hugo, HTML, JS
...

Hugo propose several option on there website.
https://gohugo.io/tools/search/
Algolia looks like a good choice for the doc hosted on github, they a free Community tiers that can fit.
A local search for offline documentation would be interesting too.
...

Yes
...

benolayinka:
#### Summary

Closes #782 

#### Changes

- Add search to docs
- Add gh-pages config so `mage docs:build` uses separate envs

Screenshot 2020-06-09 at 12 01 42
Screenshot 2020-06-09 at 12 01 36

#### Testing

Run locally

##### Regressions

...

#### Notes for Reviewers

Beware when testing - the results return links to the thethingsstack.io, so clicking on a result will take you out of your local environment. This may also be an issue in the future with versioning, but Algolia crawls every 24hrs so hopefully it updates ok.

DocSearch is free forever from Algolia. They crawl automatically, and create a hosted index, with some js and an API key to access it. The only requirement is that we leave their logo in the search bar.

Todo:
- What's the right place for the stylesheet and scripts? @htdvisser 
- We are already out of space on the top nav bar, what should we do? @pierrephz 
- Config.toml to enable only for some deployments @htdvisser 
- Manage separate index for TTES @benolayinka 

#### Checklist


benolayinka:
Looks like I didn't rebase correctly, will fix.

@pierrephz this is the out of space issue #2415

doc/themes/the-things-stack/layouts/partials/head.html
```diff
@@ -8,6 +8,7 @@
   <link rel="icon" type="image/png" href="{{ "favicon.png" | relURL }}">
   {{- $style := resources.Get "css/theme.scss" | resources.ToCSS (dict "includePaths" (slice "node_modules")) | resources.Minify | resources.Fingerprint }}
   <link rel="stylesheet" href="{{ $style.Permalink }}" integrity="{{ $style.Data.Integrity }}">
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
```
htdvisser, line 11: We need to make sure our Content Security Policy allows this domain for CSS.

doc/themes/the-things-stack/layouts/partials/nav.html
```diff
@@ -39,3 +42,12 @@
     </div>
   </div>
 </nav>
+
+<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
```
htdvisser, line 46: We need to make sure our Content Security Policy allows this domain for JS.

doc/themes/the-things-stack/layouts/partials/nav.html
```diff
@@ -39,3 +42,12 @@
...
+<script type="text/javascript"> docsearch({
+apiKey: '4372282a02970d8402d43457bf95d486',
```
htdvisser, line 48: Can we please not hardcode API keys?

doc/themes/the-things-stack/layouts/shortcodes/ttnv2.html
```diff
@@ -0,0 +1 @@
+The Things Network Stack V2
```
htdvisser, line 1: This seems unrelated to the rest of the pull request.

doc/themes/the-things-stack/layouts/partials/nav.html
benolayinka: added to params in config.toml, but I don't know what the right way is to include it from the environment. we also want to be able to change the index for tts vs ttes

doc/themes/the-things-stack/layouts/partials/nav.html
benolayinka: where is the CSP?

pierrephz:
Regarding the search bar display,
<br/>
This is an important feature for documentation, and since we already have a full navbar, I suggest to give it a full dedicated section just below the navbar, (not fixed)
<br/>
See here a mockup of what I mean @benolayinka:
<br/>
Screenshot 2020-05-26 at 13 13 30

doc/config.toml
```diff
@@ -28,6 +28,10 @@ pygmentsUseClasses = true
   github_repository = "https://github.com/TheThingsNetwork/lorawan-stack"
   github_repository_edit = "https://github.com/TheThingsNetwork/lorawan-stack/edit/master/doc/content"
   version = "v3.8.1"
+  [params.search]
+    enabled = true
+    api_key = "4372282a02970d8402d43457bf95d486"
```
kschiffer, line 33: You'll need to source this from an environment variable. See: https://gohugo.io/getting-started/configuration/#configure-with-environment-variables

API keys should never be committed since now anyone can read it and use it for themselves. So in addition to sourcing this from ENV, you should generate a new API key and invalidate this one.

doc/config.toml.tmpl
```diff
@@ -25,6 +25,10 @@ pygmentsUseClasses = true
   github_repository = "https://github.com/TheThingsNetwork/lorawan-stack"
   github_repository_edit = "https://github.com/TheThingsNetwork/lorawan-stack/edit/master/doc/content"
   version = "{{.CurrentVersion}}"
+  [params.search]
+    enabled = true
+    api_key = "4372282a02970d8402d43457bf95d486"
```
kschiffer, line 30: Idem.

doc/config.toml
benolayinka: But how do we manage our environmental variables currently? Would this be the only one?

benolayinka: According to Algolia, this API key is search only and is safe to publish. They won't issue a new key.

benolayinka:
made the hugo params environment variables. @htdvisser i need to add them to travis for build, will sync with you for that.

@kschiffer how do we use material design icons? i included an mdi icon in the search bar using a css class but it looks like we don't load the resources for it.

localhost_1313_(Laptop with HiDPI screen) (3)

localhost_1313_(Laptop with HiDPI screen) (4)

pierrephz:
Regarding the design, they are a few things that need to be changed:

### 1 - Background

Keep the same size for the container, but change the image background to this one:

bg_doc

I suggest you to use this css style:
```
.search_container{
  height: 150px;
  background-image: url(bg_doc.png);
  background-repeat: no-repeat;
  background-position: center;
  background-size: cover;
  display: flex;
  align-items: center;
  justify-content: center;
}
```

### 2 - Search bar

Make it more impactful

```
.search_bar{
  background-color: white;
  width: 55vw;
  padding: 20px;
  border-radius: 5px;
  filter: drop-shadow(0 2px 10px rgba(0, 0, 0, 0.2));
  font-size: 16px;
  color: #c7c7c7;
}
```

I did some tests and if you use those two code snippets that should be fine

image

benolayinka:
> Without changes to `.travis.yml` this will also include search in the embedded documentation of our releases, which may get deployed to `localhost` or self-hosted deployments. I don't think that's what we want.

The HUGO_PARAMS_SEARCH_ENABLED variable disables search by default

htdvisser:
But we just added it to our Travis configuration, which means that it will be set for both [the "internal" build](https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/tools/mage/docs.go#L72-L74) and [the build for GitHub Pages](https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/tools/mage/docs.go#L92).

benolayinka:
> But we just added it to our Travis configuration, which means that it will be set for both [the "internal" build](https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/tools/mage/docs.go#L72-L74) and [the build for GitHub Pages](https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/tools/mage/docs.go#L92).

Can we specify an environment just for GH pages?

htdvisser:
In Travis that isn't so easy, but @rvolosatovs is working on migrating to GitHub Actions, which makes that much easier.

benolayinka:
@pierrephz we don't have an icon library and when I included one it didn't size properly with your custom input width so I removed it. How is this?

Screenshot 2020-06-08 at 13 30 15

pierrephz:
@benolayinka, looks better

So if you cannot size the search bar, just decrease the height of the container and it should be fine

`height: 120px;` instead of `height: 150px;` I think

Also, from your screenshots, we can see that the suggestion bubble/popup from the search bar is aligned to the right. Shouldn’t it be aligned to the left?

benolayinka:
@pierrephz 

Screenshot 2020-06-09 at 12 01 36

Screenshot 2020-06-09 at 12 01 42

laurensslats:
Printscreens look good, I'm not able to run this locally though

pierrephz:
> I don't know how happy users would be if we shove a full megabyte of `doc/themes/the-things-stack/static/img/search_bg.png` down their throats. Can't we get an SVG?

sure

@benolayinka, here is the image


[bg_doc.svg.zip](https://github.com/TheThingsNetwork/lorawan-stack/files/4752797/bg_doc.svg.zip)

kschiffer:
> [**@kschiffer**](https://github.com/kschiffer) how do we use material design icons? i included an mdi icon in the search bar using a css class but it looks like we don't load the resources for it.

We currently don't source those in the docs. If you want to use them you need to apply

```diff
diff --git a/doc/themes/the-things-stack/assets/css/theme.scss b/doc/themes/the-things-stack/assets/css/theme.scss
index cc6f87359..c68a49c59 100644
--- a/doc/themes/the-things-stack/assets/css/theme.scss
+++ b/doc/themes/the-things-stack/assets/css/theme.scss
@@ -22,6 +22,14 @@
   font-style: normal;
 }
 
+@font-face {
+  font-family: 'Material Icons'
+  font-style: normal
+  font-weight: 400
+  font-display: block
+  src: url(../fonts/materialicons.woff2) format('woff2')
+}
+
 @import "bulma/sass/utilities/initial-variables.sass";
 
 $family-sans-serif: "Lato", "LatoFallback", $family-sans-serif;
@@ -168,6 +176,23 @@ $table-cell-padding: 1.2em 0.75em;
   }
 }
 
+.material-icon {
+  font-family: 'Material Icons';
+  font-weight: normal;
+  font-style: normal;
+  font-size: 1.3rem;
+  line-height: 1;
+  letter-spacing: normal;
+  text-transform: none;
+  display: inline-block;
+  white-space: nowrap;
+  word-wrap: normal;
+  direction: ltr;
+  vertical-align: middle;
+  font-feature-settings: 'liga';
+  align-items: center;
+}
+
 .hero.is-primary {
   background-image: url(../img/hero-bg.svg);
   background-size: cover;
```

And copy the font from `/pkg/webui/assets/fonts` to `/doc/themes/static/fonts`.

The icons can then be used with the `.material-icons` class.

doc/themes/the-things-stack/assets/css/theme.scss
```diff
@@ -248,4 +248,25 @@ details + blockquote {
...
+}
+
+.search-bar {
+  background-color: white;
+  width: 55vw;
+  padding: 20px;
+  border-radius: 5px;
+  filter: drop-shadow(0 2px 10px rgba(0, 0, 0, 0.2));
+  font-size: 16px;
+  left: -1px; //hack for left-justified dropdown
```
kschiffer, line 269:
```suggestion
  left: -1px; // Hack for left-justified dropdown.
```

https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/DEVELOPMENT.md#comments

doc/themes/the-things-stack/layouts/partials/search.html
```diff
@@ -0,0 +1,24 @@
...
+      </div>
+    </div>
+</section>
+
+<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
+<script type="text/javascript"> docsearch({
+apiKey: {{ .Site.Params.search.apikey }},
+indexName: {{ .Site.Params.search.index }},
+inputSelector: '#search-input',
+debug: false // Set debug to true if you want to inspect the dropdown
```
kschiffer, line 21:
```suggestion
debug: false // Set debug to true if you want to inspect the dropdown.
```

Idem.

benolayinka:
> My previous comment still needs to be addressed. We need to make sure we only set the search for online (github-pages) docs, and not for the offline documentation build. This either needs to be fixed in Travis or we need to wait for GitHub Actions.

Yep blocked on this

benolayinka:
> My previous comment still needs to be addressed. We need to make sure we only set the search for online (github-pages) docs, and not for the offline documentation build. This either needs to be fixed in Travis or we need to wait for GitHub Actions.

@htdvisser actually if we push the pre built docs with search functionality, what does it hurt? in general i'm for fixing this but if it's going to be long and painful, maybe we don't need to block this pr if the consequence is minimal

benolayinka:
> My previous comment still needs to be addressed. We need to make sure we only set the search for online (github-pages) docs, and not for the offline documentation build. This either needs to be fixed in Travis or we need to wait for GitHub Actions.

Added a gh-pages only `config.toml` so the pages build is
`hugo ... --config gh-pages.toml,config.toml`

https://gohugo.io/getting-started/configuration/#configuration-file

PR #2591, opened 2020-05-25
Configure skipping payload crypto on application-level

johanstokking:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2596

Configure skipping payload crypto on application-level

#### Changes

- Deprecate `skip_payload_crypto` on the device level, replace by nullable `skip_payload_crypto_override`
- Introduce `skip_payload_crypto` on `ApplicationLink` level

#### Testing

Added missing unit tests for skip payload crypto in AS.

Also tested via CLI/simulation.

##### Regressions

Old usage of deprecated field could cause trouble, but that is covered by a unit test

#### Notes for Reviewers

@rvolosatovs please sanity-check the registry part.

#### Checklist


api/end_device.proto
```diff
@@ -547,9 +547,13 @@ message EndDevice {
   EndDeviceAuthenticationCode claim_authentication_code = 46;
 
   // Skip decryption of uplink payloads and encryption of downlink payloads.
+  // This field is deprecated, use skip_payload_crypto_override instead.
   bool skip_payload_crypto = 51;
+  // Skip decryption of uplink payloads and encryption of downlink payloads.
+  // This field overrides the application-level setting.
+  google.protobuf.BoolValue skip_payload_crypto_override = 52;
```
rvolosatovs, line 554: FYI, in MAC settings deprecated fields like these are renamed to `X_value`.
Since `skip_payload_crypto` is also there and also is an "override", I think we should just follow the same convention as MAC settings fields.

pkg/applicationserver/registry.go
```diff
@@ -33,6 +34,109 @@ type DeviceRegistry interface {
...
+		Old: "skip_payload_crypto",
+		New: "skip_payload_crypto_override",
+		GetTransform: func(dev *ttnpb.EndDevice) {
+			if dev.SkipPayloadCryptoOverride == nil && dev.SkipPayloadCrypto {
+				dev.SkipPayloadCryptoOverride = &pbtypes.BoolValue{Value: true}
+			} else {
+				dev.SkipPayloadCrypto = dev.SkipPayloadCryptoOverride.GetValue()
+			}
+		},
+		SetTransform: func(dev *ttnpb.EndDevice, _, _ bool) error {
```
rvolosatovs, line 133: This is not backwards compatible. For example, it breaks clients, which rely on setting `skip_payload_crypto` to `true`.
See https://github.com/TheThingsNetwork/lorawan-stack/blob/12d27d825e11b5db996cdfc8d42cb44aec7522ea/pkg/networkserver/registry.go#L284-L315 for an example of how this can be handled in backwards-compatible way.

Moreover, since `SkipPayloadCrypto` is deprecated, we should unset the field instead of setting it.

pkg/applicationserver/registry.go
```diff
@@ -33,6 +34,109 @@ type DeviceRegistry interface {
 	Set(ctx context.Context, ids ttnpb.EndDeviceIdentifiers, paths []string, f func(*ttnpb.EndDevice) (*ttnpb.EndDevice, []string, error)) (*ttnpb.EndDevice, error)
 }
 
+type deprecatedDeviceField struct {
```
rvolosatovs, line 37: Lets extract all of this into an internal package instead of directly copy-pasting from NS like this.
My suggestion is:
`pkg/internal/registry` with at least following names exported:
- `DeprecatedDeviceField`
- `DeprecatedDeviceFieldMatch`
- `MatchDeprecatedDeviceFields`

Ideally I think we should also extract utilities like `WrapDeviceGetWithDeprecatedFields` etc., but that can be done as second step.

api/end_device.proto
htdvisser: In MAC settings we only allowed because it was internal API. This changes external API so we can't just go and rename fields.

htdvisser: The new field needs to be added in `pkg/ttnpb/field_mask_validation.go`

rvolosatovs: I'm not suggesting to rename any fields. To clarify, what I mean is this:

```suggestion
  google.protobuf.BoolValue skip_payload_crypto_value = 52;
```

johanstokking: 1. Why would having both `skip_payload_crypto` and `skip_payload_crypto_value` be more clear than naming what it is, i.e. an override?
2. We would get `SkipPayloadCryptoValue.Value` which stutters
3. There is no reason why we would religiously follow previous deprecated field name renames in future field names

pkg/applicationserver/registry.go
```diff
@@ -33,6 +34,125 @@ type DeviceRegistry interface {
...
+			}
+		}
+		dev, paths, err := f(dev)
+		if err != nil || dev == nil {
+			return dev, paths, err
+		}
+		for _, d := range deprecated {
+			if ttnpb.HasAnyField(paths, d.Old) {
+				paths = ttnpb.AddFields(paths, d.New)
+			}
```
johanstokking, line 104: @rvolosatovs is there a reason this isn't in NS?

If a caller sets the old field only, this makes sure the new field is set as well, after the transform ran.

johanstokking: Added this to NS registry as well

pkg/networkserver/registry.go
```diff
@@ -45,116 +46,84 @@ func logRegistryRPCError(ctx context.Context, err error, msg string) {
...
 				d.GetTransform(dev)
 			}
 		}
 		dev, paths, err := f(ctx, dev)
 		if err != nil || dev == nil {
 			return dev, paths, err
 		}
-		for _, d := range deprecated {
+		for _, d := range replaced {
+			if ttnpb.HasAnyField(paths, d.Old) {
```
rvolosatovs, line 103: This is already handled by `registry.MatchReplacedEndDeviceFields`, remove

pkg/applicationserver/registry.go
rvolosatovs: Please remove, see NS comment.

pkg/applicationserver/registry.go
```diff
@@ -33,6 +35,90 @@ type DeviceRegistry interface {
...
+			} else {
+				dev.SkipPayloadCrypto = dev.SkipPayloadCryptoOverride.GetValue()
+			}
+		},
+		SetTransform: func(dev *ttnpb.EndDevice, useOld, useNew bool) error {
+			if useOld {
+				if useNew {
+					oldValue := dev.SkipPayloadCrypto
+					newValue := dev.SkipPayloadCryptoOverride
+					if oldValue && (newValue == nil || !newValue.Value) {
```
rvolosatovs, line 109: This does not cover case, when e.g.:
```
dev.SkipPayloadCrypto = false
dev.SkipPayloadCryptoOverride = { Value: true }
```

Which should cause an error as well, since both paths are specified by the user, but values are inconsistent.

Apart from that, not a fan of `if/else` nesting, a `switch` like in NS here would be clearer IMO (and more verbose)

pkg/applicationserver/registry.go
johanstokking: It's not; `paths` is set by `f()` (L63), which can be unrelated to the output of `MatchReplacedEndDeviceFields()` (L56)

pkg/networkserver/registry.go
johanstokking: Same; `paths` is from `f()`

pkg/applicationserver/registry.go
rvolosatovs: Ah, right. Yeah, indeed that's not the case.
Still, this should not be there - "old" fields should not be used by implementations of servers, only by the clients. The path set returned here should always be either completely generated by the server or processed from user input by the server.

rvolosatovs: Yeah, you're right.

PR #2609, opened 2020-05-27
Add prompt component

bafonins:
#### Summary

This PR introduces a new `<Prompt />` component.

#### Changes

- Add `<Prompt />` component
- Add storybook entry
- Change storybook configuration to render `div` with `modal-container` attribute to allow rendering portalled modals.

#### Testing

Storybbok

##### Regressions

none

#### Notes for Reviewers

This component can be helpful when the user accidentally navigates from a page with half-filled form. Adding this because of https://github.com/TheThingsNetwork/lorawan-stack/issues/579

#### Checklist

### DIFF
--- a/pkg/webui/components/prompt/index.js
+++ b/pkg/webui/components/prompt/index.js
@@ -1,0 +1,89 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import React from 'react'
+import { Prompt as RouterPrompt } from 'react-router-dom'
+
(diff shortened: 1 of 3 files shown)


bafonins:
>  I'd prefer to use the browser's default behavior here, using `window.beforeunload()`

Im aware of the `beforeunload` event. There are problems that this component solves. From https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event:

No ability to provide custom message.

> Some browsers used to display the returned string in the confirmation dialog, enabling the event handler to display a custom message to the user. However, this is deprecated and no longer supported in most browsers.

The user can block `beforeunload` popups.

> To combat unwanted pop-ups, browsers may not display prompts created in beforeunload event handlers unless the page has been interacted with, or may even not display them at all.

This is indeed the case.

> I could imagine that this technique might actually not work because of react-router.

Additionally, as far as I know there is no way to determine the users response and hence executing any actions is not possible which can be helpful (think of performing rollbakcs, serializing current state to `localStorage`, etc).

Personally, I found out that combination of using `Prompt` for spa navigation + `beforeunload` to intercept page reload/close gives the best result.

pkg/webui/components/prompt/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+import React from 'react'
+import { Prompt as RouterPrompt } from 'react-router-dom'
+
+import PortalledModal from '@ttn-lw/components/modal/portalled'
+
+import PropTypes from '@ttn-lw/lib/prop-types'
+
+/*
+ * `<Propmt />` is used to prompt the user before navigating from the current page. This is
```
kschiffer, line 23:
```suggestion
 * `<Prompt />` is used to prompt the user before navigating from the current page. This is
```

Issue #2119, opened 2020-03-10
External Join Server user experience


johanstokking:
#### Summary

External Join Server user experience

cc @laurensslats 

#### Why do we need this?

To make improve the user experience of creating and editing devices on a Join Server.

#### What is already there? What do you see now?

There's currently an inconsistency:

1. In a regional cluster, "External Join Server" equals empty `join_server_address`.
2. However, when using TTI Join Sever, the device should be managed there. That only works when `join_server_address` is the local host name. Because of point 1, the address is empty, so this doesn't work

#### What is missing? What do you want to see?

From a user experience perspective, there are three states:

1. Using the cluster Join Server
   - `join_server_address` is the cluster address
2. Using the device already on a global join server (i.e. TTI Join Server)
   - device is already there
   - Console is used to specify NS and AS settings
   - `join_server_address` does not change, but is set
3. Using a global Join Server with unknown address
   - `join_server_address` is empty

For option 2 and 3, we probably want NS to serve a mapping of JoinEUI to Join Server address, so Console can show a drop down and pre-fill the JoinEUI and Join Server address. cc @rvolosatovs this comes from interop, and probably needs an address field in the JS interop config. If this can be done as part of closing this issue in this milestone, I'm good with that, otherwise for 2 we need to stick with manually entering the JS address.

#### Can you do this yourself and submit a Pull Request?

Can review behavior

rvolosatovs:
@johanstokking can you clarify what `join_server_address` is for at all?
> For option 2 and 3, we probably want NS to serve a mapping of JoinEUI to Join Server address, so Console can show a drop down and pre-fill the JoinEUI and Join Server address
`JoinEUI`s uniquely identify the Join Server, what is the use case for *prefilling* or *choosing* the `JoinEUI`? That data is already chosen by the device manufacturer and the user should not alter that. Do we want to provide a way for the user to override LoRa Alliance DNS and cluster configuration for joins?

What I understand from
> There's currently an inconsistency:
>
> 1. In a regional cluster, "External Join Server" equals empty `join_server_address`.
> 2. However, when using TTI Join Sever, the device should be managed there. That only works when  `join_server_address` is the local host name. Because of point 1, the address is empty, so this doesn't work

Is that the issue is specific TTI Join Server and "External Join Server" checkbox is confusing there.
[...]

Issue #2593, opened 2020-05-26
Uncheck External Join Server


laurensslats:
#### Summary
External Join Server - unclear why this is enabled by default.

Linked to: #2529

![image](https://user-images.githubusercontent.com/17290639/82881919-9da85700-9f40-11ea-948e-f2d8399acaab.png)

...

#### Why do we need this?
Unclear to people how to configure a device to a 3rd party join server
...

#### What is already there? What do you see now?

...

#### What is missing? What do you want to see?
Uncheck the box by default.
...

#### Environment
v3.8.1+

...

#### How do you propose to implement this?
Ask our console-master Bogdans
...

#### How do you propose to test this?
- Login to the stack
- Create a new device 
- Check that the "External Join Server" is not checked by default in an old application
- Check that the "External Join Server" is not checked by default in a new application
- [Regression] Test that checking the box still works as before (i.e., requests are routed to an external join server).

...

KrishnaIyer:
@laurensslats: 
```
How do you propose to test this?
Staging environment?
```

This section is actually **how** to test and not **where**. We expect a test scenario here. If that's not clear in the template, please let us know and we can update that.

Consider the following;
```
How do you propose to test this?
- Login to the stack
- Create a new device 
- Check that the "External Join Server" is not checked by default.
- [Optional: Regression] Test that checking the box still works as before (i.e., requests are routed to an external join server).
```

bafonins:
From https://github.com/TheThingsNetwork/lorawan-stack/issues/579#issuecomment-553347858:

> Checkbox whether to use external JS (default checked)

cc @johanstokking

johanstokking:
@bafonins Indeed, I changed my mind!

Issue #1567, opened 2019-11-07
Add more fine-grained feature toggles e.g. for JS only deployments


kschiffer:
#### Summary
We need more fine-grained feature toggles for JS only deployments to ensure proper UX for the console on such deployments.

#### Why do we need this?
On JS only deployments, there will be a lot of functionalities not working in the console, so these should be properly hidden or disabled.

#### What is already there? What do you see now?
Feature toggles via #1543 

#### What is missing? What do you want to see?
More fine-grained toggles for JS only deployments (and maybe other special stack component set ups)

#### Environment
Browser.

#### How do you propose to implement this?
Once I know the exact requirements for individual features of the console, it is easy to add these to the feature checks. @htdvisser, please assist me with defining these requirements.

#### Can you do this yourself and submit a Pull Request?
Yes.

htdvisser:
With the latest additions in https://github.com/TheThingsNetwork/lorawan-stack/issues/579 I think this issue is now also covered.

rvolosatovs:
See https://github.com/TheThingsNetwork/lorawan-stack/issues/2119#issuecomment-597736420
> It does not make sense to allow creating ABP devices here, nor any NS/AS settings (MAC versions, payload formats, etc), nor choosing for "external JS" (because this JS is already "external" because there is no NS and AS enabled).

johanstokking:
@laurensslats can you describe the user flow here?

bafonins:
I checked various parts of the Console in IS+JS only setup. The only problem I was able to find (having https://github.com/TheThingsNetwork/lorawan-stack/pull/2644 in place) is that the users are not able to register new gateways. https://github.com/TheThingsNetwork/lorawan-stack/pull/2785 fixes this and resolves this issue.

Issue #849, opened 2019-06-25
Multicast device support in console


rvolosatovs:
#### Summary

![](https://i.imgur.com/LfY8FNi.png)
Multicast support to Network Server was added in https://github.com/TheThingsNetwork/lorawan-stack/pull/792
Console should allow creation of multicast devices.
From the perspective of console a multicast device can be seen as just an ABP device, which does not send uplinks.
The `multicast` field on `EndDevice` proto message https://github.com/TheThingsNetwork/lorawan-stack/blob/576cac2962c6f67d2fb9de80afa8a7f7b10deb8f/api/end_device.proto#L473 has to be set to `true`(`false` by default) on creation of the device in order for the it to correctly operate as a part of multicast group.
The value cannot be changed after the device creation.

#### Why do we need this?

To provide a nice UI for our users, who want to create multicast devices

#### How do you propose to implement this?

Maybe just have a *multicast* checkbox?
Note, that OTAA devices cannot be multicast by definition.

#### Can you do this yourself and submit a Pull Request?

I'm sure @bafonins or @kschiffer will be better suited for the job.

johanstokking:
I think UX wise multicast is like ABP; it's a fixed session.

We could have three modes to create a device:
- OTAA
- ABP
- Multicast (which is like ABP but sets `multicast`)

bafonins:
Something like this?

<img width="522" alt="Screenshot 2019-07-01 at 10 48 02" src="https://user-images.githubusercontent.com/16374166/60423102-bb185800-9bed-11e9-8e51-e45011b5da5e.png">

rvolosatovs:
Yes!
Please remove/disable the *Resets Frame Counters*, since no uplinks can be sent

PR #2644, opened 2020-06-02
Use a wizard for creating end devices

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/579, https://github.com/TheThingsNetwork/lorawan-stack/issues/1567
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2119, https://github.com/TheThingsNetwork/lorawan-stack/issues/849, https://github.com/TheThingsNetwork/lorawan-stack/issues/2593

<details>

Screenshot 2020-06-03 at 18 32 43

Screenshot 2020-06-04 at 15 49 38

Screenshot 2020-06-04 at 15 49 49

Screenshot 2020-06-04 at 15 50 01

Screenshot 2020-06-04 at 15 50 28

Screenshot 2020-06-04 at 15 50 37

Screenshot 2020-06-04 at 15 50 46

Screenshot 2020-06-04 at 15 51 04

Screenshot 2020-06-04 at 15 51 20

Screenshot 2020-06-04 at 15 53 22

Screenshot 2020-06-04 at 15 53 32

Screenshot 2020-06-04 at 15 56 30

</details>

#### Changes

- Add `<Wizard />` component
- Refactor the device add view to use the device wizard
- Add a bunch of helper react hooks
- Add `skip_payload_crypto` field to the AS settings form, other that that the new device wizard should be functionally equivalent to the current device form.

#### Testing

1. Checked prompt in safari/chrome/firefox.
2. Create various devices through the device wizard.
3. Created OTAA devices in the JS only deployment setup.

##### Regressions

Device creation

#### Notes for Reviewers

There are couple of design decisions I had to make that differ from what we discussed in https://github.com/TheThingsNetwork/lorawan-stack/issues/579:

1. Dynamically showing/hiding steps based on user selection provides pretty weird ux. Because of that I decided to make a separate configuration form that requires the user to provide base end device data to show wizard steps that are required for this particular end device type.
2. Create end device only after completing the last step. This helps avoiding creating incomplete end devices, since `beforeunload` handler does not guarantee that any rollback function will be executed closing the browser tab.
3. Wizard steps are not reflected in the browsers url (`/steps/basic`, `/steps/network`, etc). I dont really see any reason for this atm, however this can be implemented by extending the `<Wizard />` component.
4. If NS and JS are not in the cluster we cannot really register device with a particular activation mode. I decided that should still allow the users to create end devices. In that case, end devices with the `none` activation mode will be registered in the IS.

Current implementation still has some rough edges, will fix that once we agree on appropriate UI.

#### Checklist


johanstokking:
Trying to test this, but I still see the old create screen, am I doing something wrong?

bafonins:
Updated screenshots.

> Reorder the activation mode; OTAA first, then ABP, then multicast, and then "Do not configure activation" (instead of "None")

> AppEUI is shown when I select 1.0.4; it should be JoinEUI for >= 1.0.4

> The Join Settings are grayed out in the wizard header

fixed

> Home NetID can be optional
> The Home NetID, Application Server ID, Application Server KEK Label and Network Server KEK Label are shown under "Network server settings" are all optional, I think it's good to mention that

those are in the `Join Settings` and the fields you mentioned are actually optional.

Screenshot 2020-06-03 at 18 32 43

> I think it's good to mention that

There is an issue about hiding these fields, see https://github.com/TheThingsNetwork/lorawan-stack/issues/2445

> Are we getting an "empty AppEUI/JoinEUI" button here to set zeros?

After we decide how to proceed with https://github.com/TheThingsNetwork/lorawan-stack/issues/2542. I guess functionality to set zeros will be implemented in the component discussed in #2542

bafonins:
> @laurensslats please list all sensible device creation options so we can test it

if this is required I can also write a bunch of e2e tests to verify that creation of all device types is possible.

johanstokking:
> those are in the `Join Settings` and the fields you mentioned are actually optional.

Yes that's good, but it helps if that's clear, so that people are not overwhelmed by these fields and not know what to do. But I'm OK with #2445 also.





> After we decide how to proceed with #2542. I guess functionality to set zeros will be implemented in the component discussed in #2542

OK

kschiffer:
Very good!

Couple of initial remarks from testing this in the browser, forgive me if you already planned to polish some of those:

##### General

*   The `Next` / `Previous` button should have the same caption as the title of the next/previous step (e.g. \`Network layer settings\`)
    *   use `keyboard_arrow_right` icon after the caption and \``keyboard_arrow_right`\` before the caption of the `Previous` button
    *   The `Complete` button should be `Add end device`
*   The step titles do not use sentence case (\`Network Layer Settings\` vs `Network layer settings`)
*   The sub titles of the step titles should be more meaningful, they currently mostly repeat the step title. I think what could work is enumerating the most important fields of the step, e.g. `End device ID, DevEUI, Name and Description`
*   If possible, disabled fields should provide info as to why they are disabled (e.g. `LoRaWAN Version` for ABP devices)
*   We should take the opportunity and review all of the field descriptions as well
    *   @benolayinka can you go through the various steps and suggest field descriptions, as well as references to relevant documentation
    *   @pgalic96 can you look into ways to link to the documentation from within field descriptions; I'm thinking of a little :information\_source: info icon next to the description text
*   Clicking the generate button of byte fields will not mark the field as touched, meaning that validation messages persist
*   I know this is likely not trivial, but it would be very nice if submit errors would make the wizard jump to the step with the field that the respective error is associated with
*   When setting `Do not configure activation`, the next step looks weird; there should be at least a form title then
    *   Even better would be the whole process would then become a regular single-step creation, like with our other entities. Consider this gold-plating. Using a form title for the step is fine as well.

##### Initial step

*   I think OTAA mode is a better default
*   I understand we cannot assume any LoRaWAN version as sane default?
*   I think the `Application Server Address` and `Network Server Address` need some inline documentation, essentially reassuring that the default value should be good in most cases (cc @benolayinka )
*   The heading `Configure new end device` should rather be `Add new end device` and persist throughout the process (so also in the wizard phase)
*   Add a sub-title to the initial form: `Preparation`

##### Basic settings

*   `Optional device description;`  -> `Optional end device description; can`

##### Network layer settings

*   Clicking the generate button on `Device address` (without touching it before) will show a validation error; seems to be an issue of this particular field only

##### Application layer settings

*   `Skip payload crypto` -> `Skip payload encoding`

Not suggesting that these all have to be fixed in this PR, but would be cool to get in as many of those as possible and otherwise following up with issues. All that being said, great work. This is a huge step forward for the end device creation UX.

I'll follow up with code review.

johanstokking:
> * I think OTAA mode is a better default

Agreed

> * I understand we cannot assume any LoRaWAN version as sane default?

Unfortunately not, no

> * The heading `Configure new end device` should rather be `Add new end device` and persist throughout the process (so also in the wizard phase)

Agreed

> * `Skip payload crypto` -> `Skip payload encoding`

It's crypto, not encoding. You can say "Skip payload encryption and decryption" alternatively. We probably need an info field here:

_If enabled, Application Server does not decrypt uplink payload and encrypt downlink payload. Instead, the upstream application-layer needs to do the LoRaWAN cryptographic operations with the AppSKey (wrapped with the Application Server KEK). This enables end-to-end encryption_

Maybe too long?

benolayinka:
This looks awesome!!

>   * @benolayinka can you go through the various steps and suggest field descriptions, as well as references to relevant documentation

@bafonins is it possible to do tooltips to the glossary? -> https://thethingsstack.io/v3.8.2/reference/glossary/

Individual terms are e.g https://thethingsstack.io/v3.8.2/reference/glossary#activation-by-personalisation

> * I think the `Application Server Address` and `Network Server Address` need some inline documentation, essentially reassuring that the default value should be good in most cases (cc @benolayinka )

Something like: 
"The address of the Application Server the end device will communicate with. Unless you have configured a custom address for this component, leave as default."

kschiffer:
> It's crypto, not encoding. You can say "Skip payload encryption and decryption" alternatively. We probably need an info field here:

:ok\_hand:  I would indeed prefer `Skip payload encryption and decryption` then.

> _If enabled, Application Server does not decrypt uplink payload and encrypt downlink payload. Instead, the upstream application-layer needs to do the LoRaWAN cryptographic operations with the AppSKey (wrapped with the Application Server KEK). This enables end-to-end encryption_

That's indeed a bit too long. We should use a short description and reference documentation instead.

> [**@bafonins**](https://github.com/bafonins) is it possible to do tooltips to the glossary? -> [https://thethingsstack.io/v3.8.2/reference/glossary/](https://thethingsstack.io/v3.8.2/reference/glossary/)

Not possible atm but let's indeed do that. Not in this PR though, this needs a separate issue and PR.

> "The address of the Application Server the end device will communicate with. Unless you have configured a custom address for this component, leave as default."

Sounds good. I indeed wonder how common it is to configure custom addresses. If it is less than 20% then I suggest to additionally hide these two settings behind a `Use custom component server hosts`\-checkbox. cc @johanstokking ? This would be in line with our [progressive disclosure principle](https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/DESIGN.md#progressive-disclosure).

bafonins:
Updated screenshots in the PR descriptions

Fixed:

[x] The `Next` / `Previous` button should have the same caption as the title of the next/previous step (e.g. \`Network layer settings\`)
[x] use `keyboard_arrow_right` icon after the caption and \``keyboard_arrow_right`\` before the caption of the `Previous` button
[x]  The `Complete` button should be `Add end device`
[x]  The step titles do not use sentence case (\`Network Layer Settings\` vs `Network layer settings`)
[x]  The sub titles of the step titles should be more meaningful, they currently mostly repeat the step title. I think what could work is enumerating the most important fields of the step, e.g. `End device ID, DevEUI, Name and Description`
[x] When setting `Do not configure activation`, the next step looks weird; there should be at least a form title then
[x] I think OTAA mode is a better default
[x] The heading `Configure new end device` should rather be `Add new end device` and persist throughout the process (so also in the wizard phase)
[x] `Optional device description;`  -> `Optional end device description; can`
[x] `Skip payload crypto` -> `Skip payload encoding`
[x] Add a sub-title to the initial form: `Preparation`

TODO:

1. Clicking the generate button of byte fields will not mark the field as touched, meaning that validation messages persist
2. It would be very nice if submit errors would make the wizard jump

pkg/webui/console/views/device-add/wizard/prompt/index.js
```diff
@@ -0,0 +1,81 @@
...
+
+import sharedMessages from '@ttn-lw/lib/shared-messages'
+import PropTypes from '@ttn-lw/lib/prop-types'
+import useBeforeUnload from '@ttn-lw/lib/hooks/use-before-unload'
+
+const m = defineMessages({
+  modalTitle: 'Abort end device creation',
+  modalMessage:
+    'Are you sure you want to abort creation of this end device? Unsaved changes in your forms will be lost.',
+  modalApproveText: 'Dismiss and proceed',
```
kschiffer, line 30:
```suggestion
  modalTitle: 'Confirm navigation',
  modalMessage:
    'Are you sure you want to leave this page and abort the end device creation? There are unsaved changes that will be lost.',
  modalApproveText: 'Confirm navigation',
```

Sorry, had to change my mind a little about the message in this particular case.

Additionally, the cancel-message should say `Stay on this page`.

pkg/webui/lib/hooks/use-before-unload.js
```diff
@@ -0,0 +1,50 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import React from 'react'
+
+const useBeforeUnload = (handler = () => {}) => {
```
kschiffer, line 17: JsDoc would be nice here.

pkg/webui/lib/render-callback.js
```diff
@@ -12,6 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-.title
-  text-margin-top($ls.m)
-  text-margin-bottom($ls.m)
+const renderCallback = ({ render, children }, context) => {
```
kschiffer, line 15: Please add JsDoc, it took me a bit to understand the usage in the bigger context.

kschiffer: Also note that the header says `Copyright 2019`.

pkg/webui/console/views/device-add/wizard/index.js
```diff
@@ -0,0 +1,193 @@
...
+        getJoinServerAddress(snapshot) === getHostnameFromUrl(jsConfig.base_url) &&
+        activationMode === ACTIVATION_MODES.OTAA
+
+      return (
+        <>
+          <Prompt
+            when={dirty && !completed}
+            shouldBlockNavigation={handleBlockNavigation}
+            onApprove={handlePromptApprove}
+          />
```
kschiffer, line 111: This prompting could really be a default feature of the wizard component. I can't think of any scenario where we don't want to notify users of unsaved changes.

It would also take away some of the complexity here.

johanstokking:
@bafonins please re-request review when ready for another round

pkg/webui/components/wizard/index.js
```diff
@@ -0,0 +1,31 @@
...
+
+import Wizard from './wizard'
+import Steps from './steps'
+import Step from './step'
+import Stepper from './stepper'
+import Form from './form'
+import { useWizardContext } from './context'
+
+Wizard.Steps = Steps
+Wizard.Steps.displayName = 'Wizard.Steps'
```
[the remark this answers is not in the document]
bafonins: Visually group compound components.

```
Wizard
  Stepper
    Step
  Steps
    Step
      Form
  
```

vs

```
Wizard
  Wizard.Stepper
    Stepper.Step
  Wizard.Steps
    Wizard.Step
      Wizard.Form
```

Screenshot 2020-06-05 at 15 43 52

pkg/webui/components/wizard/steps.js
```diff
@@ -0,0 +1,55 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import React from 'react'
+
+import PropTypes from '@ttn-lw/lib/prop-types'
+
+import { WizardContext } from './context'
+
+class Steps extends React.Component {
```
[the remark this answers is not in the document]
bafonins: To ensure that `initSteps` is called only once. For me, it seems that class component looks simpler to understand this, however I refactored into a functional component.

pkg/webui/console/views/device-add/wizard/index.js
```diff
@@ -0,0 +1,193 @@
...
+  )
+
+  const handlePromptApprove = React.useCallback(
+    location => {
+      rollbackProgress(location.pathname)
+    },
+    [rollbackProgress],
+  )
+
+  return React.createElement(
```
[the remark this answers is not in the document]
bafonins: > This whole way of rendering without JSX and hooking into the render logic of react components needs more commenting

What does `hooking into the render logic of react components` mean?
`React.createElement` is in the public API and intellisense, or https://reactjs.org/docs/react-api.html#createelement give enough information.

Can't you use a render prop instead, which will be called by the `<Wizard />` -component with the props as argument?

Ofc, but `createElement` allows nicely separating js logic from jsx

bafonins:
> It's impossible to review +4000 lines of code in detail unless I have a week to spare

I can split the PR into several. There are a bunch of fixes/adjustments to existing components + ~1000loc for tests.
Additionally, I can write a set of disposable e2e tests that we can later transfer to cypress or whatever tool we decide to use.

> what made you choose to duplicate all step forms instead of extending and reusing the forms that we already have from the device settings. There is a lot of duplication now between the two. I feel like for the sake of maintainability, the create and update forms should really be tied closely together.

1. I prefer code duplication over introducing weird and most probably leaky abstractions just for the sake reusing some repetitive parts of the code.
2. I dont want to introduce regressions to device general settings.
3. It is not yet known how https://github.com/TheThingsNetwork/lorawan-stack/issues/2047, https://github.com/TheThingsNetwork/lorawan-stack/issues/1748, https://github.com/TheThingsNetwork/lorawan-stack/issues/263 and possibly other features will be implemented

johanstokking:
I'm repeatedly getting:

```
Error: running "node_modules/.bin/webpack --config config/webpack.config.babel.js" failed with exit code 2
```

I already did `mage js:clean js:cleanDeps jsSDK:clean jsSDK:cleanDeps`. I'm getting this on `mage js:build`

pkg/webui/console/views/device-add/wizard/index.js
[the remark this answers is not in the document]
kschiffer: > What does `hooking into the render logic of react components` mean?

That was actually referring to the `renderCallback` function usage, in addition to the issue brought up here. Should have been clearer there.

I feel like using `React.createElement` directly is a quite unusual approach to define React elements, making the code unnecessarily hard to follow.

```
return (
  <Wizard
    initialValues={configuration}
    onComplete={handleWizardComplete}
    completeMessage={sharedMessage.addDevice}
  >
  {
    ({snapshot, dirty}) => { … }
  }
  </Wizard>
```

Will do exactly the same (unless I miss something), while retaining the familiar JSX syntax.

pkg/webui/lib/render-callback.js
```diff
@@ -0,0 +1,35 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Allows react components to use the render prop design pattern without explicitly
+ * specifying `render`, `children` and `children` as a function props.
+ *
+ * @param {object} param0 - = The props of a component.
```
kschiffer, line 19:
```suggestion
 * @param {object} props - The props of a component.
```

pkg/webui/lib/render-callback.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param {object} context - The context to be bassed to children.
```
kschiffer, line 20:
```suggestion
 * @param {object} context - The context to be passed to children.
```

kschiffer:
> I can split the PR into several. There are a bunch of fixes/adjustments to existing components + ~1000loc for tests.

Please don't split any further. I think I have sufficiently reviewed everything now so now it would just duplicate my work for reviewing.

> Additionally, I can write a set of disposable e2e tests that we can later transfer to cypress or whatever tool we decide to use.

Yeah, that would definitely be very useful. Let's do this separately after we decide on e2e.

> 1.  I prefer code duplication over introducing weird and most probably leaky abstractions just for the sake reusing some repetitive parts of the code.
> 2.  I dont want to introduce regressions to device general settings.
> 3.  It is not yet known how #2047, #1748, #263 and possibly other features will be implemented
> 4.  Initial implementation of the `<Wizard />` component included only a single `<Form />` component for all steps, but such solution turned out to be more verbose.

I still have concerns but trust you to assess this better than me.

bafonins:
@johanstokking Indeed, the build was broken. Now it should be fixed.

@kschiffer regarding:

> Clicking the generate button of byte fields will not mark the field as touched, meaning that validation messages persist

This is regression from updating `formik` to v2. Since v2 all helper functions are async and with the current implementation we miss a tick in the event loop. I will fix in a separate pr as the issue is already in master.

> It would be very nice if submit errors would make the wizard jump

Will do in a separate pr. This might require changes to the sdk or more code in the wizard itself. Dont want to make this PR bigger.

bafonins:
> Concerning the component hosts in the preparation step: Since there was no reaction to my question I suggest that we indeed hide the component host configuration behind a checkbox saying Use default hosts for component servers (wording suggestions welcome), being checked by default. I feel like showing these options by default will disrupt novice users unnecessarily, given that it will be quite unusual to make changes to these fields

> Also this makes me realize that Application Server address and similar should actually be Application Server host, to be more accurate. cc @johanstokking 

this issue is not specific to the device creation, but also references other parts of the Console. I created a separate issue to address those, see https://github.com/TheThingsNetwork/lorawan-stack/issues/2694

> It's not possible to create devices without activation, getting invalid field mask error

Fixed. We strip NS related fields when no configuration is selected.

> Small styling tweak: please apply a little more spacing between the stepper and the form

Fixed

johanstokking:
@laurensslats can you test key scenarios too?

benolayinka:
Are LoRaWAN versions linked to Regional Parameters versions? So if the LoRaWAN version is 1.0.2, can we limit Regional Parameters to 1.0.2A and 1.0.2B? @johanstokking @laurensslats

laurensslats:
> Are LoRaWAN versions linked to Regional Parameters versions? So if the LoRaWAN version is 1.0.2, can we limit Regional Parameters to 1.0.2A and 1.0.2B? @johanstokking @laurensslats

Yes, this is standard practice to align LoRaWAN version & RP. There might however be some cases where this is not aligned.

johanstokking:
This should be aligned indeed. We can restrict input that way. Good point.

laurensslats:
2. **Progress bar looks ugly**
Remove the description (it's really small and difficult to read). Increase the padding
Screenshot 2020-06-17 at 17 45 17

laurensslats:
3. **Hide advanced Join Server settings**
(Almost) no-one understands this, we need to hide it from normal users.

Screenshot 2020-06-17 at 17 53 34

benolayinka:
> This looks really great!
> 
> Let's make sure to add the descriptive fields (maybe they get replaced by the tooltips #2663 at some point).
> 
> 1. **Add descriptive fields:**
> 
> LoRaWAN Version
> The LoRaWAN version (MAC), as provided by the device manufacturer
> 
> Network Server Address
> The address of the Network Server the end device will communicate with. Leave as default unless you configured a custom address
> 
> Application Server Address
> Leave as default unless you configured a custom address
> 
> External Join Server
> Select if the device is pre-registered on an external join server
> 
> Join Server Address
> Leave as default unless you configured a custom address
> 
> Screenshot 2020-06-17 at 17 35 57

Let's do these as tooltips on hover. Otherwise, we have to provide a small amount of additional info in the field text which is hardcoded in the console, and a large amount of additional info in the tooltip which is linked to the glossary.

laurensslats:
> Let's do these as tooltips on hover. Otherwise, we have to provide a small amount of additional info in the field text which is hardcoded in the console, and a large amount of additional info in the tooltip which is linked to the glossary.

Ok agreed, on the condition that we deploy the tooltips at the same time as the device wizard, otherwise we only make things more complicated. @bafonins.

kschiffer:
Sorry, we have some overlap of concerns now. @laurensslats there is an [issue](https://github.com/TheThingsNetwork/lorawan-stack/issues/2664) already about the documentation strategy and @benolayinka has already posted a lot of input in there. I've assigned you as well now. We need to agree here on the exact plan before proceeding. As such, I would really recommend to decouple that from this PR, since it is already doing A LOT.

Maybe it's an idea to target merging this PR early after the next release, so we have time to separately add improvements to the form documentation to be shipped with the same release thereafter. @bafonins agreed?

I honestly also think that this is already a big improvement to the current end device creation process regardless of the documentation.

> Remove the description (it's really small, difficult to read and doesn't look too good).

Seems to be subjective, I don't agree.

> *   Increase the left and right padding of the thin line
> *   Make width 100%. Align the right side of the progress bar with the rest of the page.
> *   Move text "Basic settings" "Network layer settings" and "Join settings" 2 px down to align with circles

Agreed.

bafonins:
@laurensslats Please read the PR description. It clearly states the issues this PR addresses. As it is mentioned in the PR and discussions in https://github.com/TheThingsNetwork/lorawan-stack/issues/579, this PR mainly addresses *technical* problems with the current device create form since it fails to support a bunch of use cases that we must support.

1. Add descriptive fields:

> The LoRaWAN version (MAC), as provided by the device manufacturer

Will fix

regarding the stack component addresses, there is a separate issue for this: https://github.com/TheThingsNetwork/lorawan-stack/issues/2694

2. Improve design progress bar

> Remove the description (it's really small, difficult to read and doesn't look too good).

I can remove this, but IMO it clearly defines the purpose of the wizard step. @kschiffer 

Will try ti fix the rest of the styling issues.

3. Hide advanced Join Server settings

> Relates to #2445. This is one of the biggest issues because (almost) no-one understands these fields and it only creates confusion.

Will be fixes in https://github.com/TheThingsNetwork/lorawan-stack/issues/2445

> Maybe it's an idea to target merging this PR early after the next release, so we have time to separately add improvements to the form documentation to be shipped with the same release thereafter. @bafonins agreed?

Sure

laurensslats:
> Remove the description (it's really small, difficult to read and doesn't look too good).
OK, let's not block the PR because of this. Let's deploy and involve @pierrephz afterwards to (optionally) improve the design in a next release.

> Maybe it's an idea to target merging this PR early after the next release, so we have time to separately add improvements to the form documentation to be shipped with the same release thereafter. @bafonins agreed?
Also agreed. Let's prioritize #2445, #2575 and #2694

johanstokking:
@laurensslats please approve as your earlier request for changes is blocking merge.

> Let's deploy and involve @pierrephz afterwards to (optionally) improve the design in a next release.

Please file follow-up issue for this.

kschiffer:
> I can remove this, but IMO it clearly defines the purpose of the wizard step. [**@kschiffer**](https://github.com/kschiffer) 

As said in my other comment. I also disagree with removing this. It's true though that the size makes it somewhat hard to read depending on the screen resolution. I can have a look at that in a separate overall styling tweak PR.

bafonins:
@laurensslats 

Fixed the issues you mentioned.

>> Are LoRaWAN versions linked to Regional Parameters versions? So if the LoRaWAN version is 1.0.2, can we limit Regional Parameters to 1.0.2A and 1.0.2B?

> You didn't include this. Either include it or make a new issue for it. I'm fine with both.

I created a separate issue for this: https://github.com/TheThingsNetwork/lorawan-stack/issues/2784



>> Let's do these as tooltips on hover. Otherwise, we have to provide a small amount of additional info in the field text which is hardcoded in the console, and a large amount of additional info in the tooltip which is linked to the glossary.

> See review Ben. I'm OK with keeping the hardcoded descriptions if we remove it again when adding the tooltip. Up to you.

There are issues regarding extending input component (https://github.com/TheThingsNetwork/lorawan-stack/issues/2259) and adding tooltips (https://github.com/TheThingsNetwork/lorawan-stack/issues/2663)

PR #2647, opened 2020-06-02
Various pub/sub improvements

johanstokking:
#### Summary

Groundwork for https://github.com/TheThingsIndustries/lorawan-stack/issues/2165

#### Changes

- Reduce log level of already started integration. TTS was printing warnings in a normal flow; when the task scheduler was restarting a failed task, updating the pub/sub might fix that. The update itself also starts a task, so that there are effectively two tasks started. That's fine, but it's not worth warning for
- Clear AppSKey conditionally in a join-accept. This is useful to send the AppSKey upstream when skipping payload crypto
- On each join-accept, there's an invalidated downlink queue. When it is empty, however, there's no reason to recalculate it and replace it on NS. The queue is already gone in NS. This saves an unnecessary rpc on each join-accept

#### Testing

Tested locally with simulated devices

##### Regressions

Not aware of any

#### Checklist

### DIFF
--- a/pkg/applicationserver/applicationserver.go
+++ b/pkg/applicationserver/applicationserver.go
@@ -542,6 +542,9 @@ func (as *ApplicationServer) handleUp(ctx context.Context, up *ttnpb.Application
 
 var errFetchAppSKey = errors.Define("app_s_key", "failed to get AppSKey")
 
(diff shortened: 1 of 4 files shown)


johanstokking:
Bump @adriansmares

PR #2657, opened 2020-06-03
Edit JSON uplink message doc for clarity

### LINKED ISSUE #2293: Uplink message documentation comments unclear about gateway time and timestamp
#### Summary

The time and timestamp field inside the `uplink_message`=>`settings` object is non clearly explained. Especially the difference between these and the time and timestamp fields that are provided per gateway.

#### Why do we need this ?

Remove any doubt.

Comments that are unclear.

I want to know what is the difference between the time and timestamps that are provided per gateway, versus the time and timestamp that is provided in the settings object.

Fix the comment.

Unfortunately not because the situation is unclear to me.

See:
https://github.com/TheThingsNetwork/lorawan-stack/blame/master/doc/content/reference/application-server-data/data-formats/_index.md#L103

And compare that to:
https://github.com/TheThingsNetwork/lorawan-stack/blame/master/doc/content/reference/application-server-data/data-formats/_index.md#L120

benolayinka:
#### Summary

Closes #2293 

#### Changes

- ~~Clarify documentation for JSON Uplink Message timestamps.~~
- Remove time + timestamp from JSON uplink rx field

#### Testing

...

##### Regressions

...

#### Notes for Reviewers

...

#### Checklist

### DIFF
--- a/doc/content/integrations/data-formats/_index.md
+++ b/doc/content/integrations/data-formats/_index.md
@@ -116,8 +116,6 @@ The JSON uplink messages use the following format:
       "data_rate_index": 5,                  // LoRaWAN data rate index
       "coding_rate": "4/6",                  // LoRa coding rate
       "frequency": "868300000",              // Frequency (Hz)
-      "timestamp": 2463457000,               // Timestamp of the gateway concentrator when the accepted uplink message has been received
-      "time": "2020-02-12T15:15:45.787Z"     // ISO 8601 UTC timestamp at which the accepted uplink has been received by the gateway
     },
     "received_at": "2020-02-12T15:15..."     // ISO 8601 UTC timestamp at which the uplink has been received by the Network Server
   }
--- a/doc/content/integrations/data-formats/_index.md
+++ b/doc/content/integrations/data-formats/_index.md
@@ -116,8 +116,8 @@ The JSON uplink messages use the following format:
       "data_rate_index": 5,                  // LoRaWAN data rate index
       "coding_rate": "4/6",                  // LoRa coding rate
       "frequency": "868300000",              // Frequency (Hz)
-      "timestamp": 2463457000,               // Timestamp of the gateway concentrator when the message has been received
-      "time": "2020-02-12T15:15:45.787Z"     // ISO 8601 UTC timestamp at which the uplink has been received by the gateway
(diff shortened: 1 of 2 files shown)


doc/content/integrations/data-formats/_index.md
```diff
@@ -116,8 +116,8 @@ The JSON uplink messages use the following format:
       "data_rate_index": 5,                  // LoRaWAN data rate index
       "coding_rate": "4/6",                  // LoRa coding rate
       "frequency": "868300000",              // Frequency (Hz)
-      "timestamp": 2463457000,               // Timestamp of the gateway concentrator when the message has been received
-      "time": "2020-02-12T15:15:45.787Z"     // ISO 8601 UTC timestamp at which the uplink has been received by the gateway
+      "timestamp": 2463457000,               // Timestamp of the gateway concentrator when the accepted uplink message has been received
+      "time": "2020-02-12T15:15:45.787Z"     // ISO 8601 UTC timestamp at which the accepted uplink has been received by the gateway
```
johanstokking, line 120: Actually I think we should reset these fields here so they disappear, @rvolosatovs what do you think?

If not, it's not really "accepted" but "deduplicated", and I'm afraid it's pretty random which one is chosen here

benolayinka: Reset them where?

johanstokking: In the Network Server

rvolosatovs: Network Server discards these, so these are only seen and used by Gateway Server.

johanstokking: > Network Server does not use these, so they are only used by Gateway Server.

Gateway Server sets them as part of the individual `RxMetadata`. It remains useful in there.

> I think NS should unset them as well, in fact.

That's what I mean. It doesn't make sense anymore to have this information _after_ deduplication in the `Settings`; but they should stay in the settings

johanstokking: @benolayinka remove these two fields here for now as we'll get rid of this

Please also file issue for unsetting these field values in NS after deduplication

benolayinka:
> Is the issue filed for unsetting these fields in NS?
> 
> I don't see the (back)reference here

Filed it but I'm not sure what it means to unset, so the issue could use improvement

PR #2662, opened 2020-06-04
Improve Redis performance

### LINKED ISSUE #33: Retry failed registry transactions
**Summary:** 

Components should retry failed registry transactions if possible

**Why do we need this?**

To gracefully handle errors

The RPC fails

Retries of the transaction

Depends on the concrete case

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/1163 by @rvolosatovs

rvolosatovs:
#### Summary

Refs #33 

#### Changes

- Update Redis library
- Delete consumer on process exit
- Use a per-device lock in Network Server (lock the UID key and EUI key, if modified, on set)
- Ensure uplink deduplicator uses expiring keys

#### Testing

Integration testing, unit tests,

bunch of concurrent
```go
package main

import (
	"context"
	"flag"
	"os"
	"os/signal"
	"syscall"
	"time"

	"github.com/go-redis/redis/v7"
	ttnredis "go.thethings.network/lorawan-stack/v3/pkg/redis"
)

func main() {
	flag.Parse()

	cl := ttnredis.New(&ttnredis.Config{
		Address:       "localhost:6379",
		RootNamespace: []string{"ttn", "v3"},
	})
	logger, err := log.NewLogger(log.WithHandler(log.NewCLI(os.Stderr)))
	if err != nil {
		panic(err)
	}
	ctx := log.NewContext(context.Background(), logger)
	logger.Info("Waiting for lock...")
	if err = ttnredis.LockedWatch(ctx, cl, flag.Arg(0), flag.Arg(1), time.Minute, func(*redis.Tx) error {
		ch := make(chan os.Signal, 1)
		signal.Notify(ch, syscall.SIGKILL, syscall.SIGABRT, syscall.SIGINT)
		logger.Info("Locked. Waiting for signal to release lock...")
		logger.Infof("Got %v, release lock", <-ch)
		return nil
	}); err != nil {
		logger.Errorf("Got error: %+v", err)
		logger.Errorf("Got error cause: %+v", errors.Cause(err))
	}
}
```

##### Regressions

Update of the library can affect potentially anything, but I followed the `CHANGELOG` closely, so should be fine

#### Checklist

### DIFF
--- a/pkg/redis/redis.go
+++ b/pkg/redis/redis.go
@@ -687,9 +687,7 @@ local ok = redis.call('set', KEYS[1], '', 'px', exp, 'nx')
 if #ARGV > 1 then
 	table.remove(ARGV, 1)
 	redis.call('rpush', KEYS[2], unpack(ARGV))
-	if ok then
-		redis.call('pexpire', KEYS[2], exp)
-	end
+	redis.call('pexpire', KEYS[2], exp)
 end
 if ok then
 	return 1
--- a/pkg/redis/redis.go
+++ b/pkg/redis/redis.go
@@ -25,7 +25,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/go-redis/redis"
+	"github.com/go-redis/redis/v7"
 	"github.com/gogo/protobuf/proto"
 )
 
@@ -413,8 +413,8 @@ func DispatchTasks(r WatchCmdable, group, id string, maxLen int64, deadline time
 	if err != redis.Nil {
 		_, err := r.Pipelined(func(p redis.Pipeliner) error {
 			for i, ret := range rets {
-				toAdd := make([]redis.Z, 0, len(ret.Messages))
-				toAddNX := make([]redis.Z, 0, len(ret.Messages))
+				toAdd := make([]*redis.Z, 0, len(ret.Messages))
+				toAddNX := make([]*redis.Z, 0, len(ret.Messages))
 				toAck := make([]string, 0, len(ret.Messages))
 				for _, msg := range ret.Messages {
 					var score float64
@@ -457,12 +457,12 @@ func DispatchTasks(r WatchCmdable, group, id string, maxLen int64, deadline time
 					}
 
 					if replace {
-						toAdd = append(toAdd, redis.Z{
+						toAdd = append(toAdd, &redis.Z{
 							Member: member,
 							Score:  score,
 						})
 					} else {
-						toAddNX = append(toAddNX, redis.Z{
+						toAddNX = append(toAddNX, &redis.Z{
 							Member: member,
 							Score:  score,
 						})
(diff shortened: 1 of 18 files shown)


johanstokking:
> Refs #33

Good to start addressing that issue. What would be the effect of this on #33?

johanstokking:
Do we have tracing and insight in Redis load per operation?

I can imagine that finding matching devices by loading and decoding the entire device proto, including the entire mac state and recent uplink messages, is one of the things that we can improve; i.e. by having separate keys (hash maps?) with DevAddr, FCnt and NwkSIntKeys to find the UID, before loading the end device.

Probably out of scope for this PR seeing the diff, but would be great to add or file as issue.

rvolosatovs:
> Good to start addressing that issue. What would be the effect of this on #33?

Retrying of failed transactions will not be added just yet, but transaction failures should happen only on implementation errors (in NS for now). Locking will be able to fail though, but that will take way less resources and should be way easier to deal with - I will add some basic *lock* retry logic in this PR.

> Do we have tracing and insight in Redis load per operation?

See https://github.com/go-redis/redis/blob/master/CHANGELOG.md#v8-unreleased

Once we update our redis library, we'll be able to implement that either manually via context or perhaps using `OpenTelemetry` (not sure if that's what it does)?

> I can imagine that finding matching devices by loading and decoding the entire device proto, including the entire mac state and recent uplink messages, is one of the things that we can improve; i.e. by having separate keys (hash maps?) with DevAddr, FCnt and NwkSIntKeys to find the UID, before loading the end device.

Good point, please file an issue. This is indeed out-of-scope of this PR and especially #33

htdvisser:
@rvolosatovs I sent you tracing instructions on Slack.

johanstokking:
https://github.com/TheThingsNetwork/lorawan-stack/issues/2665

rvolosatovs:
Added the lock in [[https://github.com/TheThingsNetwork/lorawan-stack/pull/2662/commits/aa67b7edf4b9181ec3640936cfcf0ebf634355cf ](https://github.com/TheThingsNetwork/lorawan-stack/pull/2662/commits/aa67b7edf4b9181ec3640936cfcf0ebf634355cf)](https://github.com/TheThingsNetwork/lorawan-stack/pull/2662/commits/aa67b7edf4b9181ec3640936cfcf0ebf634355cf)
The core of the implementation is based on simple lock from https://redislabs.com/ebook/part-2-core-concepts/chapter-6-application-components-in-redis/6-2-distributed-locking/ however there's no busy-waiting/polling in proposed implementation. @htdvisser @johanstokking proof-reading welcome

Will promote the PR to "review-ready" once necessary configuration, ID generation is added to NS

rvolosatovs:
Note, the lock is only fair within the TTL of 1 lock, which is fine for our use case.
The very first caller to block on the lock will be the next "locker" if previous caller successfully unlocks the lock within the TTL. In case the context of the new "locker" is `Done()`, it will pass the lock to next caller in the queue in a fair way. However, if the following callers in the queue are not served within the lock's TTL they initially queried, they will requeue, and, hence lose their position in the queue and, hence, the fairness property of the queue.

pkg/redis/redis.go
```diff
@@ -668,49 +682,183 @@ type Scripter interface {
...
+
+		timeout := time.Duration(ttlMS) * time.Millisecond
+		if hasDeadline {
+			until := time.Until(dl)
+			if until < timeout {
+				timeout = until
+			}
+		}
+		if timeout < time.Second {
+			// Necessary until https://github.com/go-redis/redis/issues/1363 is resolved.
```
rvolosatovs, line 797: This will be removed once https://github.com/go-redis/redis/pull/1365 is merged and, I suppose, we migrate to v8 of the library.

rvolosatovs:
> I'm okay with this. Didn't check the implementation of the lock, but I trust that you just followed the literature.

I did not, it's made from scratch, but it is based on (some of the) principles from the article.
This is also why I asked for explicit proof-reading of the algorithm and tried to describe/explain it's inner workings several times in this PR.

PR #2667, opened 2020-06-05
Use RTT p90 with TTL for determining late schedule

johanstokking:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2178
References https://github.com/TheThingsIndustries/lorawan-stack-support/issues/24

#### Changes

- Use time-to-live for round-trip times
- Consider percentile instead of maximum observed RTT to determine whether scheduling is too late
- Minor code cleanups in connection stats

#### Testing

Unit testing

##### Regressions

We now have the following magic numbers to observe RTTs:

```go
const (
	maxRTTs = 20
	rttTTL  = 30 * time.Minute
)
```

And the following for determining if scheduling is too late:

```go
const (
	// scheduleMinRTTCount is the minimum number of observed round-trip times that are taken into account before using
	// using their statistics for calculating an absolute time or determining whether scheduling is too late.
	scheduleMinRTTCount = 5

	// scheduleLateRTTPercentile is the percentile of round-trip times that is considered for determining whether
	// scheduling is too late.
	scheduleLateRTTPercentile = 90
)
```

This changes behavior; GS becomes more permissive for scheduling downlink.

If GS becomes too permissive, i.e. when it schedules downlink while it's too late, we may need to increase `rttTTL` and decrease `scheuldeMinRTTCount`. 

#### Notes for Reviewers

With the current settings, RTTs are only considered when there have been more than 5 downlinks in the last 30 minutes and the second highest is taken into account.

This is primarily based on experienced guesswork.

#### Checklist


pkg/gatewayserver/io/rtts.go
```diff
@@ -20,55 +20,69 @@ import (
...
+	min = sorted[0].d
+	max = sorted[len(sorted)-1].d
 	if l := len(sorted); l%2 == 0 {
-		median = (sorted[l/2-1] + sorted[l/2]) / 2
+		median = (sorted[l/2-1].d + sorted[l/2].d) / 2
 	} else {
-		median = sorted[l/2]
+		median = sorted[l/2].d
 	}
+	np = sorted[int(float32(percentile)/100.0*float32(len(sorted)))-1].d
```
adriansmares, line 85: Nit: flip this with the line bellow and use `count` instead of `len(sorted)`. Also does this will panic if `count` is 1 and percentile is not 100.

pkg/gatewayserver/io/io.go
```diff
@@ -33,7 +33,9 @@ import (
 
 const (
 	bufferSize = 1 << 4
-	maxRTTs    = 1 << 5
+
```
KrishnaIyer, line 36: I think we should expose this a config. We can file an issue for this to target `develop`.

pkg/gatewayserver/io/rtts.go
adriansmares: You may also consider putting `count` above `min` in order to avoid creating `l`.

pkg/gatewayserver/io/rtts.go
```diff
@@ -20,55 +20,69 @@ import (
...
 	r.mu.Unlock()
 }
 
 // Last returns the last measured round-trip time.
-func (r *rtts) Last() (time.Duration, bool) {
+func (r *rtts) Last(ref time.Time) (time.Duration, bool) {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
-	if len(r.items) == 0 {
+	if len(r.items) == 0 || ref.Sub(r.items[len(r.items)-1].t) > r.ttl {
```
KrishnaIyer, line 57: Wouldn't this cause a seg fault if `r.items` is empty?
The logical OR `||` has the least precedence here with both brackets and the relational `>` evaluated first before the OR condition.

pkg/gatewayserver/io/rtts.go
johanstokking: > Nit: flip this with the line bellow and use `count` instead of `len(sorted)`

Doesn't matter, see https://blog.golang.org/slices-intro

> Also this will panic if `count` is 1 and percentile is not 100.

Fixed, CI caught it too fortunately

pkg/gatewayserver/io/rtts.go
johanstokking: Evaluation is left-to-right here

pkg/gatewayserver/io/io.go
johanstokking: Yeah, let's do that as-needed

johanstokking:
`compat/db` is needed because the RTTs are now stored under `down` instead of `up`, which is not compatible but it's a bugfix so it's allowed in a minor. Also, the previous worked only because of the fact that most GS deployments are single instance and the entire proto was stored under the `up` key (including downlink stats)

pkg/gatewayserver/io/rtts.go
KrishnaIyer: Ok that's actually counterintuitive. If expressions get evaluated L->R regardless, then operator precedence has no effect.

johanstokking: It does, operator precedence makes us avoid having to write `a || (b > c)` because `>` is higher than `||`, `b > c` is a group, but the `a` is evaluated before `b > c`

KrishnaIyer: Yes but without using brackets `(b > c)`  there's no way to determine which is the group. Both `(a || b) > c` and `a || (b > c)` are equally likely and this is where precedence comes in. But anyway this is off topic.

adriansmares: https://kuree.gitbooks.io/the-go-programming-language-report/content/22/text.html

PR #2673, opened 2020-06-06
Allow rendering custom button content

bafonins:
#### Summary

This PR adds functionality to render custom content in the `<Button />` component.

References and blocks https://github.com/TheThingsNetwork/lorawan-stack/pull/2644

#### Changes

- Check for `children` prop in `buttonChildren`
- Add `<Button.Icon />` component that holds button specific styles
- Pass other props in the `<SubmitButton />` mainly to pass `children` along when required.

#### Testing

Check storybook for `<Button />`

##### Regressions

`<Button />` styling.
`<SubmitButton />` functionality.

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md`. The target branch is set to `master` if the changes are fully compatible with existing API, database, configuration and CLI.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #2631, opened 2020-05-29
Stepper component improvements

bafonins:
#### Summary

This PR adds some improvements to the `<Stepper />` component.

References and blocks https://github.com/TheThingsNetwork/lorawan-stack/pull/2644

#### Changes

- Add media query to apply vertical styles to the `<Stepper />` component
- Fix prop-types warning
- Add `withInfo` addon to storybook
- Fix `Stepper.Step` displayName

#### Testing

Check interactive `<Stepper/ >` storybook

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md`. The target branch is set to `master` if the changes are fully compatible with existing API, database, configuration and CLI.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #2676, opened 2020-06-07
Add wizard component

bafonins:
#### Summary

References and blocks https://github.com/TheThingsNetwork/lorawan-stack/pull/2644

Blocked by https://github.com/TheThingsNetwork/lorawan-stack/pull/2631 and https://github.com/TheThingsNetwork/lorawan-stack/pull/2673

#### Changes

- Add `<Wizard/>` component
- Add `<Wizard />` to storybook
- Add `useCombinedRefs` hook

#### Testing

Storybook for `<Wizard />`

##### Regressions

None, as this PR only adds a new component

#### Notes for Reviewers

General features:
* The form can be submitted on each forward transition (only when inputs are valid)
* When navigating backwards, the wizard component persist the current state of the form.
* Navigation to the previous step is always allowed regardless of the current state of the step.

~~This PR uses other implementation of the wizard component from what is currently in https://github.com/TheThingsNetwork/lorawan-stack/pull/2644. It uses a single `Formik` instance whose children are each page of the multi step form, while in #2644 a new `Formik` instance is used for each step. Personally, I prefer this implementation, however it is more verbose since it requires passing props directly to `Wizard.Step`. Also, this implementation references example implementations from the `formik` repository.~~ Reverted to the original implementation as it better matches with how deal with forms.

~~`createWizardComponent` - creates dumb components that facilitate the compound components approach.~~ 

#### Checklist


pkg/webui/lib/render-callback.js
```diff
@@ -0,0 +1,35 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Allows react components to use the render prop design pattern without explicitly
+ * specifying `render`, `children` and `children` as a function props.
+ *
+ * @param {object} param0 - = The props of a component.
+ * @param {object} context - The context to be bassed to children.
```
kschiffer, line 20:
```suggestion
 * @param {object} props - The props of a component.
 * @param {object} context - The context to be passed to children.
```

PR #2677, opened 2020-06-07
Add metrics to pkg/log

### LINKED ISSUE #2634: Add telemetry to pkg/log
#### Summary

It would be nice to have some observability in `pkg/log`.

#### Why do we need this?

If we'd have some Prometheus metrics that count logs for a certain level, we could get some insights into how many warnings/errors we're logging. We can also set up alerts to notify us when error rates increase.

A log middleware that registers a `metrics.ContextualCounterVec` to keep track of log messages per level. Maybe also per `namespace`.

Can easily verify this on localhost:1885/metrics

neoaggelos:
#### Summary

Closes #2634 

#### Changes

- Add a new observability middleware
- Use the new middlewarewhen setting up the logger during stack init.

#### Testing

Tested locally by going to localhost:1885/metrics

##### Regressions

All entries have a level, so there should not be any cases where this breaks logging.

#### Checklist

### DIFF
--- a/pkg/log/middleware/observability/observability.go
+++ b/pkg/log/middleware/observability/observability.go
@@ -1,0 +1,79 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


pkg/log/middleware/observability/observability.go
```diff
@@ -0,0 +1,79 @@
...
+	metrics.MustRegister(logMetrics)
+}
+
+// Observability is a log.Handler that tracks metrics for logged messages.
+type Observability struct{}
+
+// New creates a new Observability log middleware.
+func New() log.Middleware {
+	return &Observability{}
+}
```
johanstokking, line 65: Is there a reason why `Observability` is exposed?

Issue #2655, opened 2020-06-03
Log messages for health checks


htdvisser:
#### Summary

We're too chatty with log messages for health checks.

#### Steps to Reproduce

1. Start The Things Stack
2. Send a TCP health check to the gRPC port: `echo "" | nc localhost 1884`
3. Send HTTP requests:
	- `curl localhost:1885/healthz/live`
	- `curl localhost:1885/healthz/ready`
4. Send TCP heath checks to the TLS ports: `for port in 8881 8882 8883 8884 8885 8886 8887; do echo "" | nc localhost $port; done`

#### What do you see now?

```
  WARN grpc: Server.Serve failed to create ServerTransport: connection error: desc = "transport: http2Server.HandleStreams failed to receive the preface from client: unexpected EOF" namespace=grpc
  INFO Request handled                          duration=9.07µs method=GET namespace=web remote_addr=[::1]:49210 request_id=01E9X6NSJJY48GRT1BWZ1QGFGF response_size=3 status=200 url=http://localhost:1885/healthz/live
  INFO Request handled                          duration=11.332µs method=GET namespace=web remote_addr=[::1]:49211 request_id=01E9X6NWCD1WB1VERPWD7Z6E6E response_size=3 status=200 url=http://localhost:1885/healthz/ready
  WARN Failed to setup connection               error=unexpected EOF namespace=gatewayserver/io/mqtt remote_addr=[::1]:49197
  WARN Failed to setup connection               error=unexpected EOF namespace=gatewayserver/io/mqtt remote_addr=[::1]:49198
  WARN Failed to setup connection               error=unexpected EOF namespace=applicationserver/io/mqtt remote_addr=[::1]:49199
  WARN grpc: Server.Serve failed to create ServerTransport: connection error: desc = "transport: http2Server.HandleStreams failed to receive the preface from client: unexpected EOF" namespace=grpc
2020/06/03 15:41:26 http: TLS handshake error from [::1]:49201: unexpected EOF
2020/06/03 15:41:26 http: TLS handshake error from [::1]:49202: unexpected EOF
2020/06/03 15:41:26 http: TLS handshake error from [::1]:49203: unexpected EOF
```

#### What do you want to see instead?

Considering that these kind of health checks happen a lot in production deployments, it would be better to make them less noisy.

#### How do you propose to implement this?

- Perhaps we can catch EOFs from the TLS handshake
- Perhaps we can catch EOFs from gRPC
[...]

PR #2678, opened 2020-06-08
Extend webmiddleware.Log to allow suppressing log messages for specific URL paths

neoaggelos:
#### Summary

References #2655 
#### Changes

- The `webmiddleware.Log` middleware now accepts a list of (regex) paths for which log messages are suppressed (unless the request failed).
- Refactored unit testing for `webmiddleware.Log` middleware.

#### Testing

- Unit tests
- Test locally by setting `--http.ignore-log-paths=^/metrics$` and trying the metrics endpoint (logs are suppressed) and other endpoints (logged properly)

##### Regressions

- There should be none, the ignore list is empty by default

#### Notes for Reviewers

Currently, the stack panics in case an invalid regex is passed, maybe we would like to take care of this and show a proper error message instead.

#### Checklist

### DIFF
--- a/doc/content/reference/configuration/the-things-stack.md
+++ b/doc/content/reference/configuration/the-things-stack.md
@@ -84,6 +84,10 @@ If {{% tts %}} is deployed behind a reverse proxy that does not use a private ne
 
 - `http.trusted-proxies`: CIDRs of trusted reverse proxies
 
+You can suppress log messages for successful HTTP requests (e.g. to reduce the noise caused by the health checks in a production environment).
+
+- `http.log-ignore-paths`: List of URLs for which to suppress logs of successful requests.
+
 ## Interoperability Options
 
 {{% tts %}} supports interoperability according to LoRaWAN Backend Interfaces specification. The following options are used to configure the server for this.
--- a/pkg/webmiddleware/log.go
+++ b/pkg/webmiddleware/log.go
@@ -24,7 +24,11 @@ import (
 
 // Log returns a middleware that logs requests.
 // If logger is nil, the logger will be extracted from the context.
-func Log(logger log.Interface) MiddlewareFunc {
+func Log(logger log.Interface, ignorePathsArray []string) MiddlewareFunc {
+	ignorePaths := make(map[string]struct{})
+	for _, path := range ignorePathsArray {
+		ignorePaths[path] = struct{}{}
+	}
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			logFields := log.Fields(
@@ -46,6 +50,12 @@ func Log(logger log.Interface) MiddlewareFunc {
 
 			r = r.WithContext(log.NewContext(ctx, requestLogger))
 			metrics := httpsnoop.CaptureMetrics(next, w, r)
+
+			if metrics.Code < 400 {
+				if _, ignore := ignorePaths[r.URL.Path]; ignore {
+					return
+				}
+			}
 
 			logFields = logFields.With(map[string]interface{}{
 				"status":        metrics.Code,
--- a/pkg/webmiddleware/log_test.go
+++ b/pkg/webmiddleware/log_test.go
@@ -33,83 +33,151 @@ func (ch logEntryChannel) HandleLog(e log.Entry) error {
 	return nil
 }
 
-func (ch logEntryChannel) Expect(t *testing.T, f func(log.Entry)) {
+func (ch logEntryChannel) Expect(t *testing.T, f func(*testing.T, log.Entry)) {
 	select {
 	case e := <-ch:
-		f(e)
+		f(t, e)
 	case <-time.After(time.Second):
 		t.Fatal("Missing log entry")
 	}
 }
 
+func (ch logEntryChannel) ExpectNoLog(t *testing.T) {
+	select {
+	case <-ch:
+		t.Fatal("Expected no log entry but received one")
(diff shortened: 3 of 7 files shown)


doc/content/reference/configuration/the-things-stack.md
```diff
@@ -84,6 +84,10 @@ If {{% tts %}} is deployed behind a reverse proxy that does not use a private ne
 
 - `http.trusted-proxies`: CIDRs of trusted reverse proxies
 
+You can suppress log messages for successful HTTP requests (e.g. to reduce the noise caused by the health checks in a production environment).
+
+- `http.ignore-log-paths`: List of URLs for which to suppress logs of successful requests (regex supported).
```
johanstokking, line 89: Should be `log-ignore-paths`

pkg/webmiddleware/log.go
```diff
@@ -24,7 +25,14 @@ import (
 
 // Log returns a middleware that logs requests.
 // If logger is nil, the logger will be extracted from the context.
-func Log(logger log.Interface) MiddlewareFunc {
+func Log(logger log.Interface, ignorePaths []string) MiddlewareFunc {
+	ignoreRegex := []*regexp.Regexp{}
+	for _, path := range ignorePaths {
+		ignoreRegex = append(ignoreRegex, regexp.MustCompile(path))
```
johanstokking, line 31: Don't panic here; I think that's going to be really hard for users to figure out why there was a panic in the middle of all the configuration setup for TTS. This means that `Log()` should return the error or that it warns here that a regex is invalid and execution continues

pkg/webmiddleware/log.go
```diff
@@ -24,7 +25,14 @@ import (
...
+	}
+	if len(ignorePaths) > 0 {
+		logger.WithField("paths", ignorePaths).Debug("Ignoring logs for paths")
```
johanstokking, line 34: Log messages are in imperative mood or past tense

pkg/webmiddleware/log.go
```diff
@@ -24,7 +25,14 @@ import (
 
 // Log returns a middleware that logs requests.
 // If logger is nil, the logger will be extracted from the context.
-func Log(logger log.Interface) MiddlewareFunc {
+func Log(logger log.Interface, ignorePaths []string) MiddlewareFunc {
+	ignoreRegex := []*regexp.Regexp{}
```
johanstokking, line 29: `ignoreRegexes := make([]*regexp.Regexp, 0, len(ignorePaths))`

- init slice with known size
- variable holds set so it's plural

pkg/webmiddleware/log_test.go
```diff
@@ -33,83 +33,144 @@ func (ch logEntryChannel) HandleLog(e log.Entry) error {
 	return nil
 }
 
-func (ch logEntryChannel) Expect(t *testing.T, f func(log.Entry)) {
+func (ch logEntryChannel) Expect(t *testing.T, a *assertions.Assertion, f func(*assertions.Assertion, log.Entry)) {
 	select {
 	case e := <-ch:
-		f(e)
+		f(a, e)
```
johanstokking, line 39: I think it's better convention to use `assertions.New(t)` when you need them, and pass `*testing.T` around

johanstokking:
Configuration additions that don't change previous behavior are also not required to bump minor, so this can target `master` as well in its current shape

pkg/webmiddleware/log.go
```diff
@@ -24,7 +25,18 @@ import (
 
 // Log returns a middleware that logs requests.
 // If logger is nil, the logger will be extracted from the context.
-func Log(logger log.Interface) MiddlewareFunc {
+func Log(logger log.Interface, ignorePaths []string) MiddlewareFunc {
+	ignoreRegexes := make([]*regexp.Regexp, 0, len(ignorePaths))
+	for _, path := range ignorePaths {
+		logger := logger.WithField("path", path)
```
htdvisser, line 31: I don't think we need to log that we're ignoring these paths.

pkg/webmiddleware/log.go
```diff
@@ -47,6 +59,14 @@ func Log(logger log.Interface) MiddlewareFunc {
 			r = r.WithContext(log.NewContext(ctx, requestLogger))
 			metrics := httpsnoop.CaptureMetrics(next, w, r)
 
+			if metrics.Code < 400 {
+				for _, regex := range ignoreRegexes {
+					if regex.MatchString(r.URL.Path) {
+						return
+					}
+				}
```
htdvisser, line 67: I don't know if the performance penalty of matching multiple regexes on every request is worth the added benefit of filtering out some logs. I think it's better to start with exact matches, and perhaps extend to prefixes in the future.

pkg/webmiddleware/log.go
```diff
@@ -47,6 +51,12 @@ func Log(logger log.Interface) MiddlewareFunc {
 			r = r.WithContext(log.NewContext(ctx, requestLogger))
 			metrics := httpsnoop.CaptureMetrics(next, w, r)
 
+			if metrics.Code < 400 {
+				if _, ignore := ignorePaths[r.URL.Path]; ignore {
```
neoaggelos, line 55: Using a map here instead of going through the array of the paths for every HTTP request.

Maybe could rewrite as `ignorePaths map[string]bool` and `if metrics.Code < 400 && ignorePaths[r.URL.Path]` ? Seems more of a stylistic choice.

htdvisser: I'm fine with this.

PR #2692, opened 2020-06-10
Add workaround for toast css module build issue

kschiffer:
#### Summary
This PR will fix an issues where our builds would break when the toast component is imported from more than one entry point.

#### Changes
- Replace the toast css styling with a converted stylus file

#### Testing

- Include the toast component in the oauth app somewhere (alternatively use TTES with `mage js:buildTTI`)
- Without this patch you should see only a blank render after the build is finished
- With this patch applied, everything should work as expected

##### Regressions

It could break the styling of the toast component. I checked that.

#### Notes for Reviewers
Even after countless hours of debugging, I have no idea what the actual root cause of this bug is but I managed to narrow it down so much to see that it has to do with the `'react-toastify/dist/ReactToastify.css'` that is imported by the toast component. Converting its content to a stylus file fixed the issue 🤷‍♂️.

#### Checklist

### DIFF
--- a/pkg/webui/components/toast/index.js
+++ b/pkg/webui/components/toast/index.js
@@ -14,12 +14,12 @@
 
 import React from 'react'
 import { ToastContainer as Container, Slide } from 'react-toastify'
-import 'react-toastify/dist/ReactToastify.css'
 
 import PropTypes from '@ttn-lw/lib/prop-types'
 
 import createToast from './toast'
 
+import './react-toastify.styl'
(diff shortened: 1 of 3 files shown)


kschiffer:
I believe it has to do with the hashes that are generated for css classnames, which appear to cause trouble when `*.css` modules are imported from more than one entry point. I wasn't able to resolve that general issue but the workaround will do for the time being.

PR #2695, opened 2020-06-10
Reset device payload formatters

### LINKED ISSUE #2310: Allow device payload formatters to be unspecified

pgalic96:
#### Summary

Closes #2310

Old PR by @bafonins at #2462 

#### Changes

- #2462
- Add warning about setting both uplink and downlink payload formatters to default app formatter

#### Testing

- Check redux state

##### Regressions

- Payload formatters might get affected, but as this is a new option it does not affect other options. The way to test it is to test all the payload formatters.

#### Checklist


CHANGELOG.md
```diff
@@ -11,11 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Metrics for log messages, counted per level and namespace.
 - Allow suppressing logs on HTTP requests for user-defined paths (see `--http.log-ignore-paths` option).
-- Redux state and actions reporting to Sentry
-- Serving frontend sourcemaps in production
```
kschiffer, line 15: These shouldn't be removed?

pgalic96: They were duplicates

PR #2730, opened 2020-06-12
Migrate go code checks to github actions

rvolosatovs:
#### Summary

Migrate go code checks to github actions

#### Changes

- Migrate go code checks to github actions

#### Testing

Fork

#### Checklist


.github/workflows/go.yml
```diff
@@ -0,0 +1,133 @@
...
+    - '!tools/mage/version.go'
+    - '!tools/mage/xx.js'
+
+jobs:
+  quality:
+    name: Code Quality
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Set up Go
+      uses: actions/setup-go@v2
```
htdvisser, line 52: Let's use

```yaml
with:
  go-version: '~1.14'
```

.github/workflows/go.yml
```diff
@@ -0,0 +1,133 @@
...
+        args: -h postgres -U root ttn_lorawan_is_test
+    - name: Create ttn_lorawan_is_store_test DB
+      uses: docker://postgres
+      env:
+        PGPASSWORD: root
+      with:
+        entrypoint: /usr/bin/createdb
+        args: -h postgres -U root ttn_lorawan_is_store_test
+    - name: Set up Go
+      uses: actions/setup-go@v2
```
htdvisser, line 109: Let's use

```yaml
with:
  go-version: '~1.14'
```

.github/workflows/go.yml
```diff
@@ -0,0 +1,133 @@
...
+      uses: actions/cache@v2
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
+    - name: Check out code
+      uses: actions/checkout@v2
+    - name: Make Mage
+      run: make tools/bin/mage
```
htdvisser, line 63: Perhaps we could speed things up if we also cache the `mage` binary or is that too much work?

.github/workflows/go.yml
```diff
@@ -0,0 +1,133 @@
...
```
htdvisser, line 120: Perhaps we could speed things up if we also cache the `mage` binary or is that too much work?

.github/workflows/go.yml
```diff
@@ -0,0 +1,133 @@
...
+    - name: Initialize Go cache
+      uses: actions/cache@v2
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
+    - name: Check out code
+      uses: actions/checkout@v2
+    - name: Make Mage
```
htdvisser, line 62: Maybe add a `go mod download` before this one?

.github/workflows/go.yml
```diff
@@ -0,0 +1,133 @@
...
```
htdvisser, line 119: Maybe add a `go mod download` before this one?

.github/workflows/go.yml
```diff
@@ -0,0 +1,133 @@
...
+    - name: Create ttn_lorawan_is_store_test DB
+      uses: docker://postgres
+      env:
+        PGPASSWORD: root
+      with:
+        entrypoint: /usr/bin/createdb
+        args: -h postgres -U root ttn_lorawan_is_store_test
+    - name: Set up Go
+      uses: actions/setup-go@v2
+    - name: Initialize Go cache
```
htdvisser, line 110: Shouldn't we do this after the checkout? Since we're relying on the hash of `go.sum`?

.github/workflows/go.yml
```diff
@@ -0,0 +1,133 @@
...
+    - '!tools/mage/xx.js'
+
+jobs:
+  quality:
+    name: Code Quality
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Set up Go
+      uses: actions/setup-go@v2
+    - name: Initialize Go cache
```
htdvisser, line 53: Shouldn't we do this after the checkout? Since we're relying on the hash of `go.sum`?

.github/workflows/go.yml
rvolosatovs: Is there a reason not to just use the latest?

htdvisser: Because

1. We had problems with using the "latest" version of things in the past.
1. We want to be able to *choose* to use a new Go version, and not be forced to.
1. We want to be able to re-run a build from the past and still get (almost) the same result.

PR #2748, opened 2020-06-16
Only initialize Redis Mutex for components that use it

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/pull/2200
Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2199

#### Changes

- Only initialize Redis Mutex for components that use it

#### Checklist

### DIFF
--- a/cmd/ttn-lw-stack/commands/start.go
+++ b/cmd/ttn-lw-stack/commands/start.go
@@ -19,6 +19,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


rvolosatovs:
> LGTM. Note as I mentioned this needs a Patch release.

Hmm, can you clarify? Also, in the issue you mention this is "breaking config" - how is that? There are no config changes.

KrishnaIyer:
> Also, in the issue you mention this is "breaking config" - how is that? There are no config changes.

Well we don't need this fix in `3.8.4` and workaround the problem by setting up Redis and adding that config. But doing that means "adding config that's not there for the something to work" which is breaking config.

Issue #603, opened 2019-05-01
Add i18n to storybook


bafonins:
#### Summary

See properly rendered `<Message />`s in the storybook

#### Why do we need this?

See the components as they would be rendered in the console

#### What is already there? What do you see now?

* `react-intl` dependency
* logic to generate messages 

#### What is missing? What do you want to see?

* translated messages in the storybook
* absence of `react-intl` warnings: 
```
[React Intl] Cannot format message: "event:gs.down.send", using message id as fallback.
```

#### Environment

Browser

#### How do you propose to implement this?

Similar to what we do for the console

#### Can you do this yourself and submit a Pull Request?

I can check if it is feasible and then implement.

cc @kschiffer

PR #2759, opened 2020-06-16
Fix broken components in storybook

### LINKED ISSUE #2693: Fix component stories
#### Summary
A couple of component stories are currently crashing and should be fixed.

#### Steps to Reproduce
The following stories have issues:

- Icons
- List
- Status
  - With Label
  - Without Pulse

#### What do you see now?
Crashing / buggy stories

All stories functioning properly

Via storybook

@pgalic96

pgalic96:
#### Summary

Closes #2693 

#### Changes

- Change array generation method in `List` component as the previous one was not generating items correctly
- Add `Message` component to `Status` label, this should be fixed for all components by adding internationalization (https://github.com/TheThingsNetwork/lorawan-stack/issues/603)
- Wrap the icon elements with a top level `div` to fix `Icon` story
- Fix proptype warning for `List`

#### Testing

Check storybook

##### Regressions

None

#### Checklist

### DIFF
--- a/pkg/webui/components/icon/story.js
+++ b/pkg/webui/components/icon/story.js
@@ -44,6 +44,15 @@ display value of the wrapping \`<span />\`in your local scoped css. The
 positioning will differ slightly, so the nudge props can be used to fine-tune
 the appearance.`
 
+const iconElement = icons.map(icon => {
+  return (
+    <div className={style.wrapper} key={icon}>
+      <Icon icon={icon} />
+      {icon}
+    </div>
+  )
+})
+
 storiesOf('Icon', module)
   .addDecorator((story, context) =>
     withInfo({
@@ -53,16 +62,7 @@ storiesOf('Icon', module)
       text: doc,
     })(story)(context),
   )
-  .add('Icons', () =>
-    icons.map(function(icon) {
-      return (
-        <div className={style.wrapper} key={icon}>
-          <Icon icon={icon} />
-          {icon}
-        </div>
-      )
-    }),
-  )
+  .add('Icons', () => <div>{iconElement}</div>)
   .add('Usage', () => (
     <div className={style.wrapper}>
       <div className={style.block}>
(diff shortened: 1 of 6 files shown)


pkg/webui/components/status/story.js
```diff
@@ -15,16 +15,22 @@
...
 
 const containerStyle = {
-  width: '100px',
+  width: '120px',
   display: 'flex',
   alignItems: 'center',
   justifyContent: 'space-between',
 }
 
+const m = defineMessages({
```
bafonins, line 29: There is no reason to define custom translated messages just for storybook. If the `<Status />` component does not accept raw strings, it is a bug and should fixed.

pkg/webui/components/list/index.js
```diff
@@ -34,7 +34,9 @@ class List extends React.PureComponent {
     emptyMessageValues: PropTypes.shape({}),
     footer: PropTypes.node,
     header: PropTypes.node,
-    items: PropTypes.arrayOf(PropTypes.shape({})),
+    items: PropTypes.arrayOf(
+      PropTypes.oneOfType(PropTypes.shape({}), PropTypes.string, PropTypes.number),
```
bafonins, line 38:
```suggestion
      PropTypes.oneOfType(PropTypes.shape({}), PropTypes.string),
```

pkg/webui/components/icon/story.js
```diff
@@ -44,6 +44,15 @@ display value of the wrapping \`<span />\`in your local scoped css. The
 positioning will differ slightly, so the nudge props can be used to fine-tune
 the appearance.`
 
+const iconElement = icons.map(function(icon) {
```
bafonins, line 47:
```suggestion
const iconElements = icons.map(icon => {
```

pkg/webui/components/status/story.js
pgalic96: yep there was an issue with calling `intl.formatMessage` with a string as a parameter instead of a message, now this case is handled correctly.

pkg/webui/components/list/index.js
bafonins: actually, this should be array
```js
    items: PropTypes.arrayOf(
      PropTypes.oneOfType([PropTypes.shape({}), PropTypes.string]),
```

pkg/webui/components/status/story.js
bafonins: 👌

PR #2762, opened 2020-06-17
Support SRV records for service discovery

johanstokking:
#### Summary

Groundwork for #138

#### Changes

Use SRV records as proposed in #138 in `pkg/rpcmiddleware/discovery.DialContext()`

#### Testing

Unit testing and local test environment with production DNS

#### Notes for Reviewers

Let's track usage, documentation and applicable `CHANGELOG.md` additions in #138.

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md`. The target branch is set to `master` if the changes are fully compatible with existing API, database, configuration and CLI.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #2763, opened 2020-06-17
Support SRV records for service discovery

### LINKED ISSUE #138: Cluster and Service discovery

johanstokking:
#### Summary

Closes #138
Replaces #2762

#### Changes

Use SRV records as proposed in #138 via the `pkg/rpcmiddleware/discovery.WithDialer()` gRPC dial option.

Discovery for non-gRPC services is not implemented yet, but we can do that on a case-by-case basis, and consider doing that in specific client Go modules.

#### Testing

Unit testing and local test environment with production DNS (`_ttn-v3-dcs-grpc._tcp.0.0.0.0.0.0.0.D.E.7.5.D.3.B.0.7.join.thethings.industries`)

#### Checklist


pkg/rpcmiddleware/discover/discover.go
```diff
@@ -77,64 +99,106 @@ func DefaultURL(target string, port int, tls bool) (string, error) {
...
-	return grpc.DialContext(ctx, target, allOpts...)
+
+	return grpc.WithContextDialer(func(ctx context.Context, target string) (net.Conn, error) {
+		var addresses []string
+		if _, _, err := net.SplitHostPort(target); err == nil {
+			addresses = []string{target}
+		} else {
+			if service, ok := ServiceName(role); ok {
+				_, addrs, err := options.dnsResolver.LookupSRV(ctx, service, "tcp", target)
+				if err == nil {
```
htdvisser, line 172: Shouldn't we return lookup errors (other than not found)?

```go
var dnsErr *net.DNSError
if errors.As(err, &dnsErr) {
	if !dnsErr.IsNotFound {
		return nil, err
	}
}
```

johanstokking: 84617e29b

pkg/rpcmiddleware/discover/discover.go
```diff
@@ -21,9 +21,11 @@ import (
 	"net"
 	"strings"
```
htdvisser, line 23: Nit: We don't usually put empty lines between stdlib imports.

johanstokking: `gofumports` does that

PR #2780, opened 2020-06-22
Remove version from hosted documentation path

### LINKED ISSUE #2778: Stop serving versioned documentation
#### Summary

Let's drop the version number from our hosted documentation.

#### Why do we need this ?

- Breaks links
- Bad for SEO
- Bad for search #2588 

Version numbers e.g https://thethingsstack.io/v3.8.4/

Everything at https://thethingsstack.io/...

Yes. @johanstokking to give the ok?

benolayinka:
#### Summary

Closes #2778

#### Changes

- Remove version from docs build path

#### Testing

Run locally

#### Checklist


benolayinka:
> Looks effective.
> 
> Please document in the guidelines that contributors should document the version number if things get added in a new version. This applies to features, API, configuration etc. We didn't need this because we versioned the entire docs.

Do you mean that in the documentation for a new feature, we should also include the version which that feature first appeared in? Like 'new since version x'

benolayinka:
I think there's a redirect at the root of thethingsstack.io which @htdvisser or @KrishnaIyer will need to remove to make this work

.mage/docs.go
```diff
@@ -115,8 +114,7 @@ func (d Docs) Build() (err error) {
 	if err != nil {
 		return err
 	}
-	url.Path = path.Join(url.Path, currentVersion)
-	destination := path.Join("public", currentVersion)
+	destination := "public"
 	defer func() {
 		genErr := d.generateRedirect()
```
htdvisser, line 119: @benolayinka this is that root redirect

johanstokking:
> Do you mean that in the documentation for a new feature, we should also include the version which that feature first appeared in? Like 'new since version x'

Yes

benolayinka:
> > Do you mean that in the documentation for a new feature, we should also include the version which that feature first appeared in? Like 'new since version x'
> 
> Yes. For documentation that targets `master`, that's the next patch bump, for documentation targeting `develop` that's the next minor bump.
> 
> For documentation in PRs (or changes in the making) that are open while we tag a new release, the author should bump the version.
> 
> I think the above is useful to have in our documentation guidelines.

Sounds good but where should we put it? We could make a [Front Matter](https://gohugo.io/content-management/front-matter/) field and use that to generate "New since" indicators? @htdvisser

johanstokking:
I think it should become inline in the content, but we can make a shortcode for this.

htdvisser:
That might work if it applies to the entire document, but often something just applies to a small part, perhaps a config option, a CLI flag, a field in our API (we'd also have to annotate that BTW).

Let's start with a shortcode `{{< since-version "3.9" >}}` that generates a badge/tag or something.

Later we can figure out how we can do this with generated docs (such as the API docs)

htdvisser:
Haha @johanstokking jinx

benolayinka:
Added a `{{< new-in-version >}}` shortcode which displays at 60% font size middle-aligned. @pierrephz any better styling suggestions?

Screenshot 2020-06-25 at 14 02 29

DEVELOPMENT.md
```diff
@@ -324,6 +324,7 @@ We use [`revive`](http://github.com/mgechev/revive) to lint Go code and [`eslint
 
 Please respect the following guidelines for content in our documentation site. A copy and paste template for creating new documentation can be found [here](doc/content/example-template).
 
+- Use the `{{< new-in-version >}}` shortcode to tag documentation for features added in a particular version. For documentation that targets `master`, that's the next patch bump, e.g `3.8.x`. For documentation targeting `develop` that's the next minor bump, e.g `3.x`.
```
johanstokking, line 327: Insert an example version, on how the shortcode is used

```suggestion
- Use the `{{< new-in-version "3.8.5" >}}` shortcode to tag documentation for features added in a particular version. For documentation that targets `master`, that's the next patch bump, e.g `3.8.x`. For documentation targeting `develop` that's the next minor bump, e.g `3.x.0`.
```

doc/content/example-template/_index.md
```diff
@@ -18,6 +18,16 @@ Use a requirements subheading to list requirements/prerequisites.
 1. Requirement 1
 2. Requirement 2
 
+## New Features {{< new-in-version "3.8.5">}}
+
+Use the {{< new-in-version >}} shortcode to tag documentation for features added in a particular version of {{% tts %}}. For documentation that targets `master`, that's the next patch bump, e.g `3.8.x`. For documentation targeting `develop` that's the next minor bump, e.g `3.x`. Check the CHANGELOG for the current version.
```
johanstokking, line 23:
```suggestion
Use the `{{< new-in-version "3.8.5" >}}` shortcode to tag documentation for features added in a particular version of {{% tts %}}. For documentation that targets `master`, that's the next patch bump, e.g `3.8.x`. For documentation targeting `develop` that's the next minor bump, e.g `3.x.0`. Check `tools/bin/mage version:current` for the current version.
```

.mage/docs.go
htdvisser: So if you remove the versions, you can also remove the root recdirect.

pierrephz:
> Added a `{{< new-in-version >}}` shortcode which displays at 60% font size middle-aligned. @pierrephz any better styling suggestions?

@benolayinka, Looks like 60% works, but I would suggest to display it as a superscript and use our red (#F9595A)

image

.mage/docs.go
```diff
@@ -115,34 +108,10 @@ func (d Docs) Build() (err error) {
...
-	url.Path = path.Join(url.Path, currentVersion)
-	destination := path.Join("public", currentVersion)
-	defer func() {
-		genErr := d.generateRedirect()
-		if err == nil {
-			err = genErr
-		}
-	}()
+	destination := "public"
 	return execHugo("-b", url.String(), "-d", destination, "--environment", "gh-pages")
```
htdvisser, line 112: This is the only place where `destination` is used, so no need to have it as a variable anymore.

PR #2792, opened 2020-06-25
Improve form field labels

### LINKED ISSUE #1383: Fix required field styling
#### Summary

Screenshot 2019-09-25 at 16 43 41

Related commit: https://github.com/TheThingsNetwork/lorawan-stack/commit/182bd49159dcff229f6bd22dbc9b6ddf597dfe73

#### Steps to Reproduce

1. Open any form with required fields
2. See the required indicator being misaligned

#### What do you see now?

Screenshot 2019-09-25 at 16 43 41

Screenshot 2019-09-25 at 16 46 45

No

kschiffer:
#### Summary
This quickfix PR will introduce various improvements to the field labels in our forms.

Closes #1383

#### Changes
- Use  the `::after` selector to render the "required dot"
  - This will keep the inner HTML of the label clean and enable easier selection via `cy.findByLabelText()`
- Fix a display problem that made the required dot rendered too high (in horizontal layout)
- Use proper `id` attribute for the `<input />`'s and corresponding `for` attribute for labels
- Make sure checkbox/radio group are using and passing down the `name` attribute

#### Testing

Checked in storybook

##### Regressions

Might cause log errors for invalid IDs and/or `for` attributes

#### Notes for Reviewers
This is mostly for #2764 with some other small things added while being at it.

#### Checklist

### DIFF
--- a/pkg/webui/components/checkbox/checkbox.js
+++ b/pkg/webui/components/checkbox/checkbox.js
@@ -135,12 +135,14 @@ class Checkbox extends React.PureComponent {
       checkboxProps.onFocus = this.context.onFocus
       checkboxProps.disabled = disabled || this.context.disabled
       checkboxProps.checked = this.context.getValue(name)
+      checkboxProps.id = `${this.context.name}-${name}`
       groupCls = this.context.className
     } else {
       checkboxProps.onBlur = onBlur
       checkboxProps.onFocus = onFocus
       checkboxProps.disabled = disabled
       checkboxProps.checked = checked
+      checkboxProps.id = name
     }
 
     const cls = classnames(className, style.wrapper, groupCls, {
@@ -149,7 +151,7 @@ class Checkbox extends React.PureComponent {
     })
 
     return (
-      <label className={cls}>
+      <label className={cls} htmlFor={checkboxProps.id}>
         <span className={style.checkbox}>
           <input
             type="checkbox"
--- a/pkg/webui/components/radio-button/radio.js
+++ b/pkg/webui/components/radio-button/radio.js
@@ -19,6 +19,7 @@ import classnames from 'classnames'
 import Message from '@ttn-lw/lib/components/message'
 
 import PropTypes from '@ttn-lw/lib/prop-types'
+import stringToHash from '@ttn-lw/lib/string-to-hash'
 
 import { RadioGroupContext } from './group'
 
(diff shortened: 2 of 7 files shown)


pkg/webui/components/checkbox/checkbox.js
```diff
@@ -135,12 +135,14 @@ class Checkbox extends React.PureComponent {
...
       checkboxProps.disabled = disabled || this.context.disabled
       checkboxProps.checked = this.context.getValue(name)
+      checkboxProps.id = `${this.context.name}-${name}`
       groupCls = this.context.className
     } else {
       checkboxProps.onBlur = onBlur
       checkboxProps.onFocus = onFocus
       checkboxProps.disabled = disabled
       checkboxProps.checked = checked
+      checkboxProps.id = `${name}`
```
bafonins, line 145:
```suggestion
      checkboxProps.id = name
```

pkg/webui/components/radio-button/radio.js
```diff
@@ -107,20 +108,22 @@ class RadioButton extends React.PureComponent {
...
+      radioProps.id = `${name}-${stringToHash(value)}`
     }
 
     const cls = classnames(className, style.wrapper, groupCls, {
       [style.disabled]: radioProps.disabled,
     })
 
     return (
-      <label className={cls}>
+      <label className={cls} htmlFor={radioProps.id}>
```
bafonins, line 126: this shouldnt be necessary as `<input />` is wrapped by `<label />`

pkg/webui/components/radio-button/radio.js
```diff
@@ -107,20 +108,22 @@ class RadioButton extends React.PureComponent {
       radioProps.name = this.context.name
       radioProps.disabled = disabled || this.context.disabled
       radioProps.checked = value === this.context.value
+      radioProps.id = `${this.context.name}-${stringToHash(value)}`
```
bafonins, line 111: Why `stringToHash` ?

kschiffer: In order to make the ID unique, I use a combination of the name and value. However, the value could be an arbitrary string like `Example Value`, which would break the ID, so I use the hash instead.

pkg/webui/components/radio-button/radio.js
kschiffer: I know, but it also doesn't hurt and I wanted it to be consistent with the checkbox. Please unresolve this, if you disagree.

pkg/webui/components/radio-button/radio.js
bafonins: There a couple things to that:

1. I cannot imagine a use case in the console/oauth apps where one might need to define `Example Value` as the value. We dont use native form submission and do not allow the users to use the console/oauth without js, hence we can always transform the value and adjust the result accordingly.
2. In general it doesnt look right to use `Example Value` as the value for a radio button (imho)
3. `{name}-123123123123` is just noise in the dom and I personally check dom elements when adding features (imho)
4. `stringToHash` is just some function most probably copied from the internet without any references and I dont really feel to examine it carefully to argue about its collision probability
5. There is no reason to add the id attribute in the first place as `<input />` is wrapped within the `<label />` tag which implicitly associates those two elements.
6. You can always check and remove whitespaces via small utility function within the component, there no reason to add a separate dependency for a radio button. The less dependencies our core components have, the better.

kschiffer:
#2816

johanstokking:
@kschiffer please delete branch when no longer necessary

Issue #2694, opened 2020-06-10
Improve handling stack component addresses in the Console


bafonins:
#### Summary

Currently, the UI for setting stack component addresses can be ambiguous to novice users.

#### Why do we need this?

Hide specifics under default options such that new users do not have to deal with ttn stack specifics.

#### What is already there? What do you see now?

1. Usage of `{component} Server address`
2. Application link, device general settings, gateway creation, gateway general settings - requires the users to specify stack component addresses

#### What is missing? What do you want to see?

1. Use `{component} server host` instead of  `{component} server address` as we require the users to specify only host names and not full addresses.
2. Hide input fields under a checkbox with title `Use default {component} Server host` which is initially checked. Reveal the input and allow the user to specify custom host only when the checkbox is unchecked.

#### Environment

v3.8.3

#### How do you propose to implement this?

Extend validation schema to process checkbox values.
Conditionally render text inputs on checkbox value changes

#### How do you propose to test this?

Setup application link, create gateway, edit gateway, edit end device, create end device

#### Can you do this yourself and submit a Pull Request?

yes

bafonins:
Blocked on https://github.com/TheThingsNetwork/lorawan-stack/issues/2813

kschiffer:
#3522 includes a fix for the end device registration process and the same behavior can be used for the general settings as well. The updated gateway registration will likely use the same mechanic, but this is still in planning.

I'll remove the `blocked` and `needs/ux` labels, since this is now actionable as part of #3522 at least for end devices.

johanstokking:
@bafonins what is the status here?

bafonins:
@johanstokking See my [comment](https://github.com/TheThingsNetwork/lorawan-stack/issues/3522#issuecomment-788715945) regarding JS address handling in DR form

kschiffer:
See also #3947

bafonins:
Blocked on https://github.com/TheThingsNetwork/lorawan-stack/issues/3947

PR #2801, opened 2020-06-25
Improve handling stack component addresses in the Console

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2694

<details>

Device add:
Screenshot 2020-06-30 at 15 36 12
Screenshot 2020-06-30 at 15 36 17
Screenshot 2020-06-30 at 15 36 22
Screenshot 2020-06-30 at 15 36 30

Device general settings:

Screenshot 2020-07-02 at 16 26 42
Screenshot 2020-07-02 at 16 26 50

</details>

#### Changes

- Ignore `409` error when fetching applications stats when `link` data is present.
- Add `Use default component addresses` to the device wizard and general settings.

#### Testing

create/edit end device. Did the same with ns/as/js disabled.

##### Regressions

end device forms

#### Notes for Reviewers

Ignoring `409` errors when fetching app link allows the user to see the `link` data, otherwise it is not shown (happens prior to this PR)

@laurensslats your feedback is required as well

#### Checklist


kschiffer:
Regarding the stack component hosts, it's not quite what I had in mind. With this solution, there is still quite some potential confusion for the user, since he/she has to make several unclear decisions per stack component (and whether to use an external JS).

What I think is better is to hide the whole set of stack component configs behind a single checkbox that says `Use default component hosts`, which is enabled by default. When unchecked, the inputs for the individual components appear.

This way we can reassure the user that it is likely that no change is required without confronting/confusing them with the exact inputs that would be required. The checkbox per component input is not necessary, since the prefilled value already communicates that this is a suggested default.

Rest LGTM.

bafonins:
cc @johanstokking @laurensslats

johanstokking:
I tend to agree with @kschiffer; you either go with an all-defaults "I don't care" option that's enabled by default, or you have advanced options where you enter the addresses (and the defaults are filled in).

> * Change `{component} Server address` ti `{component} Server host`.

Strictly speaking it is an address; we do allow ports here.

bafonins:
Updated descriptions and fixed component address handling in device wizard

kschiffer:
I suggest putting this PR on hold until we have a [clear strategy for our form UX](#2813). I'm afraid that if we continue adding things right now, we could end up with a lot of patchwork that will not ultimately improve the UX as a whole. 

**Please** understand this is not against the implementation itself or to slow things down unnecessarily. I'm only trying to protect you from doing obsolete changes resulting from half-baked reviews resulting from the ongoing work on #2813. The whole form UX is a sensitive matter that needs careful consideration. I think this PR has some things that we want to add eventually but also others about which I will have to do some research about.

I will put priority to come up with solutions so we can continue here.

johanstokking:
@kschiffer we need to keep working in small iterations. This PR can only wait for #2813 if that doesn't take too long. There are useful changes here as requested from users and customers, and they care little about defining a holistic view around form UX.

For example, how can this PR be tweaked to address #2813 already?

kschiffer:
I understand and I'm struggling with that as well.

> they care little about defining a holistic view around form UX.

Certainly not about defining it but about implementing it. We saw many complaints ultimately resulting from a lack of just that. I believe we need to address this in a more organized way.

> For example, how can this PR be tweaked to address [#2813](https://github.com/TheThingsNetwork/lorawan-stack/issues/2813) already?

I initially wrote these down:
1.  We should use the same `Use default component addresses`-logic for the end device general settings
2.  I would suggest to remove the checkbox / input combination for component address fields and use a field with default value instead (like we did before). I understand why you added it this way but if we use this we'll end up with a redundant second level of protection.
3.  In a separate issue / PR, we need to look into reorganizing the various component address fields into an `Advanced settings` section (just like #2815)

With these changes, it should hopefully be sustainable enough.

bafonins:
Updated PR description.

I think this discussion clearly shows that our development processes are broken.

I still dont understand what was wrong with the initial implementation, imo it addressed all the issues novice users might have when dealing with component address fields. Anyway, I applied requested changes and Im strongly against merging this.
1. It looks much worse (especially the general settings page) than it is right now
2. It adds quite some complexity to the general settings page and I dont want to maintain this
3. It introduces inconsistencies regarding component address handling in the Console.

kschiffer:
> I still dont understand what was wrong with the initial implementation, imo it addressed all the issues novice users might have when dealing with component address fields.

Let me reiterate my standpoint: we cannot continue regarding individual UX issues in isolation. If we do that we will introduce patch-work solutions that might all make sense on their own but break the UX overall. I strongly believe that if we want to solve issues like these we need to zoom out and regard the user process as a whole, which is what I am currently working on.

For example, in this PR we hide stack components addresses behind a checkbox (grouped or individually). When investigating the user story for creating entities, however, it becomes apparent that:

*   Stack component addresses can be removed from some creation forms (#2806)
*   Stack component addresses can be put into an `Advanced settings` section instead

This is the reason why I initially suggested halting this PR until we have a better picture of where we need to go with our forms. I also believe in an iterative process but I am against following it blindly, especially when we are trying to improve functionality that is not broken per se but can just be done better.

> Anyway, I applied requested changes and Im strongly against merging this.

I ultimately agree for the reasons stated above. If we really need to make a change now that at least addresses the issue of confusing novice users with component addresses, we can merge it like this.

Issue #2445, opened 2020-04-29
Hide advanced settings in console when adding devices


laurensslats:
#### Summary

Simplify the device-creation by hiding advanced settings that most users don't understand, nor need to set.

#### Why do we need this?

Most of the users don't understand these fields, in most cases these fields don't have to be set.

#### What is already there? What do you see now?

![Screenshot 2020-04-29 at 12 14 37](https://user-images.githubusercontent.com/17290639/80585233-42537980-8a13-11ea-8fd4-3dd66ab5c012.png)

#### What is missing? What do you want to see?

Hide these settings and allow the user to expand these fields. Something like this:  
 

![Artboard 2](https://user-images.githubusercontent.com/17290639/80585862-50ee6080-8a14-11ea-9a17-2b7655061f09.png)

kschiffer:
This issue is overlapping with #579.

PR #2815, opened 2020-06-30
Hide advanced settings when adding devices in the Console

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2445

Screenshot 2020-07-01 at 14 58 23

Screenshot 2020-07-01 at 14 58 13

#### Changes

- Add `<Form.Section />`
- Hide unwanted fields in the join step

#### Testing

Create otaa devices (use users w/ and w/o the right to edit device keys)

##### Regressions

`net_id` and the rest of the fields not being handled properly

#### Checklist


pkg/webui/console/views/device-add/wizard/join-settings-form/validation-schema.js
```diff
@@ -52,25 +52,31 @@ const validationSchema = Yup.object()
...
       },
     ),
-    net_id: Yup.nullableString().emptyOrLength(
-      3 * 2,
-      Yup.passValues(sharedMessages.validateLength),
-    ), // 3 Byte hex.
-    resets_join_nonces: Yup.boolean().when('$lorawanVersion', {
-      // Verify if lorawan version is 1.1.0 or higher.
-      is: version => parseLorawanMacVersion(version) >= 110,
+    net_id: Yup.nullableString().when(['$includeAdvanced'], {
```
kschiffer, line 55: I think any collapsible section should not make a difference for the validation schema since:

1.  The sections only control the visibility of certain fields and they should not have any other side effects for the fields (e.g. a filled input that got hidden should still be respected when submitting the field)
2.  There will never be required fields (without default value) inside hideable sections

This also means that when a hidden field is marked as invalid on submit, the section will automatically be expanded so that the validation error becomes visible.

In fact, I believe it will be easiest for us if instead of removing the fields of a section from the DOM altogether (which has side effects for formik), we should use a visibility styling instead (e.g. `display: none` or `visibility: hidden`).

pkg/webui/components/form/section/index.js
```diff
@@ -0,0 +1,68 @@
...
+      {expandable && (
+        <Button className={style.button} type="button" secondary naked onClick={onExpandedChange}>
+          <Icon
+            className={style.icon}
+            icon={expanded ? 'arrow_drop_up' : 'arrow_drop_down'}
+            nudgeDown
+          />
+        </Button>
+      )}
+      <div>{expanded ? children : null}</div>
```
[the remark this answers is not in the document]
kschiffer: See also my other suggestion on using css visibility instead.

pkg/webui/locales/en.json
```diff
@@ -21,192 +21,6 @@
   "components.key-value-map.index.addEntry": "Add entry",
   "components.mobile-menu.index.loggedInAs": "Logged in as <b>{userId}</b>",
   "components.navigation.side.index.hideSidebar": "Hide sidebar",
-  "components.notification.details.index.showDetails": "Show details",
```
kschiffer, line 24: That doesn't look right. Try rerunning `mage js:translations`.

pkg/webui/components/form/section/index.js
```diff
@@ -0,0 +1,68 @@
...
+import Icon from '@ttn-lw/components/icon'
+
+import Message from '@ttn-lw/lib/components/message'
+
+import PropTypes from '@ttn-lw/lib/prop-types'
+
+import style from './section.styl'
+
+const FormSection = props => {
+  const { className, title, expandable, onExpand, initiallyExpanded, children } = props
```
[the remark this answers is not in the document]
bafonins: The idea was to use `<Form.Section />` in other places where we define `<Message />`. I changed this and added the `<Form.SubTitle />` for this.

pkg/webui/components/form/section/index.js
```diff
@@ -0,0 +1,80 @@
...
+
+const m = defineMessages({
+  expand: 'Expand section',
+  collapse: 'Collapse section',
+})
+
+// `<FormCollapseSection />` aggregates a set of form fields under common title as well as adds
+// functionality to hide or show the fields.
+const FormCollapseSection = props => {
+  const { className, id, title, onCollapse, initiallyCollapsed, children } = props
```
kschiffer, line 34: Just curious, why do you require an `id`?

bafonins: for the `aria-controls` attribute

PR #2820, opened 2020-07-01
Support connecting to Redis over TLS

### LINKED ISSUE #2166: TLS Configuration for Redis
#### Summary

We should add configuration options for connecting to Redis over TLS.

#### Why do we need this?

Security.

The `redis.Options` has a `TLSConfig` field, but we don't populate it.

#### What is missing? What do you want to see? How do you propose to implement this?

We should add a bool option that will populate that `TLSConfig` field with `component.GetTLSClientConfig(ctx, ...)`

htdvisser:
#### Summary

This PR adds support for connecting to Redis over TLS.

Closes #2166

#### Changes

- [x] Bunch of refactorings for TLS configuration
- [x] Added a TLS config to Redis
- [x] Updated docs

#### Testing

- Unit tests
- Manual testing with `redis-server --tls-port 6379 --port 0 --tls-cert-file cert.pem --tls-key-file key.pem --tls-ca-cert-file cert.pem --tls-auth-clients no` and `ttn-lw-stack start --redis.tls.require --redis.tls.root-ca cert.pem`

##### Regressions

This removes auto-reloading certificate files

#### Notes for Reviewers

- I initially implemented this in the Redis config, but discovered that our TLS configuration is a bit of a mess, and duplicated in multiple places, so I decided to start consolidating that a bit.
- In the process I removed auto-reloading of certificates. We added this for ACME certs, but with the builtin ACME support, I don't think we need this anymore. Moreover, I think it would be better to reload explicitly with a HUP signal instead of trying to watch files for changes.

#### Checklist


pkg/redis/redis.go
```diff
@@ -110,15 +116,34 @@ type FailoverConfig struct {
...
+		tlsConfig := &tls.Config{
+			MinVersion: tls.VersionTLS12,
+		}
+		err := c.TLS.Client.Apply(tlsConfig)
+		if err != nil {
+			return nil, err
+		}
+		conn, err = tls.DialWithDialer(dialer, network, addr, tlsConfig)
+		if err != nil {
+			return nil, fmt.Errorf("Error connecting to %q: %w", addr, err)
```
johanstokking: `redis: dial %q: %w`

htdvisser: This shouldn't have been committed in the first place.

htdvisser:
The reason for not taking context was that it's too easy to use it incorrectly. The only place where a context would make sense would be in the `Get(Client)Certificate` with the KeyVault, but in that place you'd want a request context, not a global context.

pkg/redis/redis.go
```diff
@@ -110,15 +116,34 @@ type FailoverConfig struct {
...
+	var (
+		conn net.Conn
+		err  error
+	)
+	dialer := &net.Dialer{Timeout: timeout}
+	if c.TLS.Require {
+		tlsConfig := &tls.Config{
+			MinVersion: tls.VersionTLS12,
+		}
+		err := c.TLS.Client.Apply(tlsConfig)
```
rvolosatovs: This is sharing memory across concurrent calls (`Apply` takes a pointer receiver), which is quite a code smell and may cause bugs in the future. Moreover, we don't need to call this on each `Dial`, just once should be enough

pkg/redis/redis.go
```diff
@@ -110,15 +116,34 @@ type FailoverConfig struct {
 	MasterName string   `name:"master-name" description:"Redis Sentinel master name"`
 }
 
-func dial(ctx context.Context, network, addr string) (net.Conn, error) {
+func (c Config) dial(ctx context.Context, network, addr string) (net.Conn, error) {
```
rvolosatovs: How about
```suggestion
func (c Config) makeDial(ctx context.Context, network, addr string) (func(net.Conn, error), error) {
```

Instead, which `newRedisClient` would call?
it's a very straight-forward change, which fits directly into current implementation and would allow us to reuse TLS config and initialization as well as fix the issue below

PR #2823, opened 2020-07-02
Fix join settings validation

kschiffer:
#### Summary
This quickfix PR fixes an issue of the validation of the `Home net ID` field in the join settings forms.

#### Changes
- Make `net_id` default to an empty string

#### Testing

Manual testing.

##### Regressions

This might break setting the `net_id` prop of end devices.

#### Notes for Reviewers
Noticed this via https://sentry.io/organizations/the-things-industries/issues/1757151966/?project=2682566

Without this patch, the byte input field will receive null after submitting, which is an invalid value. Since you initially set the validation to `nullableString()`, I'm not sure whether this patch introduces any side effects. I did not notice any when testing this though.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - End device events stream restart in the Console.
 - CLI was unable to read input from pipes.
 - Timezones issue in claim authentication code form, causing time to reverse on submission.
+- Errors during submit of the join settings for end devices in the Console.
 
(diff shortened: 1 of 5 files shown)


kschiffer:
I see. Then I suggest my pushed solution. I think in some scenarios it can be good to be able to provide a encode/decode function for the field. That enables us to transform arbitrary registry values into something that the field understands while being able to keep validation and casting closer to the actual registry value.

We're already using such approach for the `<UnitInput />` but I think this is actually something that all fields should be capable of.

PR #2837, opened 2020-07-02
Optimize DevAddr matching

### LINKED ISSUE #2665: Optimize device matching
#### Summary

Optimize device matching in Network Server

#### Why do we need this?

For Redis I/O, memory usage and overall compute performance when we increasing DevAddr reuse and increasing DevAddr size.

Currently, NS ranges over devices by DevAddr, and decodes the proto. The protos are big, containing the mac state and recent uplink and downlink messages, plus there can be many protos (hundreds) of duplicate DevAddr.

We need a materialized view in Redis for matching uplinks to devices by UID, before loading the matched device.

Store relevant information for matching in dedicated keys, so basically DevAddr, FCnt, NwkSIntKeys and UID.

A sorted set and/or hash map would be useful as this enables Redis sorting and from-querying by FCnt, getting the integrity keys and UID.

Update this materialized view on each uplink and join-accept.

We can have a rolling migration with fallback to the existing matching scheme.

Unit tests, potentially with asserting the Redis keys and their values in the test.

Can review

rvolosatovs:
#### Summary

Closes #2665 

#### Changes

- Sort in-Redis
- Only get one device at a time
- Share computations between instances
- Cache match for possible reuse

#### Testing

Tested with actual device using legacy DB contents

##### Regressions

The whole data uplink flow and, consequently,, Class A downlink

#### Notes for Reviewers

This is very much WIP and absolutely not ready for review, but opening by request from @johanstokking 

#### Checklist


rvolosatovs:
Blocked by #3052

johanstokking:
@rvolosatovs what is the status here?

rvolosatovs:
As you can see, it's a really big amount of changes necessary for this PR, so I'm extracting smaller parts. 
https://github.com/TheThingsNetwork/lorawan-stack/pull/3052
https://github.com/TheThingsNetwork/lorawan-stack/pull/3169
https://github.com/TheThingsNetwork/lorawan-stack/pull/3170

I was finally able to compile and run tests yesterday on the branch and found some failures, which I'm fixing
The testing PR itself is also flaky, so it needs some work still

rvolosatovs:
I fixed and improved Redis matching implementation, as well as updated the device registry tests, which now pass.
We also now get coverage in the actual `pkg/networkserver/redis` package.
@johanstokking The Redis implementation is ready for review and I do not intend to do any more changes to the algorithm (except possible cosmetic changes).

pkg/networkserver/grpc_gsns.go
```diff
@@ -951,23 +999,8 @@ func (ns *NetworkServer) handleDataUplink(ctx context.Context, up *ttnpb.UplinkM
...
-
-				ctx = matched.Context
-				pld.FullFCnt = matched.FCnt
-				up.DeviceChannelIndex = uint32(matched.ChannelIndex)
-				up.Settings.DataRateIndex = matched.DataRateIndex
-				logger = log.FromContext(ctx).WithFields(log.Fields(
-					"data_rate_index", up.Settings.DataRateIndex,
-					"device_channel_index", up.DeviceChannelIndex,
-				))
+				// TODO: Rematch device (ISSUE REF)
```
rvolosatovs: I will still address this and make some changes to the matching utility function(s)

johanstokking:
> #### Testing
> Tested with actual device using legacy DB contents

Did you also test with multiple devices that share the same DevAddr?

---

Main reviewer is @adriansmares

pkg/networkserver/redis/lua/deviceMatch.lua
```diff
@@ -0,0 +1,85 @@
...
+--
+-- KEYS[10]	- sorted list of uid, LastFCntUp, FNwkSIntKey, MACVersion of matching devices using 32-bit frame counters with rollover
+-- KEYS[11] - sorted list of uid, LastFCntUp, FNwkSIntKey, MACVersion of matching devices using 32-bit frame counters with rollover being processed
+--
+-- KEYS[12] - list of uids of devices matching pending session DevAddr
+-- KEYS[13] - list of uids of devices matching pending session DevAddr being processed
+--
+-- KEYS[14] - list of uids of devices matching either current or pending session DevAddr not present in either KEYS[2], KEYS[3], nor KEYS[4]
+-- KEYS[15] - list of uids of devices matching either current or pending session DevAddr not present in either KEYS[2], KEYS[3], nor KEYS[4] being processed
+-- NOTE: It is expected that count of devices using 16-bit frame counters << count of devices using 32-bit frame counters.
```
htdvisser: What does this mean exactly?

htdvisser (commented):
I think this could benefit from quite a number of inline comments to explain to future contributors what exactly is going on here.

pkg/networkserver/redis/lua/deviceMatch.lua
rvolosatovs: We expect most devices to use 32-bit frame counters

htdvisser: Okay, but are there any implications if that's not the case?

rvolosatovs: Not really - the implementation is optimized for this case (which, if not now, will be true in the future as newer LoRaWAN spec versions are more widely adopted), that's it.

htdvisser: Then maybe it's better to write something that says that it's optimized for devices with 32-bit counters. The way I read the current comment, it sounds like unexpected things may happen if the property doesn't hold.

rvolosatovs:
The implementation is ready for review. @adriansmares @johanstokking 

I will go through the implementation myself once again for proof-reading (this PR has been in the works for a while...) and add comments where necessary. It would be a great help if you could point out the especially cryptic fragments in need of commentary.

I will also perform extensive integration testing on Monday.

adriansmares (approved):
LGTM.

Fundamentally I think I understood the implementation itself, and outside of the comment that @htdvisser had about the `lua` comment clarity, I don't think there is much to add.

What I am currently missing for this PR is testing how does this perform compared to the old implementation. We moved a large part of the matching in Redis itself, and I'm still not sure how this will interact with other services that we currently host in Redis. 

Redis scripts (such as the `lua` here) basically [stop-the-world](https://redis.io/commands/eval#atomicity-of-scripts) while running, and we share this Redis instance between all of our components. It may be a good idea to consider breaking down the Redis instance - something for TTI, and definitely outside the scope of this PR.

Regarding testing - what can we actually do that is meaningful ? We ruled out getting traffic from _real world clusters_, for obvious reasons, but how can we actually test this well ? What kind of simulators do we need for a test that would for example simulate 5k, 10k, 20k devices sending traffic at different intervals ? I don't think what we have on `staging1` is enough in order to test the waters of this refactor, since the situation between a cluster with 3 active devices and a cluster with 30k active devices is very, very different.

rvolosatovs:
> What I am currently missing for this PR is testing how does this perform compared to the old implementation. We moved a large part of the matching in Redis itself, and I'm still not sure how this will interact with other services that we currently host in Redis.
> 
> Redis scripts (such as the `lua` here) basically stop-the-world while running, and we share this Redis instance between all of our components. It may be a good idea to consider breaking down the Redis instance - something for TTI, and definitely outside the scope of this PR.

Regarding
> What I am currently missing for this PR is testing how does this perform compared to the old implementation

specifically - I'm creating devices on 3.9 and then switching to this branch for testing. So backwards-compatibility should be fine.

For the rest:
The Lua operations are designed (and supposed anyway) to run quite fast. We should definitely profile this, but I think for now this should not cause issues unless we have thousands of devices matching the same DevAddr (which we don't right?). But it would make sense indeed to investigate using separate Redis instances for components, it's a generally a good idea I think, especially considering that Redis is single-threaded, so having a Redis instance per-component would prove to be beneficial IMO.

> Regarding testing - what can we actually do that is meaningful ? We ruled out getting traffic from real world clusters, for obvious reasons, but how can we actually test this well ? What kind of simulators do we need for a test that would for example simulate 5k, 10k, 20k devices sending traffic at different intervals ? I don't think what we have on staging1 is enough in order to test the waters of this refactor, since the situation between a cluster with 3 active devices and a cluster with 30k active devices is very, very different.

That's my issue as well, yes. We could write a script to generate a bunch of devices and simulate a bunch of random traffic to stress-test the network as a whole? Perhaps we can get some inspiration from CyberEye.

adriansmares:
> specifically - I'm creating devices on 3.9 and then switching to this branch for testing. So backwards-compatibility should be fine.

I'm referring strictly to performance here - correctness can be fixed, but performance is a design property that can be very costly to fix later (after all, it is the reason why we have this PR in the first place).

> That's my issue as well, yes. We could write a script to generate a bunch of devices and simulate a bunch of random traffic to stress-test the network as a whole? Perhaps we can get some inspiration from CyberEye.

I don't think `simulate` from `ttn-lw-cli` is enough. It does its job and it's perfect for reproduction purposes, but it does not include a fully fledged flow. I would be far more comfortable releasing `3.10` having complete, parallel flows as part of the test. 

What I wish is:
1. Simulated device sends `JoinRequest` and waits for `JoinAccept`
2. Simulated device starts sending uplinks on different FPorts (a small set, let's say `42, 43, 44`) at different intervals (some send often, like every 10 seconds with jitter, some are more lazy and send every 5-10 minutes with jitter)
3. After a number of uplinks (let's say 500-1000 ?) the device rejoins and this starts again

I don't expect (and please correct me if I'm wrong) to have differences in 1, so measuring it is not very important there. 
What would be good to measure is the round trip time in the NS (`UplinkMessage.ReceivedAt` - `ApplicationUp.ReceivedAt` - we can do this using the AS) and the evolution of the metrics that we already have:
- How did the Redis memory usage evolve post upgrade ?
- How did the Redis CPU usage evolve post upgrade ?
- How did the NS memory usage (heap + stack + total memory reserved in the OS) usage evolve post upgrade ?
- How did the NS CPU usage evolve post upgrade ?
- How did `pps` evolve post upgrade ?
- How did info/warning/error rates evolve post upgrade ?

rvolosatovs:
As discussed in the v3 call, what would be great, is if we had a stack instance deployed in an environment, which is as close to the production environment, as possible, where:
- The stack gets a snapshot of actual production databases
- Actual production traffic is directed (duplicated) to that instance via PB
- Downlinks/sending of data to applications it either turned off or somehow mocked, such that we could still observe what is actually sent

Ideally, what we would have is an encapsulated stack, where all incoming requests are duplicated and all outgoing requests are blocked. We can then observe not only the performance etc., but also compare it side-by-side with the actual production instance in term of both performance and correctness.

rvolosatovs:
@adriansmares also note that our biggest issue is currently the amount of data sent back and forth, since that costs us money. This PR fixes the issue - the speed of matching, for example, is not really an issue - we have plenty of time to handle all of that for Class A, since minimum RX delay is 1 second and by default we use 5 seconds, which is plenty. (for other classes it does not matter)

johanstokking (approved):

PR #2839, opened 2020-07-03
Add key vault caching

### LINKED ISSUE #36: Cache unwrapping keys to reduce load on key vault
**Summary:** 

Cache unwrapping keys to reduce load on key vault

**What do you see now?**

Unwrapping keys on-the-fly, also inside loops.

Keys should not be stored (long) in memory, but they can be cached briefly within the same function.

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/1218 by @johanstokking

adriansmares:
#### Summary

Closes #36 

#### Changes

- Added a cached `KeyVault` wrapper. It currently caches only the unwrapping process, based on the KEK label and the ciphertext.
- Added the cached wrapper as part of the configuration.
- Added cache metrics to `pkg/crypto`.

#### Testing

Unit testing with a mock key vault.

##### Regressions

Enabling the caching may affect only the unwrapping process. However, I haven't seen any changed behavior during testing.

#### Notes for Reviewers

I haven't included a default configuration for this - any thoughts on what 'sane defaults' we should add ? I'd propose 5 minutes TTL with a cache size of ~1000.

#### Checklist

### DIFF
--- a/pkg/config/shared.go
+++ b/pkg/config/shared.go
@@ -131,23 +131,33 @@ type Rights struct {
 	TTL time.Duration `name:"ttl" description:"Validity of Identity Server responses"`
 }
 
+// KeyVaultCache represents the configuration for key vault caching.
+type KeyVaultCache struct {
+	Size int           `name:"size" description:"Cache size"`
+	TTL  time.Duration `name:"ttl" description:"Cache elements time to live"`
+}
+
 // KeyVault represents configuration for key vaults.
 type KeyVault struct {
 	Provider string            `name:"provider" description:"Provider (static)"`
+	Cache    KeyVaultCache     `name:"cache"`
 	Static   map[string][]byte `name:"static"`
 }
 
 // KeyVault returns an initialized crypto.KeyVault based on the configuration.
 func (v KeyVault) KeyVault() (crypto.KeyVault, error) {
+	vault := cryptoutil.EmptyKeyVault
 	switch v.Provider {
 	case "static":
 		kv := cryptoutil.NewMemKeyVault(v.Static)
 		kv.Separator = ":"
 		kv.ReplaceOldNew = []string{":", "_"}
-		return kv, nil
-	default:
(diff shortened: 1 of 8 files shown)


pkg/crypto/cryptoutil/keyvault_cache.go
```diff
@@ -0,0 +1,103 @@
...
+			Subsystem: subsystem,
+			Name:      "cache_miss",
+			Help:      "Number of cache misses",
+		},
+		[]string{"cache"},
+	),
+}
+
+func init() {
+	metrics.MustRegister(cMetrics)
```
htdvisser, line 67: We usually keep metrics and such in `metrics.go` or `observability.go`

pkg/config/shared.go
```diff
@@ -131,23 +131,33 @@ type Rights struct {
 	TTL time.Duration `name:"ttl" description:"Validity of Identity Server responses"`
 }
 
+// KeyVaultCache represents the configuration for key vault caching.
+type KeyVaultCache struct {
+	Size int           `name:"size" description:"Cache size"`
```
htdvisser, line 136: Maybe describe that 0 means off. Sometimes 0 means unlimited.

pkg/crypto/cryptoutil/keyvault_cache.go
```diff
@@ -0,0 +1,103 @@
...
+
+type cachedVault struct {
+	crypto.KeyVault
+	unwrapCache gcache.Cache
+}
+
+func NewCacheKeyVault(main crypto.KeyVault, ttl time.Duration, size int) crypto.KeyVault {
+	return &cachedVault{
+		KeyVault:    main,
+		unwrapCache: gcache.New(size).Expiration(ttl).ARC().Build(),
```
htdvisser, line 83: Should you really include `.Expiration(ttl)` if it's `0`?

PR #2850, opened 2020-07-07
Split up oauth package (account app scaffolding)

kschiffer:
#### Summary
This PR takes care of renaming the oauth app to account app.

References #1422 

#### Changes
- Rename oauth app to account app
  - Rename `oauth` package to `accountapp`
  - Rename `oauth.js` entrypoint to `account.js`
  - Rename `oauth` webui folder to `account`
  - Rename static assets
  - Update module import statements
  - Update various configs accordingly
    - `package.json` module name mapper for jest
    - webpack config for module aliases (`@oauth` to `@account`)
    - eslint alias config
    - Update stack config structure accordingly
       - config names, e.g. `TTN_LW_OAUTH_UI_JS_FILE` to `TTN_LW_ACCOUNTAPP_UI_JS_FILE`
       - default values, e.g. entrypoints from `oauth.js` to `account.js`
  - Update documentation accordingly 
    - Renaming all old `oauth` occurrences
    - Updating some legacy info
  - Rename github commit topic from `oauth` to `account`
- Some preliminary work for renaming user facing occurrences of `Login` / `Logout` to `Sign in` / `Sign out`
- Some related reducer fixes
- Updating go tests accordingly

#### Testing

I did manual testing of all views of the account app.

##### Regressions

This might break views and functionality in the account app, e.g. when old naming have not been updated

#### Notes for Reviewers
Note that the target branch is again `feature/1422-account-app`, so this is not targeting master/develop and is preliminary work for upcoming account app related PRs. The point of this is again to reduce review diffs, chunking things down into bits that are scoped and easy to review until eventually merging `feature/1422-account-app` into `develop` when all parts have been merged into it.

This PR is a primarily a refactor PR that changes almost no functionality, except for displayed messages and updated routes. Actual feature implementation will follow right after this PR.

I decided to rename `Login`/`Logout` to `Sign in`/`Sign out` since it is somewhat common wording for web applications right now. [This is a very old discussion](https://blog.benjamincharity.com/log-in-vs-login-vs-sign-in/) but ultimately I think `Sign in`/`Sign out` suits us better because it's common, friendlier and more mature.

<details><summary>Running this in develop mode</summary>

It's necessary to update the config variables to mount this in development mode:
```
export TTN_LW_CONSOLE_OAUTH_AUTHORIZE_URL="http://localhost:8080/account/authorize"
export TTN_LW_CONSOLE_OAUTH_LOGOUT_URL="http://localhost:8080/account/logout"
export TTN_LW_CONSOLE_OAUTH_TOKEN_URL="http://localhost:8080/account/token"
export TTN_LW_IS_ACCOUNTAPP_UI_CANONICAL_URL="http://localhost:8080/account"
export TTN_LW_IS_ACCOUNTAPP_UI_JS_FILE="libs.bundle.js account.js"
export TTN_LW_IS_ACCOUNTAPP_UI_JS_FILE="libs.bundle.js account.js"
```
</details>

Main reviewers: @bafonins, @htdvisser 

#### Checklist


kschiffer:
Understood. Reverting the config should be no problem I think, but to me, it would be weird to serve the account app under `/oauth`. 

Maybe we should then split the concerns more also in our routings:

*   `/account` for all routes that actually relate to the frontend account app
    *   `/account/api/auth/login`
    *   `/account/api/auth/logout`
    *   `/account/api/me`
    *   `account/*` for frontend handled routes
*   keeping `/oauth` for all other routes that deal with OAuth authorization
    *   `/oauth/logout`
    *   `/oauth/authorize`
    *   `/oauth/local-callback`
    *   `/oauth/token`
    *   `/oauth` -> redirect to `/account`

What do you think?

kschiffer:
cc @johanstokking

htdvisser:
I like the idea of splitting this into a pure OAuth part and an Account part, that gives a nice separation of concerns while also keeping everything compatible.

kschiffer:
Are you ok with renaming the `oauth` package to `accountapp` as well?

htdvisser:
If we split them, I think it would be better to keep OAuth functionality in the `oauthapp` package, and write account management features in the `accountapp` package (maybe removing the `app` suffixes)

kschiffer:
Yeah, I was considering that but wanted to avoid the refactoring hassle. But I guess you're right, this makes most sense.

johanstokking:
I agree with the conclusion here

kschiffer:
> We can't break config in the v3 major. This would currently happen to existing deployments that use the `is.oauth.*` config options. It would also happen to existing deployments that use the default value for `is.oauth.ui.css-file`, `is.oauth.ui.js-file` and `is.oauth.ui.icon-prefix` with custom (CDN) branding.

This is then a bit problematic. If we split the oauth package up, then configs like  `is.oauth.ui.css-file`, `is.oauth.ui.js-file` and `is.oauth.ui.icon-prefix` should really be `account.ui.*`.

Is there any way to use config aliases or something like that? @htdvisser

htdvisser:
There are no aliases in our config. We could fall back to different config options, like we do when configuring the Redis cache, with fallback to the Redis store config.

However, for OAuth/Account config I'm afraid it would be very unintuitive to use, because we'd also have to handle the legacy defaults, not just empty values. Honestly, I think we're much better off keeping the `is.oauth.*` options where they are (and use those for the slimmed down OAuth app), and add new `is.account.*` options for the new account app.

kschiffer:
After working on this for a bit:

*   The frontend part of the oauth app would basically consist of the authorization screen only
*   I think it's very annoying having to retain the `is.oauth.ui.*`  configs just for that
    *   It's also not good from UX perspective to present the authorization as separate entity (for the user it should be a part of the account app)
*   Also, I would really like to use one JS entry-point for both apps `account.js` to avoid more duplicated boilerplates. This won't be possible unless we change the default values for oauth.
*   I'm also a little bit unsure about the exact restrictions of how we can alter the stack configuration without breaking the CC
*   I'm a bit out of luck with finding a viable solution here and really need some input

I also had the idea to use the `OAuth.UI`  config for both the account app and (remaining) OAuth app, but then we'd need some way to get the correct canonical URL config, which is basically the only thing that would differ between the two.

PR #2854, opened 2020-07-07
Make gateway not connected errors consistent

### LINKED ISSUE #2853: Make gateway not connected errors consistent
#### Summary

When caching gateway connection stats on Redis, querying the stats of a not-connected Gateway results in a `redis not found` error, instead of `gateway not connected`

#### Steps to Reproduce

1. Start stack with cache enabled
2. Get connection stats of a gateway that is not connected.

#### What do you see now?

```
$ ttn-lw-cli gtw connection-stats gtw1
error:pkg/redis:not_found (entity not found)
    correlation_id=67ce91f04b7b41f49b78ba991da14bb8
exit status 255
```

```bash
$ cli gtw connection-stats gtw1
error:pkg/gatewayserver:not_connected (gateway `gtw1` not connected)
    gateway_uid=gtw1
    correlation_id=e17381fb1e9c4828b16fd30226465efe
exit status 255
```

#### Environment

3.8.5, with `cache.service=redis`

Catch the redis not found error

Test locally

yes

neoaggelos:
#### Summary

Closes #2853

#### Changes

- Catch Redis not found error, convert to gateway not connected.

#### Testing

Test locally

##### Regressions

There should not be any.

#### Notes for Reviewers

Basing on master, this is fixing an inconsistent error message.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Inconsistent error message responses when retrieving connection stats from GS if the gateway is not connected.
+
 ### Security
 
 ## [3.8.5] - 2020-07-06
--- a/pkg/gatewayserver/grpc.go
+++ b/pkg/gatewayserver/grpc.go
@@ -18,6 +18,7 @@ import (
 	"context"
 
 	"go.thethings.network/lorawan-stack/v3/pkg/auth/rights"
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 	"go.thethings.network/lorawan-stack/v3/pkg/unique"
 )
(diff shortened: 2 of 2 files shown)


pkg/gatewayserver/grpc.go
```diff
@@ -28,16 +29,19 @@ func (gs *GatewayServer) GetGatewayConnectionStats(ctx context.Context, ids *ttn
 		return nil, err
 	}
 
+	uid := unique.ID(ctx, ids)
 	if gs.statsRegistry != nil {
 		stats, err := gs.statsRegistry.Get(ctx, *ids)
 		if err != nil || stats == nil {
```
KrishnaIyer, line 35: Should this be `&&`?

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix inconsistent error message responses when retrieving connection stats from GS if the gateway is not connected.
```
johanstokking, line 20: Remove "Fix"; it's already listed under Fixed

pkg/gatewayserver/grpc.go
johanstokking: No, but in case of `nil` I think the `errNotConnected` should be returned. `nil` implies not found here I believe.

neoaggelos: To be honest, the `err != nil` is enough here. We never return `nil` without an error, or an error without a `nil` response. Yet, we do this to be on the safe side and avoid returning a nil response.

PR #2855, opened 2020-07-07
Support generic encryption/decryption in KeyVault

KrishnaIyer:
#### Summary

Support generic encryption/decryption in KeyVault. References https://github.com/TheThingsIndustries/lorawan-stack/pull/2087

#### Changes

- Add `Encrypt` / `Decrypt` to Keyvault.
- Add mem/empty implementation.
- Testing

#### Testing

Unit tests

##### Regressions

No

#### Checklist

### DIFF
--- a/pkg/crypto/cryptoutil/cryptoutil.go
+++ b/pkg/crypto/cryptoutil/cryptoutil.go
@@ -26,7 +26,7 @@ import (
 
 var (
 	errKEKNotFound         = errors.DefineNotFound("kek_not_found", "KEK with label `{label}` not found")
-	errKeyNotFound         = errors.DefineNotFound("key_not_found", "Key with ID `{id}` not found")
+	errKeyNotFound         = errors.DefineNotFound("key_not_found", "key with ID `{id}` not found")
 	errCertificateNotFound = errors.DefineNotFound("certificate_not_found", "certificate with ID `{id}` not found")
 )
 
--- a/pkg/crypto/cryptoutil/cryptoutil.go
+++ b/pkg/crypto/cryptoutil/cryptoutil.go
@@ -26,6 +26,7 @@ import (
 
 var (
 	errKEKNotFound         = errors.DefineNotFound("kek_not_found", "KEK with label `{label}` not found")
+	errKeyNotFound         = errors.DefineNotFound("key_not_found", "Key with ID `{id}` not found")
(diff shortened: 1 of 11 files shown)


pkg/crypto/cryptoutil/cryptoutil.go
```diff
@@ -26,6 +26,7 @@ import (
 
 var (
 	errKEKNotFound         = errors.DefineNotFound("kek_not_found", "KEK with label `{label}` not found")
+	errKeyNotFound         = errors.DefineNotFound("key_not_found", "Key with ID `{id}` not found")
```
johanstokking, line 29: Lowercase `key`

KrishnaIyer:
> Maybe a stupid question, but can't we use the existing KEKs for encryption? So that we can use them for wrapping (RFC 3394) and for encryption (AES-GCM)?

Yes they can be. But there are 2 reasons why I explicitly use a separate key
- Separation of concerns: Using the same master key to encrypt different types of data is not really a good idea. So we can use KEKs for Keys(LW Keys) and Generic Keys for generic data. If the AS in the future wanted to encrypt something that's not a LW Key, it can use the Generic Key.
- This makes encryption available to all components in the future just by using a Key ID, without the need to update the keyvault. Ex: Currently, if the GCS needs to encrypt something, we need to update `pkg/keyvault` and define a `GCSKEKLabel`.

pkg/crypto/encryption.go
```diff
@@ -0,0 +1,69 @@
...
+	gcm, err := cipher.NewGCM(cipherBlock)
+	if err != nil {
+		return nil, err
+	}
+	nonce := make([]byte, gcm.NonceSize())
+	_, err = io.ReadFull(rand.Reader, nonce)
+	if err != nil {
+		return nil, err
+	}
+	return gcm.Seal(nonce, nonce, plaintext, nil), nil
```
htdvisser, line 47: The [example](https://pkg.go.dev/crypto/cipher@go1.14.4?tab=doc#NewGCM) has the first argument to `Seal` as `nil`.

KrishnaIyer: Yeah that's in case the `nonce` is stored separately in the database. When decrypting, we need to enter the same nonce that was used during encryption. By using it as the first parameter, we append it to the cipher from where `Open` can [use](https://github.com/TheThingsNetwork/lorawan-stack/blob/5a638382966edeb0c852839243aa4905a8dfe76c/pkg/crypto/encryption.go#L64) it.

PR #2865, opened 2020-07-08
Support Gateway Secrets

KrishnaIyer:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/1979

#### Changes

- Add a secret blob to the Gateway Proto along with rights and RPCs to the gateway registry.
- Support storing/retrieving in IS
- Add CLI support to store/retrieve the secrets as plaintext.
- Modify GCS to use this value in the update-info response.

#### Testing

Tested using locally built CLI.

- [x] Add Unit Tests

CLI test
```
#  lbs-lns-token: f0253aa2bb788e9188f01cbf12d71b69
#  cups-token: 9f0ea9c3f12d63ac079c21242f57cc6d
$ ./ttn-lw-cli gateways store-secret test-gateway lbs-lns-token cups-token
Please enter the plaintext value for key lbs-lns-token:********************************
Please enter the plaintext value for key cups-token:********************************

$ ./ttn-lw-cli gateways retrieve-secret test-gateway
{
  "values": {
    "cups-token": "OWYwZWE5YzNmMTJkNjNhYzA3OWMyMTI0MmY1N2NjNmQ=",
    "lbs-lns-token": "ZjAyNTNhYTJiYjc4OGU5MTg4ZjAxY2JmMTJkNzFiNjk="
  }
}
```
- [x] Add CUPS test
[cups_tests.txt](https://github.com/TheThingsNetwork/lorawan-stack/files/4948568/tests.txt)

- [ ] Test cases using TTIG

##### Regressions

Covered in the CUPS test above.

#### Notes for Reviewers

This normally goes to `develop` but current `master` is bumped.

#### Checklist


KrishnaIyer:
@bafonins / @kschiffer / @rvolosatovs : No need for review. I can't somehow remove you here.

api/gateway.proto
```diff
@@ -127,8 +127,13 @@ message Gateway {
   google.protobuf.Duration schedule_anytime_delay = 19 [(gogoproto.stdduration) = true, (gogoproto.nullable) = true];
   // update the location of this gateway from status messages
   bool update_location_from_status = 21;
+  // An encypted secret value for this gateway.
+  // When set, this is currently sent as the lnsKey parameter in the Basic Station CUPS response. This can also be extended to store other values such as client TLS private key.
+  // This value cannot be set/retrieved in Gateway API calls. Use the StoreGatewaySecret, RetrieveGatewaySecret requests.
+  // It can however be deleted by DeleteGateway since it's part of the entity.
+  bytes secret = 22;
```
johanstokking: I think this is too simple of a field and description.

I would suggest a `message Secrets` with keyed secrets (map) with a reference to the encryption key (or the encrypted DEK). This could be reusable for entities – I can imagine use cases for `Application` as well – or inline in `Gateway`.

See https://github.com/packetbroker/api/blob/master/v3/keys.proto for some inspiration.

api/gateway.proto
```diff
@@ -285,3 +290,18 @@ message GatewayConnectionStats {
   // Statistics for each sub band.
   repeated SubBand sub_bands = 10;
 }
+
+// The plain text value of the gateway secret.
+message GatewaySecretPlainText{
+  string value = 1;
```
johanstokking: Encrypted data doesn’t have to be a string

Ideally it would be a marshaled proto that is also defined here, like `BasicStationLNSKey`, to make it fully typed.

Otherwise, consider `google.protobuf.Struct`.

pkg/identityserver/config.go
```diff
@@ -67,6 +67,7 @@ type Config struct {
 		SMTP         smtp.Config          `name:"smtp"`
 		Templates    emailTemplatesConfig `name:"templates"`
 	} `name:"email"`
+	GatewaySecretKeyID string `name:"gateway-secret-key-id" description:"ID of the key used to encrypt gateway secrets at rest"`
```
johanstokking, line 78: This would only be needed for encrypting new keys. If you store the DEK/KEK label with the encrypted data, you can easily rotate keys by changing only the config value for newly encrypted secrets.

pkg/identityserver/gateway_registry.go
```diff
@@ -64,6 +72,9 @@
 		req.FrequencyPlanIDs = []string{req.FrequencyPlanID}
 	}
 
+	// Don't allow setting the secret field while creating the gateway.
+	req.Gateway.Secret = nil
```
johanstokking: This would go with `ttnpb.ProhibitFields()`

pkg/identityserver/gateway_registry.go
```diff
@@ -116,6 +127,9 @@
 	}
 	req.FieldMask.Paths = cleanFieldMaskPaths(ttnpb.GatewayFieldPathsNested, req.FieldMask.Paths, getPaths, []string{"frequency_plan_id"})
 
+	// Don't allow getting the secret field.
+	req.FieldMask.Paths = ttnpb.ExcludeFields(req.FieldMask.Paths, "secret")
```
johanstokking: Idem

johanstokking:
Also file issue for managing the new rights in the Console. Or try yourself!

api/rights.proto
```diff
@@ -119,6 +119,10 @@ enum Right {
   RIGHT_GATEWAY_STATUS_READ = 38;
   // The right to view view gateway location.
   RIGHT_GATEWAY_LOCATION_READ = 39;
+  //The right to encrypt and store the a secret value associated with this gateway.
+  RIGHT_GATEWAY_WRITE_SECRET = 57;
+  //The right to decrypt and retrieve the LBS secret value.
+  RIGHT_GATEWAY_READ_SECRET = 58;
```
htdvisser: - "the a" secret?
- Or the LBS secret?
- Or plural secrets?
- Spaces after `//`

KrishnaIyer:
@johanstokking / @htdvisser : 

This was quite tricky to locally test but I've done E2E tests locally. Please check the tests in the description.

pkg/identityserver/config.go
KrishnaIyer: Yeah this works for decrypting since I use the [KeyID that's stored in the secret](https://github.com/TheThingsNetwork/lorawan-stack/blob/2c21c4f2460dfb0105de655fd74951423dadb447/pkg/identityserver/gateway_registry.go#L401).

But this is still needed for [encryption](https://github.com/TheThingsNetwork/lorawan-stack/blob/2c21c4f2460dfb0105de655fd74951423dadb447/pkg/identityserver/gateway_registry.go#L355) right? There's nowhere else I see where I can get this from.

api/gateway.proto
```diff
@@ -285,3 +290,18 @@ message GatewayConnectionStats {
   // Statistics for each sub band.
   repeated SubBand sub_bands = 10;
 }
+
+// The plain text value of the gateway secret.
+message GatewaySecretPlainText{
+  map<string,bytes> values = 1 [(validate.rules).map.keys.string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$" , max_len: 36}];
+}
```
johanstokking, line 297: Shouldn't we have a map with keyed secrets per entity, whose value is a marshaled message?

api/gateway.proto
```diff
@@ -285,3 +290,18 @@ message GatewayConnectionStats {
...
+
+message StoreGatewaySecretRequest{
+  GatewayIdentifiers gateway_ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (validate.rules).message.required = true];
+  GatewaySecretPlainText plain_text = 2  [(gogoproto.nullable) = false];
+}
+
+message RetrieveGatewaySecretRequest{
+  GatewayIdentifiers gateway_ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (validate.rules).message.required = true];
```
johanstokking, line 305: No need to include this in the response

api/gateway_services.proto
```diff
@@ -76,6 +76,23 @@ service GatewayRegistry {
...
   };
+
+  // SetGatewayLBSSecret is used to encrypt a plaintext value and store it along with the gateway.
+  // Requires the RIGHT_GATEWAY_WRITE_SECRET right.
+  rpc StoreGatewaySecret(StoreGatewaySecretRequest) returns (google.protobuf.Empty) {
+    option (google.api.http) = {
+      patch: "/gateways/{gateway_ids.gateway_id}/secret"
+      body: "*"
+    };
+  };
```
johanstokking, line 87: So there is still one secret per gateway entity, which gets patched in its entirety here?

How would we support multiple secrets in the future?

api/gateway.proto
johanstokking: So then, `Gateway` can contain multiple independent secrets, keyed by string, and one of them is `basicstation`, which is the encrypted marshaled `GatewayBasicStationSecretData` which has two string fields

cmd/ttn-lw-cli/commands/flags.go
```diff
@@ -129,6 +130,19 @@ func getAPIKeyID(flagSet *pflag.FlagSet, args []string, i int) string {
...
 
+func getPlainTextValue(flagSet *pflag.FlagSet, args []string, i int) string {
+	var plainTextValue string
+	if len(args) > 0+i {
+		if len(args) > 1+i {
+			logger.Warn("Multiple plain text values found in arguments, considering only the first")
+		}
+		plainTextValue = args[0+i]
+	} else {
+		plainTextValue, _ = flagSet.GetString("plain-text-value")
```
johanstokking, line 141: It's a bad practice to take secrets from flags. Can we avoid this by piping or prompt? Or would that become really unfriendly?

api/secrets.proto
```diff
@@ -0,0 +1,30 @@
...
+
+package ttn.lorawan.v3;
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+import "github.com/envoyproxy/protoc-gen-validate/validate/validate.proto";
+
+option go_package = "go.thethings.network/lorawan-stack/v3/pkg/ttnpb";
+
+// Secrets contains key-value pairs of secret values as well as the ID of the encryption key.
+message Secrets{
```
htdvisser, line 25: Nit:

```suggestion
message Secrets {
```

api/secrets.proto
```diff
@@ -0,0 +1,30 @@
...
+  map<string,bytes> values = 1 [(validate.rules).map.keys.string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$" , max_len: 36}];
+	// ID of the Key used to encrypt the secret.
```
htdvisser, line 27: Indent with spaces please

api/gateway.proto
```diff
@@ -285,3 +290,18 @@ message GatewayConnectionStats {
   // Statistics for each sub band.
   repeated SubBand sub_bands = 10;
 }
+
+// The plain text value of the gateway secret.
+message GatewaySecretPlainText{
```
htdvisser, line 295: Nit:

```suggestion
message GatewaySecretPlainText {
```

Gofmt made you lazy, @KrishnaIyer 😏

cmd/ttn-lw-cli/commands/flags.go
htdvisser: This func seems to be unused.

cmd/ttn-lw-cli/commands/gateways.go
```diff
@@ -399,6 +402,67 @@ var (
...
+		Short: "Send plaintext values to be encrypted and stored.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			gtwID, err := getGatewayID(cmd.Flags(), firstArgs(1, args...), true)
+			if err != nil {
+				return err
+			}
+			if len(args) != 2 {
+				return errNoOfArguments
+			}
+			keys := strings.Split(args[1], ",")
```
htdvisser, line 419: Why have the keys as a single argument, instead of just taking `args[1:]`?

cmd/ttn-lw-cli/commands/gateways.go
```diff
@@ -399,6 +402,67 @@ var (
...
+			})
+			if err != nil {
+				return err
+			}
+			return nil
+		},
+	}
+	gatewaysRetrieveSecretCommand = &cobra.Command{
+		Use:   "retrieve-secret [gateway-id]",
+		Short: "Retrive the decrypted Plaintext values. Secrets are printed in base64.",
```
htdvisser, line 448: I think it's a bit confusing that `retrieve-secret` uses base64 while `store-secret` doesn't.

pkg/identityserver/gateway_registry.go
```diff
@@ -40,11 +42,18 @@ var (
 		"gateway.delete", "delete gateway",
 		ttnpb.RIGHT_GATEWAY_INFO,
 	)
-)
+	evtStoreSecret = events.Define(
```
htdvisser, line 45: This would be `evtStoreGatewaySecrets` and `evtRetrieveGatewaySecrets`

pkg/identityserver/gateway_registry.go
```diff
@@ -40,11 +42,18 @@ var (
...
+	)
+	evtRetrieveSecret = events.Define(
+		"gateway.secrets.retrieve", "retrieve gateway secrets",
+		ttnpb.RIGHT_GATEWAY_INFO,
+	)
 
-var (
 	errAdminsCreateGateways = errors.DefinePermissionDenied("admins_create_gateways", "gateways may only be created by admins, or in organizations")
 	errGatewayEUITaken      = errors.DefineAlreadyExists("gateway_eui_taken", "a gateway with EUI `{gateway_eui}` is already registered as `{gateway_id}`")
+	errInvalidFieldmask     = errors.DefineInvalidArgument("invalid_fieldmask", "invalid fieldmask")
```
htdvisser, line 56: This doesn't belong here.

pkg/identityserver/gateway_registry.go
```diff
@@ -122,6 +134,13 @@ func (is *IdentityServer) getGateway(ctx context.Context, req *ttnpb.GetGatewayR
 	}
 	req.FieldMask.Paths = cleanFieldMaskPaths(ttnpb.GatewayFieldPathsNested, req.FieldMask.Paths, getPaths, []string{"frequency_plan_id"})
 
+	// Don't allow getting the secrets field.
+	if err := ttnpb.ProhibitFields(req.FieldMask.Paths,
+		"secrets",
+	); err != nil {
+		return nil, errInvalidFieldmask.WithCause(err)
+	}
```
htdvisser, line 142: This belongs in https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/ttnpb/field_mask_validation.go#L849 similar to what we do with OAuth clients.

pkg/identityserver/gateway_registry.go
```diff
@@ -183,6 +202,13 @@ func (is *IdentityServer) listGateways(ctx context.Context, req *ttnpb.ListGatew
...
```
htdvisser, line 210: This belongs in https://github.com/TheThingsNetwork/lorawan-stack/blob/master/pkg/ttnpb/field_mask_validation.go#L850 similar to what we do with OAuth clients.

pkg/identityserver/gateway_registry.go
```diff
@@ -257,6 +283,13 @@ func (is *IdentityServer) updateGateway(ctx context.Context, req *ttnpb.UpdateGa
 		return nil, err
 	}
 
+	// Don't allow getting the secrets field.
+	if err := ttnpb.ProhibitFields(req.FieldMask.Paths,
+		"secrets",
+	); err != nil {
+		return nil, errInvalidFieldmask.WithCause(err)
+	}
```
htdvisser, line 291: Same story

pkg/identityserver/gateway_registry.go
```diff
@@ -40,11 +42,18 @@ var (
 		"gateway.delete", "delete gateway",
 		ttnpb.RIGHT_GATEWAY_INFO,
 	)
-)
+	evtStoreSecret = events.Define(
+		"gateway.secrets.store", "store gateway secrets",
+		ttnpb.RIGHT_GATEWAY_INFO,
+	)
+	evtRetrieveSecret = events.Define(
```
htdvisser, line 49: This is unused

pkg/identityserver/store/secrets.go
```diff
@@ -0,0 +1,39 @@
...
+package store
+
+import "go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+
+// Secrets model.
+type Secrets struct {
+	Model
+	SoftDelete
+
+	Data []byte `gorm:"type:BYTEA"`
```
htdvisser, line 24: This doesn't feel right. I think it's better to store entity in the same way we store attributes, so something like

```go
type Secret struct {
	ID string `gorm:"type:UUID;primary_key;default:gen_random_uuid()"`

	EntityID   string `gorm:"type:UUID;index:attribute_entity_index;not null"`
	EntityType string `gorm:"type:VARCHAR(32);index:attribute_entity_index;not null"`

	Key             string `gorm:"type:VARCHAR"`
	EncryptionKeyID string `gorm:"type:VARCHAR"`
	Ciphertext      string `gorm:"type:BYTEA"`
}
```

htdvisser:
@KrishnaIyer you updated the PR while I was reviewing it, so my review may be outdated on some points. Please re-request review when the PR is not actively being edited.

cmd/ttn-lw-cli/commands/flags.go
KrishnaIyer: Yeah indeed. I forgot to remove it. We get the secret value using `gopass` similar to the passwords.

KrishnaIyer:
Moving this from Slack @johanstokking ;

Ok so currently we store/fetch secrets like so;
$ ./ttn-lw-cli gateways store-secret test-gateway lbs-lns-token cups-token
Please enter the plaintext value for key lbs-lns-token:********************************
Please enter the plaintext value for key cups-token:********************************

$ ./ttn-lw-cli gateways retrieve-secret test-gateway
{
  "values": {
    "cups-token": "OWYwZWE5YzNmMTJkNjNhYzA3OWMyMTI0MmY1N2NjNmQ=",
    "lbs-lns-token": "ZjAyNTNhYTJiYjc4OGU5MTg4ZjAxY2JmMTJkNzFiNjk="
  }
}
It's just a simple key/value and so it's handled/stored similar to Attributes.

If we introduce a type;
 We then have a list of supported secret types lbs-lns-secret , lbs-cups-secret  etc?
The Client should know the type, and fetch the values (using a prompt in the case of the CLI) and marshal it and send it to be encrypted?

So the stored secrets would be  something like
[
  "lbs-lns-secret":{"token": "value", "tls-cert":"value", "tls-key": "value"},
  "lbs-cups-secret":{}
  "some-other-secret":{}
]
And the Key of each block is the Encryption key ID. Is that what you mean?

johanstokking:
> It's just a simple key/value and so it's handled/stored similar to Attributes.

The difference is that attributes are user defined and are not driving behavior of The Things Stack, while these secrets are. These secret keys now become magic keys that we need to document everywhere.

Separate the storage of ciphertext (byte array) from what it represents (structs). The latter should be defined as proto message. Yes, the client should be aware of what type it is dealing with – that should become part of the user interface (CLI and Console) – and thus do the marshaling and unmarshaling of plaintext to the concerning message type.

KrishnaIyer:
Closing this but will keep the branch for reference. New Draft to discuss API/IS model will follow.

PR #2867, opened 2020-07-08
Fix sentry error descriptions

### LINKED ISSUE #2849: Improve sentry error fingerprints for frontend errors
#### Summary
We currently see a lot of sentry errors being grouped in an unhelpful way. We need to use better fingerprints to ensure that grouped events actually account for the same type of error.

#### Steps to Reproduce
See sentry issues like these: https://sentry.io/organizations/the-things-industries/issues/1739012974/events/d441eb3ca15d4897be4be5a8bdce8b6b/?project=2682566&referrer=github_integration

Going through the individual error events, it can be seen that all the errors are grouped on the grounds of being logged as: `Non-Error exception captured with keys: code, details, message`
However, a lot of different backend errors can ultimately result in this error message. See also: https://github.com/TheThingsNetwork/lorawan-stack/issues/2824#issuecomment-654559983

#### What do you see now?
Sentry events being grouped in an unhelpful way.

Sentry events being grouped correctly.

#### Environment
v3.8.5

I've found [this article](https://blog.sentry.io/2019/01/17/debug-tough-front-end-errors-sentry-clues) which has some helpful suggestions.

We should also make sure that we wrap the backend error into a proper `Error()` constructor.

Manually.

@pgalic96

pgalic96:
#### Summary

Closes #2849 

An example of an error event: https://sentry.io/organizations/the-things-industries/issues/1772980065/?project=2732897&query=is%3Aunresolved

#### Changes

- Add grouping of error events by error id, if one is present
- Wrap errors with `Error` object and add error id as description, or generic description if error id is not present

#### Testing

- Tested manually

##### Regressions

- Sentry logging

#### Notes for Reviewers

@kschiffer This might look ugly, but I don't see other way of doing it until we are absolutely sure that everything we get from backend is in the same format (meaning that sometimes errors don't have message, details, etc.)

#### Checklist

### DIFF
--- a/pkg/webui/console/store/middleware/logics/lib.js
+++ b/pkg/webui/console/store/middleware/logics/lib.js
@@ -25,8 +25,6 @@ import {
 
 import { clear as clearAccessToken } from '@console/lib/access-token'
 
-import { clear as clearAccessToken } from '@console/lib/access-token'
-
 const getResultActionFromType = function(typeString, status) {
   if (typeString instanceof Array) {
     if (typeString.length === 1) {
--- a/pkg/webui/console/store/middleware/logics/lib.js
+++ b/pkg/webui/console/store/middleware/logics/lib.js
@@ -19,8 +19,8 @@ import { error } from '@ttn-lw/lib/log'
 import {
   isUnauthenticatedError,
   isInvalidArgumentError,
-  isUnknown,
   getBackendErrorId,
+  isUnknown,
 } from '@ttn-lw/lib/errors/utils'
 
 import { clear as clearAccessToken } from '@console/lib/access-token'
@@ -95,20 +95,17 @@ const createRequestLogic = function(
           window.location.reload()
         } else {
           // Otherwise, dispatch the fail action and report it to Sentry.
-          if (isUnknown(e) || isInvalidArgumentError(e)) {
+          if (isInvalidArgumentError(e)) {
             Sentry.withScope(scope => {
               scope.setExtras(e)
-              if (e.message) {
-                scope.setFingerprint(getBackendErrorId(e))
-                Sentry.captureException(new Error(getBackendErrorId(e)))
-              } else if (e.payload && e.payload.message) {
-                scope.setFingerprint(getBackendErrorId(e.payload))
-                Sentry.captureException(new Error(getBackendErrorId(e.payload)))
-              } else if (isInvalidArgumentError(e)) {
-                Sentry.captureException(new Error('Invalid argument error'))
-              } else {
-                Sentry.captureException(new Error('Unknown error'))
-              }
+              const fingerprint = getBackendErrorId(e)
+              scope.setFingerprint(fingerprint)
+              Sentry.captureException(new Error(fingerprint))
+            })
+          } else if (isUnknown(e)) {
+            Sentry.withScope(scope => {
+              scope.setExtras(e)
+              Sentry.captureException(e)
             })
           }
           dispatch(failAction(e))
--- a/pkg/webui/console/store/middleware/logics/lib.js
+++ b/pkg/webui/console/store/middleware/logics/lib.js
@@ -16,7 +16,14 @@ import { createLogic } from 'redux-logic'
 import * as Sentry from '@sentry/browser'
 
 import { error } from '@ttn-lw/lib/log'
-import { isUnauthenticatedError, isInvalidArgumentError, isUnknown } from '@ttn-lw/lib/errors/utils'
+import {
+  isUnauthenticatedError,
+  isInvalidArgumentError,
+  isUnknown,
+  getBackendErrorId,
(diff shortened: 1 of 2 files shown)


pkg/webui/console/store/middleware/logics/lib.js
```diff
@@ -89,7 +94,20 @@ const createRequestLogic = function(
         } else {
           // Otherwise, dispatch the fail action and report it to Sentry.
           if (isUnknown(e) || isInvalidArgumentError(e)) {
```
kschiffer, line 96: What was actually the reason for distinguishing invalid argument errors?

pkg/webui/console/store/middleware/logics/lib.js
```diff
@@ -89,7 +94,20 @@ const createRequestLogic = function(
...
-            Sentry.captureException(failAction(e))
+            Sentry.withScope(scope => {
+              scope.setExtras(e)
+              if (e.message) {
+                scope.setFingerprint(getBackendErrorId(e))
+                Sentry.captureException(new Error(getBackendErrorId(e)))
+              } else if (e.payload && e.payload.message) {
```
kschiffer, line 102: When does that happen that a fail action is thrown? That seems odd to me, the `process()` function should dispatch API actions that should fail with backend error objects or request errors. Only the request logic itself here is wrapping the error into the fail action.

pkg/webui/lib/components/error-view.js
```diff
@@ -39,8 +40,14 @@ class ErrorView extends React.Component {
...
-      scope.setExtras(errorInfo)
-      Sentry.captureException(error)
+      scope.setExtras(error)
+      if (error.message) {
+        const errorId = getBackendErrorId(error)
+        scope.setFingerprint([errorId])
+        Sentry.captureException(new Error(errorId))
+      } else {
+        Sentry.captureException(error)
+      }
```
kschiffer, line 50: Better to use our error utils here to distinguish the error `isBackend` and `isFrontend`.

You might also consider making a new util `getErrorFingerprint(error)` which basically does all of the above and we can maintain it in one place.

pkg/webui/console/store/middleware/logics/lib.js
pgalic96: sentry.io/organizations/the-things-industries/issues/1739001198/?project=2682566&query=is%3Aunresolved  check this

pkg/webui/console/store/middleware/logics/lib.js
pgalic96: the reason is that there is a lot of errors that are part of the regular flow, `404` for example always occurs when creating new gateways which haven't been connected. We don't want this to consume our sentry subscription.

pkg/webui/console/store/middleware/logics/lib.js
pgalic96: https://sentry.io/organizations/the-things-industries/issues/1739001198/events/024287ed737f4e8590651102b6c97653/?project=2682566

Why `isBackend` is false for this error? It has all the attributes that `isBackend` checks, but it still goes into this condition `isUnknown || isInvalidArgumentError`. This and many other errors mean that we cant just ignore `isUnknown`, but the problem is that the request errors and unrecognized backend errors are mixed there, meaning Sentry will have a hard time distinguishing between them.

pgalic96:
So, while looking at the error events I came to some conclusions:

- We don't need (or shouldn't at least) separate utility for getting the error fingerprint. The error fingerprint should be derived for the backend errors, since these errors have a nice `message` property from which we can extract the fingerprint using `getBackendErrorId`. Frontend errors get caught by a built-in error catcher
- In error view it is enough to catch the backend errors, as the frontend errors that crash the console will get caught by default

pkg/webui/console/store/middleware/logics/lib.js
kschiffer: I get that for 404s, but invalid argument is not a 404?

pkg/webui/console/store/middleware/logics/lib.js
kschiffer: So what's the conclusion? Can you check where we wrongfully throw the whole action instead of just the error payload? We need an issue to fix that.

pkg/webui/console/store/middleware/logics/lib.js
pgalic96: yeah I think we decided that `invalidArgumentError` is the one that is not part of any regular flow, so it is good to report it to Sentry. The rest of the errors (404, 401, 403...) is not reported to Sentry.

pkg/webui/console/store/middleware/logics/lib.js
pgalic96: Yes we did it here, I wrapped the exception with `failAction` and reported it like that to Sentry. That is fixed now.

PR #2878, opened 2020-07-09
Fix validation of snake case keys

KrishnaIyer:
#### Summary

Fix proto validation of snake case keys. We're using an incorrect regex (most likely a copy-paste mistake).

#### Changes

- Update regex
- Generate proto

#### Testing

Tested regex separately.

##### Regressions

Update: This is backwards compatible to allow `-` as well

#### Notes for Reviewers

~This technically breaks API but it's a bug so I think it's fine.~

#### Checklist


KrishnaIyer:
> Is there a specific reason for not allowing - as well?

Nah. It's just that it's inconsistent. Updated.

PR #2883, opened 2020-07-11
Fix handling of fragmented event stream chunks in JS SDK

### LINKED ISSUE #2826: Data/event views frequently crashing after some amount of time in the Console
#### Summary

Many sentry events suggest that the event/data views in the console crash due to a `Network Error` after some amount of time. This needs to be investigated and fixed.

#### Steps to Reproduce

1.  Go to data views for gateways, end devices or applications
2.  Wait for an arbitrary amount of time (I believe it depends on the frequency of event data coming in)
3.  Observe the event component crashing and error logs appearing

#### What do you see now?

Event views being fragile and crashing after some time.

Event views being stable and reliable.

#### Environment

Seems to be ongoing for a while, this has been reported ever since we added the sentry integration to the Console.

Need more time to investigate the exact issue. Currently trying to reproduce this reliably. From the sentry logs, it appears that a certain event type/message might be the problem.

Depends on what the exact issue is.

Yes. cc @bafonins let me know if you have any insights on this.

kschiffer:
#### Summary
Closes #2826

This PR fixes an issue where incomplete event stream chunks will cause crashes in the console data views.

#### Changes
- Buffer stream chunks so that parsing can be retried when concatenated with the consequent chunks
- Refactor error logic to prevent that the stream is terminated when it does not have to be

#### Testing

Manual testing on the staging environment

##### Regressions

This could introduce errors in streaming endpoints handled by the SDK

#### Notes for Reviewers
I'm also working on some improvement on how event errors are handled in the console. We currently throw and stop the event stream on any error, when for some errors related to the `onChunk` handler, the event stream can very well be continued after that.

#### Checklist

### DIFF
--- a/sdk/js/src/api/stream/stream-node.js
+++ b/sdk/js/src/api/stream/stream-node.js
@@ -34,9 +34,9 @@ import { notify, EVENTS } from './shared'
  *
  *    // Add listeners to the stream.
  *    stream
- *      .on('start', () => console.log('conn opened'));
- *      .on('chunk', chunk => console.log('received chunk', chunk));
- *      .on('error', error => console.log(error));
+ *      .on('start', () => console.log('conn opened'))
+ *      .on('chunk', chunk => console.log('received chunk', chunk))
+ *      .on('error', error => console.log(error))
  *      .on('close', () => console.log('conn closed'))
  *
  *    // Close the stream after 20 s.
@@ -58,6 +58,7 @@ export default async function(payload, url) {
   }
 
   let reader = null
+  let buffer = ''
   axios({
     url,
     data: JSON.stringify(payload),
@@ -65,6 +66,7 @@ export default async function(payload, url) {
     responseType: 'stream',
     headers: {
       Authorization,
+      Accept: 'text/event-stream',
     },
   })
     .then(response => response.data)
@@ -74,10 +76,11 @@ export default async function(payload, url) {
 
       stream.on('data', function(data) {
         const parsed = data.toString('utf8')
-
-        for (const line of parsed.trim().split('\n')) {
-          const result = JSON.parse(line).result
-          notify(listeners[EVENTS.CHUNK], result)
+        buffer += parsed
+        const lines = buffer.split(/\n\n/)
+        buffer = lines.pop()
+        for (const line of lines) {
+          notify(listeners[EVENTS.CHUNK], JSON.parse(line).result)
         }
       })
       stream.on('end', function() {
--- a/sdk/js/src/api/stream/stream.js
+++ b/sdk/js/src/api/stream/stream.js
@@ -35,9 +35,9 @@ import 'web-streams-polyfill/dist/polyfill'
  *
  *    // Add listeners to the stream.
  *    stream
- *      .on('start', () => console.log('conn opened'));
- *      .on('chunk', chunk => console.log('received chunk', chunk));
- *      .on('error', error => console.log(error));
+ *      .on('start', () => console.log('conn opened'))
+ *      .on('chunk', chunk => console.log('received chunk', chunk))
+ *      .on('error', error => console.log(error))
  *      .on('close', () => console.log('conn closed'))
  *
  *    // Close the stream after 20 s.
@@ -65,6 +65,7 @@ export default async function(payload, url) {
     signal: abortController.signal,
     headers: {
       Authorization,
+      Accept: 'text/event-stream',
     },
   })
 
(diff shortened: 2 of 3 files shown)


sdk/js/src/api/stream/stream-node.js
```diff
@@ -68,6 +69,11 @@ export default async function(payload, url) {
     },
   })
     .then(response => response.data)
+    .catch(function(error) {
```
bafonins, line 72: Can you confirm that the issue occurs when using events in node env?

kschiffer:
> Hmm, this feels like a workaround.
> 
> Indeed it is expected that messages are sent in chunks, not necessarily line by line. Isn’t there a simple buffered reader for this, where you write to, and read line-by-line?

From what I can see the line breaks are sent inconsistently, sometimes an event JSON ends with one, sometimes it doesn't, making it hard to distinguish fragmented messages from unfragmented ones. If all events ended with a linebreak, it would be easier in the SDK to buffer the chunks until the next linebreak occurs.

johanstokking:
Looping in @htdvisser

htdvisser:
When an HTTP client requests a `text/event-stream` response, [this is caught by grpc-gateway](https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/pkg/rpcserver/rpcserver.go#L192) and we instruct it to [use a (double) newline as delimiter](https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/pkg/jsonpb/ttn.go#L38). grpc-gateway then [writes each message it receives from gRPC, writes the delimiter and flushes to the network](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/handler.go#L50-L92).

I really can't see how a browser should be able to receive non-delimited events.

johanstokking:
Isn't the problem that the browser receives an incomplete event? That the next part of the same event comes in the next chunk? Like:

```
{"A":42}\n\n{"B":
42}\n\n
```

This is expected with buffered reads anywhere, but I don't know if the underlying protocol does any magic to reconstruct fragments.

Wouldn't this be the algorithm in pseudo-code?

```
buffer = []
while chunk = network.read() {
	buffer += chunk
	while i := buffer.indexOf("\n\n"); i >= 0 {
		event = json.parse(buffer[:i])
		processEvent(event)
		buffer = buffer[i:]
	}
}
```

kschiffer:
The issue was a missing `Content-Type` header set to `text/event-stream`. After this, the double line ends arrive properly. I've added a simple buffering logic that will buffer any incomplete chunk until the next line end.

Checked both node and browser.

sdk/js/src/api/stream/stream-node.js
kschiffer: The issue was occurring there as well, the reason being the missing `Content-Type` header.

sdk/js/src/api/stream/stream.js
```diff
@@ -90,10 +92,11 @@ export default async function(payload, url) {
       }
 
       const parsed = ArrayBufferToString(value)
-
-      for (const line of parsed.trim().split('\n')) {
-        const result = JSON.parse(line).result
-        notify(listeners[EVENTS.CHUNK], result)
+      buffer += parsed
+      const lines = buffer.split(/\n\n/)
+      buffer = lines.pop()
```
kschiffer, line 97: So any possible incomplete tail is added to the buffer.

sdk/js/src/api/stream/stream-node.js
```diff
@@ -58,13 +58,15 @@ export default async function(payload, url) {
...
   let reader = null
+  let buffer = ''
   axios({
     url,
     data: JSON.stringify(payload),
     method: 'POST',
     responseType: 'stream',
     headers: {
       Authorization,
+      'Content-Type': 'text/event-stream',
```
htdvisser, line 69: Shouldn't this be an `Accept` header, since you want the response to be `text/event-stream` (and not the request)?

htdvisser: Same in the other implementation

kschiffer: That's true actually. While both works (setting the request to `text/event-stream` will trigger the response to be the same), I guess using `Accept` is more fitting here.

htdvisser:
That shouldn't happen, and I also don't see it happen in on `master` or in production...

bafonins:
Indeed, looks like the header is added by `webpack-dev-server`. In production build works as expected.

Issue #2914, opened 2020-07-15
Document `mac-settings` and `mac-state` commands


benolayinka:
#### Summary

There is currently no CLI guide for updating device parameters using API commands from https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/doc/content/reference/api/end_device.md

There is also confusion about the difference between `mac-state` and `mac-settings` endpoints.
https://github.com/TheThingsIndustries/lorawan-stack-support/issues/48

#### Why do we need this ?

To provide a standard way to configure mac settings on devices.

#### What is already there? What do you see now?

https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/doc/content/reference/api/end_device.md

#### What is missing? What do you want to see?

A guide

#### How do you propose to document this?

I'll write the documentation. @rvolosatovs can you clarify the `mac-state` vs `mac-settings` thing and recommend a way for configuring these settings?

#### Can you do this yourself and submit a Pull Request?

Yes

rvolosatovs:
See https://github.com/TheThingsIndustries/lorawan-stack-support/issues/48#issuecomment-651779700

`mac-settings` are only effective on device creation or after MAC state reset (e.g. after OTAA join or ABP FCnt reset, or ResetInd MAC command)
`mac-state` changes are volatile and only apply to the active session - once the MAC state is reset, all modifications are lost

benolayinka:
> See [TheThingsIndustries/lorawan-stack-support#48 (comment)](https://github.com/TheThingsIndustries/lorawan-stack-support/issues/48#issuecomment-651779700)
> 
> `mac-settings` are only effective on device creation or after MAC state reset (e.g. after OTAA join or ABP FCnt reset, or ResetInd MAC command)
> `mac-state` changes are volatile and only apply to the active session - once the MAC state is reset, all modifications are lost

Okay. The [MacState docs](https://thethingsstack.io/v3.8.6/reference/api/end_device/#message:MACState) say it is read-only. This isn't true?

rvolosatovs:
Not anymore indeed

benolayinka:
Okay. One last question @rvolosatovs :)

For the `mac-settings`, what's the difference between setting
`--mac-settings.rx1-delay` and `--mac-settings.desired-rx1-delay`? the MAC settings is only applied on a join anyway, right?

And for `mac-state`, what does setting `mac-state.desired-parameters` do, if the state is lost after a reset? When does desired become current?

[...]

Issue #2269, opened 2020-03-30
`factory_preset_frequencies` behavior


rvolosatovs:
#### Summary

`factory_preset_frequencies` behavior is not obvious. It's only mentioned in the BI spec and it's stated that it's "mandatory for ABP", yet nothing is said about OTAA.

#### Why do we need this?

To get some clarity and to fix ABP behavior in fixed-channel bands

#### What is already there? What do you see now?

NS assumes `factory_preset_frequencies` to be the channels device uses on reset, so if it's set to `A` and NS sends channels `B` in `CFList` of join-accept, NS assumes that device has channels `A ++ B` enabled immediately after join, which may or may not be expected by the user.

#### What is missing? What do you want to see?

a. ~Leave as-is, document~
b. ~Disallow `factory_preset_frequencies` for OTAA~
c. ~Ignore `factory_preset_frequencies` for OTAA~
d. Assume `factory_preset_frequencies` are the join channels used by OTAA device

#### Can you do this yourself and submit a Pull Request?

yes

rvolosatovs:
Note, that we have very similar issues with RX1 data rate offset, RX2 data rate index and RX1 delay. (aka other MAC parameters being sent in join-accept)

rvolosatovs:
Note, that we have very similar issues with RX1 data rate offset, RX2 data rate index and RX1 delay. (aka other MAC parameters being sent in join-accept)

rvolosatovs:
Note, that we have very similar issues with RX1 data rate offset, RX2 data rate index and RX1 delay

htdvisser:
> NS assumes that device has channels `A ++ B` enabled immediately after join

The effect of CFList is defined in PHY. For EU863-870 that is:

> _CFList is a list of five channel frequencies for the channels four to eight_

---

I like option (d) as it would allow devices to perform joins on those channels in addition to (or alternative to) the mandatory join channels.

The RX1 data rate offset and RX2 data rate are in fact defaults, so if those are specified, I think we should respect those (and use those for the join-accept).

The RX1 delay can be completely ignored for OTAA devices, because join-accept messages use JoinAcceptDelay1/2 instead of ReceiveDelay1/2.

rvolosatovs:
> I like option (d) as it would allow devices to perform joins on those channels in addition to (or alternative to) the mandatory join channels.

[...]

PR #2928, opened 2020-07-16
Add MAC settings guide

### LINKED ISSUE #2501: Document v3 MAC settings
#### Summary

Document v3 MAC settings for end devices.  Came up during #2467

#### Why do we need this ?

Document v3 MAC settings and possible use cases for them. 

`ttn-lw-cli end-devices update --help` lists the available MAC settings. There is a short description in the API reference page.

Use cases for important MAC settings. Better documentation for when and how they should be set (e.g. https://github.com/TheThingsNetwork/lorawan-stack/pull/2467#discussion_r422088779)

...

no. assigning @rvolosatovs  and @benolayinka, re-assign as needed.

benolayinka:
#### Summary

Closes #2501 #2914 

#### Changes

- Add CLI guide for updating MAC settings
- Add missing enum to device API guide
- Tiny comment fixup (`mac-state` is no longer read only.)

#### Testing

Run locally

#### Checklist

### DIFF
--- a/api/end_device.proto
+++ b/api/end_device.proto
@@ -285,7 +285,7 @@ message MACSettings {
 
 // MACState represents the state of MAC layer of the device.
 // MACState is reset on each join for OTAA or ResetInd for ABP devices.
-// This is used internally by the Network Server and is read only.
+// This is used internally by the Network Server.
 message MACState {
   option (gogoproto.populate) = false;
 
--- a/doc/content/devices/mac-settings/_index.md
+++ b/doc/content/devices/mac-settings/_index.md
@@ -1,0 +1,112 @@
+---
+title: "MAC Settings"
+description: ""
+---
+
+This section provides guidelines for configuring MAC settings for end devices from the CLI.
+
+<!--more-->
+
+{{< cli-only >}}
+
+### MAC Settings and MAC State
+
+`mac-settings` on {{% tts %}} are configurable per end device. To configure persistent MAC settings, make changes to `mac-settings.desired_<parameter>`. Updates to `mac-settings.desired_<parameter>` take effect on device creation, after OTAA join or ABP FCnt reset, ResetInd, or after MAC state reset.
+
+`mac-state` can be used to test MAC settings in the current session. To update settings for testing in the current session, make changes to the `mac-state.desired_parameters.<parameter>`. Updates to the `mac-state.desired_parameters.<parameter>` are applied on the next uplink, and lost on reset.
+
+The expected procedure for testing and updating settings is:
+
+1. Modify `mac-state.desired_parameters.<parameter>` to see changes in the current session
+2. Test that everything works as expected
+3. Modify `mac-settings.desired_<parameter>` to make the change permanent
+
+If no settings are provided on device creation or unset, defaults are taken from [Network Server Configuration]({{< ref src="/reference/configuration/network-server" >}}).
+
+### Available MAC settings
+
+Run the following command to get a list of all available MAC settings and available parameter values:
+
+```bash
+$ ttn-lw-cli end-devices set --help
+```
+
+You can also refer to the [End Device API Reference page]({{< ref "reference/api/end_device#message:MACSettings" >}}) for documentation on the available MAC settings and MAC state parameters.
+
+### Class Specific Settings
+
+Settings that are useful based on device class are:
+
+All devices:
+
+- `mac-settings.factory-preset-frequencies`
+
+Class A:
+
+- `mac-settings.desired-rx1-delay`
+- `mac-settings.desired-rx1-data-rate-offset`
+- `mac-settings.desired-rx2-data-rate-index`
+- `mac-settings.desired-rx2-frequency`
+- `mac-settings.supports-32-bit-f-cnt`
+- `mac-settings.use-adr`
+
+Class A ABP:
+
+- `mac-settings.resets-f-cnt`
+
+Class B:
+
+- `mac-settings.class-b-timeout`
+- `mac-settings.ping-slot-periodicity`
+- `mac-settings.desired-ping-slot-data-rate-index`
+- `mac-settings.desired-ping-slot-frequency`
+
+Class C:
+
+- `mac-settings.class-c-timeout`
+
+Some additional examples are included below. All settings are available at the [End Device API Reference page]({{< ref "reference/api/end_device#message:MACSettings" >}}) and can be viewed using the `ttn-lw-cli end-devices set --help` command.
+
+### Configure Factory Preset Frequencies for ABP Devices
+
+To tell {{% tts %}} which frequencies are configured in an ABP device, set the `mac-settings.factory-preset-frequencies` parameter. For example, to configure a device using the default EU868 frequencies, use the following command:
+
+```bash
+$ ttn-lw-cli devices update <device_id> --mac-settings.factory-preset-frequencies 868100000,868300000,868500000,867100000,867300000,867500000,867700000,867900000
+```
+
+>Note: For ABP devices, `mac-settings.factory-preset-frequencies` should be specified on `device create` or the settings will only take effect after MAC reset.
+
+### Set Duty Cycle
+
+To change the duty cycle, set the `desired-max-duty-cycle` parameter. For example, to set the duty cycle to 0.098%, use the following command:
+
+```bash
+$ ttn-lw-cli end-devices set <app-id> <device-id> --mac-settings.desired-max-duty-cycle DUTY_CYCLE_1024
+```
+
+>Note: See the [End Device API Reference]({{< ref "reference/api/end_device#message:MACSettings" >}}) for available fields and definitions of constants. DUTY_CYCLE_1024 represents 1/1024 ≈ 0.098%.
+
+### Enable ADR
+
+To enable ADR, set the `mac-settings.use-adr` parameter
+
+```bash
+$ ttn-lw-cli end-devices set <app-id> <device-id> --mac-settings.use-adr true 
+```
+
+### Set RX1 Delay
+
+The RX1 delay of end devices is set to 5 second by default. To change it, set the `desired-rx1-delay` parameter:
+
+```bash
+$ ttn-lw-cli end-devices set <app-id> <device-id> --mac-settings.desired-rx1-delay RX_DELAY_5
+```
+
+### Unset MAC settings
+
+The CLI can also be used to unset MAC settings (so that the default ones are used):
+
+```bash
+$ ttn-lw-cli end-devices set <app-id> <device-id> --unset mac-settings.rx1-delay
+```
--- a/doc/content/devices/mac-settings/_index.md
+++ b/doc/content/devices/mac-settings/_index.md
@@ -11,7 +11,9 @@ This section provides guidelines for configuring MAC settings for end devices fr
 
 ### MAC Settings and MAC State
 
-`mac-settings` on {{% tts %}} are configurable per end device. To configure persistent MAC settings, make changes to `mac-settings.desired_<parameter>`. Updates to `mac-settings.desired_<parameter>` take effect on device creation, after OTAA join or ABP FCnt reset, ResetInd, or after MAC state reset.
+MAC settings on {{% tts %}} are configurable per end device. To configure persistent MAC settings, make changes to `mac-settings.desired_<parameter>`. Updates to `mac-settings.desired_<parameter>` take effect on device creation, after OTAA join or ABP FCnt reset, ResetInd, or after MAC state reset.
+
+`mac-settings.<parameter>` represents what the Network Server believes is configured on the end device, and should not be changed, unless the device does not conform to spec. It may however be necessary to set `mac-settings.RX1_delay` for ABP devices where this is not configured as part of activation.
 
 `mac-state` can be used to test MAC settings in the current session. To update settings for testing in the current session, make changes to the `mac-state.desired_parameters.<parameter>`. Updates to the `mac-state.desired_parameters.<parameter>` are applied on the next uplink, and lost on reset.
 
@@ -21,7 +23,7 @@ The expected procedure for testing and updating settings is:
 2. Test that everything works as expected
 3. Modify `mac-settings.desired_<parameter>` to make the change permanent
 
-If no settings are provided on device creation or unset, defaults are taken from [Network Server Configuration]({{< ref src="/reference/configuration/network-server" >}}).
+If no settings are provided on device creation or unset, defaults are first taken from the device Frequency Plan if available, and finally from [Network Server Configuration]({{< ref src="/reference/configuration/network-server" >}}).
 
 ### Available MAC settings
 
--- a/api/api.md
+++ b/api/api.md
@@ -2310,7 +2310,7 @@ This is used internally by the Network Server.
 
 MACState represents the state of MAC layer of the device.
 MACState is reset on each join for OTAA or ResetInd for ABP devices.
-This is used internally by the Network Server and is read only.
+This is used internally by the Network Server.
 
 | Field | Type | Label | Description |
 | ----- | ---- | ----- | ----------- |
--- a/api/api.swagger.json
+++ b/api/api.swagger.json
@@ -11703,7 +11703,7 @@
           "description": "Time when the last downlink message was scheduled."
         }
       },
-      "description": "MACState represents the state of MAC layer of the device.\nMACState is reset on each join for OTAA or ResetInd for ABP devices.\nThis is used internally by the Network Server and is read only."
+      "description": "MACState represents the state of MAC layer of the device.\nMACState is reset on each join for OTAA or ResetInd for ABP devices.\nThis is used internally by the Network Server."
     },
     "v3MACVersion": {
       "type": "string",
--- a/doc/content/getting-started/migrating-from-v2/configure-mac-settings.md
+++ b/doc/content/getting-started/migrating-from-v2/configure-mac-settings.md
@@ -3,39 +3,6 @@ title: Fine-tuning MAC Settings for End Devices
 weight: 60
 ---
 
-{{< cli-only >}}
+MAC settings on {{% tts %}} are configurable per end device - see the [MAC settings guide]({{< ref "reference/api/end_device#message:MACSettings" >}}) for instructions.
 
-This section is relevant for end devices that need MAC settings fine-tuning to function properly with the Network Server of {{% tts %}}.
-
-MAC settings on {{% tts %}} are configurable per end device. They can be configured from the CLI.
-
-### Setting RX1 Delay
-
-The RX1 delay of end devices is set to 1 second by default. For some end devices, this may lead to downlink messages not being scheduled in time. Therefore, it is recommended that the RX1 delay be increased to 5 seconds:
-
-```bash
-$ ttn-lw-cli end-devices set "app-id" "device-id" --mac-settings.rx1-delay RX_DELAY_5
-```
-
-### Unsetting MAC settings
-
-The CLI can also be used to unset MAC settings (so that the default ones are used):
-
-```bash
-$ ttn-lw-cli end-devices set "app-id" "device-id" --unset mac-settings.rx1-delay
-```
-
-### Available MAC settings
-
-Run the following command to get a list of all available MAC settings:
-
-```bash
-$ ttn-lw-cli end-devices set --help
-```
-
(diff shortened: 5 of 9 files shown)


doc/content/devices/mac-settings/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Setting Duty Cycle
+
+To change the duty cycle, set the `max-duty-cycle` parameter. For example, to set the duty cycle to 0.098%, use the following command:
+
+```bash
+$ ttn-lw-cli end-devices update "app-id" "device-id" --mac-settings.max-duty-cycle DUTY_CYCLE_1024
+```
+
+>Note: See the [End Device API Reference]({{< ref "reference/api/end_device#message:MACSettings" >}}) for available fields and definitions of constants. DUTY_CYCLE_1024 represents 1/1024 ≈ 0.98%.
```
neoaggelos: Is it 0.98 or 0.098?

doc/content/devices/mac-settings/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Setting Uplink and Downlink Dwell Time
+
+To enable uplink and downlink dwell time, set the `uplink-dwell-time` or `downlink-dwell-time` parameters of `mac-state`. For example, to enable downlink dwell time in the current session, use the command:
+
+```bash
+$ ttn-lw-cli end-devices update "app-id" "device-id" --mac-state.current-parameters.downlink_dwell_time true 
+```
```
htdvisser: Maybe good to mention that you typically shouldn't be setting these except for testing the implementation of your (development) devices.

doc/content/devices/mac-settings/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Setting RX1 Delay
+
+The RX1 delay of end devices is set to 1 second by default. For some end devices, this may lead to downlink messages not being scheduled in time. Therefore, it is recommended that the RX1 delay be increased to 5 seconds:
```
htdvisser: I'm pretty sure the Network Server default is actually 5 seconds.

doc/content/devices/mac-settings/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+<!--more-->
+
+{{< cli-only >}}
+
+MAC settings on {{% tts %}} are configurable per end device.
+
+Updates to the `mac-state` have an effect during the current session, and are lost on reset.
+
+Updates to `mac-settings` take effect after device creation or after MAC state reset (e.g. after OTAA join or ABP FCnt reset, or ResetInd MAC command)
```
rvolosatovs:
```suggestion
Updates to `mac-settings` take effect on device creation or after MAC state reset (e.g. after OTAA join or ABP FCnt reset, or `ResetInd` MAC command)
```

doc/content/devices/mac-settings/_index.md
rvolosatovs: Normally users are expected to modify `mac_settings` and only modify `mac_state` for testing/development purposes

doc/content/devices/mac-settings/_index.md
rvolosatovs: Yes, 5 seconds is the default. 1 second is the default in v2.

rvolosatovs:
So, the expected procedure for tweaking device settings is:
1. Modify `mac_state.desired_parameters.X`
2. Ensure the NS configured `X`, test that everything works as expected
3. Modify `mac_settings.desired_X` to make the change permanent

Modification of `mac_state.current_parameters` should basically only happen when the device is misconfigured by the user.

Things that are definitely interesting for users to tweak are (some of them are actually required for some devices to work):

All devices:
- `mac_settings.factory_preset_frequencies` (refs #2269)

Class A:
- `mac_settings.desired_rx1_delay`
- `mac_settings.desired_rx1_data_rate_offset`
- `mac_settings.desired_rx2_data_rate_index`
- `mac_settings.desired_rx2_frequency`
- `mac_settings.supports_32_bit_f_cnt`
- `mac_settings.use_adr`

Class A ABP:
- `mac_settings.resets_f_cnt`

Class B:
- `mac_settings.class_b_timeout`
- `mac_settings.ping_slot_periodicity`
- `mac_settings.desired_ping_slot_data_rate_index`
- `mac_settings.desired_ping_slot_frequency`

Class C:
- `mac_settings.class_c_timeout`

benolayinka:
> So, the expected procedure for tweaking device settings is:
> 
> 1. Modify `mac_state.desired_parameters.X`
> 2. Ensure the NS configured `X`, test that everything works as expected
> 3. Modify `mac_settings.desired_X` to make the change permanent
> 
> Modification of `mac_state.current_parameters` should basically only happen when the device is misconfigured by the user.
> 
> Things that are definitely interesting for users to tweak are (some of them are actually required for some devices to work):
> 
> All devices:
> 
> * `mac_settings.factory_preset_frequencies` (refs #2269)
> 
> Class A:
> 
> * `mac_settings.desired_rx1_delay`
> * `mac_settings.desired_rx1_data_rate_offset`
> * `mac_settings.desired_rx2_data_rate_index`
> * `mac_settings.desired_rx2_frequency`
> * `mac_settings.supports_32_bit_f_cnt`
> * `mac_settings.use_adr`
> 
> Class A ABP:
> 
> * `mac_settings.resets_f_cnt`
> 
> Class B:
> 
> * `mac_settings.class_b_timeout`
> * `mac_settings.ping_slot_periodicity`
> * `mac_settings.desired_ping_slot_data_rate_index`
> * `mac_settings.desired_ping_slot_frequency`
> 
> Class C:
> 
> * `mac_settings.class_c_timeout`

Okay great, will update. What's the difference though between changing `--mac-settings.rx1-delay` and `--mac-settings.desired-rx1-delay`? The MAC setting are only applied on a join anyway, right?

rvolosatovs:
> The MAC setting are only applied on a join anyway, right?

There's no join for ABP devices

benolayinka:
> > The MAC setting are only applied on a join anyway, right?
> 
> There's no join for ABP devices

Okay but still have the same question - when/how is `--mac-settings.rx1-delay` applied, vs. `--mac-settings.desired-rx1-delay` ? You stated that `--mac-settings.desired-rx1-delay` goes in to effect on MAC state reset, but when does `--mac-settings.rx1-delay` take effect?

benolayinka:
Also @rvolosatovs is there a way to set dwell time persistently? The only `dwell-time` parameters I see are for `mac-state` which is lost on reset, correct? https://thethingsstack.io/v3.8.7/reference/api/end_device/#message:MACParameters

htdvisser:
Dwell time restrictions are regulatory limitations, and come from the frequency plan. Users shouldn't customise those.

rvolosatovs:
> > > The MAC setting are only applied on a join anyway, right?
> > 
> > 
> > There's no join for ABP devices
> 
> Okay but still have the same question - when/how is `--mac-settings.rx1-delay` applied, vs. `--mac-settings.desired-rx1-delay` ? You stated that `--mac-settings.desired-rx1-delay` goes in to effect on MAC state reset, but when does `--mac-settings.rx1-delay` take effect?

On MAC state reset. For ABP that happens on device creation, FCnt reset or as a consequence of `ResetInd` MAC command

benolayinka:
> > > > The MAC setting are only applied on a join anyway, right?
> > > 
> > > 
> > > There's no join for ABP devices
> > 
> > 
> > Okay but still have the same question - when/how is `--mac-settings.rx1-delay` applied, vs. `--mac-settings.desired-rx1-delay` ? You stated that `--mac-settings.desired-rx1-delay` goes in to effect on MAC state reset, but when does `--mac-settings.rx1-delay` take effect?
> 
> On MAC state reset. For ABP that happens on device creation, FCnt reset or as a consequence of `ResetInd` MAC command

How can both `--mac-settings.rx1-delay` and `--mac-settings.desired-rx1-delay` go in to effect at the same time on MAC state reset?

For example, if I set `--mac-settings.rx1-delay=RX_DELAY_5` and `--mac-settings.desired-rx1-delay=RX_DELAY_1`, what will the device use on the next MAC state reset? And why?

benolayinka:
@htdvisser there's no description of the DATA_RATE_INDEX constants or the DATA_RATE_OFFSET in api.md or the reference. Could you explain them and I'll commit the descriptions?

benolayinka:
> The data rate index is the index of the data rate as specified in Regional Parameters. The mapping from index to SF and BW is different in each region. It also depends on the region which data rates are specified in the first place.
> 
> Screenshot 2020-07-20 at 15 42 27

So the DATA_RATE_INDEX defines the uplink data rate, and the DATA_RATE_OFFSET gives the downlink data rate based on the uplink rate?

Screenshot 2020-07-20 at 15 54 51

htdvisser:
`DATA_RATE_INDEX` is used for both uplink and downlink data rates.

The downlink data rate index indeed depends on the uplink data rate, the offset, and the region (the US915 table is different).

rvolosatovs:
> For example, if I set `--mac-settings.rx1-delay=RX_DELAY_5` and `--mac-settings.desired-rx1-delay=RX_DELAY_1`, what will the device use on the next MAC state reset? And why?

For ABP device `mac-state.current-parameters.rx1-delay` would be set to 5 and `mac-state.desired-parameters.rx1-delay` would be set to 1.

I think the confusion stems from the fact that Network Server does not actually know what the device uses. Network Server *assumes* things about the device:
1. It assumes that the device is fully-compliant with the specification
2. Due to (1.) Network Server *assumes* that device correctly implements MAC commands and Network Server can configure various parameters on the device and, hence, assumes that the device will use those parameters within the session if configuration succeeds.
3. Due to (1.) Network Server *assumes* that device correctly implements join-accept handling and Network Server can configure various settings that can be configured in the join-accept and, hence, assumes that the device will use those parameters within the session if join succeeds.

So let's take, for example, RX1 delay
For OTAA, NS will send RX1 delay in the join-accept, and, if the device confirms the session (by sending an uplink within the session), it assumes that device is using RX1 delay equal to the value configured in the join-accept. For example, if value of `3` was sent in the join-accept, `mac-state.current-parameters.rx1-delay` will be set to `3`.
For ABP there is no join procedure, and, hence, the only way for Network Server know what the current RX1 delay value that the device's using is, is to:
a. ~configure it via MAC command~ (impossible to do if NS doesn't know "initial" RX1 delay - it's a circular dependency, NS needs to know current RX1 delay to configure a different RX1 delay this way.)
b. Rely on user configuration, i.e. `mac-settings.rx1-delay`

As you see `b.` is the only option here and so `mac-settings.rx1-delay` is mandatory for ABP and useless for OTAA (https://github.com/TheThingsNetwork/lorawan-stack/issues/2269#issuecomment-606504719)

benolayinka:
> > For example, if I set `--mac-settings.rx1-delay=RX_DELAY_5` and `--mac-settings.desired-rx1-delay=RX_DELAY_1`, what will the device use on the next MAC state reset? And why?
> 
> For ABP device `mac-state.current-parameters.rx1-delay` would be set to 5 and `mac-state.desired-parameters.rx1-delay` would be set to 1.
> 
> I think the confusion stems from the fact that Network Server does not actually know what the device uses. Network Server _assumes_ things about the device:
> 
> 1. It _assumes_ that the device is fully-compliant with the specification
> 2. Due to (1.) Network Server _assumes_ that device correctly implements MAC commands and Network Server can configure various parameters on the device and, hence, assumes that the device will use those parameters within the session if configuration succeeds.
> 3. Due to (1.) Network Server _assumes_ that device correctly implements join-accept handling and Network Server can configure various settings that can be configured in the join-accept and, hence, assumes that the device will use those parameters within the session if join succeeds.
> 
> So let's take, for example, RX1 delay
> For OTAA, NS will send RX1 delay in the join-accept, and, if the device confirms the session (by sending an uplink within the session), it assumes that device is using RX1 delay equal to the value configured in the join-accept. For example, if value of `3` was sent in the join-accept, `mac-state.current-parameters.rx1-delay` will be set to `3`.
> For ABP there is no join procedure, and, hence, the only way for Network Server know what the current RX1 delay value that the device's using is, is to:
> a. ~configure it via MAC command~ (impossible to do if NS doesn't know "initial" RX1 delay - it's a circular dependency, NS needs to know current RX1 delay to configure a different RX1 delay this way.)
> b. Rely on user configuration, i.e. `mac-settings.rx1-delay`
> 
> As you see `b.` is the only option here and so `mac-settings.rx1-delay` is mandatory for ABP and useless for OTAA ([#2269 (comment)](https://github.com/TheThingsNetwork/lorawan-stack/issues/2269#issuecomment-606504719))
> 
> Some key points here:
> 
> * `mac-state` is not the "device state", it's what Network Server knows about the "device state"
>   
>   * `mac-state.current-parameters` in particular is what Network Server knows is currently modifiable part "device state"
>   * `mac-state.desired-parameters` in particular is what Network Server wants modifiable part of the "device state" to be
> * everything with `desired` prefix is what Network Server will try to configure after/during device activation
> * `mac-settings` (among other things) allows to override some default values Network Server assumes about the device state, if, for example, the device is not fully-compliant with the LoRaWAN specification

awesome, thank you. does this mean that `mac-state` changes are not actually propagated to the device, but only affect NS downlinks?

doc/content/devices/mac-settings/_index.md
```diff
@@ -0,0 +1,97 @@
...
+
+<!--more-->
+
+{{< cli-only >}}
+
+### MAC Settings and MAC State
+
+`mac-settings` on {{% tts %}} are configurable per end device. To configure persistent MAC settings, make changes to `mac-settings.desired_<parameter>`. Updates to `mac-settings.desired_<parameter>` take effect on device creation, after OTAA join or ABP FCnt reset, ResetInd, or after MAC state reset.
+
+`mac-state` can be used to test MAC settings immediately. To update settings during the current session, make changes to the `mac-state.desired_parameters.<parameter>`. Updates to the `mac-state.desired_parameters.<parameter>` are lost on reset.
```
htdvisser: Maybe write that changes will typically be sent to the device in response to the next uplink, so not technically immediately.

doc/content/devices/mac-settings/_index.md
```diff
@@ -0,0 +1,97 @@
...
+$ ttn-lw-cli end-devices update --help
+```
+
+You can also refer to the [End Device API Reference page]({{< ref "reference/api/end_device#message:MACSettings" >}}) for documentation on the available MAC settings and MAC state parameters.
+
+### Class Specific Settings
+
+Settings that are useful based on device class are:
+
+All devices:
```
htdvisser, line 42: Please split blocks with empty lines (this is a `p`, the next line is an `ul`). See also the documentation guidelines.

Same below.

doc/content/devices/mac-settings/_index.md
```diff
@@ -0,0 +1,97 @@
...
+- `mac-settings.factory-preset-frequencies`
```
htdvisser, line 44: Should we write a short description of what the setting does? Also below?

Ideally imported/copied from the (generated) proto descriptions.

Would be nice if @rvolosatovs could improve documentation for these in the `.proto` files, because now this one just says "List of factory-preset frequencies.". From what I understand this gives the frequencies on which the Network Server accepts Uplink messages from ABP devices or Join Requests from OTAA devices.

benolayinka: @rvolosatovs can you give me a quick example configuration I can add, and either expand the `.proto` descriptions or paste them here so I can?

rvolosatovs: Let's make it part of https://github.com/TheThingsNetwork/lorawan-stack/issues/2269

Other than that, a quick example could be an ABP device using TTN FP frequencies as `factory-preset-frequencies`

benolayinka: is that something like `ttn-lw-cli devices update <device_id> --mac-settings.factory-preset-frequencies 868100000 868300000 868500000 867100000 867300000 867500000 867700000 867900000 868800000`

rvolosatovs: I think you have 1 too much (868800000) - it should be 8 frequencies
Also, I don't think the frequencies should be space-delimited, but I may be wrong, @htdvisser ?
Also note that for ABP this should be specified on `create`, otherwise these will only be effective after reset

htdvisser: For ABP devices that use TTN's Arduino SDK it's indeed the first 8 from that list (so not 868.8).

From the [documentation of pflag](https://pkg.go.dev/github.com/spf13/pflag?tab=doc#StringSlice) I suspect it parses a comma-separated list, which also makes sense, because CLI arguments split on space.

doc/content/devices/mac-settings/_index.md
```diff
@@ -0,0 +1,112 @@
...
+```bash
+$ ttn-lw-cli end-devices update <app-id> <device-id> --mac-settings.use-adr true 
+```
+
+### Set RX1 Delay
+
+The RX1 delay of end devices is set to 5 second by default. To change it, set the `desired-rx1-delay` parameter:
+
+```bash
+$ ttn-lw-cli end-devices update <app-id> <device-id> --mac-settings.desired-rx1-delay RX_DELAY_5
```
neoaggelos: Use the canonical command (`ttn-lw-cli end-devices set`) instead

doc/content/devices/mac-settings/_index.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+### Unset MAC settings
+
+The CLI can also be used to unset MAC settings (so that the default ones are used):
+
+```bash
+$ ttn-lw-cli end-devices update <app-id> <device-id> --unset mac-settings.rx1-delay
```
neoaggelos: also here

benolayinka:
@rvolosatovs @johanstokking need your review as codeowners (roman I think this one is more you)

api/end_device.proto
```diff
@@ -285,7 +285,7 @@ message MACSettings {
 
 // MACState represents the state of MAC layer of the device.
 // MACState is reset on each join for OTAA or ResetInd for ABP devices.
-// This is used internally by the Network Server and is read only.
+// This is used internally by the Network Server.
```
rvolosatovs, line 288: Should we specify that it is partially read only?
This field is really low-level API and is generally meant to only be modified by Network Server itself.

doc/content/devices/mac-settings/_index.md
```diff
@@ -0,0 +1,112 @@
...
+
+This section provides guidelines for configuring MAC settings for end devices from the CLI.
+
+<!--more-->
+
+{{< cli-only >}}
+
+### MAC Settings and MAC State
+
+`mac-settings` on {{% tts %}} are configurable per end device. To configure persistent MAC settings, make changes to `mac-settings.desired_<parameter>`. Updates to `mac-settings.desired_<parameter>` take effect on device creation, after OTAA join or ABP FCnt reset, ResetInd, or after MAC state reset.
```
rvolosatovs:
```suggestion
Various MAC settings on {{% tts %}} are configurable per end device. To configure persistent MAC settings, make changes to `mac-settings.desired_<parameter>`. Updates to `mac-settings.desired_<parameter>` take effect on device creation, after OTAA join or ABP FCnt reset, ResetInd, or after MAC state reset.
```

rvolosatovs: This should mention `mac-settings.current_<parameter>` as well - those are generally necessary when device is not fully compliant with the spec, but also in other cases

doc/content/devices/mac-settings/_index.md
```diff
@@ -0,0 +1,112 @@
...
+
+`mac-state` can be used to test MAC settings in the current session. To update settings for testing in the current session, make changes to the `mac-state.desired_parameters.<parameter>`. Updates to the `mac-state.desired_parameters.<parameter>` are applied on the next uplink, and lost on reset.
+
+The expected procedure for testing and updating settings is:
+
+1. Modify `mac-state.desired_parameters.<parameter>` to see changes in the current session
+2. Test that everything works as expected
+3. Modify `mac-settings.desired_<parameter>` to make the change permanent
+
+If no settings are provided on device creation or unset, defaults are taken from [Network Server Configuration]({{< ref src="/reference/configuration/network-server" >}}).
```
rvolosatovs: Or frequency plan of the device.

By order of preference:
User overrides (MAC settings) -> Frequency Plan (and regional spec) -> Network Server configuration

Issue #2605, opened 2020-05-27
Automated frontend testing


kschiffer:
#### Summary

Recent regression and malfunction incidents during our latest releases have shown us that we need a proper testing strategy for our frontends. Such testing strategy should provide a minimum confidence in the integrity of our releases, as far as the frontend is concerned. In this issue I want to try discuss the state of my research here.

#### Why do we need this?

To ensure the quality and integrity of our releases as well as improving our confidence when releasing.

#### What is already there? What do you see now?

The current testing setup is quite loose and consists of:

*   First and foremost verifying functionality manually and monkey testing
*   A few unit tests for some of our react components
*   Unit tests for some modules/functions, usually ones that have a higher complexity
*   Setting up storybook stories for our global components

#### What is missing? What do you want to see?

With all of the above having helped us attaining a certain degree of quality initially, such tests are not enough to ensure the quality sufficiently anymore. The Things Stack for LoRaWAN consists of many different components, APIs, frontend and backend logic, third party modules etc. While all of these are usually well tested in isolation, we've seen a lot of regressions in the past originating from different components not integrating properly with each other.

As a result, we need to introduce a set of end-to-end tests that will help us spotting these problems during development and as part of the CI process.

##### Testing frameworks and tools

Frontend (e2e) testing is not trivial. To be able to test under realistic conditions, it is necessary to automate actual browser interactions, inspecting outputs visually and structurally, injecting security contexts, stubbing certain API responses, all while running and interacting with a testing environment of the stack.

There are currently three serious frontend end-to-end testing frameworks/tools:

1.  [Selenium](https://www.selenium.dev/)
2.  [Cypress](https://www.cypress.io/)
3.  [Puppeteer](https://pptr.dev/)

Here's what I found out about these:

##### Selenium

*   Not really a frontend testing framework, rather a general purpose browser automation tool
*   Apparently quite slow
*   Since not a frontend testing framework per se, there is some plumbing necessary to use it as such
*   Appears to be free

##### Cypress

*   Dedicated frontend testing framework
*   Pretty straightforward API
*   Excellent documentation (!)
*   Supports Chrome, Firefox and Microsoft Edge
*   Worked out of the box for me using `yarn add`
*   Can record videos, capture screenshots that will help making sense of issues when they occur in CI ([see example from my test](https://i.imgur.com/Yrl7o2j.gif))
*   Has a nice interactive mode where you can watch the tests
[...]

PR #2939, opened 2020-07-19
Setup e2e tests for webui

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2605

#### Changes

- Extend documentation with the e2e testing methodology.
- Test coverage for e2e tests.
- Add base cypress tests with base utilities.

#### Testing

See the docs

#### Notes for Reviewers

~There are no big changes to the magefile compared to https://github.com/TheThingsNetwork/lorawan-stack/pull/2764. I only fixed some minor formatting issues. Will implement the rest after discussing the tests.~

@kschiffer 
~At this stage of this PR please review the proposed testing methodology and the tests. Ill follow up with changes to mage later. Let me know if there is anything you dont agree with or want to be clarified.~

<details>
<summary>Details</summary>

1. **Changing stack configuration.** As mentioned in the docs, UI for some views can vary depending on the stack configuration. Currently, there is no straightforward way to change the configuration between the tests except for restarting the lorawan-stack with different flags. This takes quite some time and drastically complicates the setup for e2e tests. So, as of now, we use the default configuration (with all components enabled), but edit only the injected `<script />` tag when necessary. IMO, this should suffice as it is the only source of truth for the frontend to determine the stack configuration. See tests for creating applications. Ideally, we should test what we actually deploy, but currently this does not seem feasible.
2. **Asserting on backend related messages.** Overall, I dont think that doing this is a good idea. I dont like the idea of making assertions of on specific text captions that the frontend has no control of. Such approach might force us to adjust the tests every time when, for exampple, validation order on the backend changes. However, in some specific cases this can be helpful.
3. **Asserting on API request body, redux store.** I dont really have any strong opinions on this and I can imagine that for some cases making assertions on api responses/requests can be helpful.
4. **Title assertions.** Sometimes these fail in interactive mode. Does this happen to you as well @kschiffer ?

</details>

#### Checklist


DEVELOPMENT.md
```diff
@@ -716,10 +716,161 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+|-- plugins                     Cypress plugins
+|-- support                     Cypress commands and test utilities
+|-- screenshots                 screenshots generated when running tests (4)
+`-- videos                      videos generated when running tests (5)
+```
+
+1. `pkg/cypress/integrations` contains all test specifications.
+  - Each test file must be placed into corresponding folder (`console`/`oauth`).
+  - Each test must follow the following naming: `{context}.spec.js`.
+  - One test file must have end-to-end dedicated only to a specific entity or view.
```
kschiffer:
```suggestion
  - One test file must have end-to-end tests dedicated only to a specific entity or view.
```

DEVELOPMENT.md
```diff
@@ -716,10 +716,161 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+- Extract components that appear on various views and test them separately instead of making assertions in each test where this component is used. For example, such components could be the page header and side navigation.
+- Dedicate at least one test to assert that the view displays its UI elements in place on initial load. Assert on UI changes in tests that trigger these changes.
+- Prefer duplicating entities with non-conflicting ids in tests instead of executing database teardown before each test. For example, when testing various scenaros for registering gateway, consider creating gateways with different ID's and EUI's instead of using a single gateway and drop the database before each test. Note: try to use this approach when possible, otherwise do not hesitate to restore database state before each test.
```
kschiffer:
```suggestion
- Prefer duplicating entities with non-conflicting ids in tests instead of executing database teardown before each test. For example, when testing various scenarios for registering gateway, consider creating gateways with different ID's and EUI's instead of using a single gateway and drop the database before each test. Note: try to use this approach when possible, otherwise do not hesitate to restore database state before each test.
```

DEVELOPMENT.md
```diff
@@ -716,10 +716,161 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+- Consider various stack configurations when writing end-to-end tests. Some views have different UI depending on availability of diffenent stack components. For example, the end device wizard looks different for deployments with complete cluster (NS+JS+AS) and for JS-only configuration.
```
kschiffer:
```suggestion
- Consider various stack configurations when writing end-to-end tests. Some views have different UI depending on availability of diffenent stack components. For example, the end device wizard looks different for deployments with complete cluster (NS+JS+AS) and for JS-only configuration. Likewise, sections and entire views can be enabled or disabled based on our feature toggles. If your test scenario differs based on different feature toggle conditions, make sure to probe these preconditions in your tests.
```

DEVELOPMENT.md
```diff
@@ -716,10 +716,161 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+
+We distinguish between regular end-to-end tests and smoke tests. With smoke tests we try to accumulate application state (redux store, local storage, application cache, etc.) and verify that the application works as expected. When writing smoke tests we comply with he following guidelines:
```
kschiffer:
```suggestion
##### Smoke tests

We distinguish between regular end-to-end tests and smoke tests. With smoke tests we try to accumulate application state (redux store, local storage, application cache, etc.) and verify that the application works as expected. When writing smoke tests we comply with the following guidelines:
```

cypress/integration/oauth/change-password.spec.js
```diff
@@ -0,0 +1,115 @@
...
+      password_confirm: 'ABCDefg123!',
+    }
+    cy.createUser(user)
+    cy.loginOAuth({ user_id: user.ids.user_id, password: user.password })
+    cy.visit(`${Cypress.config('oauthRootPath')}/update-password`)
+
+    cy.findByLabelText('Old password').type(user.password)
+    cy.findByLabelText('New password').type(newPassword)
+    cy.findByLabelText('Confirm password').type(`${newPassword}{enter}`)
+
```
kschiffer, line 77:
```suggestion
cy.findByLabelText('Revoke access').check()
```

To keep this working regardless of the default state.

cypress/integration/oauth/login.spec.js
```diff
@@ -0,0 +1,88 @@
...
+  it('should validate before submitting an empty form', () => {
+    cy.findByRole('button', { name: 'Login' }).click()
+
+    cy.findErrorByLabelText('User ID').as('userIdError')
+    cy.get('@userIdError')
+      .should('contain', 'User ID')
+      .and('be.visible')
+    cy.get('@userIdError')
+      .should('contain', 'is required')
+      .and('be.visible')
```
kschiffer:
```suggestion
    cy.get('@userIdError')
      .should('contain.text', 'User ID is required')
      .and('be.visible')
```

Same for all similar occurrences.

DEVELOPMENT.md
```diff
@@ -716,10 +716,161 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+
+- Smoke tests are testing complete user behavior to perform some complex and critical flow. For example, when testing registration of The Things Uno:
```
kschiffer:
```suggestion
- Smoke tests are testing complete user stories in a **wide and shallow** manner, meaning:
  - performing some complex and critical flow that touches multiple components, APIs and/or views
  - not testing different configurations or preconditions of the same flow in depth
- For example, when testing registration of The Things Uno:
 
```

DEVELOPMENT.md
```diff
@@ -716,10 +716,161 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+  1. Add test file `cypress/integration/smoke/devices/create.js`
+  2. Describe the whole user story to register the device including logging in, creating an application (or using an existing one), link the application and create the end device.
+- One smoke test should be encapsulated into a single `it` declaration.
```
kschiffer: This is a bit confusing now since you use a helper function `describeSmokeTest()`

cypress/integration/smoke/utils.js
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/* eslint-disable import/prefer-default-export */
+
+const defineSmokeTest = (description, test) => ({
+  description,
+  test,
+})
```
kschiffer, line 20: I understand the idea but I'd rather be able to use the native `describe() / it()` syntax for the smoke tests. E.g. we might want to group related smoke tests using `describe()`.

cypress/support/commands.js
```diff
@@ -0,0 +1,170 @@
...
+      body: { user_id: credentials.user_id, password: credentials.password },
+      headers: {
+        'X-CSRF-Token': headers['x-csrf-token'],
+      },
+    })
+  })
+})
+
+// Helper function to quickly login to the console programmatically.
+Cypress.Commands.add('loginConsole', credentials => {
```
kschiffer, line 39:
```suggestion
Cypress.Commands.add('loginConsole', credentials => {
```

Let's use a default argument for the credentials. For most test cases it should not be necessary to use distinct credentials.

pkg/webui/template.go
```diff
@@ -89,17 +89,26 @@ const appHTML = `
...
-      window.ASSETS_ROOT={{$assetsBaseURL}};
-      window.BRANDING_ROOT={{$brandingBaseURL}};
-      window.APP_CONFIG={{.AppConfig}};
-      window.SITE_NAME={{.SiteName}};
-      window.SITE_TITLE={{.Title}};
-			window.SITE_SUB_TITLE={{.SubTitle}};
-			window.SENTRY_DSN={{.SentryDSN}};
-			{{with .CSRFToken}}window.CSRF_TOKEN={{.}};{{end}}
-      {{with .PageData}}window.PAGE_DATA={{.}};{{end}}
+		<script>
```
kschiffer, line 92:
```suggestion
	<script>
```

pkg/webui/template.go
```diff
@@ -89,17 +89,26 @@ const appHTML = `
...
+					APP_ROOT:{{.MountPath}},
+					ASSETS_ROOT:{{$assetsBaseURL}},
+					BRANDING_ROOT:{{$brandingBaseURL}},
+					APP_CONFIG:{{.AppConfig}},
+					SITE_NAME:{{.SiteName}},
+					SITE_TITLE:{{.Title}},
+					SITE_SUB_TITLE:{{.SubTitle}},
+					SENTRY_DSN:{{.SentryDSN}},
+					{{with .CSRFToken}}CSRF_TOKEN:{{.}}{{end}},
+					{{with .PageData}}PAGE_DATA:{{.}}{{end}}
```
kschiffer: Notice the ending comma in the conditional, I think this will result in errors in some situations.

pkg/webui/template.go
```diff
@@ -89,17 +89,26 @@ const appHTML = `
...
+				};
+		  win.__ttn_config__ = config;
+
+			if (win.Cypress && win.__initStackConfig) {
```
kschiffer, line 107: The tabs seem to be off here.

DEVELOPMENT.md
```diff
@@ -716,10 +716,161 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+  1. Add test file `cypress/integration/console/applications/api-keys/create.spec.js`.
+  2. Test the behavior of the API key create view independently from any other specification.
+- Do not repeat actions via the UI that are not related to the current test context. Consider adding reusable [Cypress commands](https://docs.cypress.io/api/cypress-api/custom-commands.html) that do necessary test setup programmatically. This means that when testing any UI that is not the login specification and requies the user to be logged in, there is no need to log in through the login page, while we simply fetch the access token. Note: this does not mean that one cannot create a cypress command that performs actions via UI.
+- Extract components that appear on various views and test them separately instead of making assertions in each test where this component is used. For example, such components could be the page header and side navigation.
+- Dedicate at least one test to assert that the view displays its UI elements in place on initial load. Assert on UI changes in tests that trigger these changes.
```
kschiffer, line 873: See my remark on that one.

cypress/integration/console/applications/create.spec.js
```diff
@@ -0,0 +1,138 @@
...
+      )
+      .and('be.visible')
+    cy.findByLabelText('Linking')
+      .should('be.visible')
+      .should('have.attr', 'value', 'true')
+    cy.findByLabelText('Network Server address').should('be.visible')
+    cy.findDescriptionByLabelText('Network Server address')
+      .should('contain', 'Leave empty to link to the Network Server in the same cluster')
+      .and('be.visible')
+    cy.findByRole('button', { name: 'Create application' }).should('be.visible')
```
kschiffer, line 64: Nit: Maybe this selector config is worth an alias command `cy.findButtonByRole('Create application')`?

cypress/integration/console/shared/header.spec.js
```diff
@@ -0,0 +1,73 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+describe('Header', () => {
+  describe('Console logout', () => {
+    const logout = userName => {
+      cy.findByTestId('profile-dropdown')
```
kschiffer: More realistic selection would be something resembling

> find a dropdown containing your name in the header section

bafonins: Indeed. I think it also makes sense use `cy.within` when testing shared components. Added this to the docs.

cypress/integration/smoke/utils.js
bafonins: I added `defineSmokeTest` to integrate other features to smoke tests later on. Please read https://cypress.io/blog/2019/12/06/use-meaningful-smoke-tests/

cypress/integration/console/applications/create.spec.js
bafonins: we might consider adding this. personally, i dont really see how we can benefit from this atm

pkg/webui/template.go
bafonins: Not sure what you mean here. Can you provide code suggestion?

bafonins:
> While it is easy to pull off for simple views like Login, it becomes pretty tedious to check for all elements in views like entity overviews

There are 2 options on how solve this:
1. Tests various UI components separately. Similar to headers and side-navigation component as mentioned in the docs.
2. Do not pollute a single `it` with assertions on all UI components. There is nothing wrong in creating a bunch of `it`'s for several parts of a complex view.

> It also locks us in too much into a specific structure of the UI and its implementation  

I dont know how one can test UI without taking into account the structure of the view.

> Instead of checking UI without context, we can obtain the same validation implicitly by investing in writing more test cases, specifically test cases that are focused on the accesibility of information. E.g. for application overviews we should have test cases (based on user stories) that verify that all relevant information can be obtained. Such tests cases will also be better decoupled from implementation details.

Not sure what you mean. Can you provide an example or more details?

> We might also want to look into ways to make use of the generated screenshots, e.g. for cheap sanity-checking before making a release

Visual testing can indeed help, but I would suggest to discuss and implement this in a separate PR. This is not a simple task.

> I think it's ok to perform login and user registration programmatically, once we have tested it in a dedicated smoke test. Otherwise, we will end up with a lot of repetitive work for the test runner without gaining any value from it

Agreed. Changed this.

> I see you did not add any tests to cypress/integration/console/applications/create.spec.js that actually perform a successful submit. Just to be sure, we would plan to do some tests that test successful submitting for different scenarios as well, right? This should not be left to the smoke tests only.

Ofc, I added tests for application creation solely to showcase how we can augment the stack configuration in tests.

> Instead of creating users programmatically, why not using a seed with a default user? Considering how many tests are using authenticated views, this should speed up the tests quite a bit.

A single user is not enough. Consider, for example, testing collaborators functionality or admin functionality over the users. Additionally, in some cases we want different users to be used within the same context (`describe`) but different tests (`it`), say, for testing entity listing for various scenarios (we dont want previously created/updated entities to affect other tests. if we use a single user we have to drop the database between the tests and this takes longer than programmatically creating a new user). The question is how many users should we seed initially? Should take the approach taken in registry tests in the identity server where all necessary users are defined before running the tests? 

Also, we already seed the `admin` user. My plan is to use that user to perform specific actions like changing the state of the user, update entity rights, etc.

This is an open discussion and Im not saying that we should discuss/implement this. IMO, looking into ways to speed up tests makes sense when running e2e tests takes ~10mins. Let me know if you have a clear idea how this can be implemented within this PR addressing issues mentioned above.

> IMO, if we do integration tests, we should not limit ourselves to behavior that we can control in the frontend. The user also does not care from which component a usability issue originates from. If we write our end-to-end tests based on real-life scenarios, we should not make different assertions based on what implementation details are present in the background, just like the user wouldn't do that too.

1. We are writing end-to-end tests, not integration tests.
2. What exactly are you testing when asserting on text in the error notification for the error returned from the backend?
	- Does the backend validate requests? Thats the responsibility of the backend to test this behavior.
	- Does the backend return errors for a particular request of a certain type? Thats the responsibility of the backend to test this behavior.
	- Is the `<ErrorNotification />` component working and not breaking the view? Well, thats already covered when the component is rendered. Ideally, we should integrate cypress into storybook and run unit/integration tests for each component.
	- Does the `<ErrorNotification />` component render the error message? 1-2 tests are enough for this and this can be achieved with simply unit testing.
	- Does the `<ErrorNotification />` component render `An unknown error occurred. Please try again later.` for unknown errors? Same here.
3. I dont think that asserting on a particular _text_ gives us much more confidence than asserting on functional requirements:
 	- that the notification is shown and visible when form submission failed.
	- that the url stays the same
	- that the form state is preserved
4. Overall, I dont think that asserting on any hardcoded text is a good idea, but taking into account the number of issues we had with missing or incorrect frontend defined messages we should assert on these. We dont have the tooling to verify that the message exists if someone references `sharedMessages.someMessage` (btw typescript can solve this issue), thats why have I added support for asserting on field errors, warnings, etc.
5. In case of testing error responses, how many assertions is enough? Should we test all permutations of inputs that passes the frontend defined validation?

I suggest to find out what works better in our case. Lets define a bunch of assertions on backend returned data and experiment for a while to see how to proceed further. At the end of the day all the guidelines are just suggestions such that anyone from the dev team can quickly start writing tests that bring as much value as possible with less investment.

cypress/support/commands.js
bafonins: See https://github.com/TheThingsNetwork/lorawan-stack/pull/2939#issuecomment-662472814

pkg/webui/template.go
```diff
@@ -89,17 +89,25 @@ const appHTML = `
...
+			var config = {
+				APP_ROOT:{{.MountPath}},
+				ASSETS_ROOT:{{$assetsBaseURL}},
+				BRANDING_ROOT:{{$brandingBaseURL}},
+				APP_CONFIG:{{.AppConfig}},
+				SITE_NAME:{{.SiteName}},
+				SITE_TITLE:{{.Title}},
+				SITE_SUB_TITLE:{{.SubTitle}},
+				SENTRY_DSN:{{.SentryDSN}},
+				{{with .CSRFToken}}CSRF_TOKEN:{{.}}{{end}},
```
kschiffer: What I mean is the line has an ending comma while being in a conditional statement of the template. When there is no `PAGE_DATA`, it could result in syntax errors since there will be a trailing comma.
A solution would be to not use these conditional statements at the end of the object assignment.

tools/mage/dev.go
```diff
@@ -155,6 +241,15 @@ func (Dev) InitStack() error {
 	)
 }
 
+// StartDevStack starts TTS in end-to-end test configuration.
```
kschiffer, line 242: `in development configuration`

cypress/integration/smoke/registration/index.js
```diff
@@ -0,0 +1,78 @@
...
+  cy.findByLabelText('Email').type(user.email)
+  cy.findByLabelText('Password').type(user.password)
+  cy.findByLabelText('Confirm password').type(user.password)
+  cy.findByRole('button', { name: 'Register' }).click()
+
+  cy.findByTestId('notification')
+    .should('be.visible')
+    .should('contain', 'You have successfully registered and can login now')
+
+  // Login
```
kschiffer:
```suggestion
  // Login.
```

DEVELOPMENT.md
```diff
@@ -716,10 +716,170 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+
+In some cases it can be necessary to select DOM elements using a special selection data attribute. We use `data-test-id` for this purpose. Use this attribute to select DOM elements when more realistic means of selection are not sufficient. Use meaningful but concise ID values, such as `error-notification`.
+
+- Select DOM elements using text captions and labels when possible.
+  - Select form fields by its label via `cy.findByLabelText`, e.g. `cy.findByLabelText('User ID')`. Same for field errors,warnings and descriptions, use `cy.findErrorByLabelText`, `cy.findWarningByLabelText` and `cy.findDescriptionByLabelText`.
+  - Select buttons, links, tabs and other elements that are described by [ARIA roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#Roles) via `cy.findByRole`, e.g. `cy.findByRole('button', {name: 'Submit'})`.
+  - Select text elements via `cy.findByText`, e.g. `cy.findByText('Change password', {selector: 'h1'})`.
```
kschiffer: I would actually not encourage to make a selection (partially) based on the element type. It would be nicer if it was possible to select headings in general `h1, h2, h3, h4, h5`. Ultimately, what matters is that it is a heading, not exactly what kind of heading.

DEVELOPMENT.md
```diff
@@ -716,10 +716,170 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+  - Each test must follow the following naming: `{context}.spec.js`.
+  - One test file must have end-to-end tests dedicated only to a specific entity or view.
+2. `pkg/cypress/{app}/shared` contains all test specification not directly related to a single entity or a view. For example, `side-navigation.spec.js` or `header.spec.js` must be placed into the `cypress/console/shared` folder because both components are present on multiple views and are partially related to the stack entities. Make sure to scope cypress selections within the tested component using `cy.within`.
+3. `pkg/cypress/smoke` contains tests that simulate a complete user story trying to do almost everything a typical user would do. For example, a typical smoke test can verify that the user is able to register, login, create application and register The Things Uno. For more details and diffeence between regular end-to-end and smoke tests see the [End-to-end tests structure](#organizing-end-to-end-tests) section.
+4. and 5. `Cypress` stores screenshots and videos to the appropriate folder after running end-to-end tests. These should not be added to the repository.
+
```
kschiffer, line 863:
```suggestion
```

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
```
kschiffer, line 4: Should we use this to fix some common user types?

cypress/support/commands.js
```diff
@@ -0,0 +1,170 @@
...
+  cy.on('window:before:load', win => {
+    Object.defineProperty(win, '__initStackConfig', {
+      value: () => {
+        fnArray.forEach(fn => fn(win.__ttn_config__))
+      },
+    })
+  })
+})
+
+// Selectors
```
kschiffer:
```suggestion
// Selectors.
```

cypress/support/commands.js
```diff
@@ -0,0 +1,170 @@
...
+    .and('be.visible')
+
+  return cy.get('@error')
+})
+
+// Helper function to select field warning.
+Cypress.Commands.add('findWarningByLabelText', label => {
+  getFieldDescriptorByLabel(label).as('warning')
+
+  // Check for the warning icon
```
kschiffer:
```suggestion
  // Check for the warning icon.
```

package.json
```diff
@@ -75,6 +81,7 @@
   "dependencies": {
     "@formatjs/intl-relativetimeformat": "^2.8.2",
     "@sentry/browser": "^5.19.0",
+    "@testing-library/cypress": "^6.0.0",
```
kschiffer: This should be a `devDependency`, I guess?

pkg/webui/oauth/views/create-account/index.js
```diff
@@ -199,19 +199,22 @@ export default class CreateAccount extends React.PureComponent {
                 title={sharedMessages.userId}
                 name="user_id"
                 autoComplete="username"
+                data-test-id="user_id"
```
kschiffer: These should not be necessary anymore.

pkg/webui/oauth/views/login/index.js
```diff
@@ -147,8 +147,8 @@ export default class OAuth extends React.PureComponent {
               />
               <Form.Field
                 title={sharedMessages.password}
-                component={Input}
                 name="password"
+                component={Input}
```
kschiffer: Remove this diff.

tools/mage/dev.go
kschiffer: So actually, I would like to retain the option to run e2e tests on a separate test database. Right now, running the tests will drop the development database. If I develop, I want to keep my current DB setup independently from the e2e tests, which is why I designed it to be separated in my PR.

.github/workflows/end-to-end.yml
```diff
@@ -0,0 +1,110 @@
...
+        run: tools/bin/mage dev:dbStop dev:dbErase dev:dbStart dev:initStack dev:dbDump js:Build
+      - name: Run frontend end-to-end tests
+        run: tools/bin/mage dev:startDevStack & tools/bin/mage -v js:cypressHeadless
+      - name: Upload screenshots if failed
+        uses: actions/upload-artifact@v2
+        with:
+          name: Frontend test screenshots
+          path: cypress/screenshots
+        if: ${{ failure() }}
+      - name: Upload videos if failed
```
kschiffer: Should we also upload the coverage report?

tools/mage/general.go
```diff
@@ -33,3 +34,11 @@ func targetError(err error) error {
...
 	return nil
 }
+
+func pathExists(filename string) bool {
+	_, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
```
htdvisser: This is technically incorrect. If you don't have permission to read the dir in which you're looking for the file, it would return that the file exists, while you don't actually know. 

Future developers who see that there is a `pathExists` in this package might use it incorrectly, so I don't think it's a good idea to do it this way.

Also, you can shorten L40-43 to `return !os.IsNotExist(err)`, which makes me wonder if it's really worth that one line you might save in the places where you'd use this.

DEVELOPMENT.md
```diff
@@ -716,10 +716,169 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+Make sure to build the frontend assets and run the stack before executing end-to-end tests.
+
+`Cypress` provides two modes for running tests: headless and interactive.
+- Headless mode - will not display any browser GUI and output test progress into your terminal instead. This is helpful when one just needs see the results of the tests.
+- Interactive mode - will run an `Electron` based application together with the full-fledged browser. This is helpful when developig frontend applications as it provides hot reload, time travelling, browser extensions and DOM access.
+
+> Note: Currently, we test our frontend only in Chromium based browsers.
+
+You can run `Cypress` in the headless mode by running the following command:
+```bash
```
htdvisser, line 753: Newlines between blocks

DEVELOPMENT.md
```diff
@@ -716,10 +716,169 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+$ tools/bin/mage endToEnd:CypressHeadless
+```
+
+You can run `Cypress` in the interactive mode by running the following command:
+```bash
```
htdvisser: Newlines between blocks

DEVELOPMENT.md
```diff
@@ -716,10 +716,169 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+  </div>
+  ```
+
+  ```js
+    // Bad. This assertions will pass while not being visible to the user.
+    cy.findByTestId('test').should('exists')
+
+    // Good. This assertion will rightfully fail.
+    cy.findByTestId('test').should('be.visible')
+  ```
```
htdvisser: Strange indentation inside the code block?

DEVELOPMENT.md
```diff
@@ -716,10 +716,169 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+```
+
+#### Code coverage
+
+Code coverage can be used to verify that tests invoke code for handling edge cases.
+To generate code coverage report run:
+
+- Global text summary.
+```bash
+  npx nyc report --reporter=text-summary
```
htdvisser: No $? Also below

tools/mage/dev.go
```diff
@@ -70,10 +76,90 @@ func execDockerCompose(args ...string) error {
...
+func (Dev) SQLStop() error {
+	if mg.Verbose() {
+		fmt.Println("Stopping SQL database")
+	}
+	return execDockerCompose(append([]string{"stop"}, sqlDatabase)...)
+}
+
+// SQLMakeSnapshot stores the current cockroach data folder for later restores.
+func (Dev) SQLMakeSnapshot() error {
+	if mg.Verbose() {
```
htdvisser: Shouldn't we first make sure the database is stopped?

tools/mage/dev.go
```diff
@@ -70,10 +76,90 @@ func execDockerCompose(args ...string) error {
...
+	}
+	if err := sh.Copy(from, to); err != nil {
+		return err
+	}
+	return d.SQLStart()
+}
+
+// DBDump performs an SQL database dump of the dev database to the .cache folder.
+func (Dev) DBDump() error {
+	if mg.Verbose() {
```
htdvisser, line 119: Shouldn't we first make sure the database is running?

tools/mage/dev.go
```diff
@@ -70,10 +76,90 @@ func execDockerCompose(args ...string) error {
...
+	}
+	if err := os.MkdirAll(".cache", 0755); err != nil {
+		return err
+	}
+	return execDockerComposeWithOutput(filepath.Join(".cache", "sqldump.sql"), "exec", "-T", "cockroach", "./cockroach", "dump", devDatabaseName, "--insecure")
+}
+
+// DBRestore restores the dev database using a previously generated dump.
+func (Dev) DBRestore() error {
+	mg.Deps(Js.Deps)
```
htdvisser: Shouldn't we first make sure the database is running?

tools/mage/scripts/restore-db-dump.js
```diff
@@ -0,0 +1,55 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
```
htdvisser: I'm not a big fan of using mixed Mage / JavaScript for our dev tooling here.

tools/mage/dev.go
```diff
@@ -70,10 +76,90 @@ func execDockerCompose(args ...string) error {
...
+
+// DBDump performs an SQL database dump of the dev database to the .cache folder.
+func (Dev) DBDump() error {
+	if mg.Verbose() {
+		fmt.Println("Saving database dump")
+	}
+	if err := os.MkdirAll(".cache", 0755); err != nil {
+		return err
+	}
+	return execDockerComposeWithOutput(filepath.Join(".cache", "sqldump.sql"), "exec", "-T", "cockroach", "./cockroach", "dump", devDatabaseName, "--insecure")
```
htdvisser: Not a big fan of this, as it assumes we always use CockroachDB (which we don't; in CI we use PostgreSQL)

tools/mage/dev.go
```diff
@@ -70,10 +76,90 @@ func execDockerCompose(args ...string) error {
...
+}
+
+// DBRestore restores the dev database using a previously generated dump.
+func (Dev) DBRestore() error {
+	mg.Deps(Js.Deps)
+	if mg.Verbose() {
+		fmt.Println("Restoring database from dump")
+	}
+	return sh.Run("node", filepath.Join("tools", "mage", "scripts", "restore-db-dump.js"), "--db", devDatabaseName, "--dump", filepath.Join(".cache", "sqldump.sql"))
```
htdvisser: Not really happy with an external script that just runs

```
DROP DATABASE <NAME>
CREATE DATABASE <NAME>
```

and then the queries in `sqldump.sql`.

We can easily do that here directly.

.github/workflows/end-to-end.yml
```diff
@@ -0,0 +1,110 @@
+name: End-to-end Tests
+on:
+  pull_request:
+    paths:
+      - '*.go'
+      - '.revive.toml'
```
rvolosatovs: This should not trigger end-to-end tests

.github/workflows/end-to-end.yml
```diff
@@ -0,0 +1,110 @@
+name: End-to-end Tests
+on:
+  pull_request:
+    paths:
```
rvolosatovs: https://github.com/TheThingsNetwork/lorawan-stack/pull/2764/files#r447485316

.github/workflows/end-to-end.yml
```diff
@@ -0,0 +1,110 @@
...
+      - 'pkg/webui/**'
+      - 'sdk/js/**'
+      - 'yarn.lock'
+
+jobs:
+  end-to-end:
+    name: Frontend based (cypress)
+    runs-on: ubuntu-18.04
+    env:
+      should_cache: true # Might be useful for debugging.
```
rvolosatovs: https://github.com/TheThingsNetwork/lorawan-stack/pull/2764/files#r447485665

tools/mage/dev.go
bafonins: Shall we do that? The main use of `DBDump` is to get the database dump after `InitStack` while `InitStack` does not depend on `DBStart`.
Also, if the database is not started a proper error is displayed.

cypress/fixtures/example.json
bafonins: What do you mean?

kschiffer: Sorry, disregard this.

tools/mage/dev.go
bafonins: But this actually works in CI 🤷‍♂️

tools/mage/dev.go
```diff
@@ -55,12 +59,16 @@ func (Dev) Misspell() error {
...
+	sqlDatabase           = "cockroach"
+	redisDatabase         = "redis"
 	devDataDir            = ".env/data"
-	devDatabaseName       = "ttn_lorawan"
+	devDatabaseName       = "ttn_lorawan_dev"
 	devDockerComposeFlags = []string{"-p", "lorawan-stack-dev"}
+	databaseURI           = fmt.Sprintf("postgresql://root@localhost:26257/%s?sslmode=disable", devDatabaseName)
 )
 
+var devDatabases = []string{sqlDatabase, redisDatabase}
```
rvolosatovs: nit; this can be in the same `var` block above

tools/mage/dev.go
```diff
@@ -70,10 +78,97 @@ func execDockerCompose(args ...string) error {
...
 }
 
+func execDockerComposeWithOutput(filepath string, args ...string) error {
+	output, err := sh.Output("docker-compose", dockerComposeFlags(args...)...)
+	if err != nil {
+		return err
+	}
+	message := []byte(output)
+	err = ioutil.WriteFile(filepath, message, 0644)
+	return err
```
rvolosatovs:
```suggestion
	return ioutil.WriteFile(filepath, []byte(output), 0644)
```

tools/mage/dev.go
```diff
@@ -70,10 +78,97 @@ func execDockerCompose(args ...string) error {
 	return err
 }
 
+func execDockerComposeWithOutput(filepath string, args ...string) error {
```
rvolosatovs: This function is way too specific and is inconsistent with Go analogues:
https://github.com/TheThingsNetwork/lorawan-stack/blob/31119a0db8e9a7933e021195623f50ba5420f2df/tools/mage/go.go#L54-L72

tools/mage/dev.go
```diff
@@ -70,10 +78,97 @@ func execDockerCompose(args ...string) error {
...
+	if err = db.Exec(fmt.Sprintf("DROP DATABASE %s;", devDatabaseName)).Error; err != nil {
+		return err
+	}
+	if err = db.Exec(fmt.Sprintf("CREATE DATABASE %s;", devDatabaseName)).Error; err != nil {
+		return err
+	}
+	if err = db.Exec(string(b)).Error; err != nil {
+		return err
+	}
+
```
rvolosatovs: nit; remove newline

tools/mage/dev.go
```diff
@@ -70,10 +78,97 @@ func execDockerCompose(args ...string) error {
...
+	return nil
```
rvolosatovs:
```suggestion
	return db.Exec(fmt.Sprintf(`DROP DATABASE %s;
CREATE DATABASE %s;
%s`,
 devDatabaseName, devDatabaseName, string(b)),
).Error
```

tools/mage/go.go
```diff
@@ -71,6 +71,10 @@ func outputGo(cmd string, args ...string) (string, error) {
 	return buf.String(), nil
 }
 
+func runGoMute(args ...string) error {
```
rvolosatovs: This is only ever used once in the whole codebase, remove and just use `execGo(nil, os.Stderr, "run", args...)` explicitly

tools/mage/dev.go
```diff
@@ -155,6 +250,15 @@ func (Dev) InitStack() error {
 	)
 }
 
+// StartDevStack starts TTS in end-to-end test configuration.
+func (Dev) StartDevStack() error {
+	os.Setenv("TTN_LW_IS_DATABASE_URI", databaseURI)
+	if mg.Verbose() {
+		fmt.Println("Starting the Stack")
+	}
+	return runGoMute("./cmd/ttn-lw-stack", "start")
```
rvolosatovs: Use `filepath.Join`

tools/mage/dev.go
rvolosatovs: Split line for readability - see e.g. https://github.com/TheThingsNetwork/lorawan-stack/blob/31119a0db8e9a7933e021195623f50ba5420f2df/tools/mage/dev.go#L135-L140

tools/mage/js.go
```diff
@@ -76,7 +77,32 @@ func (js Js) runWebpack(config string, args ...string) error {
...
-	return js.runYarnCommandV("eslint", append([]string{"--color", "--no-ignore"}, args...)...)
+	return js.runYarnCommand("eslint", append([]string{"--color", "--no-ignore"}, args...)...)
+}
+
+func (js Js) waitOn() error {
+	u, err := url.Parse(js.getURL())
+	if err != nil {
+		return err
+	}
+	return js.runYarnCommand("wait-on", append([]string{
```
rvolosatovs: `append` and `[]string{}` should not be necessary here

tools/mage/js.go
```diff
@@ -76,7 +77,32 @@ func (js Js) runWebpack(config string, args ...string) error {
...
+		fmt.Sprintf("--timeout=%d", 120000),
+		fmt.Sprintf("--interval=%d", 1000),
+		fmt.Sprintf("http-get://%s/oauth", u.Host),
+	})...)
+}
+
+func (js Js) runCypress(command string, args ...string) error {
+	mg.Deps(js.waitOn)
+
+	return js.runYarnCommand("cypress", append([]string{command, fmt.Sprintf("--config-file=%s", filepath.Join("config", "cypress.json")), "--config", fmt.Sprintf("baseUrl=%s", js.getURL())}, args...)...)
```
rvolosatovs: split line

tools/mage/js.go
```diff
@@ -76,7 +77,32 @@ func (js Js) runWebpack(config string, args ...string) error {
...
+	return js.runYarnCommand("wait-on", append([]string{
+		fmt.Sprintf("--timeout=%d", 120000),
+		fmt.Sprintf("--interval=%d", 1000),
+		fmt.Sprintf("http-get://%s/oauth", u.Host),
+	})...)
+}
+
+func (js Js) runCypress(command string, args ...string) error {
+	mg.Deps(js.waitOn)
+
```
rvolosatovs: nit; remove newline

tools/mage/js.go
```diff
@@ -76,7 +77,32 @@ func (js Js) runWebpack(config string, args ...string) error {
...
+	return js.runYarnCommand("cypress", append([]string{command, fmt.Sprintf("--config-file=%s", filepath.Join("config", "cypress.json")), "--config", fmt.Sprintf("baseUrl=%s", js.getURL())}, args...)...)
+}
+
+func (js Js) getURL() string {
```
rvolosatovs: 1. Avoid `getX()`, just write `X()` - this is a convention in Go
2. `URL` is too ambiguous
```suggestion
func (js Js) frontendURL() string {
```
Or maybe you have a better suggestion?

tools/mage/js.go
```diff
@@ -76,7 +77,32 @@ func (js Js) runWebpack(config string, args ...string) error {
...
+	if js.isProductionMode() {
+		return "http://localhost:1885"
```
rvolosatovs: Would be nice to at least have the ports defined as consts above

tools/mage/dev.go
```diff
@@ -70,10 +78,97 @@ func execDockerCompose(args ...string) error {
...
+
+// DBDump performs an SQL database dump of the dev database to the .cache folder.
+func (Dev) DBDump() error {
+	if mg.Verbose() {
+		fmt.Println("Saving database dump")
+	}
+	if err := os.MkdirAll(".cache", 0755); err != nil {
+		return err
+	}
+	return execDockerComposeWithOutput(filepath.Join(".cache", "sqldump.sql"), "exec", "-T", "cockroach", "./cockroach", "dump", devDatabaseName, "--insecure")
```
rvolosatovs: Moreover, this is the only usage of `execDockerComposeWithOutput` in the whole codebase - just remove the helper and use `execDockerCompose` explicitly here

tools/mage/js.go
```diff
@@ -76,7 +82,36 @@ func (js Js) runWebpack(config string, args ...string) error {
...
+	return js.runYarnCommand("wait-on", []string{
+		fmt.Sprintf("--timeout=%d", 120000),
+		fmt.Sprintf("--interval=%d", 1000),
+		fmt.Sprintf("http-get://%s/oauth", u.Host),
+	}...)
+}
+
+func (js Js) runCypress(command string, args ...string) error {
+	mg.Deps(js.waitOn)
+	// return js.runYarnCommand("cypress", append([]string{command, fmt.Sprintf("--config-file=%s", filepath.Join("config", "cypress.json")), "--config", fmt.Sprintf("baseUrl=%s", js.frontendURL())}, args...)...)
```
rvolosatovs: no commented-out code please

cypress/integration/oauth/change-password.spec.js
```diff
@@ -0,0 +1,126 @@
...
+      .and('be.visible')
+    cy.findErrorByLabelText('Confirm password')
+      .should('contain.text', 'Confirm password is required')
+      .and('be.visible')
+
+    cy.location('pathname').should('eq', `${Cypress.config('oauthRootPath')}/update-password`)
+  })
+
+  // TODO: Remove `skip` after https://github.com/TheThingsNetwork/lorawan-stack/issues/2951.
+  it.skip('succeeds changing password when revoking access', () => {
```
kschiffer:
```suggestion
  it('succeeds changing password when revoking access', () => {
```

Issue is resovled.

cypress/integration/oauth/change-password.spec.js
```diff
@@ -0,0 +1,126 @@
...
+    cy.findByLabelText('Confirm password').type(`${newPassword}{enter}`)
+
+    cy.findByTestId('notification')
+      .should('be.visible')
+      .should('contain', 'Password changed')
+    cy.location('pathname').should('include', `${Cypress.config('oauthRootPath')}/login`)
+  })
+
+  // TODO: Remove `skip` after https://github.com/TheThingsNetwork/lorawan-stack/issues/2951.
+  it.skip('succeeds changing password without revoking access', () => {
```
kschiffer: Same.

cypress/integration/console/applications/create.spec.js
```diff
@@ -0,0 +1,134 @@
...
+    cy.visit(`${Cypress.config('consoleRootPath')}/applications/add`)
+
+    cy.findByText('Add application', { selector: 'h1' }).should('be.visible')
+    cy.findByLabelText('Application ID')
+      .should('be.visible')
+      .and('have.attr', 'placeholder')
+      .and('eq', 'my-new-application')
+    cy.findByLabelText('Application name')
+      .should('be.visible')
+      .should('be.visible')
```
kschiffer:
```suggestion
```

cypress/support/commands.js
```diff
@@ -0,0 +1,175 @@
...
+
+  // Reset cookies and local storage to avoid csrf and session state inconsitencies within tests.
+  cy.clearCookies()
+  cy.clearLocalStorage()
+})
+
+// Helper function to quickly seed the database to a fresh state using a
+// previously generated sql dump.
+Cypress.Commands.add('dropAndSeedDatabase', () => {
+  // Cy.exec('node tools/mage/scripts/restore-db-dump.js')
```
kschiffer:
```suggestion
```

DEVELOPMENT.md
```diff
@@ -716,10 +716,176 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+
+We distinguish between regular end-to-end tests and smoke tests. With smoke tests we try to accumulate application state (redux store, local storage, application cache, etc.) and verify that the application works as expected. When writing smoke tests we comply with he following guidelines:
+
+- Smoke tests are testing complete user stories in a **wide and shallow** manner, meaning:
+  - performing some complex and critical flow that touches multiple components, API's and/or views
+  - not testing different configurations or preconditions of the same flow in depth
+  - For example, when testing registration of The Things Uno:
+    1. Add test file `cypress/integration/smoke/devices/create.js`
+    2. Describe the whole user story to register the device including creating an application (or using an existing one), link the application and create the end device.
+- For example, when testing registration of The Things Uno:
```
kschiffer:
```suggestion
```

DEVELOPMENT.md
```diff
@@ -716,10 +716,176 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+```
+
+1. `pkg/cypress/integrations` contains all test specifications.
+  - Each test file must be placed into corresponding folder (`console`/`oauth`).
+  - Each test must follow the following naming: `{context}.spec.js`.
+  - One test file must have end-to-end tests dedicated only to a specific entity or view.
+2. `pkg/cypress/{app}/shared` contains all test specification not directly related to a single entity or a view. For example, `side-navigation.spec.js` or `header.spec.js` must be placed into the `cypress/console/shared` folder because both components are present on multiple views and are partially related to the stack entities. Make sure to scope cypress selections within the tested component using `cy.within`.
+3. `pkg/cypress/smoke` contains tests that simulate a complete user story trying to do almost everything a typical user would do. For example, a typical smoke test can verify that the user is able to register, login, create application and register The Things Uno. For more details and diffeence between regular end-to-end and smoke tests see the [End-to-end tests structure](#organizing-end-to-end-tests) section.
```
kschiffer:
```suggestion
3. `pkg/cypress/integration/smoke` contains tests that simulate a complete user story trying to do almost everything a typical user would do. For example, a typical smoke test can verify that the user is able to register, login, create application and register The Things Uno. For more details and diffeence between regular end-to-end and smoke tests see the [End-to-end tests structure](#organizing-end-to-end-tests) section.
```

DEVELOPMENT.md
```diff
@@ -716,10 +716,176 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+- Consider various stack configurations when writing end-to-end tests. Some views have different UI depending on availability of diffenent stack components. For example, the end device wizard looks different for deployments with complete cluster (NS+JS+AS) and for JS-only configuration. Likewise, sections and entire views can be enabled or disabled based on our feature toggles. If your test scenario differs based on different feature toggle conditions, make sure to probe these preconditions in your tests.
+
+##### Smoke tests
+
+We distinguish between regular end-to-end tests and smoke tests. With smoke tests we try to accumulate application state (redux store, local storage, application cache, etc.) and verify that the application works as expected. When writing smoke tests we comply with he following guidelines:
```
kschiffer:
```suggestion
We distinguish between regular end-to-end tests and smoke tests. While regular end-to-end tests are scoped to a specific view or component and tests those in depth, smoke tests are testing complete user stories that are critical to the overall initegrity of the application and usually comprise multiple components and views, e.g. login flow, user registration or creation of applications. When writing smoke tests we comply with he following guidelines:
```

Clarifying the difference between regular e2e and smoke tests a little more.

DEVELOPMENT.md
```diff
@@ -716,10 +716,176 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+
+- Smoke tests are testing complete user stories in a **wide and shallow** manner, meaning:
+  - performing some complex and critical flow that touches multiple components, API's and/or views
```
kschiffer:
```suggestion
  - performing some complex and critical flow that touches multiple components, APIs and/or views
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,81 @@
...
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(npx yarn cache dir)"
+      - name: Initialize Yarn module cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+      - name: Initialize messages and locales cache
```
rvolosatovs:
```suggestion
      - name: Initialize message and locale cache
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,81 @@
...
+            .cache/babel
+            .cache/messages
+            pkg/webui/console/locales/.backend
+          key: ${{ runner.os }}-messages-cache-${{ hashFiles('pkg/webui/**/*.js') }}-${{ hashFiles('config/messages.json')}}
+      - name: Initialize public folder cache
+        uses: actions/cache@v2
+        with:
+          path: public
+          key: ${{ runner.os }}-public-cache-${{ hashFiles('pkg/webui/**') }}-${{ hashFiles('sdk/js/**/*.js') }}-${{ hashFiles('config/webpack.config.babel.js') }}-${{ hashFiles('**/yarn.lock') }}-${{ hashFiles('pkg/ttnpb/field_mask_validation.go') }}
+      - name: Initialize JS SDK cache
```
rvolosatovs: Why is this only performed here and not in other workflows using JS SDK?
Please add to all relevant workflows in order to avoid inconsistencies. (can be in a separate PR)

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,81 @@
...
+            ${{ runner.os }}-yarn-
+      - name: Initialize messages and locales cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            .cache/babel
+            .cache/messages
+            pkg/webui/console/locales/.backend
+          key: ${{ runner.os }}-messages-cache-${{ hashFiles('pkg/webui/**/*.js') }}-${{ hashFiles('config/messages.json')}}
+      - name: Initialize public folder cache
```
rvolosatovs, line 34: IDEM

DEVELOPMENT.md
```diff
@@ -716,10 +716,175 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
 
 ## Testing
 
+### Unit Tests
+
+To run unit tests, observe the following mage targets:
```
rvolosatovs:
```suggestion
To run unit tests, use the following mage targets:
```
or *run*

DEVELOPMENT.md
```diff
@@ -716,10 +716,175 @@ The message definitions in `pkg/webui/locales` can be used to provide translatio
...
+$ tools/bin/mage endToEnd:CypressHeadless
+```
+
+You can run `Cypress` in the interactive mode by running the following command:
+
+```bash
+$ tools/bin/mage endToEnd:CypressInteractive
+```
+
+#### Running frontend end-to-end tests in CI
```
rvolosatovs: Not sure what this section is for, we can just read the workflow source, right?

tools/mage/dev.go
```diff
@@ -155,6 +239,15 @@ func (Dev) InitStack() error {
 	)
 }
 
+// StartDevStack starts TTS in end-to-end test configuration.
+func (Dev) StartDevStack() error {
+	os.Setenv("TTN_LW_IS_DATABASE_URI", databaseURI)
```
rvolosatovs, line 244: I would expect a running DB to be a dependency of this.
If the reason why it's not is that DB may be started externally (i.e. not via this mage file), how about a target, which would check if DB is available, and, if not, starts the DB?

rvolosatovs (approved):

rvolosatovs (approved):

.github/workflows/e2e.yml
rvolosatovs: Is there any action taken in response to this?
Are you going to update this PR? Are you going to open another PR before merging this one? Are you going to file an issue?

bafonins: I removed caching steps that do no appear in other js workflow. We can add those later if we experience any noticeable performance issues.

kschiffer: Note that I was able to bring the run time down [to about 4 minutes](https://github.com/kschiffer/lorawan-stack/runs/783055521) under optimal circumstances with the caching I had applied in the other e2e PR.

PR #2941, opened 2020-07-20
Return registry error cause code

### LINKED ISSUE #2940: AppSKey not recovered via interop
#### Summary

AppSKey is not recovered via interop

#### Steps to Reproduce

1. Setup stack the usual way
2. Register device on external Join Server
3. Unlink AS
4. Activate device
5. Link AS
6. Send uplink message

#### What do you see now?

AS doesn't recover the AppSKey from the JS via interop

```json
{
  "name": "as.up.data.drop",
  "time": "<snip>",
  "identifiers": [
    {
      "device_ids": {
        "device_id": "<snip>",
        "application_ids": {
          "application_id": "<snip>"
        },
        "dev_eui": "<snip>",
        "join_eui": "70B3D57ED0000000",
        "dev_addr": "270000<snip>"
      }
    }
  ],
  "data": {
    "namespace": "pkg/applicationserver",
    "name": "app_s_key",
    "message_format": "failed to get AppSKey",
    "cause": {
      "namespace": "pkg/joinserver",
      "name": "registry_operation",
      "message_format": "registry operation failed",
      "correlation_id": "90e316b18fd74b488ef8d5d0dd0c2cec",
      "cause": {
        "namespace": "pkg/redis",
        "name": "not_found",
        "message_format": "entity not found",
        "code": 5
      },
      "code": 13
    },
    "code": 13
  },
  "correlation_ids": [
    "<snip>"
  ],
  "origin": "<snip>",
  "context": {
    "tenant-id": "<snip>"
  },
  "visibility": {
    "rights": [
    ]
  }
}

```

AppSKey recovery via JS interop

#### Environment

3.8.6

The error code is not checked correctly; returned code is `Internal` while AS would go to JS via interop if it were `NotFound`.

See reproduction steps

Already working on it

johanstokking:
#### Summary

Closes #2940

#### Changes

- JS now returns the registry error cause's code instead of `Internal` (13). This is useful for clients to act on whether the entity exists in the registry, for example
- AS contacts interop JS when the cluster-local JS reported not found (no functional change, but now it works)
- Error comparison now doesn't take code into account if any is `Unknown`

#### Testing

Unit tests

##### Regressions

Clients relying on `Internal` code from JS device registry operations break.

#### Notes for Reviewers

@htdvisser please check `pkg/errors`

#### Checklist

### DIFF
--- a/pkg/errors/comparison.go
+++ b/pkg/errors/comparison.go
@@ -14,7 +14,7 @@
 
 package errors
 
-// Resemble returns true iff the given errors resemble, meaning that the Code,
+// Resemble returns true iff the given errors resemble, meaning that the
 // Namespace and Name of the errors are equal. A nil error only resembles nil.
 // Invalid errors or definitions (including typed nil) never resemble anything.
 func Resemble(a, b error) bool {
@@ -29,7 +29,6 @@ func Resemble(a, b error) bool {
 	if !ok {
 		return false
 	}
-	return ttnA.Code() == ttnB.Code() &&
-		ttnA.Namespace() == ttnB.Namespace() &&
+	return ttnA.Namespace() == ttnB.Namespace() &&
 		ttnA.Name() == ttnB.Name()
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Validation of keys for gateway metrics and version fields.
 - Read only access for the gateway overview page in the Console.
 - Fix an issue that frequently caused event data views crashing in the Console.
+- Application Server contacting Join Server via interop for fetching the AppSKey.
 
 ### Security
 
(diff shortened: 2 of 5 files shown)


pkg/errors/comparison.go
```diff
@@ -29,7 +31,7 @@ func Resemble(a, b error) bool {
 	if !ok {
 		return false
 	}
-	return ttnA.Code() == ttnB.Code() &&
+	return (ttnA.Code() == uint32(codes.Unknown) || ttnB.Code() == uint32(codes.Unknown) || ttnA.Code() == ttnB.Code()) &&
```
htdvisser, line 34: I don't really understand this. It shouldn't be possible to define an error with the same name in the same namespace, but give it a different code.

johanstokking: The big question is: should https://github.com/TheThingsNetwork/lorawan-stack/pull/2941/files#diff-4ba598240540ba3699477b82237d81b5R59 `BeTrue` or `BeFalse`. I think it makes more sense for this to be true.

htdvisser: Okay. Then maybe we should just only compare the namespace and name.

johanstokking: Also fine with me

PR #2943, opened 2020-07-20
Sync gateway clock via uplink token

### LINKED ISSUE #2912: Synchronize Gateway Server clocks via uplink token if needed
#### Summary

Synchronize Gateway Server clocks via uplink token if needed

#### Why do we need this?

To make sure that GS instances handling the downstream connection to gateways have time synchronization for scheduling downlink. The Gateway Server instance for upstream and downstream may be different if the UDP session stickiness of the load balancer doesn't work, either because propagation is not fast enough, or if the originating IP is not the same.

Clock synchronization only on upstream

If there's no clock synchronization when scheduling downlink, use the uplink token to synchronize the clock.

Include the absolute server time of the upstream GS instance in the uplink token, and let the GS downstream instance use that. This requires both instances to have little delta between their system clocks.

Unit testing

Yes

johanstokking:
#### Summary

Closes #2912

#### Changes

- Synchronize gateway clock via uplink token
- Minor cleanup in Basic Station frontend

#### Testing

Unit testing

##### Regressions

Should be covered by integration tests and new unit test

#### Notes for Reviewers

@KrishnaIyer is the main reviewer.

Functional changes primarily in `pkg/gatewayserver/scheduling`

@htdvisser @rvolosatovs no review necessary; API is not affected, only GS in-memory storage struct

#### Checklist


pkg/gatewayserver/scheduling/scheduler.go
```diff
@@ -252,32 +252,45 @@ var (
...
 // If there are round-trip times available, the maximum value will be used instead of ScheduleTimeShort.
-func (s *Scheduler) ScheduleAt(ctx context.Context, payloadSize int, settings ttnpb.TxSettings, rtts RTTs, priority ttnpb.TxSchedulePriority) (Emission, error) {
+func (s *Scheduler) ScheduleAt(ctx context.Context, opts Options) (Emission, error) {
 	defer trace.StartRegion(ctx, "schedule transmission").End()
 
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 	if !s.clock.IsSynced() {
-		return Emission{}, errNoClockSync.New()
+		if token := opts.UplinkToken; token != nil {
```
htdvisser, line 272: Maybe good to also check `!token.ServerTime.IsZero()`, just to catch the edge case where uplink and downlink GS are not the same version.

pkg/gatewayserver/io/basicstationlns/messages/upstream.go
```diff
@@ -181,7 +168,6 @@ func (req *JoinRequest) ToUplinkMessage(ids ttnpb.GatewayIdentifiers, bandID str
 		RSSI:               req.RadioMetaData.UpInfo.RSSI,
 		ChannelRSSI:        req.RadioMetaData.UpInfo.RSSI,
 		SNR:                req.RadioMetaData.UpInfo.SNR,
```
KrishnaIyer, line 170: You'll need to add `RCtx` as the Antenna Index here then;
```
AntennaIndex:       uint32(req.RadioMetaData.UpInfo.RCtx)
```

pkg/gatewayserver/io/basicstationlns/messages/upstream.go
```diff
@@ -353,7 +324,6 @@ func (updf *UplinkDataFrame) ToUplinkMessage(ids ttnpb.GatewayIdentifiers, bandI
 		RSSI:               updf.RadioMetaData.UpInfo.RSSI,
 		ChannelRSSI:        updf.RadioMetaData.UpInfo.RSSI,
 		SNR:                updf.RadioMetaData.UpInfo.SNR,
```
KrishnaIyer, line 326: Also here;
```
AntennaIndex:       uint32(updf.RadioMetaData.UpInfo.RCtx),
```

pkg/gatewayserver/io/basicstationlns/messages/upstream.go
johanstokking: Why "then"? The `AntennaIndex` was never set, also not via the uplink token.

But I replaced setting and getting `RCtx` using `AntennaIndex` as backing. I think that's an important fix anyway for multi-antenna Basic Station gateways, as `AntennaIndex` was left to `0`, but it was set in the uplink token which would get overwritten by GS further in the chain, with `AntennaIndex` still at `0`, breaking downlink because the timestamp is slightly off.

KrishnaIyer: "then" because you removed the uplink token handling here 😉 .

PR #2948, opened 2020-07-21
Fix Basic Station Authentication

### LINKED ISSUE #2882: Basic Station always allows unauthenticated connections when API Key is defined
#### Summary

Basic Station always allows unauthenticated connections when API Key is defined.

#### Steps to Reproduce

1. Create a Gateway.
2. Create an API Key for Gateway with Link Rights.
3. Connect to the `traffic` endpoint without authentication.

#### What do you see now?

- Unauthenticated connections are allowed. They are able to send uplinks and receive downlinks, basically spoof a real gateway.
                  
- If a gateway has an API Key associated with it, it must not be allowed to connect with key as part of the auth header.
- MQTT and gRPC already enforce this by default. UDP gateway don't need auth anyway.

The issue is that for Basic Station, if there's no Auth header, we assume the Link Rights. However, if the gateway has an API Key with Link rights defined, then the connection must provide this.
There's currently no way to check if a gateway has an API key associated with it or not.

#### Environment

`tts/ttes v3.8.5` 

There are 3 options
1. Don't allow unauthenticated connections at all (like MQTT/gRPC). This is a deviation from the LNS Spec but we already enforce this for CUPS connections.
2. Add `gs.basic-station.allow-unauthenticated` flag for debugging and set to false for production env.
3. Check if there's an API key per gateway
    - Support `bool RequireAPIKey(requiredRights []ttnpb.Right)` in the gateway registry.
    - For every connection, check if there's an API Key with `Link` rights defined in IS. If yes, make sure that the gateway has an Auth header or else refuse connection.

@johanstokking : I prefer option 2. It's easy to implement, we can support debugging on local deployments and enforce using keys in production.

Based on the implementation, connect a gateway and test the possibilities.

Yes

KrishnaIyer:
#### Summary

Closes #2882 

#### Changes

- Add new `allow-unauthenticated` flag (default is `true` to make it compatible with existing deployments.
- Integrate functions
- Test

#### Testing

Unit Tests

|`allow-unauthenticated` flag (down)/Header in Auth(right)|Yes|No|
|---|----|---|
|True|Check auth and allow connection|Assume rights and allow connection|
|False|Check auth and allow connection|Don't allow|

##### Regressions

Covered by UTs

Also checked that there's no regression in the config due to the refactoring

This PR
```
gs:
  basic-station:
    allow-unauthenticated: true
    fallback-frequency-plan-id: ""
    listen: :1887
    listen-tls: :8887
    use-traffic-tls-address: false
    ws-ping-interval: 30s
```

On `master`
```
gs:
  basic-station:
    fallback-frequency-plan-id: ""
    listen: :1887
    listen-tls: :8887
    use-traffic-tls-address: false
    ws-ping-interval: 30s
```

#### Notes for Reviewers

Requires a small PR in `lorawan-stack-aws`. Will add that.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Options to allow/deny non-admin users to create applications, gateways, etc. (the the `is.user-rights.*` options).
 - Admins now receive emails about requested user accounts that need approval.
 - Support for synchronizing gateway clocks via uplink tokens. UDP gateways may not connect to the same Gateway Server instance.
+- Option to allow unauthenticated Basic Station connections. Unset `gs.basic-station.allow-authenticated` to enforce auth check for production clusters.
 
 ### Changed
 
@@ -50,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Read only access for the gateway overview page in the Console.
 - Fix an issue that frequently caused event data views crashing in the Console.
 - Application Server contacting Join Server via interop for fetching the AppSKey.
+- Basic Station always allows unauthenticated connections when API Key is defined.
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Options to allow/deny non-admin users to create applications, gateways, etc. (the the `is.user-rights.*` options).
 - Admins now receive emails about requested user accounts that need approval.
 - Support for synchronizing gateway clocks via uplink tokens. UDP gateways may not connect to the same Gateway Server instance.
-- Option to allow unauthenticated Basic Station connections. Unset `gs.basic-station.allow-authenticated` to enforce auth check for production clusters.
+- Option to allow unauthenticated Basic Station connections. Unset `gs.basic-station.allow-unauthenticated` to enforce auth check for production clusters. Please note that unauthenticated connections in existing connections will not be allowed unless this is set.
 
 ### Changed
 
@@ -51,7 +51,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Read only access for the gateway overview page in the Console.
 - Fix an issue that frequently caused event data views crashing in the Console.
 - Application Server contacting Join Server via interop for fetching the AppSKey.
-- Basic Station always allows unauthenticated connections when API Key is defined.
 
 ### Security
 
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -246,7 +246,11 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 	if conf.BasicStation.FallbackFrequencyPlanID != "" {
 		bsCtx = frequencyplans.WithFallbackID(bsCtx, conf.BasicStation.FallbackFrequencyPlanID)
 	}
-	bsWebServer := basicstationlns.New(bsCtx, gs, conf.BasicStation.UseTrafficTLSAddress, conf.BasicStation.WSPingInterval)
+	bsWebServer := basicstationlns.New(bsCtx, gs, basicstationlns.Options{
+		UseTrafficTLSAddress: conf.BasicStation.UseTrafficTLSAddress,
+		WSPingInterval:       conf.BasicStation.WSPingInterval,
+		AllowUnauthenticated: conf.BasicStation.AllowUnauthenticated,
+	})
 	for _, endpoint := range []component.Endpoint{
 		component.NewTCPEndpoint(conf.BasicStation.Listen, "Basic Station"),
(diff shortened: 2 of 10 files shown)


pkg/gatewayserver/gatewayserver.go
```diff
@@ -246,7 +246,11 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 	if conf.BasicStation.FallbackFrequencyPlanID != "" {
 		bsCtx = frequencyplans.WithFallbackID(bsCtx, conf.BasicStation.FallbackFrequencyPlanID)
 	}
-	bsWebServer := basicstationlns.New(bsCtx, gs, conf.BasicStation.UseTrafficTLSAddress, conf.BasicStation.WSPingInterval)
+	bsWebServer := basicstationlns.New(bsCtx, gs, basicstationlns.Options{
```
adriansmares, line 249: At this point you can move the `BasicStationConfig` from `pkg/gatewayserver/config.go` to the `basicstationlns` package, and accept it in `New`.

KrishnaIyer: Yeah good suggestion. Added that now.

pkg/gatewayserver/io/basicstationlns/config.go
```diff
@@ -0,0 +1,31 @@
...
+	UseTrafficTLSAddress bool          `name:"use-traffic-tls-address" description:"Use WSS for the traffic address regardless of the TLS setting"`
+	WSPingInterval       time.Duration `name:"ws-ping-interval" description:"Interval to send WS ping messages"`
+	AllowUnauthenticated bool          `name:"allow-unauthenticated" description:"Allow unauthenticated connections"`
+}
+
+// DefaultConfig contains the default configuration.
+var DefaultConfig = Config{
+	UseTrafficTLSAddress: false,
+	WSPingInterval:       30 * time.Second,
+	AllowUnauthenticated: true,
```
htdvisser, line 30: I don't like this. We should be secure by default.
I think security is important enough to break existing deployments that allowed unauthenticated gateways by default.

cc: @johanstokking

johanstokking: I agree

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -54,7 +54,8 @@ var (
 		"listener",
 		"failed to serve Basic Station frontend listener",
 	)
-	errGatewayID = errors.DefineInvalidArgument("invalid_gateway_id", "invalid gateway id `{id}`")
+	errGatewayID       = errors.DefineInvalidArgument("invalid_gateway_id", "invalid gateway id `{id}`")
+	errAuthNotProvided = errors.DefineNotFound("auth_not_provided", "no auth provided for gateway id `{id}`")
```
johanstokking, line 58: `DefineUnauthenticated()`

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -225,15 +227,20 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 		ctx = frequencyplans.WithFallbackID(ctx, fallback)
 	}
 
-	// For gateways with valid EUIs and no auth, we provide the link rights ourselves as in the udp frontend.
 	if auth == "" {
-		ctx = rights.NewContext(ctx, rights.Rights{
-			GatewayRights: map[string]*ttnpb.Rights{
-				uid: {
-					Rights: []ttnpb.Right{ttnpb.RIGHT_GATEWAY_LINK},
+		// If the server allows unauthenticated connections (for local testing), we provide the link rights ourselves as in the udp frontend.
```
johanstokking, line 231: Don't reference another frontend here

pkg/gatewayserver/io/basicstationlns/config.go
KrishnaIyer: Ok but we don't do this for `gs.require-registered-gateways`. That's false by default and that's a security issue as well.

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
KrishnaIyer: Ok but this was there already; https://github.com/TheThingsNetwork/lorawan-stack/blob/a0bec9d2769925b4b44380c343c9e158ef33ae23/pkg/gatewayserver/io/basicstationlns/basicstationlns.go#L228

CHANGELOG.md
```diff
@@ -50,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Read only access for the gateway overview page in the Console.
 - Fix an issue that frequently caused event data views crashing in the Console.
 - Application Server contacting Join Server via interop for fetching the AppSKey.
+- Basic Station always allows unauthenticated connections when API Key is defined.
```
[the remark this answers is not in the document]
KrishnaIyer: We technically added something that fixed something. I'll try to make this clear in the `Added` section.

pkg/gatewayserver/io/basicstationlns/config.go
KrishnaIyer: Also I wanted this to land in a `backport` where this breaks.

johanstokking:
This is good to go now

PR #2957, opened 2020-07-22
Use minor version in main branches

htdvisser:
#### Summary

This updates our documentation to the new branch naming scheme.

#### Notes for Reviewers

- The issue template for backport releases could be removed, because it's now the same as regular releases (but from a previous minor branch)
- I used `blob/default` for GitHub links (I was surprised that this worked)


README.md
```diff
@@ -1,7 +1,5 @@
 # The Things Stack, an open source LoRaWAN Network Server
 
-[![Build Status](https://travis-ci.com/TheThingsNetwork/lorawan-stack.svg?branch=master)](https://travis-ci.com/TheThingsNetwork/lorawan-stack) [![Coverage Status](https://coveralls.io/repos/github/TheThingsNetwork/lorawan-stack/badge.svg?branch=master)](https://coveralls.io/github/TheThingsNetwork/lorawan-stack?branch=master)
```
htdvisser, line 3: We can't (won't) update this to always point to the latest version, so instead of confusing people, I think it's better to remove these.

PR #2959, opened 2020-07-22
Make CLI command aliases consistent

### LINKED ISSUE #2546: Make aliases of CLI commands consistent
#### Summary

While reviewing #2540 I noticed that we don't consistently alias our CLI commands, so let's start doing that.

#### Why do we need this?

Less confusion.

We typically alias CLI commands so that users don't have to think too much. For instance, we alias "list" to "ls", and "collaborators" to "collaborator", "members", "member".

We're not always consistent though. With `delete` commands we sometimes don't alias it at all, and in some places we use any combination of `del`, `remove`, `rm`.

We need to check the aliases of our CLI commands and make them consistent.

- `create`: `add`, `register` (`generate` for API keys)
- `get`: `info`
- `get-xxxs`: `xxxs`
- `set`: `update` (perhaps also change commands where the canonical name is now `update`)
- `delete`: `del`, `remove`, `rm`
- ...

neoaggelos:
#### Summary

Closes #2546 

#### Changes

- Update command aliases, as proposed in the linked issue.

#### Testing

Verify create/update/delete works locally with existing commands and new aliases

##### Regressions

For some commands, changed `update` to be an alias to `set`. This should not break anything, `update` still works.

#### Notes for Reviewers

Not updating documentation, will follow up with a separate PR.

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/applications.go
+++ b/cmd/ttn-lw-cli/commands/applications.go
@@ -186,9 +186,9 @@ var (
 		}),
 	}
 	applicationsUpdateCommand = &cobra.Command{
-		Use:     "update [application-id]",
-		Aliases: []string{"set"},
-		Short:   "Update an application",
+		Use:     "set [application-id]",
+		Aliases: []string{"update"},
+		Short:   "Set properties of an application",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			appID := getApplicationID(cmd.Flags(), args)
 			if appID == nil {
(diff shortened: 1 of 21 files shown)


cmd/ttn-lw-cli/commands/applications.go
```diff
@@ -186,9 +186,9 @@ var (
 		}),
 	}
 	applicationsUpdateCommand = &cobra.Command{
```
htdvisser, line 188: Maybe even better to rename the variables too

Issue #2913, opened 2020-07-15
Pass configuration options to frontend


htdvisser:
#### Summary

We need to find a way to pass backend configuration to the frontend.

#### Why do we need this?

- So that the frontend can give users more information about the actions they perform.
- So that the frontend can disable functionality that isn't enabled on the backend.

#### What is already there? What do you see now?

- We have some configuration in the Identity Server related to user signup (password requirements, invite only, admin approval required).
- We have configuration in the Identity Server related to disabling picture uploads

#### What is missing? What do you want to see?

- The Account frontend needs those configuration options so that it can show those password requirements, require an invitation token, or inform the user that they need to wait for admin approval.
- The Account and Console frontends need to know if picture uploads are enabled

#### How do you propose to implement this?

Option A: Pass such configuration in the HTML template. This works for the Account frontend (because it runs on the Identity Server, but not for the Console, because the Identity Server runs on a different server.

Option B: Add a method to the `ttn.lorawan.v3.Configuration` service to get config options (this has my preference).

#### How do you propose to test this?

Can be unit-tested

#### Can you do this yourself and submit a Pull Request?

Yes, if I have time, but if anyone is interested in working on this, please comment!

KrishnaIyer:
+1 for Option B.

bafonins:
Are we talking about these https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/pkg/webui/template.go#L93-L101 ?

I guess option A is the easiest. However, option B seems to be more appropriate and reliable. It is also quite helpful for mocking stack configuration for e2e tests.

kschiffer:
Agreed. Option B.

htdvisser:
With the API done, I'm handing this over to @kschiffer and/or @bafonins.

PR #2960, opened 2020-07-22
console: Add create user page

### LINKED ISSUE #2905: Create users in the Console
#### Summary

Feature request to create users in the console.

#### Why do we need this?

It's a useful feature and not everyone wants to use the CLI.

No add user button

Add user button

Maybe. @kschiffer want me to try?

benolayinka:
#### Summary

Closes #2905 

#### Changes

- Add a 'add user' form in the console

#### Testing

TBA @kschiffer 

#### Notes for Reviewers

This is a draft to get some feedback from @kschiffer @bafonins 

I don't see anything in DEVELOPMENT.md about writing tests, only running them. Can you guys point me in the right direction, and add it to the docs?

Also blocked on #2913 , for password form validation

#### Checklist


pkg/webui/console/views/admin-user-management-add/index.js
```diff
@@ -0,0 +1,108 @@
...
+
+import sharedMessages from '@ttn-lw/lib/shared-messages'
+import PropTypes from '@ttn-lw/lib/prop-types'
+
+import { createUser } from '@console/store/actions/users'
+import { attachPromise } from '@console/store/actions/lib'
+
+const m = defineMessages({
+  createSuccess: 'User created',
+  createFailure: 'Error creating user',
```
kschiffer:
```suggestion
  createFailure: 'An error occurred and the user could not be created',
```

https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/DESIGN.md#error-messages

pkg/webui/console/views/admin-user-management-add/index.js
```diff
@@ -0,0 +1,108 @@
...
+
+  @bind
+  onSubmitSuccess(response) {
+    const { navigateToList } = this.props
+
+    toast({
+      title: response.ids.user_id,
+      message: m.createSuccess,
+      type: toast.types.SUCCESS,
+    })
```
kschiffer: > In create forms, forward the user either to the newly created record or to the list of records after successful creation; do not show a toast additionally

https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/DESIGN.md#forms

pkg/webui/console/components/user-data-form/index.js
```diff
@@ -149,7 +166,7 @@ class UserForm extends React.Component {
           title={sharedMessages.userId}
           name="ids.user_id"
           component={Input}
-          disabled
+          disabled={!create}
```
kschiffer:
```suggestion
          disabled={!create}
          autofocus={create}
```

> Auto focus the first input when:
> 
> *   The form is a create form

https://github.com/TheThingsNetwork/lorawan-stack/blob/develop/DESIGN.md#forms

pkg/webui/console/components/user-data-form/index.js
```diff
@@ -123,6 +139,7 @@ class UserForm extends React.Component {
 
   render() {
     const {
+      create,
       error,
       initialValues: values,
```
kschiffer, line 168: You need to prepare the blank `initialValues` for the create form, otherwise form validation will not work properly.

Essentially you need to pass an object with empty strings per field at `LL153`. Have a look at other create forms to see what I mean.

pkg/webui/console/views/admin-user-management-add/index.js
```diff
@@ -0,0 +1,108 @@
...
+
+    navigateToList()
+  }
+
+  @bind
```
kschiffer: You only need `@bind` if you reference `this` in the method.

pkg/webui/console/views/admin-user-management-add/index.js
```diff
@@ -0,0 +1,108 @@
...
+  onSubmitFailure(response) {
+    toast({
+      title: m.createFailure,
+      message: response.message,
+      type: toast.types.ERROR,
+    })
```
kschiffer: This makes me realize that the error handling of the user form has been wrong. It should actually put the error into the component state and then use that state to display an error notification at the top of the form. So this is something that the form component takes care of by itself , you don't need any further handling of errors in the view component then.

Can you change this to work this way?

You can see how this is done in `pkg/webui/console/components/collaborator-form/index.js LL104-113, L173`.

pkg/webui/console/views/admin-user-management-add/index.js
```diff
@@ -0,0 +1,108 @@
...
+              create
+              error={null}
+              onSubmit={this.onSubmit}
+              onSubmitSuccess={this.onSubmitSuccess}
+              onSubmitFailure={this.onSubmitFailure}
+              initialValues={{
+                ids: { user_id: '' },
+                name: '',
+                description: '',
+              }}
```
kschiffer: The blank initial values should be the default prop of the user data form instead.

pkg/webui/console/views/admin-user-management-add/index.js
```diff
@@ -0,0 +1,108 @@
...
+
+  render() {
+    return (
+      <Container>
+        <PageTitle title={sharedMessages.userAdd} />
+        <Row>
+          <Col lg={8} md={12}>
+            <UserDataForm
+              create
+              error={null}
```
kschiffer:
```suggestion
```

pkg/webui/console/views/admin-user-management-add/index.js
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  render() {
+    return (
+      <Container>
+        <PageTitle title={sharedMessages.userAdd} />
+        <Row>
+          <Col lg={8} md={12}>
+            <UserDataForm
+              create
```
kschiffer: We usually use this the other way around, so that there is an `update` prop. Please use that and pass the `update` prop in the update view accordingly.

```suggestion
```

pkg/webui/lib/shared-messages.js
```diff
@@ -28,6 +28,7 @@ export default defineMessages({
   addGateway: 'Add gateway',
   addOrganization: 'Add organization',
   addPubsub: 'Add Pub/Sub',
+  addUser: 'Add user',
```
kschiffer: Let's use just `userAdd`
```suggestion
```
Make sure to update any existing references.

kschiffer:
> I don't see anything in DEVELOPMENT.md about writing tests, only running them. Can you guys point me in the right direction, and add it to the docs?

We are currently working on that, the guidelines will be updated soon. Don't worry about it for now.

pkg/webui/console/components/user-data-form/index.js
```diff
@@ -49,6 +58,10 @@ const validationSchema = Yup.object().shape({
     .oneOf(approvalStates)
     .required(sharedMessages.validateRequired),
   description: Yup.string().max(2000, Yup.passValues(sharedMessages.validateTooLong)),
+  password: Yup.string().required(sharedMessages.validateRequired),
```
kschiffer:
```suggestion
  password: Yup.string().min(8, Yup.passValues(sharedMessages.validateTooShort)).required(sharedMessages.validateRequired),
```

To make it match the validation in `oauth/views/create-account.js`

benolayinka: I think it's better not to do this for now until #2913  - this way, you get an error back from the NS explaining the failure based on the config. Until #2913 we don't have a way to make the regex match the config

pkg/webui/console/views/admin-user-management-add/index.js
benolayinka: Done, but it feels weird without a toast because it forwards to the update screen which is nearly identical, with no visual indicator of success.

kschiffer: It shouldn't go to the update screen but to the list of users.

I realized though that this can be problematic as well when there are more than one pages of entries, since you don't immediately see the new entry then. I'll have to think of ways to fix it, but until then this is the way to do it consistently.

pkg/webui/console/views/admin-user-management-add/index.js
benolayinka: Done, but the success handling is weird without the toasts. Can I leave them?

kschiffer: For the update form, yes.

benolayinka:
@kschiffer error messages currently show like this. can we create human friendly error translations?

Screenshot 2020-07-23 at 17 35 56

Also, the state required message doesn't appear if you leave state blank when creating a user - the submit button just doesn't work. Do you know why?

pkg/webui/console/components/user-data-form/index.js
kschiffer: Whatever we do there, it needs to be consistent in all forms that have password fields.

kschiffer:
> [**@kschiffer**](https://github.com/kschiffer) error messages currently show like this. can we create human friendly error translations?

In this particular case, yes, by applying: https://github.com/TheThingsNetwork/lorawan-stack/pull/2960#discussion_r459182966

The length limitation needs to be caught by the validation schema, only then we're able to show a relevant error message next to the relevant field.

> Also, the state required message doesn't appear if you leave state blank when creating a user - the submit button just doesn't work. Do you know why?

Did you try to rebase on latest `v3.9` ? There have been some issues with an old `yup` version.

pkg/webui/console/components/user-data-form/index.js
```diff
@@ -179,29 +223,55 @@ class UserForm extends React.Component {
...
+        {!update && (
+          <Form.Field
+            title={sharedMessages.confirmPassword}
+            component={Input}
+            name="confirmPassword"
+            type="password"
+            autoComplete="new-password"
+            required
+          />
+        )}
```
kschiffer, line 255:
```suggestion
        {!update && (
          <Form.Field
            title={sharedMessages.password}
            component={Input}
            name="password"
            type="password"
            autoComplete="new-password"
            required
          />
          <Form.Field
            title={sharedMessages.confirmPassword}
            component={Input}
            name="confirmPassword"
            type="password"
            autoComplete="new-password"
            required
          />
        )}
```

pkg/webui/console/views/admin-user-management-add/index.js
```diff
@@ -0,0 +1,93 @@
...
+
+const m = defineMessages({
+  createSuccess: 'User created',
+})
+
+@connect(
+  undefined,
+  {
+    createUser: attachPromise(createUser),
+    navigateToUser: userId => push(`/admin/user-management/${userId}`),
```
kschiffer:
```suggestion
    navigateToUser: userId => push(`/admin/user-management`),
```

> It shouldn't go to the update screen but to the list of users.

https://github.com/TheThingsNetwork/lorawan-stack/pull/2960#discussion_r461349222

pkg/webui/console/views/admin-user-management-add/index.js
```diff
@@ -0,0 +1,93 @@
...
+  onSubmitSuccess(response) {
+    const { navigateToUser } = this.props
+    const userId = getUserId(response)
+
+    toast({
+      title: userId,
+      message: m.createSuccess,
+      type: toast.types.SUCCESS,
+    })
+
```
kschiffer:
```suggestion
```
No toasts for create operations, we need to be consistent about that. Seeing the new entry appearing in the list of users is enough feedback. I admit there needs to be some update around that flow, but we need to do this in a streamlined and organized way. For the time being, please stick to the design guidelines and feel free to open issues about anything you find problematic about them.

kschiffer:
Also please check the CI, there are some linting issues in the JS SDK.

pkg/webui/console/views/admin-user-management-add/index.js
benolayinka: Should we redirect to the list then, rather than the update page? Run it locally and see which feels better

edit: saw your other comment

pkg/webui/console/components/user-data-form/index.js
benolayinka: any validation we put here could get out of sync with config. until #2913 is done, i think we have to remove validation from any config controlled password fields. @htdvisser ?

pkg/webui/console/components/user-data-form/index.js
benolayinka: this doesn't work unless I wrap the JSX in fragment tags. Is that better?

kschiffer: Yeah indeed, forgot to add that. Use `<> … </>`

pkg/webui/console/components/user-data-form/index.js
```diff
@@ -179,29 +223,55 @@ class UserForm extends React.Component {
...
           options={approvalStateOptions}
+          required
         />
         <Form.Field
           title={sharedMessages.admin}
           name="admin"
           component={Checkbox}
           label={m.adminLabel}
         />
+        {!update && (
```
kschiffer, line 236: Actually, it would be handy to be able to change passwords of existing users as well, no? I don't know whether the API allows that though, did you check that?

benolayinka: looks like the API supports it, but you have to provide the old password, so it probably needs a slightly different form. i can do a update/forgot password as another PR

benolayinka:
edit: nevermind, just needed default empty string props for these fields

@kschiffer there's still something wrong with the validation - if I try to submit with an empty email or state, the errors don't show - the submit button just mysteriously doesn't work. here's what I get if I submit with a totally empty form (email and state errors missing):

Screenshot 2020-08-03 at 12 23 32

PR #2961, opened 2020-07-22
Auto-completion fixes and configuration for release packages

### LINKED ISSUE #2907: Add auto-complete configuration on release packages (Snap, Brew, Deb, RPM)
#### Summary

Brew and snap have built-in configuration for auto-completion. We should use that.

#### Why do we need this?

Users can simply download from snap or brew and get auto-completion out of the box, instead of having to configure auto-completion themselves.

Auto-complete support and documentation to enable it.

Add configuration for auto-completion to be enabled by default when installing the CLI via snap or brew.

#### Environment

master 3.8.5

Look up documentation and update our build scripts to include the auto completion configuration script.

Test locally 

yes

neoaggelos:
#### Summary

Closes #2907 

#### Changes

- Keep auto-completion scripts in `config/completion/`
- Setup a CI step to regenerate the required auto-completion scripts and verify that they are up to date
- Add bash completion script to the snapcraft package
- Add bash/fish/zsh completion scripts to the brew formula (*not tested*)
- Add bash/fish/zsh completion scripts to deb/rpm packages
- Add bash/fish/zsh completion scripts to archive packages
- Override Cobra bug with fish completion script

#### Testing

- Tested goreleaser snapcraft build and auto-completion is working properly. (NOTE: This needs goreleaser v0.140.1 or newer, previous versions have a bug that make the build fail).
- Test that auto-completion scripts are added on the binary archives.
- Tested debian package
- Tested that CI checks that auto-completion scripts are OK.
- tested homebrew build on Ubuntu

#### Notes for Reviewers

@rvolosatovs: This needs `goreleaser` v0.140.1 or newer, but  `go mod tidy` also updates versions of other dependencies. Also, check new CI step.

#### Checklist


.github/workflows/general.yml
```diff
@@ -45,6 +45,8 @@ jobs:
         key: ${{ runner.os }}-tools-${{ hashFiles('tools/**') }}
     - name: Make Mage
       run: make tools/bin/mage
+    - name: Auto-completion scripts
+      run: tools/bin/mage cli:autocomplete
```
rvolosatovs: I'm fine with this if it runs very quick (i.e. below 10 seconds), if not - I don't think it belongs here

rvolosatovs:
> This needs `goreleaser` v0.140.1 or newer, but `go mod tidy` also updates versions of other dependencies. Also, check new CI step.

You might have to introduce a couple of `replace`s to avoid importing the updated `protobuf` version, but other than that updating is fine.

.github/workflows/general.yml
neoaggelos: This only takes a few seconds.

htdvisser: I'd say we only need this in (snapshot) releases.

.github/workflows/go.yml
```diff
@@ -43,6 +43,8 @@ jobs:
         key: ${{ runner.os }}-tools-${{ hashFiles('tools/**') }}
     - name: Make Mage
       run: make tools/bin/mage
+    - name: Auto-completion scripts
+      run: tools/bin/mage cli:autocomplete
```
htdvisser: I'd say we only need this in (snapshot) releases.

neoaggelos:
> Shouldn't we add this for debian etc. builds (`nfpm`), and maybe even docker containers as well?

Went ahead and added this for deb and rpm packages as well. Also tested a bit more with zsh and fish and made a small change to avoid a bug with the fish script (caused by Cobra).

It does not really make sense to add this to the Docker images.

Please review again, as there a bunch of new changes.

neoaggelos:
> Generally looks good.
> 
> I don't think there's any reason to check in the completion files, we really only need them in the release packages, so maybe just add them to `.gitignore` to avoid having to regenerate them all the time.

Agreed. I added them because it was much easier to test initially, but I suppose this is no longer needed.

htdvisser:
> unable to test the homebrew build because I don't have a Mac

Homebrew also works on Linux

neoaggelos:
Tested with Homebrew as well, ironed out a few rough edges, and fixed commit history.

.github/workflows/general.yml
rvolosatovs: I think it's quite beneficial to have this for Go builds as well

rvolosatovs:
> I don't think there's any reason to check in the completion files, we really only need them in the release packages, so maybe just add them to `.gitignore` to avoid having to regenerate them all the time.I don't think there's any reason to check in the completion files, we really only need them in the release packages, so maybe just add them to .gitignore to avoid having to regenerate them all the time.

We do not need them, but it makes packaging downstream hell of a lot easier.
Talking about distros, which are not "officially supported" - it costs us nothing to have these checked in, but the generation step does need to be a part of some common mage target, which is run on all CLI-related changes.

neoaggelos:
> IMO we should have the generated scripts checked-in and run the generation step on all Go builds.(instead of release builds) 

I also think this is a better way to go. @htdvisser can I get comment on this?

htdvisser:
It's of course a tradeoff. As @rvolosatovs wrote, it makes packaging for other distros a bit easier. On the other hand, it makes development a bit harder, because we have to think about regenerating the files if we make changes to API. I don't have a strong opinion, so it's up to you to decide.

htdvisser:
Do keep in mind that it's extremely easy to add the files in the future if we do need them whereas it's more difficult to remove them, because we don't know who might be relying on them being in the repo.

neoaggelos:
For simplicity's sake, let's keep them out of this PR then. Let's add them as needed, when needed.

Issue #1406, opened 2019-10-01
Rate Limiting


htdvisser:
#### Summary

We have to add rate limiting to our API endpoints.

Transferred from https://github.com/TheThingsIndustries/lorawan-stack/issues/546

#### Why do we need this?

- Security: Prevent abuse of endpoints (such as login)
- Performance: Prevent one entity from taking all resources

#### What is already there? What do you see now?

We currently don't have any rate limits.

#### What is missing? What do you want to see?

1. Rate limit before auth
	- Limit by remote IP address
2. Rate limit after auth
	- Limit by access token ID
	- Limit by target entity ID (if access token is authorized)
	- Different classes for different types of requests
3. Rate limits on streams (streaming RPCs, websockets, MQTT, UDP)

#### How do you propose to implement this?

For the actual implementation we can just use an existing library like https://github.com/throttled/throttled, which already has support for both in-memory and shared Redis record keeping.

The "key" for rate limiting would be `<class>:<limitBy>:<IP/AccessTokenID/TargetEntityID>`. We still need to define the rate limiting classes:

| Class | Limit by                        | Max rate    | Max burst |
| ----- | ------------------------------- | ----------- | --------- |
| token | IP                              | x per hour  | x         |
| login | IP                              | x per hour  | x         |
| ...   | IP/AccessTokenID/TargetEntityID | x per s/m/h | x         |

For (1): when we have #443, valid access tokens are cached. This means that we can put the rate limiting on the func that verifies non-cached access tokens (the token class above). We also need to put IP rate limiting on the username/password login of the OAuth server (the login class above).

For (2) we need to define for each RPC whether we limit by access token ID or by target entity ID, and which rate limiting class should be applied.

| RPC | Class |
| --- | ----- |
| ... | ...   |

htdvisser:
In https://github.com/TheThingsIndustries/lorawan-stack/issues/546 @johanstokking suggested the following classes:

- Tier 1: user ops; login, password reset, search, etc.
- Tier 2: heavy user/automated ops; listing, pagination, any access management changes **, etc.
- Tier 3: light user/automated ops; entity CRUD, downlink queue ops

** including collaborator changes, API key changes, etc, which may need revocation, email notifications, etc

johanstokking:
Network operators must have insight in why the server rejected requests because of rate limits.

htdvisser:
[...]

PR #2986, opened 2020-07-27
Clear GS connection stats registry on graceful exit

### LINKED ISSUE #2906: Gracefully disconnect Gateways before shutting down Gateway Server
#### Summary

Gracefully disconnect Gateways before shutting down Gateway Server

#### Why do we need this?

Properly clear any cached connection stats of connected gateways when the Gateway Server is updated or restarted. It is also good to avoid leaving hung connections.

The GS shuts down without taking any care of existing gateway connections

GS to properly close existing Gateway Connections before exiting.

#### Environment

3.8.5

Catch exit signal on the GS, do some cleanup. We need to see how that affects shutdown times for a large number of gateways (~ 10K). It might be better to simply clear the connection stats in the cache registry instead.

Test locally, adding unit tests should also not be too hard.

yes. Adding to Next Up, but this is useful for upgrades, so increase priority if needed.

neoaggelos:
#### Summary

Closes #2906

#### Changes

- Keep track of the Redis keys we set
- Clear all keys when Gateway Server is gracefully shutting down

#### Testing

- Added unit test for new `ClearAll` registry interface function.
- Tested locally with simulated gateways. I want to test a bit more.

##### Regressions

This should not affect the normal operation, this is only running a clean action after GS context expires.

#### Notes for Reviewers

As noted in the linked issue, trying to terminate a large number connections in the short time window where the GS is shutting down might not work very well. This PR simply clears the connection stats instead.

Storing the keys internally in the registry for two main reasons:

- No need to compute N Redis keys when shutting down.
- Not really a fan of using the `gs.connections` map for ranging over the GS active connections while the server is shutting down.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -290,6 +290,12 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 			}, component.TaskRestartOnFailure)
 	}
 
+	if gs.statsRegistry != nil {
+		go func() {
+			<-gs.Context().Done()
+			gs.statsRegistry.ClearAll()
+		}()
+	}
 	return gs, nil
 }
 
--- a/pkg/gatewayserver/redis/registry.go
+++ b/pkg/gatewayserver/redis/registry.go
@@ -17,6 +17,7 @@ package redis
 import (
 	"context"
 	"runtime/trace"
+	"sync"
 
 	ttnredis "go.thethings.network/lorawan-stack/v3/pkg/redis"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
@@ -25,7 +26,8 @@ import (
 
 // GatewayConnectionStatsRegistry implements the GatewayConnectionStatsRegistry interface.
 type GatewayConnectionStatsRegistry struct {
-	Redis *ttnredis.Client
+	Redis   *ttnredis.Client
+	allKeys sync.Map // string to struct{}
 }
 
 func (r *GatewayConnectionStatsRegistry) key(uid string) string {
@@ -34,15 +36,16 @@ func (r *GatewayConnectionStatsRegistry) key(uid string) string {
 
 // Set sets or clears the connection stats for a gateway.
 func (r *GatewayConnectionStatsRegistry) Set(ctx context.Context, ids ttnpb.GatewayIdentifiers, stats *ttnpb.GatewayConnectionStats) error {
-	uid := unique.ID(ctx, ids)
-
+	key := r.key(unique.ID(ctx, ids))
 	defer trace.StartRegion(ctx, "set gateway connection stats").End()
 
 	var err error
 	if stats == nil {
-		err = r.Redis.Del(r.key(uid)).Err()
+		err = r.Redis.Del(key).Err()
+		r.allKeys.Delete(key)
 	} else {
-		_, err = ttnredis.SetProto(r.Redis, r.key(uid), stats, 0)
+		_, err = ttnredis.SetProto(r.Redis, key, stats, 0)
+		r.allKeys.Store(key, struct{}{})
 	}
 
 	if err != nil {
@@ -59,4 +62,19 @@ func (r *GatewayConnectionStatsRegistry) Get(ctx context.Context, ids ttnpb.Gate
 		return nil, ttnredis.ConvertError(err)
 	}
 	return result, nil
+}
+
+// ClearAll deletes connection stats for all gateways set from this registry instance.
+func (r *GatewayConnectionStatsRegistry) ClearAll() error {
+	keys := []string{}
+	r.allKeys.Range(func(_key, _ interface{}) bool {
+		if key, ok := _key.(string); ok {
+			keys = append(keys, key)
+		}
+		return true
+	})
+	if err := r.Redis.Del(keys...).Err(); err != nil {
+		return ttnredis.ConvertError(err)
+	}
+	return nil
 }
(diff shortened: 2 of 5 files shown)


neoaggelos:
`compat/db` tag was added automatically (I assume because of the Registry changes), but it does not actually hurt backwards-compatibility.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -290,6 +290,12 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 			}, component.TaskRestartOnFailure)
 	}
 
+	if gs.statsRegistry != nil {
+		go func() {
+			<-gs.Context().Done()
+			gs.statsRegistry.ClearAll()
```
johanstokking, line 296: This call is not awaited, i.e. not wait group that is signaled when this is done.

It feels like we need a more general component shutdown behavior that is actually being awaited.

pkg/gatewayserver/redis/registry.go
```diff
@@ -34,15 +36,16 @@ func (r *GatewayConnectionStatsRegistry) key(uid string) string {
...
 
 	var err error
 	if stats == nil {
-		err = r.Redis.Del(r.key(uid)).Err()
+		err = r.Redis.Del(key).Err()
+		r.allKeys.Delete(key)
 	} else {
-		_, err = ttnredis.SetProto(r.Redis, r.key(uid), stats, 0)
+		_, err = ttnredis.SetProto(r.Redis, key, stats, 0)
+		r.allKeys.Store(key, struct{}{})
```
johanstokking, line 48: If we only store, and if the gateway reconnects to a different instance, and the current instance goes down, the gateway connection stats get reset. That shouldn't happen; if the gateway disconnects, it should be cleared from the map.

Also, have you considered alternative ways to keep a unique set of strings?

How about storing them in a set in Redis per GS instance?

neoaggelos: > If we only store, and if the gateway reconnects to a different instance, and the current instance goes down, the gateway connection stats get reset. That shouldn't happen; if the gateway disconnects, it should be cleared from the map.

When the gateway disconnects from this instance, the stats will be cleared anyway (while the `io.Connection` is shutting down) which would also remove the key from the map.

> Also, have you considered alternative ways to keep a unique set of strings?
>
> How about storing them in a set in Redis per GS instance?

This does feel like a better solution, and we can change `Set` to update in a transaction, so we would be largedly avoiding the issue above.

If we do have a unique instance id then this could also be a solution. I would like that ID to persist on restarts as well, but that is not really needed.

pkg/gatewayserver/gatewayserver.go
neoaggelos: Ah, indeed. I tried to imitate that with a task that waits until the GS context is cancelled, but it did not really feel right.

Is it worth adding generic support for cleanup tasks in components for this? Would we benefit in other components as well?

An alternative (which could be tricky) is to do the cleanup when the GS is starting instead (before starting any of the frontends). We can use a set in Redis for this, as suggested below. But this would not really work on retired instances.

johanstokking: Another problem with this approach is that if a GS instance crashes, or is not closed properly, we leave dirty database as well.

> An alternative (which could be tricky) is to do the cleanup when the GS is starting instead (before starting any of the frontends)

Cleaning on start is fine, but how do we deal with starting a replicated instance?

---

I think that "connected" only counts if the gateway is really connected to an instance, regardless of gateway stats. I think the real problem is that we mark the gateway "connected" when there are stats, while GS knows exactly which gateways are connected.

I think the real challenge is on the clustering side. Please file an issue for TTES for this.

---

> But this would not really work on retired instances.

This is not a problem.

--- 

This PR is attempting to close #2906 while that is about disconnecting the gateway connections. That should then automatically delete the stats entry for that connected gateway. So I think for closing #2906 you should just disconnect gateways after closing the listeners and make sure that stats are cleared.

@htdvisser any thoughts on awaiting close behavior in the shared component? We may use tasks for this, and make sure that the task scheduler awaits all running tasks before stopping the component.

pkg/gatewayserver/redis/registry.go
johanstokking: Please see above; we don't have to store the stats in a set; GS should clear the entries on disconnect.

pkg/gatewayserver/gatewayserver.go
neoaggelos: > This PR is attempting to close #2906 while that is about disconnecting the gateway connections. That should then automatically delete the stats entry for that connected gateway. So I think for closing #2906 you should just disconnect gateways after closing the listeners and make sure that stats are cleared.

Does that not take a significant time for a large number of connections?

neoaggelos: I originally commented about the delay on #2906 as well

htdvisser: > @htdvisser any thoughts on awaiting close behavior

Yeah this is a bit annoying about using context cancellation for shutdown.

For the task scheduler we can probably use an [`errgroup`](http://pkg.go.dev/golang.org/x/sync/errgroup) to wait for all tasks to finish before actually exiting.

A year ago I also wrote a proof of concept for a `cancel` func that waits for all registered functions to return before returning from `cancel`. Maybe that's something to consider.

johanstokking: > Does that not take a significant time for a large number of connections?

Yes, it will take some time to gracefully disconnect thousands of gateways and clean up their state, but that is inevitable, right? We get nice gateway disconnect events, the gateways themselves also see their connection close gracefully, etc.

> For the task scheduler we can probably use an [`errgroup`](http://pkg.go.dev/golang.org/x/sync/errgroup) to wait for all tasks to finish before actually exiting.

SGTM

neoaggelos:
Closing, focusing on #1406 for now.

PR #2987, opened 2020-07-28
doc: Improve webhook template instantiation doc

nejraselimovic:
#### Summary
Adding the instruction for using just token value as a header value.

#### Testing
Tested on a webhook template with TTS running locally.

#### Checklist

### DIFF
--- a/doc/content/integrations/webhooks/webhook-templates/instantiation.md
+++ b/doc/content/integrations/webhooks/webhook-templates/instantiation.md
@@ -25,6 +25,19 @@ headers:
 
 If the user has filled in the value of `token` with `Zpdc7jWMvYzVTeNQ`, then the resulting webhook will contain a header named `Authorization` with the value `Bearer Zpdc7jWMvYzVTeNQ`.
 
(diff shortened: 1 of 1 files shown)


doc/content/integrations/webhooks/webhook-templates/instantiation.md
```diff
@@ -25,6 +25,12 @@ headers:
 
 If the user has filled in the value of `token` with `Zpdc7jWMvYzVTeNQ`, then the resulting webhook will contain a header named `Authorization` with the value `Bearer Zpdc7jWMvYzVTeNQ`.
 
+If the header value needs to contain just `token`, then define the header as:
```
adriansmares: I'd use:
`Keep in mind that if you need to use a field directly as header value, you should wrap it with hyphens as follows:`

nejraselimovic: Maybe I should include today's issue with empty headers array, if it turns out that it's just the matter of formatting the yaml file

adriansmares (approved):

PR #2988, opened 2020-07-28
Update Material Icons

htdvisser:
#### Summary

This updates the material icons and adds other variants (eot, woff, ttf, svg)

#### Changes

- Updated font files
- Updated CSS

#### Testing

Monkey testing

##### Regressions

Don't know what those `#iefix` and `#MaterialIcons` suffixes do for `eot` and `svg`, so not sure if those are correct.

#### Checklist


kschiffer:
We use a modified version of the material icons with one added icon (possibly more in the future). If we want to reference other file formats, we have to source those from the original `.woff2`. However, I honestly don't think this is necessary, and it is also not a fix for TheThingsIndustries/lorawan-stack-support#98

johanstokking:
I think multiple formats (as in this PR) is a good idea anyway. Also I would strongly suggest to keep things as standard as possible. Which is the single custom icon that needs us to build and maintain our own font?

kschiffer:
It's this one:



It seemed easiest to me to modify the font to include this since it's basically a modification of an existing icon of the font.

It's likely that we will add more specific icons in the future. I agree it's not good to modify the original font. What we could do instead is extracting such icons into our own custom font. But until we do that, we cannot merge this PR, so I would suggest to close and open an issue about this. I think that's fine since the changes in this PR are not urgent.

EDIT: #3010

kschiffer (commented):
Suggesting to close this. See my other comment.

johanstokking:
What #3010 doesn't cover is serving the fonts in different formats, which is the purpose of this pull request, and I believe is a good idea doing.

kschiffer:
That's true but then we need to source the different formats from the original woff2 and keept that one in place. If we merge the PR like this, we lose the additional icon.

johanstokking:
I'm fine closing this, but please file an issue to offer multiple sources for icons.

---

I would still strongly prefer spending time on customer value, which in my opinion is not maintaining our own font.

What is the use case of this icon? Sorting? Can we use `unfold_more` instead?

kschiffer:
> What is the use case of this icon? Sorting? Can we use `unfold_more` instead?

Indeed, it's the indicator for an unsorted table column. `unfold_more` does not match the other two icons and it's too large.

@htdvisser please apply this diff and then it can be merged. I'll look into improving this later. 

```diff
--- a/pkg/webui/components/icon/index.js
+++ b/pkg/webui/components/icon/index.js
@@ -25,6 +25,7 @@ const hardcoded = {
   settings: 'tune',
   integration: 'call_merge',
   data: 'poll',
+  sort_order: 'unfold_more',
   sort_order_asc: 'arrow_drop_down',
   sort_order_desc: 'arrow_drop_up',
   overview: 'dashboard',
```

htdvisser:
Done

kschiffer (approved):

Issue #2950, opened 2020-07-21
Events Streams Blocking Console When Start Events Not Sent


htdvisser:
#### Summary

Since https://github.com/TheThingsNetwork/lorawan-stack/pull/2565 the events streams in the console block after clicking around in the console.

#### Steps to Reproduce

1. Go to https://tti.staging1.cloud.thethings.industries/console/gateways
2. In a single tab, open a gateway and use the breadcrumbs to go back
3. Repeat for each gateway

#### What do you see now?

At some point the console keeps loading. This seems to be caused by "too many open connections" because `/api/v3/events` requests aren't closed.

#### What do you want to see instead?

The console should stay responsive.

#### Environment

Build based on the current `master` branch 3df51cd750f57b37c3acffc28b417441babdbf30 (don't pay attention to the 3.8.5 that is shown by the console)

#### How do you propose to implement this?

I suspect this is caused by the fact that headers are only written right before the first message on the stream. If there's no message, there's no header and the browser keeps waiting.

So to investigate, I suggest to:

- Check if we write the header correctly
- Check if gRPC gateway writes and flushes the header correctly
- Check if the Echo/Mux stuff doesn't interfere

#### How do you propose to test this?

Try the reproduction steps

rvolosatovs:
On Firefox console is loading forever if there are 6 open tabs - if e.g. one of the working tabs is refreshed - one of the loading tabs unblocks, but that seems to fail sometimes.

In any case, I tried sending the start/end messages, but that did not change anything - it still hangs after there are 6 tabs.

Note: all tabs are at data view of a gateway (2 different ones)

htdvisser:
@kschiffer any ideas?

kschiffer:
> On Firefox console is loading forever if there are 6 open tabs - if e.g. one of the working tabs is refreshed - one of the loading tabs unblocks, but that seems to fail sometimes.

That is expected and [a limitation of not using HTTP/2](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events):

[...]

PR #2989, opened 2020-07-28
Send event stream start events

rvolosatovs:
#### Summary

References #2950 

#### Changes

- Reintroduce `events.stream.start`

#### Testing

Ran console

##### Regressions

Console-related

#### Notes for Reviewers

Please DO NOT depend on `events.stream.start` - console must completely ignore this event, the only purpose of it is to ensure that headers are sent.
As part of #2950 I will later investigate how we could fix this without relying on such workarounds. ( hence, also, the issue is not closed, but merely referenced)
Assigning @johanstokking since I'm almost done for today and off tomorrow

@kschiffer these entries should not be present:
2020-07-28-15:53:05-screenshot

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Event stream start and stop events are not published anymore.
 - Remove version from hosted documentation paths.
 - Gateway connection stats are now stored in a single key.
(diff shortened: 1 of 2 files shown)


htdvisser (approved):

htdvisser (approved):
LGTM. Please remove the line "Event stream start and stop events are not published anymore." from CHANGELOG instead of adding a line that they are sent.

johanstokking (approved):
I removed the change from the changelog as it's effectively reverted when it's not released yet

PR #2992, opened 2020-07-28
console: Add warning to frequency_plan_id field

### LINKED ISSUE #2805: Warn users about empty `frequency_plan_id` in the Console
#### Summary

After https://github.com/TheThingsNetwork/lorawan-stack/pull/2785 we allow creating/updating gateways without `frequency_plan_id` in the Console. While leaving empty `frequency_plan_id` is allowed we should warn the users about consequences.

#### Why do we need this?

Make users aware of limitations when creating/updating gateways without `frequency_plan_id`

The `frequency_plan_id` field which is not required:

Screenshot 2020-06-28 at 20 32 50

@KrishnaIyer what are general limitations when creating gateways without `frequency_plan_id` ?

I think the easiest is to show a warning that includes field description and the limitations. Once a frequency plan is selected we can remove the warning and show a regular description.

Check the `frequency_plan_id` field w/ and w/o value set.

yes

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2805

Screenshot 2020-07-28 at 18 39 25

#### Changes

- Show warning when frequency plan is not selected

#### Testing

Check gateway creation form and general settings page

##### Regressions

`<FetchSelect />` component

#### Checklist

### DIFF
--- a/pkg/webui/console/components/gateway-data-form/index.js
+++ b/pkg/webui/console/components/gateway-data-form/index.js
@@ -63,6 +63,8 @@ const m = defineMessages({
     'Configure gateway delay (minimum: {minimumValue}ms, default: {defaultValue}ms)',
   delayWarning:
     'Delay too short. The lower bound ({minimumValue}ms) will be used by the Gateway Server.',
+  frequencyPlanWarning:
+    'Without choosing a frequency plan, packets from the gateway will not be correctly processed',
 })
 
 const validationSchema = Yup.object().shape({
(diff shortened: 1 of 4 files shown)


KrishnaIyer:
Small Nit pick; I'd recommend `packets` over `the packets` as that's grammatically correct. 

Yes sorry this was how I wrote it in the issue so it's my fault.. :)

kschiffer (changes requested):
We shouldn't present blank forms that already show a warning and it's also not convention to allow `Select…` as valid input. Instead, it needs to be an explicit "opt-in". So we need to make `Do not use a frequency plan` a default option in the `<Select />`, which will then also trigger the warning. We can then also keep the `required` prop of the field.

PR #2993, opened 2020-07-29
Add frequency_plan_id fallback to Gateway search

htdvisser:
#### Summary

This is a quickfix for an issue that resulted in warnings about the frequency plan ID field when searching for gateways.

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

Issue #1353, opened 2019-09-12
Support multiple frequency plans in devices


rvolosatovs:
#### Summary

NS shall allow multiple frequency plans to specified for a device.

#### Why do we need this?

See https://github.com/TheThingsNetwork/lorawan-stack/issues/761

#### What is already there? What do you see now?

Exactly 1, mandatory frequency plan can be specified

#### What is missing? What do you want to see?

Support for multiple frequency plans, at least 1.
On creation/update NS shall validate frequency plan compatibility according to https://github.com/TheThingsNetwork/lorawan-stack/issues/761#issuecomment-530150946

#### How do you propose to implement this?

Adapt band-dependent functionality to use the lowest common denominator between bands, such that behavior is compatible in all *n* bands
Deprecate `frequency_plan_id`, add `repeated string frequency_plan_ids` instead.
Validate in registry SET operation.

#### Can you do this yourself and submit a Pull Request?

yes

htdvisser:
Also need to consider already registered devices that are stored with a `frequency_plan_id`.

Perhaps it's simpler to just keep `frequency_plan_id` as well as its behavior, and add a new `additional_frequency_plan_ids` that can be appended to the one in `frequency_plan_id`.

rvolosatovs:
I think that's confusing - a simple solution is to just set `frequency_plan_ids` whenever a device with `frequency_plan_id` is read.
Whenever a device with `frequency_plan_id` is written, we can set `frequency_plan_ids` instead.

htdvisser:
I don't think it's that simple actually. Your solution would mean that in the API, `frequency_plan_id` is equivalent to `frequency_plan_ids[0]` (we can't stop accepting/returning `frequency_plan_id` until v4), we'd have to allow setting/getting both, but not at the same time. In stored devices we would have to look whether `frequency_plan_ids` exists, and fall back to `frequency_plan_id` if it doesnt.

htdvisser:
Let's keep this in Backlog until we have more demand (👍's)

PR #2995, opened 2020-07-29
console: Fix field mask for gateways table

### LINKED ISSUE #2994: When clicking All (Admin) on the gateway section I get an unexpected error
Schermafbeelding 2020-07-28 om 16 07 33

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2994

#### Changes

- Change `frequency_plan_id` to `frequency_plan_ids` in field mask

#### Testing

Cannot reproduce the issue when listing gateways as admin 

#### Checklist

### DIFF
--- a/pkg/webui/console/containers/gateways-table/index.js
+++ b/pkg/webui/console/containers/gateways-table/index.js
@@ -116,7 +116,7 @@ class GatewaysTable extends React.Component {
 
       return getGatewaysList(
         params,
-        ['name', 'description', 'frequency_plan_id', 'gateway_server_address'],
+        ['name', 'description', 'frequency_plan_ids', 'gateway_server_address'],
         { withStatus: true, isSearch: tab === ALL_TAB || query.length > 0 },
       )
     }


pkg/webui/console/containers/gateways-table/index.js
```diff
@@ -116,7 +116,7 @@ class GatewaysTable extends React.Component {
 
       return getGatewaysList(
         params,
-        ['name', 'description', 'frequency_plan_id', 'gateway_server_address'],
+        ['name', 'description', 'frequency_plan_ids', 'gateway_server_address'],
```
htdvisser, line 119: I don't think this alone will be sufficient, since other parts of the console still use the deprecated `frequency_plan_id`.

bafonins: Well, this removes the warning described in the issue

htdvisser: Sure, it may remove the warning, but if the table still expects `frequency_plan_id`, you're going to have an empty column.

bafonins: Yes, but the IS returns both `frequency_plan_id` and `frequency_plan_ids`. Currently, everything in the Console is designed around `frequency_plan_id` being a string. The idea of this PR is to remove the warning.

htdvisser: https://github.com/TheThingsNetwork/lorawan-stack/pull/2993 already removed that warning on the server side, so I guess this PR is no longer needed.

kschiffer: So what to do here? How come we both use `id` and `ids` in the first place?

kschiffer (commented):
Should we close this then?

htdvisser:
I think we can close this PR and file a separate issue for adding console support for gateways with multiple frequency plans (and later for end devices refs #1353)

PR #2998, opened 2020-07-29
Make sure RxMetadata location is nil if gateway location is not public

neoaggelos:
#### Summary

Quickfix PR, make sure that `RxMetadata.Location` is nil when gateway location is not public. This came up while refactoring the GS tests.

#### Testing

Unit tests

##### Regressions

`RxMetadata.Location` is injected by the GS and the NS does not depend on it, so no regressions here.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/io.go
+++ b/pkg/gatewayserver/io/io.go
@@ -215,6 +215,8 @@ func (c *Connection) HandleUp(up *ttnpb.UplinkMessage) error {
 			if location.Source != ttnpb.SOURCE_UNKNOWN {
 				md.Location = &location
 			}
+		} else if !c.gateway.LocationPublic {
+			md.Location = nil
 		}
 	}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Authentication metadata missing from published events.
 - Under some circumstances, CLI would mistakenly import ABP devices as OTAA.
+- Gateway Server could include the gateway antenna location on messages forwarded to the Network Server even if the gateway location was not public.
 
 ### Security
 


pkg/gatewayserver/io/io.go
```diff
@@ -214,6 +214,8 @@ func (c *Connection) HandleUp(up *ttnpb.UplinkMessage) error {
 			if location.Source != ttnpb.SOURCE_UNKNOWN {
 				md.Location = &location
 			}
+		} else {
```
KrishnaIyer: There's something funky here.

1. Shouldn't this `if` in `L214` be 
```
location.Source == SOURCE_REGISTRY
```


2. This doesn't actually fix the issue right? If `c.gateway.LocationPublic` is false and there's GPS location from the frontend, that would be forwarded anyway right?

neoaggelos: 1. The check is there to avoid forwarding empty locations (where source == unknown == 0). The source can also `SOURCE_GPS` (or whatever else we may support in the future).

2. It is forwarded without this change.

neoaggelos: 1. The check is there to avoid forwarding empty locations (where source == unknown == 0). The source can also `SOURCE_GPS` (or whatever else we may support in the future).

2. It is forwarded without this change.

KrishnaIyer: 1. So if GPS is the source, you overwrite the GPS value in the MD with the location from registry? How does that work?
2. Then I don't understand this PR. It's ok for forward GPS location but not registry location if the gateway location shouldn't be public?

neoaggelos: So, `grpc` and `mqtt` frontends are receiving Protocol Buffers already, so the `RxMetadata[*].Location` field can be set already. `udp` and `basicstation` frontends do not set that field.

The question basically is: If the gateway location is not public, should we forward the location received by the frontend itself?

Also note that the antenna location in `c.gateway` is not necessarily coming from the registry, since it can be updated dynamically from incoming gateway status messages.

> 1. So if GPS is the source, you overwrite the GPS value in the MD with the location from registry? How does that work?

Indeed, this does not make much sense in hindsight. It is related to the previous question though.

I don't really have a strong opinion on this, and it seems like we should just decide on a sane behaviour and go ahead with it.

johanstokking:
@neoaggelos what exactly is the problem that you're trying to solve? Can you file an issue first?

pkg/gatewayserver/io/io.go
neoaggelos: @KrishnaIyer The documentation states that the `RxMetadata[*].Location` field is injected by the Gateway Server, and not the incoming messages. I also believe it does not make sense to forward that field when a user has set the gateway location to not public.

Can we make a decision here? This is blocking for #2999, and to be honest has been stale for a very long time.

adriansmares: `The question basically is: If the gateway location is not public, should we forward the location received by the frontend itself?` 
No. That would leak the location of the gateway and renders the location hiding useless for GPS gateways.

I also believe the value in the registry, if it exists, should override the one sent by the gateway.

KrishnaIyer: As @johanstokking mentioned, can you please create an issue for this and let's take a decision there?

neoaggelos: See https://github.com/TheThingsNetwork/lorawan-stack/issues/3088

johanstokking: > > The question basically is: If the gateway location is not public, should we forward the location received by the frontend itself?

> No. That would leak the location of the gateway and renders the location hiding useless for GPS gateways.

Agreed

> I also believe the value in the registry, if it exists, should override the one sent by the gateway.

Also agreed

pkg/gatewayserver/io/io.go
```diff
@@ -214,6 +214,8 @@ func (c *Connection) HandleUp(up *ttnpb.UplinkMessage) error {
 			if location.Source != ttnpb.SOURCE_UNKNOWN {
```
johanstokking, line 215: When does this happen? Shouldn't the location be set from the registry if and only if the source is registry?

pkg/gatewayserver/io/io.go
johanstokking: This happens if the location is public, but the antenna index doesn't match, so maybe

```suggestion
		} else if !c.gateway.LocationPublic {
```

pkg/gatewayserver/io/io.go
neoaggelos: When the location is updated automatically from status messages https://github.com/TheThingsNetwork/lorawan-stack/blob/ea45d9c824df5d1ad8595e2663eeea64ab59e870/pkg/gatewayserver/gatewayserver.go#L763-L764

johanstokking (approved):

KrishnaIyer (approved):
This is a security patch IMO and needs to be backported.

PR #3003, opened 2020-07-29
Refactor Secrets API/Model

KrishnaIyer:
#### Summary

Refactor Secrets API/Data Model. Refs #2865 

#### Changes

- Refactor API / Data Model to accommodate review comments.

#### Testing

Will test once merged to #2865 

##### Regressions

Will test once merged to #2865 

#### Notes for Reviewers

- Let's please first agree on the API/Model before I update the entire tests. 
- This will not compile since I'm doing this in a different branch. I'll fix all the issues once this is merged to the original PR branch.
- It's targeting `feature/1979-gateway-secrets`, so it's safe to merge without checks once we agree on the implementation.

#### Checklist


api/gateway.proto
```diff
@@ -291,17 +291,22 @@ message GatewayConnectionStats {
...
-  map<string,bytes> values = 1 [(validate.rules).map.keys.string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$" , max_len: 36}];
+// The plain text value of the Basic Station LNS gateway secret.
+message GatewayLBSLNSSecretPlainText{
+  string token_key = 1;
+}
+
+// The collective secret message use in RPC calls.
+message GatewaySecrets{
+  GatewayLBSLNSSecretPlainText lbs_lns_plain_text = 1 [(gogoproto.customname) = "LBSLNSPlainText"];
 }
```
johanstokking: - `GatewaySecrets` would imply it is already plain text. Maybe clarify by renaming `GatewaySecretValue`
- Either make the secret value fields a `oneof`, or don't declare `GatewaySecrets` at all, and return a `bytes` to the client, and let the client do the unmarshaling to the expected type. That is not type-safe, but gives more flexiblity to the client, in the sense that it could also store JSON buffers

api/gateway.proto
```diff
@@ -291,17 +291,22 @@ message GatewayConnectionStats {
   repeated SubBand sub_bands = 10;
 }
 
-// The plain text value of the gateway secret.
-message GatewaySecretPlainText{
-  map<string,bytes> values = 1 [(validate.rules).map.keys.string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$" , max_len: 36}];
+// The plain text value of the Basic Station LNS gateway secret.
+message GatewayLBSLNSSecretPlainText{
```
johanstokking: Spacing, also other occurrences

api/gateway.proto
```diff
@@ -291,17 +291,22 @@ message GatewayConnectionStats {
...
+  string token_key = 1;
```
johanstokking: Probably needs a length cap to avoid people storing megabytes of `token_key`

api/secrets.proto
```diff
@@ -17,13 +17,13 @@ syntax = "proto3";
...
-import "github.com/envoyproxy/protoc-gen-validate/validate/validate.proto";
 
 option go_package = "go.thethings.network/lorawan-stack/v3/pkg/ttnpb";
 
-// Secrets contains key-value pairs of secret values as well as the ID of the encryption key.
-message Secrets{
-  map<string,bytes> values = 1 [(validate.rules).map.keys.string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$" , max_len: 36}];
+// Secret contains a secret blob as well as the ID of the encryption key.
+// The blob is typically a marshalled proto message.
+message Secret{
```
johanstokking: So how does this relate to the gateway secrets?

johanstokking (commented):
Looked at API only

api/secrets.proto
KrishnaIyer: So this `Secret` contains an encrypted blob with the key ID primarily meant for storing. 

The `GatewaySecrets` is how plaintext is handled by clients and is meant for RPCs.

api/gateway.proto
KrishnaIyer: Our API Keys are 98 characters so I've capped this at 128 characters.

api/gateway.proto
KrishnaIyer: Our API Keys are 98 characters so I've capped this at 128 characters.

api/gateway.proto
KrishnaIyer: Either option would mean that we can only send one secret per call. That's fine right? 
Actually, each secret should be handled by a separate key so this is fine indeed.

pkg/identityserver/store/secret.go
```diff
@@ -0,0 +1,77 @@
...
+func init() {
+	registerModel(&Attribute{})
+}
+
+type secrets []Secret
+
+func (s secrets) toMap() map[string]*ttnpb.Secret {
+	pbSecrets := make(map[string]*ttnpb.Secret, len(s))
+	for _, secret := range s {
+		pbSecret := ttnpb.Secret{}
```
htdvisser, line 41: Nit: I think we would typically use `var ` declaration here

pkg/identityserver/store/secret.go
```diff
@@ -0,0 +1,77 @@
...
+		_ = pbSecret.Unmarshal(secret.CipherData)
```
htdvisser, line 42: I don't think this is the place to ignore errors

api/gateway.proto
```diff
@@ -131,7 +131,7 @@ message Gateway {
   // Encrypted Gateway Secrets.
   // This value cannot be set/retrieved in Gateway API calls. Use the StoreGatewaySecret, RetrieveGatewaySecret requests.
   // It can however be deleted by DeleteGateway since it's part of the entity.
-  Secrets secrets = 22;
+  map<string,Secret> secrets = 22 [(validate.rules).map.keys.string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$" , max_len: 36}];
```
htdvisser, line 134: It feels a bit weird to add this field to to our public API when it can't be read or written.

Do we really need the separate RPCs for adding/retrieving gateway secrets?
Can we not do the rights check on the field mask in the regular Get/Update RPCs?
Can we not use validation to only allow certain keys to be stored?

PR #3013, opened 2020-08-03
Use same CSRF cookie for OAuth app and API

### LINKED ISSUE #2951: CSRF token invalid error when changing password
#### Summary

It is not possible to change pasword via the OAuth app on the staging environment.

#### Steps to Reproduce

1. Go to https://tti.staging1.cloud.thethings.industries/oauth/update-password or build the production version of the lorawan-stack yourself and visit the `/oauth/update-password` route. 
2. Fill in the fields
3. Submit the form and see the error

Screenshot 2020-07-21 at 17 42 01

#### What do you see now?

An error. `CSRF token invalid`

Successful form submission and changed password.

#### Environment

v3.8.7 on `staging1`

Most probably related to changes in https://github.com/TheThingsNetwork/lorawan-stack/pull/2564

https://github.com/TheThingsNetwork/lorawan-stack/pull/2939 and manually changing password.

@kschiffer

kschiffer:
#### Summary
Closes #2951 

#### Changes
- Use common parent path `/` in the CSRF configuration of OAuth app and API
- Set `X-CSRF-Token` header for API requests in the OAuth app (was only used for `/oauth/api` routes before)

#### Testing

Manual.

##### Regressions

This could break CSRF validation in the OAuth app. I tried to test different scenarios but it's always quite hard to anticipate what various already set cookies might cause in trouble. So this needs to be especially verified in the staging environment.

#### Notes for Reviewers
The issue was that we used two different cookie paths `/api` and `/oauth` respectively, resulting in two separate CSRF cookies. Since want to use both for the same CSRF protection (the CSRF tokens exposed on `/oauth` need to be valid in `/api` as well), the cookie setup was wrong and needed to be changed so that both use the same parent path `/` as `path` prop of the cookie. That way, the same cookie can be used for both paths.

Additionally, we also weren't passing the `X-CSRF-Header` to the API requests.

#### Checklist

### DIFF
--- a/pkg/oauth/server.go
+++ b/pkg/oauth/server.go
@@ -220,7 +220,7 @@ func (s *server) RegisterRoutes(server *web.Server) {
 		}),
 	)
 
-	csrfMiddleware := middleware.CSRF("_csrf", s.config.Mount, s.config.CSRFAuthKey)
+	csrfMiddleware := middleware.CSRF("_csrf", "/", s.config.CSRFAuthKey)
 
 	api := root.Group("/api", csrfMiddleware)
 	api.POST("/auth/login", s.Login)
--- a/pkg/web/web.go
+++ b/pkg/web/web.go
@@ -226,7 +226,7 @@ func New(ctx context.Context, opts ...Option) (*Server, error) {
 	apiRouter.Use(
 		mux.MiddlewareFunc(webmiddleware.CookieAuth("_session")),
 		mux.MiddlewareFunc(webmiddleware.Conditional(
-			webmiddleware.CSRF(hashKey, csrf.CookieName("_csrf"), csrf.Path("/api"), csrf.SameSite(csrf.SameSiteStrictMode)),
+			webmiddleware.CSRF(hashKey, csrf.CookieName("_csrf"), csrf.Path("/"), csrf.SameSite(csrf.SameSiteStrictMode)),
 			csrfSkip,
 		)),
 		mux.MiddlewareFunc(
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -67,6 +67,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Application Server pub/sub integrations race condition during shutdown.
 - Console webhook templates empty headers error.
 - Console MQTT URL validation.
+- Password updates not working in the OAuth frontend.
```
htdvisser: This doesn't need to be in changelog, since it fixes something that we broke in this version, not something that was already broken in a previous version.

pkg/oauth/server.go
```diff
@@ -220,7 +220,7 @@ func (s *server) RegisterRoutes(server *web.Server) {
 		}),
 	)
 
-	csrfMiddleware := middleware.CSRF("_csrf", s.config.Mount, s.config.CSRFAuthKey)
+	csrfMiddleware := middleware.CSRF("_csrf", "/", s.config.CSRFAuthKey)
```
htdvisser, line 223: Won't this now conflict with the console?

kschiffer: The console uses `_console_csrf`.

htdvisser (approved):
LGTM. Don't have time to test this, so I trust that you tested this sufficiently.

bafonins (approved):

kschiffer:
> Don't have time to test this, so I trust that you tested this sufficiently.

> This could break CSRF validation in the OAuth app. I tried to test different scenarios but it's always quite hard to anticipate what various already set cookies might cause in trouble. **So this needs to be especially verified in the staging environment.**

PR #3017, opened 2020-08-04
Improve selecting regional parameters versions in the Console

### LINKED ISSUE #2784: Improve selecting regional parameters versions in the Console
#### Summary

See discussion starting at https://github.com/TheThingsNetwork/lorawan-stack/pull/2644#issuecomment-644146615

#### Why do we need this?

More guidance for the users.

Selects for lorawan and regional parameters versions.

Regional parameters version can be aligned with lorawan version.

#### Environment

v3.8.4

Filter `lorawan_phy_version` field based on selected `lorawan_version`

```
| lorawan version  | lorawan phy version  |
|------------------|----------------------|
|       1.0.0      |        1.0.0         |
|       1.0.1      |        1.0.1         |
|       1.0.2      |   1.0.2-a, 1.0.2-b   |
|       1.0.3      |       1.0.3-a        |
|       1.0.4      |         (?)          |
|       1.1.0      |   1.1.0-a, 1.1.0-b   |
```

cc @johanstokking @laurensslats What about `1.0.4` ?

Check different selections of `lorawan_version` in the console (both the create and general settings forms)

yes

bafonins:
Filter lorawan phy version based on lorawan version of the end device

#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2784

<details>
Screenshot 2020-08-04 at 16 26 46

Screenshot 2020-08-04 at 16 27 01

Screenshot 2020-08-04 at 16 27 20

Screenshot 2020-08-04 at 16 27 38

Screenshot 2020-08-04 at 16 28 01

Screenshot 2020-08-04 at 16 29 00

</details>

#### Changes

- Filter `lorawan_phy_version` based on `lorawan_version` of an end device.
- Select `lorawan_phy_version` when only a single option is available.

##### Regressions

Regional parameters version selection.

#### Notes for Reviewers

Not sure if we should do the same in the device general settings page since the user can still set any phy version via cli/api.

#### Checklist

### DIFF
--- a/pkg/webui/console/views/device-add/wizard/network-settings-form/index.js
+++ b/pkg/webui/console/views/device-add/wizard/network-settings-form/index.js
@@ -38,6 +38,32 @@ import {
 
 import validationSchema from './validation-schema'
 
+const lorawanPhyMap = {
+  100: [{ value: '1.0.0', label: 'PHY V1.0' }],
+  101: [{ value: '1.0.1', label: 'PHY V1.0.1' }],
+  102: [
+    { value: '1.0.2-a', label: 'PHY V1.0.2 REV A' },
+    { value: '1.0.2-b', label: 'PHY V1.0.2 REV B' },
+  ],
+  103: [{ value: '1.0.3-a', label: 'PHY V1.0.3 REV A' }],
+  104: [{ value: '1.0.0', label: 'PHY V1.0' }, { value: '1.0.1', label: 'PHY V1.0.1' }],
+  110: [
+    { value: '1.1.0-a', label: 'PHY V1.1 REV A' },
+    { value: '1.1.0-b', label: 'PHY V1.1 REV B' },
+  ],
+  0: LORAWAN_PHY_VERSIONS,
+}
+
+/**
+ * Filters available lorawan phy versions by lorawan version.
+ *
+ * @param {number} version - Lorawan version.
+ * @returns {Array} - A list of matching lorawan phy version options based on `version`.
+ */
+const getLorawanPhyByVersion = version => {
+  return lorawanPhyMap[version] || LORAWAN_PHY_VERSIONS
+}
+
 const defaultFormValues = {
   lorawan_phy_version: '',
   frequency_plan_id: '',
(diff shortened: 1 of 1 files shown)


pkg/webui/console/views/device-add/wizard/network-settings-form/index.js
```diff
@@ -38,6 +38,36 @@ import {
...
+    return [{ value: '1.0.1', label: 'PHY V1.0.1' }]
+  } else if (version === 102) {
+    return [
+      { value: '1.0.2-a', label: 'PHY V1.0.2 REV A' },
+      { value: '1.0.2-b', label: 'PHY V1.0.2 REV B' },
+    ]
+  } else if (version === 103) {
+    return [{ value: '1.0.3-a', label: 'PHY V1.0.3 REV A' }]
+  } else if (version === 104) {
+    return [{ value: '1.0.0', label: 'PHY V1.0' }, { value: '1.0.1', label: 'PHY V1.0.1' }]
```
johanstokking: @rvolosatovs please implement or guide @bafonins extending the API for allowing the RP2 format. See referenced issue for the table.

For this PR, we might need to skip this step (1.0.4), as we don't have a valid PHY version to select from.

---

On a side note; any reason not to use `switch`/`case`?

kschiffer (approved):
This resolves #2784. However, we need to avoid showing select inputs that have only one option, since this is redundant and confusing for the user. We have a similar issue with AppEUI prefixes. I'm fine with merging like this if you could open an issue about avoiding the single-option select fields here.

kschiffer:
> Not sure if we should do the same in the device general settings page since the user can still set any phy version via cli/api.

cc @johanstokking. Should we allow to do that?

johanstokking:
> > Not sure if we should do the same in the device general settings page since the user can still set any phy version via cli/api.
> 
> cc @johanstokking. Should we allow to do that?

Do the same thing in the general settings page. This is client side validation (for now); we'll enforce it later in the NS, so please prepare for it in the Console.

bafonins:
> Do the same thing in the general settings page. This is client side validation (for now); we'll enforce it later in the NS, so please prepare for it in the Console.

I created a separate issue for this https://github.com/TheThingsNetwork/lorawan-stack/issues/3026. I would suggest change handling of `lorawan_phy_version` in the general settings page together with the server side validation. If we do it now, this can confuse the users who has `lorawan_version` and `lorawan_phy_version` selected in any different way than you described in https://github.com/TheThingsNetwork/lorawan-stack/issues/2784#issuecomment-648735723.

Issue #2632, opened 2020-05-29
Document Events API


htdvisser:
#### Summary

I think it would be good to document the events API.

#### Why do we need this ?

Because we expect people to build integrations on top of it. Think gateway uptime monitoring, airtime tracking, audit logging, etc.

#### What is already there? What do you see now?

We define events througout the codebase and we have a powerful API to subscribe to events.

#### What is missing? What do you want to see?

Documentation for the Events API in the API reference. 

#### How do you propose to document this?

- Document the gRPC service similar to how we document the other gRPC services
- Document how it maps to HTTP (with server-sent events)
- Up for discussion: Add generated documentation for the events defined by The Things Stack

johanstokking:
We should also discuss to which extent is covered by the compatibility commitment in terms of whether events are expected and what the payload is.

htdvisser:
Maybe we should annotate API reference with compatibility level for each API?

johanstokking:
Yeah that’s possible. It would be helpful if we have at least to begin with:

- stable event names
- stable proto message per event name

Then, we have a few behavioral things like:

- consistency of input and output events, like receive vs forward/drop events
- deterministic set of events by situation, like entire activation flow, confirmed downlink scheduling with the confirmation, etc
- stable visibility of events; we might need to restrict things as we expand the payload or if we introduce new rights

PR #3018, opened 2020-08-04
Refactor Go API of pkg/events

htdvisser:
#### Summary

This PR refactors the Go API of `pkg/events` in preparation for #2632.

#### Changes

- `events.Define` now takes variadic options (instead of visibility rights) and returns an `events.Builder` interface (instead of a function).
- Updated all places that define events to move visibility into option.
- Updated all places that call events (which are now structs, not funcs) to use a convenience function `NewWithIdentifiersAndData` which has the same signature as the old event functions.
- Changes in the Network Server to work with the new Builders instead of closures.

#### Testing

Existing unit tests should still pass.

##### Regressions

Can't think of anything that could break at runtime, unless I made some stupid mistake in my regex find/replace

#### Notes for Reviewers

This makes a lot of small changes in many places, so let's try to get this merged ASAP to avoid conflicts.

#### Checklist

	- ⚠️ No, this is only preparation, but I think it's acceptable to do this separately.
	- Apparently the events documentation wasn't updated for `visibility` so it's correct again.

### DIFF
--- a/pkg/events/builder.go
+++ b/pkg/events/builder.go
@@ -1,0 +1,93 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package events
+
+import (
+	"context"
+	"time"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+type builder struct {
+	definition *definition
+	options    []Option
+}
+
+func (b *builder) With(options ...Option) Builder {
+	extended := &builder{
+		definition: b.definition,
+	}
+	extended.options = append(extended.options, b.options...)
+	extended.options = append(extended.options, options...)
+	return extended
+}
+
+func (b *builder) New(ctx context.Context, opts ...Option) Event {
+	evt := &event{
+		ctx: ctx,
+		innerEvent: ttnpb.Event{
+			Name:           b.definition.name,
+			Time:           time.Now().UTC(),
+			Origin:         hostname,
+			CorrelationIDs: CorrelationIDsFromContext(ctx),
+		},
+	}
+	for _, opt := range b.options {
+		opt.applyTo(evt)
+	}
+	for _, opt := range opts {
+		opt.applyTo(evt)
+	}
+	return evt
+}
+
+func (b *builder) NewWithIdentifiersAndData(ctx context.Context, ids CombinedIdentifiers, data interface{}) Event {
+	e := local(b.New(ctx))
+	if ids != nil {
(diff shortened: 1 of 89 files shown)


johanstokking (approved):
Looks good in the sample review that I did

@rvolosatovs to check the NS part still

adriansmares (approved):
LGTM but as mentioned above, leaving final word for NS changes to @rvolosatovs .

pkg/events/builder.go
```diff
@@ -0,0 +1,93 @@
...
+func (b *builder) BindData(data interface{}) Builder {
+	return b.With(WithData(data))
+}
+
+// Builder is the interface for building events from definitions.
+type Builder interface {
+	With(opts ...Option) Builder
+	New(ctx context.Context, opts ...Option) Event
+
+	// Convenience function for legacy code. Same as New(ctx, WithIdentifiers(ids), WithData(data)).
```
rvolosatovs, line 77: nit; A nicer way to implement these would be via functions, which would take `Builder` interface as parameter.
Also, these are *methods*.

pkg/events/options.go
```diff
@@ -0,0 +1,95 @@
...
+		for _, ids := range identifiers {
+			e.innerEvent.Identifiers = append(e.innerEvent.Identifiers, ids.EntityIdentifiers())
+		}
+	})
+}
+
+// WithData returns an option that sets the data of the event.
+func WithData(data interface{}) Option {
+	return optionFunc(func(e *event) {
+		e.data = data
```
rvolosatovs, line 49: this should panic if `e.data` is not nil

pkg/events/options.go
```diff
@@ -0,0 +1,95 @@
...
+				e.innerEvent.CorrelationIDs = mergeStrings(e.innerEvent.CorrelationIDs, cids)
+			}
+		}
+	})
+}
+
+// WithVisibility returns an option that sets the visibility of the event.
+func WithVisibility(rights ...ttnpb.Right) Option {
+	return optionFunc(func(e *event) {
+		e.innerEvent.Visibility = ttnpb.RightsFrom(rights...)
```
rvolosatovs, line 63: `panic` if already set

pkg/events/options.go
```diff
@@ -0,0 +1,95 @@
...
+				authentication.TokenID = tokenID
+			}
+		}
+		if md.XForwardedFor != "" {
+			xff := strings.Split(md.XForwardedFor, ",")
+			authentication.RemoteIP = strings.Trim(xff[0], " ")
+		}
+		if authentication.RemoteIP != "" || authentication.TokenID != "" ||
+			authentication.TokenType != "" || authentication.Type != "" {
+			e.innerEvent.Authentication = authentication
```
rvolosatovs, line 92: `panic` if already set

kschiffer (approved):
Code owned files LGTM

pkg/events/options.go
htdvisser: I'm just moving what was already there. The goal here is not to reimplement.

pkg/events/options.go
htdvisser: I'm just moving what was already there. The goal here is not to reimplement.

pkg/events/options.go
htdvisser: I'm just moving what was already there. The goal here is not to reimplement.

pkg/events/options.go
rvolosatovs: That is not true - before `data` could only be specified once, now it's possible to specify it multiple times - you changed the semantics

pkg/events/options.go
rvolosatovs: See above

PR #3029, opened 2020-08-05
Refactor Basic Station front end

KrishnaIyer:
#### Summary

Refs https://github.com/TheThingsIndustries/lorawan-stack/issues/2252 and 
https://github.com/TheThingsIndustries/lorawan-stack/pull/2275

#### Changes

- Refactor BasicStation frontend 
- Make naming consistent by using LBS
- Add formatter interface
- Update tests

@johanstokking is the main reviewer. I can also modify the naming in `pkg/basicstation/cups` and `pkg/gatewayconfigurationserver` if it makes sense to do so.

#### Testing

Unit testing

- [ ] Improve test coverage
- [ ] Test with Gateway

##### Regressions

Don't expect any since there are no functional changes.

#### Checklist

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -630,6 +630,7 @@ go.opencensus.io v0.22.3 h1:8sGtKOrtQqkN1bp2AtX+misvLIlOmsEsNd+9NIcPEm8=
 go.opencensus.io v0.22.3/go.mod h1:yxeiOL68Rb0Xd1ddK5vPZ/oVn4vY4Ynel7k9FzqtOIw=
 go.packetbroker.org/api/v3 v3.0.0 h1:RwpVsmLP9jRILTDBJ6ylo+g9szKxM/zWJle9XqN48bg=
 go.packetbroker.org/api/v3 v3.0.0/go.mod h1:pmw0HryGn2ElU9H+98RQb7jFUvq7feKwPGdmTPPkC2w=
+go.thethings.network/lorawan-stack v3.0.0-pre+incompatible h1:5I4+DbEyYFuWpR0jkF5NDh1HENTUFtNmhu++T4aQBHw=
 go.thethings.network/lorawan-stack-legacy/v2 v2.0.2 h1:AXv7hAtYj0dPocFF6rhWwW5WvOVSm/YNk3yihgNYcXE=
 go.thethings.network/lorawan-stack-legacy/v2 v2.0.2/go.mod h1:6ytuRb3k6/f/lJFvJpOaPS4abTDp2mdcE9EIZZV1v9U=
 go.uber.org/atomic v1.4.0/go.mod h1:gD2HeocX3+yG+ygLZcrzQJaqmWj9AIm7n08wl/qW/PE=
--- a/go.sum
+++ b/go.sum
@@ -630,7 +630,6 @@ go.opencensus.io v0.22.3 h1:8sGtKOrtQqkN1bp2AtX+misvLIlOmsEsNd+9NIcPEm8=
 go.opencensus.io v0.22.3/go.mod h1:yxeiOL68Rb0Xd1ddK5vPZ/oVn4vY4Ynel7k9FzqtOIw=
 go.packetbroker.org/api/v3 v3.0.0 h1:RwpVsmLP9jRILTDBJ6ylo+g9szKxM/zWJle9XqN48bg=
 go.packetbroker.org/api/v3 v3.0.0/go.mod h1:pmw0HryGn2ElU9H+98RQb7jFUvq7feKwPGdmTPPkC2w=
-go.thethings.network/lorawan-stack v3.0.0-pre+incompatible h1:5I4+DbEyYFuWpR0jkF5NDh1HENTUFtNmhu++T4aQBHw=
 go.thethings.network/lorawan-stack-legacy/v2 v2.0.2 h1:AXv7hAtYj0dPocFF6rhWwW5WvOVSm/YNk3yihgNYcXE=
 go.thethings.network/lorawan-stack-legacy/v2 v2.0.2/go.mod h1:6ytuRb3k6/f/lJFvJpOaPS4abTDp2mdcE9EIZZV1v9U=
 go.uber.org/atomic v1.4.0/go.mod h1:gD2HeocX3+yG+ygLZcrzQJaqmWj9AIm7n08wl/qW/PE=
--- a/pkg/gatewayserver/config.go
+++ b/pkg/gatewayserver/config.go
@@ -18,8 +18,8 @@ import (
 	"time"
 
 	"go.thethings.network/lorawan-stack/v3/pkg/config"
-	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io/basicstationlns"
 	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io/udp"
+	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io/ws"
 	"go.thethings.network/lorawan-stack/v3/pkg/types"
 )
 
@@ -31,7 +31,7 @@ type UDPConfig struct {
 
 // BasicStationConfig defines the Basic Station configuration of the Gateway Server.
 type BasicStationConfig struct {
-	basicstationlns.Config  `name:",squash"`
+	ws.Config               `name:",squash"`
 	FallbackFrequencyPlanID string `name:"fallback-frequency-plan-id" description:"Fallback frequency plan ID for non-registered gateways"`
 	Listen                  string `name:"listen" description:"Address for the Basic Station frontend to listen on"`
 	ListenTLS               string `name:"listen-tls" description:"Address for the Basic Station frontend to listen on (with TLS)"`
--- a/pkg/gatewayserver/config.go
+++ b/pkg/gatewayserver/config.go
@@ -29,7 +29,7 @@ type UDPConfig struct {
 	Listeners  map[string]string `name:"listeners" description:"Listen addresses with (optional) fallback frequency plan ID for non-registered gateways"`
 }
 
-// BasicStationConfig defines the Basic Station configuration of the Gateway Server.
+// BasicStationConfig defines the LoRa Basics Station configuration of the Gateway Server.
 type BasicStationConfig struct {
 	ws.Config               `name:",squash"`
 	FallbackFrequencyPlanID string `name:"fallback-frequency-plan-id" description:"Fallback frequency plan ID for non-registered gateways"`
--- a/pkg/gatewayserver/io/ws/format.go
+++ b/pkg/gatewayserver/io/ws/format.go
@@ -29,17 +29,17 @@ type ParsedTime struct {
 	RefTime float64
 }
 
-// Format abstracts messages to/from websocket based gateways.
-type Format interface {
+// Formatter abstracts messages to/from websocket based gateways.
+type Formatter interface {
 	// GetRouterConfig parses version messages, generates router config (for downstream) and a status message (for upstream).
 	GetRouterConfig(ctx context.Context, message []byte, bandID string, fps map[string]*frequencyplans.FrequencyPlan, receivedAt time.Time) (context.Context, []byte, *ttnpb.GatewayStatus, error)
 
 	// FromDownlink generates a downlink byte stream that can be sent over the WS connection.
 	FromDownlink(dids ttnpb.GatewayIdentifiers, rawPayload []byte, scheduledMsg *ttnpb.TxSettings, dlToken int64, dlTime time.Time, xTime int64) ([]byte, error)
 
-	// ToUplink parses Uplink ("updf")/Join Request ("jreq") messages into ttnpb.UplinkMessage.
+	// ToUplink parses Uplink/JoinRequest messages into ttnpb.UplinkMessage.
 	ToUplink(ctx context.Context, raw []byte, ids ttnpb.GatewayIdentifiers, bandID string, receivedAt time.Time, msgType string) (*ttnpb.UplinkMessage, ParsedTime, error)
 
-	// ToTxAck extracts fields from the TxConfirmation ("dntxed") message and converts it to  ttnpb.TxAcknowledgment message.
+	// ToTxAck parses fields from the TxConfirmation message and converts it to  ttnpb.TxAcknowledgment message.
 	ToTxAck(ctx context.Context, message []byte, tokens io.DownlinkTokens, receivedAt time.Time) (*ttnpb.TxAcknowledgment, ParsedTime, error)
 }
--- a/pkg/gatewayserver/io/ws/format.go
+++ b/pkg/gatewayserver/io/ws/format.go
@@ -1,0 +1,45 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ws
+
+import (
+	"context"
+	"time"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/frequencyplans"
+	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+// ParsedTime abstracts time values parsed from messages.
+type ParsedTime struct {
+	XTime   int64
+	RefTime float64
+}
+
+// Format abstracts messages to/from websocket based gateways.
+type Format interface {
+	// GetRouterConfig parses version messages, generates router config (for downstream) and a status message (for upstream).
+	GetRouterConfig(ctx context.Context, message []byte, bandID string, fps map[string]*frequencyplans.FrequencyPlan, receivedAt time.Time) (context.Context, []byte, *ttnpb.GatewayStatus, error)
+
+	// FromDownlink generates a downlink byte stream that can be sent over the WS connection.
+	FromDownlink(dids ttnpb.GatewayIdentifiers, rawPayload []byte, scheduledMsg *ttnpb.TxSettings, dlToken int64, dlTime time.Time, xTime int64) ([]byte, error)
+
+	// ToUplink parses Uplink ("updf")/Join Request ("jreq") messages into ttnpb.UplinkMessage.
+	ToUplink(ctx context.Context, raw []byte, ids ttnpb.GatewayIdentifiers, bandID string, receivedAt time.Time, msgType string) (*ttnpb.UplinkMessage, ParsedTime, error)
+
+	// ToTxAck extracts fields from the TxConfirmation ("dntxed") message and converts it to  ttnpb.TxAcknowledgment message.
+	ToTxAck(ctx context.Context, message []byte, tokens io.DownlinkTokens, receivedAt time.Time) (*ttnpb.TxAcknowledgment, ParsedTime, error)
+}
--- a/pkg/gatewayserver/io/ws/format.go
+++ b/pkg/gatewayserver/io/ws/format.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"go.thethings.network/lorawan-stack/v3/pkg/frequencyplans"
+	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/scheduling"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 )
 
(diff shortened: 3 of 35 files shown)


pkg/gatewayserver/config.go
```diff
@@ -29,9 +29,9 @@ type UDPConfig struct {
 	Listeners  map[string]string `name:"listeners" description:"Listen addresses with (optional) fallback frequency plan ID for non-registered gateways"`
 }
 
-// BasicStationConfig defines the Basic Station configuration of the Gateway Server.
+// BasicStationConfig defines the LoRa Basics Station configuration of the Gateway Server.
```
KrishnaIyer, line 32: I can also change this to LBS but we can't change the json tag for obvious reasons.

pkg/gatewayserver/io/ws/format.go
```diff
@@ -0,0 +1,45 @@
...
+// ParsedTime abstracts time values parsed from messages.
+type ParsedTime struct {
+	XTime   int64
+	RefTime float64
+}
+
+// Formatter abstracts messages to/from websocket based gateways.
+type Formatter interface {
+	// GetRouterConfig parses version messages, generates router config (for downstream) and a status message (for upstream).
+	GetRouterConfig(ctx context.Context, message []byte, bandID string, fps map[string]*frequencyplans.FrequencyPlan, receivedAt time.Time) (context.Context, []byte, *ttnpb.GatewayStatus, error)
```
johanstokking: - "router" is a pretty specific LBS term while any future websockets based protocol will implement this interface
- Can you elaborate on why the context is returned? Would it make sense to have a specific `WithContext()` here? Or at least godoc?

pkg/gatewayserver/io/ws/format.go
```diff
@@ -0,0 +1,45 @@
...
+
+	// FromDownlink generates a downlink byte stream that can be sent over the WS connection.
+	FromDownlink(dids ttnpb.GatewayIdentifiers, rawPayload []byte, scheduledMsg *ttnpb.TxSettings, dlToken int64, dlTime time.Time, xTime int64) ([]byte, error)
```
johanstokking: - `dids` is not an abbreviation, see `CONTRIBUTING.md`
- Some parameters are very LBS specific, like `xTime`. If we separate the formatter from logic that basically maintains the gateway connection, it should be a clear separation. So a solution might be that there's no stateless formatter, but that the protocol implementation has some state?
- `scheduledMsg` doesn't seem like a good name for a `*ttnpb.TxSettings`

pkg/gatewayserver/io/ws/lbslns/upstream.go
```diff
@@ -400,10 +405,56 @@ func (updf *UplinkDataFrame) FromUplinkMessage(up *ttnpb.UplinkMessage, bandID s
...
+
+// ToUplink implements Format.
+func (lbsLNS) ToUplink(ctx context.Context, raw []byte, ids ttnpb.GatewayIdentifiers, bandID string, receivedAt time.Time, msgType string) (*ttnpb.UplinkMessage, ws.ParsedTime, error) {
+	var (
+		up         *ttnpb.UplinkMessage
+		parsedTime ws.ParsedTime
+		err        error
+	)
+	switch msgType {
+	case ws.TypeUpstreamJoinRequest:
```
johanstokking: This implies that websocket protocols have different messages for join request and uplink data frames, while that is also a pretty clear LBS thing. I don't think the websocket protocol handler should care about what sort of upstream message it is.

htdvisser (approved):
That change to my code-owned `config.go` is fine. Please re-request review if you want me to check anything else.

pkg/gatewayserver/io/ws/format.go
adriansmares: The context is used in order to enrich the `log` with version fields (`station` build, `firmware` version, gateway `model`). I think that the intent is good but we should indeed document this.

adriansmares (commented):
I looked over most of the code and don't have special comments yet (LGTM mostly).

I think instead of hypothesizing what 'future' WebSocket based frontends might look like and what they would use/need, it would be better to just consider formatters to be LBS specific. We'll extract the 'common' root between LBS (and it's internal versions) and this new, never seen before WebSocket implementation once it exists.

KrishnaIyer:
This was quite some effort to take your points into account @johanstokking. But now the separation is kinda clear. This PR is quite big so let me know if I can help make it easier to review.

go.sum
```diff
@@ -626,6 +626,7 @@ go.opencensus.io v0.22.3 h1:8sGtKOrtQqkN1bp2AtX+misvLIlOmsEsNd+9NIcPEm8=
 go.opencensus.io v0.22.3/go.mod h1:yxeiOL68Rb0Xd1ddK5vPZ/oVn4vY4Ynel7k9FzqtOIw=
 go.packetbroker.org/api/v3 v3.0.0 h1:RwpVsmLP9jRILTDBJ6ylo+g9szKxM/zWJle9XqN48bg=
 go.packetbroker.org/api/v3 v3.0.0/go.mod h1:pmw0HryGn2ElU9H+98RQb7jFUvq7feKwPGdmTPPkC2w=
+go.thethings.network/lorawan-stack v3.0.0-pre+incompatible h1:5I4+DbEyYFuWpR0jkF5NDh1HENTUFtNmhu++T4aQBHw=
```
johanstokking: Remove

pkg/gatewayserver/io/ws/format.go
```diff
@@ -0,0 +1,48 @@
...
+	Address string
+	Prefix  string
+}
+
+// Formatter abstracts messages to/from websocket based gateways.
+type Formatter interface {
+	// Connect creates necessary session state for a gateway.
+	Connect(ctx context.Context, uid string) error
+	// Disconnect removes the session state for a gateway.
+	Disconnect(ctx context.Context, uid string)
```
johanstokking, line 38: If the formatter implementation is stateful, it's probably better to manage the state via construction (`New() (Formatter, error)`) and closure (`Close() error`).

Otherwise, it's unclear for the caller if `Connect()` can be used concurrently for the same instance of `Formatter` they have, or if there should be new formatters constructed.

pkg/gatewayserver/io/ws/format.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// HandleDiscover handles discovery messages from web socket based gateways.
+	// This function returns a byte stream to be sent as response to the discovery message.
+	HandleDiscover(ctx context.Context, raw []byte, server io.Server, endPoint EndPoint, receivedAt time.Time) []byte
```
johanstokking: Is there any handling here, or is it formatting only?

Can we generalize this a bit more to, for example, a configuration request? Discovery smells a lot like Basic Station

pkg/gatewayserver/io/ws/lbslns/discover.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"go.thethings.network/lorawan-stack/v3/pkg/basicstation"
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io"
+	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io/ws"
+	"go.thethings.network/lorawan-stack/v3/pkg/log"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+var errEmptyGatewayEUI = errors.Define("empty_gateway_eui", "empty gateway EUI")
```
johanstokking:
```suggestion
var errEmptyGatewayEUI = errors.DefineFailedPrecondition("empty_gateway_eui", "empty gateway EUI")
```

pkg/gatewayserver/io/ws/lbslns/downstream.go
```diff
@@ -12,20 +12,24 @@
...
 	"time"
 
 	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/scheduling"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 )
 
-var errDownlinkMessage = errors.Define("downlink_message", "could not translate downlink message")
+var (
+	errDownlinkMessage = errors.Define("downlink_message", "could not translate downlink message")
```
johanstokking: No error messages starting with "could not"; it's an error already

pkg/gatewayserver/io/ws/lbslns/format.go
```diff
@@ -0,0 +1,45 @@
...
+	tokens   io.DownlinkTokens
+	sessions ws.Sessions
+}
+
+// NewFormatter returns a new LoRa Basic Station LNS formatter.
+func NewFormatter() ws.Formatter {
+	return &lbsLNS{}
+}
+
+func (lbsLNS *lbsLNS) Connect(ctx context.Context, uid string) error {
```
johanstokking: nit; receiving name is quite long; `f` for formatter?

pkg/gatewayserver/io/ws/lbslns/format.go
```diff
@@ -0,0 +1,45 @@
...
+	return lbsLNS.sessions.NewSession(ctx, uid)
+}
+
+func (lbsLNS *lbsLNS) Disconnect(ctx context.Context, uid string) {
+	err := lbsLNS.sessions.DeleteSession(uid)
+	if err != nil {
+		logger := log.FromContext(ctx)
+		logger.WithError(err).Warn("Failed to disconnect")
+	}
```
johanstokking, line 48: Return error (implement `io.Closer` as suggsted above?)

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -63,17 +54,17 @@ type srv struct {
...
 	cfg                  Config
+	formatter            Formatter
 }
 
 func (*srv) Protocol() string            { return "basicstation" }
 func (*srv) SupportsDownlinkClaim() bool { return false }
 
-// New creates the Basic Station front end.
-func New(ctx context.Context, server io.Server, cfg Config) *echo.Echo {
+// New creates the LoRa Basics Station front end.
```
johanstokking: Not here

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -63,17 +54,17 @@ type srv struct {
...
+func New(ctx context.Context, server io.Server, formatter Formatter, cfg Config) *echo.Echo {
```
johanstokking, line 63: OK so `Formatter` implementation is a singleton.

I think that the session should be passed via context to the formatter, and not make the formatter implementation manage session lifetime via connect/disconnect.

pkg/gatewayserver/io/ws/sessions.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+// Session contains the session state for a single gateway.
+type Session struct {
+	ID int32
+}
+
+var (
+	errSessionNotFound      = errors.DefineNotFound("session_not_found", "session for uid `{uid}` not found")
+	errSessionAlreadyExists = errors.DefineAlreadyExists("session_already_exists", "matching contact info not found")
```
johanstokking: LOLWUT

pkg/gatewayserver/io/ws/sessions.go
```diff
@@ -0,0 +1,79 @@
...
+// Sessions holds state of the WS sessions.
+type Sessions struct {
+	items sync.Map
+}
+
+// NewSession creates a new session for the given UID.
+// If errSessionAlreadyExists is returned, it indicates a corrupted state or improper session termination.
+func (s *Sessions) NewSession(ctx context.Context, uid string) error {
+	if _, ok := s.items.Load(uid); ok {
+		return errSessionAlreadyExists.WithAttributes("uid", uid)
```
johanstokking: Check that `uid` attribute exists in fixed error message

pkg/gatewayserver/io/ws/sessions.go
```diff
@@ -0,0 +1,79 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+)
+
+// Session contains the session state for a single gateway.
+type Session struct {
+	ID int32
+}
+
+var (
+	errSessionNotFound      = errors.DefineNotFound("session_not_found", "session for uid `{uid}` not found")
```
johanstokking: `uid` is not a user-facing term. You can probably just omit `uid` here because the context is clear from the error namespace, otherwise replace by `gateway`

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -249,23 +211,29 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 		return err
 	}
 
+	// Handle session state
+	err = s.formatter.Connect(ctx, uid)
+	if err != nil {
+		logger.WithError(err).Warn("Failed to handle session")
+		return err
+	}
```
johanstokking: So here basically `WithSession()`, deriving a context with state that can be consumed by the formatter

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -299,217 +267,63 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
...
 
 				logger.Info("Send downlink message")
 				wsWriteMu.Lock()
-				err = ws.WriteMessage(websocket.TextMessage, msg)
+				err = ws.WriteMessage(websocket.TextMessage, dnmsg)
 				wsWriteMu.Unlock()
 				if err != nil {
 					logger.WithError(err).Warn("Failed to send downlink message")
 					conn.Disconnect(err)
+					s.formatter.Disconnect(ctx, uid)
```
johanstokking: This happens via `defer()` already right?

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -299,217 +267,63 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
...
-
 	for {
 		_, data, err := ws.ReadMessage()
 		if err != nil {
 			logger.WithError(err).Debug("Failed to read message")
 			return err
 		}
-
-		typ, err := messages.Type(data)
+		downMsg, err := s.formatter.HandleUp(ctx, data, ids, conn, time.Now())
```
johanstokking: Confusing variable name; is it a response message to the uplink message?

That is downstream, but `downMsg` seems a lot like a downlink message.

johanstokking (changes requested):

pkg/gatewayserver/io/ws/lbslns/downstream.go
```diff
@@ -52,12 +55,31 @@ func (dnmsg DownlinkMessage) MarshalJSON() ([]byte, error) {
...
+	dnmsg.Pdu = hex.EncodeToString(down.GetRawPayload())
+	dnmsg.RCtx = int64(settings.Downlink.AntennaIndex)
+	dnmsg.Diid = int64(f.tokens.Next(down.CorrelationIDs, dlTime))
+
+	// The first 16 bits of XTime gets the session ID from the upstream latestXTime and the other 48 bits are concentrator timestamp accounted for rollover.
+	var (
+		state State
+		ok    bool
+	)
+	if state, ok = ws.SessionFromContext(ctx).State.Load().(State); !ok {
```
johanstokking:
```suggestion
	state, ok := ws.SessionFromContext(ctx).State.Load().(State)
```

pkg/gatewayserver/io/ws/lbslns/downstream_test.go
```diff
@@ -12,24 +12,36 @@
...
+package lbslns
 
 import (
+	"context"
 	"testing"
 	"time"
 
+	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io/ws"
+
+	"github.com/mohae/deepcopy"
```
johanstokking: Formatting

pkg/gatewayserver/io/ws/lbslns/upstream.go
```diff
@@ -400,10 +407,139 @@ func (updf *UplinkDataFrame) FromUplinkMessage(up *ttnpb.UplinkMessage, bandID s
...
+		}
+		if err := conn.HandleUp(up); err != nil {
+			logger.WithError(err).Warn("Failed to handle upstream message")
+			return nil, err
+		}
+		session := ws.SessionFromContext(ctx)
+		lbsState := State{
+			ID: int32(jreq.UpInfo.XTime >> 48),
+		}
+		session.State.Store(deepcopy.Copy(lbsState))
```
johanstokking: What is there to copy?

pkg/gatewayserver/io/ws/sessions.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+type sessionKeyType struct{}
+
+var sessionKey sessionKeyType
+
+// Session contains the session state for a single gateway.
+type Session struct {
+	State atomic.Value
+}
```
johanstokking: Either don't wrap `atomic.Value` in a struct, or declare a `Data interface{}` and let the consumer figure out what to put in there. I prefer the latter

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -0,0 +1,340 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/unique"
+	"go.thethings.network/lorawan-stack/v3/pkg/web"
+	"go.thethings.network/lorawan-stack/v3/pkg/web/middleware"
+	"google.golang.org/grpc/metadata"
+)
+
+var (
+	errListener = errors.DefineFailedPrecondition(
+		"listener",
+		"failed to serve Basic Station frontend listener",
```
johanstokking: Not LBS in this context

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -0,0 +1,340 @@
...
+	)
+	errGatewayID      = errors.DefineInvalidArgument("invalid_gateway_id", "invalid gateway id `{id}`")
```
johanstokking: ID

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -0,0 +1,340 @@
...
+	errNoAuthProvided = errors.DefineUnauthenticated("no_auth_provided", "no auth provided for gateway id `{id}`")
```
johanstokking:
```suggestion
	errNoAuthProvided = errors.DefineUnauthenticated("no_auth_provided", "no auth provided for `{uid}`")
```

Mind `uid` here, it is available below

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -0,0 +1,340 @@
...
+			ctx = rights.NewContext(ctx, rights.Rights{
+				GatewayRights: map[string]*ttnpb.Rights{
+					uid: {
+						Rights: []ttnpb.Right{ttnpb.RIGHT_GATEWAY_LINK},
+					},
+				},
+			})
+		} else {
+			// We error here directly as there is no need make an RPC call to the IS to get a failed rights check due to no Auth.
+			return errNoAuthProvided
```
johanstokking: There's an attribute defined which is not set here

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -0,0 +1,340 @@
...
+	server               io.Server
+	webServer            *echo.Echo
+	upgrader             *websocket.Upgrader
+	useTrafficTLSAddress bool
+	wsPingInterval       time.Duration
+	cfg                  Config
+	formatter            Formatter
+}
+
+func (*srv) Protocol() string            { return "basicstation" }
```
johanstokking: `"ws"` ?

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -0,0 +1,340 @@
...
+func (*srv) SupportsDownlinkClaim() bool { return false }
+
+// New creates a new WebSocket frontend.
+func New(ctx context.Context, server io.Server, formatter Formatter, cfg Config) *echo.Echo {
+	ctx = log.NewContextWithField(ctx, "namespace", "gatewayserver/io/basicstation")
```
johanstokking: `".../io/ws"` ?

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -0,0 +1,340 @@
...
+		ctx:       ctx,
+		server:    server,
+		upgrader:  &websocket.Upgrader{},
+		webServer: webServer,
+		formatter: formatter,
+		cfg:       cfg,
+	}
+
+	webServer.GET("/router-info", s.handleDiscover)
+	webServer.GET("/traffic/:id", s.handleTraffic)
```
johanstokking: This is LBS specific and doesn't belong here

johanstokking (changes requested):
Open ends are still in `pkg/gatewayserver/io/ws/ws.go`; that's very tied to LBS still.

I don't know what the solution is really. Also I start to wonder what the overall added value of this separation is going to be.

pkg/gatewayserver/io/ws/lbslns/downstream.go
KrishnaIyer: `state` is used outside the `if`.

pkg/gatewayserver/io/ws/sessions.go
KrishnaIyer: Ok but this does something similar right? If I use `Data interface{}` then I'll need to add a mutex as well to `Session` since UL/DL paths can write to this concurrently.

johanstokking: I meant that the caller then assumes what type `Data interface{}` is, if the caller uses it from multiple goroutines and needs protection

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -0,0 +1,336 @@
...
+		webServer.Close()
+	}()
+
+	return webServer
+}
+
+func (s *srv) handleConnectionInfo(c echo.Context) error {
+	ctx := c.Request().Context()
+	logger := log.FromContext(ctx).WithFields(log.Fields(
+		"endpoint", "discover",
```
johanstokking: `discover` is a LBS term

pkg/gatewayserver/io/ws/ws_test.go
```diff
@@ -34,11 +34,11 @@ import (
...
 	"go.thethings.network/lorawan-stack/v3/pkg/frequencyplans"
 	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io"
-	. "go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io/basicstationlns"
-	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io/basicstationlns/messages"
 	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io/mock"
+	. "go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io/ws"
+	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io/ws/lbslns"
 	"go.thethings.network/lorawan-stack/v3/pkg/log"
-	pfconfig "go.thethings.network/lorawan-stack/v3/pkg/pfconfig/basicstationlns"
+	pfconfig "go.thethings.network/lorawan-stack/v3/pkg/pfconfig/lbslns"
```
johanstokking, line 41: Is it feasible to not test `ws` with `lbslns`?

KrishnaIyer: That would require generating mock messages in the `lbslns` package and using only a raw byte stream in `ws`. Not sure if it's worth the effort.

johanstokking: OK

johanstokking (approved):

PR #3030, opened 2020-08-05
Fix release actions

htdvisser:
#### Summary

This is a quick fix for the release workflows.

#### Changes

- Trigger "Release Tag" workflow on rc tags
- Move documentation environment to the documentation job

#### Testing

Trial and error I guess

### DIFF
--- a/.github/workflows/release-snapshot.yml
+++ b/.github/workflows/release-snapshot.yml
@@ -79,11 +79,6 @@ jobs:
       run: tools/bin/mage docs:deps
     - name: Build documentation
       run: tools/bin/mage docs:build
-      env:
-        HUGO_BASE_URL: https://thethingsstack.io/
-        HUGO_PARAMS_SEARCH_APIKEY: ${{ secrets.HUGO_PARAMS_SEARCH_APIKEY }}
-        HUGO_PARAMS_SEARCH_ENABLED: true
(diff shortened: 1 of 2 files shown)


johanstokking (approved):
No idea about the env but it's probably for a good reason

htdvisser:
The env was previously only on the jobs that didn't end up on GitHub pages (so basically ineffective)

rvolosatovs (commented):
This is not how this is supposed to work - we have snapshot builds for RCs.
Why was this merged? What is the reasoning behind it?
I do not see any related discussion, issue reference or even a clear answer as to "why" this is done and what other alternatives were considered.

htdvisser:
We build snapshots from "master" builds, which is currently on every commit to the `v3.9` branch.

Release candidates should be exactly like actual releases, except that they're shown as pre-releases on GitHub. Release candidates are intended to test everything, including GitHub Pages, which isn't built by the Snapshot workflow (and didn't actually work, as it turned out).

rvolosatovs:
Next time you have different opinions about how something should work, please voice them during review and not just suddenly decide to change things without discussion.
We already discussed it in https://github.com/TheThingsNetwork/lorawan-stack/pull/2744#discussion_r443411860 and you approved it, which means you explicitly accepted this approach and I do not see a reason why you would suddenly completely ignore that decision.

Issue #1707, opened 2019-12-09
Basic Station Integration: Clock misalignment after burst of uplinks


beitler:
#### Summary
lorawan-stack's gatewayserver truncates `xtime` to 32 bit and tracks the rollovers by detecting backward jumps:

https://github.com/TheThingsNetwork/lorawan-stack/blob/dc8fc5ab888f57c20657729b51760eca98452605/pkg/gatewayserver/scheduling/clock.go#L83

This function is called for each conversion between the internal 32 bit timestamp representation to `xtime`, i.e. of each uplink and downlink. The rollover detection is based on the RolloverClock.relative field which is updated on every uplink:

https://github.com/TheThingsNetwork/lorawan-stack/blob/dc8fc5ab888f57c20657729b51760eca98452605/pkg/gatewayserver/scheduling/clock.go#L49

Now, there is a race condition between `FromTimestampTime` and `Sync` which causes permanent misalignment when called in certain order. Assume the following sequence of events:
1. Uplink 1 is received with xtime=1 --> Sync(timestamp=1)
2. Uplink 2 is received with xtime=2 --> Sync(timestamp=2)
3. Uplink 3 is received with xtime=3 --> Sync(timestamp=3)
4. Downlink is scheduled in response to Uplink 1 with transmission time xtime=1+1=2 --> FromTimestampTime(timestamp=2)

In steps 1.-3. the RolloverClock.relative field progressed to the value 3. In step 4, the FromTimestampTime is called with timestamp=2 which causes a rollover to be added.

#### Steps to Reproduce

This execution order can occur when a burst of uplink messages is sent where at least one of the first uplink messages causes a downlink to be scheduled on the same gateway. uplink message bursts can occur for example, when the back-haul connection experiences intermittend connection faults as observed with cellular or WiFi links.

This scenario can reliably be reproduced by a small change in Basic Station which simulates the case of bursted uplinks: Inside the rxpolling function
https://github.com/lorabasics/basicstation/blob/master/src/ral_lgw.c#L224
move the `s2e_flushRxjobs(&TC->s2ctx)` invocation into the condition where we run out of rx jobs:
https://github.com/lorabasics/basicstation/blob/master/src/ral_lgw.c#L239
With this change, uplink messages will be accumulated in the RX queue and will only be flushed when it is full. For practical purposes reduce the size of the RX queue in s2conf to a reasonably small value.

#### What do you see now?

```
[...]

PR #3036, opened 2020-08-06
Fix gateway concentrator clock roll over

### LINKED ISSUE #2581: Basic Station Integration: Spurious xtime jumps
#### Summary
For background on `xtime` and its handling in lorawan-stack read issue #1709.
Spurious 2**32-jumps of the `xtime` value sent to the gateway in the `dnmsg` message can be observed. This permanently puts gateway and server out of sync. As a consequence, downlinks scheduled based on the wrong xtime value are rejected by the Gateway because they are too far in the future. 

#### Steps to Reproduce

1. Start a gateway process and schedule a first downlink before the first 32 bit rollover.
2. Schedule another downlink some time after the first one, but still before the first 32 bit rollover.

#### What do you see now?
The first downlink has correct xtime. For the second downlink the `xtime` value is 2**32 larger than expected. Here is a concrete example:

```
<CONCENTRATOR START>
  "updf" xtime=00e1000000721200 (0)
...
  "updf" xtime=00e10000049e2b28 (0)
  "updf" xtime=00e100000536c120 (0)
 "dnmsg" xtime=00e100000536c120 (1) < -- first downlink OK
  "updf" xtime=00e1000005cf6b00 (0)
...
  "updf" xtime=00e100002b5c4050 (0)
  "updf" xtime=00e100002bf4d668 (0)
 "dnmsg" xtime=00e100012bf4d668 (1) < -- second downlink xtime offset by 2**32
  "updf" xtime=00e100002c8d6c40 (0)
 "dnmsg" xtime=00e100012c8d6c40 (1) < -- all following downlinks offset by 2**32
  "updf" xtime=00e100002d260220 (0)     /
 "dnmsg" xtime=00e100012d260220 (1) < -´
...
```

Here is the complete station.log:
[station.log](https://github.com/TheThingsNetwork/lorawan-stack/files/4667368/station.log)

The xtime log can be extracted from it like this
```bash
head -n 300 station.log  | sed -n 's/[^{]*\({.*}\).*/\1/p' | jq -cr '[.msgtype, (select(.msgtype == "updf") | .upinfo.xtime), (select(.msgtype == "dnmsg") | .xtime ) ,.RxDelay ] | @csv' | awk -F, '{ printf "%8.8s xtime=%016lx (%1d)\n", $1, $2, $3 }'
```

I suspect the issue is rooted in the tracking of 32 bit rollovers of `xtime`. In this particular case apparently a 32 bit rollover was assumed to have happened some time between `0x00000536c120` and `0x00002bf4d668` (~11 minutes).

The class A downlink `xtime` should not be offset by 2**32 to the corresponding uplink.

#### Environment
Basic Station 2.0.4
TTN community network

johanstokking:
#### Summary

Closes #2581
References #1707

#### Changes

- Synchronize concentrator timestamp when uplink messages arrive out-of-order
- Work-around for Basic Station gateways sending uplink frames with no `xtime` (https://github.com/lorabasics/basicstation/issues/74)

#### Testing

Unit testing, see test case for #2581

##### Regressions

Not expected

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/basicstationlns/basicstationlns.go
+++ b/pkg/gatewayserver/io/basicstationlns/basicstationlns.go
@@ -431,6 +431,11 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 				logger.WithError(err).Debug("Failed to parse join-request message")
 				return err
 			}
+			// TODO: Remove (https://github.com/lorabasics/basicstation/issues/74)
+			if jreq.UpInfo.XTime == 0 {
+				logger.Warn("Received join-request without xtime, drop message")
+				break
+			}
 			recordXTime(jreq.UpInfo.XTime, up.ReceivedAt)
 			if err := conn.HandleUp(up); err != nil {
 				logger.WithError(err).Warn("Failed to handle uplink message")
@@ -447,6 +452,11 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 			if err != nil {
 				logger.WithError(err).Debug("Failed to parse uplink data frame")
 				return err
+			}
+			// TODO: Remove (https://github.com/lorabasics/basicstation/issues/74)
+			if updf.UpInfo.XTime == 0 {
+				logger.Warn("Received uplink data frame without xtime, drop message")
+				break
 			}
 			recordXTime(updf.UpInfo.XTime, up.ReceivedAt)
 			if err := conn.HandleUp(up); err != nil {
--- a/pkg/gatewayserver/io/basicstationlns/basicstationlns.go
+++ b/pkg/gatewayserver/io/basicstationlns/basicstationlns.go
@@ -334,20 +334,16 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 		}
 	}()
 
-	var syncedConcentratorTime bool
 	recordXTime := func(timestamp int64, server time.Time) {
 		// The session is the 16 MSB.
 		atomic.StoreInt32(&sessionID, int32(timestamp>>48))
-		if !syncedConcentratorTime {
-			conn.SyncWithGatewayConcentrator(
-				// The concentrator timestamp is the 32 LSB.
-				uint32(timestamp&0xFFFFFFFF),
-				server,
-				// The Basic Station epoch is the 48 LSB.
-				scheduling.ConcentratorTime(time.Duration(timestamp&0xFFFFFFFFFF)*time.Microsecond),
-			)
-			syncedConcentratorTime = true
-		}
+		conn.SyncWithGatewayConcentrator(
+			// The concentrator timestamp is the 32 LSB.
+			uint32(timestamp&0xFFFFFFFF),
+			server,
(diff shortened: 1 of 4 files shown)


pkg/gatewayserver/scheduling/clock.go
```diff
@@ -59,10 +59,13 @@ func (c *RolloverClock) SyncTime() (time.Time, bool) {
 // Sync synchronizes the clock with the given concentrator timestamp and the server time.
 func (c *RolloverClock) Sync(timestamp uint32, server time.Time) ConcentratorTime {
 	rollovers := int64(c.absolute/ConcentratorTime(time.Microsecond)) >> 32
-	if passed := int64(timestamp) - int64(c.relative); passed < 0 {
+	// If the new timestamp is lower, it is considered a roll over if it advanced more than 25%.
```
KrishnaIyer, line 62: Nit; 
```
// If the new timestamp is lower, it is considered a roll over if the difference is more than 25%.
```

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
```diff
@@ -435,6 +431,11 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 				logger.WithError(err).Debug("Failed to parse join-request message")
 				return err
 			}
+			// TODO: Remove (https://github.com/lorabasics/basicstation/issues/74)
```
adriansmares, line 434: This probably has to stay here even after 74 is going to be solved, since it's basically guaranteed that gateway vendors won't upgrade right away.

adriansmares (approved):

pkg/gatewayserver/io/basicstationlns/basicstationlns.go
johanstokking: Yep, I'm afraid so indeed.

pkg/gatewayserver/scheduling/clock.go
KrishnaIyer: Also isn't 25% a lot? We tolerate a drift of up to 17.9 mins?

johanstokking: > ```
> // If the new timestamp is lower, it is considered a roll over if the difference is more than 25%.
> ```

On a scale of 0-9, going from 8 to 1, is "the difference" 7 or 3? We're specifically testing if the value advanced more than X, so the 3.

johanstokking: > Also isn't 25% a lot? We tolerate a drift of up to 17.9 mins?

It's pretty arbitrary but this is what we ended up with choosing in an offline discussion with @htdvisser

htdvisser: The idea behind that 25% was that it is also the margin for FCnt rollovers in earlier LoRaWAN versions (MAX_FCNT_GAP).

KrishnaIyer: Yeah ok sounds good. A small note would've been useful.

PR #3049, opened 2020-08-07
Refactor component tasks

rvolosatovs:
#### Summary

Refactor component tasks
This is required for:
1. Being able to just use `StartTask` directly in NS without custom wrapping
2. Being able to properly test tasks (usage is coming in upcoming NS testing PR)

#### Changes

- Refactor component tasks and configuration

#### Testing

Integration testing

##### Regressions

Possibly incorrect jitter/backoff values used

#### Checklist

- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI according to the compatibility commitments in `README.md` for the chosen target branch.

### DIFF
--- a/pkg/component/tasks.go
+++ b/pkg/component/tasks.go
@@ -68,8 +68,18 @@ func (c *Component) RegisterTask(ctx context.Context, id string, fn TaskFunc, re
 	})
 }
 
-// StartTask starts the specified task function, optionally with restart policy and backoff.
-func (c *Component) StartTask(ctx context.Context, id string, fn TaskFunc, restart TaskRestart, jitter float64, backoff ...time.Duration) {
+type TaskStarter interface {
+	// StartTask starts the specified task function, optionally with restart policy and backoff.
+	StartTask(context.Context, string, TaskFunc, TaskRestart, float64, ...time.Duration)
+}
+
+type StartTaskFunc func(context.Context, string, TaskFunc, TaskRestart, float64, ...time.Duration)
+
+func (f StartTaskFunc) StartTask(ctx context.Context, id string, fn TaskFunc, restart TaskRestart, jitter float64, backoff ...time.Duration) {
+	f(ctx, id, fn, restart, jitter, backoff...)
+}
+
+func DefaultStartTask(ctx context.Context, id string, fn TaskFunc, restart TaskRestart, jitter float64, backoff ...time.Duration) {
 	logger := log.FromContext(ctx).WithField("task_id", id)
 	if len(backoff) == 0 {
 		backoff = defaultTaskBackoff[:]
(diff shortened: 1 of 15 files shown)


rvolosatovs:
Blocking upcoming NS testing PR (#3052 )

johanstokking (commented):
Code LGTM, but this is not a "quickfix" but needed a (technical debt) issue to close.

> Possibly incorrect jitter/backoff values used

Can you elaborate on this?

pkg/component/tasks.go
```diff
@@ -37,85 +37,123 @@ const (
...
 
+func (c *Component) NewTaskConfig(id string, fn TaskFunc, restart TaskRestart, backoff *TaskBackoffConfig) *TaskConfig {
+	return &TaskConfig{
+		Context: c.Context(),
+		ID:      id,
+		Func:    fn,
+		Restart: restart,
+		Backoff: backoff,
+	}
+}
```
htdvisser: I don't think this adds much

htdvisser (approved):
LGTM

rvolosatovs:
> Code LGTM, but this is not a "quickfix" but needed a (technical debt) issue to close.

This was necessary for NS testing, which is in turn necessary for registry optimizations, which is medium priority.

> > Possibly incorrect jitter/backoff values used
> 
> Can you elaborate on this?

Before packages used plain ("magic") constants when configuring tasks, this does not happen anymore - while refactoring I might have missed some value/set it to some incorrect value. It seemed like the default jitter value was used everywhere though, so doesn't seem likely to happen.

johanstokking:
> > Code LGTM, but this is not a "quickfix" but needed a (technical debt) issue to close.
> 
> This was necessary for NS testing, which is in turn necessary for registry optimizations, which is medium priority.

How is this comment related to my comment? That because of medium priority there's no time to file a technical debt issue? Because that is not the way it works. Priority goes via priority labels only.

johanstokking (approved):

rvolosatovs:
> > > Code LGTM, but this is not a "quickfix" but needed a (technical debt) issue to close.
> > 
> > 
> > This was necessary for NS testing, which is in turn necessary for registry optimizations, which is medium priority.
> 
> How is this comment related to my comment? That because of medium priority there's no time to file a technical debt issue? Because that is not the way it works. Priority goes via priority labels only.

I would need to stop working what I was working on(so make a huge context switch and contribute to making the main branch out-of-date - e.g. the event changes took some time to adapt to, the longer we wait, the more of such issues arise), file an issue (with "discussion" label I suppose?) and wait for a response. "just" filing an issue and immediately working on it is no different from what is done here.
While I agree that this should be have been filed in an issue, I don't see how this could realistically fit in the development process:
1. Refactor registry API as part of #2665 
2. Iteratively arrive at a viable implementation
3. Adapt tests - (until 1. and 2. are done, it makes no sense to do that, since final API is not known)
4. Find out that with existing implementation it is not possible to correctly test downlink due to testing changes (and see clear drawbacks of existing implementation, which e.g. make code more complex and disallow proper usage in NS)

--- Here the actual implementation of the feature is done and the tests are in-progress, component task API being the blocking issue
Your suggestion would be to file an issue about what I'm supposed to do (refactor component task API) and completely stop working on this feature waiting for feedback on the filed issue and find some, possibly merge-conflict-inducing, lower-priority issue to work on?

This does not seem like a viable choice to me - it directly hurts productivity.
Do you disagree?
Is there some other way how you would expect this to happen?

Issue #2047, opened 2020-02-25
Device MAC layer configuration in the console


rvolosatovs:
#### Summary

Device MAC layer should be configurable from the console.
Replaces https://github.com/TheThingsNetwork/lorawan-stack/issues/1378

#### Why do we need this?

- Class B device creation/updating from the console
- Creation/updating of end-devices using non-default MAC layer settings (e.g. The Things Uno)

#### What is already there? What do you see now?

CLI as the only means of configuring device MAC layer

#### What is missing? What do you want to see?

Support for configuring all `MACSettings` fields.

###### High

- All devices
	- [x] `rx2_data_rate_index`
	- [x] `rx2_frequency`
	- [x] `factory_preset_frequencies`

- Class A-specific (aka, all non-multicast devices)
	- [x] `rx1_delay`
	- [x] `rx1_data_rate_offset`
	- [x] `resets_f_cnt`

- Class B-specific
	- [x] `ping_slot_periodicity`

###### Medium

- Class A-specific (aka, all non-multicast devices)
	- [x] `max_duty_cycle`
	- [x] `supports_32_bit_f_cnt`
	- [x] `use_adr`
	- [x] `status_time_periodicity`
	- [x] `status_count_periodicity`

- Class B-specific
	- [x] `ping_slot_data_rate_index`
	- [x] `ping_slot_frequency`
	- [x] `beacon_frequency`

###### Low

- Class A-specific (aka, all non-multicast devices)
	- [x] `adr_margin`
	- [x] `desired_rx1_delay`
	- [x] `desired_rx1_data_rate_offset`
	- [x] `desired_rx2_data_rate_index`
	- [x] `desired_rx2_frequency`
	- [x] `desired_max_duty_cycle`
	- [x] `desired_adr_ack_limit_exponent`
	- [x] `desired_adr_ack_delay_exponent`

- Class B-specific
	- Non-multicast
		- [x] `class_b_timeout`
		- [x] `desired_ping_slot_data_rate_index`
		- [x] `desired_ping_slot_frequency`
		- [x] `desired_beacon_frequency`

- Class C-specific
	- Non-multicast
		- [x] `class_c_timeout`

> NOTE: Some of these may already be configurable (e.g. FCnt stuff), please update the checkboxes accordingly and check that these settings are places in correct places (i.e. it's not available for multicast devices)

[...]

PR #3065, opened 2020-08-10
Allow setting mac settings when creating devices in the Console

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2047

<details>
Device wizard network settings step:

OTAA:

Screenshot 2020-11-03 at 17 10 37

OTAA class B:

Screenshot 2020-11-03 at 17 10 45

ABP:

Screenshot 2020-11-03 at 17 11 06

ABP class B:

Screenshot 2020-11-03 at 17 11 13

Multicast:

Screenshot 2020-11-03 at 17 11 33

Multicast class B:

Screenshot 2020-11-03 at 17 12 01

General settings example:
Screenshot 2020-11-03 at 17 15 01

</details>

#### Changes

- Adjust the `<KeyValueMap />` component to allow using indices as keys
- Extend the network settings step in the device wizard to allow setting mac settings
- Various adjustments for the wizard component

#### Testing

Manual testing for various device types + e2e tests for end device creation.

##### Regressions

Device creation or just the network settings step in the device wizard.

#### Checklist


pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -0,0 +1,194 @@
...
+  rx1DelayDescription: 'Class A RX1 delay in seconds. RX2 delay is RX1 delay + 1 second.',
+  rx1DataRateOffsetTitle: 'RX1 Data Rate Offset',
+  rx2DataDateIndexDescription:
+    'The default RX2 data rate index value the device uses after a reset',
+  rx2DataRateIndexTitle: 'RX2 Data Rate Index',
+  rx2FrequencyTitle: 'RX2 Frequency',
+  rx2FrequencyDescription: 'Frequency for RX2 (Hz)',
+  frequencyPlaceholder: 'e.g. 869525000 for 869,525 MHz',
+  macSettings: 'MAC settings',
+  updateSuccess: 'The MAC settings have been updated successfully',
```
kschiffer:
```suggestion
  updateSuccess: 'The MAC settings updated',
```

pkg/webui/components/key-value-map/entry.js
```diff
@@ -84,12 +89,16 @@ class Entry extends React.Component {
...
   keyPlaceholder: PropTypes.message.isRequired,
   name: PropTypes.string.isRequired,
   onBlur: PropTypes.func.isRequired,
   onChange: PropTypes.func.isRequired,
   onRemoveButtonClick: PropTypes.func.isRequired,
-  value: PropTypes.shape({ key: PropTypes.string, value: PropTypes.string }).isRequired,
+  value: PropTypes.shape({
+    key: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+    value: PropTypes.any,
+  }).isRequired,
```
kschiffer:
```suggestion
  }),
```

In your implementation, the value can be `undefined`.

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -0,0 +1,194 @@
...
+  const handleResetsFCntChange = React.useCallback(evt => {
+    const { checked } = evt.target
+
+    setResetsFCnt(checked)
+  }, [])
+
+  return (
+    <Form.CollapseSection
+      id="mac-settings"
+      title={m.macSettings}
```
kschiffer:
```suggestion
      title={sharedMessages.advancedSettings}
```

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -0,0 +1,194 @@
...
+      initiallyCollapsed={initiallyCollapsed}
+    >
+      {isABP && (
```
kschiffer, line 104:
```suggestion
      {isABP && (
      <Form.SubTitle title={m.macSettings} />
```

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -0,0 +1,194 @@
...
+          warning={resetsFCnt ? m.resetWarning : undefined}
+          name="mac_settings.resets_f_cnt"
+          component={Checkbox}
+        />
+      )}
+      <Form.Field
+        title={m.rx2DataRateIndexTitle}
+        description={m.rx2DataDateIndexDescription}
+        name="mac_settings.rx2_data_rate_index.value"
+        component={Select}
```
kschiffer: Why not `type="number"` and `component={Input}` ?

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -0,0 +1,194 @@
...
+          name="mac_settings.rx1_delay.value"
+          component={Select}
+          options={rx1DelayOptions}
+        />
+      )}
+      {(isABP || isOTAA) && (
+        <Form.Field
+          title={m.rx1DataRateOffsetTitle}
+          name="mac_settings.rx1_data_rate_offset"
+          component={Select}
```
kschiffer: Why not `type="number"` and `component={Input}` ?

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -0,0 +1,194 @@
...
+    <Form.CollapseSection
+      id="mac-settings"
+      title={m.macSettings}
+      initiallyCollapsed={initiallyCollapsed}
+    >
+      {isABP && (
+        <Form.Field
+          title={m.rx1DelayTitle}
+          description={m.rx1DelayDescription}
+          name="mac_settings.rx1_delay.value"
```
kschiffer: Why not `type="number"` and `component={Input}` ?

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -0,0 +1,194 @@
...
+      />
+      <Form.Field
+        type="number"
+        min={100000}
+        step={100}
+        title={m.rx2FrequencyTitle}
+        description={m.rx2FrequencyDescription}
+        placeholder={m.frequencyPlaceholder}
+        name="mac_settings.rx2_frequency"
+        component={Input}
```
kschiffer, line 150: Can we use `encode`/ `decode` here so the user can use the common floating point notation?

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -0,0 +1,194 @@
...
+        addMessage={m.freqAdd}
+        valuePlaceholder={m.frequencyPlaceholder}
+      />
+      {isClassB && (
+        <>
+          <Form.Field
+            title={m.pingSlotPeriodicityTitle}
+            description={m.pingSlotPeriodicityDescription}
+            name="mac_settings.ping_slot_periodicity.value"
+            component={Select}
```
kschiffer, line 167: Why `Select`?

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -0,0 +1,194 @@
...
+          component={Select}
+          options={rx1DelayOptions}
+        />
+      )}
+      {(isABP || isOTAA) && (
+        <Form.Field
+          title={m.rx1DataRateOffsetTitle}
+          name="mac_settings.rx1_data_rate_offset"
+          component={Select}
+          options={dataRateOffsetOptions}
```
kschiffer: Can we add a description here too?

pkg/webui/console/components/mac-settings-section/mac-settings-section.js
```diff
@@ -0,0 +1,199 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
```
kschiffer: I guess this file is a duplicate?

pkg/webui/console/components/mac-settings-section/validation-schema.js
```diff
@@ -0,0 +1,124 @@
...
+        return Yup.object().strip()
+      }
+
+      return Yup.object({
+        value: Yup.number(),
+      })
+    }),
+    rx2_frequency: Yup.number().min(100000, {
+      ...sharedMessages.validateNumberGte,
+      values: { value: 100000 },
```
kschiffer: I guess this is a remainder from before we introduced  `yup.passValues`.

pkg/webui/console/components/mac-settings-section/validation-schema.js
```diff
@@ -0,0 +1,124 @@
...
+        otherwise: schema => schema.strip(),
+      })
+    }),
+    rx1_data_rate_offset: Yup.number().when('$activation_mode', {
+      is: mode => mode === ACTIVATION_MODES.ABP || mode === ACTIVATION_MODES.OTAA,
+      then: schema =>
+        schema.min(0, { ...sharedMessages.validateNumberGte, values: { value: 0 } }).max(7, {
+          ...sharedMessages.validateNumberLte,
+          values: { value: 7 },
+        }),
```
kschiffer: Use `yup.passValues()`

pkg/webui/console/components/mac-settings-section/validation-schema.js
```diff
@@ -0,0 +1,124 @@
...
+        otherwise: schema => schema.strip(),
+      })
+    }),
+    ping_slot_frequency: Yup.number().when('$class_b', {
+      is: true,
+      then: schema =>
+        schema.min(100000, {
+          ...sharedMessages.validateNumberGte,
+          values: { value: 100000 },
+        }),
```
kschiffer: Same.

pkg/webui/console/lib/device-utils.js
```diff
@@ -40,6 +40,12 @@ export const LORAWAN_PHY_VERSIONS = Object.freeze([
   { value: '1.1.0-b', label: 'PHY V1.1 REV B' },
 ])
 
+export const DEVICE_CLASSES = Object.freeze({
```
kschiffer: Nit: Maybe `/console/constants/` is a better place for this?

kschiffer (changes requested):
Seems like the branch misses a `formik` update to `2.1.5`, I'm getting errors otherwise.

There's also some prop type warnings for the select inputs, which use numbers as values, which is not allowed by the prop types. However, I'd suggest to actually use `number`\-type `<Input />`'s for these anyway. As far as I can see, the values are basically a range of numbers, so using a select for this is an anti-pattern, but I might miss something there (also I feel like this might have been discussed already but I don't remember the conclusion).

kschiffer:
Two questions for @rvolosatovs for my understanding:

*   is it possible to change these values after the creation of the end device?
*   Can these options only be set during creation?

rvolosatovs:
> Two questions for @rvolosatovs for my understanding:
> 
>     * is it possible to change these values after the creation of the end device?
yes

>     * Can these options only be set during creation?
no

Most of `mac-settings` options, however, are only effective on device MAC state reset - e.g. OTAA (re-)join, ABP FCnt reset, ABP device creation, ABP ResetInd MAC command

pkg/webui/console/components/mac-settings-section/index.js
bafonins: What else?

pkg/webui/console/components/mac-settings-section/index.js
bafonins: Not sure what you mean

kschiffer: `869,525` MHz frequency needs to be entered as `869525000` Hz. I suspect that it's more intuitive for the user to be able to enter the value as MHz. This can be done via the `encode` / `decode` props.

kschiffer: Actually, even better would be to use the `<UnitInput />` component here, with a default on `MHz` and `Hz` second option.

pkg/webui/console/components/mac-settings-section/index.js
kschiffer: I'm not familiar with this setting but in depth, but if the ping slot periodicity allows for a range of numbers, we should rather use the `type="number"` input.

pkg/webui/console/views/device-add/wizard/network-settings-form/index.js
```diff
@@ -103,10 +126,12 @@ const NetworkSettingsForm = props => {
 
   const validationContext = React.useMemo(
     () => ({
+      isClassB,
       activationMode,
     }),
-    [activationMode],
+    [activationMode, isClassB],
   )
```
kschiffer, line 93: I suggest removing `React.useMemo()` here, since this is a really simple function.

pkg/webui/console/views/device-general-settings/network-server-form/index.js
```diff
@@ -59,13 +68,48 @@ const NetworkServerForm = React.memo(props => {
...
+  }, [])
+
+  let activationMode = ACTIVATION_MODES.ABP
+  if (supports_join) {
+    activationMode = ACTIVATION_MODES.OTAA
+  } else if (multicast) {
+    activationMode = ACTIVATION_MODES.MULTICAST
+  }
+
   const validationContext = React.useMemo(
```
kschiffer, line 82: Same here. I don't think the composition of this context object should be of any performance concern.

kschiffer (approved):
LGTM now but I'm not very familiar with the specifics of mac settings, so I suggest waiting with merging until @rvolosatovs has greenlighted this as well.

pkg/webui/components/key-value-map/entry.js
```diff
@@ -31,43 +31,48 @@ class Entry extends React.Component {
   @bind
   handleRemoveButtonClicked(event) {
     const { onRemoveButtonClick, index } = this.props
+
```
rvolosatovs, line 46: nit; this seems like unnecessary waste of screen real estate

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -0,0 +1,185 @@
...
+          title={m.rx1DelayTitle}
+          type="number"
+          description={m.rx1DelayDescription}
+          name="mac_settings.rx1_delay.value"
+          component={Input}
+          min={1}
+          max={15}
+        />
+      )}
+      {(isABP || isOTAA) && (
```
rvolosatovs: All devices are either ABP or OTAA.
This should only be relevant for non-multicast ABP devices

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -0,0 +1,185 @@
...
+  }, [])
+
+  return (
+    <Form.CollapseSection
+      id="mac-settings"
+      title={sharedMessages.advancedSettings}
+      initiallyCollapsed={initiallyCollapsed}
+    >
+      <Form.SubTitle title={m.macSettings} />
+      {isABP && (
```
rvolosatovs, line 104: This should only be relevant for non-multicast devices

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -0,0 +1,185 @@
...
+    label: <Message content={m.pingSlotPeriodicityValue} values={{ count: value }} />,
+  }
+})
+
+const MacSettingsSection = props => {
+  const { activationMode, deviceClasses, resetsFCnt: initialFCnt, initiallyCollapsed } = props
+
+  const isClassB = deviceClasses.includes(DEVICE_CLASSES.CLASS_B)
+  const isOTAA = activationMode === ACTIVATION_MODES.OTAA
+  const isABP = activationMode === ACTIVATION_MODES.ABP
```
rvolosatovs, line 77: There should be a distinction here between non-multicast ABP and multicast ABP

pkg/webui/console/views/device-add/wizard/network-settings-form/index.js
```diff
@@ -93,6 +97,25 @@ const NetworkSettingsForm = props => {
     setResetsFCnt(checked)
   }, [])
 
+  const [deviceClasses, setDeviceClasses] = React.useState([DEVICE_CLASSES.CLASS_A])
+  const handleDeviceClassChange = React.useCallback(evt => {
+    const { checked, name } = evt.target
+
+    let cls
```
rvolosatovs: Exclusion of those 2 characters must have really sped up the development /s
```suggestion
    let class
```

pkg/webui/console/views/device-add/wizard/network-settings-form/index.js
```diff
@@ -93,6 +97,25 @@ const NetworkSettingsForm = props => {
...
+    if (name === 'supports_class_b') {
```
rvolosatovs: Reverse the order here - we always prefer class C.
Note, that by the spec device can't implement both B and C classes simultaneously.

pkg/webui/console/views/device-add/wizard/network-settings-form/index.js
```diff
@@ -93,6 +97,25 @@ const NetworkSettingsForm = props => {
...
+      cls = DEVICE_CLASSES.CLASS_B
+    } else if (name === 'supports_class_c') {
+      cls = DEVICE_CLASSES.CLASS_C
+    }
+
+    if (checked && typeof cls !== 'undefined') {
```
rvolosatovs: Device can only be in once class at a time, and that is dictated by `mac_state.device_class` and set by Network Server. Console cannot know what class the device is in if either `supports_class_c` or `supports_class_b` is `true`. iff `supports_class_c == supports_class_b == false`, console can be sure that device class is always A.

This whole thing seems too complicated for what it does, the "maximum capable" class(that's the only thing you can determine here, really) is just:
```elm
if supports_class_c then
	DEVICE_CLASSES.CLASS_C
else if supports_class_b then
	DEVICE_CLASSES.CLASS_B
else
	DEVICE_CLASSES.CLASS_A
```

pkg/webui/console/views/device-add/wizard/network-settings-form/index.js
```diff
@@ -93,6 +97,25 @@ const NetworkSettingsForm = props => {
...
+      setDeviceClasses(classes => [...classes, cls])
+    } else if (!checked && typeof cls !== 'undefined') {
+      setDeviceClasses(classes => classes.filter(c => c !== cls))
+    }
+  }, [])
+
+  const isClassB = deviceClasses.includes(DEVICE_CLASSES.CLASS_B)
```
rvolosatovs: Isn't there a way to just check the value of `supports_class_b` directly here? Why would `deviceClasses` be necessary at all?

pkg/webui/console/views/device-add/wizard/network-settings-form/index.js
```diff
@@ -143,10 +168,17 @@ const NetworkSettingsForm = props => {
         component={Select}
         options={lorawanPhyVersionOptions}
       />
+      <Form.Field
+        title={sharedMessages.supportsClassB}
+        name="supports_class_b"
+        component={Checkbox}
+        onChange={handleDeviceClassChange}
```
rvolosatovs: The device class does not change when this field value is changed. The "maximum capable class" is changed, but the actual class the device is operating in does not change - it's a dynamic value, which may change during the session, which is fully controlled by device.

pkg/webui/console/views/device-add/wizard/network-settings-form/validation-schema.js
```diff
@@ -17,21 +17,107 @@ import sharedMessages from '@ttn-lw/lib/shared-messages'
...
+            value: Yup.number()
+              .min(1, Yup.passValues(sharedMessages.validateNumberGte))
+              .max(15, Yup.passValues(sharedMessages.validateNumberLte)),
+          }),
+        otherwise: schema => schema.strip(),
       })
-    }
+    }),
+    rx1_data_rate_offset: Yup.number().when('$activationMode', {
+      is: mode => mode === ACTIVATION_MODES.ABP || mode === ACTIVATION_MODES.OTAA,
```
rvolosatovs: only non-multicast

pkg/webui/console/views/device-add/wizard/network-settings-form/validation-schema.js
```diff
@@ -17,21 +17,107 @@ import sharedMessages from '@ttn-lw/lib/shared-messages'
...
-      return schema.shape({
-        resets_f_cnt: Yup.boolean().default(false),
+  mac_settings: Yup.object({
+    rx1_delay: Yup.lazy(delay => {
+      if (!Boolean(delay) || typeof delay.value === 'undefined') {
+        return Yup.object().strip()
+      }
+
+      return Yup.object().when('$activationMode', {
+        is: ACTIVATION_MODES.ABP,
```
rvolosatovs, line 67: only non-multicast

pkg/webui/console/views/device-add/wizard/network-settings-form/validation-schema.js
```diff
@@ -17,21 +17,107 @@ import sharedMessages from '@ttn-lw/lib/shared-messages'
...
+    rx1_data_rate_offset: Yup.number().when('$activationMode', {
+      is: mode => mode === ACTIVATION_MODES.ABP || mode === ACTIVATION_MODES.OTAA,
+      then: schema =>
+        schema
+          .min(0, Yup.passValues(sharedMessages.validateNumberGte))
+          .max(7, Yup.passValues(sharedMessages.validateNumberLte)),
+      otherwise: schema => schema.strip(),
+    }),
+    resets_f_cnt: Yup.boolean().when('$activationMode', {
+      is: ACTIVATION_MODES.ABP,
```
rvolosatovs, line 86: only non-multicast

pkg/webui/console/views/device-general-settings/network-server-form/index.js
```diff
@@ -59,13 +68,48 @@ const NetworkServerForm = React.memo(props => {
...
   const [error, setError] = React.useState('')
-  const [resetsFCnt, setResetsFCnt] = React.useState(
-    (isABP && device.mac_settings && device.mac_settings.resets_f_cnt) || false,
-  )
+
   const [lorawanVersion, setLorawanVersion] = React.useState(
     parseLorawanMacVersion(device.lorawan_version),
   )
 
+  const [deviceClasses, setDeviceClasses] = React.useState(() => {
```
rvolosatovs: I don't think this is necesary

pkg/webui/console/views/device-general-settings/network-server-form/index.js
```diff
@@ -59,13 +68,48 @@ const NetworkServerForm = React.memo(props => {
...
+      deviceClasses.push(DEVICE_CLASSES.CLASS_B)
+    }
+
+    if (supports_class_c) {
+      deviceClasses.push(DEVICE_CLASSES.CLASS_C)
+    }
+
+    return deviceClasses
+  })
+  const handleDeviceClassChange = React.useCallback(evt => {
```
rvolosatovs: IDEM

rvolosatovs (dismissed):

pkg/webui/console/views/device-add/wizard/network-settings-form/index.js
kschiffer: `class` is a keyword in JavaScript.

bafonins: 1. yes
2. `class` is reserved in javascript and can be used only to define a class
3. we already use `cls` in many places throughout the codebase

rvolosatovs: If shadowing is not allowed, then that really sucks!
Anyway, just want to point out that this really stands out and is not obvious for a person with not much context about frontend. I wish there was a clearer way to denote this, but I can't think of any.

pkg/webui/console/components/mac-settings-section/index.js
bafonins: `ping_slot_periodicity` enum values can be mapped to a numeric value (see https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.9/api/api.md#enum-pingslotperiod). but is it really more convenient for the user to specify `0` to set the period to 1s, instead of just choosing `every second` in the select?

Also related https://github.com/TheThingsNetwork/lorawan-stack/issues/2258

kschiffer: Ok yeah, I overlooked how the value is composed and `<Select />` is indeed the way to go here.

bafonins:
Blocked by https://github.com/TheThingsNetwork/lorawan-stack/pull/3199

bafonins:
Blocked by https://github.com/TheThingsNetwork/lorawan-stack/pull/3203

pkg/webui/console/components/mac-settings-section/index.js
rvolosatovs: I don't see any change here

pkg/webui/console/components/mac-settings-section/index.js
rvolosatovs: Still not addressed

bafonins:
Blocked on https://github.com/TheThingsNetwork/lorawan-stack/pull/3258

rvolosatovs (changes requested):
Did not review the code as per your explicit request and only checked the UI.
1. It should be possible to unset values. Currently if I set RX1 delay to `4` I see no way to delete the value - I tried saving an empty form, but that did not work.
2. 
2020-09-30-22:55:39-screenshot
That's not how it works. All devices implement class A. There should be 2 checkboxes here - "Supports class B" and "Supports class C". It is probably fine to make them mutually exclusive, however I'd prefer an approach where both could be selected at the same time. Note, that for multicast devices either "supports class B" or "supports class C" *MUST* be selected
3. I do not like the hidden "advanced settings" approach, in general it's a good idea, but for some cases, e.g. for class B devices in Chinese frequency plan, some of these fields *MUST* be specified and it's confusing for the user to have to search through them. Perhaps you can move the required options out of the hidden menu depending on the device settings?
4. Invalid error handling. NS error is displayed in AS device creation window. As a person unfamiliar with the project I will never figure out what's wrong. What I'd like to see here, is automatic redirection to the place of the error and the field light up in red.
2020-09-30-23:03:02-screenshot

rvolosatovs:
Also, IMHO, it's better UX to make fields disabled instead of removing them depending on, for example, class selection. That way the user could point at the disabled field and receive a hint regarding why is the field disabled.

bafonins:
@rvolosatovs 

>> It should be possible to unset values. Currently if I set RX1 delay to 4 I see no way to delete the value - I tried saving an empty form, but that did not work.

fixed

>> That's not how it works. All devices implement class A. There should be 2 checkboxes here - "Supports class B" and "Supports class C". It is probably fine to make them mutually exclusive, however I'd prefer an approach where both could be selected at the same time. Note, that for multicast devices either "supports class B" or "supports class C" MUST be selected

fixed

>> Invalid error handling. NS error is displayed in AS device creation window. As a person unfamiliar with the project I will never figure out what's wrong. What I'd like to see here, is automatic redirection to the place of the error and the field light up in red.

Will implement this in a separate PR. https://github.com/TheThingsNetwork/lorawan-stack/issues/3291

>>  I do not like the hidden "advanced settings" approach, in general it's a good idea, but for some cases, e.g. for class B devices in Chinese frequency plan, some of these fields MUST be specified and it's confusing for the user to have to search through them. Perhaps you can move the required options out of the hidden menu depending on the device settings?

Now we expand the advanced settings section by default for multicast devices and make the `ping_slot_periodicity` required. Are there any other MAC settings required for other device configurations?

rvolosatovs (changes requested):
2020-10-07-21:17:02-screenshot
NetID or Net ID (@johanstokking ?), KEK labels

I still do not see the the ping slot frequency field and have to manually expand "advanced settings". That's a recipe for disaster, especially considering that NS errors are represented as JS errors (in the screenshot you get a false impression of the UX, because the request is only sent from JS setting tab and that's the selector, which lights up red)
2020-10-07-21:20:00-screenshot

rvolosatovs:
Additionally, after a creation error happens, when I go back to "Network layer settings", the "Supports Class B" checkbox is unchecked, even though it was checked before by me. That also should not happen

rvolosatovs:
Additionally, the user should be able to change the device type and MAC version by just going back through the tabs. Currently to do that, you have to restart the process, which also should not happen

rvolosatovs:
2020-10-07-21:25:59-screenshot
If no ID is set (i.e. the field is not selected) and "Network layer settings" button is pressed - nothing happens, the error also does not appear. The error only starts appearing after something is typed in the field (and, for example, cleared after)

rvolosatovs:
2020-10-07-21:29:16-screenshot
Again, not a fan of this, IMO these should just be 2 checkboxes without title. Otherwise, the title could be "LoRaWAN class capabilities"

johanstokking:
> NetID or Net ID (@johanstokking ?)

NetID

bafonins:
@rvolosatovs 

> It should be possible to unset values. Currently if I set RX1 delay to 4 I see no way to delete the value - I tried saving an empty form, but that did not work.

fixed

> Invalid error handling. NS error is displayed in AS device creation window. As a person unfamiliar with the project I will never figure out what's wrong. What I'd like to see here, is automatic redirection to the place of the error and the field light up in red.

fixed. Now the user is directed to the errored step.

> If no ID is set (i.e. the field is not selected) and "Network layer settings" button is pressed - nothing happens, the error also does not appear. The error only starts appearing after something is typed in the field (and, for example, cleared after)

fixed

> Again, not a fan of this, IMO these should just be 2 checkboxes without title. Otherwise, the title could be "LoRaWAN class capabilities"

changed `End device class` to `LoRaWAN class capabilities`

> NetID, KEK labels

fixed

> I still do not see the the ping slot frequency field and have to manually expand "advanced settings". That's a recipe for disaster, especially considering that NS errors are represented as JS errors

With this PR the user is redirected to the errored step, also I added functionality to expand the advanced settings section on errored submission.

rvolosatovs:
2020-11-19-19:05:43-screenshot
When creating an OTAA device
1. `invalid field mask` is completely opaque to the user
2. There is no way to set the required class B parameter at all

Please parse the error, and make sure the wrong field is somehow highlighted

bafonins:
> There is no way to set the required class B parameter at all

Fixed this. No the NS form properly shows class b related fields on error. Now the user is able to recover from not filling in the `ping_slot_frequency` value.

Wont add this functionality in this PR.

bafonins:
@rvolosatovs

bafonins:
@rvolosatovs

bafonins:
@rvolosatovs

rvolosatovs (approved):
Please avoid unnecessary spam notifications in the future - I was off for a few days, you can see that in the company calendar

PR #3068, opened 2020-08-10
Fix authentication metadata in events

neoaggelos:
#### Summary

Quickfix PR, fixes a small regression introduced in #3018 

#### Changes

- Include `WithAuthFromContext` as a default option.
- Add defaultOptions when defining a new event.

#### Testing

Tested locally

##### Regressions

The default options are added before the options passed on the event definition, so there should not be any. Verified by testing the visibility options.

#### Checklist

### DIFF
--- a/pkg/applicationserver/observability.go
+++ b/pkg/applicationserver/observability.go
@@ -77,6 +77,7 @@ var (
 	evtReceiveDataDown = events.Define(
 		"as.down.data.receive", "receive downlink data message",
 		events.WithVisibility(ttnpb.RIGHT_APPLICATION_TRAFFIC_READ),
+		events.WithAuthFromContext(),
 	)
 	evtDropDataDown = events.Define(
 		"as.down.data.drop", "drop downlink data message",
@@ -85,6 +86,7 @@ var (
 	evtForwardDataDown = events.Define(
 		"as.down.data.forward", "forward downlink data message",
 		events.WithVisibility(ttnpb.RIGHT_APPLICATION_TRAFFIC_READ),
+		events.WithAuthFromContext(),
 	)
 	evtLostQueueDataDown = events.Define(
 		"as.down.data.queue.lost", "lose downlink data queue",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Authentication metadata missing from published events.
+
 ### Security
 
 ## [3.9.0] - 2020-08-06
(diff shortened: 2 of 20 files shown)


pkg/events/definitions.go
```diff
@@ -39,6 +39,7 @@ func (d *definition) With(options ...Option) Builder {
 
 var defaultOptions = []Option{
 	WithVisibility(ttnpb.RIGHT_ALL),
+	WithAuthFromContext(),
```
rvolosatovs: If this is supposed to be always there, why is it an option?

pkg/events/definitions.go
```diff
@@ -71,7 +72,7 @@ func defineSkip(name, description string, skip uint, opts ...Option) Builder {
 	i18n.Define(fmt.Sprintf("%s:%s", i18nPrefix, name), description).SetSource(1 + skip)
 	initMetrics(name)
 
-	var b Builder = def
+	b := def.With(defaultOptions...)
```
rvolosatovs: IMO `defaultOptions` should not be applied here. Since this is the only way to construct an event and these are always applied, it makes no sense to have them as options - all side effects should be directly applied at construction time(i.e. in the constructor).

neoaggelos: cc @htdvisser, since this is related to changes in #3018.

pkg/events/definitions.go
htdvisser: Now that you mention this, I'm not actually sure it should be included by default. Auth information should only be added on events triggered by external calls (registry CRUD, maybe downlink pushes if we have auth context there). So it would indeed be better to set this option on those events directly.

pkg/events/definitions.go
htdvisser: If it makes @rvolosatovs feel better we can rename these things "event mutators" or something.

pkg/applicationserver/observability.go
```diff
@@ -77,6 +77,7 @@ var (
 	evtReceiveDataDown = events.Define(
 		"as.down.data.receive", "receive downlink data message",
 		events.WithVisibility(ttnpb.RIGHT_APPLICATION_TRAFFIC_READ),
+		events.WithAuthFromContext(),
```
adriansmares, line 80: This is correct.

pkg/applicationserver/observability.go
```diff
@@ -85,6 +86,7 @@ var (
 	evtForwardDataDown = events.Define(
 		"as.down.data.forward", "forward downlink data message",
 		events.WithVisibility(ttnpb.RIGHT_APPLICATION_TRAFFIC_READ),
+		events.WithAuthFromContext(),
```
adriansmares, line 89: This is NS->AS (line 440 in `applicationserver.go`), not from any user (and soon not authenticated anymore anyway).

adriansmares (approved):
LGTM ~with some small comments.~

htdvisser (approved):

PR #3072, opened 2020-08-10
Add user agent metadata to published events

### LINKED ISSUE #1173: Add more context to events
#### Summary

We could add some more context to our events.

#### Why do we need this?

Just because we have it, it may also help users when debugging.

We have support for adding "contextual" information to events, but aren't using it.

- IP address of the RPC caller
- User agent of the RPC caller
- ID of the auth token that was used on the RPC

I already looked into this some time ago, and considered adding it as details to the `Event` API. This would work well for RPC events, but would be awkward for uplink/downlink events so I didn't do that. Here's what I already had:

- Use `peer.FromContext`, and check the `x-forwarded-for` headers in the metadata

```go
var remoteAddrs []string
if md, ok := metadata.FromIncomingContext(ctx); ok {
  if forwardedForHeaders, ok := md["x-forwarded-for"]; ok {
    for _, forwardedForHeader := range forwardedForHeaders {
      for _, remoteAddr := range strings.Split(forwardedForHeader, ",") {
        remoteAddrs = append(remoteAddrs, strings.TrimSpace(remoteAddr))
      }
    }
  }
}
if peer, ok := peer.FromContext(ctx); ok {
  peerAddr := peer.Addr.String()
  if peerAddr != "pipe" { // Skip the gRPC loopback conn.
    if host, _, err := net.SplitHostPort(peerAddr); err == nil {
      peerAddr = host
    }
  }
  remoteAddrs = append(remoteAddrs, peerAddr)
}
if len(remoteAddrs) > 0 {
  details.CallerAddress = strings.Join(remoteAddrs, ", ")
}
```

- Check the metadata for a `user-agent` or `grpcgateway-user-agent`

```go
if md, ok := metadata.FromIncomingContext(ctx); ok {
  if userAgents, ok := md["grpcgateway-user-agent"]; ok { // Set by gRPC gateway.
    details.CallerUserAgent = strings.Join(userAgents, " ")
  } else if userAgents, ok := md["user-agent"]; ok {
    details.CallerUserAgent = strings.Join(userAgents, " ")
  }
}
```

- Check the metadata for a `bearer` token, extract the type and ID (but never the secret).

```go
if md := rpcmetadata.FromIncomingContext(ctx); strings.ToLower(md.AuthType) == "bearer" {
  if tokenType, tokenID, _, err := auth.SplitToken(token); err == nil {
    details.TokenType, details.TokenID = tokenType, tokenID
  }
}
```

Yes, but I don't really have time for it now.

neoaggelos:
#### Summary

Closes #1173 

#### Changes

- Add `UserAgent` field to `rpcmetadata.MD`
- Add `UserAgent` field to published events when that is available in the context.

#### Testing

Unit testing, test locally by editing a device using curl, the cli, the console and verifying the user agent on the emitted events.

##### Regressions

There should not be any

#### Notes for Reviewers

Added as a new option, included by default.

#### Checklist


pkg/events/definitions.go
```diff
@@ -39,6 +39,7 @@ func (d *definition) With(options ...Option) Builder {
 
 var defaultOptions = []Option{
 	WithVisibility(ttnpb.RIGHT_ALL),
+	WithUserAgentFromContext(),
```
neoaggelos: See discussion in #3068 for this

htdvisser: Yes. @rvolosatovs made a good point there. The same applies here: it would be better to add this directly on events that are triggered in RPCs called by user agents.

api/events.proto
```diff
@@ -43,6 +43,7 @@ message Event {
     string remote_ip = 4 [(gogoproto.customname) = "RemoteIP"];
   }
   Authentication authentication = 9;
+  string user_agent = 10;
```
htdvisser: I have to say it feels a bit weird to now have `remote_ip` inside `authentication` and `user_agent` separate. Maybe we should pull `remote_ip` out as well then.

I think doing this wouldn't violate our compatibility commitment, because (1) events aren't covered and (2) because of the bug you're fixing in #3068 we weren't sending this yet anyway.

neoaggelos: What about the opposite? Adding the `user_agent` inside `authentication`? This maintains compatbility (even if not explicitly required by the compatibility commitment)

htdvisser: Because users are not authenticating with an IP address or user agent.

pkg/events/options.go
```diff
@@ -67,13 +67,17 @@ func WithVisibility(rights ...ttnpb.Right) Option {
 // WithAuthFromContext returns an option that extracts auth information from the context when the event is created.
 func WithAuthFromContext() Option {
 	return optionFunc(func(e *event) {
-		authentication := &ttnpb.Event_Authentication{}
 		if p, ok := peer.FromContext(e.ctx); ok && p.Addr != nil && p.Addr.String() != "pipe" {
 			if host, _, err := net.SplitHostPort(p.Addr.String()); err == nil {
-				authentication.RemoteIP = host
+				e.innerEvent.RemoteIP = host
 			}
 		}
```
htdvisser: I think this now fits better in the "user agent" thing (and then maybe it's more ClientInfo than UserAgent)

htdvisser (approved):

kschiffer (approved):
Codeowned files LGTM

adriansmares (approved):

johanstokking (commented):
Awesome

PR #3077, opened 2020-08-11
Do not assume supports_join when importing devices with the CLI

neoaggelos:
#### Summary

Quickfix PR that stops assuming `supports_join` is true when importing devices with the `ttn-lw-cli end-devices create < devices.json` command.

#### Testing

Test locally

##### Regressions

This breaks OTAA devices being imported if the `supports_join` field is unset, but this should count as a bug rather than expected behaviour.

#### Notes for Reviewers

For context, https://github.com/TheThingsIndustries/lorawan-stack-migrate/pull/1#discussion_r467604230

> @johanstokking: I see. So the problem is that if you don't pass supports_join, we assume supports_join. I think that is wrong. So the problem is in the create command. Let's not work around that by explicitly setting supports_join to false. It feels like the best solution is to take the values from the input decoder literally, and don't assume defaults in that case.

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/end_devices.go
+++ b/cmd/ttn-lw-cli/commands/end_devices.go
@@ -362,12 +362,10 @@ var (
 				}
 				paths = append(paths, ttnpb.FlattenPaths(decodedPaths, endDeviceFlattenPaths)...)
 
-				if ttnpb.ContainsField("supports_join", paths) {
-					if abp && device.SupportsJoin {
-						return errActivationMode.New()
-					}
-					abp = !device.SupportsJoin
+				if abp && device.SupportsJoin {
+					logger.Warn("Reading from standard input, ignoring --abp and --multicast flags")
 				}
+				abp = !device.SupportsJoin
 			}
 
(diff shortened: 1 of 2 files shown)


cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -362,12 +362,7 @@ var (
 				}
 				paths = append(paths, ttnpb.FlattenPaths(decodedPaths, endDeviceFlattenPaths)...)
 
-				if ttnpb.ContainsField("supports_join", paths) {
-					if abp && device.SupportsJoin {
-						return errActivationMode.New()
-					}
-					abp = !device.SupportsJoin
-				}
+				abp = !device.SupportsJoin
```
htdvisser, line 368: This now ignores the explicit values of the `--abp` and `--multicast` flags (L354-356) if we're receiving anything on stdin, while it previously only ignored those flags if the field `supports_join` was actually present in the input. Not sure if that's a big problem, but maybe good to warn if one of those flags is present while also reading from stdin.

htdvisser (approved):

PR #3093, opened 2020-08-17
OAuth contextual configuration

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/issues/1659

#### Changes

- OAuth server configuration is no longer static and now is provided by a `ConfigProvider` which returns the configuration based on the context. A static provider has been added for ease of use and backwards compatibility.
- OAuth server constructor now requires a `Component` and a `ConfigProvider` instead of `context.Context` and `Config`. The component is needed in order to allow the OAuth server to use `pkg/web/oauthclient`.
- OAuth client configuration (`*oauth2.Config`) can now be overridden contextually. The original design of `pkg/web/oauthclient` is tailored towards our own OAuth server (or rather, to the normal addressing schema). The changes proposed by this PR allows the component that constructs the client to provide it's own `*oauth2.Config` at runtime using the `WithOAuth2ConfigProvider` option.
- OAuth client token callback end may now be overwritten in order to allow additional steps to be performed after the token exchange. 

#### Testing

Tested as a whole as part of closed-source [changeset](https://github.com/TheThingsIndustries/lorawan-stack/tree/feature/1659-fed-auth-db ). Changes are backwards compatible, as the default behavior (i.e. without any `Option`s) is just the one that was available before.

##### Regressions

The already existing uses of the two packages are virtually unchanged, and this has been tested as described above.

#### Notes for Reviewers

See [this federated provider](https://github.com/TheThingsIndustries/lorawan-stack/tree/feature/1659-fed-auth-db/pkg/oauth/oidc) in order to see how the changes are used. Feel free to suggest changes or alternative approaches. I've tried to reuse and integrate our already existing building blocks as best as possible.

@htdvisser and @kschiffer main reviewers.
@rvolosatovs, @bafonins, @johanstokking no need to review (I cannot manually remove the requests, sorry).

#### Checklist

### DIFF
--- a/pkg/identityserver/identityserver.go
+++ b/pkg/identityserver/identityserver.go
@@ -100,7 +100,7 @@ func New(c *component.Component, config *Config) (is *IdentityServer, err error)
 	}
 
 	is.config.OAuth.CSRFAuthKey = is.GetBaseConfig(is.Context()).HTTP.Cookie.HashKey
-	is.oauth = oauth.NewServer(is.Context(), struct {
+	is.oauth, err = oauth.NewServer(c, struct {
 		store.UserStore
 		store.UserSessionStore
 		store.ClientStore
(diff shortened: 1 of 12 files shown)


htdvisser (approved):
LGTM

Issue #2909, opened 2020-07-14
Add `mac-settings.mac-commands-in-class-b-c`


rvolosatovs:
#### Summary

Add and respect `mac-settings.mac-commands-in-class-b-c`

#### Why do we need this?

To allow configuring whether NS would send MAC commands in class B/C downlinks or not

#### What is already there? What do you see now?

MAC requests only sent in class A downlink slots

#### What is missing? What do you want to see?

Introduce `mac-settings.mac-commands-in-class-b-c`, which is `false` by default and would allow MAC requests to be sent in class B/C downlinks

#### How do you propose to implement this?

Add field to proto
When the flag is set to `true` - allow bypassing MAC command handling in class B/C

#### How do you propose to test this?

Unit tests, integration testing

#### Can you do this yourself and submit a Pull Request?

yes

rvolosatovs:
Updated issue according to the changes in https://github.com/TheThingsNetwork/lorawan-stack/pull/3103

johanstokking:
I highly doubt whether we want this. From LoRaWAN 1.0.4, MAC commands are always in class A. Adding discussion so that we don't forget to discuss this before planning it.

rvolosatovs:
> I highly doubt whether we want this. From LoRaWAN 1.0.4, MAC commands are always in class A. Adding discussion so that we don't forget to discuss this before planning it.

Indeed, I also would prefer not doing this, the issue is there to account for this feature being available before https://github.com/TheThingsNetwork/lorawan-stack/pull/3103

johanstokking:
OK, let's close it then.

Issue #2774, opened 2020-06-20
NS reports "WARN MAC command buffer not fully answered" for Class C ABP devices


ama9910:
#### Summary
When a Class C ABP device first sends an uplink, the NS sends a number of MAC commands in a series of downlink packets and expects the device to respond with confirmations in a number of uplinks.

If an application downlink is queued before the MAC confirmations are received, the NS reports "WARN MAC command buffer not fully answered" and the downlink is dropped.

Do we need to wait for all MAC answers to be returned from the device before allowing a downlink to be sent? Unlike OTAA where there is a security context issue, ABP downlinks could still be sent on known channels for instance.

We have observed some devices that do not immediately respond to the initial MAC downlinks and it can take some time before these devices finally respond and the application is able to control them.

#### Steps to Reproduce

1. Register a Class C ABP device
2. Send an uplink (f_cnt=1)
3. Send an application downlink
4. Ensure device does not respond with an uplink
5. Note the following:
    - Device did not receive the downlink
    - Backend logs show "WARN MAC command buffer not fully answered"

#### What do you see now?
Device does not receive downlink until it has answered ALL MAC commands from the NS

#### What do you want to see instead?
Where application downlink can be sent to device, downlink should be sent even if device has not confirmed all MAC requests.

#### Environment
V3.8.3
Test device is a Multitech mDot

rvolosatovs:
See Class C section of LoRaWAN 1.1 spec (L2329-2335)
![2020-06-22-11:36:04-screenshot](https://user-images.githubusercontent.com/12877905/85272550-98c1bf00-b47c-11ea-90b0-94cab7090a57.png)

This is precisely what NS does. The default interval, if not configured(`mac-settings.class-c-timeout` option), is 5 minutes.
Note, that this behavior is enforced for devices of all LoRaWAN versions, because that's a generally reasonable thing for NS to do.

Regarding device not receiving the downlink, it seems that the device's MAC layer is simply not configured correctly until MAC command exchange is finished.
Until MAC answers are received from the device, NS assumes the spec defaults (which may be overriden on per-setting basis per-device, see MACSettings), and if device is actually using different MAC layer settings, it may not receive the downlink, so it seems like expected behavior to me. If you disagree, please provide some device-specific data to debug - the initial device configuration and device event log, you can do it either here or you can send me via email - roman@thethingsnetwork.org.

ama9910:
[...]

PR #3103, opened 2020-08-19
Only schedule MAC commands in class A downlink slots

rvolosatovs:
#### Summary

References #2909 https://github.com/TheThingsNetwork/lorawan-stack/issues/2774

Incidentally, this makes proceeding with #3052 easier.

cc @ama9910

#### Changes

- Only schedule MAC commands in class A downlink slots
- In case something goes wrong with MAC queue and either responses don't fit or an unknown command is present, this is only logged and does not prevent new MAC requests from being scheduled anymore.

#### Testing

integration testing

##### Regressions

This can affect MAC command scheduling for all classes, regressions not expected.

#### Notes for Reviewers

Actual support for flag from #2909 will be added later.
Perhaps we should implement it as part of the next release to allow users to fallback to the legacy behavior? To be honest, it seems that scheduling MAC in classes B/C causes more trouble than benefits for us and our customers.
In the past we have encountered numerous issues with this and many users were confused about this behavior and consequences it brings to downlink scheduling and priorities. Perhaps we could move #2909 to Backlog instead and only implement if customers explicitly ask for it? Again, as I see it now, I don't think anyone today is going to miss this feature and users are more likely to be happy for this change, however it does change behavior. Thoughts @johanstokking ?

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- MAC commands (both requests and responses) are now only scheduled in class A downlink slots in accordance to latest revisions to LoRaWAN specification.
+
 ### Deprecated
 
 ### Removed
--- a/pkg/networkserver/downlink.go
+++ b/pkg/networkserver/downlink.go
@@ -171,35 +171,42 @@ func (ns *NetworkServer) generateDataDownlink(ctx context.Context, dev *ttnpb.En
 	}
 	maxDownLen, maxUpLen = maxDownLen-8, maxUpLen-8
 
-	var fPending bool
-	spec := lorawan.DefaultMACCommands
-	cmds := make([]*ttnpb.MACCommand, 0, len(dev.MACState.QueuedResponses)+len(dev.MACState.PendingRequests))
-	var lostResps []*ttnpb.MACCommand
+	var (
+		fPending bool
+		genState generateDownlinkState
+		cmdBuf   []byte
+	)
 	if class == ttnpb.CLASS_A {
-		for i, cmd := range dev.MACState.QueuedResponses {
-			desc := spec[cmd.CID]
-			if desc == nil {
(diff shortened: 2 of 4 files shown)


pkg/networkserver/downlink.go
```diff
@@ -171,35 +171,42 @@ func (ns *NetworkServer) generateDataDownlink(ctx context.Context, dev *ttnpb.En
...
+			logger := logger.WithField("cid", cmd.CID)
+			desc, ok := spec[cmd.CID]
+			switch {
+			case !ok:
+				logger.Error("Unknown MAC command response enqueued, set FPending")
 				maxDownLen = 0
 				fPending = true
-				break
-			}
-			if desc.DownlinkLength > maxDownLen {
```
rvolosatovs, line 187: Note the bug fixed here - `desc.DownlinkLength` only contains the command payload length, but we should also account for the CID

johanstokking:
> Perhaps we should implement it as part of the next release to allow users to fallback to the legacy behavior? To be honest, it seems that scheduling MAC in classes B/C causes more trouble than benefits for us and our customers.

I agree

> In the past we have encountered numerous issues with this and many users were confused about this behavior and consequences it brings to downlink scheduling and priorities. Perhaps we could move #2909 to Backlog instead and only implement if customers explicitly ask for it?

Yes, fine with me

> Again, as I see it now, I don't think anyone today is going to miss this feature and users are more likely to be happy for this change, however it does change behavior.

I agree

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- MAC commands (both requests and reponses) are now only schduled in class A downlink slots in accordance to latest revisions to LoRaWAN specification.
```
```suggestion
- MAC commands (both requests and responses) are now only scheduled in class A downlink slots in accordance to latest revisions to LoRaWAN specification.
```

johanstokking (approved):

ama9910:
Thanks. I don't think the new behaviour will cause any problems for Class C devices that we have implemented/tested, and good to know #2909 is still possible if needed.

Issue #3010, opened 2020-08-03
Source out custom icons into own icon font


kschiffer:
#### Summary
We currently use a modified version of the material icon font to include a necessary icon as sort order indicator. This introduces some maintainability issues when updating the font. Instead we should use our own icon font for all custom icons that we use.

#### Why do we need this?
- Improve maintainability of our web fonts
- Avoid accidental overrides of icon fonts resulting in loss of custom icons

#### What is already there? What do you see now?
Modified material icons font.

#### What is missing? What do you want to see?
- Original material icons font
- Custom tti icon font with additional icons that we need

#### How do you propose to implement this?
1. Extract custom icon to own icon font using some font creation software
2. Modify `<Icon />` component to source from the custom icon font when necessary

#### How do you propose to test this?
Manual testing

#### Can you do this yourself and submit a Pull Request?
Yes.

kschiffer:
Not doing this. We'll use the new google material icons instead.

Issue #2664, opened 2020-06-05
Revisit documentation strategy of forms in the Console


kschiffer:
#### Summary
After adding a lot of content to our documentation recently, as well as a glossary, it's time to revisit the way we assist users with documentation in our forms.

#### Why do we need this?
- Entity creation forms are quite complex (e.g. Gateways, End Devices, Webhooks, etc)
- We received quite some feedback about questions that arose while filling forms in the Console
- Having to research answers for road-blocks while using forms disrupts the UX and puts users off 
- Increase the value and accessibility of our documentation

#### What is already there? What do you see now?
- A lot of [documentation content](https://thethingsstack.io/)
- A glossary for common and complex LoRaWAN or TTS related terminology #2425 
- A react component that allows us to reference documentation from within the Console #2397

#### What is missing? What do you want to see?
- A strategy that informs us how to assist users with documentation content when filling out forms
- Applying that strategy to the console

#### How do you propose to implement this?
I have already created an issue for [displaying glossary entries in the console](#2663). I think this will already assist users a lot as it allows them to obtain additional information right in place without having to disrupt the process.

That will give us three different way to assist users with documentation:
- Inline field descriptions (currently the only thing we use)
- Glossary tooltips for field names
- Referencing the documentation (mostly as part of the field description)

There is also an `<InfoField />` component which can be used to display longer text and documentation links. It might be useful in some situations where important information has to be shown inline and it would not fit into field descriptions. Generally, I would suggest avoiding using this element.

I propose the following usage strategy for these three distinct types:
##### Inline field descriptions
- Should provide information about the particular property/field in context of the current process
- Should avoid to explain the general/overall meaning of the field/property; leave this to the glossary tooltip
- E.g. for `LoRaWAN Version` -> `The LoRaWAN version (MAC), as provided by the device manufacturer`
  - This is a good and relevant description because it informs how the field relates to the end device (and hence the creation of the end device)

##### Glossary tooltips
- Are attached to the field title (see #2663)
- Provide general information about the meaning of a field/property/term in relation to TTS or LoRaWAN
- Should be written concisely, in a way that the most important information can be understood from the first 30 words.
  - Keeping in mind that we might have to abbreviate longer glossary content with an ellipsis `…`, to not run into display issues

##### Documentation links
- Should be attached to the end of field descriptions
- Should reference detailed guides, information that cannot be explained in the brevity of a field description
  - if possible it should use hashes to point to specific positions within a documentation page
[...]

PR #3114, opened 2020-08-20
console: Add generate join eui button

benolayinka:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/issues/2310

#### Changes

- Add button to generate `0`s Join EUI

Screenshot 2020-08-20 at 17 29 41

#### Testing

Run locally

#### Notes for Reviewers

@kschiffer there's no perfect material icon for this, do/can we make one?

#### Checklist


benolayinka:
This button is actually maybe not a good idea. It's easy to press and clear a JoinEUI you entered if you don't know what it's for, and for most cases where people have a real JoinEUI it may be difficult to understand why the button is there

johanstokking:
> This button is actually maybe not a good idea. It's easy to press and clear a JoinEUI you entered if you don't know what it's for, and for most cases where people have a real JoinEUI it may be difficult to understand why the button is there

But if they add a device, and they accidentally click the button, they can re-enter it, right?

It would be more problematic if this happens during a device update, but we don't allow changing EUIs once the device is created.

I think it's good idea to have this. Maybe put `00` in the button as it's hex, but otherwise I'm happy with it.

benolayinka:
> > This button is actually maybe not a good idea. It's easy to press and clear a JoinEUI you entered if you don't know what it's for, and for most cases where people have a real JoinEUI it may be difficult to understand why the button is there
> 
> But if they add a device, and they accidentally click the button, they can re-enter it, right?

Sure, but that's annoying. Unless we can make it really clear why and when someone wouldn't have a JoinEUI and would need to use this, I think this button may just add more confusion.

I definitely think we need to update the inline explanation with very succinct instructions. These explanations are helpful for first time users who aren't familiar with these terms, and I think they should guide as clearly as possible where to find the data that goes in the box.

So in this case,

"JoinEUI identifies the Join Server and is provided by the device manufacturer. If no JoinEUI is provided (usually because your device has a programmable JoinEUI) use `0000000000000000`.

and if we wanna keep the button

"JoinEUI identifies the Join Server and is provided by the device manufacturer. If no JoinEUI is provided (usually because your device has a programmable JoinEUI) use `0000000000000000`. Click the zeros button to prefill zeros."

@kschiffer

kschiffer:
> "JoinEUI identifies the Join Server and is provided by the device manufacturer. If no JoinEUI is provided (usually because your device has a programmable JoinEUI) use `0000000000000000`.

This is good but maybe we can say " If not JoinEUI is provided (…) the field can be filled with zeros".

If we have some documentation regarding this, we should also put a doc link here.

johanstokking:
> (usually because your device has a programmable JoinEUI)

I would say "(usually for research and development)"

That's really what it is. There can be off-the-shelf devices with empty JoinEUI but that is rare, therefore "usually".

> If no[t] JoinEUI is provided

It's tricky to assume empty JoinEUI then. There should always be a JoinEUI (or AppEUI) provided, but it may be blank, and for R&D without an official block, it's best to use blank.

johanstokking (approved):
LGTM, please generate translations to fix CI

benolayinka:
> LGTM, please generate translations to fix CI

@kschiffer if there's no "00" material icon, can we create one?

johanstokking:
> @kschiffer if there's no "00" material icon, can we create one?

References https://github.com/TheThingsNetwork/lorawan-stack/issues/3010

pkg/webui/console/containers/join-eui-prefixes-input/join-eui-prefixes-input.js
```diff
@@ -179,6 +186,19 @@ class JoinEUIPrefixesInput extends React.PureComponent {
...
+          type="button"
+          className={style.buttonZeros}
+          onClick={this.handleZerosClick}
+        >
+          <Icon
+            className={style.buttonIcon}
+            onClick={this.handleZerosClick}
+            small
+            icon="exposure_zero"
+          />
```
kschiffer: Use a custom `<span>00</span>` instead of the icon and style it to use `$font-family-mono`.

pkg/webui/console/containers/join-eui-prefixes-input/join-eui-prefixes-input.js
```diff
@@ -28,6 +29,7 @@ import style from './join-eui-prefixes-input.styl'
 
 const m = defineMessages({
   empty: 'No prefix',
+  zeroInput: 'Zero input',
```
kschiffer:
```suggestion
  zeroInput: 'Fill with 00',
```

pkg/webui/lib/shared-messages.js
```diff
@@ -45,7 +45,7 @@ export default defineMessages({
   appEUI: 'AppEUI',
   appEUIJoinEUI: 'AppEUI/JoinEUI',
   appEUIDescription:
-    'The AppEUI is a global application ID that uniquely identifies the owner of the end ­device. AppEUI is used for end devices with LoRaWAN versions 1.0.0-1.0.3, JoinEUI for 1.0.4+.',
+    'The AppEUI uniquely identifies the owner of the end ­device. If no AppEUI is provided by your device manufacturer (usually for development), use 0000000000000000',
```
```suggestion
    'The AppEUI uniquely identifies the owner of the end ­device. If no AppEUI is provided by your device manufacturer (usually for development), it can be filled with zeros',
```

I still think this is better than suggesting the full value, which looks odd.

kschiffer (changes requested):
Let's keep it simple and use an actual `00` string instead of an actual icon (styled to use `$font-family-mono`).

pkg/webui/lib/shared-messages.js
benolayinka: true that it looks nicer, but it's a little bit clearer to say "use 0000000000000000," especially because they also have to program that number in to their device.

benolayinka:
Here's how it looks now. I hate to say it but my gut feeling says to remove the button @johanstokking - it's not that hard to manually enter zeros and then we don't have to explain why in most cases you shouldn't need to press the button.

Otherwise, we need this really long description text.

Screenshot 2020-08-26 at 11 04 37

kschiffer:
We decided to use the button now. Please remove the extra sentence that mentions the button. It should be clear enough from the rest of the field description. We cannot provide an overly verbose explanation of the field for all possible situations, this is to be left to the documentation. Feel free to use a `DocLink` inside the field description if such documentation already exists. We have to live with a possible last bit of unclarity until we implement a proper inline documentation strategy (#2664).

Also, note that the styling of the button is a bit off. It should look like the other icon buttons for text inputs that we have (e.g. auto-generate button for other EUIs).

pkg/webui/lib/shared-messages.js
kschiffer: The field description is not a good place to suggest concrete byte values because we always style displayed byte values in a `<SafeInspector />`. My condition to accept displaying the full value is that it is styled in a way that resembles the way that values are displayed in the `<SafeInspector />`.

IMO it's not worth the hassle since the other formulation should be clear enough in conjunction with the fill button.

benolayinka:
Updated to use generate button styling from other form fields. Will add a doclink as part of another PR

Screenshot 2020-08-26 at 13 39 42

pkg/webui/console/containers/join-eui-prefixes-input/join-eui-prefixes-input.js
```diff
@@ -178,6 +184,13 @@ class JoinEUIPrefixesInput extends React.PureComponent {
           onChange={this.handleChange}
           onBlur={this.handleBlur}
           error={error}
+          action={{
+            icon: 'autorenew',
+            type: 'button',
+            title: m.zeroInput,
+            onClick: this.handleZerosClick,
+            raw: true,
+          }}
```
kschiffer:
```suggestion
          action={{
            type: 'button',
            title: m.zeroInput,
            onClick: this.handleZerosClick,
            raw: true,
            children: (<span className={style.zeroFillButton}>00</span>)
          }}
```

We still need to use the `00` "icon".

pkg/webui/lib/shared-messages.js
```diff
@@ -45,7 +45,7 @@ export default defineMessages({
   appEUI: 'AppEUI',
   appEUIJoinEUI: 'AppEUI/JoinEUI',
   appEUIDescription:
-    'The AppEUI is a global application ID that uniquely identifies the owner of the end ­device. AppEUI is used for end devices with LoRaWAN versions 1.0.0-1.0.3, JoinEUI for 1.0.4+.',
+    'The AppEUI uniquely identifies the owner of the end ­device. If no AppEUI is provided by your device manufacturer (usually for development), fill with zeros',
```
```suggestion
    'The AppEUI uniquely identifies the owner of the end ­device. If no AppEUI is provided by your device manufacturer (usually for development), it can be filled with zeros.',
```

https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.9/DESIGN.md#general-considerations

> *   Do not use direct address of the user except for dialogs
> *   Do not use end-of-sentence periods, unless the text consists of multiple sentences

johanstokking:
This is not generate, this is fill with zeros. I don't think this icon is clear.

kschiffer (changes requested):

pkg/webui/console/containers/join-eui-prefixes-input/join-eui-prefixes-input.js
kschiffer: Sorry, wasn't clear enough on that one.

benolayinka:
Screenshot 2020-08-27 at 13 32 00

kschiffer:
I added some fixes and improvements. There were some issues related to the blur behavior of the input as a whole, also the fill button did not set the prefix back to the default.

benolayinka (commented):
Looks good to me. Do we need another reviewer?

kschiffer (approved):

kschiffer:
Should be fine, you can mere.

PR #3117, opened 2020-08-20
Change AS linking retry logic

adriansmares:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2313
Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2265

#### Changes

- Add dynamic backoff duration support to component tasks
- Use dynamic backoff duration depending on last error code that occurred during linking. Problematic error codes can go to up to 30 minutes of backoff
- Link tasks no longer stop unless explicitly stopped by a `Set` or `Delete` call, no matter the error code
- Count link failures for any non-`nil` error
- Always count link stops

#### Testing

Already existing tests cover this. The fixes presented here deal with issues that exist only at scale.

##### Regressions

This changes link behavior in order to allow links to live longer lifetimes and no longer get cancelled too early.

#### Checklist

### DIFF
--- a/pkg/applicationserver/linking.go
+++ b/pkg/applicationserver/linking.go
@@ -44,6 +44,45 @@ func (as *ApplicationServer) linkAll(ctx context.Context) error {
 	)
 }
 
+var (
+	linkHealthyDuration = 1 * time.Minute
+	linkBackoffConfig   = &component.TaskBackoffConfig{
+		Jitter: component.DefaultTaskBackoffConfig.Jitter,
+		DynamicInterval: func(ctx context.Context, executionTime time.Duration, invocation int, err error) time.Duration {
+			defaultIntervals := component.DefaultTaskBackoffConfig.Intervals
+			extendedIntervals := append(defaultIntervals,
+				1*time.Minute,
+				5*time.Minute,
+				15*time.Minute,
+				30*time.Minute,
+			)
+
+			var intervals []time.Duration
+			switch {
+			case errors.IsFailedPrecondition(err),
+				errors.IsUnauthenticated(err),
+				errors.IsPermissionDenied(err),
+				errors.IsInvalidArgument(err),
+				errors.IsAlreadyExists(err),
+				errors.IsCanceled(err):
+				intervals = extendedIntervals
+			default:
+				intervals = defaultIntervals
+			}
+
+			bi := invocation - 1
+			if bi >= len(intervals) {
+				bi = len(intervals) - 1
+			}
+			if executionTime > linkHealthyDuration {
+				bi = 0
+			}
+
+			return intervals[bi]
+		},
+	}
+)
+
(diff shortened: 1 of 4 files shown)


pkg/applicationserver/linking.go
```diff
@@ -44,6 +44,45 @@ func (as *ApplicationServer) linkAll(ctx context.Context) error {
...
+	linkHealthyDuration = 1 * time.Minute
+	linkBackoffConfig   = &component.TaskBackoffConfig{
+		Jitter: 0.3,
+		DynamicInterval: func(ctx context.Context, executionTime time.Duration, invocation int, err error) time.Duration {
+			defaultIntervals := component.DefaultTaskBackoffConfig.Intervals
+			extendedIntervals := append(defaultIntervals,
+				1*time.Minute,
+				5*time.Minute,
+				15*time.Minute,
+				30*time.Minute,
```
adriansmares, line 57: Timings as per https://github.com/TheThingsIndustries/lorawan-stack/issues/2265#issuecomment-664238105

pkg/applicationserver/linking.go
```diff
@@ -63,23 +102,10 @@ func (as *ApplicationServer) startLinkTask(ctx context.Context, ids ttnpb.Applic
...
-			switch {
-			case errors.IsFailedPrecondition(err),
-				errors.IsUnauthenticated(err),
-				errors.IsPermissionDenied(err),
-				errors.IsInvalidArgument(err):
-				log.FromContext(ctx).WithError(err).Warn("Failed to link")
-				return nil
-			case errors.IsCanceled(err),
-				errors.IsAlreadyExists(err):
-				return nil
```
adriansmares, line 76: `nil` here is very aggressive and `err` can actually be a remote error (so it's not a local cancellation, but rather the remote host cancelled his context). As such, these errors shouldn't result in the link being closed, but instead in a longer back off.

pkg/applicationserver/linking.go
```diff
@@ -226,10 +252,8 @@ func (as *ApplicationServer) link(ctx context.Context, ids ttnpb.ApplicationIden
 	registerLinkStart(ctx, l)
 	go func() {
 		<-ctx.Done()
-		if err := ctx.Err(); errors.IsCanceled(err) {
-			logger.Info("Unlinked")
-			registerLinkStop(ctx, l)
-		}
+		logger.WithError(ctx.Err()).Info("Unlinked")
```
adriansmares, line 255: We need to always count the link stops, no matter if they ended in an error or not. Otherwise we cannot determine how many links are active at a time (since errors are logged even before the link starts - for connection failure purposes).

htdvisser:
I think @rvolosatovs should also take a look given his recent work on the task logic

htdvisser (approved):

johanstokking (approved):

rvolosatovs (approved):

Issue #2965, opened 2020-07-23
CLI Auth Error - Following Change to TLS Redirection in Config


ZeroSum24:
#### Summary

Using ttn-lw-cli v3.8.7 I'm unable to run console commands against a self-managed TTN v3.8.7 deployment. This was working previously and has since stopped following the addition of the http.redirect-to-tls flag in the deployment configuration.

The console is still accessible and working as expected. I am not sure if this is a bug or a system configuration issue. Either way your help would be much appreciated!

#### Steps to Reproduce

1. Install dockerised TTN deployment on a remote server with a Let's Encrypt certificate
2. Follow the getting started guide on setting up the deployment, adding in the http.redirect-to-tls flag, set to true
3. Install the ttn-lw-cli toolset on another machine, and point the CLI to use the server domain
4. Try to run a command using the setup ttn-lw-cli toolset

**ttn-lw-stack.yml - Server:**

```
# Identity Server configuration
 is:
  # Email configuration for "REMOVED"
   email:
     sender-name: 'The Things Stack'
     sender-address: 'noreply@REMOVED'
     network:
       name: 'TTN-stack'
       console-url: 'https://REMOVED/console'
       identity-server-url: 'https://REMOVED/oauth'

 # Web UI configuration for "REMOVED":
   oauth:
      ui:
        canonical-url: 'https://REMOVED/oauth'
        is:
          base-url: 'https://REMOVED/api/v3'

 # HTTP server configuration
 http:
   cookie:
     block-key: 'REMOVED'                # generate 32 bytes (openssl rand -hex 32)
     hash-key: 'REMOVED'                 # generate 64 bytes (penssl rand -hex 64)
   metrics:
     password: 'REMOVED'               # choose a password
   pprof:
     password: 'REMOVED'                 # choose a password
   redirect-to-tls: 'true'

# Let's encrypt for "REMOVED"
 tls:
   source: 'acme'
   acme:
     dir: '/var/lib/acme'
     email: 'REMOVED'
     hosts: ['REMOVED']
     default-host: 'REMOVED'

 #If Gateway Server enabled, defaults for "REMOVED":
 gs:
   mqtt:
     public-address: 'REMOVED:1882'
     public-tls-address: 'REMOVED:8882'
   mqtt-v2:
     public-address: 'REMOVED:1881'
     public-tls-address: 'REMOVED:8881'

 #If Gateway Configuration Server enabled, defaults for "REMOVED":
 gcs:
[...]

PR #3120, opened 2020-08-20
cli: Fix login issues with explicit :443 port

neoaggelos:
#### Summary

References #2965

#### Changes

Drop the explicit `:443` port number from the oauth server address if present.

#### Testing

Test locally, with `http.redirect-to-tls` enabled and disabled. Login works as expected.

##### Regressions

There should not be any, 443 is the default HTTPS port

#### Notes for Reviewers

This does not fix the issue on the server side, it only makes sure that the CLI does not suffer from it.

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/root.go
+++ b/cmd/ttn-lw-cli/commands/root.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cmd/ttn-lw-cli/commands/root.go
```diff
@@ -133,6 +134,14 @@ func preRun(tasks ...func() error) func(cmd *cobra.Command, args []string) error
 			}
 		}
 
+		// Drop explicit `:443` port number from OAuth server address if present.
+		// Causes issues with `--http.redirect-to-tls` stack option.
+		u, err := url.Parse(config.OAuthServerAddress)
+		if u.Port() == "443" {
+			u.Host = u.Hostname()
+			config.OAuthServerAddress = u.String()
+		}
```
htdvisser: I think it's better to also check the scheme. If it's `http` on `80` or `https` on `443` we can remove the port.

htdvisser (approved):

PR #3123, opened 2020-08-23
Fix oauth logout redirect URI composition

kschiffer:
#### Summary
This quickfix fixes the way that the logout redirect URI is composed.

#### Changes
- Use the whole URI, not just the path

#### Testing

Manual

##### Regressions

This could break the logout redirect for client-initiated logouts.

#### Notes for reviewers
Would be good if you could checkout and do some manual testing as well before we merge this, to be sure.

#### Checklist

### DIFF
--- a/pkg/oauth/user.go
+++ b/pkg/oauth/user.go
@@ -17,7 +17,6 @@ package oauth
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"net/http"
 	"net/url"
 	"runtime/trace"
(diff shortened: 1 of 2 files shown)


pkg/oauth/user.go
```diff
@@ -288,7 +288,7 @@ func (s *server) ClientLogout(c echo.Context) error {
 	if err != nil {
 		return err
 	}
-	return c.Redirect(http.StatusFound, fmt.Sprintf("%s?%s", url.Path, url.RawQuery))
+	return c.Redirect(http.StatusFound, fmt.Sprintf("%s?%s", url.String(), url.RawQuery))
```
htdvisser: I'm pretty sure `url.String()` already renders the query if not empty.

htdvisser (approved):

PR #3141, opened 2020-08-25
as: Rename config to `packages`

neoaggelos:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/pull/2295.

This is not breaking config, there are no config options for application packages in TTS.

### DIFF
--- a/pkg/applicationserver/config.go
+++ b/pkg/applicationserver/config.go
@@ -54,15 +54,15 @@ type InteropConfig struct {
 
 // Config represents the ApplicationServer configuration.
 type Config struct {
-	LinkMode            string                    `name:"link-mode" description:"Mode to link applications to their Network Server (all, explicit)"`
-	Devices             DeviceRegistry            `name:"-"`
-	Links               LinkRegistry              `name:"-"`
-	MQTT                config.MQTT               `name:"mqtt" description:"MQTT configuration"`
-	Webhooks            WebhooksConfig            `name:"webhooks" description:"Webhooks configuration"`
-	PubSub              PubSubConfig              `name:"pubsub" description:"Pub/sub messaging configuration"`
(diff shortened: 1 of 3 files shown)


adriansmares (commented):
`ApplicationPackagesConfig` contains the registry configuration, so I don't see why this is backwards compatible with already existing configurations.

neoaggelos:
> `ApplicationPackagesConfig` contains the registry configuration, so I don't see why this is backwards compatible with already existing configurations.

That is not user configuration though, this is configured in https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.9/cmd/ttn-lw-stack/commands/start.go#L244-L246

pkg/applicationserver/config.go
```diff
@@ -60,7 +60,7 @@ type Config struct {
 	MQTT                config.MQTT               `name:"mqtt" description:"MQTT configuration"`
 	Webhooks            WebhooksConfig            `name:"webhooks" description:"Webhooks configuration"`
 	PubSub              PubSubConfig              `name:"pubsub" description:"Pub/sub messaging configuration"`
-	ApplicationPackages ApplicationPackagesConfig `name:"application-packages" description:"Application packages configuration"`
+	ApplicationPackages ApplicationPackagesConfig `name:"packages" description:"Application packages configuration"`
```
adriansmares: `Packages ApplicationPackagesConfig`

adriansmares (approved):
LGTM but cc @johanstokking

johanstokking (approved):

pkg/applicationserver/config.go
johanstokking: Isn't it then `name:"-"` ?

adriansmares: We need the actual name in proprietary.

PR #3152, opened 2020-08-26
Refactor AS message processors

### LINKED ISSUE #2670: Flexible JavaScript message processor input/output
#### Summary

Flexible JavaScript message processor input and output

#### Why do we need this?

This allows for adding input fields and output fields for JavaScript payload processors.

Currently we adopt The Things Network Stack V2 interface, which is extended with an `env` global variable:

```js
function Decoder(payload, f_port) {
	return {
		temperature: ((payload[0] & 0x80 ? 0xffff : 0x0000) << 16 | payload[0] << 8 | payload[1]) / 100,
		dev_eui: env.dev_eui,
		brand: env.brand,
		model: env.model,
	}
}
```

For downlink, there's an `Encoder()` function. This function expects an FPort, while the FPort should be the result of the Enocder.

A more flexible approach with an input and output object;

```js
function decode(input) {
  return {
    data: {
      key: "value"
    }
  }
}
```

This allows for returning state, warnings and errors, for example:

```js
function decode(input) {
  var state = input.state;
  state.key = "value";
  return {
    data: {
      key: "value",
      count: input.bytes.length
    },
    state: state,
    warnings: ["warning 1", "warning 2"],
    errors: ["error1"]
  }
}
```

The encoder would need to return the bytes as well as the FPort.

We need to stay backwards compatible, so we can't change the signature of `Decoder()` and `Encoder()`. We should therefore introduce new function names, and our wrapper would basically test which one is defined and call the right function with the right data. The current wrapper looks like this:

https://github.com/TheThingsNetwork/lorawan-stack/blob/b5b911f6dcbd9bbe9214c60b64252b862956b404/pkg/messageprocessors/javascript/javascript.go#L134-L141

And this would basically become something like

```js
if (typeof Decoder == "function") {
  Decoder();
} else {
  decode({ bytes: env.payload, fPort: env.f_port });
}
```

For the FPort, AS needs to accept downlink with unset FPort (`0`) and allow the encoder to set it. Only if the application and the encoder didn't set a valid FPort, AS should reject the downlink.

Unit tests will do

Can do, can review

### LINKED ISSUE #2975: Replace otto with goja
#### Summary

We should replace the unmaintained [otto](https://github.com/robertkrimen/otto) library with [goja](https://github.com/dop251/goja).

Refs #2670 and #2966

#### Why do we need this?

Because we should avoid unmaintained libraries, especially ones that we use to run arbitrary code.

johanstokking:
#### Summary

Closes #2670
Closes #2975
Groundwork for https://github.com/TheThingsNetwork/lorawan-stack/issues/263

#### Changes

- Replace `otto` by `goja`
- Introduce input/output objects that give more flexibility, with backwards compatibility
- Decode downlink messages
- Retire device repository; it wasn't used and not documented anyway
- Change message processor API for external runners; it wasn't used and doesn't fit with new structure

#### Testing

Unit testing

##### Regressions

There should be no functional regressions.

A known performance regression is that `goja` doesn't support setting memory limits (see https://github.com/dop251/goja/issues/6). This may lead to problems in environments where anyone can just enter arbitrary code that allocates a lot of memory.

#### Notes for Reviewers

Others: code owned files only

#### Checklist


pkg/applicationserver/payload.go
```diff
@@ -62,17 +63,18 @@ func (as *ApplicationServer) encodeAndEncrypt(ctx context.Context, dev *ttnpb.En
...
-	return as.decode(ctx, dev, uplink, defaultFormatters)
+	return as.decodeUplink(ctx, dev, uplink, defaultFormatters)
 }
 
-func (as *ApplicationServer) decrypt(ctx context.Context, dev *ttnpb.EndDevice, uplink *ttnpb.ApplicationUplink) error {
+func (as *ApplicationServer) decryptUplink(ctx context.Context, dev *ttnpb.EndDevice, uplink *ttnpb.ApplicationUplink) error {
 	if dev.Session == nil || dev.Session.AppSKey == nil {
 		return errNoAppSKey.New()
 	}
+	// TODO: Cache unwrapped keys (https://github.com/TheThingsNetwork/lorawan-stack/issues/36)
```
adriansmares: That issue is closed and the references can be removed.

pkg/applicationserver/payload.go
```diff
@@ -94,74 +96,96 @@ func (as *ApplicationServer) decode(ctx context.Context, dev *ttnpb.EndDevice, u
...
+		session = dev.Session
+	case dev.PendingSession != nil && bytes.Equal(dev.PendingSession.SessionKeyID, downlink.SessionKeyID):
+		session = dev.PendingSession
+	default:
+		return errUnknownSession.New()
+	}
+	if session.GetAppSKey() == nil {
+		return errNoAppSKey.New()
+	}
+	// TODO: Cache unwrapped keys (https://github.com/TheThingsNetwork/lorawan-stack/issues/36)
```
adriansmares: IDEM.

pkg/messageprocessors/javascript/javascript.go
```diff
@@ -40,113 +40,177 @@ func New() messageprocessors.PayloadEncodeDecoder {
...
-	return env
+type encodeDownlinkInput struct {
+	Data  map[string]interface{} `json:"data"`
+	FPort *uint8                 `json:"fPort"`
+}
+
+type encodeDownlinkOutput struct {
+	Bytes    []uint8  `json:"bytes"`
+	FPort    *uint8   `json:"fPort"`
+	Warnings []string `json:"warnings"`
```
adriansmares, line 51: I see no usage for these warnings. Shouldn't we at least emit some events for these ?

adriansmares:
Good job ! I have only some small remarks.

pkg/messageprocessors/javascript/javascript.go
johanstokking: Yeah I was thinking what to do with this. Stay tuned.

johanstokking: 2f6888ec6

johanstokking:
```json
{
  "name": "as.up.data.decode.warning",
  "time": "2020-08-26T17:06:03.787144Z",
  "identifiers": [
    {
      "device_ids": {
        "device_id": "dev1",
        "application_ids": {
          "application_id": "app1"
        },
        "dev_eui": "1122334455667788",
        "join_eui": "0000000000000000"
      }
    }
  ],
  "data": {
    "@type": "type.googleapis.com/ttn.lorawan.v3.ApplicationUplink",
    "session_key_id": "AXQruu2BHxR68Zel+ZN1ag==",
    "f_port": 1,
    "f_cnt": 5,
    "frm_payload": "vu8=",
    "decoded_payload": {
        "bytes": [
              190,
              239
            ]
      },
    "decoded_payload_warnings": [
      "my test warning"
    ],
    "rx_metadata": [
      {
        "gateway_ids": {
          "gateway_id": "simulate"
        },
        "time": "2020-08-26T17:06:03.508552Z",
        "timestamp": 3584956232,
        "location": {
          "source": "SOURCE_REGISTRY"
        },
        "uplink_token": "CgwKCgoIc2ltdWxhdGUQyK64rQ0aDAj7rJr6BRC4npuMAiDA4quAq2g="
      }
    ],
    "settings": {
      "data_rate": {
        "lora": {
          "bandwidth": 125000,
          "spreading_factor": 12
        }
      },
      "coding_rate": "4/5",
      "frequency": "868500000",
      "timestamp": 3584956232,
      "time": "2020-08-26T17:06:03.508552Z"
    },
    "received_at": "2020-08-26T17:06:03.562813Z",
    "confirmed": true
  },
  "correlation_ids": [
    "as:up:01EGNVQMW5PN6F7YR6S3YPCA0F",
    "gs:conn:01EGNVQMMG00DZ6SQ6NS0PXJ3C",
    "gs:uplink:01EGNVQMNABPN41PK2HC4GDAH8",
    "ns:uplink:01EGNVQMNAC31GP3KSE6H8YA99",
    "rpc:/ttn.lorawan.v3.GsNs/HandleUplink:01EGNVQMNAC95K61P55C3S95KZ",
    "rpc:/ttn.lorawan.v3.GtwGs/LinkGateway:01EGNVQMKXN63FVC5TBZ05QCHJ"
  ],
  "origin": "Johan-2.local",
  "visibility": {
    "rights": [
    ]
  }
}
```

adriansmares (approved):

htdvisser (approved):
Code-owned files (basically just `api`) LGTM. Although strictly speaking it's breaking internal API, the fact that it's not used makes this an acceptable break.

Maybe good to set code owners on `pkg/doc`, `pkg/messageprocessors` and `pkg/scripting` sometime?

kschiffer (approved):

Issue #3142, opened 2020-08-25
Refactor Network Server test structure


rvolosatovs:
#### Summary

Make "unit"s bigger, test behaviors, not implementation specifics

#### Why do we need this?

To simplify development and testing.

#### What is already there? What do you see now?

Removal of quite some internal mocks in #3052 

#### What is missing? What do you want to see?

Most of `grpc_gsns_test.go`, `grpc_asns_test.go` and `downlink_test.go` should be integrated into flow tests only mocking the "outside world", all internal components should be using "real" implementations.

#### How do you propose to implement this?

One-by-one integrate tests into `networkserver_flow_test.go`

#### How do you propose to test this?

unit tests

#### Can you do this yourself and submit a Pull Request?

yes

rvolosatovs:
Refs #918

Issue #2732, opened 2020-06-15
Unit Test are too slow


htdvisser:
#### Summary

While reviewing #2730 I noticed that some of our unit tests are **really** slow. Slow tests slow down development, so let's see if we can speed things up a bit.

#### Steps to Reproduce

Just check CI.

#### What do you see now?

Here's the top 10:

Package | Duration
-- | --
pkg/gatewayserver | 65.747
pkg/applicationserver | 53.63
pkg/applicationserver/io/pubsub/provider/nats | 20.546
pkg/gatewayserver/io/udp | 15.849
pkg/networkserver | 14.465
pkg/applicationserver/io/pubsub | 10.511
pkg/applicationserver/io/pubsub/provider/mqtt | 10.322
pkg/identityserver | 8.578
pkg/identityserver/store | 5.378
pkg/applicationserver/io/web | 3.253

#### What do you want to see instead?

These tests don't all need to take this long.

#### Environment

CI

#### How do you propose to implement this?

- Let's start by looking at why some tests take so long
- Then we can see if it's really necessary for them to take so long
- Maybe we can run some test with `t.Parallel()`
- Maybe we can use the `-short` flag for some of the builds?

rvolosatovs:
`networkserver` expects developers to use `-short` during development, `go test` runs the whole test suite and that will take time - there's no way around that, since we're dealing with e.g. RX delays in flow tests

htdvisser:
I'm a bit disappointed that you don't want to take any responsibility here, @rvolosatovs. Can none of those test be run in parallel to speed things up? Can we really not use mocked time or something?

rvolosatovs:
> I'm a bit disappointed that you don't want to take any responsibility here, @rvolosatovs. Can none of those test be run in parallel to speed things up? Can we really not use mocked time or something?

There are several tests in NS that need to be rewritten, which will be done later - it's pointless to try to speed them up.
NS already uses mock time, hides slow tests behind a `-short` and runs tests in parallel.
`go test` in Network Server will be slow by design - in flow tests we need to wait 5 seconds for the join-accept to be delivered alone. E.g. if NS has a flow test for OTAA device, which performs a join, then sends an uplink and receives a downlink, it will take *at least* 6 seconds. And we're talking about just one test case - full NS testing suite tests implementation against various PHY version, MAC version and frequency plan combinations.
Please do not run `go test` during development, run `go test -short -failfast` instead - tests during development run fast.

[...]

PR #3170, opened 2020-08-31
Testing utilities, use in Network Server

rvolosatovs:
#### Summary

Various testing utilities used by https://github.com/TheThingsNetwork/lorawan-stack/pull/3052
Refs #2732 #3142 

#### Changes

- Add various test utlities used by https://github.com/TheThingsNetwork/lorawan-stack/pull/3052
- Introduce `RunTest`, `RunSubtest` and the like. There create a new test context, etc and set a mandatory timeout on the context (either a quite liberal default is set, or user-configured). A log statement is printed for long-running tests to ease debugging of hanging tests.
- I was able to checkout some changes from https://github.com/TheThingsNetwork/lorawan-stack/pull/3052 where this new functionality is used in NS, so I added it here.

#### Checklist

### DIFF
--- a/pkg/networkserver/adr_internal_test.go
+++ b/pkg/networkserver/adr_internal_test.go
@@ -15,6 +15,7 @@
 package networkserver
 
 import (
+	"context"
 	"fmt"
 	"math"
 	"math/rand"
@@ -23,7 +24,6 @@ import (
 
 	pbtypes "github.com/gogo/protobuf/types"
 	"github.com/smartystreets/assertions"
-	"go.thethings.network/lorawan-stack/v3/pkg/log"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 	"go.thethings.network/lorawan-stack/v3/pkg/util/test"
 	"go.thethings.network/lorawan-stack/v3/pkg/util/test/assertions/should"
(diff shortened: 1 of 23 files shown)


rvolosatovs:
Setting `prio/high` as dependant issue is `prio/high`

pkg/networkserver/adr_internal_test.go
```diff
@@ -465,25 +471,22 @@ func TestAdaptDataRate(t *testing.T) {
...
+			Func: func(ctx context.Context, t *testing.T, a *assertions.Assertion) {
+				dev := CopyEndDevice(tc.Device)
+				fp := FrequencyPlan(dev.FrequencyPlanID)
+				err := adaptDataRate(ctx, dev, LoRaWANBands[fp.BandID][dev.LoRaWANPHYVersion], ttnpb.MACSettings{})
+				if !a.So(err, should.Equal, tc.Error) {
+					t.Fatalf("ADR failed with: %s", err)
+				}
+				expected := CopyEndDevice(tc.Device)
+				if tc.DeviceDiff != nil {
+					tc.DeviceDiff(expected)
```
rvolosatovs, line 486: Note the race condition here. `tc` must be copied before `test.RunSubtest` is run for all parallel test cases. That is done in https://github.com/TheThingsNetwork/lorawan-stack/pull/3052 and very hard to cherry-pick. Meanwhile I disabled parallel execution in this PR.

adriansmares: ~You've mentioned that you have disabled parallel execution, but I still see `Parallel: true` on line 476.  Is this intended ?~ See the diff bellow.

I believe you can add `tc := tc` from #3052 and it won't cause an issue during merge/rebase (you will still have to `ok` the conflict tho, but at the same time you will have a diff from the disabled parallel execution).

adriansmares (approved):

johanstokking (approved):

pkg/networkserver/adr_internal_test.go
rvolosatovs: This has to be done everywhere and for some test cases it's not just `tc := tc` that's necessary. This requires manual process, which is quite time-consuming, even with editor macros. The files are too different to simply checkout at this point and search for necessary diffs/generating appropriate diffs with `--patch` is a waste of time

Issue #2312, opened 2020-04-06
Reverse AS-NS linking direction


adriansmares:
#### Background
_Can be skipped if you are familiar with https://github.com/TheThingsIndustries/lorawan-stack/issues/941_

In the current implementation of the communication between the Network Server and Application Server, the links are established by the Application Server, which links itself to the Network Server. Over this link, uplinks are sent as a stream of `ttnpb.ApplicationUp`.

The underlying issue with this approach is that Application Servers are inherently stateful due to linking. This introduces challenges with regards to load balancing between multiple Application Servers - the distribution of links over multiple instances, maintaining/migrating links when instances are down or a network partition occurs being just a number of issues introduced by the linking state.

#### Proposed solution

In order to tackle this issue, following our offline discussions, we've decided that we would like to reverse the direction of the links. This means that instead of the Application Server being a client to the Network Server, the Network Server becomes a client of the Application Server. This simplifies the logic for both components.

Since the Application Server may have clients (MQTT/gRPC) connect to it while not being linked by a Network Server, a new PubSub service should be used as a broker between Application Server instances. When a new application subscription (connection) arrives to an Application Server, a PubSub subscription is created for the uplinks of the said application. When an uplink is received by an Application Server instance, it is published to the topic of the application.

#### Implementation details:

- `HandleUplink` should be added to a new `NsAs` service that the Application Server implements.
```proto
service NsAs {
  rpc HandleUplink(ApplicationUp) returns (google.protobuf.Empty);
}
```
> We may want to revisit https://github.com/TheThingsNetwork/lorawan-stack/issues/1523 since we are already adding a new RPC.
- The NS uses this RPC in order to push uplinks to the AS.
- The NS uses `component.GetPeer`/`component.GetPeerCon` in order to receive the peer/connection to the AS ~based on the application identifiers~.
- When the AS receives an uplink, it triggers the required webhooks and application packages. Only the AS that receives the uplink should send the webhooks and run the application packages. 
The uplink is then published to the PubSub service, from where subscription based frontends can pick it up.
- AS downlink queue operations no longer use the link connection, but rather the cluster peer connection.
- AS can use `gocloud.dev/pubsub` for PubSub communication for inter-AS PubSub.

#### Required migrations
[...]

PR #3190, opened 2020-09-04
Remove AS-NS linking in the AS

adriansmares:
#### Summary

References #2312 

#### Changes

- Add `NsAs` service that allows NS to send uplinks to AS
- Relax `ApplicationInvalidatedDownlinks` validation, since we do use empty queues to signal `FCnt` changes
- Remove AS linking mechanism completely
- Retry Server removed (as links are no longer cancelled on update)
- `RxMetadata` and `TxSettings.DataRate.Modulation` added to AS tests, as we now go through the gRPC validation middleware, and these are required in order to actually make the calls
- Frontends can now choose if they subscribe only to locally processed traffic, or the cluster-wide traffic. Cluster traffic is routed over Pub/Sub (over Redis Pub/Sub in this PR), while local traffic is simply broadcasted/routed on send

#### Testing

As the NS part is not ready, I've adapted the `mockNS` to simulate a real NS, and the tests pass (but not much integration testing can be done without the NS).

##### Regressions

No behavioral changes are expected, but tests have been adapted to the new environment.

#### Notes for Reviewers

This is the initial groundwork for the AS-side removal. There are still some `TODOs` left that we should discuss, mainly about how to handle depreciation - I've added these in #2312 .

@johanstokking this is ready for an initial review.

#### Checklist


api/applicationserver.proto
```diff
@@ -106,6 +106,15 @@ service As {
   };
 }
 
+message NsAsHandleUplinkRequest {
```
johanstokking, line 117: I think that `ApplicationUps` is more consistent (i.e. we have `Clients` which has `repeated Client`)

api/applicationserver.proto
```diff
@@ -106,6 +106,15 @@ service As {
...
+  repeated ApplicationUp application_uplinks = 1 [(validate.rules).repeated.min_items = 1];
```
johanstokking: `application_ups`

pkg/applicationserver/applicationserver.go
```diff
@@ -253,47 +238,69 @@ func (as *ApplicationServer) Roles() []ttnpb.ClusterRole {
...
+
+	return nil
 }
 
-func skipPayloadCrypto(link *link, dev *ttnpb.EndDevice) bool {
+func (as *ApplicationServer) broadcastUp(ctx context.Context, up *ttnpb.ApplicationUp) error {
+	if err := as.localDistributor.SendUp(ctx, up); err != nil {
+		return err
+	}
+	return as.clusterDistributor.SendUp(ctx, up)
```
johanstokking: Are they sequential for a reason? Are these calls blocking? If so, consider goroutines and a wait group

pkg/applicationserver/distribution/distribution.go
```diff
@@ -0,0 +1,38 @@
...
+	"context"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/applicationserver/io"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+// PubSub sends upstream traffic from publishers to subscribers.
+type PubSub interface {
+	// SendUp publishes traffic to the subscribers.
+	SendUp(context.Context, *ttnpb.ApplicationUp) error
```
johanstokking: I would assume this is blocking because of the name and the fact it returns an error. So that's ok, but please do this concurrent then

johanstokking: You might want to rename these to `Publish()` so that they match better with their `Subscribe()` counterparts and it aligns better with their godoc

pkg/applicationserver/distribution/pubsub_distributor.go
```diff
@@ -0,0 +1,72 @@
...
+	pubsub        PubSub
+	subscriptions *subscriptionMap
+}
+
+// SendUp sends traffic to the underlying Pub/Sub.
+func (d *pubSubDistributor) SendUp(ctx context.Context, up *ttnpb.ApplicationUp) error {
+	return d.pubsub.SendUp(ctx, up)
+}
+
+var errMissingIdentifiers = errors.DefineUnimplemented("missing_identifiers", "subscriptions without identifiers are not supported")
```
johanstokking: That's a failed precondition

johanstokking (commented):
Great stuff, some minor comments only

Please append fixup commits so that subsequent reviews go a bit faster

api/applicationserver.proto
adriansmares: The reasoning behind the naming was that later we may want to add other fields to this request, so `*Request` fits this better in that case.

pkg/applicationserver/applicationserver.go
adriansmares: Error handling is better if they are sequential, and both just write to a buffered channel, so I'm more inclined to leave it as is.

Edit: ~actually Redis publish is blocking, but that's ok since it's not going to throttle the local call (which is non-blocking)~.
Edit2: Actually both are blocking, but I'm still not sold on spawning extra goroutines just for this. I want to do this change when we can pinpoint this as a throttle point.

johanstokking: I'm okay with optimizing this later.

api/applicationserver.proto
johanstokking: We can do that later and stay wire compatible. But I'm fine with either

johanstokking (approved):

htdvisser:
> Remove AS linking mechanism completely

Maybe I missed something, but I thought that we had to keep the old linking mechanism for supporting external Application Servers. Did we drop that?

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -238,6 +239,9 @@ var startCommand = &cobra.Command{
 			config.AS.Devices = &asredis.DeviceRegistry{
 				Redis: redis.New(config.Redis.WithNamespace("as", "devices")),
 			}
+			config.AS.Distribution.PubSub = &asdistribredis.PubSub{
+				Redis: redis.New(config.Redis.WithNamespace("as", "traffic")),
```
htdvisser: Shouldn't this use the "cache" Redis, since we're just using it for pub/sub, and not for storage?

pkg/applicationserver/grpc.go
```diff
@@ -37,64 +37,54 @@ func (as *ApplicationServer) SetLink(ctx context.Context, req *ttnpb.SetApplicat
...
-		log.FromContext(ctx).WithError(err).Warn("Failed to cancel link")
-	}
 	_, err := as.linkRegistry.Set(ctx, *ids, nil, func(link *ttnpb.ApplicationLink) (*ttnpb.ApplicationLink, []string, error) { return nil, nil, nil })
 	if err != nil {
 		return nil, err
 	}
 	return ttnpb.Empty, nil
 }
 
+var errUnimplemented = errors.DefineUnimplemented("unimplemented", "unimplemented")
```
htdvisser: This is too generic. The description could for example say something about linking not being necessary in this deployment.

sdk/js/generated/api-definition.json
```diff
@@ -736,6 +742,7 @@
         "provider",
         "provider.mqtt",
         "provider.mqtt.client_id",
+        "provider.mqtt.headers",
```
htdvisser: This looks like an unrelated change

adriansmares:
> > Remove AS linking mechanism completely
> 
> Maybe I missed something, but I thought that we had to keep the old linking mechanism for supporting external Application Servers. Did we drop that?

My understanding was that from the beginning external linking was abandoned as soon as we opened https://github.com/TheThingsNetwork/lorawan-stack/issues/2312 due to the related scalability issues it creates.

The newer mentality imo being that 'having your external AS' is achieved by disabling crypto at application/device level and operating these crypto operations yourself (Violet-like).

pkg/applicationserver/grpc.go
adriansmares: Linking is not going to be available in any deployment, but I can mention that the method has been deprecated.

johanstokking:
> The newer mentality imo being that 'having your external AS' is achieved by disabling crypto at application/device level and operating these crypto operations yourself

Yep, that's it

PR #3194, opened 2020-09-08
doc: Add common CA certs to reference

### LINKED ISSUE #3163: Add (up-to-date) Root CA PEM file to Documentation
#### Summary

We should add a PEM file to our documentation site containing the Root CAs for typical deployments.

#### Why do we need this ?

- Basic Station Gateways require a "trust" file instead of relying on the roots in the OS.
- Some MQTT clients require a "CA" file instead of relying on the roots in the OS.
- For many users it's too difficult to find out which CA(s) they need to include for a given deployment.

We currently link to the certificate of `Let's Encrypt Authority X3` signed by `DST Root CA X3` which expires `Mar 17 2021`. This would mean that all these gateways would have to be updated with a new root before then.

I think we should put a `ca.pem` file on our documentation site that contains the most used CAs in typical deployments (i.e. following the documented deployment methods).

- Determine a set of Root CAs to support. I'd propose `DST Root CA X3` (which we already link to), `ISRG Root X1` (for future Let's Encrypt), and the roots used to sign certificates on the major cloud providers.
	- Alternatively, we could take Mozilla's common root CAs, but that file might be too big for some gateways.
- Create a `.pem` file with those certificates.
- Link to it in the docs.

I think this is something for @benolayinka or @nejraselimovic.

Adding @johanstokking @KrishnaIyer @adriansmares @rvolosatovs for discussion

benolayinka:
#### Summary

Closes #3163 

#### Testing

Run locally

#### Checklist


doc/content/reference/root-certificates/ca.pem
```diff
@@ -0,0 +1,146 @@
+DST Root CA X3
+==============
```
KrishnaIyer, line 2: This is not a valid `.pem` file. You should remove those headings and also the extra new line. Here's a [reference](https://github.com/ARMmbed/mbedtls/blob/development/tests/data_files/test-ca_cat21.crt)

If you want to tag which certificate is which, I'd recommend adding an extra `md` file somewhere and listing these in order.

KrishnaIyer (changes requested):

doc/content/reference/root-certificates/ca.pem
benolayinka: I think anything outside the certificate tags is ignored https://stackoverflow.com/questions/19578812/comments-in-a-pem-file

KrishnaIyer: That's not standard behaviour. Some clients (like `OpenSSL`) do that but not all.

htdvisser: Anything outside the boundaries is indeed ignored: https://tools.ietf.org/html/rfc7468#section-2

htdvisser: This is fine: https://tools.ietf.org/html/rfc7468#section-5.2

KrishnaIyer (approved):

PR #3200, opened 2020-09-08
Event backends resilience improvements

### LINKED ISSUE #3191: Improve event backends resilience
#### Summary

Event backends (pub/subs) need to be more resilient in case of failures. Additionally, we need more logging presence, in order to be able to debug failures.

#### Why do we need this?

In order to automatically reconnect on connection failures, and in order to allow better debugging capabilities.

https://github.com/TheThingsNetwork/lorawan-stack/blob/7f5fb0457fb872e1a152a358916922b2701aa3c7/pkg/events/cloud/cloud.go#L45-L50

https://github.com/TheThingsNetwork/lorawan-stack/blob/7f5fb0457fb872e1a152a358916922b2701aa3c7/pkg/events/redis/redis.go#L62-L66

This has the following issues:
- If an error occurs, it is never reported anywhere (even though we could log it)
- If an error occurs, the stream is down forever

- Add a context to the Redis Pub/Sub, and at least log errors from it.
- Attempt to retry the receive / resubscribe if an error occurs, with the usual backoff and jitter.
- I would favor the following pattern for context cancellation, instead of checking `ctx.Err()`:
```golang
select {
case <- ctx.Done():
  return ctx.Err()
default:
}
```

#### Environment

`v3.9.1`

Add the missing retry mechanism and improve the logging, in places where errors can occur.

Trigger 'partitions' by killing Redis / the `gocloud` Pub/Sub.

Yes. cc @htdvisser @rvolosatovs

adriansmares:
#### Summary

Closes #3191 

#### Changes

- Backends are now initialized after a component has been created
- Backends now have their own context and always log their errors, instead of relying on `err == nil`
- Backends now subscribe inside a component task
- The Go Cloud backend now also subscribes on the first `Subscribe`, similar to the behavior of the Redis backend

#### Testing

1 . Started two stacks with the following extra configuration:
```yaml
events:
  backend: redis
  redis:
    address: localhost:6379
    database: 0
    namespace:
    - ttn
    - v3
```
2. Killed my Redis container, waited for a bit, started it again.
3. Checked that published events are still visible (I've updated an application and checked for the event, but anything would do).

##### Regressions

This changes how the Redis events pub/sub subscribes to events and improves logging capabilities. Already existing unit tests should cover this.

#### Checklist

### DIFF
--- a/cmd/internal/shared/events.go
+++ b/cmd/internal/shared/events.go
@@ -28,15 +28,15 @@ import (
 )
 
 // InitializeEvents initializes the event system.
-func InitializeEvents(ctx context.Context, c *component.Component, conf config.ServiceBase) error {
+func InitializeEvents(ctx context.Context, taskStarter component.TaskStarter, conf config.ServiceBase) error {
 	switch conf.Events.Backend {
 	case "internal":
 		return nil // this is the default.
 	case "redis":
-		events.SetDefaultPubSub(redis.NewPubSub(c, conf.Events.Redis))
+		events.SetDefaultPubSub(redis.NewPubSub(ctx, taskStarter, conf.Events.Redis))
 		return nil
 	case "cloud":
-		ps, err := cloud.NewPubSub(c, conf.Events.Cloud.PublishURL, conf.Events.Cloud.SubscribeURL)
+		ps, err := cloud.NewPubSub(ctx, taskStarter, conf.Events.Cloud.PublishURL, conf.Events.Cloud.SubscribeURL)
 		if err != nil {
 			return err
 		}
--- a/cmd/internal/shared/events.go
+++ b/cmd/internal/shared/events.go
@@ -28,21 +28,21 @@ import (
 )
 
 // InitializeEvents initializes the event system.
-func InitializeEvents(ctx context.Context, c *component.Component, config config.ServiceBase) error {
-	switch config.Events.Backend {
+func InitializeEvents(ctx context.Context, c *component.Component, conf config.ServiceBase) error {
+	switch conf.Events.Backend {
 	case "internal":
 		return nil // this is the default.
 	case "redis":
-		events.SetDefaultPubSub(redis.NewPubSub(c, config.Events.Redis))
+		events.SetDefaultPubSub(redis.NewPubSub(c, conf.Events.Redis))
 		return nil
 	case "cloud":
-		ps, err := cloud.NewPubSub(c, config.Events.Cloud.PublishURL, config.Events.Cloud.SubscribeURL)
+		ps, err := cloud.NewPubSub(c, conf.Events.Cloud.PublishURL, conf.Events.Cloud.SubscribeURL)
 		if err != nil {
 			return err
 		}
 		events.SetDefaultPubSub(ps)
 		return nil
 	default:
-		return fmt.Errorf("unknown events backend: %s", config.Events.Backend)
+		return fmt.Errorf("unknown events backend: %s", conf.Events.Backend)
 	}
 }
--- a/cmd/internal/shared/events.go
+++ b/cmd/internal/shared/events.go
@@ -36,7 +36,7 @@ func InitializeEvents(ctx context.Context, c *component.Component, config config
 		events.SetDefaultPubSub(redis.NewPubSub(c, config.Events.Redis))
 		return nil
 	case "cloud":
-		ps, err := cloud.NewPubSub(ctx, config.Events.Cloud.PublishURL, config.Events.Cloud.SubscribeURL)
+		ps, err := cloud.NewPubSub(c, config.Events.Cloud.PublishURL, config.Events.Cloud.SubscribeURL)
 		if err != nil {
 			return err
 		}
--- a/cmd/internal/shared/events.go
+++ b/cmd/internal/shared/events.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 
+	"go.thethings.network/lorawan-stack/v3/pkg/component"
 	"go.thethings.network/lorawan-stack/v3/pkg/config"
 	"go.thethings.network/lorawan-stack/v3/pkg/events"
 	"go.thethings.network/lorawan-stack/v3/pkg/events/cloud"
@@ -27,12 +28,12 @@ import (
 )
 
 // InitializeEvents initializes the event system.
-func InitializeEvents(ctx context.Context, config config.ServiceBase) (err error) {
+func InitializeEvents(ctx context.Context, c *component.Component, config config.ServiceBase) error {
 	switch config.Events.Backend {
 	case "internal":
 		return nil // this is the default.
 	case "redis":
-		events.SetDefaultPubSub(redis.NewPubSub(config.Events.Redis))
+		events.SetDefaultPubSub(redis.NewPubSub(c, config.Events.Redis))
 		return nil
 	case "cloud":
 		ps, err := cloud.NewPubSub(ctx, config.Events.Cloud.PublishURL, config.Events.Cloud.SubscribeURL)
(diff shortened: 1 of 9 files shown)


cmd/internal/shared/events.go
```diff
@@ -27,15 +28,15 @@ import (
 )
 
 // InitializeEvents initializes the event system.
-func InitializeEvents(ctx context.Context, config config.ServiceBase) (err error) {
+func InitializeEvents(ctx context.Context, c *component.Component, config config.ServiceBase) error {
```
rvolosatovs: nit; not a fan of one-letter name here, since all three parameter types start with `c`, IMO this is quite confusing. Consider `comp` ?
While we're at it, `config` should be called `conf` https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.9/DEVELOPMENT.md#single-word-entities
So maybe refactor these altogether?

pkg/events/cloud/cloud.go
```diff
@@ -19,81 +19,129 @@ import (
...
 
 // PubSub with Go Cloud backend.
 type PubSub struct {
 	events.PubSub
 
-	ctx          context.Context
-	contentType  string
-	topic        *pubsub.Topic
-	subscription *pubsub.Subscription
+	c           *component.Component
```
rvolosatovs: While I'm not a fan of the naming scheme above, I consider those nits.
This one is too important to be considered a nit. `PubSub.c` gives exactly 0 context on what this field is for. Might as well call it `x`, `y`, `foo` or `field`.
```suggestion
	component   *component.Component
```

pkg/events/cloud/cloud_test.go
```diff
@@ -91,7 +100,7 @@ func TestCloudPubSub(t *testing.T) {
 		if a.So(e.Identifiers(), should.NotBeNil) && a.So(e.Identifiers(), should.HaveLength, 1) {
 			a.So(e.Identifiers()[0].GetApplicationIDs(), should.Resemble, &appID)
 		}
-	case <-time.After(time.Second):
+	case <-time.After(timeout):
```
rvolosatovs: Create a context with timeout instead and just use `ctx.Done()` here.
Take a look at https://github.com/TheThingsNetwork/lorawan-stack/blob/d703d33381ce50b1966d44b003f7798e5bc7446a/pkg/util/test/test.go#L74-L165 as well

cmd/internal/shared/events.go
adriansmares: The catch is that our whole codebase uses `c` as the common name for a `*component.Component`. While I can go for `comp`, it will be the odd one out. If we decide to change the abbreviation, that's fine by me but I'd rather fix it everywhere (and not as part of this PR, since it's out of scope). 

I did change the struct members and `conf`.

johanstokking (commented):
Avoid high coupling by passing `*component.Component`; instead define a receiving interface with what you need (i.e. `StartTask()`)

adriansmares:
> 
> 
> Avoid high coupling by passing `*component.Component`; instead define a receiving interface with what you need (i.e. `StartTask()`)

Task configuration structure (`component.TaskConfig`) is in `component` too, and so do the defaults (backoff/jitter). We may consider moving task related code (`task.go` from `pkg/component`) into its own package in order to decouple those. I'd prefer to do this in a separate PR (rather than pumping this one).

johanstokking:
I mean something like this:

```diff
diff --git a/cmd/internal/shared/events.go b/cmd/internal/shared/events.go
index 1ba33d99b..8c3aa9955 100644
--- a/cmd/internal/shared/events.go
+++ b/cmd/internal/shared/events.go
@@ -33,7 +33,7 @@ func InitializeEvents(ctx context.Context, c *component.Component, conf config.S
 	case "internal":
 		return nil // this is the default.
 	case "redis":
-		events.SetDefaultPubSub(redis.NewPubSub(c, conf.Events.Redis))
+		events.SetDefaultPubSub(redis.NewPubSub(ctx, c, conf.Events.Redis))
 		return nil
 	case "cloud":
 		ps, err := cloud.NewPubSub(c, conf.Events.Cloud.PublishURL, conf.Events.Cloud.SubscribeURL)
diff --git a/pkg/events/redis/redis.go b/pkg/events/redis/redis.go
index 0fc95e826..2676625d7 100644
--- a/pkg/events/redis/redis.go
+++ b/pkg/events/redis/redis.go
@@ -29,17 +29,17 @@ import (
 )
 
 // WrapPubSub wraps an existing PubSub and publishes all events received from Redis to that PubSub.
-func WrapPubSub(wrapped events.PubSub, c *component.Component, conf ttnredis.Config) *PubSub {
+func WrapPubSub(ctx context.Context, wrapped events.PubSub, taskStarter component.TaskStarter, conf ttnredis.Config) *PubSub {
 	ttnRedisClient := ttnredis.New(&conf)
 	eventChannel := ttnRedisClient.Key("events")
-	ctx := log.NewContextWithFields(c.Context(), log.Fields(
+	ctx = log.NewContextWithFields(ctx, log.Fields(
 		"namespace", "events/redis",
 		"channel", eventChannel,
 	))
 	ctx, cancel := context.WithCancel(ctx)
 	return &PubSub{
 		PubSub:       wrapped,
-		component:    c,
+		taskStarter:  taskStarter,
 		ctx:          ctx,
 		cancel:       cancel,
 		client:       ttnRedisClient.Client,
@@ -48,15 +48,15 @@ func WrapPubSub(wrapped events.PubSub, c *component.Component, conf ttnredis.Con
 }
 
 // NewPubSub creates a new PubSub that publishes and subscribes to Redis.
-func NewPubSub(c *component.Component, conf ttnredis.Config) *PubSub {
-	return WrapPubSub(events.NewPubSub(events.DefaultBufferSize), c, conf)
+func NewPubSub(ctx context.Context, taskStarter component.TaskStarter, conf ttnredis.Config) *PubSub {
+	return WrapPubSub(ctx, events.NewPubSub(events.DefaultBufferSize), taskStarter, conf)
 }
 
 // PubSub with Redis backend.
 type PubSub struct {
 	events.PubSub
 
-	component    *component.Component
+	taskStarter  component.TaskStarter
 	ctx          context.Context
 	cancel       context.CancelFunc
 	eventChannel string
@@ -99,7 +99,7 @@ func (ps *PubSub) subscribeTask(ctx context.Context) error {
 // Subscribe implements the events.Subscriber interface.
 func (ps *PubSub) Subscribe(name string, hdl events.Handler) error {
 	ps.subOnce.Do(func() {
-		ps.component.StartTask(&component.TaskConfig{
+		ps.taskStarter.StartTask(&component.TaskConfig{
 			Context: ps.ctx,
 			ID:      "events_redis_subscribe",
 			Func:    ps.subscribeTask,

```

cmd/internal/shared/events.go
```diff
@@ -27,21 +28,21 @@ import (
 )
 
 // InitializeEvents initializes the event system.
-func InitializeEvents(ctx context.Context, config config.ServiceBase) (err error) {
-	switch config.Events.Backend {
+func InitializeEvents(ctx context.Context, taskStarter component.TaskStarter, conf config.ServiceBase) error {
```
rvolosatovs, line 31: in this case a short `ts` would be fine as well, since it's unambiguous

pkg/events/redis/redis_test.go
```diff
@@ -53,45 +55,60 @@ func redisConfig() ttnredis.Config {
...
 }
 
-func TestRedisPubSub(t *testing.T) {
-	a := assertions.New(t)
+var timeout = (1 << 10) * test.Delay
 
-	events.IncludeCaller = true
+func TestRedisPubSub(t *testing.T) {
+	test.RunTest(t, test.TestConfig{
+		Timeout: 4 * timeout,
```
rvolosatovs, line 72: note, that `RunTest` sets a default (very liberal) timeout, so it mostly makes sense to set `Timeout` only to increase the value, otherwise we risk introducing flakiness on slow machines

rvolosatovs (approved):

PR #3203, opened 2020-09-09
Fix mage dev:DBRestore target

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2605

#### Changes

- Flush redis when running `dev:DBRestore`
- Fix error handling in `dev:DBRestore`

#### Notes for Reviewers

Currently, we restore only the sql database between tests which causes problems when adding tests for devices since entities are partially stored in redis. We have to reset redis together with the sql database.

~Not sure how this should be implemented exactly. I picked the easiest option to just flush the database via `redis-cli`. It is worth mentioning that it is also possible to dump and restore redis state from `rdb` and `aol` files.~

#### Checklist

### DIFF
--- a/tools/mage/dev.go
+++ b/tools/mage/dev.go
@@ -136,13 +136,13 @@ func (Dev) DBRestore(ctx context.Context) error {
 	}
 	db, err := store.Open(ctx, databaseURI)
 	if err != nil {
-		return nil
+		return err
 	}
 	defer db.Close()
 
 	b, err := ioutil.ReadFile(filepath.Join(".cache", "sqldump.sql"))
 	if err != nil {
-		return nil
+		return err
 	}
 	return db.Exec(fmt.Sprintf(`DROP DATABASE %s;
 		CREATE DATABASE %s;
--- a/tools/mage/dev.go
+++ b/tools/mage/dev.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


kschiffer (approved):
LGTM

johanstokking:
@rvolosatovs

tools/mage/dev.go
```diff
@@ -151,6 +151,16 @@ func (Dev) DBRestore(ctx context.Context) error {
 	).Error
 }
 
+// RedisFlush deletes all keys from redis and restore the database to its initial state.
+func (Dev) RedisFlush() error {
+	if mg.Verbose() {
+		fmt.Println("Deleting all keys from redis")
+	}
+
+	_, err := sh.Exec(nil, nil, os.Stderr, "docker-compose", dockerComposeFlags("exec", "-T", "redis", "redis-cli", "FLUSHDB")...)
```
rvolosatovs: Honestly I'm not a fan of this and I think we should only be removing keys we need to be removing here (i.e. delete in a namespace).

rvolosatovs (approved):

tools/mage/dev.go
rvolosatovs: E.g. that's how we do it when testing https://github.com/TheThingsNetwork/lorawan-stack/blob/1282e8fc385341703db4e3fa18fc593af593da8c/pkg/util/test/redis.go#L121-L135

PR #3213, opened 2020-09-10
Fix application uplinks queue memory usage

### LINKED ISSUE #2908: Per-application uplink queue buffer size
#### Summary

Use the configured buffer size as a max cutoff point, but trim Redis streams to only store non-processed uplinks.
Note, we have a stream per-application
Refs #2904 

#### Why do we need this?

To minimize memory usage for things we don't need to store

Currently all applications use the same configurable constant uplink queue buffer size

Perhaps a more reasonable approach would actually be to have an absolute cap on this and dynamically adjust it at runtime.

#### Environment

production

 maintain the amount of enqueued uplinks and amount of processed uplinks and set stream size to `min(bufferSize, enqueued-processed)` or simply delete the processed entries, but I'm not sure of what the implications of this are on performance - we need to research and profile this. Redis streams are an append-only data structure - so if e.g. the cap is 20k, then we would store last 20k uplinks per application regardless of whether they are processed or not and that's not ideal.

integration testing

yes

rvolosatovs:
#### Summary

Closes #2908

#### Changes

- Delete processed application uplinks from queues
- Categorize application uplink streams, store at most 1 invalidation and store join-accepts separately. Thanks @adrian for the suggestion.
- ~Read at most 1 uplink from queue - this ensures maximum fail-safety and minimizes allocations in NS.~ Store the last invalidation FCnt per-device UID and avoid sending invalidations with FCnt lower than that value

#### Testing

Unit testing, test with a class c device

##### Regressions

From https://redis.io/commands/xdel:
> Eventually if all the entries in a macro-node are marked as deleted, the whole node is destroyed and the memory reclaimed. This means that if you delete a large amount of entries from a stream, for instance more than 50% of the entries appended to the stream, the memory usage per entry may increment, since what happens is that the stream will start to be fragmented. However the stream performances will remain the same.

The memory usage may slightly increase for a while, but I believe it will fix our issues after that

#### Checklist

### DIFF
--- a/pkg/networkserver/redis/application_uplink_queue.go
+++ b/pkg/networkserver/redis/application_uplink_queue.go
@@ -46,8 +46,16 @@ func NewApplicationUplinkQueue(cl *ttnredis.Client, maxLen int64, group, id stri
 	}
 }
 
-func (q *ApplicationUplinkQueue) uidUplinkKey(uid string) string {
+func (q *ApplicationUplinkQueue) uidGenericUplinkKey(uid string) string {
 	return q.redis.Key("uid", uid, "uplinks")
+}
+
+func (q *ApplicationUplinkQueue) uidInvalidationKey(uid string) string {
+	return q.redis.Key(q.uidGenericUplinkKey(uid), "invalidation")
+}
+
+func (q *ApplicationUplinkQueue) uidJoinAcceptKey(uid string) string {
+	return q.redis.Key(q.uidGenericUplinkKey(uid), "join-accept")
 }
 
 const payloadKey = "payload"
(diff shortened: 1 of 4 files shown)


pkg/networkserver/redis/application_uplink_queue.go
```diff
@@ -61,9 +69,26 @@ func (q *ApplicationUplinkQueue) Add(ctx context.Context, ups ...*ttnpb.Applicat
...
 
+		var (
+			streamID             string
+			maxLen, maxLenApprox int64
+		)
+		switch up.Up.(type) {
+		case *ttnpb.ApplicationUp_JoinAccept:
+			streamID = q.uidJoinAcceptKey(uid)
+			maxLenApprox = q.maxLen
+		case *ttnpb.ApplicationUp_DownlinkQueueInvalidated:
```
adriansmares, line 77: I don't think this is ok. We cannot coalesce this per application, since we may have multiple devices in the same app.  We'll still have to keep the limit here, but that's fine since we now `XDEL`.

pkg/networkserver/redis/application_uplink_queue.go
```diff
@@ -61,9 +69,26 @@ func (q *ApplicationUplinkQueue) Add(ctx context.Context, ups ...*ttnpb.Applicat
...
+		default:
+			streamID = q.uidGenericUplinkKey(uid)
+			maxLenApprox = q.maxLen
+		}
+
 		if err = q.redis.XAdd(&redis.XAddArgs{
-			Stream:       q.uidUplinkKey(uid),
-			MaxLenApprox: q.maxLen,
+			Stream:       streamID,
+			MaxLen:       maxLen,
```
adriansmares: Is it valid to specify both ? As far as I know the command syntax is to add `~` when you want to approximate, so how would specifying both work ?

pkg/networkserver/redis/application_uplink_queue.go
rvolosatovs: Ah yes indeed, good point!

pkg/networkserver/redis/application_uplink_queue.go
rvolosatovs: One of them is always zero (a.k.a unspecified in Go semantics)

johanstokking (approved):

johanstokking:
Please wait for @adriansmares

adriansmares (approved):

PR #3225, opened 2020-09-14
Add `ns-db prune` stack subcommand

rvolosatovs:
#### Summary

References #2908 

#### Changes

- Add `ns prune` stack subcommand

#### Testing

Start 3.9.1, generate a bunch of uplinks, switch to latest 3.9, generate a bunch of uplinks and run the tool

#### Checklist

### DIFF
--- a/cmd/ttn-lw-stack/commands/redis.go
+++ b/cmd/ttn-lw-stack/commands/redis.go
@@ -1,0 +1,110 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import (
+	"github.com/spf13/cobra"
+	nsredis "go.thethings.network/lorawan-stack/v3/pkg/networkserver/redis"
+	"go.thethings.network/lorawan-stack/v3/pkg/redis"
+)
+
+func rangeRedisKeysIteration(cl *redis.Client, cursor uint64, scanKey string, f func(k string) bool) (uint64, error) {
+	ks, cursor, err := cl.Scan(cursor, scanKey, 1).Result()
+	if err != nil {
+		return 0, err
+	}
+	for _, k := range ks {
+		if !f(k) {
+			return 0, nil
+		}
+	}
+	return cursor, nil
+}
+
+func rangeRedisKeys(cl *redis.Client, scanKey string, f func(k string) bool) error {
+	cursor, err := rangeRedisKeysIteration(cl, 0, scanKey, f)
+	if err != nil {
+		return err
+	}
+	for cursor > 0 {
+		cursor, err = rangeRedisKeysIteration(cl, cursor, scanKey, f)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+var (
+	nsDBCommand = &cobra.Command{
+		Use:   "ns-db",
+		Short: "Manage Network Server database",
+	}
+	nsDBPruneCommand = &cobra.Command{
+		Use:   "prune",
+		Short: "Remove unused Network Server data",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if config.Redis.IsZero() {
+				panic("Only redis is supported by this command")
+			}
+
+			logger.Info("Connecting to Redis database...")
+			cl := NewNetworkServerApplicationUplinkQueueRedis(*config)
+			var deleted uint64
+			defer func() { logger.Debugf("%d processed stream entries deleted", deleted) }()
+			return rangeRedisKeys(cl, nsredis.ApplicationUplinkQueueUIDGenericUplinkKey(cl, "*"), func(k string) bool {
+				gs, err := cl.XInfoGroups(k).Result()
+				if err != nil {
+					logger.WithError(err).Errorf("Failed to query groups of stream %q", k)
+					return true
+				}
+				for _, g := range gs {
+					if g.Name != "ns" {
+						logger.Errorf("Unexpected consumer group with name %q found for stream %q", g.Name, k)
+						continue
+					}
+					last := "-"
+					for {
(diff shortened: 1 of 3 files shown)


cmd/ttn-lw-stack/commands/redis.go
```diff
@@ -0,0 +1,115 @@
+// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
```
rvolosatovs: will update to 2020

KrishnaIyer:
Can you add some documentation here;
- Necessary config. I don't think we need to dump the entire `ns`  config object for this tool right?
- How often/under what conditions should this tool be run?

rvolosatovs:
>     * Necessary config. I don't think we need to dump the entire `ns`  config object for this tool right?

It requires configuration in `redis` namespace sufficient to create a client, that's it.
Logger configuration is required if you want to see output.

>     * How often/under what conditions should this tool be run?

When updating from 3.9.1 or below to a version above that. It should not be necessary in other cases, but that may change and I intend to add more GC subcommands here. It is designed to provide minimum load on the DB, it runs slow on purpose and it can be run anytime.

johanstokking:
This command is very specific for one use case whose is very conditional. Since we have the compatibility commitment on CLI, this is going to stay in V3 until the end of time.

We may not have application uplink queues anymore, let alone Redis.

Therefore, can we have a more general `ns-db prune` command, like we have `is-db` commands?

rvolosatovs:
> This command is very specific for one use case whose is very conditional. Since we have the compatibility commitment on CLI, this is going to stay in V3 until the end of time.
> 
> We may not have application uplink queues anymore, let alone Redis.
> 
> Therefore, can we have a more general `ns-db prune` command, like we have `is-db` commands?

We can extract this into a separate tool, but this is specific to Redis and application uplink queues. The intention is to add more subcommands to `ns` namespace.
`ns-db prune` apart from being ugly, requires then the CLI determine what storage backend is used automatically (error prone and unnecessary) and could potentially delete a lot of things we would not want to have deleted. The idea is to have, for example, utilities to handle stuck entries in streams (i.e. delivered to a consumer, but never acked due to failure), possible lock-contention etc. In any case, the whole point here is to be specific about Redis.

rvolosatovs:
So I'm fine with extracting this into standalone `ns-redis-tool` or whatever, but definitely not `*-lw-stack ns-db prune`

rvolosatovs:
Can't we just make this subcommand hidden, such that customers can't depend on it?

johanstokking:
> `ns-db prune` apart from being ugly

What is ugly about it? We have `is-db` already, and `prune` is quite common. I find it more clear from a user perspective than `gc`

> requires then the CLI determine what storage backend is used automatically

`ttn-lw-stack` already does that based on config. It doesn't matter that NS currently only uses Redis from the user's perspective, like `is-db` is also generic and is not `ttn-lw-stack is postgresql insert-user` or something.

> and could potentially delete a lot of things we would not want to have deleted

That's going to be the magic of `prune`. And we can always make confirmations or flags should that be necessary. See https://docs.docker.com/engine/reference/commandline/system_prune/

> utilities to handle stuck entries in streams [...], possible lock-contention etc

If we need CLI to recover for these specific things, we're doing something wrong. We are not going to increase ops burden to manually fix these things and document everything.

rvolosatovs:
> What is ugly about it? We have `is-db` already, and `prune` is quite common. I find it more clear from a user perspective than `gc`

`is-db` or `ns-db` is. It should be `is db <command>` or just `is create-admin-user`, for example. The `-db` part is completely irrelevant and non-intuitive

cmd/ttn-lw-stack/commands/redis.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+	}
+	return nil
+}
+
+var (
+	nsCommand = &cobra.Command{
+		Use:     "ns",
+		Aliases: []string{"ns-db"},
+		Short:   "Manage Network Server",
```
htdvisser: You may not agree with our historical choices for naming commands, but I think we can all agree that we prefer our commands to be consistent. So please instead write:

```
Use:   "ns-db",
Short: "Manage the Network Server database",
```

htdvisser (changes requested):

cmd/ttn-lw-stack/commands/redis.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+
+			logger.Info("Connecting to Redis database...")
+			cl := NewNetworkServerApplicationUplinkQueueRedis(*config)
+			var deleted uint64
+			defer func() { logger.Debugf("%d processed stream entries deleted", deleted) }()
+			return rangeRedisKeys(cl, nsredis.ApplicationUplinkQueueUIDGenericUplinkKey(cl, "*"), func(k string) bool {
+				gs, err := cl.XInfoGroups(k).Result()
+				if err != nil {
+					logger.WithError(err).Errorf("Failed to query groups of stream '%s'", k)
```
johanstokking: Isn't `gs` undefined at this point?

cmd/ttn-lw-stack/commands/redis.go
```diff
@@ -0,0 +1,109 @@
...
+				}
+				for _, g := range gs {
+					if g.Name != "ns" {
+						logger.Errorf("Unexpected consumer group with name '%s' found for stream '%s'", g.Name, k)
```
johanstokking:
```suggestion
						logger.Errorf("Unexpected consumer group with name %q found for stream %q", g.Name, k)
```

cmd/ttn-lw-stack/commands/redis.go
```diff
@@ -0,0 +1,109 @@
...
+						continue
+					}
+					last := "-"
+					for {
+						msgs, err := cl.XRangeN(k, last, g.LastDeliveredID, 1).Result()
+						if err != nil {
+							logger.WithError(err).Errorf("Failed to XRANGE over stream '%s'", k)
```
johanstokking:
```suggestion
							logger.WithError(err).Errorf("Failed to XRANGE over stream %q", k)
```

cmd/ttn-lw-stack/commands/redis.go
```diff
@@ -0,0 +1,109 @@
...
+							return true
+						}
+						var ids []string
+						for _, msg := range msgs {
+							ids = append(ids, msg.ID)
+							last = msg.ID
+						}
+						_, err = cl.XDel(k, ids...).Result()
+						if err != nil {
+							logger.WithError(err).Errorf("Failed to XDEL from stream '%s', continue to next stream", k)
```
johanstokking:
```suggestion
							logger.WithError(err).Errorf("Failed to XDEL from stream %q, continue to next stream", k)
```

htdvisser (approved):

rvolosatovs:
@johanstokking

johanstokking (approved):

PR #3227, opened 2020-09-14
Fix inconsistent KeyValueMap validation

### LINKED ISSUE #2515: Inconsistent KeyValueMap validation
#### Summary

The `KeyValueMap` component is used across the console, for attributes and also for webhook headers. Even though the key and value should be validated with `onBlur` event, it does not happen always.
 
#### Steps to Reproduce

It is hard to find a reliable reproduction procedure since it happens sometimes, but here is the usual one:
1. Go to application create webhook view
2. Add new header entry, fill both `key` and `value`
3. Delete `value` string and move from the input field
4. Observe no validation error, while if you never entered the `value` string and moved away from input there is a validation error that the `key-value` pair must be complete

Same happens with gateway and application attributes.

#### What do you see now?

Inconsistent validation

Reliable `onBlur` validation

#### Environment

```
go run ./cmd/ttn-lw-stack version

The Things Stack for LoRaWAN: ttn-lw-stack
Version:             3.8.1-dev
Go version:          go1.14
OS/Arch:             darwin/amd64
```

Revisit `KeyValueMap` component

Yes but maybe @bafonins wants to take a look at it as well

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2515

Before:
broken-gif

With this PR:
fixed-gif

#### Changes

- Make sure that `Field` receives complete value of a `KeyValueMap` entry.

#### Testing

Tested `KeyValueMap` behavior in webhook/application/gateway and device forms.

##### Regressions

`KeyValueMap` validation

#### Checklist

### DIFF
--- a/pkg/webui/components/key-value-map/entry.js
+++ b/pkg/webui/components/key-value-map/entry.js
@@ -28,6 +28,18 @@ const m = defineMessages({
 })
 
 class Entry extends React.Component {
+  _getKeyInputName() {
+    const { name, index } = this.props
+
+    return `${name}[${index}].key`
+  }
+
+  _getValueInputName() {
+    const { name, index } = this.props
+
+    return `${name}[${index}].value`
+  }
+
   @bind
   handleRemoveButtonClicked(event) {
     const { onRemoveButtonClick, index } = this.props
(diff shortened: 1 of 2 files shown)


pkg/webui/components/key-value-map/entry.js
```diff
@@ -66,7 +78,7 @@ class Entry extends React.Component {
           placeholder={valuePlaceholder}
           type="text"
           onChange={this.handleValueChanged}
-          onBlur={onBlur}
+          onBlur={this.handleBlur}
```
kschiffer, line 102: Shouldn't this handler also be applied to both inputs? Overall, this reminds me of the blur solution that we have for the Join EUI prefix input: https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.9/pkg/webui/console/containers/join-eui-prefixes-input/join-eui-prefixes-input.js#L106-L133

kschiffer (approved):

Issue #2884, opened 2020-07-11
Make event data views more resilient of errors


kschiffer:
#### Summary
Our event views are currently crashing altogether when there is any kind of error related to the event stream. In many situations, errors are related to the parsing of one single event. In these situations we can safely keep the overall event stream running.

#### Why do we need this?
Unnecessary crashing of the entire event data view.

#### What is already there? What do you see now?
We already make use of error boundaries that should help catching errors that result from faulty handling of event data on the frontend side.

#### What is missing? What do you want to see?
Before #2883, the SDK event streams would escalate some errors far too much in a way that caused the entire stream to be closed when it would actually not be necessary. Respectively, the Console also treats any error coming from the SDK stream as fatal and closes the event stream.

We need to treat event errors as non-fatal, since they mostly relate to the parsing of one specific data chunk. A failure does not warrant closing the event stream completely.

#### Environment
v3.8.5

#### How do you propose to implement this?
My idea is
1. Refactor event store logic to handle errors in the SDK event stream when chunk parsing fails, e.g.
`notify(listeners[EVENTS.ERROR], new Error('Dismissed unparseable message chunk'))`
2. Handle such errors by adding synthetic (frontend defined) "meta" events to the console event store
3. Distinguish such events accordingly by our event components to display a custom error message for them
`isSyntheticErrorEvent(event)` and `<SyntheticErrorEvent />`
4. Remove `GET_EVENT_MESSAGE_FAILURE` action type from the list of actions that close the event stream

@bafonins please let me know if you approve of such a solution.

#### How do you propose to test this?
We still need some ideas about how to properly end-to-end test event data views in the Console. This has already bitten us too often.

#### Can you do this yourself and submit a Pull Request?
Yes.

bafonins:
Adding custom events is a good idea. I was thinking about adding events for pause, resume and clear events when implementing https://github.com/TheThingsNetwork/lorawan-stack/pull/2477, but wasnt sure about the UI. Not sure about proposed implementation, but can take a look myself when working on https://github.com/TheThingsNetwork/lorawan-stack/issues/2231

kschiffer:
> Not sure about proposed implementation, but can take a look myself when working on [#2231](https://github.com/TheThingsNetwork/lorawan-stack/issues/2231)

I have a WIP with that solution. What exactly makes you not sure about it? Just want to reassure before I continue working on this.

bafonins:
[...]

PR #3211, opened 2020-09-10
Add unique event identifiers

adriansmares:
#### Summary

Closes #580 

#### Changes

- Add unique identifiers to events

#### Testing

Unit testing should cover this.

##### Regressions

No expected regressions.

#### Notes for Reviewers

I chose `UniqueID`/`unique_id` over `ID`/`id` since we already have `Identifiers`/`identifiers`, and I thought it would be confusing to have both (even though indeed they are documented). Naming can still be changed, of course. 

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #3229, opened 2020-09-14
Event fixes and improvements (Console)

### LINKED ISSUE #2887: Data views slow down and freeze after receiving many events
#### Summary

#### Steps to Reproduce

#### What do you see now?
Freezing data views

#### Environment
v3.5.8 and older

@bafonins what do you think?

Manual testing.

Yes.

### LINKED ISSUE #2980: Show all event info in expanded event details
#### Summary
The expanded details section of the console currently only shows the `data` prop of the event. However, there is other relevant data in the root of the event object that should be shown too.

#### Why do we need this?
The user should be able to inspect all available data of an event if he/she wishes to

Only the `data` prop of the event object being shown

The whole event object being shown instead.

@bafonins @johanstokking has there been any discussion that led to this decision that I might have missed?

#### Environment
v3.8.x

Just change the details component to show `event` instead of `event.data`

Trivial.

Yes.

### LINKED ISSUE #3098: Event list does not account for multiple events with same name and timestamp
#### Summary
The event list component is prone to same key warnings, happening when occasionally multiple events with the same name and timestamps are generated.

#### Steps to Reproduce
1. Go to gateway events (it's likelier to happen here but this can happen also for applicaiton/end device events)
2. Wait for some events to arrive
3. Eventually, see a duplicate key warning in the logs

#### What do you see now?
```
react-dom.development.js?7f13:546 Warning: Encountered two children with the same key, `2020-08-18T15:25:49.148698549Z-gs.status.forward`.
```

Event keys being composed in a way that is sufficiently unique.

#### Environment
v3.9.0

We need to find a way to derive a truly UID from the event content. If some events are actually identical then this will be impossible, but on the other hand raises the question of why such events exist in the first place

Manual.

Yes. I'll fix this as part of my work on #2826

kschiffer:
#### Summary
This PR introduces fixes and improvements to event components and  views. It solves performance issues (#2887), [layout issues](https://github.com/TheThingsIndustries/lorawan-stack/issues/2260#issuecomment-682060901)), moves to show the complete event data on expand (#2980) and introduces "synthetic" status events (see #2884).

Closes #2887
Closes #2980
References #2260
References #2884

<details><summary>Screenshot</summary>

image
</details>

#### Changes
- Refactor event global and console components into a single event component
- Determine the display of the data preview column by the data type
- Add specific preview components for most common data message type
- Move to showing the event details on the side, detached from the event row
- Introduce `react-window` to display the list to only render the currently visible events
- Add utilities and definitions of synthetic events to display frontend defined messages in the event stream
- Add definitions and display logic for synthetic status events (will be used in upcoming PR)
- Add utility to `<ErrorMessage />` to display the root cause together with the top level message
- Reorganize hardcoded icons

#### Testing

I did quite some manual testing using the staging environment with live data as well as storybook. End-to-end testing for events is still not possible since we would need a workflow to simulate events inside test.

##### Regressions

This PR could cause regressions regarding:
- Event display
- Event stream error handling

#### Notes for Reviewers
The diff unfortunately became quite large, even after branching off some of the changes. Since I decided to refactor a lot of the code fundamentally, all things considered, I don't think any more splitting up will do us a favor.

- This PR already contains definitions for synthetic events that are not yet used. I decided against adding the store updates to improve the error resilience (#2884) in this PR to reduce the diff so I'll add that in a separate PR once this is merged
- I removed the color coding of events that we currently have. As can be seen in [#2260](https://github.com/TheThingsIndustries/lorawan-stack/issues/2260#issuecomment-682060901) this can cause false signals of caution for events that are actually harmless. This needs some better consideration which I will track in a new issue.
- I decided to use the message type (`event.data['@type']`) to determine the data preview since I believe this is clearer and easier to maintain than going from the event name.
  - The previews that I chose for particular message types are based on what we currently display but these need to be determined more clearly based on some feedback and discussions
- This PR unifies the fragmented, repetitive and as I think somewhat too abstracted structure that we have for event components and containers into just one `<Events />` component. I believe this improves the maintainability of this core functionality of the console.
- I have removed the error display for event errors that were caught by the error boundary as these errors are frontend-specific and not helpful for the user. Instead these errors now get logged to sentry.

#### Checklist


pkg/webui/console/components/events/index.js
```diff
@@ -0,0 +1,125 @@
...
+import Widget from './widget'
+import m from './messages'
+import { getEventId } from './utils'
+
+import style from './events.styl'
+
+const Events = React.memo(({ events, scoped, onClear, entityId }) => {
+  const [paused, setPause] = useState(false)
+  const [focus, setFocus] = useState({ eventId: undefined, visible: false })
+  const onPause = useCallback(() => setPause(!paused), [paused])
```
bafonins:
```suggestion
  const onPause = useCallback(() => setPause(paused => !paused), [])
```

pkg/webui/console/components/events/list.js
```diff
@@ -0,0 +1,114 @@
...
+    />
+  </div>
+)
+
+EmptyMessage.propTypes = {
+  entityId: PropTypes.string.isRequired,
+}
+
+const EventsList = React.memo(
+  ({ events, paused, scoped, entityId, onRowClick, activeId }) => {
```
bafonins: `paused` is not used

pkg/webui/console/components/events/previews/gateway-status.js
```diff
@@ -13,31 +13,28 @@
...
-import style from './column.styl'
+import messages from '../messages'
+
+import DescriptionList from './shared/description-list'
+import JSONPayload from './shared/json-payload'
 
-const HeaderColumn = props => {
-  const { className, content } = props
+const GatewayStatusPreview = React.memo(({ event }) => {
+  const metrics = event.data.metrics
```
bafonins:
```suggestion
  const metrics = getByPath(event, 'data.metrics')
```

pkg/webui/console/components/events/utils/index.js
```diff
@@ -0,0 +1,92 @@
...
+
+import { getDeviceId, getApplicationId } from '@ttn-lw/lib/selectors/id'
+
+import DefaultPreview from '../previews/default'
+import DefaultSyntheticEventPreview from '../previews/synthetic/default'
+import SyntheticErrorEventPreview from '../previews/synthetic/error'
+
+import { eventIconMap, dataTypeMap, applicationUpMessages } from './definitions'
+
+export const getEventId = event => `${event.time}-${event.name}`
```
bafonins, line 23: Just as a note: with https://github.com/TheThingsNetwork/lorawan-stack/pull/3211 we can use the `unique_id` field.

bafonins (changes requested):
Looks great!

1. Horizontal scroll appears on live data with expanded side navigation. (Device)
2. For some reason I get two identical events appear (app/device/gateway):

Screenshot 2020-09-21 at 14 55 59

3. Cannot clear events when paused.
4. On screen sizes 622-670px - close button on the event sidebar adds horizontal scroll and disappears from the screen
5. Please resolve prop-types warnings in storybook

pkg/webui/console/components/events/utils/index.js
kschiffer: Yes. I'll add as soon as this is merged into `v3.10`.

kschiffer:
Resolved your points.

> 1.  For some reason I get two identical events appear (app/device/gateway):

Can you double-check that this one is related to this PR. Couldn't reproduce on latest `v3.9`.

bafonins (approved):

PR #3232, opened 2020-09-15
doc: Update HTTP API instructions

benolayinka:
#### Summary

Add note about adding fields as query parameters in HTTP API requests

#### Checklist

### DIFF
--- a/doc/content/reference/api/_index.md
+++ b/doc/content/reference/api/_index.md
@@ -4,3 +4,21 @@ description: ""
 ---
 
 This is the reference for the gRPC and HTTP APIs that {{% tts %}} exposes.
+
+<!--more-->
+
+## Field Masks
+
+{{% tts %}} APIs use field masks to specify a subset of fields that should be returned by a reading request, or to specify fields that should be updated in a writing request. See Google's [Protocol Buffers reference](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.FieldMask) for more information about field masks.
+
+## HTTP Queries
+
+Additional fields may be specified in HTTP requests by appending them as query string parameters. For example, to request the `name`, `description`, and `locations` of devices in an `EndDeviceRegistry.Get` request, add these fields to the `field_mask` field. To get this data for device `dev1` in application `app1`:
+
+```bash
+curl -i -H "Authorization: Bearer NNSXS.XXXXXXXXX" https://thethings.example.com/api/v3/applications/app1/devices/dev1?field_mask=name,description,locations
+```
+
+> Fields that are empty or zero are not returned in requests, even if they are specified in a field mask. 
+
+> If you're having trouble with the HTTP API, you can always inspect requests in the Console using your browser's inspector. All of the data displayed in the Console is pulled using HTTP API requests, and this should give you some insight in to how they are formed.


api/end_device_services.proto
```diff
@@ -34,8 +34,8 @@ service EndDeviceRegistry {
     };
   };
 
-  // Get the end device with the given identifiers, selecting the fields specified
-  // in the field mask.
+  // Get the end device with the given identifiers. To return additional fields,
+  // specify them in the field mask.
```
htdvisser: This is true for almost all RPCs, so if we write it here, we should write it everywhere. I'm fine with leaving it out of here, since the field mask is already described in the request message.

sdk/js/generated/api-definition.json
```diff
@@ -734,23 +734,8 @@
         "location_solved",
         "location_solved.topic",
         "provider",
-        "provider.aws_iot",
```
htdvisser: There should not be any changes in this file

doc/content/reference/api/_index.md
```diff
@@ -4,3 +4,13 @@ description: ""
 ---
 
 This is the reference for the gRPC and HTTP APIs that {{% tts %}} exposes.
+
+<!--more-->
+
+## HTTP Queries
+
+Additional fields may be specified in HTTP requests by appending them as query string parameters. For example, to request the `name`, `description`, and `locations` of devices in an `EndDeviceRegistry.Get` request, add these fields to the `field_mask` field:
```
htdvisser: I think we should write a little more about what field masks are, and how we use them in our APIs (because we use them on Get/List, but also on Update RPCs).

See also https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.FieldMask

htdvisser (changes requested):

doc/content/reference/api/_index.md
```diff
@@ -4,3 +4,13 @@ description: ""
...
+
+```bash
+GET /api/v3/applications/{end_device_ids.application_ids.application_id}/devices/{end_device_ids.device_id}?field_mask=name,description,locations
+```
```
johanstokking: This is not a bash script, is it? What is this?

doc/content/reference/api/_index.md
johanstokking: Also `EndDeviceRegistry` doesn't exist; they're separate per component

johanstokking (changes requested):

doc/content/reference/api/_index.md
benolayinka: > Also `EndDeviceRegistry` doesn't exist; they're separate per component

Ah I see what you mean, but the IS `EndDeviceRegistry` is just called `EndDeviceRegistry` https://thethingsstack.io/reference/api/end_device/#the-enddeviceregistry-service

doc/content/reference/api/_index.md
```diff
@@ -4,3 +4,15 @@ description: ""
 ---
 
 This is the reference for the gRPC and HTTP APIs that {{% tts %}} exposes.
+
+<!--more-->
+
+## Field Masks
+
+{{% tts %}} APIs use field masks to specify a subset of fields that should be returned by a `Get` request, or to specify fields that should be updated in an `Update` request. See Google's [protocol buffer reference](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.FieldMask) for more information about field masks.
```
johanstokking:
```suggestion
{{% tts %}} APIs use field masks to specify a subset of fields that should be returned by a reading request, or to specify fields that should be updated in a writing request. See Google's [protocol buffer reference](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.FieldMask) for more information about field masks.
```

It's not only `Get`; it can also be listing, searching etc. Same for `Update`; can be setting stuff.

johanstokking: Also, Protocol Buffers is a name so title case

doc/content/reference/api/_index.md
johanstokking: You're right!

doc/content/reference/api/_index.md
```diff
@@ -4,3 +4,15 @@ description: ""
...
+
+## HTTP Queries
+
+Additional fields may be specified in HTTP requests by appending them as query string parameters. For example, to request the `name`, `description`, and `locations` of devices in an `EndDeviceRegistry.Get` request, add these fields to the `field_mask` field:
+
+`GET /api/v3/applications/{end_device_ids.application_ids.application_id}/devices/{end_device_ids.device_id}?field_mask=name,description,locations`
```
johanstokking: But what is this still, an HTTP header?

Maybe a cURL example is clearest?

johanstokking (approved):
Awesome

htdvisser (approved):

PR #3235, opened 2020-09-15
Fix target dependency of install yarn target

kschiffer:
#### Summary
This quickfix PR fixes the dependency of the `installYarn` target of our mage tooling. Before this fix, `yarn` would be reinstalled whenever the `package.json` changed, which is not necessary and resulted in fragmented builds and the necessity to load and rebuild all JS dependencies from scratch which has cost me a lot of time already.

#### Changes
- Run `installYarn` only when the yarn folder is missing in `node_modules`

#### Testing

Manual testing.

#### Notes for Reviewers
Technically we could compare the versions of the yarn installation and the version value in `package.json` but I believe this is unnecessary overengineering given that the yarn version doesn't change frequently.

#### Checklist

### DIFF
--- a/tools/mage/js.go
+++ b/tools/mage/js.go
@@ -42,7 +42,6 @@ func yarnWorkingDirectoryArg(elem ...string) string {
 func installYarn() error {
 	ok, err := target.Path(
 		filepath.Join("node_modules", "yarn"),
-		"package.json",
 	)
 	if err != nil {
 		return targetError(err)


htdvisser (approved):
LGTM

tools/mage/js.go
```diff
@@ -40,18 +40,15 @@ func yarnWorkingDirectoryArg(elem ...string) string {
 }
 
 func installYarn() error {
-	ok, err := target.Path(
-		filepath.Join("node_modules", "yarn"),
-		"package.json",
```
rvolosatovs, line 45: just remove this line and revert everything else, we use `target` package in mage files, unless using `os` directly is absolutely necessary - this is not the case here.

rvolosatovs (changes requested):

bafonins (approved):

rvolosatovs (approved):

Issue #3220, opened 2020-09-11
Limit application downlink queue size


johanstokking:
#### Summary

Limit application downlink queue size

#### Why do we need this?

To avoid that there are too many downlink messages queued in the registry

#### What is already there? What do you see now?

- No limit on number of stored application downlink messages in the queue
- No limit on the number of downlink messages that you can enqueue

#### What is missing? What do you want to see?

- Limit the number of messages that are stored (configurable per deployment)
- Limit the number of messages in a downlink queue replace or push request

#### How do you propose to implement this?

- Check stored number of downlink messages. For migration, if this exceeds the maximum, take only the first N
- Add API level limit on number of downlink messages that can be pushed or replaced

#### How do you propose to test this?

Unit testing

#### Can you do this yourself and submit a Pull Request?

Can review

rvolosatovs:
closed by https://github.com/TheThingsNetwork/lorawan-stack/pull/3236

PR #3236, opened 2020-09-15
Introduce and respect downlink-queue-capacity in Network Server

rvolosatovs:
#### Summary

Introduce and respect downlink-queue-capacity in Network Server

#### Changes

- Set a capacity for downlink queue size per-session
- Set limit on the queued application downlinks that can be sent via RPC

#### Testing

The change is trivial and just visual inspection is enough and existing
unit tests are enough

#### Checklist


pkg/networkserver/config.go
```diff
@@ -126,6 +126,7 @@ type Config struct {
 	DefaultMACSettings     MACSettingConfig             `name:"default-mac-settings" description:"Default MAC settings to fallback to if not specified by device, band or frequency plan"`
 	Interop                config.InteropClient         `name:"interop" description:"Interop client configuration"`
 	DeviceKEKLabel         string                       `name:"device-kek-label" description:"Label of KEK used to encrypt device keys at rest"`
+	DownlinkQueueCapacity  uint64                       `name:"downlink-queue-capacity" description:"Maximum downlink queue size per-session"`
```
johanstokking: Why not `int`?

pkg/networkserver/grpc_asns.go
```diff
@@ -245,8 +246,13 @@ func matchQueuedApplicationDownlinks(ctx context.Context, dev *ttnpb.EndDevice,
 	return nil
 }
 
+var errDownlinkQueueCapacityExceeded = errors.DefineResourceExhausted("downlink_queue_capacity_exceeded", "Downlink queue capacity exceeded")
+
 // DownlinkQueueReplace is called by the Application Server to completely replace the downlink queue for a device.
 func (ns *NetworkServer) DownlinkQueueReplace(ctx context.Context, req *ttnpb.DownlinkQueueRequest) (*pbtypes.Empty, error) {
+	if n := len(req.Downlinks); uint64(n) > ns.downlinkQueueCapacity {
```
adriansmares: In theory this is `2 * ns.downlinkQueueCapacity`, since we may push both the `session` and the `pending_session` at the same time, correct ?

pkg/networkserver/grpc_asns.go
```diff
@@ -310,7 +319,10 @@ func (ns *NetworkServer) DownlinkQueueReplace(ctx context.Context, req *ttnpb.Do
 
 // DownlinkQueuePush is called by the Application Server to push a downlink to queue for a device.
 func (ns *NetworkServer) DownlinkQueuePush(ctx context.Context, req *ttnpb.DownlinkQueueRequest) (*pbtypes.Empty, error) {
-	if len(req.Downlinks) == 0 {
+	switch n := len(req.Downlinks); {
+	case uint64(n) > ns.downlinkQueueCapacity:
+		return nil, errDownlinkQueueCapacityExceeded.New()
+	case n == 0:
 		return ttnpb.Empty, nil
 	}
```
johanstokking: More readable and shorter version:

```suggestion
	if n := len(req.Downlinks); uint64(n) > ns.downlinkQueueCapacity {
		return nil, errDownlinkQueueCapacityExceeded.New()
	} else if n == 0 {
		return ttnpb.Empty, nil
	}
```

pkg/networkserver/grpc_asns.go
```diff
@@ -310,7 +319,10 @@ func (ns *NetworkServer) DownlinkQueueReplace(ctx context.Context, req *ttnpb.Do
 
 // DownlinkQueuePush is called by the Application Server to push a downlink to queue for a device.
 func (ns *NetworkServer) DownlinkQueuePush(ctx context.Context, req *ttnpb.DownlinkQueueRequest) (*pbtypes.Empty, error) {
-	if len(req.Downlinks) == 0 {
+	switch n := len(req.Downlinks); {
+	case uint64(n) > ns.downlinkQueueCapacity:
```
adriansmares: Same here (?)

pkg/networkserver/grpc_asns.go
rvolosatovs: Yep, I was making this change while you were reviewing https://github.com/TheThingsNetwork/lorawan-stack/pull/3236/commits/ad7ef7ccaafd952bb305c6b0e56b796334ec51a3

pkg/networkserver/grpc_asns.go
johanstokking: Another more readable version that has less of a show off factor of what you can do with switch/case:

```suggestion
	if uint64(len(req.Downlinks)) > ns.downlinkQueueCapacity {
		return nil, errDownlinkQueueCapacityExceeded.New()
	}
	if len(req.Downlinks) == 0 {
		return ttnpb.Empty, nil
	}
```

Note that `len()` is not counting anything

api/end_device.proto
```diff
@@ -51,7 +51,7 @@ message Session {
   // Time when the session started. Network Server only.
   google.protobuf.Timestamp started_at = 8 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
   // Queued Application downlink messages. Stored in Application Server and Network Server.
-  repeated ApplicationDownlink queued_application_downlinks = 9;
+  repeated ApplicationDownlink queued_application_downlinks = 9 [(validate.rules).repeated.max_items = 100000];
```
johanstokking: Don't set the limit here if it can be configured in NS to a potentially higher value that would not be effective.

That also really raises the question what the need is for `uint64`.

Set it in the request so that you cannot enqueue more than N downlink messages at once. Clients can adapt for that.

htdvisser: Bit confusing to mix 10K (like in changelog) and 100K (here)

htdvisser (approved):
LGTM

adriansmares:
Don't forget that this closes https://github.com/TheThingsNetwork/lorawan-stack/issues/3220.

What I do miss here is basically the 'pruning of sessions that are already too big' - or does `ns-db` handle that already ?

api/end_device.proto
rvolosatovs: Yeah, that was supposed to be in the request - wrong message

pkg/networkserver/grpc_asns.go
rvolosatovs: I don't care about showing off, what I do care about is readability and maintainability.
In my opinion `switch` is the most readable option when there are 2 or more cases.
IMO `else if` structure should be used only in very specific cases.

I refactored into `else if`

rvolosatovs:
> Don't forget that this closes #3220.
> 
> What I do miss here is basically the 'pruning of sessions that are already too big' - or does `ns-db` handle that already ?

I'm afraid we cannot just remove enqueued downlinks, since that could break applications.

api/messages.proto
```diff
@@ -343,5 +343,5 @@ message MessagePayloadFormatters {
 
 message DownlinkQueueRequest {
   EndDeviceIdentifiers end_device_ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false];
-  repeated ApplicationDownlink downlinks = 2;
+  repeated ApplicationDownlink queued_application_downlinks = 2 [(validate.rules).repeated.max_items = 100000];
```
rvolosatovs: ignore

pkg/networkserver/networkserver.go
```diff
@@ -160,18 +161,23 @@ type NetworkServer struct {
...
 
-const downlinkProcessTaskName = "process_downlink"
+const (
+	downlinkProcessTaskName = "process_downlink"
+	maxInt                  = int(^uint(0) >> 1)
+)
 
 // New returns new NetworkServer.
 func New(c *component.Component, conf *Config, opts ...Option) (*NetworkServer, error) {
+	panic(maxInt)
```
rvolosatovs: ignore

adriansmares (approved):

johanstokking (approved):

PR #3257, opened 2020-09-21
doc: Add device best practices

benolayinka:
#### Summary

Add best practices section for devices

#### Testing

Run locally

#### Checklist

### DIFF
--- a/doc/content/devices/best-practices/_index.md
+++ b/doc/content/devices/best-practices/_index.md
@@ -12,9 +12,9 @@ LoRaWAN devices should always comply to the [LoRaWAN specification](https://www.
 
 The goal of these best practices is to optimize individual devices (especially for battery consumption) and the network as a whole (to reliably and efficiently serve more devices with the same number of gateways).
 
-## Unnecessary Join Requests
+## Eliminate Unnecessary Join Requests
 
-The LoRaWAN specification warns especially against systematic rejoin in case of network failure. A device should keep the result of an activation in permanent storage if the device is expected to be turned off during its lifetime.
+The LoRaWAN specification warns specifically against systematic rejoin in case of network failure. A device should keep the result of an activation in permanent storage if the device is expected to be power-cycled during its lifetime.
 
 A device may temporarily lose connection with the network for many reasons: if the network server or gateways are suffering from an outage, if there’s no coverage in the area, etc. Whenever a device rejoins, it consumes the closest gateway’s airtime to emit the downlink - and if many devices in an area are rejoining at the same time (e.g. in case of a temporary gateway outage), this leads to network bloating in the area.
 
@@ -22,7 +22,9 @@ A device may temporarily lose connection with the network for many reasons: if t
 
 Limit the frequency of your transmissions to the minimum possible. Can you sample data once a day? Or even less?
 
-Encode messages as efficiently as possible. Shorter messages means shorter transmission time.
+Encode messages as efficiently as possible. Shorter messages mean shorter transmission time.
+
+Avoid confirmed uplink messages. Confirmed uplinks should only be used in the case where 100% assurance of transmission is necessary, e.g. alarms.
 
 The duty cycle of radio devices is often regulated by government. If this is the case, the duty cycle is commonly set to 1%, but make sure to check the regulations of your local government to be sure.
 
@@ -30,7 +32,13 @@ In Europe, duty cycles are regulated by section 7.2.3 of the ETSI EN300.220 stan
 
 Additionally, the LoRaWAN specification dictates duty cycles for the join frequencies, the frequencies devices of all LoRaWAN-compliant networks use for over-the-air activations (OTAA) of devices. In most regions this duty cycle is set to 1%.
 
+## Expect Packet Loss
+
+You should expect packet loss up to 10%. Implement Forward Error Correction if that's a problem.
+
 ## Synchronization, Backoff, and Jitter
+
+([See LoRaWAN Specification 1.0.3, line 1065](https://lora-alliance.org/sites/default/files/2018-07/lorawan1.0.3.pdf)).
 
 Synchronization of devices happens if end devices respond to a large-scale external event - for example, hundreds of end devices that are connected to the same power source and the power is switched off and on again, or hundreds of end devices that are connected to the same gateway, and the firmware of the gateway needs to be updated.
 
@@ -52,11 +60,43 @@ But what if you have another site with thousands of these devices? Then the 10 s
 
 An implementation like this prevents persistent failures of sites and the network as a whole and helps speed up recovery after outages.
 
-## Additional Best Practices
+## Use a Good Random Number Generator
 
-- Save device parameters between regular power cycles
-- Use a true random number generator (especially for jitter and channel hopping)
-- Use [OTAA]({{< ref "/reference/glossary#otaa" >}}) instead of [ABP]({{< ref "/reference/glossary#abp" >}})
-- Optimize [data rate]({{< ref "/reference/glossary#data-rate" >}})
-- Try to use [ADR]({{< ref "/reference/glossary#adr" >}}) for non moving devices, and ADR for moving devices
-- Avoid non-essential downlinks
+True randomness in a device's random number generator is especially important for preventing network congestion. If devices share a seed for a pseudorandom number generator, they will choose the same random numbers. Devices should use a unique seed such as the device address.
+
+Bad randomization will result in an uneven distribution of channels selected for random channel selection, causing subpar network performance. ([LoRaWAN Specification 1.0.3, line 244](https://lora-alliance.org/sites/default/files/2018-07/lorawan1.0.3.pdf)).
+
+Bad randomization will also result in transmission synchronization if devices respond to a large scale external event (for example, if they are all powered on at the same time).
+
+## Use ADR for Stationary Devices
+
+For devices that don't move, the LoRaWAN specification recommends allowing the Network Server to control the data rate to minimize power consumption.
+
+For moving devices, ADR should not be used since RF conditions will likely change, but since many moving devices are temporarily stationary, it is possible to save additional power by requesting ADR only during the time a device is stationary. ([LoRaWAN Specification 1.0.3, line 438](https://lora-alliance.org/sites/default/files/2018-07/lorawan1.0.3.pdf)).
+
+You may also use application specific knowledge to predict when ADR is appropriate. A tracking device can detect when it is moving, for example. A parked car sensor can detect when a parked car will affect RF conditions, and should fall back to another strategy. 
+
+## Use OTAA
+
+OTAA devices perform a join-procedure with the network, during which a dynamic Device Address is assigned and security keys are negotiated with the device. Activation by Personalization (ABP) requires hardcoding the Device Address as well as the security keys in the device, which is insecure. ABP also has the downside that devices can not switch network providers without manually changing keys in the device.
+
+## Power Cycles
+
+Devices should save network parameters between regular power cycles. This includes session parameters like `DevAddr`, session keys, `FCnt`, and nonces. This allows the device to easily Join, as keys and counters remain synchronized.
+
+Devices should also randomize initial power on delay (i.e. Join). See [Synchronization, Backoff, and Jitter](#synchronization-backoff-jitter)
+
+## Frame Counters
+
+Devices must increment the frame counter after each uplink and downlink. Devices should use 32 bit counters for FCntUp and FCntDwn to prevent replay attacks.
+
+## Ack
+
+It is possible that you will not receive an ACK for every confirmed uplink or downlink. A good rule of thumb is to wait for at least **three** missed ACK's to assume link loss.
+
+In the case of link loss, do the following:
+
+- Set TX power to the maximum allowed/supported, and try again
+- Decrease data rate step by step, and try again
+- Reset to default channels, and try again
+- Send periodic join requests with backoff
--- a/doc/content/devices/best-practices/_index.md
+++ b/doc/content/devices/best-practices/_index.md
@@ -1,0 +1,62 @@
+---
+title: "Best Practices"
+description: ""
+weight:
+---
+
+This section contains best practices for building connected devices which use {{% tts %}} as a Network Server.
+
+<!--more-->
+
+LoRaWAN devices should always comply to the [LoRaWAN specification](https://www.lora-alliance.org/lorawan-for-developers), with special regard to minimizing unnecessary join requests and duty cycle limitations.
+
+The goal of these best practices is to optimize individual devices (especially for battery consumption) and the network as a whole (to reliably and efficiently serve more devices with the same number of gateways).
+
+## Unnecessary Join Requests
+
+The LoRaWAN specification warns especially against systematic rejoin in case of network failure. A device should keep the result of an activation in permanent storage if the device is expected to be turned off during its lifetime.
+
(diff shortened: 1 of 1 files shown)


doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,53 @@
...
+
+The duty cycle of radio devices is often regulated by government. If this is the case, the duty cycle is commonly set to 1%, but make sure to check the regulations of your local government to be sure.
+
+In Europe, duty cycles are regulated by section 7.2.3 of the ETSI EN300.220 standard. This standard defines the following sub-bands and their duty cycles:
+
+- g (863.0 – 868.0 MHz): 1%
+- g1 (868.0 – 868.6 MHz): 1%
+- g2 (868.7 – 869.2 MHz): 0.1%
+- g3 (869.4 – 869.65 MHz): 10%
+- g4 (869.7 – 870.0 MHz): 1%
```
htdvisser: I'm quite sure this is outdated, there's a new spec.

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,53 @@
...
+weight:
+---
+
+This section contains best practices for building connected devices which use {{% tts %}} as a Network Server.
+
+<!--more-->
+
+LoRaWAN devices should always comply to the [LoRaWAN specification](https://www.lora-alliance.org/lorawan-for-developers), with special regard to minimizing join requests and duty cycle limitations.
+
+## Minimize Join Requests
```
htdvisser: I don't really like this heading as a "best practice". Joins are not bad, only unnecessary joins are.

htdvisser (commented):
Perhaps we should make clear what the goal of these best practices is. We want to optimize both individual devices (mostly battery consumption) and the network as a whole (being able to reliably and efficiently serve more devices with the same number of gateways).

Some more best practices:

- Try to keep parameters between regular power cycles
- Try to use a true random number generator (for jitter, but also channel hopping)
- Try to use OTAA instead of ABP
- Try to avoid synchronization (by using backoff and jitter)
- Try to limit transmissions (both frequency and duration, discuss duty cycle)
- Try to minimize payload size (by using efficient encoding)
- Try to optimize data rate
- Try to use ADR for static devices, and not use ADR for moving devices (devices can use sensors to detect if they are static or moving)
- Try to avoid downlink (joins, confirmed, application downlink) if not essential
- ...

doc/content/devices/best-practices/_index.md
johanstokking: Also this gets too much attention here. I don't think it adds much, just refer to the spec and mention the regions it applies to.

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,53 @@
...
+
+Let’s take an example device that starts in Join mode when it powers on, and reverts to Join mode after being disconnected from the network. There are hundreds of such devices in a field, and one gateway that covers this field.
+
+The power source for the devices is switched on, and the gateway immediately receives the noise of hundreds of simultaneous Join requests. LoRaWAN gateways can deal quite well with noise, but this is just too much, and the gateway can’t make any sense of it. No Join requests are decoded, so no Join requests are forwarded to the network and no Join requests are accepted.
+
+Exactly, or approximately 10 seconds later (the devices either have pretty accurate clocks, or they’re all equally inaccurate), the gateway again receives the noise of hundreds of simultaneous Join requests, and still can’t make anything of it. This continues every 10 seconds after that, and the entire site stays offline.
```
johanstokking:
```suggestion
If the retry interval is a fixed duration, e.g. 10 seconds, the gateway again receives the noise of hundreds of simultaneous Join requests, and still can’t make anything of it. This continues every 10 seconds after that, and the entire site stays offline.
```

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,62 @@
...
+An implementation like this prevents persistent failures of sites and the network as a whole and helps speed up recovery after outages.
+
+## Additional Best Practices
+
+- Save device parameters between regular power cycles
+- Use a true random number generator (especially for jitter and channel hopping)
+- Use [OTAA]({{< ref "/reference/glossary#otaa" >}}) instead of [ABP]({{< ref "/reference/glossary#abp" >}})
+- Optimize [data rate]({{< ref "/reference/glossary#data-rate" >}})
+- Try to use [ADR]({{< ref "/reference/glossary#adr" >}}) for non moving devices, and ADR for moving devices
+- Avoid non-essential downlinks
```
htdvisser: We avoid bullet lists in our documentation. I can help writing some text later this week.

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Synchronization, Backoff, and Jitter
+
+Synchronization of devices happens if end devices respond to a large-scale external event - for example, hundreds of end devices that are connected to the same power source and the power is switched off and on again, or hundreds of end devices that are connected to the same gateway, and the firmware of the gateway needs to be updated.
+
+Let’s take an example device that starts in Join mode when it powers on, and reverts to Join mode after being disconnected from the network. There are hundreds of such devices in a field, and one gateway that covers this field.
+
+The power source for the devices is switched on, and the gateway immediately receives the noise of hundreds of simultaneous Join requests. LoRaWAN gateways can deal quite well with noise, but this is just too much, and the gateway can’t make any sense of it. No Join requests are decoded, so no Join requests are forwarded to the network and no Join requests are accepted.
```
KrishnaIyer, line 47: I personally haven't encountered this scenario. The most probable alternative anecdote is when devices perform "unnecessary joins" and they all do it at the same time.

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+The duty cycle of radio devices is often regulated by government. If this is the case, the duty cycle is commonly set to 1%, but make sure to check the regulations of your local government to be sure.
+
+In Europe, duty cycles are regulated by section 7.2.3 of the ETSI EN300.220 standard.
+
+Additionally, the LoRaWAN specification dictates duty cycles for the join frequencies, the frequencies devices of all LoRaWAN-compliant networks use for over-the-air activations (OTAA) of devices. In most regions this duty cycle is set to 1%.
+
+## Synchronization, Backoff, and Jitter
+
+Synchronization of devices happens if end devices respond to a large-scale external event - for example, hundreds of end devices that are connected to the same power source and the power is switched off and on again, or hundreds of end devices that are connected to the same gateway, and the firmware of the gateway needs to be updated.
```
KrishnaIyer, line 43: It also happens if `N` devices send uplinks exactly every `x` minutes/hours. If this happens to be after a mass rejoin, then the network sees `N` uplinks every `x` minutes and no activity in between.

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+Let’s take an example device that starts in Join mode when it powers on, and reverts to Join mode after being disconnected from the network. There are hundreds of such devices in a field, and one gateway that covers this field.
+
+The power source for the devices is switched on, and the gateway immediately receives the noise of hundreds of simultaneous Join requests. LoRaWAN gateways can deal quite well with noise, but this is just too much, and the gateway can’t make any sense of it. No Join requests are decoded, so no Join requests are forwarded to the network and no Join requests are accepted.
+
+If the retry interval is a fixed duration, e.g. 10 seconds, the gateway again receives the noise of hundreds of simultaneous Join requests, and still can’t make anything of it. This continues every 10 seconds after that, and the entire site stays offline.
```
KrishnaIyer, line 49: It's more probable that some devices do get through but the loss rate is very high.

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Jitter
+
+This situation can be improved by using jitter. Instead of sending a Join request every 10 seconds, the devices send a Join request 10 seconds after the previous one, plus or minus a random duration of 0-20% of this 10 seconds. This jitter percentage needs to be truly random, because if all devices use the same pseudorandom number generator, they will still be synchronized, as they will all pick the same “random” number.
```
KrishnaIyer, line 53: > same pseudorandom number generator

Same external source of randomisation. If using an internal source, then the range of randomness (random number space) should be big enough to be actually random.

KrishnaIyer (changes requested):
Pretty good first pass. I think the tone of this should be different from our other guides that people can "follow" sequentially. This is more like a list of recommendations. 
Can you switch to passive voice for this?

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,62 @@
...
+weight:
+---
+
+This section contains best practices for building connected devices which use {{% tts %}} as a Network Server.
+
+<!--more-->
+
+LoRaWAN devices should always comply to the [LoRaWAN specification](https://www.lora-alliance.org/lorawan-for-developers), with special regard to minimizing unnecessary join requests and duty cycle limitations.
+
+The goal of these best practices is to optimize individual devices (especially for battery consumption) and the network as a whole (to reliably and efficiently serve more devices with the same number of gateways).
```
rvolosatovs, line 13:
```suggestion
The goal of these best practices is to optimize various aspects of individual device operation (with emphasis on minimizing power consumption) and the performance of the network as a whole (to reliably and efficiently serve more devices with the same number of gateways).
```

I think the explanation of network optimizations can also be rephrased to sound more professional. Perhaps, talk about scalability in regards to the gateway count?
Also, maybe a good idea to split this up in a proper paragraph and clearly state the properties we optimize for instead of putting it in parentheses?

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Unnecessary Join Requests
+
+The LoRaWAN specification warns especially against systematic rejoin in case of network failure. A device should keep the result of an activation in permanent storage if the device is expected to be turned off during its lifetime.
```
rvolosatovs:
```suggestion
The LoRaWAN specification advises specifically against periodic rejoin in case of network failure. A device should keep the result of an activation in permanent storage if the device is expected to be power-cycled during its lifetime.
```

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+A device may temporarily lose connection with the network for many reasons: if the network server or gateways are suffering from an outage, if there’s no coverage in the area, etc. Whenever a device rejoins, it consumes the closest gateway’s airtime to emit the downlink - and if many devices in an area are rejoining at the same time (e.g. in case of a temporary gateway outage), this leads to network bloating in the area.
```
rvolosatovs, line 19:
```suggestion
A device may temporarily lose connection with the network for many reasons: if the network server or gateways are suffering from an outage, if there’s no coverage in the area, etc. Whenever a device rejoins, it consumes the closest gateway’s airtime to emit the downlink - and if many devices in an area are rejoining at the same time (e.g. in case of a temporary gateway outage), this leads to network congestion in the area.
```

The airtime we care about the most is indeed the downlink, but there's no downlink in any of the cases you list at the start (i.e. if NS is down, if gateways are down or there's no coverage)

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Limit Transmission Length, Payload Size, and Duty Cycle
+
+Limit the frequency of your transmissions to the minimum possible. Can you sample data once a day? Or even less?
```
rvolosatovs, line 23:
```suggestion
Limit the periodicity of transmissions to the minimum. Can you sample data once a day? Or even less?
```

frequency is confusing, because all packets are transmitted on particular frequency.

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+Encode messages as efficiently as possible. Shorter messages means shorter transmission time.
```
rvolosatovs:
```suggestion
Optimize message encoding for size. Shorter messages mean shorter transmission time.
```

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+The duty cycle of radio devices is often regulated by government. If this is the case, the duty cycle is commonly set to 1%, but make sure to check the regulations of your local government to be sure.
+
+In Europe, duty cycles are regulated by section 7.2.3 of the ETSI EN300.220 standard.
```
rvolosatovs, line 31:
```suggestion
For example, in Europe, duty cycles are regulated by section 7.2.3 of the ETSI EN300.220 standard.
```

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+Additionally, the LoRaWAN specification dictates duty cycles for the join frequencies, the frequencies devices of all LoRaWAN-compliant networks use for over-the-air activations (OTAA) of devices. In most regions this duty cycle is set to 1%.
```
rvolosatovs, line 33:
```suggestion
Additionally, the LoRaWAN specification dictates duty cycles for the join frequencies - the frequencies LoRaWAN-compliant devices use for over-the-air activation (OTAA). In most regions this duty cycle is set to 1%.
```

doc/content/devices/best-practices/_index.md
```diff
@@ -0,0 +1,62 @@
...
+
+An implementation like this prevents persistent failures of sites and the network as a whole and helps speed up recovery after outages.
+
+## Additional Best Practices
+
+- Save device parameters between regular power cycles
+- Use a true random number generator (especially for jitter and channel hopping)
+- Use [OTAA]({{< ref "/reference/glossary#otaa" >}}) instead of [ABP]({{< ref "/reference/glossary#abp" >}})
+- Optimize [data rate]({{< ref "/reference/glossary#data-rate" >}})
+- Try to use [ADR]({{< ref "/reference/glossary#adr" >}}) for non moving devices, and ADR for moving devices
```
rvolosatovs:
```suggestion
- Try to use [ADR]({{< ref "/reference/glossary#adr" >}}) for immobile devices
```
What does `, and ADR for moving devices` mean?

Also, see the related wording/explanation in the spec (1.0.3 in this case)
2020-09-23-23:34:53-screenshot

rvolosatovs (changes requested):
It's a good start, but I think we can greatly improve the terminology and wording used. In my opinion, in it's current state this document is fine for a hobbyist project, but we need to present a more professional image.
Please also align this with terminology/wording used in the LoRaWAN spec. We can actually take a lot of best-practices straight out-of-there, or even make references to the spec.

benolayinka:
> It's a good start, but I think we can greatly improve the terminology and wording used. In my opinion, in it's current state this document is fine for a hobbyist project, but we need to present a more professional image.
> Please also align this with terminology/wording used in the LoRaWAN spec. We can actually take a lot of best-practices straight out-of-there, or even make references to the spec.

Appreciate the feedback. In general, I never want to substitute a more complex word for a simpler one with exactly the same meaning, especially not to sound more professional - rather the opposite. Documentation should be friendly and as accessible to as many people as possible. Here's a nice example -> https://plainlanguage.gov/guidelines/words/use-simple-words-phrases/

doc/content/devices/best-practices/_index.md
benolayinka: i'll read the spec and try to fill them out. it's good for me anyway.

htdvisser:
@avbentem just mentioned https://lora-developers.semtech.com/library/tech-papers-and-guides/the-book/joining-and-rejoining on the forum. Maybe good to add some cross-references. And perhaps there's some more useful recommendations we can add.

benolayinka:
Replaced by https://github.com/TheThingsIndustries/lorawan-stack-docs/pull/11

PR #3262, opened 2020-09-22
Fix NS panic when RX2 parameters cannot be computed

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2360

#### Changes

- Fix NS panic when RX2 parameters cannot be computed

#### Checklist

### DIFF
--- a/pkg/networkserver/downlink.go
+++ b/pkg/networkserver/downlink.go
@@ -1089,7 +1089,7 @@ func (ns *NetworkServer) attemptClassADataDownlink(ctx context.Context, dev *ttn
 		transmitAt time.Time
 		maxDR      band.DataRate
 	)
-	if attemptRX1 && rx1DRIdx > dev.MACState.CurrentParameters.Rx2DataRateIndex {
+	if attemptRX1 && rx1DRIdx > dev.MACState.CurrentParameters.Rx2DataRateIndex || !attemptRX2 {
 		transmitAt = slot.RX1()
 		maxDR = rx1DR
 	} else {


adriansmares (approved):
So the `rx2DR` was `nil` due the invalid data rate index, which lead to the nil `maxDR` which caused the panic ?

rvolosatovs:
> So the `rx2DR` was `nil` due the invalid data rate index, which lead to the nil `maxDR` which caused the panic ?

Exactly.

Regarding the RX1 DR question, which you since removed - we'd return on https://github.com/TheThingsNetwork/lorawan-stack/blob/96ca2a2ee7f41b071f94db61a5b4f78be77e6723/pkg/networkserver/downlink.go#L1076-L1085, but I guess you already figured it out

johanstokking:
This doesn't fully close https://github.com/TheThingsIndustries/lorawan-stack/issues/2360, as it doesn't recover panics. Is there a reason not to recover panics if something fails because of some edge case? cc @htdvisser

rvolosatovs:
The panics should not happen. `panic` == execution cannot continue and the only sensible option is to fail. Everything else should be handled as errors.

FTR, if NS would recover here, it'd end up in an infinite loop and we don't want that.

johanstokking:
> The panics should not happen. `panic` == execution cannot continue and the only sensible option is to fail.

"execution" is scoped to handling the uplink message, and "to fail" is to drop the uplink message. That can be achieved by recovering the panic and logging it so we can fix it. But it doesn't unnecessarily take down the entire Network Server and throws away everything that was in progress.

rvolosatovs:
So if we're talking about recovering from bugs in this particular case - sure, the downlink task can be wrapped in a `recover` block, just like we do with RPCs. It makes a lot of sense to do that, since the panic then is specific to the device. But we should add a recover to the whole NS itself.

htdvisser:
You can't `recover` from uncaught panics in another goroutine, but I think it would be good if we could recover in some scoped/isolated places. We already have recovery in an uplink RPC (as gRPC middleware), so this would be about recovering the downlink task. Perhaps it would even make sense to build it into `pkg/component/tasks.go` directly, so that it applies to any background task.

rvolosatovs:
> You can't `recover` from uncaught panics in another goroutine, but I think it would be good if we could recover in some scoped/isolated places. We already have recovery in an uplink RPC (as gRPC middleware), so this would be about recovering the downlink task. Perhaps it would even make sense to build it into `pkg/component/tasks.go` directly, so that it applies to any background task.

Yep, that's exactly what I did in https://github.com/TheThingsNetwork/lorawan-stack/pull/3264

Issue #2542, opened 2020-05-14
Remove `No prefix` field if nothing is to be shown


laurensslats:
#### Summary

No-prefix - Unclear what it is and how it can be set.

#### Why do we need this?

Only if people set it by themselves, which only happens occasionally.

#### What is already there? What do you see now?

![image](https://user-images.githubusercontent.com/17290639/81901135-54f6b280-95be-11ea-8af9-bc588bf9a2fe.png)

#### What is missing? What do you want to see?

Remove the `No prefix` field if there's nothing to show.

#### Environment

v3.7.2

#### How do you propose to implement this?

Only show it when there's a known prefix.

#### Can you do this yourself and submit a Pull Request?

@kschiffer

bafonins:
I dont think we should hide the prefixes select.

1. We fetch prefixes from JS and depending on the users internet connection those can arrive after some delay. This can result in initially hiding the select and then showing it.
2. There is always at least one prefix available - the empty prefix, `0000000000000000`. During the last call we discussed that we want the users to be aware that empty `join_eui` is a viable,allowed option that developers can use for testing devices (cc @johanstokking)

That being said, I dont think we should hide the select.

kschiffer:
I disagree:

1.  If the Join EUI prefixes are a prerequisite to be able to enter the Join EUI, we need to fetch those before activating the Join EUI field as a whole. This is regardless of whether there actually are prefixes available or not and is a combined constraint of the whole Join EUI field (which then _can_ consist select and text-input)
2.  2.A select input element without options (or just one option) makes no sense and only confuses users, which is the reason this issue was created

laurensslats:
To consider:
No-one uses pre-fixes as of now nor requested this feature. This might have to do with the fact that nearly no-one knows what this is, nor how it can be set. 

> There is always at least one prefix available - the empty prefix, `0000000000000000`. During the last call we discussed that we want the users to be aware that empty join_eui is a viable, allowed option that developers can use for testing devices

If we go down this path, we need to provide 2 pre-fix options at any time (I guess the following?)
Default `26000000/7`
Experimental nodes: `00000000/7`

In addition we need documentation of:
1. What are pre-fixes 
2. When to use `default`, when to use `experimental nodes` and why
3. How to manually set pre-fixes 

_I'm personally not aware of the benefits of using an empty pre-fix over the default one._

If we don't have the above documentation in place, we only cause confusion and it asks resources from our support team because people will ask questions about it.

[...]

PR #3269, opened 2020-09-23
Remove `No prefix` field if nothing is to be shown 

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2542

#### Changes

- Hide select when no prefixes available

#### Testing

1. Without any prefixes configured in js
2. With prefixes configured in js

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- Join EUI prefixes select on empty prefixes configuration in Join Server.
+
 ### Fixed
 
 - Broken link to setting device location in the device map widget.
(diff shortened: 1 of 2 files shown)


kschiffer (approved):
LGTM. 

Regarding the fetching behavior, I think that ultimately it would be better to fetch the JoinEUI prefixes on the form level already and pass the results directly to `<JoinEuiPrefixesInput />`. Right now, the user is able to perform inputs before the prefixes arrived. This could be prevented by also disabling the form when it is in fetching state, which would however result in styling flickers on initial render. 

If the prefixes are loaded on the initial step then it is very likely that they have arrived by the time the user proceeded to the `Basic Settings` step.

I suggest merging this as it is but keeping the issue open, so we can address this later, unless you want to address this right away.

PR #3277, opened 2020-09-23
Join Server Application Activation Setting Registry

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2322

#### Changes

- Add Join Server Application Activation Setting Registry
- Add `cryptoutil.UnwrapKeyEnvelope`
- Export `test.FormatError`
- Use `*ttnpb.KeyEnvelope` in `cryptoutil`
- Enforce that KEK is used if configured. Case when KEK label is non-empty and the KEK is missing in the Join Server now causes an error.

#### Testing

~In-progress~

#### Checklist


rvolosatovs:
@johanstokking feel free to take a look at the implementation already(it's pretty much final, unless I find bugs during testing), while I'm busy with testing

rvolosatovs:
@johanstokking the existing tests fail, because they depend on the fact that if configured KEK is missing, JS stores keys in plain-text. I do not think that it is in any way a sensible behavior - JS should never store keys in plain-text, especially if it's configured not to do so.
Do you agree with the proposed change?
Shall I refactor the existing tests then?

johanstokking:
@rvolosatovs if there's no encryption key for storage configured, what do you propose without breaking configuration?

The reason JS may store keys in the clear is that for some simple and small deployments, protecting the disk suffices. If you're using the Getting Started guide to setup a local network with a locally running Redis, there's very little added value in using KEKs.

rvolosatovs:
> @rvolosatovs if there's no encryption key for storage configured, what do you propose without breaking configuration?
> 
> The reason JS may store keys in the clear is that for some simple and small deployments, protecting the disk suffices. If you're using the Getting Started guide to setup a local network with a locally running Redis, there's very little added value in using KEKs.

If no key is configured - no encryption, however currently JS simply ignores the fact that key is configured if not present and that is the issue I'm pointing out here. In my opinion it's a bug, so it's fine to fix it here.

rvolosatovs:
So I mean things like these https://github.com/TheThingsNetwork/lorawan-stack/blob/b6b33e43d8232819f6f5046caa00544ea1137d10/pkg/joinserver/joinserver.go#L451-L454
I propose simply:
- If KEK label is configured - encrypt (and if corresponding KEK is missing - error)
- If KEK label is not configured - don't encrypt

johanstokking:
I see. Yes, that makes sense.

rvolosatovs:
> I see. Yes, that makes sense.

Ok, so with this approach, currently local joins fail with

```
  INFO Join not accepted                        dev_eui=DEADBEEF01020304 error=error:pkg/joinserver:wrap_key (failed to wrap key) error_cause=error:pkg/crypto/cryptoutil:kek_not_found (KEK with label `ns:localhost` not found) grpc_method=HandleJoin grpc_service=ttn.lorawan.v3.NsJs join_eui=01020304DEADBEEF label=ns:localhost namespace=joinserver request_id=01EMF5GVCHCKTYDS8EWW7PPWSR
```

What shall we do here then?
Since address-bound KEK is the default, I think it makes sense to check if address-bound KEK is present in the vault and if it is, encrypt - otherwise send plaintext. That means we'd have to extend `KeyVault` to expose `Contains(kekLabel string) (bool, error)` method

johanstokking (approved):

johanstokking (changes requested):
I was too fast here, sorry about that.

> @rvolosatovs Since address-bound KEK is the default, I think it makes sense to check if address-bound KEK is present in the vault and if it is, encrypt - otherwise send plaintext. That means we'd have to extend `KeyVault` to expose `Contains(kekLabel string) (bool, error)` method

I'm not sure if we can and should add `Contains()` as some implementations may not support this or become quite expensive. Currently, wrapping is skipped when the key vault says "not found", which I believe is good behavior.

I don't entirely understand how this relates with AAS: that is not addressed based. If that application KEK doesn't exist, indeed we need to error, i.e. not continue and return plaintext.

pkg/joinserver/grpc_application_activation_settings_registry.go
```diff
@@ -0,0 +1,117 @@
...
+
+	pbtypes "github.com/gogo/protobuf/types"
+	"go.thethings.network/lorawan-stack/v3/pkg/auth/rights"
+	"go.thethings.network/lorawan-stack/v3/pkg/crypto/cryptoutil"
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+type applicationActivationSettingsRegistryServer struct {
+	JS       *JoinServer
```
adriansmares, line 28: Isn't it better to pass the registry and key vault here instead of tying it to the whole `JoinServer` struct ?

adriansmares (approved):
LGTM but please wait for @johanstokking

pkg/joinserver/grpc_application_activation_settings_registry.go
rvolosatovs: 1. no
2. that's consistent with all the gRPC servers in this package

htdvisser (approved):

johanstokking:
@rvolosatovs is this ready for another pass? If so, request review.

I want the API in for 3.10, so adding a release blocker.

pkg/joinserver/grpc_application_activation_settings_registry.go
```diff
@@ -0,0 +1,117 @@
...
+	_, err := srv.JS.applicationActivationSettings.SetByID(ctx, req.ApplicationIdentifiers, nil, func(stored *ttnpb.ApplicationActivationSettings) (*ttnpb.ApplicationActivationSettings, []string, error) {
+		if stored == nil {
+			return nil, nil, errApplicationActivationSettingsNotFound.New()
+		}
+		return nil, nil, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return ttnpb.Empty, err
```
johanstokking:
```suggestion
	return ttnpb.Empty, nil
```

johanstokking (approved):

PR #3288, opened 2020-09-28
Support LBS LNS Gateway Secrets 

KrishnaIyer:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/1979

#### Changes

- Add Secret message to the API
- Add `LBSLNSSecret` to Gateway store
- Support CRUD
- Update GCS to use decrypted secret instead of from credentials
- Update tests

#### Testing

- [x] Unit tests
- [x] Integration tests with CUPS API

##### Regressions

- [x] Test existing CUPS functions

#### Notes for Reviewers

- Since `LBSLNSSecret` is now a field in `Gateway`, we cannot use a no-echo prompt like we do for User passwords. Any suggestions?

#### Checklist


KrishnaIyer:
CI is fixed and is ready for review.. :)

pkg/identityserver/config.go
```diff
@@ -77,6 +77,9 @@ type Config struct {
 		SMTP         smtp.Config          `name:"smtp"`
 		Templates    emailTemplatesConfig `name:"templates"`
 	} `name:"email"`
+	Gateways struct {
+		EncryptionKeyID string `name:"encryption-key-id" description:"ID of the key used to encrypt gateway secrets at rest"`
```
htdvisser, line 81: New config needs to be documented in `reference/config` and mentioned in `CHANGELOG`.

pkg/identityserver/store/gateway.go
```diff
@@ -63,12 +64,16 @@ type Gateway struct {
 	UpdateLocationFromStatus bool `gorm:"default:false not null"`
 
 	Antennas []GatewayAntenna
+
+	LBSLNSSecret []byte `gorm:"type:BYTEA;column:lbs_lns_secret"`
```
htdvisser, line 68: New database fields need to be mentioned in `CHANGELOG`, and there needs to be a note about running the schema migration.

pkg/identityserver/store/gateway.go
```diff
@@ -116,6 +121,15 @@ var gatewayPBSetters = map[string]func(*ttnpb.Gateway, *Gateway){
 			pb.Antennas[i] = antenna.toPB()
 		}
 	},
+	lbsLNSSecretField: func(pb *ttnpb.Gateway, gtw *Gateway) {
+		blocks := bytes.SplitN(gtw.LBSLNSSecret, lbsLNSSecretSeparator, -1)
```
htdvisser: That -1 doesn't look right. I think it should be 2 instead.

pkg/identityserver/store/gateway.go
```diff
@@ -165,6 +179,15 @@ var gatewayModelSetters = map[string]func(*Gateway, *ttnpb.Gateway){
 			gtw.Antennas[i] = antenna
 		}
 	},
+	lbsLNSSecretField: func(gtw *Gateway, pb *ttnpb.Gateway) {
+		if pb.LBSLNSSecret != nil {
```
htdvisser, line 185: And if it is nil, you should clear `gtw.LBSLNSSecret`.

pkg/identityserver/store/gateway.go
```diff
@@ -116,6 +121,15 @@ var gatewayPBSetters = map[string]func(*ttnpb.Gateway, *Gateway){
 			pb.Antennas[i] = antenna.toPB()
 		}
 	},
+	lbsLNSSecretField: func(pb *ttnpb.Gateway, gtw *Gateway) {
+		blocks := bytes.SplitN(gtw.LBSLNSSecret, lbsLNSSecretSeparator, -1)
+		if len(blocks) == 2 {
```
htdvisser, line 126: And if it is not, you should clear `pb.LBSLNSSecret`.

pkg/identityserver/gateway_registry.go
```diff
@@ -77,6 +78,18 @@ func (is *IdentityServer) createGateway(ctx context.Context, req *ttnpb.CreateGa
 		req.FrequencyPlanIDs = []string{req.FrequencyPlanID}
 	}
 
+	if req.LBSLNSSecret != nil {
+		if is.config.Gateways.EncryptionKeyID == "" {
+			return nil, errGatewaySecretEncryptionKey
```
htdvisser: This doesn't make sense. If it's not configured, we shouldn't return an error of type "not found". Perhaps make an `EncryptionKeyID` of `""` indicate "no encryption"? I think we also have that in the device registries.

pkg/identityserver/gateway_registry.go
```diff
@@ -153,6 +174,17 @@ func (is *IdentityServer) getGateway(ctx context.Context, req *ttnpb.GetGatewayR
 		return nil, err
 	}
 
+	if gtw.LBSLNSSecret != nil {
+		if gtw.LBSLNSSecret.KeyID == "" {
+			return nil, errGatewaySecretEncryptionKey.WithAttributes("id", gtw.LBSLNSSecret.KeyID)
```
htdvisser: See previous comment

pkg/identityserver/gateway_registry.go
```diff
@@ -254,6 +286,19 @@ func (is *IdentityServer) listGateways(ctx context.Context, req *ttnpb.ListGatew
 		if rights.RequireGateway(ctx, gtw.GatewayIdentifiers, ttnpb.RIGHT_GATEWAY_INFO) != nil {
 			gtws.Gateways[i] = gtw.PublicSafe()
 		}
+
+		if rights.RequireGateway(ctx, gtw.GatewayIdentifiers, ttnpb.RIGHT_GATEWAY_READ_SECRETS) != nil {
```
htdvisser: Let's not validate the rights on every gateway unless the caller actually requests this field.

pkg/identityserver/gateway_registry.go
```diff
@@ -283,6 +328,21 @@ func (is *IdentityServer) updateGateway(ctx context.Context, req *ttnpb.UpdateGa
 			return nil, err
 		}
 	}
+	if ttnpb.HasAnyField(req.FieldMask.Paths, "lbs_lns_secret") && req.LBSLNSSecret != nil {
```
htdvisser: Why that `req.LBSLNSSecret != nil`? Is it not allowed to unset this field?

htdvisser (changes requested):
Reviewed code-owned files. BTW: we need code owners on `pkg/basicstation`, otherwise nobody will feel responsible for reviewing those files.

pkg/identityserver/store/gateway.go
KrishnaIyer: Ah yeah. The plan was to allow `:` in the secret indeed.

pkg/identityserver/gateway_registry.go
KrishnaIyer: should we allow that?

pkg/identityserver/gateway_registry.go
KrishnaIyer: Should we allow unencrypted values to be stored in a field where we specifically say it's encrypted at rest?

htdvisser: I don't think that "encrypted at rest" is relevant at the API level. In most deployments (including TTI Cloud) the entire database is encrypted at rest. This is just another layer of encryption that operators can choose to enable, just as with the root keys in the device registries.

pkg/identityserver/config.go
KrishnaIyer: https://github.com/TheThingsIndustries/lorawan-stack-docs/pull/25

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Meta information and status events in the event views in the Console.
 - Support for setting the frame counter width of an end device in the Console.
 - Include consumed airtime metadata in uplink messages and join requests (see `uplink_message.consumed_airtime` field).
+- Store and retrieve LBS LNS Secrets from database.
+  - This requires a database schema migration (`ttn-lw-stack is-db migrate`) because of the added column.
```
htdvisser, line 19: And don't forget to mention the new config that may need to be added in existing deployments

htdvisser (approved):
One thing left, then ✅

CHANGELOG.md
htdvisser: I already wrote that in my previous review by the way

johanstokking (approved):

api/rights.proto
```diff
@@ -155,6 +159,8 @@ enum Right {
 
   // The pseudo-right for all (current and future) possible rights.
   RIGHT_ALL = 55;
+
+  // Next value; 59
```
adriansmares: Nit: `:`

pkg/identityserver/gateway_registry.go
```diff
@@ -77,6 +79,21 @@ func (is *IdentityServer) createGateway(ctx context.Context, req *ttnpb.CreateGa
...
 
+	if req.LBSLNSSecret != nil {
+		value := req.LBSLNSSecret.Value
+		if is.config.Gateways.EncryptionKeyID != "" {
+			value, err = is.KeyVault.Encrypt(ctx, req.LBSLNSSecret.Value, is.config.Gateways.EncryptionKeyID)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			logger := log.FromContext(ctx)
```
adriansmares: Nit: you can merge these two here, and bellow.

pkg/identityserver/store/gateway.go
adriansmares, line 68: Consider a default value here - otherwise old databases will complain that they cannot migrate already existing gateways. 

_I forgot this at least twice already_

adriansmares (approved):

pkg/identityserver/store/gateway.go
KrishnaIyer: Isn't that only for `not null` fields @htdvisser ?

htdvisser: Yeah. Nullable columns have null as default "value".

Issue #2740, opened 2020-06-15
Streamline entity title sections in the Console


bafonins:
#### Summary

Currently, entity title section in the Console are inconsistent.

#### Why do we need this?

Consistent UI across the Console.

#### What is already there? What do you see now?

Application title section:
![Screenshot 2020-06-15 at 17 17 56](https://user-images.githubusercontent.com/16374166/84669175-3d984580-af2d-11ea-8167-a676f57ad4b4.png)

Device title section:
![Screenshot 2020-06-15 at 17 18 11](https://user-images.githubusercontent.com/16374166/84669177-3e30dc00-af2d-11ea-8bc1-b051d768c312.png)

Gateway title section:
![Screenshot 2020-06-15 at 17 18 23](https://user-images.githubusercontent.com/16374166/84669181-3ec97280-af2d-11ea-9594-2beebf9dc16d.png)

Organization title section:
![Screenshot 2020-06-15 at 17 18 42](https://user-images.githubusercontent.com/16374166/84669185-3f620900-af2d-11ea-80b4-7c05da3fdde8.png)

#### What is missing? What do you want to see?

1. Creation date is missing in the device title section, while it is present title sections for other entities. Im not sure if we really need this, since we display `Created at` in the overview pages for each entity.
2. Missing `ID` sub-title:
![68497283-7e700380-0254-11ea-8c01-0907a8e4e909](https://user-images.githubusercontent.com/16374166/84669699-d29b3e80-af2d-11ea-9509-ae6f96891b25.png)
I find this quite helpful.
3. Inconsistent status location and messages (gateway and device title pages).
4. Description location (device title section displays description in a different way than title sections for other entities)

cc @kschiffer 

#### Environment

3.8.4

#### How do you propose to implement this?

Refactor `<EntityTitleSection />` and its usage for applications/devices/gateways and organizatons, as well as `<GatewayConnectionStatus />`

#### Can you do this yourself and submit a Pull Request?

yes

kschiffer:
Please wait with the implementation until I followed up with new designs for the title sections. Will try to squeeze that in this month.

kschiffer:
@bafonins I've added the new designs to Zeplin.

bafonins:
Blocked on https://github.com/TheThingsNetwork/lorawan-stack/issues/2333

bafonins:
[...]

PR #3289, opened 2020-09-28
Add downlink count for end devices in the Console

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2740

<details>
Screenshot 2020-09-28 at 15 31 04
Screenshot 2020-09-28 at 15 31 40

</details>

#### Changes

- Extract downlink frame count from the `ns.down.data.schedule.attempt` event.
- Adjust styles for `<MessagesCount />`

#### Testing

Tested with actual devices on ch staging

#### Notes for Reviewers

@rvolosatovs Please confirm that the `ns.down.data.schedule.attempt` event is the one to check for downlink count. Also, can we use this (or any other) event to bump `last seen` timer for an end device in the console related to downlinks? Currently, we do this only for uplinks (`ns.up.data.receive`, `ns.up.join.receive`, `ns.up.rejoin.receive`)

#### Checklist

### DIFF
--- a/pkg/webui/console/components/entity-title-section/content/messages-count/messages-count.styl
+++ b/pkg/webui/console/components/entity-title-section/content/messages-count/messages-count.styl
@@ -14,3 +14,6 @@
 
 .container
   margin-left: $ls.s
+
+  &:not(:first-of-type)
+    margin-left: $ls.xxs
--- a/pkg/webui/console/store/reducers/devices.js
+++ b/pkg/webui/console/store/reducers/devices.js
@@ -17,6 +17,8 @@ import { mergeWith, merge } from 'lodash'
 import { getCombinedDeviceId, combineDeviceIds } from '@ttn-lw/lib/selectors/id'
 import getByPath from '@ttn-lw/lib/get-by-path'
 
+import { parseLorawanMacVersion } from '@console/lib/device-utils'
+
 import {
   GET_DEV,
   GET_DEVICES_LIST_SUCCESS,
@@ -33,6 +35,7 @@ const defaultState = {
 
 const heartbeatEvents = ['ns.up.data.receive', 'ns.up.join.receive', 'ns.up.rejoin.receive']
 const uplinkFrameCountEvent = 'ns.up.data.process'
+const downlinkFrameCountEvent = 'ns.down.data.schedule.attempt'
 
 const mergeDerived = (state, id, derived) =>
   Object.keys(derived).length > 0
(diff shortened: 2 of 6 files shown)


rvolosatovs:
> @rvolosatovs Please confirm that the `ns.down.data.schedule.attempt` event is the one to check for downlink count.

It is indeed the event you should be using, however there are 2 distinct downlink counters.
Pre 1.1 there's only one and only one counter should be displayed.
In 1.1 and later versions - there are 2 counters. If `FPort` in downlink is equal to `0` - then it's the "network" frame counter, otherwise it's the "application" frame counter. Both counters have to be displayed.

> Also, can we use this (or any other) event to bump last seen timer for an end device in the console related to downlinks? Currently, we do this only for uplinks (ns.up.data.receive, ns.up.join.receive, ns.up.rejoin.receive)

No, you can only update last seen on uplinks.

pkg/webui/console/components/entity-title-section/content/messages-count/messages-count.styl
```diff
@@ -14,3 +14,6 @@
 
 .container
   margin-left: $ls.s
+
+  &:not(:first-of-type)
+    margin-left: $ls.xxs
```
kschiffer, line 19:
```suggestion
.container:first-of-type
    margin-left: $ls.s

.container:not(:first-of-type)
    margin-left: $ls.xxs
```

pkg/webui/console/store/reducers/devices.js
```diff
@@ -120,10 +121,13 @@ const devices = function(state = defaultState, { type, payload, event }) {
 
       // Detect uplink process event to update uplink frame count state.
```
kschiffer:
```suggestion
      // Detect uplink/downlink process events to update uplink/downlink frame count state.
```

kschiffer (approved):

bafonins:
> In 1.1 and later versions - there are 2 counters. If FPort in downlink is equal to 0 - then it's the "network" frame counter, otherwise it's the "application" frame counter. Both counters have to be displayed.

I suggest we stick to the current implementation, but display only the "application" frame counter for 1.1+ devices by cheking the `f_port` value. @rvolosatovs What do you think?

The problem is that I cant really test this with a 1.1 device. I can file an issue for future adjustments.

rvolosatovs:
> I suggest we stick to the current implementation, but display only the "application" frame counter for 1.1+ devices by cheking the `f_port` value. @rvolosatovs What do you think?

SGTM, that's the more relevant one of the two from the perspective of the user.
The other one is mostly useful for debugging of the device operation.
Please file a backlog issue about showing it as well.

bafonins:
@rvolosatovs Is there any way we can distinguish between `< 1.1` and `>= 1.1` end devices by just looking at the event payload?

rvolosatovs:
> @rvolosatovs Is there any way we can distinguish between `< 1.1` and `>= 1.1` end devices by just looking at the event payload?

Depends on the event, but general answer is "no".
In 1.1 there are MAC commands(`Rekey` and `Reset`) that change the MAC version of the device within a session. Console will have to update the MAC version in response to these after the initial GET of the device.
Perhaps, we could introduce a "MAC version change" event, similar to how we have one for class change.
We should also probably have a `session start` event, which would be triggered on first uplink confirming the session or on creation for ABP, that would contain MAC version among other things. @johanstokking what do you think?

johanstokking:
> We should also probably have a `session start` event, which would be triggered on first uplink confirming the session or on creation for ABP, that would contain MAC version among other things

This makes the event stream stateful and I don't think we should do that. If you want to differentiate the display per LoRaWAN version, which makes perfect sense and aligns with what we already do elsewhere, it should be from fetching the end device.

PR #3293, opened 2020-09-29
AS End Device Fetcher

neoaggelos:
#### Summary

References #2669

#### Changes

- Add an AS cache for retrieving device information from other registries, for use as metadata. Kept separately from the AS device registry.
- Currently, only retrieve location information, as needed for #2669

#### Testing

- Test locally
- Unit tests

##### Regressions

- Increases memory usage in AS, can be leveraged by using a Redis cache instead

#### Notes for Reviewers

This is groundwork for #2669, extracted as a separate PR.

#### Checklist

- [ ] Scope: The referenced issue is addressed, there are no unrelated changes.
- [X] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [X] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [X] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #3294, opened 2020-09-29
Set location metadata in uplink messages forwarded by the AS

### LINKED ISSUE #2669: Device location not included in metadata
#### Summary
Device latitude, longitude & altitude are not included in metadata received in the MQTT integration (and presumably other integrations)

#### Steps to Reproduce
1. Register device and set location in console
2. Create an MQTT integration
3. Send uplink
4. Note that location information is not included in message received in MQTT client

#### What do you see now?
Message as described in documentation - https://thethingsstack.io/v3.8.1/integrations/data-formats/

Similar to V2. Include device latitude, longitude and altitude in metadata

...

#### Environment
3.8.2 PAYG instance. Node-RED MQTT Client

Include fields for device location

We'll see Lat/Lon fields in metadata

Unfortunately not

neoaggelos:
#### Summary

Closes #2669 

#### Changes

- Add `locations` field in `ApplicationUplink` messages.
- Set field when handling uplink message in the AS.

#### Testing

- Tested locally, working on unit tests.

#### Checklist


neoaggelos:
Blocked by #3293

api/messages.proto
```diff
@@ -202,7 +202,9 @@ message ApplicationUplink {
 
   bool confirmed = 11;
 
-  // next: 13
+  // End device location metadata, set by the Application Server while handling the message.
+  map<string,Location> locations = 13;
```
adriansmares: Why the `map` instead of `repeated` ? `Location` contains the source.

adriansmares (commented):
Target your cache PR in order to distinguish the changes.

pkg/identityserver/end_device_registry.go
```diff
@@ -98,8 +99,10 @@ func (is *IdentityServer) createEndDevice(ctx context.Context, req *ttnpb.Create
 }
 
 func (is *IdentityServer) getEndDevice(ctx context.Context, req *ttnpb.GetEndDeviceRequest) (dev *ttnpb.EndDevice, err error) {
-	if err = rights.RequireApplication(ctx, req.EndDeviceIdentifiers.ApplicationIdentifiers, ttnpb.RIGHT_APPLICATION_DEVICES_READ); err != nil {
-		return nil, err
+	if err := clusterauth.Authorized(ctx); err != nil {
```
htdvisser: This should not be necessary. The IS already has middleware that assigns `ttnpb.AllClusterRights` when using cluster auth.

johanstokking:
Lots of reviewers here; @neoaggelos who should do what?

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -269,6 +270,18 @@ var startCommand = &cobra.Command{
...
+				var cache gcache.Cache
+				if cfg.Size <= 0 {
+					cache = gcache.New(-1).Build()
+				} else {
+					cache = gcache.New(cfg.Size).LFU().Build()
+				}
+				config.AS.DeviceCache.Cache = applicationserver.NewMemoryDeviceCache(cache)
+			case "":
+			default:
+			}
```
johanstokking: We initialize the concrete implementations in a config package, not in the start command. See many other use cases.

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -269,6 +270,18 @@ var startCommand = &cobra.Command{
...
 			}
+			switch cfg := config.AS.DeviceCache; cfg.Backend {
+			case "memory":
+				var cache gcache.Cache
+				if cfg.Size <= 0 {
+					cache = gcache.New(-1).Build()
+				} else {
+					cache = gcache.New(cfg.Size).LFU().Build()
+				}
+				config.AS.DeviceCache.Cache = applicationserver.NewMemoryDeviceCache(cache)
```
johanstokking: Also don't bother with a cache implementation on this level; make `NewMemoryDeviceCache` responsible for that

neoaggelos:
> Lots of reviewers here; @neoaggelos who should do what?

This is because of CODEOWNERS. This PR is currently blocked waiting for https://github.com/TheThingsNetwork/lorawan-stack/pull/3293

pkg/applicationserver/device_cache.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// DeviceCache caches device protos.
+type DeviceCache interface {
+	Get(ctx context.Context, ids ttnpb.EndDeviceIdentifiers) (*ttnpb.EndDevice, error)
+	Set(ctx context.Context, ids ttnpb.EndDeviceIdentifiers, dev *ttnpb.EndDevice, ttl time.Duration) error
+}
+
+type memoryDeviceCache struct {
+	gcache.Cache
```
johanstokking: Since we're in `pkg/applicationserver`, don't embed this type as that exposes all methods of `gcache.Cache` on `memoryDeviceCache`, leading to future errors in the future.

pkg/applicationserver/applicationserver.go
```diff
@@ -1055,3 +1065,33 @@ func (as *ApplicationServer) GetMQTTConfig(ctx context.Context) (*config.MQTT, e
...
 	return &config.MQTT, nil
 }
+
+func (as *ApplicationServer) getCachedEndDevice(ctx context.Context, ids ttnpb.EndDeviceIdentifiers) (*ttnpb.EndDevice, error) {
+	if as.deviceCache != nil {
+		dev, err := as.deviceCache.Get(ctx, ids)
+		if err != nil {
+			return dev, nil
+		}
+	}
```
johanstokking: `getCachedEndDevice()` implies that it uses a cache, but that is conditional here.

Better rename this to `fetchEndDevice()` and godoc that it may use the cache?

pkg/applicationserver/applicationserver.go
```diff
@@ -1055,3 +1065,33 @@ func (as *ApplicationServer) GetMQTTConfig(ctx context.Context) (*config.MQTT, e
...
+	cc, err := as.GetPeerConn(ctx, ttnpb.ClusterRole_ENTITY_REGISTRY, ids)
+	if err != nil {
+		return nil, err
+	}
+	client := ttnpb.NewEndDeviceRegistryClient(cc)
+	dev, err := client.Get(ctx, &ttnpb.GetEndDeviceRequest{
+		EndDeviceIdentifiers: ids,
+		FieldMask: pbtypes.FieldMask{
+			Paths: []string{
+				"locations",
```
johanstokking: This is lots of magic; how does the caller know that only `locations` are fetched? That happens to be the first use case of getting the device from IS. But this should be passed as method argument.

johanstokking (changes requested):
I'm not sure if I like the high coupling in AS this way. For inspiration, please see the tenant fetcher which has a cache.

You might want to change the cache interface to a generic "getter" with callback to fetch the entity if needed. The concrete implementation can then always use the callback to get the entity (no cache), use a memory cache, etc.

johanstokking:
Wait, is https://github.com/TheThingsNetwork/lorawan-stack/pull/3293 the better version of this?

neoaggelos:
> Wait, is #3293 the better version of this?

https://github.com/TheThingsNetwork/lorawan-stack/pull/3294#issuecomment-700526324. #3293 is the device fetcher extracted as a separate PR, so that this can focus on the proto changes instead.

adriansmares:
> 
> 
> > Wait, is #3293 the better version of this?
> 
> [#3294 (comment)](https://github.com/TheThingsNetwork/lorawan-stack/pull/3294#issuecomment-700526324). #3293 is the device fetcher extracted as a separate PR, so that this can focus on the proto changes instead.

If you make such chains of PRs, please actually make them target one another instead of making it look like this PR introduces changes that it doesn't. While you indeed added the `blocking`/`blocks` labels and comments, the boundaries between the PRs are unclear and as such the reviewers do not actually know which change in the code belongs which. 

If you would have targeted your branch from #3293 here, instead of `3.10`, only the API protos changes would have been visible (exactly what you wanted).

GitHub would have also automatically changed the target of this branch back to `3.10` once #3293 was merged, so you don't end up with lingering references.

api/messages.proto
neoaggelos: This is the same format as in `EndDevice` https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.10/api/end_device.proto#L437.

api/messages.proto
```diff
@@ -202,7 +202,10 @@ message ApplicationUplink {
 
   bool confirmed = 11;
 
-  // next: 13
+  // End device location metadata, set by the Application Server while handling the message.
+  map<string,Location> locations = 14;
```
neoaggelos, line 213: This is to account for https://github.com/TheThingsNetwork/lorawan-stack/pull/3298/files#diff-5e41eaf89519c9fe313dc525c2d3712dR207

johanstokking: This link doesn't seem to go to a related target, can you elaborate?

htdvisser: I guess he meant the `google.protobuf.Duration consumed_airtime = 13` added in #3298.

htdvisser (approved):

johanstokking (approved):

pkg/applicationserver/applicationserver.go
```diff
@@ -935,6 +939,13 @@ func (as *ApplicationServer) handleUplink(ctx context.Context, ids ttnpb.EndDevi
 		uplink.LastAFCntDown = dev.Session.LastAFCntDown
 	}
 
+	isDev, err := as.endDeviceFetcher.Get(ctx, ids, "locations")
+	if err != nil {
+		log.FromContext(ctx).WithError(err).Warn("Failed to retrieve end device locations")
```
johanstokking: Nit; we have a `logger` here already

johanstokking (approved):

adriansmares (approved):

PR #3301, opened 2020-09-30
Fix freezes and further improve performance of event views

### LINKED ISSUE #2887: Data views slow down and freeze after receiving many events

kschiffer:
#### Summary
This PR fixes lags, freezes and improves the performance of the event views by introducing various performance optimizations.

Closes #2887 

#### Changes
- Speed up render times of `<Event />` component
  - Introduce `<SafeInspector.Light />` to avoid big overheads of the regular safe inspector when rendered inside events
  - Replace expensive usage of `getByPath()` object prop utility with vanilla implementation
  - Replace expensive usage of `<DateFormat />` with simple vanilla implementation
  - Make event error boundary a `PureComponent`
- Improve performance of store logic
  - Introduce an event limit to avoid memory issues (currently set to 2000 events, ~3MB)
  - Introduce a buffered dispatch to the event store, to debounce expensive processing and store update operations, which caused significant lag in very active event streams (especially when scrolling the list)
    - Refactor affected store logic to handle multiple events per dispatch
- Other tweaks
  - Slightly adjust the overscan count for `react-window`
  - Fix a small styling issue of the `<SafeInspector />`
  - Remove superfluous top-level error checks for end devices, gateways and organization (missed doing that in #3229)

#### Testing

Manual testing in staging environment and injected redux state

##### Regressions

This might introduce issues regarding
  - Event list integrity
  - Display issues of the data previews

#### Notes for Reviewers
From my own profiling and testing, this improves the stability and speeds up the performance of the event view significantly. Regarding performance optimization there's obviously a never-ending room for further tweaks, though I think this is the best trade-off between reasonable performance optimization and avoiding adding too many obtrusive code changes.

Regarding 2000 event limit: this seemed like a sane limitation to me for now, at which the console still runs smoothly. We might want to fine-tune this value later when we have more feedback and metrics. Generally, it's important to establish for the user that the Console event view is not the correct place to source long-term event data.

#### Checklist


kschiffer:
Closing to split up into individual PRs.

PR #3325, opened 2020-10-06
Split up oauth package into oauth and account (Account app scaffolding)

kschiffer:
#### Summary
This PR takes care of splitting up the oauth package into the oauth package (token exchange and authorization) to  account app (authentication and account management), while retaining the config structure to stay conform to our compatibility commitment.

References #1422 

#### Changes
- Split up oauth app and account app
  - Extract authentication logic from `oauth` package to  new `account` package
  - Rename `oauth.js` entrypoint to `account.js`
  - Rename `oauth` webui folder to `account`
  - Rename static assets
  - Update module import statements
  - Update various configs accordingly
    - `package.json` module name mapper for jest
    - webpack config for module aliases (`@oauth` to `@account`)
    - eslint alias config
    - Retain stack config structure
  - Rename github commit topic from `oauth` to `account`
- Some related reducer fixes
- Updating go tests accordingly

#### Testing

I did manual testing and updated the unit tests.

##### Regressions

This is a pretty extensive PR so there might be issues introduced to numerous aspects:
- Authentication, login and logout
- Token exchange and authorization
- Broken redirects
- Broken views

#### Notes for Reviewers

Here's a little recap of the plan since this is going on for a while now:
- We want to introduce the "Account" app to be a dedicated app dealing with user authentication, profile settings, session management, client authorization management and other concerns revolving around the user account
- The `oauth` package will remain in place but shrinked down to merely authorization concerns (oauth flow, token exchange, client authorization)
- The Account app will not be an OAuth client (like the Console) but authorize API action via the session token. Session cookie authorization was added via [Enable authorization via session cookie by kschiffer · Pull Request #2564 · TheThingsNetwork/lorawan-stack · GitHub](https://github.com/TheThingsNetwork/lorawan-stack/pull/2564)
- The frontend of the `oauth` server will be reduced to merely the authorization screen (to authorize OAuth clients)
- Since we cannot break existing configuration, I decided to use the `is.oauth.*` configuration for **both** the `oauth` and `account` package.
  - This means that both web servers use the same `/oauth` base route by default
  - It is intended, however, to overwrite the default to use `/account` instead
  - On the frontend we'll also use the same entry point `account.js` for both frontends
  - We'll also use the same "branding" for both frontends, since we want the experience to unified through the "Account" app, meaning that the remaining authorization screen of the `oauth` server will be handled by the `account.js` endpoint as well.
- This PR is a necessary scaffolding to be able to rebase the already done frontend work of the Account app onto a working basis

Note that the target branch is again `feature/1422-account-app`, so this is not targeting master/develop and is preliminary work for upcoming account app related PRs. The point of this is again to reduce review diffs, chunking things down into bits that are scoped and easy to review until eventually merging `feature/1422-account-app` into `v3.10` when all parts have been merged into it.

It's necessary to update the config variables to mount this in development mode:
```
export export TTN_LW_IS_OAUTH_UI_JS_FILE="libs.bundle.js account.js"
```

Main reviewers: @bafonins @htdvisser 
@adriansmares please double-check whether these changes introduce any issues regarding federated auth.

#### Checklist


cmd/internal/shared/identityserver/config.go
```diff
@@ -28,15 +28,17 @@ import (
...
 			TemplateData: webui.TemplateData{
 				SiteName:      "The Things Stack for LoRaWAN",
+				Title:         "Account",
 				Language:      "en",
 				CanonicalURL:  shared.DefaultOAuthPublicURL,
 				AssetsBaseURL: shared.DefaultAssetsBaseURL,
-				IconPrefix:    "oauth-",
-				CSSFiles:      []string{"oauth.css"},
-				JSFiles:       []string{"oauth.js"},
+				IconPrefix:    "account-",
```
htdvisser: Changing this means that deployments with custom branding will have to copy/rename their custom branding files.

pkg/oauth/server.go
```diff
@@ -22,6 +22,7 @@ import (
 
 	echo "github.com/labstack/echo/v4"
 	"github.com/openshift/osin"
+	sess "go.thethings.network/lorawan-stack/v3/pkg/account/session"
```
htdvisser: Why is this import alias needed?

pkg/oauth/server.go
```diff
@@ -223,18 +223,10 @@ func (s *server) RegisterRoutes(server *web.Server) {
...
 	csrfMiddleware := middleware.CSRF("_csrf", "/", s.config.CSRFAuthKey)
 
-	api := root.Group("/api", csrfMiddleware)
-	api.POST("/auth/login", s.Login)
-	api.POST("/auth/logout", s.Logout, s.requireLogin)
-	api.GET("/me", s.CurrentUser, s.requireLogin)
-
 	page := root.Group("", csrfMiddleware)
-	page.GET("/login", webui.Template.Handler, s.redirectToNext)
 	page.GET("/logout", s.ClientLogout)
```
htdvisser, line 230: Maybe add a note that this route is in use by existing OAuth clients, meaning that it's public API and should therefore not be removed in v3.

pkg/oauth/server.go
```diff
@@ -223,18 +223,10 @@ func (s *server) RegisterRoutes(server *web.Server) {
...
 	page.GET("/authorize", s.Authorize(webui.Template.Handler), s.redirectToLogin)
 	page.POST("/authorize", s.Authorize(webui.Template.Handler), s.redirectToLogin)
-	page.GET("/", webui.Template.Handler, s.redirectToLogin)
-	page.GET("/*", webui.Template.Handler)
```
htdvisser, line 237: I think we (used to) send emails linking to `/oauth`, so if we remove this, we're breaking those links, so let's keep the redirect

pkg/account/user.go
```diff
@@ -0,0 +1,118 @@
...
+		return err
+	}
+	session, err := s.store.CreateSession(ctx, &ttnpb.UserSession{
+		UserIdentifiers: userIDs,
+		SessionSecret:   hashedSecret,
+	})
+	if err != nil {
+		return err
+	}
+	events.Publish(evtUserLogout.NewWithIdentifiersAndData(ctx, userIDs, nil))
```
htdvisser, line 97: This doesn't look like the place to send a `evtUserLogout`

pkg/account/session/observability.go
```diff
@@ -0,0 +1,41 @@
...
+		events.WithVisibility(ttnpb.RIGHT_USER_ALL),
+		events.WithAuthFromContext(),
+		events.WithClientInfoFromContext(),
+	)
+	evtUserLogout = events.Define(
+		"account.user.logout", "logout user",
+		events.WithVisibility(ttnpb.RIGHT_USER_ALL),
+		events.WithAuthFromContext(),
+		events.WithClientInfoFromContext(),
+	)
```
htdvisser: These do not seem to be used in this package

pkg/account/observability.go
```diff
@@ -0,0 +1,41 @@
...
+		events.WithVisibility(ttnpb.RIGHT_USER_ALL),
+		events.WithAuthFromContext(),
+		events.WithClientInfoFromContext(),
+	)
+	evtUserLoginFailed = events.Define(
+		"oauth.user.login_failed", "login user failure",
+		events.WithVisibility(ttnpb.RIGHT_USER_ALL),
+		events.WithAuthFromContext(),
+		events.WithClientInfoFromContext(),
+	)
```
htdvisser: This doesn't seem to be used in this package.

htdvisser (commented):
Did a quick initial review. Will take a closer look another time.

johanstokking:
> Here's a little recap of the plan since this is going on for a while now:
> [...]

Totally agreed with this.

adriansmares (commented):
Took a look at the changes to `pkg/oauth` and I consider this to be compatible with the federated auth (which will probably migrate to `pkg/account`).

htdvisser:
Nothing to add to my previous review

pkg/oauth/server.go
kschiffer: But the route will be handled by the account server. Am I missing something here?

htdvisser: Okay, then it's fine. I thought the account server was mounted somewhere else.

htdvisser (approved):
Go stuff LGTM

PR #3326, opened 2020-10-06
Update gateway antenna location only when auth is available

### LINKED ISSUE #3138: Gateway location update from status messages does not work when gateway is connected without authentication
#### Summary

Upon receiving a location from a gateway status message, GS performs an `GatewayRegistry.Update` call to the IS. RPC metadata from the context are forwarded if available (i.e. if the gateway has connected with authentication -- this is MQTT, gRPC, or BasicStation), otherwise falls back to cluster auth.

However, `GatewayRegistry.Update` does not accept cluster auth as valid authentication.

#### Steps to Reproduce

1. Connect with a gateway over UDP
2. Send a status message that includes a location, e.g. `{"stat":{"time":"2020-08-25 10:07:24 UTC","lati":37,"long":30,"alti":2,"rxnb":0,"rxok":0,"rxfw":0,"ackr":0,"dwnb":0,"txnb":0}}`

#### What do you see now?

```
  WARN Finished unary call                      duration=6.332433ms error=error:pkg/auth/rights:insufficient_gateway_rights (insufficient rights for gateway `gtw2`) error_correlation_id=10a5ae5615d8427b86c380cdec783001 error_name=insufficient_gateway_rights error_namespace=pkg/auth/rights grpc_code=PermissionDenied grpc_method=Update grpc_service=ttn.lorawan.v3.GatewayRegistry missing=[RIGHT_GATEWAY_SETTINGS_BASIC] namespace=grpc request_id=01EGJFHVXT6FVAFKKRQFJPVB0W uid=gtw2
  WARN Finished unary call                      duration=7.933779ms error=rpc error: code = PermissionDenied desc = error:pkg/auth/rights:insufficient_gateway_rights (insufficient rights for gateway `gtw2`) error_correlation_id=10a5ae5615d8427b86c380cdec783001 error_name=insufficient_gateway_rights error_namespace=pkg/auth/rights grpc_code=PermissionDenied grpc_method=Update grpc_service=ttn.lorawan.v3.GatewayRegistry namespace=grpc
  WARN Failed to update antenna location        error=error:pkg/auth/rights:insufficient_gateway_rights (insufficient rights for gateway `gtw2`) gateway_eui=0101010101010102 gateway_uid=gtw2 namespace=gatewayserver/io/udp protocol=udp tenant_id=default uid=gtw2

```

The gateway antenna location should be updated.

#### Environment

3.9.1

Three options:

- Support `clusterauth` in `updateGateway`
- Support `clusterauth` in `updateGateway` only when `field mask is `[]string{"antennas"}`
- Add a new RPC `UpdateGatewayAntennas` with clusterauth.

- Unit tests in the GS mock registry, to verify clusterauth is forwarded
- Unit tests in the IS to verify that clusterauth works
- Test locally with a UDP gateway.

yes

neoaggelos:
#### Summary

Closes #3138

#### Changes

- Accept RIGHT_GATEWAY_LINK in GatewayRegistry.Update if only locations path is passed
- Do not use clusterauth for location updates from GS.
- Update Console messages to reflect this change.

#### Testing

Unit tests, pending tests with an actual gateway

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ For details about compatibility between different releases, see the **Commitment
 - CUPS Server only accepts The Things Stack API Key for token auth.
 - Improve MQTT Pub/Sub task restart conditions and error propagation.
 - Pausing event streams is not saving up arriving events during the pause anymore.
+- Gateway server can now update the gateway location only if the gateway is authenticated.
 
 ### Deprecated
 
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -804,9 +804,7 @@ func (gs *GatewayServer) handleLocationUpdates(conn connectionEntry) {
 	var err error
 	var callOpt grpc.CallOption
 	callOpt, err = rpcmetadata.WithForwardedAuth(ctx, gs.AllowInsecureForCredentials())
-	if errors.IsUnauthenticated(err) {
-		callOpt = gs.WithClusterAuth()
-	} else if err != nil {
+	if err != nil {
 		return
 	}
 	registry, err := gs.getRegistry(ctx, &conn.Gateway().GatewayIdentifiers)
--- a/pkg/identityserver/gateway_registry.go
+++ b/pkg/identityserver/gateway_registry.go
@@ -318,9 +318,12 @@ func (is *IdentityServer) listGateways(ctx context.Context, req *ttnpb.ListGatew
 
 func (is *IdentityServer) updateGateway(ctx context.Context, req *ttnpb.UpdateGatewayRequest) (gtw *ttnpb.Gateway, err error) {
 	if err = rights.RequireGateway(ctx, req.GatewayIdentifiers, ttnpb.RIGHT_GATEWAY_SETTINGS_BASIC); err != nil {
-		return nil, err
+		// Allow setting only the location field with the RIGHT_GATEWAY_LINK right.
+		isLink := rights.RequireGateway(ctx, req.GatewayIdentifiers, ttnpb.RIGHT_GATEWAY_LINK) == nil
+		if topLevel := ttnpb.TopLevelFields(req.FieldMask.Paths); !isLink || len(topLevel) != 1 || topLevel[0] != "antennas" {
+			return nil, err
+		}
 	}
-
 	// Backwards compatibility for frequency_plan_id field.
 	if ttnpb.HasAnyField(req.FieldMask.Paths, "frequency_plan_id") {
 		if !ttnpb.HasAnyField(req.FieldMask.Paths, "frequency_plan_ids") {
--- a/api/api.md
+++ b/api/api.md
@@ -2797,7 +2797,7 @@ Gateway is the message that defines a gateway on the network.
 | `enforce_duty_cycle` | [`bool`](#bool) |  | Enforcing gateway duty cycle is recommended for all gateways to respect spectrum regulations. Disable enforcing the duty cycle only in controlled research and development environments. |
 | `downlink_path_constraint` | [`DownlinkPathConstraint`](#ttn.lorawan.v3.DownlinkPathConstraint) |  |  |
 | `schedule_anytime_delay` | [`google.protobuf.Duration`](#google.protobuf.Duration) |  | Adjust the time that GS schedules class C messages in advance. This is useful for gateways that have a known high latency backhaul, like 3G and satellite. |
-| `update_location_from_status` | [`bool`](#bool) |  | update the location of this gateway from status messages |
+| `update_location_from_status` | [`bool`](#bool) |  | Update the location of this gateway from status messages. This only works for gateways connecting with authentication; gateways connected over UDP are not supported. |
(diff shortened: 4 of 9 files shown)


pkg/gatewayserver/gatewayserver.go
```diff
@@ -841,7 +832,7 @@ func (gs *GatewayServer) handleLocationUpdates(conn connectionEntry) {
 							"antennas",
 						},
 					},
-				}, callOpt)
+				}, gs.WithClusterAuth())
```
KrishnaIyer: I don't know if this is a good idea. Anyone can now spoof the location of any UDP gateway since we don't check for auth.

htdvisser (changes requested):
I agree with @KrishnaIyer. We should only do this for gateways that provide sufficient auth. 

Since most gateways will use an API key with only `RIGHT_GATEWAY_LINK`, I would suggest to update `(*IdentityServer).updateGateway` to accept the `RIGHT_GATEWAY_LINK` right if only the `antennas` field is in the field mask.

adriansmares (commented):
Nothing to add.

kschiffer (approved):
Codeowned files LGTM

neoaggelos:
Lots of reviewers requested here because of multiple codeowners.

@kschiffer please review the text in the console
@htdvisser please review IS changes, documentation

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improve LinkADRReq scheduling condition computation and, as a consequence, downlink task efficiency.
 - CUPS Server only accepts The Things Stack API Key for token auth.
 - Improve MQTT Pub/Sub task restart conditions and error propagation.
+- GS is now consistent in automatically updating gateway locations only if the gateway is connected via an authenticated frontend (gRPC, MQTT, or BasicStation).
```
```suggestion
- Gateway Server is now updates the gateway location only if the gateway is authenticated.
```

htdvisser (approved):

KrishnaIyer:
Just a question; Did we explore the possibility to make the dynamic location part of the stats and make the clients fetch that? Do we need to update the entity registry for this? 
The reasoning being that we clear the stats on disconnect and we don't have the last known location?

pkg/identityserver/gateway_registry.go
```diff
@@ -316,11 +316,19 @@ func (is *IdentityServer) listGateways(ctx context.Context, req *ttnpb.ListGatew
...
+	}
+	return rights.RequireGateway(ctx, req.GatewayIdentifiers, ttnpb.RIGHT_GATEWAY_SETTINGS_BASIC)
+}
+
 func (is *IdentityServer) updateGateway(ctx context.Context, req *ttnpb.UpdateGatewayRequest) (gtw *ttnpb.Gateway, err error) {
-	if err = rights.RequireGateway(ctx, req.GatewayIdentifiers, ttnpb.RIGHT_GATEWAY_SETTINGS_BASIC); err != nil {
+	if err = checkUpdateGatewayRights(ctx, req); err != nil {
 		return nil, err
 	}
-
```
johanstokking: This is a bit narrow and doesn't need it's own function.

How about:

```go
func (is *IdentityServer) updateGateway(ctx context.Context, req *ttnpb.UpdateGatewayRequest) (gtw *ttnpb.Gateway, err error) {
	if err := rights.RequireGateway(ctx, req.GatewayIdentifiers, ttnpb.RIGHT_GATEWAY_SETTINGS_BASIC); err != nil {
		// Allow setting only the location field with the RIGHT_GATEWAY_LINK right.
		isLink := rights.RequireGateway(ctx, req.GatewayIdentifiers, ttnpb.RIGHT_GATEWAY_LINK) == nil
		if topLevel := ttnpb.TopLevelFields(req.FieldMask.Paths); len(topLevel) != 1 || topLevel[0] != "location" || !isLink {
			return nil, err
		}
	}
```

neoaggelos:
> The reasoning being that we clear the stats on disconnect and we don't have the last known location?

Yup, this is the only catch there. To be honest, I am not sure if this would help or just cause confusion to the users.

Another alternative would be to have multiple location sources (same as we do for end devices), but that seems unnecessarily complex and needs API changes, so let's avoid it for something minor like this.

pkg/identityserver/gateway_registry.go
neoaggelos: The point is to preserve the error message being about the gateway settings rights, not the gateway link rights.

Seems a tad more sane to me, but what you are suggesting is also fine if we do not really care about that.

johanstokking: > The point is to preserve the error message being about the gateway settings rights, not the gateway link rights.

Did you see the code above?

neoaggelos: Ah, missed the `isLink`.

pkg/webui/console/containers/gateway-location-form/gateway-location-form.js
```diff
@@ -30,7 +30,7 @@ const m = defineMessages({
   locationDescription: 'The location of this gateway may be publicly displayed',
   updateLocationFromStatus: 'Update from status messages',
   updateLocationFromStatusDescription:
-    'Update the location of this gateway based on incoming status messages',
+    'Update the location of this gateway based on incoming status messages (only works for authenticated gateway connections)',
```
kschiffer: This should rather be a warning that appears when the checkbox is activated. Also, this should link to some documentation that explains the preconditions for the location updates to be functional in more detail.

@benolayinka @nejraselimovic do we have something like that already?

My suggestion to not stall this PR: remove this addition and open an issue about this.

kschiffer (approved):
See my remark.

johanstokking (approved):

PR #3329, opened 2020-10-07
Optimize render performance of event components

kschiffer:
#### Summary
This PR introduces performance improvements for the render functions of event components, to speed up the render times of these frequently (re-)rendered components.

References #2887 

#### Changes
- Speed up render times of `<Event />` component
  - Introduce `<SafeInspector.Light />` to avoid big overheads of the regular safe inspector when rendered inside events
  - Replace expensive usage of `getByPath()` object prop utility with vanilla implementation
  - Replace expensive usage of `<DateFormat />` with simple vanilla implementation
  - Make event error boundary a `PureComponent`
- Other tweaks
  - Slightly adjust the overscan count for `react-window`
  - Fix a small styling issue of the `<SafeInspector />`

#### Testing

Manual testing in staging environment and injected redux state.

##### Regressions

This might introduce display issues of the data previews.

#### Notes for Reviewers
From my own profiling and testing, this improves the stability and speeds up the performance of the event view quite a bit. Regarding performance optimization, there's obviously a never-ending room for further tweaks, though I think this is the best trade-off between reasonable performance optimization and avoiding adding too many obtrusive code changes.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - CLI support for Storage Integration (see `ttn-lw-cli end-devices storage` and `ttn-lw-cli applications storage` commands).
 - Network Server does not retry rejected `NewChannelReq` data rate ranges or rejected `DLChannelReq` frequencies anymore.
 - Functionality to allow admin users to list all organizations in the Console.
+- Performance optimizations for event views in the Console.
 
 ### Changed
 
--- a/pkg/webui/components/safe-inspector/chunk-array.js
+++ b/pkg/webui/components/safe-inspector/chunk-array.js
@@ -1,0 +1,21 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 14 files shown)


bafonins (approved):

kschiffer:
After testing this some more:

After using only the performance tab of chrome developer tools, I used [`storybook-addon-performance`](https://github.com/atlassian-labs/storybook-addon-performance) to measure the impact more meaningfully.

*   For some reason, there is a high variance between samples of up to 50%
    *   I've used samples that render 100 copies of the same `<Event />` component for 10 times
*   This is mainly due to some individual `<Event />` renders taking much longer than the rest (even when using exact same props)
    *   I've yet to find the exact cause, which is quite difficult
    *   From checking performance samples, I get the idea that this might be caused by `react-intl` sometimes causing lag when loading the intl context. It's however hard to verify. Some performance samples suggest that there is a general throttling happening affecting the whole call stack of an individual `<Event />` render. See [here](https://user-images.githubusercontent.com/5710611/96499473-5d29dc80-1245-11eb-81ad-9b0e7b216c12.png) and [here](https://user-images.githubusercontent.com/5710611/96499663-9d895a80-1245-11eb-9583-ea596a832393.png)  
        (two renders in the middle)
    *   All `<Event />` components that the `<EventList />` renders are tapping the react intl context multiple times
    *   This could be mitigated by pre-loading and passing down the translations of all possible messages that the event component can display
    *   This issue remains the cause of render lags during fast scrolling in the event view
    *   This also makes it very hard to produce meaningful performance measurements
*   The performance can be further improved by removing the date conversion from the server to the local date. This operation includes parsing and offsetting the date, which profiling showed to be quite expensive. This also made me wonder whether we really want to show local dates for events, since this might cause confusion when users will compare dates throughout different timezones.

Since I cannot trust the test results sufficiently, I don't want to introduce these changes since they might actually be futile. It is likely that these changes improve the performance but they, unfortunately, won't address the much bigger cause of render lags which I described above.

PR #3330, opened 2020-10-07
Improve performance and load resilience of event store logic

### LINKED ISSUE #2887: Data views slow down and freeze after receiving many events

kschiffer:
#### Summary
This PR fixes lags, freezes and improves the performance of the event views by introducing optimizations to the event store logic.

References #3329 
Closes #2887 

#### Changes

- Improve performance of store logic
  - Introduce an event limit to avoid memory issues (currently set to 2000 events, ~3MB)
  - Introduce a buffered dispatch to the event store, to debounce expensive processing and store update operations, which caused significant lag in very active event streams (especially when scrolling the list)
    - Refactor affected store logic to handle multiple events per dispatch
- Other tweaks
  - Remove superfluous top-level error checks for end devices, gateways and organization (missed doing that in #3229)

#### Testing

Manual testing in staging environment and injected redux state.
##### Regressions

This might introduce issues regarding event list integrity.

#### Notes for Reviewers
Regarding 2000 event limit: this seemed like a sane limitation to me for now, at which the console still runs smoothly. We might want to fine-tune this value later when we have more feedback and metrics. Generally, it's important to establish for the user that the Console event view is not the correct place to source long-term event data.

See also #3329 and #3301 

#### Checklist

### DIFF
--- a/pkg/webui/console/store/selectors/events.js
+++ b/pkg/webui/console/store/selectors/events.js
@@ -32,7 +32,7 @@ export const createEventsInterruptedSelector = entity =>
   function(state, entityId) {
     const store = selectEventsStore(state.events[entity], entityId)
 
-    return store ? store.interrupted : false
+    return Boolean(store.interrupted)
   }
 
 export const createEventsErrorSelector = entity =>
@@ -40,6 +40,13 @@ export const createEventsErrorSelector = entity =>
     const store = selectEventsStore(state.events[entity], entityId)
 
     return store ? store.error : undefined
+  }
+
+export const createEventsTruncatedSelector = entity =>
+  function(state, entityId) {
+    const store = selectEventsStore(state.events[entity], entityId)
+
+    return Boolean(store.truncated)
   }
 
 export const createLatestEventSelector = function(entity) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - CLI support for Storage Integration (see `ttn-lw-cli end-devices storage` and `ttn-lw-cli applications storage` commands).
 - Network Server does not retry rejected `NewChannelReq` data rate ranges or rejected `DLChannelReq` frequencies anymore.
 - Functionality to allow admin users to list all organizations in the Console.
+- Limitation of displayed and stored events in the Console to 2000.
 
 ### Changed
 
--- a/pkg/webui/console/components/events/events.styl
+++ b/pkg/webui/console/components/events/events.styl
@@ -95,6 +95,18 @@ $event-container-height = 40px
     margin: 0
     list-style: none
 
+.truncated
+  border-normal(top)
+  display: flex
+  align-items: center
+  justify-content: center
+  font-size: $fs.s
+  color: $tc-warning
+  padding: $cs.xxs 0
+
+  span
+    margin-right: $cs.xxs
+
 .event
   height: $event-container-height
   background-color: transparent
--- a/pkg/webui/console/components/events/index.js
+++ b/pkg/webui/console/components/events/index.js
@@ -18,6 +18,7 @@ import classnames from 'classnames'
 import hamburgerMenuClose from '@assets/misc/hamburger-menu-close.svg'
 
 import Button from '@ttn-lw/components/button'
+import Icon from '@ttn-lw/components/icon'
 
 import Message from '@ttn-lw/lib/components/message'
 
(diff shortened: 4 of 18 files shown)


pkg/webui/console/store/selectors/events.js
```diff
@@ -42,6 +42,13 @@ export const createEventsErrorSelector = entity =>
     return store ? store.error : undefined
   }
 
+export const createEventsTruncatedSelector = entity =>
+  function(state, entityId) {
+    const store = selectEventsStore(state.events[entity], entityId)
+
+    return store ? store.truncated : undefined
```
bafonins: nit
```suggestion
    return Boolean(store.truncated)
```

pkg/webui/console/store/middleware/logics/events.js
```diff
@@ -104,12 +105,33 @@ const createEventsConnectLogics = function(reducerName, entityName, onEventsStar
...
+        const {
+          addToBuffer: addToEventBuffer,
+          clearBuffer: clearEventBuffer,
+        } = createBufferedProcess(events => {
+          const processedEvents = events
+            .filter(
+              // See https://github.com/TheThingsNetwork/lorawan-stack/pull/2989
+              event => event.name !== 'events.stream.start' && event.name !== 'events.stream.stop',
+            )
+            .sort((a, b) => (a.time > b.time ? -1 : 1))
```
bafonins: The main reason to sort events is that we have no guarantees that they arrive in order and this does not solve this problem. Consider the following case:

Buffer `b` of size 2;
Events `e1`, `e2` and `e3`
- `e1.time = t1`, `e1.time_arrived = t1'`
- `e2.time = t2`, `e2.time_arrived = t2'`
- `e3.time = t3`, `e3.time_arrived = t3'`

Such that `t3 > t2 > t1` and `t3' > t1' > t2'` (So the event `e1` arrives after `e2`)

The events should be rendered in the following order: `[e3, e2, e1]`, however with this implementation we get `[e2, e1, e3]`.

bafonins: This actually happened to me:
Screenshot 2020-10-08 at 17 10 42

pkg/webui/console/store/middleware/logics/events.js
```diff
@@ -104,12 +105,33 @@ const createEventsConnectLogics = function(reducerName, entityName, onEventsStar
...
+
+          if (processedEvents.length > 0) {
```
bafonins: nit; I would check for this before filtering and sorting, so we dont recreate an empty array when calling `filter`

pkg/webui/console/store/reducers/devices.js
```diff
@@ -114,18 +115,24 @@ const devices = function(state = defaultState, { type, payload, event }) {
...
         }
       }
 
       // Detect uplink process event to update uplink frame count state.
-      else if (event.name === uplinkFrameCountEvent) {
-        const id = getCombinedDeviceId(event.identifiers[0].device_ids)
-        return mergeDerived(state, id, {
-          uplinkFrameCount: getByPath(event, 'data.payload.mac_payload.full_f_cnt'),
-        })
+      const uplinkFrameCountEvent = events.find(event => event.name === uplinkFrameCountEventName)
```
bafonins: I no longer see uplink count on the end device overview page. Not sure what caused it

pkg/webui/lib/create-buffered-process.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * `createBufferedProcess` creates a buffered event ingest to combine and
+ * process multiple payloads at once that can otherwise cause lags when
+ * processed individually in rapid succession.
+ *
+ * @param {Function} process - The action to be performed on the combined
+ * payloads before the buffer is cleared.
+ * @param {number} maxItems - The maximum number of items that the buffer can
+ * hold before it will dispatch, regardless of the delay. Default is 20.
```
bafonins: Actually it can hold and dispatch `maxItems` + 1

bafonins (commented):
Like the idea. Can you share perf metrics/tools you used to measure performance improvements?

Also see my comments

pkg/webui/console/store/middleware/logics/events.js
kschiffer: This one is to check whether there is any event left after removing `events.stream.start/stop`. Otherwise, there should be no empty event arrays arriving here.

kschiffer:
Unfortunately, I cannot make sure that the changes here introduce an improvement that offsets the cost of making the logic more complicated. By profiling under development conditions with live data, there are too many outside factors that distort the results. It is necessary to go about this more systematically, by setting up a performance testing environment that allows a more meaningful comparison of different logics under production conditions. I will open PRs with more individual changes once I was able to do this. I reached a similar conclusion for #3329 as well (https://github.com/TheThingsNetwork/lorawan-stack/pull/3329#issuecomment-712410891).

I will close this for now and open new PRs with regards to limiting the maximum number of commits as well as removing the unnecessary error notifications.

PR #3331, opened 2020-10-07
Introduce CompatibilityVersion in Redis registry

rvolosatovs:
#### Summary

Introduce `CompatibilityVersion` in Redis registry.
Follow-up of https://github.com/TheThingsNetwork/lorawan-stack/pull/2837

#### Changes

- Introduce `CompatibilityVersion` in Redis registry.

#### Testing

Integration testing with physical device

#### Notes for Reviewers

Current Network Server registry implementation is not completely
backwards-compatible due to an optimization. Disable this optimization
by default. Once I am done with the migration tool, we can set
`CompatibilityVersion` to `3.10` or higher and then we can benefit from
the write optimization. At that point I will add the configuration to
config as well - for now there's no use for it. (this could only be used
on new deployments with empty registries and we don't plan that any time
soon)

#### Checklist

### DIFF
--- a/pkg/networkserver/redis/registry.go
+++ b/pkg/networkserver/redis/registry.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/blang/semver"
 	"github.com/go-redis/redis/v7"
 	pbtypes "github.com/gogo/protobuf/types"
 	"github.com/oklog/ulid/v2"
@@ -58,6 +59,8 @@ const (
 type DeviceRegistry struct {
 	Redis   *ttnredis.Client
 	LockTTL time.Duration
+	// CompatibilityVersion denotes the lowest possible stack version the registry should be compatible with.
+	CompatibiltyVersion semver.Version
 
 	entropyMu *sync.Mutex
 	entropy   io.Reader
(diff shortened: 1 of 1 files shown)


rvolosatovs:
To be clear: this results in all relevant non-zero fields to be written on each device write. The list of these fields can be seen in the diff.

pkg/networkserver/redis/registry.go
```diff
@@ -58,6 +59,8 @@ const (
 type DeviceRegistry struct {
 	Redis   *ttnredis.Client
 	LockTTL time.Duration
+	// CompatibilityVersion denotes the lowest possible stack version the registry should be compatible with.
+	CompatibiltyVersion semver.Version
```
johanstokking, line 63: Where is this set?

adriansmares (commented):
Nothing to add.

pkg/networkserver/redis/registry.go
rvolosatovs: See OP:
> At that point I will add the configuration to config as well - for now there's no use for it. (this could only be used on new deployments with empty registries and we don't plan that any time soon)

This will be set when there's a reason to do so.

adriansmares: While indeed we won't indeed deploy a new cluster too soon, there are customers that run the enterprise stack themselves (or that can pickup the AMI for the Marketplace) which will start on a clean registry. So there is a need for it.

rvolosatovs: Ok, but then we have to discuss this, because this is something we probably want to have shared between components, maybe not - adding this to the config now means that we cannot change it later and I'm not very confident to do this just yet. We don't even know if this is the right approach to handling this issue.
So what I'm saying is let's file an issue, discuss and then implement accordingly.
This PR addresses the fact that 3.10 is currently broken and so is a bugfix. I do not want to add new features here.

adriansmares (approved):
Approved, but please create the associated issue.

PR #3343, opened 2020-10-09
Improve Pub/Sub MQTT driver context propagation

### LINKED ISSUE #3338: Change Paho MQTT WaitTimeout usage in Pub/Sub
#### Summary

We should make sure the client is closed after a failed `WaitTimeout` call - otherwise we are leaking the client which is actually still connecting. Consider the following connection code:

https://github.com/TheThingsNetwork/lorawan-stack/blob/a56e90acbd1e0e01bc0a6f5f509feda390933698/pkg/applicationserver/io/pubsub/provider/mqtt/provider.go#L148-L151

As the operation 'timed out', we immediately return an error and consider the connection to be failed - but this is not what happens actually:

https://github.com/eclipse/paho.mqtt.golang/blob/ba85050a1f239f4e954dc95920213db51f937df1/token.go#L78-L90

If the timeout finishes, the operation is **not** cancelled. This means that we leak a working client during connection if the timeout is hit.

This can happen under heavy load caused by multiple applications attempt to reconnect their link, because the [backoff](https://github.com/TheThingsNetwork/lorawan-stack/pull/3337) is not working correctly.

#### Steps to Reproduce

1. Stress the AS with a heavy amount of link failures.
2. Attempt to connect a Pub/Sub integration.
3. The Paho connection procedure does not get scheduled fast enough to actually occur before the timeout, or connecting simply takes too much.
4. `OpenConnection` timeouts and returns.
5. Some time later, the client actually connects.

#### What do you see now?

The MQTT connection is leaked.

The MQTT connection should always be closed on error.

#### Environment

`3.9.4`

Always call `client.Disconnect` if error is not `nil`.

Hard to test this correctly. Since it's not even reproducible correctly, testing it has the same issue.

Yes. cc @johanstokking . 

I also believe this is backport worthy to 3.9.x, as we will continue to experience this (and the lack of backoff) consistently until we upgrade to 3.10

adriansmares:
#### Summary

Closes #3338

#### Changes

- Upgrade Paho to `ba85050a1f23`
- Remove connection / publishing / subscription timeouts. Use context cancellations instead, as Paho [recently](https://github.com/eclipse/paho.mqtt.golang/pull/447) added support for it (only await, no cancellation for started operations unfortunately).
- Always disconnect the client if the connection attempt fails
- Linking and Pub/Sub tasks no longer stop on registry errors (except `NotFound` errors) - we've actually observed this in production when the link to Redis stops, and the task stops
- Do not cancel contexts with `nil` error. This makes code that uses `ctx.Done` and `ctx.Err` to miss the cancellation
- Add keep-alive to MQTT connections

#### Testing

Unit testing covers this.

##### Regressions

As this touches the MQTT driver timeouts, and a bit of the linking/pub/sub startup logic, regressions can occur there. However, the changes actually make us more lenient in task restarts (since we now actually propagate the errors better).

#### Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,8 +16,8 @@ replace gopkg.in/DATA-DOG/go-sqlmock.v1 => gopkg.in/DATA-DOG/go-sqlmock.v1 v1.3.
 // Versions higher trigger google/protobuf update past v1.3.5.
 replace gocloud.dev => gocloud.dev v0.19.0
 
-// Remove this once https://github.com/eclipse/paho.mqtt.golang/pull/451 is merged.
-replace github.com/eclipse/paho.mqtt.golang => github.com/johanstokking/paho.mqtt.golang v1.2.1-0.20200917115254-cf2daad9593e
+// Use the latest master in order to avoid accidental downgrades to v1.2.0.
+replace github.com/eclipse/paho.mqtt.golang => github.com/eclipse/paho.mqtt.golang v1.2.1-0.20200918111050-ba85050a1f23
 
 require (
 	contrib.go.opencensus.io/exporter/prometheus v0.1.0
@@ -32,7 +32,7 @@ require (
 	github.com/disintegration/imaging v1.6.2
 	github.com/dlclark/regexp2 v1.2.1 // indirect
 	github.com/dop251/goja v0.0.0-20200824171909-536f9d946569
-	github.com/eclipse/paho.mqtt.golang v1.2.1-0.20200609161119-ca94c5368c77
+	github.com/eclipse/paho.mqtt.golang v1.2.1-0.20200918111050-ba85050a1f23
 	github.com/envoyproxy/protoc-gen-validate v0.4.0
 	github.com/felixge/httpsnoop v1.0.1
 	github.com/fsnotify/fsnotify v1.4.9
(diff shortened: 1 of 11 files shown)


pkg/applicationserver/io/pubsub/provider/mqtt/driver.go
```diff
@@ -264,10 +260,8 @@ func (s *subscription) Close() error {
 		return nil
 	}
 	token := s.client.Unsubscribe(s.topic)
-	if !token.WaitTimeout(timeout) {
-		return errUnsubscribeFailed.WithCause(context.DeadlineExceeded)
-	} else if token.Error() != nil {
-		return errUnsubscribeFailed.WithCause(token.Error())
+	if err := waitToken(s.ctx, token); err != nil {
```
johanstokking, line 263: I can imagine that the caller calls `Close()` when the context is already canceled, immediately returning `waitToken()`. I wonder what the value of `ctx` is in `Close()`?

Should we use a new context with short timeout here, or are we really sure we can use the context?

adriansmares: `s.ctx` is the context that was used to create the subscription, which is the context that was used to open the provider connection. You are correct that on a pub/sub cancellation, `waitToken` will return immediately. 

But if that context is closed/expired - the connection is getting closed as well, so is there any value in waiting ? Waiting an unsubscribe here just to kill the connection afterwards seems a bit pointless. 

We can of course extend `context.Background()`, but do we get much out of it ?

go.mod
```diff
@@ -16,9 +16,6 @@ replace gopkg.in/DATA-DOG/go-sqlmock.v1 => gopkg.in/DATA-DOG/go-sqlmock.v1 v1.3.
 // Versions higher trigger google/protobuf update past v1.3.5.
 replace gocloud.dev => gocloud.dev v0.19.0
 
-// Remove this once https://github.com/eclipse/paho.mqtt.golang/pull/451 is merged.
-replace github.com/eclipse/paho.mqtt.golang => github.com/johanstokking/paho.mqtt.golang v1.2.1-0.20200917115254-cf2daad9593e
```
htdvisser, line 20: Probably smart to keep a replacement to the latest master, otherwise a dependency upgrade may accidentally revert it to v1.2.0.

pkg/applicationserver/io/pubsub/provider/mqtt/driver.go
johanstokking: OK fine to use `s.ctx` then.

htdvisser (approved):

johanstokking (approved):

PR #3344, opened 2020-10-09
Allow reusing DevAddr when required

rvolosatovs:
#### Summary

Allow reusing DevAddr when required

#### Changes

- Allow reusing DevAddr when required
- Limit the amount of DevAddr generation retries if a new, unused
  DevAddr cannot be generated
- Properly handle cases when DevAddr is reused when rematching device on
  data uplink

#### Testing

Integration testing with physical devices

#### Notes for Reviewers

This allows setting `ns.dev-addr-prefixes` to `DEADBEEF/32`, for example
to have all devices use the same address(`DEADBEEF`). Note, that we still do not
expect to reuse DevAddr and so when this happens, ERROR log level is
used.

#### Checklist

### DIFF
--- a/pkg/networkserver/grpc_gsns.go
+++ b/pkg/networkserver/grpc_gsns.go
@@ -857,72 +857,79 @@ func (ns *NetworkServer) handleDataUplink(ctx context.Context, up *ttnpb.UplinkM
 				return nil, nil, errOutdatedData.New()
 			}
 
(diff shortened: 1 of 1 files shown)


rvolosatovs:
Also note, that currently setting `ns.dev-addr-prefixes` to `XXXXXXXX/32` results in an infinite loop on join

adriansmares (approved):

PR #3363, opened 2020-10-14
console: Fix creation of multicast devices

### LINKED ISSUE #3361: Cannot create multicast end devices in the Console
#### Summary

It is not possible to create multicast end devices via the Console.
CLI works fine.

#### Steps to Reproduce

1. Go to the end device wizard
2. Create multicast device
3. Go to the general settings of the new end device
4. See that it is registered as an ABP device

#### What do you see now?

Multicast devices are registered as ABP

Multicast devices must be registered as multicast devices

Most probably related to validation schema of the network settings form

Manually create multicast devices

yes

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/3361

#### Changes

- Allow settings the `multicast` entry in NS.
- Add validation to require class C for multicast devices (multicast devices cannot be class A and class B will be added in https://github.com/TheThingsNetwork/lorawan-stack/pull/3065)

#### Testing

Created and edited multicast devices.

#### Notes for Reviewers

Class B and e2e tests will added with https://github.com/TheThingsNetwork/lorawan-stack/pull/3065

#### Checklist


pkg/webui/console/views/device-add/wizard/network-settings-form/validation-schema.js
```diff
@@ -21,7 +21,13 @@ const validationSchema = Yup.object({
...
   lorawan_version: Yup.string().required(sharedMessages.validateRequired),
   lorawan_phy_version: Yup.string().required(sharedMessages.validateRequired),
-  supports_class_c: Yup.boolean().default(false),
+  supports_class_c: Yup.boolean().when(['$activationMode'], (mode, schema) => {
+    if (mode === ACTIVATION_MODES.MULTICAST) {
+      return schema.oneOf([true], sharedMessages.validateRequired)
+    }
+
+    return schema.default(false)
+  }),
```
kschiffer: This needs to work in a way that the `Supports Class C` checkbox will be checked by default while the checkbox will be disabled. Please also add a field description in these cases that the device has to support class c when using multicast.

sdk/js/generated/device-entity-map.json
```diff
@@ -415,7 +415,7 @@
   },
   "max_frequency": ["ns", "ns"],
   "min_frequency": ["ns", "ns"],
-  "multicast": ["ns", "read_only"],
```
kschiffer, line 418: This file is generated from `sdk/js/generated/device-field-masks.json`. Changes to this file need to be sourced from there, otherwise this change will be lost, when `device-entity-map.json` is regenerated.

As far as I remember, `device-field-masks.json` is currently still generated by hand using via `pkg/ttnpb/field_mask_validation.go`

pkg/webui/console/views/device-add/wizard/network-settings-form/index.js
```diff
@@ -117,6 +117,7 @@ const NetworkSettingsForm = props => {
         title={sharedMessages.supportsClassC}
         name="supports_class_c"
         component={Checkbox}
+        required={isMulticast}
```
kschiffer:
```suggestion
```

kschiffer (changes requested):

pkg/webui/console/views/device-add/wizard/network-settings-form/index.js
bafonins: why?

pkg/webui/console/views/device-add/wizard/network-settings-form/validation-schema.js
bafonins: > Please also add a field description in these cases that the device has to support class c when using multicast.

This is technically incorrect. Multicast devices must support either class b or c, we currently do not have functionality to create class b compatible devices via the console, but it is possible via the CLI. Support for class B will be added in https://github.com/TheThingsNetwork/lorawan-stack/pull/3065 as it requires fields from `mac_settings`.
Adding description that a multicast device _has to support class c_ raises more confusion for both experienced and novice users than adds any value.

Currently, the problem is that the end device wizard has option to register multicast devices but it is not working. With this PR merged we at least can allow creating multicast devices via the console by allowing only class c compatible devices regardless of https://github.com/TheThingsNetwork/lorawan-stack/pull/3065 being merged within the same release as this fix or in the next release.

pkg/webui/console/views/device-add/wizard/network-settings-form/index.js
kschiffer: Marking checkboxes as required communicates that the option has to be checked consciously in order to continue. This is usually used to ensure that the user is aware of a certain condition or consequence (e.g. agreeing to ToS). In this case, I don't think we need this explicit confirmation and instead, we'll prefill and lock the checkbox field with the mandatory option. Therefore, it's also not necessary to mark the checkbox field as required, just like it also isn't under normal circumstances.

pkg/webui/console/views/device-add/wizard/network-settings-form/validation-schema.js
kschiffer: In any case, I'd prefer a solution that does not force the user to click a checkbox to proceed. If we know this option has to be checked then we should preselect that value and prevent changes to the field by disabling it. The confusion you mention will occur just as much when the user is faced with a generic validation message that the field is required.

Anyway, since the logic will be changed soon anyway via #3065, it's not worth spending more time on perfecting this now.

kschiffer (approved):

PR #3391, opened 2020-10-21
Fix logout redirect loops for clients without logout redirect URIs

### LINKED ISSUE #3371: Setting up OAuth clients without logout redirect URI can lead to redirect loops
#### Summary
When oauth clients are setup without `--logout-redirect-uri`, it can lead to the redirect URI being set as `/` which will cause infinite redirect loops to the logout endpoint of the OAuth backend.

#### Steps to Reproduce

1. Setup a client without `--logout-redirect-uri`
2. Logout from the console
3. Experience an `ERR_TOO_MANY_REDIRECTS` error, which will be communicated differently depending on the browser

#### What do you see now?
Redirect loops happening when clients don't use a logout redirect URI

No redirect loops happening as a result.

#### Environment
v3.10.x

If no redirect URI is set, the oauth client will use it's own mount path `/oauth` as redirect URI. There seems to be a bug which transforms this value to ``.

A unit test should be added to catch issues like these.

Yes.

kschiffer:
#### Summary
Closes #3371

#### Changes
- Use the OAuth mount path as default redirect URI

#### Testing

Manual and unit tests.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - End device import in the Console crashing in Firefox.
 - Creation of multicast end devices in the Console.
 - Overwriting values in the end device wizard in the Console.
+- Redirect loops when logging out of the Console if the Console OAuth client had no logout redirect URI(s) set.
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


htdvisser (approved):
LGTM

PR #3393, opened 2020-10-21
Skip token exchange when no state cookie is present

### LINKED ISSUE #3351: Skip OAuth callback if an auth cookie already exists
#### Summary
When the user tries to re-run the OAuth callback endpoint (e.g. when clicking the browser's back button after logging in) they will see an error message (`Bad request`). It would be better UX to just skip the token exchange process and redirect straight back to the application root.

#### Why do we need this?
It's unnecessary to try running the token exchange (and consequently displaying an error) if the access cookie already exists.

An error message `Bad request`.

A simple redirect to the app root.

Simply add a check whether the access cookie already exists and skip the token exchange by redirecting to the application root instead.

Unit tests.

Yes.

kschiffer:
#### Summary
This PR changes the `oauthclient` token exchange handler so that it skips the token exchange if a valid auth cookie is present but no state cookie. This situation often occurs when users use the back button of the browser after logging in (e.g. to the Console)

Closes #3351 

#### Changes
- Check for the auth cookie when no state cookie is present and redirect to the client's mount path

#### Testing

Manual.

##### Regressions

This might introduce issues regarding the OAuth token exchange.

#### Checklist

### DIFF
--- a/pkg/web/oauthclient/callback.go
+++ b/pkg/web/oauthclient/callback.go
@@ -51,6 +51,15 @@ func (oc *OAuthClient) HandleCallback(c echo.Context) error {
 
 	stateCookie, err := oc.getStateCookie(c)
 	if err != nil {
+		// Running the callback without state cookie often occurs when re-running
+		// the callback after successful token exchange (e.g. using the browser's
+		// back button after logging in). If there is a valid auth cookie, we just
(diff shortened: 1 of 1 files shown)


htdvisser (approved):
LGTM

Issue #2681, opened 2020-06-09
Show error descriptions for 5xx server errors


bafonins:
#### Summary

Currently, console/oauth app shows a generic error description on 5xx errors returned from the backend. We should provide more details on what exactly failed. 

#### Why do we need this?

See https://github.com/TheThingsIndustries/lorawan-stack/issues/2151

#### What is already there? What do you see now?

1. Error boundaries that take care of catching errors and rendering the error view.
2. Error utilities that parse errors

#### What is missing? What do you want to see?

Add checks to the error view for 5xx errors and add translations for the error description.

As an example:

* 500: there is an internal problem with the server {address}
* 502: there is a problem with the proxy {address}
* 503: the server {address} is temporarily unavailable
* 504, 522, 524: the connection to {address} timed out

#### Environment

v3.8.3

#### How do you propose to implement this?

Similar to how we handle 404 not found errors

#### How do you propose to test this?

Configure NS,AS,JS,IS as enabled, but do not start the actual stack components. Then open the device overview page. The JS SDK should make requests to the disabled component and display the error view.

#### Can you do this yourself and submit a Pull Request?

yes

bafonins:
This could be combined with https://github.com/TheThingsNetwork/lorawan-stack/issues/3975

kschiffer:
Should also be addressed by #4480

PR #3397, opened 2020-10-22
Navigate to failed step in end device wizard

### LINKED ISSUE #3291: Navigate to failed step in end device wizard
#### Summary

We should navigate the user to wizard step that failed.

#### Why do we need this?

We make api requests only when submitting the last form in the end device wizard. Since the whole device creation flow is quite complicated and frontend validation might be not up to date to the actual validation rules defined on the backend, any request can fail. Currently, we display errors without navigating the user to the actual failed step.

Consider a simple case when a user trying to create a new device (OTAA) with already existing ID. With our current implementation the error will be presented at the Join Settings step, while the actual step caused the error is the Basic Settings step.

1. End device wizard
2. API in the JS SDK

1. Functionality in `<Wizard />` to navigate to arbitrary step.
2. Functionality to check the component that returned an error.

TBA

e2e tests, manual testing

yes

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/3291

wizard-error

<details>
<summary>Submit bar fix screenshots</summary>
Screenshot 2020-10-22 at 14 02 22
Screenshot 2020-10-26 at 16 11 23
Screenshot 2020-10-26 at 16 11 29

</details>

#### Changes

- Add step id and rely on it instead of step index
- Extend wizard actions
- Export stack configuration map from the js sdk
- Augment errors originating in the js sdk to include additional information about failed request (method, url, failed component, params)
- Change `net ID` to `NetID` in the join settings form (see https://github.com/TheThingsNetwork/lorawan-stack/pull/3065#issuecomment-705409095)
- Fix `<SubmitBar />` buttons for small screens

##### Regressions

Navigation in the end device wizard.

#### Notes for Reviewers

1. `request_details` in errors - Can also be helpful for https://github.com/TheThingsNetwork/lorawan-stack/issues/2681.
2. Current implementation works fine for various scenarios I tested, however it is not possible to test all possible inputs so there might be issues with navigating to the actual failed step because some values are accumulated from previous steps (e.g. device `ids`). Proper solution to avoid such uncertainty would be to submit forms on each step instead of doing this only after the last one.

#### Checklist

### DIFF
--- a/pkg/webui/components/wizard/wizard.js
+++ b/pkg/webui/components/wizard/wizard.js
@@ -25,6 +25,8 @@ const FIRST_STEP = 0
 // Action types.
 const INIT = 'INIT'
 const GO_TO_STEP = 'GO_TO_STEP'
+const SET_ERROR = 'SET_ERROR'
+const COMPLETE_STEP = 'COMPLETE_STEP'
 
 const reducer = (state, action) => {
   switch (action.type) {
@@ -34,22 +36,33 @@ const reducer = (state, action) => {
         steps: action.steps,
       }
     case GO_TO_STEP:
-      const { snapshots: oldSnapshots, currentStep: oldStep } = state
-      const { values, step: currentStep } = action
-
-      // Replace current step values when navigating between wizard steps.
-      if (values) {
+      return {
+        ...state,
+        currentStepId: action.step,
+        error: undefined,
       }
+    case SET_ERROR:
+      return {
+        ...state,
+        error: action.error,
+        currentStepId: action.step,
+      }
+    case COMPLETE_STEP:
+      const { snapshots: oldSnapshots, currentStepId: oldStepId, steps } = state
+      const { values } = action
+
+      const oldStepIndex = steps.findIndex(({ id }) => id === oldStepId)
+      // Replace current step values when navigating between wizard steps.
       const snapshots = [
-        ...oldSnapshots.slice(0, oldStep - 1),
+        ...oldSnapshots.slice(0, oldStepIndex),
         values,
-        ...oldSnapshots.slice(oldStep),
+        ...oldSnapshots.slice(oldStepIndex + 1),
       ]
 
       return {
         ...state,
-        currentStep,
         snapshots,
+        error: undefined,
       }
     default:
       return state
@@ -60,7 +73,7 @@ const Wizard = React.forwardRef((props, ref) => {
   const { initialStepId, onComplete, initialValues, completeMessage } = props
 
   const [state, dispatch] = React.useReducer(reducer, {
-    // Active step in the wizard.
+    error: undefined,
     currentStepId: initialStepId,
     // A list of all steps in the wizard.
     steps: [],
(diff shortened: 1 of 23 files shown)


sdk/js/src/api/http.js
```diff
@@ -112,7 +113,16 @@ class Http {
...
       if ('response' in err && err.response && 'data' in err.response) {
-        throw err.response.data
+        const error = cloneDeep(err.response.data)
+        error.request_details = {
+          url: get(err, 'response.config.url'),
+          method: get(err, 'response.config.method'),
+          params: get(err, 'response.config.params'),
+          base_url: get(err, 'response.config.baseURL'),
+          stack_component: parsedComponent,
+        }
```
kschiffer: Not too happy about this. I'd prefer to leave the error responses unmanipulated.

IMO, a better option would be to change the output for device CRUD functions in the SDK so that when an error occurs, they return an object of errors keyed by component ID. Currently, we just return the first error that occurred.

sdk/js/src/api/http.js
```diff
@@ -141,7 +151,7 @@ class Http {
    * Extracts The Things Stack component abbreviation from the endpoint.
    *
    * @param {string} endpoint - The endpoint got for a request method.
-   * @returns {string} One of {is|as|gs|js|ns}.
+   * @returns {string} One of {is|as|gs|js|ns|edtc|qrg|gcs}.
```
kschiffer:
```suggestion
   * @returns {string} The stack component abbreviation.
```

We will also forget to update this next time for sure.

pkg/webui/components/wizard/wizard.js
```diff
@@ -34,88 +36,120 @@ const reducer = (state, action) => {
...
-  const { currentStep, steps, snapshots } = state
-
-  const stepsCount = steps.length
+  const { currentStepId, steps, snapshots, error } = state
 
   const stepsInit = React.useCallback(steps => {
     dispatch({ type: INIT, steps })
   }, [])
+  const completeStep = React.useCallback(values => {
+    console.log('kekster mani suka')
```
kschiffer: Remove.

kschiffer (changes requested):
The changes regarding the next/prev button layouts in small screensizes don't seem to be committed here.

Please implement it like this:
- Retain the text of the "next step" and "Submit" button
- Hide the text of the "back" button
- Overwrite the default value of the `min-width` for the back button to `auto`

sdk/js/src/api/http.js
bafonins: > Not too happy about this.

Me neither, but its the simplest solution that can be changed at any moment, does not require any api changes in the console/sdk and uses functionality that we already have in the sdk. Also, this provides additional debug info (http method, full url and target stack component can be helpful).

> IMO, a better option would be to change the output for device CRUD functions in the SDK so that when an error occurs, they return an object of errors keyed by component ID. 

Then I would expect device CRUD methods to return successful results in a similar shape.
```
{
  is: {... device fields stored in is ...},
  as: {... device fields stored in as ...},
  ...
}
```

We can adjust this by changing signatures of device crud methods, for example
```
async create(applicationId, device, mask = Marshaler.fieldMaskFromPatch(device))
// can be changed to:
async create(applicationId, device, mask = Marshaler.fieldMaskFromPatch(device), mergeResult = true)
// or
async create(applicationId, device, mask = Marshaler.fieldMaskFromPatch(device), { mergeResult = true })
// or
async create(applicationId, device, { mask = Marshaler.fieldMaskFromPatch(device), mergeResult = true })
```

Basically when `mergeResult` is `true` then both merged result and first error is returned, otherwise we return results split by component, see above. This will require some changes to end device redux primitives.

I can add utility functions to `lib/errors/utils.js` to get the target component and use those instead of directly accessing `request_details` in the device-add wizard. So later we can switch to other implementation (we can check for failed component by error message or anything else)

kschiffer: Indeed, I was thinking of a solution like that. Can you put this into a new issue and add a `TODO:` here?

Note that when manipulating the error object, these things will be part of the displayed error details in the Console. Meaning that they won't represent the backend error 1:1 anymore. This might confuse others when trying to debug.

kschiffer (approved):
See my comment about adding a `TODO:`

PR #3415, opened 2020-10-27
Various improvements related to Redis task queues

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2698
Required by https://github.com/TheThingsNetwork/lorawan-stack/pull/3398

#### Changes

- Update Redis to v8 (out of notable changes - it now fully supports
  context cancellation and deadlines) ~(#3438 )~
- Make Redis task dispatch lazy and consumer-initiated, i.e. when the work
  queues are empty, one of the consumers performs the dispatch of tasks.
  This ensures correctness, and simplifies scaling as well as
  implementation. Under very heavy load this could potentially lead to bad
  performance, since then pop/dispatch script could task a long time to
  execute. As previously mentioned by @adriansmares, we should
  investigate designating separate Redis instances to different
  components/features to mitigate this if it ever becomes a problem.
- Clean-up Network Server downlink tasks

#### Testing

Unit tests, integration testing

##### Regressions

Potentially downlink tasks could break, but I am performing extensive
testing to avoid that
Aditionally, there might be some unexpected side-effects from the Redis
lib update

#### Notes for Reviewers

~Would be nice to test this in staging and deploy as part of first proper 3.10 release.~

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [x] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #3398, opened 2020-10-22
Use NsAs interface in Network Server instead of linking

rvolosatovs:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2312 #3263

#### Changes

- Remove `LinkApplication` from NS API
- Use `NsAs.HandleUplink`
- Only send FPort>0 uplinks to AS
- ~Optimize Redis task queue impl~ (see #3415 )

#### Testing

Unit tests, integration testing

##### Regressions

Application uplink flow

#### Notes for Reviewers

@adriansmares pushing downlinks to AS does not work for me, so it's quite difficult to test properly:
```
 DEBUG Finished unary call                      duration=18.953473ms error=rpc error: code = NotFound desc = error:pkg/redis:not_found (entity not found) error_correlation_id=c64c7144914c463692e8bf84f95e9444 error_name=not_found error_namespace=pkg/redis grpc_code=NotFound grpc_method=DownlinkQueuePush grpc_service=ttn.lorawan.v3.AppAs namespace=grpc
error:pkg/redis:not_found (entity not found)
```

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [x] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #3417, opened 2020-10-27
Merge 3.10

rvolosatovs:
Merge 3.10


rvolosatovs:
Ideally I would like to include https://github.com/TheThingsNetwork/lorawan-stack/pull/3415 here as well, such that #3398 can be unblocked and so made ready for review

Issue #2871, opened 2020-07-09
Reduce size of EndDevice objects in Redis


htdvisser:
#### Summary

We need to reduce the size of the EndDevice objects we store in Redis.

#### Why do we need this?

1. Because all of this is stored in memory, and with 250kB per end device operators are going to have very expensive infrastructure bills.
2. Because we're currently still fetching full EndDevice protos for a non-unique DevAddr, which leads to a lot of network traffic, which should also be reduced.

#### What is already there? What do you see now?

Looking at the heavy parts, I see

- `mac_state.recent_uplinks`
  - Both `raw_payload` and decoded `payload`. If we need the payloads in the first place, we definitely don't need both
  - `rx_metadata.*.packet_broker.hops`. This is already pretty heavy, and we'll have it for every duplicate. I also don't think we actually use it.
  - `correlation_ids`. I don't think we actually use those.
- `mac_state.recent_downlinks`
  - `correlation_ids`. I don't think we actually use those.
- `recent_adr_uplinks`.
  - I don't think we need `raw_payload` and decoded `payload`.
  - From `rx_metadata.*` I think we only need RSSI and SNR
- `recent_uplinks`. Is this just a copy of `mac_state.recent_uplinks`?
  - Same story as for `mac_state.recent_uplinks`
- `recent_downlinks`. Is this just a copy of `mac_state.recent_downlinks`?
  - Same story as for `mac_state.recent_downlinks`

#### What is missing? What do you want to see?

Remove the bloat.

#### Environment

v3.8.5

#### How do you propose to implement this?

- Filter the fields before storing
- Don't store double data

#### How do you propose to test this?

- Observe the size of stored protos, observe the `redis_client_*` metrics.

rvolosatovs:
Removing data we currently provide is breaking API, so I don't think we can do that in this major.
`recent_uplinks`/`recent_downlinks` have different meaning than `mac_state.recent_uplinks`/`mac_state.recent_downlinks` respectively. `mac_state.recent_uplinks` only contains last 20 data uplinks within a session, `recent_uplinks` contains everything - e.g. join-requests etc. Moreover, e.g. NS uses both `recent_uplinks` and `mac_state.recent_uplinks`
While we might not use various data, we cannot remove it, because our users might.
So unless it's not possible to `GET` a field via device registry, we cannot remove it, I'm afraid.

I think we can remove the payloads though indeed.

[...]

PR #3421, opened 2020-10-28
Optimize Network Server device storage

rvolosatovs:
#### Summary

Optimize Network Server device storage. Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2871

#### Changes

- Get rid of `recent_uplinks`, `recent_downlinks` and
  `recent_adr_uplinks`
- Add MACState.LastADRChangeFCntUp
- Fix DevStatus FCnt-bound calculation
- Avoid storing raw payloads of frames

#### Testing

Made a OTAA class C device join and send traffic on merge base, switched to head, verified that traffic still flows and rejoin works

##### Regressions

Potentially ADR and downlink flow could break.

#### Notes for reviewers

I did not address all the suggestions in the original issue, but rather the heaviest fields and the low-hanging fruit. I will come up with a migration tool, which would clean-up the legacy fields, then we can profile it, see how this performs and go from there (i.e. to optimize this further).

#### Checklist


pkg/networkserver/mac/adr.go
```diff
@@ -145,10 +145,21 @@ func txPowerStep(phy *band.Band, from, to uint8) float32 {
...
-	if len(dev.RecentADRUplinks) == 0 {
+	if dev.MACState == nil || len(dev.MACState.RecentUplinks) == 0 {
 		return nil
 	}
 
+	adrUplinks := dev.MACState.RecentUplinks
+	for i := len(dev.MACState.RecentUplinks) - 1; i >= 0; i-- {
+		up := dev.MACState.RecentUplinks[i]
+		if (up.Payload.MType == ttnpb.MType_UNCONFIRMED_UP || up.Payload.MType == ttnpb.MType_CONFIRMED_UP) &&
+			up.Payload.GetMACPayload().FCnt >= dev.MACState.LastADRChangeFCntUp {
```
johanstokking: Shouldn't this be comparing full FCnt? Assuming that the payload is not being changed.

johanstokking (approved):

PR #3428, opened 2020-11-01
Do not redirect to port 443 if a different port is configured

### LINKED ISSUE #3118: `--http.redirect-to-tls` option always redirects to port 443
#### Summary

Using stack with `--http.redirect-to-tls` always redirects traffic to port 443.

#### Steps to Reproduce

1. `ttn-lw-stack start --http.redirect-to-tls`
2. `curl https://localhost:8885`

Several parts come into play here:

1. https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.9/pkg/web/web.go#L198
2. https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.9/pkg/webmiddleware/redirect.go#L65
3. https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.9/pkg/component/web.go#L61

In (1), the value from the map is always used, which is `0` for non-existing entries. In (2), when the port number is `0`, then the hostname is returned without any port, which causes the redirect to the default HTTPS port instead (443).

#### What do you see now?

```
$ curl https://localhost:8885 -k
<a href="https://localhost/">Found</a>.

$ curl https://localhost:8885 -k -L
curl: (7) Failed to connect to localhost port 443: Connection refused

$ http://localhost:1885 -k
<a href="https://localhost:8885/">Found</a>.

$ curl http://localhost:1885 -k -L
curl: (7) Failed to connect to localhost port 443: Connection refused
```

Redirect to the correct HTTPS port. 

#### Environment

3.9.1

This needs delicate handling in order to avoid breaking existing installations. We cannot simply redirect to the HTTPS port all the time, since that will break Docker installations (where stack is listening on container port `:8885` but traffic is received/sent over the port `:443`.

Overall, this is a bug, but we cannot simply fix it, because it could end up breaking existing deployments (e.g. by suddenly redirecting traffic to port 8885). I think the cleanest solution to this would be to add a `--http.redirect-to-tls-port=<integer>`, defaulting to `443`. This is backwards compatible with existing configs, and can be used to always redirect to the correct HTTPS port.

Test locally and in a Docker deployment.

yes, but let's discuss before diving into this.

neoaggelos:
#### Summary

Closes #3118 

#### Changes

- When `--http.redirect-to-tls` is set, only configure `redirectConfig.Port` if it is not set already.

#### Testing

- Test locally with `curl`, test with built Docker image.
- Test with `--http.redirect-to-tls` --> redirects to port 443, preserving existing behaviour
- Test with `--http.redirect-to-tls --http.redirect-to-host=my.host` --> redirects to `https://my.host`, preserving existing behaviour
- Test with `--http.redirect-to-tls --http.redirect-to-host=my.host:8885` --> redirects to `https://my.host:8885` (without this PR, the stack redirects to `https://my.host`)

##### Regressions

There should not be any. The fixed bug would only occur when the HTTPS server was accessed at a port other than 443 with `redirect-to-tls` enabled. The comment on the CHANGELOG should be enough to clear out any confusion.

#### Notes for Reviewers

This seems to be the least "breaking" way to deal with the bug discussed in the issue.

It would be best if we could get this in `3.10.0` (or wait for `3.11.0`). I do not think it's good to introduce this change on a patchlevel release.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ For details about compatibility between different releases, see the **Commitment
 
 - Next button title in the end device wizard in the Console.
 - Navigation to the user edit page after creation in the Console.
+- The port number of the `--http.redirect-to-host` option was ignored when `--http.redirect-to-tls` was used. This could lead to situations where the HTTPS server would always redirect to port 443, even if a different one was specified.
+  - If the HTTPS server is available on `https://thethings.example.com:8443`, the following config is required: `--http.redirect-to-tls --http.redirect-to-host=thethings.example.com:8443`.
 
 ### Security
 
--- a/pkg/web/web.go
+++ b/pkg/web/web.go
@@ -194,8 +194,11 @@ func New(ctx context.Context, opts ...Option) (*Server, error) {
 	}
 	if options.redirectToHTTPS != nil {
 		redirectConfig.Scheme = func(string) string { return "https" }
-		redirectConfig.Port = func(current uint) uint {
-			return uint(options.redirectToHTTPS[int(current)])
+		// Only redirect to HTTPS port if no port redirection has been configured
+		if redirectConfig.Port == nil {
+			redirectConfig.Port = func(current uint) uint {
+				return uint(options.redirectToHTTPS[int(current)])
+			}
 		}
 	}
 


htdvisser (approved):
LGTM. Please make sure the CHANGELOG entry ends up in unreleased, instead of v3.10.0

kschiffer (approved):

PR #3436, opened 2020-11-03
console: Added hint to user entries

### LINKED ISSUE #3429: Add `(This is you)` hint to user entries in user-management tables
#### Summary

Tables that show user entries should append a `(This is you)` message next to the user id, to improve clarity.

See https://github.com/TheThingsNetwork/lorawan-stack/issues/1563

#### Why do we need this?

Improve clarity for the user.

Plain display of the user id.

A hint that an entry refers to the current user when navigating to `/console/admin/user-management`.

#### Environment

v3.10

See https://github.com/TheThingsNetwork/lorawan-stack/pull/3407

Manual testing

@asmulko

asmulko:
#### Summary
Closes #3429 

#### Changes

- Added hint to user entries in user-management table.
- Updated collaborators table to use shared message (This is you)

Previously:

Now:

#### Testing

- Manually tested

#### Checklist


pkg/webui/console/containers/users-table/index.js
```diff
@@ -133,14 +150,15 @@ export default class UsersTable extends Component {
...
-        headers={headers}
+        headers={this.headers}
         addMessage={sharedMessages.userAdd}
         tableTitle={<Message content={sharedMessages.users} />}
         getItemsAction={this.getUsersList}
         searchItemsAction={this.getUsersList}
         baseDataSelector={this.baseDataSelector}
         pageSize={pageSize}
         searchable
+        {...this.props}
```
pgalic96: I think it's better to avoid using spread operators for props, especially if it initially works and we defined all props that are supplied to the component. Is this right @kschiffer ?

pgalic96 (approved):
Looks good, please use imperative verb for commit (Add instead of added). Check contributing guidelines on commit messages: https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.10/CONTRIBUTING.md

Also it is fine to just use the number without `#` for the branch name.

Also see one review comment and maybe wait for a review by @bafonins or @kschiffer as well.

pkg/webui/console/containers/users-table/index.js
kschiffer:
```suggestion
```

It depends. @asmulko, you likely added this because we do this in other table components as well. I was more lax regarding this until experiencing some issues. When spreading props, it's important to make sure that no unrelated props are being passed down, since that can lead to nasty bugs. To easily avoid that, it's usually better to be explicit about what props to pass (as in not using spread). In this particular case, there's also no use in passing the props.

kschiffer (approved):
LGTM, see my comment about removing the spread.

asmulko:
Thanks for the comments, fixed CI and removed unnecessary props

Issue #2446, opened 2020-04-29
Support resetting device state in Network Server


rvolosatovs:
#### Summary

Support resetting device state in Network Server

#### Why do we need this?

To support resetting the state after e.g. ABP device reset

#### What is already there? What do you see now?

Only possibility is to set stuff explicitly via device registry SET

#### What is missing? What do you want to see?

Functionality to reset the state in NS, just like it does on device creation.

#### How do you propose to implement this?

Just like NS already does when creating an ABP device

#### Can you do this yourself and submit a Pull Request?

yes

htdvisser:
This prevents some users from migrating from v2 to v3, so setting medium priority.

rvolosatovs:
@johanstokking coming back to the discussion we had during sync, this issue addresses specifically decreasing FCnt via device registry and resetting MAC state in response to that, such that a button in the console would then set the FCnt of the device to 0 or other value for reset

rvolosatovs:
After taking another look at this, it makes much more sense to introduce a separate RPC for this. PR's coming soon!

Issue #3433, opened 2020-11-03
Integrate Device Repository


johanstokking:
#### Summary

Integrate the LoRaWAN Device Repository: https://github.com/TheThingsNetwork/lorawan-devices.

This issue spans onboarding new end devices.

References #2813 

@neoaggelos is the implementer, work with @htdvisser for [index and search](https://blevesearch.com).

@kschiffer please design and implement Console UX for browsing end devices by brand, model and potentially firmware and hardware versions. Note that there will be thousands of end devices.

#### Why do we need this?

To make onboarding LoRaWAN devices easy.

#### What is already there? What do you see now?

Device creation wizard and general settings page for advanced device settings. This is quite overwhelming and error prone to users who have a "LoRaWAN device" in their hands and they only know the vendor and model.

#### What is missing? What do you want to see?

The user journey would be as follows:

1. User clicks **Add end device**
2. User gets to browse devices by brand, model, firmware version (if there are multiple) and hardware version (if there are multiple [supported hardware versions in the firmware version](https://github.com/TheThingsNetwork/lorawan-devices/blob/782727a8a99c150fda616b1293980b20e6796ee6/vendor/example/windsensor.yaml#L16-L18)). Below, the user can still start the current wizard for manual creation
3. When the user selected their brand, model and version, they still have to enter the device identifiers (for OTAA: JoinEUI, DevEUI and root key(s); for ABP: DevEUI, DevAddr and session keys). Whether the end device is OTAA or ABP is in the device repository. Also the MAC version is in the device repository, so only the relevant root key(s) and session keys are entered by the user
4. Create the end device from the template, also setting the payload formatter to repository

#### How do you propose to implement this?

1. Remove `EndDeviceVersion` message from the API; this is unused and superseded by `EndDeviceTemplate`. Consider also removing `EndDeviceBrand` and `EndDeviceModel`, unless you need it for the Device Repository API (see point 3 below)
2. Create a new first-class component Device Repository
3. Implement API, config and a service implementation for browsing end devices from the repository. This is much like the existing Webhook Templates (which are also being loaded as YAML), although we probably need **search**, **pagination**, loading a profile and codecs by `EndDeviceVersionIdentifiers` and serving static assets (like photos). As reference:
   1. API: https://github.com/TheThingsNetwork/lorawan-stack/blob/6f15ec162081c282a370008af834811d0df76efa/api/applicationserver_web.proto
[...]

Issue #2434, opened 2020-04-29
Resetting frame counts for end devices in the Console


kschiffer:
#### Summary
Users should be able to reset the end device frame counters.

#### Why do we need this?
Close feature gap with the v2 console

#### What is already there? What do you see now?
I guess the backend already supports this, but I need some info on that (cc @rvolosatovs)

#### What is missing? What do you want to see?
A way to trigger the frame counter reset in the console.

#### Environment
v3.7.1

#### How do you propose to implement this?
I'd say this could be a simple button inside of the Network Layer tab in the end device settings.

The reset button should be enabled when `session` is not set.

To reset, the Console will just have to set `session.last_f_cnt_up` to `0`.

#### Can you do this yourself and submit a Pull Request?
Yes, but I need some hint as to what API to use for this.

rvolosatovs:
Are we talking about the uplink frame counters here?
If so, resetting uplink frame counters must always include MAC state reset as well, which is more tricky and requires custom logic in Network Server. In any case Network Server automatically determines and handles uplink frame counter resets for devices, for which `resets_f_cnt == true`, I don't think we need an explicit way to reset uplink frame counters from the console. cc @johanstokking

johanstokking:
I think having a button "reset MAC state" with an info message that this also resets frame counters, is a good thing.

This is for devices that are not LoRaWAN compliant and loose their internal state on a power cycle. So yes, it is more than frame counters; it's resetting the whole thing, via the Console.

rvolosatovs:
> This is for devices that are not LoRaWAN compliant and loose their internal state on a power cycle. So yes, it is more than frame counters; it's resetting the whole thing, via the Console.

The devices, which are not LoRaWAN compliant in this case will send an uplink with frame counter lower than the previous one - Network Server will notice that and reset the MAC state without any user input.
See https://github.com/TheThingsNetwork/lorawan-stack/blob/0acff7cd288d114b43f6b178667cdcfb6a11c4ac/pkg/networkserver/grpc_gsns.go#L346-L416

While I agree that this is nice to have, this is definitely not a priority in my opinion and is a "goldplating" issue, since this is already handled automatically by the Network Server.

johanstokking:
> The devices, which are not LoRaWAN compliant in this case will send an uplink with frame counter lower than the previous one - Network Server will notice that and reset the MAC state without any user input.

[...]

PR #3437, opened 2020-11-03
Add Reset RPC to NS device registry

rvolosatovs:
#### Summary

Add Reset RPC to NS device registry. Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2446

#### Changes

- Add Reset RPC to NS device registry
- Support nil in test.FormatError
- Print attributes as part of test.FormatError
- Add `devices reset` command to CLI
- Avoid setting deprecated fields on GET
- Remove `TODO`, which we're not going *TODO* (https://github.com/TheThingsNetwork/lorawan-stack/issues/3433#issuecomment-721102872)

#### Testing

Unit tests and CLI integration testing

#### Notes for Reviewers

Adding a separate RPC for this made much more sense than handling this as part of FCnt decrease on SET. The endpoint works just like usual NS device GET, except it requires write right and it resets state. So clients can use the endpoint to reset state and immediately get new values of fields.

#### Checklist


rvolosatovs:
Blocking https://github.com/TheThingsNetwork/lorawan-stack/issues/2434

api/networkserver.proto
```diff
@@ -73,6 +73,14 @@ service NsEndDeviceRegistry {
     };
   };
 
+  // Reset resets device state.
+  rpc Reset(ResetEndDeviceRequest) returns (EndDevice) {
```
johanstokking: It is not clear to me from the rpc name (`Reset()`) what the scope is.

Is it resetting the session? What is the state? MAC state?

api/end_device.proto
```diff
@@ -610,6 +610,13 @@ message SetEndDeviceRequest {
   google.protobuf.FieldMask field_mask = 2 [(gogoproto.nullable) = false];
 }
 
+message ResetEndDeviceRequest {
+  EndDeviceIdentifiers end_device_ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (validate.rules).message.required = true];
+  // The names of the end device fields that should be returned.
+  // See the API reference for which fields can be returned by the different services.
+  google.protobuf.FieldMask field_mask = 2 [(gogoproto.nullable) = false];
+}
```
johanstokking: Even though it's documented that `field_mask` is for the "get" part, it's not clear otherwise that there's something to get. API clients may assume that the field mask indicates what is being reset.

Suggesting `RestAndGetEndDeviceRequest` or something

api/networkserver.proto
johanstokking: Maybe `ResetFactoryDefaults()` ?

rvolosatovs: The idea is that this resets all NS-controlled mutable state, so it's close to recreating the device

johanstokking (approved):

PR #3444, opened 2020-11-06
Fix Network Server field mask validation and Console usage

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/3437

#### Changes

- Removed `recent_uplinks` and `recent_downlinks` from the NS end device registry field mask validation. The fields are no longer used post #3437 
- Replaced the said fields with their `mac_state` and `pending_mac_state` variants in the field mask validation
- Console now uses `mac_state.recent_uplinks` for recent uplink status instead

#### Testing

I've reproduced this by trying to access a device in the Console. So did just that.

##### Regressions

The removed fields were already unavailable after #3437, so this just makes validation catch up to the registry behavior. No other regressions are expected.

#### Notes for Reviewers

I didn't add the fields to the `Set` RPC since they weren't there before. We can relax that later, should it be needed.

#### Checklist


pkg/ttnpb/field_mask_validation.go
```diff
@@ -313,6 +313,8 @@ var nsEndDeviceReadFieldPaths = [...]string{
 	"mac_state.queued_join_accept.request.selected_mac_version",
 	"mac_state.queued_responses",
 	"mac_state.rx_windows_available",
+	"mac_state.recent_downlinks",
```
rvolosatovs, line 315: please make sure this is sorted alphabetically

pkg/ttnpb/field_mask_validation.go
```diff
@@ -515,6 +517,8 @@ var nsEndDeviceReadFieldPaths = [...]string{
 	"pending_mac_state.queued_join_accept.request.selected_mac_version",
 	"pending_mac_state.queued_responses",
 	"pending_mac_state.rx_windows_available",
+	"pending_mac_state.recent_downlinks",
```
rvolosatovs, line 519: IDEM

rvolosatovs (approved):

pkg/ttnpb/field_mask_validation.go
adriansmares: Should I sort only `mac_state` lines only, or the whole slice ? I wanted to do this initially but the diff was large so I thought I'm missing something about the sorting.

rvolosatovs: The whole slice is supposed to be sorted and if it's not - we should fix that

rvolosatovs: Note the case-insensitive sorting!

rvolosatovs: In `vim`: `:sort i`

kschiffer (approved):
Codeowned files LGTM.

johanstokking (approved):

rvolosatovs (approved):

htdvisser (approved):
Nothing to add for `.go` files.

PR #3263, opened 2020-09-22
Remove AS-NS linking in the AS and NS

adriansmares:
#### Summary

Closes #2312 
References #3190 
References #3260 
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/3145

#### Changes

- Reintroduce the changes added by #3190

#### Testing

Covered by #3190 

##### Regressions

Covered by #3190 

#### Notes for Reviewers

I merged #3190 too fast, which left the `v3.10` branch in an unusable state. As such, I've reverted the merge in #3260 in order to make the branch usable again. 

This PR was generated using `git revert -m 1 62f404a75c37a93b431b0133927bcfd773404c4f` on 62f404a75.

As it was already approved, this PR should be used as a base for the NS changes, once @rvolosatovs has enough time to pick them up. We can then merge them into this branch and then into `v3.10`.

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #3439, opened 2020-11-04
Fix context lifetimes in asynchronous processing

johanstokking:
#### Summary

Fix context lifetime in asynchronous processing

Needed for https://github.com/TheThingsNetwork/lorawan-stack/pull/3438

#### Changes

- Decouple the request context lifetime on asynchronous message processing

#### Testing

Unit testing

##### Regressions

Should be none.

In fact, this may resolve race conditions when an asynchronous process depends on the lifetime of a closed request context.

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #3445, opened 2020-11-06
Application Server link setup fixes

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/3263

#### Changes

- Fix the `SetLink` gRPC endpoint - it was adding a non existing field mask field on link creation
- Use an empty link internally, when the link does not exist. This is for convenience - it ensures that link setup is not required in order to receive or send traffic. The sematics of `GetLink` have not been affected by this.
- Fix AS tests based on the https://github.com/TheThingsNetwork/lorawan-stack/pull/3415 `redis-go` update.

#### Testing

Local testing of downlink operations (based on @rvolosatovs report [here](https://github.com/TheThingsNetwork/lorawan-stack/pull/3398#issuecomment-717526781))

##### Regressions

This changes the link retrieval logic, but shouldn't cause malfunctioning.

#### Checklist

### DIFF
--- a/pkg/applicationserver/applicationserver.go
+++ b/pkg/applicationserver/applicationserver.go
@@ -262,7 +262,7 @@ func (as *ApplicationServer) Subscribe(ctx context.Context, protocol string, ids
 
 // Publish processes the given upstream message and then publishes it to the application frontends.
 func (as *ApplicationServer) Publish(ctx context.Context, up *ttnpb.ApplicationUp) error {
-	link, err := as.linkRegistry.Get(ctx, up.ApplicationIdentifiers, []string{
+	link, err := as.getLink(ctx, up.ApplicationIdentifiers, []string{
 		"default_formatters",
 		"skip_payload_crypto",
 	})
@@ -336,7 +336,7 @@ func (as *ApplicationServer) downlinkQueueOp(ctx context.Context, ids ttnpb.EndD
 	if err != nil {
 		return err
 	}
-	link, err := as.linkRegistry.Get(ctx, ids.ApplicationIdentifiers, []string{
+	link, err := as.getLink(ctx, ids.ApplicationIdentifiers, []string{
 		"default_formatters",
 		"skip_payload_crypto",
 	})
@@ -473,7 +473,7 @@ func (as *ApplicationServer) DownlinkQueueList(ctx context.Context, ids ttnpb.En
 	if err != nil {
 		return nil, err
 	}
-	link, err := as.linkRegistry.Get(ctx, ids.ApplicationIdentifiers, []string{
+	link, err := as.getLink(ctx, ids.ApplicationIdentifiers, []string{
 		"default_formatters",
 		"skip_payload_crypto",
 	})
--- a/pkg/applicationserver/applicationserver.go
+++ b/pkg/applicationserver/applicationserver.go
@@ -119,8 +119,8 @@ func New(c *component.Component, conf *Config) (as *ApplicationServer, err error
 			ttnpb.PayloadFormatter_FORMATTER_JAVASCRIPT: javascript.New(),
 			ttnpb.PayloadFormatter_FORMATTER_CAYENNELPP: cayennelpp.New(),
 		}),
-		clusterDistributor: distribution.NewPubSubDistributor(ctx, conf.Distribution.Timeout, conf.Distribution.PubSub),
-		localDistributor:   distribution.NewLocalDistributor(ctx, conf.Distribution.Timeout),
+		clusterDistributor: distribution.NewPubSubDistributor(ctx, c, conf.Distribution.Timeout, conf.Distribution.PubSub),
+		localDistributor:   distribution.NewLocalDistributor(ctx, c, conf.Distribution.Timeout),
 		interopClient:      interopCl,
 		interopID:          conf.Interop.ID,
 		endDeviceFetcher:   conf.EndDeviceFetcher.Fetcher,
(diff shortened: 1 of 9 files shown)


adriansmares:
Changed the context adaptation introduced in https://github.com/TheThingsNetwork/lorawan-stack/pull/3439/files to be done on `publishUp` instead of `publishUp` call sites. This ensures we don't miss call sites (and fixes flakiness).

pkg/applicationserver/applicationserver.go
```diff
@@ -301,6 +301,7 @@ func (as *ApplicationServer) processUp(ctx context.Context, up *ttnpb.Applicatio
 }
 
 func (as *ApplicationServer) publishUp(ctx context.Context, up *ttnpb.ApplicationUp) error {
+	ctx = as.FromRequestContext(ctx)
```
johanstokking: I don't find this a logical change. By convention, the caller provides the context.

adriansmares: The caller provides the context indeed, but having to do `as.FromRequestContext` each time you call `publishUp` is error prone, and before this PR a call site was missed. Nonetheless, I'll change this back to call sites.

johanstokking: Not each time: only when the caller is actually aware of the fact that the context decouples.

pkg/applicationserver/applicationserver.go
```diff
@@ -290,7 +290,7 @@ func (as *ApplicationServer) processUp(ctx context.Context, up *ttnpb.Applicatio
 		return nil
 	}
 
-	if err := as.publishUp(ctx, up); err != nil {
+	if err := as.publishUp(as.FromRequestContext(ctx), up); err != nil {
```
adriansmares: For the record - the context here is from the RPC call that NS makes - it will not outlive the publishing process.

johanstokking: Then why would you decouple this?

In fact, if the NS wants to cancel the publishing process, it should do so via the context.

adriansmares: This PR targets the AS link removal branch - the NS uses unary RPCs to push uplinks here. The context here is automatically cancelled as soon as the RPC callback is finished.

johanstokking: Yes, I know, but we use context decoupling in places where it gets passed on to another goroutine, i.e. when sending it as part of `ContextualApplicationUp` on a channel. As long as that doesn't happen, there's no need to decouple the context, and in fact we shouldn't.

adriansmares: But that's exactly what `publishUp` does indirectly:

`publishUp` calls `Publish` on the local distributor:
https://github.com/TheThingsNetwork/lorawan-stack/blob/e1431e984880da5c89da4f6ae3b109e320ec415e/pkg/applicationserver/applicationserver.go#L303-L304

`localDistributor` is a local distributor:
https://github.com/TheThingsNetwork/lorawan-stack/blob/e1431e984880da5c89da4f6ae3b109e320ec415e/pkg/applicationserver/applicationserver.go#L123

`localDistributor` `Publish` calls `subscriptionSet.Publish` and `subscriptionMap.Publish`:
https://github.com/TheThingsNetwork/lorawan-stack/blob/e1431e984880da5c89da4f6ae3b109e320ec415e/pkg/applicationserver/distribution/local_distributor.go#L42-L53

`subscriptionMap` is just a `map[ttnpb.ApplicationIdentifiers]*subscriptionSet`, and `subscriptionSet.Publish`  will create the `ContextualApplicationUp`:
https://github.com/TheThingsNetwork/lorawan-stack/blob/e1431e984880da5c89da4f6ae3b109e320ec415e/pkg/applicationserver/distribution/subscription_set.go#L98-L111

Fundamentally each time you do `publishUp` the uplink will end up in another goroutine. This is the whole reason for which I proposed that the decoupling is done as part of `publishUp` in the first place.

johanstokking: Exactly for the code above, you should decouple the context as late as possible.

The caller wants to be able to cancel the publish, like canceling the call context that is used here:

https://github.com/TheThingsNetwork/lorawan-stack/blob/e1431e984880da5c89da4f6ae3b109e320ec415e/pkg/applicationserver/distribution/subscription_set.go#L104

If you do what you did, i.e. derive from the AS context in `publishUp()`, then this `ctx` is the component context already. How would you cancel waiting for having the message sent on a channel that way?

Therefore, call `FromRequestContext()` as late as possible, i.e. only when setting `Context` of `ContextualApplicationUp`, and use the request context for everything else.

adriansmares: Ok, I understand what you mean now.

adriansmares:
@johanstokking I've updated the decoupling to occur as late as possible, as discussed above.

pkg/applicationserver/distribution/subscription_set.go
```diff
@@ -97,7 +98,7 @@ func (s *subscriptionSet) Subscribe(ctx context.Context, protocol string, ids *t
 // Publish publishes the upstream traffic to the subscribers.
 func (s *subscriptionSet) Publish(ctx context.Context, up *ttnpb.ApplicationUp) error {
 	ctxUp := &io.ContextualApplicationUp{
-		Context:       ctx,
+		Context:       s.FromRequestContext(ctx),
```
johanstokking: Is this the reason to pass the `Component` all the way here?

I think we get too high coupling of `Component`; why not declare a receiving interface?

adriansmares: I couldn't think of a good name for this particular decoupling - `RequestDecoupler` sounds a bit off. Thoughts ?

johanstokking: Yeah why not?

adriansmares: Done :+1:

johanstokking (approved):

PR #3446, opened 2020-11-06
Remove Network Server related fields from application links

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/3263
References https://github.com/TheThingsNetwork/lorawan-stack/pull/3445

#### Changes

- Remove the `network_server_address` and `api_key` fields from `ApplicationLink`, as they are not used but their contents are currently marked as required (due to length requirements)

#### Testing

Unit tests.

##### Regressions

As this removes API fields, CLI fields will be affected by this. I've added the removed flags manually and hid them in order to keep user scripts from breaking.

#### Checklist


johanstokking (approved):

api/applicationserver.proto
```diff
@@ -31,14 +31,8 @@ package ttn.lorawan.v3;
...
-  // The address of the external Network Server where to link to.
-  // The typical format of the address is "host:port". If the port is omitted,
-  // the normal port inference (with DNS lookup, otherwise defaults) is used.
-  // Leave empty when linking to a cluster Network Server.
-  string network_server_address = 1 [(validate.rules).string.pattern = "^(?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*(?:[A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])(?::[0-9]{1,5})?$|^$"];
-  // The API key to use to link the Application Server to Network Server.
-  // This API key needs to have RIGHT_APPLICATION_LINK.
-  string api_key = 2 [(gogoproto.customname) = "APIKey", (validate.rules).string.min_len = 1];
+  reserved 1; // Deprecated: network_server_address
+  reserved 2; // Deprecated: api_key
```
htdvisser, line 35: I'm pretty sure this violates our compatibility commitment, as existing scripts that rely on the CLI or API will now get validation errors when getting the link (because these fields are no longer allowed to be in the field mask). Similarly, when setting the link I think the server should ignore the fields instead of erroring (possibly returning a warning in the header/trailer).

htdvisser (commented):
I'm also pretty sure this breaks the console, which currently still requests `network_server_address` and `api_key`.

adriansmares:
> I'm also pretty sure this breaks the console, which currently still requests `network_server_address` and `api_key`.

This is true and I can submit a PR about this soon.

> I'm pretty sure this violates our compatibility commitment, as existing scripts that rely on the CLI or API will now get validation errors when getting the link (because these fields are no longer allowed to be in the field mask). Similarly, when setting the link I think the server should ignore the fields instead of erroring (possibly returning a warning in the header/trailer).

I consider (and please feel free to plug in your opinions on this) that the `Get`/`Set` methods do not fall under

`We will not break the API towards gateways and applications within the major version. This includes how gateways communicate (with Gateway Server) and how applications work with data (with Application Server)`

As they do not work with how the AS sends uplinks / receives downlinks from the external world (`data` in this paragraph).

`We will not break the public command-line interface and configuration within the major version. This means that you can safely build scripts and migrate configuration.`

This is handled by the hidden flags that have been added by this PR.

`We will not break the API between components and events within minor versions. So at least the same minor versions of components are compatible with each other.`

As this targets `3.11` I don't think this PR violates the guarantee.

htdvisser:
> `We will not break the public command-line interface and configuration within the major version. This means that you can safely build scripts and migrate configuration.`
> 
> This is handled by the hidden flags that have been added by this PR.

But it does break if someone uses a v3.9 CLI to talk to a v3.11 AS

adriansmares:
> > `We will not break the public command-line interface and configuration within the major version. This means that you can safely build scripts and migrate configuration.`
> > This is handled by the hidden flags that have been added by this PR.
> 
> But it does break if someone uses a v3.9 CLI to talk to a v3.11 AS

Does this mean that we interpret `Command-line interface` as the CLI binary itself will continue to work withing the same major, not the CLI parameters (the 'interface') themselves ? 

I'm asking this because the `public command-line interface` seems to refer to the parameters / semantics of the binary itself, not necessarily using an old binary (otherwise why is `public` mentioned ?).

johanstokking:
> @htdvisser But it does break if someone uses a v3.9 CLI to talk to a v3.11 AS

Indeed, if users have a bootstrap script that creates an app, API key and configures a link, then that will break. What is better here; fail because the mechanism changed (AS is linked implicitly), or ignore these fields and return a warning?

> I'm asking this because the `public command-line interface` seems to refer to the parameters / semantics of the binary itself, not necessarily using an old binary (otherwise why is `public` mentioned ?).

It refers to non-hidden and non-`EXPERIMENTAL` commands. So `application link set --api-key ...` does fall under the compatibility commitment.

htdvisser:
I think we should allow the fields in the field masks, but ignore them, possibly with a warning that the fields are no longer needed. Nothing needs to break, because the desired outcome (working NS-AS communication for the app) is still the same.

api/applicationserver.proto
johanstokking: So should we keep them here?

adriansmares: The last commit leaves these fields as part of the allowed field mask and removes them before calling the store, while generating a warning about the field's deprecated status.

htdvisser (approved):
Owned files LGTM

PR #3448, opened 2020-11-08
Use yarn cache in CI

bafonins:
#### Summary

`yarn` cache was added to the js workflow in [this commit](https://github.com/TheThingsNetwork/lorawan-stack/commit/980c8e1719adf462ffebc042fc67509e3b5ab96a#diff-9992b6f0e329a6dc49913c62f6efe2102f7a64f55d4bdab3d47df86292db5e87), however there is no check for cache hits. this PR adds exactly this.

#### Changes

- Check for cache hit before installing dependencies. 
- Some clean up for gateway and application views to trigger js related jobs.

#### Testing

see CI

##### Regressions

Im not exactly sure whether `resore-keys` entry is needed in this case, but all examples both in the `cache` action repo and in discussions `restore-keys` is present. This might cause issues.

#### Checklist

### DIFF
--- a/.github/workflows/js.yml
+++ b/.github/workflows/js.yml
@@ -28,6 +28,7 @@ jobs:
       id: yarn-cache-dir-path
       run: echo "::set-output name=dir::$(npx yarn cache dir)"
     - name: Initialize Yarn module cache
+      id: yarn-cache
       uses: actions/cache@v2
       with:
         path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
@@ -63,6 +64,7 @@ jobs:
     - name: Build JS SDK
       run: tools/bin/mage jsSDK:clean jsSDK:build
     - name: Install JS dependencies
+      if: steps.yarn-cache.outputs.cache-hit != 'true'
       run: tools/bin/mage js:deps
       timeout-minutes: 5
     - name: Generate JS translations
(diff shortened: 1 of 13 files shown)


kschiffer:
> Im not exactly sure whether `resore-keys` entry is needed in this case, but all examples both in the `cache` action repo and in discussions `restore-keys` is present. This might cause issues.

It's not needed but it should speed up `js.deps` a bit when there was a cache miss.

kschiffer (approved):

.github/workflows/js.yml
```diff
@@ -61,6 +62,7 @@ jobs:
     - name: Build JS SDK
       run: tools/bin/mage jsSDK:clean jsSDK:build
     - name: Install JS dependencies
+      if: steps.yarn-cache.outputs.cache-hit != 'true'
```
htdvisser, line 67: Would it make sense to do the same for `jsSDK:deps` above?

htdvisser: Same in other files

bafonins: From testing various scenarios there is not much improvement when checking for cache hit for js sdk dependencies. It also seems that there are some issues in js/js sdk mage targets that cause weird diff issues in CI in some cases. Ill investigate this further in a separate PR together with improvements for `js:build` target.

I added yarn cache check to `release-snapshot` and `general` workflows as well.

kschiffer:
Btw note [this](https://github.com/TheThingsNetwork/lorawan-stack/runs/1392219479?check_suite_focus=true#step:10:43):

> Command "dump" is deprecated, cockroach dump will be removed in a subsequent release.

Also, note that you'll likely want to skip the `js:build` step altogether if there's a correct cache of the `public` folder, since this target ignores the `public` folder. But I guess this is also what you want to take care of in the next PR.

kschiffer (approved):

htdvisser (approved):
I think this PR needs to be updated after https://github.com/TheThingsNetwork/lorawan-stack/pull/3484

Other than that, LGTM

PR #3464, opened 2020-11-11
Use `start-single-node` command for CockroachDB

### LINKED ISSUE #3465: Use `start-single-node` command when using CockroachDB as database
#### Summary

From CockroachDB 20.2 onwards, the `cockroach start` command is no longer supported, and `start-single-node` must be used instead.

#### Steps to Reproduce

1. Pull the latest `cockroachdb/cockroach` image
2. `docker-compose up -d`

#### What do you see now?

The database never comes online, since the `start` command is no longer supported.

We need to use `start-single-node` instead of `start`

#### Environment

Any version. This has to do with the database init command, not TTS version.

Change `start` to `start-single-node` in `docker-compose.yml`. Update documentation accordingly, and make sure that people with existing deployments that upgrade to the latest CockroachDB image know how to fix it as well.

`docker-compose up -d` should bring up the database.

Yes.

neoaggelos:
#### Summary

Closes #3465 

#### Changes

- Use `start-single-node` instead of `start`
- Run the end-to-end tests GitHub workflow when `docker-compose.yml` file changes, which also helps with making sure this change fixes the end-to-end tests issues we have experienced.

#### Testing

`docker-compose up -d` works, end-to-end tests work.

##### Regressions

I would assume that this would break very old cockroachdb versions that lack the `start-single-node` command. I could not find a changelog for the version or date where `start-single-node` was introduced, but I highly doubt there are any deployments that would use such an old version.

#### Notes for Reviewers

@htdvisser @benolayinka In some capacity, I think we should add a note in the changelog for this, even though the TTS version is irrelevant. Please suggest how that could look like.

#### Checklist

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -14,6 +14,7 @@ on:
       - 'sdk/js/**'
       - 'yarn.lock'
       - 'cypress/**'
+      - 'docker-compose.yml'
 
 jobs:
   end-to-end:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   # If using CockroachDB:
   cockroach:
     image: cockroachdb/cockroach
-    command: start --http-port 26256 --insecure
+    command: start-single-node --http-port 26256 --insecure
     restart: unless-stopped
     volumes:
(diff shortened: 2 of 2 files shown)


htdvisser (approved):

bafonins (approved):
👍

PR #3466, opened 2020-11-11
Add validation to token credentials

KrishnaIyer:
#### Summary

Add validation to token credentials. Resultant of debugging https://github.com/TheThingsIndustries/lorawan-stack/issues/2420#issue-737928329

#### Changes

- Add validation to token credentials
- Update test

#### Testing

UT 

##### Regressions

Not expected since UT covers these.

#### Checklist


pkg/basicstation/cups/messages.go
```diff
@@ -84,6 +90,17 @@ func TokenCredentials(trust *x509.Certificate, authorization string) ([]byte, er
...
 	// TODO: Refactor when client side TLS is supported https://github.com/TheThingsNetwork/lorawan-stack/issues/137
 	out = append(out, emptyClientCert...)
+	token := authorization
+	if strings.HasPrefix(authorization, "Bearer") {
+		t := strings.Split(authorization, " ")
+		if len(t) != 2 {
+			return nil, errInvalidAuthToken.WithAttributes("token", authorization)
+		}
+		token = t[1]
+	}
```
htdvisser: Why do we need to strip the `Bearer` part? What if someone wants to connect their gateway to a different network that does require `Bearer`?

pkg/basicstation/cups/messages.go
```diff
@@ -84,6 +90,17 @@ func TokenCredentials(trust *x509.Certificate, authorization string) ([]byte, er
...
+	if _, _, _, err := auth.SplitToken(token); err != nil || len(token) != apiKeyLength {
+		return nil, errInvalidAuthToken.WithAttributes("token", authorization)
+	}
```
htdvisser: Why do we need to restrict to The Things Stack tokens? What if someone wants to connect their gateway to a different network server?

pkg/basicstation/cups/messages.go
```diff
@@ -84,6 +90,17 @@ func TokenCredentials(trust *x509.Certificate, authorization string) ([]byte, er
...
 	out = append(out, []byte(fmt.Sprintf("%s%s%s", "Authorization: ", authorization, "\r\n"))...)
```
htdvisser: To prevent mistakes, why don't we do `strings.TrimRight(authorization, "\r\n")` here? That way we can be sure that we never add any extra CR or LF characters.

pkg/basicstation/cups/messages.go
KrishnaIyer: We don't. We just strip it for validating the token. `authorization` is the original variable that's used unchanged if the validation is successful.

pkg/basicstation/cups/messages.go
KrishnaIyer: Ok but then we cannot run any validation except the CRLF trimRight.

htdvisser: TBH, I don't think this is the place to validate anything. All we need to do here is make sure that there's only a single CRLF in here.

htdvisser (approved):

Issue #3458, opened 2020-11-10
Transfer full entity rights to another user on soft-delete of user


pgalic96:
#### Summary

When soft-deleting a user, specify another user to transfer rights for entities that do not have another user that has `_ALL` rights for them.

#### Why do we need this?

Prevent having entities that no user has full rights to.

#### What is already there? What do you see now?

User delete logic that simply soft-deletes the user without checking any linked models.

#### What is missing? What do you want to see?

An option in CLI and console to specify another user  which would get full rights to user entities (either all of them or just the ones that do not have another user that has full rights on them).

#### Environment

ttn-lw-stack `v3.10`

#### How do you propose to implement this?

- Expand user delete option to check memberships for all user owned entities
- Transfer rights to the specified user

#### How do you propose to test this?

- Unit tests, manual

#### Can you do this yourself and submit a Pull Request?

Yes or @htdvisser

pgalic96:
@johanstokking @htdvisser 

I have one concern with this approach:

- Do we want to have an option where user could specify the receiver of `RIGHTS_ALL` per entity? I can see this as something that could be useful if we want to avoid forcing users to assign the same user to be an owner of all concerned entities, possibly an entity where they weren't even a collaborator before the delete? On the other hand, this can get quite messy if this needs to be done for a lot of entities.

As this happens only with the entities that the user is the sole owner of, we can also just put this responsibility on the user, by displaying a warning (and they can reassign manually themselves if they want to avoid this situation)

johanstokking:
@pgalic96 let's go with one inheriting user for the rights of the deleting user. We may want to come back on this later, but not for now.

pgalic96:
@htdvisser I had the following idea:

`Delete` method should not have any parameters in the body (as per `delete` http method), therefore including user id for the transfer would not make sense here.

The solution is to create a new RPC `TransferOwnedEntityRights` with user id of the receiver as parameter that would perform this task. It would be performed as part of `delete` CLI command. Then as part of `Delete` RPC we would check if the user to be deleted still has entities that are solely owned, and if this is the case we would output something like `This action will leave following entities orphaned: [entity-ids]. Please re-assign rights to these entities.` This would also mean that the `receiver-id` flag would be optional for the CLI command. Thoughts?

htdvisser:
I think it could be useful to have a dedicated RPC to give some user the same rights as another user. Such an RPC could also be useful for admins to quickly copy all rights from one user to another.

[...]

Issue #2735, opened 2020-06-15
Pub/sub error for non-existent application


benolayinka:
#### Summary

I did some pub/sub testing, and created a couple applications and nats/mqtt pub/sub integrations in them. I deleted the apps, but in the console, I'm still getting weird integration errors for non-existent apps:

```
WARN Integration failed                       application_uid=my-new-application error=nats: no servers available for connection namespace=applicationserver/io/pubsub pub_sub_id=test
```

But the app "my-new-application" doesn't exist and I have no integrations in any other app

<img width="1347" alt="Screenshot 2020-06-15 at 13 00 03" src="https://user-images.githubusercontent.com/6963436/84650287-36127580-af08-11ea-8614-4b1a95de72bc.png">

<img width="1347" alt="Screenshot 2020-06-15 at 13 00 09" src="https://user-images.githubusercontent.com/6963436/84650351-57736180-af08-11ea-98be-14e9be21008f.png">

#### Steps to Reproduce

I'm not 100% sure this is what happened, but I can send any logs or anything else that might be useful as well

1. Create an app
2. Create a pub/sub nats integration
3. Delete the app

#### What do you see now?

```
WARN Integration failed                       application_uid=my-new-application error=nats: no servers available for connection namespace=applicationserver/io/pubsub pub_sub_id=test
```

#### Environment

Mac OS X, running dev stack

#### Can you do this yourself and submit a Pull Request?

No

adriansmares:
I've investigated this and it appears to be happening due to the fact that deleting an application (be it with the console or the CLI) is not 'recursive' - i.e. any associated objects such as Pub/Sub integrations or Webhook integration remain active.

I see two options here:
- CLI and Console recursively walk through entities (Webhooks, Pub/Sub, links) and delete them. 
- The AS does this for them.

The second option changes behavior, but I think it still respects our API guarantees. I personally prefer 2 since we avoid lots of round trips. Thoughts on this @johanstokking ?

bafonins:
> CLI and Console recursively walk through entities (Webhooks, Pub/Sub, links) and delete them.

I dont think that the client should be responsible for making sure that all links to an entity (application in this case) should be deleted. This adds quite some complexity and is tedious to handle

johanstokking:
> I personally prefer 2 since we avoid lots of round trips. Thoughts on this @johanstokking ?

Indeed AS should do this

adriansmares:
[...]

PR #3467, opened 2020-11-11
Purge entities from db

### LINKED ISSUE #1703: RPCs for Purging Entities as Admin
#### Summary

Since our soft-delete prevents entities from being recreated, it would be very useful to have (admin-only) RPCs that completely purge an entity and all associated DB models.

#### Why do we need this?

Automatic GC/purge (#604) isn't happening yet, and users still want to be able to recreate entities.

See #604

Until we have that, manual purge.

Additional RPCs, possibly only for gRPC and CLI to start with.

Yes

pgalic96:
#### Summary

Closes #1703 

#### Changes

- Add RPC for users, organizations, gateways and applications
- Add cli commands for purging entities

#### Testing

Manual, unit testing

##### Regressions

Should not be any as this is a new functionality

#### Notes for Reviewers

There are no rights checks and rights transfer for organizations and users on purge (it will be part of soft-delete checks per https://github.com/TheThingsNetwork/lorawan-stack/issues/3458 ), as this is an admin-only functionality. 
#### Checklist


api/application_services.proto
```diff
@@ -79,6 +79,14 @@ service ApplicationRegistry {
       delete: "/applications/{application_id}"
     };
   };
+
+  // Purge the application. This will release the application ID for reuse.
+  // All end devices must be deleted from the application before it can be deleted.
```
htdvisser, line 84: Maybe good to note that the application owner is responsible for clearing data from any (external) integrations, such as the storage integration and other platforms that may store and expose (historical) data by application ID.

api/gateway_services.proto
```diff
@@ -83,6 +83,13 @@ service GatewayRegistry {
       delete: "/gateways/{gateway_id}"
     };
   };
+
+  // Purge the gateway. This will release all field values for reuse.
```
htdvisser: Let's be explicit here instead of saying "all field values".

Again, maybe good to note that the gateway owner is responsible for clearing data from any (external) integrations that may store and expose (historical) data by gateway ID.

api/user_services.proto
```diff
@@ -80,6 +80,13 @@ service UserRegistry {
       delete: "/users/{user_id}"
     };
   };
+
+  //Purge the user. This will release the user ID for reuse.
```
htdvisser: Again, maybe good to note that the user is responsible for clearing data from any (external) integrations that may store and expose (historical) data by organization or user ID (those are in the same namespace).

api/organization_services.proto
```diff
@@ -71,6 +71,13 @@ service OrganizationRegistry {
       delete: "/organizations/{organization_id}"
     };
   };
+
+  // Purge the organization. This will release the organization ID for reuse.
```
htdvisser, line 75: Again, maybe good to note that the organization owner is responsible for clearing data from any (external) integrations that may store and expose (historical) data by organization or user ID (those are in the same namespace).

cmd/ttn-lw-cli/commands/applications.go
```diff
@@ -251,6 +251,28 @@ var (
...
+	applicationsPurgeCommand = &cobra.Command{
+		Use:     "purge [application-id]",
+		Aliases: []string{"permanent-delete", "hard-delete"},
+		Short:   "Purge an application",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			appID := getApplicationID(cmd.Flags(), args)
+			if appID == nil {
+				return errNoApplicationID
+			}
+
```
htdvisser, line 264: I think it may be smart to warn users for what they're about to do, and ask for extra confirmation (either interactively or with a flag).

Same for the other purge commands, and later also in the Console.

pkg/identityserver/organization_registry.go
```diff
@@ -243,6 +250,28 @@ func (is *IdentityServer) deleteOrganization(ctx context.Context, ids *ttnpb.Org
 	return ttnpb.Empty, nil
 }
 
+func (is *IdentityServer) purgeOrganization(ctx context.Context, ids *ttnpb.OrganizationIdentifiers) (*types.Empty, error) {
+	if !is.IsAdmin(ctx) {
+		return nil, errAdminsPurgeOrganizations
+	}
+	err := is.withDatabase(ctx, func(db *gorm.DB) error {
```
htdvisser, line 257: Organizations also have API keys.

pkg/identityserver/user_registry.go
```diff
@@ -619,6 +626,32 @@ func (is *IdentityServer) deleteUser(ctx context.Context, ids *ttnpb.UserIdentif
 	return ttnpb.Empty, nil
 }
 
+func (is *IdentityServer) purgeUser(ctx context.Context, ids *ttnpb.UserIdentifiers) (*types.Empty, error) {
+	if !is.IsAdmin(ctx) {
+		return nil, errAdminsPurgeUsers
+	}
+	err := is.withDatabase(ctx, func(db *gorm.DB) error {
```
htdvisser, line 633: Users also have API keys.

pkg/identityserver/store/oauth_store.go
```diff
@@ -140,6 +140,18 @@ func (s *oauthStore) DeleteAuthorization(ctx context.Context, userIDs *ttnpb.Use
 	return err
 }
 
+func (s *oauthStore) DeleteUserAuthorizations(ctx context.Context, userIDs *ttnpb.UserIdentifiers) error {
```
htdvisser, line 143: Better to make this about deleting everything. Not only `ClientAuthorization`s, but also all `AuthorizationCode`s and `AccessToken`s.

pkg/identityserver/store/contact_info_store.go
```diff
@@ -233,3 +233,18 @@ func (s *contactInfoStore) Validate(ctx context.Context, validation *ttnpb.Conta
...
+	entity, err := s.findDeletedEntity(ctx, entityID, "id")
+	if err != nil {
+		return err
+	}
+	var contactInfo []ContactInfo
+	query := s.query(ctx, ContactInfo{}).Where(&ContactInfo{EntityID: entity.PrimaryKey(), EntityType: entityTypeForID(entityID)})
+	err = query.Find(&contactInfo).Error
+	if err != nil {
+		return err
+	}
```
htdvisser: I don't think this is necessary. We don't need to return any errors if no contact info found.

pkg/identityserver/store/api_key_store.go
```diff
@@ -132,3 +132,18 @@ func (s *apiKeyStore) UpdateAPIKey(ctx context.Context, entityID ttnpb.Identifie
...
+	entity, err := s.findDeletedEntity(ctx, entityID, "id")
+	if err != nil {
+		return err
+	}
+	var keyModels []APIKey
+	query := s.query(ctx, APIKey{}).Where(&APIKey{EntityID: entity.PrimaryKey(), EntityType: entityTypeForID(entityID)})
+	err = query.Find(&keyModels).Error
+	if err != nil {
+		return err
+	}
```
htdvisser: I don't think this is necessary. We don't need to return any errors if no API keys found.

pkg/identityserver/store/organization_store.go
```diff
@@ -158,3 +158,35 @@ func (s *organizationStore) DeleteOrganization(ctx context.Context, id *ttnpb.Or
...
+		if err := s.replaceAttributes(ctx, "organization", orgModel.ID, orgModel.Attributes, nil); err != nil {
+			return err
+		}
+	}
+
+	err = s.purgeEntity(ctx, id)
+	if err != nil {
+		return err
+	}
+	// Purge account after purging user because it is necessary for user query
```
htdvisser: I suppose you meant organization here

pkg/identityserver/store/membership_store.go
```diff
@@ -267,3 +267,37 @@ func (s *membershipStore) SetMember(ctx context.Context, id *ttnpb.OrganizationO
...
+	}
+	var memberships []Membership
+	query := s.query(ctx, Membership{}).Where(&Membership{
+		EntityID:   entity.PrimaryKey(),
+		EntityType: entityTypeForID(entityID),
+	})
+	err = query.Find(&memberships).Error
+	if err != nil {
+		return err
+	}
```
htdvisser: It's not necessary to first find and then delete.

cmd/ttn-lw-cli/commands/applications.go
```diff
@@ -41,6 +41,9 @@ func applicationIDFlags() *pflag.FlagSet {
 }
 
 var errNoApplicationID = errors.DefineInvalidArgument("no_application_id", "no application ID set")
+var errNoConfirmation = errors.DefineInvalidArgument("no_confirmation", "action not confirmed")
+
+var applicationPurgeWarning = "This action will permanently delete the application and all related data (API keys, rights, attributes etc.)"
```
johanstokking: Idiomatic would be to group them in a `var ( ... )` block

johanstokking: BTW, are we using vars for messages like these elsewhere in the CLI?

If so, they would be `const`s, but I don't think we do this

cmd/ttn-lw-cli/commands/gateways.go
```diff
@@ -48,6 +48,8 @@ func gatewayIDFlags() *pflag.FlagSet {
 
 var errNoGatewayID = errors.DefineInvalidArgument("no_gateway_id", "no gateway ID set")
 
+var gatewayPurgeWarning = "This action will permanently delete the gateway and all related data (API keys, antennas, attributes etc.)"
```
johanstokking: Idem and onwards

cmd/ttn-lw-cli/commands/root.go
```diff
@@ -207,6 +209,35 @@ func optionalAuth() error {
 	return nil
 }
 
+func confirmChoice(warning string) bool {
```
johanstokking: We probably need to have a non-interactive option here too, i.e. with a `--force` flag or something

johanstokking (commented):
Only reviewed code owned files

cmd/ttn-lw-cli/commands/applications.go
pgalic96: I haven't seen it used like this anywhere else.

pgalic96: I grouped it inside one file but let me know if there is a better way of doing this.

pkg/identityserver/gateway_registry.go
```diff
@@ -398,6 +405,35 @@ func (is *IdentityServer) deleteGateway(ctx context.Context, ids *ttnpb.GatewayI
...
+		err = store.GetContactInfoStore(db).DeleteEntityContactInfo(ctx, ids)
+		if err != nil {
+			return err
+		}
+		return store.GetGatewayStore(db).PurgeGateway(ctx, ids)
+	})
+	if err != nil {
+		return nil, err
+	}
+	events.Publish(evtDeleteGateway.NewWithIdentifiersAndData(ctx, ids, nil))
```
bafonins:
```suggestion
	events.Publish(evtPurgeGateway.NewWithIdentifiersAndData(ctx, ids, nil))
```

otherwise `gateway.delete` is dispatched

bafonins (commented):
1. After purging an application, requests to fetch app link and collaborators returns 200, while fetching api-keys results in 403.
2. I can still add integrations on purged application (e.g. webhook integration)

pkg/identityserver/gateway_registry.go
pgalic96: I believe the second issue is due to the integrations being stored in AS and its part of a separate issue (https://github.com/TheThingsNetwork/lorawan-stack/issues/2735). Should we block this PR on this issue?

pkg/identityserver/organization_registry.go
```diff
@@ -243,6 +250,33 @@ func (is *IdentityServer) deleteOrganization(ctx context.Context, ids *ttnpb.Org
...
+		err := store.GetContactInfoStore(db).DeleteEntityContactInfo(ctx, ids)
+		if err != nil {
+			return err
+		}
+		// delete related API keys before purging the organization
+		err = store.GetAPIKeyStore(db).DeleteEntityAPIKeys(ctx, ids)
+		if err != nil {
+			return err
+		}
+		err = store.GetMembershipStore(db).DeleteAccountMembers(ctx, ids.GetOrganizationOrUserIdentifiers())
```
kschiffer, line 267:
```suggestion
        // Delete related memberships before purging the organization
        err = store.GetMembershipStore(db).DeleteAccountMembers(ctx, ids.GetOrganizationOrUserIdentifiers())
```

Nit: keep your commenting consistent. Also, from our [DEVELOPMENT.md](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.10/DEVELOPMENT.md#comments) guidelines:

> All comments should be English sentences, starting with a capital letter and ending with a period.

pkg/identityserver/store/application_store_test.go
```diff
@@ -128,5 +128,16 @@ func TestApplicationStore(t *testing.T) {
...
+		err = store.PurgeApplication(ctx, &ttnpb.ApplicationIdentifiers{ApplicationID: "foo"})
+
+		a.So(err, should.BeNil)
+
+		// Check that application ids are released after purge
+		_, err = store.CreateApplication(ctx, &ttnpb.Application{
+			ApplicationIdentifiers: ttnpb.ApplicationIdentifiers{ApplicationID: "foo"},
+		})
+
+		a.So(err, should.BeNil)
```
kschiffer, line 152: You might also want to check whether associated API Keys and/or memberships were deleted.

pkg/identityserver/store/contact_info_store_test.go
```diff
@@ -132,5 +140,6 @@ func TestContactInfoValidation(t *testing.T) {
 		if a.So(info, should.HaveLength, 1) {
 			a.So(info[0].ValidatedAt, should.NotBeNil)
 		}
+
```
kschiffer:
```suggestion
```

pkg/identityserver/user_registry.go
```diff
@@ -619,6 +626,37 @@ func (is *IdentityServer) deleteUser(ctx context.Context, ids *ttnpb.UserIdentif
...
+		}
+		err = store.GetMembershipStore(db).DeleteAccountMembers(ctx, ids.GetOrganizationOrUserIdentifiers())
+		if err != nil {
+			return err
+		}
+		err = store.GetOAuthStore(db).DeleteUserAuthorizations(ctx, ids)
+		if err != nil {
+			return err
+		}
+		return store.GetUserStore(db).PurgeUser(ctx, ids)
```
kschiffer, line 655: What about user sessions?

kschiffer (changes requested):

cmd/ttn-lw-cli/commands/root.go
```diff
@@ -207,6 +209,49 @@ func optionalAuth() error {
...
+var (
+	applicationPurgeWarning  = "This action will permanently delete the application and all related data (API keys, rights, attributes etc.)"
+	gatewayPurgeWarning      = "This action will permanently delete the gateway and all related data (API keys, antennas, attributes etc.)"
+	organizationPurgeWarning = `This action will permanently delete the organization and all 
+	related data (API keys, rights, attributes etc.). 
+	It might also cause entities to be orphaned if this organization is the only one that has full rights on the entity.`
+	userPurgeWarning = `This action will permanently delete the user and all related data
+ 	(API keys, entity rights, attributes etc.).
+		It might also cause entities to be orphaned if this user is the only one that has full rights on the entity.`
+)
```
htdvisser: - I don't think we ever put messages together like this, but if you think this helps keeping things consistent, then okay. But maybe better to use a sepearate file for things like it?
- The line breaks are not quite right. There are additional whitespace characters that shouldn't be there. In addition, I think we usually split long lines over multiple `Warn` calls.

cmd/ttn-lw-cli/commands/root.go
```diff
@@ -207,6 +209,49 @@ func optionalAuth() error {
...
+
+func confirmChoice(warning string, force bool) bool {
+	if force {
+		return true
+	}
+	if warning != "" {
+		logger.Info(warning)
```
htdvisser: If it's a warning, use `Warn`

htdvisser (commented):
@bafonins I suspect this is because of caching. The Identity Server caches memberships and rights, and every cluster without Identity Server also caches memberships and rights. 

It's possible to clear the first one by implementing `DeleteEntityMembers` and `DeleteAccountMembers` in `pkg/identityserver/store/membership_cache.go`. These funcs woulf first call the underlying store, and then clear the cache. Clearing the cache is a bit complicated, as we'd have to `SCAN` the cache for matching keys, which is not ideal.

We currently don't have a solution for clearing the cache of other clusters. Related discussion in https://github.com/TheThingsNetwork/lorawan-stack/issues/2735#issuecomment-729259663

pkg/identityserver/store/application_store_test.go
pgalic96: The issue with testing this is that it's not part of `Purge` application store method. It is a part of the `Purge` method in the application registry. I think what is ensured with the separate store functions is that with the same input, the behavior is correct. In the store it makes sense to ensure that the fields like `Picture`, `Attribute`, and `Antenna` are cleared properly because that is not covered with tests. Not sure if having the functionality to test the full flow should reside in the unit tests.

pgalic96:
Extended the logic to cover user sessions and contact info validations. Extended the tests with antennas and attributes. Would cache clearing also be part of this PR, if its not getting too big already?

cmd/ttn-lw-cli/commands/root.go
```diff
@@ -207,6 +209,40 @@ func optionalAuth() error {
 	return nil
 }
 
+func confirmChoice(warnings []string, force bool) bool {
+	if force {
+		return true
+	}
+	if len(warnings) > 0 {
+		for _, warning := range warnings {
+			logger.Info(warning)
```
htdvisser: If it's a warning, use `Warn`.

See my previous review: https://github.com/TheThingsNetwork/lorawan-stack/pull/3467#discussion_r526695724

htdvisser (approved):
Small thing left, rest looks good to me.

kschiffer (approved):
LGTM

bafonins (approved):

bafonins:
You might want to add the new api to changelog

PR #3470, opened 2020-11-12
Limit event views in the Console to 2000 events per stream

### LINKED ISSUE #2887: Data views slow down and freeze after receiving many events

kschiffer:
#### Summary
This PR is a forgotten followup from #3330. It limits the number of events per stream to 2000 in the Console, to prevent memory and performance issues in the event views.

image

Closes #2887 

#### Changes
- Set a hard limit for events to 2000, after that old limits will be truncated
- Add some logic and styles to communicate truncated events

#### Testing

Manual testing on the staging environment

##### Regressions

This touches store logic for events, so it could introduce arbitrary issues with event ingestion and display in the Console.

#### Checklist

### DIFF
--- a/pkg/webui/console/components/events/index.js
+++ b/pkg/webui/console/components/events/index.js
@@ -15,9 +15,11 @@
 import React, { useState, useCallback } from 'react'
 import classnames from 'classnames'
 
+import EVENT_STORE_LIMIT from '@console/constants/event-store-limit'
 import hamburgerMenuClose from '@assets/misc/hamburger-menu-close.svg'
 
 import Button from '@ttn-lw/components/button'
+import Icon from '@ttn-lw/components/icon'
 
 import Message from '@ttn-lw/lib/components/message'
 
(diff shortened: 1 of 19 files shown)


pkg/webui/console/components/events/index.js
```diff
@@ -32,81 +33,91 @@ import { getEventId } from './utils'
...
-    [focus],
-  )
+const Events = React.memo(
+  ({ events, scoped, paused, onClear, onPauseToggle, entityId, truncated }) => {
+    const [focus, setFocus] = useState({ eventId: undefined, visible: false })
+    const onPause = useCallback(() => onPauseToggle(paused), [onPauseToggle, paused])
+    const handleRowClick = useCallback(
+      eventId => {
+        if (eventId !== focus.eventId) {
+          setFocus({ eventId, visible: eventId })
```
bafonins, line 44:
```suggestion
          setFocus({ eventId, visible: true })
```

bafonins (approved):

PR #3473, opened 2020-11-12
Skip payload crypto if AppSKey is unknown to AS

johanstokking:
#### Summary

Skip payload crypto if AppSKey is unknown to the Application Server

#### Changes

This enhances the check for skipping payload crypto in Application Server. Formerly, only the end device and link configuration was checked. Now, if the Application Server has the AppSKey (or it can unwrap it), payload crypto is not skipped.

The reason for this is that when using a Join Server that doesn't wrap the AppSKey (i.e. the cluster-local Join Server, or Semtech Join Server), you don't want the Application Server to skip payload crypto. Another reason is that if the Join Server does wrap the AppSKey but with a KEK that is only known to the Application Server, the Application Server will have to unwrap the AppSKey. For example, when using the Global Join Server (that uses the cluster KEK) and having `skip_payload_crypto` set, Application Server would pass on the AppSKey with a KEK label for the routing cluster, which is useless for the application.

#### Testing

Unit testing

##### Regressions

None

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Changed
 
 - Limitation of displayed and stored events in the Console to 2000.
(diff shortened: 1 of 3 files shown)


adriansmares (approved):
LGTM.

Issue #2533, opened 2020-05-13
Remove usage of default HTTP Client or Transport


htdvisser:
#### Summary

We should stop using the default HTTP client, and instead use one we get from the `component.(*Component)` so that we can pass it TLS client config (such as trusted CAs).

#### What is already there? What do you see now?

Some places still use `http.DefaultClient`

#### What is missing? What do you want to see?

We should use an http Client that we can configure

#### How do you propose to implement this?

- Add a `func (c *Component) HTTPTransport() http.RoundTripper`
- Add a `func (c *Component) HTTPClient() *http.Client`
- Use those in places where we make HTTP calls
	- `pkg/applicationserver/io/packages/loradms/v1`
	- `pkg/applicationserver`
	- `pkg/fetch`
	- `pkg/identityserver/picture`
	- `pkg/interop`
	- `pkg/web/oauthclient`
	- maybe other places

PR #3475, opened 2020-11-13
LoRaCloud DAS parsing fixes

### LINKED ISSUE #3356: LoRaCloud DAS Unauthorized errors should be handled gracefully
#### Summary

Currently, if an invalid API key is used with LoRaCloud DAS, the remote server will respond with a body containing only `Unauthorized`. As we always attempt to unmarshal the body as JSON, we fail and generate an unnecessary warning.

#### Steps to Reproduce

1. Use an invalid API key (a random one would do) with LoRaCloud DAS
2. Send an uplink

#### What do you see now?

The application package attempts to send the uplink to LoRaCloud DAS, and DAS responds with `Unauthorized`. This generates the following warning:
```
2020-10-13T13:04:09.948+02:00	WARN Failed to handle message error=invalid character 'U' looking for beginning of value namespace=applicationserver/io/packages
2020-10-13T13:04:10.088+02:00	WARN Failed to process message code=502 error=error:pkg/applicationserver/io/web:request (request failed with status `502`) namespace=applicationserver
```

Warnings which are less cryptic.

#### Environment

`v3.9.4`

We should unmarshal only on 200 error codes, and check the `Content-Type` header perhaps.
I think the log level should also be dropped, as users can add basically any host as a LoRaCloud DAS endpoint, so we don't really gain much insight from these warnings.

Repeat the steps above and check that the warnings are gone / at least useful.

Yes, but @neoaggelos feel free to reassign.

### LINKED ISSUE #3455: Add HTTP client timeout to the LoRaCloud DAS client
#### Summary

The HTTP client used by LoRaCloud DAS needs HTTP read/write timeouts.

#### Why do we need this?

As the client is shared between applications and tenants, the underlying connection pool can be consumed by invalid endpoints.

The package uses `http.DefaultClient`.

The client should use an `http.Client` with a timeout (similar to the webhooks).

#### Environment

`v3.10.0`

Update the package to use a custom client.

Unit tests.

Yes.

adriansmares:
#### Summary

Closes #3455 
Closes #3356

#### Changes

- Removed LoRaCloud DAS unused object definitions. As the service public API evolves contiguously, their API definitions do not, and our request parsers lag behind due to their use of sum-types in JS. As such, we now parse only the part of the responses that specifies the downlink to be scheduled on the device, and nothing more. A nice side effect of this is that we pass the response from the service directly, thus always being up to date with their newer fields.
- Make HTTP requests `context.Context` aware.
- Introduce a custom `http.Client` with a 30 second timeout until #2533 is implemented, as we've seen cases in which the client hangs in CH.
- Responses with non-2xx status code are no longer parsed and simply generate an error

#### Testing

- As suggested by @Oliv4945, I've used the `gettime` command on the Murata Modem. This used to trigger an parsing error before the fixes.
- Tested requests with an invalid API key and checked stack logs.

#### Checklist


johanstokking (approved):

Oliv4945:
Thanks !

PR #3486, opened 2020-11-16
Remove usage of default HTTP client and transport

### LINKED ISSUE #2533: Remove usage of default HTTP Client or Transport

neoaggelos:
#### Summary

Closes #2533

#### Changes

- Adds `component.HTTPClient(context.Context) (*http.Client, error)` and `component.HTTPTransport(context.Context) (http.RoundTripper, error)`
- Extend `fetcher.FromHTTP` to accept a custom transport.
- Add new configuration `Transport http.RoundTripper` for all fetchers.
- (as) Extend `io.Server` with `HTTPClient()` method and use it for webhooks and the LoRaCloud package.
- (oauthclient) Use the component HTTPClient.

#### Testing

Unit tests, test locally.

##### Regressions

Possible, since different parts of the codebase dealt with HTTP clients and transports in all sorts of ways.

#### Notes for Reviewers

Please make sure your code-owned files look OK.

@johanstokking `pkg/interop` client intentionally not changed, since it would require quite some refactoring which I would prefer to do in a separate PR.
@kschiffer specifically, what is a scenario to verify that `oauthclient` works as expected?

#### Checklist

### DIFF
--- a/cmd/ttn-lw-stack/commands/start.go
+++ b/cmd/ttn-lw-stack/commands/start.go
@@ -190,6 +190,17 @@ var startCommand = &cobra.Command{
 
 		redisConsumerID := redis.Key(host, strconv.Itoa(os.Getpid()))
 
+		transport, err := c.HTTPTransport(ctx)
+		if err != nil {
+			return err
+		}
+		if conf := config.ServiceBase.FrequencyPlans; conf.Transport == nil {
+			conf.Transport = transport
+		}
+		if conf := config.ServiceBase.Interop.SenderClientCA; conf.Transport == nil {
+			conf.Transport = transport
+		}
+
 		if start.IdentityServer {
 			logger.Info("Setting up Identity Server")
 			if config.IS.OAuth.UI.TemplateData.SentryDSN == "" {
--- a/pkg/applicationserver/config.go
+++ b/pkg/applicationserver/config.go
@@ -150,10 +150,13 @@ func (c WebhooksConfig) NewWebhooks(ctx context.Context, server io.Server) (web.
 	case "":
 		return nil, nil
 	case "direct":
+		client, err := server.HTTPClient(ctx)
+		if err != nil {
+			return nil, err
+		}
+		client.Timeout = c.Timeout
 		target = &web.HTTPClientSink{
-			Client: &http.Client{
-				Timeout: c.Timeout,
-			},
+			Client: client,
 		}
 	default:
 		return nil, errWebhooksTarget.WithAttributes("target", c.Target)
--- a/pkg/applicationserver/io/packages/loradms/v1/package.go
+++ b/pkg/applicationserver/io/packages/loradms/v1/package.go
@@ -17,7 +17,6 @@ package loraclouddevicemanagementv1
 import (
 	"context"
 	"fmt"
-	"net/http"
 	"time"
 
 	"github.com/gogo/protobuf/types"
@@ -39,14 +38,9 @@ import (
 const PackageName = "lora-cloud-device-management-v1"
 
 // DeviceManagementPackage is the LoRa Cloud Device Management application package.
-//
-// TODO: Once https://github.com/TheThingsNetwork/lorawan-stack/issues/2533 is implemented,
-// change io.Server usage to an interface that extends io.Server, but with the HTTPClient()
-// method added, and remove the custom client.
 type DeviceManagementPackage struct {
 	server   io.Server
 	registry packages.Registry
-	client   *http.Client
 }
 
 // RegisterServices implements packages.ApplicationPackageHandler.
@@ -121,7 +115,12 @@ func (p *DeviceManagementPackage) sendUplink(ctx context.Context, up *ttnpb.Appl
 	logger := log.FromContext(ctx)
 	eui := objects.EUI(*up.DevEUI)
 
-	client, err := api.New(p.client, api.WithToken(data.token), api.WithBaseURL(data.serverURL))
+	httpClient, err := p.server.HTTPClient(ctx)
+	if err != nil {
+		logger.WithError(err).Debug("Failed to create HTTP client")
+		return err
+	}
+	client, err := api.New(httpClient, api.WithToken(data.token), api.WithBaseURL(data.serverURL))
 	if err != nil {
 		logger.WithError(err).Debug("Failed to create API client")
 		return err
@@ -312,9 +311,6 @@ func New(server io.Server, registry packages.Registry) packages.ApplicationPacka
 	return &DeviceManagementPackage{
 		server:   server,
 		registry: registry,
-		client: &http.Client{
-			Timeout: 30 * time.Second,
-		},
 	}
 }
 
--- a/pkg/component/http.go
+++ b/pkg/component/http.go
@@ -1,0 +1,47 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package component
+
+import (
+	"context"
+	"net/http"
+	"time"
+)
+
+// defaultHTTPClientTimeout is the default timeout for the HTTP client.
+const defaultHTTPClientTimeout = 10 * time.Second
+
+// HTTPClient returns a new *http.Client with a default timeout and a configured transport.
+func (c *Component) HTTPClient(ctx context.Context) (*http.Client, error) {
+	tr, err := c.HTTPTransport(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return &http.Client{
+		Timeout:   defaultHTTPClientTimeout,
+		Transport: tr,
+	}, nil
+}
+
+// HTTPTransport returns a new http.RoundTripper with TLS client configuration.
+func (c *Component) HTTPTransport(ctx context.Context) (http.RoundTripper, error) {
+	tlsConfig, err := c.GetTLSClientConfig(ctx)
+	if err != nil {
+		return nil, err
+	}
+	tr := http.DefaultTransport.(*http.Transport).Clone()
+	tr.TLSClientConfig = tlsConfig
+	return tr, nil
+}
--- a/pkg/fetch/http.go
+++ b/pkg/fetch/http.go
@@ -64,10 +64,16 @@ func (f httpFetcher) File(pathElements ...string) ([]byte, error) {
 }
 
 // FromHTTP returns an object to fetch files from a webserver.
-func FromHTTP(rootURL string, cache bool) (Interface, error) {
-	transport := http.DefaultTransport
+func FromHTTP(transport http.RoundTripper, rootURL string, cache bool) (Interface, error) {
+	if transport == nil {
+		transport = http.DefaultTransport
+	}
 	if cache {
-		transport = httpcache.NewMemoryCacheTransport()
+		transport = &httpcache.Transport{
+			Transport:           transport,
+			Cache:               httpcache.NewMemoryCache(),
+			MarkCachedResponses: true,
+		}
 	}
 	var root *url.URL
 	if rootURL != "" {
--- a/pkg/networkserver/networkserver.go
+++ b/pkg/networkserver/networkserver.go
@@ -191,6 +191,13 @@ func New(c *component.Component, conf *Config, opts ...Option) (*NetworkServer, 
 			return c.GetTLSClientConfig(ctx)
 		}
 		interopConf.BlobConfig = c.GetBaseConfig(ctx).Blob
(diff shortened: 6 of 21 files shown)


cmd/ttn-lw-stack/commands/start.go
```diff
@@ -170,11 +170,28 @@ var startCommand = &cobra.Command{
 
 		redisConsumerID := redis.Key(host, strconv.Itoa(os.Getpid()))
 
+		transport, err := c.HTTPTransport(ctx)
+		if err != nil {
+			logger.WithError(err).Error("Failed to setup HTTP transport")
```
johanstokking: Should we continue execution?

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -269,6 +289,12 @@ var startCommand = &cobra.Command{
 					Redis: redis.New(config.Redis.WithNamespace("as", "io", "webhooks")),
 				}
 			}
+			if cfg := config.AS.Webhooks.Templates; cfg.Transport == nil {
+				cfg.Transport = transport
+			}
+			if cfg := config.AS.Interop.InteropClient; cfg.Transport == nil {
+				cfg.Transport = transport
+			}
```
johanstokking: I think it's quite error prone to set transport settings here.

Can we were this in the component (context)?

pkg/applicationserver/config.go
```diff
@@ -129,10 +129,12 @@ func (c WebhooksConfig) NewWebhooks(ctx context.Context, server io.Server) (web.
 	case "":
 		return nil, nil
 	case "direct":
+		client, err := server.HTTPClient(ctx, component.WithTimeout(c.Timeout))
+		if err != nil {
+			log.FromContext(ctx).WithError(err).Error("Failed to configure HTTP client")
```
johanstokking: How about returning the error?

pkg/applicationserver/io/packages/loradms/v1/package.go
```diff
@@ -219,13 +224,15 @@ func (p *DeviceManagementPackage) Package() *ttnpb.ApplicationPackage {
 }
 
 // New instantiates the LoRa Cloud Device Management package.
-func New(server io.Server, registry packages.Registry) packages.ApplicationPackageHandler {
+func New(ctx context.Context, server io.Server, registry packages.Registry) packages.ApplicationPackageHandler {
+	client, err := server.HTTPClient(ctx, component.WithTimeout(defaultClientTimeout))
+	if err != nil {
+		log.FromContext(ctx).WithError(err).Error("Failed to create HTTP client")
```
johanstokking: Idem

johanstokking (dismissed):
Stopped review halfway. Please rework and mark as ready when done

pkg/applicationserver/io/packages/loradms/v1/package.go
johanstokking: What do you mean with "tested locally" if you didn't see this panicking?

pkg/component/http.go
```diff
@@ -0,0 +1,71 @@
...
+const (
+	// defaultTimeout is the default timeout for the HTTP client.
+	defaultTimeout = 10 * time.Second
+)
+
+// HTTPClientOpt applies an option to the HTTP client.
+type HTTPClientOpt func(*http.Client)
+
+// WithTimeout sets the HTTP client timeout.
+func WithTimeout(t time.Duration) HTTPClientOpt {
```
adriansmares: Not sure if I like this.

First of all, this becomes `component.WithTimeout`, which conflicts with actual component options. 
Second, I think it's a bit of an anti-pattern to override this everywhere - We added `HTTPClient` in order to have uniform HTTP client settings everywhere. I'd prefer that this option is used only in tests at most.

pkg/component/http.go
```diff
@@ -0,0 +1,71 @@
...
+		c.Transport = t
+	})
+}
+
+// HTTPClient returns a new *http.Client with a timeout and a configured
+// transport. Uses the http.RoundTripper returned from HTTPTransport(). In case
+// of an error, an *http.Client with http.DefaultTransport is returned, along
+// with the error. It is up to the caller to decide how to handle the error.
+func (c *Component) HTTPClient(ctx context.Context, opts ...HTTPClientOpt) (*http.Client, error) {
+	transport, err := c.HTTPTransport(ctx)
```
adriansmares: Error handling for `err` here is missing.

pkg/component/http.go
```diff
@@ -0,0 +1,71 @@
...
+	return client, err
+}
+
+// HTTPTransport returns a new http.RoundTripper with TLS client configuration.
+// If the call fails, http.DefaultTransport is returned along with the error,
+// and it is up to the caller to decide how to handle the error.
+func (c *Component) HTTPTransport(ctx context.Context) (http.RoundTripper, error) {
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	tlsConfig, err := c.GetTLSClientConfig(ctx)
+	if err == nil {
```
adriansmares: Return the error, don't hide it.

adriansmares (changes requested):

kschiffer (commented):
> @kschiffer specifically, what is a scenario to verify that oauthclient works as expected?

Testing whether client authorization e.g. of the Console still works. I tested this both with and without TLS and it worked. Codeowned files also LGTM.

pkg/component/http.go
htdvisser: We should probably also check how this interacts with the context deadline that we use in individual requests.

pkg/component/http.go
htdvisser: Technically not missing, because it is returned in the end.

But I would indeed prever doing error handling here, and returning `client, nil` in the end of the func.

pkg/component/http.go
htdvisser: Technically not hidden, because it is returned in the end.

But I would indeed prever doing error handling here, and returning `transport, nil` in the end of the func.

pkg/fetch/http.go
```diff
@@ -64,10 +64,16 @@ func (f httpFetcher) File(pathElements ...string) ([]byte, error) {
...
-		transport = httpcache.NewMemoryCacheTransport()
+		transport = &httpcache.Transport{
+			Transport:           transport,
+			Cache:               httpcache.NewMemoryCache(),
+			MarkCachedResponses: true,
+		}
+	}
+	if transport == nil {
+		transport = http.DefaultTransport
 	}
```
htdvisser: Shouldn't this be above the `if cache {`?

neoaggelos: It shouldn't make a difference, since `httpcache.Transport` will use `http.DefaultTransport` if transport is `nil`. I can see how that may be confusing though, so I moved it before the `if cache {` block.

cmd/ttn-lw-stack/commands/start.go
adriansmares: This should indeed halt the execution, please `return err`.

pkg/applicationserver/io/packages/loradms/v1/package.go
```diff
@@ -35,17 +34,14 @@ import (
 	"google.golang.org/grpc"
 )
 
-const packageName = "lora-cloud-device-management-v1"
+const (
```
adriansmares: Remove the diff (the `()`).

pkg/component/http.go
```diff
@@ -0,0 +1,50 @@
...
+
+package component
+
+import (
+	"context"
+	"net/http"
+	"time"
+)
+
+const (
```
adriansmares: Nit: You may remove the parenthesis.

adriansmares (approved):
I like this a lot more now.

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -170,6 +170,20 @@ var startCommand = &cobra.Command{
 
 		redisConsumerID := redis.Key(host, strconv.Itoa(os.Getpid()))
 
+		transport, err := c.HTTPTransport(ctx)
+		if err != nil {
+			return err
+		}
+		if cfg := config.ServiceBase.DeviceRepository; cfg.Transport == nil {
```
rvolosatovs:
```suggestion
		if conf := config.ServiceBase.DeviceRepository; conf.Transport == nil {
```
https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.10/DEVELOPMENT.md#single-word-entities

pkg/component/http.go
```diff
@@ -0,0 +1,48 @@
...
+func (c *Component) HTTPClient(ctx context.Context) (*http.Client, error) {
+	transport, err := c.HTTPTransport(ctx)
+	if err != nil {
+		return nil, err
+	}
+	client := &http.Client{
+		Timeout:   defaultHTTPClientTimeout,
+		Transport: transport,
+	}
+	return client, nil
```
rvolosatovs: nit;
```suggestion
	return &http.Client{
		Timeout:   defaultHTTPClientTimeout,
		Transport: transport,
	}, nil
```

pkg/component/http.go
```diff
@@ -0,0 +1,48 @@
...
+	"net/http"
+	"time"
+)
+
+// defaultHTTPClientTimeout is the default timeout for the HTTP client.
+const defaultHTTPClientTimeout = 10 * time.Second
+
+// HTTPClient returns a new *http.Client with a default timeout and a configured transport.
+func (c *Component) HTTPClient(ctx context.Context) (*http.Client, error) {
+	transport, err := c.HTTPTransport(ctx)
```
rvolosatovs: nit;
```suggestion
	tr, err := c.HTTPTransport(ctx)
```

pkg/component/http.go
```diff
@@ -0,0 +1,48 @@
...
+	client := &http.Client{
+		Timeout:   defaultHTTPClientTimeout,
+		Transport: transport,
+	}
+	return client, nil
+}
+
+// HTTPTransport returns a new http.RoundTripper with TLS client configuration.
+func (c *Component) HTTPTransport(ctx context.Context) (http.RoundTripper, error) {
+	transport := http.DefaultTransport.(*http.Transport).Clone()
```
rvolosatovs: nit; this is not used before L46 and it does not fail - do this on L45

pkg/networkserver/networkserver.go
```diff
@@ -175,13 +175,21 @@ func New(c *component.Component, conf *Config, opts ...Option) (*NetworkServer,
...
 	var interopCl InteropClient
 	if !conf.Interop.IsZero() {
 		interopConf := conf.Interop
 		interopConf.GetFallbackTLSConfig = func(ctx context.Context) (*tls.Config, error) {
 			return c.GetTLSClientConfig(ctx)
 		}
 		interopConf.BlobConfig = c.GetBaseConfig(ctx).Blob
+		if interopConf.Transport == nil {
+			interopConf.Transport = transport
+		}
```
rvolosatovs:
```suggestion
	var interopCl InteropClient
	if !conf.Interop.IsZero() {
		interopConf := conf.Interop
		interopConf.GetFallbackTLSConfig = func(ctx context.Context) (*tls.Config, error) {
			return c.GetTLSClientConfig(ctx)
		}
		interopConf.BlobConfig = c.GetBaseConfig(ctx).Blob
		if interopConf.Transport == nil {
			tr, err := c.HTTPTransport(ctx)
			if err != nil {
				return nil, err
			}
			interopConf.Transport = tr
		}
```

rvolosatovs (changes requested):

rvolosatovs (approved):

neoaggelos:
This needs codeowner approval to merge.

pkg/applicationserver/io/packages/loradms/v1/package.go
```diff
@@ -307,13 +306,10 @@ func (p *DeviceManagementPackage) Package() *ttnpb.ApplicationPackage {
 }
 
 // New instantiates the LoRa Cloud Device Management package.
-func New(server io.Server, registry packages.Registry) packages.ApplicationPackageHandler {
+func New(ctx context.Context, server io.Server, registry packages.Registry) packages.ApplicationPackageHandler {
```
KrishnaIyer: This isn't related right?

KrishnaIyer (commented):
Code owned files LGTM.

To what extent is this tested?

> Regressions
Possible, since different parts of the codebase dealt with HTTP clients and transports in all sorts of ways.

That doesn't seem very convincing to me.

neoaggelos:
> > Regressions
> > Possible, since different parts of the codebase dealt with HTTP clients and transports in all sorts of ways.
> 
> That doesn't seem very convincing to me.

Indeed, most of this is not covered by our unit tests (but should be covered by the end to end tests). I'd rather this is merged early for 3.12 instead, to avoid disruptions with 3.11

adriansmares (approved):

KrishnaIyer (approved):
Code owned files LGTM

htdvisser (approved):

PR #3525, opened 2020-11-25
console: Add hover menu to sidebar in collapsed state

### LINKED ISSUE #1157: Add a hover menu to the sidebar navigation, when collapsed
#### Summary

The sidebar menu is not sufficiently accessible when collapsed; subitems cannot be accessed while the menu is collapsed. It should provide a hover/overlay menu, showing the available sub-entries on hover (and via keyboard focus) when collapsed.

#### Why do we need this?

To enable better usage and improve accessibility and responsiveness in small screen sizes.

Ability to collapse the sidebar.

Adding hover/overlay menus to make sub-entries accessible in this mode.

#### Environment

Console.

Basically use the `<Dropdown />`\-component to render the full title of the entry and any possible sub-entries on hover. Just having the icons is not enough.

Yes but help welcome @pgalic96 @mjamescompton

References #1172

asmulko:
#### Summary
Closes #1157

#### Changes

- Added hover menu to sidebar menu in collapsed (minimized) state.
- Updated side menu parent button to highlight if one of the children is active.

Before:
before1
before2

Now:
after_1
after2

#### Testing

Manually tested

#### Checklist

### DIFF
--- a/pkg/webui/components/dropdown/dropdown.styl
+++ b/pkg/webui/components/dropdown/dropdown.styl
@@ -32,6 +32,17 @@ ul.dropdown
     margin-top: 0
     height: .4rem
 
+  li.dropdown-header-item
+    border-normal('bottom')
+    display: block
+    margin-bottom: 0
+    font-weight: $fw.bold
+
+    span
+      line-height: 1
+      display: block
+      padding: $cs.m
+
   li.dropdown-item
     display: block
     margin-bottom: 0
--- a/pkg/webui/components/dropdown/index.js
+++ b/pkg/webui/components/dropdown/index.js
@@ -46,15 +46,18 @@ Dropdown.defaultProps = {
   onItemsClick: () => null,
 }
 
-const DropdownItem = function({ icon, title, path, action, exact }) {
+const DropdownItem = ({ icon, title, path, action, exact, showActive, tabIndex }) => {
   const iconElement = icon && <Icon className={style.icon} icon={icon} nudgeUp />
+  const activeClassName = classnames({
+    [style.active]: showActive,
+  })
   const ItemElement = action ? (
-    <button onClick={action} onKeyPress={action} role="tab" tabIndex="0">
+    <button onClick={action} onKeyPress={action} role="tab" tabIndex={tabIndex}>
       {iconElement}
       <Message content={title} />
     </button>
   ) : (
-    <NavLink activeClassName={style.active} to={path} exact={exact}>
+    <NavLink activeClassName={activeClassName} to={path} exact={exact} tabIndex={tabIndex}>
       {iconElement}
       <Message content={title} />
     </NavLink>
@@ -71,6 +74,8 @@ DropdownItem.propTypes = {
   exact: PropTypes.bool,
   icon: PropTypes.string.isRequired,
   path: PropTypes.string,
+  showActive: PropTypes.bool,
+  tabIndex: PropTypes.string,
   title: PropTypes.message.isRequired,
 }
 
@@ -78,8 +83,23 @@ DropdownItem.defaultProps = {
   action: undefined,
   exact: false,
   path: undefined,
+  showActive: true,
+  tabIndex: '0',
+}
+
+const DropdownHeaderItem = ({ title }) => (
+  <li className={style.dropdownHeaderItem}>
+    <span>
+      <Message content={title} />
+    </span>
+  </li>
+)
+
+DropdownHeaderItem.propTypes = {
+  title: PropTypes.message.isRequired,
 }
 
 Dropdown.Item = DropdownItem
+Dropdown.HeaderItem = DropdownHeaderItem
 
 export default Dropdown
--- a/pkg/webui/components/navigation/side/item/index.js
+++ b/pkg/webui/components/navigation/side/item/index.js
@@ -12,11 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import React, { Fragment } from 'react'
+import React from 'react'
 import classnames from 'classnames'
 import bind from 'autobind-decorator'
 import { withRouter } from 'react-router-dom'
 
+import Dropdown from '@ttn-lw/components/dropdown'
 import Icon from '@ttn-lw/components/icon'
 
 import Message from '@ttn-lw/lib/components/message'
@@ -65,6 +66,7 @@ export class SideNavigationItem extends React.PureComponent {
 
   handleExpandCollapsableItem() {
     this.setState({ isExpanded: !this.state.isExpanded })
+    document.activeElement.blur()
   }
 
   componentDidMount() {
@@ -85,8 +87,14 @@ export class SideNavigationItem extends React.PureComponent {
     }
   }
 
+  handleItemClick = event => {
+    if (event && event.target) {
+      event.target.blur()
+    }
+  }
+
   render() {
-    const { className, children, title, depth, icon, path, exact, isActive } = this.props
+    const { className, children, title, depth, icon, path, exact, isActive, location } = this.props
     const { isExpanded } = this.state
     const { isMinimized, onLeafItemClick } = this.context
 
@@ -106,6 +114,8 @@ export class SideNavigationItem extends React.PureComponent {
             isExpanded={isExpanded}
             isMinimized={isMinimized}
             children={children}
+            currentPathName={location.pathname}
+            onDropdownItemsClick={this.handleItemClick}
           />
         ) : (
           <LinkItem
@@ -115,6 +125,7 @@ export class SideNavigationItem extends React.PureComponent {
             exact={exact}
             path={path}
             depth={depth}
+            onDropdownItemsClick={this.handleItemClick}
           />
         )}
       </li>
@@ -127,69 +138,105 @@ const CollapsableItem = ({
   onClick,
   isActive,
   isExpanded,
-  isMinimized,
   title,
   icon,
   depth,
-}) => (
-  <Fragment>
-    <button
-      className={classnames(style.button, {
-        [style.buttonActive]: isActive,
-      })}
-      type="button"
-      data-hook="side-nav-item-button"
-      onClick={onClick}
-    >
-      {icon && <Icon icon={icon} className={style.icon} />}
-      <Message content={title} className={style.message} />
-      <Icon
-        icon="keyboard_arrow_down"
-        className={classnames(style.expandIcon, {
-          [style.expandIconOpen]: isExpanded,
+  currentPathName,
+  onDropdownItemsClick,
+}) => {
+  const subItems = children.map(item => {
+    return {
+      title: item.props.title,
+      path: item.props.path,
+      icon: item.props.icon,
+    }
+  })
+
+  const subItemActive = subItems.some(item => item.path === currentPathName)
+
+  return (
+    <>
+      <button
+        className={classnames(style.button, {
+          [style.buttonActive]: isActive,
+          [style.linkActive]: subItemActive,
         })}
-      />
-    </button>
-    <SideNavigationList isMinimized={isMinimized} depth={depth + 1} isExpanded={isExpanded}>
-      {children}
-    </SideNavigationList>
-  </Fragment>
-)
+        type="button"
+        onClick={onClick}
+      >
+        {icon && <Icon icon={icon} className={style.icon} />}
+        <Message content={title} className={style.message} />
+        <Icon
+          icon="keyboard_arrow_down"
+          className={classnames(style.expandIcon, {
+            [style.expandIconOpen]: isExpanded,
+          })}
+        />
+      </button>
+      <Dropdown className={style.flyOutList} onItemsClick={onDropdownItemsClick}>
+        <Dropdown.HeaderItem title={title.defaultMessage} />
+        {subItems.map(item => (
+          <Dropdown.Item key={item.path} title={item.title} path={item.path} icon={item.icon} />
+        ))}
+      </Dropdown>
+      <SideNavigationList depth={depth + 1} isExpanded={isExpanded} className={style.subItems}>
+        {children}
+      </SideNavigationList>
+    </>
+  )
+}
 
 CollapsableItem.propTypes = {
   children: PropTypes.node,
+  currentPathName: PropTypes.string.isRequired,
   depth: PropTypes.number.isRequired,
   icon: PropTypes.string,
   isActive: PropTypes.bool.isRequired,
   isExpanded: PropTypes.bool.isRequired,
-  isMinimized: PropTypes.bool,
   onClick: PropTypes.func.isRequired,
+  onDropdownItemsClick: PropTypes.func,
   title: PropTypes.message.isRequired,
 }
 
 CollapsableItem.defaultProps = {
   children: undefined,
   icon: undefined,
-  isMinimized: false,
+  onDropdownItemsClick: () => null,
 }
 
-const LinkItem = ({ onClick, title, icon, exact, path }) => (
-  <NavigationLink
-    onClick={onClick}
-    className={style.link}
-    activeClassName={style.linkActive}
-    exact={exact}
-    path={path}
-  >
-    {icon && <Icon icon={icon} className={style.icon} />}
-    <Message content={title} className={style.message} />
-  </NavigationLink>
-)
+const LinkItem = ({ onClick, title, icon, exact, path, onDropdownItemsClick }) => {
+  const handleLinkItemClick = React.useCallback(
+    event => {
+      document.activeElement.blur()
+      onClick(event)
+    },
+    [onClick],
+  )
+
+  return (
+    <>
+      <NavigationLink
+        onClick={handleLinkItemClick}
+        className={style.link}
+        activeClassName={style.linkActive}
+        exact={exact}
+        path={path}
+      >
+        {icon && <Icon icon={icon} className={style.icon} />}
+        <Message content={title} className={style.message} />
+      </NavigationLink>
+      <Dropdown className={style.flyOutList} onItemsClick={onDropdownItemsClick}>
+        <Dropdown.Item title={title} path={path} showActive={false} icon={''} tabIndex="-1" />
+      </Dropdown>
+    </>
+  )
+}
 
 LinkItem.propTypes = {
   exact: PropTypes.bool.isRequired,
   icon: PropTypes.string,
   onClick: PropTypes.func,
+  onDropdownItemsClick: PropTypes.func,
   path: PropTypes.string,
   title: PropTypes.message.isRequired,
 }
@@ -198,6 +245,7 @@ LinkItem.defaultProps = {
   icon: undefined,
   path: undefined,
   onClick: () => null,
+  onDropdownItemsClick: () => null,
 }
 
 export default withRouter(bind(SideNavigationItem))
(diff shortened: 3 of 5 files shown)


pkg/webui/components/navigation/side/item/item.styl
```diff
@@ -29,6 +30,15 @@
         height: auto
         justify-content: space-evenly
 
+      &:focus-within,
```
asmulko, line 34: I'm really not sure about this one and need help.
One of the requirements was to support accessing child elements via 'keyboard focus', so I did it with focus-within.
However adding this css property resulted in hover menu not closing after click.

Could anyone please suggest a proper solution to this issue or what am I missing?

Thanks!

pkg/webui/components/navigation/side/item/index.js
```diff
@@ -15,7 +15,8 @@
 import React, { Fragment } from 'react'
 import classnames from 'classnames'
 import bind from 'autobind-decorator'
-import { withRouter } from 'react-router-dom'
+import { NavLink, withRouter } from 'react-router-dom'
+import { some } from 'lodash'
```
bafonins: Use standard [some](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/some).

pkg/webui/components/navigation/side/item/item.styl
```diff
@@ -118,3 +128,41 @@
...
+    right: -1px
+    z-index: -1
+
+  a
+    display: inline-block
+    height: 100%
+    width: 100%
+
+.fly-out-item
+  font-weight: 600
```
bafonins:
```suggestion
  font-weight: $fw.bold
```

pkg/webui/components/navigation/side/item/item.styl
```diff
@@ -118,3 +128,41 @@
...
+
+  &-divider
+    height: 0
+    margin: 4px 0
+    overflow: hidden
+    border-top: 1px solid #dbdbdb
```
bafonins: do not hardcode colors, see here https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.10/pkg/webui/styles/variables/generic.styl#L15-L39

pkg/webui/components/navigation/side/item/item.styl
```diff
@@ -29,6 +30,15 @@
         height: auto
         justify-content: space-evenly
 
+      &:focus-within,
+      &:hover
+        .fly-out-list
+          display: block
+          visibility: visible
+          transform: translate3d(43px, -47px, 0px)
```
bafonins: Why `translate3d`? Isnt [enough](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/translate) ?

bafonins: Such solution does not account for page scroll:

Screenshot 2020-11-25 at 14 51 43

pkg/webui/components/navigation/side/item/item.styl
```diff
@@ -29,6 +30,15 @@
...
+          padding: 10px
```
bafonins: https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.10/pkg/webui/styles/variables/generic.styl#L43-L72

pkg/webui/components/navigation/side/item/item.styl
bafonins: See also [mixins](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.10/pkg/webui/styles/mixins.styl#L15-L63)

pkg/webui/components/navigation/side/item/index.js
```diff
@@ -127,63 +129,98 @@ const CollapsableItem = ({
...
-    onClick={onClick}
-    className={style.link}
-    activeClassName={style.linkActive}
-    exact={exact}
-    path={path}
-  >
-    {icon && <Icon icon={icon} className={style.icon} />}
-    <Message content={title} className={style.message} />
-  </NavigationLink>
+  <Fragment>
```
bafonins:
```suggestion
  <>
```

pkg/webui/components/navigation/side/item/index.js
```diff
@@ -127,63 +129,98 @@ const CollapsableItem = ({
...
+      <Message content={title} className={style.message} />
+    </NavigationLink>
+    <ul className={style.flyOutList}>
+      <li className={style.flyOutItem}>
+        <NavLink activeClassName={style.flyOutItemActiveLink} to={path} exact={exact}>
+          <Message content={title.defaultMessage} />
+        </NavLink>
+      </li>
+    </ul>
+  </Fragment>
```
bafonins:
```suggestion
  </>
```

pkg/webui/components/navigation/side/item/index.js
```diff
@@ -127,63 +129,98 @@ const CollapsableItem = ({
...
+        <Icon
+          icon="keyboard_arrow_down"
+          className={classnames(style.expandIcon, {
+            [style.expandIconOpen]: isExpanded,
+          })}
+        />
+      </button>
+      <ul className={style.flyOutList}>
+        {subItems.length > 0 && (
+          <Fragment>
```
bafonins:
```suggestion
          <>
```

pkg/webui/components/navigation/side/item/index.js
```diff
@@ -127,63 +129,98 @@ const CollapsableItem = ({
...
+            <li className={style.flyOutItem}>{title.defaultMessage}</li>
+            <li className={style.flyOutItemDivider} />
+            {subItems.map(item => (
+              <li key={item.path}>
+                <NavLink activeClassName={style.flyOutItemActiveLink} to={item.path} exact={false}>
+                  <Message content={item.title} />
+                </NavLink>
+              </li>
+            ))}
+          </Fragment>
```
bafonins:
```suggestion
          </>
```

bafonins (changes requested):

pkg/webui/components/navigation/side/item/index.js
```diff
@@ -127,63 +129,98 @@ const CollapsableItem = ({
...
+      {icon && <Icon icon={icon} className={style.icon} />}
+      <Message content={title} className={style.message} />
+    </NavigationLink>
+    <ul className={style.flyOutList}>
+      <li className={style.flyOutItem}>
+        <NavLink activeClassName={style.flyOutItemActiveLink} to={path} exact={exact}>
+          <Message content={title.defaultMessage} />
+        </NavLink>
+      </li>
+    </ul>
```
kschiffer: Please use our existing `<Dropdown />` component and make necessary adjustments to it, if need be.

pkg/webui/components/navigation/side/item/item.styl
kschiffer: I think you'll need to blur the element programmatically after the route change. E.g:

```
const handleLinkClick = e => {
  
  // Do route change stuff...
  
  // Blur the element.
  e.target.blur()
}
```

pkg/webui/components/navigation/side/item/item.styl
kschiffer: Indeed, `transform` is not the way to go here (and rarely ever). You'll need to use `position: absolute` with `left: calc(3rem + 1px)`. That way, it will stay attached to the menu entry, regardless of the scrolling position. To make this work, you'll also need to remove some of the `overflow: hidden` styles in the parent containers. If you do so, make sure that it does not introduce other layout issues, e.g. for in mobile screen sizes.

pkg/webui/components/navigation/side/item/item.styl
```diff
@@ -118,3 +128,41 @@
...
+  visibility hidden
+  position: fixed
+  min-width: 150px
+  padding: 4px 1px
+  background-color: #fff
+  box-shadow: 2px 1px 3px rgba(0,0,0,0.1)
+  border: 1px solid #eee
+  border-left: 0
+  border-radius: 0 3px 3px 0
+  list-style: none
```
kschiffer: *   Avoid using absolute pixel values and use relative `rem` values instead
*   Use mixins to solve common styling problems (as suggested by Bogdans already)
*   Use variables to use consistent values (e.g. for border radius: `$br.s`, colors, border, etc)

pkg/webui/components/navigation/side/item/item.styl
```diff
@@ -118,3 +128,41 @@
 
   &-open
     transform: translateY(-50%) rotate(180deg)
+
+.fly-out-list
+  visibility hidden
```
kschiffer:
```suggestion
  display: none
```

Note also that you forgot the `:` here.

kschiffer (changes requested):
Functionality-wise, this is working pretty well already. There are some implementation issues though as noted in the comments. There should also be some hover styling, but that should be addressed when you use `<Dropdown />` as noted in the comments.

pkg/webui/components/navigation/side/item/item.styl
asmulko: @bafonins could you please comment on this as well? Would this be the preferred solution or there's another way to do this? For example: autofocusing elements on page we get navigated to.

pkg/webui/components/navigation/side/item/index.js
```diff
@@ -127,63 +129,86 @@ const CollapsableItem = ({
...
         })}
-      />
-    </button>
-    <SideNavigationList isMinimized={isMinimized} depth={depth + 1} isExpanded={isExpanded}>
-      {children}
-    </SideNavigationList>
-  </Fragment>
-)
+        type="button"
+        data-hook="side-nav-item-button"
```
bafonins: This can be removed

pkg/webui/components/dropdown/dropdown.styl
```diff
@@ -32,6 +32,25 @@ ul.dropdown
     margin-top: 0
     height: .4rem
 
+  li.dropdown-header-item
+    display: block
+    margin-bottom: 0
+    font-weight: $fw.bold
+
+    border-normal('bottom')
```
bafonins: We prefer using mixins before defining any styles on an element.

bafonins (changes requested):
Indeed, it seems that blurring an element programmatically is the easiest solution.

```js
  const handleItemClick = React.useCallback(
    event => {
      if (event && event.target) {
        event.target.blur()
      }

      onClick(event)
    },
    [onClick],
  )
```

Should do the trick. Just pass this callback in both `<CollapsableItem />` and `<LinkItem />` components.

asmulko:
> Indeed, it seems that blurring an element programmatically is the easiest solution.
> 
> ```js
>   const handleItemClick = React.useCallback(
>     event => {
>       if (event && event.target) {
>         event.target.blur()
>       }
> 
>       onClick(event)
>     },
>     [onClick],
>   )
> ```
> 
> Should do the trick. Just pass this callback in both `<CollapsableItem />` and `<LinkItem />` components.

I just found out that it's possible to pass `onItemsClick` prop to the `Dropdown` components and it did the trick with the bit of code you provided.

kschiffer:
A lot better now. There is still one issue though: When clicking on the icon (left of the dropdown) instead of the menu entry, the dropdown will stay visible.

I will go through the code again in a bit.

pkg/webui/components/navigation/side/item/item.styl
```diff
@@ -118,3 +126,15 @@
...
+  display: none
+
+  &::before
+    content: ''
+    position: absolute
+    top: -30px
+    bottom: -30px
+    left: -10px
+    right: -30px
+    z-index: -1
```
kschiffer, line 141: While this is an interesting approach, it is too intrusive. You can achieve a similar effect by using css transitions with delay, so that the fly out menu will not hide immediately, but after a small cooldown (~.5s)

pkg/webui/components/navigation/side/item/item.styl
```diff
@@ -29,6 +30,13 @@
         height: auto
         justify-content: space-evenly
 
+      &:focus-within,
+      &:hover
+        .fly-out-list
+          display: block
+          left: calc(3rem + 1px)
+          transform: translate(0px, -46px)
```
kschiffer: Again, don't use `transform: translate()` to achieve layout effects. This should really be used for special cases only (such as e.g. rotating elements). When you have to use `transform` to achieve layouts, it is a good indicator that there's a broader styling issue.

You can achieve the positioning without transform here. The key is setting `position: relative` to the parent. Try the following diff:

```diff
diff --git a/pkg/webui/components/navigation/side/item/item.styl b/pkg/webui/components/navigation/side/item/item.styl
index b2724d4f1..2c1604679 100644
--- a/pkg/webui/components/navigation/side/item/item.styl
+++ b/pkg/webui/components/navigation/side/item/item.styl
@@ -15,6 +15,7 @@
 .item
   +media-query-min($bp.s)
     &-minimized
+      position: relative
       .message,
       .expand-icon
       .sub-items
@@ -34,8 +35,8 @@
       &:hover
         .fly-out-list
           display: block
+          top: 0
           left: calc(3rem + 1px)
-          transform: translate(0px, -46px)
 
 .button
   reset-button()
diff --git a/pkg/webui/components/navigation/side/side.styl b/pkg/webui/components/navigation/side/side.styl
index 975b050e2..06823f743 100644
--- a/pkg/webui/components/navigation/side/side.styl
+++ b/pkg/webui/components/navigation/side/side.styl
@@ -99,6 +99,12 @@ $minimize-button-height = 5rem
         .icon
           margin-right: 0
 
+    .navigation-list
+      overflow: visible
+
+    .drawer
+      overflow: visible
+
 .header
   border-normal(bottom)
   sidebar-transition(padding)
@@ -154,3 +160,4 @@ $minimize-button-height = 5rem
 .scroll-lock
   +media-query($bp.s)
     overflow: hidden
+
```

pkg/webui/components/navigation/side/item/index.js
```diff
@@ -106,6 +113,8 @@ export class SideNavigationItem extends React.PureComponent {
             isExpanded={isExpanded}
             isMinimized={isMinimized}
             children={children}
+            currentPathName={location.pathname}
+            onDropdownItemsClick={this.handleItemClick}
           />
         ) : (
           <LinkItem
```
kschiffer, line 121: I think you'll have to apply the same blur on click logic to the LinkItem as well, otherwise the dropdown will stay visible when the user clicks on this LinkItem instead of the dropdown item.

kschiffer (changes requested):

pkg/webui/components/navigation/side/item/index.js
asmulko: Was not able to get this working with `event.target`, but `document.activeElement` did the job. Let me know if this needs an update as well.

pkg/webui/components/dropdown/dropdown.styl
```diff
@@ -32,6 +32,24 @@ ul.dropdown
...
+      reset-button()
+
+    & > button
+      box-sizing: border-box
+      line-height: 1
+      display: block
+      text-decoration: none
+      padding: $cs.m
+      width: 100%
+      cursor: auto
```
kschiffer: Nit:
```suggestion
    button
      reset-button()
      box-sizing: border-box
      line-height: 1
      display: block
      text-decoration: none
      padding: $cs.m
      width: 100%
      cursor: auto
```

pkg/webui/components/dropdown/index.js
```diff
@@ -71,15 +74,30 @@ DropdownItem.propTypes = {
...
 DropdownItem.defaultProps = {
   action: undefined,
   exact: false,
   path: undefined,
+  showActive: true,
+}
+
+const DropdownHeaderItem = ({ title }) => (
+  <li className={style.dropdownHeaderItem}>
+    <button role="tab" tabIndex="-1">
```
kschiffer: This is semantically not a button, since it cannot be clicked. You should use `span` or `div` rather.

pkg/webui/components/navigation/side/item/item.styl
```diff
@@ -118,3 +125,9 @@
 
   &-open
     transform: translateY(-50%) rotate(180deg)
+
+.fly-out-list
+  left: calc(3rem + 1px)
+  top: 0
+  transition: visibility $ad.s
```
kschiffer: After trying this out, I have to say that your original solution (with the area around the fly-out) is actually preferable here, mainly because this introduces weird delays when hovering along the sidebar icons, wich would be quite hacky to circumvent.

So, let's revert this back. Mea culpa!

kschiffer (changes requested):
Almost good to go!

pkg/webui/components/navigation/side/item/item.styl
kschiffer: See my recent comment on this. Your solution here was actually fitting.

pkg/webui/components/dropdown/index.js
```diff
@@ -46,15 +46,18 @@ Dropdown.defaultProps = {
...
+    [style.active]: showActive,
+  })
   const ItemElement = action ? (
     <button onClick={action} onKeyPress={action} role="tab" tabIndex="0">
       {iconElement}
       <Message content={title} />
     </button>
   ) : (
-    <NavLink activeClassName={style.active} to={path} exact={exact}>
+    <NavLink activeClassName={activeClassName} to={path} exact={exact}>
```
kschiffer:
```suggestion
    <NavLink activeClassName={activeClassName} to={path} exact={exact} tabIndex={tabIndex}>
```

You have to pass the `tabIndex` prop here, and in the `button` element above. See my comment below.

pkg/webui/components/dropdown/index.js
```diff
@@ -46,15 +46,18 @@ Dropdown.defaultProps = {
   onItemsClick: () => null,
 }
 
-const DropdownItem = function({ icon, title, path, action, exact }) {
+const DropdownItem = ({ icon, title, path, action, exact, showActive }) => {
```
kschiffer:
```suggestion
const DropdownItem = ({ icon, title, path, action, exact, showActive, tabIndex }) => {
```

pkg/webui/components/navigation/side/item/index.js
```diff
@@ -127,69 +138,105 @@ const CollapsableItem = ({
...
+        className={style.link}
+        activeClassName={style.linkActive}
+        exact={exact}
+        path={path}
+      >
+        {icon && <Icon icon={icon} className={style.icon} />}
+        <Message content={title} className={style.message} />
+      </NavigationLink>
+      <Dropdown className={style.flyOutList} onItemsClick={onDropdownItemsClick}>
+        <Dropdown.Item title={title} path={path} showActive={false} icon={''} />
```
kschiffer:
```suggestion
        <Dropdown.Item title={title} path={path} showActive={false} icon={''} tabIndex="-1" />
```

Let's use `tabIndex` -1 here, since otherwise, users have to tab twice per entry. You need to pass this in the dropdown item component.

pkg/webui/components/navigation/side/item/item.styl
```diff
@@ -118,3 +125,17 @@
 
   &-open
     transform: translateY(-50%) rotate(180deg)
+
+.fly-out-list
+  display: none
+  left: calc(3rem + 1px)
```
kschiffer:
```suggestion
  left: 3rem
```

Looks nicer.

kschiffer (approved):
Awesome. See my remarks with some last tweaks.

pkg/webui/components/navigation/side/item/index.js
asmulko: Great point

bafonins (approved):
Looks good 👍

kschiffer (approved):
👍

PR #3530, opened 2020-11-26
Device Repository

### LINKED ISSUE #3433: Integrate Device Repository

neoaggelos:
#### Summary

Closes #3433 

#### Changes

- Add new DeviceRepository API and messages under `api/devicerepository.proto`
- Replace `EndDeviceBrand` and `EndDeviceModel` messages and move them to `api/devicerepository.proto`
- Add a new DeviceRepository component
-  Extend list of allowed Field Mask Paths for the new RPCs.
- Add new `BandID` field in IS end devices.
- Everything related to the device repository store (fetching brands/models/templates/codecs) has been extracted to #3543

#### Testing

- Unit testing
- Test locally

##### Regressions

There should not be any

#### Notes for Reviewers

The API messages closely follow the format of the [LoRaWAN devices schema](https://lorawan-schema.org/draft/devices/1/). For enum values (e.g. band IDs, LoRaWAN version), then we convert to the TTS. 

Using `brand` and `model` everywhere, because `BrandID` already exists in our protos and in the database.

#### Checklist


api/devicerepository.proto
```diff
@@ -0,0 +1,232 @@
...
+package ttn.lorawan.v3;
+
+option go_package = "go.thethings.network/lorawan-stack/v3/pkg/ttnpb";
+
+message EndDeviceBrand {
+  string brand_id = 1 [(gogoproto.customname) = "BrandID"];
+  string name = 2;
+
+  uint32 private_enterprise_number = 3;
+  repeated string organization_unique_identifiers = 4;
```
johanstokking: Add some comments here

api/devicerepository.proto
```diff
@@ -0,0 +1,232 @@
...
+  uint32 vendor_id = 5 [(gogoproto.customname) = "VendorID"];
```
johanstokking: `lora_alliance_vendor_id`

api/devicerepository.proto
```diff
@@ -0,0 +1,232 @@
...
+  string logo = 8;
+}
+
+message EndDeviceModel {
+  string brand_id = 1 [(gogoproto.customname) = "BrandID"];
+  string model_id = 2 [(gogoproto.customname) = "ModelID"];
+  string name = 3;
+  string description = 4;
+
+  message Version {
```
johanstokking: `HardwareVersion`

api/devicerepository.proto
```diff
@@ -0,0 +1,232 @@
...
+
+  message FirmwareVersion {
+    message Profile {
+      string profile_id = 1 [(gogoproto.customname) = "ProfileID"];
+      bool lorawan_certified = 2 [(gogoproto.customname) = "LoRaWANCertified"];
+      string codec_id = 3 [(gogoproto.customname) = "CodecID"];
+    }
+    string version = 1;
+    uint32 numeric = 2;
+    repeated string hardware_versions = 3;
```
johanstokking: `supported_hardware_versions`

api/devicerepository.proto
```diff
@@ -0,0 +1,232 @@
...
+    map<string, Profile> profiles = 4;
+  }
+  repeated FirmwareVersion firmware_versions = 6;
+  repeated string sensors = 7;
+  message Dimensions {
+    float width = 1;
+    float height = 2;
+    float diameter = 3;
+    float length = 4;
```
johanstokking: These are `FloatValue`s too

api/devicerepository.proto
```diff
@@ -0,0 +1,232 @@
...
+      google.protobuf.FloatValue min = 1;
+      google.protobuf.FloatValue max = 2;
+    }
+    Limits temperature = 1;
+    Limits relative_humidity = 2;
+  }
+  OperatingConditions operating_conditions = 11;
+
+  string ip_code = 12 [(gogoproto.customname) = "IPCode"];
+  repeated string key_provisioning = 13;
```
johanstokking: This can be an enum

api/devicerepository.proto
```diff
@@ -0,0 +1,232 @@
...
+  string key_security = 14;
```
johanstokking: This too

api/devicerepository.proto
```diff
@@ -0,0 +1,232 @@
...
+  repeated FirmwareVersion firmware_versions = 6;
+  repeated string sensors = 7;
+  message Dimensions {
+    float width = 1;
+    float height = 2;
+    float diameter = 3;
+    float length = 4;
+  }
+  Dimensions dimensions = 8;
+  float weight = 9;
```
johanstokking: `FloatValue`

api/devicerepository.proto
```diff
@@ -0,0 +1,232 @@
...
+
+message ListEndDeviceBrandsRequest {
+  // Optionally limit to a specific brand ID
+  string brand_id = 1 [(gogoproto.customname) = "BrandID"];
+
+  // Limit (for pagination)
+  google.protobuf.UInt32Value limit = 2;
+
+  // Offset (for pagination)
+  google.protobuf.UInt32Value offset = 3;
```
johanstokking: Align this with other List requests

api/devicerepository.proto
```diff
@@ -0,0 +1,232 @@
...
+    repeated Compliance radio_equipment = 2;
+  }
+  Compliances compliances = 19;
+
+  repeated string additional_radios = 20;
+}
+
+message ListEndDeviceBrandsRequest {
+  // Optionally limit to a specific brand ID
+  string brand_id = 1 [(gogoproto.customname) = "BrandID"];
```
johanstokking: Limit listing brands by a specific brand? How does that work?

api/devicerepository.proto
```diff
@@ -0,0 +1,232 @@
...
+  string search = 5;
+
+  google.protobuf.FieldMask field_mask = 6 [(gogoproto.nullable) = false];
+}
+
+message ListEndDeviceModelsRequest {
+  // Query end devices from a specific brand
+  string brand_id = 1 [(gogoproto.customname) = "BrandID"];
+  // Query end devices matching a specific model
+  string model_id = 2 [(gogoproto.customname) = "ModelID"];
```
johanstokking: Idem

api/devicerepository.proto
```diff
@@ -0,0 +1,232 @@
...
+
+message ListEndDeviceModelsResponse {
+  repeated EndDeviceModel models = 1;
+  uint32 total = 2;
+  uint32 count = 3;
+  uint32 offset = 4;
+}
+
+message MessagePayloadFormatter {
+  string codec = 1;
```
johanstokking: It should be clear here that it's JavaScript. We want to be able to support other script types as well. It's probably easiest if you use the same formatter type and parameter couple here.

api/devicerepository.proto
```diff
@@ -0,0 +1,232 @@
...
+
+  rpc GetTemplate(EndDeviceVersionIdentifiers) returns (EndDeviceTemplate) {
+    option (google.api.http) = {
+      get: "/dr/brands/{brand_id}/models/{model_id}/{firmware_version}/{band_id}/template"
+    };
+  };
+
+  rpc GetUplinkDecoder(EndDeviceVersionIdentifiers) returns (MessagePayloadFormatter) {
+    option (google.api.http) = {
+      get: "/dr/brands/{brand_id}/models/{model_id}/{firmware_version}/{band_id}/formatters/uplink_decoder"
```
johanstokking: Do we use underscores in the path?

pkg/devicerepository/store/enums.go
```diff
@@ -0,0 +1,81 @@
...
+	"AU915-928": band.AU_915_928,
+	"CN470-510": band.CN_470_510,
+	"AS923":     band.AS_923,
+	"KR920-923": band.KR_920_923,
+	"IN865-867": band.IN_865_867,
+	"RU864-870": band.RU_864_870,
+}
+
+// BandIDToRegion is the inverse mapping of RegionToBandID.
+var BandIDToRegion map[string]string
```
johanstokking: What do we need this for?

Also why are these mappings exposed?

johanstokking (commented):
It would be preferable if you extract this into (1) the DR component with tooling, config, start command etc, (2) the API (can be part of 1), and (3) the logic. You may want to target a large feature branch that we then merge in one go.

api/devicerepository.proto
kschiffer: If this is meant to be a filter by a substring, then this is not necessary for the implementation the Console.

api/devicerepository.proto
kschiffer: If this is meant to be a filter by a substring, then this is not necessary for the implementation the Console.

kschiffer (commented):
The proto draft looks good to me and is in line with how we plan to use the API in the Console.

johanstokking:
@neoaggelos please request review when ready

api/devicerepository.proto
```diff
@@ -0,0 +1,388 @@
...
+  // Device battery information.
+  Battery battery = 10;
+  message OperatingConditions {
+    message Limits {
+      // Min value of operating conditions range.
+      google.protobuf.FloatValue min = 1;
+      // Max value of operating conditions range.
+      google.protobuf.FloatValue max = 2;
+    }
+    // Temperature operating conditions.
```
johanstokking: Indicate that this is in Celcius

api/devicerepository.proto
```diff
@@ -0,0 +1,388 @@
...
+    Limits temperature = 1;
+    // Relative humidity operating conditions.
```
johanstokking: Indicate that this is a fraction (i.e. 0..1)

api/devicerepository.proto
```diff
@@ -0,0 +1,388 @@
...
+  // List of any additional radios included in the device.
+  repeated string additional_radios = 20;
+}
+
+message GetEndDeviceBrandRequest {
+  // Application identifiers.
+  ApplicationIdentifiers application_ids = 1 [(gogoproto.customname) = "ApplicationIDs"];
+
+  // Brand identifier, as defined in the Device Repository.
+  string brand_id = 21 [(gogoproto.customname) = "BrandID"];
```
johanstokking: Why 21?

api/devicerepository.proto
```diff
@@ -0,0 +1,388 @@
...
+  // List of standards the device is compliant with.
+  Compliances compliances = 19;
+
+  // List of any additional radios included in the device.
+  repeated string additional_radios = 20;
+}
+
+message GetEndDeviceBrandRequest {
+  // Application identifiers.
+  ApplicationIdentifiers application_ids = 1 [(gogoproto.customname) = "ApplicationIDs"];
```
johanstokking: What are these for? Also in other get requests

api/devicerepository.proto
```diff
@@ -0,0 +1,388 @@
...
+  string formatter_parameter = 2;
+}
+
+service DeviceRepository {
+  rpc ListBrands(ListEndDeviceBrandsRequest) returns (ListEndDeviceBrandsResponse) {
+    option (google.api.http) = {
+      get: "/dr/brands"
+
+      additional_bindings {
+        get: "/dr/applications/{application_ids.application_id}/brands"
```
johanstokking: What's the use case?

api/devicerepository.proto
```diff
@@ -0,0 +1,388 @@
...
+
+      additional_bindings {
+        get: "/dr/applications/{application_ids.application_id}/brands/{brand_id}"
+      }
+    };
+  }
+
+  rpc ListModels(ListEndDeviceModelsRequest) returns (ListEndDeviceModelsResponse) {
+    option (google.api.http) = {
+      get: "/dr/brands/{brand_id}/models"
```
johanstokking: How are we getting `search` in?

pkg/devicerepository/config.go
```diff
@@ -0,0 +1,41 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/devicerepository/store"
+)
+
+// Config represents the DeviceRepository configuration.
+type Config struct {
+	Store store.Store `name:"-"`
+
+	Static    map[string][]byte `name:"-"`
+	Directory string            `name:"directory" description:"Retrieve devices from the filesystem"`
+	URL       string            `name:"url" description:"Retrieve devices from a web server"`
```
johanstokking: Make the source explicit with a config key.

The implicit way of figuring out which source to use based on an order of precedence is legacy.

pkg/devicerepository/config.go
```diff
@@ -0,0 +1,41 @@
...
+
+	AssetsBaseURL string `name:"assets-base-url" description:"The base URL for assets"`
+	RequireAuth   bool   `name:"require-auth" description:"Require authentication for the device repository"`
```
johanstokking: What does this mean really?

pkg/devicerepository/grpc.go
```diff
@@ -0,0 +1,247 @@
...
+// ListBrands implements the ttnpb.DeviceRepositoryServer interface.
+func (dr *DeviceRepository) ListBrands(ctx context.Context, req *ttnpb.ListEndDeviceBrandsRequest) (*ttnpb.ListEndDeviceBrandsResponse, error) {
+	if dr.config.RequireAuth {
+		if req.ApplicationIDs == nil {
+			return nil, errNoAuthentication.New()
+		}
+		if err := rights.RequireApplication(ctx, *req.ApplicationIDs, ttnpb.RIGHT_APPLICATION_DEVICES_READ); err != nil {
+			return nil, err
+		}
+	}
```
johanstokking: What is the use case of _not_ requiring auth?

pkg/identityserver/store/end_device.go
```diff
@@ -40,6 +40,7 @@ type EndDevice struct {
 	ModelID         string `gorm:"type:VARCHAR"`
 	HardwareVersion string `gorm:"type:VARCHAR"`
 	FirmwareVersion string `gorm:"type:VARCHAR"`
+	BandID          string `gorm:"type:VARCHAR"`
```
johanstokking, line 43: In `EndDevice`, this now competing with `FrequencyPlanID`. That field is not stored in IS, but in NS. I do think it is valuable to keep this here. @rvolosatovs thoughts?

pkg/ttnpb/field_mask_validation.go
```diff
@@ -866,6 +866,12 @@ var AllowedFieldMaskPathsForRPC = map[string][]string{
 	"/ttn.lorawan.v3.UserRegistry/List":                omitFields(UserFieldPathsNested, "password", "temporary_password"),
 	"/ttn.lorawan.v3.UserRegistry/Update":              omitFields(UserFieldPathsNested, "password", "password_updated_at"),
 	"/ttn.lorawan.v3.EntityRegistrySearch/SearchUsers": omitFields(UserFieldPathsNested, "password", "temporary_password"),
+
+	// Device Repository:
+	"/ttn.lorawan.v3.DeviceRepository/GetBrand":   EndDeviceBrandFieldPathsNested,
+	"/ttn.lorawan.v3.DeviceRepository/ListBrands": EndDeviceBrandFieldPathsNested,
+	"/ttn.lorawan.v3.DeviceRepository/GetModel":   EndDeviceModelFieldPathsNested,
+	"/ttn.lorawan.v3.DeviceRepository/ListModels": EndDeviceModelFieldPathsNested,
```
johanstokking: We don't need the others?

api/devicerepository.proto
neoaggelos: This is for authentication purposes. We cannot require application rights without application identifiers.

pkg/ttnpb/field_mask_validation.go
neoaggelos: The other RPCs do not accept field mask paths.

api/devicerepository.proto
neoaggelos: With a querystring parameter, e.g. `?search=uno`

api/devicerepository.proto
johanstokking: Right, but why make it open?

pkg/devicerepository/grpc.go
neoaggelos: Running a standalone Device Repository server, without the need to have an IS.

api/devicerepository.proto
neoaggelos: What do you mean by "make it open"?

bafonins:
Went through protos and tested dr API provided by @neoaggelos. Looks good to me!

pkg/devicerepository/grpc.go
johanstokking: There shall be an Identity Server in that case, with a mock application or anything.

Using TTS Device Repository component as a work-around for other use cases is fine, but with the strict precondition that it does not affect any code in The Things Stack, and certainly not security related.

api/devicerepository.proto
johanstokking: See below

johanstokking:
Authentication cannot be open, please remove that, see above.

When ready to review again, please request here.

neoaggelos:
@johanstokking Removed `RequireAuth` option completely.

pkg/devicerepository/config.go
```diff
@@ -32,13 +32,17 @@ type Config struct {
 	Blob         config.BlobPathConfig `name:"blob"`
 
 	AssetsBaseURL string `name:"assets-base-url" description:"The base URL for assets"`
-	RequireAuth   bool   `name:"require-auth" description:"Require authentication for the HTTP and gRPC endpoints"`
+}
+
+// StoreConfig represents configuration for the Device Repository store.
+type StoreConfig struct {
+	Store store.Store `name:"-"`
 }
```
johanstokking: Why?

pkg/devicerepository/grpc.go
```diff
@@ -46,18 +46,16 @@ func (dr *DeviceRepository) assetURL(brandID, path string) string {
...
 // ListBrands implements the ttnpb.DeviceRepositoryServer interface.
 func (dr *DeviceRepository) ListBrands(ctx context.Context, req *ttnpb.ListEndDeviceBrandsRequest) (*ttnpb.ListEndDeviceBrandsResponse, error) {
-	if dr.config.RequireAuth {
-		if req.ApplicationIDs == nil {
-			return nil, errNoAuthentication.New()
-		}
-		if err := rights.RequireApplication(ctx, *req.ApplicationIDs, ttnpb.RIGHT_APPLICATION_DEVICES_READ); err != nil {
-			return nil, err
-		}
+	if req.ApplicationIDs == nil {
```
johanstokking: Make these non-nullable

api/devicerepository.proto
```diff
@@ -173,13 +173,28 @@ message EndDeviceModel {
     // Links to other device videos.
     repeated string other = 2;
   }
+  // Device video.
+  string video = 16;
   // Device videos.
-  Videos videos = 16;
+  Videos videos = 17;
```
johanstokking: So what is it now?

johanstokking (commented):
LGTM but please see my last comment here

api/devicerepository.proto
```diff
@@ -183,18 +183,16 @@ message EndDeviceModel {
...
       string name = 1;
       // Reseller regions.
       repeated string region = 2;
       // Reseller URL.
       string url = 3 [(gogoproto.customname) = "URL"];
-    }
   }
   // Reseller URLs.
-  repeated ResellerURL reseller_urls = 20;
+  repeated Reseller resellers = 20 [(gogoproto.customname) = "Resellers"];
```
johanstokking: Note that this custom name is not necessary

cmd/ttn-lw-stack/commands/config.go
```diff
@@ -53,22 +55,24 @@ type Config struct {
...
+	GS:               shared_gatewayserver.DefaultGatewayServerConfig,
+	NS:               shared_networkserver.DefaultNetworkServerConfig,
+	AS:               shared_applicationserver.DefaultApplicationServerConfig,
+	JS:               shared_joinserver.DefaultJoinServerConfig,
+	Console:          shared_console.DefaultConsoleConfig,
+	GCS:              shared_gatewayconfigurationserver.DefaultGatewayConfigurationServerConfig,
+	DTC:              shared_devicetemplateconverter.DefaultDeviceTemplateConverterConfig,
+	QRG:              shared_qrcodegenerator.DefaultQRCodeGeneratorConfig,
+	PBA:              shared_packetbrokeragent.DefaultPacketBrokerAgentConfig,
+	DeviceRepository: shared_devicerepository.DefaultDeviceRepositoryConfig,
```
johanstokking: Inconsistent; must be DR

pkg/devicerepository/grpc.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	for _, model := range response.Models {
+		if photos := model.Photos; photos != nil {
+			photos.Main = dr.assetURL(model.BrandID, photos.Main)
+			for idx, photo := range photos.Other {
+				photos.Other[idx] = dr.assetURL(model.BrandID, photo)
+			}
+		}
+	}
```
johanstokking: Please extract this to avoid repeating this. Also it's likely that other assets are added here in the future

pkg/identityserver/store/end_device.go
johanstokking: This is still open

cmd/ttn-lw-stack/commands/config.go
neoaggelos: There is existing (but previously unused) `device-repository.*` config, so renaming that to `dr` is typically a breaking change. Personally, I like `DR` better as well, is it ok to change it in this case?

pkg/identityserver/store/end_device.go
neoaggelos: Since FrequencyPlanID maps to single a BandID, but a single BandID can be used by multiple Frequency Plans, I do not know what the action should be here. Note that this BandID is only used to fetch the correct profile from the device repository.

Maybe @rvolosatovs can provide some more context?

pkg/devicerepository/grpc.go
neoaggelos: Extract as in just a separate function, or something else? Did not do that initially because we tend to avoid declaring functions that are only used once in the codebase.

cmd/ttn-lw-stack/commands/config.go
rvolosatovs: You can also rename the field, but leave the tag as `device-repository`. Maybe even setup an alias?

pkg/identityserver/store/end_device.go
rvolosatovs: Sorry, but I lack context myself to say anything as it's unclear to me why band ID would be required to fetch a device profile. Also I don't know what IS has to do with this.
If you could explain what is happening here and why I could give you my thoughts.

In any case, device profile message in the backend interfaces contains `RFRegion`, which corresponds to our frequency plan ID perfectly fine I don't see why we'd ever want to use just the band ID here, since we can always derive band ID from the frequency plan ID.

neoaggelos: Also see https://github.com/TheThingsNetwork/lorawan-devices/blob/master/vendor/the-things-products/the-things-uno.yaml#L9-L17. In short, for the Device Repository there is a device profile per region.

cmd/ttn-lw-stack/commands/config.go
johanstokking: Indeed. This refers to `DeviceRepository` here in code

Anyway, I'm fine with deprecating the old config. It wasn't used anyway. But if you can keep it compatible, I'm good with that too.

pkg/devicerepository/grpc.go
johanstokking: This was already being done twice

pkg/devicerepository/grpc.go
```diff
@@ -0,0 +1,208 @@
...
+	if len(response.Models) == 0 {
+		return nil, errModelNotFound.WithAttributes("brand_id", req.BrandID, "model_id", req.ModelID)
+	}
+	model := response.Models[0]
+	if photos := model.Photos; photos != nil {
+		photos.Main = dr.assetURL(model.BrandID, photos.Main)
+		for idx, photo := range photos.Other {
+			photos.Other[idx] = dr.assetURL(model.BrandID, photo)
+		}
+	}
```
johanstokking: Make this use the func and adapt if necessary

pkg/identityserver/store/end_device.go
johanstokking: @rvolosatovs 

>In any case, device profile message in the backend interfaces contains `RFRegion`, which corresponds to our frequency plan ID perfectly fine I don't see why we'd ever want to use just the band ID here, since we can always derive band ID from the frequency plan ID.

It corresponds to the Band ID, that's the issue here.

So my question is, is the `BandID` in the IS authoritive or informative? If the former, should NS respect this value, and validate frequency plan IDs, or should we keep this decoupled? Decoupling may get inconsistent devices: band EU863-870 with frequency plan US902-915 FSB 2.

Please consider this from the developer perspective when dealing with our APIs.

neoaggelos: This `BandID` is used (by the AS) to retrieve the appropriate end device codecs from the Device Repository. Two things are of note here: First, AS is not aware of the Frequency Plan ID used by the device (and so it cannot infer the Band ID). Secondly, the frequency plan ID for the device is chosen as a separate step.

Indeed, validating the BandID with respect to the Frequency Plan ID would make sense when creating the device. However, the Network Server is also unaware of the version ids.

What should be the course of action here?

rvolosatovs: I think it's not the cleanest approach, but it makes more sense to validate the frequency plan in NS, especially from developer/customer standpoint, so I vote for authoritative

johanstokking:
Raising priority here if we want this part of the TTN V3 initial release.

@neoaggelos what is the status here?

If you don't request reviews, we're not looking at it.

neoaggelos:
Rebased over latest v3.11 and resolved conflicts with the newest API changes.

cmd/ttn-lw-stack/commands/config.go
htdvisser: Since this config was never used or documented, I don't think we violate compatibility by changing it to `dr`.

pkg/devicerepository/config.go
```diff
@@ -0,0 +1,48 @@
...
+type Config struct {
+	Store StoreConfig `name:"store"`
+
+	ConfigSource string                `name:"config-source" description:"Source of the Device Repository (static, directory, url, blob)"`
+	Static       map[string][]byte     `name:"-"`
+	Directory    string                `name:"directory" description:"OS filesystem directory, which contains Device Repository package"`
+	URL          string                `name:"url" description:"URL, which contains Device Repository package"`
+	Blob         config.BlobPathConfig `name:"blob"`
+
+	AssetsBaseURL string `name:"assets-base-url" description:"The base URL for assets"`
```
htdvisser: For consistency with the other descriptions, you should probably say "for Device Repository assets"

api/devicerepository.proto
```diff
@@ -0,0 +1,364 @@
...
+  // Device videos.
+  Videos videos = 16;
+
+  // Device information page URL.
+  string product_url = 17 [(gogoproto.customname) = "ProductURL"];
+  // Device datasheet URL.
+  string datasheet_url = 18 [(gogoproto.customname) = "DatasheetURL"];
+
+  message Reseller {
+      // Reseller name.
```
htdvisser: Inconsistent indentation

cmd/ttn-lw-stack/commands/config.go
johanstokking: This is not addressed @neoaggelos

cmd/ttn-lw-stack/commands/config.go
```diff
@@ -53,22 +55,24 @@ type Config struct {
 	DTC              devicetemplateconverter.Config    `name:"dtc"`
 	QRG              qrcodegenerator.Config            `name:"qrg"`
 	PBA              packetbrokeragent.Config          `name:"pba"`
+	DeviceRepository devicerepository.Config           `name:"device-repository"`
```
johanstokking: This can be `dr` too

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -127,6 +131,7 @@ var startCommand = &cobra.Command{
 				start.DeviceTemplateConverter = true
 				start.QRCodeGenerator = true
 				start.PacketBrokerAgent = true
+				start.DeviceRepository = true
```
johanstokking, line 134: Also under `startDefault` below

htdvisser (commented):
Generally LGTM.

- Missing CODEOWNERS for `pkg/devicerepository` (I did not review that package)
- Missing validation on some fields in request messages
- For frontend it may be nice to add the new enums to `pkg/ttnpb/i18n.go`

pkg/devicerepository/config.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// StoreConfig represents configuration for the Device Repository store.
+type StoreConfig struct {
+	Store store.Store `name:"-"`
+}
```
johanstokking: 1. Why is this wrapping `store.Store`?
2. Why is this referenced on L26 with a named field, but nothing can and should be configured here?

pkg/devicerepository/devicerepository.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// New returns a new *DeviceRepository.
+func New(c *component.Component, conf *Config) (*DeviceRepository, error) {
+	dr := &DeviceRepository{
+		Component: c,
+		ctx:       log.NewContextWithField(c.Context(), "namespace", "devicerepository"),
+		config:    conf,
+
+		store: conf.Store.Store,
```
johanstokking, line 51: What is the benefit?

pkg/devicerepository/grpc.go
```diff
@@ -0,0 +1,202 @@
...
+// withDefaultBrandFields appends default ttnpb.EndDeviceBrand fields.
+func withDefaultBrandFields(paths []string) []string {
+	return ttnpb.AddFields(paths, "brand_id")
+}
+
+func (dr *DeviceRepository) assetURL(brandID, path string) string {
+	if path == "" || dr.config.AssetsBaseURL == "" {
+		return path
+	}
+	return strings.TrimRight(dr.config.AssetsBaseURL, "/") + "/vendor/" + brandID + "/" + path
```
johanstokking: nit; use `fmt.Sprintf` or even URL building

pkg/ttnpb/field_mask_validation.go
```diff
@@ -338,6 +340,7 @@ var nsEndDeviceReadFieldPaths = [...]string{
 	"version_ids.firmware_version",
 	"version_ids.hardware_version",
 	"version_ids.model_id",
+	"version_ids.band_id",
```
johanstokking: Keep this sorted

pkg/ttnpb/field_mask_validation.go
```diff
@@ -68,6 +69,7 @@ var isEndDeviceWriteFieldPaths = []string{
...
```
johanstokking: Idem

pkg/ttnpb/field_mask_validation.go
```diff
@@ -48,6 +48,7 @@ var isEndDeviceReadFieldPaths = []string{
...
```
johanstokking: Idem

pkg/ttnpb/field_mask_validation.go
```diff
@@ -428,6 +431,7 @@ var RPCFieldMaskPaths = map[string]RPCFieldMaskPathValue{
 			"version_ids.firmware_version",
 			"version_ids.hardware_version",
 			"version_ids.model_id",
+			"version_ids.band_id",
```
johanstokking: Check all occurrences

johanstokking (changes requested):
@rvolosatovs please review NS changes. The `version_ids.band_id` should now be authoritive and `frequency_plan_id` should be in the given band. I'm also okay with an issue in Backlog for this.

It seems to me that this is not what's now in NS: only that the given band ID is valid, and that it's being stored by NS too.

pkg/devicerepository/config.go
neoaggelos: Configuration is added in #3543. The `Store` is there for the unit tests.

pkg/identityserver/store/end_device.go
neoaggelos: Added a relevant check in `pkg/networkserver/grpc_deviceregistry.go`, but I see that https://github.com/TheThingsNetwork/lorawan-stack/pull/3680 is changing everything there. I think it is best to revert (to avoid merge conflicts) add a TODO for this.

pkg/devicerepository/config.go
johanstokking: So there's a store and there are sources, and the two are unrelated? Or the source is used by the store?

And you need to wrap the store only for unit tests?

neoaggelos: The source is used by the store for initialization.

pkg/devicerepository/config.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"go.thethings.network/lorawan-stack/v3/pkg/config"
+	"go.thethings.network/lorawan-stack/v3/pkg/devicerepository/store"
+)
+
+// Config represents the DeviceRepository configuration.
+type Config struct {
+	Store StoreConfig `name:"store"`
+
+	ConfigSource string                `name:"config-source" description:"Source of the Device Repository (static, directory, url, blob)"`
```
htdvisser: I thought I already commented on this (maybe it's in one of the other pull requests), but since we're not violating compatibility commitment by changing unused and undocumented config, I think we can also write `Source` instead of `ConfigSource`, since this is about the source of the device repository, not the source of the config of the device repository.

api/devicerepository.proto
```diff
@@ -0,0 +1,364 @@
...
+  // List of any additional radios included in the device.
+  repeated string additional_radios = 21;
+}
+
+message GetEndDeviceBrandRequest {
+  // Application identifiers.
+  ApplicationIdentifiers application_ids = 1 [(gogoproto.customname) = "ApplicationIDs", (gogoproto.nullable) = false];
+
+  // Brand identifier, as defined in the Device Repository.
+  string brand_id = 2 [(gogoproto.customname) = "BrandID"];
```
htdvisser: missing validation

api/devicerepository.proto
```diff
@@ -0,0 +1,364 @@
...
+  // Page number for pagination. 0 is interpreted as 1.
+  uint32 page = 3;
+
+  // Order (for pagination)
+  string order_by = 4 [(validate.rules).string = {
+    in: ["", "brand_id", "-brand_id", "name", "-name"]
+  }];
+
+  // Search for brands matching a query string.
+  string search = 5;
```
htdvisser: missing validation

api/devicerepository.proto
```diff
@@ -0,0 +1,364 @@
...
+  // Field mask paths.
+  google.protobuf.FieldMask field_mask = 6 [(gogoproto.nullable) = false];
+}
+
+message GetEndDeviceModelRequest {
+  // Application identifiers.
+  ApplicationIdentifiers application_ids = 1 [(gogoproto.customname) = "ApplicationIDs", (gogoproto.nullable) = false];
+
+  // Brand identifier, as defined in the Device Repository.
+  string brand_id = 2 [(gogoproto.customname) = "BrandID"];
```
htdvisser: missing validation

api/devicerepository.proto
```diff
@@ -0,0 +1,364 @@
...
+  // Model identifier, as defined in the Device Repository.
+  string model_id = 3 [(gogoproto.customname) = "ModelID"];
```
htdvisser: missing validation

api/devicerepository.proto
```diff
@@ -0,0 +1,364 @@
...
+  // Field mask paths.
+  google.protobuf.FieldMask field_mask = 4 [(gogoproto.nullable) = false];
+}
+
+message ListEndDeviceModelsRequest {
+  // Application identifiers.
+  ApplicationIdentifiers application_ids = 1 [(gogoproto.customname) = "ApplicationIDs", (gogoproto.nullable) = false];
+
+  // List end devices from a specific brand.
+  string brand_id = 2 [(gogoproto.customname) = "BrandID"];
```
htdvisser: missing validation

api/devicerepository.proto
```diff
@@ -0,0 +1,364 @@
...
+  uint32 limit = 3 [(validate.rules).uint32.lte = 1000];
+  // Page number for pagination. 0 is interpreted as 1.
+  uint32 page = 4;
+  // Order end devices
+  string order_by = 5 [(validate.rules).string = {
+    in: ["", "brand_id", "-brand_id", "model_id", "-model_id", "name", "-name"]
+  }];
+
+  // List end devices matching a query string.
+  string search = 6;
```
htdvisser: missing validation

pkg/devicerepository/config.go
neoaggelos: Yes, this was first suggested in #3543 and fixed there.

johanstokking (approved):
We need validation in the API as @htdvisser said. This can be limited to the request messages (for now). See https://github.com/lorawan-schema/draft-devices/blob/master/schema.json for regexps

htdvisser (approved):

Issue #2813, opened 2020-06-30
Improve form UX


kschiffer:
#### Summary
This umbrella issue keeps track of the progress of various improvements that are necessary to improve the UX of our forms.

#### Why do we need this?
- Users are overwhelmed by the complexity and demanded inputs of the forms
  - Users don't understand
    - what the demanded LoRaWAN terms mean for them in practice
    - where they can find this information in order to fill the form
  - Users are confused about why certain unfamiliar information is (or appears to be) required when creating entities
- There are inconsistencies in how we use
  - Field descriptions
  - Field placeholders
  - Field default values
- The design guidelines form section needs to be updated accordingly

#### What is already there? What do you see now?
- Forms section in our design guidelines
- Glossary items for many field titles
- Use of wizard pattern and accordions for complex forms (End device creation and general settings)

#### What is missing? What do you want to see?
- [ ] Updating design guidelines for forms (and conform webui to it) ⟶  #2664 
- [ ] Use tooltips for field names that reference information from our doc glossary ⟶ #2663
- [ ] Reorganizing complex forms into sections (collapsible accordions) ⟶ #2807 
- [ ] Remove irrelevant fields from entity creation forms ⟶ #2806 
- [ ] Use progressive disclosure to hide settings that are for advanced use cases only ⟶ #2694, #2445 

#### Can you do this yourself and submit a Pull Request?
This is a joined effort. I suggest everyone to put some priority into these issues (if not the case already), since these are the source of a lot of friction in our UX.

johanstokking:
I fully agree, but;

> * Users don't understand
>   
>   * what the demanded LoRaWAN terms mean for them in practice
>   * where they can find this information in order to fill the form
> * Users are confused about why certain unfamiliar information is (or appears to be) required when creating entities

This is going to change drastically once we have the device repository in place. It will be a matter of selecting the vendor, the model, and potentially the region and (if applicable) the firmware version. That's it.

So even though we should make things easy to input, the best thing is to avoid users to input this stuff in the first place.

kschiffer:
Indeed. I've added issues regarding the device repository to the OP. The information of the device repository will play a big role to the overall improvement of the form UX.

johanstokking:
What is the status here?

@kschiffer how far are you with the new screens and can you share ideas here?

kschiffer:
[...]

PR #3538, opened 2020-11-27
Various form layout updates (incl. vertical layout)

### LINKED ISSUE #3529: Retire horizontal form layout and use vertical layout only
#### Summary
As part of the form UX improvements (#2813), we will move towards vertical form layout for all our forms (meaning to display the field label above the field).

#### Why do we need this?
There are a couple of reasons to make this shift:
- Horizontal layout is harder to scan, since there's a high distance between the field label and field input
- Horizontal layout puts a strain on users for having to remember the field title (and purpose) while making an input
- Horizontal layout has some layout inconsistencies/issues for checkboxes, due to the fact that checkboxes usually don't need a field title (they just use the label next to the checkbox)
- Horizontal layout is not as common as vertical layout
- Horizontal layout puts a width limit on labels
- There's also a technical benefit of not having to deal with two different layout modes for fields (as we do now)

Luckily, our forms already allow for vertical layouts (like we currently use in the login form). Right now the horizontal layout is the default.

- Removing the `horizontal` prop flag in our `<Form />` component
- Removing additional styling that corresponds to the horizontal layout
- Some alternative layout solutions for
  - Displaying checkboxes without field title (and use the checkbox label instead)
  - Same for `<Input.Toggled />`
  - Still enabling the use of tooltip icons for these fields too
- Some styling adjustments
  - Form sub titles need to be more distinct from field labels 
  - Overall spacing adjustments
  - Allow for different field widths (e.g. for fields such as "Port number")

I've looked through all of our forms and don't see any big issues that require broader restructure to make this change to the vertical form layout.

This is a matter of some styling and display logic changes. I think the most challenging here is to enable some different stylings in the `<Form.Field />` component without stirring it up too much. All in all, this should not be a big issue though.

These changes should not affect our existing tests, since they'll mostly use semantic selectors, which we won't change. I don't think this change needs any new tests introduced.

Yes.

kschiffer:
#### Summary
Closes #3529 
References #2813 

image

#### Changes
- Remove horizontal layout
  - Remove horizontal props from forms and inputs (except checkbox/radio button groups)
  - Remove styling for horizontal forms
- Replace literal headings with `<Form.SubTitle />`
- Allow fields without title (to be used e.g. by checkboxes)
- Allow extending checkboxes with children prop (to add glossary icon)
- Introduce `inputWidth` prop for some inputs (to easily set various input widths for fields)
- Various related styling fixes and tweaks
- Remove `tabindex` for glossary icons
- Some low-hanging-fruit fixes to some forms
  - Updated messages
  - Updated form structure and field order
- Updated stories

#### Testing

Manual testing as well as existing e2e tests.

##### Regressions

This should not introduce any regressions, since this is mostly styling focused

#### Notes for Reviewers
See #3529 and #2813 for details

#### Checklist


pkg/webui/components/checkbox/checkbox.js
```diff
@@ -171,6 +174,7 @@ class Checkbox extends React.PureComponent {
           <span className={style.checkmark} />
         </span>
         {label && <Message className={style.label} content={label} />}
+        {children}
```
bafonins: Any reason for adding this? Cant find any usage of `<Checkbox />` with `children` being set.

pkg/webui/components/link/index.js
```diff
@@ -135,6 +135,7 @@ const DocLink = function(props) {
     secondary,
     primary,
     disabled,
+    tabIndex,
```
bafonins, line 142: `tabIndex` is not defined in prop-types

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ For details about compatibility between different releases, see the **Commitment
 ### Added
 
 ### Changed
+- Form field layouts in the Console (field labels are now dispayed above the field).
```
bafonins, line 18: new line missing

bafonins (commented):
Looks good 👌

Is the difference between field and submit bar lengths intentional? This is how the application form looks on my screen:
Screenshot 2020-12-08 at 17 31 24

pkg/webui/components/checkbox/checkbox.js
kschiffer: Ah yeah, we'll use this later on to be able to display tooltip icons after a checkbox label. If you want I can remove it here and re-add it later, when we use it.

bafonins: I would prefer adding this together with tooltips

bafonins (approved):

PR #3543, opened 2020-11-29
Device Repository Store Implementation

neoaggelos:
#### Summary

This PR implements the Store functionality of the Device Repository component. Extracted from (and targetting) #3530 so that it can be reviewed separately.

#### Changes

- [X] Implement parsing of the Device Repository files via a `store.Store` interface.
- [X] Add `remoteStore`, which implements the `store.Store` interface (retrieving brands/models/templates and codecs) using a fetcher. Sorting/Searching and pagination are not supported.
- [X] Add a `bleveStore` which implements the `store.Store` interface. Internally, it uses a `remoteStore`, and adds indexing/searching/sorting capabilities on top of it. Will follow-up separately in #3433 regarding the implementation.
- [X] Add a new `ttn-lw-stack dr create-index` command, which generates the index files required by bleve.
- [x] Refresh used index files dynamically (could be as simple as fetch new index zip, unarchive, lock mutex, replace bleve index and unlock index).

#### Testing

- [X] Unit test `remoteStore` (list brands, models, get template, get codecs)
- [ ] Unit test `bleveStore`

#### Notes for Reviewers

This is the fetching/parsing implementation, extracted from #3530 to review separately.

@rvolosatovs End Device Template MAC Settings https://github.com/TheThingsNetwork/lorawan-stack/pull/3543/files#diff-18b497f7fcc8e24c81b6921778b0f89d44ffc163cf356e9a7bd7fe691b026865R274-R423
@htdvisser Store and indexing implementation. 

#### Checklist


johanstokking:
@neoaggelos what do you want me to review at this stage?

neoaggelos:
@johanstokking Please review the following:

- The store interface and the naive implementation (`remoteStore`)
- Configuration
- The process of creating, fetching and refreshing the device repository package zip. Currently using a placeholder personal GitHub repository for the package. This should be part of the https://github.com/TheThingsNetwork/lorawan-devices CI, which should periodically build and update the device repo package. Perhaps the easiest solution would be to build the package and store at an S3 bucket.
- Have a look at the indexed store implementation.

Some comments for the `bleveStore` implementation:
- bleve supports a `bleve.IndexAlias` interface, which allows atomically switching the underlying index being used. This is cleaner as far as bleve is concerned, but does not work well for us, since it makes the logic of handling multiple indexes in the same working directory harder (especially across stack restarts).
- Brands and models are indexed using their ids as keys. Storing the protobuf objects as JSON enables searching via profile/sensors/compliances/brand info/model info/etc. Also see relevant comments in #3433 for this decision.

All parts have been tested extensively, please review while I work on unit tests as well.

Also, @rvolosatovs please comment review the mac settings of the generated end device templates.

pkg/devicerepository/store/remote/schema.go
```diff
@@ -0,0 +1,436 @@
...
+	if p.Rx2DataRateIndex > 0 {
+		dev.MACSettings.Rx2DataRateIndex = &ttnpb.DataRateIndexValue{
+			Value: p.Rx2DataRateIndex,
+		}
+		paths = append(paths, "mac_settings.rx2_data_rate_index")
+	}
+	if p.Rx2Frequency > 0 {
+		dev.MACSettings.Rx2Frequency = &pbtypes.UInt64Value{
+			Value: uint64(p.Rx2Frequency * 100000),
+		}
```
rvolosatovs, line 409: `paths` addition missing here?

pkg/devicerepository/store/remote/schema.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+	if p.Supports32BitFCnt {
+		dev.MACSettings.Supports32BitFCnt = &pbtypes.BoolValue{
+			Value: true,
+		}
+		paths = append(paths, "mac_settings.supports_32_bit_f_cnt")
+	}
+	if fs := p.FactoryPresetFrequencies; fs != nil && len(fs) > 0 {
```
rvolosatovs:
```suggestion
	if fs := p.FactoryPresetFrequencies; len(fs) > 0 {
```

pkg/devicerepository/store/remote/util.go
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+package remote
+
+import "go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+
+// DutyCycleFromFloat converts a float value (0 < dc < 1) to a ttnpb.AggregatedDutyCycle
+// enum value. The enum value is rounded-down to the closest value, which means
+// that dc == 0.3 will return ttnpb.DUTY_CYCLE_4 (== 0.25).
+func DutyCycleFromFloat(dc float64) ttnpb.AggregatedDutyCycle {
```
rvolosatovs: No need to introduce new file just for putting one helper function, which is only used once in the whole codebase

pkg/devicerepository/store/remote/schema.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+	if p.PingSlotFrequency > 0 {
+		dev.MACSettings.PingSlotFrequency = &pbtypes.UInt64Value{
+			Value: uint64(p.PingSlotFrequency * 100000),
+		}
+		paths = append(paths, "mac_settings.ping_slot_frequency")
+	}
+	if p.PingSlotPeriod > 0 {
+		dev.MACSettings.PingSlotPeriodicity = &ttnpb.PingSlotPeriodValue{
+			Value: pingSlotPeriodToPB[p.PingSlotPeriod],
```
rvolosatovs, line 384: Where does this value come from? Using seconds here is not the standard "LoRaWAN way"

rvolosatovs:
Please be more specific about the particular MAC-related functionality you have questions about

pkg/devicerepository/store/remote/schema.go
neoaggelos: For context, The `PingSlotPeriod` value used here is coming from the Device Repository where it is validated against [the following schema](https://lorawan-schema.org/draft/devices/1):

```
    "pingSlotPeriod": {
      "enum": [32, 64, 128, 256, 512, 1024, 2048, 4096],
      "description": "Ping slot period (seconds)"
    },
```

In #3433, we decided that values read by the device repository can be assumed to comply with the schema. Is there any better to go about with this?

neoaggelos:
> Please be more specific about the particular MAC-related functionality you have questions about

Please check if the MAC settings that are added to the end device template will actually work as expected, or if extra configuration is needed for them.

johanstokking:
Looping in @htdvisser on the bleve part

cmd/internal/shared/devicerepository/config.go
```diff
@@ -14,15 +14,28 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/devicerepository/store/bleve"
+)
 
 // DefaultDeviceRepositoryConfig is the default configuration for the Device Repository.
 var DefaultDeviceRepositoryConfig = devicerepository.Config{
 	ConfigSource: "url",
-	URL:          "https://raw.githubusercontent.com/TheThingsNetwork/lorawan-devices/master",
+	// TODO: This is for initial development only. Replace after we decide how the
+	// package is built and where it is stored.
+	URL: "https://raw.githubusercontent.com/neoaggelos/lorawan-devices-index/master",
```
johanstokking: References https://github.com/TheThingsNetwork/lorawan-stack/issues/3433#issuecomment-737078549

cmd/ttn-lw-stack/commands/dr.go
```diff
@@ -0,0 +1,72 @@
...
+			if err := bleve.CreatePackage(ctx, fetch.FromFilesystem(source), source, output, overwrite); err != nil {
+				return err
+			}
+			logger.WithField("path", output).Info("Successfully created index")
+			return nil
+		},
+	}
+	drInitCommand = &cobra.Command{
+		Use:   "init",
+		Short: "Initialize device repository",
```
johanstokking:
```suggestion
		Short: "Initialize Device Repository",
```

johanstokking: From a user's perspective, what is initializing the Device Repository?

pkg/devicerepository/store/bleve/bleve.go
```diff
@@ -0,0 +1,176 @@
...
+			return nil, err
+		}
+	}
+	if err := s.initStore(); err != nil {
+		return nil, err
+	}
+
+	go func() {
+		select {
+		case <-s.ctx.Done():
```
johanstokking: No need for select/case here

pkg/devicerepository/store/bleve/config.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"time"
+)
+
+// Config is configuration for bleveStore.
+type Config struct {
+	WorkingDirectory string `name:"working-directory" description:"Local directory where temporary Device Repository files are stored"`
+
+	AutoInit bool           `name:"auto-init" description:"Initialize device repository automatically when starting the stack"`
+	Refresh  *time.Duration `name:"refresh" description:"Interval for refreshing device repository"`
```
johanstokking:
```suggestion
	RefreshInterval  time.Duration `name:"refresh-interval" description:"Interval for refreshing Device Repository"`
```

pkg/devicerepository/store/bleve/bleve.go
```diff
@@ -0,0 +1,176 @@
...
+			s.modelsIndex.Close()
+			s.brandsIndex.Close()
+		}
+	}()
+
+	if t := c.Refresh; t != nil {
```
johanstokking:
```suggestion
	if d := c.RefreshInterval; d > 0 {
```

You still don't want people to specify `0s`

pkg/devicerepository/store/bleve/bleve.go
```diff
@@ -0,0 +1,176 @@
...
+
+	brandsIndex   bleve.Index
+	brandsIndexMu sync.RWMutex
+	modelsIndex   bleve.Index
+	modelsIndexMu sync.RWMutex
+
+	workingDirectory string
+
+	fetcher   fetch.Interface
+	refreshCh <-chan time.Time
```
johanstokking: Why is this here?

pkg/devicerepository/store/bleve/bleve.go
```diff
@@ -0,0 +1,176 @@
...
+var (
+	errNoWorkingDirectory = errors.DefineInvalidArgument("no_working_directory", "no working directory specified")
+	errNoFetcherConfig    = errors.DefineInvalidArgument("no_fetcher_config", "no index fetcher configuration specified")
+)
+
+// NewStore returns a new device repository store with indexing capabilities (using bleve).
+func (c Config) NewStore(ctx context.Context, f fetch.Interface) (store.Store, error) {
+	if c.WorkingDirectory == "" {
+		return nil, errNoWorkingDirectory.New()
+	}
```
johanstokking: Default to https://golang.org/pkg/os/#Getwd ?

pkg/devicerepository/store/bleve/bleve.go
```diff
@@ -0,0 +1,176 @@
...
+			for {
+				select {
+				case <-s.ctx.Done():
+					return
+				case <-s.refreshCh:
+					logger := log.FromContext(ctx)
+
+					logger.Debug("Refreshing Device Repository")
+					if err := s.initStore(); err != nil {
+						logger.WithError(err).Error("Failed to refresh device repository")
```
johanstokking:
```suggestion
						logger.WithError(err).Error("Failed to refresh Device Repository")
```

Please find/replace case sensitive

pkg/devicerepository/store/bleve/bleve.go
```diff
@@ -0,0 +1,176 @@
...
+	s.storeMu.Lock()
+	defer s.storeMu.Unlock()
+	return (&archiver{}).Unarchive(b, s.workingDirectory)
+}
+
+var (
+	errIndexTimeout = errors.DefineFailedPrecondition("index_timeout", "Timed out while opening index. Make sure the index is not being used by another process")
+)
+
+func (s *bleveStore) openIndexWithTimeout(path string, timeout time.Duration) (bleve.Index, error) {
```
johanstokking: Idiomatic would be for the caller to pass a context and use the deadline.

pkg/devicerepository/store/bleve/bleve.go
```diff
@@ -0,0 +1,176 @@
...
+		return err
+	}
+
+	s.storeMu.Lock()
+	defer s.storeMu.Unlock()
+	return (&archiver{}).Unarchive(b, s.workingDirectory)
+}
+
+var (
+	errIndexTimeout = errors.DefineFailedPrecondition("index_timeout", "Timed out while opening index. Make sure the index is not being used by another process")
```
johanstokking: And return `context.DeadlineExceeded` instead of a new error like here

pkg/devicerepository/store/bleve/store.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	searchRequest := bleve.NewSearchRequest(bleve.NewConjunctionQuery(queries...))
+	searchRequest.Size = int(req.Limit)
+	if req.Page == 0 {
+		req.Page = 1
+	}
+	searchRequest.From = int((req.Page - 1) * req.Limit)
+
+	searchRequest.Fields = []string{"BrandPB"}
```
johanstokking: So this is a full-text search on JSON, or is Bleve smart enough to look at field values?

pkg/devicerepository/store/bleve/packager.go
```diff
@@ -0,0 +1,158 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/devicerepository/store/remote"
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/fetch"
+	"go.thethings.network/lorawan-stack/v3/pkg/jsonpb"
+	"go.thethings.network/lorawan-stack/v3/pkg/log"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+type indexableBrand struct {
+	BrandPB  string // *ttnpb.EndDeviceBrand marshaled into JSON string
```
johanstokking: So is a JSON string, not PB

pkg/devicerepository/store/bleve/store.go
```diff
@@ -0,0 +1,192 @@
...
+	s.brandsIndexMu.RUnlock()
+
+	brands := make([]*ttnpb.EndDeviceBrand, 0, len(result.Hits))
+	for _, hit := range result.Hits {
+		s, ok := hit.Fields["BrandPB"].(string)
+		if !ok {
+			return nil, errCorruptedIndex.New()
+		}
+		brand := &ttnpb.EndDeviceBrand{}
+		if err := jsonpb.TTN().Unmarshal([]byte(s), brand); err != nil {
```
johanstokking: Isn't it much cheaper to keep the original struct in memory? Now you're unmarshaling JSON for each hit, on every search request? With a fancy autocomplete mechanism, that could be dozens per second per user searching for a device.

pkg/devicerepository/store/bleve/store.go
```diff
@@ -0,0 +1,192 @@
...
+	s.modelsIndexMu.RUnlock()
+
+	models := make([]*ttnpb.EndDeviceModel, 0, len(result.Hits))
+	for _, hit := range result.Hits {
+		s, ok := hit.Fields["ModelPB"].(string)
+		if !ok {
+			return nil, errCorruptedIndex.New()
+		}
+		model := &ttnpb.EndDeviceModel{}
+		if err := jsonpb.TTN().Unmarshal([]byte(s), model); err != nil {
```
johanstokking: Idem

pkg/devicerepository/store/bleve/zip.go
```diff
@@ -0,0 +1,88 @@
...
+	"archive/zip"
+	"bytes"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+)
+
+// archiver archives and extracts data from zip archives.
+type archiver struct{}
```
johanstokking: Why is this a struct?

pkg/devicerepository/store/bleve/zip.go
```diff
@@ -0,0 +1,88 @@
...
+func (a *archiver) Unarchive(b []byte, destinationDirectory string) error {
+	rd := bytes.NewReader(b)
+	archive, err := zip.NewReader(rd, rd.Size())
+	if err != nil {
+		return err
+	}
+
+	for _, file := range archive.File {
+		destination := path.Join(destinationDirectory, file.Name)
+		if err := os.MkdirAll(path.Dir(destination), 0755); err != nil {
```
johanstokking: `0644` is probably good enough

pkg/devicerepository/store/remote/remote.go
```diff
@@ -0,0 +1,299 @@
...
+	return &store.GetBrandsResponse{
+		Count:  end - start,
+		Offset: start,
+		Total:  uint32(len(brands)),
+		Brands: brands[start:end],
+	}, nil
+}
+
+var (
+	errUnknownBrand = errors.DefineNotFound("unknown_brand", "unknown brand `{brand_id}`")
```
johanstokking:
```suggestion
	errBrandNotFound = errors.DefineNotFound("brand_not_found", "brand `{brand_id}` not found")
```

pkg/devicerepository/store/remote/remote_test.go
```diff
@@ -0,0 +1,584 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/devicerepository/store/remote"
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/fetch"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	"go.thethings.network/lorawan-stack/v3/pkg/util/test/assertions/should"
+)
+
+var (
+	data = map[string][]byte{
+		"vendor/index.yaml": []byte(`
```
johanstokking: I know we do this elsewhere in the codebase, but please consider making use of a local `testdata` dir which is designed for this in go

pkg/devicerepository/store/remote/schema.go
```diff
@@ -0,0 +1,448 @@
...
+	MaxDutyCycle              float64             `yaml:"maxDutyCycle"`
+	Supports32BitFCnt         bool                `yaml:"supports32bitFCnt"`
+}
+
+var (
+	errUnknownBand = errors.DefineNotFound("unknown_band", "unknown band `{band_id}`")
+	errNoProfile   = errors.DefineNotFound("no_profile", "device does not support region `{region}`")
+
+	errUnknownMACVersion                = errors.DefineNotFound("unknown_mac_version", "unknown LoRaWAN version `{mac_version}`")
+	errUnknownRegionalParametersVersion = errors.DefineNotFound("unknown_regional_parameters_version", "unknown Regional Parameters version `{phyVersion}`")
```
johanstokking:
```suggestion
	errUnknownRegionalParametersVersion = errors.DefineNotFound("unknown_regional_parameters_version", "unknown Regional Parameters version `{phy_version}`")
```

pkg/devicerepository/store/remote/schema.go
```diff
@@ -0,0 +1,448 @@
...
+)
+
+// ToTemplatePB returns a ttnpb.EndDeviceTemplate from an end device profile.
+func (p EndDeviceProfile) ToTemplatePB(ids *ttnpb.EndDeviceVersionIdentifiers, info *ttnpb.EndDeviceModel_FirmwareVersion_Profile) (*ttnpb.EndDeviceTemplate, error) {
+	phyVersion, ok := regionalParametersToPB[p.RegionalParametersVersion]
+	if !ok {
+		return nil, errUnknownRegionalParametersVersion.WithAttributes("phyVersion", p.RegionalParametersVersion)
```
johanstokking: Here also

pkg/devicerepository/store/remote/schema.go
johanstokking: You can remove `Unknown` and `unknown_` here

pkg/devicerepository/store/remote/schema.go
johanstokking: For more context, this comes from the LoRaWAN Backend Interfaces' Device Profile

pkg/devicerepository/store/bleve/store.go
neoaggelos: Quoting https://github.com/TheThingsNetwork/lorawan-stack/issues/3433#issuecomment-734128445:

> ... it is not trivial to load them back from the index. E.g. for an ttnpb.EndDevice object, we would have to retrieve every single nested field individually, and then recreate the original proto, which I would rather avoid....

This means that, instead of `BrandPB`, we would need to individually load all fields as an interface first, then manually set in the result proto. This seems very tedious to me.

For reference, here is what the fields for an end device model are (NOTE: These contain all key names, including `EU_863_870`, even though they are not part of the proto).

```
$ ./bleve check modelsIndex.bleve
checking fields: [ModelPB.brand_id ModelPB.model_id ModelPB.name ModelPB.description ModelPB.firmware_versions.version ModelPB.firmware_versions.numeric ModelPB.firmware_versions.profiles.EU_863_870.profile_id ModelPB.firmware_versions.profiles.EU_863_870.lorawan_certified ModelPB.firmware_versions.profiles.EU_863_870.codec_id ModelPB.ip_code ModelPB.key_security ModelPB.hardware_versions.version ModelPB.hardware_versions.numeric ModelPB.product_url ModelPB.datasheet_url BrandID ModelPB.hardware_versions.part_number BrandName ModelID ModelName _all ModelPB.firmware_versions.supported_hardware_versions ModelPB.firmware_versions.profiles.AS_923.profile_id ModelPB.firmware_versions.profiles.AS_923.lorawan_certified ModelPB.firmware_versions.profiles.AS_923.codec_id ModelPB.firmware_versions.profiles.US_902_928.profile_id  (... and lots more ....)]
....
```

Converting to and from JSON is not the most performant option, but improving on it is not trivial. I am OK with creating an issue to deal with this later, and adding a TODO comment instead of blocking on this.

pkg/devicerepository/store/bleve/store.go
neoaggelos: This is only the list of fields that bleve will return for documents that match the query, not the list of fields on which we search. Bleve will search all fields.

pkg/devicerepository/store/remote/enums.go
```diff
@@ -0,0 +1,71 @@
...
+	"RP001-1.0.3-RevA": ttnpb.PHY_V1_0_3_REV_A,
+	// TODO: Add Regional Parameters for LoRaWAN version 1.0.4 (https://github.com/TheThingsNetwork/lorawan-stack/issues/3513)
+	// "RP002-1.0.0": ttnpb.PHY_UNKNOWN,
+	// "RP002-1.0.1": ttnpb.PHY_UNKNOWN,
+	"RP001-1.1":      ttnpb.PHY_V1_1_REV_A,
+	"RP001-1.1-RevB": ttnpb.PHY_V1_1_REV_B,
+}
+
+// pingSlotPeriodToPB maps LoRaWAN schema ping slot period to ttnpb.PingSlotPeriod enum values.
+var pingSlotPeriodToPB = map[uint32]ttnpb.PingSlotPeriod{
```
rvolosatovs, line 54: This is invalid. See https://github.com/lorawan-schema/draft-devices/pull/9

rvolosatovs (changes requested):

pkg/devicerepository/store/remote/schema.go
rvolosatovs: The spec just specifies the value type as "number", are you sure the value is meant to be in seconds and not the exponent?

pkg/devicerepository/store/remote/schema.go
```diff
@@ -0,0 +1,460 @@
...
+			Value: dutyCycleFromFloat(dc),
+		}
+		paths = append(paths, "mac_settings.desired_max_duty_cycle")
+	}
+
+	dev.MACState = &ttnpb.MACState{
+		DesiredParameters: ttnpb.MACParameters{},
+	}
+	if p.MaxEIRP > 0 {
+		dev.MACState.DesiredParameters.MaxEIRP = p.MaxEIRP
```
rvolosatovs, line 436: This does not persist between sessions - currently NS uses either FP or band defaults, @johanstokking should we add a MAC setting for it?

pkg/devicerepository/store/remote/schema.go
```diff
@@ -0,0 +1,460 @@
...
+		t := time.Duration(p.ClassBTimeout) * time.Second
+		dev.MACSettings.ClassBTimeout = &t
+		paths = append(paths, "mac_settings.class_b_timeout")
+	}
+	if p.ClassCTimeout > 0 {
+		t := time.Duration(p.ClassCTimeout) * time.Second
+		dev.MACSettings.ClassCTimeout = &t
+		paths = append(paths, "mac_settings.class_c_timeout")
+	}
+	if p.PingSlotDataRateIndex > 0 {
```
rvolosatovs: `0` is a valid value, also note that this and all other `ping.*` parameters should always be set when `SupportsClassB` is `true` according to LoRaWAN Backend interfaces

pkg/devicerepository/store/remote/schema.go
```diff
@@ -0,0 +1,460 @@
...
+		}
+		paths = append(paths, "mac_settings.ping_slot_frequency")
+	}
+	if p.PingSlotPeriod > 0 {
+		dev.MACSettings.PingSlotPeriodicity = &ttnpb.PingSlotPeriodValue{
+			Value: pingSlotPeriodToPB[p.PingSlotPeriod],
+		}
+		paths = append(paths, "mac_settings.ping_slot_periodicity")
+	}
+	if p.Rx1Delay > 0 {
```
rvolosatovs: `0` (mapping to `1`) is a valid value I suppose @johanstokking ?

pkg/devicerepository/store/remote/schema.go
```diff
@@ -0,0 +1,460 @@
...
+		dev.MACSettings.Rx1Delay = &ttnpb.RxDelayValue{
+			Value: p.Rx1Delay,
+		}
+		paths = append(paths, "mac_settings.rx1_delay")
+	}
+	if p.Rx1DataRateOffset > 0 {
```
rvolosatovs: `0` is a valid value

pkg/devicerepository/store/remote/schema.go
```diff
@@ -0,0 +1,460 @@
...
+		dev.MACSettings.Rx1DataRateOffset = &pbtypes.UInt32Value{
+			Value: p.Rx1DataRateOffset,
+		}
+		paths = append(paths, "mac_settings.rx1_data_rate_offset")
+	}
+	if p.Rx2DataRateIndex > 0 {
```
rvolosatovs: `0` is a valid value

pkg/devicerepository/store/remote/schema.go
```diff
@@ -0,0 +1,460 @@
...
+	}
+	if fs := p.FactoryPresetFrequencies; len(fs) > 0 {
+		dev.MACSettings.FactoryPresetFrequencies = make([]uint64, 0, len(fs))
+		for _, freq := range fs {
+			dev.MACSettings.FactoryPresetFrequencies = append(dev.MACSettings.FactoryPresetFrequencies, uint64(freq*100000))
+		}
+		paths = append(paths, "mac_settings.factory_preset_frequencies")
+	}
+	if dc := p.MaxDutyCycle; dc > 0 {
+		dev.MACSettings.DesiredMaxDutyCycle = &ttnpb.AggregatedDutyCycleValue{
```
rvolosatovs:
```suggestion
		dev.MACSettings.MaxDutyCycle = &ttnpb.AggregatedDutyCycleValue{
```

pkg/devicerepository/store/remote/enums.go
johanstokking: Fixed this in published schema

pkg/devicerepository/store/remote/schema.go
johanstokking: That validation is also enforced by the Device Repository so you can rely on the presence of those values when `SupportsClassB` is set

pkg/devicerepository/store/remote/schema.go
johanstokking: Yes, that is a valid value that defaults to 1

pkg/devicerepository/store/bleve/zip.go
neoaggelos: This is a directory, hence `0755`. It is impossible to create sub-directories otherwise.

cmd/ttn-lw-stack/commands/dr.go
```diff
@@ -0,0 +1,42 @@
...
+
+package commands
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var (
+	drCommand = &cobra.Command{
+		Use:   "dr",
```
johanstokking: Is `dr-db` more consistent, with `is-db` and `ns-db`?

pkg/devicerepository/grpc.go
```diff
@@ -50,8 +52,8 @@ func (dr *DeviceRepository) ListBrands(ctx context.Context, req *ttnpb.ListEndDe
 	if err := rights.RequireApplication(ctx, req.ApplicationIDs, ttnpb.RIGHT_APPLICATION_DEVICES_READ); err != nil {
 		return nil, err
 	}
-	if req.Limit == 0 {
-		req.Limit = 1000
+	if req.Limit > defaultLimit || req.Limit <= 0 {
```
johanstokking: Nit; `Limit` is unsigned

```suggestion
	if req.Limit > defaultLimit || req.Limit == 0 {
```

pkg/devicerepository/grpc.go
```diff
@@ -102,8 +104,8 @@ func (dr *DeviceRepository) ListModels(ctx context.Context, req *ttnpb.ListEndDe
...
```
johanstokking: idem

pkg/devicerepository/store/bleve/bleve.go
```diff
@@ -0,0 +1,95 @@
...
+	"time"
+
+	"github.com/blevesearch/bleve"
+	"go.thethings.network/lorawan-stack/v3/pkg/devicerepository/store"
+	"go.thethings.network/lorawan-stack/v3/pkg/devicerepository/store/remote"
+	"go.thethings.network/lorawan-stack/v3/pkg/fetch"
+)
+
+// defaultTimeout is the timeout while trying to open the index. This is to avoid
+// blocking on the index open call, which will hung indefinetetly if the index
```
johanstokking:
```suggestion
// blocking on the index open call, which hangs indefinitely if the index
```

pkg/devicerepository/store/bleve/bleve.go
```diff
@@ -0,0 +1,95 @@
...
+	defer cancel()
+	s.modelsIndex, err = openIndex(ctx, path.Join(wd, modelsIndexPath))
+	if err != nil {
+		return nil, err
+	}
+	go func() {
+		<-s.ctx.Done()
+		s.modelsIndex.Close()
+		s.brandsIndex.Close()
+	}()
```
johanstokking: Would it be cleaner to have a `Close() error` on the store instead?

pkg/devicerepository/store/bleve/bleve_test.go
```diff
@@ -0,0 +1,392 @@
...
+					a.So(err, should.BeNil)
+				}
+				a.So(result, should.Resemble, tc.result)
+			})
+		}
+	})
+
+	t.Run("Template", func(t *testing.T) {
+
+	})
```
johanstokking: TODO or what?

pkg/devicerepository/store/bleve/create.go
```diff
@@ -0,0 +1,186 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/fetch"
+	"go.thethings.network/lorawan-stack/v3/pkg/jsonpb"
+	"go.thethings.network/lorawan-stack/v3/pkg/log"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+type indexableBrand struct {
+	BrandPB  string // *ttnpb.EndDeviceBrand marshaled into JSON string
+	ModelsPB string // []*ttnpb.EndDeviceModel marshaled into JSON string
```
johanstokking: Maybe I missed it, but we are doing full-text search on JSON strings?

pkg/devicerepository/store/bleve/create.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// Initialize fetches the Device Repository package file and generates index files.
+func (c Config) Initialize(ctx context.Context, fetcher fetch.Interface, overwrite bool) error {
+	wd, err := getWorkingDirectory(c.SearchPaths)
+	if err != nil {
+		return err
+	}
+
+	b, err := fetcher.File("master.zip")
```
johanstokking: Wasn't it `package.zip`?

pkg/devicerepository/store/bleve/create.go
```diff
@@ -0,0 +1,186 @@
...
+		Paths: ttnpb.EndDeviceBrandFieldPathsNested,
+	})
+	if err != nil {
+		return err
+	}
+
+	brandsBatch := brandsIndex.NewBatch()
+	modelsBatch := modelsIndex.NewBatch()
+	for _, brand := range brands.Brands {
+		log.FromContext(ctx).WithField("brand_id", brand.BrandID).Debug("Indexing")
```
johanstokking: Imperative mood in log messages. And you probably want a `logger` variable before the loop

pkg/devicerepository/store/bleve/create.go
```diff
@@ -0,0 +1,186 @@
...
+		models, err := s.GetModels(store.GetModelsRequest{
+			Paths:   ttnpb.EndDeviceModelFieldPathsNested,
+			BrandID: brand.BrandID,
+		})
+		if errors.IsNotFound(err) {
+			// Skip vendors without any models
+			continue
+		} else if err != nil {
+			return err
+		}
```
johanstokking:
```suggestion
		if err != nil {
			if errors.IsNotFound(err) {
				// Skip vendors without any models
				continue
			}
			return err
		}
```

Is faster and looks better

pkg/devicerepository/store/bleve/store.go
johanstokking: I don't get it really. You are searching through structs, right? There can be a pointer in there to an already unmarshaled entity, right?

tools/mage/dev.go
```diff
@@ -228,6 +228,9 @@ func (Dev) InitStack() error {
 	if err := runGo("./cmd/ttn-lw-stack", "is-db", "init"); err != nil {
 		return err
 	}
+	if err := runGo("./cmd/ttn-lw-stack", "dr", "init"); err != nil {
```
johanstokking: See the inconsistency

tools/mage/header.yml
```diff
@@ -1,6 +1,7 @@
 rules:
 - include: [.go,.js,.make,Makefile,.styl,.proto]
-  exclude: [.snap,.json,.goreleaser.release.yml,.goreleaser.snapshot.yml]
+  exclude: [.snap,.json,.goreleaser.release.yml,.goreleaser.snapshot.yml,
+    pkg/devicerepository/store/remote/testdata/vendor/foo-vendor]
```
johanstokking: Can we glob match `testdata` under `pkg` here?

cmd/ttn-lw-stack/commands/dr.go
neoaggelos: Yes, it is more consistent with the rest of the commands, but there is no database here.

pkg/devicerepository/store/bleve/create.go
neoaggelos: Yes, indeed. In particular, the `ModelsPB` field here will only be used for searching brands based on information for their models. (e.g. searching for "The things uno" will return the `the-things-products` brand).

pkg/devicerepository/store/bleve/store.go
neoaggelos: Well, it does not seem possible to store an actual struct (e.g. `*ttnpb.EndDeviceBrand`) and load it back from the index. We could instead add an in-memory cache for this, to avoid unmarshaling JSON repeatedly?

tools/mage/header.yml
neoaggelos: Under the hood this is just a `strings.Contains()` check. Perhaps `testdata` would be enough?

pkg/devicerepository/store/bleve/create.go
neoaggelos: Changed to simply retrieve the latest master branch from https://github.com/TheThingsNetwork/lorawan-devices

tools/mage/header.yml
johanstokking: I'm fine with that too; the point is to have a pattern here to exclude `testdata`.

cmd/ttn-lw-stack/commands/dr.go
johanstokking: From a user's perspective, there is a database.

api/devicerepository.proto
```diff
@@ -0,0 +1,366 @@
...
+  message Videos {
+    // Link to main device video.
+    string main = 1;
+    // Links to other device videos.
+    repeated string other = 2;
+  }
+  // Device video.
+  string video = 16;
+  // Device videos.
+  Videos videos = 17;
```
johanstokking: I still don't get `video` field 16

api/devicerepository.proto
```diff
@@ -0,0 +1,366 @@
...
+  message Reseller {
+      // Reseller name.
+      string name = 1;
+      // Reseller regions.
+      repeated string region = 2;
+      // Reseller URL.
+      string url = 3 [(gogoproto.customname) = "URL"];
+  }
+  // Reseller URLs.
+  repeated Reseller resellers = 20 [(gogoproto.customname) = "Resellers"];
```
johanstokking: It looks like this branch isn't correctly rebased. Please do so and re-request review. Also please check the files here before doing so, to verify that the diff is good for review.

api/devicerepository.proto
neoaggelos: It seems to be part of the example windsensor device, but indeed I do not see the `video` field to be a part of the schema itself. 

https://github.com/TheThingsNetwork/lorawan-devices/blob/master/vendor/example/windsensor.yaml#L119

Will remove, but we need to fix the example as well.

neoaggelos:
Regarding the constant marshaling/unmarshaling of protos, given that the bleve index does not allow loading back structs in a single call, I think it it probably simpler to avoid storing JSON strings on the index completely, and instead:

- Use the index to retrieve the brand ID (or model ID) only.
- Load the brand/model information from the underlying store using its ID.
- use a gcache for `*ttnpb.EndDeviceBrand`s and `*ttnpb.EndDeviceModel`s.

api/devicerepository.proto
neoaggelos: https://github.com/TheThingsNetwork/lorawan-devices/pull/53

Dockerfile
```diff
@@ -6,6 +6,8 @@ RUN chmod 755 /bin/ttn-lw-stack
 COPY ttn-lw-cli /bin/ttn-lw-cli
 RUN chmod 755 /bin/ttn-lw-cli
 
+RUN mkdir -p /srv/ttn-lorawan/device-repository && /bin/ttn-lw-stack dr-db init
```
htdvisser: I don't know if this is a good idea, as every (re-)build of the Docker image will result in a potentially different device repository.

If we go this way I think it would be better to have the device repository as a git submodule of the lorawan-stack repository, and build the index from that. That way we know exactly which commit of the device repository will end up in our images.

We'd do something similar for the frequency plans repository and webhook template repository (in a separate pull request).

neoaggelos: I would rather we avoid making significant changes to the repository and the release process in this particular PR. The Dockerfile is pretty much tied with our release workflow either way. In that sense, I think it is better to do something like this for now:

Add a CI step for fetching the device-repository archive from GitHub, and then use that for the Dockerfile, e.g.

```
ADD device-repository/ /tmp/device-repository
RUN mkdir -p /srv/ttn-lorawan/device-repository && /bin/ttn-lw-stack dr-db init --dr.config-source=directory --dr.directory=/tmp/device-repository
```

And create a separate issue for handling the frequency plans, webhook templates and the device repository repositories.

htdvisser: Opened a PR: https://github.com/TheThingsNetwork/lorawan-stack/pull/3671

neoaggelos: Updated configuration and Dockerfile accordingly, assuming that `data/lorawan-devices` is a submodule for the Device Repository

johanstokking: What is the verdict?

pkg/devicerepository/config.go
```diff
@@ -19,30 +19,45 @@ import (
...
 func (c Config) NewStore(ctx context.Context, blobConf config.BlobConfig) (store.Store, error) {
 	if c.Store.Store != nil {
 		return c.Store.Store, nil
 	}
-	return &store.NoopStore{}, nil
+
+	return c.Store.Bleve.NewStore(ctx)
+}
+
+var errFetcherConfig = errors.DefineInvalidArgument("fetcher_config", "only the directory configuration source is supported")
```
johanstokking: This is going to backfire badly if we forget to update this message. Please harmonize with error messages of similar cases.

johanstokking (commented):
I don't see any bad mistakes at this point.

Why is this still draft?

Dockerfile
johanstokking: @neoaggelos

neoaggelos: With https://github.com/TheThingsNetwork/lorawan-stack/pull/3671 merged, I will rebase this PR and update the Dockerfile

Dockerfile
```diff
@@ -12,6 +12,11 @@ RUN rm -rf /srv/ttn-lorawan/lorawan-frequency-plans/.git
 COPY data/lorawan-webhook-templates /srv/ttn-lorawan/lorawan-webhook-templates
 RUN rm -rf /srv/ttn-lorawan/lorawan-webhook-templates/.git
 
+COPY data/lorawan-devices /tmp/lorawan-devices
+RUN rm -rf /tmp/lorawan-devices/.git && \
```
htdvisser: This `rm` is not needed, as this dir doesn't end up in the final Docker image

Dockerfile
```diff
@@ -28,6 +33,8 @@ COPY public /srv/ttn-lorawan/public
 
 COPY --from=builder /srv/ttn-lorawan/lorawan-frequency-plans /srv/ttn-lorawan/lorawan-frequency-plans
 COPY --from=builder /srv/ttn-lorawan/lorawan-webhook-templates /srv/ttn-lorawan/lorawan-webhook-templates
+COPY --from=builder /srv/ttn-lorawan/device-repository /srv/ttn-lorawan/device-repository
+RUN chown thethings:thethings -R /srv/ttn-lorawan/device-repository
```
htdvisser: Is this necessary? The Things Stack only needs to read these files, right?

Dockerfile
```diff
@@ -12,6 +12,11 @@ RUN rm -rf /srv/ttn-lorawan/lorawan-frequency-plans/.git
 COPY data/lorawan-webhook-templates /srv/ttn-lorawan/lorawan-webhook-templates
 RUN rm -rf /srv/ttn-lorawan/lorawan-webhook-templates/.git
 
+COPY data/lorawan-devices /tmp/lorawan-devices
+RUN rm -rf /tmp/lorawan-devices/.git && \
+  mkdir -p /srv/ttn-lorawan/device-repository && \
```
htdvisser: This dirname may be a bit confusing, maybe we should make clear that it contains the index, not the entire repository.

cmd/internal/shared/devicerepository/config.go
```diff
@@ -14,12 +14,21 @@
...
 // DefaultDeviceRepositoryConfig is the default configuration for the Device Repository.
 var DefaultDeviceRepositoryConfig = devicerepository.Config{
-	ConfigSource: "url",
-	URL:          "https://raw.githubusercontent.com/TheThingsNetwork/lorawan-devices/master",
+	ConfigSource: "directory",
+	Directory:    "data/lorawan-devices",
+
+	Store: devicerepository.StoreConfig{
+		Bleve: bleve.Config{
+			SearchPaths: []string{"device-repository", "/srv/ttn-lorawan/device-repository"},
```
htdvisser: Same story as in my Dockerfile comment: This dirname may be a bit confusing, maybe we should make clear that it contains the index, not the entire repository.

pkg/devicerepository/config.go
```diff
@@ -19,30 +19,45 @@ import (
...
 type Config struct {
 	Store StoreConfig `name:"store"`
 
-	ConfigSource string                `name:"config-source" description:"Source of the Device Repository (static, directory, url, blob)"`
-	Static       map[string][]byte     `name:"-"`
-	Directory    string                `name:"directory" description:"OS filesystem directory, which contains Device Repository package"`
-	URL          string                `name:"url" description:"URL, which contains Device Repository package"`
-	Blob         config.BlobPathConfig `name:"blob"`
+	ConfigSource string `name:"config-source" description:"Source of the Device Repository (directory)"`
+	Directory    string `name:"directory" description:"OS filesystem directory, which contains Device Repository package"`
```
htdvisser: 1. As I commented in the other PR, I don't think we're breaking compatibility if we make changes, as this wasn't ever used or documented. Therefore I think we can rename this to `Source` (because it's the device repository source, not the config source).
2. Maybe "Device Repository package" should be "Device Repository source"

Dockerfile
neoaggelos: It is necessary for bleve.

.gitignore
```diff
@@ -92,6 +92,7 @@ node_modules/
 /public
 /stories
 /pkg/webui/locales/.backend
+/device-repository-index
```
htdvisser: Since we already have `/data/lorawan-devices`, would it make sense to use `/data/lorawan-devices-index`?

Dockerfile
```diff
@@ -12,6 +12,10 @@ RUN rm -rf /srv/ttn-lorawan/lorawan-frequency-plans/.git
 COPY data/lorawan-webhook-templates /srv/ttn-lorawan/lorawan-webhook-templates
 RUN rm -rf /srv/ttn-lorawan/lorawan-webhook-templates/.git
 
+COPY data/lorawan-devices /tmp/lorawan-devices
+RUN mkdir -p /srv/ttn-lorawan/device-repository-index && \
```
htdvisser: For consistency with `lorawan-devices`, can we use `lorawan-devices-index`?

cmd/internal/shared/devicerepository/config.go
```diff
@@ -14,12 +14,21 @@
...
 // DefaultDeviceRepositoryConfig is the default configuration for the Device Repository.
 var DefaultDeviceRepositoryConfig = devicerepository.Config{
-	ConfigSource: "url",
-	URL:          "https://raw.githubusercontent.com/TheThingsNetwork/lorawan-devices/master",
+	Source:    "directory",
+	Directory: "data/lorawan-devices",
+
+	Store: devicerepository.StoreConfig{
+		Bleve: bleve.Config{
+			SearchPaths: []string{"device-repository-index", "/srv/ttn-lorawan/device-repository-index"},
```
htdvisser: For consistency with `lorawan-devices` in `/data`, can we check in `data` and use `lorawan-devices-index`?

.gitignore
```diff
@@ -92,6 +92,7 @@ node_modules/
 /public
 /stories
 /pkg/webui/locales/.backend
+/lorawan-devices-index
```
htdvisser: I thought I suggested `/data/lorawan-devices-index`? Why not do that?

htdvisser (approved):
Small thing, rest lgtm.

Issue #3546, opened 2020-11-30
Support LBS Timesync message


KrishnaIyer:
#### Summary

Support [LBS Timesync](https://doc.sm.tc/station/tcproto.html#time-synchronization) message

Refs #3515 

#### Why do we need this?

To allow more accurate time sync for LBS gateways with PPS or GPS based sync capabilities.

#### What is already there? What do you see now?

`timesync` message not supported.

#### What is missing? What do you want to see?

- Support downstream `timesync` as a response to upstream `timesync`
- Send timesync periodically (ex: every 30 mins).

#### Environment

`TTS 3.10.2`

#### How do you propose to implement this?

- Implement the message and figure out what to do with `txtime` and `gpstime`.
- For the periodic message, use a new ticker.

#### How do you propose to test this?

UT + LBS gateway with a PPS signal.

#### Can you do this yourself and submit a Pull Request?

Yups

PR #3547, opened 2020-11-30
Fix RTT handling for LBS gateways

### LINKED ISSUE #3515: Revisit RTT calculation for LBS gateways
#### Summary

Revisit RTT calculation for LBS gateways

#### Why do we need this?

We use the LBS `RefTime` parameter returned by the gateway to calculate RTT. This may have issues or the way we are using it is incorrect (under certain conditions).

We use the LBS `RefTime` to calculate RTT. It's definition;
> The RefTime field is calculated from the last received MuxTime, adjusted by the time interval on the router between the arrival of MuxTime and the sending of RefTime. Since downlink traffic is less frequent than uplink messages, it is likely that a MuxTime may be reused to construct multiple RefTime fields. This means that round-trip measurements are composed of the latency of the last downlink message and the most recent uplink message.

There are 2 (potential) issues using a time managed by the gateway
1. We don't exactly know what happens in case of infrequent downlinks. Specifically this;
> This means that round-trip measurements are composed of the latency of the last downlink message and the most recent uplink message.

The way I read it, the gateway will internally offset these values and return a sane RefTime but I don't really know.

2. If the gateway's internal clock is faulty/out of sync, it's possible we measure very large values of RTT and hence never use this gateway for downlinks.

That we are able to account for this.

#### Environment

`v3.10 and LBS gateways`

Should we perhaps Calculate RTT only for TxAcks and use the server time as reference (like UDP)?

UT + Gateway testing

Yup but let's discuss.

KrishnaIyer:
#### Summary

Closes #3515 

Moved `timesync` message support to #3546 

#### Changes

- Measure RTT only for valid TxAcks.
- Add sanity check for round trip time
- Update tests

#### Testing

UT (testing on gateway ongoing)

##### Regressions

Traffic test on physical gateways should cover regressions.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ For details about compatibility between different releases, see the **Commitment
 
 - Simulated uplinks visibility in webhook messages.
 - Retransmission handling.
+- RTT recording for LBS gateways. The maximum round trip delay for RTT calculation is configurable via `--gs.basic-station.max-valid-round-trip-delay`.
 
 ### Security
 
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -271,7 +271,7 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 	}{
 		{
 			Name:      "basicstation",
-			Formatter: lbslns.NewFormatter(),
+			Formatter: lbslns.NewFormatter(conf.BasicStation.MaxValidRoundTripDelay),
 			listenerConfig: listenerConfig{
 				fallbackFreqPlanID: conf.BasicStation.FallbackFrequencyPlanID,
 				listen:             conf.BasicStation.Listen,
--- a/pkg/gatewayserver/io/ws/lbslns/upstream.go
+++ b/pkg/gatewayserver/io/ws/lbslns/upstream.go
@@ -431,11 +431,13 @@ func (f *lbsLNS) HandleUp(ctx context.Context, raw []byte, ids ttnpb.GatewayIden
 		"upstream_type", typ,
 	))
 
-	recordTime := func(refTime float64, xTime int64, server time.Time) {
-		sec, nsec := math.Modf(refTime)
-		if sec != 0 {
-			ref := time.Unix(int64(sec), int64(nsec*1e9))
-			conn.RecordRTT(server.Sub(ref), server)
+	recordTime := func(recordRTT bool, refTime float64, xTime int64, server time.Time) {
+		if recordRTT {
+			sec, nsec := math.Modf(refTime)
+			if sec != 0 {
+				ref := time.Unix(int64(sec), int64(nsec*1e9))
+				conn.RecordRTT(server.Sub(ref), server)
+			}
 		}
 		conn.SyncWithGatewayConcentrator(
 			// The concentrator timestamp is the 32 LSB.
@@ -444,6 +446,11 @@ func (f *lbsLNS) HandleUp(ctx context.Context, raw []byte, ids ttnpb.GatewayIden
 			// The Basic Station epoch is the 48 LSB.
 			scheduling.ConcentratorTime(time.Duration(xTime&0xFFFFFFFFFF)*time.Microsecond),
 		)
+	}
+
+	getTimeFromFloat64 := func(timeInFloat float64) time.Time {
+		sec, nsec := math.Modf(timeInFloat)
+		return time.Unix(int64(sec), int64(nsec*1e9))
 	}
 
 	switch typ {
@@ -485,7 +492,7 @@ func (f *lbsLNS) HandleUp(ctx context.Context, raw []byte, ids ttnpb.GatewayIden
 			ID: int32(jreq.UpInfo.XTime >> 48),
 		}
 		session.DataMu.Unlock()
-		recordTime(jreq.RefTime, jreq.UpInfo.XTime, receivedAt)
+		recordTime(false, jreq.RefTime, jreq.UpInfo.XTime, receivedAt)
 
 	case TypeUpstreamUplinkDataFrame:
 		var updf UplinkDataFrame
@@ -512,7 +519,7 @@ func (f *lbsLNS) HandleUp(ctx context.Context, raw []byte, ids ttnpb.GatewayIden
 			ID: int32(updf.UpInfo.XTime >> 48),
 		}
 		session.DataMu.Unlock()
-		recordTime(updf.RefTime, updf.UpInfo.XTime, receivedAt)
+		recordTime(false, updf.RefTime, updf.UpInfo.XTime, receivedAt)
 
 	case TypeUpstreamTxConfirmation:
 		var txConf TxConfirmation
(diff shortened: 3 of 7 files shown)


pkg/gatewayserver/gatewayserver.go
```diff
@@ -270,8 +270,9 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
 		listenerConfig listenerConfig
 	}{
 		{
-			Name:      "basicstation",
-			Formatter: lbslns.NewFormatter(),
+			Name: "basicstation",
+			// TODO: Add to config in v3.11 (https://github.com/TheThingsNetwork/lorawan-stack/issues/3548)
+			Formatter: lbslns.NewFormatter(10 * time.Second),
```
johanstokking: Please carefully read the compatibility commitment. This is not necessary

johanstokking (commented):
This is not making the LNS implementation more maintainable.

If we only care about measuring RTTs on downlink, why not just put the `MuxTime` in the downlink message, and observe `RefTime` in TX ack?

Then remove RTT stuff from join-request and uplink data frame and only sync concentrator time there.

KrishnaIyer:
> This is not making the LNS implementation more maintainable.

I've simplified the implementation significantly. 

> If we only care about measuring RTTs on downlink, why not just put the MuxTime in the downlink message, and observe RefTime in TX ack?

Yes this is what I do. The `MuxTime` needs to be stored somewhere to fetch on the next `TxAck` and check. Originally, I put that in the session state. Now I fetch that from the downlink tokens.

> Then remove RTT stuff from join-request and uplink data frame and only sync concentrator time there.

Yes this is also what I do. Please see
https://github.com/TheThingsNetwork/lorawan-stack/blob/7f4a72efe19967f85312dd16a710226a45d11a8a/pkg/gatewayserver/io/ws/lbslns/upstream.go#L444
https://github.com/TheThingsNetwork/lorawan-stack/blob/7f4a72efe19967f85312dd16a710226a45d11a8a/pkg/gatewayserver/io/ws/lbslns/upstream.go#L498
https://github.com/TheThingsNetwork/lorawan-stack/blob/7f4a72efe19967f85312dd16a710226a45d11a8a/pkg/gatewayserver/io/ws/lbslns/upstream.go#L525

pkg/gatewayserver/io/ws/lbslns/upstream.go
```diff
@@ -407,16 +407,25 @@ func (updf *UplinkDataFrame) FromUplinkMessage(up *ttnpb.UplinkMessage, bandID s
...
+	logger := log.FromContext(ctx)
+	sec, nsec := math.Modf(conf.RefTime)
+	refTime := time.Unix(int64(sec), int64(nsec*1e9))
+	var (
+		cids  []string
+		ok    bool
+		txAck *ttnpb.TxAcknowledgment
+		delta time.Duration
+	)
+	if cids, delta, ok = tokens.Get(uint16(conf.Diid), refTime); ok {
```
johanstokking: Downlink tokens work with server time

pkg/gatewayserver/io/ws/lbslns/upstream.go
```diff
@@ -407,16 +407,25 @@ func (updf *UplinkDataFrame) FromUplinkMessage(up *ttnpb.UplinkMessage, bandID s
...
+		txAck = &ttnpb.TxAcknowledgment{
+			CorrelationIDs: cids,
+			Result:         ttnpb.TxAcknowledgment_SUCCESS,
+		}
 	} else {
-		logger := log.FromContext(ctx)
 		logger.WithField("diid", conf.Diid).Debug("Tx acknowledgement either does not correspond to a downlink message or arrived too late")
 	}
-	return &txAck
+	return txAck, delta
```
johanstokking: `delta` is undefined here

pkg/gatewayserver/io/ws/lbslns/upstream.go
```diff
@@ -512,28 +522,35 @@ func (f *lbsLNS) HandleUp(ctx context.Context, raw []byte, ids ttnpb.GatewayIden
...
+
 		session := ws.SessionFromContext(ctx)
 		session.DataMu.Lock()
 		session.Data = State{
 			ID: int32(txConf.XTime >> 48),
 		}
 		session.DataMu.Unlock()
-		recordTime(txConf.RefTime, txConf.XTime, receivedAt)
+
+		if delta > f.maxRoundTripDelay {
```
johanstokking, line 537: The server time minus `RefTime` _is_ the delta.

`RefTime` is in the server time domain as it's sent as `MuxTime`.

Why are downlink tokens used for this?

pkg/gatewayserver/io/ws/lbslns/upstream.go
```diff
@@ -432,6 +441,7 @@ func (f *lbsLNS) HandleUp(ctx context.Context, raw []byte, ids ttnpb.GatewayIden
 	))
 
 	recordTime := func(refTime float64, xTime int64, server time.Time) {
+		// Set refTime to 0 to skip recording RTT
```
johanstokking: I don't really like this but ok

pkg/gatewayserver/io/ws/lbslns/upstream.go
```diff
@@ -431,11 +431,13 @@ func (f *lbsLNS) HandleUp(ctx context.Context, raw []byte, ids ttnpb.GatewayIden
 		"upstream_type", typ,
 	))
 
-	recordTime := func(refTime float64, xTime int64, server time.Time) {
-		sec, nsec := math.Modf(refTime)
-		if sec != 0 {
-			ref := time.Unix(int64(sec), int64(nsec*1e9))
-			conn.RecordRTT(server.Sub(ref), server)
+	recordTime := func(recordRTT bool, refTime float64, xTime int64, server time.Time) {
+		if recordRTT {
```
johanstokking, line 435: 🙏

pkg/gatewayserver/io/ws/lbslns/upstream.go
```diff
@@ -446,6 +448,11 @@ func (f *lbsLNS) HandleUp(ctx context.Context, raw []byte, ids ttnpb.GatewayIden
 		)
 	}
 
+	getTimeFromFloat64 := func(timeInFloat float64) time.Time {
```
johanstokking: Probably declare this in the package with `(time.Time, bool)` returns

pkg/gatewayserver/io/ws/lbslns/upstream.go
```diff
@@ -431,11 +431,13 @@ func (f *lbsLNS) HandleUp(ctx context.Context, raw []byte, ids ttnpb.GatewayIden
...
 	))
 
-	recordTime := func(refTime float64, xTime int64, server time.Time) {
-		sec, nsec := math.Modf(refTime)
-		if sec != 0 {
-			ref := time.Unix(int64(sec), int64(nsec*1e9))
-			conn.RecordRTT(server.Sub(ref), server)
+	recordTime := func(recordRTT bool, refTime float64, xTime int64, server time.Time) {
+		if recordRTT {
+			sec, nsec := math.Modf(refTime)
```
johanstokking: And use here

adriansmares (commented):
Nothing to add.

CHANGELOG.md
```diff
@@ -26,6 +25,9 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Fixed
 
+- RTT recording for LBS gateways.
+    The maximum round trip delay for RTT calculation is configurable via `--gs.basic-station.max-valid-round-trip-delay`.
```
htdvisser: This newline+indent has no effect. Just keep it on one line.

htdvisser (approved):
Config changes LGTM.

johanstokking (approved):

PR #3553, opened 2020-12-01
Format end device entity map in js sdk

bafonins:
#### Summary

It seems that `device-entity-map.json` has wrong formatting.

#### Changes

- Format `device-entity-map.json` after running `node ./sdk/js/util/device-field-mask-mapper.js`

##### Regressions

...

#### Notes for Reviewers

Most probably caused by [this commit](https://github.com/TheThingsNetwork/lorawan-stack/commit/19d53ad5d2ca17422c1f17e409f5acd3462b1c9d#diff-9992b6f0e329a6dc49913c62f6efe2102f7a64f55d4bdab3d47df86292db5e87). We had no issues before because `jsSDK:DeviceFieldMasks` wasnt executed in CI.

#### Checklist

### DIFF
--- a/sdk/js/generated/device-entity-map.json
+++ b/sdk/js/generated/device-entity-map.json
@@ -1,450 +1,1542 @@
 {
-  "application_server_address": [["is", "js"], ["is", "js"]],
-  "attributes": ["is", "is"],
-  "created_at": [["is", "ns"], "read_only"],
-  "description": ["is", "is"],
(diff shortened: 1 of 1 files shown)


kschiffer:
This is a generated file that should not be formatted, or at least it should not affect CI if it is not formatted. Did this cause diffs in CI, or what is the reason to apply this?

PR #3560, opened 2020-12-03
Fix nested side navigation focus

### LINKED ISSUE #3551: Expand nested side navigation entries on focus
#### Summary

Currently, when navigating via keyboard the 

ezgif com-gif-maker

#### Why do we need this?

Focused element must be visible.

Currently, the user can focus on the main entry and press `<space>` to expand the dropdown with nested items. However, if the user continues navigating focused items are hidden.

#### Environment

v3.10.2 `ttn-lw-stack`

I see 2 options:
1. If the user focuses on nested and hidden item in the side navigation, the entry should expand.
2. Set `tabIndex` to `-1` on nested items when the dropdown is collapsed. 

Manually

@asmulko

asmulko:
#### Summary
Closes #3551

#### Changes

- `tabIndex` is set to` -1` on nested items when the dropdown is collapsed

#### Testing

Manually tested

#### Checklist

### DIFF
--- a/pkg/webui/components/navigation/side/item/index.js
+++ b/pkg/webui/components/navigation/side/item/index.js
@@ -42,6 +42,8 @@ export class SideNavigationItem extends React.PureComponent {
     icon: PropTypes.string,
     /** A flag specifying whether the side navigation item is active or not. */
     isActive: PropTypes.bool,
+    /** A flag specifying whether the linkable item is to be tabbable or not. */
+    isLinkTabbable: PropTypes.bool,
     location: PropTypes.location.isRequired,
     /** The path of the linkable side navigation item. */
     path: PropTypes.string,
@@ -55,6 +57,7 @@ export class SideNavigationItem extends React.PureComponent {
     exact: false,
     icon: undefined,
     isActive: false,
+    isLinkTabbable: undefined,
     depth: 0,
     path: undefined,
   }
@@ -86,7 +89,17 @@ export class SideNavigationItem extends React.PureComponent {
   }
 
   render() {
-    const { className, children, title, depth, icon, path, exact, isActive } = this.props
+    const {
+      className,
+      children,
+      title,
+      depth,
+      icon,
+      path,
+      exact,
+      isActive,
+      isLinkTabbable,
+    } = this.props
     const { isExpanded } = this.state
     const { isMinimized, onLeafItemClick } = this.context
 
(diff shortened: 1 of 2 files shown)


kschiffer:
Sorry, but after going through the diff, it became apparent to me that the solution suggested in #3551 using `tabindex` is actually not the correct solution for this. Instead, we should use simple styling to achieve the desired effect:

```diff
diff --git a/pkg/webui/components/navigation/side/list/list.styl b/pkg/webui/components/navigation/side/list/list.styl
index effe245e6..8b850437b 100644
--- a/pkg/webui/components/navigation/side/list/list.styl
+++ b/pkg/webui/components/navigation/side/list/list.styl
@@ -29,6 +29,10 @@
 
   &-nested > li
     height: 0
+    transition: visibility $ad.m, height $ad.m ease-in-out
+
+  &-nested:not(.list-expanded) > li
+    visibility: hidden
 
   &-expanded > li
     height: 3.1rem
```

Using `visibility` we can prevent the item from becoming focused (since invisible elements cannot be focused) while being able to preserve the animation (which wouldn't be possible with `display: none`).

kschiffer (changes requested):
See my comment above.

asmulko:
> Sorry, but after going through the diff, it became apparent to me that the solution suggested in #3551 using `tabindex` is actually not the correct solution for this. Instead, we should use simple styling to achieve the desired effect:
> 
> ```diff
> diff --git a/pkg/webui/components/navigation/side/list/list.styl b/pkg/webui/components/navigation/side/list/list.styl
> index effe245e6..8b850437b 100644
> --- a/pkg/webui/components/navigation/side/list/list.styl
> +++ b/pkg/webui/components/navigation/side/list/list.styl
> @@ -29,6 +29,10 @@
>  
>    &-nested > li
>      height: 0
> +    transition: visibility $ad.m, height $ad.m ease-in-out
> +
> +  &-nested:not(.list-expanded) > li
> +    visibility: hidden
>  
>    &-expanded > li
>      height: 3.1rem
> ```
> 
> Using `visibility` we can prevent the item from becoming focused (since invisible elements cannot be focused) while being able to preserve the animation (which wouldn't be possible with `display: none`).

Agreed, looks way cleaner than the `tabIndex` solution I've implemented. Thank you!

kschiffer (approved):

PR #3595, opened 2020-12-15
Add simple markdown to backend error messages

### LINKED ISSUE #3581: Render backend error messages with simple markown
#### Summary
We should render error messages properly, using a simple markdown engine.

#### Why do we need this?
Error messages from the backend are often using simple markdown syntax. We should render this properly, instead of passing through the raw text message to the user.

`<ErrorNotification />` component

Some processing of the message to decorate it with HTML markup, based on the markdown of the error message.

From what I can see, the only thing that is really used is the \` character, to denote values. So, I believe that integrating an external markdown module just for that is overkill. Instead, we should implement a simple parser that will replace these with `<pre />` tags. But please do a little digging whether \` is really the only thing that is used.

Good to add this to storybook and maybe a simple unit test for the parser.

@asmulko

asmulko:
#### Summary

Closes #3581

#### Changes
- Added simple parser that replaces backend error message `back quotes` with `<pre></pre>` tag
- Added simple unit tests for the parser
- Added [dompurify](https://www.npmjs.com/package/dompurify) dependency
- Added example in storybook
errorWithMarkup

#### Testing

- Unit tests
- Manually tested in storybook

#### Checklist

### DIFF
--- a/cypress/integration/console/shared/collaborators/add.spec.js
+++ b/cypress/integration/console/shared/collaborators/add.spec.js
@@ -96,7 +96,7 @@ describe('Collaborators', () => {
 
       cy.findByTestId('error-notification')
         .should('be.visible')
-        .findByText(`user \`${userNotExist}\` not found`)
+        .findByText(userNotExist)
         .should('be.visible')
       cy.visit(
         `${Cypress.config('consoleRootPath')}/applications/${applicationId}/collaborators/add`,
@@ -124,7 +124,7 @@ describe('Collaborators', () => {
 
       cy.findByTestId('error-notification')
         .should('be.visible')
-        .findByText(`organization \`${orgNotExist}\` not found`)
+        .findByText(orgNotExist)
         .should('be.visible')
       cy.visit(
         `${Cypress.config('consoleRootPath')}/applications/${applicationId}/collaborators/add`,
@@ -169,7 +169,7 @@ describe('Collaborators', () => {
 
       cy.findByTestId('error-notification')
         .should('be.visible')
-        .findByText(`user \`${userNotExist}\` not found`)
+        .findByText(userNotExist)
         .should('be.visible')
       cy.visit(`${Cypress.config('consoleRootPath')}/gateways/${gatewayId}/collaborators/add`)
     })
@@ -195,7 +195,7 @@ describe('Collaborators', () => {
 
       cy.findByTestId('error-notification')
         .should('be.visible')
-        .findByText(`organization \`${orgNotExist}\` not found`)
+        .findByText(orgNotExist)
         .should('be.visible')
       cy.visit(`${Cypress.config('consoleRootPath')}/gateways/${gatewayId}/collaborators/add`)
     })
@@ -240,7 +240,7 @@ describe('Collaborators', () => {
 
       cy.findByTestId('error-notification')
         .should('be.visible')
-        .findByText(`user \`${userNotExist}\` not found`)
+        .findByText(userNotExist)
         .should('be.visible')
       cy.visit(`${Cypress.config('consoleRootPath')}/organizations/${testOrgId}/collaborators/add`)
     })
@@ -253,7 +253,7 @@ describe('Collaborators', () => {
 
       cy.findByTestId('error-notification')
         .should('be.visible')
-        .findByText('account of type `organization` can not collaborate on `organization`')
+        .findByText(/can not collaborate/i)
         .should('be.visible')
       cy.visit(`${Cypress.config('consoleRootPath')}/organizations/${testOrgId}/collaborators/add`)
     })
@@ -266,7 +266,7 @@ describe('Collaborators', () => {
 
       cy.findByTestId('error-notification')
         .should('be.visible')
-        .findByText(`organization \`${orgNotExist}\` not found`)
+        .findByText(orgNotExist)
         .should('be.visible')
       cy.visit(`${Cypress.config('consoleRootPath')}/organizations/${testOrgId}/collaborators/add`)
     })
--- a/pkg/webui/components/notification/index.js
+++ b/pkg/webui/components/notification/index.js
@@ -26,7 +26,7 @@ import Details from './details'
 
 import style from './notification.styl'
 
-const Notification = function({
+const Notification = ({
   content,
   className,
   title,
@@ -41,7 +41,7 @@ const Notification = function({
   actionMessage,
   buttonIcon,
   'data-test-id': dataTestId,
-}) {
+}) => {
   const classname = classnames(style.notification, className, {
     [style.error]: error,
     [style.warning]: warning,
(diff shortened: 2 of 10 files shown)


asmulko:
With failed e2e tests I see that I've misunderstood the implementation of how the errors are generated, will update the PR accordingly

cypress/integration/console/shared/collaborators/add.spec.js
```diff
@@ -96,7 +96,7 @@ describe('Collaborators', () => {
 
       cy.findByTestId('error-notification')
         .should('be.visible')
-        .findByText(`user \`${userNotExist}\` not found`)
+        .findByText(`${userNotExist}`)
```
kschiffer:
```suggestion
        .findByText(userNotExist)
```

pkg/webui/lib/replace-back-quotes.js
```diff
@@ -0,0 +1,27 @@
...
+ * Returns a string with backquotes replaced with `<pre></pre>` tag.
+ *
+ * @param {string} value - The string value that needs backquotes replaced.
+ * @returns {string} The string value without all backquotes substituted by
+ * <pre></pre> tag.
+ */
+const replaceBackQuotes = value =>
+  value.replace(/`([^`[]+)`/g, (_, item) => {
+    return `<pre>${item}</pre>`
+  })
```
kschiffer: Since we have only a simple case with back quotes. I think instead of transforming into a string, you could create a small utility component, or [add a small package](https://github.com/iansinnott/react-string-replace) that returns proper nodes. That way, we can save the whole `dangerouslySetInnerHTML` and sanitization.

pkg/webui/lib/components/message/index.js
```diff
@@ -31,13 +39,25 @@ const renderContent = (content, component, props) => {
...
+            }}
+          />
+        ) : (
+          content
+        )}
+      </Component>
+    )
   }
 
   return content
```
kschiffer, line 38: The plain `content` return needs the check as well.

pkg/webui/lib/components/message/index.js
```diff
@@ -14,13 +14,21 @@
...
+import replaceBackQuotes from '@ttn-lw/lib/replace-back-quotes'
 
 import style from './message.styl'
 
-const renderContent = (content, component, props) => {
+const renderHtmlContent = content => {
+  const contentWithHtml = content.map(subContent => subContent.map(replaceBackQuotes))
+
+  return sanitize(contentWithHtml)
+}
```
kschiffer: Instead of doing this, you could use `import { useIntl } from 'react-intl'` to convert the message into a translated string programmatically. Then you could use a string-string-replace logic (see below) to convert that into proper react nodes, which you could then return.

kschiffer (changes requested):
Trickier than I thought. See my remark on using a direct replacement into react nodes (instead of HTML strings). I think this would simplify the solution quite a bit.

pkg/webui/lib/replace-back-quotes.js
kschiffer: Also, contrary to my initial suggestion, please use `<code></code>` instead, since this needs to be inline.

asmulko: It does simplify the solution a lot, thanks for your comments! Let me know if anything else needs refactoring

pkg/webui/components/notification/index.js
```diff
@@ -65,7 +65,7 @@ const Notification = function({
         <div className={style.content}>
           {title && <Message className={style.title} content={title} component="h4" />}
           <div>
-            <Message content={content} values={messageValues} />
+            <Message content={content} values={messageValues} hasMarkdown={Boolean(error)} />
```
kschiffer:
```suggestion
            <Message content={content} values={messageValues} convertBackticks={Boolean(error)} />
```

kschiffer: Let's call this `convertBackticks`, since we're not really parsing any markdown other than converting backticks.

pkg/webui/lib/components/message/index.js
```diff
@@ -70,6 +74,13 @@ const Message = function({
     return renderContent(content, component, rest)
   }
 
+  if (hasMarkdown) {
+    const contentWithMarkdown = formatMessage(content, vals)
+    return reactStringReplace(contentWithMarkdown, /`([^`[]+)`/g, (match, i) => (
+      <code key={i}>{match}</code>
+    ))
+  }
```
kschiffer:
```suggestion
  if (hasMarkdown) {
    const contentWithMarkdown = formatMessage(content, vals)
    return renderContent(reactStringReplace(contentWithMarkdown, /`([^`[]+)`/g, (match, i) => (
      <code key={i}>{match}</code>
    )), component, rest)
  }
```

This should work too, no? Otherwise, it's not possible to use the `component` or `className` prop.

pkg/webui/lib/components/error-message/index.js
```diff
@@ -19,7 +19,7 @@ import Message from '@ttn-lw/lib/components/message'
 import PropTypes from '@ttn-lw/lib/prop-types'
 import { hasCauses, getBackendErrorRootCause, toMessageProps } from '@ttn-lw/lib/errors/utils'
 
-const ErrorMessage = function({ content, withRootCause, className, ...rest }) {
+const ErrorMessage = ({ content, withRootCause, className, ...rest }) => {
   const baseProps = {
```
kschiffer, line 28: Please also add a check here whether the content is a backend error message and if yes, apply the `convertBackticks` props to the `<Message />` as well.

kschiffer (changes requested):
Please also add the following styling to `pkg/webui/styles/main.styl`

```stylus
code
  font-family: $font-family-mono
  font-size: $fs.s
  background-color: #ececec // Save this value as new variable $c-text-backdrop
  padding: 0.1rem $cs.xxs
  color: $tc-black
```

pkg/webui/lib/components/message/index.js
asmulko: Good catch, thanks!

pkg/webui/lib/components/error-message/index.js
```diff
@@ -17,12 +17,18 @@ import React from 'react'
...
+  toMessageProps,
+  isBackend,
+} from '@ttn-lw/lib/errors/utils'
 
-const ErrorMessage = function({ content, withRootCause, className, ...rest }) {
+const ErrorMessage = ({ content, withRootCause, className, ...rest }) => {
   const baseProps = {
     className,
     firstToUpper: true,
+    convertBackticks: Boolean(isBackend(content)),
```
kschiffer, line 31: We should be safe to assume that functions that start with `is` return a boolean. If that's not the case then the utility function needs to be fixed.

pkg/webui/lib/components/message/index.js
```diff
@@ -95,6 +110,10 @@ Message.propTypes = {
    * passed through without any modifications.
    */
   content: PropTypes.message.isRequired,
+  /**
+   * Flag specifying whether the the backticks should be converted to <code></code> tag.
```
kschiffer:
```suggestion
   * Flag specifying whether the the backticks should be converted to `<code/>` tag.
```

kschiffer (approved):

Issue #3569, opened 2020-12-08
Network Server database migration tool


rvolosatovs:
#### Summary

Network Server database migration tool

#### Why do we need this?

To really use optimizations added in 3.10

#### What is already there? What do you see now?

Fallback to legacy behavior

#### What is missing? What do you want to see?

A tool for migration

#### How do you propose to implement this?

Loop over keys and migrate

#### How do you propose to test this?

Setup 3.10, run migration, update to 3.11 and ensure everything still works

#### Can you do this yourself and submit a Pull Request?

yes

rvolosatovs:
Blocking 3.11

Issue #3552, opened 2020-11-30
Fix handling of duplicates and devices changing matching categories


rvolosatovs:
#### Summary

Fix handling of duplicates and devices changing matching categories.
In current implementation matching is broken for devices changing categories while matching. E.g. device moving from pending session to current.

#### Steps to Reproduce

1. Register an OTAA device
2. Let the device join
3. Send first uplink in session via multiple gateways

#### What do you see now?

Observe that only the first uplink is actually processed, duplicates are dropped

#### What do you want to see instead?

See all duplicates properly handled

#### How do you propose to implement this?

Rework some matching features, it's already in-progress for a while, you'll see in PR

#### How do you propose to test this?

Repeat reproduction steps and observe desired result

#### Can you do this yourself and submit a Pull Request?

yes

Issue #3254, opened 2020-09-18
Avoid matching pending session and FCnt reset when uplink ACK is set in storage backend


rvolosatovs:
#### Summary

Do not fetch devices, which cannot match due to ACK bit set.

#### Why do we need this?

To improve performance

#### What is already there? What do you see now?

In #2837 huge changes were made related to matching.
Currently NS checks the ACK bit directly when ranging over the returned devices from Redis.

#### What is missing? What do you want to see?

Redis script can handle it itself and avoid fetching these redundant devices.

#### How do you propose to implement this?

Parameterize lua matching script

#### How do you propose to test this?

Integration testing

#### Can you do this yourself and submit a Pull Request?

yes

PR #3603, opened 2020-12-19
Rework Network Server matching

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/3552
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/3569
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/3254

#### Changes

- Simplify the matching algorithm
- Optimize matching
	- Avoid matching pending sessions when ACK bit is set
	- Avoid matching FCnt rollover/reset for devices, which don't support 32-bit FCnt scheme *and* not support resets
- Rework algorithm
	- Workers may now process duplicate entries, each entry is processed at least once
	- Sort pending session matches by join time
	- Sort current session matches by LSB only
- Improve correctness
	- Relevant device states are copied on uplink - after the uplink is received device states may change, but we must consider the "old" state when matching and locking is not favorable here.
- Fix various bugs
- Reorganize store:
	- Now we store optimized msgpack-encoded device views (different whether it is pending or current session)

#### Testing

Extensive integration testing, tested migrations with existing production data set

##### Regressions

Everything uplink and class A-downlink related, since this PR modifies the core feature of the Network Server.

#### Checklist

### DIFF
--- a/pkg/networkserver/grpc_gsns.go
+++ b/pkg/networkserver/grpc_gsns.go
@@ -355,7 +355,7 @@ func (ns *NetworkServer) matchAndHandleDataUplink(ctx context.Context, dev *ttnp
 					}
 				}
 				if nbTrans < 2 || lastAt.IsZero() {
-					log.FromContext(ctx).Debug("Repeated FCnt value, but frame is not a retransmission, skip")
+					log.FromContext(ctx).Debug("Repeated FCnt value, but frame is not a retransmission")
 					return nil, false, nil
 				}
 				maxDelay := maxRetransmissionDelay(dev.MACState.CurrentParameters.Rx1Delay)
@@ -367,7 +367,7 @@ func (ns *NetworkServer) matchAndHandleDataUplink(ctx context.Context, dev *ttnp
 					"retransmission_number", nbTrans,
 				))
 				if delay > maxDelay {
-					log.FromContext(ctx).Warn("Retransmission delay exceeds maximum, skip")
+					log.FromContext(ctx).Warn("Retransmission delay exceeds maximum")
 					return nil, false, nil
 				}
 
@@ -383,7 +383,7 @@ func (ns *NetworkServer) matchAndHandleDataUplink(ctx context.Context, dev *ttnp
 
 	// NOTE: We assume no dwell time if current value unknown.
 	if dev.MACState.LoRaWANVersion.IgnoreUplinksExceedingLengthLimit() && len(up.RawPayload)-5 > int(dr.MaxMACPayloadSize(dev.MACState.CurrentParameters.UplinkDwellTime.GetValue())) {
-		log.FromContext(ctx).Debug("Uplink length exceeds maximum, skip")
+		log.FromContext(ctx).Debug("Uplink length exceeds maximum")
 		return nil, false, nil
 	}
 
@@ -397,17 +397,17 @@ func (ns *NetworkServer) matchAndHandleDataUplink(ctx context.Context, dev *ttnp
 			session = dev.PendingSession
 		}
 		if session.NwkSEncKey == nil || len(session.NwkSEncKey.Key) == 0 {
-			log.FromContext(ctx).Warn("Device missing NwkSEncKey in registry, skip")
+			log.FromContext(ctx).Warn("Device missing NwkSEncKey in registry")
 			return nil, false, nil
 		}
 		key, err := cryptoutil.UnwrapAES128Key(ctx, session.NwkSEncKey, ns.KeyVault)
 		if err != nil {
-			log.FromContext(ctx).WithField("kek_label", session.NwkSEncKey.KEKLabel).WithError(err).Warn("Failed to unwrap NwkSEncKey, skip")
+			log.FromContext(ctx).WithField("kek_label", session.NwkSEncKey.KEKLabel).WithError(err).Warn("Failed to unwrap NwkSEncKey")
 			return nil, false, nil
 		}
 		cmdBuf, err = crypto.DecryptUplink(key, pld.DevAddr, cmacFMatchResult.FullFCnt, cmdBuf, len(pld.FOpts) > 0)
 		if err != nil {
-			log.FromContext(ctx).WithError(err).Warn("Failed to decrypt uplink, skip")
+			log.FromContext(ctx).WithError(err).Warn("Failed to decrypt uplink")
 			return nil, false, nil
 		}
 	}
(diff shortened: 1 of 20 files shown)


rvolosatovs:
Blocking 3.11 (#3610)

pkg/networkserver/redis/registry.go
```diff
@@ -135,638 +151,324 @@ func (r *DeviceRegistry) GetByEUI(ctx context.Context, joinEUI, devEUI types.EUI
...
-		if err := msgpack.Unmarshal([]byte(v), res); err != nil {
-			return err
+		if s == string(noUplinkMatchMarker) {
+			return errNoUplinkMatch.New()
 		}
-
+		ctx := log.NewContextWithField(ctx, "match_key", matchResultKey)
+		res := &uplinkMatchResult{}
+		if err := unmarshalMsgpack([]byte(s), res); err != nil {
+			panic(s)
```
johanstokking, line 270: This doesn't seem right

pkg/networkserver/grpc_gsns.go
```diff
@@ -209,185 +204,186 @@ func (ns *NetworkServer) matchAndHandleDataUplink(ctx context.Context, dev *ttnp
...
+	type sessionMatchType uint8
+	const (
+		currentOriginalMatch sessionMatchType = iota
+		currentRetransmissionMatch
+		currentResetMatch
+		pendingMatch
+	)
+	var matchType sessionMatchType
+	switch {
+	case !pld.Ack &&
```
johanstokking, line 231: Start adding comments here, i.e. what is this case, the session confirmation?

pkg/networkserver/grpc_gsns.go
```diff
@@ -209,185 +204,186 @@ func (ns *NetworkServer) matchAndHandleDataUplink(ctx context.Context, dev *ttnp
...
-					},
-					CorrelationIDs: append(dev.MACState.PendingApplicationDownlink.CorrelationIDs, up.CorrelationIDs...),
-				},
+			case dev.Session.LastFCntUp == 0 && dev.SupportsJoin && len(dev.MACState.RecentUplinks) == 1,
+				dev.Session.LastFCntUp == 0 && !dev.SupportsJoin && len(dev.MACState.RecentUplinks) == 0:
+
+				ctx = log.NewContextWithField(ctx, "f_cnt_reset", false)
+				matchType = currentOriginalMatch
+
+			default: // cmacFMatchResult.FullFCnt == dev.Session.LastFCntUp
```
johanstokking, line 322: Is the comment implicit?
A comment would help here

rvolosatovs: There are already 2 cases above:
`			case cmacFMatchResult.FullFCnt < dev.Session.LastFCntUp:`
`			case cmacFMatchResult.FullFCnt > dev.Session.LastFCntUp:`
Not sure what you mean and how I can improve this.
Can you give me a suggestion for a comment?
In this case FCnt is repeated and it's not the first uplink in the session

rvolosatovs: You should also notice the `matchType`, which should clarify what's going on

pkg/networkserver/redis/registry.go
rvolosatovs: Debugging artifact

pkg/networkserver/grpc_gsns.go
johanstokking: Mostly code comments would do

rvolosatovs: Sorry, it's not clear to me. What do you mean exactly?
I'd rather not pollute code with redundant comments - if the code is not self-explanatory, then I'd rather fix the code. What is not clear in the implementation exactly?

This is a switch of 3 cases:
- A > B
- A < B
- A == B

I commented the `A==B` part, since that is hidden behind a `default:` statement and the switch is quite long.
What else should be explained?

johanstokking: With the comment on this line, I was asking whether this code comment was a left-over or that it was descriptive.

With adding comments I didn't mean these cases [but these](https://github.com/TheThingsNetwork/lorawan-stack/pull/3603#discussion_r547894219), my bad. I see that you [already put comments there](https://github.com/rvolosatovs/lorawan-stack/blob/fix/matching/pkg/networkserver/grpc_gsns.go#L271-L276) so this thread can be resolved.

johanstokking (approved):
I don't see mistakes, but needs approval from @adriansmares on the Lua scripts

pkg/networkserver/redis/registry.go
```diff
@@ -76,8 +65,42 @@ func (r *DeviceRegistry) Init(ctx context.Context) error {
...
+	var buf bytes.Buffer
+	enc.Reset(&buf)
+	enc.UseCompactFloats(true)
+	enc.UseCompactInts(true)
+	enc.UseCustomStructTag("json")
+
+	err := enc.Encode(v)
+	b := buf.Bytes()
+
+	msgpack.PutEncoder(enc)
```
adriansmares, line 81: Why not `defer` after `GetEncoder` ? I understand that this is close to the original `msgpack.Marshal` but I don't think it hurts to do this.

pkg/redis/generate_scripts.go
```diff
@@ -59,7 +59,7 @@ var (
 )`))
 
 func isCommentLine(s string) bool {
```
adriansmares, line 61: This generator may be removed once we move to `go` `1.16` via file embedding. Consider creating an issue about it.

References https://github.com/golang/go/issues/41191

adriansmares (approved):
Reviewed all files - LGTM. Bellow are some small nits.

htdvisser (commented):
Nothing to add

pkg/networkserver/redis/registry.go
rvolosatovs: This is copy-paste of original implementation and I would prefer to keep in that way. Why exactly would we want to refactor this into using `defer` except for cosmetic pleasure?

adriansmares: `defer` has the advantage of returning the encoder even when a `panic` occurs. But in this particular case I don't expect a panic, so feel free to merge as is (I did approve this).

rvolosatovs: We'd be better off not adding the value back into the pool if this happens and just let it be GC'ed, since the state of `enc` in that case is unknown and it may be dirty

rvolosatovs:
FYI, https://github.com/TheThingsNetwork/lorawan-stack/blob/2e5e9dfab39b7adb909f09a7a7e366c57dee6395/pkg/networkserver/grpc_gsns.go#L259 introduced in this PR is broken and matching will always fallback, since `fallthrough` does not work the way I expected it to (it behaves like C `switch` without the `break` at the end of the case). cc @adriansmares I guess?

Note, that there are a few invalid `fallthrough` usages throughout the codebase.

johanstokking:
Thanks @rvolosatovs, that's really appreciated!

PR #3608, opened 2020-12-22
Support CUPS redirection

KrishnaIyer:
#### Summary

Add CUPS redirection support. Refs; https://github.com/TheThingsIndustries/lorawan-stack/issues/2088

#### Changes

- Setup Target CUPS URI and Key via gateway proto.
- Remove CUPS information from Gateway attributes.
- Modify GCS to use these
- Testing

#### Testing

- [x] UT
- [x] CUPS API Test
- [x] Initial Test with gateway

##### Regressions

- [x] E2E test with gateway (once review comments are addressed)

#### Notes for Reviewers

Please check [here](https://github.com/TheThingsIndustries/lorawan-stack/issues/2088#issuecomment-749479749) for more context.

@htdvisser  is the main reviewer since he wrote the original CUPS implementation.

#### Checklist


johanstokking (approved):
LGTM

Any field mask validation changes required?

api/gateway.proto
```diff
@@ -148,7 +148,14 @@ message Gateway {
   // The entire field must be used in RPCs since sub-fields are validated wrt to each other. Direct selection/update of sub-fields only are not allowed.
   // Use the top level field mask `claim_authentication_code` even when updating single fields.
   GatewayClaimAuthenticationCode claim_authentication_code = 23;
-  // next: 24
+  // CUPS URI for LoRa Basics Station CUPS redirection.
+  // The CUPS Trust field will be automatically fetched from the cert chain presented by the target server.
+  string target_cups_uri = 24 [(gogoproto.customname) = "TargetCUPSURI", (validate.rules).string.pattern = "^(?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*(?:[A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])(?::[0-9]{1,5})?$|^$"];
```
htdvisser: - Is there a specific reason for not allowing `https://` in front? Without a scheme, we can't really call this a URI.
- If not, can we just use a `(validate.rules).string.uri = true` here?

pkg/basicstation/cups/update_info.go
```diff
@@ -188,12 +192,14 @@ func (s *Server) UpdateInfo(c echo.Context) error {
...
 			md := metadata.New(map[string]string{
 				"id":            ids.GatewayID,
-				"authorization": fmt.Sprintf("Bearer %s", gtw.Attributes[cupsCredentialsAttribute]),
+				"authorization": fmt.Sprintf("Bearer %s", string(gtw.TargetCUPSKey.Value)),
 			})
 			if ctxMd, ok := metadata.FromIncomingContext(ctx); ok {
 				md = metadata.Join(ctxMd, md)
 			}
 			ctx = metadata.NewIncomingContext(ctx, md)
+			// This ensures that the newly registered gateway will make a CUPS request to the `gtw.TargetCUPSURI`.
```
htdvisser: I would reword this a bit, as we don't know for sure what the gateway will do. We do know that this will cause the server to return the CUPS URI.

pkg/basicstation/cups/update_info.go
```diff
@@ -229,68 +235,75 @@ func (s *Server) UpdateInfo(c echo.Context) error {
...
-	}
-	if s.allowCUPSURIUpdate && gtw.Attributes[cupsURIAttribute] != req.CUPSURI {
-		res.CUPSURI = gtw.Attributes[cupsURIAttribute]
+	_, host, port, err := parseAddress("https", req.CUPSURI)
+	if err != nil {
+		return errInvalidCUPSURI.WithAttributes("uri", req.CUPSURI)
 	}
+	cupsAddress := fmt.Sprintf("%s:%s", host, port)
+
+	if s.allowCUPSURIUpdate && gtw.TargetCUPSURI != "" && gtw.TargetCUPSURI != cupsAddress {
```
htdvisser: Why are we ignoring the scheme here?

pkg/basicstation/cups/update_info.go
```diff
@@ -229,68 +235,75 @@ func (s *Server) UpdateInfo(c echo.Context) error {
...
-	if credentials := gtw.Attributes[cupsCredentialsAttribute]; credentials != "" {
-		logger.WithField("cups_uri", gtw.Attributes[cupsURIAttribute]).Debug("Getting trusted certificate for CUPS...")
-		cupsTrust, err := s.getTrust(gtw.Attributes[cupsURIAttribute])
+		logger.Debug("Get trusted certificate for CUPS")
+		cupsTrust, err := s.getTrust(gtw.TargetCUPSURI)
 		if err != nil {
 			return err
 		}
-		cupsCredentials, err := TokenCredentials(cupsTrust, credentials)
+		cupsCredentials, err := TokenCredentials(cupsTrust, string(gtw.TargetCUPSKey.Value))
```
htdvisser: Maybe we could also use a flow to update (rotate) CUPS credentials?

api/gateway.proto
KrishnaIyer: `https` was implied but good to make that explicit.

pkg/basicstation/cups/update_info.go
KrishnaIyer: Good suggestion. Done.

api/gateway.proto
```diff
@@ -148,7 +148,17 @@ message Gateway {
...
+  // CUPS URI for LoRa Basics Station CUPS redirection.
+  // The CUPS Trust field will be automatically fetched from the cert chain presented by the target server.
+  string target_cups_uri = 24 [(gogoproto.customname) = "TargetCUPSURI", (validate.rules).string.uri = true];
+  // CUPS Key for LoRa Basics Station CUPS redirection.
+  // If redirecting to another instance of TTS, use the CUPS API Key for the gateway on the target instance.
+  // Requires the RIGHT_GATEWAY_READ_SECRETS for reading and RIGHT_GATEWAY_WRITE_SECRETS for updating this value.
+  Secret target_cups_key = 25 [(gogoproto.customname) = "TargetCUPSKey"];
+  // If set, credentials set in `target_cups_key` will be sent to the gateway on the next `/update-info` request.
+  // Used to rotate CUPS credentials.
+  bool update_cups_key = 26 [(gogoproto.customname) = "UpdateCUPSKey"];
```
htdvisser: Why is this necessary? The update-info request already contains a checksum of the credentials, so the server can just compare the values, right?

pkg/basicstation/cups/update_info.go
```diff
@@ -229,68 +235,77 @@ func (s *Server) UpdateInfo(c echo.Context) error {
...
 
-	if gtw.Attributes[cupsURIAttribute] == "" {
-		gtw.Attributes[cupsURIAttribute] = req.CUPSURI
-	}
-	if s.allowCUPSURIUpdate && gtw.Attributes[cupsURIAttribute] != req.CUPSURI {
-		res.CUPSURI = gtw.Attributes[cupsURIAttribute]
+	_, _, _, err = parseAddress("https", req.CUPSURI)
+	if err != nil {
+		return errInvalidCUPSURI.WithAttributes("uri", req.CUPSURI)
 	}
```
htdvisser: The fact that the gateway made this request in the first place, probably means that it had a valid URI already, right?

pkg/basicstation/cups/update_info.go
```diff
@@ -229,68 +235,77 @@ func (s *Server) UpdateInfo(c echo.Context) error {
...
 
-	if credentials := gtw.Attributes[cupsCredentialsAttribute]; credentials != "" {
-		logger.WithField("cups_uri", gtw.Attributes[cupsURIAttribute]).Debug("Getting trusted certificate for CUPS...")
-		cupsTrust, err := s.getTrust(gtw.Attributes[cupsURIAttribute])
+	if (s.allowCUPSURIUpdate && gtw.TargetCUPSURI != "" && gtw.TargetCUPSURI != req.CUPSURI) || gtw.UpdateCUPSKey {
```
htdvisser: With this you're allowing users to bypass `allowCUPSURIUpdate` by setting `UpdateCUPSKey=true`

KrishnaIyer: Good catch 👍

pkg/basicstation/cups/update_info.go
KrishnaIyer: It's technically possible for it to make a request to a valid URI but set a nonsense value to the `cupsURI` field in the request body. But this would fail in fetching the trust so I've improved the error logging down the line.

adriansmares (approved):

pkg/basicstation/cups/update_info.go
```diff
@@ -188,12 +193,14 @@ func (s *Server) UpdateInfo(c echo.Context) error {
...
 				"id":            ids.GatewayID,
-				"authorization": fmt.Sprintf("Bearer %s", gtw.Attributes[cupsCredentialsAttribute]),
+				"authorization": fmt.Sprintf("Bearer %s", string(gtw.TargetCUPSKey.Value)),
 			})
 			if ctxMd, ok := metadata.FromIncomingContext(ctx); ok {
 				md = metadata.Join(ctxMd, md)
 			}
 			ctx = metadata.NewIncomingContext(ctx, md)
+			// This makes the server return the target CUPS URI and credentials to the gateway.
+			req.CUPSURI = ""
```
htdvisser, line 197: Maybe I asked this before, but is this really necessary? If the CUPS URI in the request already matches the desired CUPS URI, we don't need to update it, right?

pkg/basicstation/cups/update_info.go
```diff
@@ -229,68 +236,81 @@ func (s *Server) UpdateInfo(c echo.Context) error {
 		gtw.Attributes = make(map[string]string)
 	}
 
-	if gtw.LBSLNSSecret == nil {
-		return errLNSCredentials.WithAttributes("gateway_uid", uid)
-	}
-
-	if s.requireExplicitEnable || gtw.Attributes[cupsAttribute] != "" {
+	if s.requireExplicitEnable && gtw.Attributes[cupsAttribute] != "" {
```
htdvisser: This is incorrect. If we require explicit enabling, we need to check for the attribute. If the attribute is missing or falsy, we return the error. If we don't require explicit enabling, but there is an attribute that explicitly disables it, we also return the error.

pkg/basicstation/cups/update_info.go
```diff
@@ -229,68 +236,81 @@ func (s *Server) UpdateInfo(c echo.Context) error {
...
+		cupsCredentials, err := TokenCredentials(cupsTrust, string(gtw.TargetCUPSKey.Value))
 		if err != nil {
 			return err
 		}
 		if crc32.ChecksumIEEE(cupsCredentials) != req.CUPSCredentialsCRC {
+			logger.Debug("Update CUPS Credentials")
 			res.CUPSCredentials = cupsCredentials
+			res.CUPSURI = req.CUPSURI
+		}
+	} else {
```
htdvisser: I think this entire if/elif/else block can be structured a bit better:

```go
if s.allowCUPSURIUpdate && gtw.TargetCUPSURI != "" {
  // we are allowed to and may want to configure another CUPS
  if gtw.TargetCUPSKey.Value == nil || gtw.TargetCUPSKey.Value == nil {
    // but for that we do need credentials
    return errTargetCUPSCredentials.New()
  }
  if gtw.TargetCUPSURI != req.CUPSURI {
    // we need to update the CUPS URI
    res.CUPSURI = gtw.TargetCUPSURI
  }
  cupsTrust, err := s.getTrust(gtw.TargetCUPSURI)
  // ...
  cupsCredentials, err := TokenCredentials(cupsTrust, string(gtw.TargetCUPSKey.Value))
  // ...
  if crc32.ChecksumIEEE(cupsCredentials) != req.CUPSCredentialsCRC {
    res.CUPSURI = req.CUPSURI
  }
} else {
  // we are not allowed to configure another CUPS, or we don't have one
  // so we configure LNS
```

pkg/basicstation/cups/update_info.go
```diff
@@ -229,68 +236,81 @@ func (s *Server) UpdateInfo(c echo.Context) error {
...
+			return errServerTrust.WithCause(err).WithAttributes("address", req.CUPSURI)
+		}
+		cupsCredentials, err := TokenCredentials(cupsTrust, string(gtw.TargetCUPSKey.Value))
 		if err != nil {
 			return err
 		}
 		if crc32.ChecksumIEEE(cupsCredentials) != req.CUPSCredentialsCRC {
+			logger.Debug("Update CUPS Credentials")
 			res.CUPSCredentials = cupsCredentials
+			res.CUPSURI = req.CUPSURI
```
htdvisser: We don't need to update the CUPS URI, because it's already as expected

pkg/basicstation/cups/update_info.go
KrishnaIyer: So if s.requireExplicitEnable is false and  gtw.Attributes[cupsAttribute] is "false", then we should error?

> s.requireExplicitEnable is false

In English: No need to explicitly enable CUPS, aka CUPS is available for all gateways.

> gtw.Attributes[cupsAttribute] is "false"

In English: Explicitly disable CUPS for this gateway. So we return an error?

To be honest I don't actually understand why this logic was needed in the first place.

pkg/basicstation/cups/update_info.go
KrishnaIyer: So with the "create gateway if not exists" flow, the gateway connects without any authentication. So by this, we force a CUPS redirect back to the same URI but with the newly created auth credentials sent in the gateway response.

pkg/basicstation/cups/update_info.go
htdvisser: It was originally intended as a feature flag. I'm fine with removing it.

pkg/basicstation/cups/update_info.go
KrishnaIyer: hmm. That doesn't account for the case where we need to only do a credentials update. `gtw.TargetCUPSURI` would need to be set update only the credentials. This is why I avoided this scheme.
So, for this, we'd need to ask users to se `gtw.TargetCUPSURI` to the current value in addition to setting the `gtw.TargetCUPSKey`. Not sure if we want that?

pkg/basicstation/cups/update_info.go
KrishnaIyer: Yup so this option is now deprecated; https://github.com/TheThingsNetwork/lorawan-stack/pull/3608/commits/49ab44754075ed5531517831868abf5a69321465

KrishnaIyer:
End to End tests are ✅ . Here's the process

#### Procedure

1. Create 2 servers with TLS endpoints/certificates setup.
2. Server 1: 
   - Register gateway.
   - Create a CUPS API Key `CUPS_KEY_1`.
3. Server 2: 
   - Register gateway.
   - Create a CUPS API Key `CUPS_KEY_1`.
   - Create an LNS API Key `LNS_KEY`.
   - Set the LNS Key for the gateway
   ```bash
    $ SECRET=$(echo $LNS_KEY | xxd -p -u | perl -pe 's/\n//')
    $  ttn-lw-cli gateways update <gateway-id> --lbs_lns_secret.value $SECRET
   ```
4. Server 1:
   - Setup Redirection with `target-cups-uri` as `https://<server-2-address>:443`. 
   - Use `cups-key-2` to set the `target-cups-key`
   ```bash
    $ SECRET=$(echo $CUPS_KEY_2 | xxd -p -u | perl -pe 's/\n//')
    $ ttn-lw-cli gateways update test-ttig-bc4 --target_cups_uri=<uri>:443 --target_cups_key.value $SECRET
   ```
  
5. Configure gateway to `https:<server-1-address>:443` using `CUPS_KEY_1`.
6. Gateway connects to `Server 1`. Server sends the updated credentials 
```
Jan 19 10:16:35  tti-lw-stack[1586]: DEBUG Configure CUPS                           cups_uri=server-2:443 gateway_eui=xxx http.method=POST http.path=/update-info namespace=web request_id=01EWD2ASZ00BB901W0CZYBFF9S
```
7. Gateway validates creds and connects to  `https:<server-2-address>:443` using `CUPS_KEY_2`.
8. `Server 2` returns the LNS Credentials.
9. Gateway connects to the LNS endpoint of `Server 2`.

pkg/basicstation/cups/config.go
```diff
@@ -40,7 +40,6 @@ type ServerConfig struct {
 // NewServer returns a new CUPS server from this config on top of the component.
 func (conf ServerConfig) NewServer(c *component.Component, customOpts ...Option) *Server {
 	opts := []Option{
-		WithExplicitEnable(conf.ExplicitEnable),
```
htdvisser, line 43: This is not what deprecated means. If we remove it, we chould mark the config as removed (ineffective). If we only deprecate it, we should still support it until we remove it.

pkg/basicstation/cups/messages.go
```diff
@@ -85,7 +85,7 @@ func TokenCredentials(trust *x509.Certificate, authorization string) ([]byte, er
 	out = append(out, trust.Raw...)
 	// TODO: Refactor when client side TLS is supported https://github.com/TheThingsNetwork/lorawan-stack/issues/137
 	out = append(out, emptyClientCert...)
-	out = append(out, []byte(fmt.Sprintf("%s%s%s", "Authorization: ", strings.TrimRight(authorization, "\r\n"), "\r\n"))...)
+	out = append(out, []byte(fmt.Sprintf("%s%s%s", "Authorization: Bearer ", strings.TrimRight(authorization, "\r\n"), "\r\n"))...)
```
htdvisser: This change makes the comment on the func incorrect. The comment still says _contents of the Authorization header_, but instead this is now always injecting "Bearer".

Why do we even do that? Not all networks that we might want to redirect to will use Bearer tokens.

pkg/basicstation/cups/config.go
KrishnaIyer: Ah I see. It's `removed` then. I've updated the description and changelog.

pkg/basicstation/cups/messages.go
KrishnaIyer: But TTS _does_ need "Bearer" (at least using the API Keys). I've used a different approach here; https://github.com/TheThingsNetwork/lorawan-stack/pull/3608/commits/105ae1675c902fc6ea5fb7640b2a905cc7d68d6e.

htdvisser (approved):

Issue #3605, opened 2020-12-21
Add direct database access for end-to-end tests


kschiffer:
#### Summary
For some test cases, it is beneficial to prepare the database with direct access, rather than through API requests. We should provide the means to do so in our end-to-end test tooling.

#### Why do we need this?
- Speed up tests (e.g. by retrieving access tokens directly from the database instead of via OAuth roundtrip)
- Enable some test-cases that need database states that cannot be easily instated (e.g. contact info validation, password updates using forgot-password-token)

#### What is already there? What do you see now?
Clearing and seeding databases using `mage` commands.

#### What is missing? What do you want to see?
Direct access to the database using a Postgres client within JS

#### How do you propose to implement this?
We originally had such functionality in the early end-to-end groundwork test PR to drop and seed the database. But @htdvisser rightfully noted that this could be achieved simpler using the respective mage targets that we already had in our tooling. However, we cannot achieve any flexible SQL manipulation using only the `mage` targets.

Hence, my idea is to simply reintroduce the Postgres client we used back then and extend the functionality. Basically, what we need to be able to do is to execute plain SQL commands from within a test.

#### Can you do this yourself and submit a Pull Request?
Yes.

cc @bafonins

kschiffer:
This has been added via #3683

Issue #3540, opened 2020-11-27
Convert functions to arrow functions


kschiffer:
#### Summary
We recently introduced an eslint warning for not using arrow functions. We should resolve such warnings in one big refactor PR, so that the warning output of eslint becomes unclogged.

#### Why do we need this?
- Annoying verbose and noisy warning output of `eslint`
- Annoying warnings and markings in IDEs
- Noise in PRs when these conversions are committed on top of changes

#### What is already there? What do you see now?
Eslint config update and the decision to move to arrow functions exclusively.

#### What is missing? What do you want to see?
All applicable functions being converted to arrow functions and hence eslint warnings about arrow functions being resolved.

#### How do you propose to implement this?
There are tools to help doing this automatically, e.g. https://github.com/cmstead/js-refactor
We might want to use such tools to streamline other code styling inconsistencies as well.

#### How do you propose to test this?
Unnecessary.

#### Can you do this yourself and submit a Pull Request?
@bafonins please pick this up or reassign.

bafonins:
>  We should resolve such warnings in one big refactor PR, so that the warning output of eslint becomes unclogged.

I dont think that global refactor is required. IMO, this should be refactored gradually.

> Annoying warnings and markings in IDEs

You can disable this in your ide.

> Noise in PRs when these conversions are committed on top of changes

In most cases refactoring plain function to arrow function is 2 line diff. I do not consider this to be a big issue.

Overall, we should move towards using arrow functions gradually and changed within feature/fix PR's. Also, this should be a joint effort of everyone working on `webui`.

PR #3613, opened 2020-12-22
Account app updates and cleanups

kschiffer:
#### Summary
This PR introduces fixes and updates to the current scaffold, which will prepare it for the upcoming PR on profile settings. It also contains refactors that will conform it to our current code style-guide, including the exclusive use of functional components, usage of arrow components. As such, the goal is also to reduce the diff of upcoming PRs.

#### Changes
- Convert all functions to arrow functions #3540 
- Convert all class components to functional components with hooks
- Remove context based usage of `env` (via `withEnv()` and `<EnvProvider />`)
- Update change password functionality
  - Use container class (to be used later-on in the authenticated view as well)
  - Use password validation rules from the IS (see #1067)
  - Add `updateUserPassword` rpc to JS SDK
- Add `requestPromiseMiddleware`
- Add redux primitives and logic for IS config
- Add SDK instance singleton module
- Some logic updates (using `createRequestLogic`)
- Optimize update password flow (e.g. redirect to login automatically)
- Fix routing
  - Handle root route via frontend (instead of backend redirect)
  - Use `?n=…` query parameter only for known authenticated routes 
  - Use not found page otherwise
- Replace usage of API tree with direct usage of the SDK instance
- Various small related fixes

#### Testing

I'm currently working on #3605 to be able to add some more tests, but would like this to be reviewed in the meantime already.

##### Regressions

All original OAuth functionality might experience regression with this.

#### Notes for Reviewers
Sorry for the big diff, it is mainly due to the functional component and arrow function updates. This could have been done successively as well but I really want the AA to be proper right from the start.

Again note that this PR targets the `feature/1422-account-app` feature branch, which I kept rebased on latest master.

#### Checklist


pkg/webui/account/views/overview/index.js
```diff
@@ -0,0 +1,68 @@
...
+
+@connect(
+  state => ({
+    user: selectUser(state),
+  }),
+  {
+    logout,
+  },
+)
+export default class Overview extends React.PureComponent {
```
kschiffer, line 37: Note that this will be completely replaced in the next PR, which I why I did not update this one to a functional component.

pkg/webui/account/views/create-account/index.js
```diff
@@ -91,151 +86,107 @@ const getSuccessMessage = state => {
...
+          {siteName}
+          <br />
+          <Message content={m.createANewAccount} component="strong" />
+        </h1>
+        <hr className={style.hRule} />
+        <Form
+          onSubmit={handleSubmit}
+          initialValues={initialValues}
+          error={error}
+          validationSchema={validationSchema}
```
bafonins, line 147: Any reasons for not using https://github.com/TheThingsNetwork/lorawan-stack/blob/feature/1422-account-app-cleanups/pkg/webui/lib/create-password-validation-schema.js for this form?

pkg/webui/account/views/forgot-password/index.js
```diff
@@ -58,81 +55,66 @@ const initialValues = { user_id: '' }
...
+          name="user_id"
+          component={Input}
+          autoFocus
+          required
+        />
+        <Form.Submit
+          component={SubmitButton}
+          message={m.send}
+          className={style.submitButton}
+          alwaysEnabled
```
bafonins: Cant find the `alwaysEnabled` prop defined anywhere

pkg/webui/account/views/forgot-password/index.js
```diff
@@ -58,81 +55,66 @@ const initialValues = { user_id: '' }
...
+        />
+        <Button.Link naked secondary message={sharedMessages.cancel} to="/login" />
```
bafonins: 1. Navigate to `/console`
2. Land on the login page and go to the forgot password view
3. Submit or just navigate back
4. Login
5. Land on the account app overview instead of the Console

Not sure if we should preserve the initial users location, but is seems weird that we do that for registration and not for the forgot password view.

pkg/webui/account/views/front/index.js
```diff
@@ -43,6 +47,11 @@ const FrontView = () => {
             <Route path="/validate" component={Validate} />
             <Route path="/code" component={Code} />
             <Redirect exact from="/" to="/login" />
+            {authRoutes.map(({ path, exact }) => (
+              <Route path={path} exact={exact} key={path}>
+                <Redirect to={`/login?n=${selectApplicationRootPath()}${location.pathname}`} />
```
bafonins, line 53: What is the purpose of this redirect?

pkg/webui/account/views/front/index.js
```diff
@@ -43,6 +47,11 @@ const FrontView = () => {
             <Route path="/validate" component={Validate} />
             <Route path="/code" component={Code} />
             <Redirect exact from="/" to="/login" />
+            {authRoutes.map(({ path, exact }) => (
```
bafonins, line 51: Currently, there is only one route in `authRouter` which is for `/`. When the user navigates to `/oauth` these lines are never reached because of L49

pkg/webui/account/views/front/index.js
```diff
@@ -26,10 +26,14 @@ import UpdatePassword from '@account/views/update-password'
 import FrontNotFound from '@account/views/front-not-found'
 import Validate from '@account/views/validate'
 import Code from '@account/views/code'
+import { authRoutes } from '@account/views/landing'
```
bafonins: IMO, moving this to a separate file is better.

bafonins (commented):
Submit section on forms looks a bit weird on tablets:

Screenshot 2020-12-27 at 16 33 35


Page content on the account overview completely disappears on phones:

Screenshot 2020-12-27 at 16 35 39

pkg/webui/account/views/front/index.js
kschiffer: The purpose is to redirect routes that require authentication back to the login page, with a query parameter that redirectes to the original route after login. E.g. `/profile-settings` (will be added later) would redirect back to the login page. All other routes that do not correspond to anything will show a 404.

Note that the `<Front />` handler only operates when the user is not authenticated.

pkg/webui/account/views/front/index.js
kschiffer: That's true. There will be more routes added with the next PRs which will make this setup purposeful. It's fine that `/` is handled by L49, since we don't need a redirect query parameter for it.

kschiffer:
Thanks for the review. I've addressed all your comments.

pkg/webui/lib/create-password-validation-schema.js
```diff
@@ -0,0 +1,56 @@
...
+      password => hasMinLength(password, requirements.min_length),
+    )
+    .test(
+      'max-length',
+      { message: sharedMessages.validateTooLong, values: { max: requirements.max_length } },
+      password => hasMaxLength(password, requirements.max_length),
+    )
+    .test(
+      'min-special',
+      { message: sharedMessages.validateSpecial, values: { special: requirements.min_special } },
```
bafonins, line 35: `validateSpecial` is not defined in `shared-messages`

pkg/webui/lib/create-password-validation-schema.js
```diff
@@ -0,0 +1,56 @@
...
+      password => hasSpecial(password, requirements.min_special),
+    )
+    .test(
+      'min-upper',
+      { message: sharedMessages.validateUppercase, values: { upper: requirements.min_uppercase } },
```
bafonins, line 40: `validateUppercase` is not defined in `shared-messages`

pkg/webui/lib/create-password-validation-schema.js
```diff
@@ -0,0 +1,56 @@
...
+      password => hasUpper(password, requirements.min_uppercase),
+    )
+    .test(
+      'min-digit',
+      { message: sharedMessages.validateDigit, values: { digit: requirements.min_digits } },
```
bafonins, line 45: `validateDigit` is not defined in `shared-messages`

pkg/webui/account/views/validate/index.js
```diff
@@ -45,70 +46,74 @@ const m = defineMessages({
...
+      } catch (error) {
+        if (isNotFoundError(error)) {
+          handleError(createFrontendError(m.tokenNotFoundTitle, m.tokenNotFoundMessage))
+        } else {
+          handleError(error)
+        }
       }
     }
-  }
+    makeRequest()
```
bafonins, line 83: Any reason for extracting the whole logic to a local function?

kschiffer: Because we need to run async logic there and it's not allowed to use an async function directly as argument for `useEffect()`, since the effect function must either return a cleanup function or nothing (but would then return a promise).

pkg/webui/account/store/middleware/user.js
```diff
@@ -12,28 +12,38 @@
...
+  },
+})
+
+const getUserLogic = createRequestLogic({
+  type: user.GET_USER,
+  process: async ({ action, getState }) => {
+    const {
+      meta: { selector },
+    } = action
+    console.log(action)
```
bafonins: Remove

bafonins:
1. Go to `/oauth`
2. Fill in login/password and submit the form
3. Results in an endless loop without being logged in to the account app

Configuration I used for testing:

```
go run ./cmd/ttn-lw-stack start all \
--log.level debug \
--is.oauth.ui.js-file libs.bundle.js,account.js \
--console.ui.js-file libs.bundle.js,console.js \
--console.ui.assets-base-url http://localhost:8080/assets \
--console.ui.canonical-url http://localhost:8080/console \
--console.ui.is.enabled --console.ui.is.base-url http://localhost:8080/api/v3 \
--console.ui.as.enabled --console.ui.as.base-url http://localhost:8080/api/v3 \
--console.ui.ns.enabled --console.ui.ns.base-url http://localhost:8080/api/v3 \
--console.ui.js.enabled --console.ui.js.base-url http://localhost:8080/api/v3 \
--console.ui.gs.enabled --console.ui.gs.base-url http://localhost:8080/api/v3 \
--console.oauth.logout-url=http://localhost:8080/oauth/logout \
--console.oauth.authorize-url=http://localhost:8080/oauth/authorize \
--console.oauth.token-url=http://localhost:8080/oauth/token
```

kschiffer:
Hmm that's odd. I cannot reproduce this. Does the same thing happen in incognito mode? It also doesn't seem to be reflected by the E2Es.

bafonins:
> Does the same thing happen in incognito mode?

yes, same thing.

> It also doesn't seem to be reflected by the E2Es.

Try adding `cy.wait(10000)` at the end of `succeeds logging in when using valid credentials` test. Same issue happens when running cypress.

I guess we should revisit the way we assert page load

kschiffer:
Got it now, you're missing `--is.oauth.ui.is.base-url http://localhost:8080/api/v3` in your configuration. Since this is a different host, the session cookie will not be sent to the IS making it return a 401.

Try like this:

```shell
go run ./cmd/ttn-lw-stack start \
--log.level debug \
--is.oauth.ui.js-file libs.bundle.js,account.js \
--console.ui.js-file libs.bundle.js,console.js \
--console.ui.assets-base-url http://localhost:8080/assets \
--console.ui.canonical-url http://localhost:8080/console \
--is.oauth.ui.canonical-url http://localhost:8080/oauth \
--is.oauth.ui.is.enabled --is.oauth.ui.is.base-url http://localhost:8080/api/v3 \
--console.ui.is.enabled --console.ui.is.base-url http://localhost:8080/api/v3 \
--console.ui.as.enabled --console.ui.as.base-url http://localhost:8080/api/v3 \
--console.ui.ns.enabled --console.ui.ns.base-url http://localhost:8080/api/v3 \
--console.ui.js.enabled --console.ui.js.base-url http://localhost:8080/api/v3 \
--console.ui.gs.enabled --console.ui.gs.base-url http://localhost:8080/api/v3 \
--console.oauth.logout-url=http://localhost:8080/oauth/logout \
--console.oauth.authorize-url=http://localhost:8080/oauth/authorize \
--console.oauth.token-url=http://localhost:8080/oauth/token
```

These are the additional ENVs I use on top [of the usual dev config setup](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.10/DEVELOPMENT.md#development-configuration):
```
export TTN_LW_IS_OAUTH_UI_JS_FILE="libs.bundle.js account.js"
export TTN_LW_IS_OAUTH_UI_IS_BASE_URL="http://localhost:8080/api/v3"
```

cmd/internal/shared/identityserver/config.go
```diff
@@ -28,15 +28,17 @@ import (
...
 				SiteName:      "The Things Stack for LoRaWAN",
+				Title:         "Account",
 				Language:      "en",
 				CanonicalURL:  shared.DefaultOAuthPublicURL,
 				AssetsBaseURL: shared.DefaultAssetsBaseURL,
 				IconPrefix:    "oauth-",
-				CSSFiles:      []string{"oauth.css"},
-				JSFiles:       []string{"oauth.js"},
+				CSSFiles:      []string{"account.css"},
+				JSFiles:       []string{"account.js"},
```
htdvisser: This is a breaking change for deployments that use custom JS/CSS files (which I don't think is very common). In any case, make sure to mention in CHANGELOG that deployments need to change their config if they customized these.

pkg/identityserver/identityserver.go
```diff
@@ -106,10 +108,18 @@ func New(c *component.Component, config *Config) (is *IdentityServer, err error)
 		store.ClientStore
 		store.OAuthStore
 	}{
-		UserStore:        store.GetUserStore(is.db),
-		UserSessionStore: store.GetUserSessionStore(is.db),
 		ClientStore:      store.GetClientStore(is.db),
 		OAuthStore:       store.GetOAuthStore(is.db),
+		UserStore:        store.GetUserStore(is.db),
+		UserSessionStore: store.GetUserSessionStore(is.db),
```
htdvisser: Unnecessary diff

htdvisser (commented):
Please update CODEOWNERS so that there are code owners on `pkg/account`

pkg/identityserver/identityserver.go
htdvisser: I'm still seeing the diff in the latest version of the PR

htdvisser (changes requested):
This still doesn't have CODEOWNERS on `pkg/account` as I requested in an earlier review. Please address previous comments before re-requesting review.

kschiffer:
> This still doesn't have CODEOWNERS on `pkg/account` as I requested in an earlier review. Please address previous comments before re-requesting review.

Sorry, missed a push there.

CODEOWNERS
```diff
@@ -58,6 +58,7 @@
 /pkg/identityserver @htdvisser @bafonins @kschiffer
 /pkg/email @htdvisser @bafonins @kschiffer
 /pkg/oauth @htdvisser @bafonins @kschiffer
+/pkg/account @htdvisser @bafonins @kschiffer
```
johanstokking, line 61: This is my only code owned file and I'm fine with this if this was discussed before.

Maybe loop in @asmulko or @mjamescompton instead of @htdvisser to reduce his scope

johanstokking (approved):
Please let me know if you want me to review something specifically, and if so, what

CODEOWNERS
htdvisser: We should indeed take a better look at the packages that are code-owned by me. In this case it's fine to add me me. (1) It's pretty much all code that was moved from `pkg/oauth`, so I already owned it. (2) It's pretty small and (3) I don't expect it to change often.

kschiffer:
> Please let me know if you want me to review something specifically, and if so, what

I don't think there's anything else for you to review in this.

pkg/webui/account/store/middleware/user.js
```diff
@@ -12,28 +12,37 @@
...
+})
+
+const getUserLogic = createRequestLogic({
+  type: user.GET_USER,
+  process: async ({ action, getState }) => {
+    const {
+      meta: { selector },
+    } = action
+    const userId =
+      'payload' in action && action.payload.id ? action.payload.id : selectUserId(getState())
```
bafonins, line 40: it seems that id` is required here. any use case when it can be missing?

pkg/webui/account/store/reducers/identity-server.js
```diff
@@ -0,0 +1,31 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { handleActions } from 'redux-actions'
+
+import { GET_IS_CONFIGURATION_SUCCESS } from '@console/store/actions/identity-server'
```
bafonins: I dont think we should reference console and account apps. Some repetition wont hurt until we implement some proper api to share store related functionality between apps.

bafonins: Actually, there are is actions defined in `@account/store/actions/identity-server`

pkg/webui/account/views/create-account/index.js
```diff
@@ -91,151 +87,124 @@ const getSuccessMessage = state => {
...
-      })
-    } catch (error) {
-      this.setState({
-        error: error.response.data,
-      })
-    } finally {
-      setSubmitting(false)
-    }
+  const [error, setError] = useState(undefined)
+  const passwordRequirements = useSelector(selectPasswordRequirements)
```
bafonins, line 97: You can get the is configuration directly from `useRequest` on L87

pkg/webui/account/views/login/index.js
```diff
@@ -59,112 +57,107 @@ const validationSchema = Yup.object().shape({
...
-    return next.split('?')[0]
+  if (location.state && location.state.info) {
+    info = location.state.info
+  } else if (!next || (next !== appRoot && !Boolean(error))) {
+    info = m.loginToContinue
   }
 
-  return next
+  return (
+    <React.Fragment>
```
bafonins: fragment not needed

pkg/webui/account/views/login/index.js
```diff
@@ -59,112 +57,107 @@ const validationSchema = Yup.object().shape({
...
+            <Form.Submit
+              component={SubmitButton}
+              message={sharedMessages.login}
+              className={style.submitButton}
+            />
+            <Button.Link
+              to={`/register${location.search}`}
+              secondary
+              message={m.createAccount}
+              className={style.registerButton}
```
bafonins: `register-button` class does not exist

pkg/webui/account/views/front/index.js
```diff
@@ -27,9 +29,12 @@ import FrontNotFound from '@account/views/front-not-found'
...
+import { selectApplicationRootPath } from '@ttn-lw/lib/selectors/env'
+import PropTypes from '@ttn-lw/lib/prop-types'
+
 import style from './front.styl'
 
-const FrontView = () => {
+const FrontView = ({ location }) => {
   return (
     <div className={style.container}>
       <section className={style.content}>
```
bafonins, line 40: On the next line `.header` class does not exist.

pkg/webui/console/components/user-data-form/index.js
```diff
@@ -79,53 +74,14 @@ const validationSchema = Yup.object().shape({
...
 @injectIntl
 class UserForm extends React.Component {
   constructor(props) {
     super(props)
 
     const { update, passwordRequirements } = props
     this.validationSchema = update
-      ? validationSchema
+      ? baseValidationSchema
       : createPasswordValidationSchema(passwordRequirements)
```
bafonins: We need to merge password validation with validation rules for other fields as well. With this implementation:
1. Form values are no properly casted
2. Multiple fields are missing validation

pkg/webui/lib/components/full-view-error/error.js
```diff
@@ -33,10 +32,13 @@ import {
...
 import statusCodeMessages from '@ttn-lw/lib/errors/status-code-messages'
 import PropTypes from '@ttn-lw/lib/prop-types'
+import { selectApplicationRootPath } from '@ttn-lw/lib/selectors/env'
 
 import style from './error.styl'
 
-const FullViewErrorInner = function({ error, env }) {
+const appRoot = selectApplicationRootPath()
+
+const FullViewErrorInner = function({ error }) {
```
bafonins:
```suggestion
const FullViewErrorInner = ({ error }) => {
```

pkg/webui/lib/components/full-view-error/error.js
```diff
@@ -88,22 +90,19 @@ const FullViewErrorInner = function({ error, env }) {
   )
 }
 
-const FullViewErrorInnerWithEnv = withEnv(FullViewErrorInner)
-
 const FullViewError = function({ error, header }) {
```
bafonins:
```suggestion
const FullViewError = ({ error, header }) => {
```

pkg/webui/account.js
```diff
@@ -15,40 +15,35 @@
...
-const rootElement = document.getElementById('app')
+if (selectSentryDsnConfig()) {
+  Sentry.init(sentryConfig)
+}
 
 const render = () => {
   const App = require('./account/views/app').default
 
   DOM.render(
-    <EnvProvider env={env}>
```
bafonins, line 42: We still have base components that rely on `withEnv` hoc, e.g. `<BaseAnchorLink />` and `<DocLink />` that can be used in the account app. We should keep the `<EnvProvider />` as long as it is used in `@ttn-lw/conponents/`.

bafonins (changes requested):

htdvisser (approved):

pkg/webui/account/store/middleware/user.js
kschiffer: Once the user is logged in, we can omit the user id and get it from the store. This does not work initially during the init logic. Since after the initialization the user is stored, there should be no situation where the user id is missing.

pkg/webui/account/store/reducers/identity-server.js
kschiffer: Yup. Forgot to rename the import.

pkg/webui/account/views/create-account/index.js
kschiffer: I can but it would need an extra selector to extract the password requirements from the is config object (and not from the root of the store). I don't think it's worth the hassle.

pkg/webui/account.js
kschiffer: Searching in `pkg/webui/components` reveals that `<BaseAnchorLink />` and `<DocLink />` seem to be the only one's that still use `withEnv`. I'll add a commit to replace this with env selectors.

kschiffer:
@bafonins thanks for the review. I've addressed all your comments and also did another rebase on latest `v3.11`. I had to add [some additional commits](https://github.com/TheThingsNetwork/lorawan-stack/pull/3613/commits/a817d1de1bc44cc67d668b89ece9f5b91db04f18) to properly merge online status functionality, which was introduced via #3629.

bafonins (approved):
Manually went and tested through various pages of the account app. Didnt review the code completely, only changes since the last review.

PR #3616, opened 2020-12-23
Entity rights transfer

### LINKED ISSUE #3458: Transfer full entity rights to another user on soft-delete of user

pgalic96:
#### Summary

Closes #3458 

I suppose this should end up in 3.11 because it is also changing existing delete method to return error in case there are orphaned entities present. (let me know if this is right otherwise will change to `3.10`)

#### Changes

- Add `TransferOwnedEntityRights` RPC 
- Add `FindSingleOwnerMemberships` store method that returns a list of `ttnpb.Identifier`s for which the user is the only owner (full rights on entity)
- Add `TransferEntityRights` store method that transfers user rights for a list of entities to another user
- Add a check in the `Delete` user registry method which calls `FindSingleOwnerMemberships` and returns error if there are any entities
- Add `transfer-entity-rights` cli command

#### Testing

unit tests, manual

##### Regressions

- there shouldn't be any as this is a new functionality

#### Notes for Reviewers

@htdvisser I had a lot of problems querying memberships based on the rights, mainly because of its type (int32 array) and the fact there are multiple rights that are considered owning rights. I solved it using `pq.Array`, however I needed to use this long `x OR y OR z OR n` query because the query doesn't support array of arrays to check if the rights are in the array of owning rights. If you have any idea how to make this query more concise let me know.

Also both store methods are designed to be used as separate RPCs later (list owned entities, and transfering rights for list of entities to another user), and I also tried to account for organizations as well.

#### Checklist


cmd/ttn-lw-cli/commands/users.go
```diff
@@ -60,6 +68,19 @@ func getUserID(flagSet *pflag.FlagSet, args []string) *ttnpb.UserIdentifiers {
 	return &ttnpb.UserIdentifiers{UserID: userID}
 }
 
+func getReceiverID(flagSet *pflag.FlagSet, args []string) *ttnpb.UserIdentifiers {
+	var userID string
+	if len(args) > 0 {
+		userID = args[1]
```
johanstokking: Panics if `len(args) == 1`

Also, you're using this in combination with `getUserID()` which warns if there are multiple identifiers passed, but that seems to be intended for `transfer-owner-rights`.

api/user_services.proto
```diff
@@ -91,6 +91,12 @@ service UserRegistry {
       delete: "/users/{user_id}/purge"
     };
   };
+  // Transfers solely owned entity rights to another user
+  rpc TransferOwnedEntityRights(TransferEntityRightsRequest) returns (google.protobuf.Empty) {
```
johanstokking: What is solely owned? Rights that only the user identified by `user_ids` possesses?

What should be clear from the description (and rpc name) is what it does, i.e. does it:
1. Transfer only rights that `user_ids` has, and nobody else has?
2. Transfer all rights that `user_ids` has

johanstokking (commented):
Only reviewed CLI and API

api/user_services.proto
pgalic96: yeah I didn't come up with a concise name that explains the issue. `Transfering rights for all entities where the user is the only collaborator that has full rights on the entity` would explain it but not in a concise way.

htdvisser: I actually like @johanstokking's (2). If we just make it transfer all rights, that will be a lot easier to explain.

api/user.proto
```diff
@@ -122,6 +122,11 @@ message UpdateUserPasswordRequest {
   bool revoke_all_access = 4;
 }
 
+message TransferEntityRightsRequest {
```
htdvisser: I would rename this to `TransferUserRightsRequest`.

pkg/identityserver/store/membership_store.go
```diff
@@ -294,3 +295,75 @@ func (s *membershipStore) DeleteAccountMembers(ctx context.Context, id *ttnpb.Or
...
+}
+
+func (s *membershipStore) TransferEntityRights(ctx context.Context, id *ttnpb.OrganizationOrUserIdentifiers, receiverID *ttnpb.OrganizationOrUserIdentifiers, entityList []ttnpb.Identifiers) error {
+	defer trace.StartRegion(ctx, "transfer entity rights").End()
+	for _, entityID := range entityList {
+		userRights, err := s.GetMember(ctx, id, entityID)
+		if err != nil {
+			return err
+		}
+		err = s.SetMember(ctx, receiverID, entityID, userRights)
```
htdvisser: Maybe good to merge with existing rights instead of replacing them. Even though this func is currently only used when we know the receiver doesn't already have rights on the entity, this may be confusing for future work.

pkg/identityserver/store/membership_store.go
```diff
@@ -294,3 +295,75 @@ func (s *membershipStore) DeleteAccountMembers(ctx context.Context, id *ttnpb.Or
...
+	accountQuery := s.query(ctx, Account{}).
+		Select(`"accounts"."id"`).
+		Where(fmt.Sprintf(`"accounts"."account_type" = '%s' AND "accounts"."uid" = ?`, id.EntityType()), id.IDString()).
+		QueryExpr()
+	userMembershipQuery := s.query(ctx, &Membership{}).
+		Select(`"memberships"."entity_id","memberships"."entity_type"`).
+		Where(`"memberships"."account_id" = (?)`, accountQuery).
+		Where(`"memberships"."rights" = (?) OR "memberships"."rights" = (?) OR "memberships"."rights" = (?) OR "memberships"."rights" = (?)`,
+			pq.Array([]ttnpb.Right{ttnpb.RIGHT_ALL}),
+			pq.Array([]ttnpb.Right{ttnpb.RIGHT_GATEWAY_ALL}),
```
htdvisser: Are we forgetting OAuth client entities here?

pkg/identityserver/store/membership_store.go
```diff
@@ -294,3 +295,75 @@ func (s *membershipStore) DeleteAccountMembers(ctx context.Context, id *ttnpb.Or
...
+			pq.Array([]ttnpb.Right{ttnpb.RIGHT_APPLICATION_ALL}),
+			pq.Array([]ttnpb.Right{ttnpb.RIGHT_ORGANIZATION_ALL})).
+		QueryExpr()
+	query := s.query(ctx, &Membership{}).
+		Select(`"memberships"."entity_id", "memberships"."entity_type"`).
+		Where(`("memberships"."entity_id","memberships"."entity_type") IN (?)`, userMembershipQuery).
+		Where(`"memberships"."rights" = (?) OR "memberships"."rights" = (?) OR "memberships"."rights" = (?) OR "memberships"."rights" = (?)`,
+			pq.Array([]ttnpb.Right{ttnpb.RIGHT_ALL}),
+			pq.Array([]ttnpb.Right{ttnpb.RIGHT_GATEWAY_ALL}),
```
htdvisser: Are we forgetting OAuth client entities here?

pkg/identityserver/store/membership_store.go
```diff
@@ -294,3 +295,75 @@ func (s *membershipStore) DeleteAccountMembers(ctx context.Context, id *ttnpb.Or
...
+func (s *membershipStore) FindSingleOwnerMemberships(ctx context.Context, id *ttnpb.OrganizationOrUserIdentifiers) ([]ttnpb.Identifiers, error) {
+	defer trace.StartRegion(ctx, "find single full rights memberships ").End()
+	accountQuery := s.query(ctx, Account{}).
+		Select(`"accounts"."id"`).
+		Where(fmt.Sprintf(`"accounts"."account_type" = '%s' AND "accounts"."uid" = ?`, id.EntityType()), id.IDString()).
+		QueryExpr()
+	userMembershipQuery := s.query(ctx, &Membership{}).
+		Select(`"memberships"."entity_id","memberships"."entity_type"`).
+		Where(`"memberships"."account_id" = (?)`, accountQuery).
+		Where(`"memberships"."rights" = (?) OR "memberships"."rights" = (?) OR "memberships"."rights" = (?) OR "memberships"."rights" = (?)`,
```
htdvisser: I'm not sure `= (?)` is the correct way to check this. It's possible that users have the `_ALL` right combined with other rights if they first had the other rights, and then were granted the `_ALL` right.

api/user_services.proto
pgalic96: @htdvisser By transferring all the rights you mean the full (owner) rights? Because if you mean all the entity rights then the user might not have the rights to set the collaborators for the entity and this doesn't seem like something we should bypass.

pgalic96:
@htdvisser I want to make you aware of an issue with this command. Transfering a set of entities enters a deadlock after performing the transfer action for the first entity in the array (not seeing any queries in logs for any other entities except the first one in the array). It happens on this line:
https://github.com/TheThingsNetwork/lorawan-stack/blob/0aecf6cdd22fa17efc492628146525fa1221d77a/pkg/identityserver/user_registry.go#L715
This line is never executed:
https://github.com/TheThingsNetwork/lorawan-stack/blob/5d458e21ccaecce37393301d0f16cafbbc26ea0b/pkg/identityserver/store/membership_store.go#L227

Do you know what might be the reason for this?

cmd/ttn-lw-cli/commands/users.go
```diff
@@ -415,6 +451,33 @@ var (
 			return nil
 		},
 	}
+	usersTransferRightsCommand = &cobra.Command{
+		Use:     "transfer-rights [user-id] [receiver-id]",
+		Aliases: []string{"transfer", "move", "rights-move"},
```
htdvisser: Let's not add these aliases. Perhaps in the future we want to `transfer` or `move` users somewhere else, which would then give us conflicts.

pkg/auth/rights/require.go
```diff
@@ -66,6 +66,29 @@ var (
 	)
 )
 
+// RequireEntity checks that context contains the required rights for the
+// entity.
+func RequireEntity(ctx context.Context, id *ttnpb.EntityIdentifiers, required ...ttnpb.Right) (err error) {
```
htdvisser: I don't really like this, as it's too easy to use it in the wrong way.

api/user.proto
```diff
@@ -128,6 +128,11 @@ message UpdateUserPasswordRequest {
   bool revoke_all_access = 4;
 }
 
+message TransferUserRightsRequest {
+  UserIdentifiers user_ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (validate.rules).message.required = true];
+  UserIdentifiers receiver_ids = 2 [(gogoproto.nullable) = false, (validate.rules).message.required = true];
```
htdvisser:
```suggestion
  UserIdentifiers sender_ids = 1 [(gogoproto.nullable) = false, (validate.rules).message.required = true];
  UserIdentifiers receiver_ids = 2 [(gogoproto.nullable) = false, (validate.rules).message.required = true];
```

I'm actually not a big fan of `sender`/`receiver` (you don't send rights, you transfer them), and would prefer `from`/`to` (you transfer rights from user A to user B)

pkg/ttnpb/rights.go
```diff
@@ -65,6 +65,7 @@ var (
 	AllClusterRights      = &Rights{}
 	AllAdminRights        = &Rights{}
 	AllRights             = &Rights{}
+	AllCollaboratorRights = &Rights{}
```
htdvisser: Considering that this is only used in a single place, I don't think we should add this here.

htdvisser:
Re-requesting review to look at https://github.com/TheThingsNetwork/lorawan-stack/pull/3616#issuecomment-780074806 later.

pgalic96:
@htdvisser answered most of your feedback, tested that the issue from my comment still persists.

htdvisser:
We've decided not to do this. 

This is trying to solve a problem that only happened 1 or 2 times, and can also easily be fixed with admin intervention (restoring the deleted user).

The solution is currently too inefficient. While it might work well for a couple of entities, it will be extremely slow for users with many memberships.

It's just not worth the effort at this time.

Issue #2825, opened 2020-07-02
Better account for connection losses in the Console


kschiffer:
#### Summary
We currently don't handle the loss of internet connection very well in the Console. This needs to be improved.

#### Why do we need this?
Connection losses result in [unhandled exceptions](https://sentry.io/organizations/the-things-industries/issues/1676736688/?project=2682566) and other than a message informing about the application being offline, it behaves as if it is still online.

#### What is already there? What do you see now?
- A toast notification informing about the loss of connection
- An offline status appearing in the footer

#### What is missing? What do you want to see?
Several things:
- Avoid sending API requests when the app is offline
  - Use data that is already in the store instead
  - Use a service worker to provide caching and a true offline mode as we do in DCS
- Provide a bolder visual hint that the app is running in offline mode (the current offline tag in the footer is barely noticeable)
  - This should also inform about the fact that cached values are used until the connection is re-established
- Look into preventing any avoidable exception resulting from the app being offline

#### Environment
v3.8.4

#### How do you propose to implement this?
@mjamescompton has good insights into implementing this from his work on the DCS

#### How do you propose to test this?
Manual and end-to-end tests.

#### Can you do this yourself and submit a Pull Request?
I think @mjamescompton is best suited for this.

htdvisser:
Also related to https://github.com/TheThingsNetwork/lorawan-stack/issues/2681

kschiffer:
Closed via #4400

PR #3629, opened 2020-12-31
Improve Console handling of offline states and connection issues

kschiffer:
#### Summary
This PR improves the Console's handling of connection issues as well as detecting and reacting to offline states.

image
image

References #2825

#### Changes
- Introduce new store logic for different "online status"
  - Use enum instead of boolean for the "online status": `ONLINE`, `OFFLINE` and `CHECKING`
  - Detect network and timeout errors within request logics and set the connection state to `CHECKING` accordingly
  - Issue a connection probe by requesting the `auth_info` endpoint and set the online state to `ONLINE` or `OFFLINE` accordinigly
  - When `OFFLINE`, issue periodic reconnect attempts and reset the online status back to `ONLINE` if successful
  - Prevent any  "ill-fated" requests within request logics, while the online status is `OFFLINE` and issue reconnect attempts instead
- Make `withRequest` HOC and `useRequest` hook aware of the online status and re-issue request actions if the application has gone back `ONLINE`
- Add a timeout config for the JS SDK axios: `10000ms` to avoid running requests unnecessarily long when the connection is gone or wonky
- Fix possible unhandled promise rejection in the JS SDK `stream.js` handler, which currently [spams our sentry logs](https://sentry.io/organizations/the-things-industries/issues/1661144927/)
- Update views and components for new shape of `onlineStatus`

#### Testing

Manual testing. This can be done by e.g.
- Shutting down the stack server
- Disabling the internet connection while keeping the Wi-Fi connection running. I used my phone as Wi-Fi hotspot and disabled mobile data. This allows for realistic testing in conjunction with using the staging environments.

##### Regressions

This change affects the `createLogicMiddleware`, which we use to wrap all store logics, as such, an introduced bug could affect all logics.

#### Notes for Reviewers
Note that e.g. shutting down the Wi-Fi altogether or using the network tab of dev tools is not enough to test this properly. We have logic in place that checks the `navigator.onLine` prop or uses the `online`/`offline` window events and sets the online state in the store accordingly. These events are connected to the OS and whether it is currently connected to a network. The offline mode in the network tab of dev tools is using this API as well.

However, another more realistic scenario is the internet connection breaking up while still appearing properly connected for the OS. In such cases, we cannot use the browser events above. The only way to detect a broken connection then is to intercept requests that return network errors. Which is what this PR does.

#### Checklist


pkg/webui/console/store/actions/lib.js
```diff
@@ -55,8 +55,27 @@ export const attachPromise = actionOrActionCreator => {
...
     const action = actionOrActionCreator(...args)
     return decorateAction(action)
   }
 }
+
+/**
+ * Helper function to retrieve the result action types based
+ * on the request action type.
+ *
+ * @param {object|string} typeString - The request action type.
```
bafonins: Is it an array or an object?

Note that
```js
const obj = {};
obj instanceof Array // is `false`
```

```suggestion
 * @param {Array<...>|string} typeString - The request action type.
```

pkg/webui/console/store/actions/lib.js
```diff
@@ -55,8 +55,27 @@ export const attachPromise = actionOrActionCreator => {
...
+ * @param {string} status - The result type, either `SUCCESS` or `FAILURE`.
+ * @returns {string} - The result action type.
+ */
+export const getResultActionFromType = (typeString, status) => {
+  if (typeString instanceof Array) {
```
bafonins: What is the use case of passing an array here?

pkg/webui/components/offline-status/index.js
```diff
@@ -44,34 +46,45 @@ const handleMessage = (message, type) => {
...
+    ? sharedMessages.online
+    : isChecking
+    ? sharedMessages.connectionIssues
+    : sharedMessages.offline
+
   return (
-    <span className={classnames(style.status, { [style.online]: isOnline })}>
-      <Icon className={style.icon} icon={isOnline ? 'info' : 'error'} />
-      <Message content={isOnline ? sharedMessages.online : sharedMessages.offline} />
+    <span className={classnames(style.status, style[onlineStatus])}>
```
bafonins: Using `style[onlineStatus]` can cause styling issues later, especially that there no tests for this component. If anyone changes values of `ONLINE_STATUS` this breaks.

I would prefer using something like this:
```js
{
  [style.online]: isOnline,
  [style.offline]: isOffline,
  [style.checking]: isChecking,
}

File: pkg/webui/console/store/middleware/logics/lib.js [hunk 4]
```diff
@@ -79,41 +110,47 @@ const createRequestLogic = function(
...
-              Sentry.captureException(e)
-            })
+        } else if (isNetworkError(e) || isTimeoutError(e)) {
+          // If there was a network error, it could mean that the network
+          // connection is currently interrupted. Setting the online state to
+          // `checking` will trigger respective connection management logics.
+          if (!selectIsCheckingStatus(getState()) && action.type !== ATTEMPT_RECONNECT) {
+            // We only need to set the status and trigger the connection checks
+            // if the online status was `online` previously.
+            dispatch(setOnlineStatus(ONLINE_STATUS.CHECKING))
```
bafonins:
```suggestion
            dispatch(setStatusChecking())
```

Not sure if `setOnlineStatus` is even needed

pkg/webui/console/store/middleware/logics/status.js
```diff
@@ -0,0 +1,104 @@
...
+
+import * as status from '@console/store/actions/status'
+
+import { selectIsOfflineStatus } from '@console/store/selectors/status'
+
+const isRoot = selectIsConfig().base_url
+
+const initialInterval = 5000
+let interval = initialInterval
+const connectionCheck = (dispatch, done) => async () => {
```
bafonins: Can we remove `async` here?

pkg/webui/console/store/middleware/logics/status.js
```diff
@@ -0,0 +1,104 @@
...
+  dispatch(status.attemptReconnect())
+  done()
+}
+
+let periodicCheck
+
+const connectionManagementLogic = createLogic({
+  type: status.SET_CONNECTION_STATUS,
+  process: async ({ action, getState }, dispatch, done) => {
```
bafonins:
```suggestion
  process: async ({ action }, dispatch, done) => {
```

pkg/webui/console/store/middleware/logics/status.js
```diff
@@ -0,0 +1,104 @@
...
+  validate: ({ action, getState }, allow, reject) => {
+    if (selectIsOfflineStatus(getState()) && navigator.onLine) {
+      return allow(action)
+    }
+    if (Boolean(periodicCheck)) {
+      clearTimeout(periodicCheck)
+    }
+    reject()
+  },
+  process: async ({ action, getState }, dispatch, done) => {
```
bafonins:
```suggestion
  process: async (_, dispatch, done) => {
```

pkg/webui/console/store/middleware/logics/status.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    done()
+  },
+})
+
+const connectionCheckLogic = createLogic({
+  type: status.ATTEMPT_RECONNECT,
+  // Additionally to periodic reconnects, freshly incoming request actions will
+  // also trigger reconnection attempts, which is why this acction is throttled
```
bafonins:
```suggestion
  // also trigger reconnection attempts, which is why this action is throttled
```

pkg/webui/console/store/middleware/logics/status.js
```diff
@@ -0,0 +1,104 @@
...
+      return allow(action)
+    }
+    if (Boolean(periodicCheck)) {
+      clearTimeout(periodicCheck)
+    }
+    reject()
+  },
+  process: async ({ action, getState }, dispatch, done) => {
+    try {
+      await axios.get(`${isRoot}/auth_info`, { timeout: 4500 })
```
bafonins, line 81: Why 4500ms here and 5000ms on L44?

bafonins (changes requested):

pkg/webui/console/store/actions/lib.js
kschiffer: I actually just moved this here from `/webui/console/store/middleware/logics/lib.js`. I don't remember why I included the possibility to use arrays too. I'll remove it since it does not seem to be used anywhere.

pkg/webui/console/store/middleware/logics/status.js
kschiffer: The attempts here can be retried in 5 second intervals, so I wanted to ensure that the old request has finished before the retry fires.

bafonins (approved):

Issue #2798, opened 2020-06-25
Migrate away from gogo/protobuf


rvolosatovs:
#### Summary

https://github.com/gogo/protobuf is not mantained any more https://github.com/gogo/protobuf/issues/691 (currently)

#### Why do we need this?

It's our dependency, which is incompatible with new `golang/protobuf` version, which more and more packages depend on, hence we need to replace the `golang/protobuf` version, depending on outdated versions of our direct dependencies and potentially even breaking packages this way

#### What is already there? What do you see now?

`gogo/protobuf` dependency

#### What is missing? What do you want to see?

Figure this out

#### How do you propose to implement this?

Figure out if a new maintainer will appear or different plugin with feature parity? 
Use just vanilla protobuf?

#### How do you propose to test this?

tests

#### Can you do this yourself and submit a Pull Request?

yes

rvolosatovs:
Validation plugin we're using dropped support for GoGo
https://github.com/envoyproxy/protoc-gen-validate/pull/340

htdvisser:
I think the best way forward is to follow the ecosystem and migrate away from gogo/protobuf. With more and more of our other dependencies moving away from gogo, I think it will become increasingly difficult to keep using it. Of course it's going to be a lot of work to migrate, so if we do it, we need to come up with a good plan.

@rvolosatovs probably knows more about the custom options that are set in our `gogottn` generator, but here's what I found for the explicit options in our proto files:

- We could start by removing the `gogoproto.customname`, `gogoproto.stdtime` and `gogoproto.stdduration`, and `goproto_enum_prefix` options. Those are relatively easy to remove, since the Go compiler will immediately complain about any resulting issues.
- Removing `gogoproto.embed` option would mean that we can no longer access embedded fields (the Go compiler will help us find those), and that messages no longer satisfy some interfaces (this may be more difficult).
- The `gogoproto.nullable` option will be much more work, because we will have to start using the getters, and add nil-checks. Resulting problems may not get caught by the Go compiler. Possible workaround would be to temporarily make those fields private, then rewrite to getters/setters and finally making the fields public again.
[...]

PR #3647, opened 2021-01-08
Add FieldIsZero, validate non-zero fields in middleware

rvolosatovs:
#### Summary

In it's current state field mask validation only partially restricts the fields that can be set.
For example if RPC supports paths `A` and `A.B`, sending value with `A.C` non-empty and field mask `A` should result in an error - the RPC implementations currently must perform such checks, but that is error-prone and hard to maintain (in case of NS `Set` RPC this is several hundreds lines of checks)- this PR addresses that.
An example of real-word case where this is very relevant is key envelopes or MAC state.

#### Changes

- Implement `FieldIsZero` on `EndDevice` and related messages
- If `FieldIsZero` is implemented for a message and fieldmask is specified - ensure that disallowed fields are zero in the interceptor.

#### Testing

Tested creating a new class C device and sending a bunch of uplinks. Could not test downlink flow, because application downlink is broken cc @adriansmares 

##### Regressions

All clients, which are currently sending invalid messages will be broken - that is expected and desired.
That probably means that CLI should also ensure only the allowed fields are sent. cc @htdvisser 

#### Checklist


pkg/rpcmiddleware/validator/validator.go
```diff
@@ -60,17 +60,45 @@ func convertError(err error) error {
...
+			if ok {
+				continue
+			}
+			if isZero == nil {
+				if v, ok := msg.(interface {
+					FieldIsZero(string) bool
+				}); ok {
+					isZero = v.FieldIsZero
+				} else {
+					// TODO: Use reflection to validate a (sub-)field by path
```
rvolosatovs: Do we want to do this?
If so, I will file an issue and add a reference according to the guidelines, otherwise I will remove this `else`
Ideally these methods should be generated, but that should be done after #2798 is solved

pkg/rpcmiddleware/validator/validator.go
```diff
@@ -60,17 +86,53 @@ func convertError(err error) error {
 	return grpc.Errorf(codes.InvalidArgument, err.Error())
 }
 
-var errNoValidator = errors.DefineUnimplemented("no_validator", "validator not defined")
+var (
+	errNoValidator = errors.DefineUnimplemented("no_validator", "validator not defined")
+	errNonZeroPath = errors.DefineInvalidArgument("non_zero_path", "path '{path}' is not zero")
```
johanstokking: nit; backticks

pkg/rpcmiddleware/validator/validator.go
```diff
@@ -60,17 +86,53 @@ func convertError(err error) error {
...
+				return !strings.HasPrefix(tail[j], prefix)
+			})] {
+				if isZero == nil {
+					if v, ok := msg.(interface {
+						FieldIsZero(string) bool
+					}); ok {
+						isZero = v.FieldIsZero
+					} else {
+						// TODO: Use reflection to validate a subfield by path
+						break
```
johanstokking: So if `msg` is of type `EndDevice`, and `FieldIsZero` is implemented on `MACState`, it doesn't get called unless `EndDevice` implements some umbrella `FieldIsZero`? I think it makes sense to check the nearest `FieldIsZero` of `sp` here.

If we don't do this, this `FieldIsZero` support on `msg` should be done before the loop.

pkg/ttnpb/join.go
```diff
@@ -0,0 +1,151 @@
...
+package ttnpb
+
+import (
+	"fmt"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/types"
+)
+
+// FieldIsZero returns whether path p is zero.
+func (v *JoinRequest) FieldIsZero(p string) bool {
```
johanstokking, line 24: This doesn't look very maintainable. How are we going to do this?

htdvisser (commented):
As @johanstokking also wrote, hand-writing those `FieldIsZero` funcs is not going to be very maintainable. Can we perhaps generate those funcs?

rvolosatovs:
@htdvisser @johanstokking please see https://github.com/TheThingsNetwork/lorawan-stack/pull/3647#discussion_r553969216 I already wrote when opening the PR
> Ideally these methods should be generated, but that should be done after #2798 is solved

pkg/ttnpb/join.go
rvolosatovs: Was already answered in https://github.com/TheThingsNetwork/lorawan-stack/pull/3647#discussion_r553969216

pkg/rpcmiddleware/validator/validator.go
rvolosatovs: > I think it makes sense to check the nearest FieldIsZero of sp here.

1. Since these methods will be auto-generated(see https://github.com/TheThingsNetwork/lorawan-stack/pull/3647#discussion_r553969216), all relevant messages will have this implemented
2. We do not have a generic way to access value of type `T` under arbitrary path `a.b.c`, so that is not possible unless we loop over all field tree in the struct using reflection and I don't think we should do that for all RPCs on each call.
3. `FieldIsZero` check is done inside the loop to avoid unnecessary type-assert on each RPC. This check is only performed when there's a path, which needs to be zero - there's no reason to check `FieldIsZero`, if we don't need to call it.

rvolosatovs:
blocking https://github.com/TheThingsIndustries/lorawan-stack/issues/2417

johanstokking (commented):
In general I'm okay with the current state.

I'm assuming that as long as the struct is part of the test table and the list of all fields is up to date, there's a guard against forgetting the zero checks per field. Right?

---

Just out of curiosity; why can't we do this with reflection? We would have to look at the request type, traverse the fields recursively, and once the field is found, see which type it is and check if the value is zero.

rvolosatovs:
> I'm assuming that as long as the struct is part of the test table and the list of all fields is up to date, there's a guard against forgetting the zero checks per field. Right?

Yes

> Just out of curiosity; why can't we do this with reflection? We would have to look at the request type, traverse the fields recursively, and once the field is found, see which type it is and check if the value is zero.

1. The traversing part is not trivial and would depend on custom gogo logic even with reflection, since the path we get are protobuf paths and we have to convert them to Go paths. Unless #2798 is done I think it's wasted effort to implement it this way
2. Reflection is slow and we'd do this on each RPC with potentially 100s of fields to validate.
3. Generating would be simpler and more efficient, but requires #2798

johanstokking (approved):
OK I'm fine with this, thanks for clarifying @rvolosatovs

htdvisser (approved):

Issue #3606, opened 2020-12-21
End-to-end test coverage


kschiffer:
#### Summary

This umbrella issue keeps track of our end-to-end test coverage.

*   [ ] Applications
    *   [x] CRUD #3650
    *   [x] Checking overview (information displayed as expected) #3663
    *   [x] Payload formatters (set all types, remove) #3841 
    *   [ ] Integrations
        *   [ ] MQTT (Check credentials)
        *   [x] Webhooks (CRUD, with and without webhook template)
        *   [x] Pub/sub (CRUD) #3720
    *   [x] Collaborators (CRUD) #3443, #3646
    *   [x] API Keys (CRUD) #3452
*   [x] End devices
    *   [x] CRUD 
        *   [x] Create
        *   [x] Read
        *   [x] Update
        *   [x] Delete
    *   [x] Import
    *   [x] Checking overview (information displayed as expected) #3663
    *   [x] Scheduling downlinks #3883
    *   [x] Location (setting and removing) @asmulko on this atm
    *   [x] Payload formatters (set types, remove, check precedence over app level PFs) #3841 
    *   [x] Claim authentication code (setting and removing)
    *   [x] General Settings (test various sub-forms, removing gateway)
*   [ ] Gateways
    *   [x] CRUD #3672
    *   [x] Checking overview (information displayed as expected) #3663
    *   [x] Location (setting and removing) #3681
        *   [ ] Confirm updating via status message
    *   [x] Collaborators (CRUD) #3443, #3646
    *   [x] API Keys (CRUD) #3452
*   [x] Organizations
    *   [x] CRUD #3670
    *   [x] Checking overview (information displayed as expected) #3663
    *   [x] Collaborators (CRUD) #3443, #3646
    *   [x] API Keys (CRUD) #3452
*   [x] Authentication and authorization
    *   [x] Login
    *   [x] Logout (Console and OAuth app)
    *   [x] User registration
    *   [x] Change password (when authenticated)
    *   [x] Change password (using temporary password)
    *   [x] Authorizing an OAuth client
    *   [x] Verifying contact info
    *   [x] Retrieving authorization code (`/oauth/code`)
*   [ ] Miscellaneous
    *   [ ] Feature toggles
        *   [ ] Check various rights setups and confirm:
            *   [ ] Correct visibility of items
            *   [ ] Correct write restriction of items
    *   [ ] Data / Event streams (all entities)
[...]

PR #3663, opened 2021-01-13
Add overview e2e tests

asmulko:
#### Summary

References #3606 

#### Changes

- Added overview tests for applications, gateways, organizations

#### Checklist

### DIFF
--- a/cypress/integration/console/shared/overview/overview.spec.js
+++ b/cypress/integration/console/shared/overview/overview.spec.js
@@ -1,0 +1,206 @@
+// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+describe('Overview', () => {
+  const userId = 'main-overview-test-user'
+  const user = {
+    ids: { user_id: userId },
+    primary_email_address: 'view-overview-test-user@example.com',
+    password: 'ABCDefg123!',
+    password_confirm: 'ABCDefg123!',
+  }
+
+  before(() => {
+    cy.dropAndSeedDatabase()
+    cy.createUser(user)
+  })
+
+  describe('Application', () => {
+    const applicationId = 'app-overview-test'
+    const application = {
+      ids: { application_id: applicationId },
+      name: 'Application Test Name',
+      description: `Application Test Description`,
+    }
+
+    before(() => {
+      cy.loginConsole({ user_id: userId, password: user.password })
+      cy.createApplication(application, userId)
+      cy.clearLocalStorage()
+      cy.clearCookies()
+    })
+
+    beforeEach(() => {
+      cy.loginConsole({ user_id: user.ids.user_id, password: user.password })
+    })
+
+    it('displays UI elements in place', () => {
+      cy.visit(`${Cypress.config('consoleRootPath')}/applications/${applicationId}`)
+
+      cy.findByText(`${application.name}`, { selector: 'h1' }).should('be.visible')
+      cy.findByRole('link', { name: /0 End device/ }).should('be.visible')
+      cy.findByRole('link', { name: /1 Collaborator/ }).should('be.visible')
+      cy.findByRole('link', { name: /0 API key/ }).should('be.visible')
+
+      cy.findByRole('row', { name: new RegExp(applicationId) }).should('be.visible')
+
+      cy.findByTestId('events-widget').should('be.visible')
+      cy.findByTestId('devices-table').should('be.visible')
+
+      cy.findByTestId('error-notification').should('not.exist')
+    })
+  })
+
+  describe('Gateway', () => {
+    const gatewayId = 'gateway-overview-test'
+    const gateway = {
+      ids: { gateway_id: gatewayId, eui: '0000000000000000' },
+      name: 'Gateway Test Name',
+      description: 'Gateway Test Description',
+      gateway_server_address: 'test-address',
+      frequency_plan_id: 'EU_863_870',
+    }
+
+    before(() => {
+      cy.loginConsole({ user_id: userId, password: user.password })
+      cy.createGateway(gateway, userId)
+      cy.clearLocalStorage()
+      cy.clearCookies()
+    })
+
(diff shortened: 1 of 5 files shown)


cypress/integration/console/shared/overview/overview.spec.js
```diff
@@ -0,0 +1,168 @@
...
+      cy.findByRole('columnheader', { name: 'Last updated at' }).should('be.visible')
+
+      // Events
+      cy.findByRole('link', { name: /See all activity/ }).should('be.visible')
+      cy.findByRole('link', { name: `Waiting for events from ${applicationId}…` })
+
+      // End Device
+      cy.findByRole('link', { name: /Import end devices/ }).should('be.visible')
+      cy.findByRole('link', { name: /Add end device/ }).should('be.visible')
+    })
```
kschiffer, line 62: Please also assert that there's not error notification anywhere.

cypress/integration/console/shared/overview/overview.spec.js
```diff
@@ -0,0 +1,168 @@
...
+      // Title
+      cy.findByText(`${application.name}`, { selector: 'h1' }).should('be.visible')
+      cy.findByRole('link', { name: /End device/ }).should('be.visible')
+      cy.findByRole('link', { name: /Collaborator/ }).should('be.visible')
+      cy.findByRole('link', { name: /API key/ }).should('be.visible')
+
+      // General Information
+      cy.findByRole('columnheader', { name: 'General information' }).should('be.visible')
+      cy.findByRole('columnheader', { name: 'Application ID' }).should('be.visible')
+      cy.findByRole('cell', { name: /app-overview-test/ }).should('be.visible')
```
kschiffer:
```suggestion
      cy.findByRole('cell', { name: /app-overview-test/ }).should('be.visible')
      cy.findByRole('cell', { name: /Application Test Name/ }).should('be.visible')
```

Also consider `new RegExp(application.ids.application_id)` instead.

cypress/integration/console/shared/overview/overview.spec.js
```diff
@@ -0,0 +1,168 @@
...
+      cy.visit(`${Cypress.config('consoleRootPath')}/applications/${applicationId}`)
+
+      // Title
+      cy.findByText(`${application.name}`, { selector: 'h1' }).should('be.visible')
+      cy.findByRole('link', { name: /End device/ }).should('be.visible')
+      cy.findByRole('link', { name: /Collaborator/ }).should('be.visible')
+      cy.findByRole('link', { name: /API key/ }).should('be.visible')
+
+      // General Information
+      cy.findByRole('columnheader', { name: 'General information' }).should('be.visible')
```
kschiffer: Take care not to be too stiff about display assertions.

I think it's not really relevant for the test case whether this is in a `columnheader` or `cell`. Likewise, I'd suggest not asserting on column header text such as `Application ID` or `Created at`. This is detail for me, meaning that the asserting on the exact text here is irrelevant, similarly to how the user does also not assert to find a text that says exactly `General information` on the overview page, instead he/she asserts to find concrete values of the application somewhere.

It's not always easy to find the right balance with this but please think of ways to avoid us having to refactor tests for minor changes.

cypress/integration/console/shared/overview/overview.spec.js
```diff
@@ -0,0 +1,168 @@
...
+      cy.findByRole('columnheader', { name: 'Application ID' }).should('be.visible')
+      cy.findByRole('cell', { name: /app-overview-test/ }).should('be.visible')
+      cy.findByRole('columnheader', { name: 'Created at' }).should('be.visible')
+      cy.findByRole('columnheader', { name: 'Last updated at' }).should('be.visible')
```
kschiffer: These are not relevant enough. If anything, you should check whether the correct dates are displayed (but I also think this is not necessary).

cypress/integration/console/shared/overview/overview.spec.js
```diff
@@ -0,0 +1,168 @@
...
+
+      // Events
+      cy.findByRole('link', { name: /See all activity/ }).should('be.visible')
+      cy.findByRole('link', { name: `Waiting for events from ${applicationId}…` })
```
kschiffer: Instead of checking these texts, you can use a `data-test-id="event-widget"` to make sure that the event widget is displayed, which is what we actually care about with regards to the test case, as opposed to checking that some exact texts are rendered.

cypress/integration/console/shared/overview/overview.spec.js
kschiffer: Also, admittedly, we lack a decision of a clear guideline on how to write such \`displays UI elements in place\` test cases.

cypress/integration/console/shared/overview/overview.spec.js
```diff
@@ -0,0 +1,168 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
```
kschiffer:
```suggestion
// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
```

kschiffer (changes requested):

kschiffer:
What about end devices?

cypress/integration/console/shared/overview/overview.spec.js
```diff
@@ -0,0 +1,208 @@
...
+      cy.loginConsole({ user_id: user.ids.user_id, password: user.password })
+    })
+
+    it('displays UI elements in place', () => {
+      cy.visit(`${Cypress.config('consoleRootPath')}/applications/${applicationId}`)
+
+      cy.findByText(`${application.name}`, { selector: 'h1' }).should('be.visible')
+      cy.findByRole('link', { name: /End device/ }).should('be.visible')
+      cy.findByRole('link', { name: /Collaborator/ }).should('be.visible')
+      cy.findByRole('link', { name: /API key/ }).should('be.visible')
```
kschiffer: Let's also add a check that the correct numbers are displayed for these.

cypress/integration/console/shared/overview/overview.spec.js
```diff
@@ -0,0 +1,208 @@
...
+  describe('End Devices', () => {
+    const applicationId = 'app-end-devices-overview-test'
+    const application = {
+      ids: { application_id: applicationId },
+      name: 'Application End Devices Test Name',
+      description: `Application End Devices Test Description`,
+    }
+
+    const endDeviceId = 'end-device-overview-test'
+    const endDevice = {
```
kschiffer, line 148:
```suggestion
    const endDeviceRequestBody = {
```

Maybe cleaner to split this up:

```
const endDevice = { ... }
const endDeviceFieldMask = { ... }
const endDeviceRequestBody = {
  end_device: endDevice,
  field_mask: endDeviceFieldMask,
}
```

And maybe also worth thinking about using the SDK for these use cases. (For later concern)

kschiffer (changes requested):

kschiffer (approved):
Good work. Thank you!

Issue #3577, opened 2020-12-09
Review length limits of API fields


htdvisser:
#### Summary

We need to review the length limits of API fields that can be set in request messages.

#### What is missing? What do you want to see?

- Maximum length on all `string` and `bytes` fields
- Maximum length on all `map` fields
- Maximum length on all `repeated` fields

#### How do you propose to implement this?

For example, on `attributes`:

```
(validate.rules).map = {
  max_pairs: 10,
  keys: {
    string: {
      pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$",
      max_len: 36
    }
  },
  values: {
    string: {
      max_len: 200
    }
  }
}
```

And on `contact_info`:

```
(validate.rules).repeated = {
  max_items: 10,
}
```

On `rights`:

```
(validate.rules).repeated = {
  unique: true,
  items: {
    enum: {
      defined_only: true
    }
  }
}
```

#### Can you do this yourself and submit a Pull Request?

I'll make a start, and assign others for their "owned" APIs.

htdvisser:
I think this is good now.

PR #3669, opened 2021-01-15
Make max length validation more strict

htdvisser:
#### Summary

This pull request addresses #3577 by adding max length validations on `string`, `bytes` fields, as well as `maps` and `repeated` fields.

#### Testing

For most fields I validated that our production databases do not contain any values that are even close to these limits.

##### Regressions

This could break consistency with validation rules in clients such as the Console.

#### Notes for Reviewers

@johanstokking please see if any additional limits need to be set in JS and EDTC API.

@rvolosatovs please see if any additional limits need to be set in NS (EndDevice) APIs.

#### Checklist

- ⚠️ Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
	- As with anything relating to API this could break clients that rely on buggy (non validated) behavior.
- ⚠️ Documentation: Relevant documentation is added or updated.
	- This will be updated the next time we sync API descriptors to the documentation site


htdvisser:
@kschiffer or @bafonins: please take a look and see if any changes are needed in the Console as a result of these changes. If not, please comment and remove your review request.

api/gateway.proto
```diff
@@ -244,7 +259,13 @@ message GatewayAntenna {
   float gain = 1;
   // location is the antenna's location.
   Location location = 2 [(gogoproto.nullable) = false, (validate.rules).message.required = true];
-  map<string,string> attributes = 3 [(validate.rules).map.keys.string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$" , max_len: 36}];
+  map<string,string> attributes = 3 [
+    (validate.rules).map = {
+      max_pairs: 10,
```
kschiffer, line 271: Why only 10?

htdvisser: Like any other limit we set, it's pretty arbitrary. In our production databases we don't see any entities with more than 10 attributes. Most entities don't have any attributes. Entities that do have attributes, have between 1 and 6. So let's turn the question around: Why more than 10?

kschiffer: I guess for how we use attributes currently it's fine. I remember some discussion around using labels e.g. for tags, in which case I can see more than 10 attributes being beneficial. As long as we can increase the limit later on, it's fine I'd say.

kschiffer:
I went through and pushed some validation rule updates accordingly.

rvolosatovs (approved):

bafonins (approved):

htdvisser:
I'll merge this with a failing Go test, because that one comes from https://github.com/TheThingsNetwork/lorawan-stack/pull/3679

Issue #3489, opened 2020-11-17
Refactor fetch package


johanstokking:
#### Summary

Refactor the common `pkg/fetch`.

#### Why do we need this?

To allow for fallback to another source, i.e. when a web server goes down. Another source could be another web location, or a local directory, or any other source.

#### What is already there? What do you see now?

Fetching is currently configured for a single source, and the source itself doesn't support a fallback mechanism.

The HTTP fetcher uses a memory cache transport.

#### What is missing? What do you want to see?

We need to specify multiple alternative HTTP locations that are tried sequentially, randomly or concurrently. We would use this to access geographically distributed or otherwise mirrored data sources that contain the same content.

We also need to be able to specify multiple layers that are tried, for example, first any HTTP source, then a local directory, then a cache with potentially outdated data.

If the HTTP source is configured with multiple URLs, all would fail or time-out in order to try a fallback layer.

Whenever something gets fetched from the primary layer, the cache would be updated.

Keep the compatibility commitment for configuration in mind; all configuration is forward-compatible.

#### How do you propose to test this?

Unit testing with multiple HTTP test servers and local `testdata` dir. We need different scenarios tested.

#### Can you do this yourself and submit a Pull Request?

Can review

neoaggelos:
Self-note; let's consider adding a `FileReader(paths ...string) (io.Reader, error)` method, which should improve handling for large files.

johanstokking:
> Self-note; let's consider adding a `FileReader(paths ...string) (io.Reader, error)` method, which should improve handling for large files.

What would be the purpose of this?

htdvisser:
Before starting work on this, please also read the draft design for the `io.FS` interface that's coming in Go 1.16: https://go.googlesource.com/proposal/+/master/design/draft-iofs.md

The expectation is that we'll soon see several implementations (OS filesystem, HTTP filesystem, S3 filesystem, ...) that implement that interface, so probably smart to go with something similar.

PR #3671, opened 2021-01-18
Add submodules for external repositories

htdvisser:
#### Summary

This pull request adds Git submodules for external repositories that were previously fetched from GitHub at runtime using `pkg/fetch`. The goal of this PR is to have a default/fallback for these repositories in our release images. We expect that production deployments will instead use (read-only) volume mounts for each repository, that are periodically updated by an external (cron) job.

Related to #3489 and the discussion in https://github.com/TheThingsNetwork/lorawan-stack/pull/3543#discussion_r558066489.

#### Changes

- Add submodules for frequency plans, device repository and webhook templates
- Include repository contents in Docker images
- Adapt tooling, CI, documentation

#### Testing

- [x] Will test Mage tooling when the next PR is merged to the device repository
- [x] Will test Docker images
- [ ] CI will happen after merge, as my fork is currently not ready for testing

##### Regressions

Existing deployments that use the default configuration (with fetching from GitHub), will now only get updates to external repos on new releases.

#### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,12 @@ RUN chmod 755 /bin/ttn-lw-stack
 COPY ttn-lw-cli /bin/ttn-lw-cli
 RUN chmod 755 /bin/ttn-lw-cli
 
+COPY data/lorawan-frequency-plans /srv/ttn-lorawan/lorawan-frequency-plans
+RUN rm -rf /srv/ttn-lorawan/lorawan-frequency-plans/.git
+
+COPY data/lorawan-webhook-templates /srv/ttn-lorawan/lorawan-webhook-templates
+RUN rm -rf /srv/ttn-lorawan/lorawan-webhook-templates/.git
+
 FROM alpine:3.12
 
 RUN addgroup -g 886 thethings && adduser -u 886 -S -G thethings thethings
@@ -19,6 +25,9 @@ COPY --from=builder /bin/ttn-lw-cli /bin/ttn-lw-cli
 RUN ln -s /bin/ttn-lw-cli /bin/cli
 
 COPY public /srv/ttn-lorawan/public
+
+COPY --from=builder /srv/ttn-lorawan/lorawan-frequency-plans /srv/ttn-lorawan/lorawan-frequency-plans
+COPY --from=builder /srv/ttn-lorawan/lorawan-webhook-templates /srv/ttn-lorawan/lorawan-webhook-templates
 
 EXPOSE 1700/udp 1881 8881 1882 8882 1883 8883 1884 8884 1885 8885 1887 8887
 
--- a/.github/ISSUE_TEMPLATE/release.md
+++ b/.github/ISSUE_TEMPLATE/release.md
@@ -20,6 +20,13 @@ This is a checklist for releases. This is filled in by both the releaser and the
   $ git checkout -b release/v3.${minor}.${patch}
   ```
 
+- [ ] Update submodules
+  ```bash
+  $ mage git:pullSubmodules
+  $ git add data
+  $ git commit -m "data: Update external repositories" # if any changes.
+  ```
+
 - [ ] Update the `CHANGELOG.md` file
   - [ ] Change the **Unreleased** section to the new version and add date obtained via `date +%Y-%m-%d` (e.g. `## [3.2.1] - 2019-10-11`)
   - [ ] Check if we didn't forget anything important
--- a/.github/workflows/release-snapshot.yml
+++ b/.github/workflows/release-snapshot.yml
@@ -17,6 +17,7 @@ jobs:
       uses: actions/checkout@v2
       with:
         fetch-depth: 0
+        submodules: true
     - name: Login to Docker Hub
       uses: docker/login-action@v1
       with:
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -16,6 +16,9 @@ jobs:
     steps:
     - name: Check out code
       uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        submodules: true
     - name: Login to Docker Hub
       uses: docker/login-action@v1
       with:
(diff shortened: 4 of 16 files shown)


Dockerfile
```diff
@@ -20,13 +29,19 @@ RUN ln -s /bin/ttn-lw-cli /bin/cli
...
+
 EXPOSE 1700/udp 1881 8881 1882 8882 1883 8883 1884 8884 1885 8885 1887 8887
 
 RUN mkdir /srv/ttn-lorawan/public/blob
 
 VOLUME ["/srv/ttn-lorawan/public/blob"]
 
-ENV TTN_LW_BLOB_LOCAL_DIRECTORY=/srv/ttn-lorawan/public/blob \
+ENV TTN_LW_AS_WEBHOOKS_TEMPLATES_URL=/srv/ttn-lorawan/lorawan-webhook-templates \
+    TTN_LW_FREQUENCY_PLANS_URL=/srv/ttn-lorawan/lorawan-frequency-plans \
```
neoaggelos: This should be `TTN_LW_FREQUENCY_PLANS_DIRECTORY`, right?

htdvisser:
We probably need to discuss whether we want those environment variables in the Dockerfile in the first place. Setting environment variables means that the values from config files get ignored, as environment takes precedence. We've already experienced this issue before with the other options that we set in environment. This technically means that we're breaking config compatibility for Docker deployments, so I'm afraid we can't really do this after all.

What we could do instead is to set defaults for the `directory` config option in The Things Stack, update the "detection mechanism" to check if that directory exists before falling back to to the `url`. What do you guys think?

rvolosatovs (approved):

rvolosatovs:
> We probably need to discuss whether we want those environment variables in the Dockerfile in the first place. Setting environment variables means that the values from config files get ignored, as environment takes precedence. We've already experienced this issue before with the other options that we set in environment. This technically means that we're breaking config compatibility for Docker deployments, so I'm afraid we can't really do this after all.
> 
> What we could do instead is to set defaults for the `directory` config option in The Things Stack, update the "detection mechanism" to check if that directory exists before falling back to to the `url`. What do you guys think?

1. We should instruct users to use env variables for Docker deployments, which is IMO already good enough, as long as everything is documented
2. Do we support multiple config files? If so, a simple solution would be to include a default config file in the image setting these, such that the user could load more config files or set env and those would take precedence in case of a conflict.

johanstokking (approved):

johanstokking:
> * We should instruct users to use env variables for Docker deployments, which is IMO already good enough, as long as everything is documented

That's a complete change of the current [Getting Started](https://www.thethingsindustries.com/docs/getting-started/installation/configuration/).

Also, I don't think we should prefer env over yaml for Docker deployments.

> * Do we support multiple config files? If so, a simple solution would be to include a default config file in the image setting these, such that the user could load more config files or set env and those would take precedence in case of a conflict.

I'm quite sure we don't support multiple config files.

---

We might need to do this at runtime; make the default value dependent on the discovered environment.

rvolosatovs:
Why not just have multiple configs and have the last ones take precedence? That's how a huge percentage of ecosystem works as well.

johanstokking:
> Why not just have multiple configs and have the last ones take precedence? That's how a huge percentage of ecosystem works as well.

Actually we do support this: https://github.com/TheThingsNetwork/lorawan-stack/blob/477f990239e30f906d80c538bdd61673e7fad890/pkg/config/config.go#L266-L283

So that means that things like `--config user.yaml --config docker.yaml` should work. This "reverse order" thing that we have is wrong and should change in V4, but we need to get rid of `pkg/config` anyway in the next major.

We would need to add a config that's always in the image, after passing the mounted config file from the user:

https://github.com/TheThingsNetwork/lorawan-stack/blob/fdbc4f2a0b9eda375df25c4621b951d18174ef50/docker-compose.yml#L39

htdvisser:
@rvolosatovs @johanstokking I've removed the ENV from the Dockerfile, since I believe this would violate our compatibility commitment regarding config files. Instead, I've updated the "source detection mechanism" (which is deprecated, but still used by default) to check directory presence before falling back to the next option. When an existing config file specifies a directory, that takes precedence, and therefore this would not violate our compatibility commitment. When a source is explicitly configured, the detection mechanism is bypassed anyway.

https://github.com/TheThingsNetwork/lorawan-stack/pull/3671/files/381d05efd73969d600a1e9b641f9e9569a41bd56..4a423e33ba87f85bee3a147036b5b48ba4d169bf

rvolosatovs (approved):
I still think that simply passing the necessary config to the binary as the last element is a way cleaner solution

neoaggelos (commented):
I think the submodules should also be pulled for the end to end tests workflow (at least for the Device Repository they are needed).

htdvisser:
@neoaggelos the e2e tests seemed to be happy, so apparently it's not necessary in this PR. If in a future PR our CI relies on submodules, then they can be initialized in that PR.

johanstokking (approved):
Only reviewed https://github.com/TheThingsNetwork/lorawan-stack/pull/3671/files/381d05efd73969d600a1e9b641f9e9569a41bd56..4a423e33ba87f85bee3a147036b5b48ba4d169bf now

Dockerfile
```diff
@@ -6,6 +6,15 @@ RUN chmod 755 /bin/ttn-lw-stack
 COPY ttn-lw-cli /bin/ttn-lw-cli
 RUN chmod 755 /bin/ttn-lw-cli
 
+COPY data/lorawan-devices /srv/ttn-lorawan/lorawan-devices
+RUN rm -rf /srv/ttn-lorawan/lorawan-devices/.git
```
neoaggelos: This should also be avoided. The `lorawan-devices` repo contains lots of image assets that we do **not** want to bundle in release images.

EDITED: do not

htdvisser: Do want or do not want?

This just deletes the Git history, which we don't need in release images.

neoaggelos: Do not add `lorawan-devices` in the Docker image at all. Copying only the required assets and initializing the index is handled in #3543

PR #3676, opened 2021-01-19
Add option to disable user registrations

htdvisser:
#### Summary

This pull request adds a config option to enable/disable user registrations.

#### Why do we need this?

In some networks, users are created by admin users, or in case of The Things Stack Enterprise, are created using the OIDC integration. In such cases, you'd want to be able to disable regular (unauthenticated) user registrations.

#### Changes

- Add config option
- Check config on user create

#### Testing

Unit testing

#### Notes for Reviewers

@kschiffer how much work would it be to hide the "Create an account" from the login screen?

@kschiffer also please consider this in your Account App work.

#### Checklist

	- [x] Will open PR


kschiffer:
> [**@kschiffer**](https://github.com/kschiffer) how much work would it be to hide the "Create an account" from the login screen?

As far as I can see, this is part of the exposed IS configuration, which the frontend can access. So it's not a big deal. I can push this to this PR in a bit.

htdvisser:
@kschiffer it wasn't. Just I added a commit to make it part of the IS config response.

kschiffer:
I've updated the OAuth app to not show the user registration option when the IS is configured accordingly. Instead of using the exposed IS config, I think it's actually better to pass this information directly through the template. Otherwise, it's a bit weird having to launch a separate API request to render the login page.

We might actually want to pass the password requirement (or even the whole IS configuration object) like that as well. But that's for later concern.

You can also revert [606523c](https://github.com/TheThingsNetwork/lorawan-stack/pull/3676/commits/606523cecaa306855c55d50cb0fb359ebfc993c8) if you want to.

kschiffer (approved):

bafonins (approved):

johanstokking (approved):

PR #3679, opened 2021-01-19
Optimize join-accept storage

rvolosatovs:
#### Summary

Optimize join-accept storage

#### Changes

- Only store necessary data

#### Testing

Should be covered by existing unit tests

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [x] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #3680, opened 2021-01-19
Allow session import in Network Server

### LINKED ISSUE #2472: Allow creation of devices with preexisting MAC state in Network Server
#### Summary

Allow creation of devices with preexisting MAC state in Network Server

#### Why do we need this?

For easier migration support

MAC state gets generated and overwrites whatever is specified

Merging the generated MAC state and specified fields

Merge both

yes

### LINKED ISSUE #4000: Allow selective MAC state overrides on creation
#### Summary

Allow users to create devices with MAC state by only partially selecting the "overrides" of the state and let NS merge generate the rest and merge.

#### Why do we need this?

To simplify usability

Currently NS discards all fields in `mac_state` namespace and overrides with generated MAC state on creation

Generate MAC state, apply overrides

See above

Unit tests

yes

rvolosatovs:
#### Summary

Closes #2472
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4000

#### Changes

- Allow session import in Network Server

#### Testing

Unit tests, extensive integration testing by @neoaggelos and CyberEye (https://github.com/TheThingsIndustries/lorawan-stack-support/issues/357)

##### Regressions

Potentially create and update functionality could break, so we should do extensive testing here.

#### Notes for Reviewers

~I will split this PR into smaller parts~(#3919 )

#### Checklist


rvolosatovs:
Blocked by #3679

rvolosatovs:
Blocking TheThingsNetwork/lorawan-stack-migrate#14

rvolosatovs:
@neoaggelos I am still working on testing, but perhaps you could test https://github.com/TheThingsNetwork/lorawan-stack-migrate/pull/14 with this branch already and assist in the process?

pkg/networkserver/grpc_deviceregistry.go
```diff
@@ -275,392 +269,1752 @@ func (ns *NetworkServer) Get(ctx context.Context, req *ttnpb.GetEndDeviceRequest
...
+}
+
+func setKeyIsZero(m map[string]*ttnpb.EndDevice, get func(*ttnpb.EndDevice) *ttnpb.KeyEnvelope, path string) bool {
+	if dev, ok := m[path+".key"]; ok {
+		if ke := get(dev); !ke.Key.IsZero() {
+			return false
+		}
+	}
+	if dev, ok := m[path+".encrypted_key"]; ok {
+		if ke := get(dev); len(ke.EncryptedKey) == 0 {
```
neoaggelos: I think this should be the opposite?

```suggestion
		if ke := get(dev); len(ke.EncryptedKey) != 0 {
```

pkg/ttnpb/fieldmask_utils.go
```diff
@@ -222,6 +222,17 @@ func FieldsWithPrefix(prefix string, paths ...string) []string {
 	return ret
 }
 
+// FieldsWithoutPrefix returns the paths under prefix with said prefix removed.
+func FieldsWithoutPrefix(prefix string, paths ...string) []string {
+	ret := make([]string, 0, len(paths))
+	for _, p := range paths {
+		if i := strings.Index(p, prefix+"."); i >= 0 {
+			ret = append(ret, p[i+1:])
```
neoaggelos: Also this should be
```suggestion
			ret = append(ret, p[i+len(prefix)+1:])
```

pkg/networkserver/grpc_deviceregistry.go
```diff
@@ -275,392 +269,1752 @@ func (ns *NetworkServer) Get(ctx context.Context, req *ttnpb.GetEndDeviceRequest
...
+	if dev, ok := m[path+".encrypted_key"]; ok {
+		if ke := get(dev); len(ke.EncryptedKey) != 0 {
+			return false
+		}
+	}
+	return true
+}
+
+func setKeyEqual(m map[string]*ttnpb.EndDevice, getA, getB func(*ttnpb.EndDevice) *ttnpb.KeyEnvelope, pathA, pathB string) bool {
+	if a, b := getA(m[pathA+".key"]).GetKey(), getB(m[pathB+".key"]).GetKey(); a == nil && b != nil || a != nil && b == nil || !a.Equal(*b) {
```
neoaggelos:
```suggestion
	if a, b := getA(m[pathA+".key"]).GetKey(), getB(m[pathB+".key"]).GetKey(); a == nil && b != nil || a != nil && b == nil || b != nil && !a.Equal(*b) {
```

pkg/networkserver/grpc_deviceregistry.go
```diff
@@ -275,392 +269,1752 @@ func (ns *NetworkServer) Get(ctx context.Context, req *ttnpb.GetEndDeviceRequest
...
+	}
 
-		if req.EndDevice.LoRaWANVersion.Compare(ttnpb.MAC_V1_1) >= 0 {
-			if err := ttnpb.RequireFields(sets,
-				"session.keys.nwk_s_enc_key.key",
-				"session.keys.s_nwk_s_int_key.key",
-			); err != nil {
-				return nil, nil, errInvalidFieldMask.WithCause(err)
+	var evt events.Event
+	dev, ctx, err := ns.devices.SetByID(ctx, st.Device.EndDeviceIdentifiers.ApplicationIdentifiers, st.Device.EndDeviceIdentifiers.DeviceID, st.GetFields(), st.SetFunc(func(ctx context.Context, stored *ttnpb.EndDevice) error {
```
neoaggelos, line 2194: In its current state, this fails for devices without a session with the following error:

```
 ERROR Failed to set device in registry         error=error:pkg/redis:store (store error) error_cause='ping_slot_frequency' has no subfields, but [value] were specified grpc.method=Set grpc.service=ttn.lorawan.v3.NsEndDeviceRegistry namespace=grpc peer.address=[::1]:55570 peer.real_ip=::1 request_id=01EXTCNTRKTPFPG16B8HPA28NR

```

After investigating a little, this is returned in https://github.com/rvolosatovs/lorawan-stack/blob/edc7d17780db04fc8fbfa252467cbabc8cf7bd85/pkg/networkserver/redis/registry.go#L726

because `gets` contains `mac_settings.ping_slot_frequency.value`, but `updated.MACSettings.PingSlotFrequency` is `nil`, and the pb setter does not like that. 

Any fix may also need to be applied in lines 541 and 563 as well.

rvolosatovs: Can you provide a device model I can reproduce it with? I'm able to create devices with no session just fine locally.
Also, `mac_settings.ping_slot_frequency.value` is not a valid path.

neoaggelos: This is an example that fails:

```
{"ids":{"device_id":"uno","application_ids":{"application_id":"akolaitis-testing"},"dev_eui":"0101010101010101","join_eui":"0101010101010101"},"created_at":"0001-01-01T00:00:00Z","updated_at":"0001-01-01T00:00:00Z","name":"uno","lorawan_version":"1.0.2","lorawan_phy_version":"1.0.2-b","frequency_plan_id":"EU_863_870_TTN","supports_join":true,"root_keys":{"app_key":{"key":"01010101010101010101010101010101"}},"mac_settings":{"supports_32_bit_f_cnt": true}}
```

It works if no MAC settings are set:

```
{"ids":{"device_id":"uno","application_ids":{"application_id":"akolaitis-testing"},"dev_eui":"0101010101010101","join_eui":"0101010101010101"},"created_at":"0001-01-01T00:00:00Z","updated_at":"0001-01-01T00:00:00Z","name":"uno","lorawan_version":"1.0.2","lorawan_phy_version":"1.0.2-b","frequency_plan_id":"EU_863_870_TTN","supports_join":true,"root_keys":{"app_key":{"key":"01010101010101010101010101010101"}}}
```

rvolosatovs:
Setting prio to high, because we have customers waiting for this.

rvolosatovs:
Creation of devices should now be fully working and tested, next will be the update unit tests, possible bugfixes and little missing details. @neoaggelos can you test again?

rvolosatovs:
By giving this another look, I don't think we need a minor bump, since both gRPC and JSON APIs should actually be fully compatible. The only change is that now we allow more `value` subfields in fieldmasks.
@bafonins can you validate that long-deprecated `mac_state.desired_parameters.ping_slot_data_rate_index` (and current parameter analog of it) is not used by the console?

rvolosatovs:
Blocked by #3919

rvolosatovs:
@neoaggelos I haven't tested this with latest changes from #3919 , but this should be ready for testing.

rvolosatovs:
The implementation is ready to review
I will proceed with adding unit tests, meanwhile @neoaggelos could you coordinate integration testing (involving CyberEye as well)?
Apart from migration from v2, the more flows we test here, the better. That means, creating devices with various settings, as well as updating them.
The goal would be to ensure that only valid devices can be created here, which would save us a lot of support time and prevent a potential variety of weird bugs caused by (internal) device state being invalid

pkg/networkserver/grpc_deviceregistry.go
```diff
@@ -275,392 +269,1969 @@ func (ns *NetworkServer) Get(ctx context.Context, req *ttnpb.GetEndDeviceRequest
...
+	if dev, ok := m[path+".encrypted_key"]; ok {
+		if ke := get(dev); len(ke.EncryptedKey) != 0 {
+			return false
+		}
+	}
+	return true
+}
+
+func setKeyEqual(m map[string]*ttnpb.EndDevice, getA, getB func(*ttnpb.EndDevice) *ttnpb.KeyEnvelope, pathA, pathB string) bool {
+	if a, b := getA(m[pathA+".key"]).GetKey(), getB(m[pathB+".key"]).GetKey(); a == nil && b != nil || a != nil && b == nil || a != nil && b != nil && !a.Equal(*b) {
```
johanstokking, line 558:
```go
a == nil && b != nil || a != nil && b == nil
```

If only Go had a logical XOR operator

pkg/networkserver/grpc_deviceregistry.go
```diff
@@ -275,392 +269,1969 @@ func (ns *NetworkServer) Get(ctx context.Context, req *ttnpb.GetEndDeviceRequest
...
+		return false
+	}
+	if a, b := getA(m[pathA+".kek_label"]).GetKEKLabel(), getB(m[pathB+".kek_label"]).GetKEKLabel(); a != b {
+		return false
+	}
+	return true
+}
+
+// Set implements NsEndDeviceRegistryServer.
+func (ns *NetworkServer) Set(ctx context.Context, req *ttnpb.SetEndDeviceRequest) (*ttnpb.EndDevice, error) {
```
johanstokking, line 920: This one is now 1570 lines long without any comments for logical sections. Also, there are some long lists of field paths; how do we maintain those? Can't those be put in a `var` with a godoc with instructions on how to keep that list up to date (like what the rule of for having that path there)?

pkg/networkserver/grpc_deviceregistry.go
```diff
@@ -275,392 +269,1969 @@ func (ns *NetworkServer) Get(ctx context.Context, req *ttnpb.GetEndDeviceRequest
...
+		)
+	}
+
+	if err := st.ValidateIfNotZeroThenNotZero("supports_join",
+		"ids.join_eui",
+	); err != nil {
+		return nil, err
+	}
+
+	if err := st.ValidateIfZeroThenZero("supports_join",
```
johanstokking: Like the list below

pkg/networkserver/grpc_deviceregistry.go
```diff
@@ -275,392 +269,1969 @@ func (ns *NetworkServer) Get(ctx context.Context, req *ttnpb.GetEndDeviceRequest
...
+				"session.keys.s_nwk_s_int_key.kek_label",
+			)
+			getTransforms = append(getTransforms, func(dev *ttnpb.EndDevice) {
+				dev.Session.SNwkSIntKey = &ttnpb.KeyEnvelope{
+					Key: k,
+				}
+			})
+		}
+	}
+	if st.Device.PendingSession != nil {
```
johanstokking, line 1582: This is never `true` in tests

pkg/networkserver/grpc_deviceregistry.go
```diff
@@ -275,392 +269,1969 @@ func (ns *NetworkServer) Get(ctx context.Context, req *ttnpb.GetEndDeviceRequest
...
+				dev.PendingSession.SNwkSIntKey = &ttnpb.KeyEnvelope{
+					Key: k,
+				}
+			})
 		}
-		if req.EndDevice.PendingMACState != nil {
-			if ttnpb.HasAnyField(sets, "pending_mac_state.current_parameters.adr_data_rate_index") && req.EndDevice.PendingMACState.CurrentParameters.ADRDataRateIndex > phy.MaxADRDataRateIndex {
-				return nil, nil, errInvalidFieldValue.WithAttributes("field", "pending_mac_state.current_parameters.adr_data_rate_index")
+	}
+	if st.Device.PendingMACState.GetQueuedJoinAccept() != nil {
```
johanstokking, line 1648: This is never true in tests

pkg/networkserver/grpc_deviceregistry.go
```diff
@@ -275,392 +269,1969 @@ func (ns *NetworkServer) Get(ctx context.Context, req *ttnpb.GetEndDeviceRequest
...
+
+	var (
+		// hasPendingSession indicates whether the effective device model contains a non-zero pending session.
+		hasPendingSession bool
+
+		// hasQueuedJoinAccept indicates whether the effective device model contains a non-zero queued join-accept.
+		hasQueuedJoinAccept bool
+	)
+	if err := st.ValidateSetFields(func(m map[string]*ttnpb.EndDevice) (bool, string) {
+		var hasPendingMACState bool
```
johanstokking, line 1910: This callback is unreached in tests

pkg/networkserver/grpc_deviceregistry.go
rvolosatovs: https://github.com/TheThingsNetwork/lorawan-stack/pull/3680/commits/b2f531e1db27caebf8cd3ec4a3c4b7e4ae85fc16 what do you think about this?

johanstokking: Yes, that's much better

htdvisser (approved):
Nothing to add here.

rvolosatovs:
In it's current state, NS requires mandatory fields like `mac_state.lorawan_version` to be set on create, although we generate MAC state for non-OTAA devices and this one can be derived from top-level `lorawan_version`. Still have to figure out how to merge these nicely, meanwhile the migration should work fine with this, since it generates the MAC state client-side.

rvolosatovs:
FYI, it would be very nice to fix #2269 before we release this feature to prevent invalid devices being created in the registry

adriansmares (approved):

rvolosatovs:
@benolayinka please update the docs once this is merged

johanstokking (approved):

PR #3685, opened 2021-01-20
Allow resetting `<Select />` value

bafonins:
#### Summary

This PR allows resetting `<Select />` value to `null` when `undefined` or `''` is provided as value.

#### Changes

- Use arrow function
- Pass `null` when no value is provided

#### Notes for Reviewers

Noticed this when working on DR.

#### Checklist

### DIFF
--- a/pkg/webui/components/select/index.js
+++ b/pkg/webui/components/select/index.js
@@ -164,7 +164,7 @@ class Select extends React.PureComponent {
         className={cls}
         inputId={id}
         classNamePrefix="select"
-        value={getValue(translatedOptions, value)}
+        value={getValue(translatedOptions, value) || null}
         options={translatedOptions}
         onChange={this.onChange}
         onBlur={this.onBlur}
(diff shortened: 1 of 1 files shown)


pkg/webui/components/select/index.js
```diff
@@ -164,7 +164,7 @@ class Select extends React.PureComponent {
         className={cls}
         inputId={id}
         classNamePrefix="select"
-        value={getValue(translatedOptions, value)}
+        value={getValue(translatedOptions, value) || null}
```
kschiffer, line 167: In theory, `''` can be a valid value. Maybe it's better to use null for `undefined` only, [since that can also not be set properly by formik](https://github.com/formium/formik/issues/2332).

kschiffer (approved):

pkg/webui/components/select/index.js
bafonins: well, `''` can indeed be a valid value, but I dont think we should account for that. In any case where backend expects `''` for a certain field represented by `<Select />` a dummy/hardcoded value should be used and necessary conversion can be done in the submit function or validation schema.

Ill merge as it is now, if we decide to allow `''` in select it is pretty simple fix.

PR #3689, opened 2021-01-21
Support parsing of DevAddrPrefix type

KrishnaIyer:
#### Summary

Small fix needed for https://github.com/TheThingsIndustries/lorawan-stack-support/pull/279#discussion_r558924999

#### Changes

- Parse `DevAddrPrefix` type from flag.

#### Testing

Tested in staging. Ref; https://github.com/TheThingsIndustries/lorawan-stack-support/pull/279#discussion_r561746480

##### Regressions

NA

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/internal/util/flags.go
+++ b/cmd/ttn-lw-cli/internal/util/flags.go
@@ -28,6 +28,7 @@ import (
 	"github.com/spf13/pflag"
 	"go.thethings.network/lorawan-stack/v3/pkg/errors"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	ttstypes "go.thethings.network/lorawan-stack/v3/pkg/types"
 )
 
 // DeprecateFlag deprecates a CLI flag.
@@ -683,6 +684,16 @@ func setField(rv reflect.Value, path []string, v reflect.Value) error {
 									GatewayID: v.Index(i).String(),
 								},
 							}))
+						}
+					case ft.Elem().PkgPath() == "go.thethings.network/lorawan-stack/v3/pkg/types" &&
+						ft.Elem().Name() == "DevAddrPrefix" && vt.Elem().Kind() == reflect.String:
+						for i := 0; i < v.Len(); i++ {
+							var prefix ttstypes.DevAddrPrefix
+							err := prefix.UnmarshalText([]byte(v.Index(i).String()))
+							if err != nil {
+								return err
+							}
+							slice.Index(i).Set(reflect.ValueOf(prefix))
 						}
 					case ft.Elem().Kind() == reflect.Int32 && vt.Elem().Kind() == reflect.String:
 						if reflect.PtrTo(ft.Elem()).Implements(textUnmarshalerType) {
--- a/cmd/ttn-lw-cli/internal/util/flags.go
+++ b/cmd/ttn-lw-cli/internal/util/flags.go
@@ -28,7 +28,6 @@ import (
 	"github.com/spf13/pflag"
 	"go.thethings.network/lorawan-stack/v3/pkg/errors"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
-	ttstypes "go.thethings.network/lorawan-stack/v3/pkg/types"
 )
 
 // DeprecateFlag deprecates a CLI flag.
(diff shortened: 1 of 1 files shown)


htdvisser (dismissed):
To make this a bit more future-proof, I'd rather see this support any `encoding.TextUnmarshaler` type, like we do for enums below.

johanstokking (approved):

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -684,6 +684,14 @@ func setField(rv reflect.Value, path []string, v reflect.Value) error {
...
 						}
+					case ft.Elem().PkgPath() == "go.thethings.network/lorawan-stack/v3/pkg/types" &&
+						ft.Elem().Name() == "DevAddrPrefix" && vt.Elem().Kind() == reflect.String:
+						for i := 0; i < v.Len(); i++ {
+							err := slice.Index(i).Addr().Interface().(encoding.TextUnmarshaler).UnmarshalText([]byte(v.Index(i).String()))
+							if err != nil {
+								return err
+							}
+						}
 					case ft.Elem().Kind() == reflect.Int32 && vt.Elem().Kind() == reflect.String:
```
htdvisser: Can't we avoid the block above if we simply change this?

```suggestion
					case vt.Elem().Kind() == reflect.String:
```

KrishnaIyer: Yup. That works.

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -684,7 +684,7 @@ func setField(rv reflect.Value, path []string, v reflect.Value) error {
 								},
 							}))
 						}
-					case ft.Elem().Kind() == reflect.Int32 && vt.Elem().Kind() == reflect.String:
+					case vt.Elem().Kind() == reflect.String:
 						if reflect.PtrTo(ft.Elem()).Implements(textUnmarshalerType) {
```
htdvisser: Would be even nicer to pull that `if` into the `case`. And I guess we can then do the same on L526.

htdvisser (approved):
LGTM, but can be improved even more

cmd/ttn-lw-cli/internal/util/flags.go
rvolosatovs:
```suggestion
					case reflect.PtrTo(ft.Elem()).Implements(textUnmarshalerType) && vt.Elem().Kind() == reflect.String:
```
to be specific

KrishnaIyer: Done. The reason I didn't do this because we lose the `does not implement encoding.TextUnmarshaler` error and instead receive the default `is not convertible to` error.

cmd/ttn-lw-cli/internal/util/flags.go
```diff
@@ -522,14 +522,10 @@ func setField(rv reflect.Value, path []string, v reflect.Value) error {
...
 					field.Set(v)
-				case ft.Kind() == reflect.Int32 && vt.Kind() == reflect.String:
-					if reflect.PtrTo(ft).Implements(textUnmarshalerType) {
-						err := field.Addr().Interface().(encoding.TextUnmarshaler).UnmarshalText([]byte(v.String()))
-						if err != nil {
-							return err
-						}
-					} else {
-						return fmt.Errorf(`%s does not implement encoding.TextUnmarshaler`, ft.Name())
+				case ft.Kind() == reflect.Int32 && vt.Kind() == reflect.String && reflect.PtrTo(ft).Implements(textUnmarshalerType):
```
htdvisser: This `ft.Kind() == reflect.Int32` doesn't look quite right, as that would only make this work for enums and other `int32` types that implement `encoding.TextUnmarshaler`.

htdvisser (changes requested):

cmd/ttn-lw-cli/internal/util/flags.go
KrishnaIyer: Originally, I proposed a simple change that works for one specific case which won't affect others. After some review comments, I'm combining cases to make this more efficient. I don't really have much context on the regressions that this change may cause.
Can you provide a final suggestion on what can be done here? Should I just return to #3689 (review)?

htdvisser: I think after removing `ft.Kind() == reflect.Int32 &&` this should be fine.

htdvisser (approved):
LGTM.

As for regression testing:

- Make sure you can still update the state of an OAuth client (to verify that enum unmarshaling still works -- the L525 change stuff)
- Make sure you can still update the rights of an OAuth client (to verify that enum slice unmarshaling still works -- the L687/683 stuff)

KrishnaIyer:
Tested for regressions

```bash
$ ./ttn-lw-cli version                                                    
Version:             3.11.0-rc1-dev
Go version:          go1.15.7
OS/Arch:             darwin/amd64
```

```bash
$ ./ttn-lw-cli clients update console --rights RIGHT_ALL,RIGHT_ORGANIZATION_APPLICATIONS_LIST  --state STATE_APPROVED
{
  "ids": {
    "client_id": "console"
  },
  "created_at": "2021-02-08T12:54:53.439Z",
  "updated_at": "2021-02-08T13:07:34.890Z",
  "state": "STATE_APPROVED",
  "rights": [
    "RIGHT_ALL",
    "RIGHT_ORGANIZATION_APPLICATIONS_LIST"
  ]
}
```

PR #3701, opened 2021-01-22
Add profile settings to Account App

kschiffer:
#### Summary
References #1422 

This PR introduces profile settings to the Account App.

image

#### Changes
- Add profile settings view
  - Profile settings panel
    - Change user data and set/change profile pictures
    - Enable disable profile picture settings based on stack configuration
    - Delete the user account
  - Change password panel
- Add profile card to the AA overview page
- Update end-to-end tests
- Related fixes

#### Testing

Manual and e2e

##### Regressions

This is new functionality so it should not produce any regressions.

#### Checklist


htdvisser (commented):
code-owned files LGTM.

It is however important to note that the current implementation of the Identity Server only lets admins change the primary email address of a user, so this will not work for regular users.

In the future, the idea is that users can change their primary email address to any of the (verified) email addresses that they already have in their list of contact info, so before we can let them update their primary email address, we first need to let them manage their contact info.

htdvisser:
I created https://github.com/TheThingsNetwork/lorawan-stack/issues/3711 for changing the primary email address.

kschiffer:
> It is however important to note that the current implementation of the Identity Server only lets admins change the primary email address of a user, so this will not work for regular users.

Weird. I was able to change the email also as non-admin without problems. ~~Still, I've disabled the field now for non-admins.~~

htdvisser:
Looks like I was mistaken, and we actually allow changing it.

https://github.com/TheThingsNetwork/lorawan-stack/blob/c8e90b4a10ad29f40d6a0557a2067a7f94ae00ff/pkg/identityserver/user_registry.go#L360-L372

https://github.com/TheThingsNetwork/lorawan-stack/blob/c8e90b4a10ad29f40d6a0557a2067a7f94ae00ff/pkg/identityserver/user_registry.go#L408-L432

pkg/webui/lib/selectors/profile-picture.js
```diff
@@ -0,0 +1,71 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export const isValidProfilePictureObject = profilePicture =>
+  Boolean(profilePicture) && Boolean(profilePicture.sizes)
+
+export const getProfilePictureSizes = profilePicture => {
+  if (!isValidProfilePictureObject(profilePicture))
```
bafonins:
```suggestion
  if (!isValidProfilePictureObject(profilePicture)) {
    throw new Error('Invalid profile picture object')
  }
```

bafonins: Also, why throwing an error here and not returning an empty array? I dont think that throwing in a selector is a good idea.

sdk/js/src/service/users.js
```diff
@@ -100,7 +100,17 @@ class Users {
         field_mask: Marshaler.fieldMask(mask),
       },
     )
-    return Marshaler.unwrapUser(response)
+
+    const result = Marshaler.unwrapUser(response)
+
+    // Get new profile picture value if the old one was deleted
+    if (mask.includes('profile_picture') || mask.includes('primary_email_address')) {
```
bafonins, line 108: How is `primary_email_address` related here?

pkg/webui/account/containers/profile-settings-form/validation-schema.js
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        return schema
+      },
+    ),
+  name: Yup.string()
+    .min(2, Yup.passValues(sharedMessages.validateTooShort))
+    .max(50, Yup.passValues(sharedMessages.validateTooLong)),
+  primary_email_address: Yup.string()
+    .email(sharedMessages.validateEmail)
```
bafonins, line 48: Screenshot 2021-02-03 at 11 55 52

Error message in not really helpful here

pkg/webui/lib/selectors/profile-picture.js
```diff
@@ -0,0 +1,71 @@
...
+  return sizes[0]
+}
+
+export const getClosestProfilePictureBySize = (profilePicture, size) => {
+  const sizes = getProfilePictureSizes(profilePicture)
+  if (sizes[size]) {
+    return sizes[size]
+  }
+  if (sizes.length === 1) {
+    return Object.values(sizes)[0]
```
bafonins: This line returns a size instead of a profile picture

.goreleaser.release.yml
```diff
@@ -144,7 +144,7 @@ brews:
   - name: ttn-lw-stack
     ids:
       - stack
-    github:
+    tap:
```
bafonins: is this related to this PR?

api/applicationserver_web.proto
```diff
@@ -111,7 +111,7 @@ message ApplicationWebhook {
 
   // The API key to be used for downlink queue operations.
   // The field is provided for convenience reasons, and can contain API keys with additional rights (albeit this is discouraged).
-  string downlink_api_key = 17 [(gogoproto.customname) = "DownlinkAPIKey", (validate.rules).string.max_len = 64];
+  string downlink_api_key = 17 [(gogoproto.customname) = "DownlinkAPIKey", (validate.rules).string.max_len = 128];
```
bafonins: Is this change related to this pr?

pkg/webui/account/containers/profile-settings-form/index.js
```diff
@@ -0,0 +1,292 @@
...
+  const handleDelete = useCallback(async () => {
+    try {
+      await dispatch(attachPromise(deleteUser)())
+
+      // The hard redirect will conclude the deletion by deleting the
+      // (now invalid) session cookie and redirecting back to the login screen.
+      // The `account-deleted` query will cause a success notification to be
+      // shown on the login screen.
+      const appRoot = selectApplicationRootPath()
+      window.location = `${appRoot}/login?account-deleted`
```
bafonins, line 182: Not really robust approach. Redirect with a state seems to be more appropriate here

pkg/webui/account/containers/profile-settings-form/index.js
```diff
@@ -0,0 +1,292 @@
...
+  const useGravatarConfig = useSelector(selectUseGravatarConfiguration)
+  const disableUploadConfig = useSelector(selectDisableUploadConfiguration)
+  const initialProfilePictureSource =
+    useGravatarConfig && isGravatarProfilePicture(user.profile_picture) ? 'gravatar' : 'upload'
+
+  const [profilePictureSource, setProfilePictureSource] = useState(initialProfilePictureSource)
+  const [error, setError] = useState()
+  const [emailAddress, setEmailAddress] = useState(user.primary_email_address)
+
+  const { debouncedFunction: debouncedSetEmail, cancel: debounceCancel } = debounce(
```
bafonins: This creates a new debounced function on each rerender. Consider using `React.useMemo`

pkg/webui/account/containers/profile-settings-form/index.js
```diff
@@ -0,0 +1,292 @@
...
+    setEmailAddress,
+    500,
+  )
+
+  const validationContext = {
```
bafonins: Consider using `React.useMemo`.

pkg/webui/account/containers/profile-settings-form/index.js
```diff
@@ -0,0 +1,292 @@
...
+import style from './profile-settings-form.styl'
+
+const imageDataTransform = data => {
+  // Handle empty value (when file was removed)
+  if (data === '') return null
+
+  // Handle selected image file
+  const matches = data.match(/^data:(.*);base64,(.*)$/)
+  // Discard invalid values
+  if (matches.length < 3) return null
```
bafonins: `matches` can be `null`

pkg/webui/account/containers/profile-settings-form/index.js
```diff
@@ -0,0 +1,292 @@
...
+        setError(error)
+        setSubmitting(false)
+      }
+    },
+    [dispatch, user, validationContext],
+  )
+
+  const handleDelete = useCallback(async () => {
+    try {
+      await dispatch(attachPromise(deleteUser)())
```
bafonins: This looks weird

pkg/webui/components/profile-picture/index.js
```diff
@@ -0,0 +1,61 @@
...
+  getClosestProfilePictureBySize,
+  isValidProfilePictureObject,
+} from '@ttn-lw/lib/selectors/profile-picture'
+
+import styles from './profile-picture.styl'
+
+const ProfilePicture = ({ profilePicture, className, size }) => {
+  const imageRef = useRef()
+  const handleImageError = useCallback(error => {
+    error.target.src = missingPropfilePicture
```
bafonins:
```suggestion
    error.target.src = missingProfilePicture
```

pkg/webui/components/profile-picture/index.js
```diff
@@ -0,0 +1,61 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import React, { useRef, useCallback } from 'react'
+import classnames from 'classnames'
+
+import missingPropfilePicture from '@assets/img/placeholder/missing-profile-picture.png'
```
bafonins:
```suggestion
import missingProfilePicture from '@assets/img/placeholder/missing-profile-picture.png'
```

pkg/webui/lib/selectors/profile-picture.js
```diff
@@ -0,0 +1,71 @@
...
+  if (!isValidProfilePictureObject(profilePicture))
+    throw new Error('Invalid profile picture object')
+  return profilePicture.sizes
+}
+
+export const getLargestAvailableProfilePicture = profilePicture => {
+  const sizes = getProfilePictureSizes(profilePicture)
+  const sizeValues = Object.keys(profilePicture.sizes)
+
+  return sizes[0] || sizes[Math.max(...sizeValues)]
```
bafonins: isnt `sizes[0]` the original picture?

pkg/webui/lib/selectors/profile-picture.js
```diff
@@ -0,0 +1,71 @@
...
+export const getSmallestAvailableProfilePicture = profilePicture =>
+  getClosestProfilePictureBySize(profilePicture, 64)
+
+export const getOriginalSizeProfilePicture = profilePicture => {
+  const sizes = getProfilePictureSizes(profilePicture)
+
+  return sizes[0]
+}
+
+export const getClosestProfilePictureBySize = (profilePicture, size) => {
```
bafonins, line 35: Closest to `> 'size'` or `< 'size'` ?

bafonins (changes requested):
After successfully setting gravatar and refreshing the page its gone and no profile picture is shown in profile settings

sdk/js/src/service/users.js
kschiffer: In case of gravatar, the PP is attached to the `primary_email_address`, so if that changed, the profile picture changes as well. I will make the comment a bit clearer wrt this.

pkg/webui/account/containers/profile-settings-form/index.js
kschiffer: We cannot do that here because we need to reinitialize the AA with a refresh after the account was deleted and all sessions and access tokens were removed. This basically equals logging out, which we also do with a hard refresh, which to me is the safest way to reinitialize, as opposed to clearing all state manually.

.goreleaser.release.yml
kschiffer: No, rebasing issue...

pkg/webui/account/containers/profile-settings-form/index.js
kschiffer: Please be more constructive. "This looks weird" is not actionable.
```suggestion
      await dispatch(attachPromise(deleteUser()))
```
Would be an alternative here, it doesn't really matter.

bafonins: Here is more constructive suggestion:

Please consider this gist:
```js
// outside of the component
const promisifiedDeleteUser = attachPromise(deleteUser)

...

// in delete callback
await dispatch(promisifiedDeleteUser())
```

pkg/webui/lib/selectors/profile-picture.js
kschiffer: The object structure looks like this:

```javascript

{ 
  sizes: {
    '0': '/assets/blob/profile_pictures/admin.01EXQ1Y3VA3Q2A4CY80WRWTMGJ/original.jpg',
    '64': '/assets/blob/profile_pictures/admin.01EXQ1Y3VA3Q2A4CY80WRWTMGJ/64.jpg',
    '128': '/assets/blob/profile_pictures/admin.01EXQ1Y3VA3Q2A4CY80WRWTMGJ/128.jpg',
     '256': '/assets/blob/profile_pictures/admin.01EXQ1Y3VA3Q2A4CY80WRWTMGJ/256.jpg',
     '512': '/assets/blob/profile_pictures/admin.01EXQ1Y3VA3Q2A4CY80WRWTMGJ/512.jpg'
   }
}
```

So this would return the PP URL, as expected.

pkg/webui/lib/selectors/profile-picture.js
kschiffer: This one is not used anyways, so I'll remove it.

pkg/webui/lib/selectors/profile-picture.js
kschiffer: I've updated the function, it will return whatever is closer to the requested size. Since the PPs use powers of two, I didn't include any logic for the case that there are two sizes with the same distance, since it should not occur.

pkg/webui/account/containers/profile-settings-form/index.js
bafonins: This causes `handleSubmit` to be recreated on each rerender.

pkg/webui/account/containers/profile-settings-form/index.js
```diff
@@ -0,0 +1,312 @@
...
+  const validationContext = {
+    disableUploadConfig,
+    useGravatarConfig,
+    initialProfilePictureSource,
+  }
+
+  // Cancel debounced queue when unmounting.
+  useEffect(() => () => debounceCancel(), [debounceCancel])
+
+  const handleEmailAddressChange = useCallback(
```
bafonins: `handleEmailAddressChange` is not really needed. you can use `debouncedSetEmail` directly.

pkg/webui/account/containers/profile-settings-form/index.js
```diff
@@ -0,0 +1,312 @@
...
+          accept={['.jpg', '.jpeg', '.png']}
+          decode={imageDataDecoder}
+          dataTransform={imageDataTransform}
+          imageClassName={style.uploadImagePreview}
+          required={useGravatarConfig}
+          mayRemove
+          image
+        />
+      )}
+      {useGravatarConfig && profilePictureSource === 'gravatar' && (
```
bafonins, line 236: When `use_gravatar` and `disable_upload` are `true` this part is rendered and this is shown to the user:

Screenshot 2021-02-08 at 16 37 45

> You can upload a different profile picture by selecting the option above.

There is no option above to upload profile picture.

pkg/webui/lib/selectors/profile-picture.js
```diff
@@ -0,0 +1,53 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export const isValidProfilePictureObject = profilePicture =>
+  Boolean(profilePicture) && Boolean(profilePicture.sizes)
+
+export const getProfilePictureSizes = profilePicture => {
+  if (!isValidProfilePictureObject(profilePicture)) {
+    return []
```
bafonins:
```suggestion
    return {}
```

cypress/integration/account/profile-settings.spec.js
```diff
@@ -0,0 +1,158 @@
...
+
+const user = {
+  ids: { user_id: 'profile-settings-test-user' },
+  name: 'Test User',
+  primary_email_address: 'test-user@example.com',
+  password: 'ABCDefg123!',
+  password_confirm: 'ABCDefg123!',
+}
+
+describe('Account App profile settings', () => {
```
bafonins, line 23: Please consider adding tests with `disable_upload = true` and `use_gravatar = false`. Just UI assertions should be enough.

pkg/webui/account/containers/profile-settings-form/index.js
bafonins: I see. My initial comment was not about using hard reload with `window.location`, but about the approach for displaying info notification via query parameter. I just want to emphasize that this shouldnt become the normal way to handle such cases.

Im fine with keeping this for now, but later for similar cases this should be implemented properly with more robust approach.

pkg/webui/account/containers/profile-card/index.js
```diff
@@ -0,0 +1,54 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
```
bafonins:
```suggestion
// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
```

pkg/webui/account/containers/profile-card/profile-card.styl
```diff
@@ -0,0 +1,53 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
```
bafonins:
```suggestion
// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
```

pkg/webui/account/containers/profile-picture/index.js
```diff
@@ -0,0 +1,28 @@
+// Copyright © 2020 The Things Network Foundation, The Things Industries B.V.
```
bafonins, line 1:
```suggestion
// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
```

bafonins (changes requested):

bafonins (approved):

kschiffer:
@johanstokking / @rvolosatovs please review, otherwise I cannot merge.

johanstokking (approved):
Code owned files LGTM (only `tools/mage/dev.go`)

PR #3737, opened 2021-02-03
Fix concurrent token requests after token expiry 

kschiffer:
#### Summary
This quickfix PR addresses an issue that could lead to multiple token requests being launched at the same time, resulting in all such requests failing except the first one. This likely one of the causes for [this sentry event](https://sentry.io/organizations/the-things-industries/issues/1661144927/?project=2682566) and [this one](https://sentry.io/organizations/the-things-industries/issues/1850591405).

#### Changes
- Make sure that the token request only runs once at a time and other callers will wait for the request to finish
- Add a couple of end-to-end tests around access token failure scenarios

#### Testing

Manual and end-to-end tests.

#### Notes for Reviewers
This issue happens e.g. when the access token expires and the Console makes a new request to the token source. We obtain new access tokens "on demand", when the Console tries to make a backend request. When multiple requests are launched at the same time (e.g. when, upon navigation change, reconnecting to a stream and retrieving an entity), the Console would also launch multiple requests to obtain the new token. After the token was retrieved by one request, the other one would fail, since the old access token is now deleted. This failed request then triggers a sentry log event.

#### Checklist

### DIFF
--- a/pkg/webui/lib/access-token.js
+++ b/pkg/webui/lib/access-token.js
@@ -14,27 +14,43 @@
 
 import * as cache from './cache'
 
-export default retrieveToken => async () => {
-  const storedToken = cache.get('accessToken')
-  let token
+export default fetchToken => {
+  let tokenPromise
+  let finishedRetrieval = true
 
-  if (!storedToken || Date.parse(storedToken.expiry) < Date.now()) {
-    // If we don't have a token stored or it's expired, we want to retrieve it.
-
-    // Remove stored, invalid token.
-    clear()
-
-    // Retrieve new token and store it.
-    const response = await retrieveToken()
-    token = response.data
+  const retrieveToken = async () => {
+    const response = await fetchToken()
+    const token = response.data
     cache.set('accessToken', token)
+    finishedRetrieval = true
+
     return token
   }
 
-  // If we have a stored token and its valid, we want to use it.
-  return storedToken
+  return () => {
+    const token = cache.get('accessToken')
+
+    if (!token || Date.parse(token.expiry) < Date.now()) {
+      // If we don't have a token stored or it's expired, we want to retrieve it.
+
+      // Prevent issuing more than one request at a time.
+      if (finishedRetrieval) {
+        finishedRetrieval = false
+
+        // Remove stored, invalid token.
+        clear()
+
+        // Retrieve new token and store it.
+        tokenPromise = retrieveToken()
+      }
+      return tokenPromise
+    }
+
+    // If we have a stored token and its valid, we want to use it.
+    return token
+  }
 }
 
-export function clear() {
+export const clear = () => {
   cache.remove('accessToken')
 }


cypress/integration/console/auth-token-handling.spec.js
```diff
@@ -0,0 +1,112 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import stringToHash from '../../../pkg/webui/lib/string-to-hash'
+
+describe('Auth token handling', () => {
```
bafonins: 1. These tests do not test fixes made in this PR. If you cherry pick only https://github.com/TheThingsNetwork/lorawan-stack/pull/3737/commits/da474ef69a27f6ddfc6d0793735d5bd949a54863 to the latest `v3.11` all tests pass.
2. There are no UI tests. IMO, such tests should be defined within smoke tests for some concrete user flow

pkg/webui/lib/access-token.js
```diff
@@ -14,27 +14,41 @@
...
-  // If we have a stored token and its valid, we want to use it.
-  return storedToken
+  return () => {
+    const token = cache.get('accessToken')
+
+    if (!token || Date.parse(token.expiry) < Date.now()) {
+      // If we don't have a token stored or it's expired, we want to retrieve it.
+
+      // Remove stored, invalid token.
+      clear()
```
bafonins: Each call executes `clear` while token fetching is pending, while only one `clear` is required (is it even required?).

bafonins (commented):
Manually tested, seems to be working. Please see my comments

pkg/webui/lib/access-token.js
kschiffer: I'll move it inside the `if`. I think it's good to clear the token, since it's known to be invalid and the retrieval of the new token might fail, in which case the expired token would remain.

cypress/integration/console/auth-token-handling.spec.js
kschiffer: 1.  Yeah, these tests do not cover the particular failure scenario fixed here, which is somewhat hard to recreate (and to detect as well), should have made that clearer.
2.  There are to some extend UI tests, since they test whether the UI behaves correctly in these scenarios (no wrong redirects, no error messages). These tests are not connected to a concrete user flow, since they're basically over-arching functionality of the frontend. It only tests the behavior in some rare but possible states that the application can get in.

kschiffer: If you prefer, I can remove the tests and we can discuss this in a separate PR.

bafonins: Im fine merging the fix as it is right now. But for the tests I would really prefer updating our testing guidelines to account for such scenarios.

bafonins (approved):

Issue #604, opened 2019-05-01
GC or Recovery of deleted entities


htdvisser:
#### Summary

Some entities in the IS are soft-delete. We want to be able to clean them up (and release their IDs) or recover them.

#### Why do we need this?

Sometimes entities are deleted by accident, in which case we want to restore them. After a cooldown period we want to be able to re-register new entities with the same IDs.

#### What is already there? What do you see now?

On delete, entities are marked with a `deleted_at`

#### What is missing? What do you want to see?

- [x] It should be possible to search for or to list entities that have been deleted
- [x] It should be possible to recover deleted entities
- [ ] Deleted entities should be permanently deleted after a cooldown period

#### How do you propose to implement this?

- In ListXXX/GetXXX requests we could add a `bool include_deleted`
- For restoring deleted entities we can...
    - restore deleted entities on Get/Update 
    - restore deleted entities with a new API call
- For cleaning up deleted entities we can...
    - add a background process to the IS that periodically GCs deleted entities
    - add an API endpoint for GCing deleted entities, which can be called from a cron job.

#### Can you do this yourself and submit a Pull Request?

Sure

virtualguy:
@htdvisser this would be great to have for re-provisioning devices.

Is there any work around to reuse a dev_eui other than manual manipulation of redis+DB?

htdvisser:
This particular issue is more about applications and gateways. Sometimes users want to start over and decide to delete their application/gateway, not knowing that the ID will not be released.

Right now the solution is to re-register with a different ID, but sometimes users really like their IDs...

Note that for gateways, the EUIs are released on delete (since https://github.com/TheThingsNetwork/lorawan-stack/pull/1843).

---

End Devices are not soft-delete, so if you delete those they're gone, and you can re-register them.

The tricky part for End Devices is that they have to be deleted from 4 databases (the Join Server, Application Server, Network Server and Identity Server). The console and command-line interface already do this, but if you have a multi-cluster deployment, and delete your end devices in the "wrong" console, the end devices may still be registered in the other JS/AS/NS. We've also seen users who use the API directly and only delete devices from the Identity Server, again keeping the registrations in JS/AS/NS.

virtualguy:
Gotcha, makes sense. Sounds like we have a different issue as the NS was complaining of device already existing during create from the CLI tool. I'll try reproduce and open a separate ticket

htdvisser:
Most of this is implemented, and will land in `v3.12`.

[...]

PR #3739, opened 2021-02-04
Soft-Deleted Entity Handling

### LINKED ISSUE #1121: Restoring Deleted Applications
#### Summary
When you delete an application and decides to add it back, if fails because it complains that the ID already exists.

#### Steps to Reproduce

1. Remove existing application
2. Add it back using CLI or 
3. Add it back using API

#### What do you see now?
```
{
    "code": 6,
    "message": "error:pkg/identityserver/store:id_taken (ID already taken)",
    "details": [
        {
            "namespace": "pkg/identityserver/store",
            "name": "id_taken",
            "message_format": "ID already taken",
            "correlation_id": "e68a49d8b874408d97c8c57b15b0838e"
        }
    ]
}
```

Successfullly add back the application

#### Environment

amazon linux using TTN CLI and Postman using API Call

When user try to insert a deleted application, update the deleted flag to false

...

### LINKED ISSUE #1903: Allow listing of and searching for soft-deleted entities
#### Summary

It would be helpful if we could add a flag to the List and Search requests to include soft-deleted entities.

#### Why do we need this?

This would let a user (or admin) restore them or permanently delete them (#1121, #604, #1703)

- We do soft-delete, so deleted entities are just entities with a non-nil deleted_at field.
- Deleted entities are never included in list or search responses.
- GETting or PUTting to them results in a 404.

A way to list/search for soft-deleted entities.

Later (out of scope for this issue) a way to permanently delete, or restore the entity.

- API: Add a `include_deleted` in List and Search requests.
- API: Add a `deleted_at` to entities that have soft-delete.
- IS: Update pb setters and model setters for models that have soft-delete.
- IS: Update entity stores to query for deleted entities if requested.

We may also have to update the rights logic (queries) to include deleted items

htdvisser:
#### Summary

This PR improves handling of soft-deleted entities in the Identity Server.

Refs #604 
Closes #1903 
Closes #1121

#### Changes

- API: Add `deleted_at` field to API of entities that can be soft-deleted
- API: Add ~`include_deleted`~ `deleted` to `List` and `Search` RPCs in registries with soft-delete
- API: Add `Restore` RPCs to registries with soft-delete
	- This requires the DELETE right, meaning that anyone who can delete the application, can also restore it.
- API: Add missing `Purge` RPC to OAuth client registry
- CONF: Add "restore window" config option that sets how long after deleting an entity it can still be restored.
- IS: Implement listing deleted entities
- IS: Implement searching for deleted entities
- IS: Implement restoring deleted entities
- IS: Implement missing `Purge` RPC of OAuth client registry
- ~Console: Implement listing / searching for deleted entities~

#### Testing

- Unit tests
- CLI
- Console (in progress)

##### Regressions

I think we may see some inconsistent behavior in the rights checks when caching is enabled. Not sure what we can do about that.

#### Notes for Reviewers

There's still some work to do in future PRs:

- API, IS: Add RPCs for listing soft-deleted entities for which the cooldown window has expired, meaning that users are no longer allowed to restore them
	- The initial implementation will probably use cluster auth for this. No idea what we can do for external integrations, but I think that's out of scope here anyway.

#### Checklist


kschiffer:
UX wise, this should ideally work like this:

*   We should use the trash metaphor and add a "Trash" tab to the entity tables
*   In the "Trash" tab, every table row should have two buttons "Restore" and "Permanently delete"
*   The "Trash" tab should have a "Empty trash" button to quickly purge all items
*   The confirmation toast for entity deletions needs an "Undo" button
*   The confirmation modal needs to be updated to remove "This action cannot be undone!" warnings

This is quite some changes which I would recommend doing in a separate PR. Maybe it's best to limit this PR to backend changes only? I can then take your Console work as foundation for a follow-up PR.

pkg/identityserver/organization_registry.go
```diff
@@ -250,6 +260,31 @@ func (is *IdentityServer) deleteOrganization(ctx context.Context, ids *ttnpb.Org
...
+		org, err := orgStore.GetOrganization(store.WithSoftDeleted(ctx, true), ids, softDeleteFieldMask)
+		if err != nil {
+			return err
+		}
+		if org.DeletedAt == nil {
+			panic("store.WithSoftDeleted(ctx, true) returned result that is not deleted")
+		}
+		if time.Since(*org.DeletedAt) > is.configFromContext(ctx).Delete.Restore {
+			return errRestoreWindowExpired.New()
+		}
```
kschiffer, line 278: This appears like an unnecessary limitation to me. Why shouldn't it be possible to restore the entity until it is actually GC'ed? (Idem for the other entity's restore methods)

pkg/identityserver/application_registry.go
```diff
@@ -161,6 +168,9 @@ func (is *IdentityServer) listApplications(ctx context.Context, req *ttnpb.ListA
 			return nil, err
 		}
 	}
+	if req.IncludeDeleted {
+		ctx = store.WithSoftDeleted(ctx, false)
+	}
```
kschiffer: With regards to the UX that I outlined above, what would be needed is actually an exclusive filter. Something like `IsDeleted` which then only gets deleted or not deleted records.

kschiffer (changes requested):
Codeowned files LGTM. See my remark on the Console bit.

htdvisser:
I've updated the PR according to @kschiffer's suggestions. The API now takes a `deleted` parameter that makes the IS only return deleted entities if true. The Console changes have been taken out of this PR.

> @kschiffer:
> This appears like an unnecessary limitation to me. Why shouldn't it be possible to restore the entity until it is actually GC'ed? (Idem for the other entity's restore methods)

The reason here is external services. The idea is that those will periodically ask the IS about deleted (and unrecoverable) entities, so that those services can also purge their databases. This means that we should allow for some time between entities being unrecoverable, and being really gone.

kschiffer (approved):
Changes LGTM, haven't tested CLI.

johanstokking (commented):
API LGTM

johanstokking (approved):
API LGTM now with a green button

rvolosatovs (approved):

Issue #3221, opened 2020-09-11
Make events pub/sub format configurable


adriansmares:
#### Summary

The event backends should have a configurable format.

#### Why do we need this?

In order to lower runtime costs on encoding/decoding JSON, and to lower the amount of network traffic generated by our events.

#### What is already there? What do you see now?

Marshal/unmarshal using `encoding/json`. `go-cloud` offers internally `protobuf` usage, but this is not available over settings.

#### What is missing? What do you want to see?

Configurable marshaling, between `json` and `protobuf`.

#### Environment

`v3.9.2`

#### How do you propose to implement this?

- Add a configuration field to backends for `format` 
- Make this format be part of the factory function.
- Use the selected format at runtime.

#### How do you propose to test this?

Check using an appropiate pub/sub client if the format used is correct.

#### Can you do this yourself and submit a Pull Request?

Yes. cc @htdvisser

htdvisser:
When we initially built the events system back in 2018, we implemented the "internal" and "redis" backends as a reference only. The idea was that we'd migrate to a different backend such as RabbitMQ, Cloud Pub/Sub or NSQ/NATS for more "serious" deployments.

The "cloud" backend already uses protobuf as a default (and has support for json as well). 

And to be honest, I don't think the biggest scalability problem is the encoding (JSON vs Protobuf), but rather the fact that all events consumers (i.e. all console instances) subscribe to the firehose of all events and filter them locally, instead of routing events on the pub/sub broker or through some separate router (such as in Packet Broker).

adriansmares:
> And to be honest, I don't think the biggest scalability problem is the encoding (JSON vs Protobuf), but rather the fact that all events consumers (i.e. all console instances) subscribe to the firehose of all events and filter them locally, instead of routing events on the pub/sub broker or through some separate router (such as in Packet Broker).

That's most likely correct - but keep in mind that we had some interesting performance gains by simply changing the JSON library in some cases - I'd really like to see if `protobuf` would be better, at least initially.

But indeed - the real problem is the routing is way too trivial at the moment, so refactoring the events backend is inevitable - especially since the Event Server will need `exactly-once` semantics for HA (which we don't have right now).

htdvisser:
[...]

PR #3748, opened 2021-02-08
Refactor Events Package

htdvisser:
#### Summary

This PR refactors the events package to be more efficient, especially when using the Redis backend.

This is related to, but does not yet address https://github.com/TheThingsNetwork/lorawan-stack/issues/3221.

#### Changes

- Change Subscriber interface to take entity identifiers
- Replace the "default" pubsub with a "noop" pubsub
- Implement a new "basic" pubsub
- Change Redis pubsub backend to use multiple channels, and only subscribe to Redis channels when there are actual subscriptions
- Change Redis pubsub backend to use protobuf instead of json

#### Testing

- [x] Kept the existing unit tests for the existing backends as much the same as possible
- [x] Monkey testing with CLI and Console
- [x] Will try to write a common test suite that we can use to test all backend implementations.

##### Regressions

- The most common use of events is the event stream in the Console, and I've tested that.

#### Notes for Reviewers

I'm afraid this is a bit too late for v3.11, even though it would have been nice to get it in there. What we could do, is keep the existing `redis` backend, and instead add the current Redis implementation as a separate backend. Then we can put it behind a config flag, and we could still add it in v3.11.1.

#### Checklist

	- [ ] Needs minor bump


pkg/events/redis/redis.go
```diff
@@ -17,67 +17,98 @@ package redis
...
-	sub := ps.client.Subscribe(ctx, ps.eventChannel)
-	logger.Info("Subscribed")
+
+	ps.mu.Lock()
+	ps.sub = ps.client.Subscribe(ctx)
+	patterns := make([]string, len(ps.subscriptions))
+	for pattern := range ps.subscriptions {
+		patterns = append(patterns, pattern)
+	}
+	logger.WithField("patterns", patterns).Debug("PSUBSCRIBE")
```
adriansmares: Maybe the messages should be a bit more descriptive ?

pkg/events/redis/redis.go
```diff
@@ -86,28 +117,108 @@ func (ps *PubSub) subscribeTask(ctx context.Context) error {
...
+
+		ps.PubSub.RemoveSubscription(sub)
+
+		ps.mu.Lock()
+		defer ps.mu.Unlock()
+
+		for _, pattern := range sub.patterns {
+			ps.subscriptions[pattern]--
+			if ps.subscriptions[pattern] == 0 {
+				log.FromContext(ctx).WithField("pattern", pattern).Debug("PUNSUBSCRIBE")
```
adriansmares: Same here.

adriansmares (approved):

johanstokking (approved):
Nothing to add

htdvisser:
Re-requesting review from @adriansmares for the added test suite

adriansmares (approved):

Issue #3749, opened 2021-02-08
Simulate uplink and schedule downlink based on device repository examples


johanstokking:
#### Summary

Simulate uplink and schedule downlink based on examples declared in the device repository.

#### Why do we need this?

To make it really easy to test integrations.

#### What is already there? What do you see now?

Onboarding devices via the device repository.

#### What is missing? What do you want to see?

In the **Messages** tab, when the end device has a version set, retrieve uplink examples from the device repository and put the example names in a dropdown and schedule the binary payload. It can also be a pre-fill for the input field, so that the user can see what the binary payload is.

Same for scheduling downlink.

#### How do you propose to implement this?

I suppose this needs an API extension to list the codec information with the payload examples. When those can be retrieved by the end device's version identifiers, the dropdowns can be populated.

#### How do you propose to test this?

Test with devices with and without examples.

#### Can you do this yourself and submit a Pull Request?

Can test

neoaggelos:
Had already done some work on this during initial implementation, parts of which were also used for the Wall Of Fame during the conference.

My proposal would be to extend `MessagePayloadFormatter` with two new fields:
- `CodecID`: this comes from the Device Repository, and can be used by users to refer to the Device Repository
- `Examples`: this parses the codec examples from the Device Repository. For the `Output` of each example, a `protobuf.Struct` is used.

This also means that we add field mask paths to the `GetPayloadFormatterRequest`. By default, only the `formatter` and `formatter_parameter` fields are returned, maintaining backwards API compatibility.

johanstokking:
> My proposal would be to extend `MessagePayloadFormatter` with two new fields

Why?

neoaggelos:
> > My proposal would be to extend `MessagePayloadFormatter` with two new fields
> 
> Why?

I am open to alternatives, but I think it's a simple solution: The `ttnpb.MessagePayloadFormatter` message is used by the `Get*Codec()` RPCs only. The AS can keep using it to retrieve the codecs and encode/decode payloads, and the Console can then use it to retrieve both the codec (if needed) and the codec examples.

johanstokking:
Right, I was confused with `MessagePayloadFormatters` (plural). Sounds good to me.

neoaggelos:
API has been implemented in https://github.com/TheThingsNetwork/lorawan-stack/pull/3772, removing my assignment and the `needs/api` tag.

Issue #3694, opened 2021-01-21
Device Repository API field validation


neoaggelos:
#### Summary

Add validations for all Device Repository API

#### Why do we need this?

We want to have strict field validations across the whole API

#### What is already there? What do you see now?

With #3530. validation for identifiers and a few select fields of the request messages. 

#### What is missing? What do you want to see?

Validation rules for all fields.

#### Environment

v3.11

#### How do you propose to implement this?

Validate all API field according to the LoRaWAN Devices Schema https://lorawan-schema.org/draft/devices/1/

#### How do you propose to test this?

Existing unit tests cover API functionality. End to end tests initialize the device repository index, so they will cover failures with data from the Device Repository

#### Can you do this yourself and submit a Pull Request?

Yes

htdvisser:
Since @neoaggelos isn't going to finish this before the v3.14.1 deadline, I'll move it to v3.14.2.

PR #3772, opened 2021-02-09
Device Repository API for Codec Examples

neoaggelos:
#### Summary

API and backend for #3749.

#### Changes

- Extend `MessagePayloadFormatter` with codec ID and examples fields.
- Extend `GetPayloadFormatterRequest` with examples.
- Parse codec examples from the Device Repository. 

#### Testing

- Local testing `?field_mask=examples,codec_id` with devices from the Device Repository.
- Unit tests.

##### Regressions

None, default field mask paths maintain API backwards-compatibility

#### Notes for Reviewers

Adding this for the March milestone, but opening as a draft PR since I already had the development branch lying around, and it would be a lot more time-consuming to visit it again in a month's time from now. Tests are still TODO and there are still some rough edges, will cc people once this is ready to review.

#### Checklist


neoaggelos:
I would prefer to have this in `3.12.0`, since it introduces API changes, and the linked issue is for Q2.

api/devicerepository.proto
```diff
@@ -308,6 +312,20 @@ message MessagePayloadFormatter {
...
+  string codec_id = 3 [(gogoproto.customname) = "CodecID", (validate.rules).string = {pattern: "^([a-z0-9](?:[-]?[a-z0-9]){2,}|)?$" , max_len: 36}];
+
+  message Example {
+    string description = 1;
+    message Input {
+      uint32 f_port = 1 [(validate.rules).uint32.lte = 255];
+      bytes frm_payload = 2 [(gogoproto.customname) = "FRMPayload"];
+    };
+    Input input = 2;
+    google.protobuf.Struct output = 3;
```
johanstokking: Why is the input following the input parameters, but the output is not? I.e. where are the warnings and errors?

api/devicerepository.proto
```diff
@@ -308,6 +312,20 @@ message MessagePayloadFormatter {
...
   string formatter_parameter = 2;
+
+  string codec_id = 3 [(gogoproto.customname) = "CodecID", (validate.rules).string = {pattern: "^([a-z0-9](?:[-]?[a-z0-9]){2,}|)?$" , max_len: 36}];
+
+  message Example {
+    string description = 1;
+    message Input {
+      uint32 f_port = 1 [(validate.rules).uint32.lte = 255];
+      bytes frm_payload = 2 [(gogoproto.customname) = "FRMPayload"];
+    };
```
johanstokking: This is the input for decoding uplink and decoding downlink. How about encoding downlink?

pkg/devicerepository/store/remote/schema.go
```diff
@@ -444,15 +444,23 @@ func (p EndDeviceProfile) ToTemplatePB(ids *ttnpb.EndDeviceVersionIdentifiers, i
...
 }
 
-// EndDeviceCodec is the format of the `vendor/<vendor>/<codec-id>.yaml` files.
+type EndDeviceCodecExample struct {
+	Description string `yaml:"description"`
+	Input       struct {
+		FPort uint32 `yaml:"fPort"`
+		Bytes []byte `yaml:"bytes"`
+	} `yaml:"input"`
+	Output map[string]interface{} `yaml:"output"`
```
johanstokking: How does this work for `DownlinkEncoder`?

johanstokking (changes requested):

johanstokking:
@neoaggelos the merge window of 3.12.0 closes today. It's too tight to request a review on a draft PR the day before that window closes. But let's try to get this in.

api/devicerepository.proto
neoaggelos: Changed to using a Struct for the input as well.

pkg/devicerepository/grpc.go
```diff
@@ -172,32 +172,29 @@ func (dr *DeviceRepository) GetTemplate(ctx context.Context, req *ttnpb.GetTempl
...
+func getCodec(ctx context.Context, req *ttnpb.GetPayloadFormatterRequest, f func(store.GetCodecRequest) (*ttnpb.MessagePayloadFormatter, error)) (*ttnpb.MessagePayloadFormatter, error) {
 	if clusterauth.Authorized(ctx) != nil {
 		if err := rights.RequireApplication(ctx, req.ApplicationIDs, ttnpb.RIGHT_APPLICATION_DEVICES_READ); err != nil {
 			return nil, err
 		}
 	}
-	return dr.store.GetUplinkDecoder(req.VersionIDs)
+	return f(store.GetCodecRequest{
+		VersionIDs: req.VersionIDs,
+		Paths:      ttnpb.AddFields(req.FieldMask.Paths, "formatter", "formatter_parameter"),
```
neoaggelos: This ensures API backwards compatibility, which means that it is OK even if this does not come with 3.12.0.

Moving to 3.12.1 milestone.

api/devicerepository.proto
johanstokking: TBH, I think this becomes a bit nasty. We know the input and output parameters here so they can be typed. We just need to differentiate between encoding and decoding. Why do you use `Struct` here?

pkg/devicerepository/store/remote/schema.go
```diff
@@ -17,7 +17,7 @@ package remote
 import (
 	"time"
 
-	pbtypes "github.com/gogo/protobuf/types"
+	"github.com/gogo/protobuf/types"
```
johanstokking: Revert

pkg/devicerepository/store/interface.go
```diff
@@ -55,12 +55,10 @@ type GetModelsResponse struct {
...
 }
 
-// DefinitionIdentifiers is a request to retrieve an end device template for an end device definition.
-type DefinitionIdentifiers struct {
-	BrandID,
-	ModelID,
-	FirmwareVersion,
-	BandID string
+// GetCodecRequest is a request to retrieve the codec of
+type GetCodecRequest struct {
```
johanstokking: Can't this be an interface?

johanstokking (commented):
Looks pretty much done; why is this still draft?

neoaggelos:
Unit tests are done, this is only pending some final testing.

api/devicerepository.proto
```diff
@@ -303,11 +307,51 @@ message ListEndDeviceModelsResponse {
...
+  google.protobuf.Struct data = 1;
+  repeated string warnings = 2;
+  repeated string errors = 3;
+}
+
+message MessagePayloadDecoder {
   // Payload formatter type.
   PayloadFormatter formatter = 1 [(validate.rules).enum.defined_only = true];
   // Parameter for the formatter, must be set together.
   string formatter_parameter = 2;
```
htdvisser, line 327: While we're at it, should we set a maximum length here?

api/devicerepository.proto
```diff
@@ -303,11 +307,51 @@ message ListEndDeviceModelsResponse {
   repeated EndDeviceModel models = 1;
 }
 
-message MessagePayloadFormatter {
+message EncodedMessagePayload {
+  uint32 f_port = 1 [(validate.rules).uint32.lte = 255];
+  bytes frm_payload = 2 [(gogoproto.customname) = "FRMPayload"];
```
htdvisser, line 312: Should we set a maximum length validation rule here?

api/devicerepository.proto
```diff
@@ -303,11 +307,51 @@ message ListEndDeviceModelsResponse {
...
+  repeated string warnings = 3;
+  repeated string errors = 4;
```
htdvisser: Should we set validation rules for maximum length (both of the list, and of each list element)?

api/devicerepository.proto
```diff
@@ -303,11 +307,51 @@ message ListEndDeviceModelsResponse {
...
+}
+
+message DecodedMessagePayload {
+  google.protobuf.Struct data = 1;
+  repeated string warnings = 2;
+  repeated string errors = 3;
```
htdvisser: Should we set validation rules for maximum length (both of the list, and of each list element)?

api/devicerepository.proto
```diff
@@ -303,11 +307,51 @@ message ListEndDeviceModelsResponse {
...
+message MessagePayloadDecoder {
   // Payload formatter type.
   PayloadFormatter formatter = 1 [(validate.rules).enum.defined_only = true];
   // Parameter for the formatter, must be set together.
   string formatter_parameter = 2;
+
+  string codec_id = 3 [(gogoproto.customname) = "CodecID", (validate.rules).string = {pattern: "^([a-z0-9](?:[-]?[a-z0-9]){2,}|)?$" , max_len: 36}];
+
+  message Example {
+    string description = 1;
```
htdvisser: Should we set a maximum length validation rule here?

api/devicerepository.proto
```diff
@@ -303,11 +307,51 @@ message ListEndDeviceModelsResponse {
...
+    EncodedMessagePayload input = 2;
+    DecodedMessagePayload output = 3;
+  }
+  // Examples
```
htdvisser: This is kind of obvious, right?

api/devicerepository.proto
```diff
@@ -303,11 +307,51 @@ message ListEndDeviceModelsResponse {
...
+  repeated Example examples = 4;
```
htdvisser: Should we add a validation rule for the maximum length of the list?

api/devicerepository.proto
```diff
@@ -303,11 +307,51 @@ message ListEndDeviceModelsResponse {
...
+}
+
+message MessagePayloadEncoder {
+  // Payload formatter type.
+  PayloadFormatter formatter = 1 [(validate.rules).enum.defined_only = true];
+  // Parameter for the formatter, must be set together.
+  string formatter_parameter = 2;
```
htdvisser, line 343: Should we set a maximum length validation rule here?

api/devicerepository.proto
```diff
@@ -303,11 +307,51 @@ message ListEndDeviceModelsResponse {
...
+
+  string codec_id = 3 [(gogoproto.customname) = "CodecID", (validate.rules).string = {pattern: "^([a-z0-9](?:[-]?[a-z0-9]){2,}|)?$" , max_len: 36}];
+
+  message Example {
+    string description = 1;
```
htdvisser: Should we set a maximum length validation rule here?

api/devicerepository.proto
```diff
@@ -303,11 +307,51 @@ message ListEndDeviceModelsResponse {
...
+    DecodedMessagePayload input = 2;
+    EncodedMessagePayload output = 3;
+  }
+  // Examples
+  repeated Example examples = 4;
```
htdvisser: Should we add a validation rule for the maximum length of the list?

htdvisser (commented):
Reviewed `api` and added some suggestions.

api/devicerepository.proto
neoaggelos: These are sourced from the device repository https://github.com/TheThingsNetwork/lorawan-devices/blob/master/lib/draft/schema.json#L240-L250 where no maximum length is enforced.

There is already an issue to add validation rules for the device repository API (https://github.com/TheThingsNetwork/lorawan-stack/issues/3694) for other message fields as well.

I am adding validation rules while I'm at it, but the proper place for these rules would be the schema of the Device Repository.

api/devicerepository.proto
neoaggelos: Leaving unset, since there are no validations rules in the `MACPayload` message either https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.12/api/lorawan.proto#L138

api/devicerepository.proto
johanstokking: Only the code files themselves don't have a length check.

FPort and FRMPayload length are validated by the device repository.

There is (currently) no limitation of the number of examples, the length of the example description and the elements in the input and output fields.

I tend to agree that we should keep the validation in the device repository. This API is only returning fields anyway.

htdvisser (approved):
Okay, if these messages are indeed only used as response messages, validation isn't that important. But still, by not adding validation now, we risk that someone decides to implement a new RPC that _does_ use these messages in a request, and then we won't have proper validation. 

Other than that, I don't have anything to add.

johanstokking (approved):

Issue #56, opened 2019-01-29
Send ApplicationDownlinkSent event from Tx acknowledgement


johanstokking:
**Summary:** 

Send DownlinkSent event based on the Tx acknowledgement received from a Gateway Server.

**Why do we need this?**

To provide insight to the downstream to applications and integrations.

**What is already there? What do you see now?**

Tx acknowledgement is received by the GS.

**What is missing? What do you want to see?**

- The Tx acknowledgement should be sent from the GS to the NS
- The NS should figure out which application downlink belonged to that Tx acknowledgement, based on the correlation IDs. We probably need the GS to use (temporary) tokens to find the right correlation IDs

**How do you propose to implement this?**

- Add new rpc in `GsNs` for the Tx acknowledgement
- Send an event from GS to NS with the correlation IDs
- NS to send the DownlinkSent event to AS

**What can you do yourself and what do you need help with?**

@rvolosatovs if GS just gives you a correlation ID, you can't find the downlink message, right? I guess it's gone also anyway from the queue?

Should the GS hold the downlink message (in memory) and send the sent downlink message back to the NS?

How actually does the NS find the concerning downlink message in the case of a downlink NACK, because then it's also gone from the queue right?

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/1476 by @johanstokking

johanstokking:
Blocking #76

neoaggelos:
This requires the API change of #76 which is going in 3.12, updated the milestone for this issue accordingly.

neoaggelos:
Quote from https://github.com/TheThingsNetwork/lorawan-stack/issues/76#issuecomment-774503756

> .... NS omits fields from the DownlinkMessage that is sent to the Gateway Server (e.g. end device identifiers).
>
> So, I think it is best to keep all this information on the NS side instead: Have NS store scheduled downlinks temporarily. When GS reports a TxAcknowledgment, NS checks if there is a matching scheduled downlink. If there is no match, skip (maybe print a warning). Otherwise, retrieve the complete downlink message, merge the `TxSettings` reported by the GS and then take care of forwarding to the AS.

Is this still the way to go for this issue? Set `ttn:v3:ns:sent-downlink:<CorrelationID>`, try to match that when a TxAck arrives, merge the transmission metadata from GS and forward the ApplicationDownlinkSent message?

[...]

PR #3773, opened 2021-02-09
Send TxEvents upstream

### LINKED ISSUE #76: Send Tx events upstream
**Summary:** 

GS can report TX acknowledgement of a specific `ttnpb.DownlinkMessage` to the NS, for the purpose of improved observability.

TX acknowledgement now used to signal presence of JIT queue, nothing else.

- [ ] API for signaling TX events from GS to NS. Add an rpc in `GsNs`, i.e. `ReportTxAcknowledgement()` next to the existing `HandleUplink()`. GS reports the original `DownlinkMessage` as well as some metadata (gateway ID and `TxAcknowledgment`)
- [ ] NS may want to keep statistics, i.e. set the `TxAcknowledgement` in the recent downlink messages. I'm fine with a follow up issue on this
- [ ] NS publishes the downlink sent and failed events, based on the result code, to AS. AS already has infrastructure to decrypt and process further

neoaggelos:
#### Summary

Closes #76

#### Changes

- Extend `GsNs` interface with `ReportTxAcknowledgment()` RPC
- Forward TxAck events from GS frontends to the NS, along with the sent downlink message (which includes transmission settings).
- For udp/lbs gateways, the downlink tokens are used to match sent downlink messages. For gRPC/MQTT gateways, the `TxAcknowledgment` must contain the correlation IDs of the downlink message being acked.
- Add new `gs.down.tx.ack.forward` and `gs.down.tx.ack.drop` events.
- (TODO) handle transmission errors on the GS frontends (e.g. https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.11/pkg/gatewayserver/io/udp/udp.go#L374)

#### Testing

- Test locally with a UDP gateway and verify that TxAck events are received, and that `ReportTxAcknowledgment` is called for the upstream host.
- Unit tests.

##### Regressions

None. Unit tests verify that existing flows do not break.

#### Notes for Reviewers

Using `io.DownlinkTokens` for MQTT and gRPC gateways means that there is a maximum number `1 << 4` downlink messages that can be acked. This should not be a problem, but it is an implementation restriction that is not mandatory by the protocol itself.

@johanstokking Please have a look at the implementation and comment accordingly. The `TxAcknowledgment.Result` field maps directly to the TX_ACK type enums from the UDP packet forwarder, but this does not allow the frontend to signal errors during the scheduling of the downlink message itself (e.g. https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.11/pkg/gatewayserver/io/udp/udp.go#L355). Should we forward a TxAck with `UNKNOWN_ERROR` to the GS for these, or add a new channel (e.g. `TxErr`)? We need to make sure that this does not mess up the Round trip times calculation in the GS side.

#### Checklist


api/messages.proto
```diff
@@ -159,6 +162,16 @@ message TxAcknowledgment {
     GPS_UNLOCKED = 8;
   }
   Result result = 2 [(validate.rules).enum.defined_only = true];
+
+  // Copy of the acknowledged downlink message. Set by the Gateway Server.
+  // For the UDP and LBS frontends, the downlink token of the TxAck message is used.
+  // For the gRPC and MQTT frontends, the correlation IDs are used.
```
johanstokking: Remove, this is not relevant for the API

api/messages.proto
```diff
@@ -159,6 +162,16 @@ message TxAcknowledgment {
...
+  DownlinkMessage downlink_message = 3;
+}
+
+message GatewayTxAcknowledgment {
+  TxAcknowledgment tx_ack = 1;
+  GatewayIdentifiers gateway_ids = 2 [(gogoproto.customname) = "GatewayIDs"];
```
johanstokking: Reverse these for consistency

pkg/gatewayserver/gatewayserver.go
```diff
@@ -707,14 +713,15 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
 			ctx = events.ContextWithCorrelationID(ctx, fmt.Sprintf("gs:tx_ack:%s", events.NewCorrelationID()))
-			msg.CorrelationIDs = append(msg.CorrelationIDs, events.CorrelationIDsFromContext(ctx)...)
 			if msg.Result == ttnpb.TxAcknowledgment_SUCCESS {
 				registerSuccessDownlink(ctx, gtw, protocol)
 			} else {
 				registerFailDownlink(ctx, gtw, msg, protocol)
 			}
-			// TODO: Send Tx acknowledgement upstream (https://github.com/TheThingsNetwork/lorawan-stack/issues/76)
-			continue
+			if d := msg.DownlinkMessage; d != nil {
```
johanstokking: When is it `nil`?

pkg/gatewayserver/io/downlink_token.go
```diff
@@ -34,25 +38,45 @@ type DownlinkTokens struct {
...
 }
 
-// Next returns a new downlink token.
-func (t *DownlinkTokens) Next(correlationIDs []string, time time.Time) uint16 {
+// Next returns a new downlink token for a downlink message.
+func (t *DownlinkTokens) Next(msg *ttnpb.DownlinkMessage, time time.Time) uint16 {
 	key := uint16(atomic.AddUint32(&t.last, 1))
 	pos := key % downlinkTokenItems
+
+	copied := deepcopy.Copy(msg).(*ttnpb.DownlinkMessage)
```
johanstokking: Leave this up to the caller, if needed

pkg/gatewayserver/io/downlink_token.go
```diff
@@ -34,25 +38,45 @@ type DownlinkTokens struct {
...
 	item := t.items[pos]
 	if item.key != token {
 		return nil, 0, false
 	}
-	return item.correlationIDs, time.Sub(item.time), true
+	return item.msg, time.Sub(item.time), true
+}
+
+// GetWithCorrelationIDs is like Get, but uses the correlation IDs instead of the downlink token.
+func (t DownlinkTokens) GetWithCorrelationIDs(cids []string, time time.Time) (*ttnpb.DownlinkMessage, time.Duration, bool) {
```
johanstokking: I think that we should use a signaling CID here (`gs:down:XXX`), and not all. Or use the downlink token as a new CID (`gs:down:token:XXX`), and look for that?

pkg/gatewayserver/io/downlink_token.go
```diff
@@ -34,25 +38,45 @@ type DownlinkTokens struct {
...
+nextItem:
+	for _, item := range t.items {
+		if len(cids) != len(item.correlationIDs) {
+			continue
+		}
+		for index, cid := range item.correlationIDs {
+			if cids[index] != cid {
+				continue nextItem
+			}
```
johanstokking: The problem is that we require an exact match of all CIDS, in the right order. This is asking for trouble, for no good reason imo

pkg/gatewayserver/io/downlink_token.go
```diff
@@ -34,25 +38,45 @@ type DownlinkTokens struct {
 	items [downlinkTokenItems]downlinkToken
 }
 
-// Next returns a new downlink token.
-func (t *DownlinkTokens) Next(correlationIDs []string, time time.Time) uint16 {
+// Next returns a new downlink token for a downlink message.
+func (t *DownlinkTokens) Next(msg *ttnpb.DownlinkMessage, time time.Time) uint16 {
 	key := uint16(atomic.AddUint32(&t.last, 1))
 	pos := key % downlinkTokenItems
```
johanstokking, line 45: It's not that there's a limit of 1<<4, but that it rolls over after that. This is because there is no need to keep track of many downlink messages, but you do want to allow the gateway to send TX acknowledgments of older frames.

pkg/gatewayserver/io/mqtt/mqtt.go
```diff
@@ -148,6 +154,7 @@ func (c *connection) setup(ctx context.Context) (err error) {
 			case <-ctx.Done():
 				return
 			case down := <-c.io.Down():
+				c.tokens.Next(down, time.Now())
```
johanstokking: So here you might want to add a correlation ID with that token

pkg/gatewayserver/io/mqtt/mqtt.go
```diff
@@ -319,6 +326,11 @@ func (c *connection) deliver(pkt *packet.PublishPacket) {
 			logger.WithError(err).Warn("Failed to unmarshal Tx acknowledgment message")
 			return
 		}
+		if ack.DownlinkMessage == nil {
+			if down, _, ok := c.tokens.GetWithCorrelationIDs(ack.GetCorrelationIDs(), time.Now()); ok {
```
johanstokking: And look for that specific correlation ID here

johanstokking: Or parse the token from the CID and use the method to get the message by token, which is probably cleaner

pkg/gatewayserver/io/udp/udp.go
```diff
@@ -299,8 +299,9 @@ func (s *srv) handleUp(ctx context.Context, state *state, packet encoding.Packet
 			}
 		}
 		var rtt *time.Duration
-		if cids, delta, ok := state.tokens.Get(uint16(packet.Token[0])<<8|uint16(packet.Token[1]), packet.ReceivedAt); ok {
-			msg.TxAcknowledgment.CorrelationIDs = cids
+		if downlink, delta, ok := state.tokens.Get(uint16(packet.Token[0])<<8|uint16(packet.Token[1]), packet.ReceivedAt); ok {
+			msg.TxAcknowledgment.DownlinkMessage = downlink
+			msg.TxAcknowledgment.CorrelationIDs = downlink.CorrelationIDs
```
johanstokking, line 304: What is the added value of `CorrelationIDs` over reading `DownlinkMessage.CorrelationIDs`?

pkg/networkserver/grpc_gsns.go
```diff
@@ -1274,3 +1274,9 @@ func (ns *NetworkServer) HandleUplink(ctx context.Context, up *ttnpb.UplinkMessa
 	logger.Debug("Unmatched MType")
 	return ttnpb.Empty, nil
 }
+
+// ReportTxAcknowledgment is called by the Gateway Server when a tx acknowledgment is received.
+func (ns *NetworkServer) ReportTxAcknowledgment(ctx context.Context, up *ttnpb.GatewayTxAcknowledgment) (_ *pbtypes.Empty, err error) {
+	// TODO: Forward TxAcknowledgment to the Application Server (https://github.com/TheThingsNetwork/lorawan-stack/issues/76)
```
johanstokking: @rvolosatovs this is probably a simple call, can you chime in here?

rvolosatovs:
Assigning myself to add the NS part

api/networkserver.proto
```diff
@@ -48,6 +48,7 @@ service AsNs {
 // The GsNs service connects a Gateway Server to a Network Server.
 service GsNs {
   rpc HandleUplink(UplinkMessage) returns (google.protobuf.Empty);
+  rpc ReportTxAcknowledgment(GatewayTxAcknowledgment) returns (google.protobuf.Empty);
```
johanstokking: Good to add some docs here as well, while you're at it

pkg/gatewayserver/io/downlink_token.go
```diff
@@ -34,25 +40,49 @@ type DownlinkTokens struct {
...
+
+// ParseTokenFromCorrelationIDs parses the correlation ID
+func (t DownlinkTokens) ParseTokenFromCorrelationIDs(cids []string) (uint16, bool) {
+	for _, cid := range cids {
+		match := parseTokenRegex.FindString(cid)
+		if match == "" {
+			continue
+		}
+		token, err := strconv.ParseUint(strings.TrimPrefix(match, "gs:down:token:"), 10, 16)
+		if err != nil {
```
johanstokking: nit; you now use the `gs:down:token:` magic const twice. It's nicer to do a group regex match so you get the number directly

pkg/gatewayserver/io/mqtt/mqtt.go
```diff
@@ -319,6 +328,13 @@ func (c *connection) deliver(pkt *packet.PublishPacket) {
 			logger.WithError(err).Warn("Failed to unmarshal Tx acknowledgment message")
 			return
 		}
+		if ack.DownlinkMessage == nil {
```
johanstokking: So do we or don't we support the downlink message to be passed by the client?

johanstokking (commented):
Did an early review, I see it's still in progress.

pkg/gatewayserver/gatewayserver.go
neoaggelos: The `DownlinkMessage` may or may not be set by the frontend (e.g. when there was no downlink token match).

pkg/gatewayserver/io/udp/udp.go
neoaggelos: We must also set `CorrelationIDs` to avoid breaking the API.

pkg/gatewayserver/io/mqtt/mqtt.go
neoaggelos: The proposed implementation forwards the `DownlinkMessage` as-is if it is provided by the gateway, otherwise it tries to match using the correlation ID of the downlink token. Should we always ignore the `DownlinkMessage` sent by the gateway instead?

johanstokking: > Should we always ignore the `DownlinkMessage` sent by the gateway instead?

Yes, and put in the API doc that it's set by the Gateway Server.

johanstokking:
@neoaggelos please mark as ready for review and I'll do a full review.

pkg/networkserver/grpc_gsns.go
neoaggelos: This is part of #56 instead of #76, so let's do in a separate PR.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Added
 
+- Gateway Server forwards `TxAck` packets to the Network Server.
```
johanstokking: What does `TxAck` mean for the user?

api/networkserver.proto
```diff
@@ -47,7 +47,10 @@ service AsNs {
 
 // The GsNs service connects a Gateway Server to a Network Server.
 service GsNs {
+  // HandleUplink forwards an uplink message to the Network Server.
```
johanstokking: No need to follow godoc here (`<symbol> <verb>`). In fact, it looks a bit weird in the generated docs.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -707,14 +713,15 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
+				d.CorrelationIDs = append(d.CorrelationIDs, events.CorrelationIDsFromContext(ctx)...)
```
johanstokking: Now, the correlation IDs are set _after_ invoking the register events, so that they don't contain them. Why is that?

pkg/gatewayserver/io/downlink_token_test.go
```diff
@@ -15,34 +15,52 @@
...
 	a := assertions.New(t)
 	tokens := DownlinkTokens{}
 
+	msgs := make([]*ttnpb.DownlinkMessage, 0, downlinkTokenItems*2)
 	all := []uint16{}
 	for i := 0; i < downlinkTokenItems*2; i++ {
-		cids := []string{fmt.Sprintf("message_%d", i)}
-		all = append(all, tokens.Next(cids, time.Unix(int64(i), 0)))
+
+		msgs = append(msgs, ttnpb.NewPopulatedDownlinkMessage(test.Randy, true))
```
johanstokking: Please don't use populators in new test code

pkg/gatewayserver/io/downlink_token_test.go
```diff
@@ -15,34 +15,52 @@
...
 				a.So(ok, should.BeFalse)
 			}
 		}
 	}
 }
+
+func TestCorrelationIDs(t *testing.T) {
+	tokens := DownlinkTokens{}
+
+	token := tokens.Next(ttnpb.NewPopulatedDownlinkMessage(test.Randy, true), time.Now())
```
johanstokking: Idem

pkg/gatewayserver/io/grpc/grpc.go
```diff
@@ -132,8 +134,13 @@ func (s *impl) LinkGateway(link ttnpb.GtwGs_LinkGatewayServer) error {
...
-			if msg.TxAcknowledgment != nil {
-				if err := conn.HandleTxAck(msg.TxAcknowledgment); err != nil {
+			if ack := msg.TxAcknowledgment; ack != nil {
+				if token, ok := s.tokens.ParseTokenFromCorrelationIDs(ack.GetCorrelationIDs()); ok {
+					if down, _, ok := s.tokens.Get(token, time.Now()); ok {
+						ack.DownlinkMessage = down
+					}
+				}
+				if err := conn.HandleTxAck(ack); err != nil {
 					logger.WithError(err).Warn("Failed to handle Tx acknowledgement")
```
johanstokking:
```suggestion
					logger.WithError(err).Warn("Failed to handle Tx acknowledgment")
```

pkg/gatewayserver/observability.go
```diff
@@ -209,26 +218,44 @@ var gsMetrics = &messageMetrics{
 		},
 		[]string{protocol},
 	),
+	downlinkTxAckForwarded: metrics.NewContextualCounterVec(
```
johanstokking: This is what `downlinkTxSucceeded` is for

pkg/gatewayserver/observability.go
```diff
@@ -209,26 +218,44 @@ var gsMetrics = &messageMetrics{
...
+		prometheus.CounterOpts{
+			Subsystem: subsystem,
+			Name:      "downlink_tx_ack_forwarded_total",
+			Help:      "Total number of forwarded tx acknowledgments",
+		},
+		[]string{host, protocol},
+	),
+	downlinkTxAckDropped: metrics.NewContextualCounterVec(
```
johanstokking: This is what `downlinkTxFailed` is for

pkg/networkserver/grpc_gsns.go
johanstokking: Then update the issue number here

johanstokking (commented):
Almost there

api/networkserver.proto
```diff
@@ -47,7 +47,10 @@ service AsNs {
...
   rpc HandleUplink(UplinkMessage) returns (google.protobuf.Empty);
+  // ReportTxAcknowledgment forwards a TxAcknowledgment packet to the Network Server.
+  rpc ReportTxAcknowledgment(GatewayTxAcknowledgment) returns (google.protobuf.Empty);
```
htdvisser, line 61: I think this breaks (5) of our compatibility commitment. If a `v3.11.0+n` Gateway Server starts calling this RPC on a `v3.11.0` Network Server, we'll get Unimplemented errors (that get spammed to Sentry), so I think this PR needs to go to the `v3.12` branch.

johanstokking: Yep I agree, I missed that. I breaks rule 5:

>We will not break the API between components and events within minor versions. So at least the same minor versions of components are compatible with each other.

adriansmares (commented):
Nothing to add.

pkg/gatewayserver/observability.go
neoaggelos: This metric relates to TxAck messages that are successfully forwarded to the Network Server (not necessarily meaning that the downlink was sent successfully), so it seems to be different than `downlinkTxSucceeded`.

pkg/gatewayserver/observability.go
neoaggelos: Same as `downlinkTxSucceeded` vs `downlinkTxAckForwarded`

api/networkserver.proto
neoaggelos: Changed to target 3.12

pkg/gatewayserver/observability.go
htdvisser: I think that this metric (as well as the event) is very confusing. I actually can't think of any reason why users would be interested in knowing whether the TX ACK was successfully forwarded to the Network Server or not. The only thing they're interested in is the fact that there was a TX ACK in the first place.

johanstokking: Indeed. Note that `downlinkTxSucceeded` is also not being used, and I was hoping that this PR would start using it.

johanstokking: 1. Fire `downlinkTxSucceeded` and `downlinkTxFailed` events
2. Consider removing the added `downlinkTxAckForwarded` and `downlinkTxAckDropped` events. I'm not against keeping them, because we can then see the balance between received messages and forwarded/dropped messages and see if they add up, but it should be very clear to the user what is what

johanstokking (changes requested):

pkg/gatewayserver/observability.go
neoaggelos: Hmm, alright. Removed the new metrics and sticking to the existing `downlinkTxSucceeded` and `downlinkTxFailed` metrics.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -663,6 +663,10 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
 					} else {
 						registerForwardStatus(ctx, gtw, msg, host.name)
 					}
+				case *ttnpb.TxAcknowledgment:
+					if err := host.handler.HandleTxAck(ctx, gtw.GatewayIdentifiers, msg); err != nil {
+						logger.WithField("host", host.name).WithError(err).Debug("Drop TxAck")
```
johanstokking:
```suggestion
						logger.WithField("host", host.name).WithError(err).Debug("Drop Tx acknowledgment")
```

api/networkserver.proto
```diff
@@ -47,7 +47,10 @@ service AsNs {
 
 // The GsNs service connects a Gateway Server to a Network Server.
 service GsNs {
+  // Called by the Gateway Server when an uplink message arrives.
   rpc HandleUplink(UplinkMessage) returns (google.protobuf.Empty);
+  // Called by the Gateway Server when a tx acknowledgment arrives.
```
johanstokking:
```suggestion
  // Called by the Gateway Server when a Tx acknowledgment arrives.
```

pkg/gatewayserver/io/udp/config.go
```diff
@@ -32,11 +32,11 @@ type Config struct {
 	// PacketBuffer defines how many packets are buffered to handlers before it overflows.
 	PacketBuffer int `name:"packet-buffer" description:"Buffer size of unhandled packets"`
 	// DownlinkPathExpires defines for how long a downlink path is valid. A downlink path is renewed on each pull data and
-	// TX acknowledgement packet.
+	// TX acknowledgment packet.
```
johanstokking:
```suggestion
	// Tx acknowledgment packet.
```

pkg/gatewayserver/io/grpc/grpc_test.go
```diff
@@ -471,6 +476,32 @@ func TestTraffic(t *testing.T) {
...
+					}
+				}
+
+				select {
+				case ack := <-conn.TxAck():
+					a.So(ack.DownlinkMessage, should.Resemble, tc.Message)
+					a.So(ack.Result, should.Equal, ttnpb.TxAcknowledgment_SUCCESS)
+				case <-time.After(timeout):
+					if tc.TxAckErrorAssertion == nil {
+						t.Fatal("Timeout waiting for TxAck")
```
johanstokking:
```suggestion
						t.Fatal("Timeout waiting for Tx acknowledgment")
```

pkg/gatewayserver/io/grpc/grpc_test.go
```diff
@@ -471,6 +476,32 @@ func TestTraffic(t *testing.T) {
...
+				select {
+				case upCh <- &ttnpb.GatewayUp{
+					TxAcknowledgment: &ttnpb.TxAcknowledgment{
+						CorrelationIDs: cids,
+						Result:         ttnpb.TxAcknowledgment_SUCCESS,
+					},
+				}:
+				case <-time.After(timeout):
+					if tc.TxAckErrorAssertion == nil {
+						t.Fatal("Receive unexpected timeout while sending TxAcknowledgment")
```
johanstokking:
```suggestion
						t.Fatal("Receive unexpected timeout while sending Tx acknowledgment")
```

api/messages.proto
```diff
@@ -145,6 +145,9 @@ message DownlinkMessage {
 }
 
 message TxAcknowledgment {
+  // Correlation IDs for the downlink message.
+  // Set automatically by the UDP and LBS frontends.
+  // For gRPC and the MQTT v3 frontends, the correlation IDs must match the ones of the downlink message the TxAck message refers to.
```
johanstokking:
```suggestion
  // For gRPC and the MQTT v3 frontends, the correlation IDs must match the ones of the downlink message the Tx acknowledgment message refers to.
```

api/gatewayserver.proto
```diff
@@ -36,7 +36,7 @@ message GatewayUp {
   repeated UplinkMessage uplink_messages = 1;
   // Gateway status produced by the gateway.
   GatewayStatus gateway_status = 2;
-  // A transmission acknowledgement or error.
+  // A transmission acknowledgment or error.
```
johanstokking:
```suggestion
  // A Tx acknowledgment or error.
```

pkg/gatewayserver/io/udp/config.go
```diff
@@ -32,11 +32,11 @@ type Config struct {
...
 	// Gateways typically pull data every 5 seconds.
 	DownlinkPathExpires time.Duration `name:"downlink-path-expires" description:"Time after which a downlink path to a gateway expires"`
 	// ConnectionExpires defines for how long a connection remains valid while no pull data, push data or TX
```
johanstokking:
```suggestion
	// ConnectionExpires defines for how long a connection remains valid while no pull data, push data or Tx
```

pkg/gatewayserver/io/ws/ws_test.go
```diff
@@ -1244,7 +1277,7 @@ func TestRTT(t *testing.T) {
 					select {
 					case ack := <-gsConn.TxAck():
 						if ack.Result != ttnpb.TxAcknowledgment_SUCCESS {
-							t.Fatalf("Tx Acknowledgement failed")
+							t.Fatalf("Tx Acknowledgment failed")
```
johanstokking:
```suggestion
							t.Fatalf("Tx acknowledgment failed")
```

johanstokking (approved):
LGTM. Only make the term consistent for our future selves and for grepping logs etc.

htdvisser (approved):
Nothing to add

PR #3787, opened 2021-02-11
Update node modules

kschiffer:
#### Summary
This PR updates our node modules as much as possible, both normal and development dependencies.

#### Changes
- Update all packages to the latest version, or as much as possible without breaking functionality or pipeline
- Update all configs accordingly
  - Rename `.babelrc` to `babel.config.json`, which is the recommended [naming](https://babeljs.io/docs/en/configuration#whats-your-use-case), `.babelrc` is not suggested anymore
- Update the js codebase formatting as a result of the eslint and prettier updates
  - This includes applying the result of `eslint --fix`
  - This includes updating JsDoc comments to include documentation of object properties
- [Rewrite `<LocationMap />` component](https://github.com/TheThingsNetwork/lorawan-stack/pull/3787/files#diff-5ef9c2b92a525147bfe070dd9e63dae54dd511a685bd2d6661690bdf27762073) to fix breaking changes
- Update `dependabot` schedule to monthly
- Remove some dependencies that we were not using (anymore), e.g. `snapshot-diff`
- Use the same devDependencies/tooling for JS SDK and webui
  - Saves us from keeping two seperate toolings updates
  - Good thing to put the configuration of these closer together

#### Testing

Manual testing and CI.

##### Regressions

Possible but hard to enumerate. Passing CI and e2es should ensure that there are no critical issues introduced.

#### Notes for Reviewers
I'm doing this since we recently updated our dependabot config to check for outdated dependencies regularly. To avoid getting a flood of dependabot PRs, I want to use this PR to bring all dependencies to the latest state in one go. After this we can then use the monthly dependabot PRs to keep everything up-to-date.

The update to latest `prettier` introduced some changes to how anonymous functions are formatted (with a space between function symbol and parentheses). I did not find any way to change this preference, so I guess we have to accept it. We'll replace those with the arrow notation anyway. There's also a slight change in how line-breaks are calculated. Unfortunately these changes make the diff quite noisy.

Some information on other packages that I could not update to the latest version:

#### history
- currently no support for the `basename` prop. would mean verbose refactor or hacky workarounds to get it working with the latest version.

#### webpack v5
- not mature enough to update
- Too many breaking changes and many plugins don't support it yet (or won't ever) e.g. ChunkHash plugin
- generally updating webpack is a world of pain and should be done in dedicated manner and in good consideration

#### stylus-loader v5
- needs webpack v5

#### webpack-cli
- causes unresolvable crash with many plugins

#### babel-plugin-react-intl -> formatjs
- this breaks functionality that we need to display external backend messages, e.g. 

#### copy-webpack-plugin v7
- minimum supported webpack version is 5

#### html-webpack-plugin v5
- minimum supported webpack version is 5

#### cypress
I did not update cypress since this will already be done by #3822, so no need to duplicate efforts here.

#### storybook
While working on this, I noticed that our storybook setup is currently broken. I prefer fixing this separately, together with updating storybook itself.

#### Checklist


htdvisser (approved):
Mage changes LGTM

bafonins (commented):
> Possible but hard to enumerate.

This seems to be a good indication that we should avoid such PRs with massive updates, especially when we are considering core dependencies like `babel` and others. Moreover, such huge diffs (and diffs to `package.json`) make it really hard for reviews to go through the changes.

> The update to latest prettier introduced some changes to how anonymous functions are formatted (with a space between function symbol and parentheses). I did not find any way to change this preference, so I guess we have to accept it.

Indeed, there is no way to disable this if we want to keep our `prettier+eslint` setup. See [prettier v2 changelog](https://prettier.io/blog/2020/03/21/2.0.0.html#always-add-a-space-after-the-function-keyword-3903httpsgithubcomprettierprettierpull3903-by-j-f1httpsgithubcomj-f1-josephfrazierhttpsgithubcomjosephfrazier-sosukesuzukihttpsgithubcomsosukesuzuki-thorn0httpsgithubcomthorn0-7516httpsgithubcomprettierprettierpull7516-by-bakkothttpsgithubcombakkot)

I suggest to split this PR into separate. This can look like:
1. Update `eslint`, `eslint` plugins, `prettier` and any other lint/fmt dependencies + format the code
2. `babel`, `jest` and other build related deps (like `webpack` plugins)
3. `react`, other UI related dependencies (like `react-select`, `react-toastify` and others), `redux`, etc.
4. `storybook`
5. rest

I do understand that this requires more extra work on this, but taking into account that this PR already has some merge conflicts this is justified. Additionally we shouldnt accept such PRs as the norm.

Splitting this PR would really make it much easier for me to review as well as lower the risk of breaking anything.

kschiffer:
Closing this then.

PR #3792, opened 2021-02-12
Add dynamic style for byte field

### LINKED ISSUE #3784: App Key text field is too narrow
#### Summary
When registering a new device the App Key field in the form is too narrow to show the whole key, the last byte overflows. There is no indication that there's another couple of characters off to the right of the field. This is causing trouble for users when they generate a key and copy it out as they're missing off the last byte.
...

#### Steps to Reproduce

1. Start to register a new device
2. Run through the first step to select a device template
3. Click the generate app key button next to the field

---

#### What do you see now?
---
You'll get something like this:

Register-end-device-al-thing-The-Things-Network
---
Note that there's a "hidden" final byte at the right:
extra byte

...

A slightly wider text box! This seems to happen at all screen sizes.

...

#### Environment
Tested in Chrome/Linux with public v3 and private instances

asmulko:
#### Summary

Closes #3784

#### Changes

- added dynamic style that calculates `max-width` for byte input based on `max` prop
- reduced spacing between byte pairs

Before:
before

After:
after

#### Testing

Manually tested

#### Checklist

### DIFF
--- a/pkg/webui/components/input/index.js
+++ b/pkg/webui/components/input/index.js
@@ -57,12 +57,14 @@ class Input extends React.Component {
     }).isRequired,
     label: PropTypes.string,
     loading: PropTypes.bool,
+    max: PropTypes.number,
     onBlur: PropTypes.func,
     onChange: PropTypes.func,
     onEnter: PropTypes.func,
     onFocus: PropTypes.func,
     placeholder: PropTypes.message,
     readOnly: PropTypes.bool,
+    showPerChar: PropTypes.bool,
     title: PropTypes.message,
     type: PropTypes.string,
     valid: PropTypes.bool,
@@ -78,16 +80,19 @@ class Input extends React.Component {
     component: 'input',
     disabled: false,
     error: false,
-    inputWidth: 'm',
+    /** Default `inputWidth` value is set programmatically based on input type. */
+    inputWidth: undefined,
     icon: undefined,
     label: undefined,
     loading: false,
+    max: undefined,
     onFocus: () => null,
     onBlur: () => null,
     onChange: () => null,
     onEnter: () => null,
     placeholder: undefined,
     readOnly: false,
+    showPerChar: false,
     title: undefined,
     type: 'text',
     valid: false,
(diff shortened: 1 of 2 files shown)


bafonins:
IMO, this doesnt look right.

field

Can we use `inputWidth` prop to increase input width for 16byte long inputs?

kschiffer (changes requested):
- Instead of using the new `autoWidth` prop, let's adjust the styling for `inputWidth="auto"` for `<Input.Byte />` components.
- Let's then also adjust the default value for `inputWidth` to "auto" (for `<Input.Byte />`).
- There shouldn't be any changes to `<Form.Field />` necessary
- It appears like the auto width calculation needs to be adjusted to the new spacing
- In case of the `AppKey` input, since the PR adjusted the spacing of the byte values, it appears like we don't need to use autoWidth here because it looks better with the width aligned to the other fields, and it is possible now since it does not cover the last byte anymore.

pkg/webui/components/input/index.js
```diff
@@ -150,8 +152,13 @@ class Input extends React.Component {
     const { focus } = this.state
 
     let Component = component
+    let inputStyle
     if (type === 'byte') {
       Component = ByteInput
+      const { max } = this.props
+      if (max && (inputWidth === 'auto' || max > 16)) {
```
kschiffer:
```suggestion
      if (max && (inputWidth === 'auto')) {
```

I don't think we need that, we should rather make sure that we set `inputWidth` actively.

kschiffer (changes requested):
What about

> Let's then also adjust the default value for `inputWidth` to `"auto"` (for `<Input.Byte />`).

?

So we would make this default, except for the `AppKey` input, where we would overwrite the default (since it's sort of a special case). After you change the default value, make sure it does not break any layouts in our forms.

pkg/webui/components/input/index.js
```diff
@@ -78,10 +79,11 @@ class Input extends React.Component {
     component: 'input',
     disabled: false,
     error: false,
-    inputWidth: 'm',
+    inputWidth: undefined,
```
kschiffer, line 84: Please add a comment that the default value is set programmatically based on input type.

kschiffer (changes requested):
Looking good now, please make sure the auto width is also applied to the AppEUI field:
image

pkg/webui/components/input/index.js
```diff
@@ -180,9 +194,16 @@ class Input extends React.Component {
       [style.textarea]: type === 'textarea',
     })
 
+    const byteProps =
+      type === 'byte'
+        ? {
+            showPerChar,
+          }
+        : null
```
kschiffer:
```suggestion
    const passedProps = { 
      ...rest,
      ...(type === 'byte' ? { showPerChar } : {})
    }
```

And then pass that to the component.

kschiffer (approved):

PR #3796, opened 2021-02-13
Add highlight for frame counter in live data events

### LINKED ISSUE #3716: Highlight Frame Counter value of message events in Live Data views
#### Summary
Highlight Frame Counter (FCnt) value of message events in Live Data views (gateway, application, device).
Highlight/accent with a white rectangle just like is done for DevAddr and payload fields.

#### Why do we need this?
This will make reading framecounter values easier to locate and read.

Framcounter values are already displayed but are more difficult to locate and read at a glance.

Highlight Frame Counter (FCnt) values for message events in Live Data views.

asmulko:
#### Summary

Closes #3716 

#### Changes

- added highlight prop for description list item
- added highlight for frame Counter (FCnt) in live data events

Before:
before

After:
after

#### Testing

Manually tested

#### Checklist

### DIFF
--- a/pkg/webui/console/components/events/previews/shared/description-list/description-list.styl
+++ b/pkg/webui/console/components/events/previews/shared/description-list/description-list.styl
@@ -42,3 +42,12 @@
   &:after
     content: '>'
 
+.highlight
+  line-height: 1.9rem;
+  border: 1px solid $c-divider-dark;
+  font-family: 'IBM Plex Mono', Consolas, Monaco, 'Andale Mono', 'Ubuntu Mono', monospace
+  display: inline
+  background-color: white
+  height: 100%
+  padding: 0 $cs.xs
+  border-radius: $br.s
--- a/pkg/webui/console/components/events/previews/uplink-message.js
+++ b/pkg/webui/console/components/events/previews/uplink-message.js
@@ -49,8 +49,8 @@ const UplinkMessagePreview = React.memo(({ event }) => {
   return (
     <DescriptionList>
       <DescriptionList.Byte title={messages.devAddr} data={devAddr} />
+      <DescriptionList.Item title={messages.fCnt} data={fCnt} highlight />
       <DescriptionList.Item title={messages.fPort} data={fPort} />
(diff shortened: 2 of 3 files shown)


pkg/webui/console/components/events/previews/shared/description-list/description-list.styl
```diff
@@ -42,3 +42,12 @@
   &:after
     content: '>'
 
+.highlight
```
bafonins, line 45: There is nothing wrong with this approach and it seems to work as expected, however now we have 2 places where byte like outline is defined (in `<Byte />` and `<DescriptionList />`. If one changes the styling for one component (say the `<Byte />` as it is used in many places in the Console) we can end up with inconsistent styling within the `<Event />` component.

@kschiffer Is there any other way to highlight the `fCnt` value in the events list? Mb by making it bold (similar to how values are rendered in the V2 Console)?

kschiffer: I agree with you that we're risking some inconsistencies here. I think we should regard this fix here as a preliminary solution. Let's use this solution now and I'll open another issue about increasing the recognizability of common data points in the event streams in general.

pkg/webui/console/components/events/previews/uplink-message.js
```diff
@@ -49,7 +49,7 @@ const UplinkMessagePreview = React.memo(({ event }) => {
     <DescriptionList>
       <DescriptionList.Byte title={messages.devAddr} data={devAddr} />
       <DescriptionList.Item title={messages.fPort} data={fPort} />
-      <DescriptionList.Item title={messages.fCnt} data={fCnt} />
+      <DescriptionList.Item title={messages.fCnt} data={fCnt} highlight />
```
kschiffer:
```suggestion
      <DescriptionList.Item title={messages.fCnt} data={fCnt} highlight />
      <DescriptionList.Item title={messages.fPort} data={fPort} />
```

Let's also change the order here, which I think is more in line with the order of importance.

kschiffer (approved):

Issue #3786, opened 2021-02-11
Snapshots for visual testing in our end-to-end tests


kschiffer:
#### Summary
In this issue I want to suggest an improvement for how we conduct visual testing as part of our end-to-end test. The idea is to use snapshot testing instead of asserting on UI elements directly.

#### Why do we need this?
Current visual testing has issues such as:
- Locks us in into a specific arrangement of the UI (down to the exact wording of texts)
- Produces false negatives when UI changes structurally
- Is tedious and ambiguous to set up
- They are superficial tests that won't catch many obvious-to-the-eye issues

#### What is already there? What do you see now?
With visual testing, I mean the parts of our tests that ensure that the UI is rendered correctly. Currently we do that by asserting on the visibility of the UI elements that are rendered on a page (see [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.11/cypress/integration/console/gateways/create.spec.js#L35-L68) e.g.), as well as asserting on the exact text that these elements display.

#### What is missing? What do you want to see?
There are libraries that facilitate visual testing using snapshots. Snapshots are either literal image snapshots, or copies of text or markup that an implementation is checked against. We can use this method to make visual testing more flexible yet more precise. When working on visual updates, we can create snapshots (via cypress) and let cypress compare the visual output against those screenshots for subsequent tests.

Obviously there are caveats for this methodology as well. E.g. one issue of snapshot testing is the danger that snapshots get carelessly updated when their test-cases fail. It's important to maintain snapshot testing consciously and that everyone working on end-to-end tests knows how snapshot tests work.

I still think that this downside does not offset the benefits that we gain from using snapshot tests (as described above).

#### How do you propose to implement this?
There are many cypress plugins like https://github.com/jaredpalmer/cypress-image-snapshot which we can use to implement snapshot testing. The idea would be to replace all `renders UI elements in place`-tests with such snapshot tests. We could then also more efficiently add visual tests for different visual states.

#### Can you do this yourself and submit a Pull Request?
Yes, after we made a decision.

@bafonins @asmulko please share your thoughts about this and/or greenlight if you agree.

bafonins:
> Locks us in into a specific arrangement of the UI (down to the exact wording of texts)

We do want to assert that elements are in the right place with correct text displayed, right?

> Produces false negatives when UI changes structurally

Can you provide an example of such false negatives?

[...]

PR #3822, opened 2021-02-17
Improve e2e setup for device repository tests in the Console

bafonins:
#### Summary

This PR adds more tests for the DR form in the Console.

#### Changes

- Make use of fixtures to stub requests when fetching brands, models and templates
- Remove usage of 3rd party device templates
- Update `cypress` to `6.5.0`
- Fix some tests after `cypress` update
- Added more e2e tests for end device creation via the DR form.

#### Testing

Run e2e tests, CI

#### Notes for Reviewers

I decided to update `cypress` to `v6+` to make use of [cy.intercept](https://docs.cypress.io/api/commands/intercept.html) which is better alternative to `cy.route`. Also, updating `cypress` to the latest version seems like a good idea considering that currently were are 2 major versions behind.

It seems that `cypress` now has different way of asserting that elements are visible and since we use customized checkboxes and radio buttons the `be.visible` assertion does not work anymore. I changed these to `exist`. Also see https://github.com/TheThingsNetwork/lorawan-stack/issues/3786. It also seems that `findByText` chained with `should('not.be.visible')` changed its behavior in `v6+`. Changed these to `should('not.exist')` as well.

There are still tests that use the actual brands, models and templates directly fetched from the device repository. I think we should test whether products from `the-things-products` can be registered, especially taking into account that we are responsible for maintaining these in DR and we have full control of the entries there. Might be worth adding tests for `the-things-industries` (generic node) as well.

Replaced usage of `cy.route` with `cy.intercept`.

@kschiffer Please check changes made to `cypress/integrations/account/*`.

#### Checklist


package.json
```diff
@@ -129,7 +129,7 @@
     "yup": "^0.29.1"
   },
   "resolutions": {
-    "cypress": "4.11.0"
+    "cypress": "6.5.0"
   },
```
kschiffer, line 133: This might not be necessary anymore with the new version. Did you check that?

pkg/webui/console/views/device-add/repository/device-registration/index.js
```diff
@@ -69,81 +69,29 @@ const Registration = props => {
...
-                : sharedMessages.appKeyDescription
-            }
-            component={Input.Generate}
-            disabled={!mayEditKeys}
-            mayGenerateValue={mayEditKeys}
-            onGenerateValue={generate16BytesKey}
-            glossaryId={glossaryId.APP_KEY}
-          />
-          {lwVersion >= 110 && (
+      <div data-test-id="device-registration">
```
kschiffer, line 77: The diff confuses me. Was there anything changed apart from wrapping everything into a div?

cypress/integration/account/profile-settings.spec.js
```diff
@@ -247,8 +251,8 @@ describe('Account App profile settings without gravatar and upload', () => {
         cy.findByRole('button', { name: 'Expand' }).click()
       })
 
-    cy.findByText('Gravatar', { exact: false }).should('not.be.visible')
-    cy.findByText('Upload image', { exact: false }).should('not.be.visible')
+    cy.findByText('Gravatar', { exact: false }).should('not.exist')
+    cy.findByText('Upload image', { exact: false }).should('not.exist')
```
kschiffer, line 255: Shouldn't this select the label directly, which is visible?

kschiffer (approved):
Not happy with having to use `should('exist')` in some cases now, but I guess there is not much we can do.

> [**@kschiffer**](https://github.com/kschiffer) Please check changes made to `cypress/integrations/account/*`.

LGTM

package.json
bafonins: Yes, had issues in CI without this.

pkg/webui/console/views/device-add/repository/device-registration/index.js
bafonins: No, only the wrapping div.

PR #3857, opened 2021-02-24
Fix footer layout

### LINKED ISSUE #3851: Footer layout is broken in mobile screen sizes
#### Summary
Recent updates and extensions to the Console/Account App footer introduced some styling issues in mobile screen sizes:
image

#### Steps to Reproduce
1. Go to the Console
2. Enable mobile mode in dev tools
3. Scroll down to see the footer

#### What do you see now?
Jagged layout.

Unjagged layout.

Use icon only layout for the support button. This will likely do the trick.

Manual testing.

@asmulko

asmulko:
#### Summary

Closes #3851

#### Changes

- Hide support button text for small screen sizes

Before:
before

After:
a1
a2

#### Testing
- Manually tested

#### Checklist

### DIFF
--- a/pkg/webui/components/button/index.js
+++ b/pkg/webui/components/button/index.js
@@ -28,7 +28,7 @@ import ButtonIcon from './button-icon'
 
 import style from './button.styl'
 
-function assembleClassnames({
+const assembleClassnames = ({
   message,
   danger,
   warning,
@@ -42,7 +42,8 @@ function assembleClassnames({
   error,
   raw,
   disabled,
-}) {
+  responsiveLabel,
+}) => {
   if (unstyled) {
     return className
   }
@@ -58,6 +59,7 @@ function assembleClassnames({
     [style.large]: large,
     [style.raw]: raw,
     [style.disabled]: disabled,
+    [style.responsiveLabel]: responsiveLabel,
   })
 }
 
@@ -69,7 +71,7 @@ const buttonChildren = props => {
   ) : (
     <>
       {icon ? <ButtonIcon icon={icon} type="left" /> : null}
-      {message ? <Message content={message} /> : null}
+      {message ? <Message content={message} className={style.linkButtonMessage} /> : null}
     </>
   )
 
@@ -131,7 +133,7 @@ Button.defaultProps = {
   onBlur: undefined,
 }
 
-Button.Link = function(props) {
+Button.Link = props => {
   const { disabled, titleMessage } = props
   const buttonClassNames = assembleClassnames(props)
   const { to } = props
@@ -147,7 +149,7 @@ Button.Link = function(props) {
 }
 Button.Link.displayName = 'Button.Link'
 
-Button.AnchorLink = function(props) {
+Button.AnchorLink = props => {
   const { target, title, name } = props
   const htmlProps = { target, title, name }
   const buttonClassNames = assembleClassnames(props)
@@ -227,6 +229,8 @@ const commonPropTypes = {
   autoFocus: PropTypes.bool,
   /** A message to be evaluated and passed to the <button /> element. */
   title: PropTypes.message,
+  /** A flag specifying whether the button `message` should be responsive. */
+  responsiveLabel: PropTypes.bool,
 }
 
 buttonChildren.propTypes = {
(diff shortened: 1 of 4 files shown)


pkg/webui/components/footer/index.js
```diff
@@ -31,6 +33,10 @@ const m = defineMessages({
   footer: "You are the network. Let's build this thing together.",
 })
 
+const isMinimized =
+  Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0) <=
```
bafonins, line 37: `window.innerWidth >= document.documentElement.clientWidth` will always be `true`.
`innerWidth` and `clientWidth` are equal when there is no scrollbar, or `innerWidth > clientWidth` because `clientWidth` accounts for scrollbar and computes only the visible part of the screen (document)

pkg/webui/components/footer/index.js
```diff
@@ -31,6 +33,10 @@ const m = defineMessages({
   footer: "You are the network. Let's build this thing together.",
 })
 
+const isMinimized =
```
bafonins: We can make the `<Link />` component responsive by itself (say via `minimize` prop or something like this).
Another option would be render different version of `<Link.Anchor />` based on screen size via css media queries.

pkg/webui/components/footer/index.js
```diff
@@ -51,12 +57,18 @@ const Footer = ({ className, documentationLink, links, supportLink, onlineStatus
...
+                ? sharedMessages.documentationAbbreviation
+                : sharedMessages.documentation
+            }
+          />
         </Link.Anchor>
       )}
       {supportLink && (
         <Button.AnchorLink
-          message={sharedMessages.getSupport}
+          message={isMinimized ? null : sharedMessages.getSupport}
```
kschiffer:
```suggestion
          message={isMinimized ? undefined : sharedMessages.getSupport}
```

This will make the button use the correct styling for only icon buttons.

pkg/webui/components/footer/index.js
kschiffer: I agree. We usually don't do responsive styling programmatically, unless forced to. You can use a combination of attached classnames and rendering different elements and show/hide them based on screen size (using css media queries).

pkg/webui/components/footer/index.js
```diff
@@ -51,12 +57,18 @@ const Footer = ({ className, documentationLink, links, supportLink, onlineStatus
       <span className={style.version}>v{process.env.VERSION}</span>
       {documentationLink && (
         <Link.Anchor className={style.documentation} secondary href={documentationLink} external>
-          <Message content={sharedMessages.documentation} />
+          <Message
+            content={
+              isMinimized && supportLink
+                ? sharedMessages.documentationAbbreviation
+                : sharedMessages.documentation
```
kschiffer: Let's not use an abbreviation here. It might work in English but might be troublesome in other languages. Let's instead hide the version tag. Use a tight media query for this so it only get's hidden when it becomes necessary.

kschiffer (changes requested):

bafonins (changes requested):
Please center the icon
Screenshot 2021-03-02 at 17 08 37
Screenshot 2021-03-02 at 17 08 45

pkg/webui/components/button/index.js
```diff
@@ -69,7 +71,11 @@ const buttonChildren = props => {
   ) : (
     <>
       {icon ? <ButtonIcon icon={icon} type="left" /> : null}
-      {message ? <Message content={message} /> : null}
+      {message ? (
+        <span className={style.linkButtonMessage}>
```
bafonins: Do you really need this wrapping `span`?

```js
<Message className={style.linkButtonMessage} content={message} />
```

should work just fine

bafonins (approved):

pkg/webui/components/button/index.js
```diff
@@ -227,6 +229,8 @@ const commonPropTypes = {
   autoFocus: PropTypes.bool,
   /** A message to be evaluated and passed to the <button /> element. */
   title: PropTypes.message,
+  /** A flag specifying whether the button `message` should be responsive. */
+  minimize: PropTypes.bool,
```
kschiffer:
```suggestion
  /** A flag specifying whether the button `message` should be responsive. */
  responsiveLabel: PropTypes.bool,
```

kschiffer (approved):

PR #3862, opened 2021-02-25
Fix event name and metadata of OAuth events

htdvisser:
#### Summary

This fixes the event that's triggered on user login, and adds `remote_ip` and `user_agent` metadata.

Refs https://github.com/TheThingsIndustries/lorawan-stack/issues/2601

#### Testing

- Some unit testing for the new middleware
- Monkey testing for the events

#### Checklist


pkg/oauth/observability.go
```diff
@@ -20,6 +20,21 @@ import (
 )
 
 var (
+	// EvtUserLogin indicates a user login.
+	EvtUserLogin = events.Define(
+		"oauth.user.login", "login user successful",
```
kschiffer, line 25: To me, logging in and out is more the scope of `account` which also holds the `session` package. The oauth package only hooks into it to enable the `ClientLogout`.

So my suggestion would be to move this back to `account/observability.go` and import it in the `oauth` package, and to use `account.user.{login|logout}` instead.

kschiffer (changes requested):
LGTM but please see my comment on the event scope.

pkg/oauth/observability.go
htdvisser: That was what I did, but that creates import cycles as `pkg/account` imports `pkg/oauth`. And with the event names starting with `oauth.` I think it's not a big problem to have the events in this package.

kschiffer: Yeah, I remember a similar problem when working on the AA. Would it help to make `session` a top level package, and add the events to its observability package?

I don't want to make too much of a big deal out of it. I just find it important to avoid associating the `oauth` package with authentication anymore.

I'll approve and leave it at your discretion.

kschiffer (approved):

PR #3872, opened 2021-03-01
Fix absolute time schedule conflict handling

johanstokking:
#### Summary

Fixes https://github.com/TheThingsIndustries/lorawan-stack-support/issues/225

#### Changes

Changes the error code that Gateway Server returns on a downlink path from `ResourceExhausted` to `AlreadyExists`.

Network Server sees `ResourceExhausted` as non-retryable error for scheduling downlink messages with an absolute time. This means that currently, if any downlink path has a conflict with an already scheduled downlink message, the Network Server will not try other gateways. This pull request changes that behavior: `AlreadyExists` is retryable and will result in other downlink paths being tried.

#### Testing

Unit testing

##### Regressions

Should be none.

#### Checklist


KrishnaIyer (approved):
Yeah `AlreadyExists` does make more sense for this.

rvolosatovs (approved):

PR #3873, opened 2021-03-01
Fix GCS Panic on UDP gateway conf generation

KrishnaIyer:
#### Summary

Fix Panic while generating SX1301 config for frequency plans without radio configuration.

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2602

#### Changes

- If there are no radios configured the frequency plan (ex: ISM2400), use an empty SX1301 block.

#### Testing

- Run stack locally
- Create gateway with 
- Query `global_conf.json`

```json
{
  "SX1301_conf": {
    "lorawan_public": false,
    "clksrc": 0,
    "antenna_gain": 0
  },
  "gateway_conf": {
    "gateway_ID": "50A0CBFFFE800000",
    "server_address": "localhost",
    "serv_port_up": 1700,
    "serv_port_down": 1700,
    "servers": [
      {
        "gateway_ID": "50A0CBFFFE800000",
        "server_address": "localhost",
        "serv_port_up": 1700,
        "serv_port_down": 1700,
        "serv_enabled": true
      }
    ]
  }
}
```

##### Regressions

NA

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ For details about compatibility between different releases, see the **Commitment
 - Event name on user login.
 - Application uplink queue handling in Network Server.
 - Application Server session desynchronization with the Network Server. The Application Server will now attempt to synchronize the end device session view on downlink queue operational errors. This fixes the `f_cnt_too_low` and `unknown_session` errors reported on downlink queue push and replace.
+- Panic while generating SX1301 config for frequency plans without radio configuration.
 
 ### Security
 
--- a/pkg/pfconfig/cpf/cpf.go
+++ b/pkg/pfconfig/cpf/cpf.go
@@ -85,9 +85,12 @@ func BuildLorad(gtw *ttnpb.Gateway, fps *frequencyplans.Store) (*LoradConfig, er
 	if err != nil {
 		return nil, err
 	}
-	sx1301Conf, err := shared.BuildSX1301Config(fp)
-	if err != nil {
-		return nil, err
+	sx1301Conf := &shared.SX1301Config{}
+	if len(fp.Radios) != 0 {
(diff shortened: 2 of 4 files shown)


pkg/pfconfig/shared/shared.go
```diff
@@ -277,6 +277,9 @@ var defaultTxLUTConfigs = []TxLUTConfig{
 
 // BuildSX1301Config builds the SX1301 configuration for the given frequency plan.
 func BuildSX1301Config(frequencyPlan *frequencyplans.FrequencyPlan) (*SX1301Config, error) {
+	if len(frequencyPlan.Radios) == 0 {
```
rvolosatovs: This ignores all the other fields that belong here, right? Shouldn't this result in an error instead or at least generate a partially correct config?
As I see it now this is just silently failing and that can't be right.
Maybe check for Radios before calling this function?

rvolosatovs (dismissed):

pkg/pfconfig/shared/shared.go
johanstokking: Imo this `BuildSX1301Config` shouldn't be called if there aren't radios. The caller is responsible to check _what_ radios there are before calling functions that are about a specific concentrator.

KrishnaIyer: The reason I implemented it this way is because this is now common for all types of config (UDP/LBS/CPF). With the suggested approach, I had to replicate this check thrice. But done.

johanstokking (approved):

pkg/pfconfig/shared/shared.go
johanstokking: It's good to be explicit

PR #3906, opened 2021-03-08
Disable UDP IP firewall by default

### LINKED ISSUE #3894: Disable UDP remote address check by default
#### Summary

Disable the UDP remote address check by default.

#### Why do we need this?

Because the number of gateways that are changing remote addresses seems to be increasing rapidly, sometimes even for the upstream and downstream session, which leads to an unusable gateway.

The `gs.udp.addr-change-block` is 1 minute

Make it `0` to disable it.

1. Change default configuration: https://github.com/TheThingsNetwork/lorawan-stack/blob/51606a79fd661c2740e13ef06a1bcaa07c1b064a/pkg/gatewayserver/io/udp/config.go#L57. When 0, the address check isn't enabled at all (https://github.com/TheThingsNetwork/lorawan-stack/blob/51606a79fd661c2740e13ef06a1bcaa07c1b064a/pkg/gatewayserver/io/udp/udp.go#L58). BTW, while you're at it, remove that `== true` thing that got my eyes bleeding
2. Add deprecation notice in changelog
3. Check that our deployment scripts do not set this setting so that it will get configured anyway

This is already tested in some production environments

Can review

KrishnaIyer:
#### Summary

Disable UDP IP firewall by default. Closes #3894 

#### Changes

- Set the default value of 0 (0 is equivalent to `0 * time.Second` for `time.Duration`)
- Add note on security of new setting.

#### Testing

No new functions added. Only config value. 

##### Regressions

I don't think there'd be any.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/udp/udp.go
+++ b/pkg/gatewayserver/io/udp/udp.go
@@ -58,7 +58,7 @@ func Serve(ctx context.Context, server io.Server, conn *net.UDPConn, config Conf
 	if config.AddrChangeBlock > 0 {
 		firewall = NewMemoryFirewall(ctx, config.AddrChangeBlock)
 	}
-	if config.RateLimiting.Enable == true {
+	if config.RateLimiting.Enable {
 		firewall = NewRateLimitingFirewall(firewall, config.RateLimiting.Messages, config.RateLimiting.Threshold)
 	}
 	s := &srv{
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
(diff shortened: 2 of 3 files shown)


pkg/gatewayserver/io/udp/udp.go
```diff
@@ -58,7 +58,7 @@ func Serve(ctx context.Context, server io.Server, conn *net.UDPConn, config Conf
 	if config.AddrChangeBlock > 0 {
 		firewall = NewMemoryFirewall(ctx, config.AddrChangeBlock)
 	}
-	if config.RateLimiting.Enable == true {
+	if config.RateLimiting.Enable {
```
johanstokking, line 61: 🙏

johanstokking (approved):

PR #3911, opened 2021-03-09
Add field to require secure gateway connections

### LINKED ISSUE #1436: Optionally require authenticated gateway connection
#### Summary

Optionally require authenticated links from gateways.

#### Why do we need this?

To disallow unauthenticated connections, like UDP.

GS keeps track of UDP source addresses which detects and disallows spoofing gateway EUIs while the real UDP-linked gateway is still connected.

We should not allow any unauthenticated gateway links, i.e. UDP on a per-gateway setting basis.

- Add a `require_authentication` bool to `Gateway`, store in IS
- Get this information when connecting the gateway in GS
- Check if the concerning frontend is authenticated. If not, and if `require_authentication` is set, stop the connecting procedure

Can review

htdvisser:
#### Summary

This adds a field to gateways that allows users to indicate that their gateways are only to connect over authenticated connections.

Closes #1436

#### Changes

- [x] Added API field and DB column
- [x] Updated GS connection logic to check that field
- [x] Add checkbox to Console
- [x] Updated console e2e tests

Before:
before_checkbox

After:
after_checkbox

#### Testing

- Manually tested
- Updated e2e tests

##### Regressions

...

#### Notes for Reviewers

...

#### Checklist


htdvisser:
Assigning @asmulko for the Console part:

- Adding a checkbox to `/console/gateways/add`: **Require Authenticated Connection** with description "This gateway may only connect if it uses an authenticated Basic Station or MQTT connection." Probably fits best under **Gateway Server address*.
- Similarly on the `/console/gateways/:gateway_id:/general-settings`

htdvisser:
@johanstokking @KrishnaIyer: While working on this, I noticed that the accept logic for UDP gateways is less than optimal, especially for connections that we reject. 

If we reject an UDP gateway connection (because the gateway doesn't exist, or because it requires authenticated connections), we still send an ACK message to the gateway (so the gateway can't know that something is wrong). However, on every next UDP packet sent by the gateway on the same "udp connection"  triggers a new Connect flow, which makes at least 2 RPCs to the Identity Server, and several database queries. Are you aware of this? Isn't there something we can do better here?

johanstokking:
> However, on every next UDP packet sent by the gateway on the same "udp connection" triggers a new Connect flow, which makes at least 2 RPCs to the Identity Server, and several database queries. Are you aware of this? Isn't there something we can do better here?

I think we can do better here indeed. We probably want to debounce this. We should still occasionally retry while the gateway is being added by the owner. This needs a new issue though.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -468,6 +479,7 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids
 				"schedule_anytime_delay",
 				"schedule_downlink_late",
 				"update_location_from_status",
+				"require_authenticated_connection",
```

johanstokking (approved):

adriansmares (approved):

KrishnaIyer (approved):
Implementation LGTM. As we discussed on Slack, this option will essentially override the `gs.basic-station.allow-unauthenticated` and `gs.tabs-hubs.allow-unauthenticated` per gateway.
gs_options



Please add that to the changelog or the configuration reference of the documentation.

pkg/webui/lib/shared-messages.js
```diff
@@ -294,6 +294,9 @@ export default defineMessages({
   refresh: 'Refresh',
   replaceWebhook: 'Replace webhook',
   removeCollaborator: 'Remove collaborator',
+  requireAuthenticatedConnection: 'Require authenticated connection',
+  requireAuthenticatedConnectionDescription:
+    'This gateway may only connect if it uses an authenticated Basic Station or MQTT connection',
```
```suggestion
    'Controls whether this gateway may only connect if it uses an authenticated Basic Station or MQTT connection',
```

This needs to be reworded to avoid confusion.

kschiffer: Actually the wording of all checkbox descriptions are misleading. I'll open an issue about this.

kschiffer (approved):
LGTM but the checkbox description needs to be reworded. See comment.

Issue #3771, opened 2021-02-09
Configure routing policies


johanstokking:
#### Summary

Configure routing policies.

For listing available networks and tenants, we need https://github.com/packetbroker/iam/issues/2.

#### Why do we need this?

For networks to configure routing policies with other networks

#### What is already there? What do you see now?

This should currently be done with [`pbctl`](https://github.com/packetbroker/pb#configure-routing-policies), but it would be nice to be able to configure routing policies via the CLI and Console.

#### What is missing? What do you want to see?

The ability to configure routing policies between the local Packet Broker network or tenant, and any other network or tenant.

#### How do you propose to implement this?

- Add API to Packet Broker Agent for routing policy management
- Implement CLI support
- Implement Console support. Let's start with manually entering the NetID and tenant ID. With https://github.com/packetbroker/iam/issues/2, we can have a dropdown of public networks and tenants (but still manual entry must be possible)

#### How do you propose to test this?

Integration testing

#### Can you do this yourself and submit a Pull Request?

Can do API and CLI support.

@kschiffer please already think about UX. See linked `pbctl` commands to get an idea of what can be done.

johanstokking:
@kschiffer please start thinking about the UX in the form of wireframes.

The API is designed. See https://github.com/TheThingsNetwork/lorawan-stack/pull/3912/files#diff-4371dadc483aa14e23af85a7baf0e1d57b6d0c67fcf203b94412e13b224cc541 and then especially those `rpc`s.

My idea is as follows:

- Administrators can manage peering for their network. In The Things Stack, this is per deployment. In The Things Stack Cloud, this is per tenant. That is, no application or device level settings (just yet)
- The configuration is technically two-fold: forwarder and home network configuration. Forwarders have gateway infrastructure, home networks have end devices. Most users will play either role, but technically we need to make a distinction here

The UI elements are:

1. Top-level menu item for Packet Broker. My suggestion would be to use the term Packet Broker to push that brand. So the menu item would read **Packet Broker** and not peering
2. In the Console overview page, show the status of Packet Broker (t.b.d., but for now whether the network has an active registration)
3. The Packet Broker settings page. Here, the user can:
   - Register and deregister with Packet Broker, and see their registration info (`/pba/registration` bindings). When there is no registration (`NotFound`), disable all components on the page
[...]

PR #3912, opened 2021-03-09
API to manage peering

johanstokking:
#### Summary

API for #3771

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2798

#### Changes

Add `PacketBrokerAgent` service

#### Testing

N/A

##### Regressions

N/A

#### Notes for Reviewers

- Since we're moving away from gogo, I think it's best not to use gogo for new API. Therefore, `pkg/types.NetID` is not mapped. The `customname` is still necessary for gRPC gateway that expects normalized names
- Elsewhere, we use `bytes` for NetID, but that doesn't work in paths, which is pretty much a necessity in getting and setting routing policies per home network. Therefore it's `uint32`, but we need to decide if this is a good idea and what the alternative would be

#### Checklist


htdvisser (approved):

johanstokking:
I'll reopen when the feature is ready

Issue #2492, opened 2020-05-07
Rethink Gateway Accept Logic


htdvisser:
#### Summary

We may have to change the way the Gateway Server handles new gateway connections. We discussed this on Slack some time ago, and I also commented some of it in https://github.com/TheThingsIndustries/lorawan-stack/pull/1975#issuecomment-613356493, but I forgot to create an actual issue for it.

#### Why do we need this?

Because with the current implementation we may experience some performance issues.

#### What is already there? What do you see now?

The current implementation of the UDP frontend in the Gateway Server is very packet-oriented. This means that on every received UDP packet, it tries to find an existing connection for the EUI in the packet, or create a new connection otherwise. This could mean that on networks with `require-registered-gateways`, it would do a lookup (resulting in a NotFound) for every UDP packet. This could also mean that a single `IP:port` can send packets containing many different EUIs, and practically DoS the Gateway Server.

#### What is missing? What do you want to see?

The implementation of legitimate gateways running the UDP packet forwarder, is actually relatively "connectionful", which we can see from the fact that the source IP:port stays the same for long times. As a result, we don't actually need to try setting up a new connection whenever we receive a UDP packet with a different EUI. We only need to do so when we see an UDP packet from an `IP:port` that we haven't seen before.

If we do it this way, we can quite easily add some extra checks before connecting gateways such as limiting the total number of connections per IP, rate-limiting new connections per IP, or even queueing new connections per IP. We can also reduce the load on networks with `require-registered-gateways`, since we only have to lookup the gateway once when the "connection" starts instead of on every packet.

#### How do you propose to implement this?

- Keep a `map[RemoteAddr]*UDPConn` instead of a `map[EUI]*state`
	- The lifetime is for as long as the connection doesn't time out, so we won't have to retry connecting a gateway on each UDP packet we receive.
	- The `UDPConn` has a buffered channel for raw UDP packets, so we have "natural" rate limiting and flood protection.
- When the GS receives a UDP packet for an existing connection (it's in the map), it just writes the packet to the connection's channel for raw UDP packets.
	- After writing the entry to the channel, the GS can immediately continue reading the next UDP packet.
	- This should be done with a `select`, since the connection's buffered channel may be full and we still want to keep handling other connections.
- When the GS receives a UDP packet for a new connection (not yet in the map), create a new entry in the map, and add the `UDPConn` to some "accept queue"
[...]

PR #3920, opened 2021-03-12
Rate Limiting

neoaggelos:
#### Summary

References #1406

#### Changes

- Introduce new `ratelimit` package, with a base `RateLimiter` interface, as well as middleware for gRPC, http, echo.
- Introduce rate limiting configuration for components.
- Rate limit gRPC unary methods (by remote IP and method name; supports config to override per method)
- Rate limit gRPC new streams and incoming stream traffic (by remote IP and method name; supports config to override per method)
- Rate limit GCS using HTTP middleware (by remote IP and request URL)
- Rate limit AS MQTT connections and incoming traffic (i.e. downlinks) by remote IP. 
- Rate limit GS MQTT frontend (new connections by remote IP, traffic by remote address)
- Rate limit GS UDP frontend (by remote IP address)
- Rate limit GS gRPC frontend
- Rate limit GS BasicStation frontend (connections by remote IP, traffic by gateway UID)
- Unit tests

#### Testing

- Unit tests
- Extensive integration testing
  - [X] Test rate limits for gRPC unary methods and streams using CLI
  - [X] Test config overrides for gRPC methods
  - [X] Test with a real UDP gateway
  - [ ] Test with a basicstation gateway
  - [ ] Test with a gRPC gateway (with simulate CLI commands)
  - [ ] Test AS MQTT connections (reject connections, reject downlinks)
  - [x] Test GCS endpoints

##### Regressions

There should not be any, all config is fully backwards-compatible.

The Redis dependencies of throttled are forcing an update of protobuf to 1.4.2. As a temporary workaround, I have created a fork where these dependencies are removed, but we need to figure out a proper solution.

#### Notes for Reviewers

@bafonins @kschiffer @pgalic96 @KrishnaIyer @adriansmares GitHub requested review automatically. No need to go over the whole PR, only make sure your code-owned files look OK.

#### Checklist


cmd/ttn-lw-stack/commands/start.go
```diff
@@ -292,11 +330,18 @@ var startCommand = &cobra.Command{
 					Redis: redis.New(config.Redis.WithNamespace("as", "io", "webhooks")),
 				}
 			}
-			fetcher, err := config.AS.EndDeviceFetcher.NewFetcher(c)
+			config.AS.MQTT.RateLimiting.Limiter, err = config.AS.MQTT.RateLimiting.Connections.New(c.Context())
+			if err != nil {
+				return shared.ErrInitializeGatewayServer.WithCause(err)
```
htdvisser: I don't think `ErrInitializeGatewayServer` is appropriate here

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -315,6 +367,18 @@ func (c *connection) CanWrite(info *auth.Info, topicParts ...string) bool {
...
 func (c *connection) deliver(pkt *packet.PublishPacket) {
 	logger := log.FromContext(c.io.Context()).WithField("topic", pkt.TopicName)
+
+	if c.limit() {
+		// TODO: observability
+		err := errTrafficRateLimitExceeded.WithAttributes(
+			"application_uid", unique.ID(c.io.Context(), c.io.ApplicationIDs()),
+			"remote_addr", c.mqtt.RemoteAddr().String(),
+		)
+		logger.WithError(err).Error("Terminating connection")
```
htdvisser: This shouldn't be an `Error` log; those are reserved for things that shouldn't happen (possible bugs). A `Warn` log is better here.

pkg/config/config.go
```diff
@@ -585,6 +586,8 @@ func (m *Manager) setDefaults(prefix string, flags *pflag.FlagSet, config interf
 				}
 				flags.StringP(name, shorthand, def, description)
 
+			case []ratelimit.Profile:
```
htdvisser, line 589: Why is this necessary? Can't we just tell the config package to ignore the config with `name:"-"`?

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -69,9 +94,35 @@ func (s *srv) accept() error {
 			return err
 		}
 
+		remoteAddr := mqttConn.RemoteAddr().String()
+		remoteIP, _, err := net.SplitHostPort(remoteAddr)
+		if err != nil {
+			remoteIP = remoteAddr // TODO: this is a hacky fallback
+		}
```
htdvisser: If this ever happens, there's something very wrong in the standard library.

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -69,9 +94,35 @@ func (s *srv) accept() error {
...
-			conn := &connection{server: s.server, mqtt: mqttConn, format: s.format}
+			remoteAddr := remoteAddr
+			ctx := log.NewContextWithFields(s.ctx, log.Fields("remote_addr", remoteAddr))
+			conn := &connection{
+				server: s.server,
+				mqtt:   mqttConn,
+				format: s.format,
+				limit: func() bool {
+					// TODO: filter by application UID and remote IP (and possibly auth token ID) instead
+					limit, _ := s.trafficLimiter.RateLimit(remoteAddr, "")
```
htdvisser: Isn't it much simpler to just rate limit on this particular connection?

pkg/gatewayserver/io/udp/udp.go
```diff
@@ -96,6 +108,16 @@ func (s *srv) read() (err error) {
 			}
 			return err
 		}
+
+		// TODO: Should UDP port be a part of this?
```
htdvisser: We need to re-think the connection logic of UDP gateways anyway (https://github.com/TheThingsNetwork/lorawan-stack/issues/2492). Then we can rate limit _new_ "connections" by IP address, and rate limit messages on each individual "connection".

pkg/ratelimit/http.go
```diff
@@ -0,0 +1,69 @@
...
+
+// HTTPRemoteIP returns a key that varies by remote IP, honoring X-Forwarded-For headers.
+func HTTPRemoteIP(r *http.Request) string {
+	if xff := r.Header.Get("x-forwarded-for"); xff != "" {
+		// TODO: only for trusted proxies
+		s := strings.Split(xff, ",")[0]
+		if s != "" {
+			return strings.Trim(s, " ")
+		}
+	}
```
htdvisser: We already have the `ProxyHeaders` middleware that sets `X-Real-IP` when a trusted proxy sets those headers (and otherwise unsets it)

pkg/ratelimit/grpc.go
```diff
@@ -0,0 +1,125 @@
...
+
+// GRPCVaryBy returns a rate limit key from a request context and the full method name.
+type GRPCVaryBy func(ctx context.Context, fullMethod string) string
+
+// GRPCRemoteIP returns the remote IP of a gRPC request.
+func GRPCRemoteIP(ctx context.Context, fullMethod string) string {
+	if md := rpcmetadata.FromIncomingContext(ctx); md.XForwardedFor != "" {
+		xff := strings.Split(md.XForwardedFor, ",")
+		return strings.Trim(xff[0], " ")
+	}
```
htdvisser: We have the `ProxyHeaders` middleware that sets `x-real-ip` when a trusted proxy sets those headers (and otherwise unsets it)

pkg/config/config.go
neoaggelos: The `[]ratelimit.Profile` is user configuration, but only accepted from a config file and not from command line arguments.

pkg/applicationserver/io/mqtt/mqtt.go
johanstokking: Also it should be imperative

neoaggelos:
Testing is still in progress, but the implementation should be ready for review. It might be better to review along with the documentation reference (https://github.com/TheThingsIndustries/lorawan-stack-docs/pull/290).

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -69,8 +75,23 @@ func (s *srv) accept() error {
 			return err
 		}
 
+		remoteAddr := mqttConn.RemoteAddr().String()
+		remoteIP, _, _ := net.SplitHostPort(remoteAddr)
+		key := fmt.Sprintf("as:accept:mqtt:ip:%s", remoteIP)
```
johanstokking: Is this safe?

```suggestion
		rateLimitKey := mqttConn.RemoteAddr().String()
		if host, _, err := net.SplitHostPort(rateLimitKey); err == nil {
			rateLimitKey = host
		}
		key := fmt.Sprintf("as:accept:mqtt:ip:%s", rateLimitKey)
```

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -69,8 +75,23 @@ func (s *srv) accept() error {
...
+
+		if limit, _ := s.server.RateLimiter().RateLimit(key, []string{"as:accept"}); limit {
```
johanstokking: `RateLimit()` probably wants to accept `...string`. See below.

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -69,8 +75,23 @@ func (s *srv) accept() error {
...
+			err := errRateLimitExceeded.WithAttributes("remote_addr", remoteAddr)
+			// TODO: this will lead to a flood of log messages
+			log.FromContext(s.ctx).WithError(err).Warn("Drop connection")
```
johanstokking: Please address

pkg/ratelimit/rate_limit.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"context"
+
+	"github.com/throttled/throttled"
+	"go.thethings.network/lorawan-stack/v3/pkg/log"
+)
+
+// RateLimiter can be used to rate limit access to a resource (using a key string).
+type RateLimiter interface {
+	RateLimit(key string, class []string) (bool, *Result)
```
johanstokking: The parameters and the return values are a bit off.

First, there's no godoc to explain them. What is class? Why not classes? Why not `classes ...string`? I guess the first return value is to rate limit. But what is the second return value for? Can it be `nil`? What does it mean with regard to the first return value?

Second, did you consider a pattern where there's a method that returns `error` when rate limited is engaged, which can be returned? That safes some hassle with declaring errors all the time. An example of this is like `RequireApplication()` rights.

Usage would be:

```go
if err := ratelimit.RequirePass(key, "as:accept"); err != nil {
	return err
}
```

Or something like this, you get the idea.

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -69,8 +75,23 @@ func (s *srv) accept() error {
...
+			if err := mqttConn.Close(); err != nil {
+				log.FromContext(s.ctx).WithError(err).Warn("Close connection failed")
+			}
+			return err
+		}
+
 		go func() {
-			ctx := log.NewContextWithFields(s.ctx, log.Fields("remote_addr", mqttConn.RemoteAddr().String()))
+			remoteAddr := remoteAddr
```
johanstokking: `remoteAddr` may have changed before the goroutine reaches this line. Pass the `remoteAddr` as variable to the func

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -252,6 +275,10 @@ func (c *connection) Connect(ctx context.Context, info *auth.Info) (context.Cont
 		return nil, err
 	}
 	ctx = c.io.Context()
+
+	_, authTokenID, _, _ := ttsauth.SplitToken(string(info.Password))
+	c.limitKey = fmt.Sprintf("as:down:app:%s:token:%s", uid, authTokenID)
```
johanstokking: I'm quite sure this needs a fallback too. We really don't want `authTokenID` to be empty here.

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -315,6 +342,18 @@ func (c *connection) CanWrite(info *auth.Info, topicParts ...string) bool {
 
 func (c *connection) deliver(pkt *packet.PublishPacket) {
 	logger := log.FromContext(c.io.Context()).WithField("topic", pkt.TopicName)
+
+	if limit, _ := c.server.RateLimiter().RateLimit(c.limitKey, []string{"as:down"}); limit {
+		_, authTokenID, _, _ := ttsauth.SplitToken(string(c.session.AuthInfo().Password))
```
johanstokking: Same

pkg/basicstation/cups/server.go
```diff
@@ -196,7 +197,7 @@ func NewServer(c *component.Component, options ...Option) *Server {
 
 // RegisterRoutes implements web.Registerer
 func (s *Server) RegisterRoutes(web *web.Server) {
-	web.POST("/update-info", s.UpdateInfo)
+	web.POST("/update-info", s.UpdateInfo, ratelimit.EchoMiddleware(s.component.RateLimiter(), "http:cups"))
```
johanstokking: Probably `basicstation:cups`?

pkg/component/rate_limit.go
```diff
@@ -0,0 +1,21 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package component
+
+import "go.thethings.network/lorawan-stack/v3/pkg/ratelimit"
+
+func (c *Component) RateLimiter() ratelimit.RateLimiter {
```
johanstokking: `ratelimit.RateLimiter` stutters. Maybe `ratelimit.Interface` or `ratelimit.RL`?

pkg/component/web.go
```diff
@@ -70,6 +71,7 @@ func (c *Component) initWeb() error {
 
 	if c.config.HTTP.PProf.Enable {
 		g := web.RootRouter().NewRoute().Subrouter()
+		g.Use(ratelimit.HTTPMiddleware(c.RateLimiter(), "http:admin"))
```
johanstokking: Probably `probe` is better here. `admin` is quite confusing as it implies there's a user involved.

pkg/component/web.go
```diff
@@ -84,6 +86,7 @@ func (c *Component) initWeb() error {
 
 	if c.config.HTTP.Metrics.Enable {
 		g := web.RootRouter().NewRoute().Subrouter()
+		g.Use(ratelimit.HTTPMiddleware(c.RateLimiter(), "http:admin"))
```
johanstokking: Idem

pkg/component/web.go
```diff
@@ -95,12 +98,14 @@ func (c *Component) initWeb() error {
...
 	if c.config.HTTP.Health.Enable {
 		g := web.RootRouter().NewRoute().Subrouter()
+		g.Use(ratelimit.HTTPMiddleware(c.RateLimiter(), "http:admin"))
 		if c.config.HTTP.Health.Password != "" {
 			g.Use(mux.MiddlewareFunc(webmiddleware.BasicAuth(
 				"health",
 				webmiddleware.AuthUser(healthUsername, c.config.HTTP.Health.Password),
 			)))
 		}
+		g.Use(ratelimit.HTTPMiddleware(c.RateLimiter(), "http:admin"))
```
johanstokking: Idem

pkg/gatewayserver/io/grpc/grpc.go
```diff
@@ -108,6 +112,13 @@ func (s *impl) LinkGateway(link ttnpb.GtwGs_LinkGatewayServer) error {
 
 	go func() {
 		for {
+			key := fmt.Sprintf("gs:up:gtw:%s", uid)
+			if limit, _ := s.server.RateLimiter().RateLimit(key, []string{"gs:up"}); limit {
```
johanstokking: Why do we need `gs:up` twice?

pkg/gatewayserver/io/mqtt/mqtt.go
```diff
@@ -296,6 +321,15 @@ func (c *connection) CanWrite(info *auth.Info, topicParts ...string) bool {
 
 func (c *connection) deliver(pkt *packet.PublishPacket) {
 	logger := log.FromContext(c.io.Context()).WithField("topic", pkt.TopicName)
+
+	if limit, _ := c.server.RateLimiter().RateLimit(c.limitKey, []string{"gs:up"}); limit {
+		// TODO: observability
```
johanstokking: Remove, address or file issue. This PR isn't draft anymore.

pkg/gatewayserver/io/udp/udp.go
```diff
@@ -108,6 +111,16 @@ func (s *srv) read() (err error) {
 		}
 		now := time.Now()
 
+		// TODO: verify that this works behind a reverse proxy
```
johanstokking: Idem

johanstokking (changes requested):
Revert changes to `data/lorawan-webhook-templates` and please check diff on GitHub before requesting review (or making the PR ready for review). You'd see those unrelated changes. Also there are quite some `TODO`s left that aren't addressed here.

I didn't do a full review yet; this needs another pass. Please request review when my remarks are addressed or answered.

go.mod
```diff
@@ -25,6 +25,10 @@ replace github.com/eclipse/paho.mqtt.golang => github.com/eclipse/paho.mqtt.gola
 // Versions higher trigger google/protobuf update past v1.3.5.
 replace github.com/onsi/gomega => github.com/onsi/gomega v1.10.0
 
+// Optional dependencies of throttled/v2 update golang/protobuf past v1.3.5.
+// TODO: replace personal fork
```
htdvisser: Forked: https://github.com/TheThingsIndustries/throttled

pkg/component/web.go
htdvisser: Let's just call this `pprof`; that's also what we call it in config.

pkg/component/web.go
htdvisser: Let's just call this `metrics`; that's also what we call it in config.

pkg/component/web.go
```diff
@@ -95,12 +98,14 @@ func (c *Component) initWeb() error {
 
 	if c.config.HTTP.Health.Enable {
 		g := web.RootRouter().NewRoute().Subrouter()
+		g.Use(ratelimit.HTTPMiddleware(c.RateLimiter(), "http:admin"))
```
htdvisser: 1. Let's just call this `health`; that's also what we call it in config.
2. Let's only rate limit when password protection is enabled

pkg/component/web.go
htdvisser: Isn't this double?

htdvisser (commented):
Please also add `pkg/ratelimit` to CODEOWNERS

pkg/applicationserver/io/mqtt/mqtt.go
neoaggelos: Removed for now. Is it worth adding a metric for this?

pkg/ratelimit/rate_limit.go
neoaggelos: Switched to a `Resource` interface and added missing documentation. The second value is rate limiting metadata, and are used to set response headers for HTTP and gRPC requests. This has the advantage that we define all rate limiting keys in a single place.

Also added a `ratelimit.Require` method for cases where we do not really care about such metadata (basically anything that does not involve a request context).

Should we specialize this into `ratelimit.RequireGatewayUp()`, `ratelimit.RequireApplicationDown()`, ....? I am a bit hesitant because that would basically recompute the rate limiting key for every iteration even when it is fixed.

pkg/applicationserver/io/mqtt/mqtt.go
neoaggelos: Updated. Is this the case for `mqttConn` as well?

pkg/gatewayserver/io/grpc/grpc.go
neoaggelos: The resource classes are not prepended to the rate limiting key. This is to avoid confusion in cases where multiple resource classes are set (e.g. for gRPC requests)

pkg/ratelimit/echo.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// EchoMiddleware is an Echo middleware that rate limits HTTP requests by remote IP and request URL.
+func EchoMiddleware(limiter Interface, class string) echo.MiddlewareFunc {
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) (err error) {
+			limit, result := limiter.RateLimit(echoRequestResource(c, class))
+			result.SetHTTPHeaders(c.Response().Header())
+			if limit {
+				// TODO: observability
```
htdvisser: Reference issue

htdvisser (approved):

pkg/ratelimit/rate_limit.go
johanstokking: > Should we specialize this into `ratelimit.RequireGatewayUp()`, `ratelimit.RequireApplicationDown()`, ....? I am a bit hesitant because that would basically recompute the rate limiting key for every iteration even when it is fixed.

No we shouldn't indeed

pkg/applicationserver/io/mqtt/mqtt.go
johanstokking: What is resolved about this?

Fine to resolve if you disagree, but then at least comment here.

pkg/applicationserver/io/mqtt/mqtt.go
johanstokking: No. And actually, this doesn't seem necessary in this case, as both variables are declared in this scope, i.e. within the loop.

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -69,8 +71,18 @@ func (s *srv) accept() error {
 			return err
 		}
 
-		go func() {
-			ctx := log.NewContextWithFields(s.ctx, log.Fields("remote_addr", mqttConn.RemoteAddr().String()))
+		remoteAddr := mqttConn.RemoteAddr().String()
+		ctx := log.NewContextWithFields(s.ctx, log.Fields("remote_addr", remoteAddr))
```
johanstokking, line 75: `ctx` is declared here

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -81,7 +93,7 @@ func (s *srv) accept() error {
 				mqttConn.Close()
 				return
 			}
-		}()
+		}(ctx)
```
johanstokking: So no need to pass it

pkg/applicationserver/io/mqtt/mqtt.go
```diff
@@ -252,6 +266,13 @@ func (c *connection) Connect(ctx context.Context, info *auth.Info) (context.Cont
 		return nil, err
 	}
 	ctx = c.io.Context()
+
+	authTokenID := ""
+	if _, v, _, err := ttsauth.SplitToken(string(info.Password)); err != nil && v != "" {
+		authTokenID = v
+	}
+	c.resource = ratelimit.ApplicationMQTTDownResource(ctx, ids, authTokenID)
```
johanstokking, line 275: So if `authTokenID` is empty, the rate limiting pool is shared?

What if the client is DoSing with a random password?

Shouldn't we hash the password instead, if splitting the token doesn't work?

pkg/gatewayserver/io/mqtt/mqtt.go
```diff
@@ -84,7 +95,7 @@ func (s *srv) accept() error {
 				mqttConn.Close()
 				return
 			}
-		}()
+		}(ctx)
```
johanstokking: Idem

pkg/gatewayserver/io/udp/udp.go
```diff
@@ -107,11 +110,18 @@ func (s *srv) read() (err error) {
 			return err
 		}
 		now := time.Now()
+		ctx := log.NewContextWithField(s.ctx, "remote_addr", addr.String())
+
+		if err := ratelimit.Require(s.server.RateLimiter(), ratelimit.GatewayUDPTrafficResource(addr)); err != nil {
+			if ratelimit.Require(s.limitLogs, ratelimit.NewCustomResource(addr.IP.String())) != nil {
```
johanstokking: Shouldn't this be `== nil`?

pkg/ratelimit/rate_limit.go
```diff
@@ -0,0 +1,87 @@
...
+			return limiter.RateLimit(resource)
+		}
+	}
+	return l.defaultLimiter.RateLimit(resource)
+}
+
+// Require checks that the rate limit for a Resource has not been exceeded.
+func Require(limiter Interface, resource Resource) error {
+	if limit, _ := limiter.RateLimit(resource); limit {
+		return errRateLimitExceeded.New()
```
johanstokking: It could be helpful to include some metadata here as error attributes

pkg/ratelimit/rate_limit.go
```diff
@@ -0,0 +1,87 @@
...
+	for _, c := range resource.Classes() {
+		if limiter, ok := l.limiters[c]; ok {
+			return limiter.RateLimit(resource)
+		}
+	}
+	return l.defaultLimiter.RateLimit(resource)
+}
+
+// Require checks that the rate limit for a Resource has not been exceeded.
+func Require(limiter Interface, resource Resource) error {
```
johanstokking, line 81: Not sure what to think when I see `ratelimit.Require()`... that is requires rate limiting to be engaged or the opposite?

What do you think @htdvisser ?

johanstokking (commented):
@KrishnaIyer please review the `pkg/gatewayserver/io/...` stuff, especially the websockets implementation and testing

@neoaggelos aren't we missing rate limiting on AS webhooks?

pkg/gatewayserver/io/udp/udp.go
```diff
@@ -69,6 +68,10 @@ func Serve(ctx context.Context, server io.Server, conn *net.UDPConn, config Conf
 	if config.RateLimiting.Enable {
 		firewall = NewRateLimitingFirewall(firewall, config.RateLimiting.Messages, config.RateLimiting.Threshold)
 	}
+	limitLogs, err := limitLogsConfig.New(ctx, limitLogsSize)
+	if err != nil {
+		panic(err)
```
adriansmares: Return `err`. No need to `panic`.

pkg/applicationserver/io/mqtt/mqtt.go
adriansmares: I don't think this whole check is a good idea in the first place. If the connection attempt is proxified, the rate limiting becomes virtually global.

pkg/gatewayserver/io/mqtt/mqtt.go
```diff
@@ -72,8 +73,18 @@ func (s *srv) accept() error {
 			return err
 		}
 
+		remoteAddr := mqttConn.RemoteAddr().String()
```
adriansmares, line 76: Are we sure this does what we think it does ? `RemoteAddr` contains the port as well, and as far as I know the port will differ on each connection attempt. Also this is open to collisions due to reverse-proxies.

pkg/gatewayserver/io/udp/udp.go
neoaggelos: Opted to `panic` because this is coming from internal configuration, so an error message is not really helpful.

pkg/gatewayserver/io/mqtt/mqtt.go
neoaggelos: The address port is stripped in `ratelimit.GatewayAcceptMQTTConnectionResource`.

> Also this is open to collisions due to reverse-proxies.

On the client side, we cannot really do much for this (since for example all gateways behind a NAT will always have the same remote IP address). 

If MQTT (or UDP) endpoints are served behind a reverse proxy, then indeed this will not work. I believe TTN and TTSC are not affected by this, since the endpoints are not behind a reverse proxy.

It is definitely worth adding a note for this on the documentation.

cc @htdvisser

htdvisser: As I already wrote to @neoaggelos, we should be able to configure "trusted" IP addresses that bypass rate limiting for such situations. 

For now, we should just document that rate limiting must be done by the proxy (and disabled in The Things Stack) if a reverse proxy is used.

In the future we may support PROXY protocol to get the real client IP from load balancers that support it.

neoaggelos: Perhaps a better solution would be to take the gateway ID into consideration for the rate limiting key? An issue would be that the MQTT frontend does too many things before terminating the connection in that case.

pkg/applicationserver/io/mqtt/mqtt.go
neoaggelos: This is only ever reached if `c.server.Subscribe(ctx, "mqtt", &ids, true)` succeeds, which means that the client has supplied valid auth.

pkg/gatewayserver/io/mqtt/mqtt.go
adriansmares: The gateway ID is not really available at this level - this is the TCP accept routine, what is 'one the wire' is still unknown at this layer.

I'm ok with the trusted source IPs + documentation, and agree that there isn't much that can be done at this layer.

kschiffer (approved):
Codeowned files LGTM

pkg/gatewayserver/io/udp/udp.go
adriansmares: I understand that the intent is to stop the stack, by force if necessary, if the configuration is invalid. But I'd prefer that `limitLogsSize` is validated as part of the configuration itself, instead of here. You can return the error and never start the stack at that point.

I don't think there is any reason to have `panic(err)` in production code.

johanstokking (approved):

johanstokking:
Great work @neoaggelos !

PR #3922, opened 2021-03-15
Add checks for admin and universal rights in context

johanstokking:
#### Summary

Groundwork for #3771

#### Changes

Add ~`RequireAuthenticated()`,~ `RequireIsAdmin()` and `RequireUniversal()`

#### Testing

Unit testing

##### Regressions

None

#### Notes for Reviewers

This is integrated in `pkg/auth/rights`. There was some distinction necessary between entity rights and authentication info, which provides non-entity rights. I'm open to suggestions in naming things.

Example usage:

```go
func (s *pbaServer) Register(ctx context.Context, _ *pbtypes.Empty) (*ttnpb.PacketBrokerRegistration, error) {
	if err := rights.RequireIsAdmin(ctx); err != nil {
		return nil, err
	}
	...
}
```

#### Checklist

### DIFF
--- a/pkg/auth/rights/require.go
+++ b/pkg/auth/rights/require.go
@@ -24,6 +24,22 @@ import (
 
 // Errors for no/insufficient rights.
 var (
+	ErrUnauthenticated = errors.DefineUnauthenticated(
+		"unauthenticated",
+		"unauthenticated",
+	)
+	ErrNoUniversalRights = errors.DefinePermissionDenied(
+		"no_universal_rights",
+		"no universal rights",
+	)
+	ErrInsufficientUniversalRights = errors.DefinePermissionDenied(
+		"insufficient_universal_rights",
+		"insufficient universal rights",
+	)
+	ErrNoAdmin = errors.DefinePermissionDenied(
+		"no_admin",
+		"no admin",
+	)
 	ErrNoApplicationRights = errors.DefinePermissionDenied(
 		"no_application_rights",
 		"no rights for application `{uid}`",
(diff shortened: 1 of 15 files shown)


pkg/auth/rights/require.go
```diff
@@ -66,9 +82,57 @@ var (
 	)
 )
 
+// RequireAuthenticated checks that the context is authenticated.
+func RequireAuthenticated(ctx context.Context) error {
+	authInfo, err := AuthInfo(ctx)
+	if err != nil {
+		return err
+	}
```
johanstokking: @htdvisser this is not entirely in sync with `(*IdentityServer).RequireAuthenticated()`: there it fetches the user and checks whether the user is valid.

Should we do that here too, or is that happening implicitly somewhere? If we need to do that, how, as we don't have an entity fetcher in the context.

See

https://github.com/TheThingsNetwork/lorawan-stack/blob/fb3fc93055d138fcae9bbfaa4c3da9b2579a70fb/pkg/identityserver/entity_access.go#L299-L303

htdvisser: It's quite important to check user state in the `RequireAuthenticated`, otherwise an unapproved user in a network that requires admin approval can perform actions.

Maybe a better question is: do we actually need this? In the issue you're addressing (https://github.com/TheThingsNetwork/lorawan-stack/issues/3771) you only seem to need the `RequireIsAdmin`.

johanstokking: The only place that I was going to use this for is [`GetRegistration()`](https://github.com/TheThingsNetwork/lorawan-stack/pull/3912/files#diff-4371dadc483aa14e23af85a7baf0e1d57b6d0c67fcf203b94412e13b224cc541R174) so that any authenticated user can see whether PB is enabled and what the network name is, DevAddr blocks etc. Just informative.

I can do a few things:

1. Raise required privileges for viewing registration info to admin
2. Introduce a new universal right for viewing registration info (and maybe also for the other rpcs), assuming that granted universal rights imply an approved and not suspended user (?)
3. Get the user state here, but how?

johanstokking: I think the cleanest is to ditch `RequireAuthenticated` and use `RequireUniversal` with new rights, or require admin rights for everything.

htdvisser: For the initial implementation, let's just require admin. We can always loosen that later.

KrishnaIyer (approved):
Code owned test files LGTM

pkg/auth/rights/require.go
johanstokking: OK, `RequireAuthenticated` is gone

htdvisser (approved):

PR #3961, opened 2021-03-19
Improve Application Server end device fetcher network partition resilience

### LINKED ISSUE #3942: Improve Application Server end device fetcher network partition resilience
#### Summary

The end device fetcher used by the Application Server in order to retrieve the locations needs to be more robust when the IS peer is not available.

#### Why do we need this?

Since the location 'injection' is on the hotpath, we need to ensure that we don't freeze the AS when the IS is not available.

The cached end device fetcher caches successful requests.

- The context used to call the fetcher should have a timeout. I propose a 5 second timeout, and we may make it configurable in the future. 
- We should deduplicate the requests via a `singleflight` group - see https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.11/pkg/messageprocessors/devicerepository/devicerepository.go .
- We should cache errors as well as successful requests.

#### Environment

`3.11.3`

- Before calling the end device fetcher, create a context with a deadline of 5 seconds.
- Deduplicate requests using the same UID strategy that is used for caching.
- Have an error cache.

This is actually pretty hard to test. Nonetheless, one may start only the IS component and then rest of the stack separately, and attempt to fly traffic after the IS component is killed.

Yes

adriansmares:
#### Summary

Closes #3942

#### Changes

- Cache end device fetch errors as part of the cached end device fetcher.
- Add and use a `singleflight.Group` end device fetcher.
- Add and use a timeout end device fetcher, that automatically adds a timeout to the context used for the `Get` call.
- Add a circuit breaker end device fetcher - when a burst of errors occurs, the circuit breaker opens and no `Get` calls are made - only an error is returned. The circuit breaker automatically closes and allows new attempts to be made after a specified timeout.

#### Testing

Unit testing.

##### Regressions

This affects the end device fetcher functionality of the Application Server. The blast radius of the changes is that the end device location is not going to be visible in the upstream messages. We can test this by simulating a network partition between AS and IS and seeing how traffic is affected.

#### Notes for Reviewers

The circuit breaker is outside the scope of the initial issue, but nonetheless I felt that the timeout and error caching will not help for windows of 15 minutes during which the IS is not available (with a timeout of 5 seconds per call, we're limited at 0.2 pps since the cache is per device).

#### Checklist

### DIFF
--- a/pkg/applicationserver/device_fetcher.go
+++ b/pkg/applicationserver/device_fetcher.go
@@ -69,6 +69,11 @@ type cachedEndDeviceFetcher struct {
 	cache   gcache.Cache
 }
 
+type endDeviceFetcherCacheEntry struct {
+	err error
+	dev *ttnpb.EndDevice
+}
+
 // NewCachedEndDeviceFetcher wraps an EndDeviceFetcher with a local cache.
 func NewCachedEndDeviceFetcher(fetcher EndDeviceFetcher, cache gcache.Cache) EndDeviceFetcher {
 	return &cachedEndDeviceFetcher{fetcher, cache}
@@ -76,15 +81,16 @@ func NewCachedEndDeviceFetcher(fetcher EndDeviceFetcher, cache gcache.Cache) End
 
 // Get implements the EndDeviceFetcher interface.
 func (f *cachedEndDeviceFetcher) Get(ctx context.Context, ids ttnpb.EndDeviceIdentifiers, fieldMaskPaths ...string) (*ttnpb.EndDevice, error) {
-	uid := unique.ID(ctx, ids)
-	key := fmt.Sprintf("%s:%s", uid, strings.Join(fieldMaskPaths, ","))
-	dev, err := f.cache.Get(key)
-	if err != nil {
-		dev, err := f.fetcher.Get(ctx, ids, fieldMaskPaths...)
-		if err == nil {
-			f.cache.Set(key, dev)
-		}
-		return dev, err
+	key := endDeviceKey(ctx, ids, fieldMaskPaths...)
+	e, err := f.cache.Get(key)
+	if entry, ok := e.(*endDeviceFetcherCacheEntry); err == nil && ok {
+		return entry.dev, entry.err
 	}
-	return dev.(*ttnpb.EndDevice), err
+	dev, err := f.fetcher.Get(ctx, ids, fieldMaskPaths...)
+	f.cache.Set(key, &endDeviceFetcherCacheEntry{err, dev})
+	return dev, err
+}
+
+func endDeviceKey(ctx context.Context, ids ttnpb.EndDeviceIdentifiers, fieldMaskPaths ...string) string {
+	return fmt.Sprintf("%s:%s", unique.ID(ctx, ids), strings.Join(fieldMaskPaths, ","))
 }
--- a/pkg/applicationserver/device_fetcher.go
+++ b/pkg/applicationserver/device_fetcher.go
@@ -25,6 +25,7 @@ import (
 	pbtypes "github.com/gogo/protobuf/types"
 	"go.thethings.network/lorawan-stack/v3/pkg/component"
 	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/log"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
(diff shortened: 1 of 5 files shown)


pkg/applicationserver/device_fetcher.go
```diff
@@ -69,22 +73,129 @@ type cachedEndDeviceFetcher struct {
...
+	}
+	// At this point we have a number of failures that is above the threshold, and any attempts are recent.
+	// The circuit breaker is open.
+	return errCircuitBreakerOpen.New()
+}
+
+func (f *circuitBreakerEndDeviceFetcher) observeError(err error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	if err != nil {
```
htdvisser: I think we should consider the error type here. It's relatively easy to trigger NotFound errors by deleting an end device from the IS without deleting it from the AS. If you'd coordinate this across multiple devices you can relatively easily trigger the circuit breaker for other users in the network.

pkg/applicationserver/device_fetcher.go
```diff
@@ -69,22 +73,129 @@ type cachedEndDeviceFetcher struct {
...
+	if f.failures < f.threshold {
+		return nil
+	}
+	// If the attempt timeout expired, consider the circuit as being closed.
+	if time.Now().Sub(f.lastFailedAttempt) > f.timeout {
+		return nil
+	}
+	// At this point we have a number of failures that is above the threshold, and any attempts are recent.
+	// The circuit breaker is open.
+	return errCircuitBreakerOpen.New()
```
johanstokking, line 175: Can we write a log entry here with `WARN` level, same with `INFO` as when it closes (implicitly, for the first time)? That would be very handy for figuring out what is going on.

adriansmares:
I've rebased the PR but only the last commit (279f60883) contains changes.

The logs have been added and we can expect the log stream to look as follows:

```
== RUN   TestEndDeviceFetcher/CircuitBreaker
=== RUN   TestEndDeviceFetcher/CircuitBreaker/InitialClosed
=== RUN   TestEndDeviceFetcher/CircuitBreaker/InitialBurst
  WARN Circuit breaker open                     circuit=end_device_fetcher count=10 error=error:pkg/applicationserver_test:unavailable (server unavailable) test_name=TestEndDeviceFetcher
=== RUN   TestEndDeviceFetcher/CircuitBreaker/BreakerOpen
=== RUN   TestEndDeviceFetcher/CircuitBreaker/SecondBurst
  WARN Circuit breaker open                     circuit=end_device_fetcher count=11 error=error:pkg/applicationserver_test:unavailable (server unavailable) test_name=TestEndDeviceFetcher
  WARN Circuit breaker open                     circuit=end_device_fetcher count=12 error=error:pkg/applicationserver_test:unavailable (server unavailable) test_name=TestEndDeviceFetcher
=== RUN   TestEndDeviceFetcher/CircuitBreaker/BreakerClosed
  INFO Circuit breaker closed                   circuit=end_device_fetcher previous_count=12 test_name=TestEndDeviceFetcher
```

The circuit breaker counts now only errors with the following codes:
- Cancelled
- Deadline Exceeded
- Aborted
- Internal
- Unavailable

pkg/applicationserver/device_fetcher.go
```diff
@@ -174,15 +175,39 @@ func (f *circuitBreakerEndDeviceFetcher) circuitOpen() error {
...
 	f.mu.Lock()
 	defer f.mu.Unlock()
-	if err != nil {
-		f.lastFailedAttempt = time.Now()
-		f.failures++
-	} else {
-		f.lastFailedAttempt = time.Time{}
-		f.failures = 0
+	switch {
+	case infrastructureError:
```
johanstokking: What's the benefit of the var?

pkg/applicationserver/device_fetcher.go
```diff
@@ -174,15 +175,39 @@ func (f *circuitBreakerEndDeviceFetcher) circuitOpen() error {
...
+		{
```
johanstokking: No braces needed

johanstokking (approved):

pkg/applicationserver/device_fetcher.go
adriansmares: The error code check is done outside the mutex, to minimize retention. ~I can't tell if this is premature or not.~

Edit: it is premature, reverting.

johanstokking (approved):

PR #3967, opened 2021-03-22
Fix payload size limits in AU915

### LINKED ISSUE #3938: Incorrect maximum downlink sizes in AU915
#### Summary
The downlink sizes in https://github.com/TheThingsNetwork/lorawan-stack/blob/d4791aeafa01f8f85eb258c534a5706a1d0c4a74/pkg/band/au_915_928.go#L95 are based on older versions of the regional parameters, which appear to have a clerical error in the repeater-compatible size list. Repeater compatibility requires ensuring that the packets can be encapsulated in a large (242 byte payload / 250 byte MACPayload / 255 byte PHYPayload) lorawan packet. This means reducing the maximum user payload size by 20 bytes on the high end only. But by mistake, the editor seems to have reduced ALL the sizes by 20 bytes. This mistake seems to be fixed in the latest regional parameters.

I think TTN should correct the values 41 and 117 in the table, to 61 and 137. This is consistent with the latest regional parameters, and with all of the other regions in the older regional parameters. And devices should be happy to accept the larger sizes, since they MUST be capable of accepting larger sizes when a repeater is not in use.

#### Steps to Reproduce
Try to send a maximum size downlink in AU915

#### What do you see now?
The maximum downlink sizes are too small, by 20 bytes, plus a further 5 bytes due to an unrelated bug (*edit* see #3939).

The maximum downlink sizes should ideally match the sizes in the RP002-1.0.2 document.

#### Environment
AU915 subband 2

Increment the affected values by 20.

Attempt to send a maximum size downlink

I could - but you may decide to reject the idea and use the published incorrect values.

rvolosatovs:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/3938

#### Changes

- Fix payload size limits in AU915

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ For details about compatibility between different releases, see the **Commitment
 - Incorrect documentation url for event details data formats.
 - Search functionality for applications, gateways and organizations in the Console.
 - Error handling of end device template formats for the application overview page in the Console.
+- Payload size limits for AU915 data rates 8 and 9, which are now consistent with Regional Parameters RP002-1.0.2.
 
 ### Security
 
--- a/pkg/band/au_915_928.go
+++ b/pkg/band/au_915_928.go
@@ -92,8 +92,8 @@ func init() {
 			ttnpb.DATA_RATE_5: makeLoRaDataRate(7, 125000, makeConstMaxMACPayloadSizeFunc(230)),
 			ttnpb.DATA_RATE_6: makeLoRaDataRate(8, 500000, makeConstMaxMACPayloadSizeFunc(230)),
 
-			ttnpb.DATA_RATE_8:  makeLoRaDataRate(12, 500000, makeConstMaxMACPayloadSizeFunc(41)),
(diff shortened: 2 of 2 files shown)


rvolosatovs:
FYI, the values currently used are in accordance with 1.1-b spec for devices with repeater. We do not support repeaters anyway, so the suggested values from https://github.com/TheThingsNetwork/lorawan-stack/issues/3938 should be used, which are valid in 1.1-b too, for devices *without* a repeater

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Fixed
 
 - Incorrect documentation url for event details data formats.
+- Payload size limits for AU915 data rates 8 and 9, which are now consistent with latest version of Regional Parameters specification.
```
KrishnaIyer: Can you mention the exact version for future reference?

KrishnaIyer (approved):
Don't have the spec on hand but I trust that this is right.

PR #3976, opened 2021-03-24
Add payload formatter testing to the Console

### LINKED ISSUE #2232: Payload format testing
#### Summary
We should provide testing functionality for our payload format settings in the Console.

#### Why do we need this?
- to facilitate debugging and development for our users
- to close v2/v3 feature gap

Payload format settings views for application and end device level.

Functionality similar to the v2 payload format testing:
image
I think we can basically do a 1:1 adaption of this.

#### Environment
v3.7.0

1. Add backend API to test set payload formatters (at least that's how it currently works in v2)
2. Update JS SDK
3. Extend payload format container component to use testing capabilities

I can do the frontend part. Who can pick up the backend part? cc @htdvisser Please assign accordingly.

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2232

<details>

Application-level payload formatters:

Screenshot 2021-04-07 at 12 55 31
Screenshot 2021-04-07 at 12 55 26
Screenshot 2021-04-07 at 12 56 08
Screenshot 2021-04-07 at 12 56 15

End device payload formatters:

Screenshot 2021-04-07 at 13 01 41
Screenshot 2021-04-07 at 13 01 31
Screenshot 2021-04-07 at 13 01 23
Screenshot 2021-04-07 at 12 58 20
Screenshot 2021-04-07 at 12 57 57
Screenshot 2021-04-07 at 12 57 47
Screenshot 2021-04-07 at 12 57 34
Screenshot 2021-04-07 at 12 56 42
Screenshot 2021-04-07 at 12 56 37

</details>

#### Changes

- Extend `<PayloadFormattersForm />` to show the testing form
- Extend JS SDK api 
- Improve `<CodeEditor />` styles
  - Add `inline` and `error` style 
- Fix storybook issue
- Add `inline` version of `<CodeEditor />`

#### Testing

Manual testing

#### Notes for Reviewers

1. Absence of e2e testing
Using `cypress` to test functionality of `<CodeEditor />` is complicated. Among other issues, the component stores its value as a set of nested html elements which makes problematic to assert on editor value. Another option would be to make use of something like https://github.com/cypress-io/snapshot to save html snapshots, however we had experience with using snapshots before and it didnt prove to be useful. Ill think of some options to test this while this PR is in review.
2. UI mismatch with the v2 Console
Our current setup for forms does not allow to reconstruct UI from the v2 Console without nasty hacks, however if we ignore the html spec saying that `<form />` element cannot be nested in another `<form />` it is possible. The main issue here is to have proper validation for `payload` and `f_port` fields.
3. Only encode testing for downlinks
It matches what the v2 Console provides. @adriansmares Do you think it is necessary to provide option to test the `decodeDownlink` as well?

@kschiffer 
Please let me know if you have any ideas how we can improve UX for testing functionality. We might consider not closing with this PR https://github.com/TheThingsNetwork/lorawan-stack/issues/2232, but discuss UI/UX improvements there for later PRs.

#### Checklist


adriansmares:
>     3\. It matches what the v2 Console provides. @adriansmares Do you think it is necessary to provide option to test the `decodeDownlink` as well?

Yes. To give a bit more context on this:

Downlinks are not stored in their undecrypted+undecoded form by the AS, but instead the encrypted binary payload is saved in the NS. When someone uses `DownlinkQueueList`, the encrypted payloads are decrypted and then decoded using `decodeDownlink`. 

But testing it manually shouldn't really be a requirement, and instead we should test if the `decodeDownlink` is the inverse of `encodeDownlink`. How perhaps this should work is that when someone tests `encodeDownlink`, we have something like a checkmark that says `decodeDownlink is the inverse function of encodeDownlink: passed`. How this test works:

1. User inputs his functions.
2. He provides a `decoded_payload` in order to see the output of `encodeDownlink`. 
3. In the background, we take the binary result and run it via `decodeDownlink`, and check that the `decoded_payload` matches.

adriansmares (approved):
Functionality LGTM. I think it would be a good idea to add the 'inverse function' test for `decodeDownlink`, but for me it's not mandatory.

kschiffer:
Good work. Regarding the UX, I've created a wireframe for what I think it a better solution.

- [Uplink](https://user-images.githubusercontent.com/5710611/112874329-1af1aa80-90c3-11eb-8458-066c27719853.png)
- [Downlink](https://user-images.githubusercontent.com/5710611/112875787-eda5fc00-90c4-11eb-8ba7-1ab20c1b75ee.png)

As you can see, this is mostly a layout matter to display setup and testing horizontally next to each other and to rearrange the fields a little. This will make developing and fixing formatter functions a lot easier. Also the inputs for the payload and test result are now more semantic (safe inspector for byte result, code editor for json result).

Let me know if you think it's possible to include it like this in this PR, otherwise I'll look how we can split things up. Implementation-wise (wrt form mechanics) it shouldn't make much of a difference.

I'll wait with code review until we made a decision here. One thing I noticed though: why is this only implemented for device level PFs?

bafonins:
@kschiffer I updated UI to match the wireframes

> One thing I noticed though: why is this only implemented for device level PFs?

The API requires a device ID, see https://github.com/TheThingsNetwork/lorawan-stack/issues/2232#issuecomment-801197454. Note, that the users can still test application level formatters by selecting `Use application payload formatter`.

pkg/webui/components/code-editor/code-editor.styl
```diff
@@ -21,6 +21,18 @@
...
+    height: $default-input-height
+
+    &.focus
+      border-color: $c-active-blue
+
+    &.error
+      border-color: $c-error
+
+    :global .ace_marker-layer .ace_active-line
+      background: transparent
```
kschiffer: I think all this inline styling and logic can be removed since we're not using it.

pkg/webui/components/form/field/container.js
```diff
@@ -0,0 +1,39 @@
...
+
+import style from './field.styl'
+
+const FieldContainer = ({ horizontal, children }) => {
+  return (
+    <div className={classnames(style.fieldContainer, { [style.horizontal]: horizontal })}>
+      {children}
+    </div>
+  )
+}
```
kschiffer:
```suggestion
const FieldContainer = ({ horizontal, children }) => (
  <div className={classnames(style.fieldContainer, { [style.horizontal]: horizontal })}>
    {children}
  </div>
)
```

Nit

pkg/webui/console/components/payload-formatters-form/index.js
```diff
@@ -40,18 +44,28 @@ const m = defineMessages({
...
+  { label: m.repository, value: TYPES.REPOSITORY },
+]
+const formatterOptions = formatterOptionsWithReset.slice(1, formatterOptionsWithReset.length)
+
 const validationSchema = Yup.object().shape({
-  [FIELD_NAMES.RADIO]: Yup.string()
+  [FIELD_NAMES.SELECT]: Yup.string()
     .oneOf(Object.values(TYPES))
     .required(sharedMessages.validateRequired),
   [FIELD_NAMES.JAVASCRIPT]: Yup.string().when('types-radio', {
```
kschiffer:
```suggestion
  [FIELD_NAMES.JAVASCRIPT]: Yup.string().when('types-select', {
```

pkg/webui/console/components/payload-formatters-form/test-form/index.js
```diff
@@ -0,0 +1,198 @@
...
+            <Form.Field
+              title={m.bytePayload}
+              name="payload"
+              type="byte"
+              component={Input}
+              className={style.payload}
+              unbounded
+            />
+          ) : (
+            <Form.Field
```
kschiffer, line 122: This field currently validates on change but it should validate on blur only.

pkg/webui/console/containers/device-payload-formatters/uplink.js
```diff
@@ -104,13 +123,46 @@ class DevicePayloadFormatters extends React.PureComponent {
...
+    const { appId, devId, decodeUplink, device } = this.props
+    const { f_port, payload, formatter, parameter } = data
+    const { version_ids } = device
+
+    const { uplink } = await decodeUplink(appId, devId, {
+      uplink: {
+        f_port,
+        frm_payload: hexToBase64(payload),
+        // `rx_metadata` and `settings` fields are required by the validation middleware in AS.
+        // These fields wont affect the result of decoding an uplink message.
```
kschiffer:
```suggestion
        // These fields won't affect the result of decoding an uplink message.
```

sdk/js/src/service/application-server.js
```diff
@@ -0,0 +1,65 @@
+// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
```
kschiffer:
```suggestion
// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
```

kschiffer (changes requested):
Please also add the following
- A notification in the end device payload formatters including a link leading to the application level formatters, when the option `Use application default` is selected
  - `Click here to modify the default formatter for this application`
- A notification in the application level formatters reading `To test-run this payload formatter, use the "Payload Formatter" tab of any end device registered inside this application`
  - This should not be visible if "None" is selected

pkg/webui/console/components/payload-formatters-form/index.js
```diff
@@ -161,67 +225,112 @@ class PayloadFormattersForm extends React.Component {
...
-        >
-          <Form.Field
-            name={FIELD_NAMES.RADIO}
-            title={m.formatterType}
-            component={Radio.Group}
-            onChange={this.onTypeChange}
-            warning={type === TYPES.DEFAULT ? m.appFormatterWarning : undefined}
-            horizontal
+      <Row>
+        <Col sm={12} lg={this._showTestSection() ? 6 : 12}>
```
kschiffer, line 261:
```suggestion
        <Col sm={12} lg={this._showTestSection() ? 6 : 12}>
```

Are we using this underscore consistently?

pkg/webui/console/containers/device-payload-formatters/uplink.js
```diff
@@ -30,25 +35,39 @@ import sharedMessages from '@ttn-lw/lib/shared-messages'
...
 import {
   selectSelectedDeviceId,
   selectSelectedDeviceFormatters,
+  selectSelectedDevice,
 } from '@console/store/selectors/devices'
 
+const m = defineMessages({
+  appFormatterText:
+    'Click <Link>here</Link> to modify the default uplink formatter for this application',
+})
```
kschiffer: Let's use one message for both PF types. `Click <Link>here</Link> to modify the default payload formatter for this application`

kschiffer (approved):

pkg/webui/console/components/payload-formatters-form/index.js
bafonins: What do you mean?

kschiffer: The method uses an underscore prefix. I was just wondering when this should be used.

bafonins: We already use that approach, e.g.
https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.12/pkg/webui/components/key-value-map/entry.js#L31-L41
https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.12/pkg/webui/console/components/join-eui-prefixes-input/index.js#L71-L81

The reason I added prefix to `showTestSection` is to visually separate regular methods that, say, can be passed down to children and strictly utilitarian method that is used only within the component, should not be passed anywhere else and needs access to component props.

kschiffer: Consider adding this to the coding guidelines so we can ensure that this is done consistently.

PR #4011, opened 2021-03-31
Add LNS API Key to gateway general settings page

### LINKED ISSUE #3929: Add option to fill LNS Secret of a Gateway via the Console
#### Summary

Add option to fill LNS Secret of a Gateway via the Console.

#### Why do we need this?

Currently, we fill LNS secrets via the CLI which requires quite a number of steps. To make this feature more usable, it's better to add it to the console.

Here's the current process for setting the LNS Secret of a gateway is as follows;
1. Create an LNS API Key with `Link as Gateway Rights`
2. Copy that API Key 
3. Login to the CLI (if steps 1 and 2 are done using the console)
4. Format the API Key as a Hex value
```bash
$ SECRET=$(echo -n <LNS API KEY> | xxd -ps -u -c 8192)
```
5. Set the `lbs-lns-secret.value` field of the gateway object
```bash
$ ttn-lw-cli gateways update <gateway-id> --lbs-lns-secret.value $SECRET
```
Documented here; https://www.thethingsindustries.com/docs/gateways/lora-basics-station/cups/#configure-cups-to-send-the-lns-api-key

Provide an option in the console to do this easily.

#### Environment

```
tts v3.11.2
```

We discussed 2 options
1. Fully automated
 - Have a button (`Create and Set LNS API Key`)
 - When the user presses this button, the above (CLI) steps are performed by the console directly, including creating the API Key and is set in the gateway without the user having to do anything.

2. Partially automated
 - Have a text input (`Set LNS API Key`)
 - The user creates the key with (hopefully) the correct rights and copies that key and enters it here. The Console then only does the conversion to hex and sets the value.

I'm not sure which option is better but that's because I'm a UX dummy.

I think the appropriate place would be in the **General Settings** page of a **Gateway** under the **Gateway Server Address** option;
Screenshot 2021-03-16 at 12 11 52

Things to consider
1. This field is only necessary for LoRa Basics Station gateways. It will be ignored by the backend if set for others. Since there is no way for TTS to know if a gateway is LBS or not, it's ok to support this for all gateways.
2. We may want to set this field at the time of gateway creation (if the user chooses to do so). This is I think for later to consider as part of the gateway wizard. 

🐒 

And I can actually write cypress E2E for this.

Nyet. Can help with review and tests.

KrishnaIyer:
#### Summary

Closes #3929 

#### Changes

- Add Add LNS API Key to gateway general settings page
- Convert to/from Base64 where necessary.

#### Testing

Locally tested and added some lines to Cypress
Screenshot 2021-03-31 at 12 01 49

Full tests; https://drive.google.com/file/d/1YjInUYomNoO_QchAL-uFLt0aJONenxdE/view?usp=sharing
(Note: This test was done before rebasing `v3.12` so the `Require Authentication field is missing`).

##### Regressions

None that I can think of

#### Notes for Reviewers

- I know nothing about JS so I don't think this isn't optimal code.
- I didn't find an easy way to do the casting using `Yup` since I don't have experience with that package.

#### Checklist


pkg/webui/console/views/gateway-general-settings/basic-settings-form/index.js
```diff
@@ -60,12 +60,19 @@ const BasicSettingsForm = React.memo(props => {
...
+    }
     return validationSchema.cast(initialValues)
   }, [gateway])
 
   const onFormSubmit = React.useCallback(
     async (values, { resetForm, setSubmitting }) => {
       const castedValues = validationSchema.cast(values)
+      if (castedValues.lbs_lns_secret !== undefined && castedValues.lbs_lns_secret.value !== undefined){
+        castedValues.lbs_lns_secret.value = btoa(castedValues.lbs_lns_secret.value)
+        console.log(castedValues.lbs_lns_secret.value)
```
bafonins: Remove logging

pkg/webui/console/views/gateway-general-settings/basic-settings-form/index.js
```diff
@@ -131,6 +138,12 @@ const BasicSettingsForm = React.memo(props => {
         component={Checkbox}
         label={sharedMessages.enabled}
         description={sharedMessages.requireAuthenticatedConnectionDescription}
+        />
+        <Form.Field
+        title={sharedMessages.lbsLNSSecret}
```
bafonins, line 142: Please format `tools/bin/mage js:fmt`

pkg/webui/console/views/gateway-general-settings/basic-settings-form/index.js
```diff
@@ -60,12 +60,19 @@ const BasicSettingsForm = React.memo(props => {
       ...gateway,
       attributes: mapAttributesToFormValue(gateway.attributes),
     }
+    if (initialValues.lbs_lns_secret !== undefined && initialValues.lbs_lns_secret.value !== undefined){
```
bafonins:
```suggestion
    if (Boolean(initialValues.lbs_lns_secret) && initialValues.lbs_lns_secret.value !== undefined){
```

cypress/integration/console/gateways/edit.spec.js
```diff
@@ -98,6 +98,10 @@ describe('Gateway general settings', () => {
     cy.findByLabelText('Automatic updates')
       .should('exist')
       .and('have.attr', 'value', 'false')
+    cy.findByLabelText('LoRa Basics Station LNS Authentication Key')
+    .should('contain', 'The Authentication Key for Lora Basics Station LNS connections. This field is ignored for other gateways.')
```
bafonins:
```js
    cy.findDescriptionByLabelText('LoRa Basics Station LNS Authentication Key')
      .should(
        'contain',
        'The Authentication Key for Lora Basics Station LNS connections. This field is ignored for other gateways.',
      )
      .and('be.visible')
```

pkg/webui/console/views/gateway-general-settings/basic-settings-form/index.js
```diff
@@ -131,6 +138,12 @@ const BasicSettingsForm = React.memo(props => {
         component={Checkbox}
         label={sharedMessages.enabled}
         description={sharedMessages.requireAuthenticatedConnectionDescription}
+        />
+        <Form.Field
+        title={sharedMessages.lbsLNSSecret}
+        description={sharedMessages.lbsLNSSecretDescription}
+        name="lbs_lns_secret.value"
```
bafonins, line 144: Can we validate values for this field on the client?

bafonins (changes requested):

pkg/webui/console/views/gateway-general-settings/basic-settings-form/index.js
KrishnaIyer: 🤦 I even regexed `console.log` to make sure I had removed these.

pkg/webui/console/views/gateway-general-settings/basic-settings-form/index.js
KrishnaIyer: Yeah good one. So the API allows `2048` bytes. But since we convert to `Base64` we need an additional 33% (at max) capacity. So 66% of 2048 = 1351,68 so I've set the max length at `1350`.

kschiffer (approved):

KrishnaIyer:
Waiting for @bafonins to approve before merging.

pkg/webui/console/views/gateway-general-settings/basic-settings-form/index.js
bafonins: 👍 Please add this as description for L42 in `basic-settings-form/validation-schema.js `

bafonins (approved):

PR #4026, opened 2021-04-06
Fix mac and phy version handling in the Console

bafonins:
#### Summary

This PR fixes mac and phy version handling in device general settings page.

#### Changes

- Change mac/phy version values
- Adjust tests for `parseLorawanMacVersion`

#### Testing

unit, e2e and manual testing

#### Notes for Reviewers

Currently, NS returns mac/phy version values as `MAC_...` and `PHY_...` which causes `Lorawan version` and `Regional parameters` to be empty.

#### Checklist


kschiffer (approved):
How did this sneak through CI?

bafonins:
@kschiffer NS still accepts the old format of mac/phy versions, e.g. `1.0.0`, but returns the new format. We currently do not have exhaustive device gs e2e tests.

@rvolosatovs Can you approve that this is expected behavior?

rvolosatovs:
This is what the spec demands, marshaling proto enums into JSON strings. @htdvisser @johanstokking is this going correct? If we are deviating from spec anyhow, where is that documented?

Actual JSON encoding format can be seen here https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.12/pkg/ttnpb/testdata/ttnpb_encoding_golden.md#L327-L850

Note the difference between JSONPB and JSON

htdvisser:
We have indeed been deviating from the spec in several places by using `(gogoproto.customtype)` and explicitly implemented (un)marshalers. Some binary fields are rendered to JSON as hex instead of base64 (DevAddrs, EUIs, etc.) and some even have completely custom JSON rendering (DevAddr prefixes, EUI prefixes, some enums).

I don't think any of these deviations are documented anywhere.

Until v3.12 I don't think there was a difference between JSON and JSONPB, as far as I know, grpc-gateway always just used JSONPB.

asmulko (approved):

PR #4049, opened 2021-04-12
Fix panic in Downlink event handling

KrishnaIyer:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2673

#### Changes

- Pass a local copy of the down message to the events handler.

#### Testing

We can test this on staging. But this seems simple enough.

##### Regressions

NA

#### Checklist

### DIFF
--- a/pkg/gatewayserver/grpc_nsgs.go
+++ b/pkg/gatewayserver/grpc_nsgs.go
@@ -82,7 +82,8 @@ func (gs *GatewayServer) ScheduleDownlink(ctx context.Context, down *ttnpb.Downl
 		}
 		ctx = events.ContextWithCorrelationID(ctx, events.CorrelationIDsFromContext(conn.Context())...)
 		down.CorrelationIDs = append(down.CorrelationIDs, events.CorrelationIDsFromContext(ctx)...)
-		registerSendDownlink(ctx, conn.Gateway(), down, conn.Frontend().Protocol())
+		msg := *down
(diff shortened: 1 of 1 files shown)


adriansmares (approved):

pkg/gatewayserver/grpc_nsgs.go
```diff
@@ -82,7 +82,8 @@ func (gs *GatewayServer) ScheduleDownlink(ctx context.Context, down *ttnpb.Downl
 		}
 		ctx = events.ContextWithCorrelationID(ctx, events.CorrelationIDsFromContext(conn.Context())...)
 		down.CorrelationIDs = append(down.CorrelationIDs, events.CorrelationIDsFromContext(ctx)...)
-		registerSendDownlink(ctx, conn.Gateway(), down, conn.Frontend().Protocol())
+		msg := *down
+		registerSendDownlink(ctx, conn.Gateway(), &msg, conn.Frontend().Protocol())
```
johanstokking: What this does is only copy `ttnpb.DownlinkMessage`, including pointers to the same memory space, so this doesn't resolve the issue.

---

10 lines above it says that `down` will be owned by the connection. I think that the actual issue is that `down` isn't actually owned by the connection: it is being used after `ScheduleDown()`, like here.

So I'd suggest renaming `down` on L75 to `connDown` and use that on L76-L77, and don't change anything else.

johanstokking (changes requested):

johanstokking (approved):

PR #4052, opened 2021-04-13
Add API key expiry

### LINKED ISSUE #2228: Add expiry to API keys
#### Summary

We should let users set an expiry date on API keys. Optional network configuration could set a maximum lifetime on API keys.

#### Why do we need this?

Security.

- Add an expiry date, similar to what we have for access tokens.

pgalic96:
#### Summary

Closes #2228 

#### Changes

- Add field `ExpiresAt` to the `APIKey` model.
- Extend CLI commands to accept `--expiry` flag
- Add a check for expired API Key in `AuthInfo`

#### Testing

- Unit tests
- Manual

##### Regressions

- Anything related to API keys

#### Notes for Reviewers

I am trying to figure out why a test in `OrganizationAccess` is failing. It is exactly the same as the passing tests in other `XAccess` registries so it's confusing. Shouldn't be anything crucial so I think this is ready for review.

#### Checklist


pkg/identityserver/entity_access.go
```diff
@@ -354,6 +358,7 @@ func (is *IdentityServer) RequireAuthenticated(ctx context.Context) error {
 		}
 	}
 	if apiKey := authInfo.GetAPIKey(); apiKey != nil {
+
```
kschiffer:
```suggestion
```

pkg/identityserver/identityserver_test.go
```diff
@@ -78,12 +78,18 @@ func init() {
...
+			expiredTime := time.Now().Add(-1 * time.Hour)
 			population.APIKeys[id] = append(
 				apiKeys,
 				&ttnpb.APIKey{
 					Name:   "key without rights",
 					Rights: []ttnpb.Right{ttnpb.RIGHT_SEND_INVITES},
 				},
+				&ttnpb.APIKey{
+					Name:      "expired key",
+					Rights:    []ttnpb.Right{ttnpb.RIGHT_SEND_INVITES},
```
kschiffer:
```suggestion
					Rights:    []ttnpb.Right{ttnpb.RIGHT_USER_ALL},
```

Nit: to test the expiration in isolation this should at least have sufficient rights to perform the action otherwise.

kschiffer (approved):
Codeowned LGTM.

Should we add console support in this PR too? Shouldn't be a big deal.

api/application.proto
```diff
@@ -111,6 +111,7 @@ message CreateApplicationAPIKeyRequest {
   ApplicationIdentifiers application_ids = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (validate.rules).message.required = true];
   string name = 2 [(validate.rules).string.max_len = 50];
   repeated Right rights = 3 [(validate.rules).repeated.items.enum.defined_only = true];;
+  string expiry = 4 [(validate.rules).string = {pattern: "^\\d{4}\\-(0?[1-9]|1[012])\\-(0?[1-9]|[12][0-9]|3[01])$"}];
```
htdvisser: It's the job of the frontend to convert a YYYY-MM-DD (or local format) to a timestamp, so:

```suggestion
  google.protobuf.Timestamp expires_at = 4 [(gogoproto.stdtime) = true];
```

Also in the other places.

cmd/ttn-lw-cli/commands/applications_claim.go
```diff
@@ -88,6 +98,7 @@ key is provided, a new API key will be created.`,
 
 func init() {
 	applicationClaimAuthorize.Flags().String("api-key", "", "")
+	applicationClaimAuthorize.Flags().String("expiry", "", "API key expiry date (YYYY-MM-DD)")
```
htdvisser: This should probably be `api-key-expiry`, and the comment should indicate that it's only applicable when generating a new key.

pkg/identityserver/api_key_utils.go
```diff
@@ -29,8 +31,22 @@ var apiKeyHashSettings auth.HashValidator = pbkdf2.PBKDF2{
...
+func parseAndValidateAPIKeyExpiry(expiry string) (*time.Time, error) {
+	expiryDate, err := time.Parse("2006-01-02", expiry)
+	if err != nil {
+		return nil, errInvalidDateFormat
+	}
+	if expiryDate.Before(time.Now()) {
+		return nil, errExpiryDateInPast
+	}
+	return &expiryDate, nil
+}
```
htdvisser: This needs to be moved to the frontends (in this case, the Go code should go to the CLI).

htdvisser (changes requested):

cmd/ttn-lw-cli/commands/applications_access.go
```diff
@@ -263,11 +251,32 @@ var (
...
+
+			if expiry != "" {
+				expiryDate, err := time.Parse("2006-01-02", expiry)
+				if err != nil {
+					return errInvalidDateFormat
+				}
+				if expiryDate.Before(time.Now()) {
+					return errExpiryDateInPast
+				}
+			}
```
htdvisser: - You do this quite often, maybe extract it into a func?
- Should we take midnight (the beginning of the day) or a minute to midnight (the end of the day)?
- What timezone is used?

cmd/ttn-lw-cli/commands/applications_claim.go
```diff
@@ -38,15 +40,36 @@ key is provided, a new API key will be created.`,
...
-					ttnpb.RIGHT_APPLICATION_DEVICES_WRITE,
-					ttnpb.RIGHT_APPLICATION_DEVICES_WRITE_KEYS,
-				)
+				apiKey, err := ttnpb.NewApplicationAccessClient(is).CreateAPIKey(ctx, &ttnpb.CreateApplicationAPIKeyRequest{
+					ApplicationIdentifiers: *appID,
+					Name:                   "Device Claiming",
+					Rights: []ttnpb.Right{ttnpb.RIGHT_APPLICATION_DEVICES_READ,
+						ttnpb.RIGHT_APPLICATION_DEVICES_READ_KEYS,
+						ttnpb.RIGHT_APPLICATION_DEVICES_WRITE,
+						ttnpb.RIGHT_APPLICATION_DEVICES_WRITE_KEYS},
```
htdvisser:
```suggestion
						ttnpb.RIGHT_APPLICATION_DEVICES_WRITE_KEYS,
					},
```

cmd/ttn-lw-cli/commands/applications_claim.go
```diff
@@ -38,15 +40,36 @@ key is provided, a new API key will be created.`,
...
-				apiKey, err := createApplicationAPIKey(ctx, *appID, "Device Claiming",
-					ttnpb.RIGHT_APPLICATION_DEVICES_READ,
-					ttnpb.RIGHT_APPLICATION_DEVICES_READ_KEYS,
-					ttnpb.RIGHT_APPLICATION_DEVICES_WRITE,
-					ttnpb.RIGHT_APPLICATION_DEVICES_WRITE_KEYS,
-				)
+				apiKey, err := ttnpb.NewApplicationAccessClient(is).CreateAPIKey(ctx, &ttnpb.CreateApplicationAPIKeyRequest{
+					ApplicationIdentifiers: *appID,
+					Name:                   "Device Claiming",
+					Rights: []ttnpb.Right{ttnpb.RIGHT_APPLICATION_DEVICES_READ,
```
htdvisser:
```suggestion
					Rights: []ttnpb.Right{
						ttnpb.RIGHT_APPLICATION_DEVICES_READ,
```

pkg/identityserver/api_key_utils.go
```diff
@@ -43,11 +47,25 @@ func GenerateAPIKey(ctx context.Context, name string, rights ...ttnpb.Right) (ke
...
-	key = &ttnpb.APIKey{
-		ID:     generatedID,
-		Key:    hashedKey,
-		Name:   name,
-		Rights: rights,
+	// Check that expiry date passed is not a zero time or nil value
+	if expiresAt != nil && !expiresAt.IsZero() {
+		if expiresAt.Before(time.Now()) { // Check if an expired date is passed.
+			return nil, "", errExpiryDateInPast
+		}
```
htdvisser: I don't think this check belongs here. Maybe better to add a `(validate.rules).timestamp.gt_now = true` to the `.proto` files so that it's validated that way?

pkg/identityserver/store/api_key_store.go
```diff
@@ -45,6 +45,7 @@ func (s *apiKeyStore) CreateAPIKey(ctx context.Context, entityID ttnpb.Identifie
 		Name:       key.Name,
 		EntityID:   entity.PrimaryKey(),
 		EntityType: entityTypeForID(entityID),
+		ExpiresAt:  key.ExpiresAt,
```
htdvisser, line 76: How about UpdateAPIKey? Shouldn't users be able to change this after creation?

cmd/ttn-lw-cli/commands/applications_access.go
pgalic96: >What timezone is used?

the timezone is UTC, since it's the server time. should we make it timezone aware? I can think of some edge cases in the current implementation where the api key could expire before the user expects it to

>Should we take midnight (the beginning of the day) or a minute to midnight (the end of the day)?

We could also take the minute exactly from the api key creation?

pkg/identityserver/store/api_key_store.go
pgalic96: From this comment https://github.com/TheThingsNetwork/lorawan-stack/pull/4052#discussion_r614718052 I got the impression we want it only when creating the key? I don't think we need to allow updating the expiration date when users can just set a new key, but let me know if you meant something else.

htdvisser: That CLI command generates an API key for claiming. My comment there was about the fact that passing the expiry flag only makes sense when letting the command generate an API key, not when setting an existing API key.

I think in general users will want to be able to edit the expiry date of an API key.

cmd/ttn-lw-cli/commands/applications_access.go
htdvisser: The server indeed uses UTC, but clients (like also the Console) may use local time.

After giving this some more thought I think we should just let users pass the entire date+time instead of only the date; that way they can decide.

pgalic96:
@htdvisser This PR scope expanded now that I added fieldmasks to the UpdateAPIKey RPC's. This obviously breaks the console edit api key functionality cc @kschiffer

htdvisser (changes requested):
Adding field masks like this breaks our compatibility commitment, because older versions of the CLI can no longer update or delete API keys. We need to add some logic in the IS that, if the field mask of the update request is empty, the update RPC behaves exactly the same as before, so perhaps by setting the field mask to `name,rights` if it's empty.

pgalic96:
@htdvisser 

https://github.com/TheThingsNetwork/lorawan-stack/pull/4052/files#diff-ff7c6e5ca66110e66fe17681b27504602b065200d0d9cd146438cce0f4fbf5a3R157-R160

this also means the console support for updating api keys is not broken anymore.

Not sure why the end-to-end tests fail, locally they work fine.

cmd/ttn-lw-cli/commands/applications_access.go
```diff
@@ -375,10 +383,12 @@ func init() {
 	applicationAPIKeys.AddCommand(applicationAPIKeysGet)
 	applicationAPIKeysCreate.Flags().String("name", "", "")
 	applicationAPIKeysCreate.Flags().AddFlagSet(applicationRightsFlags)
+	applicationAPIKeysCreate.Flags().String("api-key-expiry", "", "API key expiry date (YYYY-MM-DD:HH:mm)")
```
htdvisser: Other timestamps in the CLI use `YYYY-MM-DDTHH:MM:SSZ`.

pkg/identityserver/api_key_utils.go
```diff
@@ -43,11 +44,22 @@ func GenerateAPIKey(ctx context.Context, name string, rights ...ttnpb.Right) (ke
...
+			Rights:    rights,
+			ExpiresAt: expiresAt,
+		}
+	} else {
+		key = &ttnpb.APIKey{
+			ID:     generatedID,
+			Key:    hashedKey,
+			Name:   name,
+			Rights: rights,
+		}
```
htdvisser: Unless you also meant to check for `expiresAt.IsZero()`, this isn't needed.

pkg/identityserver/store/api_key_store.go
```diff
@@ -125,12 +154,25 @@ func (s *apiKeyStore) UpdateAPIKey(ctx context.Context, entityID *ttnpb.EntityId
 		}
 		return nil, err
 	}
-	if len(key.Rights) == 0 {
+	// Support for older cli versions which do not set field masks for UpdateAPIKey.
```
htdvisser: This has nothing to do with the CLI, it's about API compatibility with previous versions of The Things Stack.

pkg/identityserver/api_key_utils.go
pgalic96: ah yea this is a leftover from when there was a zero value

htdvisser (approved):

PR #4058, opened 2021-04-15
Add uint32 notation for device address

### LINKED ISSUE #4014: Add option for C-Style uint32_t `0x00000000` representation for device address field
#### Summary
In the Console for a device, for the 'Device address' field add the uint32_t format 0x000000 as option.

In the Console for a device the 'Device address' field can be copied.
In C/C++ code the device address is usually defined as uint32_t type which requires the format 0x000000.
But this format is currently not available as one of the options.

#### Why do we need this?
Users coding in C/C++ cannot use a copy/pasted device address value because it is in the wrong format so they have to manually change the the format of the pasted device address value after each copy/paste action.
This is very impractical. Copy/paste should just work without having to make any manual changes after the paste.

By default the device address is shown as a sequence of bytes separated by a space:
capture 2021-04-02 14·00·42

It is also possible to display and copy it in C/C++ msb and lsb formats:
capture 2021-04-02 14·00·47

capture 2021-04-02 14·00·50

Add a format option for uint32_t type with a value looking like: 0x26011FD6

#### Environment
_Not relevant._

Add the format option as described.

\-

No.

asmulko:
#### Summary

Closes #4014 

#### Changes

- Added C-Style uint32_t representation for device address field

Before:
No option

After:
after_1

#### Testing

- Manually tested

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ For details about compatibility between different releases, see the **Commitment
 
 - Payload formatter testing functionality in the Console.
 - Options in the Identity Server to reject passwords that contain the user ID (`is.user-registration.password-requirements.reject-user-id`) or common passwords (`is.user-registration.password-requirements.reject-common`).
+- C-Style uint32_t representation for end device address field.
 
 ### Changed
 
--- a/pkg/webui/components/data-sheet/story.js
+++ b/pkg/webui/components/data-sheet/story.js
@@ -46,6 +46,13 @@ const testData = [
     header: 'Activation Info',
     items: [
       { key: 'Device EUI', value: '1212121212', type: 'byte', sensitive: false },
+      {
+        key: 'Device EUI with Uint32_t',
+        value: '1212121212',
+        type: 'byte',
+        sensitive: false,
+        enableUint32: true,
+      },
       { key: 'Join EUI', value: '1212121212', type: 'byte', sensitive: false },
       {
         key: 'Value with Nesting',
--- a/pkg/webui/components/safe-inspector/index.js
+++ b/pkg/webui/components/safe-inspector/index.js
@@ -54,6 +54,15 @@ const m = defineMessages({
   byteOrder: 'Switch byte order',
 })
 
+const MSB = 'msb'
+const LSB = 'lsb'
+const UINT32_T = 'uint32_t'
+const representationRotateMap = {
+  [MSB]: LSB,
+  [LSB]: UINT32_T,
+  [UINT32_T]: MSB,
+}
+
 @injectIntl
 export class SafeInspector extends Component {
   static propTypes = {
@@ -63,6 +72,8 @@ export class SafeInspector extends Component {
     data: PropTypes.string.isRequired,
     /** Whether the component should resize when its data is truncated. */
     disableResize: PropTypes.bool,
+    /** Whether uint32_t notation should be enabled for byte representation. */
+    enableUint32: PropTypes.bool,
     /** Whether the data can be hidden (like passwords). */
     hideable: PropTypes.bool,
     /** Whether the data is initially visible. */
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ For details about compatibility between different releases, see the **Commitment
 
 - Payload formatter testing functionality in the Console.
 - Options in the Identity Server to reject passwords that contain the user ID (`is.user-registration.password-requirements.reject-user-id`) or common passwords (`is.user-registration.password-requirements.reject-common`).
+- C-Style uint32_t representation for device address field
```
bafonins:
```suggestion
- C-Style uint32_t representation for end device address field.
```

pkg/webui/components/safe-inspector/index.js
```diff
@@ -63,6 +63,8 @@ export class SafeInspector extends Component {
     data: PropTypes.string.isRequired,
     /** Whether the component should resize when its data is truncated. */
     disableResize: PropTypes.bool,
+    /** Whether uint32_t notation should be enabled for byte representation. */
+    enableUint32: PropTypes.bool,
```
bafonins, line 76: Please add a new entry to storybook

pkg/webui/components/safe-inspector/index.js
```diff
@@ -134,7 +137,16 @@ export class SafeInspector extends Component {
 
   @bind
   handleSwapToggle() {
-    this.setState(prev => ({ msb: !prev.msb }))
+    const { representation } = this.state
+    const { enableUint32 } = this.props
+
+    if (representation === 'msb') {
+      this.setState({ representation: 'lsb' })
```
bafonins: Consider something like this instead of nested ifs

```js
// Outside of the component.
const MSB = 'msb'
const LSB = 'lsb'
const UINT32_T = 'uint32_t'
const representationRotateMap = {
  [MSB]: LSB,
  [LSB]: UINT32_T,
  [UINT32_T]: MSB,
}

// Inside the component
_getNextRepresentation(current) {
  const { enableUint32 } = this.props
	  
  const next = representationRotateMap[current]
  return next === UINT32_T && !enableUint32 ? representationRotateMap[next] : next
}

@bind
handleSwapToggle() {
  this.setState(({ representation }) => ({
    representation: this._getNextRepresentation(representation),
  }))
}
```

pkg/webui/components/safe-inspector/index.js
```diff
@@ -232,8 +244,12 @@ export class SafeInspector extends Component {
     if (isBytes) {
       const chunks = chunkArray(data.toUpperCase().split(''), 2)
       if (!byteStyle) {
-        const orderedChunks = msb ? chunks : chunks.reverse()
-        formattedData = display = orderedChunks.map(chunk => `0x${chunk.join('')}`).join(', ')
+        if (representation === 'uint32_t') {
```
bafonins: Please introduce constants for representation types
```suggestion
        if (representation === UINT32_T) {
```

bafonins (changes requested):

pkg/webui/components/data-sheet/story.js
```diff
@@ -46,6 +46,13 @@ const testData = [
     header: 'Activation Info',
     items: [
       { key: 'Device EUI', value: '1212121212', type: 'byte', sensitive: false },
+      {
```
bafonins, line 49: Currently, storybook is broken. Please add:
```js
  window.__ttn_config__ = {
    APP_CONFIG: {
      documentation_base_url: 'https://thethingsstack.io',
    },
  }
```

to `config/storybook/preview-head.html`

bafonins (approved):
Looks good. Please my comment before merging

pkg/webui/components/safe-inspector/index.js
```diff
@@ -54,6 +54,15 @@ const m = defineMessages({
...
 })
 
+const MSB = 'msb'
+const LSB = 'lsb'
+const UINT32_T = 'uint32_t'
+const representationRotateMap = {
+  [MSB]: LSB,
+  [LSB]: UINT32_T,
+  [UINT32_T]: MSB,
+}
```
kschiffer, line 64: This implementation is rather weird to me. I'd prefer having a regular array that is being cylced

```javascript
const representations = [MSB, LSB, UINT32_T]
```

The array should be composed based on the provided props (we could even use the array itself as a prop) and then keep track of the index via state. This will also be more flexible should we add more/other representations in the future.

kschiffer (changes requested):

kschiffer (approved):

Issue #4019, opened 2021-04-06
Multi frame request support for LoRaCloud GLS


adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/4018

We should support sending multi frame requests via GLS, using the recently sent uplinks.

#### Why do we need this?

In order to have better location queries, since multi frame should provide more accurate results.

#### What is already there? What do you see now?

The `QUERY_TOARSSI` query type, which does a single frame request using the `TOARSSI` algorithm.

#### What is missing? What do you want to see?

1. A new configuration option that should specify if the query should be single frame or multi frame (not a new query type).
2. A new configuration option for how many of the recent frames we should we use for the multi frame queries (if they are enabled). We should also provide a way to automatically determine this information, based on the first byte of the payload.
3. A new configuration option for how recent (a duration) should the recent uplinks be, when used by the multi frame query.
4. On uplink handling, we should check if we should do a single frame request, as we currently do, or use multiple frames using the new configuration.

#### Environment

`v3.12`

#### How do you propose to implement this?

Add the requested configuration to the package (new fields), and then use them as part of the uplink handling.

#### How do you propose to test this?

Send multiple uplinks and check if the query is done. You can use the simulation functionality.

#### Can you do this yourself and submit a Pull Request?

Yes. cc @johanstokking

adriansmares:
Blocked by https://github.com/TheThingsNetwork/lorawan-stack/issues/4018

htdvisser:
Since @adriansmares is off today, and it isn't in progress yet, this isn't going to make today's deadline, so moved to v3.12.2.

PR #4061, opened 2021-04-15
Store recent data application uplinks in the Application Server 

### LINKED ISSUE #4018: Store recent data application uplinks in the Application Server
#### Summary

The Application Server should store, in a FIFO fashion, the application data uplinks sent by the end device.

#### Why do we need this?

Certain integrations may want to query the last received data uplinks. Since we would like to avoid enabling the storage integration just for this, we could bring back the `recent_uplinks` field of `ttnpb.EndDevice`:

https://github.com/TheThingsNetwork/lorawan-stack/blob/e37d779528c17aabf33d9149c5603e31e9206217/api/end_device.proto#L534

The storage integration keeps these uplinks in long term storage.

The Application Server should store, either as part of the end device, or of the session, the last 16 application data uplinks.

#### Environment

`v3.12`

Add the API field, save the uplinks as part of `as.handleUplink`, and allow frontends to query the recent uplinks via `io.Server`.

Send some real uplinks from and end device and check that they are stored.

Yes. cc @johanstokking

adriansmares:
#### Summary

Closes #4018

#### Changes

- Break down the `io.Server` interface into multiple sub-interfaces, as it was getting too big.
- Add a data application uplink registry. 

#### Testing

Unit testing.

##### Regressions

This PR adds a new store, for which unit tests are added. The usage of the functionality will be tested better as part of https://github.com/TheThingsNetwork/lorawan-stack/issues/4019.

#### Notes for Reviewers

This implementation diverges from the proposal due to the following reasons:
- We always fetch the full `ttnpb.EndDevice` from Redis, which means that we would always be pulling the stored uplinks while accessing an end device. This would increase the network traffic for minimal gains. _We could indeed avoid this by storing different fields in different keys, but that's added complexity for little gain._
- The device registry is available over gRPC, and we'd rather avoid exposing this storage to users, who will most likely misuse it. _We could simply disallow the field in the allowed field mask, indeed._
- We have better control over the keys, and can actually check how much storage is used by this functionality, in memory-critical situations.

#### Checklist

### DIFF
--- a/pkg/applicationserver/config.go
+++ b/pkg/applicationserver/config.go
@@ -71,6 +71,7 @@ type Config struct {
 	LinkMode         string                    `name:"link-mode" description:"Deprecated - mode to link applications to their Network Server (all, explicit)"`
 	Devices          DeviceRegistry            `name:"-"`
 	Links            LinkRegistry              `name:"-"`
+	UplinkStorage    UplinkStorageConfig       `name:"uplink-storage" description:"Application uplinks storage configuration"`
 	Distribution     DistributionConfig        `name:"distribution" description:"Distribution configuration"`
 	EndDeviceFetcher EndDeviceFetcherConfig    `name:"fetcher" description:"End Device fetcher configuration"`
 	MQTT             config.MQTT               `name:"mqtt" description:"MQTT configuration"`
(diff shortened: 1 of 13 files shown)


pkg/applicationserver/config.go
```diff
@@ -92,6 +93,12 @@ var (
 	errWebhooksTarget   = errors.DefineInvalidArgument("webhooks_target", "invalid webhooks target `{target}`")
 )
 
+// UplinkStorageConfig defines the configuration of the application uplinks storage used by integrations.
+type UplinkStorageConfig struct {
+	Registry ApplicationUplinkRegistry `name:"-"`
+	Limit    int64                     `name:"limit" description:"How many application uplinks should be stored"`
```
johanstokking:
```suggestion
	Limit    int64                     `name:"limit" description:"Number of application uplinks should be stored"`
```

johanstokking (approved):

KrishnaIyer (approved):

htdvisser (approved):

PR #4068, opened 2021-04-16
LoRaCloud GLS multi frame request support

### LINKED ISSUE #4019: Multi frame request support for LoRaCloud GLS

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/4061
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4019

#### Changes

- Add multi frame requests API support and updated tests
- Add multi frame requests support to the LoRaCloud GLS package

#### Testing

Unit testing.

##### Regressions

Additional unit testing will improve coverage (WIP).

#### Notes for Reviewers

~This is work in progress - the only part that is missing is the Console support. But the backend logic may already be reviewed.~

~The PR targets `feature/4018-recent-uplink-storage` in order to ease the job of the reviewers (as it builds up upon the changes from that PR).~

#### Checklist


pkg/applicationserver/io/packages/loragls/v3/api/objects.go
```diff
@@ -85,30 +87,65 @@ type SingleFrameRequest struct {
...
+	return Gateway{
+			GatewayID: gtwUID,
 			Latitude:  m.Location.Latitude,
 			Longitude: m.Location.Longitude,
 			Altitude:  float64(m.Location.Altitude),
-		})
-		r.Frame = append(r.Frame, Uplink{
-			GatewayID: m.GatewayID,
+		}, Uplink{
+			GatewayID: gtwUID,
```
johanstokking: Do they need to know this, i.e. are we required to leak the UID? If not, let's refrain from doing so. I think they only need these values to be unique per request to correlate them in the algorithm, but they don't have to be stable over time and certainly not include potentially sensitive information (customer name, site names etc).

johanstokking (commented):
LGTM, minor comments. Need changelog entry also

johanstokking (approved):
Backend approved

adriansmares:
I've marked this ready for (console) review. The form has been updated but the behavior is not correct, and I'm not entirely sure how to fix it.

The initial value for the `queryState` / `multiFrame` states is created before the actual package data is loaded. As soon as the package data is actually retrieved, the `initialValues` will contain the correct values (and state for that matter), but the states are stuck to the default values until the user interacts with the select / checkbox.

A possible solution that I thought about is to always conditionally render based on `initialValues`, and force a re-render using a dummy state, but I'm not sure if that's the best solution.

cc @bafonins / @kschiffer

pkg/webui/console/containers/lora-cloud-gls-form/index.js
```diff
@@ -49,21 +51,50 @@ const m = defineMessages({
   setLoRaCloudToken: 'Set LoRa Cloud token',
   deleteWarning:
     'Are you sure you want to delete the LoRaCloud Geolocation token? This action cannot be undone.',
+  queryType: 'Query type',
+  queryTypeDescription: 'What kind of geolocation query should be used',
+  multiFrame: 'Multiframe',
+  multiFrameDescription: 'Enable multiframe lookups',
```
kschiffer:
```suggestion
  multiFrameDescription: 'Enable multiframe lookups to improve accuracy',
```

pkg/webui/console/containers/lora-cloud-gls-form/index.js
```diff
@@ -49,21 +51,50 @@ const m = defineMessages({
...
+  multiFrameWindowSize: 'Multiframe window size',
+  multiFrameWindowSizeDescription:
+    'How many historical message to send as part of the request. A window size of 0 automatically determines this based on the first byte of the payload.',
```
kschiffer:
```suggestion
    'How many historical messages to send as part of the request. Using 0 will automatically determine this based on the first byte of the payload.',
```

pkg/webui/console/containers/lora-cloud-gls-form/index.js
```diff
@@ -49,21 +51,50 @@ const m = defineMessages({
...
+  multiFrameTimeWindow: 'Multiframe time window',
+  multiFrameTimeWindowDescription: 'How recent the historical messages should be, in minutes.',
```
kschiffer:
```suggestion
  multiFrameTimeWindowDescription: 'The maximum age of considered historical messages in minutes',
```

pkg/webui/console/containers/lora-cloud-gls-form/index.js
```diff
@@ -49,21 +51,50 @@ const m = defineMessages({
...
 const validationSchema = Yup.object()
   .shape({
     data: Yup.object().shape({
       token: Yup.string().required(sharedMessages.validateRequired),
-      query: Yup.string().oneOf(['TOARSSI']),
+      query: Yup.string()
+        .oneOf(LORACLOUD_GLS_QUERY_VALUES)
+        .required(sharedMessages.validateRequired),
+      multi_frame: Yup.boolean().required(sharedMessages.validateRequired),
+      multi_frame_window_size: Yup.number()
```
bafonins: As far as I understand `multi_frame_window_size` and `multi_frame_window_age` should be set only when `multi_frame` is `true`. If thats the case, please strip these values within the validation schema.

```js
multi_frame_window_size: Yup.number().when('multi_frame', {
  is: true,
  then: schema => schema
          .min(0, Yup.passValues(sharedMessages.validateNumberGte))
          .max(16, Yup.passValues(sharedMessages.validateNumberLte))
  otherwise: schema => schema.strip()
})
```

pkg/webui/console/containers/lora-cloud-gls-form/index.js
```diff
@@ -49,21 +51,50 @@ const m = defineMessages({
...
+        .min(0, Yup.passValues(sharedMessages.validateNumberGte))
+        .max(16, Yup.passValues(sharedMessages.validateNumberLte))
+        .required(sharedMessages.validateRequired),
+      multi_frame_window_age: Yup.number()
```
bafonins: Same as described for `multi_frame_window_size`

pkg/webui/console/containers/lora-cloud-gls-form/index.js
```diff
@@ -124,6 +155,15 @@ const LoRaCloudGLSForm = () => {
     throw error
   }
 
+  const initialQuery = initialValues.data.query
+  const [queryType, setQueryType] = useState(initialQuery)
+
+  const initialMultiFrame = initialValues.data.multi_frame
+  const [multiFrame, setMultiFrame] = useState(initialMultiFrame)
```
bafonins: nit
```suggestion
  const [multiFrame, setMultiFrame] = useState(initialValues.data.multi_frame)
```

Since `initialMultiFrame` is not used in any other place within the component

bafonins (changes requested):
Currently, `multi_frame_window_age`, `multi_frame_window_size` are always sent to the backend even when the `multi_frame` is not checked. 

Consider casting form values within the `handleSubmit` callback.

```js
const castedValues = validationSchema.cast(values)
await dispatch(
  promisifiedSetAppPkgDefaultAssoc(appId, LORA_CLOUD_GLS.DEFAULT_PORT, {
    package_name: LORA_CLOUD_GLS.DEFAULT_PACKAGE_NAME,
    ...castedValues,
  }),
)

```

Combined with conditional validation this will remove unwanted fields in the payload.

kschiffer (changes requested):
When using `useState()` in react, that only initializes the value on first render, not every rerender. To keep the state and store in sync you have to use `useEffect()` to update the state whenever the initial value changed. Also, if you integrate the default values into the validation schema, you can use casting to generate the new default values when multiframe gets enabled.

<details><summary>Here's a diff that will make it work</summary>

```diff
diff --git a/pkg/webui/console/containers/lora-cloud-gls-form/index.js b/pkg/webui/console/containers/lora-cloud-gls-form/index.js
index 7976c3407..8d38ea948 100644
--- a/pkg/webui/console/containers/lora-cloud-gls-form/index.js
+++ b/pkg/webui/console/containers/lora-cloud-gls-form/index.js
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.

-import React, { useCallback, useState } from 'react'
+import React, { useCallback, useState, useEffect, useRef } from 'react'
 import { defineMessages } from 'react-intl'
 import { useDispatch, useSelector } from 'react-redux'

@@ -74,30 +74,33 @@ const validationSchema = Yup.object()
       token: Yup.string().required(sharedMessages.validateRequired),
       query: Yup.string()
         .oneOf(LORACLOUD_GLS_QUERY_VALUES)
+        .default(LORACLOUD_GLS_QUERY_TYPES.TDOARSSI)
         .required(sharedMessages.validateRequired),
-      multi_frame: Yup.boolean().required(sharedMessages.validateRequired),
-      multi_frame_window_size: Yup.number()
-        .min(0, Yup.passValues(sharedMessages.validateNumberGte))
-        .max(16, Yup.passValues(sharedMessages.validateNumberLte))
-        .required(sharedMessages.validateRequired),
-      multi_frame_window_age: Yup.number()
-        .min(1, Yup.passValues(sharedMessages.validateNumberGte))
-        .max(7 * 24 * 60, Yup.passValues(sharedMessages.validateNumberLte))
+      multi_frame: Yup.boolean()
+        .default(false)
         .required(sharedMessages.validateRequired),
+      multi_frame_window_size: Yup.number().when('multi_frame', {
+        is: true,
+        then: Yup.number()
+          .min(0, Yup.passValues(sharedMessages.validateNumberGte))
+          .max(16, Yup.passValues(sharedMessages.validateNumberLte))
+          .default(0)
+          .required(sharedMessages.validateRequired),
+        otherwise: Yup.number().strip(),
+      }),
+      multi_frame_window_age: Yup.number().when('multi_frame', {
+        is: true,
+        then: Yup.number()
+          .min(1, Yup.passValues(sharedMessages.validateNumberGte))
+          .max(7 * 24 * 60, Yup.passValues(sharedMessages.validateNumberLte))
+          .default(1440)
+          .required(sharedMessages.validateRequired),
+        otherwise: Yup.number().strip(),
+      }),
     }),
   })
   .noUnknown()

-const defaultValues = {
-  data: {
-    token: '',
-    query: LORACLOUD_GLS_QUERY_TYPES.TDOARSSI,
-    multi_frame: false,
-    multi_frame_window_size: 0,
-    multi_frame_window_age: 1440,
-  },
-}
-
 const promisifiedSetAppPkgDefaultAssoc = attachPromise(setAppPkgDefaultAssoc)
 const promisifiedDeleteAppPkgDefaultAssoc = attachPromise(deleteAppPkgDefaultAssoc)

@@ -105,13 +108,14 @@ const LoRaCloudGLSForm = () => {
   const [error, setError] = useState('')
   const appId = useSelector(selectSelectedApplicationId)
   const selector = ['data']
+  const formRef = useRef(null)

   const dispatch = useDispatch()
   const defaultAssociation = useSelector(state =>
     selectApplicationPackageDefaultAssociation(state, LORA_CLOUD_GLS.DEFAULT_PORT),
   )
   const packageError = useSelector(selectGetApplicationPackagesError)
-  const initialValues = validationSchema.cast(defaultAssociation || defaultValues)
+  const initialValues = validationSchema.cast(defaultAssociation)

   const handleSubmit = useCallback(
     async values => {
@@ -158,11 +162,20 @@ const LoRaCloudGLSForm = () => {
   const initialQuery = initialValues.data.query
   const [queryType, setQueryType] = useState(initialQuery)

-  const initialMultiFrame = initialValues.data.multi_frame
-  const [multiFrame, setMultiFrame] = useState(initialMultiFrame)
-  const handleMultiFrameChange = useCallback(evt => setMultiFrame(evt.target.checked), [
-    setMultiFrame,
-  ])
+  const [multiFrame, setMultiFrame] = useState()
+  const handleMultiFrameChange = useCallback(
+    evt => {
+      setMultiFrame(evt.target.checked)
+      const { setValues, values } = formRef.current
+      setValues(validationSchema.cast(values))
+    },
+    [setMultiFrame],
+  )
+
+  useEffect(() => {
+    // Keep multiframe state in sync with global store.
+    setMultiFrame(initialValues.data.multi_frame)
+  }, [initialValues.data.multi_frame])

   return (
     <RequireRequest
@@ -173,7 +186,7 @@ const LoRaCloudGLSForm = () => {
         validationSchema={validationSchema}
         initialValues={initialValues}
         onSubmit={handleSubmit}
-        enableReinitialize
+        formikRef={formRef}
       >
         <Form.Field
           component={Input}
@@ -188,6 +201,7 @@ const LoRaCloudGLSForm = () => {
           description={m.queryTypeDescription}
           name="data.query"
           options={LORACLOUD_GLS_QUERY_LABELS}
+          disabled={LORACLOUD_GLS_QUERY_LABELS.length === 1}
           onChange={setQueryType}
           required
         />
@@ -211,6 +225,7 @@ const LoRaCloudGLSForm = () => {
                   type="number"
                   min={0}
                   max={16}
+                  inputWidth="xs"
                   required
                 />
                 <Form.Field
@@ -221,6 +236,7 @@ const LoRaCloudGLSForm = () => {
                   type="number"
                   min={1}
                   max={7 * 24 * 60}
+                  inputWidth="xs"
                   required
                 />
               </>
```
</details>

bafonins:
@adriansmares The problem here is that `React.useState` is initialized only once and hence in 
```js
  const initialMultiFrame = initialValues.data.multi_frame
  const [multiFrame, setMultiFrame] = useState(initialMultiFrame) // `initialMultiFrame` is always `false` initially
```

`multiFrame` value in react state is only updated when `handleMultiFrameChange` is called.

The simplest solution here is to fetch values for both `das` and `gls` forms when the user initially navigates to `/console/applications/test-app2/integrations/lora-cloud` (so in `application-integrations-lora-cloud` component) and not when the user actully expands the form.

Another option would be to use the `withRequest` hoc in both `gls` and `das` forms to fetch values and render the loading indicator before rendering the form.

You can also react to changes to `defaultAssociation` inside `React.useEffect` and call `setMultiFrame` directly when `defaultAssociation` changes

kschiffer:
> The simplest solution here is to fetch values for both `das` and `gls` forms when the user initially navigates to `/console/applications/test-app2/integrations/lora-cloud` (so in `application-integrations-lora-cloud` component) and not when the user actully expands the form.

This is ultimately the way to go I'd say, but let's track this in [a new issue](https://github.com/TheThingsNetwork/lorawan-stack/issues/4099) and keep the diff small in this PR.

kschiffer (approved):
LGTM now.

bafonins (changes requested):
When deleting token, the form should be reset to its default state and do not persist any values. You can use `formRef.current.resetForm({values: ...})` to achieve this within the `handleDelete` callback.

Currently, when deleting existing token `data.token` persists the token and `data.multi_frame` is checked.

bafonins (approved):

PR #4069, opened 2021-04-17
Add read only option for rights selection

### LINKED ISSUE #3991: Preset rights addition to rights setting in console
#### Summary
Add profiles for preset rights in console at collaborator rights for individual users or organizations. 

#### Why do we need this?
The request is to simplify using the existing right-structure available in console to make configuring read-only easier. Therefor it is requested to add: 
- Grant read-only rights

In Console currently, predefined settings are available for:
- Grant all current and future rights
- Grant individual rights

As simple method to add read-only rights.

Add "read-only rights" to the preset options.

Verify if no other rights have been assigned.

NO

asmulko:
#### Summary

Closes #3991

#### Changes

- Added read-only option for rights selection

Before:
before_readonly_1

After:
after_readonly_1

#### Testing

- Manually tested

#### Checklist

### DIFF
--- a/pkg/webui/console/components/rights-group/index.js
+++ b/pkg/webui/console/components/rights-group/index.js
@@ -16,6 +16,7 @@ import React from 'react'
 import bind from 'autobind-decorator'
 import { defineMessages, injectIntl } from 'react-intl'
 import classnames from 'classnames'
+import { isEqual } from 'lodash'
 
 import Checkbox from '@ttn-lw/components/checkbox'
 import Notification from '@ttn-lw/components/notification'
@@ -37,6 +38,7 @@ const m = defineMessages({
     "This {entityType} has a wildcard right that you don't have. The {entityType} can therefore only be removed entirely.",
   grantType: 'Grant type',
   allCurrentAndFutureRights: 'Grant all current and future rights',
+  grantAllReadOnlyRights: 'Grant all read-only rights',
   selectIndividualRights: 'Grant individual rights',
 })
 
@@ -75,8 +77,9 @@ const computeProps = props => {
     value.includes(RIGHT_ALL) ||
     derivedPseudoRight.some(derivedRight => value.includes(derivedRight))
 
-  // Determine the current grant type.
-  const grantType = hasPseudoRightGranted ? 'pseudo' : 'individual'
+  // Determine whether read-only rights are granted.
+  const readOnlyRights = derivedRights.filter(right => right.includes('_READ'))
+  const hasReadOnlyRights = isEqual(readOnlyRights.sort(), value.sort())
 
   return {
     outOfOwnScopeIndividualRights,
@@ -84,7 +87,7 @@ const computeProps = props => {
     derivedPseudoRight,
     derivedRights,
     hasPseudoRightGranted,
-    grantType,
+    hasReadOnlyRights,
     ...props,
   }
 }
@@ -106,8 +109,6 @@ class RightsGroup extends React.Component {
      * rights for.
      */
     entityTypeMessage: PropTypes.message.isRequired,
-    /** The right grant type. */
-    grantType: PropTypes.oneOf(['pseudo', 'individual']).isRequired,
     /**
      * Whether the entity has a pseudo right that the current use does not
      * have.
@@ -115,6 +116,8 @@ class RightsGroup extends React.Component {
     hasOutOfOwnScopePseudoRight: PropTypes.bool.isRequired,
     /** Whether the entity has a pseudo right granted. */
     hasPseudoRightGranted: PropTypes.bool.isRequired,
+    /** Whether the entity has read-only rights. */
+    hasReadOnlyRights: PropTypes.bool.isRequired,
     /**
      * The intl object provided by injectIntl of react-intl, used to translate
      * messages.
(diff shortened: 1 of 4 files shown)


pkg/webui/console/components/rights-group/index.js
```diff
@@ -186,12 +190,17 @@ class RightsGroup extends React.Component {
...
   handleGrantTypeChange(val) {
-    const { onChange, pseudoRight } = this.props
+    const { onChange, pseudoRight, derivedRights } = this.props
     const { individualRightValue } = this.state
 
     if (val === 'pseudo') {
       onChange([pseudoRight])
+    } else if (val === 'readonly') {
+      this.setState({ selectReadOnly: true })
+      const readOnlyRights = derivedRights.filter(right => right.includes('_READ'))
```
bafonins, line 200: What about `_INFO` rights? These also grant read rights to entities as well

kschiffer, line 200: It's not only the rights that contain the word `READ`, unfortunately. It's necessary to go through the list and add rights individually. In that regard, does it make sense to add a new pseudo right for read only rights, @htdvisser? Otherwise we'll have to maintain a list in the console and update it any time a new read-only right is added.

If we don't add a pseudo right, we should add a (maintained) constant module that contains the read only rights and then use this list to intersect it with the available rights that can be granted within the rights-group component (like you currently do with the filter).

pkg/webui/console/components/rights-group/index.js
```diff
@@ -75,16 +77,17 @@ const computeProps = props => {
     value.includes(RIGHT_ALL) ||
     derivedPseudoRight.some(derivedRight => value.includes(derivedRight))
 
-  // Determine the current grant type.
-  const grantType = hasPseudoRightGranted ? 'pseudo' : 'individual'
+  // Determine whether read-only rights are granted.
+  const readOnlyRights = derivedRights.filter(right => right.includes('_READ'))
```
bafonins, line 81: What about `_INFO` rights? These also grant read rights to entities as well

kschiffer (changes requested):
Let's wait with this until we determined how to source the read-only rights.

pkg/webui/console/components/rights-group/index.js
```diff
@@ -75,16 +77,17 @@ const computeProps = props => {
...
+  const hasReadOnlyRights = isEqual(readOnlyRights.sort(), value.sort())
```
bafonins, line 82: You dont really need to filter first, then sort each list and compare every element. You can replace this by something like this:
```suggestion
  const hasReadOnlyRights = !hasPseudoRightGranted && value.every(right => right.includes('_READ'))
```

pkg/webui/console/components/rights-group/index.js
```diff
@@ -205,19 +214,32 @@ class RightsGroup extends React.Component {
...
 
     const selectedCheckboxesCount = individualRightValue.filter(right => !right.endsWith('_ALL'))
       .length
     const totalCheckboxesCount = derivedRights.length
     const allSelected = selectedCheckboxesCount === totalCheckboxesCount
     const indeterminate =
       selectedCheckboxesCount !== 0 && selectedCheckboxesCount !== totalCheckboxesCount
-    const allDisabled = grantType === 'pseudo' || disabled || hasOutOfOwnScopePseudoRight
+    const allDisabled =
+      grantType === 'pseudo' || grantType === 'readonly' || disabled || hasOutOfOwnScopePseudoRight
```
bafonins, line 242: Please introduce an enum for each grant type.
```js
const GRANT_TYPES = {
  INDIVIDUAL: 'individual',
  ...
}
```

pkg/webui/console/components/rights-group/index.js
```diff
@@ -205,19 +214,32 @@ class RightsGroup extends React.Component {
...
       derivedPseudoRight,
       derivedRights,
+      hasPseudoRightGranted,
+      hasReadOnlyRights,
     } = this.props
-    const { individualRightValue } = this.state
+    const { individualRightValue, selectReadOnly } = this.state
+
+    // Determine the current grant type.
+    const initialReadOnlyValue = (selectReadOnly === undefined && hasReadOnlyRights) || false
```
bafonins, line 225: Isnt this value initially always `false`?

bafonins (changes requested):

pkg/webui/console/components/rights-group/index.js
htdvisser: I don't think we should add a pseudo right for "read only" rights, because in most cases that isn't actually what you want (if you don't trust your friend enough to give them write access, you probably also don't want them to read your device keys).

htdvisser:
For the record: I don't think we should do this at all. We want our users to think about what rights they want to give collaborators or integrations, not just blindly give full read access to everything because they're too lazy to check a couple of boxes.

If we want to make peoples lives easier, I'd much rather see something like:

1. Copying collaborator rights from another application/gateway on create
2. Copying collaborator rights from an existing collaborator of the entity
3. Using a wizard (like we had with application linking) that creates an API key with link rights for gateways

kschiffer:
Good points. We should not add this like this then. I'll close this PR and make a comment in the issue about how to proceed with this.

PR #4072, opened 2021-04-19
Cache GS calls to the IS for UDP gateways

neoaggelos:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/issues/2437

#### Changes

- Wrap `(io.Server).FillGatewayContext` for the UDP frontend and cache results.
- Add a `GatewayIdentifiersCache` option in `gatewayserver.Config`. The TTL for entries is currently hard-coded to 1 minute.

#### Testing

- Unit tests

##### Regressions

- Changes to the Entity Registry will render entries in the cache out of date. This may cause temporary issues until the cache key expires.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/udp/server.go
+++ b/pkg/gatewayserver/io/udp/server.go
@@ -1,0 +1,60 @@
+// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package udp
+
+import (
+	"context"
+
+	"github.com/bluele/gcache"
+	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/io"
+	"go.thethings.network/lorawan-stack/v3/pkg/log"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+type gsWithIdentifiersCache struct {
+	io.Server
+	cache gcache.Cache
+}
+
+type fillGatewayContextResult struct {
+	ctx context.Context
+	ids ttnpb.GatewayIdentifiers
+	err error
+}
+
+// FillGatewayContext implements the io.Server interface, and caches the identifiers fetched by the Entity Registry.
+func (s *gsWithIdentifiersCache) FillGatewayContext(ctx context.Context, ids ttnpb.GatewayIdentifiers) (context.Context, ttnpb.GatewayIdentifiers, error) {
+	if cached, err := s.cache.Get(ids); err == nil {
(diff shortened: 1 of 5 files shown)


pkg/gatewayserver/io/udp/server.go
```diff
@@ -0,0 +1,60 @@
...
+	ctx context.Context
+	ids ttnpb.GatewayIdentifiers
+	err error
+}
+
+// FillGatewayContext implements the io.Server interface, and caches the identifiers fetched by the Entity Registry.
+func (s *gsWithIdentifiersCache) FillGatewayContext(ctx context.Context, ids ttnpb.GatewayIdentifiers) (context.Context, ttnpb.GatewayIdentifiers, error) {
+	if cached, err := s.cache.Get(ids); err == nil {
+		if result, ok := cached.(*fillGatewayContextResult); ok {
+			return result.ctx, result.ids, result.err
```
johanstokking, line 41: You are now returning a completely different context here, potentially a different request from a different frontend etc, or any other mixup can occur. That breaks logging, security, lifetime, everything basically. I hope you understand that this doesn't make any sense.

This should first be refactored so that there are context fillers that are cached. Those context fillers can then be applied to the request context.

Please see also https://blog.golang.org/context

johanstokking (changes requested):

pkg/gatewayserver/io/udp/server.go
neoaggelos: Indeed, I know that in the general case storing and retrieving the context from a cache is not a "correct" solution.

However, in this particular case, the change is scoped to the UDP frontend only, and the context is the global context of the UDP frontend, not a request context: https://github.com/TheThingsNetwork/lorawan-stack/blob/ccc881d90107d3c3a4bafa509d13ab8063b8f465/pkg/gatewayserver/io/udp/udp.go#L161).

I can agree though that refactoring the context fillers themselves instead should be the preferrable solution here.

adriansmares (commented):
Maybe it would not be a bad idea to have `FillGatewayContext` refactored to return `[]fillcontext.Filler` instead of a new context. 

I personally believe that we're trying to shoehorn our 'cache around interface' pattern, and instead should consider implementing caching this inside `*GatewayServer`. You can have an extra parameter, `cacheIdentifiers bool`, which would be true for some frontends, and false to other frontends. 

You can even have a wrapper that exposes `io.Server`, wrapping `*GatewayServer` and a `bool` and hides the parameter (curries) the `FillGatewayContext`. Then you don't even need to update the call sites for the frontends.

KrishnaIyer (commented):
I think that we should keep the frontends agnostic of the IS calls. 
There are three locations that we call the IS
- `FillGatewayContext` (only for UDP/LBS to fetch the tenant and the ID)
- `Connect`
  - `RequireGateway` to assert rights
  - A `GET` fetch the gateway.

What I think could be done is to implement caching on the GS level (as @adriansmares suggested). 
- Implement a Cache Registry on the GS that can be enabled/disabled and TTL can be set.
- We completely decouple the GS and the IS by making the Cache Registry implement the `GatewayRegistry` interface; the GS only needs the `GetGateways` function.
- We return always this interface when the GS calls the `getRegistry`.
- When the GS calls the `registry.Get` we check the cache for the IDs; if it's a miss, then we query the IS, update the cache.
- Basically, GS always talks to the Cache Registry. The Cache Registry calls the IS if needed.
- We still need to figure out 
   - What to do with `rights.RequireGateways` since that needs an IS connection.
   - What to do with auth when the IS is unreachable
   - My solution is to cache the last valid token that passed the auth check and check incoming connections against that until the TTL or the IS becomes available.

PR #4075, opened 2021-04-20
Pass end device version identifiers to Javascript payload formatters

### LINKED ISSUE #4009: End device version identifiers (brand, model, versions) in streaming messages
#### Summary
It would be useful for downstream applications as well as for payload formatters to get additional information about the device type of where the message originated.

...

#### Why do we need this?
In current v3, Payload Formatters of known devices will place a "decoded_payload" into the "uplink_message", but applications that receive this message has to resort to "object pattern matching" to figure out what device type it came from. Such method is somewhat fragile, and it would be much better to get this in the "uplink_message" directly, rather than somewhat harder "lookup from device_id" which I assume is possible.

For unknown device types, the Payload Formatter that one can add in Console can not AFAIK, look up which device or device type that sent the message. This pushes independent payload formatter for each device, which is seemingly wasteful and error-prone.
...

In the messages that external applications are subscribed to, there is no information of how the "decoded_payload" was constructed.
In the JavaScript function signature for decodeUplink (et al), there are only the raw bytes, and that makes it nearly impossible to do application level decoding as soon as more than one device type is used, other than putting it into every single device.
...

In messages, I would like to have an additional section "device_info", with "brand", "model", "hw_version" and "fw_version".
```json
{
  "device_info" : {
    "brand" : "dragino",
    "model" : "lht65",
    "hw_version" : "unknown",
    "fw_version" : "1.8"
  }
}
```
...

I don't (yet) know the internals of the stack and have no opinion.

...

Again, I don't know anything about the stack's internals or how the tests are built up or executed at the moment.
...

Not at the moment.

...

neoaggelos:
#### Changes

- Pass end device version identifiers in Javascript payload formatters

#### Testing

Unit tests, test locally

#### Notes for Reviewers

- Local testing does not help with finding out if there is any serious performance penalty due to the extra parameters. 
- @johanstokking Should we support legacy formatters as well? Or simply document that people need to use the new payload formatter definitions (`decodeUplink`, `decodeDownlink`, `encodeDownlink`) payload formatters if they want to access the version identifiers?

#### Checklist

### DIFF
--- a/pkg/messageprocessors/javascript/javascript.go
+++ b/pkg/messageprocessors/javascript/javascript.go
@@ -40,9 +40,31 @@ func New() messageprocessors.PayloadEncodeDecoder {
 	}
 }
 
+type endDeviceVersionIdentifiers struct {
+	BrandID         string `json:"brandID"`
+	ModelID         string `json:"modelID"`
+	FirmwareVersion string `json:"firmwareVersion"`
+	HardwareVersion string `json:"hardwareVersion"`
+	BandID          string `json:"bandID"`
+}
+
+func fromPB(version *ttnpb.EndDeviceVersionIdentifiers) endDeviceVersionIdentifiers {
+	v := endDeviceVersionIdentifiers{}
+	if version == nil {
+		return v
+	}
+	v.BrandID = version.BrandID
+	v.ModelID = version.ModelID
+	v.FirmwareVersion = version.FirmwareVersion
+	v.HardwareVersion = version.HardwareVersion
+	v.BandID = version.BandID
+	return v
+}
+
 type encodeDownlinkInput struct {
-	Data  map[string]interface{} `json:"data"`
-	FPort *uint8                 `json:"fPort"`
+	Data    map[string]interface{}      `json:"data"`
+	FPort   *uint8                      `json:"fPort"`
+	Version endDeviceVersionIdentifiers `json:"version"`
 }
 
 type encodeDownlinkOutput struct {
@@ -72,8 +94,9 @@ func (h *host) EncodeDownlink(ctx context.Context, ids ttnpb.EndDeviceIdentifier
 	}
 	fPort := uint8(msg.FPort)
 	input := encodeDownlinkInput{
-		Data:  data,
-		FPort: &fPort,
+		Data:    data,
+		FPort:   &fPort,
+		Version: fromPB(version),
 	}
 
 	// Fallback to legacy Encoder() function for backwards compatibility with The Things Network Stack V2 payload functions.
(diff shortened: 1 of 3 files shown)


adriansmares (approved):
I don't think the legacy function formats should support this. We want people to move away from them, not to have two function styles going around. 

Other than that LGTM.

pkg/messageprocessors/javascript/javascript.go
```diff
@@ -40,9 +40,31 @@ func New() messageprocessors.PayloadEncodeDecoder {
 	}
 }
 
+type endDeviceVersionIdentifiers struct {
+	BrandID   string `json:"brandID"`
+	ModelID   string `json:"modelID"`
+	FWVersion string `json:"fwVersion"`
+	HWVersion string `json:"hwVersion"`
```
johanstokking: Why abbreviate?

pkg/messageprocessors/javascript/javascript.go
```diff
@@ -85,7 +108,7 @@ func (h *host) EncodeDownlink(ctx context.Context, ids ttnpb.EndDeviceIdentifier
 				return encodeDownlink(input);
 			}
 			return {
-				bytes: Encoder(input.data, input.fPort),
+				bytes: Encoder(input.data, input.fPort, input.version),
```
johanstokking: Remove; we don't backport this to the legacy V2 formatters.

pkg/messageprocessors/javascript/javascript.go
```diff
@@ -143,12 +168,19 @@ func (h *host) DecodeUplink(ctx context.Context, ids ttnpb.EndDeviceIdentifiers,
...
+					fwVersion: input.version.fwVersion,
+					hwVersion: input.version.hwVersion,
+				},
 			}
 			if (typeof decodeUplink === 'function') {
 				return decodeUplink(input);
 			}
 			return {
-				data: Decoder(input.bytes, input.fPort)
+				data: Decoder(input.bytes, input.fPort, input.version)
```
johanstokking: Idem

johanstokking (commented):
This is not closing https://github.com/TheThingsNetwork/lorawan-stack/issues/4009: that issue is about including the version identifiers in streaming messages (i.e. `ApplicationUp`).

neoaggelos:
> This is not closing #4009: that issue is about including the version identifiers in streaming messages (i.e. `ApplicationUp`).

Okay, I was confused by the following:

> In the JavaScript function signature for decodeUplink (et al), there are only the raw bytes, and that makes it nearly impossible to do application level decoding as soon as more than one device type is used, other than putting it into every single device.

So do we actually want this PR, then? Or simply close and replace with a different one that simply has the AS append the version identifiers in the uplink message?

johanstokking:
> So do we actually want this PR, then? Or simply close and replace with a different one that simply has the AS append the version identifiers in the uplink message?

Let's close it. The "Codec API" (that is, the JavaScript function signature for encoding and decoding) is going to be standardized by the LoRa Alliance. We're going to push for the current format to be already compatible. Let's not add anything (for now).

PR #4084, opened 2021-04-22
Add Events Store and Redis implementation

htdvisser:
#### Summary

This pull request adds an `events.Store` interface and implementation on top of Redis.

Refs:

- https://github.com/TheThingsIndustries/lorawan-stack/issues/394
- https://github.com/TheThingsIndustries/lorawan-stack/issues/1213
- https://github.com/TheThingsIndustries/lorawan-stack/issues/1804
- https://github.com/TheThingsIndustries/lorawan-stack/pull/2620

#### Changes

- Change the API of `events.Subscriber` to take a list of event names instead of a glob.
- Add an `events.Store` interface that extends `events.PubSub`:

```go
// Store extends PubSub implementations with storage of historical events.
type Store interface {
	PubSub
	// FindRelated finds events with matching correlation IDs.
	FindRelated(ctx context.Context, correlationID string) ([]Event, error)
	// FetchHistory fetches the tail (optional) of historical events matching the given names (optional) and identifiers (mandatory) after the given time (optional).
	FetchHistory(ctx context.Context, names []string, ids []*ttnpb.EntityIdentifiers, after *time.Time, tail int) ([]Event, error)
	// SubscribeWithHistory is like FetchHistory, but after fetching historical events, this continues sending live events until the context is done.
	SubscribeWithHistory(ctx context.Context, names []string, ids []*ttnpb.EntityIdentifiers, after *time.Time, tail int, hdl Handler) error
}
```

- Implement `events.Store` interface with a new `redis.PubSubStore`
- Add config options to enable Redis events store

#### Testing

- Extended existing events pubsub test suite with store methods
- 🐒 test with CLI and Console

##### Regressions

None expected, and if there are any, the entire feature can be disabled with config (it's disabled by default).

#### Notes for Reviewers

This addresses points 1-4 from the "Architecture" section of the proposal in https://github.com/TheThingsIndustries/lorawan-stack/pull/2620, with some deviations:

- Unlike proposed in point 1 of the Architecture section, we still support Redis pubsub channels, but we can perhaps also remove that.
- Unlike proposed in points 2-3 of the Architecture section, `pkg/events/redis.PubSubStore` stores and publishes events directly instead of first going through a separate component. This is mainly to save bandwidth.
- Event metadata is slightly different than in point 3.3 of the Architecture section:
	- We include the entity identifiers in the Entity Event Stream for consistency
	- We add an `event_sparse` key that contains a sparse Event proto with some fields that we need in almost all use cases (context, timestamps and event visibility rights), without having to store/fetch the (large) full payload. This also allows us to set different expiry on the full payloads, while still being able to present most information of those events.
- Point 3.4 and 3.5 have not been implemented yet.

Still TODO:

- [x] Add RPC for finding related events (by correlation ID)
- [ ] Documentation
- [ ] Changelog

#### Checklist


htdvisser:
@kschiffer / @bafonins:

From now on the Console can use the `tail` and `after` fields in the event stream request to get some history before the live events. If history is not enabled on the deployment, a warning header is returned by the backend.

Perhaps also important for the Console is that with historical events the `data`, `correlation_ids`, `origin`, `authentication`, `remote_ip` and `user_agent` fields may not be available. Is the Console able to deal with that, or is some work needed?

pkg/events/redis/codec.go
```diff
@@ -0,0 +1,148 @@
...
+
+const (
+	eventUIDKey    = "event_uid"
+	eventNameKey   = "event_name"
+	entityIDKey    = "entity_id"
+	entityTypeKey  = "entity_type"
+	eventSparseKey = "event_sparse"
+)
+
+func encodeEventMeta(evt events.Event, id *ttnpb.EntityIdentifiers) []string {
```
adriansmares: I would prefer to propagate the errors here.

pkg/events/redis/redis.go
```diff
@@ -116,22 +132,28 @@ func (ps *PubSub) subscribeTask(ctx context.Context) error {
...
+				evtPB = &ttnpb.Event{}
+				err := decodeEventData(msg.Payload, evtPB)
+				if err != nil {
+					logger.WithError(err).Warn("Failed to decode event payload")
+					continue
+				}
+			case strings.HasPrefix(msg.Payload, metaEncodingPrefix):
+				m := strings.Split(strings.TrimPrefix(msg.Payload, metaEncodingPrefix), " ")
+				var err error
+				evtPB, err = (&PubSubStore{PubSub: ps}).LoadEvent(ctx, m[0])
```
adriansmares: You can preallocate this.

pkg/events/redis/store.go
```diff
@@ -0,0 +1,386 @@
...
+	}
+	var evtPB ttnpb.Event
+	if err = decodeEventData(data, &evtPB); err != nil {
+		return nil, err
+	}
+	return &evtPB, nil
+}
+
+func xMessageHasEventName(names ...string) func(msg redis.XMessage) bool {
+	if len(names) == 0 {
```
adriansmares, line 126: Consider converting the list to a `map[string]struct{}` in order to avoid looping the whole thing during matches. This may be something to consider for other places where this type of matching is done (loop through strings and check equality).

kschiffer:
> From now on the Console can use the `tail` and `after` fields in the event stream request to get some history before the live events. If history is not enabled on the deployment, a warning header is returned by the backend.

:ok\_hand: 

> Perhaps also important for the Console is that with historical events the `data`, `correlation_ids`, `origin`, `authentication`, `remote_ip` and `user_agent` fields may not be available. Is the Console able to deal with that, or is some work needed?

From looking at the code, it will cause problems when `data` is not defined since the Console expects these to be defined for certain event data types, so this needs to be adjusted. Should be a simple fix though that we can include together with enabling historical events in the Console.

johanstokking (approved):
Nothing to add

htdvisser:
@johanstokking and @adriansmares please take another look. I've added two commits to allow searching for events by Correlation ID: https://github.com/TheThingsNetwork/lorawan-stack/pull/4084/files/8c50657282b317359437c4d3404021a64e607f42..fb632cdd770595e14fa47aab3251db8eb501e6d9

pkg/events/grpc/grpc.go
```diff
@@ -119,6 +135,59 @@ func (srv *EventsServer) Stream(req *ttnpb.StreamEventsRequest, stream ttnpb.Eve
...
+				// Data is private
+				// CorrelationIDs is private
+				Origin: evtProto.Origin,
+				// Context is private
+				Visibility: evtProto.Visibility,
+				// Authentication is private
+				// RemoteIP is private
+				// UserAgent is private
+				UniqueID: evtProto.UniqueID,
+			})
```
htdvisser, line 185: Point for discussion: Should (for example) a Gateway Owner be able to see end device events related to a downlink that they see on their gateway, even if they don't have rights on that application?

I think it would be useful to at least see _that_ something happened, and to see which application or end device caused some downlink, so that the gateway owner could come in contact with the application owner if something is wrong.

The current implementation therefore lets you see such events, without including unnecessary (private) fields.

adriansmares: I think this needs a configuration toggle at least, since it allows people to 'fish' metadata to which they normally don't have access to. 

Our correlation IDs contain some entropy in them but we cannot guarantee our users have entropy into theirs.

adriansmares (approved):

pkg/events/grpc/grpc.go
johanstokking: For private networks, it should really be no problem that end devices identifiers are available in gateway events. I think indeed that's very useful to see what's going on.

On the community network this might be more of an issue, but let's just try it and see how it goes.

Let's not add configuration, that needs documentation, default config, deployment templates etc, for things that we don't know is necessary to configure.

cmd/ttn-lw-cli/commands/events.go
```diff
@@ -57,41 +63,99 @@ var eventsCommand = &cobra.Command{
...
+		for evt := range events {
+			io.Write(os.Stdout, config.OutputFormat, evt)
+		}
+
+		return ctx.Err()
+	},
+}
+
+var eventsFindRelatedCommand = &cobra.Command{
+	Use:     "find-related [CorrelationID]",
```
johanstokking:
```suggestion
	Use:     "find-related [correlation-id]",
```

johanstokking (approved):
Reviewed https://github.com/TheThingsNetwork/lorawan-stack/pull/4084/files/8c50657282b317359437c4d3404021a64e607f42..fb632cdd770595e14fa47aab3251db8eb501e6d9?file-filters%5B%5D=.go&file-filters%5B%5D=.proto

PR #4086, opened 2021-04-22
Various fixes for end device creation from Console and Device Repository

### LINKED ISSUE #4082: JS SDK ignores MAC state and MAC settings when importing devices
#### Summary

When importing devices via the Console, any MAC Settings and MAC state fields are discarded.

#### Steps to Reproduce

1. Go to Applications > Application > Import End Devices
2. Select "The Things Stack JSON" format and upload a JSON file where `mac_state` or `mac_settings` fields are set 
3. Import device

#### What do you see now?

No errors, but the MAC state and the MAC settings from the imported file are discarded.

This is because the JS SDK considers these fields to be read only (https://github.com/TheThingsNetwork/lorawan-stack/blob/69c4f0f4516357595b7d31281955e658637384f9/sdk/js/generated/device-entity-map.json#L612-L614) which was but no longer is the case.

The `device-entity-map.json` file is generated dynamically by CI with the `tools/bin/mage jssdk:devicefieldmasks` comand, but it is using `device-field-masks.json` as a source, which is not updated automatically (and is currently out of date).

The MAC settings and MAC state should be set after importing the device.

#### Environment

3.12.2

Create tooling that syncs the contents of `device-entity-map.json` based on the generated list of allowed field mask paths. This should be a simple enough tool, taking the list of field mask paths allowed per RPC, creating the `device-field-masks.json` dynamically and then executing the logic of `util/device-field-mask-mapper.js`.

CI should fail if the `device-entity-map.json` is out of date. Looking at `device-fields-masks.json`, it appears that we can skip it entirely, since it is basically constructed like so:

```
{
  ns: {
    get: AllowedFieldMaskPaths["ttn.lorawan.v3.NsEndDeviceRegistry/Get"],
    set: AllowedFieldMaskPaths["ttn.lorawan.v3.NsEndDeviceRegistry/Set"],
  },
  .....
}
```

CI should make sure that the device-entity-map.json file is up to date. End-to-end tests should also verify that nothing breaks at random in the Console.

Yes, unless I am missing any non-obvious steps. cc @bafonins

neoaggelos:
#### Summary

Closes #4082 

#### Changes

- Generate the `device-entity-map.json` file dynamically from the allowed field mask paths by the API.
- Fix `mage jsSDK:deviceFieldMasks` target to generate the file when the field mask paths have changed.
- Fix handling of empty `mac_state.current_parameters` field when creating devices
- Fix invalid Rx frequencies.

#### Testing

- Rebuild console with the new `device-entity-map.json` definitions and verify that the Console does not ignore MAC settings and MAC state fields.
- CI
- End to end tests
- Local testing

##### Regressions

There should not be any, the allowed fields should be up to date since they are generated by the API.

#### Notes for Reviewers

@bafonins I am not very familiar with the JS SDK. Is there anything I might have missed?

#### Checklist


neoaggelos:
Converting to draft until the End to end tests failures are resolved.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Fixed
 
 - Occasional crashes in the ratelimit middleware.
+- Importing end devices from the Console would occasionaly ignore some device MAC settings fields.
```
```suggestion
- Importing end devices from the Console would occasionally ignore some device MAC settings fields.
```

kschiffer (changes requested):
This appears to make problems when creating ABP devices via the device repository, which sets the MAC state directly. I don't have time to help you out with this now. @bafonins maybe?

johanstokking (approved):

kschiffer (approved):
LGTM

PR #4101, opened 2021-04-29
Fix display of gateway downlink events in the Console

### LINKED ISSUE #4078: Warnings in the TTS Console for `gs.down.send` event
#### Summary
We are seeing a warning, i.e., `There was an error and events cannot be displayed and The raw data can be viewed by clicking this row` in the gateway live data of `TTSC` Console for the `gs.down.send` event.
#### Steps to Reproduce

1. Register an end-device and Gateway in TTSC Console.
2. Schedule a `downlink`.
3. You could see the `successfully schedule data downlink` event in the end-device i.e., `ns.down.data.schedule.success`
4. Navigate to the gateway live data, observe `gs.down.send` event is shown as a warning i.e., `There was an error and events cannot be displayed. The raw data can be viewed by clicking this row`

#### What do you see now?

- The warning is seen whenever a `downlink` is scheduled from the Network Server, but we can see the raw event by clicking on it (refer to the image).
- Observed across all TTSC clusters.
image
image
image

 `gs.down.send` event.
#### Environment
TTS v3.12.1
Not sure
...
No.

kschiffer:
#### Summary
This PR will fix display issues of gateway downlink events.

References #4078 

#### Changes
- Distinguish gateway from end device downlink messages and use distinct display logic for gateways
- Fix wrong icon used for gateway downlink events
- Update event data used in event story

#### Testing

Manual testing using realistic data within storybook

##### Regressions

Shouldn't be any, but if then they would be related to previews for downlink messages in the Console event view.

#### Notes for Reviewers
This doesn't close #4078 since there appears to be another issue with downlink ack messages, for which I need more example data to fix. This will however already fix issues with downlink messages for gateways as described in the issue.

#### Checklist


pkg/webui/console/components/events/previews/downlink-message.js
```diff
@@ -38,6 +38,27 @@ const DownLinkMessagePreview = React.memo(({ event }) => {
   }
 
   if ('request' in data) {
+    const { name } = event
+    if (name.startsWith('gs')) {
+      const gatewayEUI = event.identifiers[0].gateway_ids.EUI
```
bafonins:
```suggestion
      const gatewayEUI = event.identifiers[0].gateway_ids.eui
```

bafonins (approved):

PR #4102, opened 2021-04-29
Do not print rate limiting error logs for RPC calls

neoaggelos:
#### Summary

Do not print error message logs for errors of type ResourceExhausted. This is to avoid filling up the logs.

#### Changes

- Override interceptors for HTTP and gRPC servers, and do not print messages if the error is of type ResourceExhausted
- Override interceptor for gRPC clients, because it is used by the gRPC gateway.

#### Testing

Locally

##### Regressions

None. Rate limited requests can still be observed with the metrics.

##### Notes for Reviewers

Targetting 3.12.2 since this is a pretty straightforward change.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
+- Do not print error line logs for rate limited gRPC and HTTP API requests.
+
 ### Deprecated
 
 ### Removed
--- a/pkg/rpcmiddleware/rpclog/rpclog.go
+++ b/pkg/rpcmiddleware/rpclog/rpclog.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"strings"
 
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
 	"go.thethings.network/lorawan-stack/v3/pkg/log"
 	"google.golang.org/grpc/grpclog"
 )
@@ -99,3 +100,7 @@ func (l *ttnGrpcLogger) Fatalf(format string, args ...interface{}) {
 func (l *ttnGrpcLogger) V(int) bool {
 	return true
 } // TODO: Use when log.Interface supports this
+
+func shouldSuppressError(err error) bool {
+	return errors.IsResourceExhausted(err)
+}
--- a/pkg/rpcmiddleware/rpclog/server_interceptors.go
+++ b/pkg/rpcmiddleware/rpclog/server_interceptors.go
@@ -41,6 +41,9 @@ func UnaryServerInterceptor(ctx context.Context, opts ...Option) grpc.UnaryServe
 				return resp, err
 			}
 		}
+		if shouldSuppressError(err) {
+			return resp, err
+		}
 
 		onceFields = onceFields.WithField(
 			"duration", time.Since(startTime).Round(time.Microsecond*100),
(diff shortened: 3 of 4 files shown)


KrishnaIyer (approved):

htdvisser (commented):
LGTM, but let's make sure that we don't unintentionally stop logging errors that we're actually interested in, but are (perhaps incorrectly) defined as `ResourceExhausted`:

- `no_username_available` in `pkg/account/oidc` (Enterprise)
- `channel_closed` in `pkg/applicationserver/distribution/redis`
- `buffer_full` in `pkg/applicationserver/io`
- `queue_full` in `pkg/applicationserver/io/web`
- `duty_cycle` in `pkg/gatewayserver/scheduling`

johanstokking:
Please rebase

neoaggelos:
Looking at the code:

- `duty_cycle` is always logged (DEBUG level) 
- `queue_full` is always logged (WARN level)
- `buffer_full` is always logged (WARN level)
- `channel_closed` is always logged (WARN level)
- `no_usernames_available` is returned by the echo handler, so I suppose is the only one that would be affected by this change. @adriansmares what do you think?

pkg/rpcmiddleware/rpclog/client_interceptors.go
```diff
@@ -33,6 +34,9 @@ func UnaryClientInterceptor(ctx context.Context, opts ...Option) grpc.UnaryClien
 
 		startTime := time.Now()
 		err := invoker(newCtx, method, req, reply, cc, opts...)
+		if errors.IsResourceExhausted(err) {
```
johanstokking: Please extract this to an unexposed function in `rpclog` to return whether to rate limit logs.

pkg/rpcmiddleware/rpclog/client_interceptors.go
```diff
@@ -33,6 +33,9 @@ func UnaryClientInterceptor(ctx context.Context, opts ...Option) grpc.UnaryClien
 
 		startTime := time.Now()
 		err := invoker(newCtx, method, req, reply, cc, opts...)
+		if shouldSuppressError(err) {
+			return err
+		}
```
htdvisser: I don't think we should suppress rate limit errors on the client side, because clients will most likely need to take action (slow down), so they shouldn't ignore these.

neoaggelos: The problem here is that this is used by the gRPC gateway. If we do not do this, then we get one log line per error.

neoaggelos: Removed this, let's keep an eye out and return if it becomes an issue.

CHANGELOG.md
```diff
@@ -59,6 +61,8 @@ For details about compatibility between different releases, see the **Commitment
 - LoRa Basics Station `router_config` message omits hardware specific fields.
 - Showing "Last seen" information in end device tables (replacing "Created").
 
+### Deprecated
```
johanstokking: This doesn't seem right

johanstokking (approved):

johanstokking:
Why isn't this merged?

neoaggelos:
Needs codeowner, cc @htdvisser / @kschiffer

htdvisser (approved):

PR #4116, opened 2021-05-03
Disable map link click in overview

asmulko:
#### Summary

Map overview widget should not navigate to location settings via map click, if map is already set (@kschiffer)

#### Changes

- Disabled `Link` click in map widget, when `markers` are set  

#### Testing

- Manually tested

#### Checklist

### DIFF
--- a/pkg/webui/components/map/widget/index.js
+++ b/pkg/webui/components/map/widget/index.js
@@ -15,6 +15,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
+import Link from '@ttn-lw/components/link'
 import LocationMap from '@ttn-lw/components/map'
 import WidgetContainer from '@ttn-lw/components/widget-container'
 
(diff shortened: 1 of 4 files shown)


pkg/webui/components/map/widget/index.js
```diff
@@ -73,13 +73,14 @@ export default class MapWidget extends React.Component {
...
   render() {
-    const { path } = this.props
+    const { path, markers } = this.props
 
     return (
       <WidgetContainer
         title={sharedMessages.location}
         toAllUrl={path}
         linkMessage={sharedMessages.changeLocation}
+        disableLink={markers && markers.length > 0}
```
kschiffer: This solution is a bit ambiguous. `disableLink` suggests that it disables all links (also the one in the top right hand corner). Actually, I think it's better to move the wrapping link out of the widget container to the passed child prop directly. That gives back full control over the behavior of the wrapped content.

kschiffer (changes requested):

kschiffer (approved):

PR #4117, opened 2021-05-03
Update device delete confirmation message

### LINKED ISSUE #4100: Change the pop-up message when deleting the device from the console.
#### Summary
While deleting the end-device from the console, the pop-up message in the console displays that we cannot reuse the `device ID`.

#### Steps to Reproduce
1. Create a device in the console
2. Delete the device via console

#### What do you see now?
While deleting the device via console, you will observe the pop-up message as `Are you sure you want to delete "xxx"? This action cannot be undone and it will not be possible to reuse the end device ID`.
device-id_deletion

In TTS V3, when an end device is deleted, the ID and EUI are released, and we can reuse them. Remove the notification as it is no more needed.
Ref: https://www.thethingsindustries.com/docs/reference/id-eui-constraints/

#### Environment
TTS v3.12.1

...

...

No. @kschiffer

asmulko:
#### Summary

Closes #4100 

#### Changes

- Updated device delete confirmation message

Before:
device_delete_before

After:
device_delete_after

#### Testing

- Manually tested

#### Checklist

### DIFF
--- a/pkg/webui/console/views/device-general-settings/identity-server-form/index.js
+++ b/pkg/webui/console/views/device-general-settings/identity-server-form/index.js
@@ -38,8 +38,7 @@ import validationSchema from './validation-schema'
 
 const messages = defineMessages({
   deleteDevice: 'Delete end device',
-  deleteWarning:
-    'Are you sure you want to delete "{deviceId}"? This action cannot be undone and it will not be possible to reuse the end device ID.',
+  deleteWarning: 'Are you sure you want to delete "{deviceId}"? This action cannot be undone.',
 })
 
 const IdentityServerForm = React.memo(props => {
--- a/pkg/webui/locales/en.json
+++ b/pkg/webui/locales/en.json
@@ -468,7 +468,7 @@
   "console.views.device-claim-authentication-code.index.updateSuccess": "Claim authentication updated",
   "console.views.device-claim-authentication-code.index.validateCode": "Claim authentication code must consist only of numbers and letters",
   "console.views.device-general-settings.identity-server-form.index.deleteDevice": "Delete end device",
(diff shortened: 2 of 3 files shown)


pkg/webui/console/views/device-general-settings/identity-server-form/index.js
```diff
@@ -38,8 +38,7 @@ import validationSchema from './validation-schema'
 
 const messages = defineMessages({
   deleteDevice: 'Delete end device',
-  deleteWarning:
-    'Are you sure you want to delete "{deviceId}"? This action cannot be undone and it will not be possible to reuse the end device ID.',
+  deleteWarning: 'Are you sure you want to delete "{deviceId}"?',
```
```suggestion
  deleteWarning: 'Are you sure you want to delete "{deviceId}"? This action cannot be undone.',
```

I think it's still good to make users aware that the IDs are typically not reusable right away but after a retention time only.

cc @htdvisser

kschiffer (changes requested):
Please also adjust the messages for deleting all other entities (gateways, organizations, applications, users).

pkg/webui/console/views/device-general-settings/identity-server-form/index.js
htdvisser: See https://www.thethingsindustries.com/docs/reference/id-eui-constraints/#deleted-entities

> When an end device is deleted, the ID and EUI are released and can be reused

kschiffer: I see! See my updated change suggestion then @asmulko

kschiffer (changes requested):
~~This is still adjusting the device modal only.~~

> Please also adjust the messages for deleting all other entities (gateways, organizations, applications, users).

kschiffer:
Disregard the above. Misunderstanding, I'll approve.

kschiffer (approved):

PR #4120, opened 2021-05-03
End device version identifiers in upstream uplink messages

### LINKED ISSUE #4009: End device version identifiers (brand, model, versions) in streaming messages

neoaggelos:
#### Summary

Closes #4009

#### Changes

- Add `VersionIDs` in `UplinkMessage` proto message.
- Set `VersionIDs` in Application Server before forwarding message

#### Testing

Locally, MQTT, HTTP webhooks, with and without version identifiers.

#### Checklist


api/identifiers.proto
```diff
@@ -95,3 +95,11 @@ message EntityIdentifiers {
   }
 }
 
+// Identifies an end device model with version information.
+message EndDeviceVersionIdentifiers {
+  string brand_id = 1 [(gogoproto.customname) = "BrandID", (validate.rules).string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$", max_len: 36, ignore_empty: true}];
```
adriansmares, line 105: Nit: Remove `customname`. Same for `model_id`.

api/messages.proto
```diff
@@ -223,7 +223,10 @@ message ApplicationUplink {
   // End device location metadata, set by the Application Server while handling the message.
   map<string,Location> locations = 14;
 
-  // next: 15
+  // End device version identifiers, set by the Application Server while handling the message.
+  EndDeviceVersionIdentifiers version_ids = 15 [(gogoproto.customname) = "VersionIDs"];
```
adriansmares, line 227: Nit: Remove `customname`.

adriansmares (approved):

api/identifiers.proto
neoaggelos: This pretty much means that all `.ModelID` references will need to be updated to `ModelId`, which is a rather large diff in the context of this PR. I think it is better to make this change in a separate PR (since it touches multiple components).

adriansmares: That's fine with me, but please keep in mind that we're in the process of phasing out `gogoproto`, so it's actually ok to do such modifications, given that they are just renames.

pkg/applicationserver/applicationserver.go
```diff
@@ -1117,6 +1117,9 @@ func (as *ApplicationServer) handleUplink(ctx context.Context, ids ttnpb.EndDevi
 	} else {
 		uplink.Locations = isDev.GetLocations()
 	}
+	if dev.VersionIDs != nil {
+		uplink.VersionIDs = dev.VersionIDs
```
htdvisser, line 903: If this actually works, I think we have a bug somewhere, because on L1114 you don't actually request `version_ids`, so those should always be `nil`.

neoaggelos: The `version_ids` field is fetched from the AS device registry, not the IS.

api/identifiers.proto
htdvisser: I'm fine with keeping that out of scope for this PR.

api/messages.proto
htdvisser: I'm fine with keeping that out of scope for this PR.

pkg/applicationserver/applicationserver.go
htdvisser: Ah sorry, I looked in the wrong place.

htdvisser (approved):

adriansmares (approved):

PR #4125, opened 2021-05-04
Disable uplink simulation when skipping payload crypto

### LINKED ISSUE #3913: Disable uplink simulation and downlink message sending when skipping payload crypto
#### Summary

Disable uplink simulation and downlink message sending when skipping payload crypto

#### Why do we need this?

When simulating uplink, the simulated payload would still be encrypted. I don't think it's of great added value to support uplink simulation with encrypted FRMPayload.

When sending downlink, AS doesn't have a FCntDown so it always fails. Also, the same problem applies: it would already be encrypted.

The features are enabled even when end-to-end encryption (skip payload crypto) is enabled on either end device or application level.

Let's begin with returning errors from the API. We might add nice Console support too, but that's a bit goldplating as this is pretty much and edge case.

- For uplink simulation, when fetching device identifiers, also fetch `skip_payload_crypto` for the end device and the application. When enabled, return an error
- When sending downlink messages, this might be a little trickier, because we use the normal API that integrations also use to push/replace downlink. I'm not sure what's the best way to implement there: that might need to go via Console

Locally

Can review and test

asmulko:
#### Summary

Closes #3913 

#### Changes

- Disabled device uplink simulation and downlink message sending when skipping payload crypto.

**Before:**
simulate_before_1
simulate_before_2

**After:**
simulate_after_1
simulate_after_2

#### Testing

- Added e2e tests

#### Checklist

### DIFF
--- a/cypress/integration/console/devices/messaging/edit.spec.js
+++ b/cypress/integration/console/devices/messaging/edit.spec.js
@@ -87,6 +87,35 @@ describe('End device messaging', () => {
 
       cy.findByTestId('error-notification').should('not.exist')
     })
+
+    it('disables uplink simulation when skip payload crypto is enabled', () => {
+      cy.visit(
+        `${Cypress.config(
+          'consoleRootPath',
+        )}/applications/${applicationId}/devices/${endDeviceId}/messaging/uplink`,
+      )
+
+      const response = {
+        skip_payload_crypto_override: true,
+        session: {},
+      }
+
+      cy.intercept(
+        'GET',
+        `as/applications/${applicationId}/devices/${endDeviceId}?field_mask=version_ids,formatters,skip_payload_crypto,skip_payload_crypto_override,session,pending_session`,
+        response,
+      )
+
+      cy.findByTestId('notification')
+        .should('be.visible')
+        .findByText(`Simulation is disabled for devices that skip payload crypto`)
+        .should('be.visible')
+
+      cy.findByLabelText('FPort').should('be.disabled')
+      cy.findByLabelText('Payload').should('be.disabled')
+
+      cy.findByRole('button', { name: 'Simulate uplink' }).should('be.disabled')
+    })
   })
 
   describe('Downlink', () => {
@@ -106,6 +135,38 @@ describe('End device messaging', () => {
         .findByText(
           `Downlinks can only be scheduled for end devices with a valid session. Please make sure your end device is properly connected to the network.`,
         )
+        .should('be.visible')
+
+      cy.findByLabelText('Replace downlink queue').should('be.disabled')
+      cy.findByLabelText('Push to downlink queue (append)').should('be.disabled')
+      cy.findByLabelText('FPort').should('be.disabled')
+      cy.findByLabelText('Payload').should('be.disabled')
+      cy.findByLabelText('Confirmed downlink').should('be.disabled')
+
+      cy.findByRole('button', { name: 'Schedule downlink' }).should('be.disabled')
+    })
+
+    it('disables downlink messaging when skip payload crypto is enabled', () => {
+      cy.visit(
+        `${Cypress.config(
+          'consoleRootPath',
+        )}/applications/${applicationId}/devices/${endDeviceId}/messaging/downlink`,
+      )
+
+      const response = {
+        skip_payload_crypto_override: true,
+        session: {},
+      }
+
+      cy.intercept(
+        'GET',
+        `as/applications/${applicationId}/devices/${endDeviceId}?field_mask=version_ids,formatters,skip_payload_crypto,skip_payload_crypto_override,session,pending_session`,
+        response,
+      )
+
+      cy.findByTestId('notification')
+        .should('be.visible')
+        .findByText(`Simulation is disabled for devices that skip payload crypto`)
         .should('be.visible')
 
       cy.findByLabelText('Replace downlink queue').should('be.disabled')
--- a/pkg/webui/console/components/downlink-form/downlink-form.js
+++ b/pkg/webui/console/components/downlink-form/downlink-form.js
@@ -58,7 +58,7 @@ const validationSchema = Yup.object({
   ),
 })
 
-const DownlinkForm = ({ appId, devId, device, downlinkQueue }) => {
+const DownlinkForm = ({ appId, devId, device, downlinkQueue, skipPayloadCrypto }) => {
   const [error, setError] = useState('')
   const handleSubmit = useCallback(
     async (vals, { setSubmitting, resetForm }) => {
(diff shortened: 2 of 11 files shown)


cypress/integration/console/devices/messaging/edit.spec.js
```diff
@@ -116,5 +145,37 @@ describe('End device messaging', () => {
 
       cy.findByRole('button', { name: 'Schedule downlink' }).should('be.disabled')
     })
+
+    it('fails sending downlink message without when simulation is disabled', () => {
```
kschiffer:
```suggestion
    it('disables downlink messaging when skip payload crypto is enabeld', () => {
```

cypress/integration/console/devices/messaging/edit.spec.js
```diff
@@ -87,6 +87,35 @@ describe('End device messaging', () => {
 
       cy.findByTestId('error-notification').should('not.exist')
     })
+
+    it('fails sending uplink message without when simulation is disabled', () => {
```
kschiffer:
```suggestion
    it('disables uplink simulation when skip payload crypto is enabeld', () => {
```

pkg/webui/console/components/uplink-form/uplink-form.js
```diff
@@ -83,12 +84,16 @@ const UplinkForm = props => {
...
+      {device.skip_payload_crypto_override && (
+        <Notification content={sharedMessages.deviceSimulationDisabledWarning} warning small />
+      )}
       <IntlHelmet title={m.simulateUplink} />
       <Form
         error={error}
         initialValues={initialValues}
         validationSchema={validationSchema}
         onSubmit={handleSubmit}
+        disabled={device.skip_payload_crypto_override}
```
kschiffer: Is the `override` prop enough, or should this also check for the plain `skip_payload_crypto`?

pkg/webui/lib/shared-messages.js
```diff
@@ -121,6 +121,7 @@ export default defineMessages({
   deviceDescPlaceholder: 'Description for my new end device',
   deviceIdPlaceholder: 'my-new-device',
   deviceNamePlaceholder: 'My new end device',
+  deviceSimulationDisabledWarning: 'Simulation is disabled for devices that skip payload crypto.',
```
```suggestion
  deviceSimulationDisabledWarning: 'Simulation is disabled for devices that skip payload crypto',
```

[https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.13/DESIGN.md#general-considerations](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.13/DESIGN.md#general-considerations)

> *   Do not use end-of-sentence periods, unless the text consists of multiple sentences

kschiffer (changes requested):

pkg/webui/console/views/device/index.js
```diff
@@ -64,6 +69,8 @@ import style from './device.styl'
     const devId = props.match.params.devId
     const appId = selectSelectedApplicationId(state)
     const device = selectSelectedDevice(state)
+    const link = selectApplicationLink(state)
+    const applicationLinkfetching = selectApplicationLinkFetching(state)
```
kschiffer:
```suggestion
    const applicationLinkFetching = selectApplicationLinkFetching(state)
```

pkg/webui/console/views/device/index.js
```diff
@@ -147,6 +158,12 @@ export default class Device extends React.Component {
     env: undefined,
   }
 
+  componentDidMount() {
+    const { appId, getLink } = this.props
+
+    getLink(appId, ['skip_payload_crypto'])
+  }
+
```
kschiffer: Not necessary to do that here. You can add the dispatch to the `withRequest` HOC on LL94:

```
getDevice(appId, devId, selector, { ignoreNotFound: true })
getLink(appId, ['skip_payload_crypto'])
```

Or you can make a `loadDeviceData` prop in the redux connector that combines the actions. We do that e.g. in `/console/views/application/index.js LL79`.

pkg/webui/console/views/device/index.js
```diff
@@ -74,12 +81,14 @@ import style from './device.styl'
       maySendUplink: checkFromState(maySendUplink, state),
       fetching: selectDeviceFetching(state),
       error: selectDeviceError(state),
+      applicationLinkfetching: applicationLinkfetching || !link,
```
kschiffer: Just add this to the `fetching` prop on L82, the `withRequest` HOC will do the rest.

```javascript
fetching: selectDeviceFetching(state) || selectApplicationLinkFetching(state)
```

kschiffer: Also avoid adding other dependencies to fetching props (like whether a link is present here). The selector should provide everything necessary to determine whether something is fetched.

pkg/webui/console/components/downlink-form/downlink-form.js
```diff
@@ -88,17 +88,21 @@ const DownlinkForm = ({ appId, devId, device, downlinkQueue }) => {
   }
 
   const validSession = device.session || device.pending_session
+  const deviceSimulationDisabled = device.skip_payload_crypto_override || skipPayloadCrypto
```
kschiffer:
```suggestion
  const payloadCryptoSkipped = device.skip_payload_crypto_override || skipPayloadCrypto
  const deviceSimulationDisabled = !validSession || payloadCryptoSkipped
```

pkg/webui/console/components/downlink-form/downlink-form.js
```diff
@@ -140,6 +144,11 @@ DownlinkForm.propTypes = {
     push: PropTypes.func.isRequired,
     replace: PropTypes.func.isRequired,
   }).isRequired,
+  skipPayloadCrypto: PropTypes.bool,
+}
+
+DownlinkForm.defaultProps = {
+  skipPayloadCrypto: false,
 }
```
kschiffer: I'm not really happy with assuming `false` by default, since this value is coming from the store. If we determine the absence of the `skip_payload_crypto` prop as indicating `false` then that should be done on the selector level and the prop should be marked as required here instead.

pkg/webui/console/store/selectors/applications.js
```diff
@@ -90,3 +90,9 @@ export const selectApplicationLinkFormatters = state => {
 
   return link.default_formatters
 }
+
+export const selectApplicationLinkSkipPayloadCrypto = state => {
+  const link = selectApplicationLink(state) || {}
+
+  return link.skip_payload_crypto
```
kschiffer:
```suggestion
  return link.skip_payload_crypto || false
```

Even though I'm not in the matter enough to determine whether this is a desirable setting. 

@htdvisser if there is no link or the `skip_payload_crypto` prop missing in the AS response, can we assume that payload crypto is not skipped then?

kschiffer: Also, how can the user configure `skip_payload_crypto` on the application link level?

kschiffer (changes requested):

kschiffer (approved):

Issue #3642, opened 2021-01-05
Retry failed confirmed downlinks


rvolosatovs:
#### Summary

Retry failed confirmed downlinks

#### Why do we need this?

To ensure same behavior as v2 and make customers happy

#### What is already there? What do you see now?

No retry

#### What is missing? What do you want to see?

Retry mechanism for failed downlinks

#### How do you propose to implement this?

If ack is not received in uplink, prepend the downlink to the queue by default.

#### How do you propose to test this?

integration testing

#### Can you do this yourself and submit a Pull Request?

yes

virtualguy:
Any chance this is going to squeeze into an upcoming 3.12.x release? I somehow completely missed the difference in behaviour from v2 to v3 and having to replace and rebuild the entire queue when a DownlinkNack happens is a huge pain. The intention was to use a confirmed downlink to ensure message ordering

jpmeijers:
From what I see the current behaviour is to indefinitely retry confirmed downlinks, causing headaches for people with devices that do not acknowledge these downlinks. See https://github.com/TheThingsNetwork/lorawan-stack/issues/4289

PR #4126, opened 2021-05-04
Application Server session recovery improvements

### LINKED ISSUE #4067: Improve Application Server end device session recovery algorithm
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/3880
References https://github.com/TheThingsIndustries/lorawan-stack-support/issues/395

The AS should converge the `dev.Session` / `dev.PendingSession` while running downlink queue transactions, similar to how it already does this on `DownlinkQueuePush` / `DownlinkQueueReplace`.

#### Why do we need this?

Since the NS is allowed to increase the frame counter by scheduling downlinks while the AS still has to process the downlink queue invalidations / nacks,  the frame counters may be too low. As such, in places where we 'recalculate' the queue, we should expect that the NS can reject the queue, and deal with the error accordingly. 

`as.buildSessionsFromError`, which updates the `dev.Session` / `dev.PendingSession` based on the errors reported by the Network Server.

`as.buildSessionsFromError` should also be used on downlink queue transaction failures.

#### Environment

`v3.12`

This issue should have two targets:
- Simplify the whole double queue logic that we have in place. There are lots of assumptions there (what was the old session, what is the new session) that should be simplified.
- On downlink queue transaction errors, fix the sessions and re-run the transactions.

Class C uplink acks trigger this frequently. Let's test the changes using this particular use case (queue recalculation while ACKs are scheduled in class C).

Yes.

adriansmares:
#### Summary

Closes #4067 

#### Changes

- Broke down `as.downlinkQueueOp` as it was huge and did too many things at once:
  -  Observation logic has been moved to `registerReceiveDownlinks` / `as.registerDropDownlinks` / `as.registerForwardDownlinks`
  - The retry loop (that basically syncs the session between AS and NS) is now part of `as.attemptDownlinkQueueOp`
  - The downlink encode + encryption is now part of `as.encodeAndEncryptDownlinks`
- Flattened `as.handleJoinAccept` near the 'invalidate downlinks', as the nesting made it very hard to see what was happening.
- As part of the JoinAccept handling, the invalidated downlinks may be decrypted using what is called a 'ternary session'. A ternary session is simply the old `dev.PendingSession`. This occurs when the device attempted to join the network twice, but didn't confirm the session via an uplink.
- `as.matchSession` no longer recalculates the queue when it doesn't recognize the session (i.e. it's not `dev.Session`, nor `dev.PendingSession`). The assumption here is that the Network Server has thrown away the 'old' queue for what is probably a long time, as session confirmation swaps the pending session queue with the current queue. If the AS is in the position in which it never actually saw this session before, there is no way the queue in the Network Server is not empty (as there is no other entity that could have touched the session queue, or the pending session queue).
- Downlink queue invalidations are now handled via pushes, instead of replaces. The Network Server 'evicts' this downlink queue upon invalidation, so we have no reason to actually replace the queue. This protects us from cases during which we wrongfully replace valid items inserted by an application between the invalidation and the handling.
- Downlink nacks are now handled via pushes, in order to avoid replacing a valid queue with an old queue. I understand that this behavior is changed from what we did before, but I bring the following arguments:
  - The NS is already allowed to schedule the next downlinks in the queue - we're just racing with it if we attempt to put this item at the front of the queue.
  - When https://github.com/TheThingsNetwork/lorawan-stack/issues/3642 will be done, the NS will automatically retry the uplink 'at the front of the queue'. 
  - We generally lost this item often, as the `AFCntDown` was generally too low while handling the nack anyway.
  - The push is not destructive towards the queue.
- It should no longer be possible for the AS to fail downlink queue operations. Any downlink queue operations that we do now go via `as.attemptDownlinkQueueOp`, which will automatically sync the session between the AS and the NS.

#### Testing

I've tested this locally with a 1.0.3A device in class A without any issues. My queue was carried across rejoins (something which was unstable before), and I didn't encounter at any points dropped items.

~Nonetheless this requires better class C testing, as class C was generally the 'catalyst' for these issues. I will pick this up soon, but the work may already be reviewed.~

##### Regressions

This PR changes the logic used by the following situations:
- The queue changes on `JoinAccept`
- The queue changes on `DownlinkQueueInvalidated`
- The queue changes on `DownlinkNack`

All of these have been tested in class A, ~but still require class C testing~.

Edit: Class C is done and looks promising.

#### Notes for Reviewers

@johanstokking is the main reviewer.

I consider this PR to be close to water tight. But please poke holes if you see any.

#### Checklist

### DIFF
--- a/pkg/applicationserver/applicationserver.go
+++ b/pkg/applicationserver/applicationserver.go
@@ -36,7 +36,6 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/cluster"
 	"go.thethings.network/lorawan-stack/v3/pkg/component"
 	"go.thethings.network/lorawan-stack/v3/pkg/config"
-	"go.thethings.network/lorawan-stack/v3/pkg/crypto"
 	"go.thethings.network/lorawan-stack/v3/pkg/crypto/cryptoutil"
 	"go.thethings.network/lorawan-stack/v3/pkg/errors"
 	"go.thethings.network/lorawan-stack/v3/pkg/events"
@@ -450,23 +449,55 @@ func (as *ApplicationServer) buildSessionsFromError(ctx context.Context, dev *tt
 
 const maxDownlinkQueueOperationAttempts = 50
 
-func (as *ApplicationServer) downlinkQueueOp(ctx context.Context, ids ttnpb.EndDeviceIdentifiers, items []*ttnpb.ApplicationDownlink, op func(ttnpb.AsNsClient, context.Context, *ttnpb.DownlinkQueueRequest, ...grpc.CallOption) (*pbtypes.Empty, error)) error {
-	ctx = events.ContextWithCorrelationID(ctx, fmt.Sprintf("as:downlink:%s", events.NewCorrelationID()))
-	for _, item := range items {
-		item.CorrelationIDs = append(item.CorrelationIDs, events.CorrelationIDsFromContext(ctx)...)
-	}
-	logger := log.FromContext(ctx)
-	for _, item := range items {
-		registerReceiveDownlink(ctx, ids, item)
-	}
-	peer, err := as.GetPeer(ctx, ttnpb.ClusterRole_NETWORK_SERVER, &ids)
-	if err != nil {
-		return err
-	}
+func (as *ApplicationServer) attemptDownlinkQueueOp(ctx context.Context, ids ttnpb.EndDeviceIdentifiers, dev *ttnpb.EndDevice, items []*ttnpb.ApplicationDownlink, op func(ttnpb.AsNsClient, context.Context, *ttnpb.DownlinkQueueRequest, ...grpc.CallOption) (*pbtypes.Empty, error), link *ttnpb.ApplicationLink, peer cluster.Peer) ([]string, error) {
 	pc, err := peer.Conn()
 	if err != nil {
-		return err
+		return nil, err
 	}
+	mask := make([]string, 0, 2)
+	attempt := 0
+	for {
+		ctx := log.NewContextWithField(ctx, "attempt", attempt)
+
+		sessions := make([]*ttnpb.Session, 0, 2)
+		if dev.Session != nil {
+			sessions = append(sessions, dev.Session)
+			mask = append(mask, "session.last_a_f_cnt_down")
+		}
+		if dev.PendingSession != nil {
+			// Downlink can be encrypted with the pending session while the device first joined but not confirmed the
+			// session by sending an uplink.
+			sessions = append(sessions, dev.PendingSession)
+			mask = append(mask, "pending_session.last_a_f_cnt_down")
+		}
+		if len(sessions) == 0 {
+			return nil, errNoDeviceSession.New()
+		}
+
+		encryptedItems, err := as.encodeAndEncryptDownlinks(ctx, dev, link, items, sessions)
+		if err != nil {
+			return nil, err
+		}
+		_, err = op(ttnpb.NewAsNsClient(pc), ctx, &ttnpb.DownlinkQueueRequest{
+			EndDeviceIdentifiers: ids,
+			Downlinks:            encryptedItems,
+		}, as.WithClusterAuth())
+		if err == nil {
+			return mask, nil
+		}
+		if attempt >= maxDownlinkQueueOperationAttempts || as.skipPayloadCrypto(ctx, link, dev, nil) {
+			return nil, err
+		}
+		mask, err = as.buildSessionsFromError(ctx, dev, err)
+		if err != nil {
+			return nil, err
+		}
+		attempt++
+	}
+}
+
+func (as *ApplicationServer) downlinkQueueOp(ctx context.Context, ids ttnpb.EndDeviceIdentifiers, items []*ttnpb.ApplicationDownlink, op func(ttnpb.AsNsClient, context.Context, *ttnpb.DownlinkQueueRequest, ...grpc.CallOption) (*pbtypes.Empty, error)) error {
+	ctx = events.ContextWithCorrelationID(ctx, fmt.Sprintf("as:downlink:%s", events.NewCorrelationID()))
 	link, err := as.getLink(ctx, ids.ApplicationIdentifiers, []string{
 		"default_formatters",
 		"skip_payload_crypto",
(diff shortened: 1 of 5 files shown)


pkg/applicationserver/applicationserver.go
```diff
@@ -450,30 +449,70 @@ func (as *ApplicationServer) buildSessionsFromError(ctx context.Context, dev *tt
...
+			return nil, err
+		}
+		_, err = op(ttnpb.NewAsNsClient(pc), ctx, &ttnpb.DownlinkQueueRequest{
+			EndDeviceIdentifiers: ids,
+			Downlinks:            encryptedItems,
+		}, as.WithClusterAuth())
+		if err == nil {
+			return mask, nil
+		}
+		if attempt >= maxDownlinkQueueOperationAttempts || as.skipPayloadCrypto(ctx, link, dev, nil) {
```
johanstokking, line 503: First attempt has `attempt == 0`, while `maxDownlinkQueueOperationAttempts` is a counter. So having that value set to `1` results in two attempts.

pkg/applicationserver/applicationserver.go
```diff
@@ -450,30 +449,70 @@ func (as *ApplicationServer) buildSessionsFromError(ctx context.Context, dev *tt
...
-		return err
-	}
+func (as *ApplicationServer) attemptDownlinkQueueOp(ctx context.Context, ids ttnpb.EndDeviceIdentifiers, dev *ttnpb.EndDevice, items []*ttnpb.ApplicationDownlink, op func(ttnpb.AsNsClient, context.Context, *ttnpb.DownlinkQueueRequest, ...grpc.CallOption) (*pbtypes.Empty, error), link *ttnpb.ApplicationLink, peer cluster.Peer) ([]string, error) {
 	pc, err := peer.Conn()
 	if err != nil {
-		return err
+		return nil, err
 	}
+	mask := make([]string, 0, 2)
+	attempt := 0
```
johanstokking:
```suggestion
	attempt := 1
```

pkg/applicationserver/applicationserver.go
```diff
@@ -780,255 +728,65 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids ttnpb.End
...
-				joinAccept.AppSKey = nil
+				items = append(items, msg)
+			}
+			joinAccept.InvalidatedDownlinks = nil
+			if len(items) == 0 {
+				return dev, mask, nil
 			}
+
+			if session := dev.Session; session != nil {
+				session.LastAFCntDown = items[0].FCnt - 1
```
johanstokking: Why is this? I thought that `LastAFCntDown` would be the last used `FCntDown`, i.e. `items[len(items)-1].FCnt`.

BTW, are we sure `FCnt` is never `0`?

pkg/applicationserver/applicationserver.go
```diff
@@ -1171,94 +902,93 @@ func (as *ApplicationServer) handleDownlinkQueueInvalidated(ctx context.Context,
...
+			if err != nil {
+				return nil, nil, err
+			}
+
+			// Decrypt the message as it will be sent to upstream after handling it.
+			if err := as.decryptDownlinkMessage(ctx, ids, msg, link); err != nil {
+				return nil, nil, err
+			}
+
+			items := []*ttnpb.ApplicationDownlink{msg}
```
johanstokking, line 1002: Should we already bump the FCnt here?

johanstokking (commented):
Looks great. Love all the red blocks of code

pkg/applicationserver/applicationserver.go
adriansmares: I had a long comment here, which I will leave out for historical reasons, but the TLDR is probably that it would be better to just push the uplinks only to the pending session, if possible. I'll check out if I can do it cleanly.

---

There are two mechanisms here, which actually probably should end up as godoc.

1. We call them invalidated downlinks, but the queue itself has not been cleaned for the `dev.Session` -  we just receive a copy. See https://github.com/TheThingsNetwork/lorawan-stack/blob/c0e437afc7fccb9d506e05f73b4a2d2c3db54d13/pkg/networkserver/downlink.go#L1667-L1690

2. `as.AttempDownlinkQueueOp` will replace both queues (as you cannot replace only one queue). As such, it has to compute the FCnt of each uplink again. If we don't lower this `FCnt` before calling it, each time a device joins, the `FCnt` of each element will increase by the queue size. 

Take this example:
- You schedule 3 downlinks to your device, with FCnts 42, 43, 44. `LastAFCntDown` is 44.
- Your device rejoins. We receive these downlinks in `InvalidatedDownlinks`. 
- We have to do a `replace` operation in order to create the queue for the pending session.
- When we calculate the downlinks for `dev.Session`, we start from 44, so we'll have 45, 46, 47.

If your device rejoins too many times at some point we will run out of FCnts. As such, we attempt to keep the `LastAFCntDown` steady.

pkg/applicationserver/applicationserver.go
adriansmares: `attemptDownlinkQueueOp` will bump it automatically.

adriansmares:
I've refactored the `JoinAccept` procedure to be replace-free, doing a push on the pending session instead. I've also fixed an NS error which didn't return the session diagnostics. This fixes a specific scenario during which downlinks were lost when the device joined after a confirmed downlink was scheduled.

LE: Class C tests are also pretty good. I no longer observe downlinks being dropped spuriously while mixing confirmed and unconfirmed downlinks.

pkg/applicationserver/applicationserver.go
johanstokking: > * You schedule 3 downlinks to your device, with FCnts 42, 43, 44. `LastAFCntDown` is 44.
> [...]
> * When we calculate the downlinks for `dev.Session`, we start from 44, so we'll have 45, 46, 47.
> 
> If your device rejoins too many times at some point we will run out of FCnts. As such, we attempt to keep the `LastAFCntDown` steady.

Wait, the `LastAFCntDown` survives a session? That doesn't work with rollovers in the frame.

adriansmares: No - the new `dev.PendingSession` starts with `LastAFCntDown=0`, and then we push the old downlink queue to this pending session. So if you had `n` downlinks in `dev.Session`, your `dev.PendingSession` will contain `n` downlinks, with framecounters between `1` and `n`, and `LastAFCntDown=n`.

The behavior that I described where the counter keeps increasing for `dev.Session` is no longer applicable in any shape or form, after 33c6276310b003a1a18f9c63706e27978451e976. Now on `JoinAccept` we don't even touch the downlink queue of `dev.Session`, we just create a new one for `dev.PendingSession`, for which the counter starts at 0.

johanstokking (approved):

PR #4128, opened 2021-05-05
Add MACSettings.DesiredMaxEIRP field

### LINKED ISSUE #4090: Allow configuring the MaxEIRP as part of device MAC settings
#### Summary

Allow configuring MaxEIRP for devices as part of the MAC settings

#### Why do we need this?

Currently, MaxEIRP can only be set in `MACState.DesiredParameters`. This is OK for ABP devices, but impossible for OTAA devices, since MAC state can only be configured if the device already has a session.

MaxEIRP is set in `MACState.DesiredParameters`. When initializing the MAC state of a device, the Network Server will use the default MaxEIRP setting from the respective band (https://github.com/TheThingsNetwork/lorawan-stack/blob/0e95225c3688a6d42e4f3bd2463f38258863bb91/pkg/networkserver/mac/utils.go#L221-L228).

This is also related to the Device Repository, where `maxEIRP` can be configured per device profile.

Ability to configure MaxEIRP for OTAA devices.

#### Environment

3.12.1

Introduce `MACSettings.MaxEIRP`. Network Server respects that value (if set) when initializing a new MAC state.

Unit tests, creating devices from the device repository

Yes, but I am not sure if this is the correct way to do it. Adding to 3.12.2 because this needs to be addressed along with #4086

neoaggelos:
#### Summary

Closes #4090 

#### Changes

- Introduce `DeviceEIRPValue` struct (containing an enum value of type `DeviceEIRP`)
- Introduce `MACSettings.DesiredMaxEirp` field.
- Consider MaxEirp value for `mac.DeviceDesiredMaxEIRP`
- Miscellaneous changes required for validations and setting the DesiredMaxEirp field from the CLI.

#### Testing

- Unit tests
- Setting `desired_max_eirp` field from the CLI and verify that the `DesiredParameters.MaxEIRP` value is set properly after device joins.

##### Regressions

There should not be any, the default value for the new setting is `nil` which preserves the old behaviour.
 
#### Notes for Reviewers

- Not using gogoproto, so the field is `DesiredMaxEirp` instead of `DesiredMaxEIRP`.
- The validation and `FieldIsZero()` parts probably need a review from @rvolosatovs to make sure that I did not miss anything.

#### Checklist


johanstokking (approved):
LGTM

pkg/ttnpb/testdata/ttnpb_encoding_golden.md
```diff
@@ -805,6 +805,22 @@ JSONPB | ttnpb.DataRateOffsetValue | {4 {} 0} | 4
...
+JSONPB | ttnpb.DeviceEIRP | DEVICE_EIRP_20 | "DEVICE_EIRP_20"
+JSONPB | ttnpb.DeviceEIRP | DEVICE_EIRP_21 | "DEVICE_EIRP_21"
+JSONPB | ttnpb.DeviceEIRP | DEVICE_EIRP_24 | "DEVICE_EIRP_24"
+JSONPB | ttnpb.DeviceEIRP | DEVICE_EIRP_26 | "DEVICE_EIRP_26"
+JSONPB | ttnpb.DeviceEIRP | DEVICE_EIRP_27 | "DEVICE_EIRP_27"
+JSONPB | ttnpb.DeviceEIRP | DEVICE_EIRP_29 | "DEVICE_EIRP_29"
+JSONPB | ttnpb.DeviceEIRP | DEVICE_EIRP_30 | "DEVICE_EIRP_30"
+JSONPB | ttnpb.DeviceEIRP | DEVICE_EIRP_33 | "DEVICE_EIRP_33"
+JSONPB | ttnpb.DeviceEIRP | DEVICE_EIRP_36 | "DEVICE_EIRP_36"
+JSONPB | ttnpb.DeviceEIRP | DEVICE_EIRP_8 | "DEVICE_EIRP_8"
```
htdvisser, line 855: I would also expect `ttnpb.DeviceEIRPValue` lines here

htdvisser (commented):
I really don't know much about how these custom wrappers work, so maybe good to wait for review from @rvolosatovs.

adriansmares (approved):
Nothing to add.

rvolosatovs (approved):

Issue #4053, opened 2021-04-13
Reduce payload formatter script size


johanstokking:
#### Summary

Replaces https://github.com/TheThingsIndustries/lorawan-stack/issues/2675

Reduce the maximum script size for payload formats to 4 KB.

#### Why do we need this?

Because it takes too much memory otherwise.

We will allow larger scripts via the Device Repository. We have optimizations for that planned, see https://github.com/TheThingsIndustries/lorawan-stack/issues/2264.

#### What is already there? What do you see now?

No limits. We already see scripts larger than 200 KB.

#### What is missing? What do you want to see?

Limits

#### How do you propose to implement this?

This applies to user defined scripts on the application and device level. This does not affect the message processor parameter field (for now), as we still allow larger values from device repository.

Setting this to a new minor as this is a somewhat breaking behavioral API change for users. Some scripts are for example already stored, and they cannot be updated or changed, and new devices and applications cannot be assigned a script that existing devices or applications are using.

#### How do you propose to test this?

Local testing

#### Can you do this yourself and submit a Pull Request?

Can review

descartes:
On v2 when the servers were under extreme pressure the JS decoder would time out so the HTTP Integration, MQTT & Data Storage only had the raw payload - didn't bother me, I use the JS decoder for debugging only - but it did tend to cause a stir for the TTN users.

It would be foolish to say that this could never happen on v3, and 4KB is more than plenty of code which will keep processing times down, but is it worth having a status field so that if decoding does time out or otherwise fail, that it's clear in the relayed JSON that decoded_payload is going to be empty?

johanstokking:
@descartes good suggestion to include this somehow in the payload, potentially with the already existing `errors` field of `decoded_payload`.

We actually see V3 also already failing on large scripts. We measure the 100 ms execution timeout to include the evaluation time of the script, so with large scripts and CPU pressure, that fails just like in V2. Although we use a different JavaScript engine, this is a very similar design in V3.

PR #4133, opened 2021-05-05
Limit maximum size for payload formatter scripts

neoaggelos:
#### Summary

References #4053 

#### Changes

- Add `as.formatters.max-script-size` configuration option for maximum allowed size of formatter paramaters (defaults to 4KB).
- Return an error when setting payload formatter scripts on devices or applications with size that exceeds the maximum based on the configuration.

#### Testing

- Unit tests
- Local testing with CLI and Console.

#### Notes for Reviewers

This may cause confusion to people using formatters if "things no longer work". Added a lengthy CHANGELOG entry and a link to the relevant issue on GitHub for this.

#### Checklist


CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Added
 
+- Maximum size for user-defined payload formatter scripts.
```
johanstokking, line 32: This should be under **Changed**

cmd/internal/shared/applicationserver/config.go
```diff
@@ -76,4 +76,7 @@ var DefaultApplicationServerConfig = applicationserver.Config{
 			Workers: 16,
 		},
 	},
+	Formatters: applicationserver.FormattersConfig{
+		MaxScriptSize: 4096,
```
johanstokking: I think this should be `MaxParameterLength` to be consistent with the naming in AS

johanstokking (commented):
Isn't it feasible to add this to API validation here? 

https://github.com/TheThingsNetwork/lorawan-stack/blob/f396beccdb93fa1af80c4539d9fff4e0a1e1a379/api/messages.proto#L375-L380

The validator only kick in on incoming requests: that is, when setting the value.

pkg/applicationserver/config.go
```diff
@@ -66,12 +66,17 @@ type EndDeviceFetcherCircuitBreakerConfig struct {
 	Threshold int           `name:"threshold" description:"Number of failed fetching attempts after which the circuit breaker opens"`
 }
 
+type FormattersConfig struct {
+	MaxScriptSize int `name:"max-script-size" description:"Maximum allowed size for payload formatter scripts"`
```
htdvisser: While we're at it, would this also be the moment to make the script timeout configurable?

htdvisser (commented):
Not much to add

neoaggelos:
> Isn't it feasible to add this to API validation here?

The downside to this is that it is then hardcoded and impossible to override or change (even for custom deployments where might be OK to use larger payload formatters). Is this something that we actually want? If not, then let's just set validation rules in the API.

adriansmares:
> 
> 
> > Isn't it feasible to add this to API validation here?
> 
> The downside to this is that it is then hardcoded and impossible to override or change (even for custom deployments where might be OK to use larger payload formatters). Is this something that we actually want? If not, then let's just set validation rules in the API.

It's not entirely exclusive. For downlink queues for example, it's both: API has a limit 10000, while the configuration is 1000.

So maybe something like 16KB at API level, and then the configuration may lower it to 1KB. As @htdvisser said, the script timeout is also another important variable here - you could set the limit to 99999999999, but the parser wouldn't finish before the timeout.

johanstokking:
@neoaggelos please request review if ready for review.

neoaggelos:
> So maybe something like 16KB at API level, and then the configuration may lower it to 1KB.

Added the 16KB limit at API level as well, but I think this could become a source of confusion for users.

api/messages.proto
```diff
@@ -373,11 +373,11 @@ message MessagePayloadFormatters {
   // Payload formatter for uplink messages, must be set together with its parameter.
   PayloadFormatter up_formatter = 1 [(validate.rules).enum.defined_only = true];
   // Parameter for the up_formatter, must be set together.
```
htdvisser: I agree that it may be confusing, so perhaps just update the comment to say that in addition to this 16kB limit, deployment configuration may further restrict the size of these parameters (as well as the average execution duration of a formatter)

johanstokking (approved):
Nothing to add

adriansmares (approved):

phaabe:
Does this interfere with https://github.com/TheThingsNetwork/lorawan-devices?

johanstokking:
> Does this interfere with [TheThingsNetwork/lorawan-devices](https://github.com/TheThingsNetwork/lorawan-devices)?

No, for the Device Repository we (still) allow larger scripts.

The reason why we reduce the size here, is that we want to optimize storage in the Application Server. The Application Server stores the payload formatters potentially per end device in a memory store (Redis) in base64, so allowing 100 KB scripts isn't scalable.

htdvisser (approved):

PR #4146, opened 2021-05-07
Generate release notes from CHANGELOG

htdvisser:
#### Summary

With this PR we should no longer have to edit the release notes on GitHub after a release. Instead, the release workflow will automatically select the right section from the CHANGELOG and use that.

#### Testing

Not enough, because this only runs when we tag a release 😒 

- We know that the "Get version from tag" script works, because we use it in other workflows
- I tested the `awk` command locally

##### Regressions

- If it errors, we can just fix it and re-tag.
- If it has incorrect output, we can still manually edit the release notes on GitHub.

#### Checklist

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -4,7 +4,6 @@ on:
   push:
     tags:
       - 'v3.[0-9]+.[0-9]+'
-      - 'v3.[0-9]+.[0-9]+-rc[0-9]'
 
 env:
   NODE_ENV: production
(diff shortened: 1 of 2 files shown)


.github/workflows/release-tag.yml
```diff
@@ -72,15 +72,22 @@ jobs:
...
+      id: version
+      uses: actions/github-script@v3
       with:
-        result-encoding: string
         script: |
-          const tag = context.ref.slice('refs/tags/v'.length);
-          const tagParts = tag.split('.');
-          require('fs').writeFileSync('/tmp/release-notes.md', `[Release notes](https://github.com/TheThingsNetwork/lorawan-stack/blob/${tag}/CHANGELOG.md#${tagParts[0]}${tagParts[1]}${tagParts[2]}---${{ steps.date.outputs.value }})`);
+          const tag = context.ref.slice("refs/tags/".length);
+          const majorMinorPatchRegex = /^v([0-9]+)\.([0-9]+)\.([0-9]+)$/;
```
htdvisser, line 80: This doesn't actually work for `-rc` tags, but for those we already had incorrect results.

Perhaps we should just stop running this workflow on `-rc` tags?

johanstokking: What means doesn't work, failure or wrong output?

I'm fine with running this workflow only on non-beta/rc/etc releases.

If we need a better way to distribute snapshots, we can do that, but not here.

johanstokking (approved):

.github/workflows/release-tag.yml
htdvisser: Ok, updated the tag filter so that it doesn't run.

PR #3986, opened 2021-03-26
Add arm64 Docker images

htdvisser:
#### Summary

This pull request adds linux/arm64 Docker images to snapshot builds.

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/510
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/112

#### Notes for Reviewers

I think it's best to not change the actual release builds just yet, and first do some testing with the snapshots. If that's okay with you, I'll create an for copying this to the release tag workflows.

The `docker_manifests` config of GoReleaser doesn't work on snapshot releases, but we may be able to use that in tag workflows.

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #4147, opened 2021-05-07
Release multi-arch Docker images

htdvisser:
#### Summary

This updates the release workflow to create multi-arch Docker images including amd64, arm64 (that was already added to snapshots in #3986) and armv7 (for Raspberry Pi's).

#### Testing

Most of this was already tested with the snapshots (#3986). The `docker_manifests` is new, and we don't really have a nice way of testing this without tagging a release.

##### Regressions

If this errors, we'd need to fix it and re-tag the release, which isn't ideal, but won't break production either.

#### Checklist

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -19,6 +19,10 @@ jobs:
       with:
         fetch-depth: 0
         submodules: true
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
     - name: Login to Docker Hub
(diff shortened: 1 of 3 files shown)


johanstokking (commented):
Do we need to append the architecture to the tag?

Docker supports multiple architectures with the same tag, right? See https://hub.docker.com/_/alpine?tab=tags&page=1&ordering=last_updated

htdvisser:
It would indeed be nicer if we could only push the multi-arch image, but [apparently that's not possible with GoReleaser](https://goreleaser.com/customization/docker_manifest/#how-it-works).

johanstokking (approved):

Issue #2231, opened 2020-03-25
Event stream filters in console


kschiffer:
#### Summary
We should provide filtering features for our data view / event stream.

#### Why do we need this?
- Closing v2 feature gap
- Supporting advance use cases for data monitoring
- Mitigate event flooding

#### What is already there? What do you see now?
Event streaming for all our entities.

#### What is missing? What do you want to see?
Filtering options to improve utility of the event stream

I've done a wireframe about this a while ago:
![Event filters](https://user-images.githubusercontent.com/5710611/49792655-95039700-fd33-11e8-8ac9-b2c55ef0904a.png)
Some related discussion can be viewed here: https://github.com/TheThingsIndustries/lorawan-stack/issues/1071

We should discuss which forms of filters are viable.

#### Environment
v3.7.0

#### How do you propose to implement this?
I think we can make use of a special store selector `selectFilteredEvents(filter)` which selects event based on a defined filter object.

#### Can you do this yourself and submit a Pull Request?
Anyone please feel free to pickup.

kschiffer:
With regards to [https://github.com/TheThingsNetwork/lorawan-stack/issues/3713#issuecomment-779297814,](https://github.com/TheThingsNetwork/lorawan-stack/issues/3713#issuecomment-779297814,) we need to expand the functionality here a bit. There are scenarios where applications have hundreds or thousands of end-devices. This will crash the Console altogether when the event view is open. Hence, it is necessary to enforce some filter for heavy-load situations.

Such filters could be:

*   Limit the entity scope for application event streams to max 20 (or so) end devices
*   Only show the `forward` event type for uplinks downlinks and similar outfiltering of auxiliary event

We could use different levels of enforced filtering and use it for some predefined "events per second"-thresholds.

Ideas welcome cc @bafonins @htdvisser @johanstokking

htdvisser:
The number of registered end devices isn't necessarily an indication of the amount of traffic, so I don't think it would be a good idea to do different things depending on the number of registered end devices.

My suggestion would be to extend the API to allow filtering on event name. Every events view in the Console would then give the user some UI where they can check/uncheck events they're interested in. The event names that are selected by default can then be adapted to show the most relevant events for a given entity. 

For the Application overview, we can then filter on `application.**` and `as.**`, so that we only see application layer events, not all Network Server MAC events (for those, you'd have to open a specific end device)

kschiffer:
Blocked on #3817

[...]

PR #4104, opened 2021-04-29
Update UI related dependencies

asmulko:
#### Summary

References #3934

#### Changes
- Updated UI related dependencies

#### Testing

- Manually tested

#### Notes for Reviewers

Based on #3787

Cherry-picked commits:
- 1d9e0d83648fdc4abfb53aa9dd6f4d8de9f0eb61
- 6ac01fa475e5ed978e7571f5bae79cf568c29e9c
- 466863f13b6731dbbb5fea794d1dbabe520d5252

#### Checklist
- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #4150, opened 2021-05-10
Filter out verbose events in the Console

kschiffer:
#### Summary

This PR will apply a toggle-able verbosity filter to the event stream which will by default filter out all events that are not typically of direct interest.

References #2231

With verbose mode deactivated (default) only the following events will be ingested:

```javascript
export const EVENT_VERBOSE_FILTERS = [
  'as.*.drop',
  'as.down.data.forward',
  'as.up.data.forward',
  'gs.down.send',
  'gs.gateway.connect',
  'gs.gateway.disconnect',
  'gs.status.receive',
  'gs.up.receive',
  'js.join.accept',
  'js.join.reject',
  'ns.mac.*.answer.reject',
  '*.warning',
  '*.fail',
  'organization.*',
  'user.*',
  'gateway.*',
  'application.*',
  'end_device.*',
  'client.*',
  'oauth.*',
]
```

#### Changes

*   Add a constant module for event filters
*   Update store logic for applying filters to event streams
*   Update the `<Events />` component to enable filtering out verbose events
*   Update entity event containers accordingly

#### Testing

Manual testing in the staging environment

##### Regressions

This could affect event ingestion in the Console

#### Notes for Reviewers

*   I did not add the filter for organization events, since the filter would have no effect. For completeness I added the redux primitives for organizations though.
*   I have implemented this filter at the ingestion level within the reducer, meaning that filtered events will not be added to the store. This means that historical verbose events can not be toggled, since they were already discarded. I've decided to go this route to also lift some weight on the whole event store logic (see also #2887). With this implementation it will be possible to examine events for entities with a higher event frequency.
    *   Later on, other filter could be applied on top of this but filtered out on the component level so that the visibility can be toggled for past events as well.

#### Checklist

*   [x] Scope: The referenced issue is addressed, there are no unrelated changes.
*   [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
*   [x] Documentation: Relevant documentation is added or updated.
*   [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
*   [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.


bafonins (commented):
So, is this a temporary solution for https://github.com/TheThingsNetwork/lorawan-stack/issues/2231? What is the plan there?

I experience significant delay when opening event details for both development and production builds. This does not happen to me on the latest `v3.12` branch as well as on staging Console

Also, Im getting the following warning when opening the events tab:
```
Could not load worker TypeError: name.split is not a function
    at AppConfig.exports.moduleUrl (ace.js:3722)
    at WorkerClient.$createWorkerFromOldConfig (ace.js:19056)
    at new WorkerClient (ace.js:19034)
    at Mode.createWorker (json.js?818b:300)
    at EditSession.$startWorker (ace.js:9735)
    at EditSession.$onChangeMode (ace.js:9691)
    at EditSession.setMode (ace.js:9655)
    at ReactAce.libs_0dc51107365406cd4f90../node_modules/react-ace/lib/ace.js.ReactAce.componentDidMount (ace.js:73)
    at ReactAce.eval (react-hot-loader.development.js?c2cb:707)
    at commitLifeCycles (react-dom.development.js?7f13:20664) 
    at ReactAce (eval at ES6ProxyComponentFactory (webpack-internal:///./node_modules/react-hot-loader/dist/react-hot-loader.development.js), <anonymous>:14:7)
    at div
    at CodeEditor (eval at ES6ProxyComponentFactory (webpack-internal:///./node_modules/react-hot-loader/dist/react-hot-loader.development.js), <anonymous>:14:7)
    at eval (webpack-internal:///./pkg/webui/console/components/events/details/raw.js:168:24)
    at div
    at EventDetails (webpack-internal:///./pkg/webui/console/components/events/details/index.js:306:24)
```

kschiffer:
> So, is this a temporary solution for [#2231](https://github.com/TheThingsNetwork/lorawan-stack/issues/2231)? What is the plan there?

Yes this is a temporary fix for the event stream being too verbose. This is to improve the UX already until the new event UX is designed and implemented.

> I experience significant delay when opening event details for both development and production builds. This does not happen to me on the latest `v3.12` branch as well as on staging Console
> 
> Also, Im getting the following warning when opening the events tab:

I think this was introduced via #4104, see also securingsincity/react-ace#732

I downgraded `react-ace` back. Can you see if this resolves the issue? I didn't have a big delay either way but feel like the downgraded version is quicker indeed.

bafonins (approved):
Indeed, downgrading `react-ace` solves perf issues

PR #4154, opened 2021-05-11
Fix handling of Zero EUI gateway connections

KrishnaIyer:
#### Summary

Fix handling of Zero EUI gateway connections

#### Changes

- Reject connections with Zero EUI. `ids.IsZero()` only checks if EUI is nil and not if the value is zero.

#### Testing

Didn't test but I think this is simple enough.

##### Regressions

Don't expect any.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -384,7 +384,7 @@ var (
 // This method should only be used for request contexts.
 func (gs *GatewayServer) FillGatewayContext(ctx context.Context, ids ttnpb.GatewayIdentifiers) (context.Context, ttnpb.GatewayIdentifiers, error) {
 	ctx = gs.FillContext(ctx)
-	if ids.IsZero() || ids.EUI.IsZero() {
+	if ids.IsZero() || ids.EUI != nil && ids.EUI.IsZero() {
 		return nil, ttnpb.GatewayIdentifiers{}, errEmptyIdentifiers.New()
 	}
 	if ids.GatewayId == "" {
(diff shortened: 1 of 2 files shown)


pkg/gatewayserver/gatewayserver.go
```diff
@@ -384,7 +384,7 @@ var (
 // This method should only be used for request contexts.
 func (gs *GatewayServer) FillGatewayContext(ctx context.Context, ids ttnpb.GatewayIdentifiers) (context.Context, ttnpb.GatewayIdentifiers, error) {
 	ctx = gs.FillContext(ctx)
-	if ids.IsZero() {
+	if ids.IsZero() || ids.EUI.IsZero() {
```
adriansmares: If `ids.IsZero` is false, `ids.GatewayId != "" || ids.EUI != nil`, so `ids.EUI == nil` can happen if `ids.GatewayId != ""`. So you may `panic`.

KrishnaIyer: Yeah saw that and fixed it..:)

KrishnaIyer: Yeah saw that and fixed it..:)

adriansmares (approved):

johanstokking (approved):

pkg/gatewayserver/gatewayserver.go
```diff
@@ -384,7 +384,7 @@ var (
 // This method should only be used for request contexts.
 func (gs *GatewayServer) FillGatewayContext(ctx context.Context, ids ttnpb.GatewayIdentifiers) (context.Context, ttnpb.GatewayIdentifiers, error) {
 	ctx = gs.FillContext(ctx)
-	if ids.IsZero() {
+	if ids.IsZero() || ids.EUI != nil && ids.EUI.IsZero() {
```
htdvisser, line 387: In the future I would really prefer using parentheses, even in such simple logic expressions. Without them, it's too easy to make mistakes, and too difficult to spot such mistakes while debugging during an incident. We want to optimize for readability, and less experienced (open source) contributors also need to understand, without a doubt, what's going on.

KrishnaIyer: I use parenthesis instinctively, even while making this change. In past PRs, @johanstokking has asked me to remove parentheses since it's `C` like so I removed it here to avoid comments.

PR #4156, opened 2021-05-11
Lower log level of missing class A downlink

adriansmares:
#### Summary

This lowers the log level of the `Failed to generate class A downlink, skip class A downlink slot` error to warning.

I think this occurs when the `ADRAckReq` / uplink acks are sent via the network initiated downlink slot (class B / class C), while the class A downlink slot is also available. 

Both tasks can exist concurrently, but if the class B / C tasks runs before the class A task, the class A task may have no work to do when it actually runs.

### DIFF
--- a/pkg/networkserver/downlink.go
+++ b/pkg/networkserver/downlink.go
@@ -1178,7 +1178,7 @@ func (ns *NetworkServer) attemptClassADataDownlink(ctx context.Context, dev *ttn
 		)
 	}
 	if err != nil {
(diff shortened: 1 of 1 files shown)


johanstokking:
> I think this occurs when the `ADRAckReq` / uplink acks are sent via the network initiated downlink slot (class B / class C), while the class A downlink slot is also available.

Moving forward, I don't think we should set `ADRAckReq` acknowledgments in non-class-A downlink.

johanstokking (approved):

PR #4164, opened 2021-05-12
Make GS events more useful

htdvisser:
#### Summary

This PR is a proposal for making the events from the GS more useful.

#### Changes

1. Remove the `gs.status.forward` event. This event was triggered multiple times for each status message if there are multiple upstreams in the GS, so in some cases even 3 times (cluster, packetbroker, ttnv2). It's just noisy and doesn't add anything, so why send it?
2. Don't send a `gs.up.drop` event if the NS returns an "expected" error (such as "not found" or "duplicate"). In such cases the GS did actually forward the uplink to the NS, so there's no reason to say that it was dropped.
3. Don't send a `gs.up.drop` event if the DevAddr in an uplink doesn't match the upstream "filter". The fact that we don't forward something to the `ttnv2` upstream (for example) doesn't mean that we drop it completely. Perhaps we can (in the future) bring this event back, but only trigger it if an uplink isn't forwarded _anywhere_.
4. Include the `host` in the `gs.up.forward` event, so that gateway owners can see where their uplinks go, instead of just seeing multiple forward events without details.

#### Testing

Unfortunately I don't have a gateway to test this at the moment.

##### Regressions

I don't expect any.

#### Notes for Reviewers

Please comment if you agree on (1), (2), (3) and (4). I can remove commits if we don't agree on them.

#### Checklist


KrishnaIyer (approved):
I think this is a good idea,

1. ✅ 
2. ✅ 
3. Also fine. 
> Perhaps we can (in the future) bring this event back, but only trigger it if an uplink isn't forwarded anywhere.

That's not so easy though. We'll need a small refactoring [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/ed23609a6164b9be59bd25c9e7887b86ddd4a3e5/pkg/gatewayserver/gatewayserver.go#L746-L764) and maybe add an err channel per handler and log this if every handler returned an error. Let's think of this later if necessary.

4. ✅

pkg/gatewayserver/gatewayserver.go
```diff
@@ -664,12 +664,13 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
 					}
-					if err := host.handler.HandleUplink(ctx, gtw.GatewayIdentifiers, ids, msg); err != nil {
+					switch err := host.handler.HandleUplink(ctx, gtw.GatewayIdentifiers, ids, msg); codes.Code(errors.Code(err)) {
+					case codes.Canceled, codes.DeadlineExceeded,
+						codes.Unknown, codes.Internal,
+						codes.Unimplemented, codes.Unavailable:
 						drop(ids, errHostHandle.WithCause(err).WithAttributes("host", host.name))
-						break
 					}
 					registerForwardUplink(ctx, gtw, msg.UplinkMessage, host.name)
```
johanstokking: So with for instance `DeadlineExceeded`, there's a drop and a forward registration. That doesn't seem right. Shouldn't `registerForwardUplink()` be called if the error code is `OK`?

htdvisser: This should have been in the `default` case; fixed that.

kschiffer:
Agree with all points.

PR #4173, opened 2021-05-12
Add previews for events with google.protobuf.Value data

htdvisser:
#### Summary

This adds previews for event data of type `google.protobuf.Value`.

#### Testing

Locally. It looks like this:

Screen Shot 2021-05-12 at 15 37 33

Note the `["description"]` for the "Update gateway" event, and the `cluster` for the "Forward uplink message" event.

#### Checklist

### DIFF
--- a/pkg/webui/console/components/events/previews/value.js
+++ b/pkg/webui/console/components/events/previews/value.js
@@ -1,0 +1,39 @@
+// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


pkg/webui/console/components/events/previews/value.js
```diff
@@ -0,0 +1,37 @@
...
+import JSONPayload from './shared/json-payload'
+
+import style from './previews.styl'
+
+const Value = React.memo(({ event }) => {
+  const { data } = event
+  if (Array.isArray(data.value) || typeof data.value === 'object') {
+    return <JSONPayload data={data.value} />
+  }
+  return <Message className={style.plainText} content={data.value} />
```
kschiffer: `<Message />` is only for i18n messages. To just show the plain value, you should use

```
<DescriptionList.Item key="data.value" data={data.value} />
```

Then make sure to wrap the whole output around `<DescriptionList.Item title={sharedMessages.value} />`

(also the JSON output should be wrapped around this)

kschiffer (changes requested):
Small tweaking necessary.

htdvisser:
@kschiffer can you take over, or re-assign to someone else? I probably won't be able to finish this before the implementation deadline next week, and for you it's probably only a couple minutes work.

kschiffer (approved):

pkg/webui/console/components/events/previews/value.js
kschiffer: I eventually decided against using the `title={sharedMessages.value}`, since it can be confusing for some values.

PR #4174, opened 2021-05-12
Load historical events in the Console

htdvisser:
#### Summary

This PR makes the console load historical events.

This is a follow-up of https://github.com/TheThingsNetwork/lorawan-stack/pull/4084.

#### Changes

- Request a tail of 10 events when starting the events stream
- Prevent duplicate events from getting stored (when opening/closing/opening the events stream while requesting history).

#### Testing

Local testing:

Screen Shot 2021-05-12 at 16 27 35

Note that, while requesting a tail of 10, the verbosity filter hides some events, so we only see 7 events.

In https://github.com/TheThingsNetwork/lorawan-stack/pull/4084#issuecomment-828273930 @kschiffer was worried that something could go wrong when the `data` field of an event is missing (because it was evicted from the historical events cache). The "Receive uplink message" at 16:16:45 proves that this is not a problem. I also checked the code to make sure this won't crash:

https://github.com/TheThingsNetwork/lorawan-stack/blob/5d44e36897c921e864e3e05be5f68b6c9cf61d81/pkg/webui/console/components/events/utils/index.js#L49-L57

#### Notes for Reviewers

...

#### Checklist

### DIFF
--- a/pkg/webui/console/store/reducers/events.js
+++ b/pkg/webui/console/store/reducers/events.js
@@ -53,7 +53,7 @@ const MemoizedRegExp = memoize(RegExp)
 
 const addEvent = (state, event) => {
   const { events, filter } = state
-  const { paused, streamStartTime } = state
+  const { paused } = state
 
   if (paused && !event.name.startsWith('synthetic')) {
     return {}
@@ -73,12 +73,6 @@ const addEvent = (state, event) => {
   // timestamped before it. This is to avoid showing events before the synthetic
   // resumption event.
   if (events[0] && events[0].name === EVENT_STATUS_RESUMED && event.time < events[0].time) {
-    return {}
-  }
-
-  // When requesting historical events (with tail / after) it's possible that we
-  // still have some of those in the store, so we don't add them again.
-  if (event.time < streamStartTime && events.find(e => e.unique_id === event.unique_id)) {
     return {}
   }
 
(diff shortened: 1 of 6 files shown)


htdvisser:
@kschiffer can you take over, or re-assign to someone else? I probably won't be able to finish this before the implementation deadline next week, and for you it's probably only a couple minutes work.

pkg/webui/console/store/reducers/events.js
```diff
@@ -76,6 +76,12 @@ const addEvent = (state, event) => {
     return {}
   }
 
+  // When requesting historical events (with tail / after) it's possible that we
+  // still have some of those in the store, so we don't add them again.
+  if (event.time < streamStartTime && events.find(e => e.unique_id === event.unique_id)) {
+    return {}
+  }
+
```
kschiffer: It is currently necessary to do this since the `after` prop in the event stream request is not 100% accurate likely due to minimal differences in the clocks. This leads to the very last historical event still coming through.

Once this is resolved we can remove this check. I've already limited the costly `events.find()` to historical events only.

cc @htdvisser

kschiffer:
I've made some adjustments. @bafonins please review.

adriansmares (approved):

bafonins (approved):

kschiffer:
Seems like there's some GitHub Action issue currently causing jobs to be displayed double :shrug: . I'll merge now.

Issue #3250, opened 2020-09-17
Migrate from GORM to Bun


htdvisser:
#### Summary

We should replace GORM v1 with something else.

#### Why do we need this?

GORM v1 is no longer maintained.

#### What is already there? What do you see now?

We're using GORM v1

#### What is missing? What do you want to see?

Stop using GORM v1

#### How do you propose to implement this?

We discussed two options:

1. Upgrade to GORM v2. There are some significant changes, so upgrading requires quite some effort. Still mostly maintained by the same person, but at least it's now hosted in a GitHub organization with other members.
2. Switch to another ORM or query builder library. We've been experimenting with [Bun](https://github.com/uptrace/bun) and are quite happy with how that works. Maintained by the same person who maintains the Redis library we're also using. 

#### How do you propose to test this?

With the new `storetest` package we can test the new store implementation against the same tests we wrote for our current store implementation.

htdvisser:
Will get back to this after #3467 is merged.

htdvisser:
I'll take another look at this.

htdvisser:
Slow progress, so not going to finish this in Q1. Pushing to Q2.

htdvisser:
For v3.20.0 I plan to replace our old DB initialization/migration tooling with a new `pkg/identityserver/migrations` that uses [Bun migrations](https://bun.uptrace.dev/guide/migrations.html). We will require existing deployments to upgrade their databases to v3.19 (the last version that uses the old GORM auto-migrations) before upgrading to v3.20 (the first version that will use the new Bun migrations).

htdvisser:
The new Bun store implementation was merged in https://github.com/TheThingsNetwork/lorawan-stack/pull/5642, but behind a feature flag. https://github.com/TheThingsNetwork/lorawan-stack/issues/5643 now tracks promoting the new implementation to "stable" and removing the old GORM store implementation.

PR #4180, opened 2021-05-17
DevEUI Issuer 

### LINKED ISSUE #3750: Issue DevEUI from range with maximum per application
#### Summary

Issue DevEUIs from a configurable range, with a configurable maximum per application.

#### Why do we need this?

For people running The Things Stack and own a IEEE MAC block to issue DevEUIs for devices that still need to be programmed. This is also useful for The Things Network community network, and reducing the chance of registering a device with a real DevEUI.

The requirement to enter a DevEUI.

A button to issue a new DevEUI.

- Configure IS with a IEEE MAC block; i.e. `70B3D57ED0000000/36`
- Configure a maximum of DevEUIs per application, i.e. 50
- Add rpc to issue a new DevEUI for an application, that checks if the maximum hasn't been reached, atomically incrementing an IS-local value

Discussion point is: do we want this?

My take is yes, to reduce pollution. We may consider this for AppEUI/JoinEUI too, like we did previously with V2.

Unit and integration testing

Will test

pgalic96:
#### Summary

References #3750

#### Changes

- Add DevEUI address block configuration.
- Add RPC for issuing DevEUI per application.
- Extend application table and create new `eui_block` table.

#### Testing

Manually tested edge cases and the functionality.

Unit tests.

##### Regressions

None as this is a new functionality.

#### Notes for Reviewers

Tried to keep queries as condensed as possible to reduce the number of possible race conditions.

#### Checklist


cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -516,6 +516,22 @@ var (
 					device.DevEui = devID.DevEui
 				}
 			}
+			is, err := api.Dial(ctx, config.IdentityServerGRPCAddress)
+			if err != nil {
+				return err
+			}
+			requestDevEUI, _ := cmd.Flags().GetBool("request-dev-eui")
+			if requestDevEUI {
+				logger.Debug("request-dev-eui flag set, requesting a DevEUI from the block for the application")
```
johanstokking:
```suggestion
				logger.Debug("request-dev-eui flag set, requesting a DevEUI")
```

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -1253,6 +1264,7 @@ func init() {
 	endDevicesCreateCommand.Flags().Bool("abp", false, "configure end device as ABP")
 	endDevicesCreateCommand.Flags().Bool("with-session", false, "generate ABP session DevAddr and keys")
 	endDevicesCreateCommand.Flags().Bool("with-claim-authentication-code", false, "generate claim authentication code of 4 bytes")
+	endDevicesCreateCommand.Flags().Bool("request-dev-eui", false, "request a new DevEUI from the configured address block")
```
johanstokking:
```suggestion
	endDevicesCreateCommand.Flags().Bool("request-dev-eui", false, "request a new DevEUI")
```

"the configured address block" is a bit confusing for the user as there's no configuration in the CLI for this.

pkg/identityserver/application_registry.go
```diff
@@ -60,11 +60,18 @@ var (
 		events.WithAuthFromContext(),
 		events.WithClientInfoFromContext(),
 	)
+	evtIssueDevEUIForApplication = events.Define(
+		"application.issue_dev_eui", "issue dev eui for application",
```
johanstokking:
```suggestion
		"application.issue_dev_eui", "issue DevEUI for application",
```

pkg/identityserver/application_registry.go
```diff
@@ -60,11 +60,18 @@ var (
...
+		events.WithVisibility(ttnpb.RIGHT_APPLICATION_INFO),
+		events.WithAuthFromContext(),
+		events.WithClientInfoFromContext(),
+	)
 )
 
 var (
 	errAdminsCreateApplications = errors.DefinePermissionDenied("admins_create_applications", "applications may only be created by admins, or in organizations")
 	errAdminsPurgeApplications  = errors.DefinePermissionDenied("admins_purge_applications", "applications may only be purged by admins")
+	errDevEUIIssuingNotEnabled  = errors.DefineInvalidArgument("dev_eui_issuing_not_enabled", "dev eui issuing not configured")
```
johanstokking:
```suggestion
	errDevEUIIssuingNotEnabled  = errors.DefineInvalidArgument("dev_eui_issuing_not_enabled", "DevEUI issuing not configured")
```

pkg/identityserver/application_registry.go
```diff
@@ -333,6 +340,29 @@ func (is *IdentityServer) purgeApplication(ctx context.Context, ids *ttnpb.Appli
 	return ttnpb.Empty, nil
 }
 
+func (is *IdentityServer) requestDevEUI(ctx context.Context, ids *ttnpb.ApplicationIdentifiers) (*ttnpb.DevEUI, error) {
```
johanstokking: Let's be consistent and call it "issue" everywhere

pkg/identityserver/application_registry.go
```diff
@@ -333,6 +340,29 @@ func (is *IdentityServer) purgeApplication(ctx context.Context, ids *ttnpb.Appli
...
+	if err := rights.RequireApplication(store.WithSoftDeleted(ctx, false), *ids, ttnpb.RIGHT_APPLICATION_DEVICES_WRITE); err != nil {
+		return nil, err
+	}
+	if !is.config.DevEUIBlock.Enabled {
+		return nil, errDevEUIIssuingNotEnabled
```
johanstokking:
```suggestion
		return nil, errDevEUIIssuingNotEnabled.New()
```

pkg/identityserver/config.go
```diff
@@ -96,6 +96,11 @@ type Config struct {
 	Delete struct {
 		Restore time.Duration `name:"restore" description:"How long after soft-deletion an entity can be restored"`
 	} `name:"delete"`
+	DevEUIBlock struct {
+		Enabled          bool   `name:"enabled" description:"Enable DevEUI address issuing from IEEE MAC block"`
+		MaxAddressPerApp int    `name:"max-address-per-app" description:"Maximum DevEUI addresses to be issued per application"`
+		AddressBlock     string `name:"address-block" description:"DevEUI address block with prefix length"`
```
johanstokking: Don't we support `DevAddrPrefix` as the type?

pkg/identityserver/config.go
```diff
@@ -96,6 +96,11 @@ type Config struct {
 	Delete struct {
 		Restore time.Duration `name:"restore" description:"How long after soft-deletion an entity can be restored"`
 	} `name:"delete"`
+	DevEUIBlock struct {
+		Enabled          bool   `name:"enabled" description:"Enable DevEUI address issuing from IEEE MAC block"`
+		MaxAddressPerApp int    `name:"max-address-per-app" description:"Maximum DevEUI addresses to be issued per application"`
```
johanstokking:
```suggestion
		MaxPerApplication int    `name:"max-per-application" description:"Maximum DevEUI addresses to be issued per application"`
```

johanstokking: Or `ApplicationLimit`

pkg/identityserver/store/eui_block.go
```diff
@@ -0,0 +1,32 @@
...
+
+package store
+
+type EUIBlock struct {
+	Model
+
+	Type           string `gorm:"unique_index:eui_block_index;type:VARCHAR(10);"`
+	StartAddress   *EUI64 `gorm:"type:VARCHAR(16);column:start_address"`
+	EndAddress     *EUI64 `gorm:"type:VARCHAR(16);column:end_address"`
+	CurrentAddress *EUI64 `gorm:"type:VARCHAR(16);column:current_address"`
```
johanstokking: If we want to use this also for gateway EUIs, the term "address" isn't applicable

I'm not an IS code owner, but why pointers?

johanstokking (commented):
Mostly looked at API. Storage seems pretty ok too, maybe some code quality things there, leaving that up to IS code owners.

api/application.proto
```diff
@@ -55,6 +55,10 @@ message Applications {
   repeated Application applications = 1;
 }
 
+message DevEUI {
```
adriansmares: The naming is problematic here, since we would be inconsistent with other usages. I think `IssueDevEUIResponse` would clear that up.

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -516,6 +516,22 @@ var (
...
 				}
 			}
+			is, err := api.Dial(ctx, config.IdentityServerGRPCAddress)
+			if err != nil {
+				return err
+			}
+			requestDevEUI, _ := cmd.Flags().GetBool("request-dev-eui")
+			if requestDevEUI {
+				logger.Debug("request-dev-eui flag set, requesting a DevEUI from the block for the application")
+				appID := getApplicationID(cmd.Flags(), args)
```
adriansmares: `nit`: I think this is already in `devID`, and maybe the EUI should be filled there too.

pkg/identityserver/store/eui_block.go
adriansmares: Instead of address, `Identifier` may be used (the `I` in `EUI`).

In other entities these EUI fields may be nullable, ergo the pointers, but I think this shouldn't be required here.

pkg/identityserver/store/eui_block_store.go
```diff
@@ -0,0 +1,161 @@
...
+		return err
+	}
+	// If app dev eui limit not configured, skip this step.
+	if appDevEUILimit == 0 {
+		return nil
+	}
+	// Atomically check and increment dev eui counter if its below threshold.
+	result := s.query(ctx, Application{}).
+		Where(Application{ApplicationID: appModel.ApplicationID}).
+		Where(fmt.Sprintf(`"applications"."dev_eui_counter" < '%d'`, appDevEUILimit)).
```
adriansmares: This is safe, but still gives me nightmares.

I think ```Where(`"applications"."dev_eui_counter" < ?`, appDevEUILimit).``` would be better.

pkg/identityserver/store/eui_block_store.go
```diff
@@ -0,0 +1,161 @@
...
+			Eui        EUI64
+			ExistingId string
+		}
+		if err := query.Scan(&result).Error; err != nil {
+			if gorm.IsRecordNotFoundError(err) {
+				return nil, errMaxGlobalEUILimitReached
+			}
+			return nil, err
+		}
+		if err := s.query(ctx, EUIBlock{}).
```
adriansmares: This is not atomic and can race between the Identity Servers that are just starting up, I think.

adriansmares (commented):
While the intent and implementations are quite good, I think we're creating complexity out of thin air by forcing ourselves to do logic with actual EUI blocks in the database. 

We could do everything far safer and atomically if the blocks were saved as `int64` and all of the operations where done directly in the database, and then converted to `DevEUI` while going back to protobuf.

Another thing is that IS initialization is race-y between two parallel Identity Server instances - a transaction should be used there.

pkg/identityserver/config.go
pgalic96: You mean `EUI64Prefix`, right?

pgalic96:
@adriansmares I believe most of the race condition sections are fixed, with the exception of the one here:

https://github.com/TheThingsNetwork/lorawan-stack/pull/4180/commits/929ba6604b2dd87fcd3349f39a10a9cd975da6d0#diff-9b40838043d19eaaddc019509ac9c523d2da41c3ccfe6ee9bd95048b9e630e79R83-R103

The idea is to read the current counter value and then increment, without the need to know the incremented value but the value before. GORM doesn't support the `RETURNING` statement after update (https://github.com/go-gorm/gorm/issues/4088). This could also be prevented by using `FOR UPDATE` lock, which GORM v2 has a nice support for but we are still on v1 (https://github.com/TheThingsNetwork/lorawan-stack/issues/3250). We could still probably go around this by using raw SQL but the question is whether we want to overengineer this solution in that way. I got the impression that we don't cc @htdvisser .

pkg/identityserver/config.go
```diff
@@ -96,6 +97,11 @@ type Config struct {
 	Delete struct {
 		Restore time.Duration `name:"restore" description:"How long after soft-deletion an entity can be restored"`
 	} `name:"delete"`
+	DevEUIBlock struct {
+		Enabled          bool                 `name:"enabled" description:"Enable DevEUI address issuing from IEEE MAC block"`
+		ApplicationLimit int                  `name:"application-limit" description:"Maximum DevEUI addresses to be issued per application"`
+		AddressBlock     ttntypes.EUI64Prefix `name:"address-block" description:"DevEUI address block with prefix length"`
```
johanstokking: This isn't an "address". This should simply be prefix.

johanstokking (commented):
Only reviewed code owned files: API, CLI and config

johanstokking (approved):
API, CLI and config LGTM

pkg/identityserver/identityserver.go
```diff
@@ -102,6 +102,19 @@ func New(c *component.Component, config *Config) (is *IdentityServer, err error)
 		return nil, err
 	}
 
+	if is.config.DevEUIBlock.Enabled {
+		err := is.withDatabase(is.Context(), func(db *gorm.DB) error {
+			err = store.GetEUIStore(db).CreateEUIBlock(is.Context(), "dev_eui", is.config.DevEUIBlock.Prefix)
```
adriansmares: Please use a transaction here. See https://github.com/TheThingsNetwork/lorawan-stack/blob/c0e437afc7fccb9d506e05f73b4a2d2c3db54d13/pkg/identityserver/store/store.go#L224-L254 .

pkg/identityserver/application_registry.go
```diff
@@ -333,6 +340,29 @@ func (is *IdentityServer) purgeApplication(ctx context.Context, ids *ttnpb.Appli
...
 
+func (is *IdentityServer) issueDevEUI(ctx context.Context, ids *ttnpb.ApplicationIdentifiers) (*ttnpb.IssueDevEUIResponse, error) {
+	if err := rights.RequireApplication(store.WithSoftDeleted(ctx, false), *ids, ttnpb.RIGHT_APPLICATION_DEVICES_WRITE); err != nil {
+		return nil, err
+	}
+	if !is.config.DevEUIBlock.Enabled {
+		return nil, errDevEUIIssuingNotEnabled.New()
+	}
+	res := &ttnpb.IssueDevEUIResponse{}
+	err := is.withDatabase(ctx, func(db *gorm.DB) error {
```
adriansmares, line 351: Use a transaction here too - you shouldn't need the spin loop anymore I think, if you do.

adriansmares (approved):

johanstokking:
> #### Summary
> Closes #3750

If Console isn't included then this isn't closing the issue.

kschiffer (approved):

Issue #4095, opened 2021-04-28
Rate limit identical messages


johanstokking:
#### Summary

Rate limit identical messages coming from the same gateway.

#### Why do we need this?

Sometimes we see issues with the same message being forwarded many times by the same gateway.

This is often an issue in the gateway forwarder or concentrator, that concentrator registers aren't cleared correctly or that the forwarder software is simply in a loop.

#### What is already there? What do you see now?

We have rate limiting of UDP packets per session, and there is general rate limiting support in GS.

#### What is missing? What do you want to see?

Packet based rate limiting. This shouldn't be (only) in the UDP frontend. It should be per PHYPayload per gateway antenna, and it should expire after a short time. That is, in the order of tens of seconds.

#### How do you propose to implement this?

Since this is per gateway connection, there doesn't have to be persistency across multiple GS instances.

I would propose using the existing rate limiting infrastructure. The rate limiting key should contain the PHYPayload (plain, hash or CRC) and antenna index, and expire after 10 seconds. I think it's best to implement this in `(*GatewayServer).handleUpstream()` when reading from the `Up()` channel.

#### How do you propose to test this?

Unit testing

#### Can you do this yourself and submit a Pull Request?

Can review

neoaggelos:
If this is a known behaviour, and the problem is that gateways keep sending the same message over and over, then I suppose it is better to stick to a simpler solution of simply keeping track of the last message sent by the gateway and discard duplicates, similar to what [v2 is doing](https://github.com/TheThingsNetwork/gateway-connector-bridge/blob/master/middleware/deduplicate/deduplicate.go) ?

johanstokking:
Yes, that's probably the simplest solution.

Also note that in LoRaWAN, retransmissions are a feature. The same (un)confirmed uplink message can be sent multiple times, with the same FCnt, i.e. the same PHYPayload. Frequency hopping does apply though. Also, RX2 applies: the end device has to wait at least 2 seconds for the next frame.

So, if you want to work with a simple duplication detector on the last message, take into account either a) duplicate messages within 2 seconds or b) duplicate messages that are using the same frequency.

johanstokking:
Another related issue is that some gateways are sending duplicate messages of the same uplink frame:

1. Same PHYPayload
2. Different frequency
3. Different RSSI; e.g. -44 vs -100. The right one is the message with the highest RSSI
[...]

PR #4181, opened 2021-05-18
Drop duplicate uplinks based on payload, frequency and antenna index

neoaggelos:
#### Summary

References #4095 

#### Changes

- Store the last uplink message received from an `*io.Connection`
- When a new uplink is received, compare it with the last stored message (payload, antenna index, freqeuency). Discard the message if it's a duplicate.

#### Testing

Unit testing. Test locally and observe that normal traffic still goes through without issues.

##### Regressions

...

#### Notes for Reviewers

@johanstokking The implementation follows the suggestion in #4095, but perhaps we should consider doing this in the `*io.Connection` side instead. This would prevent forwarding messages that we know will be discarded later. This is also related to the uplink count stats (with the current implementation, we count duplicate uplinks in the gateway stats.

@htdvisser `registerDropUplink` is _very_ chatty in this situation. Some ideas would be: (a) use a separate metric (b) emit an event only once per X seconds. What do you think?

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/io.go
+++ b/pkg/gatewayserver/io/io.go
@@ -15,6 +15,7 @@
 package io
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"sync/atomic"
@@ -99,6 +100,15 @@ type Connection struct {
 
 	statsChangedCh chan struct{}
 	locCh          chan struct{}
+
+	lastUplink            *uplinkMessage
+	lastRepeatUpEventTime time.Time
+}
+
+type uplinkMessage struct {
+	payload   []byte
+	frequency uint64
+	antennas  []uint32
 }
 
 var (
@@ -183,8 +193,15 @@ func (c *Connection) Gateway() *ttnpb.Gateway { return c.gateway }
 
 var errBufferFull = errors.DefineInternal("buffer_full", "buffer is full")
 
+// Interval between emitting consecutive gs.up.repeat events for the same gateway connection.
+const consecutiveRepeatUpEventsInterval = time.Minute
+
 // HandleUp updates the uplink stats and sends the message to the upstream channel.
 func (c *Connection) HandleUp(up *ttnpb.UplinkMessage) error {
+	if c.discardRepeatedUplink(up) {
+		return nil
+	}
+
 	var ct scheduling.ConcentratorTime
 	if up.Settings.Time != nil {
 		ct = c.scheduler.SyncWithGatewayAbsolute(up.Settings.Timestamp, up.ReceivedAt, *up.Settings.Time)
@@ -669,4 +686,49 @@ func (c *Connection) notifyStatsChanged() {
 	case c.statsChangedCh <- struct{}{}:
 	default:
 	}
+}
+
+func uplinkMessageFromProto(pb *ttnpb.UplinkMessage) *uplinkMessage {
+	up := &uplinkMessage{
+		payload:   pb.GetRawPayload(),
+		frequency: pb.GetSettings().Frequency,
+		antennas:  make([]uint32, 0, len(pb.GetRxMetadata())),
+	}
+	for _, md := range pb.GetRxMetadata() {
+		up.antennas = append(up.antennas, md.GetAntennaIndex())
+	}
+	return up
+}
+
+func isRepeatedUplink(this *uplinkMessage, that *uplinkMessage) bool {
+	if this == nil || that == nil || this.frequency != that.frequency || len(this.antennas) != len(that.antennas) || !bytes.Equal(this.payload, that.payload) {
+		return false
+	}
+	for idx, antenna := range this.antennas {
+		if that.antennas[idx] != antenna {
+			return false
+		}
+	}
+	return true
+}
+
+// discardRepeatedUplink will discard repeated uplinks from faulty gateway
+// implementations. It returns true if the uplink message is the same as the
+// last uplink message that was received by the connection.
+//
+// discardRepeatedUplink is not goroutine safe.
+func (c *Connection) discardRepeatedUplink(up *ttnpb.UplinkMessage) bool {
+	uplink := uplinkMessageFromProto(up)
+	shouldDiscard := isRepeatedUplink(c.lastUplink, uplink)
+	c.lastUplink = uplink
+	if shouldDiscard {
+		shouldEmitEvent := false
+		if time.Since(c.lastRepeatUpEventTime) >= consecutiveRepeatUpEventsInterval {
+			log.FromContext(c.ctx).Debug("Dropped repeated gateway uplink")
+			shouldEmitEvent = true
+			c.lastRepeatUpEventTime = time.Now()
+		}
+		registerRepeatUp(c.ctx, shouldEmitEvent, c.gateway, c.frontend.Protocol())
+	}
+	return shouldDiscard
 }
--- a/pkg/gatewayserver/io/observability.go
+++ b/pkg/gatewayserver/io/observability.go
@@ -1,0 +1,66 @@
+// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package io
+
+import (
+	"context"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"go.thethings.network/lorawan-stack/v3/pkg/events"
+	"go.thethings.network/lorawan-stack/v3/pkg/metrics"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+type messageMetrics struct {
+	repeatedUplinks *metrics.ContextualCounterVec
+}
+
+// Describe implements prometheus.Collector.
+func (m *messageMetrics) Describe(ch chan<- *prometheus.Desc) {
+	m.repeatedUplinks.Describe(ch)
+}
+
+// Collect implements prometheus.Collector.
+func (m *messageMetrics) Collect(ch chan<- prometheus.Metric) {
+	m.repeatedUplinks.Collect(ch)
+}
+
+var ioMetrics = &messageMetrics{
+	repeatedUplinks: metrics.NewContextualCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: "gs",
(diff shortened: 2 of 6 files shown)


johanstokking:
> @johanstokking The implementation follows the suggestion in #4095, but perhaps we should consider doing this in the `*io.Connection` side instead. This would prevent forwarding messages that we know will be discarded later.

Yeah that makes sense.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -433,6 +435,26 @@ type connectionEntry struct {
 	*io.Connection
 	upstreamDone chan struct{}
 	tasksDone    *sync.WaitGroup
+	lastMessage  *ttnpb.UplinkMessage
+}
+
+func isDuplicateUplink(this *ttnpb.UplinkMessage, that *ttnpb.UplinkMessage) bool {
+	switch {
+	case this == nil || that == nil:
```
johanstokking: Isn't this an implementation error?

neoaggelos: `this` will be nil for the first uplink of the connection. `that` will be whatever the frontend sends, I think it's best not to panic here for messages if the frontend is buggy. What do we gain if we skip this?

htdvisser:
> @htdvisser `registerDropUplink` is _very_ chatty in this situation. Some ideas would be: (a) use a separate metric (b) emit an event only once per X seconds. What do you think?

I think a circuit breaker style approach would work quite well here, where we only trigger an event once when it kicks in (when it starts dropping duplicates).

pkg/gatewayserver/io/io.go
```diff
@@ -201,6 +204,11 @@ func (c *Connection) HandleUp(up *ttnpb.UplinkMessage) error {
 		)).Debug("Synchronized server absolute time only")
 	}
 
+	if isDuplicateUplink(c.lastUplinkMessage, up) {
+		return nil
+	}
+	c.lastUplinkMessage = deepcopy.Copy(up).(*ttnpb.UplinkMessage)
```
johanstokking: `deepcopy.Copy()` is a very expensive operation that we should avoid in the hot path.

The only thing we need to consider here is the PHYPayload, frequency and antenna indices right? Why not put those in a struct and store in a `atomic.Value` and make `isDuplicateUplink` a method of `*Connection` that takes the relevant fields (PHYPayload + RX metadata) as arguments?

adriansmares (commented):
Nothing to add. After the storage 'slim down' this should be fine.

neoaggelos:
> I think a circuit breaker style approach would work quite well here, where we only trigger an event once when it kicks in (when it starts dropping duplicates).

`evtDropUp` is defined in the top-level `gatewayserver` package, so it is not available in `gatewayserver/io`. I would suggest that we define a separate event in `gatewayserver/io` (maybe `evtDuplicateUp`?). This would allow to include a `repeated` field in the event data, which could help with identifying how many times the repeated message was received.

Thoughts, @htdvisser, @johanstokking ? This seems clearer than trying to find ways to forward the error to the Gateway Server component.

johanstokking:
1. Publishing `evtDuplicateUp` isn't a circuit breaker. It would be circuit breaker if we had start/stop events for duplicates, where new duplicates wouldn't trigger any events. In the stop event, we _can_ put a number of duplicates that have been filtered
2. If we include all duplicates in a `repeated` field, we move the problem from publishing many events to publishing massive events with potentially thousands of duplicates
3. `evtDropUp` is indeed intended per-message and is used to ensure that receive and drop/forward are in balance (taking into account the upstream hosts)

pkg/gatewayserver/io/io.go
```diff
@@ -99,6 +100,15 @@ type Connection struct {
...
 	locCh          chan struct{}
+
+	lastUp            *uplinkMessage
+	lastRepeatUpEvent time.Time
+}
+
+type uplinkMessage struct {
+	payload     []byte
+	frequency   uint64
+	antennaMask uint64
```
johanstokking: Why a mask?

pkg/gatewayserver/io/io.go
```diff
@@ -661,3 +681,44 @@ func (c *Connection) notifyStatsChanged() {
...
+	case !bytes.Equal(this.payload, that.payload):
+		return false
+	}
+	return true
+}
+
+// Interval between emitting consecutive gs.up.repeat events for the same gateway connection.
+const consecutiveRepeatUpEventsInterval = time.Minute
+
+func (c *Connection) checkRepeatedUplink(up *uplinkMessage) (repeated, shouldEmitEvent bool) {
```
johanstokking: nit; let the caller (`HandleUp()`) not bother with the data structure (`uplinkMessage`) that is being used internally.

johanstokking: This is also not just a check (which implies no state is changed), whereas state (`lastUp`) is actually being changed. This is important for our future selves; this method isn't goroutine safe.

pkg/gatewayserver/io/observability.go
```diff
@@ -0,0 +1,67 @@
...
+	),
+}
+
+var evtRepeatUp = events.Define(
+	"gs.up.repeat", "received repeated uplink message from gateway",
+	events.WithVisibility(ttnpb.RIGHT_GATEWAY_TRAFFIC_READ),
+	events.WithDataType(&ttnpb.GatewayIdentifiers{}),
+)
+
+func registerRepeatUpMetric(ctx context.Context, protocol string) {
```
johanstokking: Remove `Metric`; the caller doesn't care whether there's a metric and/or event

pkg/gatewayserver/io/observability.go
```diff
@@ -0,0 +1,67 @@
...
+	ioMetrics.repeatedUplinks.WithLabelValues(ctx, protocol).Inc()
+}
+
+func registerRepeatUpEvent(ctx context.Context, gtw *ttnpb.Gateway) {
```
johanstokking: Idem

pkg/gatewayserver/io/observability.go
neoaggelos: The idea here is that metrics are always increased, but we only emit events once per minute for each gateway.

pkg/gatewayserver/io/io.go
```diff
@@ -185,6 +195,16 @@ var errBufferFull = errors.DefineInternal("buffer_full", "buffer is full")
 
 // HandleUp updates the uplink stats and sends the message to the upstream channel.
 func (c *Connection) HandleUp(up *ttnpb.UplinkMessage) error {
+	// Discard repeated uplinks from faulty gateway implementations.
+	if repeated, shouldEmitEvent := c.checkRepeatedUplink(uplinkMessageFromProto(up)); repeated {
+		registerRepeatUpMetric(c.ctx, c.frontend.Protocol())
+		if shouldEmitEvent {
+			log.FromContext(c.ctx).Debug("Dropped repeated gateway uplink")
+			registerRepeatUpEvent(c.ctx, c.gateway)
+		}
```
johanstokking: I see. Then;

```suggestion
		registerRepeatUp(c.ctx, shouldEmitEvent, c.frontend.Protocol())
		if shouldEmitEvent {
			log.FromContext(c.ctx).Debug("Dropped repeated gateway uplink")
		}
```

pkg/gatewayserver/io/io.go
neoaggelos: This is to reduce memory allocations per uplink message.

Is it safe to assume that antenna index will be < 63, or is this a prime example of premature (and largely unwarranted) optimization?

pkg/gatewayserver/io/io.go
```diff
@@ -661,3 +678,50 @@ func (c *Connection) notifyStatsChanged() {
...
 	}
 }
+
+func uplinkMessageFromProto(pb *ttnpb.UplinkMessage) *uplinkMessage {
+	up := &uplinkMessage{
+		payload:   pb.GetRawPayload(),
+		frequency: pb.GetSettings().Frequency,
+	}
+	for _, md := range pb.GetRxMetadata() {
+		up.antennaMask = up.antennaMask | (1 << md.GetAntennaIndex())
```
johanstokking: nit

```suggestion
		up.antennaMask |= 1 << md.GetAntennaIndex()
```

pkg/gatewayserver/io/io.go
```diff
@@ -661,3 +678,50 @@ func (c *Connection) notifyStatsChanged() {
...
+//
+// discardRepeatedUplink is not goroutine safe.
+func (c *Connection) discardRepeatedUplink(up *ttnpb.UplinkMessage) bool {
+	uplink := uplinkMessageFromProto(up)
+	shouldDiscard := isRepeatedUplink(c.lastUplink, uplink)
+	c.lastUplink = uplink
+	if shouldDiscard {
+		now := time.Now()
+		shouldEmitEvent := false
+		if now.Sub(c.lastRepeatUpEventTime) >= consecutiveRepeatUpEventsInterval {
```
johanstokking: nit

```suggestion
		shouldEmitEvent := false
		if time.Since(c.lastRepeatUpEventTime) >= consecutiveRepeatUpEventsInterval {
```

pkg/gatewayserver/io/observability.go
```diff
@@ -0,0 +1,66 @@
...
+	),
+}
+
+var evtRepeatUp = events.Define(
+	"gs.up.repeat", "received repeated uplink message from gateway",
+	events.WithVisibility(ttnpb.RIGHT_GATEWAY_TRAFFIC_READ),
+	events.WithDataType(&ttnpb.GatewayIdentifiers{}),
+)
+
+func registerRepeatUp(ctx context.Context, emitEvent bool, gtw *ttnpb.Gateway, protocol string) {
```
johanstokking, line 57: Consider exposing this on `Server` so that all observability is concentrated in `pkg/gatewayserver/observability.go`

neoaggelos: Even if we expose this on `io.Server`, the `*io.Connection` does not keep a reference to the connected server (not sure if this is by design, or it was just never needed previously).

It felt simpler to declare the metric in `pkg/gatewayserver/io` and keep the diff small (though I understand that it is probably not the cleanest solution), instead of declaring it in `pkg/gatewayserver` and trying to find a way to pass it down to the connection.

johanstokking: Ah right. No I think that this isn't even possible because of the import cycle. I'm OK with this then. Please re-request review when ready.

pkg/gatewayserver/io/io.go
adriansmares: Although in practice this is most likely safe to assume, there is nothing to guarantee it, so it's not ideal. A `map[uint32]struct{}` has a bit of a memory footprint, but could work.

If you want to keep this mask solution, at least document it (especially the assumptions).

johanstokking (approved):
Great stuff, thanks

adriansmares (approved):

pkg/gatewayserver/io/io.go
neoaggelos: Well, if we are gonna do `map[uint32]struct{}`, we must as well just do the `[]uint32` instead. Better safe than sorry.

PR #4189, opened 2021-05-19
Open bleve index in read-only mode

neoaggelos:
#### Summary

Open Bleve index files in read-only mode and remove 5 second timeout, which causes The Things Stack to sometimes fail to initialize in environments with slower disk speeds (e.g. a Raspberry Pi).

#### Changes

- Open bleve in read only mode.
- Remove 5 second timeout when opening index.

#### Testing

Unit tests, test locally.

##### Regressions

There should not be any. The index is only used for queries.

#### Notes for Reviewers

This is simple enough to add in time for 3.13.0, which resolves with running The Things Stack in a Raspberry Pi. This is an internal change, so I am not sure if something is needed in the changelog.

#### Checklist

### DIFF
--- a/pkg/devicerepository/store/bleve/bleve.go
+++ b/pkg/devicerepository/store/bleve/bleve.go
@@ -79,24 +79,10 @@ func (c Config) NewStore(ctx context.Context) (store.Store, error) {
 var errCannotOpenIndex = errors.DefineNotFound("cannot_open_index", "cannot open index")
 
 func openIndex(ctx context.Context, path string) (bleve.Index, error) {
-	var (
-		err   error
-		index bleve.Index
-	)
-	done := make(chan struct{}, 1)
-	defer close(done)
 	log.FromContext(ctx).WithField("path", path).Debug("Loading index")
-	go func() {
-		index, err = bleve.Open(path)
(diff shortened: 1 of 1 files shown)


pkg/devicerepository/store/bleve/bleve.go
```diff
@@ -79,24 +79,10 @@ func (c Config) NewStore(ctx context.Context) (store.Store, error) {
...
-	}()
-	select {
-	case <-done:
-		if err != nil {
-			return nil, errCannotOpenIndex.WithAttributes("path", path).WithCause(err)
-		}
-		return index, nil
-	case <-ctx.Done():
-		return nil, ctx.Err()
+	index, err := bleve.OpenUsing(path, map[string]interface{}{"read_only": true, "bolt_timeout": "10s"})
```
adriansmares: I would go for 60 seconds for `bolt_timeout`, since we actually just upped the timeout from 5s to 10s with this change.

References https://github.com/etcd-io/bbolt/blob/f6be82302843a215152f5a1daf652c1ee5503f85/db.go#L1020-L1023

adriansmares (approved):

Issue #359, opened 2019-03-26
Frontend assets are too big


htdvisser:
**Summary:** 

While working on Gzip (#350) I noticed that our static assets are very big. 

**Steps to Reproduce:** 

1. Just look in the `public` dir.

**What do you see now?**

Several MBs that users need to download when opening the OAuth page or Console.

**What do you want to see instead?**

Less

**How do you propose to implement this?**

- We can shed quite some weight by optimizing the images
- I hope @kschiffer and @bafonins have some ideas for the JS

**What can you do yourself and what do you need help with?**

I can submit a PR with optimized images, the rest is up to @kschiffer and @bafonins.

kschiffer:
As far as the OAuth bundle goes, I can see that there are some dependencies attached that the OAuth app is not actually using (eg. `react-select`, `clipboard.js`). We should try looking into optimizations like tree-shaking to make sure we don't have dead code in the bundles. 
Especially the OAuth app should be really slim, as there is really not much happening in there.

kschiffer:
After putting the webpack `stats.json` into [webpack visualizer](https://chrisbateman.github.io/webpack-visualizer/), I can see that in the OAuth bundle, a lot of space is consumed e.g. by `formik` and `rxjs` which is a dependency of `redux-logic`. To really reduce bundle size, we need to try and avoid big bundles which we only use marginally.

bafonins:
The issue is with the size of the `react-ace` dependency. We plan do code split the payload formats/events pages to reduce the bundle size.

htdvisser:
Looks like we're not the only ones that dislike the size of the Ace editor. Have you considered using a smaller one like https://github.com/satya164/react-simple-code-editor instead?

bafonins:
Indeed, we can try using `react-simple-code-editor`.

bafonins:
Remark for later:

We can make use of [browserslist](https://github.com/browserslist/browserslist) that is used internally by babel. This can reduce the bundle by removing specific polyfills for browsers we do not support. Additionally, we can define separate instances for the development and production modes to speed up the webui build.

PR #4192, opened 2021-05-19
Fix out of memory issues for device import in the Console

### LINKED ISSUE #4178: Running out of memory when importing many end devices from the console.
#### Summary

When importing a JSON file with lots of devices (> 200) via the Console, there is a huge increase in memory usage, which leads to page crashes and the import process to fail. 

#### Steps to Reproduce

1. Create a large export file
2. Import End Devices > Choose File > Import

#### What do you see now?

Delay and lag while the file is being parsed, import process fails after adding a few devices due to running out of memory.

image

Note that the memory usage is 3.5GB, where as the export file is close to 1MB in size.

The import process should succeeed.

#### Environment

3.12.3

This is probably a memory leak somewhere.

No, can provide an export file that causes the issue, can help with testing.

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4178

#### Changes

- Reset code-editor `undoManager` for device import
- Some small clean up

#### Testing

Manual

#### Notes for Reviewers

The issue is caused by `ace-editor` persisting all values in memory to allow undo/redo functionality within the editor. With this fix in place I was able to import 600 devices without any issues/crashes and not exceeding 400mb of memory.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ For details about compatibility between different releases, see the **Commitment
 - The CLI now properly returns a non-zero exit status code on invalid commands.
 - Gateway connection requests with zero EUI are rejected.
 - End device payload formatter reset to `FORMATTER_NONE` in the Console.
+- Memory issues when importing end devices in the Console.
 
 ### Security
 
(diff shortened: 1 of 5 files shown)


kschiffer (approved):
I think it's really time to move away from ace. It's just way too heavy and overloaded for our use cases. It's odd that undo cannot be deactivated globally or it not being activated at all when the editor is in read only mode...

Btw in the dev console, I see the device import also generates a lot of `Cannot find source map for …` warnings. These should not occur in production though.

bafonins:
Related issue https://github.com/TheThingsNetwork/lorawan-stack/issues/359#issuecomment-514543159

PR #4210, opened 2021-05-26
Do not return downlink token if the downlink is nil

KrishnaIyer:
#### Summary

Do not return downlink token if the downlink is nil. Fixes the panic reported on Slack.

#### Changes

- If the `downlink` message is nil, return false.

#### Testing

I think if CI passes this is good.

##### Regressions

Don't see any.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/downlink_token.go
+++ b/pkg/gatewayserver/io/downlink_token.go
@@ -57,7 +57,7 @@ func (t *DownlinkTokens) Next(msg *ttnpb.DownlinkMessage, time time.Time) uint16
 func (t DownlinkTokens) Get(token uint16, time time.Time) (*ttnpb.DownlinkMessage, time.Duration, bool) {
 	pos := token % downlinkTokenItems
 	item := t.items[pos]
-	if item.key != token {
+	if item.key != token || item.msg == nil {
 		return nil, 0, false
 	}
 	return item.msg, time.Sub(item.time), true
--- a/pkg/gatewayserver/io/downlink_token.go
+++ b/pkg/gatewayserver/io/downlink_token.go
@@ -57,7 +57,7 @@ func (t *DownlinkTokens) Next(msg *ttnpb.DownlinkMessage, time time.Time) uint16
 func (t DownlinkTokens) Get(token uint16, time time.Time) (*ttnpb.DownlinkMessage, time.Duration, bool) {
 	pos := token % downlinkTokenItems
 	item := t.items[pos]
-	if item.key != token || item.msg == nil {
+	if item.key != token || item.msg == nil || item.time.IsZero() {
 		return nil, 0, false
 	}
 	return item.msg, time.Sub(item.time), true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ For details about compatibility between different releases, see the **Commitment
 - Downlink queue eviction on FCnt mismatch.
 - End device payload formatter view crashing in the Console.
 - End device overview frequently crashing in the Console.
+- Panic on empty downlink in zero indexed downlink token.
 
 ### Security
 


johanstokking (commented):
This is `nil` in and `nil` out. I don't know why `DownlinkTokens` would support putting `nil` messages but not returning them. How do they end up in the downlink tokens in the first place? Why isn't the caller of `Get()` allowing for a `nil` result?

Note that downlink tokens are not only used to store downlink messages, but also for round-trip times. So `nil, DURATION, true` is a useful result of `Get()`.

KrishnaIyer:
> Note that downlink tokens are not only used to store downlink messages, but also for round-trip times. So nil, DURATION, true is a useful result of Get().

Hmm? Where is a situation where downlink (that is supposedly sent to the gateway) is nil but there's an entry for `downlinkToken.time`? How do we send a nil downlink and then check the corresponding Tx ACK for RTT?

> How do they end up in the downlink tokens in the first place?

Because when the key is 0 and the slice is empty, we just return an empty `downlinkToken` whose `msg` is nil.
https://github.com/TheThingsNetwork/lorawan-stack/blob/19f5503f16c986f1c71c7b72f38f4a0a677a591e/pkg/gatewayserver/io/downlink_token.go#L59

We could indeed just check for nil value on the caller, but I don't see any valid case for `nil, DURATION, true`.

What's worse, since we're using the default value `time.Time` in the RTT, we would be skewing the RTT value.

pkg/gatewayserver/io/udp/udp.go
```diff
@@ -328,7 +328,8 @@ func (s *srv) handleUp(ctx context.Context, state *state, packet encoding.Packet
 			}
 		}
 		var rtt *time.Duration
-		if downlink, delta, ok := state.tokens.Get(uint16(packet.Token[0])<<8|uint16(packet.Token[1]), packet.ReceivedAt); ok {
+		downlink, delta, ok := state.tokens.Get(uint16(packet.Token[0])<<8|uint16(packet.Token[1]), packet.ReceivedAt)
+		if ok && downlink != nil {
 			msg.TxAcknowledgment.DownlinkMessage = downlink
 			msg.TxAcknowledgment.CorrelationIDs = downlink.CorrelationIDs
 			rtt = &delta
```
KrishnaIyer: I don't see a case where there's no downlink but a corresponding time stamp is stored to check the TxAck against.

adriansmares:
@johanstokking keep in mind that your assumption is that the value _was_ put there via `Next`. This is not always the case.

If `uint16(packet.Token[0])<<8|uint16(packet.Token[1])` is a multiple of `downlinkTokenItems` (or simply 0), and the connection is new (maybe due to a reconnect, like the spurious ones we're currently encountering) the following condition will be false:
https://github.com/TheThingsNetwork/lorawan-stack/blob/3bc2c96246a7421fe7bba320f2e0a378a6e400c5/pkg/gatewayserver/io/downlink_token.go#L60
But there is no item there - it's just the zero value of `downlinkToken`. Both `down` and `time` will be `nil`/`0`. 

In my opinion 
https://github.com/TheThingsNetwork/lorawan-stack/blob/3bc2c96246a7421fe7bba320f2e0a378a6e400c5/pkg/gatewayserver/io/downlink_token.go#L60
is too weak, because I don't get `something, something, false` if I call `Get` with a token of `0` on a newly initialized `DownlinkTokens`.

johanstokking (commented):
Apologies, please revert the original diff. There is no `nil` in `nil` out; an uninitialized `downlinkTokens` at position `0` (initial state) results in this behavior and this should be guarded for by `DownlinkTokens`.

pkg/gatewayserver/io/downlink_token.go
```diff
@@ -57,7 +57,7 @@ func (t *DownlinkTokens) Next(msg *ttnpb.DownlinkMessage, time time.Time) uint16
 func (t DownlinkTokens) Get(token uint16, time time.Time) (*ttnpb.DownlinkMessage, time.Duration, bool) {
 	pos := token % downlinkTokenItems
 	item := t.items[pos]
-	if item.key != token {
+	if item.key != token || item.msg == nil {
```
adriansmares: This is fine, but you may also check if the time is 0.

adriansmares (approved):

pkg/gatewayserver/io/downlink_token.go
KrishnaIyer: Indeed. I don't see a case where that should be possible but better to do that here.

PR #4229, opened 2021-06-01
Improve cookie security

htdvisser:
#### Summary

This PR improves our cookie security by

1. No longer allowing browser credentials (cookies) in CORS requests.
2. Using `SameSite=Strict` and Secure attributes on cookies

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2760

#### Changes

- Remove the conditional around the CORS headers
- Remove the `AllowCredentials: true`
- Make CSRF skip explicit, now that we don't have a CORS skip anymore.
- Add Secure and SameSite attributes to cookies
- Various consistency fixes

#### Testing

Locally, but that doesn't actually cover cross-origin, so we still need to test that somehow.

##### Regressions

Cross-Origin requests (from cluster Console to Identity Server) may fail because of the removed `AllowCredentials: true`, but I think that should be okay because the React app sets the Authorization header, not the browser.

#### Checklist

### DIFF
--- a/pkg/account/session/session.go
+++ b/pkg/account/session/session.go
@@ -16,6 +16,7 @@ package session
 
 import (
 	"context"
+	"net/http"
 	"runtime/trace"
 	"time"
 
@@ -50,6 +51,7 @@ func (s *Session) authCookie() *cookie.Cookie {
 		Name:     authCookieName,
 		Path:     "/",
 		HTTPOnly: true,
+		SameSite: http.SameSiteLaxMode,
 	}
 }
 
--- a/pkg/web/cookie/cookie.go
+++ b/pkg/web/cookie/cookie.go
@@ -39,6 +39,24 @@ type Cookie struct {
 
 	// HTTPOnly restricts the cookie to HTTP (no javascript access).
 	HTTPOnly bool
+
+	// SameSite restricts usage of the cookie in cross-site requests.
+	SameSite http.SameSite
+}
+
+func (d *Cookie) new(r *http.Request) http.Cookie {
+	sameSite := d.SameSite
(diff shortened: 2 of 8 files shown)


pkg/web/cookie/cookie.go
```diff
@@ -97,23 +111,9 @@ func (d *Cookie) Remove(w http.ResponseWriter, r *http.Request) {
...
-	// Additionally, explicitly remove the cookie also from the current path.
-	// This can be necessary to also remove old cookies when the cookie paths
-	// have been changed.
-	http.SetCookie(w, &http.Cookie{
-		Name:     d.Name,
-		HttpOnly: d.HTTPOnly,
-		Value:    tombstone,
-		Expires:  time.Unix(1, 0),
-		MaxAge:   0,
-	})
```
htdvisser, line 118: This is no longer needed. We haven't issued such cookies for a really long time.

bafonins (approved):

pkg/account/session/session.go
```diff
@@ -50,6 +51,7 @@ func (s *Session) authCookie() *cookie.Cookie {
 		Name:     authCookieName,
 		Path:     "/",
 		HTTPOnly: true,
+		SameSite: http.SameSiteLaxMode,
```
kschiffer, line 54: So shouldn't this be `http.SameSiteStrictMode` then?

htdvisser: No, because if you navigate from the `nam1` Console to the `eu1` Account app in order to edit your account settings, you don't want to be greeted with a login screen because your browser doesn't want to send the cookie.

kschiffer (approved):

PR #4237, opened 2021-06-06
Fix PBA report deadlock

adriansmares:
#### Summary

Fix PBA deadlock when an operation (`handleUplink`/`handleDownlink`) deadlocks while attempting to report the status.
This occurs if for example the context is cancelled due to some kind of error:
- An error occurs, and the `subscribeDownlink` / `subscribeUplink` task starts to stop (i.e. cancels the context).
- The reporting workers stop, as their context expired. So does the goroutine that picks up the reports and sends them to workers.
- A downlink/uplink worker attempts to report that the context was cancelled, but the goroutine that distributes the reports has been stopped (since the context stopped).
- Deadlock, as `wg.Wait()` in the task will never finish, since the main workers cannot finish as they cannot write their reports.
#### Changes

- Select the context while writing to the report channel.
- Cancel the context in the forwarders, for symmetry reasons.

#### Testing

Unit testing.

##### Regressions

This ensures that channel writes cannot deadlock, but no other changes are expected.

#### Notes for Reviewers

If you are fine with the changes, please merge this as well and deploy it. We've seen this happen in production.

#### Checklist

### DIFF
--- a/pkg/packetbrokeragent/agent.go
+++ b/pkg/packetbrokeragent/agent.go
@@ -660,7 +660,10 @@ func (a *Agent) handleDownlinkMessage(ctx context.Context, down *packetbroker.Ro
 				Error: packetbroker.DownlinkMessageProcessingError_DOWNLINK_INTERNAL,
 			}
 		}
-		reportCh <- report
+		select {
+		case <-ctx.Done():
+		case reportCh <- report:
+		}
 	}()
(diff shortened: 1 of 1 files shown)


htdvisser (approved):

johanstokking:
Thanks for this, good catch.

PR #4245, opened 2021-06-08
Speed up end-to-end tests

kschiffer:
#### Summary
This PR will speed up our end-to-end test by introducing more caching mechanisms and also skipping entire runs if there was a passed run associated with the current file hashes.

#### Changes
- Add babel cache
- Add yarn cache
- Add SQL dump cache
- Add device repository DB cache
- Add additional jobs enabling to skip redundant runs entirely

#### Testing

I've tested this manually on my private fork: https://github.com/kschiffer/lorawan-stack/actions/workflows/e2e.yml

##### Regressions

The caching could theoretically be invalidated incorrectly and cause false positives / false negatives.

#### Notes for Reviewers

The `determine-if-required` job is set up in a way to cater also for parallelization of the end-to-end test that we are planning to setup. I've moved these checks to separate jobs so that it will work properly as well also when parallelizing the runs using the `matrix` strategy in GitHub actions. The whole mechanism works by associating passed runs with the file hashes of any file that could affect the end-to-end test and using this stored data to determine whether a run can be skipped since it was already done before with the same file hashes.

One other thing to note is that since we cache the outcome of the e2e tests based on these files:
```
hashFiles('*.go', 'go.mod', 'go.sum', 'pkg/**', 'tools/**', 'config/**', 'package.json', 'pkg/webui/**', 'sdk/js/**', 'yarn.lock', 'cypress/**', 'docker-compose.yml')
```
We will have to update this set of files in the workflow when introducing new files that could affect the outcome of the e2e tests.

#### Checklist

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -31,15 +31,15 @@ jobs:
           submodules: false
       - name: Get hash of code files
         id: get-hash
-        run: echo "::set-output name=hash::${{ hashFiles('*.go', 'go.mod', 'go.sum', 'pkg/**', 'tools/**', 'config/**', '!config/stack/ttn-lw-stack.yml', 'package.json', 'pkg/webui/**', 'sdk/js/**', 'yarn.lock', 'cypress/**', 'docker-compose.yml') }}"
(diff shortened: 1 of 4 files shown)


kschiffer:
Note that the current e2e run was still slow since nothing was cached yet, subsequent runs will be faster.

*   With only go changes, the runs will be around 12m
*   With no code changes, the runs will be around 2m
*   With JS changes, the runs will be around 20m
*   Worst case: With changes to the DB schema, go, JS and mage changes, the run can still be around 35m

bafonins (approved):

htdvisser (approved):
LGTM

Issue #4198, opened 2021-05-21
Add "last X minutes" parameter to Storage Integration API


moramos13:
#### Summary

On V2 Storage Integration API it was possible to add a "last X minutes" parameter on the API call.
...

#### Why do we need this?

This will be a good feature because currently the LNS sends all the stored uplinks to application, where are processed and the redundant data is erased. Also, to send all the messages stored on the Storage Integration to the application on every API call is a waste network resources.
...

#### What is already there? What do you see now?

Currently, only this calls can be made: 

GET /api/v3/as/applications/{end_device_ids.application_ids.application_id}/devices/{end_device_ids.device_id}/packages/storage/{type}

GET /api/v3/as/applications/{application_ids.application_id}/packages/storage/{type}
...

#### What is missing? What do you want to see?

A parameter specifying that it's only needed to retrieve messages on the "last X minutes" .
...

#### Environment

...

#### How do you propose to implement this?

...

#### How do you propose to test this?

...

#### Can you do this yourself and submit a Pull Request?

...

descartes:
Whilst I can calculate the 'after' date, many of my more simplistic original scripts pull the data on a regular basis and just use a bit of overlap - so this would be useful. It also facilitiates a "I just want to see the last couple of hours uplinks for debugging purposes" type utility.

If this gets scheduled for a much later release, in the meanwhile I'll probably end up with some sample code to do this using the after field for Python, PHP & JavaScript anyway as I migrate my toolkit

nejraselimovic:
@johanstokking what do you think about this?

johanstokking:
I can imagine that this is very useful indeed.

@neoaggelos please triage.

PR #4246, opened 2021-06-09
Query most recent application messages from Storage Integration (API + CLI)

neoaggelos:
#### Summary

API and CLI for #4198

#### Changes

- Add new `last` parameter to the `GetStoredApplicationUpRequest` message.
- Add new `--last` command line argument to CLI commands.

#### Testing

Tested CLI, API is tested along with the implementation.

##### Regressions

None

#### Notes for Reviewers

I don't think it's worth adding an `oneof` field for this one (which also breaks the API).

#### Checklist


api/applicationserver_integrations_storage.proto
```diff
@@ -50,9 +51,9 @@ message GetStoredApplicationUpRequest {
 
   // Limit number of results.
   google.protobuf.UInt32Value limit = 4;
-  // Query upstream messages after this timestamp only.
+  // Query upstream messages after this timestamp only. Cannot be used in conjunction with last.
   google.protobuf.Timestamp after = 5 [(gogoproto.nullable) = true, (gogoproto.stdtime) = true];
```
johanstokking: Maybe this is a good moment to get rid of the `gogoproto` tags here and explicitly check for empty value in code.

cmd/ttn-lw-cli/commands/storage_integration_util.go
```diff
@@ -62,6 +64,17 @@ func getStoredUpRequest(flags *pflag.FlagSet) (*ttnpb.GetStoredApplicationUpRequ
 	if err != nil {
 		return nil, err
 	}
+	if flags.Changed("last") {
+		if req.After != nil || req.Before != nil {
+			return nil, fmt.Errorf("--last cannot be used with --after or --before")
+		}
```
johanstokking: This is validation and should happen before getting and using the flag values.

api/applicationserver_integrations_storage.proto
```diff
@@ -65,6 +66,9 @@ message GetStoredApplicationUpRequest {
   // The names of the upstream message fields that should be returned. See the API reference
   // for allowed field names for each type of upstream message.
   google.protobuf.FieldMask field_mask = 9 [(gogoproto.nullable) = false];
+
+  // Query upstream messages that have arrived in the last minutes or hours. Cannot be used in conjunction with after and before.
+  google.protobuf.Duration last = 10;
```
htdvisser, line 70: You can enforce that "either/or" constraint by putting it in a `oneof`.

htdvisser (commented):
I'm not a big fan of this. I think that our API should have one way to do something. I really don't see why we can't expect applications to keep track of the last time they fetched messages from storage, and use that in the `after`.

Other than that, ok.

api/applicationserver_integrations_storage.proto
johanstokking: How to keep that compatible? It's a oneof with two fields (`after` and `before`) vs one field (`last`).

johanstokking:
So if I understand correctly, `last` is basically setting `after = now() - last`?

I see only two minor advantages of adding this:
- Alignment with the V2 storage integration, which (only) supported this method of time querying
- Allowing (stateless) clients not to provide an absolute time, which may cause issues with clock drift and messing up timezones

I'm okay adding this, just good to be aware of the above.

api/applicationserver_integrations_storage.proto
htdvisser: Good point. I guess it'll just be messy 😞

htdvisser:
1. The goal of V3 was never to have the same API as V2. 
2. If applications are fine with getting "the last 10 minutes or so" I'm sure they won't break if they have a clock drift of a couple hundred milliseconds.
3. We use timestamps and UTC everywhere in our API, why should we suddenly do something different here?

johanstokking:
Yeah I agree, therefore _minor_ advantages.

It's really arguments like mentioned here: https://github.com/TheThingsNetwork/lorawan-stack/issues/4198#issuecomment-845810946. Basic developer experience improvements.

So I'm still happy to get this in.

htdvisser:
I think the only reason to do this is to let lazy developers be lazy. It's really not that much work to do `after = now() - last`.

As the comment you linked already says:

> If this gets scheduled for a much later release, in the meanwhile I'll probably end up with some sample code to do this using the after field for Python, PHP & JavaScript anyway as I migrate my toolkit

So I'm still against.

cmd/ttn-lw-cli/commands/storage_integration_util.go
```diff
@@ -56,24 +56,37 @@ func getStoredUpRequest(flags *pflag.FlagSet) (*ttnpb.GetStoredApplicationUpRequ
 	var err error
 	req := &ttnpb.GetStoredApplicationUpRequest{}
 
-	req.After, err = getTimestampFlags(flags, "after")
+	if flags.Changed("last") && hasTimestampFlags(flags, "after") || hasTimestampFlags(flags, "before") {
```
johanstokking: Operator precedence doesn't work this way

cmd/ttn-lw-cli/commands/storage_integration_util.go
```diff
@@ -54,14 +56,38 @@ func getStoredUpRequest(flags *pflag.FlagSet) (*ttnpb.GetStoredApplicationUpRequ
...
+			return nil, err
+		}
+	}
+
+	if flags.Changed("last") {
+		d, err := flags.GetDuration("last")
+		if err != nil {
+			return nil, err
+		}
+		req.After, err = pbtypes.TimestampProto(time.Now().Add(-d))
```
johanstokking: So now we have a last, but we don't use it in the CLI? Why not?

neoaggelos: Well, the CLI is not lazy enough to use `last` instead of just setting `req.After`.

Most importantly, this works with earlier server versions as well (where using `last` would silently be ignored, leading to more user confusion).

johanstokking: Why is using `last` lazy? The key difference is relative time vs absolute time; with `last`, like the user intended with `--last`, the server time is the reference.

As for the compatibility argument, it is natural that new functionality cannot be used on old versions. Our compatibility commitment does not provide a best effort approach to work-around new functionality. So, using a new CLI with old TTS, and passing `--last` is as ineffective as using the old CLI without `--last`.

neoaggelos: My concern is more about user confusion rather than compatibility requirements. Using `last` is fine by me, updated accordingly.

johanstokking (approved):

adriansmares (approved):

PR #4256, opened 2021-06-10
Clear EUI in gateway rights query

KrishnaIyer:
#### Summary

Clear EUI in gateway rights query. This fixes a hard to track bug which prevented the Gateway EUI from being editable.

#### Changes

- Set the EUI to nil before querying a Gateway while checking rights.

#### Testing

Manually tested with a local stack. 

##### Regressions

I don't think there's any but @htdvisser knows best.

#### Checklist

### DIFF
--- a/pkg/identityserver/rights.go
+++ b/pkg/identityserver/rights.go
@@ -163,7 +163,9 @@ func (is *IdentityServer) ClientRights(ctx context.Context, cliIDs ttnpb.ClientI
 }
 
 // GatewayRights returns the rights the caller has on the given gateway.
+// The query for the gateway only considers the Gateway ID and not the EUI (if provided).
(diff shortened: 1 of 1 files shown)


kschiffer (approved):

pkg/identityserver/rights.go
```diff
@@ -168,6 +168,9 @@ func (is *IdentityServer) GatewayRights(ctx context.Context, gtwIDs ttnpb.Gatewa
 	if err != nil {
 		return nil, err
 	}
+	// Clear the EUI to ensure that the query only considers the Gateway ID.
```
htdvisser: I think I'd prefer to have even before `is.getRights`. Even though `getRights` doesn't really look at it, I think we should avoid any confusion.

PR #4260, opened 2021-06-11
Remove usage of isolated yarn dependency in tooling

kschiffer:
#### Summary
Quickfix PR to remove the usage of `yarn` as dev dependency. 

#### Changes
- Remove separate `yarn` install from mage tooling
- Remove yarn as dev dependency from `package.json`

#### Testing

Manual and CI

#### Notes for Reviewers

Iirc, we used to fix the yarn version before, due to issues with inconsistent file hashes between builds. We don't store the `public` files of the webui as part of the binaries anymore, so it should not be an issue anymore. Most important is that hashes stay the same for the same input files for builds in the same environment. This should be the case as long as the yarn version remains the same.

Otherwise, I see no reason to complicate the build process by installing and running a separate `yarn` package, which also slows (fresh) builds by about 2 minutes. Note also that `yarn` is already [mentioned](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.13/DEVELOPMENT.md#development-environment) as prerequisite for running the build tooling in DEVELOPMENT.md

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,8 +72,7 @@
     "webpack-dev-server": "^3.11.0",
     "webpack-plugin-hash-output": "^3.1.0",
     "webpack-shell-plugin": "https://github.com/cdeutsch/webpack-shell-plugin.git#bee537d",
-    "yargs": "^16.2.0",
-    "yarn": "^1.22.10"
+    "yargs": "^16.2.0"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


htdvisser (approved):
LGTM

michalborkowski96 (approved):

bafonins (approved):

PR #4264, opened 2021-06-11
Add frontend caches to release workflow

### LINKED ISSUE #3480: Speed up e2e execution in CI
#### Summary

Currently, e2e tests already take ~16minutes to execute (~21mins with several end device e2e tests, see https://github.com/TheThingsNetwork/lorawan-stack/pull/3065). Taking into account that most of the console/oauth app is still lacking any tests. Later with more e2e tests CI pipeline will take even more time. 

#### Why do we need this?

Make sure we do not end up with long CI execution because of e2e tests.

A bunch of e2e tests and e2e workflow that executes all tests on almost any PR.

Reduced CI time for e2e tests execution.

1.
* execute e2e tests in parallel
	- consider [Cypress Dashboard](https://www.cypress.io/dashboard/)
* split e2e tests into groups and run them in separate jobs (e.g. run console, oauth and smoke tests separately)
* Only run smoke tests on pull requests and execute all e2e tests on releases

2. ✅ Cache or/and speed up `js:build` target https://github.com/TheThingsNetwork/lorawan-stack/issues/2661
3. ✅ Optimise `cy.loginConsole` to avoid going through the whole oauth flow to get the auth token. https://github.com/TheThingsNetwork/lorawan-stack/pull/3978

CI

yes

kschiffer:
#### Summary
This quickfix PR adds caching for frontend builds to the snapshot github action workflow. This will speed up builds as well as provide caches for PR builds.

Closes #3480 

#### Changes
- Add public folder cache to `release-snapshot` workflow
- Add babel cache to `release-snapshot` workflow
- Simplify frontend steps in `release-snapshot` workflow
- Simplify cache key for public folder

#### Checklist

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -15,6 +15,7 @@ on:
       - 'yarn.lock'
       - 'cypress/**'
       - 'docker-compose.yml'
+      - '.github/workflows/e2e.yml'
 
 jobs:
   determine-if-required:
@@ -77,20 +78,6 @@ jobs:
           key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
           restore-keys: |
             ${{ runner.os }}-yarn-
-      - name: Initialize babel cache
-        id: babel-cache
-        uses: actions/cache@v2
-        with:
-          path: .cache/babel
-          key: ${{ runner.os }}-babel-cache-${{ hashFiles('config/babel.config.json', 'config/webpack.config.babel.js') }}
-          restore-keys: |
-            ${{ runner.os }}-babel-cache-
-      - name: Initialize public folder cache
-        id: public-cache
-        uses: actions/cache@v2
-        with:
-          path: public
-          key: public-cache-${{ hashFiles('pkg/webui/**') }}-${{ hashFiles('sdk/js/**/*.js', 'sdk/js/generated/*.json') }}-${{ hashFiles('config/webpack.config.babel.js') }}-${{ hashFiles('yarn.lock', 'sdk/js/yarn.lock')}}
       - name: Initialize SQL dump cache
         id: db-cache
         uses: actions/cache@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e.yml
```diff
@@ -90,7 +90,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: public
-          key: public-cache-${{ hashFiles('pkg/webui/**') }}-${{ hashFiles('sdk/js/**/*.js', 'sdk/js/generated/*.json') }}-${{ hashFiles('config/webpack.config.babel.js') }}-${{ hashFiles('yarn.lock', 'sdk/js/yarn.lock')}}
```
htdvisser, line 93: Would it make sense to move this step a little closer to where it's actually used? So above L145?

.github/workflows/e2e.yml
```diff
@@ -15,6 +15,7 @@ on:
       - 'yarn.lock'
       - 'cypress/**'
       - 'docker-compose.yml'
+      - '.github/e2e.yml'
```
htdvisser:
```suggestion
      - '.github/workflows/e2e.yml'
```

.github/workflows/release-snapshot.yml
```diff
@@ -42,6 +42,20 @@ jobs:
...
           ${{ runner.os }}-yarn-
+    - name: Initialize babel cache
+      id: babel-cache
+      uses: actions/cache@v2
+      with:
+        path: .cache/babel
+        key: ${{ runner.os }}-babel-cache-${{ hashFiles('config/babel.config.json', 'config/webpack.config.babel.js') }}
+        restore-keys: |
+          ${{ runner.os }}-babel-cache-
+    - name: Initialize public folder cache
```
htdvisser: Would it make sense to move this step a little closer to where it's actually used? So above `name: Build frontend`?

htdvisser (commented):
Would it make sense to also add this caching to the `release-tag` workflow?

.github/workflows/e2e.yml
kschiffer: Other than a cosmetic improvement it could save a couple seconds when an earlier step fails. :shrug:

htdvisser: Yeah, it would indeed just be to keep related things together.

kschiffer:
@htdvisser I've moved the caching steps and also introduced some simplifications for the frontend build steps of the release workflow. Also my editor suggested a different indentation, which I believe is correct. You should set GitHub to hide whitespace changes for the review.

.github/workflows/release-snapshot.yml
```diff
@@ -13,105 +13,113 @@ jobs:
...
+        id: babel-cache
+        uses: actions/cache@v2
+        if: steps.public-cache.outputs.cache-hit != 'true'
+        with:
+          path: .cache/babel
+          key: ${{ runner.os }}-babel-cache-${{ hashFiles('config/babel.config.json', 'config/webpack.config.babel.js') }}
+          restore-keys: |
+            ${{ runner.os }}-babel-cache-
+      - name: Build frontend
+        run: tools/bin/mage js:build
```
htdvisser, line 103: What happens here if the cache doesn't have a direct hit, but is still able to restore another build? Wouldn't omitting the `clean` result in old files accumulating in the public folder?

kschiffer: The public cache has no restore keys so it will only work with direct hits (in which case the `js:build` step is not run anyway). The babel cache will use the `.cache` folder which is picked up by babel when transpiling the modules. There will be no effects for the `public` folder. It is also necessary to omit the `js:clean` command to leverage the babel cache, which is otherwise deleted by that command.

htdvisser (approved):

Issue #3741, opened 2021-02-04
Webhook doesn't work for "Downlink sent" option


TheIronNinja:
#### Summary
Webhook doesn't work for "Downlink sent" option

#### Steps to Reproduce
Connect the application to a controled webhook so you can see which packets are sent and which not.
Using an existing device and gateway, simulate or generate a downlink.

#### What do you see now?
The "Downlink Scheduled" packet is sent, but once it's sent the webhook isn't triggered. I can see that the packet is sent from the Console view of the device and from the live view of the Gateway, so I know the packet is being sent and received.

#### What do you want to see instead?
The webhook being called whenever a Downlink is sent from the gateway.

#### Environment
The Things Stack v3.10.0 (I've checked the changelog and this hasn't been addressed yet so it should still be relevant)

KrishnaIyer:
@neoaggelos : Do we already support sending TX acks to the application server?

adriansmares:
This needs https://github.com/TheThingsNetwork/lorawan-stack/issues/56, which is blocked on https://github.com/TheThingsNetwork/lorawan-stack/issues/76 due to priorities.

neoaggelos:
Closed with https://github.com/TheThingsNetwork/lorawan-stack/pull/4282.

PR #4282, opened 2021-06-17
Send ApplicationDownlinkSent upstream messages from TxAcks

### LINKED ISSUE #56: Send ApplicationDownlinkSent event from Tx acknowledgement

neoaggelos:
#### Summary

Closes #56, https://github.com/TheThingsNetwork/lorawan-stack/issues/3741

#### Changes

- Add `SessionKeyId` field to `DownlinkMessage` messages.
- NS stores "in-flight" downlink messages.
- When a TxAck is received from the GS, it tries to match with the original scheduled downlink. If found, an `ApplicationDownlinkSent` message is forwarded to the application.

#### Testing

Unit tests for downlink matching.

Test with actual device and verify that ApplicationDownlinkSent messages are received.

#### Notes for Reviewers

NS stores scheduled downlinks with a TTL of 60 seconds by default. I am not sure if this is too high or low.

#### Checklist


pkg/networkserver/redis/scheduled_downlink_matcher.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/go-redis/redis/v8"
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/networkserver/internal/time"
+	ttnredis "go.thethings.network/lorawan-stack/v3/pkg/redis"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+// Duration before metadata for scheduled downlinks expire. TxAcknowledgements received
+// after this TTL will be considered invalid.
+const ttl = time.Minute
```
johanstokking: This variable is named too generically for this package. Also, godoc is invalid.

johanstokking (approved):
LGTM

Please wait for @adriansmares's approval too

pkg/networkserver/downlink.go
```diff
@@ -1241,6 +1253,7 @@ func (ns *NetworkServer) attemptClassADataDownlink(ctx context.Context, dev *ttn
 			Payload:              genDown.Payload,
 			RawPayload:           genDown.RawPayload,
 			DownlinkEvents:       genState.EventBuilders,
+			SessionKeyID:         dev.GetSession().GetSessionKeyID(),
```
adriansmares: This is correct, but I would have loved if we made the session key ID part of `genDown` (So add it to `generatedDownlink`).

pkg/networkserver/grpc_gsns.go
```diff
@@ -1276,6 +1276,37 @@ func (ns *NetworkServer) HandleUplink(ctx context.Context, up *ttnpb.UplinkMessa
...
 // ReportTxAcknowledgment is called by the Gateway Server when a tx acknowledgment arrives.
 func (ns *NetworkServer) ReportTxAcknowledgment(ctx context.Context, up *ttnpb.GatewayTxAcknowledgment) (_ *pbtypes.Empty, err error) {
-	// TODO: Send ApplicationDownlinkSent event from Tx acknowledgment (https://github.com/TheThingsNetwork/lorawan-stack/issues/56)
+	ack := up.GetTxAck()
+	if ack.GetResult() != ttnpb.TxAcknowledgment_SUCCESS {
+		return ttnpb.Empty, nil
+	}
+	down, err := ns.scheduledDownlinkMatcher.Match(ctx, ack)
+	if err != nil {
+		return nil, err
```
adriansmares, line 1289: I think we will see quite some `not_found` errors on the GS side if we report the local `not_found` error if the TTL expires. Nonetheless, I think this is not terrible.

Consider making the timeout part of the NS config - at least we can configure this in production then.

pkg/networkserver/redis/scheduled_downlink_matcher.go
```diff
@@ -0,0 +1,87 @@
...
+// ScheduledDownlinkMatcher is an implementation of networkserver.ScheduledDownlinkMatcher.
+type ScheduledDownlinkMatcher struct {
+	Redis *ttnredis.Client
+}
+
+func (m *ScheduledDownlinkMatcher) cidToKey(cid string) string {
+	return m.Redis.Key("cid", cid)
+}
+
+var parseDownlinkIDRegex = regexp.MustCompile(`^ns:downlink:([0-9a-zA-Z]+)$`)
```
adriansmares, line 41: This is the edgiest case, but what if someone intentionally adds `ns:downlink:mystaticthinghere` as a correlation ID ? Since the order of the correlation IDs is not defined, this may act weirder than it seems.

I think hashing the raw payload would be better, but I'm open to suggestions.

adriansmares (approved):
I think this works and is fine as is, but I have some small comments.

htdvisser (approved):
API addition LGTM.

adriansmares (approved):

PR #4283, opened 2021-06-17
Update connection stats with field mask paths

neoaggelos:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2655

#### Changes

- Change signature of `*io.Connection.Stats()` to return connection stats and list of field mask paths
- If stats registry already contains a proto, only the specified field mask paths are updated.

#### Testing

Existing unit tests for stats not breaking. Unit tests for presence of appropriate field mask paths.

Having separate GS instances handling uplink/downlink is very hard to reproduce locally.

##### Regressions

There should not be any.

#### Notes for Reviewers

- `sub_bands` is only added if GS instance is handling downlink (otherwise the instance that handles uplink would always overwrite with zero values). If we ever have utilization statistics for uplink, we probably have to revisit this.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -755,20 +755,21 @@ func (gs *GatewayServer) updateConnStats(conn connectionEntry) {
 	ctx := conn.Context()
 	logger := log.FromContext(ctx)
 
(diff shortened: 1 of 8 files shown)


johanstokking (approved):
Great, thanks

PR #4286, opened 2021-06-18
Avoid `SameSite=None` in cookies when not using TLS

kschiffer:
#### Summary
This quickfix will fix OAuth login for connections that don't use TLS (e.g. in development). Cookies with `SameSite=None` have to use the `Secure` flag, otherwise they can be rejected by browsers. `Secure` can however only be set when the connection uses TLS.

#### Changes
- Use `SameSite=Lax` when not using TLS in the OAuth state cookie
- Log a warning when cookies are set with `SameSite=None` without using TLS.

#### Testing

Manual testing.

#### Checklist

### DIFF
--- a/pkg/web/cookie/cookie.go
+++ b/pkg/web/cookie/cookie.go
@@ -18,6 +18,7 @@ import (
 	"net/http"
 	"time"
 
+	"go.thethings.network/lorawan-stack/v3/pkg/log"
 	"go.thethings.network/lorawan-stack/v3/pkg/webmiddleware"
 )
 
@@ -48,6 +49,8 @@ func (d *Cookie) new(r *http.Request) http.Cookie {
 	sameSite := d.SameSite
 	if sameSite == 0 {
 		sameSite = http.SameSiteLaxMode
+	} else if sameSite == http.SameSiteNoneMode && r.URL.Scheme == "http" {
+		log.FromContext(r.Context()).WithField("namespace", "web").Warn("SameSite=None will be rejected when not using TLS.")
 	}
 	return http.Cookie{
 		Name:     d.Name,
--- a/pkg/web/oauthclient/cookie_state.go
+++ b/pkg/web/oauthclient/cookie_state.go
@@ -35,13 +35,19 @@ func init() {
 }
 
 // StateCookie returns the cookie storing the state of the console.
-func (oc *OAuthClient) StateCookie() *cookie.Cookie {
+func (oc *OAuthClient) StateCookie(tls bool) *cookie.Cookie {
+	sameSite := http.SameSiteNoneMode
+	if !tls {
+		// SameSite=None will be rejected when not using TLS, so we have to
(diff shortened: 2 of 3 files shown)


pkg/web/cookie/cookie.go
```diff
@@ -48,6 +49,8 @@ func (d *Cookie) new(r *http.Request) http.Cookie {
 	sameSite := d.SameSite
 	if sameSite == 0 {
 		sameSite = http.SameSiteLaxMode
+	} else if sameSite == http.SameSiteNoneMode && r.URL.Scheme == "http" {
```
johanstokking, line 52: Isn't the front proxy using `http` to backend services? Won't that be seen as `http` here?

johanstokking (commented):
How can we avoid these regressions next time?

pkg/web/cookie/cookie.go
htdvisser: Proxies set `X-Forwarded-Proto` and `X-Forwarded-Host` headers which are extracted by the `ProxyHeaders` middleware. Our `RequestURL` middleware then fills the scheme and host in `r.URL`.

kschiffer: Hm. So how to address this issue then?

htdvisser: After giving this some more thought, I don't think we should "magically" downgrade security because a browser is (for whatever reason) using HTTP instead of HTTPS. If any other package tells `pkg/web/cookie` that it wants `SameSite=None`, we should respect that. So let's not make this change here.

pkg/web/oauthclient/cookie_state.go
```diff
@@ -35,13 +35,19 @@ func init() {
...
 
 // StateCookie returns the cookie storing the state of the console.
-func (oc *OAuthClient) StateCookie() *cookie.Cookie {
+func (oc *OAuthClient) StateCookie(tls bool) *cookie.Cookie {
+	sameSite := http.SameSiteNoneMode
+	if !tls {
+		// SameSite=None will be rejected when not using TLS, so we have to
+		// use SameSite=Lax in this case.
+		sameSite = http.SameSiteLaxMode
+	}
```
htdvisser, line 44: Again, I don't think this should be "magic", so I'd prefer to make this a `CrossSiteCookie bool` in `pkg/web/oauthclient.Config`. It can be `false` by default, and we can set it to `true` in multi-cluster deployments and for OIDC.

htdvisser (changes requested):

pkg/web/cookie/cookie.go
kschiffer: But what to do then for non TLS connections? Logging into the console is not possible currently due to this without TLS.

htdvisser: At the moment, the only place where we use `SameSite=None` is the OAuth client, and I've already suggested a solution for that.

Issue #3522, opened 2020-11-24
New end device registration flow (Console)


kschiffer:
#### Summary
The plan of adding the device repository (#3433) and the progressive insight from our current device registration flow in the Console, as well as insights from user interviews, give us the means to rethink device registration as a whole.

![image](https://user-images.githubusercontent.com/5710611/100092494-716d8480-2e56-11eb-9cab-18709d8df79d.png)
(See all annotated wireframes [here](https://github.com/TheThingsNetwork/lorawan-stack/files/5589482/201124_end_device_creation_update.pdf))
#### Why do we need this?
- Current end device registration flow is problematic
  - We use a staged disclosure (wizard pattern) which;
    - does not follow the users mental model of device registration
    - does not allow for quick review/verification of input from earlier stages
  - We overwhelm users with too many options, some of which are not relevant during device registration
  - We don't provide enough context and background on the form fields, making it hard for new users to do simple onboarding tasks
- We will add a new mode of device registration which is powered by the device repository (#3433)
- We will however also keep supporting manual device registration, to allow for registration of devices which are not part of the DR

#### What is already there? What do you see now?
- Current device registration flow
- The [new device repository](https://github.com/TheThingsNetwork/lorawan-devices) and WIP on [integrating it with TTS](https://github.com/TheThingsNetwork/lorawan-stack/issues/3433)
- UX research findings around end device registration
- Specification of the new flow (wireframes)

#### What is missing? What do you want to see?
- Implementation of necessary API to retrieve device repository info (currently WIP via #3433)
- Implementation of the new end device registration flow in the console, which entails two parts
  - Implementation of the new DR flow
  - Implementation of the new manual device registration flow

#### How do you propose to implement this?
Please see the wireframes here:
[201124_end_device_creation_update.pdf](https://github.com/TheThingsNetwork/lorawan-stack/files/5589482/201124_end_device_creation_update.pdf)

These should give a detailed plan on how the new flow is going to work. As you might have noticed, there are some overarching changes on how forms work, namely:
- We will retire the horizontal form layout and use the vertical layout throughout (label above field)
- We will remove most field descriptions and placeholders
- We will add tooltip toggles for all LoRaWAN fields and many other fields that need explanation for novice users
- We will move action buttons for fields outside the field and use a text label instead of icons

[...]

Issue #3945, opened 2021-03-18
Remove titles for checkbox inputs


kschiffer:
#### Summary
We moved to a vertical form layout via #3529. With this, it's no longer necessary to use titles for checkboxes but to only use the label.
#### Why do we need this?
This will reduce the cognitive load of forms and also shrink them in size.

#### What is already there? What do you see now?
Checkboxes like these:
![image](https://user-images.githubusercontent.com/5710611/111611224-817deb00-87dc-11eb-9fe7-5ee1597d7fc6.png)

#### What is missing? What do you want to see?
Checkboxes like these:
![image](https://user-images.githubusercontent.com/5710611/111611782-1ed91f00-87dd-11eb-9e5f-93d8ad9ae84d.png)

#### How do you propose to implement this?
Trivial. Mostly this comes down to replacing the `title` prop with the `label` prop. However the wording needs to be double checked. It needs to be in imperative tense and describe what the result of checking the checkbox is. To help with this, take the following sentence and complete it with the checkbox label:
`When this option is enabled, we will …`
e.g. `When this option is enabled, we will -> Enforce duty cycles`

#### How do you propose to test this?
Not necessary, this is not a functional change. We'll likely have to update some of the visual e2e tests with new labels.

#### Can you do this yourself and submit a Pull Request?
@asmulko please take the lead.

kschiffer:
While at it, please also check the descriptions of checkbox inputs. They are sometimes misleading, e.g. 

![](https://user-images.githubusercontent.com/5710611/111613091-72983800-87de-11eb-9835-2e3658710959.png)

This could be misunderstood in a way that this is already activated. This should rather be `Controls whether the status of this gateway may be publicly displayed`. I think there are more of such issues throughout our forms.

kschiffer:
Note that wrt to tooltips, this also means that since tooltips are usually displayed next to the field title, that we need to add a way for tooltips to be displayed next to the checkbox label instead.

NicolasMrad:
this has to be done individually while updating the forms. it cannot be generalised that easy, as mentioned by @kschiffer

PR #4291, opened 2021-06-21
Refactor end device creation form in the Console

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/3522

<details>

without activation
abp-advanced
multicast
multicast-advanced
abp
otaa-advanced
otaa

</details>

#### Changes

- Remove `<Wizard />` from device creation form
- Implement the new manual form according to wireframes in https://github.com/TheThingsNetwork/lorawan-stack/issues/3522

#### Testing

Manual and e2e tests

#### Notes for Reviewers

This PR implements the new design of the creation form. Some features from #3522 are not implemented.
I tried to incorporate the existing form fields with the new design, please let me know if changes are required.

#### Checklist


kschiffer:
Very nice work! Before going through the code, here are already some remarks from testing this:

*   The device ID should be prefilled with the dev EUI, like it does with DR onboarding
*   The class C timeout field is missing when selecting Class C or Class B and C
*   The layout for factory preset frequencies is a bit off (the delete button falls off to the right)
*   Let's add missing tooltips for the few fields that don't have one yet, @benolayinka can you provide the content? Does not need to be in this PR though unless timing allows it.
*   Please always use the new checkbox style, which is without `title` prop and only use `label`. This will result in a layout like in the wireframes. See also #3945.
*   `Do not configure activation` is currently not working well, since it removes also the fields above it, which may not happen. I unfortunately missed this issue in the wireframes. After talking about this with @johanstokking, we can actually remove this option entirely, since it is an edge-case that should really be handled via API or CLI rather. Providing this option is also suggests that not configuring activation is an encouraged flow, which it isn't.

I'll follow up with some other minor styling fine-tuning suggestions and code review.

johanstokking:
> * After talking about this with @johanstokking, we can actually remove this option entirely, since it is an edge-case that should really be handled via API or CLI rather. Providing this option is also suggests that not configuring activation is an encouraged flow, which it isn't.

Agreed

benolayinka:
> * The device ID should be prefilled with the dev EUI, like it does with DR onboarding

I still think the user should first be asked to choose a device name, and then the ID should be prefilled by slugifying the name. otherwise, the name field is useless, because the user has already chosen a semantically meaningful id. https://www.thethingsnetwork.org/forum/t/discussion-point-v3-console-is-too-complicated/47282/51?u=benolayinka

if you do keep it the way it is, at least prefill the ID with the DevEUI character by character (i.e onChange), because otherwise it's not immediately clear when the user leaves the DevEUI field why characters suddenly appear in the ID field.

bafonins:
@kschiffer Resolved most of your comments.

> Do not configure activation is currently not working well, since it removes also the fields above it, which may not happen. I unfortunately missed this issue in the wireframes. After talking about this with @johanstokking, we can actually remove this option entirely, since it is an edge-case that should really be handled via API or CLI rather. Providing this option is also suggests that not configuring activation is an encouraged flow, which it isn't.

How this should work then? Should we disable device creation when NS and JS are disabled?

@benolayinka Can you provide missing tooltip details?

kschiffer:
> How this should work then? Should we disable device creation when NS and JS are disabled?

I think that would imply this too, yes. Agreed, @johanstokking ?

We could otherwise allow this option only when NS and JS are disabled. That means that the option would be selected by default, so that also the FP and version fields are hidden by default. In that case that would not be a problem since they would be hidden from the start and also cannot reappear since there would be no other activation option available.

cypress/integration/console/devices/create-manual.spec.js
```diff
@@ -12,17 +12,9 @@
...
-import {
-  ConfigurationStep,
-  BasicSettingsStep,
-  NetworkLayerStep,
-  JoinSettingsStep,
-  ApplicationLayerStep,
-} from './utils'
-
-describe('End device manual create', () => {
+describe('End device manual create 2', () => {
```
kschiffer:
```suggestion
describe('End device manual create', () => {
```

pkg/webui/console/views/device-add/manual/form/validation-schema.js
```diff
@@ -0,0 +1,447 @@
...
+        factory_preset_frequencies: Yup.lazy(frequencies => {
+          if (!Boolean(frequencies)) {
+            return Yup.array().strip()
+          }
+
+          return Yup.array()
+            .default([])
+            .test(
+              'is-valid-frequency',
+              sharedMessages.validateFreqNumberic,
```
kschiffer:
```suggestion
              sharedMessages.validateFreqNumeric,
```

pkg/webui/console/views/device-add/manual/form/validation-schema.js
```diff
@@ -0,0 +1,447 @@
...
+              return Yup.string().required(sharedMessages.validateRequired)
+            }
+
+            return Yup.string()
+          }
+
+          return Yup.string().strip()
+        }),
+        class_b_timeout: Yup.lazy(value => {
+          const hasClassBTimeout = isClassB || mode === ACTIVATION_MODES.MULTICAST
```
kschiffer: Why is this dependant on multicast? Multicast could also mean just Class C, in which case this field would be stripped.

pkg/webui/console/views/device-add/manual/form/validation-schema.js
```diff
@@ -0,0 +1,447 @@
...
+            return Yup.number().strip()
+          }
+
+          return Yup.number()
+            .min(0, Yup.passValues(sharedMessages.validateNumberGte))
+            .max(15, Yup.passValues(sharedMessages.validateNumberLte))
+        }),
+        ping_slot_periodicity: Yup.lazy(value => {
+          if (isClassB) {
+            if (mode === ACTIVATION_MODES.MULTICAST || mode === ACTIVATION_MODES.ABP) {
```
kschiffer, line 232:
```suggestion
            if (mode === ACTIVATION_MODES.ABP) {
```

I might miss something but multicast should be irrelevant here since it will already imply that either Class B or Class C or both are set.

pkg/webui/console/views/device-add/manual/form/validation-schema.js
```diff
@@ -0,0 +1,447 @@
...
+  ),
+  multicast: Yup.boolean()
+    .transform(() => undefined)
+    .when(['$nsEnabled', '_activation_mode'], (nsEnabled, activationMode, schema) => {
+      if (!nsEnabled || activationMode === ACTIVATION_MODES.NONE) {
+        return schema.strip()
+      }
+
+      if (activationMode === ACTIVATION_MODES.OTAA || activationMode === ACTIVATION_MODES.ABP) {
+        return schema.transform(() => undefined).default(false)
```
kschiffer, line 307: Just curious, why is this transform necessary?

pkg/webui/console/views/device-add/manual/form/advanced-settings.js
```diff
@@ -0,0 +1,281 @@
...
+        <Radio
+          label={sharedMessages.multicast}
+          value={ACTIVATION_MODES.MULTICAST}
+          disabled={!nsEnabled}
+        />
+        <Radio label={messages.activationModeNone} value={ACTIVATION_MODES.NONE} />
+      </Form.Field>
+      {!isNone && (
+        <Form.Field
+          title={messages.classCapabilities}
```
kschiffer: Note that in case of multicast, the field title needs to change to `LoRaWAN Class for multicast downlinks`.

kschiffer:
> I still think the user should first be asked to choose a device name, and then the ID should be prefilled by slugifying the name. otherwise, the name field is useless, because the user has already chosen a semantically meaningful id.

Note that both the repository-based and manual device registration forms do not use the `Name` field anymore.

> if you do keep it the way it is, at least prefill the ID with the DevEUI character by character (i.e onChange), because otherwise it's not immediately clear when the user leaves the DevEUI field why characters suddenly appear in the ID field.

I would find this to be distracting when entering the Dev EUI. Let's also do the following to improve clarity:

*   Use a prefix `eui-` for the generated device ID value, which should make the relation clearer
*   Use a field description (not tooltip) stating `This value is automatically prefilled using the DevEUI`
*   Add to the tooltip text: `We prefill this value using the previously entered DevEUI but you can use any other unique value you want.`
*   Automatically text-select the entire value on focus (via mouse click) if it uses a prefilled value

Just to be clear, my main rationale for prefilling this value is improving form efficiency while suggesting sane defaults. If this turns out to be problematic still, we can instead use a button to generate the device ID value in a separate PR.

kschiffer (changes requested):
Some other things I noticed:

- Let's open an issue about hard-coding (or somehow sourcing) NS default values so they can be suggested as mac settings
- The field `Use network's Rx default settings` is missing
- I believe the multicast option should also enable to set Rx values, since it effectively means ABP, but I might be wrong cc @johanstokking 

Let's make sure we track missing implementation bits in separate issues.

I think it would be very beneficial if either @johanstokking or @htdvisser could run this branch and double check the form functionality and especially the "Advanced settings" section, so we have some extra eye-pairs on it.

pkg/webui/console/views/device-add/manual/form/validation-schema.js
bafonins: That is to add the `required` requirement here for abp and multicast class B end devices.

benolayinka:
here are the tooltips I saw missing, with slight improvements for existing ones (lwVersion and rpVersion)

devEuiLocation: For pre-provisioned devices, the DevEUI must be provided by the manufacturer. It is usually printed on the packaging. For programmable devices, you may generate a random EUI, and program the same EUI in the device. Be aware that the DevEUI must be unique within your tenant.
devEuiAbsence: For pre-provisioned devices, the DevEUI is hardcoded and must be received from the manufacturer. For programmable devices, you may generate a random EUI. Be aware that the DevEUI must be unique within your tenant.

joinEuiLocation: For pre-provisioned devices, the JoinEUI must be provided by the end device manufacturer or by the owner of the Join Server you will use. For programmable devices, use 0000000000000000, and be sure to enter the same JoinEUI in the device.
joinEuiAbsence: For pre-provisioned devices, the JoinEUI is hardcoded and must be received from the manufacturer. If your device is programmable, use 0000000000000000, and be sure to enter the same JoinEUI in the device.

applicationIdDescription: A unique identifier for your application.
applicationIdLocation: You can make it up, so be creative. IDs can not be reused by multiple applications.'

rx1DelayDescription: The amount of time after an uplink that the downlink window opens.
rx1DelayLocation: In The Things Stack, the default is 5 seconds, to accomodate for high-latency backhauls and/or Packet Broker.

rx2DataRateDescription: The data rate used for the RX2 window.
rx2DataRateLocation: For OTAA devices, this is configured as part of join. For ABP devices, a matching value must be programmed in the device.

classBTimeoutDescription: The amount of time after which the network server will assume a message is lost, if not confirmed.
classBTimeoutLocation: This should be set to a value less than the time between two ping slots (ping slot periodicity).

pingSlotPeriodicityDescription: The amount of time between two receive windows (ping slots).

lwVersionAbsence: Contact your manufacturer or reseller, since specifying the wrong version can lead to complex issues when the Network Server provides the end device with unsupported configuration (MAC) commands. If you absolutely can not find this information, use MAC V1.0.2, the most common LoRaWAN version, but check the event logs for errors.

regParameterAbsence: Contact your manufacturer or reseller to obtain the correct specification. Specifying a wrong version can lead to complex issues when the Network Server provides the end device with unsupported configuration (MAC) commands. If you absolutely can not find this information, use PHY V1.0.2 REV B, the most common Regional Parameters, but check the event logs for errors.

pkg/webui/console/views/device-add/manual/form/validation-schema.js
bafonins: to get the default value automatically

kschiffer:
<details><summary>Please also apply these styling tweaks:</summary>

```diff
diff --git a/pkg/webui/components/form/section/index.js b/pkg/webui/components/form/section/index.js
index 3a8567585..58761da9c 100644
--- a/pkg/webui/components/form/section/index.js
+++ b/pkg/webui/components/form/section/index.js
@@ -16,9 +16,11 @@ import React from 'react'
 import classnames from 'classnames'
 import { defineMessages, useIntl } from 'react-intl'

-import Form, { useFormContext } from '@ttn-lw/components/form'
+import { useFormContext } from '@ttn-lw/components/form'
 import Icon from '@ttn-lw/components/icon'

+import Message from '@ttn-lw/lib/components/message'
+
 import PropTypes from '@ttn-lw/lib/prop-types'

 import style from './section.styl'
@@ -42,7 +44,7 @@ const FormCollapseSection = props => {
   }, [collapsed, onCollapse])

   return (
-    <div className={classnames(style.section, className)}>
+    <div className={className}>
       <button
         className={style.button}
         type="button"
@@ -52,7 +54,7 @@ const FormCollapseSection = props => {
         aria-controls={id}
         disabled={disabled}
       >
-        <Form.SubTitle className={style.title} title={title} />
+        <Message content={title} className={style.title} />
         <Icon className={style.icon} icon={collapsed ? 'expand_down' : 'expand_up'} />
       </button>
       <div className={classnames(style.content, { [style.expanded]: !collapsed })} id={id}>
diff --git a/pkg/webui/components/form/section/section.styl b/pkg/webui/components/form/section/section.styl
index 76dc83f02..7affd1fbc 100644
--- a/pkg/webui/components/form/section/section.styl
+++ b/pkg/webui/components/form/section/section.styl
@@ -12,9 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.

-.section
-  margin-bottom: $ls.xxs
-
 .button
   reset-button()
   padding: $cs.s
@@ -22,10 +19,9 @@

 .title
   margin: 0
-  margin-right: $cs.xs
   display: inline-block
-  color: $tc-black
   user-select: auto
+  font-weight: $fw.bold

 .icon
   color: $c-icon-fill
diff --git a/pkg/webui/console/components/mac-settings-section/index.js b/pkg/webui/console/components/mac-settings-section/index.js
index c53aa1799..ce39a3666 100644
--- a/pkg/webui/console/components/mac-settings-section/index.js
+++ b/pkg/webui/console/components/mac-settings-section/index.js
@@ -41,7 +41,7 @@ const m = defineMessages({
   factoryPresetFreqTitle: 'Factory Preset Frequencies',
   freqAdd: 'Add Frequency',
   frequencyPlaceholder: 'e.g. 869525000 for 869,525 MHz',
-  macSettings: 'MAC settings',
+  advancedMacSettings: 'Advanced MAC settings',
   pingSlotFrequencyDescription: 'Frequency of the class B ping slot (Hz)',
   pingSlotFrequencyTitle: 'Ping Slot Frequency',
   pingSlotPeriodicityDescription: 'Periodicity of the class B ping slot',
@@ -84,10 +84,9 @@ const MacSettingsSection = props => {
   return (
     <Form.CollapseSection
       id="mac-settings"
-      title={sharedMessages.advancedSettings}
+      title={m.advancedMacSettings}
       initiallyCollapsed={initiallyCollapsed}
     >
-      <Form.SubTitle title={m.macSettings} />
       <Form.Field
         title={sharedMessages.frameCounterWidth}
         name="mac_settings.supports_32_bit_f_cnt"
diff --git a/pkg/webui/lib/shared-messages.js b/pkg/webui/lib/shared-messages.js
index 4e19ecee1..ddd6285ac 100644
--- a/pkg/webui/lib/shared-messages.js
+++ b/pkg/webui/lib/shared-messages.js
@@ -35,7 +35,6 @@ export default defineMessages({
   addressPlaceholder: 'host',
   addWebhook: 'Add webhook',
   admin: 'Admin',
-  advancedSettings: 'Advanced settings',
   all: 'All',
   allAdmin: 'All (Admin)',
   altitude: 'Altitude',
```

</details>

johanstokking:
> > How this should work then? Should we disable device creation when NS and JS are disabled?
> 
> I think that would imply this too, yes. Agreed, @johanstokking ?

Yes, for activation, that is. For ABP, JS isn't needed.

johanstokking:
This is a great work.

I agree with @kschiffer's comments in https://github.com/TheThingsNetwork/lorawan-stack/pull/4291#issuecomment-874609211.

Other comments:
1. I overlooked the expand title **Show advanced ...** the first time. Should there be a radio button instead, that says **Use default settings (OTAA, Class A, default timings, default cluster settings)** and another radio button **Custom LoRaWAN and cluster settings** ?
2. Small heading named **Cluster settings** above **Use external LoRaWAN backend services**. The others have headings and when checking the box, this becomes really a section that deserves a small header

kschiffer:
> 1.  I overlooked the expand title **Show advanced ...** the first time. Should there be a radio button instead, that says **Use default settings (OTAA, Class A, default timings, default cluster settings)**

I think it suffices to increase the _affordance_ of this element a bit by displaying the `Advanced ...` element as a link. The problem is that is currently looks like a section heading. In a way though, overlooking this section is somewhat desired since it should only become apparent when a user is aware of specific settings and hence scanning the UI for hints on where such settings can be found. I'll provide another diff to update this.

> and another radio button **Custom LoRaWAN and cluster settings** ?

What would be hidden behind that? Class, timings and hosts? I think the way it is currently defined in the wireframes is a clear enough (with `Use network's Rx and frequency defaults` checkbox). I believe though that we can divert a little from the wireframes to make it clearer. Namely:

*   The RX settings **and** `Factory preset frequencies`  should be hidden behind a checkbox reading `Use network's Rx and frequency defaults` (note the diverted wording from the WF). This checkbox is checked by default.
*   Contrary to the wireframe, we can use the same checkbox (and underlying fields) **for OTAA as well**.
*   For mutlicast devices, the same applies, but the **Rx1 fields are removed** (no uplinks when using MC)
*   The checkbox should have a tooltip that explains about what Rx settings are and how the deployment uses certain default settings that should be applicable to most use cases and conform to the applicable RPs. cc @benolayinka 
*   Implementation-wise, checking this checkbox equals leaving all Rx fields empty
*   Also when unchecking the checkbox, it is possible to leave individual Rx fields empty, in which case the server defaults are used

Also note that some fields that have been transfered over from the old form are incorrectly capitalized, e.g. `Rx1 Data Rate Offset`, should be `Rx1 data rate offset`.

> Small heading named **Cluster settings** above **Use external LoRaWAN backend services**. The others have headings and when checking the box, this becomes really a section that deserves a small header

So simply add a title prop with `Cluster settings` to the checkbox field @bafonins. I'm ok with this in spite of #3945, since the title would function as a section title.

pkg/webui/console/views/device-add/manual/form/advanced-settings.js
```diff
@@ -43,6 +43,7 @@ const m = defineMessages({
   classC: 'Class C (Continuous)',
   classBandC: 'Class B and class C',
   useExternalServers: 'Use external LoRaWAN backend servers',
+  multicastClassCapabilities: 'LoRaWAN Class for multicast downlinks',
```
kschiffer:
```suggestion
  multicastClassCapabilities: 'LoRaWAN class for multicast downlinks',
```

kschiffer (changes requested):
Almost ready. Please make the following changes:

- Add a tooltip to the `Use network's Rx and frequency defaults` checkbox:
```
The network uses a set of default values for Rx delays and frequencies that the end device uses. These are based on the recommendations made for the respective band. In most cases these defaults will be correct for your setup. If you wish to use different settings, you can uncheck this checkbox and use custom values.
```

- Add a tooltip to the `Cluster settings` checkbox:
```
By default, the server components of the current cluster are used. However, for advanced use cases, it is possible to register this end device to different Network Server and/or Join Server.
```

- Add a tooltip to the `Factory preset frequencies` option:
```
What is it?
Factory preset frequencies are hard-coded channel frequencies to provide for the Network Server when the end device uses frequencies that divert from the defaults of the band specification. This is uncommon but can be the case for some special devices.

Where do I find this?
If your device uses non-default channel frequencies, this information is likely found in the specification sheet or manual of your end device. Otherwise please contact your manufacturer or reseller.

What if I cannot find this?
If your device uses non-default channel frequencies and these frequencies are not passed to the Network Server, the messages sent between the network and the end device on such non-default frequencies are likely to drop.
```
@johanstokking / @benolayinka please confirm these messages.

<details><summary>Please also apply this:</summary>

```diff
diff --git a/pkg/webui/components/form/section/section.styl b/pkg/webui/components/form/section/section.styl
index 7affd1fbc..1c7b7686d 100644
--- a/pkg/webui/components/form/section/section.styl
+++ b/pkg/webui/components/form/section/section.styl
@@ -21,7 +21,7 @@
   margin: 0
   display: inline-block
   user-select: auto
-  font-weight: $fw.bold
+  text-decoration: underline
 
 .icon
   color: $c-icon-fill
@@ -29,6 +29,7 @@
   user-select: none
 
 .content
+  margin-top: $ls.xxs
   display: none
   overflow: hidden
```

</details>

johanstokking:
@kschiffer sounds good to me, one typo though (Network Server missing last r)

Do note that this only applies to ABP devices really

kschiffer:
> one typo though

Fixed.

> Do note that this only applies to ABP devices really

Ok, we currently show it for both. @bafonins please hide `Factory preset frequencies` then when OTAA is selected.

htdvisser:
> Do note that this only applies to ABP devices really

I actually checked the Network Server yesterday, and it also seems to use it for OTAA devices (for the join frequencies).

i3mSuperMan:
@kschiffer /@bafonins, Is it possible to hide the `Factory preset frequencies` setting in the console for the ABP device when the **US/AU frequency plan** is selected? As per the discussion [here](https://github.com/TheThingsNetwork/lorawan-stack/issues/2269#issuecomment-810076361), we prohibit the creation/update of the `Factory preset frequencies` field for devices in **US-like bands**.

kschiffer:
Good one @i3mSuperMan, if we disallow it on the backend then it should be reflected in the backend too. Which exact bands are affected by that @adriansmares? Just all US/AU bands?

@bafonins I'm also ok with doing that in a separate PR, since we would really like to land this in 3.14.0.

Modifications should be:

*   Disable (gray out) the field for both manual and DR-based registration
*   Remove the current field description entirely
*   Add a field description (not tooltip) only when the field is disabled, reading: `For technical reasons this option is unavailable for certain US/AU bands`

johanstokking:
> > Do note that this only applies to ABP devices really
> 
> I actually checked the Network Server yesterday, and it also seems to use it for OTAA devices (for the join frequencies).

Yes we do, but we don't need Console support for this.

> As per the discussion [here](https://github.com/TheThingsNetwork/lorawan-stack/issues/2269#issuecomment-810076361), we prohibit the creation/update of the `Factory preset frequencies` field for devices in **US-like bands**.

And this is true as well. You should hide this option if the band of the frequency plan is:

- `AU_915_928`
- `CN_470_510`
- `US_902_928`

bafonins:
> Is it possible to hide the Factory preset frequencies setting in the console for the ABP device when the US/AU frequency plan is selected?

I would prefer looking into this in a separate PR

kschiffer (approved):
image
Please remove the field descriptions for `Ping slot periodicity`, `Rx2 frequency` and `Factory preset frequencies`

Note that the PR currently removes the `snapcraft.login.gpg` file

Please also double check if we can remove any more styles or components as a result of removing the old manual end device registration.

pkg/webui/console/views/device-add/manual/form/advanced-settings.js
```diff
@@ -0,0 +1,299 @@
...
+        onChange={onDeviceClassChange}
+        options={isMulticast ? multicastClassOptions : allClassOptions}
+        tooltipId={tooltipIds.CLASSES}
+      />
+      <Form.Field
+        label={messages.defaultNetworksSettings}
+        name="_default_ns_settings"
+        component={Checkbox}
+        onChange={handleDefaultNsSettings}
+        tooltipId={tooltipIds.NETWORK_RX_DEFAULTS}
```
kschiffer, line 140: Note that this tooltip is not shown because the field does not use a title. So to remedy quickly, let's use the title `Network defaults` for this. This should be replaced then as soon as we can add tooltips to checkboxes as well.

PR #4298, opened 2021-06-22
api: Add base64 note to lbs_lns_secret

### LINKED ISSUE #399: Mention "lbs_lns_secret" should be base64 encoded
#### Summary
Add to description of `lbs_lns_secret` that it should be `base64` encoded.  Ref:https://www.thethingsindustries.com/docs/reference/api/gateway/#message:Gateway
#### Why do we need this?
Mention appropriate usage in docs
Information on what is `lbs_lns_secret`
Information on encoding format/type of the secret
...
No. @benolayinka

benolayinka:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack-docs/issues/399

#### Changes

- Add note that `lbs_lns_secret` is base64 encoded.

#### Checklist

### DIFF
--- a/api/gateway.proto
+++ b/api/gateway.proto
@@ -155,7 +155,7 @@ message Gateway {
   google.protobuf.Duration schedule_anytime_delay = 19 [(gogoproto.stdduration) = true, (gogoproto.nullable) = true];
   // Update the location of this gateway from status messages. This only works for gateways connecting with authentication; gateways connected over UDP are not supported.
   bool update_location_from_status = 21;
-  // The LoRa Basics Station LNS secret.
+  // The LoRa Basics Station LNS secret, base64 encoded.
   // This is either an auth token (such as an API Key) or a TLS private certificate.
   // Requires the RIGHT_GATEWAY_READ_SECRETS for reading and RIGHT_GATEWAY_WRITE_SECRETS for updating this value.
   Secret lbs_lns_secret = 22 [(gogoproto.customname) = "LBSLNSSecret"];


api/gateway.proto
```diff
@@ -155,7 +155,7 @@ message Gateway {
   google.protobuf.Duration schedule_anytime_delay = 19 [(gogoproto.stdduration) = true, (gogoproto.nullable) = true];
   // Update the location of this gateway from status messages. This only works for gateways connecting with authentication; gateways connected over UDP are not supported.
   bool update_location_from_status = 21;
-  // The LoRa Basics Station LNS secret.
+  // The LoRa Basics Station LNS secret, base64 encoded.
```
htdvisser, line 158: This is incorrect. The `lbs_lns_secret` field is a `Secret` message, not base64.

_inside_ the `Secret` there's indeed a `bytes value` that, when using JSON, is encoded as base64. This shouldn't have to be documented in a proto comment (because it's already proto spec). 

Probably better to update the documentation template to mention base64 for any `bytes` field (there are some exceptions for fields with `gogoproto.customtype`, those we can indeed mention explicitly).

htdvisser (changes requested):

johanstokking:
What is the status here?

api/gateway.proto
benolayinka: > This is incorrect. The `lbs_lns_secret` field is a `Secret` message, not base64.
> 
> _inside_ the `Secret` there's indeed a `bytes value` that, when using JSON, is encoded as base64. This shouldn't have to be documented in a proto comment (because it's already proto spec).
> 
> Probably better to update the documentation template to mention base64 for any `bytes` field (there are some exceptions for fields with `gogoproto.customtype`, those we can indeed mention explicitly).

this won't solve this particular problem, because the definition of `lbs_lns_secret` at https://www.thethingsindustries.com/docs/reference/api/gateway/#message:Gateway has a broken link to type `Secret`. i think that this instance is not crucial, but we have to address it as part of a much needed API doc overhaul.

benolayinka:
Closed for https://github.com/TheThingsIndustries/lorawan-stack/issues/2717

PR #4301, opened 2021-06-23
Fix antenna gain

### LINKED ISSUE #2006: Support multiple gateway antennas
#### Summary

Support multiple antennas for a single gateway

#### Why do we need this?

Some gateways may have more than one antennas

Support for a single antenna per gateway. Even if multiple antennas exist, only the first one is used and the rest are ignored.

Relevant issues #48, #1820

Handle all gateway antennas.

#### Environment

3.5.3

Update code in the relevant TODO sections where this issue is referenced

Cannot pick up currently, adding to backlog

### LINKED ISSUE #4299: Antenna gain is often not considered and gets cleared
#### Summary

Antenna gain is not considered for unauthenticated gateways: all UDP, and depending on config, other gateways too.

In addition, when the gateway status message contains the gateway location, the antenna gain gets cleared.

#### Steps to Reproduce

Part A:

1. Create gateway with one antenna, configure a gain value, and configure that the gateway should update its location from status
2. Connect gateway via UDP
3. Send downlink message with maximum EIRP for the band

Part B:

1. Let gateway send GPS coordinates as part of the gateway status message
2. Get gateway antennas from IS with CLI

#### What do you see now?

Part A: the TX power is not adjusted for antenna gain
Part B: the antenna gain is gone

Part A: the antenna gain must be taken into account, also if the gateway is unauthenticated
Part B: the antenna gain should not be overwritten by location information from gateway status messages

1. Change API so that `GatewayAntenna.location` is nullable. References https://github.com/TheThingsNetwork/lorawan-stack/issues/2798
2. Let IS return the antennas and their gain, but not their location, when the call to `GatewayRegistry.Get()` isn't authenticated (i.e. if GS can't forward the gateway's authentication). This resolves part A
3. When the status message contains GPS coordinates, GS may update the antennas as long as it preserves the gain initially fetched on connect

Integration testing; this is mostly interaction between GS and IS

Yes

johanstokking:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2006
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4299

#### Changes

- Account for antenna gain when the gateway isn't authenticated
- Preserve antenna gain when the location is in the status message
- Make antenna location nullable (in case the gain is known but the location isn't)

#### Testing

Unit testing

##### Regressions

None

#### Notes for Reviewers

@htdvisser @adriansmares please only look at code owned files

@neoaggelos please review GS as @KrishnaIyer is unavailable

@bafonins @kschiffer no need to review

There's other low hanging fruit included here too

#### Checklist


htdvisser (approved):

adriansmares (approved):
Code owned files LGTM.

neoaggelos (approved):

PR #4303, opened 2021-06-23
Add option to use cross site cookies for oauth

kschiffer:
#### Summary
This quickfix adds a config option to the `oauthclient` package to allow for more fine-grained control over cookie security.
This will also solve a bug in which it is impossible to sign into the Console when not using TLS. This is due to currently using `SameSite=none` for the OAuth state cookie, which some browsers will block when not using the `Secure` flag.

#### Changes
- Add config option `CrossSiteCookies` to `oauthclient.Config`
- Add default for this option `false`
- Use option to determine `SameSite` option for the state cookie of the OAuth process

#### Testing

Manual testing.

##### Regressions

As we have seen before, changes to our cookie logic can have many hard to foresee issues. So @htdvisser please also test.

#### Checklist

### DIFF
--- a/pkg/web/oauthclient/config.go
+++ b/pkg/web/oauthclient/config.go
@@ -24,6 +24,8 @@ type Config struct {
 	ClientID     string `name:"client-id" description:"The OAuth client ID"`
 	ClientSecret string `name:"client-secret" description:"The OAuth client secret" json:"-"`
 
+	CrossSiteCookie bool `name:"cross-site-cookie" description:"Whether to make OAuth cookies accessible cross-site"`
+
 	StateCookieName string `name:"-"`
 	AuthCookieName  string `name:"-"`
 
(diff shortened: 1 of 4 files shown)


pkg/web/oauthclient/config.go
```diff
@@ -24,6 +24,8 @@ type Config struct {
 	ClientID     string `name:"client-id" description:"The OAuth client ID"`
 	ClientSecret string `name:"client-secret" description:"The OAuth client secret" json:"-"`
 
+	CrossSiteCookie bool `name:"cross-site-cookie" description:"Whether to make OAuth cookies accessible cross-site"`
```
htdvisser, line 27: This needs to be mentioned in the CHANGELOG, also the fact that multi-cluster deployments need to set this to `true` in order to support OAuth clients that use POST callbacks. Please also create an issue for setting that to `true` in https://github.com/TheThingsIndustries/lorawan-stack-aws (doesn't need to be configurable, since those deployments always use HTTPS).

htdvisser (approved):

PR #4311, opened 2021-06-24
Infer location from decoded payload and publish location solved

### LINKED ISSUE #4176: Publish location solved event based on decoded payload
#### Summary

Publish location solved event based on decoded payload.

#### Why do we need this?

To channel location solving on a single channel; the location solved application upstream message type.

This is helpful for upstream subscribers to have one source to subscribe to.

Also, this decouples application payload from location data, which is better for privacy. Currently, users have to configure the TTN Mapper integration for all their uplink message data, whereas only location data is interesting for TTN Mapper. cc @jpmeijers

When there's a geolocation integration like LoRa Cloud Geolocation for LoRa Edge devices, the `location_solved` message type is published by Application Server.

However, when the location coded in the application payload, like GPS coordinates, the location is only made available in a non-unified way through the `uplink_message` message type.

Recognize location information from decoded uplink. This is based on predefined fields in the decoded payload.

In Application Server, after decoding the payload, recognize fields like `latitude`, `longitude` (degrees), `altitude`, `hdop` and `accuracy` (meters). CayenneLPP coordinates should also be recognized. This is to be compatible with what TTN Mapper currently needs. See https://ttnmapper.org/faq/ under "How can I contribute data".

Integrations like TTN Mapper still rely on RX metadata, which is currently not part of `location_solved` messages. Therefore we cannot remove `uplink_message` from the TTN Mapper's webhook template (https://github.com/TheThingsNetwork/lorawan-webhook-templates/blob/e03f663772ffd04682f7a3029d1f16fb79ccb628/ttnmapper.yml#L23) but I believe this should be the aim on the longer term.

I'm also curious to hearing @jpmeijers his opinion on this matter.

Unit testing

Yes

johanstokking:
#### Summary

Closes #4176 

cc @jpmeijers

#### Changes

- Infer location from decoded payload. Took some inspiration from TTN Mapper
- General housekeeping in AS

#### Testing

Unit testing

##### Regressions

None expected

#### Notes for Reviewers

Only functional change is in https://github.com/TheThingsNetwork/lorawan-stack/commit/b6863bb4e73d5473d96f1f2aad17ba9e4becd635

#### Checklist


adriansmares (approved):

pkg/applicationserver/payload.go
```diff
@@ -193,3 +194,81 @@ func (as *ApplicationServer) decodeDownlink(ctx context.Context, dev *ttnpb.EndD
...
+		}{
+			{"lat", "lon", "alt"},
+			{"lat", "lng", "alt"},
+			{"lat", "long", "alt"},
+			{"latitude", "longitude", "altitude"},
+			{"Latitude", "Longitude", "Altitude"},
+			{"latitudeDeg", "longitudeDeg", "altitude"},
+			{"latitudeDeg", "longitudeDeg", "height"},
+			{"gps_lat", "gps_lng", "gps_alt"},
+			{"gps_lat", "gps_lng", "gpsalt"},
```
jpmeijers: If I understand this code correctly, this will only allow these 9 possible sets of keys, and no other combinations of them. Won't it be better to check one field at a time, trying all it's possibilities, before checking the next field?

jpmeijers: ie, say I have the fields `latitude`, `long` and `altitude`, then this algorithm won't accept it.

johanstokking: Yes indeed. I wanted to make this explicit, to not trigger magic behavior when these keys aren't really consistent (like `latitudeDeg` and `gps_lng`)

jpmeijers:
I still think we should split the parsing of the decoded payload out into a module which we can share, and I can help maintain. There are some edge cases that are currently not handled here.

For example a Digital Matter Yabby or Oyster device sends a `fixFailed` field. If this field is `true`, the location that is in the `latitude` and `longitude` fields is a cached location which can be very old. We've had issues where this caused invalid coverage to be draw - like a link between Melbourne and Sydney (700km) at -80dBm.

https://github.com/ttnmapper/ingress-api/blob/master/utils/payload_fields_parser.go#L195-L197 and https://github.com/ttnmapper/ingress-api/blob/master/utils/payload_fields_parser.go#L213-216

If we don't handle edge cases like these in the TTI location inference, the `location-solved` webhook will be useless for TTN Mapper.

pkg/applicationserver/payload.go
jpmeijers: Yes I see your point. My opinion is slightly different: I try my best to obtain a valid location, no matter how "broken" the payload parser may be.

From TTN Mapper's side we recommend a specific set of keys:
https://docs.ttnmapper.org/FAQ.html#using-a-gps-tracker-device-transmitting-gps-coordinates-in-the-lora-payload
> The JSON object should contain the keys “latitude”, “longitude” and “altitude” and one of “hdop”, “accuracy” or “sats”

Something more to note here is that sats, hdop and accuracy are handled as three different things. They are inherently three different ways to indicate how reliable the gps location is. Putting hdop into a field called "accuracy" can cause confusion, as an accuracy of 3 metres is good, but an hdop of 3 is bad.

johanstokking:
Right. I see also that this codec is in the device repository; https://github.com/TheThingsNetwork/lorawan-devices/blob/master/vendor/digital-matter/oyster.js. Now, isn't that an issue with this codec? Shouldn't it become `cachedLatitudeDeg`? I mean, what is the point of GPS coordinates if there's no fix?

jpmeijers:
> Now, isn't that an issue with this codec? Shouldn't it become `cachedLatitudeDeg`? I mean, what is the point of GPS coordinates if there's no fix?

Yes agreed. But that means we need to lecture codec writers on properly naming fields. Digital Matter isn't the first, and won't be the last that does something like this. The second example was Tabs - one that can be seen as a reference implementation by some.

johanstokking:
OK, there will be a module that we can both import. I'll change this PR to use that module.

I really want to avoid more magic in there that checks arbitrary fields whether the location is valid or not.

In the particular case of Oyster and Yabby, I would strongly prefer putting coordinates under `cached` and adding a warning that the fix failed. Filed https://github.com/TheThingsNetwork/lorawan-devices/issues/233

johanstokking:
@jpmeijers please see https://pkg.go.dev/go.thethings.network/lorawan-application-payload. I welcome any suggestions in https://github.com/TheThingsNetwork/lorawan-application-payload.

PR #4312, opened 2021-06-24
Increase size limit for payload formatter scripts

### LINKED ISSUE #4278: Available payload formatter size is too small
#### Summary
Available Payload formatter size is too small for some commonly used devices.

#### Why do we need this?
The default payload formatter's size of Elsys devices is too big for The Things Stack. Same for KLAX, Milesight and NKE devices.

The size of Elsys' payload decoder is ±1000 bytes, The Things Stack allows up to 4 kbytes. 
Reference: https://www.elsys.se/en/elsys-payload/

This was no issue in V2.

When Elsys' payload formatter is added, the below error is shown:

Screenshot 2021-06-16 at 16 08 03

```
Invalid value of field default_formatters.up_formatter_parameter

{
  "code": 3,
  "message": "error:pkg/applicationserver:field_value (invalid value of field `default_formatters.up_formatter_parameter`)",
  "details": [
    {
      "namespace": "pkg/applicationserver",
      "name": "field_value",
      "message_format": "invalid value of field `{field}`",
      "attributes": {
        "field": "default_formatters.up_formatter_parameter"
      },
      "correlation_id": "0b76c2d282e14e909081c8ec997d4778",
      "cause": {
        "namespace": "pkg/applicationserver",
        "name": "formatter_script_too_large",
        "message_format": "formatter script size exceeds maximum allowed size",
        "attributes": {
          "max_size": 4096,
          "size": 7964
        },
        "code": 3
      },
      "code": 3
    }
  ],
  "request_details": {
    "url": "/as/applications/test-application/link",
    "method": "put",
    "stack_component": "as"
  }
}
```

~Increase available size to 8 or 16 kB?~

@neoaggelos see https://github.com/TheThingsNetwork/lorawan-stack/issues/4278#issuecomment-865832090

#### Environment
v3.13.1

Get input from Jaime on the minimum required payload formatters size.

Upload payload formatters of >4 kbytes on test local setup.

Need input form Jaime and from product team.

neoaggelos:
#### Summary

References #4053 
Closes #4278

#### Testing

...

##### Regressions

...

#### Notes for Reviewers

I think I saw somewhere a standard limit of 64KB being mentioned, but went with the 40KB limit proposed in the issue.

Let's choose a value and avoid more PRs like this in the future.

#### Checklist


htdvisser (commented):
As @johanstokking wrote in https://github.com/TheThingsNetwork/lorawan-stack/issues/4278#issuecomment-864061169:

> Currently, there's a LoRa Alliance task force under TC about Codec API where a limit is being defined. I think we'll end up somewhere in the range of 32 to 40 KB.

So I think that for now we should go for 32kB, and depending on the outcome of LoRa Alliance discussion increase it to 40kB.

johanstokking (approved):
I also said https://github.com/TheThingsNetwork/lorawan-stack/issues/4278#issuecomment-865834606, echoing @adriansmares's proposal. That's implemented here.

The LoRa Alliance decision is going to take a while because we also want to settle on a library that gets pre-loaded, i.e. Node's `Buffer` polyfill, maybe some crypto functions and maybe some CRC functions. Once that is defined, we have a better idea of what is sensible for the device maker to implement themselves, and come up with a sane limit.

Now, until we have that, I would in fact relax this for now, because this is really limiting at the moment. So let's settle on 40 KB in API and 40 KB in AS allowed maximum. Whenever we have `Buffer`, crypto and CRC32, we will once again reduce the AS allowed maximum, aligning with the LoRa Alliance technical recommendation. So this is good to go as-is.

cmd/internal/shared/applicationserver/config.go
```diff
@@ -77,6 +77,6 @@ var DefaultApplicationServerConfig = applicationserver.Config{
 		},
 	},
 	Formatters: applicationserver.FormattersConfig{
-		MaxParameterLength: 4096,
+		MaxParameterLength: 32768,
```
johanstokking: So make this 40 KB too

adriansmares (approved):

PR #4314, opened 2021-06-24
Include network identifiers in upstream messages

### LINKED ISSUE #4076: Include network identifiers in upstream metadata
#### Summary

Include network identifiers in upstream metadata.

#### Why do we need this?

For subscribing networks to identify from which network the message originates. This is specifically useful for messaging based application-layer integrations, like webhooks.

Currently, when users build an HTTP endpoint to accept messages from the AS webhook integration, they have to include the originating network in the path, because it's not part of the metadata.

There is `X-Tts-Domain`, but that is intended for downlink callbacks. I would discourage using that for origin identification. Also, that is the request header and not the actual message, so if the accepting endpoint passes on the message as-is on an internal queue, this doesn't work well.

When receiving traffic that got routed via Packet Broker, the metadata includes `forwarder_net_id`, `forwarder_tenant_id` and `forwarder_cluster_id`. The community indicated that information is missing in metadata of gateways that are directly connected to the routing cluster that is handling the traffic.

I would like to see an identification of the origin network in upstream messages.

I would like to see a network identification that follows LoRaWAN or LoRa Alliance best practices as much as possible. The main identifier is the NetID, which is already configured in the Network Server.

From LoRaWAN Backend Interfaces, there's also the "NSID"; and identifier of the Network Server within the namespace of the NetID. This is intended for different NS deployments that use the same NetID: effectively the "tenant ID" in The Things Stack.

Besides NetID and tenant ID, Packet Broker distinguishes client networks by cluster ID. This is also part of the Packet Broker Agent configuration already:

https://github.com/TheThingsNetwork/lorawan-stack/blob/b9e3712f32b748b31718896b3e83e153cdc55c6f/pkg/packetbrokeragent/config.go#L31-L33

> Packet Broker will combine tenant ID and cluster ID in the NSID (`tenant-id@cluster-id`) when it gets LoRaWAN Backend Interfaces 1.1 support.

So, I would suggest that we include `net_id`, `tenant_id` and `cluster_id` in upstream messages. This aligns directly with Packet Broker metadata and it provides an, in my view, necessary classification that we would not have with only NetID + NSID.

Add `NetworkIdentifiers` message that gets added to `UplinkMessage`, set by NS based on the context, and AS copies that to `ApplicationUp`.

I'm not sure if NS already has an identifier of the cluster ID in the context or configuration, but this is needed anyhow for per-cluster tenant configuration in TTSE.

To decide is:

1. That we want this in the first place
2. That we use the terms `tenant_id` and `cluster_id`, and not a more opaque `ns_id` although that is more LoRaWAN Backend Interfaces 1.1 compliant
3. That we introduce a new identifiers message `NetworkIdentifiers`. I'm also fine with `DeploymentIdentifiers` or any other flavor. It wouldn't be related to entity identifiers.

Unit tests + some basic integration testing

Can review

neoaggelos:
#### Summary

Closes #4076 

#### Changes

- Add new `NetworkIdentifiers` message.
- Set `NetworkIdentifers` when enqueueing `ApplicationUplink` messages to the Application Server.
- Add `ns.cluster-id` configuration option.
The changes follow the proposed implementation in the linked issue.

#### Testing

Locally

##### Regressions

The uplink message metadata may be getting too much with this change.

#### Notes for Reviewers

Draft until documentation and config PRs are done.

@johanstokking For the NetID, we use `types.NetID` with Gogoproto, and Packet Broker uses `uint32` also. Added the definition as `string` (so that it is JSON marshaled to a hex string). Which type should we use here?

#### Checklist


api/identifiers.proto
```diff
@@ -108,3 +108,13 @@ message EndDeviceVersionIdentifiers {
   string firmware_version = 4 [(validate.rules).string.max_len = 32];
   string band_id = 5 [(validate.rules).string.max_len = 32, (gogoproto.customname) = "BandID"];
 }
+
+// Identifies a Network Server.
+message NetworkIdentifiers {
+  // LoRa Alliance NetID.
+  string net_id = 1;
```
johanstokking: This is TTS API, so this should be `bytes` and with the `gogoproto.customtype` tag. I know we're stepping away from gogoproto, but we need to stay consistent with both protobuf and the JSON API. Due to the compatibility commitment, we will make sure that JSON NetID will stay represented in hex.

pkg/networkserver/networkserver.go
```diff
@@ -107,6 +107,8 @@ type NetworkServer struct {
 	netID      types.NetID
 	newDevAddr newDevAddrFunc
 
+	networkIdentifiers *ttnpb.NetworkIdentifiers
```
johanstokking: This isn't necessary, just declare `clusterID` under `netID` on L108

johanstokking (approved):

johanstokking:
Since this doesn't break API, this can go to v3.13

johanstokking:
What is the status here?

adriansmares (approved):

PR #4327, opened 2021-06-28
Make field masks nullable

johanstokking:
#### Summary

References #2798 

#### Changes

- Remove `(gogoproto.nullable) = false` from all `field_mask` fields
- Adapt code
- Minor cleaning and housekeeping

#### Testing

CI

##### Regressions

I might not have caught all exotic uses of the field mask. I'll be checking additional usages.

TTS may panic when we do not properly read the request and if the client did not specify field masks. This is oftentimes only in request contexts and those panics should be recovered. We may also write to a nil field mask in outgoing requests, and that may actually lead to crashes. So we need to test things properly.

#### Notes for Reviewers

Please sanity check and take a sample and properly review that.

#### Checklist


neoaggelos (approved):
Code-owned files LGTM

KrishnaIyer (approved):
Code Owned files LGTM

adriansmares:
Can `req.FieldMask` be `nil` in the following cases ? left hand side

https://github.com/TheThingsNetwork/lorawan-stack/blob/776a6912322ce72bcfaf08bf2d1fe1e1ec3957e7/pkg/identityserver/gateway_registry.go#L191
This one actually is not ok, `cleanFieldMaskPaths` is below.

https://github.com/TheThingsNetwork/lorawan-stack/blob/776a6912322ce72bcfaf08bf2d1fe1e1ec3957e7/pkg/identityserver/gateway_registry.go#L301
Same.

https://github.com/TheThingsNetwork/lorawan-stack/blob/776a6912322ce72bcfaf08bf2d1fe1e1ec3957e7/pkg/identityserver/gateway_registry.go#L191

Same

https://github.com/TheThingsNetwork/lorawan-stack/blob/776a6912322ce72bcfaf08bf2d1fe1e1ec3957e7/pkg/identityserver/gateway_registry.go#L301

Same

https://github.com/TheThingsNetwork/lorawan-stack/blob/776a6912322ce72bcfaf08bf2d1fe1e1ec3957e7/pkg/identityserver/gateway_registry.go#L445

Same

https://github.com/TheThingsNetwork/lorawan-stack/blob/776a6912322ce72bcfaf08bf2d1fe1e1ec3957e7/pkg/identityserver/registry_search.go#L146

Same

adriansmares (commented):
I've went through all of the files and saw some left-hand-side references. Please see above.

johanstokking:
Yes, I saw those too, but those are all after `ttnpb.HasAnyField` passes, so the slice is not expected to be empty.

adriansmares (approved):

htdvisser (approved):

Issue #3806, opened 2021-02-15
Add LR-FHSS support


Oliv4945:
#### Summary

RP2-1.0.2 LoRaWAN® Regional Parameters added LR-FHSS datarates, it would be great to be able to use them with The Things Stack.

#### Can you do this yourself and submit a Pull Request?

No :(

johanstokking:
These are the things that need to be done:

- [ ] New modulation `LRFHSS`: https://github.com/TheThingsNetwork/lorawan-stack/blob/b9e3712f32b748b31718896b3e83e153cdc55c6f/api/lorawan.proto#L242-L260. @rvolosatovs 
- [ ] Add LR-FHSS data rates to EU 863-870 MHz, US 902-928 MHz and AU 915-928 MHz, see RP2 1.0.2. Needs #3150. Note that LR-FHSS is only uplink. @rvolosatovs 
- [x] Support LR-FHSS metadata from gateways. This is only supported for UDP gateways for the moment, as things are being finalized to be integrated in Basic Station from what I understand. Attached is the protocol definition. @KrishnaIyer  [1]

---

[1] Fields added:

```
·         “datr”: The datarate is described as the combination of the Modulation Type, and the Operating Channel Width in kHz.

It is formatted as a string like: “M0CW137”

With:

§  M0: Modulation Type 0 as described in LR-FHSS specification (GMSK, 488Hz)



§  CW137: The OCW as described in the Regional parameters, in kHz


·         “hpw”: A new field for LR-FHSS only, which is the number of hopping steps (grid steps), as described in LR-FHSS specification

It is formatted as a number taking values 52 or 8 depending on the grid index: 0 and 1 respectively.

·         “modu”: The modulation descriptor is set to “LR-FHSS”

·         “codr”: The coding rate, as a string “1/3, 2/3, 5/6…”

·         “freq”: The center frequency of the LR-FHSS channel, as described in the packet forwarder configuration file.

o   A LR-FHSS channel will be described to the gateway as, for example:

"chan_0": { "freq": 867625000, "ocw": "137000", "obw": "488" },

"chan_1": { "freq": 868125000, "ocw": "336000", "obw": "488" },

"chan_2": { "freq": 868600000, "ocw": "137000", "obw": "488" }

o   If a LR-FHSS uplink is received on chan_0, the “freq” field will contain 867625000 in the above example.

[...]

PR #4351, opened 2021-07-02
LR-FHSS API additions

### LINKED ISSUE #4348: LR-FHSS Gateway entity support
#### Summary

The `Gateway` entity should have a field that signals that it supports LR-FHSS

#### Why do we need this?

In order to mark gateways that support features such as LR-FHSS. This allows us to avoid introducing separate frequency plans for each capability combination.

No API for this.

A new field.

#### Environment

`v3.14`

Add the required API fields.

Yes.

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4348
References https://github.com/TheThingsNetwork/lorawan-stack/issues/3806

#### Changes

- Add `supports_lrfhss` field to the `Gateway` entity, and add support for storage in the IS
- Add the `LRFHSS` modulation

#### Testing

Unit testing.

##### Regressions

API additions shouldn't result in regressions.

#### Checklist


api/gateway.proto
```diff
@@ -161,6 +161,9 @@ message Gateway {
   // Require an authenticated gateway connection. This prevents the gateway from using the UDP protocol and requires authentication when using other protocols.
   bool require_authenticated_connection = 27;
 
+  // The gateway supports the LR-FHSS uplink channels.
+  bool supports_lrfhss = 28;
+
   // next: 28
```
johanstokking:
```suggestion
  // next: 29
```

htdvisser (commented):
This needs a CHANGELOG entry, other than that, LGTM

johanstokking (commented):
Only looked at API

I think it's good that we start declaring gateway capabilities in `Gateway`, just like we have with `EndDevice`. That's going to be useful when filtering frequency plans, configuring the gateway and mapping the gateway. We know little about LR-FHSS, so maybe we get downlink support later, maybe we get a "V2" version. Should we put this under a `LRFHSS` message in `Gateway`, with now only 1 field `supported`, so we don't end up with a bunch of first level fields to give meaning to `supports_lrfhss`?

Also, GS and GCS may need this for configuring gateways (i.e. generating and serving frequency plans). Therefore, for unauthenticated gateways, GS needs to get this field with cluster auth. So I think it should be set in `PublicSafe()`.

kschiffer (approved):
Codeowned LGTM

johanstokking (approved):

PR #4355, opened 2021-07-05
Return plaintext values for update gateway

### LINKED ISSUE #4066: LNS Auth field in the Gateway Settings page displays garbage value
#### Summary

LNS Auth field in the Gateway Settings page displays garbage value

#### Steps to Reproduce

1. Update the LBS LNS Key or clear it.
2. Press Save changes

image

#### What do you see now?

Some garbage value is printed.

Print the LNS Key value (if set) or be empty.

#### Environment

```
tts v3.12.0
```

Should be something in the type conversion. This was actually tested in the original PR; https://github.com/TheThingsNetwork/lorawan-stack/pull/4011

But there might have been a regression introduced.

Run the stack and test this.

Yes

KrishnaIyer:
#### Summary

Closes #4066 

#### Changes

- Return plaintext values of secrets set in `UpdateGateway`.
- Remove unused error message.

#### Testing

Local stack

Screenshot 2021-07-05 at 11 27 16

##### Regressions

Don't expect any since I don't expect clients to use the encrypted value that this RPC used to return for anything.

#### Notes for Reviewers

@htdvisser If there's a prettier way to do this, let me know.

#### Checklist

### DIFF
--- a/pkg/identityserver/gateway_registry.go
+++ b/pkg/identityserver/gateway_registry.go
@@ -552,7 +552,7 @@ func (is *IdentityServer) updateGateway(ctx context.Context, req *ttnpb.UpdateGa
 		gtw.LBSLNSSecret.Value = ptLBSLNSSecret
 	}
 	if len(ptTargetCUPSKeySecret) != 0 {
-		gtw.TargetCUPSKey.Value = ptCACSecret
+		gtw.TargetCUPSKey.Value = ptTargetCUPSKeySecret
 	}
 
 	return gtw, nil
--- a/pkg/identityserver/gateway_registry.go
+++ b/pkg/identityserver/gateway_registry.go
@@ -64,11 +64,10 @@ var (
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/identityserver/gateway_registry.go
```diff
@@ -464,6 +467,7 @@ func (is *IdentityServer) updateGateway(ctx context.Context, req *ttnpb.UpdateGa
 			return nil, err
 		} else if req.LBSLNSSecret != nil {
 			value := req.LBSLNSSecret.Value
+			ptLBSLNSSecret = req.LBSLNSSecret.Value
```
htdvisser, line 470: Shouldn't this just replace the `value` above?

Same question for the other secrets below.

htdvisser (approved):
lgtm

pkg/identityserver/gateway_registry.go
KrishnaIyer: No because the `value` gets overwritten by the encrypted secret if an encryption key is defined.

kschiffer (approved):

PR #4374, opened 2021-07-06
Relax validation for gateway identifiers to Packet Broker

### LINKED ISSUE #4373: Custom validation of gateway IDs in `GsPba`
#### Summary

Custom validation of gateway IDs in `GsPba`.

#### Why do we need this?

To allow non-TTS validated gateway identifiers between the internal API of GS and PBA. This is needed when the GS uses a non-IS entity registry, i.e. in case of TTN V2 offloading.

More specifically, we need to allow underscores in the gateway identifiers.

Currently, `GsPba.UpdateGateway` uses the `Gateway` message. This includes `GatewayIdentifiers` which comes with regex validation that forbids underscores.

The internal API should use a more relaxed validation.

Introduce a new message that uses the same field numbers and types. Only the relevant types are needed; see the comments in `GsPba`. The gateway identifier's ID has the same regex but including the `_`.

CI covers everything. We don't need a test case for `_` identifiers.

Yes

johanstokking:
#### Summary

Relax validation for gateway identifiers to Packet Broker

Closes #4373 

#### Changes

Replace `Gateway` by `PacketBrokerGateway` that supports underscores in `gateway_id`.

#### Testing

CI should be good enough

##### Regressions

None expected

#### Checklist


api/packetbrokeragent.proto
```diff
@@ -41,8 +41,29 @@ service GsPba {
...
+      items: { string: { max_len: 64 } }
+    }
+  ];
+  repeated GatewayAntenna antennas = 13 [
+    (validate.rules).repeated.max_items = 8
+  ];
+  bool status_public = 14;
+  bool location_public = 15;
+  bool update_location_from_status = 21;
+}
```
KrishnaIyer: nit; perhaps rearrange them in numerical order?

api/packetbrokeragent.proto
```diff
@@ -52,8 +73,7 @@ message UpdatePacketBrokerGatewayRequest {
   google.protobuf.FloatValue tx_rate = 4;
 
   // The names of the gateway fields that are considered for update.
-  // Supported values are: antennas, contact_info, frequency_plan_id, frequency_plan_ids, ids, status_public and
-  // location_public.
+  // Supported values are: antennas, contact_info, frequency_plan_ids, ids, status_public and location_public.
```
KrishnaIyer: Nit; I think this comment isn't needed anymore since now the supported values are fields of 
 the `PacketBrokerGateway` message except the identifiers.

pkg/gatewayserver/upstream/packetbroker/packetbroker.go
```diff
@@ -106,13 +106,20 @@ func (h *Handler) ConnectGateway(ctx context.Context, ids ttnpb.GatewayIdentifie
 	pbaClient := ttnpb.NewGsPbaClient(pbaConn)
 
 	gtw := conn.Gateway()
+	antennas := make([]*ttnpb.GatewayAntenna, len(gtw.Antennas))
+	for i, ant := range gtw.Antennas {
+		antennas[i] = &ant
+	}
```
KrishnaIyer, line 112: What is this for? Doesn't `antennas = append(antennas,&ant)` work? Items are appended in the order in which they are passed right?

KrishnaIyer (approved):
Code owned files LGTM. A few nits.

pkg/gatewayserver/upstream/packetbroker/packetbroker.go
johanstokking: Yeah same amount of lines, same order; since we know the index this is slightly faster because `append()` would have to bounds check (or grow when the capacity isn't set).

PR #4383, opened 2021-07-08
Prepare for Protoc migration

htdvisser:
#### Summary

This pull requests makes changes to prepare for our big protoc migration.

Refs #2798 

#### Changes

First I'm removing the post-processing steps after protoc: no more `goimports` and no more `unconvert`.

I'm now also using different Docker images for the different things we generate. That way we don't have to use the exact versions of the different plugins that happen to be in the Docker image. ~For the time being I'm using Docker images from `ghcr.io/htdvisser/protoc`, but I'll probably move that to our TheThingsIndustries organization.~

I also added a folder with third party proto files, since we shouldn't assume that these are always present on every development machine.

We also switched over from https://github.com/TheThingsIndustries/protoc-gen-gogottn to a vanilla `protoc-gen-gogo`, with the custom options now specified in the `.proto` files. This will allow us to remove options from each file one by one.

#### Testing

Just regenerate the protos and watch 🤞 

##### Regressions

I don't expect any. In each commit I regenerated the protos to make sure that there wouldn't be any unexpected changes.

#### Checklist


johanstokking (commented):
This is a great start.

> I'm using Docker images from `ghcr.io/htdvisser/protoc`, but I'll probably move that to our TheThingsIndustries organization.

Please!

johanstokking (approved):

Issue #4290, opened 2021-06-21
Change some default values in the Console


johanstokking:
#### Summary

Change some default values in the Console

#### Why do we need this?

To make sure that we have saner defaults when people don't change the values.

#### What is already there? What do you see now?

1. RX1 delay defaults to 1 second when not entered. This happens at least when creating ABP devices
2. Gateway status is not public by default. There is no explanation what this means
3. Gateway location is not public by default. There is no explanation what this means

#### What is missing? What do you want to see?

1. Default RX1 delay should be 5 seconds, and this should also be displayed in the input box
2. Gateway status should be public by default. There should also be a short description that only if the location is public, the gateway status is shown on maps. Mapping services (Packet Broker, TTN Mapper) may have their own access policies (not sure if we need to mention the mapping services by name or just explain the concept)
3. Gateway location should be public by default. See (2) for explanation

#### How do you propose to implement this?

This should probably be baked in the Console, but I can also imagine that we're missing a default value in the NS (?). Also check how CLI handles this

#### How do you propose to test this?

Local

#### Can you do this yourself and submit a Pull Request?

No

htdvisser:
(2) and (3) are done with #4296

johanstokking:
Great! Are assignment, labels and milestone up to date here?

htdvisser:
Yeah, I'll see if I can still do (1) before the v3.13.3 deadline, otherwise I'll reschedule or re-assign.

htdvisser:
@johanstokking I don't think we should change that RX1 delay from 1 to 5 seconds just like that.

Instead, I would like to propose that we create a toggle between "Device starts with LoRaWAN default settings" and "Device starts with frequency plan settings", and then fill those into the following fields:

- RX1 Delay
- RX1 DR Offset
- RX2 DR Index
- RX2 Frequency
- Factory Preset Frequencies

What do you think?

cc: @kschiffer @bafonins

johanstokking:
> "Device starts with frequency plan settings", and then fill those into the following fields:
> 
> * RX1 Delay
> * RX1 DR Offset
> * RX2 DR Index
> * RX2 Frequency
> * Factory Preset Frequencies

RX1 delay isn't part of the frequency plan.

The default and desired RX1 delay is determined as follows:

[...]

PR #4384, opened 2021-07-08
NS API to retrieve default MAC settings

neoaggelos:
#### Summary

API for https://github.com/TheThingsNetwork/lorawan-stack/issues/4290

#### Changes

- Introduce `GetDefaultMACSettings` RPC for the `Ns` service, available also via `GET /ns/default_mac_settings`
- API accepts frequency plan ID and LoRaWAN PHY Version, builds the default (and desired) MAC settings and returns the result.

#### Testing

Locally, unit tests.

##### Regressions

Some changes in existing resolve crashes when the `*ttnpb.EndDevice` is nil, and expose some previously internal functions, but I don't think this should cause any regressions

#### Notes for Reviewers

@johanstokking @htdvisser  How should `FactoryPresetFrequencies` be handled? It is a simple list of frequencies, therefore assuming that they can be used both for uplink and downlink. However, channels are not always symmetric. How should this field be handled? Just leave empty, or fallback to the frequencies that can be used both for uplink and downlink only?

The default behaviour is to return all fields if no field mask paths are specified. This simplifies usage a lot.

No authentication checks are made for the RPC, do we want any?

Targetting 3.14.0 for API, so that Console work can be completed in time for 3.14.1

#### Checklist


api/networkserver.proto
```diff
@@ -30,6 +33,13 @@ message GenerateDevAddrResponse {
   bytes dev_addr = 1 [(gogoproto.customtype) = "go.thethings.network/lorawan-stack/v3/pkg/types.DevAddr"];
 }
 
+// Request of GetDefaultMACSettings.
+message GetDefaultMACSettingsRequest {
+  string frequency_plan_id = 1 [(gogoproto.customname) = "FrequencyPlanID", (validate.rules).string.max_len = 64];
```
neoaggelos, line 51: I know we are moving away from gogoproto, but for some reason grpc gateway still generates this as `FrequencyPlanID`, causing the following errors:

```
pkg/ttnpb/networkserver.pb.gw.go:76:11: protoReq.FrequencyPlanID undefined (type GetDefaultMACSettingsRequest has no field or method FrequencyPlanID)
pkg/ttnpb/networkserver.pb.gw.go:124:11: protoReq.FrequencyPlanID undefined (type GetDefaultMACSettingsRequest has no field or method FrequencyPlanID)
```

Added the `FrequencyPlanID` custom name tag for this PR for simplicity and to keep the diff small.

pkg/networkserver/grpc.go
```diff
@@ -17,13 +17,70 @@ package networkserver
...
+		return nil, err
+	}
+	classBTimeout := mac.DeviceClassBTimeout(nil, ns.defaultMACSettings)
+	classCTimeout := mac.DeviceClassCTimeout(nil, ns.defaultMACSettings)
+	adrMargin := mac.DeviceADRMargin(nil, ns.defaultMACSettings)
+	statusTimePeriodicity := mac.DeviceStatusTimePeriodicity(nil, ns.defaultMACSettings)
+	statusCountPeriodicity := mac.DeviceStatusCountPeriodicity(nil, ns.defaultMACSettings)
+
+	settings := &ttnpb.MACSettings{
+		ClassBTimeout:         &classBTimeout,
```
johanstokking: This would all be filled in through `ttnpb.HasAnyField`

johanstokking (commented):
> @johanstokking @htdvisser How should `FactoryPresetFrequencies` be handled? It is a simple list of frequencies, therefore assuming that they can be used both for uplink and downlink. However, channels are not always symmetric. How should this field be handled? Just leave empty, or fallback to the frequencies that can be used both for uplink and downlink only?

Factory preset frequencies are only enabling uplink channels. By default, these are the join channels.

> The default behaviour is to return all fields if no field mask paths are specified. This simplifies usage a lot.

This is inconsistent. Clients need to be specific about the field mask. A CLI flag like `--all` also works client side.

> No authentication checks are made for the RPC, do we want any?

This is probably not necessary, but it may become helpful in rate limiting. Maybe require `RIGHT_APPLICATION_DEVICES_READ`?

api/networkserver.proto
```diff
@@ -30,6 +33,13 @@ message GenerateDevAddrResponse {
...
+  PHYVersion lorawan_phy_version = 2 [(validate.rules).enum.defined_only = true];
+  google.protobuf.FieldMask field_mask = 3;
```
htdvisser: I don't know if the field mask really adds that much here. Maybe just leave it out, we can always add it in the future if we say that the default behavior is to return everything.

pkg/networkserver/grpc.go
```diff
@@ -17,13 +17,131 @@ package networkserver
...
+	},
+}
+
+func (ns *NetworkServer) GetDefaultMACSettings(ctx context.Context, req *ttnpb.GetDefaultMACSettingsRequest) (*ttnpb.MACSettings, error) {
+	fp, phy, err := FrequencyPlanAndBand(req.FrequencyPlanID, req.LorawanPhyVersion, ns.FrequencyPlans)
+	if err != nil {
+		return nil, err
+	}
+	settings := &ttnpb.MACSettings{}
+	for _, p := range req.GetFieldMask().GetPaths() {
```
htdvisser: This entire thing would be a lot less complicated without all that field masking, so I really think we should just get rid of that.

htdvisser (commented):
> How should `FactoryPresetFrequencies` be handled?

I'm not quite sure how factory-preset-frequencies is used in the NS. I always thought it was just uplink frequencies (which happen to be the downlink frequencies in EU868-like bands).

> No authentication checks are made for the RPC, do we want any?

We're not always consistent with this, but I think we should do the same as `ListFrequencyPlans` here, which also doesn't check auth.

neoaggelos (commented):
Removed field masks for the request, I also think it is more trouble and complexity than it's worth.

For the `FactoryPresetFrequencies`, I am using the default `UplinkChannels` of the related band.

pkg/networkserver/grpc.go
```diff
@@ -17,13 +17,65 @@ package networkserver
...
+	fp, phy, err := FrequencyPlanAndBand(req.FrequencyPlanID, req.LorawanPhyVersion, ns.FrequencyPlans)
+	if err != nil {
+		return nil, err
+	}
+	classBTimeout := mac.DeviceClassBTimeout(nil, ns.defaultMACSettings)
+	classCTimeout := mac.DeviceClassCTimeout(nil, ns.defaultMACSettings)
+	adrMargin := mac.DeviceADRMargin(nil, ns.defaultMACSettings)
+	statusTimePeriodicity := mac.DeviceStatusTimePeriodicity(nil, ns.defaultMACSettings)
+	statusCountPeriodicity := mac.DeviceStatusCountPeriodicity(nil, ns.defaultMACSettings)
+	factoryPresetFrequencies := make([]uint64, 0, len(phy.UplinkChannels))
```
adriansmares: You can use `DeviceDefaultChannels` and it will give you the band defaults. Note that leaving this empty has the same effect.

api/networkserver.proto
johanstokking: Yes that is inevitable in this case

johanstokking:
> For the `FactoryPresetFrequencies`, I am using the default `UplinkChannels` of the related band.

This should stay empty. There are no effective default factory preset frequencies. If there are none defined (desired case), then NS uses default LoRaWAN behavior to negotiate channels.

johanstokking (approved):
I'm also okay with removing the `field_mask`

I think this is good when factory preset frequencies are removed.

adriansmares (approved):

PR #4395, opened 2021-07-12
For duplicate data uplinks, keep the one with highest RSSI

### LINKED ISSUE #4095: Rate limit identical messages

neoaggelos:
#### Summary

Closes #4095 (refs https://github.com/TheThingsNetwork/lorawan-stack/issues/4095#issuecomment-838443803)

#### Changes

- Introduce `io.DropDuplicatesByRSSI()`, which de-duplicates a list of gateway uplinks. When two or more uplinks contain the same payload, only the one with the highest RSSI is kept.
- Use it in gRPC and UDP frontends.

#### Testing

Unit tests

##### Regressions

Adds a few more memory allocations per uplink, but it should not be a problem.

#### Notes for Reviewers

`DropDuplicatesByRSSI` is a terrible name.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/io.go
+++ b/pkg/gatewayserver/io/io.go
@@ -17,6 +17,7 @@ package io
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"fmt"
 	"sync/atomic"
 	"time"
@@ -733,4 +734,41 @@ func (c *Connection) discardRepeatedUplink(up *ttnpb.UplinkMessage) bool {
 		registerRepeatUp(c.ctx, shouldEmitEvent, c.gateway, c.frontend.Protocol())
 	}
 	return shouldDiscard
+}
+
+type rssiAndIndex struct {
+	rssi  float32
+	index int
+}
+
+// UniqueUplinkMessagesByRSSI returns the given list of gateway uplink messages after discarding
+// duplicates by RSSI. Two gateway uplink messages are considered duplicates if the RawPayload
+// is identical, and the RSSI values differ. In these cases, only the gateway uplink message
+// with the highest RSSI will be included in the result.
+//
+// UniqueUplinkMessagesByRSSI will allocate a new list of uplink messages, but will not copy the uplink
+// messages themselves.
+func UniqueUplinkMessagesByRSSI(uplinks []*ttnpb.UplinkMessage) []*ttnpb.UplinkMessage {
+	if len(uplinks) < 2 {
+		return uplinks
+	}
+
+	maxRSSI := make(map[string]rssiAndIndex, len(uplinks))
+	deduplicated := make([]*ttnpb.UplinkMessage, 0, len(uplinks))
+	for _, uplink := range uplinks {
+		md := uplink.GetRxMetadata()
+		if len(md) == 0 {
+			deduplicated = append(deduplicated, uplink)
+			continue
+		}
+		key := base64.StdEncoding.EncodeToString(uplink.GetRawPayload())
+		if s, ok := maxRSSI[key]; ok && s.rssi < md[0].RSSI {
+			deduplicated[s.index] = uplink
+			maxRSSI[key] = rssiAndIndex{md[0].RSSI, s.index}
+		} else if !ok {
+			deduplicated = append(deduplicated, uplink)
+			maxRSSI[key] = rssiAndIndex{md[0].RSSI, len(deduplicated) - 1}
+		}
+	}
+	return deduplicated
 }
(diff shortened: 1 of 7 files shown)


pkg/gatewayserver/io/io.go
```diff
@@ -734,3 +735,49 @@ func (c *Connection) discardRepeatedUplink(up *ttnpb.UplinkMessage) bool {
...
 }
+
+// DropDuplicatesByRSSI returns the given list of gateway uplink messages after discarding
+// duplicates by RSSI. Two gateway uplink messages are considered duplicates if the RawPayload
+// is identical, and the RSSI values differ. In these cases, only the gateway uplink message
+// with the highest RSSI will be included in the result.
+//
+// DropDuplicatesByRSSI will allocate a new list of uplink messages, but will not copy the uplink
+// messages themselves.
+func DropDuplicatesByRSSI(uplinks []*ttnpb.UplinkMessage) []*ttnpb.UplinkMessage {
```
adriansmares: `UniqueUplinkMessagesByRSSI`.

adriansmares (approved):

pkg/gatewayserver/io/io.go
```diff
@@ -734,3 +735,49 @@ func (c *Connection) discardRepeatedUplink(up *ttnpb.UplinkMessage) bool {
...
+	// Quick out
+	if len(uplinks) < 2 {
+		return uplinks
+	}
+
+	// Create map for max RRSI by payload (base64)
```
johanstokking:
```suggestion
	// Create map for max RSSI by payload (base64)
```

pkg/gatewayserver/io/io.go
```diff
@@ -734,3 +735,49 @@ func (c *Connection) discardRepeatedUplink(up *ttnpb.UplinkMessage) bool {
...
+
+// DropDuplicatesByRSSI returns the given list of gateway uplink messages after discarding
+// duplicates by RSSI. Two gateway uplink messages are considered duplicates if the RawPayload
+// is identical, and the RSSI values differ. In these cases, only the gateway uplink message
+// with the highest RSSI will be included in the result.
+//
+// DropDuplicatesByRSSI will allocate a new list of uplink messages, but will not copy the uplink
+// messages themselves.
+func DropDuplicatesByRSSI(uplinks []*ttnpb.UplinkMessage) []*ttnpb.UplinkMessage {
+	// Quick out
```
johanstokking:
```suggestion
```

pkg/gatewayserver/io/io.go
```diff
@@ -734,3 +735,49 @@ func (c *Connection) discardRepeatedUplink(up *ttnpb.UplinkMessage) bool {
...
+		if rssi, ok := maxRSSI[key]; !ok || rssi < md[0].RSSI {
+			maxRSSI[key] = md[0].RSSI
+		}
+	}
+	if len(maxRSSI) == len(uplinks) {
+		return uplinks
+	}
+
+	// Discard uplinks when the same payload with a larger RSSI was found
+	deduplicated := make([]*ttnpb.UplinkMessage, 0, len(maxRSSI))
```
johanstokking: Capacity needs to be `len(uplinks)`; there can be more uplinks with no metadata than deduplicated uplinks with RSSI

johanstokking (commented):
Why not do this in one loop? If there's no metadata, you can already append to a slice of deduplicated messages, right?

pkg/gatewayserver/io/udp/udp.go
```diff
@@ -295,7 +295,8 @@ func (s *srv) handleUp(ctx context.Context, state *state, packet encoding.Packet
 			logger.WithError(err).Warn("Failed to unmarshal packet")
 			return err
 		}
-		for _, up := range msg.UplinkMessages {
+		fmt.Println(len(io.UniqueUplinkMessagesByRSSI(msg.UplinkMessages)), len(msg.UplinkMessages))
```
adriansmares: This probably shouldn't be here.

adriansmares (approved):
LGTM but remove :point_up: .

johanstokking (approved):

Issue #3947, opened 2021-03-18
New gateway registration flow (Console)


kschiffer:
#### Summary
Congruently to #3522, we also want to revisit the gateway registration flow, which is currently too verbose and has many options that are usually not relevant during registration. Additionally, we can use templates of the device repository once we add support for gateways.

#### Why do we need this?
Gateway onboarding is currently not smooth.

#### What is already there? What do you see now?
Gateway registration using the gateway registration form.

#### What is missing? What do you want to see?
- Many UX improvements
  - Remove irrelevant fields
  - Revisit use flow 
  - #3823 
- Onboarding assistance from the device repository

#### How do you propose to implement this?
This issue is currently still a bit of a stub. I will follow up timely with wireframes to make this issue actionable. Implementation will likely happen in two stages:
1. Clean up of the current onboarding without device repository
2. Addition of onboarding via device repository to provide two onboarding modes, congruently with end device onboarding

#### How do you propose to test this?
While doing so, we should also extend the gateway end-to-end tests accordingly.

#### Can you do this yourself and submit a Pull Request?
I'll follow up with the wireframes and then we can decide who picks up the implementation.

johanstokking:
> * Clean up of the current onboarding without device repository

Let's start with this indeed.

Do you need a list of fields that are advanced, with their default values? That would eliminate already some fields that are always shown now.

johanstokking:
Other things in scope:

- Ability to set antenna placement (indoor or outdoor) (refs https://github.com/TheThingsNetwork/lorawan-stack/issues/4324)
- Filter frequency plans (refs https://github.com/TheThingsNetwork/lorawan-stack/issues/3823, probably needs gateways in device repository)

kschiffer:
I have finalized the new UX for gateway registration:
[220517_integrated_gateway_onboarding_01.pdf](https://github.com/TheThingsNetwork/lorawan-stack/files/8715860/220517_integrated_gateway_onboarding_01.pdf)

I've taken care to include only fields that make sense to set during onboarding and let everything else to be set in the general settings of the gateway.

The new onboarding experience combines both claiming and normal registration into one form. To be able to determine claimability via gateway EUI, we need a respective endpoint in the backend that @KrishnaIyer is working on.

This also includes planning for a new frequency plan selection experience that will show recommended plans first:
https://github.com/TheThingsNetwork/lorawan-frequency-plans/issues/51

[...]

PR #4396, opened 2021-07-12
Gateway opt out of Packet Broker forwarding

### LINKED ISSUE #3931: Gateway opt out for Packet Broker peering
#### Summary

Opt out specific gateways for peering through Packet Broker

#### Why do we need this?

Some users in The Things Network might not want to open their gateways for peering with commercial networks.

Also, some commercial networks may not want some gateways to peer with other networks.

With #3771, routing policies can be configured for the entire network or tenant.

A setting per gateway to opt out of peering.

- Add a new field to `Gateway`: `bool disable_packetbroker_forwarding` or something. I don't think it should be a `BoolValue`; it would really suppress the GS forward configuration, and not override it if GS doesn't have GS forwarding to PB configured
- On gateway connect, let GS fetch the field. GS already fetches a bunch of fields so that's pretty easy. https://github.com/TheThingsNetwork/lorawan-stack/blob/27720c5608ab7897c5dcb349ba5559ee86f48f8f/pkg/gatewayserver/gatewayserver.go#L462-L470
- We probably want to poll the Entity Registry for changes. Not only for this value, but also for the other fields, and let the IO middleware account for config changes.
- We would need to wire in a check for this value in the PB upstream handler, or better yet, before it gets sent to the upstream handler. We can also consider disabling the PB upstream handler for the connection, but if we want to make this dynamic, we need to be able to start and stop it, which is quite tricky

To discuss:

Should we combine the debouncing of updating gateway location, with some periodic rpc from GS to ER that updates the gateway location and gets gateway settings, in one go? cc @htdvisser @neoaggelos 

Local testing

@neoaggelos to implement

neoaggelos:
#### Summary

Closes #3931 

#### Changes

- Add `DisablePacketBrokerForwading` field to gateways.
- Skip `packetbroker` upstream host if needed.
- Console support

Screenshot from 2021-07-14 13-02-57

#### Testing

:monkey: 

##### Regressions

There should not be any.

#### Notes for Reviewers

- Default behaviour is backwards compatible, gateways forward to Packet Broker by default.
- Gateway needs to reconnect after updating config.
- 
#### Checklist


johanstokking:
> * Default behaviour is backwards compatible, gateways forward to Packet Broker by default.

Good

> * Gateway needs to reconnect after updating config.

Good

> * An alternate implementation I considered was adding a `DisabledForwarders` field, containing a list of forwarder names that the gateway would not start handlers for. However, the implementation was unnecessarily complex, so I went with a simple `bool` instead.

Yes that is fine

pkg/gatewayserver/gatewayserver.go
```diff
@@ -453,6 +453,7 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids
 				"schedule_downlink_late",
 				"status_public",
 				"update_location_from_status",
+				"disable_packet_broker_forwarding",
```

pkg/gatewayserver/gatewayserver.go
```diff
@@ -661,6 +662,10 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
 
 	hosts := make([]*upstreamHost, 0, len(gs.upstreamHandlers))
 	for name, handler := range gs.upstreamHandlers {
+		if name == "packetbroker" && gtw.DisablePacketBrokerForwarding {
+			logger.Info("Disable Packet Broker Forwarding")
```
johanstokking:
```suggestion
			logger.Info("Disable Packet Broker forwarding")
```

johanstokking (commented):
Besides testing and Console support, needs also changelog entry

Otherwise looks good!

pkg/gatewayserver/gatewayserver.go
KrishnaIyer: Can we downgrade this to a `Debug` or not print this at all?

pkg/identityserver/store/fieldmask_paths.go
```diff
@@ -61,6 +61,7 @@ const (
 	stateField                          = "state"
 	statusPublicField                   = "status_public"
 	supportsLRFHSSField                 = "supports_lrfhss"
+	disablePacketBrokerForwardingField  = "disable_packet_broker_forwarding"
```
KrishnaIyer: This also needs to be sorted.

KrishnaIyer (approved):
Looks good with a few small comments.

johanstokking:
I'm also okay with the changes.

What do we do with Console support?

neoaggelos:
@kschiffer Please review the relevant console changes. This is basically one more boolean option in the Gateway General Settings form.

adriansmares (approved):

pkg/webui/lib/field-description-messages.js
```diff
@@ -176,6 +176,9 @@ const m = defineMessages({
 
   updateGtwLocationFromStatusDescription:
     'When checked, the location of this gateway will be updated from status messages. This only works for gateways connecting with authentication; gateways connected over UDP are not supported.',
+
+  disablePacketBrokerForwardingDescription:
+    'When checked, uplink messages received from this gateway will not be forwarded to the Packet Broker. This option takes effect only after the gateway reconnects.',
```
kschiffer:
```suggestion
    'When checked, uplink messages received from this gateway will not be forwarded to Packet Broker. This option takes effect only after the gateway reconnects.',
```

pkg/webui/console/views/gateway-general-settings/basic-settings-form/index.js
```diff
@@ -201,6 +201,14 @@ const BasicSettingsForm = React.memo(props => {
         component={Input}
         autoComplete="on"
       />
+      <Form.Field
+        title={sharedMessages.packetBroker}
```
kschiffer, line 205:
```suggestion
        title={sharedMessages.packetBroker}
```

This should be more like "Peering via Packet Broker" and the default should be "Enabled" which is ticked.

kschiffer (changes requested):

pkg/webui/console/views/gateway-general-settings/basic-settings-form/index.js
neoaggelos: The gateway has a `disable_packet_broker_forwarding` field (makes things easier, since it will be `false` by default).

I am not sure how best to translate that into an "enabled" checkbox. Negating the checkbox feels like a not clean solution.

htdvisser (approved):

pkg/webui/console/views/gateway-general-settings/basic-settings-form/index.js
kschiffer: Ok, knowing that this will be updated soon, (#3947) I'm fine merging it this way.

kschiffer (approved):

PR #4402, opened 2021-07-13
Deprecate the AWS IoT provider status

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/pull/2580

#### Changes

- Remove the AWS IoT provider status, in order to lower the diff between OS and Enterprise


pkg/applicationserver/config.go
```diff
@@ -148,9 +148,6 @@ func (c PubSubConfig) toProto() *ttnpb.AsConfiguration_PubSub {
 	if status, ok := c.Providers["nats"]; ok {
 		providers.Nats = toStatus(status)
 	}
-	if status, ok := c.Providers["awsiot"]; ok {
```
adriansmares, line 151: This is already the case in Enterprise. See https://github.com/TheThingsIndustries/lorawan-stack/blob/9912dddafff50e27882e93711ebb3007c22e1047/pkg/applicationserver/config.go#L150-L159

johanstokking (approved):

Issue #4347, opened 2021-07-02
Add regional parameters version to TxRequest


adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/4196

We should add a `PHYVersion` to `TxRequest`

#### Why do we need this?

As signaled in https://github.com/TheThingsNetwork/lorawan-stack/issues/4196 , data rate indices may map to different SF/BW across different regional parameters version. As such, we need the version in order to allow the Gateway Server to translate the indices to the right values.

#### What is already there? What do you see now?

`rx1_data_rate_index` and `rx2_data_rate_index`, but no `PHYVersion`.

#### What is missing? What do you want to see?

```proto
PHYVersion band_version = 11;
```

The Network Server shall provide the band version, while the Gateway Server shall use them to derive the SF/BW.

#### Environment

`v3.14`

#### How do you propose to implement this?

- Add the proto definitions
- The NS should fill in the version and ID while generating the `TxRequest`
https://github.com/TheThingsNetwork/lorawan-stack/blob/2e5e9dfab39b7adb909f09a7a7e366c57dee6395/pkg/networkserver/downlink.go#L1236
https://github.com/TheThingsNetwork/lorawan-stack/blob/2e5e9dfab39b7adb909f09a7a7e366c57dee6395/pkg/networkserver/downlink.go#L1409
- The PBA should fill these too (https://github.com/packetbroker/api/issues/31)
https://github.com/TheThingsNetwork/lorawan-stack/blob/2e5e9dfab39b7adb909f09a7a7e366c57dee6395/pkg/packetbrokeragent/translation.go#L625
- The GS should take them into account while translating the data rate index
https://github.com/TheThingsNetwork/lorawan-stack/blob/2e5e9dfab39b7adb909f09a7a7e366c57dee6395/pkg/gatewayserver/io/io.go#L440

API changes should target `v3.14.0`, while functional changes may come in subsequent patches.

#### How do you propose to test this?

Test the SF/BW of a downlink sent via AU915 1.0.0 downlink.

#### Can you do this yourself and submit a Pull Request?

I can do the API and NS part, cc @KrishnaIyer for GS and @johanstokking for PBA.

johanstokking:
Cool, @KrishnaIyer please start with API soon, don't close this issue but re-assign once API is there:

- @KrishnaIyer to implement in GS
- @adriansmares to implement in NS
[...]

PR #4406, opened 2021-07-14
Use LoRaWAN phy version is GS

KrishnaIyer:
#### Summary

Use LoRaWAN phy version is GS. Covers the GS part of #4347 

#### Changes

- Get the phy based on the version in `TxRequest`.
- Update tests.

#### Testing

- UT
- Once we have the NS and/or PBA we should do some integration testing, esp with the AU/AS freq plans.

##### Regressions

Actual testing to ensure that downlinks work.

#### Notes for Reviewers

~⚠️ We shouldn't merge this without the NS/PBA parts.~

~Also I think that we should do this in `v3.14.1` since `v3.14.0` is already quite big.~

#### Checklist


pkg/gatewayserver/io/io.go
```diff
@@ -407,10 +407,17 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
...
 		}
 	}
-	phy, err := band.GetByID(fp.BandID)
+
+	phyFromFP, err := band.GetByID(fp.BandID)
+	if err != nil {
+		return false, false, 0, err
+	}
+
+	phy, err := phyFromFP.Version(request.GetLorawanPhyVersion())
```
adriansmares: I think this should fall back to `1.1-a` on `nil`, for backwards compatibility reasons.

adriansmares (commented):
NS part has been added in https://github.com/TheThingsNetwork/lorawan-stack/pull/4349, which is merged.

pkg/gatewayserver/io/io.go
KrishnaIyer: I considered it. But I think that it's better to fail downlinks. If we just fallback and allow them, it'll be difficult to debug when customers come back with mismatched downlink issues. 
Also a `v3.14.x` GS is not required to be compatible with a `v3.13.x` NS right? 
What do you think @johanstokking?

johanstokking: We should stay compatible with the old behavior, if it were only for Packet Broker and LoRaWAN Backend Interfaces support. This field can be optional, and if it isn't set, do not version the PHY and assume the latest version is used.

KrishnaIyer: Ok sounds good. Changes coming up.

johanstokking (approved):

PR #4407, opened 2021-07-14
Rate limit GetIdentifiersForGatewayEUI requests by gateway EUI

neoaggelos:
#### Summary

Apply rate limits for `GatewayRegistry/GetIdentifiersForGatewayEUI` RPCs per EUI.

#### Testing

Unit tests

#### Checklist

### DIFF
--- a/pkg/ttnpb/gateway.go
+++ b/pkg/ttnpb/gateway.go
@@ -1,0 +1,22 @@
+// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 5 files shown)


pkg/ttnpb/gateway.go
```diff
@@ -0,0 +1,24 @@
...
+
+package ttnpb
+
+func (r *GetGatewayIdentifiersForEUIRequest) EntityType() string { return "gateway" }
+
+func (r *GetGatewayIdentifiersForEUIRequest) IDString() string {
+	if r == nil {
+		return ""
+	}
+	return "eui-" + r.Eui.String()
```
htdvisser: This is not correct. We did that prefix trick in V2, but we don't do this in V3 anymore.

htdvisser (changes requested):

pkg/ttnpb/gateway.go
```diff
@@ -0,0 +1,24 @@
...
+
+package ttnpb
+
+func (r *GetGatewayIdentifiersForEUIRequest) EntityType() string { return "gateway" }
+
+func (r *GetGatewayIdentifiersForEUIRequest) IDString() string {
+	if r == nil {
+		return ""
+	}
+	return r.Eui.String()
```
htdvisser, line 21: I'm sorry, but this is still not correct. The EUI is not the ID string in the "gateway" namespace.

How about defining a new

```go
type RateLimitKeyer interface {
	RateLimitKey() string
}
```

and let the rate limit middleware try using that (still falling back to the IDStringer)?

htdvisser (changes requested):

htdvisser (approved):

PR #4410, opened 2021-07-14
Store schema version in ns-db migrate command

### LINKED ISSUE #4112: `ns-db migrate` takes a long time for larger redis databases
#### Summary
Background: we use AWS Elasticache for Redis for ns-db. The "Cache Size" reported in the AWS Console is 804MB, and if we export that to `.rdb` format it's about 450MB.

Running `ns-db migrate` takes approximately 20 or so minutes, even when the database is already up to date (with `--log.level debug`, we are able to see all records skipped because they are already up to date).

This translates into an extra 20 minutes of downtime when the application is started, since we run the migrations on startup. This amount of downtime is not suitable for us as we have very strict uptime requirements.

For the meantime, we have changed our deployment such that we no longer run the migration on startup. However, it means we must now manually run the migrations, which is undesirable. Ideally, we wouldn't have to keep track of this manually.

#### Steps to Reproduce

1. Have a large Redis DB.
2. Run `ttn-lw-stack ns-db migrate`
3. Wait.....

Logs (edited slightly for formatting and redact IDs just to be safe):

```
+ ttn-lw-stack --log.level debug ns-db migrate
INFO Connecting to Redis database...         
DEBUG Skip valid key  key=ttn:v3:ns:devices:uid:REDACTED
DEBUG Skip valid key  key=ttn:v3:ns:devices:uid:REDACTED
DEBUG Skip valid key  key=ttn:v3:ns:devices:eui:REDACTED
DEBUG Skip valid key  key=ttn:v3:ns:devices:uid:REDACTED
DEBUG Skip valid key  key=ttn:v3:ns:devices:uid:REDACTED
DEBUG Skip valid key  key=ttn:v3:ns:devices:addr:REDACTED
# ... and so on, for ~10k or more keys
```

`ns-db migrate` to be much faster, mainly in the case where all keys are already up-to-date. The faster the better (of course), but as a target, maybe <5 minutes?

#### Environment

Docker container `FROM thethingsnetwork/lorawan-stack:latest`.

I notice [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.12/cmd/ttn-lw-stack/commands/ns_db.go#L123) that the migration seems to map a function over every record in the database.

I had a couple naiive thoughts about this; I'm sure someone who knows what they're doing will have better ideas:

- parallelise the operation over several cores w/ goroutines & GOMAXPROCS (same, but faster)
- record the schema version inside the keys and ask redis "select * where schema_version < $current_schema" (please excuse the pseudo-SQL, I don't have any experience querying Redis). (query fewer rows)
- record the state of the migrations inside the DB and only run if the recorded schema version < $current_schema
- re-order the `switch` statement so that the regex for checking valid keys runs first (atm it runs ~7 regex matches that fail before realising the key is already valid -- this would be a micro-optimisation but figure I mention it anyways).

It would depend on my availability, and I would probably need a large amount of help from others.

neoaggelos:
#### Summary

Closes #4112 

#### Changes

- Read schema version and only run migrations if needed 
- Record schema version after a successful migration run.
- Add `--force` flag to override this logic and always run migrations (if required)

#### Testing

Locally

#### Notes for Reviewers

`ns_db.go` already contains quite a lot of logic and raw Redis commands, so I went with a very simple implementation approach. 

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ For details about compatibility between different releases, see the **Commitment
 
 - When a gateway uplink message contains duplicate data uplinks, only the one with the highest RSSI are forwarded.
 - The HTTP port now allows HTTP/2 connections over cleartext (h2c).
+- `ttn-lw-stack ns-db migrate` command records the schema version and only performs migrations if on a newer version.
+  - Use the `--force` flag to force perform migrations.
 
 ### Deprecated
 
--- a/cmd/ttn-lw-stack/commands/ns_db.go
+++ b/cmd/ttn-lw-stack/commands/ns_db.go
@@ -16,11 +16,13 @@ package commands
 
 import (
 	"regexp"
+	"strconv"
 	"time"
 
 	"github.com/go-redis/redis/v8"
 	"github.com/spf13/cobra"
 	"github.com/vmihailenco/msgpack/v5"
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
 	nsredis "go.thethings.network/lorawan-stack/v3/pkg/networkserver/redis"
 	ttnredis "go.thethings.network/lorawan-stack/v3/pkg/redis"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
@@ -92,6 +94,18 @@ var (
 
 			logger.Info("Connecting to Redis database...")
 			cl := NewNetworkServerDeviceRegistryRedis(*config)
+
+			if force, _ := cmd.Flags().GetBool("force"); !force {
+				needMigration, err := checkLatestSchemaVersion(cl)
+				if err != nil {
+					return err
+				}
+				if !needMigration {
+					logger.Info("Database schema version is already in latest version")
+					return nil
+				}
+			}
+
 			var migrated uint64
 			defer func() { logger.Debugf("%d keys migrated", migrated) }()
 
@@ -120,7 +134,7 @@ var (
 			addrRegexp3_11_Current := regexp.MustCompile(cl.Key("addr", devAddrRegexpStr, "current$"))
 			addrRegexp3_11_CurrentFields := regexp.MustCompile(cl.Key("addr", devAddrRegexpStr, "current", "fields$"))
 			addrRegexp3_11_PendingFields := regexp.MustCompile(cl.Key("addr", devAddrRegexpStr, "pending", "fields$"))
-			return rangeRedisKeys(ctx, cl, cl.Key("*"), 1, func(k string) (bool, error) {
+			err := rangeRedisKeys(ctx, cl, cl.Key("*"), 1, func(k string) (bool, error) {
 				logger := logger.WithField("key", k)
 				switch {
 				case uidRegexp3_10_Fields.MatchString(k):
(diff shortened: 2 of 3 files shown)


adriansmares (commented):
While this is definitely a step in the right direction, I think it would've been better to consider the version of the NS database 'schema' instead of the current stack version. Something such as:

Version 0 - no migration occurred
Version 1 - the current migration that we have on `ns-db migrate`
Version 2 and higher - migrations to be added in the future

In this way you won't run this migration on every release, since right now on every `binaryVersion` change the migration will be run, which is not really required - it is meant to be done only once.

neoaggelos:
> While this is definitely a step in the right direction, I think it would've been better to consider the version of the NS database 'schema' instead of the current stack version. Something such as:
> 
> Version 0 - no migration occurred
> Version 1 - the current migration that we have on `ns-db migrate`
> Version 2 and higher - migrations to be added in the future
> 
> In this way you won't run this migration on every release, since right now on every `binaryVersion` change the migration will be run, which is not really required - it is meant to be done only once.

The only complexity that this adds is what the upgrade path from each schema version to the next. Can migrations be skipped / squashed, or do they have to run in order?

But I agree, let's define a `networkserver.SchemaVersion int`, and make sure to bump it with future changes to the schema. If the upgrade path ever becomes not-straightforward, we can see how to deal with it then.

adriansmares:
>  Can migrations be skipped / squashed, or do they have to run in order?

The idea is that they are run in order, only once. Same for the AS migrations.

pkg/networkserver/networkserver.go
```diff
@@ -55,6 +55,9 @@ const (
 
 	// networkInitiatedDownlinkInterval is the minimum time.Duration passed before a network-initiated(e.g. Class B or C) downlink following an arbitrary downlink.
 	networkInitiatedDownlinkInterval = time.Second
+
+	// SchemaVersion is the Network Server database schema version. Bump when a migration is required.
+	SchemaVersion = 1
```
adriansmares: In theory the schema is an implementation detail of the store, so this probably is meant to go to the `nsredis` package.

adriansmares (approved):

htdvisser (commented):
I haven't been involved in the discussion about this issue, but please keep in mind that this will not work correctly when doing zero-downtime migrations, especially when migrating while the old version of the Network Server is still running.

The migration may skip new keys that are being written by the old Network Server, or the old Network Server may write keys that have already been migrated, possibly resulting in data loss. This is not necessarily a new problem, but at least until now it would be fixed by the next migration run.

However, when this PR gets merged, those keys will stay outdated, because at the end of the migration, the migration tool will register the new "schema version", but it can't actually guarantee that all keys are actually up-to-date.

I don't really have a solution, but it's something we should probably think about.

johanstokking:
There are two types of migrations that we should distinguish:

1. **Optimizations and deprecating fields**: the old database is compatible with the new version. The migration should run _after_ the NS is updated, so there are no old NSs still writing old fields. The new NS can use the old schema. Very important here is that people _cannot skip minors_ as we may remove usage of long deprecated fields in NS
2. **Breaking changes**: the old database is not compatible with the new and we need a migration first. These are migrations we need for IS. We should really avoid these types of migrations for NS, and as fas as I know, we never needed this.

So if we make sure that we document this correctly as part of upgrading, we should be good.

adriansmares (approved):

htdvisser (commented):
Nothing to add.

PR #4413, opened 2021-07-15
LRFHSS support in the gateway server

KrishnaIyer:
#### Summary

Refs; #3806 

#### Changes

- Add LRFHSS fields. Here's a summary of the changed/added fields

|Field|Spec|Comment|
|---|---|---|
|`modu`| New option `LR-FHSS`||
|`freq`| If the packet frequency matches with an LR-FHSS channel defined in the gateway configuration, the value here will be the channel center frequency.  If the packet frequency doesn’t match with an LR-FHSS channel defined in the gateway configuration, or if no channel defined, it will be the raw frequency with (device offset + frequency offset) | No changes needed |
|`datr`| Format `MxCWxxx` |Support marshalling and unmarshaling of fields|
|`codr`|New values| The data type remains the same so no changes needed|
|`hpw`|New field||
|`rsig.chan`|Value|Interpreted as LR-FHSS Channel based on `modu`. No changes needed|
|`rsig.rssic`|Value|Interpreted as RSSI of LR-FHSS packet based on `modu`. No changes needed|
|`rsig.foff`|Value|Interpreted as Frequency offset of LR-FHSS packet based on `modu`. No changes needed|
|`rsig.fdri`|New field||
 

**Refs**; https://github.com/TheThingsNetwork/lorawan-stack/issues/3806#issuecomment-823093084
**Spec**;  [PROTOCOL.md](https://github.com/TheThingsNetwork/lorawan-stack/files/6839472/PROTOCOL.md)

- Update tests 

#### Testing

UT only. I don't have a gateway.

##### Regressions

Don't expect any.

#### Notes for Reviewers

The following are still open

1. In the `packet_forwarder_PROTOCOL.txt` (https://github.com/TheThingsNetwork/lorawan-stack/issues/3806#issuecomment-823093084), The Modulation for LRFHSS has an additional OBW field (Operating band width?  `OBW in Hz`). But our API doesn't have this. Is this needed?
```
datr | string | LR-FHSS datarate identifier (eg. M0BW488CW137k, Modulation Type, OBW in Hz, OCW in kHz)
```
2. There's no mention of [marshaling/unmarshing with quotes](https://github.com/TheThingsNetwork/lorawan-stack/blob/7b27bae530eb096aae7a3430a96ec9250e325fee/pkg/util/datarate/data_rate.go#L40) in the spec for the LR-FHSS modulation. I've just copied what's done for LoRa since the spec looks the same for them
```
 datr | string | LoRa datarate identifier (eg. SF12BW500, bandwidth in KHz)
 datr | string | LR-FHSS datarate identifier (eg. M0BW488CW137k, Modulation Type, OBW in Hz, OCW in kHz)
```
3. The `hpw` and `rsig.fdri` values aren't currently used. Should we add them to the `RxMetadata`?
4. There's no TOA computation algorithm specified in the spec. I've made an update to the [TOA computation](https://github.com/TheThingsNetwork/lorawan-stack/blob/4b97dda7d653527f85cb939fcd8449d217f5fc00/pkg/toa/toa.go#L36-L37) to prevent a panic.

#### Checklist


johanstokking:
I'm asking our friends from Semtech to comment here.

mcoracin:
Hello,
Thank you for sharing this with us.
In fact since we first talked about LR-FHSS support, few fields have changed.
Please find below a description of what we are planning to release.
Sorry for the changes it will require on your side.
Best regards,
Michael

- `modu`: set to “LR-FHSS”
- `freq`: there will be 2 cases here:
>- If the packet frequency matches with an LR-FHSS channel defined in the gateway configuration, the value here will be the channel center frequency.
>- If the packet frequency doesn’t match with an LR-FHSS channel defined in the gateway configuration, or if no channel defined, it will be the raw frequency with (device offset + frequency offset).
- `datr`: a string in the format M0CW137, where M0 is the modulation type (only 0 supported for now), and CW is the Operating Channel Width in kHz
- `hpw`: LR-FHSS specific: a number describing the number of steps of the LR-FHSS grid (8 or 52 currently, as described in the LR-FHSS modulation spec)
- `rssic`: the channel RSSI
- `lsnr`: not defined yet for LR-FHSS, but may be in the future.
- `foff`: 2 cases, as for “freq”:
>- (frequency offset + device offset) if packet matched with a defined LR-FHSS channel
>- 0 otherwise
- `fdri`: LR-FHSS specific: the frequency drift in Hz between start and end of the packet

KrishnaIyer:
Thanks for your inputs 

As for my questions.
1. Answered
2. This is just Go json marshalling of strings with quotes so this is fine. 
3. We should add these fields to RxMetadata right @johanstokking ?
4. I think this is also fine.

johanstokking:
> * We should add these fields to RxMetadata right @johanstokking ?

We need something for `hpw` (what does it stand for?) and `fdri` (`FrequencyDrift`) indeed. The rest we seem to have or ain't applicable yet.

KrishnaIyer:
This is ready @johanstokking

api/metadata.proto
```diff
@@ -78,12 +78,16 @@ message RxMetadata {
   bytes uplink_token = 15;
   // Index of the gateway channel that received the message.
   uint32 channel_index = 17 [(validate.rules).uint32 = {lte: 255}];
+  // LR-FHSS hopping width (grid), in kHz.
+  string lrfhss_hopping_width = 19;
+  // Frequency drift in Hz between start and end of an LR-FHSS packet (signed).
+  int32 lrfhss_frequency_drift = 20;
```
johanstokking: Either put these in a new `LRFHSS` sub message, or remove `lrfhss_` and see if they will be reused for other modulations. The latter is what we've been doing so far here.

johanstokking (commented):
LGTM, one small thing

api/metadata.proto
```diff
@@ -78,12 +78,16 @@ message RxMetadata {
   bytes uplink_token = 15;
   // Index of the gateway channel that received the message.
   uint32 channel_index = 17 [(validate.rules).uint32 = {lte: 255}];
+  // Hopping width (grid) for LR-FHSS in kHz.
+  string hopping_width = 19;
```
johanstokking: The spec says:

```
 hpw  | number | LR-FHSS hopping channel width (grid number of steps)
```

So it's a number (not string) and the unit is steps (not KHz). Also, if it were KHz, we would be using Hz as everywhere in our API.

So where does this come from?

Also can you add a test case for unmarshaling a message?

KrishnaIyer: Ah yeah right.
```
 hpw  | string | LR-FHSS hopping channel width (grid), in kHz
```
That's from the old spec file [`packet_forwarder_PROTOCOL.txt`](https://github.com/TheThingsNetwork/lorawan-stack/issues/3806#issuecomment-823093084)  in the issue.

This has indeed changed in the new spec.

pkg/ttnpb/udp/packet_data.go
```diff
@@ -40,7 +40,7 @@ type RxPacket struct {
 	Modu string       `json:"modu"`           // Modulation identifier "LORA", "FSK" or "LR-FHSS"
 	DatR datarate.DR  `json:"datr"`           // LoRa datarate, FSK datarate or LR-FHSS datarate
 	CodR string       `json:"codr"`           // LoRa or LR-FHSS ECC coding rate identifier
-	Hpw  string       `json:"hpw"`            // LR-FHSS hopping channel width (grid), in kHz
+	Hpw  uint32       `json:"hpw"`            // Hopping width; a number describing the number of steps of the LR-FHSS grid
```
johanstokking:
```suggestion
	HPW  uint32       `json:"hpw,omitempty"`  // Hopping width; a number describing the number of steps of the LR-FHSS grid
```

johanstokking: `0` is an invalid number, it's only `8` or `52` (for now), and it's not always given.

johanstokking (approved):

adriansmares (approved):

PR #4416, opened 2021-07-16
LoRaCloud form fixes

### LINKED ISSUE #4099: Improve LoRaCloud DAS settings data fetching logic (Console)
#### Summary
We should update the way we fetch the necessary data for LoRaCloud DAS forms.

#### Why do we need this?
The necessary data for the forms is now fetched on the form container level. This leads to a couple of issues such as the form always reloading the data when collapsed and expanded, or the data being fetched only when the form is loaded, resulting in an unnecessary loading step (where the spinner is shown).
See also: https://github.com/TheThingsNetwork/lorawan-stack/pull/4068

Data being fetched on the form container level

Data being fetched already on the settings page `/console/applications/test-app2/integrations/lora-cloud`.

Data being fetched as described above and then passed to the forms. We can then also demote the form containers to plain components, since they wouldn't be fetching data anymore. It's also possible then to get rid of some of the[ state keeping logic we got inside the forms](https://github.com/TheThingsNetwork/lorawan-stack/pull/4068/files#diff-1fabe08344d3f4a6dd15b07983dcbe9c8b985a14fa5ca96e3295558169f3dfb7R187-R190) which resulted from the data not being present initially.

Manual tests should be enough here.

@asmulko

### LINKED ISSUE #4267: LoRa Cloud user experience and minor functional issues
#### Summary

LoRa Cloud user experience and minor functional issues

Documentation improvements are already tracked in https://github.com/TheThingsIndustries/lorawan-stack-docs/issues/266

#### Steps to Reproduce

This spans LoRa Cloud Device and Application Services (DAS) and Geolocation integration. Setup is standard, see for instance https://www.thethingsindustries.com/docs/integrations/cloud-integrations/ubidots/lora-edge-geolocation/#prepare-a-lora-cloud-setup as reference.

#### What do you see now?

1. **Console**: When first expanding the DAS token view and afterwards the GLS token view in the integration tab, then the DAS token is gone. The Geolocation token has a similar issue
2. The "automatic multi-frame" mode (looking at first byte, use value as `N` with maximum of 16) is very implicitly configurable by configuring a Count of `0`
3. Device location isn't updated in the Identity Server (or not visible in the Console)

1. The tokens should probably stay in the view state
2. Should we make this more explicit, with a checkbox?
3. Successful geolocation responses should result in the device moving on the map in the Console

Other requested features:

A. Include the request body in the service data. This could be useful to debug geolocation output, by knowing the input of the request. This may also become useful with other integrations.

1. This should just be a frontend thing; we don't need new API support for this. If people previously configured a token but did not enable the integration, the token is gone. It's just that when staying on the screen and disabling and re-enabling things, the tokens should come back
2. Checkbox to explicitly enable "automatic multi-frame", and when disabled, one can enter a count of at least 1?
3. The location of the device should be updated in IS with cluster auth

A. Consider including the request body always in the service data event type

Integration testing

Can test and review

kschiffer:
#### Summary

This PR addresses the outstanding Console issues of #4267. 

image

Closes #4267 
Closes #4099

#### Changes
- Move fetching from container to view

#### Testing

- Manual testing

##### Regressions

This could introduce issues to the LoRaCloud DAS forms in the Console.

#### Checklist


adriansmares (approved):

pkg/webui/console/containers/lora-cloud-gls-form/index.js
```diff
@@ -191,95 +192,113 @@ const LoRaCloudGLSForm = () => {
...
+          {multiFrame && (
+            <>
+              <Form.Field
+                component={Checkbox}
+                label={m.determineWindowSizeAutomatically}
+                description={m.automaticMultiFrameDescription}
+                onChange={handleAutomaticWindowSizeChange}
+                decode={decodeDetermineMultiframeAutomatically}
+                encode={encodeDetermineMultiframeAutomatically}
+                name="data.multi_frame_window_size"
```
bafonins, line 255: See L262 having the same field `name`. Is this intentional?

pkg/webui/console/containers/lora-cloud-gls-form/index.js
```diff
@@ -191,95 +192,113 @@ const LoRaCloudGLSForm = () => {
...
-              required
-            />
-            {multiFrame && (
-              <>
+      <Form.Field
+        component={Input}
+        title={sharedMessages.token}
+        description={m.tokenDescription}
+        name="data.token"
+        required
```
bafonins, line 225: When submitting the form without `token` there is no error rendered under this field.

bafonins (approved):
see my comments

pkg/webui/console/containers/lora-cloud-gls-form/index.js
kschiffer: Yes. Both fields in conjunction map to `data.multi_frame_window_size`.

bafonins (approved):

PR #4425, opened 2021-07-20
Disconnect gateways when relevant fields change

### LINKED ISSUE #4411: Disconnect gateway when gateway settings change
#### Summary

Disconnect the gateway when relevant gateway settings change in Identity Server.

References https://github.com/TheThingsNetwork/lorawan-stack/pull/4357
References https://github.com/TheThingsNetwork/lorawan-stack/pull/4396

#### Why do we need this?

To make changes effective while the gateway is connected. These are changes to frequency plan, public status and location, whether location is updated from status messages, whether authentication is required, whether Packet Broker forwarding is disabled, changes to the downlink path constraint, whether duty cycle is enforced and advanced settings for late scheduling and delays.

Currently, you need to disconnect the gateway for changes to the above become effective. The only thing that works magically is location updates to Packet Broker, as the Packet Broker upstream periodically fetches the location from Identity Server.

A background task to fetch the gateway settings periodically. If there are changes to any of the relevant settings, a disconnect should be triggered. The Things Stack expects the gateway to reconnect. The new settings will then be applied.

1. Remove the background gateway fetching that is currently in the Packet Broker upstream
2. Start a new goroutine with ticker (with jitter) to get the gateway from Packet Broker. This should include at least the following field mask:

```
antennas
downlink_path_constraint
enforce_duty_cycle
frequency_plan_id
frequency_plan_ids
location_public
require_authenticated_connection
schedule_anytime_delay
schedule_downlink_late
status_public
update_location_from_status
```

3. If there are changes to the gateway state as compared to the state when the gateway connected, disconnect the gateway
4. Check if all frontends handle [`Disconnect()`](https://github.com/TheThingsNetwork/lorawan-stack/blob/4757e6a946bfaf1784d5c96b232981aabb805e01/pkg/gatewayserver/io/io.go#L186-L189), especially UDP

We can unit test this with a mock frontend. The individual frontends may need to be tested with their own specific disconnection behavior towards gateways; UDP works very different from Basic Station.

Yes

johanstokking:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4411

#### Changes

- Watch gateway regularly, disconnect when relevant fields change
- Return a connection error early in the UDP frontend; disconnections from the server are cleaned up by the garbage collector
- Remove automatically fetching the gateway in the PB upstream

#### Testing

🐒 

gRPC example:

```
g/errors", "grpc.method": "LinkGateway", "grpc.service": "ttn.lorawan.v3.GtwGs", "grpc_code": "Canceled", "namespace": "grpc"}
error:pkg/gatewayserver:gateway_changed (gateway changed in registry)
    correlation_id=d45799648f2747efb70ac96f3f60d572
```

MQTT and Basic Station have proper disconnect support which is already tested. I believe that the UDP disconnection logic works too.

#### Notes for Reviewers

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -49,6 +49,7 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/upstream/ns"
 	"go.thethings.network/lorawan-stack/v3/pkg/gatewayserver/upstream/packetbroker"
 	"go.thethings.network/lorawan-stack/v3/pkg/log"
+	"go.thethings.network/lorawan-stack/v3/pkg/random"
 	"go.thethings.network/lorawan-stack/v3/pkg/rpcmetadata"
 	"go.thethings.network/lorawan-stack/v3/pkg/rpcmiddleware/hooks"
 	"go.thethings.network/lorawan-stack/v3/pkg/rpcmiddleware/rpclog"
@@ -512,6 +513,7 @@ func (gs *GatewayServer) Connect(ctx context.Context, frontend io.Frontend, ids 
 	registerGatewayConnect(ctx, ids, frontend.Protocol())
 	logger.Info("Connected")
 
+	gs.startDisconnectOnChangeTask(connEntry)
 	go gs.handleUpstream(connEntry)
 	if gs.statsRegistry != nil {
 		go gs.updateConnStats(connEntry)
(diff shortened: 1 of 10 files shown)


pkg/gatewayserver/gatewayserver.go
```diff
@@ -546,6 +548,86 @@ func (gs *GatewayServer) GetConnection(ctx context.Context, ids ttnpb.GatewayIde
...
+func requireDisconnect(connected, current *ttnpb.Gateway) bool {
+	if !sameAntennaLocations(connected.Antennas, current.Antennas) {
+		// Gateway Server may update the location from status messages. If the locations aren't the same, but if the new
+		// location is a GPS location, do not disconnect the gateway. This is to avoid that updating the location from a
+		// gateway status message results in disconnecting the gateway.
+		if len(current.Antennas) > 0 && current.Antennas[0].Location != nil && current.Antennas[0].Location.Source != ttnpb.SOURCE_GPS {
+			return true
+		}
+	}
+	if connected.DownlinkPathConstraint != current.DownlinkPathConstraint ||
```
adriansmares, line 560: This looks really hard to maintain (the `nil` checks got me eye balling), but I don't have a good suggestion on how to refactor it.

adriansmares (approved):

pkg/gatewayserver/gatewayserver.go
johanstokking: Agreed. I was thinking of instantiating two new `Gateway`s and using `SetFields()` to apply the same field mask, with `connected` and `current` as sources. Then we could safely use `Equals()`. But, we're getting rid of the (generated) `Equals()` as it's a gogoproto option.

KrishnaIyer (commented):
This adds some complexity which I'm not sure about.

How about we push this to the users? Would it make sense to add a `Gs.DisconnectGateway` RPC? 

When certain fields are updated, we can ask the user to force a disconnection.

On the console of course, we could have a `Save and disconnect` button that calls this RPC after the update of some fields.

KrishnaIyer:
ah this got merged while I was commenting.

johanstokking:
> This adds some complexity which I'm not sure about.

What is the complexity?

> How about we push this to the users? Would it make sense to add a `Gs.DisconnectGateway` RPC?

We don't push anything to users. We just automatically disconnect the gateway because the connection state in Gateway Server has become invalid.

What we could do is inform users in the Console that changes to the gateway may be effective only after the gateway reconnects, which can take up to 10 minutes.

> When certain fields are updated, we can ask the user to force a disconnection.

Isn't this complexity? How do we do with CLI and how do we do with API clients?

KrishnaIyer:
> What is the complexity?

So we are watching changes in the IS and making some updates. If we allow users to disconnect themselves using an API, we don't have to do that.

> We just automatically disconnect the gateway because the connection state in Gateway Server has become invalid.

I get that but the invalid state is a result of user action (updating a relevant IS field). Instead of polling the IS, we "interrupt" the GS connection based on an RPC, which is also a user action. This PR already has the disconnection logic. My point is that it could be triggered by the user.

> How do we do with CLI and how do we do with API clients?

I could imagine something similar to `GetConnectionStats` where we check if check the rights of the caller and disconnect the gateway. Then add some docs on which fields require a disconnection.

This is just an alternative which we don't have to consider anymore.

PR #4426, opened 2021-07-20
Use TLS client config instead of server

KrishnaIyer:
#### Summary

Use TLS client config instead of server. This is used to dial CUPS/LNS servers and fetch the CA. Before the fix, this would error with servers with self-signed certs.

This requires some AWS PRs.
- https://github.com/TheThingsIndustries/lorawan-stack-aws/pull/474
- A follow up for Docker; https://github.com/TheThingsIndustries/lorawan-stack-aws/pull/476

#### Changes

- Initialize the server's `tlsConfig` field with the component's Client and not Server TLS config.

#### Testing

Local stack using a `ca.pem` file.

##### Regressions

I don't think there's any.

#### Checklist

### DIFF
--- a/pkg/basicstation/cups/config.go
+++ b/pkg/basicstation/cups/config.go
@@ -61,6 +61,8 @@ func (conf ServerConfig) NewServer(c *component.Component, customOpts ...Option)
 			}),
 		)
 	}
+	// The Server.tlsConfig is used when dialing a CUPS or an LNS server to query its certificate chain.
+	// When dialing servers with self-signed certs, the Root CA of target server must either be trusted by the system or added explicitly via the `--tls.root-ca` option.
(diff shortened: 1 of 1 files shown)


pkg/basicstation/cups/config.go
```diff
@@ -61,7 +61,7 @@ func (conf ServerConfig) NewServer(c *component.Component, customOpts ...Option)
 			}),
 		)
 	}
-	if tlsConfig, err := c.GetTLSServerConfig(c.Context()); err == nil {
+	if tlsConfig, err := c.GetTLSClientConfig(c.Context()); err == nil {
```
htdvisser: This is very confusing when passing it to `NewServer`, so it could benefit from a little code comment.

htdvisser (approved):

Issue #3770, opened 2021-02-09
Smoothen root key provisioning and device claiming


johanstokking:
#### Summary

Smoothen the provisioning of root keys when creating new devices, and claiming end devices. This should be combined in one flow that covers 99% of the use cases.

Replaces #2777 
Replaces https://github.com/TheThingsIndustries/lorawan-stack/issues/2250
References https://github.com/TheThingsIndustries/lorawan-stack/issues/2345

#### Why do we need this?

For users, creating and claiming end devices can be confusing. The goal is often the same: onboarding new end devices.

I'd like to cover the following cases:

Where are the keys | What to enter | What mechanism to use | Where does it go | JoinEUI
--- | --- | --- | --- | ---
[1] Exposed to the user | Root keys | Creation | Cluster-local Join Server | Cluster-local JoinEUI
[2] Exposed to the user | Root keys | Creation * | The Things Join Server | `70B3D57ED0000000/36`
[3] The Things Join Server | Claim authentication code | Claiming ** / *** | N/A | `70B3D57ED0000000/36`
[4] Cluster-local Join Server | Claim authentication code | Claiming *** | N/A | Any
[5] Semtech Join Server | PIN | Claiming *** | N/A | `58A0CB0000000000/24`

\* this should be configured via config, and should be based on a JoinEUI
\** which claiming mechanism should be used is configurable but that is out of scope of this issue. References https://github.com/TheThingsIndustries/lorawan-stack/issues/2347
\*** scanning a QR code can also work

#### What is already there? What do you see now?

Currently, there's a checkbox "External Join Server". Checking this assumes that the device is already claimed on the external JS, but it's not clear that this step is necessary. With that disabled, only use case 1 is supported.

#### What is missing? What do you want to see?

Support all the provisioning and claim flows above.

#### How do you propose to implement this?

I think we need a combination of:

1. Front-end configuration of whether it is possible to create a device on The Things Join Server (use case 2). Generally speaking, we probably want a dropdown list of JoinEUI ranges that map to a Join Server address that is in the same realm as the cluster. That is, as The Things Join Server is part of The Things Stack Cloud, authentication from The Things Network does not work, but from The Things Stack Cloud does work
[...]

PR #4432, opened 2021-07-21
Add `AppJs` interface for applications to get the AppSKey

johanstokking:
#### Summary

Needed for https://github.com/TheThingsIndustries/lorawan-stack-aws-integration/issues/30
Preparation for #3770 

#### Changes

- Add `AppJs` interface for applications to get the AppSKey from the Join Server. This works with normal application rights authentication. This is needed for applications that use end-to-end encryption and that lost the AppSKey or did not receive the wrapped AppSKey from NS
- Decoupled the authorization method (cluster auth, TLS client auth, new app rights auth) from the actual logic
- `cluster.Authorized()` no longer panics when called and the cluster unary hook isn't configured for the concerning rpc. Cluster auth can now optionally be used, like in this use case

#### Testing

Unit testing

##### Regressions

None expected

#### Notes for Reviewers

@htdvisser please check the cluster auth changes. I believe we had the panic originally to signal a development error, assuming that endpoints are either using cluster auth or using entity access auth. This is no longer the case.

#### Checklist


pkg/auth/cluster/cluster.go
```diff
@@ -77,11 +77,10 @@ func verifySource(ctx context.Context, validKeys [][]byte) error {
 }
 
 // Authorized returns whether the context has been identified as a cluster call.
-// It panics if it does not inherit from `NewContext`.
 func Authorized(ctx context.Context) error {
 	ok, isStored := ctx.Value(clusterAuthKey).(bool)
 	if !isStored {
-		panic("call source not verified; did you register the cluster authentication hook for this call?")
+		return errNoClusterKey.New()
```
htdvisser: This panic should be restored. It was used to indicate that the implementation of an RPC checks for cluster auth, but the hook that injects cluster auth, is not executed in the chain.

See also

https://github.com/TheThingsNetwork/lorawan-stack/blob/3fac99e1a10caad41238b8bf159da19942207b3f/pkg/identityserver/identityserver.go#L152-L175

pkg/auth/cluster/cluster_test.go
```diff
@@ -29,9 +29,7 @@ import (
 func TestAuthorized(t *testing.T) {
 	a := assertions.New(t)
 
-	if !a.So(func() { cluster.Authorized(context.Background()) }, should.Panic) {
-		t.Fatal("Authorized should fail if there is no authentication data")
-	}
+	a.So(cluster.Authorized(context.Background()), should.NotBeNil)
```
htdvisser: This was correct, please restore the old code.

adriansmares (approved):
Just to see if I understood everything right:

The intent is that the `AsJs` interface should be authentic-able via `clusterauth`, while `AppJs` should not.

In that case you can just control this via parameters to `JoinServer.GetAppSKey`, and the gRPC server implementations provide these parameters (`allowClusterAuth`/`allowAPIKey`). Then you don't need to disable the `panic` - you just control via parameters which authentication styles you want.

The JS changes otherwise look good.

adriansmares (changes requested):
See above.

pkg/auth/cluster/cluster.go
johanstokking: I know what it was used for, but I have a problem with this, namely that we cannot check if cluster auth was used.

Panics are for developer mistakes, but I don't see what's the mistake is. We have places in our service layer where different, entirely unrelated authentication types can be used: TLS client auth, entity access rights and cluster auth.

So I disagree with panicking here, but you probably have a better reason that I'm not aware of, so I will make the authentication depending on the origin.

adriansmares (approved):

htdvisser (approved):
LGTM

PR #4433, opened 2021-07-21
DevEUI generation console support

### LINKED ISSUE #3750: Issue DevEUI from range with maximum per application

pgalic96:
#### Summary

Closes #3750 

Screenshot 2021-07-21 at 11 12 21 AM

#### Changes

- Redesign action buttons next to form fields
- Add dev eui config to template data
- Add DevEUI generation button to dev eui input in dev creation forms

#### Testing

- manual

##### Regressions

- shouldn't be any. it is a new function.

#### Notes for Reviewers

There is only 1 error that can arise from this API call, and that is if the global address block limit of dev eui's issued is reached. This is handled with `setError`and in that case the field is not populated.

Screenshot:

Screenshot 2021-07-21 at 11 12 44 AM

#### Checklist


api/application.proto
```diff
@@ -63,6 +63,8 @@ message Application {
   repeated ContactInfo contact_info = 7 [(validate.rules).repeated.max_items = 10];
 
   // next: 9
```
htdvisser: This should become `next: 10`, and please keep it on the bottom.

htdvisser (commented):
LGTM.

I noticed that you have a `column:dev_eui_counter` on `DevEUIAddressCounter int`; you can safely rename it to `DevEUICounter` for better consistency.

pkg/webui/console/api/index.js
```diff
@@ -140,6 +140,7 @@ export default {
     update: tts.Applications.updateById.bind(tts.Applications),
     eventsSubscribe: tts.Applications.openStream.bind(tts.Applications),
     getMqttConnectionInfo: tts.Applications.getMqttConnectionInfo.bind(tts.Applications),
+    generateDevEui: tts.Applications.issueDevEui.bind(tts.Applications),
```
bafonins: So `EUI` of `Eui`?

bafonins (approved):

pkg/webui/lib/shared-messages.js
```diff
@@ -353,6 +353,7 @@ export default defineMessages({
   uplinkMessage: 'Uplink message',
   uplinksReceived: 'Uplinks received',
   unexposed: 'Unexposed',
+  used: '{currentValue} / {maxValue} used',
```
```suggestion
  used: '{currentValue}/{maxValue} used',
```

pkg/webui/console/views/device-add/repository/device-registration/index.js
```diff
@@ -45,15 +46,40 @@ const m = defineMessages({
...
     onIdSelect(idInputRef)
   }, [idInputRef, onIdSelect])
 
+  const handleGenerate = React.useCallback(async () => {
+    try {
+      const result = await generateDevEUI()
+      setDevEUIGenerated(true)
+      euiInputRef.current.focus()
+      return result
+    } catch (error) {}
```
kschiffer: You're not handling the error here.

pkg/webui/components/input/index.js
```diff
@@ -229,9 +237,12 @@ class Input extends React.Component {
         </div>
         {hasAction && (
           <div className={style.actions}>
-            <Button className={style.button} disabled={disabled} {...action} />
+            <Button className={style.button} {...action} disabled={disabled || actionDisable} />
           </div>
         )}
+        {Boolean(indicatorMessage) && (
+          <Message className={style.indicator} component="label" content={indicatorMessage} />
+        )}
```
kschiffer: I think it's not a good idea to add this very particular/one-off functionality to the input itself. Rather we should go with a flexible approach here. What you can do is add a possibility to `<Form.Field />` to accept children that will be rendered next to the input.

pkg/webui/console/views/device-add/manual/form/form.js
```diff
@@ -153,6 +155,21 @@ const ManualForm = props => {
...
+
+  const handleGenerate = React.useCallback(async () => {
+    try {
+      const result = await generateDevEUI(appId)
+      setDevEUIGenerated(true)
+      setCurrentCounter(currentCounter + 1)
+      euiInputRef.current.focus()
+      return result.dev_eui
+    } catch (error) {
+      setError(error)
```
kschiffer: This should not be rendered as a form error. Rather, the message next to the button should switch to `DevEUI generation exceeded`, or `Unknown error occurred during generation` (for all other errors).

pkg/webui/console/views/device-add/manual/form/form.js
```diff
@@ -353,29 +374,24 @@ const ManualForm = props => {
...
+          onGenerateValue={handleGenerate}
+          actionDisable={devEUIGenerateDisabled}
+          inputRef={euiInputRef}
+          indicatorMessage={{
+            ...sharedMessages.used,
+            values: {
+              currentValue: applicationDevEUICounter,
+              maxValue: env.devEUIConfig.applicationLimit,
+            },
+          }}
```
kschiffer: See my other message on this prop. We should not pass this as prop but render this directly and pass it to the `<Form.Field />` as `children`.

Please also color the `0/50 used` message in red, once it is full, e.g. `50/50 used`.

kschiffer (changes requested):
I see that we currently disable the `Generate` button when all DevEUIs have been used, but also when no dev address generation was set up. In that case, ~~do we want to keep the current functionality rather (generating a random byte string)~~?  
 

EDIT: Just discussed it with @johanstokking and we indeed do not want random byte strings in any case.

So in that case, when there was no DevEUI generation set up, please hide the whole `Generate` button and also the `0/0 used` message.

pkg/webui/components/input/index.js
kschiffer: For the sake of getting this into `v3.14.0` I'm fine with doing this iteratively.

pkg/webui/components/input/generate.js
```diff
@@ -58,6 +59,7 @@ GenerateInput.propTypes = {
     ...Button.propTypes,
   }),
   generateTitle: PropTypes.message,
+  indicatorMessage: PropTypes.message,
```
kschiffer: You can remove this then.

pkg/webui/components/input/index.js
```diff
@@ -154,6 +160,9 @@ class Input extends React.Component {
       showPerChar,
       forwardedRef,
       inputRef,
+      actionDisable,
+      indicatorMessage,
```
kschiffer: Remove.

kschiffer (approved):

kschiffer (changes requested):
Sorry, accidentally approved. Please resolve UX related issues, other functional things I'm fine to do after the release.

pkg/webui/console/views/device-add/manual/connect.js
```diff
@@ -32,11 +37,14 @@ const mapStateToProps = state => ({
   mayEditKeys: checkFromState(mayEditApplicationDeviceKeys, state),
   prefixes: selectJoinEUIPrefixes(state),
   createDevice: (appId, device) => api.device.create(appId, device),
+  generateDevEUI: appId => api.application.generateDevEUI(appId),
```
kschiffer: See my other comment. Define this within the component rather, using `useCallback`.

pkg/webui/console/views/device-add/repository/connect.js
```diff
@@ -27,22 +27,28 @@ import {
...
 
 const mapStateToProps = state => ({
   appId: selectSelectedApplicationId(state),
+  applicationDevEUICounter: selectApplicationDevEUICount(state),
   prefixes: selectJoinEUIPrefixes(state),
   template: selectDeviceTemplate(state),
   templateFetching: selectDeviceTemplateFetching(state),
   templateError: selectDeviceTemplateError(state),
   createDevice: (appId, device) => api.device.create(appId, device),
+  generateDevEUI: appId => api.application.generateDevEUI(appId),
```
kschiffer: Note that this will likely cause a rerender whenever the state is recalculated (which will happen a lot when the app received many events) since the value is always reassigned. It would be better to set this via `setCallback()` within the component.

kschiffer (changes requested):

kschiffer (approved):

api/application.proto
htdvisser: It looks like this is still open

htdvisser (approved):
Small thing left, rest LGTM.

PR #4437, opened 2021-07-22
cli: Fix version checking period

michalborkowski96:
#### Summary
Fix version check period

#### Changes
Version checking wouldn't respect specified period

#### Testing
Local tests

##### Regressions
None

#### Checklist

### DIFF
--- a/.goreleaser.snapshot.yml
+++ b/.goreleaser.snapshot.yml
@@ -15,7 +15,7 @@ builds:
       - -w
       - -X go.thethings.network/lorawan-stack/v3/pkg/version.BuildDate={{.Date}}
       - -X go.thethings.network/lorawan-stack/v3/pkg/version.GitCommit={{.ShortCommit}}
-      - -X go.thethings.network/lorawan-stack/v3/pkg/version.TTN={{.Version}}
+      - -X go.thethings.network/lorawan-stack/v3/pkg/version.TTN={{ trimprefix .Branch "v" }}-SNAPSHOT-{{ .ShortCommit }}
     env:
       - CGO_ENABLED=0
     goos:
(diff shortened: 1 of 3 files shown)


johanstokking:
> Version checking wouldn't respect specified period

Only on failure, right? In that case, the backoff that we use for dialing external services is used.

~Or am I missing something here?~

https://github.com/TheThingsIndustries/lorawan-stack/pull/2759/files/69aee2fd4eedf3d4faf2cfbf9d012d94ffd587ed#r674225848

This suggestion is correct; the original context should be used, not the context with the timeout.

johanstokking:
Suggesting:

```diff
diff --git a/pkg/component/version_check.go b/pkg/component/version_check.go
index 5ac42e008..d97e7052e 100644
--- a/pkg/component/version_check.go
+++ b/pkg/component/version_check.go
@@ -37,18 +37,18 @@ func versionCheckTask(ctx context.Context, clientProvider httpClientProvider) *T
 		Context: ctx,
 		ID:      "version_check",
 		Func: func(ctx context.Context) error {
-			ctx, cancel := context.WithTimeout(ctx, versionCheckTimeout)
+			checkCtx, cancel := context.WithTimeout(ctx, versionCheckTimeout)
 			defer cancel()
 
-			client, err := clientProvider.HTTPClient(ctx)
+			client, err := clientProvider.HTTPClient(checkCtx)
 			if err != nil {
 				return err
 			}
-			update, err := version.CheckUpdate(ctx, version.WithClient(client))
+			update, err := version.CheckUpdate(checkCtx, version.WithClient(client))
 			if err != nil {
 				return err
 			}
-			version.LogUpdate(ctx, update)
+			version.LogUpdate(checkCtx, update)
 
 			select {
 			case <-ctx.Done():
```

michalborkowski96:
@johanstokking 

> Only on failure, right?

No.

`DialTaskBackoffIntervalFunc` specifies that a task should be repeated every 10 seconds. This is on top of the explicit wait. So the task would repeat every `versionCheckPeriod + 10 * time.Second`.

michalborkowski96:
As a solution I have removed the explicit wait, and used `backoff` of `versionCheckPeriod`, so now the task scheduler is responsible for making sure that task is repeated every `versionCheckPeriod`.

johanstokking (commented):
Yeah I see that, but now if a version check fails, it takes 12 hours to try the next, whereas previously, it would be retried shortly. I find 10 seconds slack on top of 12 hours less of a problem than logging warning when the version can't be checked. Because the latter is bad config or a connectivity issue of which the operator wants to be aware.

htdvisser (approved):

PR #4443, opened 2021-07-23
Cache Root CA for TLS clients

### LINKED ISSUE #1432: Cache root CA
#### Summary

Cache root CA

#### Why do we need this?

To avoid reading it every time building a TLS client configuration.

Reading `RootCA` file in `pkg/component.GetTLSClientConfig()` since #1431 

Cache the file read.

We could wire in the file fetcher and let the caching happen there.

Yes

neoaggelos:
#### Summary

Closes #1432

#### Changes

- Load Root CA once for tlsconfig.Client.
- Initiailze tls client config during component start (and also interop client initialization)

#### Testing

Existing unit tests.

##### Regressions

There should not be any.

#### Notes for Reviewers

@htdvisser I considered adding `panic`s when `RootCAPemBytes` is empty but `RootCA` is not, but eventually decided against it. Also considered encapsulating the caching using the `FileReader` config, but that leads to lazy loading the root CA, which I am not a huge fan of.

@johanstokking Perhaps we can use something similar for client certificates and keys also?

Are there any potential issues that arise from keeping the root CA in memory?

#### Checklist

### DIFF
--- a/pkg/config/tlsconfig/config.go
+++ b/pkg/config/tlsconfig/config.go
@@ -19,6 +19,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"io/ioutil"
+	"sync"
 	"sync/atomic"
 
 	"go.thethings.network/lorawan-stack/v3/pkg/errors"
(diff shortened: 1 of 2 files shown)


htdvisser (commented):
Why only pull out reading the file? Can't we load and cache the entire `*x509.CertPool`?

Also, wouldn't it be better to move this entire thing behind a `sync.Once` instead of relying on the first caller to call `LoadRootCA`?

So add a `certPool *x509.CertPool` and a `loadCertPoolOnce` *sync.Once` to `Client`, and in `ApplyTo` add a 

```go
c.loadCertPoolOnce.Do(func(){
	// load the ca file
	// use system cert pool or create new one
	// add the ca to the pool
	// set `c.certPool`
})
```

neoaggelos:
> Why only pull out reading the file? Can't we load and cache the entire `*x509.CertPool`?

The implementation _applies_ the Root CA from a file (if any) in to an existing TLS config. The TLS config itself may already contain other certs, which we should not break. This is why I decided against caching the `*x509.CertPool`

pkg/config/tlsconfig/config.go
```diff
@@ -107,6 +108,9 @@ type FileReader interface {
 // Client is client-side configuration for server TLS.
 type Client struct {
 	FileReader         FileReader `json:"-" yaml:"-" name:"-"`
+	loadRootCA         sync.Once  `json:"-" yaml:"-" name:"-"`
+	rootCABytes        []byte     `json:"-" yaml:"-" name:"-"`
+	rootCABytesError   error      `json:"-" yaml:"-" name:"-"`
```
htdvisser: I'm quite sure that unexported fields are already ignored by json, yaml and the config package.

htdvisser (approved):

PR #4454, opened 2021-07-27
Disable "Factory preset frequencies" field for US-like bands (Console)

### LINKED ISSUE #4431: Disable "Factory preset frequencies" field for US-like bands (Console)
#### Summary
The device creation form in the Console should not offer to set factory preset frequencies when the user has selected `AU_915_928`, `CN_470_510`, `US_902_928`.

#### Why do we need this?
See the discussions here:
https://github.com/TheThingsNetwork/lorawan-stack/pull/4291#issuecomment-882259971

and here:
https://github.com/TheThingsNetwork/lorawan-stack/issues/2269#issuecomment-810076361

Factory preset frequency settings in manual and DR-based end device onboarding.

Modifications should be:

- Disable (gray out) the field for both manual and DR-based registration, when the above frequencies were selected
- Remove the current field description entirely
- Add a field description (not tooltip) only when the field is disabled, reading: `For technical reasons this option is unavailable for certain US/AU bands`

Trivial.

Might consider adding a test case for this.

@bafonins

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4431
Screenshot 2021-07-28 at 15 56 20

#### Changes

- Adjust `<KeyValueMap />` component to check for the `disabled` prop
- Disable `factory_preset_frequencies` for certain bands, i.e. `AU_915_928`, `CN_470_510`, `US_902_928`

#### Testing

Manual

#### Checklist

### DIFF
--- a/pkg/webui/console/lib/device-utils.js
+++ b/pkg/webui/console/lib/device-utils.js
@@ -156,3 +156,12 @@ export const getLastSeen = device => {
     }
   }
 }
+
+/**
+ * @param {string} freqPlan - End device frequency plan.
+ * @returns {boolean} - Whether end device frequency plan has a CFList type of ChMask (channel mask).
+ */
+export const hasCFListTypeChMask = (freqPlan = '') =>
+  freqPlan.startsWith('US_902_928') ||
+  freqPlan.startsWith('AU_915_928') ||
+  freqPlan.startsWith('CN_470_510')
--- a/pkg/webui/console/views/device-add/manual/form/advanced-settings.js
+++ b/pkg/webui/console/views/device-add/manual/form/advanced-settings.js
@@ -28,7 +28,7 @@ import sharedMessages from '@ttn-lw/lib/shared-messages'
 import PropTypes from '@ttn-lw/lib/prop-types'
 import tooltipIds from '@ttn-lw/lib/constants/tooltip-ids'
 
-import { ACTIVATION_MODES } from '@console/lib/device-utils'
+import { ACTIVATION_MODES, hasCFListTypeChMask } from '@console/lib/device-utils'
 
 import messages from '../../messages'
 
@@ -44,6 +44,8 @@ const m = defineMessages({
   classBandC: 'Class B and class C',
   useExternalServers: 'Use external LoRaWAN backend servers',
   multicastClassCapabilities: 'LoRaWAN class for multicast downlinks',
+  factoryFreqWarning:
+    'In LoRaWAN, factory preset frequencies are only supported for bands with a CFList type of frequencies',
 })
 
 const pingSlotPeriodicityOptions = Array.from({ length: 8 }, (_, index) => {
@@ -78,6 +80,7 @@ const AdvancedSettingsSection = props => {
     activationMode,
     onDefaultNsSettingsChange,
     defaultNsSettings,
+    freqPlan,
   } = props
 
   const isOTAA = activationMode === ACTIVATION_MODES.OTAA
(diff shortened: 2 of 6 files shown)


pkg/webui/console/views/device-add/manual/form/advanced-settings.js
```diff
@@ -44,6 +44,7 @@ const m = defineMessages({
   classBandC: 'Class B and class C',
   useExternalServers: 'Use external LoRaWAN backend servers',
   multicastClassCapabilities: 'LoRaWAN class for multicast downlinks',
+  factoryFreqWarning: 'For technical reasons this option is unavailable for certain US/AU bands',
```
kschiffer:
```suggestion
  factoryFreqWarning: 'For technical reasons, setting factory preset frequencies is unavailable for certain bands (frequency plans)',
```

pkg/webui/console/lib/device-utils.js
```diff
@@ -156,3 +156,7 @@ export const getLastSeen = device => {
     }
   }
 }
+
+export const isUsBand = (band = '') => band.startsWith('US_902_928')
+export const isAuBand = (band = '') => band.startsWith('AU_915_928')
+export const isCnBand = (band = '') => band.startsWith('CN_470_510')
```
kschiffer: Please double-check @johanstokking. We will disallow the setting of factory preset settings for all such bands.

kschiffer (approved):

pkg/webui/console/lib/device-utils.js
johanstokking: So what the technical difference is between bands that do support factory preset frequencies and bands that do not support them, is that the former uses a CFList type of Frequencies, and the latter uses a CFList type of ChMask (channel mask). When there is a channel mask, the frequencies aren't configured by frequency in Hertz, but by index. The factory preset frequencies is really the frequencies in Hertz, so it requires bands with a CFList type of Frequencies.

So, maybe

```suggestion
export const hasCFListTypeChMask = (freqPlan = '') => freqPlan.startsWith('US_902_928') || freqPlan.startsWith('AU_915_928') || freqPlan.startsWith('CN_470_510')
```

Maybe?

Do note that frequency plan IDs start with the band ID, but bands can be string compared. If you pass a frequency plan ID in this context, do not call it `band` but `freqPlan` and use `startsWith()`, otherwise keep using `band` and use string comparison.

pkg/webui/console/views/device-add/manual/form/advanced-settings.js
johanstokking:
```suggestion
  factoryFreqWarning: 'In LoRaWAN, factory preset frequencies are only supported for bands with a CFList type of frequencies',
```

pkg/webui/console/views/device-add/manual/form/advanced-settings.js
```diff
@@ -87,6 +89,7 @@ const AdvancedSettingsSection = props => {
     deviceClass === DEVICE_CLASS_MAP.CLASS_B_C || deviceClass === DEVICE_CLASS_MAP.CLASS_B
   const isClassC =
     deviceClass === DEVICE_CLASS_MAP.CLASS_B_C || deviceClass === DEVICE_CLASS_MAP.CLASS_C
+  const disableFactoryPresetFreq = isAuBand(freqPlan) || isCnBand(freqPlan) || isUsBand(freqPlan)
```
johanstokking: Ah ok so here `freqPlan` is passed to a function that takes `band`. Let's not mix these terms because they're not the same.

johanstokking: So here basically it would be `const disableFactoryPresetFreq = hasCFListTypeChMask(freqPlan)` which is quite elegant and makes it easier to add more bands to that list later

pkg/webui/console/views/device-add/manual/form/advanced-settings.js
kschiffer: Edited to correct casing.

johanstokking (approved):

PR #4455, opened 2021-07-27
Fix TTL in Redis mutex tests

adriansmares:
#### Summary

```
--- FAIL: TestMutex (0.26s)
    doc.go:66: 
        /home/runner/work/lorawan-stack/lorawan-stack/pkg/redis/redis_test.go:540
        Expected: true
        Actual:   false
FAIL
FAIL	go.thethings.network/lorawan-stack/v3/pkg/redis	1.344s
```

#### Changes

- Increase the timeouts. They were way too small for the latency that occurs in CI.
- Fix the TTL used for comparisons. See code.

### DIFF
--- a/pkg/redis/redis_test.go
+++ b/pkg/redis/redis_test.go
@@ -499,14 +499,14 @@ func TestProtoDeduplicator(t *testing.T) {
 func TestMutex(t *testing.T) {
 	a, ctx := test.New(t)
 
-	ctx, cancel := context.WithTimeout(ctx, 2*time.Second+(1<<8)*test.Delay)
+	ttl := (1 << 10) * test.Delay
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second+ttl)
 	defer cancel()
 
 	cl, flush := test.NewRedis(ctx, "redis_test")
 	defer flush()
 	defer cl.Close()
 
-	ttl := (1 << 8) * test.Delay
 	key := cl.Key("test1")
 
 	err := LockMutex(ctx, cl, key, "test-id-1", ttl)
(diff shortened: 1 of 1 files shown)


pkg/redis/redis_test.go
```diff
@@ -537,7 +537,9 @@ func TestMutex(t *testing.T) {
 	case <-ctx.Done():
 		t.Fatalf("Timed out while waiting for LockMutex with a deadline to return")
 	case err := <-timeoutErrCh:
-		a.So(errors.IsDeadlineExceeded(err), should.BeTrue)
+		if !a.So(errors.IsDeadlineExceeded(err), should.BeTrue) {
```
adriansmares, line 540: The error here actually was:
```
--- FAIL: TestMutex (1.03s)
    doc.go:66: 
        /home/adriansmares/go/src/go.thethings.network/lorawan-stack/pkg/redis/redis_test.go:540
        Expected: true
        Actual:   false
    redis_test.go:541: error:pkg/errors:net_operation (read tcp 127.0.0.1:55354->127.0.0.1:6379: i/o timeout)
```
Basically the commands couldn't be written in time, and `i/o timeout` is not a deadline exceeded error (error code wise). The deadline exceeded that is expected here is for a `BLPop` operation, and that one will have the right code. But you cannot fail a `BLPop` operation if you don't even manage to send it in time.

pkg/redis/redis_test.go
```diff
@@ -577,7 +579,7 @@ func TestMutex(t *testing.T) {
 	}
 	a.So(lockTTL, should.BeGreaterThan, 0)
 	a.So(lockTTL, should.BeLessThanOrEqualTo, ttl)
-	a.So(listTTL, should.BeLessThanOrEqualTo, lockTTL)
+	a.So(listTTL, should.BeLessThanOrEqualTo, ttl)
```
adriansmares, line 582: The `listTTL` is actually reset to `ttl` on mutex unlock, not the TTL of the lock itself. This is according to the implementation.

johanstokking (approved):
Great, thanks

PR #4460, opened 2021-07-28
Device Repository API field validation and various fixes

### LINKED ISSUE #3694: Device Repository API field validation

neoaggelos:
#### Summary

Closes #3694

#### Changes

- Add field validation rules for all Device Repository API messages
- Fix minor bug with model datasheet URL
- Remove some gogoproto tags (custom names cannot be removed at the time, breaks the gRPC gateway).
- Add integration tests for Device Repository from GitHub (disabled by default, running only when the `slowtests` tag is set), validating all brands and models.

#### Testing

- Integration tests

##### Regressions

None.

#### Notes for Reviewers

Validation follows the [schema](https://lorawan-schema.org/draft/devices/1/).

`TestGithubDeviceRepository` validates all changes made with this PR do not cause any troubles with the Device Repository. It is disabled by default. We might want to run it from time to time. It cannot go in the `lorawan-devices` repository, since it is using API messages, creating a dependency cycle.

#### Checklist


pkg/devicerepository/grpc.go
```diff
@@ -60,9 +60,18 @@ func (dr *DeviceRepository) ensureBaseAssetURLs(models []*ttnpb.EndDeviceModel)
...
+	}
+	return rights.RequireApplication(ctx, *ids, ttnpb.RIGHT_APPLICATION_DEVICES_READ)
+}
+
+var errMissingApplicationIds = errors.DefineInvalidArgument("missing_application_identifiers", "missing application identifiers")
+
 // ListBrands implements the ttnpb.DeviceRepositoryServer interface.
 func (dr *DeviceRepository) ListBrands(ctx context.Context, req *ttnpb.ListEndDeviceBrandsRequest) (*ttnpb.ListEndDeviceBrandsResponse, error) {
-	if err := rights.RequireApplication(ctx, req.ApplicationIds, ttnpb.RIGHT_APPLICATION_DEVICES_READ); err != nil {
+	if err := requireApplication(ctx, req.GetApplicationIds()); err != nil {
```
johanstokking, line 69: We should catch whether `ApplicationIds` is unset in validation right?

api/devicerepository.proto
```diff
@@ -224,12 +229,15 @@ message EndDeviceModel {
...
+  repeated string additional_radios = 21 [(validate.rules).repeated = {
+    unique: true,
+    items: {string: {max_len: 50}},
+  }];
 }
 
 message GetEndDeviceBrandRequest {
   // Application identifiers.
-  ApplicationIdentifiers application_ids = 1 [(gogoproto.nullable) = false];
+  ApplicationIdentifiers application_ids = 1;
```
johanstokking: Set required here

johanstokking (commented):
What is the reason to declare validation rules in this API?

I'm not conceptually against it, but now we need to maintain validation definitions in two separate places. As long as we only use the API for returning data, we don't need validation, do we?

neoaggelos:
I'm fine with only keeping the fixes (and gogoproto cleanup) changes and reverting validations for maximum length, etc.

(They did help with identifying a number of small issues and inconsistencies in the Device Repository, but I still think that the proper place for this check should be the CI of the Device Repository and not API vaildations)

adriansmares (commented):
I also think it would indeed be best that the validation occurs in the DR repository. Changing the validation at API level on the TTS side is way more intrusive than handling it in the DR repository.

johanstokking (commented):
>I'm fine with only keeping the fixes (and gogoproto cleanup) changes and reverting validations for maximum length, etc.

Agreed. I understand it was useful, but we shouldn't let any light between validation on either side. TTS must fully trust whatever is in the Device Repository.

I'm fine keeping the test though. It would be good if you can add some godoc to it.

htdvisser (approved):
I'm fine with keeping the validation rules in the proto, and then simply not call the Validate method if we trust the files in the device repository.

htdvisser:
Since @neoaggelos isn't going to finish this before the v3.14.1 deadline, I'll move it to v3.14.2.

johanstokking (approved):
OK let's get this in then.

Please revert the DR change itself; probably a relic from rebasing from 3.13 to 3.14?

adriansmares (approved):

PR #4464, opened 2021-07-29
Prevent removing last _ALL right from entity collaborator

htdvisser:
#### Summary

This pull request changes the SetCollaborator RPCs of the Identity Server to ensure that every entity has at least one collaborator with the `_ALL` right.

We previously thought that users deleting their own account was a problem and worked on handling that (https://github.com/TheThingsNetwork/lorawan-stack/issues/3458 / https://github.com/TheThingsNetwork/lorawan-stack/pull/3616; eventually dropped), but in fact it's much more common for users to accidentally revoke their collaborator access from their entities, which should be solved by this PR.

Refs https://www.thethingsnetwork.org/forum/t/move-your-ttig-to-v3-yes/49287/59?u=htdvisser

#### Testing

I've extended the existing tests a little, but couldn't reliably cover the error itself because of the randomly populated database. Luckily we can still monkey-test this:

Screen Shot 2021-07-29 at 14 13 27

⚠️ The error message was updated after I took the screenshot. The message is now

> every gateway needs at least one collaborator with all rights

##### Regressions

This could make it impossible to update collaborators of entities that already don't have a collaborator with the `_ALL` right.

#### Notes for Reviewers

This doesn't fully cover accidental loss of access to entities when rights are granted through organizations, but I think it's already better than having no protection at all.

#### Checklist

### DIFF
--- a/pkg/identityserver/application_access.go
+++ b/pkg/identityserver/application_access.go
@@ -27,6 +27,7 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/identityserver/store"
 	"go.thethings.network/lorawan-stack/v3/pkg/log"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	"go.thethings.network/lorawan-stack/v3/pkg/unique"
 )
 
 var (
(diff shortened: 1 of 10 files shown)


pkg/identityserver/application_access.go
```diff
@@ -227,6 +228,8 @@ func (is *IdentityServer) getApplicationCollaborator(ctx context.Context, req *t
 	return res, nil
 }
 
+var errApplicationNeedsCollaborator = errors.DefineFailedPrecondition("application_needs_collaborator", "every application needs at least one collaborator with all rights")
```
kschiffer, line 231: Can't we use one message with an attribute for the entity type for all messages?

kschiffer (approved):
Please don't remove the `Changes` section in the PR description. It helps me to verify/check off bits of changes.

Changes LGTM, please see my note on the messages.

pkg/identityserver/application_access.go
htdvisser: Sure, but then we'll have a random entity type variable in the message that we can't translate.

htdvisser:
> Please don't remove the Changes section in the PR description. It helps me to verify/check off bits of changes.

Sorry. This is one change, exactly as described in the summary. And then copy-pasted from gateways to applications, clients and organizations.

PR #4473, opened 2021-08-02
Add metrics to GCS (CUPS)

KrishnaIyer:
#### Summary

Add metrics to GCS (CUPS). Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2639

#### Changes

- Add new observability package.
- Add some basic metrics

#### Testing

Local stack

Screenshot 2021-08-02 at 10 09 42

##### Regressions

NA

#### Checklist

### DIFF
--- a/pkg/basicstation/cups/observability.go
+++ b/pkg/basicstation/cups/observability.go
@@ -16,11 +16,9 @@ package cups
 
 import (
 	"context"
-	"regexp"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"go.thethings.network/lorawan-stack/v3/pkg/errors"
-	"go.thethings.network/lorawan-stack/v3/pkg/log"
 	"go.thethings.network/lorawan-stack/v3/pkg/metrics"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/basicstation/cups/observability.go
```diff
@@ -0,0 +1,105 @@
...
+	if pkg == "" {
+		pkg = "unknown"
+	}
+	if model == "" {
+		model = "unknown"
+	}
+	if station == "" {
+		station = "unknown"
+	}
+	cupsMetrics.requestSucceeded.WithLabelValues(ctx, request, pkg, model, station).Inc()
```
htdvisser: I'm not very happy with this. These fields are free-form and could quite easily be used by an attacker to create a DoS on our monitoring infrastructure by resource exhaustion (creating millions of time series)

KrishnaIyer:
Ok so based on the existing connected gateways, I've added a list of allowed values. The `package` field is optional as most gateways don't send it. The `model` field can be extracted from `station`.

We log the station value in debug before storing it so that in case there are gateways that is not in the list, we can check debug logs and add them.

KrishnaIyer:
Based on our offline discussion, the `succeeded` metrics doesn't contain labels anymore.

htdvisser (approved):

PR #4477, opened 2021-08-02
Add option to connect LBS gateways to plaintext LNS

KrishnaIyer:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2831

#### Changes

- Skip sending LNS credentials to the gateway if port `1887` is used.

#### Testing

- UT and test server + local TTS instance for LNS

##### Regressions

Tested those as well.

#### Notes for Reviewers

I would agree if you think this part is funky; 
https://github.com/TheThingsNetwork/lorawan-stack/blob/239f1eadac3e1e06b9c4c35344569785fbc86214/pkg/basicstation/cups/server.go#L237-L240

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ For details about compatibility between different releases, see the **Commitment
 ### Added
 
 - New config option `--as.packages.timeout` to control the message processing timeout of application packages.
+- Option to connect LBS gateways to plaintext LNS.
+  - This should only be used for local testing.
+  - If a TTS instance is used as the LNS, the `require_authenticated_connection` field for the gateway must be unset and the `--gs.basic-station.allow-unauthenticated` option for the GS must be set.
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


htdvisser (commented):
I don't really like this. Downgrading security should be explicit, and many users won't know that ports `188x` are plaintext. Also, you can already use plaintext by explicitly specifying `ws://` in the address, right?

KrishnaIyer:
> you can already use plaintext by explicitly specifying ws:// in the address, right?

This is what I prefer to have but the `gateway_server_address` is of the form `host:port` and does not allow the scheme to be specified. I guess we could expand the regex there or have a new `LNS URI` field on the gateway which overrides the `gateway_server_address` field.

Which do you think is better?

htdvisser:
I'm fine with extending the regex of `gateway_server_address` to allow a scheme.

PR #4485, opened 2021-08-03
Skip the application uplink queue if the peer is available

### LINKED ISSUE #4360: Introduce NS-AS communication fast path
#### Summary

The NS should skip queueing messages to the AS if the associated peer is available.

Currently, uplinks that should be submitted to the AS are always queued in a Redis stream, which is then asynchronously drained. This introduces quite some latency even though the peer is available and the stream could have been skipped.

The 'quite some latency' mentioned above is based on the metrics from https://github.com/TheThingsNetwork/lorawan-stack/pull/4258 / https://github.com/TheThingsNetwork/lorawan-stack/pull/4258 . The latency is heavily correlated to the number of sources (NS) and sinks available (AS instances).

#### Why do we need this?

To lower the NS-AS latency, in order to allow integrations to respond to uplink messages within `RX1`.

The uplinks are enqueued via `ns.enqueueApplicationUplinks`. The queue is then drained via `ns.processApplicationUplinkTask`.

I would like to see a 'fast-path' in `ns.enqueueApplicationUplinks` that attempts to contact the peer directly and submit the uplinks, instead of always enqueuing them. If the direct drain fails, only then the uplinks should be enqueued.

#### Environment

`v3.14`.

`ns.processApplicationUplinkTask` already contains most of the logic used for drainage. We can extract it from there and use it on both paths (fast path / slow path).

Unit tests should already cover this, and the latency gains can be tested via the metrics from the PRs above.

Yes. Pinging @johanstokking for an opinion on this - if all is clear please remove the `discuss` label.

adriansmares:
#### Summary

Closes #4360 

#### Changes

- The Network Server application uplink queue is now skipped if the Network Server successfully retrieves the Application Server peer and submits the uplinks.
  - The submission is done in a separate task. The concurrency is limited to 1024 tasks. When all tasks have been used, the NS automatically enqueues the uplinks.

#### Testing

##### Regressions

We'll have more uplinks in memory after this change, but that's still faster than the queue path.

#### Checklist

### DIFF
--- a/pkg/networkserver/networkserver.go
+++ b/pkg/networkserver/networkserver.go
@@ -33,6 +33,7 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/rpcmiddleware/rpclog"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 	"go.thethings.network/lorawan-stack/v3/pkg/types"
+	"golang.org/x/sync/semaphore"
 	"google.golang.org/grpc"
 )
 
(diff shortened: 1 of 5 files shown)


johanstokking (approved):
Beautiful

pkg/networkserver/networkserver.go
```diff
@@ -55,6 +56,11 @@ const (
 
 	// networkInitiatedDownlinkInterval is the minimum time.Duration passed before a network-initiated(e.g. Class B or C) downlink following an arbitrary downlink.
 	networkInitiatedDownlinkInterval = time.Second
+
+	// maxUplinkSubmissionConcurrency represents the maximum number of concurrent tasks that submit uplinks to the Application Server
+	// on the fast path. The fast path is achieved by skipping the enqueuing process and submitting the uplinks directly.
+	// When the concurrency limit is reached, uplinks are enqueued instead.
+	maxUplinkSubmissionConcurrency = 1024
```
johanstokking, line 63: Do we want some metrics on this as well?

PR #4489, opened 2021-08-04
List unlisted networks and tenants with a policy defined

### LINKED ISSUE #4388: List effective routing policies of non-listed networks
#### Summary

List effective routing policies of non-listed networks

#### Why do we need this?

To see all routing policies that apply

Currently, the Console fetches first all the (listed) networks, for instance:

`https://tti.eu1.cloud.thethings.industries/api/v3/pba/networks?page=1&limit=20&name_contains=&tenant_id_contains=&with_routing_policy=false`

Then, the Console fetches all the effective routing policies:

`https://tti.eu1.cloud.thethings.industries/api/v3/pba/forwarders/policies?page=1&limit=1000`

The Console then matches the policies of the second response with the networks returned by the first response. This is good and by initial design.

However, there are more effective routing policies in the second response that are not being shown, because those networks are not listed and are therefore not included in the first response.

I would like to see all effective forwarder routing policies.

I think we should do the following three tabs:

1. **All policies**. This tab should show the result of passing `with_routing_policy=true` to the `/api/v3/pba/networks` call. This will show all networks with an effective routing policy
2. **Non-default policies**. This is the same as what is already there
3. **All listed home networks**. This should show the contents of the current **All** tab

I can configure a few things on staging if that is helpful

Can test

### LINKED ISSUE #4447: Configure listed explicitly on Packet Broker registration
#### Summary

Configure whether a network is listed in Packet Broker explicitly in the Console and CLI.

#### Why do we need this?

Networks can only discover one another for peering configuration if their network is publicly listed in Packet Broker.

The reason why listed exists is to avoid that private networks get exposed automatically, leading to potential privacy issues.

So, this is a manual action. In TTSOS, this works with the `pba.registration` configuration. In TTSE, this references https://github.com/TheThingsIndustries/lorawan-stack/issues/2616.

There should be a slider to control whether the network is listed.

Just like there is a slider for registration, there should be a slider for whether the network is listed. This should show the value from Packet Broker, and a change should update the registration in Packet Broker. The default value when registering at Packet Broker should be taken from the current configuration, like now.

#### Environment

CLI + Console

Add new API to control whether the network is listed.

🐒 

Yes, but probably not the Console part. Will reassign when necessary.

johanstokking:
#### Summary

Closes #4388 

#### Changes

As suggested by @kschiffer in https://github.com/TheThingsNetwork/lorawan-stack/issues/4388#issuecomment-889816073, there's no change in the frontend required to list unlisted networks and tenants if there's a policy defined. This is now supported by new functionality in Packet Broker IAM.

#### Testing

Local integration testing

##### Regressions

None here. If anything, it'll be in IAM.

#### Checklist


htdvisser (approved):

johanstokking:
Wrong issue referenced, this closed https://github.com/TheThingsNetwork/lorawan-stack/issues/4388

Issue #4496, opened 2021-08-05
Internal error server when max user limit is reached in create user API


ilcato:
#### Summary

Calling create user API after reaching max number of users for the tenant results in a generic _Internal Server Error_ instead of a managed error.

#### Steps to Reproduce

1. Create a new _The Things Stack Discovery - FREE TIER_ instance 
2. Create 3 users (the maximum) using the console
3. Try to create the fourth user via the REST API 

#### What do you see now?

The response body is similar to the following:
```
{
    "code": 13,
    "message": "error:pkg/identityserver/store:transaction_recovered (Internal Server Error)",
    "details": [
        {
            "@type": "type.googleapis.com/ttn.lorawan.v3.ErrorDetails",
            "namespace": "pkg/identityserver/store",
            "name": "transaction_recovered",
            "message_format": "Internal Server Error",
            "correlation_id": "1479de1787be438bb25961aa85aeaed6",
            "code": 13
        }
    ]
}
```

#### What do you want to see instead?

A clear error message stating the reason of the error (e.g: "Maximum number of users reached") instead of the generic "Internal Server Error".

#### Environment

The Things Stack Cloud v3.14

#### How do you propose to implement this?

.

#### How do you propose to test this?

Following the previous steps.

#### Can you do this yourself and submit a Pull Request?

No

htdvisser:
This should be fixed by https://github.com/TheThingsNetwork/lorawan-stack/pull/4511 and https://github.com/TheThingsIndustries/lorawan-stack/pull/2872

PR #4511, opened 2021-08-10
Make pkg errors always use pointers

htdvisser:
#### Summary

This pull request resolves the "internal server error" part of https://github.com/TheThingsNetwork/lorawan-stack/issues/4496. 

Apparently we're recovering from a panic caused by `runtime error: comparing uncomparable type errors.Error`. We typically shouldn't compare errors (from `pkg/errors`) with `==`, but GORM seems to do that internally 🤷.

At some point in the future we should probably refactor `pkg/errors`, also to bring it more in line with Go 1.13 errors. But for now, I think it's sufficient to do a small refactor that makes sure we always use pointers to our error structs.

#### Changes

- Change all methods on `Error` and `Definition` to use pointers
- Update the rest of the codebase accordingly
- Small unrelated change (but still related to the issue): Stop logging errors other than database errors in the IS store.

#### Testing

Updated existing unit tests, tested by reproducing the issue.

##### Regressions

Maybe if we're doing something weird with typed nil errors.

#### Notes for Reviewers

...

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ For details about compatibility between different releases, see the **Commitment
 - End device access with limited rights in the Console.
 - Parsing of ID6 encoded EUIs from Basic Station gateways.
 - Warnings about unknown fields when getting or searching for gateways.
+- Internal Server Errors from `pkg/identityserver/store`.
 
 ### Security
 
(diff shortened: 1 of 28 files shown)


adriansmares (approved):

johanstokking (approved):
Nice work 🪠

kschiffer (approved):
Codeowned LGTM

PR #4525, opened 2021-08-12
Multiple Network Server task consumers 

### LINKED ISSUE #4361: Spawn multiple tasks in the Network Server
#### Summary

We currently spawn only one task per NS instance in order to schedule downlinks and to drain uplinks. 

#### Why do we need this?

These are more or less IO-bound operations, so we should spawn more in order to improve latency.
This is also safe, as the tasks are designed to be run in parallel (from any NS Instance).

https://github.com/TheThingsNetwork/lorawan-stack/blob/1a69b973744552c8df0fbaaf33f7846aae430e00/pkg/networkserver/networkserver.go#L249-L260

Multiple tasks for each type (downlink processing / uplink drainage).

#### Environment

`v3.14`.

Introduce a setting (for me, only one suffices) that allows us to configure how many tasks to have running at once, and wrap the for loop above in order to start `n` tasks of each time.

Leaving this one up to @neoaggelos . cc @johanstokking if you see any issue with this.

neoaggelos:
#### Summary

Closes #4361.

Introduces new configuration, but defaults are fully backwards compatible.

#### Changes

Pretty much the issue description.

- Explicitly specify consumer ID in calls to `TaskQueue.Pop()`, and update `nsredis.ApplicationUplinkQueue`, `nsredis.DownlinkTaskQueue` accordingly.
- Add configuration for spawning multiple task queues in the same Network Server instance.
- Update unit tests to test with multiple concurrent consumers. Some minor changes to the test specification, see inline comments.

#### Testing

Unit tests, also test traffic with a test device.

##### Regressions

Only if the NS task queue logic contains bugs, but we have not seen any in the past ~ 1 year, so probably not.

#### Notes for Reviewers

See inline comments.

#### Checklist


pkg/networkserver/internal/test/shared/application_uplink_queue.go
```diff
@@ -276,6 +321,7 @@ func handleApplicationUplinkQueueTest(ctx context.Context, q ApplicationUplinkQu
 		), should.BeTrue),
 
 		!a.So(assertAdd(ctx,
+			genericApp1Ups[1],
 			genericApp1Ups[2],
```
neoaggelos, line 324: Previous test specification implictly assumes that our queue is strictly FIFO, and this may not be the case for multiple consumers.

Updated test to reflect this, simply by avoiding mixing uplinks from multiple applications.

pkg/networkserver/internal/test/shared/downlink_task_queue.go
```diff
@@ -164,20 +176,28 @@ func handleDownlinkTaskQueueTest(ctx context.Context, q DownlinkTaskQueue) {
...
+			select {
+			case err := <-errCh:
+				if !a.So(err, should.BeNil) {
+					t.FailNow()
+				}
+
+			case <-ctx.Done():
+				t.Fatal("Timed out waiting for Pop to return")
+			}
+			nextPop[s.consumerID] <- struct{}{}
```
adriansmares, line 204: The test failure is caused by this I think.

This will allow the goroutine that `Pop`s to restart the loop, but we don't cancel it at the end. As such, it keeps listening in the background, eating the connections of the client.

Consider having `popCtx` be a cancellable and cleanup the goroutines at the end of the test (similar to the uplink queue test changes).

The number of Redis connections for the Redis client created by `pkg/util/test` is also 10 per CPU, which I think is going to cause troubles for the tests with 16 workers (CI has one CPU at most 1 I think). Consider increasing `NumWorkers` in `pkg/util/test.NewRedis`, and leaving the highest test to 8.

johanstokking:
Leaving review to @adriansmares

neoaggelos:
Redis tests are passing, but we now get timeouts for the `pkg/networkserver` tests.

I am having trouble reproducing this locally, tests either pass or timeout in unrelated flakes (e.g. in `TestDeviceRegistry`).

htdvisser (commented):
Code-owned `cmd/ttn-lw-stack/commands/start.go` LGTM.

The failing tests seem to be exactly related to the functionality that's changing here, so I don't think it's a flake.

adriansmares:
The failing tests have been fixed, see a981cca9a7c93fae216a07ccdd015d6973698d16 .

neoaggelos:
Along with latest fixes from @adriansmares this is finished, and needs a review to merge.

neoaggelos:
CI failed again but on an unrelated flake, so I think this is fine now.

KrishnaIyer (approved):
Looks good to me. Didn't spot anything obviously wrong.

Good work 🚀

adriansmares (approved):

PR #4528, opened 2021-08-13
Remove Nullable option on repeated fields

KrishnaIyer:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2874

#### Changes

- Remove the nullable option on embedded repeated fields
- Update usage
- Fix CLI

#### Testing

I think UT should cover this. But let me know if I need to test something specifically.

##### Regressions

I think UT should catch this but 🤷 

#### Checklist


KrishnaIyer:
This test failure was a complete PITA to debug; https://github.com/TheThingsNetwork/lorawan-stack/pull/4528/commits/5a7a9f86fe5abe92092da9247fcdca1d125859d8
(At least now I dug into the NS a bit).

So the issue here is that when [fetching the GS peer by ID](https://github.com/TheThingsNetwork/lorawan-stack/blob/5a7a9f86fe5abe92092da9247fcdca1d125859d8/pkg/networkserver/networkserver_util_internal_test.go#L789-L793) for the ClassB/C downlink in the tests, the key of the map is `ttnpb.GatewayIdentifiers`. 
When the `gogoproto` option was removed the `XXX_sizecache` field gets populated. So the Gateway IDs that are used as map key and the Gateway IDs passed from the downlink flow differ in this field. Since they don't match exactly, it fails to retrieve a peer and the downlink doesn't get scheduled.

Based on what I see, this is only due to the way we setup the NS test and shouldn't actually impact the actual flow. Let's anyway test Class C downlinks in staging before we deploy this.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Removed
 
 - Packet Broker mutual TLS authentication; only OAuth 2.0 is supported now.
+- Nullable option on embedded repeated fields.
```
johanstokking: This is not really relevant for the changelog

johanstokking (commented):
As for the NS test, shouldn't we use the UID as map key instead of `ttnpb.GatewayIdentifiers`? I think we have the same issue as with `==` equals checks

adriansmares (approved):

htdvisser (approved):
Nothing to add

PR #4531, opened 2021-08-16
Improve device import UX

### LINKED ISSUE #4370: End device import UX improvements
#### Summary
The current implementation of the device import has some UX shortcomings that should be fixed in order to make device migration smoother.

#### Why do we need this?
Improve utility and usability of the end device import feature.

Possibility to import devices using a template provided by the stack.

- Better error handling
  - Display the originating end device when errors occur during the import process
  - Display JSON syntax errors
- Display better information about the concluded process
  - Inform about the number of devices that could not be imported, together with the reason for it

Trivial.

We currently don't run end-to-end tests on the device import feature. We should also consider adding this.

I'll assign myself for now but might reassign based on availability.

kschiffer:
#### Summary
This PR improves the UX of importing end devices in the Console by applying a range of usability fixes.

Closes #4370 

<details>

image
image
image
image
image

</details>

#### Changes

- Improve situational awareness during device import
  - Add hint towards syntax or format errors when template conversion fails
  - Inform about failed device import on a per-device basis
- Decrease cognitive load with progressive disclosure
  - Move component and claim settings into advanced settings section
  - Show file import fields only after format selection
- Add contextual references to the documentation
- Add tooltip for setting claim authentication code
- Improve clarity of wording on labels
- Add functionality to abort import process
- Replace hardcoded text within progress bar component with i18n messages
- Change stream logic to enable starting stream only after event handlers have been set (via `stream.open()` method)
- Various small related styling fixes

#### Testing

Manual testing.

#### Notes for Reviewers
I had to modify the stream logic in the SDK because otherwise, messages could arrive before the handlers were set. This happened for the device import when the process errored out immediately.

@nejraselimovic @benolayinka we'd then also need to update the documentation accordingly, at least with new screenshots.

#### Checklist


pkg/webui/console/components/device-import-form/index.js
```diff
@@ -23,27 +23,30 @@ import FileInput from '@ttn-lw/components/file-input'
 import Checkbox from '@ttn-lw/components/checkbox'
 import SubmitBar from '@ttn-lw/components/submit-bar'
 import SubmitButton from '@ttn-lw/components/submit-button'
+import Link from '@ttn-lw/components/link'
 
 import Message from '@ttn-lw/lib/components/message'
 
 import DeviceTemplateFormatSelect from '@console/containers/device-template-format-select'
 
+import TOOLTIP_IDS from '@ttn-lw/lib/constants/tooltip-ids'
```
bafonins:
```suggestion
import tooltipIds from '@ttn-lw/lib/constants/tooltip-ids'
```

Lets be consistent across the code base. Otherwise, please change this import in other places as well.

pkg/webui/console/containers/device-importer/index.js
```diff
@@ -212,38 +308,121 @@ export default class DeviceImporter extends Component {
...
     const { redirectToList } = this.props
     const operationMessage = step === 'conversion' ? m.converting : m.creating
     let statusMessage = m.operationInProgress
     if (status === 'error') {
       statusMessage = m.operationHalted
-    } else if (status === 'finished') {
+    } else if (status === 'finished' && !aborted) {
       statusMessage = m.operationFinished
+    } else {
+      statusMessage = m.operationAborted
```
bafonins, line 339: Not sure if the issue is related to this line, but this how the UI looks for me when end device import is in progress. I dont think we should show `Operation aborted` in this case

Screenshot 2021-08-16 at 14 48 16

pkg/webui/console/containers/device-importer/index.js
```diff
@@ -185,24 +262,43 @@ export default class DeviceImporter extends Component {
...
+    // Start batch device creation.
+    this.setState({
+      step: 'creation',
+    })
+    this.appendToLog('Creating end devices…')
+
+    try {
+      this.createStream = api.device.bulkCreate(appId, devices)
 
       await new Promise((resolve, reject) => {
```
bafonins:
```suggestion
      await new Promise((resolve) => {
```

`reject` is not used

bafonins (changes requested):
When importing large json I get such UI for some time before the file is processed:
Screenshot 2021-08-16 at 14 58 27

I think the `Creating end devices…` line is redundant:
Screenshot 2021-08-16 at 15 35 59

kschiffer:
Thanks! I've addressed your comments.

pkg/webui/console/containers/device-importer/index.js
```diff
@@ -128,15 +168,42 @@ export default class DeviceImporter extends Component {
...
-  handleError(error) {
-    const json = JSON.stringify(error, null, 2)
-    this.setState(({ log }) => ({ error, status: 'error', log: `${log}\n${json}` }))
+  handleCreationError(error) {
+    this.logError(error)
+    const { convertedDevices, devicesProcessed } = this.state
+    const currentDevice = convertedDevices.indexOf !== -1 ? convertedDevices[devicesProcessed] : {}
+    const currentDeviceId =
+      'end_device' in currentDevice ? getDeviceId(currentDevice.end_device) : 'unknown device ID'
+    this.setState(({ devicesProcessed, convertedDevices, deviceErrors }) => ({
```
bafonins: there are no changes to `convertedDevices` 
```suggestion
    this.setState(({ devicesProcessed, deviceErrors }) => ({
```

pkg/webui/console/containers/device-importer/index.js
```diff
@@ -128,15 +168,42 @@ export default class DeviceImporter extends Component {
...
   }
 
   @bind
-  handleError(error) {
-    const json = JSON.stringify(error, null, 2)
-    this.setState(({ log }) => ({ error, status: 'error', log: `${log}\n${json}` }))
+  handleCreationError(error) {
+    this.logError(error)
+    const { convertedDevices, devicesProcessed } = this.state
+    const currentDevice = convertedDevices.indexOf !== -1 ? convertedDevices[devicesProcessed] : {}
```
bafonins:
```suggestion
    const currentDevice = convertedDevices.indexOf(...) !== -1 ? convertedDevices[devicesProcessed] : {}
```
`indexOf` is a function

bafonins (changes requested):

pkg/webui/console/containers/device-importer/index.js
```diff
@@ -212,38 +310,125 @@ export default class DeviceImporter extends Component {
...
+                  />
+                )}
+                <Message
+                  component="span"
+                  content={m.incompleteStatus}
+                  values={{ count: deviceErrors.length }}
+                />
+                <ul>
+                  {deviceErrors.map(({ deviceId, error }) => (
+                    <li key={deviceId} className={style.deviceErrorEntry}>
```
bafonins, line 420: This might cause the following error:
```
Warning: Encountered two children with the same key, `unknown device ID`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

pkg/webui/console/containers/device-importer/index.js
```diff
@@ -128,15 +168,43 @@ export default class DeviceImporter extends Component {
...
 
   @bind
-  handleError(error) {
-    const json = JSON.stringify(error, null, 2)
-    this.setState(({ log }) => ({ error, status: 'error', log: `${log}\n${json}` }))
+  handleCreationError(error) {
+    this.logError(error)
+    const { convertedDevices, devicesProcessed } = this.state
+    const currentDevice =
+      convertedDevices.indexOf(devicesProcessed) !== -1 ? convertedDevices[devicesProcessed] : {}
```
bafonins: What exactly is happening here? `convertedDevices` is an array of converted devices, while `devicesProcessed` is just a number. As far as I can tell, `currentDevice` is always an empty object because `indexOf` returns -1. Am I missing something?

pkg/webui/console/containers/device-importer/index.js
```diff
@@ -128,15 +168,43 @@ export default class DeviceImporter extends Component {
...
+      'end_device' in currentDevice ? getDeviceId(currentDevice.end_device) : 'unknown device ID'
+    this.setState(({ devicesProcessed, deviceErrors }) => ({
+      devicesProcessed: devicesProcessed + 1,
+      deviceErrors: [...deviceErrors, { deviceId: currentDeviceId, error }],
+    }))
+  }
+
+  @bind
+  logError(error) {
+    if (typeof error === 'object' && error !== null) {
```
bafonins:
```suggestion
    if (isObject(error)) {
```

bafonins (commented):
Please see my comments

pkg/webui/console/containers/device-importer/index.js
kschiffer: Yeah. `indexOf()` is not actually the right method here. A simple check against the length of the array should suffice. What this does is map the error to the corresponding device ID, using the index of the currently created device. This is necessary since the error itself does not contain information about what exact device ID it is connected to.

I'll fix and clarify this section.

bafonins (approved):

PR #4532, opened 2021-08-16
Query number of upstream messages from Storage Integration

neoaggelos:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/pull/2873

This PR contains the API and CLI changes.

#### Changes

- Add `GetStoredApplicationUpCount` RPC API.
- Implement `ttn-lw-cli app storage count [application-id]` command.
- Implement `ttn-lw-cli dev storage count [application-id] [device-id]` command.

#### Testing

Locally

##### Regressions

None

#### Notes for Reviewers

This is already reviewed in referenced PR.

#### Checklist


api/applicationserver_integrations_storage.proto
```diff
@@ -82,6 +80,43 @@ message GetStoredApplicationUpRequest {
...
+  // Count uplinks on a specific FPort only.
+  google.protobuf.UInt32Value f_port = 6;
+
+  // Count upstream messages that have arrived in the last minutes or hours. Cannot be used in conjunction with after and before.
+  google.protobuf.Duration last = 7;
+}
+
+
+message GetStoredApplicationUpCountResponse {
+  map<string,uint32> count = 1;
```
johanstokking, line 118: The map key here is the type, right? Please document that here

johanstokking (commented):
See also https://github.com/TheThingsIndustries/lorawan-stack/issues/2509#issuecomment-900087387

api/applicationserver_integrations_storage.proto
neoaggelos: This map key is the device ID, so the output looks like:

```json
{"count": {"dev1": 100, "dev2": 200}}
```

All requests are scoped to a single application.

htdvisser (commented):
I'm missing CHANGELOG here for the new API and new CLI command.

adriansmares (approved):

api/applicationserver_integrations_storage.proto
KrishnaIyer: We could actually echo the type in the response to help out the caller but we can add that later as well.

api/applicationserver_integrations_storage.proto
```diff
@@ -82,6 +80,44 @@ message GetStoredApplicationUpRequest {
   google.protobuf.Duration last = 10;
 }
 
+message GetStoredApplicationUpCountRequest {
+  // Count upstream messages from all end devices of an application. Cannot be used in conjunction with end_device_ids.
+  ApplicationIdentifiers application_ids = 1;
+  // Count upstream messages from a single end device. Cannot be used in conjunction with application_ids.
+  EndDeviceIdentifiers end_device_ids = 2;
```
KrishnaIyer, line 87: We could use the `EntityIdentifiers` and assert the type to one of these two right?

cmd/ttn-lw-cli/commands/storage_integration.go
```diff
@@ -114,15 +139,46 @@ var (
 			return getStoredUp(cmd, args, client, os.Stdout)
 		},
 	}
+	applicationsStorageCountCommand = &cobra.Command{
+		Use:   "count [application-id]",
+		Short: "Retrieve stored upstream messages",
```
KrishnaIyer: I find this a bit confusing (and I guess so will users).
```
count [application-id] [device-id]
```
just returns a count where as

```
count [application-id]
```
Retrieves the actual messages.

KrishnaIyer (commented):
I've a few question on the CLI but the rest looks good. Should we maybe wrap the "Unimplemented" error and present it nicely when someone makes this request with an OS backend?

api/applicationserver_integrations_storage.proto
adriansmares: The HTTP bindings don't work with oneof fields, that's why these fields are explicit

cmd/ttn-lw-cli/commands/storage_integration.go
adriansmares: This is a copy paste failure I believe

neoaggelos: Fixed

KrishnaIyer (approved):

Issue #4346, opened 2021-07-02
Refactor to band package to maintain individual band versions


adriansmares:
#### Summary

`pkg/band` should be refactored such that the current `downgrade` system is removed. We should have a version of each band per regional parameters version, instead of downgrading the latest version

#### Why do we need this?

The linear relation between versions no longer exist after RP0002 has been released. It has also become increasingly hard to maintain (and validate) the downgrade system due to the newer versions.

#### What is already there? What do you see now?

Each band in `pkg/band` is currently specifying the parameters of  the `RP001_V1_1_REV_B` version, and earlier versions are accessed by 'downgrading' the band to an earlier version.

#### What is missing? What do you want to see?

I would like each individual band to have its own subpackage, containing a version for each relevant regional parameters version.

#### Environment

`v3.14`

#### How do you propose to implement this?

- Export the current band definitions, on a per version basis, to a JSON file. This is to be used as ground truth for future tests.
- Introduce the new band subpackages by breaking down the bands themselves.
- Use the ground truth (i.e. the old definitions, saved in the JSON file) to verify that we didn't introduce regressions.

#### How do you propose to test this?

See above.

#### Can you do this yourself and submit a Pull Request?

Yes.

PR #4540, opened 2021-08-17
Add LR-FHSS data rate coding rates

### LINKED ISSUE #4466: Put coding rate in band's data rate
#### Summary

Declare coding rate in band's data rates

#### Why do we need this?

Unlike with LoRa, with LR-FHSS, the data rate is partly defined by the coding rate.

I think that this is needed for https://github.com/TheThingsNetwork/lorawan-stack/issues/3806

Gateway independently sets the PHY's LoRa coding rate on downlink.

With LR-FHSS this isn't possible; it depends on the data rate, but this is not declared in the band.

The coding rate isn't defined in the band.

I think we should declare the coding rate as part of the LR-FHSS data rate so that it can be get by index.

That is, add it as a field to `message LRFHSSDataRate` API.

We don't have to define the LoRa coding rate in `LoRaDataRate` just yet, let's not do that for now either.

Unit tests to get the full data rate by index and back

Can review

adriansmares:
#### Summary

References #4466
References https://github.com/TheThingsNetwork/lorawan-stack/issues/3806 
References https://github.com/TheThingsNetwork/lorawan-stack/issues/4346

#### Changes

- Add the coding rate to the LR-FHSS data rates (API)
- In GS/PBA use the coding rate from the data rate, when possible

#### Testing

Unit testing.

##### Regressions

None expected.

#### Notes for Reviewers

I've attempted to refactor away the `LoRaCodingRate` from the `band` package and to add the coding rate to the `LoRa` data rate. This is a very invasive change, and must occur in a minor, as a `3.x.y` GS may not send a full LoRa data rate to a `3.x.(y + 1)` PBA. As such, I didn't include it in this PR.

#### Checklist


pkg/pfconfig/lbslns/lbslns.go
```diff
@@ -308,7 +308,6 @@ func getDataRatesFromBandID(id string) (DataRates, error) {
 
 	for i, dr := range phy.DataRates {
 		if loraDR := dr.Rate.GetLora(); loraDR != nil {
-			loraDR.GetSpreadingFactor()
```
KrishnaIyer, line 311: 🤦

pkg/toa/toa.go
```diff
@@ -144,3 +142,24 @@ func computeFSK(payloadSize int, frequency uint64, bitRate uint32, crc bool) (ti
 		return 0, errFrequency.WithAttributes("frequency", frequency)
 	}
 }
+
+func computeLRFHSS(phyPayloadLength int, codingRate string, crc bool) (time.Duration, error) {
```
KrishnaIyer, line 146: Could you link the spec here if there's a public link?

KrishnaIyer (approved):
LGTM.

This PR should target the `v3.15` branch right?

htdvisser (approved):
Code-owned changes good to me 👍

pkg/toa/toa.go
johanstokking: We can do that, but this is defined in LoRaWAN Regional Parameters RP2 1.0.2, not in a specific datasheet like the others. We don't refer to LoRaWAN specs all the time in our codebase, so I guess that's the reason not to mention it.

johanstokking (approved):
This is more elegant than I could have hoped for

johanstokking:
> I've attempted to refactor away the `LoRaCodingRate` from the `band` package and to add the coding rate to the `LoRa` data rate. This is a very invasive change, and must occur in a minor, as a `3.x.y` GS may not send a full LoRa data rate to a `3.x.(y + 1)` PBA. As such, I didn't include it in this PR.

I think this is fine for now indeed.

PR #4547, opened 2021-08-18
Device network initiated downlink interval

### LINKED ISSUE #3333: Make network-initiated downlink interval configurable
#### Summary

Make network-initiated downlink interval configurable

#### Why do we need this?

To allow for network-initiated downlink rate-limiting by the customer

Hardcoded value in NS (https://github.com/TheThingsNetwork/lorawan-stack/blob/11cf05ba98cf6df44f064658ea4393183b0e7126/pkg/networkserver/networkserver.go#L57)

Make this value configurable

Add this as field in device MAC settings

integration testing

yes

neoaggelos:
#### Summary

Closes #3333.

#### Changes

- Add `network_initiated_downlink_interval` field to MAC settings (~do not use~ **use** gogoproto for duration, similar to other MAC settings, to avoid having to deal with *pbtypes.Duration in CLI command line flags).
- Use interval from MAC settings (if set), otherwise fallback to default value of 1 second.

#### Testing

:monkey:

##### Regressions

Changes are fully backwards compatible, existing devices have no interval set so will keep using the default value.

#### Notes for Reviewers

This issue is small enough, so worked on it to get it over with, since it requires some probably not too obvious changes in the `pkg/ttnpb` package for `FieldIsZero` and validations.

#### Checklist


adriansmares (approved):

api/end_device.proto
```diff
@@ -270,6 +270,8 @@ message MACSettings {
   // Maximum EIRP (dBm).
   // If unset, the default value from regional parameters specification will be used.
   DeviceEIRPValue desired_max_eirp = 30;
+  // The minimum duration passed before a network-initiated(e.g. Class B or C) downlink following an arbitrary downlink.
+  google.protobuf.Duration network_initiated_downlink_interval = 31 [(gogoproto.stdduration) = true];
```
johanstokking: I think we use `class_b_c` in other places; that would be instead of `network_initiated`. I know it's not consistent with NS terminology, but this is API so that should stay consistent with API, that is more important than introducing NS terms here.

neoaggelos:
This needs code-owner approval and is good to merge.

api/end_device.proto
```diff
@@ -270,6 +270,8 @@ message MACSettings {
   // Maximum EIRP (dBm).
   // If unset, the default value from regional parameters specification will be used.
   DeviceEIRPValue desired_max_eirp = 30;
+  // The minimum duration passed before a network-initiated(e.g. Class B or C) downlink following an arbitrary downlink.
```
KrishnaIyer, line 273: I'd reframe this as 
```
The minimum duration that must pass before a network-initiated (e.g. Class B or C) downlink is sent following an arbitrary downlink.
```
This feels like a `downlink spacing` or a `scheduling delay` rather than an `interval` but that's just be me.

Btw, what is an arbitrary downlink?

pkg/networkserver/mac/utils.go
```diff
@@ -571,6 +571,18 @@ outer:
 	return chs
 }
 
+const defaultClassBCDownlinkInterval = time.Second
+
+func DeviceClassBCDownlinkInterval(dev *ttnpb.EndDevice, defaults ttnpb.MACSettings) time.Duration {
+	if t := dev.GetMacSettings().GetClassBCDownlinkInterval(); t != nil {
```
KrishnaIyer, line 577: For devices that don't have this field in the registry, does this just return nil and we skip to `L583`?

Or do existing devices need a migration to set this field in the DB and then set a default value? 

This is just for me since I don't know much about how non existent fields are handled when you get them from Redis. cc: @adriansmares

KrishnaIyer (commented):
API and implementation look good to me. Small questions

pkg/networkserver/mac/utils.go
neoaggelos: This is `nil` for existing devices, indeed, no migrations required.

KrishnaIyer: Perfect thanks.

api/end_device.proto
adriansmares: You have device initiated (class A) and network initiated (class B/C). Arbitrary here is meant to catch both.

KrishnaIyer: Indeed but the word "arbitrary" is not defined in the LoRaWAN spec as a collective term and contrarily you cannot send an arbitrary downlink since their types are predefined (A/B/C). 
I'd have rephrased the following 
```
following an arbitrary downlink.
```
as
```
following a downlink.
```

But this is a minor point.

KrishnaIyer (approved):

PR #4557, opened 2021-08-24
Fix byte input width for Safari in the Console

### LINKED ISSUE #3784: App Key text field is too narrow

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/3784

<details>

before:

before

after:

after

</details>

#### Testing

Manual

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ For details about compatibility between different releases, see the **Commitment
 - Avoid rendering blank pages in the Console for certain errors.
 - Blank page crashes in the Console for certain browsers that do not fully support `Intl` API.
 - End device session keys handling in the Console.
+- Byte input width in Safari in the Console.
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


kschiffer (approved):

kschiffer:
Worth adding a changelog entry for.

PR #4573, opened 2021-09-03
Evict downlink queue of the pending session on join accept

adriansmares:
#### Summary

References https://sentry.io/organizations/the-things-industries/issues/2288551302/events/025faecbfcce495b9afee709f7b319e6/?project=2682566

#### Changes

- Clear the pending session downlink queue in the Network Server while generating the new pending session. 

#### Testing

Local testing.

##### Regressions

Since the `session_key_id` of these downlinks would be all over the place (basically one for each previous join attempt), these downlinks would be dropped anyway. At least now we don't risk filling the queue with useless downlinks.

#### Notes for Reviewers

I thought we were doing this. We were not doing this, and as a result devices that were stuck in a join loop (JoinAccept being generated but not reaching the device) would end up having new downlinks being pushed contiguously.

#### Checklist


KrishnaIyer (approved):
LGTM

PR #4574, opened 2021-09-03
Increase hash size used during uplink matching and deduplication

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack-support/issues/573

#### Changes

- Change hashing algorithm used during deduplication and matching to `SHA3-224`, from `FNV-1a`
- Lower matching cache TTL to the collection window (200 ms + 1 second = 1.2 seconds by default), down from one minute

#### Testing

Not applicable.

##### Regressions

Not applicable.

#### Notes for Reviewers

In v2, deduplication used to hash the payload using `MD5`, which is a 128 bit hash. In v3, after https://github.com/TheThingsNetwork/lorawan-stack/pull/2837 we are using `FNV-1a`, which is a 64 bit hash. I think we are seeing collisions, but I don't see a better way to test this than to disable the hashing completely. 

I would argue that a small LoRaWAN frame (think of about 1 byte per application payload) has a very low entropy, so maybe it's not impossible that this what we are seeing here.

#### Checklist

### DIFF
--- a/pkg/networkserver/grpc_gsns.go
+++ b/pkg/networkserver/grpc_gsns.go
@@ -755,7 +755,7 @@ func (ns *NetworkServer) handleDataUplink(ctx context.Context, up *ttnpb.UplinkM
 		matched *matchResult
 		ok      bool
 	)
-	const matchTTL = time.Minute
+	var matchTTL = ns.collectionWindow(ctx)
 	if err := ns.devices.RangeByUplinkMatches(ctx, up, matchTTL,
 		func(ctx context.Context, match *UplinkMatch) (bool, error) {
 			ctx = log.NewContextWithFields(ctx, log.Fields(
--- a/pkg/networkserver/redis/redis.go
+++ b/pkg/networkserver/redis/redis.go
@@ -17,10 +17,10 @@ package redis
 
 import (
 	"encoding/base64"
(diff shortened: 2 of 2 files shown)


KrishnaIyer (commented):
Does this require an `ns-db` migration for existing hashes? 
Also a note in the Changelog would be useful.

adriansmares:
> Does this require an `ns-db` migration for existing hashes?
> Also a note in the Changelog would be useful.

The keys which use the payload hash are ephemeral (they lived only one minute, and will do for a lot less after this PR).

descartes:
I've been directed here from a ticket I filed on behalf of a TTN user that had provided proof of packet loss. I have some understanding of the discussion above and on face value it appears to be related to deduplicating packets as they arrive in a much tigher timeframe than the OP's 5 minutes. Is this saying that for smaller packets there is a risk from the hashing that they may match another unrelated devices uplink (ie, another device anywhere within the eco-system). 

The technical fix is good, please be aware that someone has to explain this with some context. I can see how you arrived at this situation as I have a number of T shirts spanning back decades, but others may not be so forgiving.

If you'd prefer to cover this offline, PM me on the forum - @descartes.

KrishnaIyer (approved):

PR #4575, opened 2021-09-06
Add desired mac settings to end device general settings in the Console

### LINKED ISSUE #4290: Change some default values in the Console

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4290 and https://github.com/TheThingsNetwork/lorawan-stack/issues/2047

<details>

# OTAA:

## default

otaa-ns-settings

## class B

otaa-class-b-ns-settings

## class C

otaa-class-c-ns-settings

# ABP:

## default

abp-ns-settings

## default lorawan version 1.1.0

abp-lw-110-ns-settings

## class B

abp-class-b-ns-settings

## class C

abp-class-c-ns-settings

# MULTICAST:

## class B

multicast-class-b-ns-settings

## class C

multicast-class-c-ns-settings

</details>

#### Changes

- Add desired mac settings to end device general settings
- Update field titles
- Add tooltips for missing fields
- Remove descriptions for fields with tooltip content
- Add starting point for e2e tests

#### Testing

Manual

#### Checklist


pkg/webui/console/views/device-general-settings/network-server-form/index.js
```diff
@@ -62,6 +71,18 @@ const NetworkServerForm = React.memo(props => {
...
+  const [freqPlan, setFreqPlan] = React.useState(device.frequency_plan_id)
+  const freqPlanRef = React.useRef()
+  const handleFreqPlanChange = React.useCallback(band => {
+    setFreqPlan(band.value)
+  }, [])
+  const [phyVersion, setPhyVersion] = React.useState(device.lorawan_phy_version)
+  const phyVersionRef = React.useRef()
+  const handlePhyVersionChange = React.useCallback(version => {
+    setPhyVersion(version)
+  }, [])
```
```suggestion
  const handlePhyVersionChange = React.useCallback(setPhyVersion, [])
```

kschiffer:
I believe this needs attention from @johanstokking, to make sure the fields added here are correct and logical (specifically having both desired and actual values). So I'd suggest putting this on hold until he's back. Code-wise, it looks good to me.

johanstokking:
This looks good to me for ABP.

However, for OTAA, I miss most of the settings. I clarified https://github.com/TheThingsNetwork/lorawan-stack/issues/4290#issuecomment-888136034; for OTAA we need to configure a bunch of desired parameters as well.

bafonins:
@johanstokking So how should this look for OTAA?
Why should we allow editing only the `_desired` fields for OTAA devices and not the actual parameters, e.g. `mac_settings.rx2_data_rate_index` and `mac_settings.desired_rx2_data_rate_index`?

johanstokking:
> Why should we allow editing only the `_desired` fields for OTAA devices and not the actual parameters, e.g. `mac_settings.rx2_data_rate_index` and `mac_settings.desired_rx2_data_rate_index`?

Because RX2 data rate is part of the join-accept in OTAA. So it is already negotiated between NS and the end-device. Allowing users to customize it in the NS breaks downlink. The current value should not be changed.

With ABP there's no negotiation on this, so it should be configurable.

The desired value can always be configured; through MAC commands over downlink, the RX data rate index can be changed.

bafonins:
@johanstokking  just updated screenshots. Can you take a look and check if all the fields are present for each activation mode?

johanstokking:
Looks great.

@adriansmares can you check this as well please?

adriansmares (approved):
Looks good to me too. 

Should we mark every `current_parameters` as being current (`RX1 delay` -> `Current RX1 delay`) ?  I realize this is scope creep so feel free to do it in another PR.

johanstokking:
I'm following @adriansmares' judgement here and it looks like he already approved.

adriansmares (approved):

kschiffer (approved):

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -119,56 +184,130 @@ const MacSettingsSection = props => {
...
+                component={Input}
+                min={1}
+                max={15}
+                inputWidth="xxs"
+              />
+            )}
+            <Form.Field
+              title={m.desiredRx1DelayTitle}
+              type="number"
+              name="mac_settings.desired_rx1_delay"
```
kschiffer, line 213:
```suggestion
              tooltipId={tooltipIds.RX1_DELAY}
              name="mac_settings.desired_rx1_delay"
```

We can use the same tooltip.

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -119,56 +184,130 @@ const MacSettingsSection = props => {
...
+                max={7}
+                tooltipId={tooltipIds.DATA_RATE_OFFSET}
+              />
+            )}
+            <Form.Field
+              title={m.desiredRx1DataRateOffsetTitle}
+              type="number"
+              inputWidth="xxs"
+              name="mac_settings.desired_rx1_data_rate_offset"
+              component={Input}
```
kschiffer, line 239:
```suggestion
              component={Input}
              tooltipId={tooltipIds.DATA_RATE_OFFSET}
```

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -119,56 +184,130 @@ const MacSettingsSection = props => {
...
             max={15}
+            tooltipId={tooltipIds.RX2_DATA_RATE_INDEX}
+            inputWidth="xxs"
           />
+        )}
+        <Form.Field
+          title={m.desiredRx2DataRateIndexTitle}
+          type="number"
+          name="mac_settings.desired_rx2_data_rate_index"
+          component={Input}
```
kschiffer, line 274:
```suggestion
          tooltipId={tooltipIds.RX2_DATA_RATE_INDEX}
          component={Input}
```

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -119,56 +184,130 @@ const MacSettingsSection = props => {
...
           />
+        )}
+        <Form.Field
+          type="number"
+          min={100000}
+          step={100}
+          title={m.desiredRx2FrequencyTitle}
+          placeholder={m.frequencyPlaceholder}
+          name="mac_settings.desired_rx2_frequency"
+          component={Input}
```
kschiffer, line 301:
```suggestion
          tooltipId={tooltipIds.RX2_FREQUENCY}
          component={Input}
```

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -119,56 +184,130 @@ const MacSettingsSection = props => {
...
+        {!isOTAA && (
           <Form.Field
-            title={m.resetsFCnt}
-            onChange={handleResetsFCntChange}
-            warning={resetsFCnt ? m.resetWarning : undefined}
-            name="mac_settings.resets_f_cnt"
-            tooltipId={tooltipIds.RESETS_F_CNT}
-            component={Checkbox}
+            title={m.maxDutyCycle}
+            name="mac_settings.max_duty_cycle"
```
kschiffer, line 309: This one is missing a tooltip. Can we add one? You can use the same for `desired_max_duty_cycle` as well then.

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -189,16 +352,128 @@ const MacSettingsSection = props => {
...
+                component={Input}
+              />
+            )}
+            <Form.Field
+              type="number"
+              min={100000}
+              step={100}
+              title={m.desiredPingSlotFrequencyTitle}
+              placeholder={m.frequencyPlaceholder}
+              name="mac_settings.desired_ping_slot_frequency"
```
kschiffer, line 411:
```suggestion
              name="mac_settings.desired_ping_slot_frequency"
              tooltipId={tooltipIds.PING_SLOT_FREQUENCY}
```

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -189,16 +352,128 @@ const MacSettingsSection = props => {
...
+                component={Input}
+                type="number"
+                min={0}
+                max={15}
+              />
+            )}
+            <Form.Field
+              title={m.desiredPingSlotDataRateTitle}
+              name="mac_settings.desired_ping_slot_data_rate_index"
+              component={Input}
```
kschiffer, line 432:
```suggestion
              tooltipId={tooltipIds.PING_SLOT_DATA_RATE_INDEX}
              component={Input}
```

kschiffer, line 432:
```suggestion
              tooltipId={tooltipIds.PING_SLOT_DATA_RATE_INDEX}
              component={Input}
```

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -189,16 +352,128 @@ const MacSettingsSection = props => {
...
+        </>
+      )}
+      {isABP && (
+        <>
+          <Form.FieldContainer horizontal>
+            <Form.Field
+              className={style.smallField}
+              title={m.statusCountPeriodicity}
+              name="mac_settings.status_count_periodicity"
+              component={Input}
```
kschiffer: Can we also add a tooltip for this one?
cc @johanstokking can you provide?

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -189,16 +352,128 @@ const MacSettingsSection = props => {
...
+              type="number"
+              inputWidth="xxs"
+            />
+            <Form.Field
+              className={style.smallField}
+              title={m.statusTimePeriodicity}
+              name="mac_settings.status_time_periodicity"
```
kschiffer: Same here.

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -189,16 +352,128 @@ const MacSettingsSection = props => {
...
           <Form.Field
-            type="number"
-            min={100000}
-            step={100}
-            title={m.pingSlotFrequencyTitle}
-            description={m.pingSlotFrequencyDescription}
-            placeholder={m.frequencyPlaceholder}
-            name="mac_settings.ping_slot_frequency"
-            component={Input}
+            name="mac_settings.use_adr"
```
kschiffer: Please add a tooltip for this:

> Controls whether the device uses adaptive data rate. This will allow the network to adjust the employed data rate based on signal to noise ratio. This adaptively optimizes energy consumption, bandwith and transmission power.

Add a documentation link to https://www.thethingsindustries.com/docs/reference/adr

kschiffer (commented):
Please see my comment on adding more tooltips for the fields. If possible please add those before merging. There are a couple `desired_*` fields that can use the same tooltip as the normal field.  
Please source the missing tooltip texts from @johanstokking's comment [here](https://github.com/TheThingsNetwork/lorawan-stack/issues/4290#issuecomment-887790141).

pkg/webui/console/components/mac-settings-section/index.js
```diff
@@ -189,16 +352,128 @@ const MacSettingsSection = props => {
...
+            component={Checkbox}
+            label={m.useAdr}
+            onChange={handleUseAdrChange}
           />
+          {useAdr && (
+            <>
+              <Form.Field
+                title={m.adrMargin}
+                name="mac_settings.adr_margin"
```
kschiffer: Please add a tooltip for this:

> Signal-to-noise ratio (SNR) margin in dB that is taken into account in the Adaptive Data Rate (ADR) algorithm to optimize the data rate of the end device. A higher margin requires the end device to have a better SNR before the Network Server instructs a higher data rate for the end device to use.

(via @johanstokking )

Add a documentation link to [https://www.thethingsindustries.com/docs/reference/adr](https://www.thethingsindustries.com/docs/reference/adr)

PR #4588, opened 2021-09-06
Log drain errors only if the send did not fail

adriansmares:
#### Summary

References https://sentry.io/organizations/the-things-industries/issues/2628889957/?project=2682566&query=is%3Aunresolved

#### Changes

- Do not log the uplink send errors as `ERROR` unless they are related to the task logic

#### Testing

N/A

##### Regressions

N/A

#### Notes for Reviewers

We report an error even if the Application Server cannot actually process the uplinks. This PR makes it such that such errors are logged as `WARN`, while the rest are `ERROR`.

#### Checklist


pkg/networkserver/grpc_asns.go
```diff
@@ -99,10 +100,17 @@ func (ns *NetworkServer) processApplicationUplinkTask(ctx context.Context, consu
 		}
 
 		cl := ttnpb.NewNsAsClient(conn)
+		var sendErr bool
```
KrishnaIyer, line 103: Heavy Nit: `sendErr` sounds like an error that needs to be sent, like we use `retErr` for errors to be returned.

KrishnaIyer (approved):

PR #4591, opened 2021-09-07
Individual band versions

### LINKED ISSUE #4346: Refactor to band package to maintain individual band versions

adriansmares:
#### Summary

Closes #4346 

#### Changes

- Introduce bands ground truth via JSON serialization
  - Most of the data has been serialized, minus 3 functions (`GenerateChMasks`/`ParseChMask` and `Beacon.ComputeFrequency`)
  - This means that you should have only to review the first commit, which introduces this serialization and verification process
- Bands are now stored in a an `ID-Version` fashion - it is not possible to access bands without specifying a version (or cheating and requesting the last one, explicitly)
- The call sites of `band.GetByID` have been updated to use `band.Get` (explicit version) or at least `band.GetLatest` (latest version)

#### Testing

Unit testing, specifically I've exported the already existing bands as JSON into `testdata`. There are only 3 fields which cannot be exported (they are rather complex functions). The new individual band definitions then are compared against these JSON definitions.

##### Regressions

There should be none if the testing methodology is right.

#### Notes for Reviewers

Please verify your code owned files in the first 4 commits - you may skip the commits that add individual band definitions. 

The only downgrade which I could have gotten wrong is `disableChMaskCntl5`, which changes `GenerateChMasks` to `makeGenerateChMask72(false)` from `makeGenerateChMask72(true)`. This happens in the `AU915` and `US915` bands in versions `RP1_V1_0_2B` or lower. The rest is copy paste at heart.

#### Checklist


htdvisser (approved):
Code-owned files in the first 4 commits look good to me. Haven't reviewed testdata (don't know why that is code-owned).

johanstokking:
> The only downgrade which I could have gotten wrong is `disableChMaskCntl5`, which changes `GenerateChMasks` to `makeGenerateChMask72(false)` from `makeGenerateChMask72(true)`. This happens in the `AU915` and `US915` bands in versions `RP1_V1_0_2B` or lower. The rest is copy paste at heart.

What could have gone wrong here? I sample-reviewed this and it looks good.

Indeed, ChMaskCntl 5 has been introduced in LoRaWAN Regional Parameters RP1 1.0.3 revision A. It allows for more efficient ChMask control when using 8 channels I believe.

adriansmares:
> 
> What could have gone wrong here? I sample-reviewed this and it looks good.
> 
> Indeed, ChMaskCntl 5 has been introduced in LoRaWAN Regional Parameters RP1 1.0.3 revision A. It allows for more efficient ChMask control when using 8 channels I believe.

Any other field is checked from the `testdata`, but the `ChMaskCntl 5` support is a function that I couldn't dump. As such, if I failed copy pasting the right parameters to `makeGenerateChMask72` it would behave erroneously and the `testdata` wouldn't catch it. If you sampled the bands and I didn't miss it, we're fine.

johanstokking (approved):

PR #4600, opened 2021-09-08
Fix firewall message rate limitting

adriansmares:
#### Summary

This PR fixes the `Packet Filtered` message spam when a gateway attempts to send too many packets per second.

#### Changes

- Log the `Packet Filtered` message only if the error is not an resource exhausted one, or the rate limitting allows it.

#### Testing

N/A

##### Regressions

N/A

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/udp/udp.go
+++ b/pkg/gatewayserver/io/udp/udp.go
@@ -171,10 +171,9 @@ func (s *srv) handlePackets() {
 
 			if s.firewall != nil {
 				if err := s.firewall.Filter(packet); err != nil {
-					if errors.IsResourceExhausted(err) && ratelimit.Require(s.limitLogs, ratelimit.NewCustomResource(eui.String())) == nil {
-						break
+					if !errors.IsResourceExhausted(err) || ratelimit.Require(s.limitLogs, ratelimit.NewCustomResource(eui.String())) == nil {
+						logger.WithError(err).Warn("Packet filtered")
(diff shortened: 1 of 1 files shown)


pkg/gatewayserver/io/udp/udp.go
```diff
@@ -171,10 +171,9 @@ func (s *srv) handlePackets() {
 
 			if s.firewall != nil {
 				if err := s.firewall.Filter(packet); err != nil {
-					if errors.IsResourceExhausted(err) && ratelimit.Require(s.limitLogs, ratelimit.NewCustomResource(eui.String())) == nil {
-						break
+					if !errors.IsResourceExhausted(err) || ratelimit.Require(s.limitLogs, ratelimit.NewCustomResource(eui.String())) == nil {
```
KrishnaIyer, line 174: So we never want to log the `ResourceExhausted` error?

Wouldn't the following work for our case?
```
errors.IsResourceExhausted(err) && ratelimit.Require(s.limitLogs, ratelimit.NewCustomResource(eui.String())) != nil
```

adriansmares: If the error is not `ResourceExhausted`, we always log (`||` shorts).

If the error is `ResourceExhausted`, the first part of the `if` is false, so we check the rate limitting. If that one is fine, we log.

KrishnaIyer (approved):

PR #4605, opened 2021-09-09
RPC to fetch supported PHY versions for Band ID

### LINKED ISSUE #4518: Expose `band` pkg via API to enable sourcing FP constraints
#### Summary
We need to come up with a way to expose the restrictions of frequency plans with regards to MAC/PHY versions.

#### Why do we need this?
E.g. the Australian FPs do not support MAC/PHY 1.0.0. However, the Console still allows setting this combination because the frontend is unaware of such restrictions. This leads to frustration when users fill out the end device registration form and hitting this limitation only upon submitting.

image

- End device registration form

- API to query from the Console to get the `band -> regional parameters version -> available` mapping

Probably trivial. cc @adriansmares 

Manual testing and updating e2es.

Need help from @adriansmares (or whoever else feels confident with this).

KrishnaIyer:
#### Summary

RPC to fetch supported PHY versions for Band ID. Closes #4518 

#### Changes

- Add a new `GetSupportedPhyVersions` to the configuration server.
- Implement RPC and add tests.

#### Testing

UT

##### Regressions

NA

#### Notes for Reviewers

The implementation can change a little after https://github.com/TheThingsNetwork/lorawan-stack/pull/4591 but the API will not.

#### Checklist


adriansmares (commented):
This LGTM overall, just one question:

Do we want to manually fetch the versions of the band individually, or should we have a RPC that returns _all_ bands and their associated versions ?

cc @kschiffer

kschiffer:
> Do we want to manually fetch the versions of the band individually, or should we have a RPC that returns _all_ bands and their associated versions ?

Yeah, so in the Console the user selects the MAC and PHY version before selecting the FP. So to be able to use this RPC, we'd have to change the order of the fields. I think this is a good idea anyway given the structure of the constraints.

Being able to retrieve all bands and associated versions at once would save us a request whenever the selection changes, making the form a bit more responsive. So if it's easy to add, I'd prefer having that as well.

KrishnaIyer:
> Being able to retrieve all bands and associated versions at once would save us a request whenever the selection changes, making the form a bit more responsive. So if it's easy to add, I'd prefer having that as well.

@kschiffer If I understand you correctly, you'd want an additional RPC to fetch all bands or modify this one to do that?

Either one is quite easy to do so I'll update the PR based on what works best for you.

kschiffer:
With regards to the usage for the Console, just having an RPC to retrieve all bands and associated versions is sufficient.

KrishnaIyer:
Ok thanks. That's done now

adriansmares (approved):

pkg/band/rpc.go
```diff
@@ -0,0 +1,56 @@
...
+
+package band
+
+import (
+	"context"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+// TODO: Replace this with the `band.All` logic in https://github.com/TheThingsNetwork/lorawan-stack/pull/4591
```
htdvisser, line 23: Please add a note on that PR to actually do this.

htdvisser (commented):
LGTM

htdvisser (approved):
LGTM

PR #4607, opened 2021-09-09
Optimize Application Server worker pools

### LINKED ISSUE #4606: Optimize Application Server worker pools
#### Summary

_A worker pool is just a pool of goroutines for the purpose of this issue._

The Application Server has multiple worker pools:
- The worker pool used by the webhooks frontend
- The worker pool used by the application packages frontend - each package has a pool

These pools are _static_, the number of workers is always the same.

#### Why do we need this?

As the load on the application server increases, the network administrator has one main knob to tune - the number of workers for each pool. This is rather inefficient - the utilization of the workers is opaque, and it is not clear when the limits are hit / messages are lost.

For the webhooks frontend, we have `QueuedSink`, which has a static worker count.

For the application packages frontend, we have the following pipeline:
- An uplink arrives via a channel
- The AS checks which application packages are applicable for this uplink (i.e. which ones are enabled, such as LoRaCloud DAS/DMS)
- Now that the AS knows which package is applicable, it pushes the uplink via a channel to the worker pool of the package
  - Each package has its own pool, in order to avoid splitting the pool between packages

I'm interested in a worker pool that is like the one used in the Gateway Server, and the Packet Broker Agent, which has the following semantics:

- Workers automatically shutdown after having no work available for a period of time
- Workers are created on demand, if there is no worker currently available to process the uplink

An extra requirement is that we carry around the original context, but we decouple it (i.e. we keep every value in it, but the lifetime becomes that of the `component.Component`).

We don't have generics yet, but we can have a generic worker pool that passes around `interface{}`. Eventually, even the queue in the GS/PBA can be replaced with this queue.

#### Environment

`v3.15`

I would like this queue to be a top level package, and that the hand rolled implementations are replaced by it.

The implementation works as follows:
- Publishers interact with the pool using a `Publish` function, instead of using a channel directly.
  - `Publish` itself will spawn the workers if needed.
- Subscribers exist as handlers. Specifically, someone that wants to consume data from the publishers must create a handler _factory_. The factory function is used in order to create the handler that actually consumes the message.
  - The idea here is that we would like to have some kind of state per worker, and this factory allows us to do this. The main reason why I'm not providing the handler function directly is that I would like that each individual worker used by the webhooks frontend has it's own `*http.Client`.

Current unit tests cover the _correctness_ of the queue. 

What we really need here for testing is _metrics_: workers started and stopped, and the number of processed / failed messages.

Yes.

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4606

#### Changes

- Add `pkg/workerpool`, which contains the generic worker pool that spawns workers on demand
- Refactor the application packages frontend to use the `pkg/workerpool` pool instead of the manually spawned tasks
  - A small deviation that you will see here is that there is a worker pool added between the `io.Subscription` and the actual worker pools of each package. This is intentional and allows us to decide to which package we need to send a particular uplink to in parallel. 
- Refactor the webhooks frontend to use the `pkg/workerpool` pool instead of the manually spawned goroutines

#### Testing

##### Regressions

This changes the piping used by the Application Server worker pools. The unit tests cover this.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ For details about compatibility between different releases, see the **Commitment
 - Update to Go 1.17.
 - LBS timestamp rollover threshold.
 - Layout of error pages.
+- The Application Server worker pools may now drop workers if they are idle for too long.
 
 ### Deprecated
 
--- a/config/messages.json
+++ b/config/messages.json
@@ -2420,15 +2420,6 @@
       "file": "templates.go"
     }
   },
-  "error:pkg/applicationserver/io/web:queue_full": {
-    "translations": {
-      "en": "the queue is full"
-    },
-    "description": {
-      "package": "pkg/applicationserver/io/web",
-      "file": "webhooks.go"
-    }
-  },
   "error:pkg/applicationserver/io/web:rate_limit_exceeded": {
     "translations": {
       "en": "rate limit exceeded"
--- a/config/messages.json
+++ b/config/messages.json
@@ -8477,6 +8477,15 @@
       "file": "cookie.go"
     }
   },
+  "error:pkg/workerpool:pool_full": {
+    "translations": {
+      "en": "the worker pool is full"
+    },
+    "description": {
+      "package": "pkg/workerpool",
+      "file": "workerpool.go"
+    }
+  },
   "event:account.user.login_failed": {
     "translations": {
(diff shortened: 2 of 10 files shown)


KrishnaIyer (approved):
Fantastic. Looks good to me.

PR #4608, opened 2021-09-10
Redirect back to login when OAuth authorization request is denied

### LINKED ISSUE #4564: Cancelling The Things ID login leads to "Forbidden" error
#### Summary
Cancelling login via The Things ID leads to "Forbidden" error

#### What do you see now?

If you click **Cancel**, you get:

Instead of this error, a user should probably be redirected to the login page

#### Environment
TTS Community Edition

@kschiffer?

kschiffer:
#### Summary

This PR improves the UX when canceling the OAuth authorization request by redirecting back to the login page instead of rendering an error.

Closes #4564

#### Changes
- Make the `oauthclient` utility redirect to the `rootURL` when the OIDC-authentication was aborted. This will lead to another redirect back to the login page.

#### Testing

Manual testing by modifying the database to not make the console client skip the authorization, similar to how we do it in our end-to-end tests.

##### Regressions

I don't think there should be any.

#### Checklist

### DIFF
--- a/pkg/web/oauthclient/callback.go
+++ b/pkg/web/oauthclient/callback.go
@@ -59,6 +59,16 @@ func (oc *OAuthClient) HandleCallback(c echo.Context) error {
 	if err := c.Bind(&response); err != nil {
 		return err
 	}
+	if response.Error == "access_denied" {
+		// The `access_denied` error usually means that the authorization request
+		// was actively denied by the user. In that case, it is best to redirect
+		// back to the page root, which will enable to restart the authorization
+		// process if wished.
+		ctx := c.Request().Context()
+		config := oc.configFromContext(ctx)
+		oc.removeStateCookie(c)
+		return c.Redirect(http.StatusFound, config.RootURL)
+	}
 	if err := response.ValidateContext(c.Request().Context()); err != nil {
 		return err
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ For details about compatibility between different releases, see the **Commitment
 - Not rendering site header and footer for error pages in some situations.
 - Not providing a copy button for error pages in some situations.
 - Improved errors for invalid URLs.
+- Rendering an error in the Console after the OIDC-authentication was cancelled instead of redirecting back to the login page.
 
 ### Security
 
--- a/cypress/integration/smoke/authorization/index.js
+++ b/cypress/integration/smoke/authorization/index.js
@@ -65,13 +65,11 @@ const abortAuthorization = defineSmokeTest(
     cy.findByLabelText('Password').type(`${user.password}{enter}`)
 
     // Deny authorization.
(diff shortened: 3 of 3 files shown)


pkg/oauth/oauth.go
```diff
@@ -136,6 +136,13 @@ func (s *server) Authorize(authorizePage echo.HandlerFunc) echo.HandlerFunc {
 			switch c.Request().Method {
 			case http.MethodPost:
 				ar.Authorized, _ = strconv.ParseBool(req.PostForm.Get("authorize"))
+				if !ar.Authorized {
+					events.Publish(evtAccessTokenDeleted.NewWithIdentifiersAndData(req.Context(), &session.UserIdentifiers, nil))
+					err = s.store.DeleteSession(req.Context(), &session.UserIdentifiers, session.SessionId)
+					if err != nil && !errors.IsNotFound(err) {
+						return err
+					}
+				}
```
htdvisser: This doesn't look quite right. If on the authorization screen I click 'reject', there shouldn't be an access token to delete, and I shouldn't be logged out either.

kschiffer: EDIT: Disregard

pkg/web/oauthclient/callback.go
```diff
@@ -39,8 +38,18 @@ type oauthAuthorizeResponse struct {
...
+		if res.Error == "access_denied" {
+			// The `access_denied` error usually means that the authorization request
+			// was actively denied by the user. In that case, it is best to redirect
+			// back to the page root, which will enable to restart the authorization
+			// process if wished.
+			ctx := c.Request().Context()
+			config := oc.configFromContext(ctx)
+			oc.removeStateCookie(c)
+			return c.Redirect(http.StatusFound, config.RootURL)
+		}
```
htdvisser: I don't think this code belongs in the validator of the response message. I think it would be better to return a specific access denied error, and matching on that in `HandleCallback`

htdvisser (changes requested):

pkg/web/oauthclient/callback.go
kschiffer: I think there are two occasions where (in our setup) `access_denied` errors happen, which is:

1.  if the authorization was denied in the authorization screen of the Account App
2.  [If the OIDC login is canceled](https://github.com/TheThingsIndustries/the-things-id/blob/c82310c7a8e695e7dcb2c0f069f67254c213b711/src/oidc/oidc.controller.ts#L138-L141)

In both situations, it is good to redirect back to the root instead of rendering an error.

Since we don't use error codes in the OAuth server within TTID, we could currently only match the specific error via the error description `User aborted interaction`.

htdvisser (approved):
The code is not incorrect, so I'll approve it.

I don't think this is the right user experience, since we're hiding errors from them. But @kschiffer is the UX expert, so 🤷‍♂️

PR #4627, opened 2021-09-15
Fix state cookie Max-Age

kschiffer:
#### Summary
This quickfix PR fixes the set `Max-Age` property of the state cookie.

#### Changes

- Fix `MaxAge` calculation in the cookie util
- ~~Set `MaxAge` to 0 (meaning unlimited) for the OAuth state cookie~~

#### Testing

Manual testing.

#### Notes for Reviewers
Two things here:
- The calculation of the time period in the cookie util was wrong, it is expecting an integer of seconds, but got Nanoseconds / 1000 instead. This caused the state cookie to be valid for years, instead of 10 minutes as set in the oauth package.
- However, also 10 minutes is not desirable, since it would mean that the login screen would lose effectiveness after 10 minutes, leading to a error, since the state cookie has since expired. I think it is fine to not set any expiry for the state cookie. The official OAuth specification also makes no note on any recommended expiry.

So in practice, this change should not have any effect unless the login screen is kept open for years. Still, I wanted to fix this quirk.

#### Checklist

### DIFF
--- a/pkg/web/oauthclient/cookie_state.go
+++ b/pkg/web/oauthclient/cookie_state.go
@@ -17,7 +17,6 @@ package oauthclient
 import (
 	"encoding/gob"
 	"net/http"
-	"time"
 
 	echo "github.com/labstack/echo/v4"
 	"go.thethings.network/lorawan-stack/v3/pkg/random"


pkg/web/cookie/cookie.go
```diff
@@ -52,7 +52,7 @@ func (d *Cookie) new(r *http.Request) http.Cookie {
 	return http.Cookie{
 		Name:     d.Name,
 		Path:     d.Path,
-		MaxAge:   int(d.MaxAge.Nanoseconds() / 1000),
+		MaxAge:   int(d.MaxAge.Seconds()),
```
htdvisser: Oh seriously? This has been here since 2018?
https://github.com/TheThingsNetwork/lorawan-stack/commit/9b3570a9cd25599b8d0f6999ae3c32b9a014f83d

pkg/web/oauthclient/cookie_state.go
```diff
@@ -44,7 +43,7 @@ func (oc *OAuthClient) StateCookie() *cookie.Cookie {
 		Name:     oc.config.StateCookieName,
 		HTTPOnly: true,
 		Path:     oc.getMountPath(),
-		MaxAge:   10 * time.Minute,
+		MaxAge:   0,
```
htdvisser: I don't think this is a good idea. If you don't finish your login within a certain duration, we should not allow it anymore.

pkg/web/cookie/cookie.go
kschiffer: Yes... Only affected the state cookie though, so not really dramatic.

pkg/web/oauthclient/cookie_state.go
kschiffer: Why?

You have to regard all of this in the context of our own internal OAuth clients that use the oauthclient package (such as the Console), for which we apply some special policies already since they are trusted by us. I don't think expiring the login is of any benefit and only introduces friction. I don't want users to be able to use the Console login only within 10 minutes of loading the page and I also don't want them to be consequently greeted by an error message telling them that the login failed because of some expiration that makes no sense to them (since for them logging into the OAuth provider and the Console is one and the same thing).

I get that for a normal OAuth flow that is initiated by a third party that we would apply different measures but not for our own internal clients.

Otherwise, the only way I see this implemented in a UX friendly way is the `oauthclient` package automatically redirecting to the Console root URL that would initiate the login flow again and directly log the user in, since the user is already logged into the Account App. I chose not to do this because it could result in redirect loops when the state cookie is not present for other reasons than expiry.

kschiffer:
Currently blocked on #4700

kschiffer:
With #4700 merged. We can now also merge this one.

kschiffer:
Nevermind, this seems to have been resolved already in the meantime.

PR #4634, opened 2021-09-16
Publish Tx Ack failure result as string

### LINKED ISSUE #4336: Improvise debug details of the gs.down.tx.fail event
#### Summary
The “value” in the data object of the `gs.down.tx.fail` event currently doesn’t provide useful information to know/identify the cause of the failure.
Ref: https://github.com/TheThingsIndustries/lorawan-stack-support/issues/457

#### Steps to Reproduce
....

#### What do you see now?
The value in the data object of the `gs.down.tx.fail` event corresponds to the “Result” enum: https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.13.1/api/messages.proto#L153-L163

It should give the information on the cause for the downlink failure.

#### Environment
TTS V3.13.2

...

...
No, @KrishnaIyer

KrishnaIyer:
#### Summary

Closes #4336 

#### Changes

- Use the `.String()` method of the result in the event.

#### Testing

Local

##### Regressions

NA

#### Checklist

### DIFF
--- a/pkg/gatewayserver/observability.go
+++ b/pkg/gatewayserver/observability.go
@@ -202,7 +202,7 @@ var gsMetrics = &messageMetrics{
 			Name:      "downlink_tx_failed_total",
 			Help:      "Total number of unsuccessfully emitted downlinks",
 		},
-		[]string{protocol},
+		[]string{protocol, "result"},
 	),
 }
 
@@ -335,5 +335,5 @@ func registerSuccessDownlink(ctx context.Context, gtw *ttnpb.Gateway, protocol s
 
 func registerFailDownlink(ctx context.Context, gtw *ttnpb.Gateway, ack *ttnpb.TxAcknowledgment, protocol string) {
 	events.Publish(evtTxFailureDown.NewWithIdentifiersAndData(ctx, gtw, ack.Result.String()))
-	gsMetrics.downlinkTxFailed.WithLabelValues(ctx, protocol).Inc()
+	gsMetrics.downlinkTxFailed.WithLabelValues(ctx, protocol, ack.Result.String()).Inc()
 }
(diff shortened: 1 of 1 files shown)


pkg/gatewayserver/observability.go
```diff
@@ -334,6 +334,6 @@ func registerSuccessDownlink(ctx context.Context, gtw *ttnpb.Gateway, protocol s
 }
 
 func registerFailDownlink(ctx context.Context, gtw *ttnpb.Gateway, ack *ttnpb.TxAcknowledgment, protocol string) {
-	events.Publish(evtTxFailureDown.NewWithIdentifiersAndData(ctx, gtw, ack.Result))
```
adriansmares, line 337: Can we add `result` as a label to `downlinkTxFailed` and record the result please ?

adriansmares (approved):

pkg/gatewayserver/observability.go
KrishnaIyer: Yeah I thought of that but then we add potentially 8 new series of data and we've never really used this for debugging. Doesn't hurt I guess.

adriansmares: It would be nice for https://github.com/TheThingsNetwork/lorawan-stack/issues/18

pkg/gatewayserver/observability.go
```diff
@@ -202,7 +202,7 @@ var gsMetrics = &messageMetrics{
 			Name:      "downlink_tx_failed_total",
 			Help:      "Total number of unsuccessfully emitted downlinks",
 		},
-		[]string{protocol},
+		[]string{protocol, "result"},
```
adriansmares, line 205: Nit: constant like `host` or `protocol`.

adriansmares (approved):

pkg/gatewayserver/observability.go
KrishnaIyer: Technically we use those constants for repeated labels.

PR #4642, opened 2021-09-20
Validate data rate indices only on change

adriansmares:
#### Summary

References https://sentry.io/organizations/the-things-industries/issues/2660268652/events/4831c2b73818491f845ee73dc3c0bf03/?environment=ttn-prod-au1&project=2682566

#### Changes

- Validate the ADR Data Rate Index and ADR Tx Power Index only if they are actually going to be changed
  - This allows us to schedule channel changes even if the device is currently not requiring data rate / TX power changes
- Make sure that the loop that unrolles the data rate changes always terminates
- Allow `NbTrans` / channel mask changes even if we cannot change the data rate index / TX power index

#### Notes for Reviewers

This is a rather tricky issue and I'm not sure how to really touch it. This PR stems from our discussion on Slack that we shouldn't do ADR in non-125KHz channels yet, but at the same time we do need to schedule the `LinkADRReq` in order to enable/disable some channels.

Take the following diff between the current (left) and desired (right) MAC state of a `1.0.2B` device in `AU915_FSB_2`:
https://www.diffchecker.com/OF3E58L1

With the change that I've proposed, we will send a `LinkADRReq` that enables / disables the channels as seen in the diff, but which doesn't change the ADR data rate index / TX power. The catch ? All of the enabled uplink channels have `MaxDataRateIndex = 5`, and we are sending a `LinkADRReq` with `DataRateIndex = 6`. 

What should the device do with such a command ? Hop onto the 125 KHz channels ? Reject the `DataRateIndex` / `TxPowerIndex` ? I start to think that our desired channels are actually problematic here, since perhaps the desired channels should be in the same KHz range.

#### Checklist

### DIFF
--- a/pkg/networkserver/mac/link_adr.go
+++ b/pkg/networkserver/mac/link_adr.go
@@ -116,32 +116,55 @@ func generateLinkADRReq(ctx context.Context, dev *ttnpb.EndDevice, phy *band.Ban
 		nbTrans    uint32
 	)
 	minDataRateIndex, maxDataRateIndex, ok := channelDataRateRange(dev.MacState.DesiredParameters.Channels...)
-	if !ok ||
-		dev.MacState.DesiredParameters.AdrTxPowerIndex > uint32(phy.MaxTxPowerIndex()) ||
-		dev.MacState.DesiredParameters.AdrDataRateIndex > phy.MaxADRDataRateIndex {
+	if !ok {
 		return linkADRReqParameters{}, false, ErrCorruptedMACState.
-			WithAttributes(
-				"ok", ok,
-				"adr_tx_power_idx", dev.MacState.DesiredParameters.AdrTxPowerIndex,
-				"phy_max_tx_power", phy.MaxTxPowerIndex(),
-				"adr_data_rate_idx", dev.MacState.DesiredParameters.AdrDataRateIndex,
-				"phy_max_adr_data_rate_index", phy.MaxADRDataRateIndex,
-			).
 			WithCause(ErrChannelDataRateRange)
 	}
-	if dev.MacState.DesiredParameters.AdrDataRateIndex < minDataRateIndex || dev.MacState.DesiredParameters.AdrDataRateIndex > maxDataRateIndex {
-		return linkADRReqParameters{}, false, ErrCorruptedMACState.
-			WithAttributes(
-				"adr_data_rate_index", dev.MacState.DesiredParameters.AdrDataRateIndex,
-				"min_data_rate_index", minDataRateIndex,
-				"max_data_rate_index", maxDataRateIndex,
-			).
-			WithCause(ErrChannelDataRateRange)
+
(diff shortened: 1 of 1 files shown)


johanstokking:
> All of the enabled uplink channels have `MaxDataRateIndex = 5`, and we are sending a `LinkADRReq` with `DataRateIndex = 6`.

This isn't a problem right?

But why would we disallow 250 and 500 KHz data rates through `LinkADRReq`?

We shouldn't be instructing the device to use those data rates through ADR, but if the device doesn't use ADR it may use those channels by its own discretion.

adriansmares:
> > All of the enabled uplink channels have `MaxDataRateIndex = 5`, and we are sending a `LinkADRReq` with `DataRateIndex = 6`.
> 
> This isn't a problem right?

Is this valid per spec ? How does the end device interpret this ?

johanstokking:
> Is this valid per spec ? How does the end device interpret this ?

To not use that data rate in future uplink frames at all, regardless of ADR.

adriansmares:
In that case I think that this fix is adequate. It allows us to control the channels without requesting any ADR changes from the device.

adriansmares:
To answer the question regarding the disabled channels - this is per NS desired channels generation at https://github.com/TheThingsNetwork/lorawan-stack/blob/0e4ff01fb521922d1b4a556597f79b5d87fe2f20/pkg/networkserver/mac/utils.go#L524
The other 500KHz channels are disabled since they are not part of the uplink channels of the [frequency plan](https://github.com/TheThingsNetwork/lorawan-frequency-plans/blob/master/AU_915_928_FSB_2.yml).

Should we also enable uplinks via the `LoRaStandardChannel` / `FSKChannel` if the frequency plan supports it ? That would enable uplinks via the `917500000` channel in this particular case.

johanstokking:
> The other 500KHz channels are disabled since they are not part of the uplink channels of the [frequency plan](https://github.com/TheThingsNetwork/lorawan-frequency-plans/blob/master/AU_915_928_FSB_2.yml?rgh-link-date=2021-09-22T16%3A42%3A21Z).

Ah right, that is correct behavior then.

> Should we also enable uplinks via the `LoRaStandardChannel` / `FSKChannel` if the frequency plan supports it ? That would enable uplinks via the `917500000` channel in this particular case.

Hmm. I would say, if the frequency plan supports it, then yes.

But I'm not sure what the consequence is; probably a bunch of `NewChannelReq` and `LinkADRReq` with channel mask? And then channel hopping using the LoRa standard channel? FSK isn't really used so let's not bother with that.

adriansmares:
I've looked into adding the LoRa Standard Channel to the desired channels list and it's indeed not as easy as it may sound, since this is an uplink-only channel. I'll have to add support for uplink only channels when I add LR-FHSS anyway so I'll revisit this then.

@johanstokking please review the change here and let's visit the uplink-only channels in another PR.

johanstokking (approved):
Looks good!

Issue #14, opened 2019-01-29
Clean up component registries when applications got deleted


johanstokking:
**Summary:** 

When an application is deleted from IS, the NS and AS registries should delete the application and associated devices.

**Steps to Reproduce:** 

1. User deletes application
2. AS deletes all devices and deletes application

**What do you see now?**

Application gets deleted but devices stay there.

**What do you want to see instead?**

Devices be gone.

**How do you propose to implement this?**

An asynchronous garbage collection process in the NS and AS that deletes stale applications and its devices. This does not require an event stream, or even worse a connection, from IS to NS/AS, but rather have them be a client and check periodically if the applications are still there. This means that, as long as the garbage process has not run, unique device identifiers (i.e. JoinEUI, DevEUI) remain in the database, making it impossible to register a new device with the same identifiers.

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/740 by @johanstokking

nejraselimovic:
@johanstokking I tried creating an end device in the Console, deleting it, then re-adding it with the identical info, and I encountered no issues. Does this mean this issue isn't relevant anymore?

KrishnaIyer:
> I tried creating an end device in the Console, deleting it, then re-adding it with the identical info, and I encountered no issues. 

That's because the console takes care of also deleting the device in the NS and AS at the same time as the IS iirc. 
This is still possible to do via the CLI or our API so I don't think this is already covered.

johanstokking:
@nejraselimovic the end device delete and add with same identifiers works already indeed.

This issue is about deleting an application in the IS, and not having that deletion propagated to AS, NS and JS. These components then still hold unique identifiers (like DevEUI), that cannot be managed anymore because the application and its security context is gone from IS.

I'm reassigning this to @pgalic96 because he has been working on purging entities, which is a related topic.

NicolasMrad:
https://discuss.thethingsindustries.com/t/design-document-for-application-data-garbage-control/613

PR #4653, opened 2021-09-21
Application Server garbage collection

### LINKED ISSUE #2735: Pub/sub error for non-existent application

pgalic96:
#### Summary

Closes #2735
References #14 

#### Changes

- Add `as-db cleanup` command
- Add `Range` to registry interfaces and redis implementation
- Add IS store methods for fetching all applications / devices

#### Testing

To be added, that is why I am making this a draft PR

##### Regressions

This is a new command with new functionality, if the complements are not correctly calculated we can delete live data.

#### Notes for Reviewers

Looking for some feedback on implementation while I am writing tests. I am also still trying to find the right level of abstraction (how to structure the cleaners across various registries in interface).

#### Checklist


cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,141 @@
...
+	asiopsredis "go.thethings.network/lorawan-stack/v3/pkg/applicationserver/io/pubsub/redis"
+	"go.thethings.network/lorawan-stack/v3/pkg/applicationserver/io/web"
+	asiowebredis "go.thethings.network/lorawan-stack/v3/pkg/applicationserver/io/web/redis"
+	asredis "go.thethings.network/lorawan-stack/v3/pkg/applicationserver/redis"
+	"go.thethings.network/lorawan-stack/v3/pkg/identityserver/store"
+	"go.thethings.network/lorawan-stack/v3/pkg/redis"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+func toApplicationSet(list []*ttnpb.ApplicationIdentifiers) (set map[*ttnpb.ApplicationIdentifiers]struct{}) {
```
adriansmares: These sets should have unique identifiers in them, not pointers to identifiers. The pointers won't even compare correctly (https://play.golang.org/p/gzaizaea0JV)

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,141 @@
...
+	asDBCleanupCommand = &cobra.Command{
+		Use:   "cleanup",
+		Short: "Clean stale Application Server application data",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if config.Redis.IsZero() {
+				panic("Only Redis is supported by this command")
+			}
+
+			logger.Info("Connecting to Identity Server database...")
+			db, err := store.Open(ctx, config.IS.DatabaseURI)
```
adriansmares: This will work in deployments where the Identity Server is local, but not in the ones where it is remote. I think we will have to add this method as an RPC that uses cluster authentication.

The component stores are fine (you may assume that you can just open them via Redis).

pkg/applicationserver/io/web/redis/registry.go
```diff
@@ -53,6 +53,10 @@ type WebhookRegistry struct {
 	Redis *ttnredis.Client
 }
 
+func (r *WebhookRegistry) allKey(ctx context.Context) string {
```
adriansmares: This key/index does not exist right now, so you will either need a migration or to actually scan the database.

Checkout how `ns-db` does this here:
https://github.com/TheThingsNetwork/lorawan-stack/blob/865c1a18686d1c1d9d93240e5bf539db2b6fd7df/cmd/ttn-lw-stack/commands/ns_db.go#L49
https://github.com/TheThingsNetwork/lorawan-stack/blob/865c1a18686d1c1d9d93240e5bf539db2b6fd7df/cmd/ttn-lw-stack/commands/redis.go#L73

cmd/ttn-lw-stack/commands/as_db.go
pgalic96: I would say then this whole `as_db` command should be part of an AS RPC that is invoked from somewhere else. If we need to use cluster auth, we will have to do this procedure where we have access to the component and the running cluster.

But then do we invoke this as part of `ttn-lw-cli`? Or do we enhance `as_db` with an ability to attach to the current running component / cluster ?

pgalic96:
@adriansmares I moved the `ns_db` range implementation to redis, and used it to do the scanning, seems to work. Let me know if you like this implementation. Will re-request review as a notification, no rush since it's not a live PR yet.

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,138 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/applicationserver/io/web"
+	asiowebredis "go.thethings.network/lorawan-stack/v3/pkg/applicationserver/io/web/redis"
+	asredis "go.thethings.network/lorawan-stack/v3/pkg/applicationserver/redis"
+	"go.thethings.network/lorawan-stack/v3/pkg/identityserver/store"
+	"go.thethings.network/lorawan-stack/v3/pkg/redis"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+func toApplicationSet(list []*ttnpb.ApplicationIdentifiers) (set map[ttnpb.ApplicationIdentifiers]struct{}) {
+	set = make(map[ttnpb.ApplicationIdentifiers]struct{})
```
adriansmares: This is still not ideal - using the structs as keys is not really allowed (if you check the actual struct you will see some `XXX_*` fields in there which are filled by marshallers and will cause equality checks to fail).

I'd still recommend to use strings and `unique.ID` + contexts.

adriansmares:
> @adriansmares I moved the `ns_db` range implementation to redis, and used it to do the scanning, seems to work. Let me know if you like this implementation. Will re-request review as a notification, no rush since it's not a live PR yet.

The range implementation is much better - it's fine to continue with this.

cmd/ttn-lw-stack/commands/as_db.go
adriansmares: Pinging @htdvisser @johanstokking for thoughts on this. Making it an AS RPC would cause the performance of that particular AS instance to plummet most likely, which is why I preferred to have this range process separate from the real components, but at the same time calling the IS from a `-db` command is also not ideal either.

api/application_services.proto
```diff
@@ -119,6 +119,8 @@ service ApplicationRegistry {
       post: "/applications/{application_id}/dev-eui"
     };
   };
+
+  rpc ListAll(google.protobuf.Empty) returns (ListAllApplicationsResponse);
```
htdvisser: 1. Without pagination this really isn't going to scale in large networks, so I think we should use that.
2. Can't we use the existing `List` RPC? That already says that _If no user or organization is given, this returns the applications the caller has access to_, so we could make that return everythin if it's called with cluster auth.

api/end_device_services.proto
```diff
@@ -96,6 +96,8 @@ service EndDeviceRegistry {
       delete: "/applications/{application_ids.application_id}/devices/{device_id}"
     };
   };
+
+  rpc ListAll(google.protobuf.Empty) returns(ListAllEndDevicesResponse);
```
htdvisser: There's no way this is going to scale. We can't return hundreds of thousands of end devices in a single response.

htdvisser (commented):
Maybe `as-db` isn't the right place for this, and we should instead put it under some "batch" or "background job" command. But I don't have anything against making RPC calls from a background job.

This is still a bit race-y. On large databases, this command will take quite a long time, so if someone creates a new application between our IS call and our database range, we may accidentally delete it. Maybe it would be better to _first_ range over our own database, and then ask the IS if the application is in the deleted-but-too-late-to-be-restored window.

pgalic96:
I removed some required fields from application and end device protos, I guess that would mean that this PR can land in 3.16 only.

htdvisser:
I'm not sure. It doesn't look like a breaking change actually.

In any case, this isn't going to land in v3.15.1, so changing the milestone.

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,176 @@
...
+			return nil, nil, err
+		}
+		clusterOpts = append(clusterOpts, cluster.WithTLSConfig(tls))
+	}
+	c, err := cluster.New(ctx, &config.Cluster, clusterOpts...)
+	if err != nil {
+		return nil, nil, err
+	}
+	c.Join()
+	time.Sleep(20 * time.Millisecond)
```
adriansmares: I think this is slightly optimistic, and instead we should look a couple of times the `GetPeerConn` (+delay).

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,176 @@
...
+		if err := tlsConf.Client.ApplyTo(tls); err != nil {
+			return nil, nil, err
+		}
+		clusterOpts = append(clusterOpts, cluster.WithTLSConfig(tls))
+	}
+	c, err := cluster.New(ctx, &config.Cluster, clusterOpts...)
+	if err != nil {
+		return nil, nil, err
+	}
+	c.Join()
```
adriansmares: Don't forget to check the error, and maybe defer a `Leave`.

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	asDBCleanupCommand = &cobra.Command{
+		Use:   "cleanup",
+		Short: "Clean stale Application Server application data",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if config.Redis.IsZero() {
+				panic("Only Redis is supported by this command")
+			}
+			// Initialize AS registry cleaners (together with their local app/dev sets).
+			logger.Info("Initiating pubsub client")
```
adriansmares: Try to be consistent with PubSub - I think we used `PubSub` everywhere.

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,176 @@
...
+				}
+				pageCounter++
+			}
+			// Cleanup data from AS registries.
+			logger.Info("Cleaning pub sub registry")
+			err = pubsubCleaner.CleanData(ctx, deviceIdentityServerSet)
+			if err != nil {
+				return err
+			}
+			logger.Info("Cleaning app packages registry")
```
adriansmares: `app packages` -> `application packages`

pkg/applicationserver/cleanup.go
```diff
@@ -0,0 +1,67 @@
...
+	)
+	return err
+}
+
+func (cleaner *RegistryCleaner) DeleteComplement(ctx context.Context, devSet map[string]struct{}) error {
+	for ids := range devSet {
+		devIds, err := unique.ToDeviceID(ids)
+		if err != nil {
+			return err
+		}
```
adriansmares, line 49: Use `unique.WithContext` afterwards.
```
	ctx := ctx
	ctx, err = unique.WithContext(ctx, uid)
	if err != nil {
		return nil, err
	}
```
In order to fill the context with the unique data. This should be done after each `unique.ToDeviceID`.

adriansmares (commented):
Looks pretty good, but I would love to somehow see a testing procedure for this.

pgalic96:
Regarding testing; Mock response from IS, populating registries with extra application data, running the command and checking what's left in registries would be a good test, the problem is that some/most of these registries do not test the underlying registry interface methods but the tests are tied to the RPC's which this command is not using. Will setup some new test procedures.

pkg/unique/unique.go
```diff
@@ -41,6 +41,10 @@ func ID(ctx context.Context, id ttnpb.IDStringer) (res string) {
 	return res
 }
 
+func GenericID(ctx context.Context) (res string) {
```
htdvisser: This is missing godoc.

pkg/identityserver/application_registry.go
```diff
@@ -213,9 +228,12 @@ func (is *IdentityServer) listApplications(ctx context.Context, req *ttnpb.ListA
 		return nil, err
 	}
 
-	for i, app := range apps.Applications {
-		if rights.RequireApplication(ctx, *app.GetIds(), ttnpb.RIGHT_APPLICATION_INFO) != nil {
-			apps.Applications[i] = app.PublicSafe()
+	// If request comes from cluster, skip trimming of application info.
+	if !clusterAuth {
```
htdvisser, line 244: Should we at least restrict the field mask for cluster auth?

Like we do in `updateEndDevice` (`end_device_registry.go`)

pkg/identityserver/application_registry.go
```diff
@@ -156,29 +157,36 @@ func (is *IdentityServer) getApplication(ctx context.Context, req *ttnpb.GetAppl
...
 	var includeIndirect bool
-	if req.Collaborator == nil {
-		authInfo, err := is.authInfo(ctx)
-		if err != nil {
-			return nil, err
+	var clusterAuth bool
+	// If request comes from cluster (list all applications), skip caller rights check.
+	if clusterauth.Authorized(ctx) == nil && req.Collaborator == nil {
+		clusterAuth = true
+		ctx = store.WithSoftDeleted(ctx, false)
```
htdvisser: Shouldn't the caller decide this?

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,207 @@
...
+	pbtypes "github.com/gogo/protobuf/types"
+	"github.com/spf13/cobra"
+	"go.thethings.network/lorawan-stack/v3/pkg/applicationserver/io/web"
+	"go.thethings.network/lorawan-stack/v3/pkg/cluster"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	"go.thethings.network/lorawan-stack/v3/pkg/unique"
+	"google.golang.org/grpc"
+)
+
+func NewClusterComponentConnection(ctx context.Context, config Config, role ttnpb.ClusterRole) (*grpc.ClientConn, cluster.Cluster, error) {
```
htdvisser: - Missing godoc (also the funcs below)
- This looks like it can / will be used by other commands than as-db, so should it be in another file?

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,207 @@
...
+	return deviceIdentityServerSet, nil
+}
+
+var (
+	// Define limit for pagination (maximum defined in protos).
+	limit = uint32(1000)
+	// Max retries to join cluster
+	maxRetries = 5
+	// Delay between retries to join cluster
+	delay       = 500
```
htdvisser: Why isn't this a time.Duration?

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	return nil, nil, err
+}
+
+func FetchIdentityServerApplicationSet(ctx context.Context, client ttnpb.ApplicationRegistryClient, clusterAuth grpc.CallOption) (map[string]struct{}, error) {
+	applicationIdentityServerSet := make(map[string]struct{})
+	pageCounter := uint32(1)
+	// Iterate over application list paginated requests and add them to the IS app map.
+	for {
+		res, err := client.List(ctx, &ttnpb.ListApplicationsRequest{
```
htdvisser: Should we add a little delay so that we won't hit the rate limit?

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	return applicationIdentityServerSet, nil
+}
+
+func FetchIdentityServerEndDeviceSet(ctx context.Context, client ttnpb.EndDeviceRegistryClient, clusterAuth grpc.CallOption) (map[string]struct{}, error) {
+	deviceIdentityServerSet := make(map[string]struct{})
+	pageCounter := uint32(1)
+	// Iterate over application list paginated requests and add them to the IS app map.
+	for {
+		res, err := client.List(ctx, &ttnpb.ListEndDevicesRequest{
```
htdvisser: Should we add a little delay so that we won't hit the rate limit?

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,207 @@
...
+			}
+			// Create cluster and grpc connection with identity server.
+			conn, cl, err := NewClusterComponentConnection(ctx, *config, ttnpb.ClusterRole_ENTITY_REGISTRY)
+			if err != nil {
+				return err
+			}
+			defer func() {
+				logger.Debug("Leaving cluster...")
+				if err := cl.Leave(); err != nil {
+					logger.WithError(err).Error("Could not leave cluster")
```
htdvisser, line 73: You may want to `return` after this.

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,207 @@
...
+				}
+			}
+			return nil
+		},
+	}
+)
+
+func init() {
+	Root.AddCommand(asDBCommand)
+	asDBCommand.AddCommand(asDBCleanupCommand)
```
htdvisser, line 157: Should we perhaps add a `--dry-run` flag as well, so that we can see what the command plans to do?

pkg/applicationserver/cleanup.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"context"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/cleanup"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	"go.thethings.network/lorawan-stack/v3/pkg/unique"
+)
+
+type RegistryCleaner struct {
```
htdvisser, line 26: Missing godoc. Also on the funcs below.

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,207 @@
...
+			client := ttnpb.NewApplicationRegistryClient(conn)
+			applicationIdentityServerSet, err := FetchIdentityServerApplicationSet(ctx, client, cl.Auth())
+			if err != nil {
+				return err
+			}
+			devClient := ttnpb.NewEndDeviceRegistryClient(conn)
+			deviceIdentityServerSet, err := FetchIdentityServerEndDeviceSet(ctx, devClient, cl.Auth())
+			if err != nil {
+				return err
+			}
```
htdvisser: I've said before that I think it's dangerous to do it this way, because we may delete things that were created while command runs.

pgalic96: This is done after the local sets are computed (`RangeToLocalSet`). If there are extra applications created they will be part of the IS set but not of the local sets (unless there is a following flow: app `app-x` data created in AS -> local set in AS computed -> IS set computed -> application `app-x` created in IS . I think this state is probably unreachable, but please confirm this.

htdvisser: Ah sorry, I now see that that's hidden in those `NewXXXCleaner` funcs.

pkg/identityserver/application_registry.go
pgalic96: Should we reuse the `deleted` field in case of cluster auth, i.e. if its set we add `WithSoftDeleted(ctx, false)`? The only concern with this is that it is slightly different than what this field is indicating right now (`store.WithSoftDeleted(ctx, true)`). Otherwise we would need to introduce a new request field which breaks the compatibility commitment.

pgalic96: It wouldn't necessarily break the compatibility commitment if its just a new field, but what I wanted to say is that we have `deleted` field which indicates whether we want to **only** include deleted entities and now it seems like a very hacky approach to add another field like `merge_deleted_and_non_deleted` which wouldn't break compatibility commitment but wouldn't make sense if we can just refactor this to `deleted` and `deleted_only` fields.

pkg/applicationserver/io/packages/redis/registry.go
```diff
@@ -82,6 +83,10 @@ func (r *ApplicationPackagesRegistry) Init(ctx context.Context) error {
 	return nil
 }
 
+func (r *ApplicationPackagesRegistry) allKey(ctx context.Context) string {
```
adriansmares: Remove this

adriansmares (approved):
Reviewed `pkg/applicationserver` and the `cmd` changes. They look good.

I'd recommend a `rebase` + `autosquash` and then a review from @johanstokking.

pgalic96:
will wait for approval from @htdvisser before squashing

htdvisser:
No, please go ahead with rebase and history cleanup. I'll review after.

pgalic96:
done

htdvisser (approved):
Code-owned files LGTM

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -0,0 +1,148 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+var (
+	// Define limit for pagination (maximum defined in protos).
+	limit = uint32(1000)
+	// Max retries to join cluster
+	maxRetries = 5
+	// Delay between retries to join cluster
+	delay       = time.Duration(500) * time.Millisecond
```
johanstokking: These are declared in `cmd/ttn-lw-stack/commands` but they seem very specific to a particular (set of) command(s).

If this is not configurable via flags or a config file, why are they declared at this level and why aren't they consts?

cmd/ttn-lw-stack/commands/utils.go
```diff
@@ -0,0 +1,129 @@
...
+	c, err := cluster.New(ctx, &config.Cluster, clusterOpts...)
+	if err != nil {
+		return nil, nil, err
+	}
+	if err := c.Join(); err != nil {
+		return nil, nil, err
+	}
+	var cc *grpc.ClientConn
+	for i := 0; i < maxRetries; i++ {
+		time.Sleep(delay)
```
johanstokking, line 55: So this is not obvious that these symbols are just for this use case. There can be so many things to retry or delay.

Take as function parameters maybe?

cmd/ttn-lw-stack/commands/utils.go
```diff
@@ -0,0 +1,129 @@
...
+			return nil, err
+		}
+		for _, app := range res.Applications {
+			applicationIdentityServerSet[unique.ID(ctx, app.GetIds())] = struct{}{}
+		}
+		if len(res.Applications) < int(limit) {
+			break
+		}
+		pageCounter++
+		time.Sleep(paginationDelay)
```
johanstokking, line 85: Is this against rate limiting or flattening the load curve? I guess that rate limiting is off for cluster auth, so I presume the latter?

cmd/ttn-lw-stack/commands/utils.go
```diff
@@ -0,0 +1,129 @@
...
+		cc, err = c.GetPeerConn(ctx, role, nil)
+		if err == nil {
+			return cc, c, nil
+		}
+	}
+	return nil, nil, err
+}
+
+// FetchIdentityServerApplicationSet returns the set of all non-expired applications in the Identity Server.
+func FetchIdentityServerApplicationSet(ctx context.Context, client ttnpb.ApplicationRegistryClient, clusterAuth grpc.CallOption) (map[string]struct{}, error) {
```
johanstokking: Why are we returning a set if the caller is likely only interested in a slice of unique identifiers?

pkg/applicationserver/cleanup.go
```diff
@@ -0,0 +1,75 @@
...
+	err := cleaner.DevRegistry.Range(ctx, []string{"ids"}, func(ctx context.Context, ids ttnpb.EndDeviceIdentifiers, dev *ttnpb.EndDevice) bool {
+		cleaner.LocalSet[unique.ID(ctx, ids)] = struct{}{}
+		return true
+	},
+	)
+	return err
+}
+
+// DeleteComplement deletes registry application data of all devices in the device set.
+func (cleaner *RegistryCleaner) DeleteComplement(ctx context.Context, devSet map[string]struct{}) error {
```
johanstokking: Why take a map here?

pkg/identityserver/application_registry.go
johanstokking: Is this addressed?

johanstokking (commented):
Only checked CLI and AS. Comments are mostly code quality related, functionality looks otherwise good!

cmd/ttn-lw-stack/commands/utils.go
pgalic96: Yes indeed, I will change this to a flag parameter so we can see what works best.

pkg/identityserver/application_registry.go
pgalic96: Yes, see https://github.com/TheThingsNetwork/lorawan-stack/pull/4653/files#diff-0a302077283e72a9d2eea5fc0ceaf857f4e7c6afe47a86eac138b5d5318722faR164

cmd/ttn-lw-stack/commands/as_db.go
pgalic96: I hardcoded `limit` as its obvious we want to fetch large batches in this method. I moved `maxRetries` and `delay` as function parameters (and scoped them to the `as_db` function) I don't think these two params need to be configurable, it is used only for obtaining peer connection.

cmd/ttn-lw-stack/commands/utils.go
```diff
@@ -0,0 +1,121 @@
...
+			ApplicationIds: nil,
+			FieldMask:      &pbtypes.FieldMask{Paths: []string{"ids"}},
+			Limit:          uint32(1000),
+			Page:           pageCounter,
+		}, clusterAuth)
+		if err != nil {
+			return nil, err
+		}
+		deviceList = append(deviceList, res.EndDevices...)
+		if len(res.EndDevices) < 1000 {
```
johanstokking: I don't think this is correct. The limit is really a limit; "not more than", not necessarily "exactly". Do we have this logic in other places as well?

The preferred approach would be breaking the loop when the received item count is zero or when the total received item count equals or exceeds the total indicated in the response. That value is in a response header though, not in `res`.

Also, limits are defined as consts but there's a literal used in this function twice.

cmd/ttn-lw-stack/commands/utils.go
```diff
@@ -0,0 +1,121 @@
...
+	return deviceList, nil
+}
+
+func setToArray(set map[string]struct{}) []string {
+	keys := make([]string, len(set))
+	i := 0
+	for k := range set {
+		keys[i] = k
+		i++
+	}
```
johanstokking, line 121:
```suggestion
	keys := make([]string, 0, len(set))
	for k := range set {
		keys = append(keys, k)
	}
```

pkg/applicationserver/cleanup.go
```diff
@@ -0,0 +1,81 @@
...
+
+// CleanData cleans registry device data.
+func (cleaner *RegistryCleaner) CleanData(ctx context.Context, isSet map[string]struct{}) error {
+	complement := cleanup.ComputeSetComplement(isSet, cleaner.LocalSet)
+	devIds := make([]string, len(complement))
+	i := 0
+	for id := range complement {
+		devIds[i] = id
+		i++
+	}
```
johanstokking, line 75:
```suggestion
	devIds := make([]string, 0, len(complement))
	for id := range complement {
		devIds = append(devIds, id)
	}
```

johanstokking (approved):

johanstokking:
Approved with comments. The small code changes aren't strictly necessary but if you address them, please address in all places.

Not sure if this pagination thing needs to be addressed; @htdvisser any opinion on this? We don't want to break the loop too early.

cmd/ttn-lw-stack/commands/utils.go
pgalic96: I was actually looking into this and this is what I got from the stack overflow thread:

```
In most situations it probably won't make much of a difference, but it's not much more work,
 and in my tests (using a map with 1,000,000 random int64 keys and then generating the array
 of keys ten times with each method), 
it was about 20% faster to assign members of the array directly than to use append.

Although setting the capacity eliminates reallocations, 
append still has to do extra work to check if you've reached capacity on each append.
```

pgalic96: Might be relevant if we do operations with 100,000 + keys

johanstokking: Ah, ok, thanks for sharing

cmd/ttn-lw-stack/commands/utils.go
```diff
@@ -0,0 +1,123 @@
...
+	"crypto/tls"
+	"time"
+
+	pbtypes "github.com/gogo/protobuf/types"
+	"go.thethings.network/lorawan-stack/v3/pkg/cluster"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	"google.golang.org/grpc"
+)
+
+const defaultPaginationDelay = 1000
```
johanstokking: Limit, right?

johanstokking (approved):

cmd/ttn-lw-stack/commands/utils.go
pgalic96: ah yeah

Issue #4659, opened 2021-09-23
Migrate individual worker pools to pkg/workerpool


adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/4606
References https://github.com/TheThingsNetwork/lorawan-stack/pull/4658

We should migrate the remaining worker pools to `pkg/workerpool`.

#### Why do we need this?

This migration allows us to observe better the status of workers, as `pkg/workerpool` comes with batteries included when it comes to metrics (number of busy/idle workers, evolution of number of workers, queue size, processing latency, and work drops).

#### What is already there? What do you see now?

Individual implementations of the pools.

#### What is missing? What do you want to see?

Use `pkg/workerpool` instead of duplicated implementations.

- [x] UDP Upstream Packet Handlers
- [x] PBA Worker Pools

#### Environment

`v3.15.1`

#### How do you propose to implement this?

Adapt the handlers to the format expected by `pkg/workerpool`.

#### How do you propose to test this?

Already existing unit tests should cover this.

#### Can you do this yourself and submit a Pull Request?

Yes

PR #4663, opened 2021-09-23
Use worker pool in upstream handlers

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/4659

#### Changes

- Allow worker pools to be waited, and the queuing/minimal worker count to be disabled
- Use `pkg/workerpool` for GS upstream handlers
- Fix GS tests to actually listen for `TxAck`s coming from the NS mock
  - This would deadlock otherwise on `wg.Wait`, since a handler would be blocked on the channel write (and in tests channel writes don't use select)

#### Testing

Unit testing

##### Regressions

Unit testing should cover this

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -23,11 +23,8 @@ import (
 	"math"
 	"net"
 	"net/http"
-	"os"
-	"runtime/debug"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	pbtypes "github.com/gogo/protobuf/types"
@@ -56,6 +53,7 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 	"go.thethings.network/lorawan-stack/v3/pkg/types"
 	"go.thethings.network/lorawan-stack/v3/pkg/unique"
+	"go.thethings.network/lorawan-stack/v3/pkg/workerpool"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 )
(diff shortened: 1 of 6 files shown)


pkg/gatewayserver/gatewayserver.go
```diff
@@ -752,11 +728,25 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
...
+			QueueSize:     -1,
+		})
+		if err != nil {
+			conn.Disconnect(err)
+			return
+		}
+		defer wp.Wait()
+		host.pool = wp
 		hosts = append(hosts, host)
-		defer host.handleWg.Wait()
```
adriansmares, line 759: This should've deadlocked months ago when we introduced TxAcks, since the tests didn't actually read the NS channel, only the events -> this caused a handler to be blocked indefinitely on `host.handler.HandleTxAck`, because the channel was full because the tests didn't check for it.

~I don't see why this `Wait()` didn't deadlock before, this is weird.~

This didn't deadlock before because `HandleTxAck` does an RPC call, which allows the caller (the handler) to stop before the server (the NS mock) stops. The difference is that `pkg/workerpool` decouples the lifetime of the context of the publisher (which is equal to the lifetime of a gateway connection) from lifetime of the processing (doing the RPC). 

This change (the lifetime decoupling) is actually quite good I think, since gateways sometimes disconnect while the NS matches the uplinks, thus stopping the matching process. This should alleviate [this](https://sentry.io/organizations/the-things-industries/issues/2607823479/?environment=ttich-prod-au1&project=2682566&query=is%3Aunresolved) particular Sentry issue.

johanstokking (approved):

PR #4666, opened 2021-09-23
Use worker pools in PBA

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/4659

#### Changes

- Remove the manual implementations of worker pools from PBA
- Start multiple subscription tasks in parallel

#### Testing

Unit testing.

##### Regressions

Shouldn't occur, as this is mainly changing pipes.

#### Notes for Reviewers

I will run some tests tomorrow by connecting to PB, but the code should be ready for review.

#### Checklist

### DIFF
--- a/pkg/packetbrokeragent/agent.go
+++ b/pkg/packetbrokeragent/agent.go
@@ -446,7 +446,6 @@ func (a *Agent) forwarderPublisher(conn *grpc.ClientConn) workerpool.Handler {
 
 func (a *Agent) subscribeDownlink(ctx context.Context) error {
 	ctx = log.NewContextWithFields(ctx, log.Fields(
-		"namespace", "packetbrokeragent",
 		"forwarder_net_id", a.netID,
 		"forwarder_tenant_id", a.subscriptionTenantID,
 		"forwarder_cluster_id", a.clusterID,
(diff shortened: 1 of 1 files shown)


johanstokking:
> * There is only one gRPC connection to the Data Plane, that is shared between both publishers and subscribers. This undoes this commit: [ec2d3b9](https://github.com/TheThingsNetwork/lorawan-stack/commit/ec2d3b9772d5680e8eac99d21564180bf7c81544) . Let me know if we want to bring back this behavior (empty data plane address means that we don't attempt to dial at all)

The old behavior was there for two reasons which are still valid today:
1. PBA isn't always configured with a Data Plane connection. PBA can also be used only for administration and policy config etc. So having an empty Data Plane connection shouldn't start subscribe/publish tasks and dialing shouldn't happen
2. Having a connection per direction or role allows for some load balancing; everything would be tunneled over one connection. Having more connections between PBA and PB allows PB to better balance load. We don't really need a connection pool today, but we might in the near future.

adriansmares:
I have updated the PR to use the original task + channel approach, and also added a constant for the number of subscription tasks. Let me know if this variant is better.

This PR currently targets `feature/workerpool-tests-benchmark` of https://github.com/TheThingsNetwork/lorawan-stack/pull/4669 because that PR simplifies the handler syntax.

pkg/packetbrokeragent/agent.go
```diff
@@ -786,18 +710,14 @@ func (a *Agent) getSubscriptionFilters() []*packetbroker.RoutingFilter {
...
 		"home_network_cluster_id", a.homeNetworkClusterID,
 		"group", a.clusterID,
 	))
 	logger := log.FromContext(ctx)
 
-	conn, err := a.dialContext(ctx, a.dataPlaneAddress,
-		grpc.WithBlock(),
-		grpc.FailOnNonTempDialError(true),
-	)
+	conn, err := a.dialContext(ctx, a.dataPlaneAddress)
```
johanstokking: This should likely be reverted as `Subscribe()` relies on the availability of the connection

pkg/packetbrokeragent/agent.go
```diff
@@ -1089,15 +955,11 @@ func (a *Agent) handleUplinkMessage(ctx context.Context, up *packetbroker.Routed
...
-		"namespace", "packetbrokeragent",
 		"home_network_net_id", a.netID,
 		"home_network_cluster_id", a.homeNetworkClusterID,
 	))
 
-	conn, err := a.dialContext(ctx, a.dataPlaneAddress,
-		grpc.WithBlock(),
-		grpc.FailOnNonTempDialError(true),
-	)
+	conn, err := a.dialContext(ctx, a.dataPlaneAddress)
```
johanstokking: Idem

pkg/packetbrokeragent/agent.go
```diff
@@ -310,20 +311,28 @@ func New(c *component.Component, conf *Config, opts ...Option) (*Agent, error) {
...
 	}
 	if a.forwarderConfig.Enable && a.dataPlaneAddress != "" {
-		c.RegisterTask(newTaskConfig("pb_publish_uplink", a.publishUplink))
-		c.RegisterTask(newTaskConfig("pb_subscribe_downlink", a.subscribeDownlink))
+		for i := 0; i < publishTaskCount; i++ {
+			c.RegisterTask(newTaskConfig("pb_publish_uplink", a.publishUplink))
+		}
+		for i := 0; i < subscribeTaskCount; i++ {
+			c.RegisterTask(newTaskConfig("pb_subscribe_downlink", a.subscribeDownlink))
+		}
```
johanstokking: So now we will have multiple connections; one per subscription and per publisher, right? With standard config, that's already 10 connections.

We don't really need this now, also because the gRPC client internals may use multiple TCP connections. What we do need is multiple streams for subscriptions, but the connection may be the same.

pkg/packetbrokeragent/agent.go
adriansmares: I don't think they are required, the tests use a remote gRPC server and they do work.

adriansmares:
I've reworked the last commit to start multiple subscription streams in parallel, while sharing the same client/gRPC channel.

The current behavior on error is to cancel the whole parent task - this causes a whole new gRPC connection to be created if one of the subscription fails. The advantage here is that we force a reconnection.

We could also just ask individual streams to reconnect on failure, but maybe it's better to start with a clean connection - let me know what you think.

johanstokking:
> The current behavior on error is to cancel the whole parent task - this causes a whole new gRPC connection to be created if one of the subscription fails. The advantage here is that we force a reconnection.

This is already being handled by the gRPC client.

adriansmares:
> > The current behavior on error is to cancel the whole parent task - this causes a whole new gRPC connection to be created if one of the subscription fails. The advantage here is that we force a reconnection.
> 
> This is already being handled by the gRPC client.

Removed then. Individual streams will now independently reconnect.

pkg/packetbrokeragent/agent.go
johanstokking: OK, so just so you know how this works; dialing is now non-blocking. The connection is established in the background. Temporary dialing errors, including TLS handshake issues but there are many more, are not caught. Any usage of the client, fails immediately. If you can cope with those failures because those tasks restart, fine, but configuration errors of PBA are not immediately visible in the logs because the gRPC client is silently reconnecting (with DEBUG level in TTS).

johanstokking (approved):

pkg/packetbrokeragent/agent.go
adriansmares: Thanks, I understand the reason now :pray:. Reverted in all call sites.

PR #4672, opened 2021-09-27
Report resource exhausted errors using warnings

adriansmares:
#### Summary

References https://sentry.io/organizations/the-things-industries/issues/2628907782/?project=2682566&sort=freq&statsPeriod=14d
References https://sentry.io/organizations/the-things-industries/issues/2607823479/events/7649caace2234c15a997e544d04c3f23/tags/?project=2682566&statsPeriod=14d

#### Changes

- Log resource exhausted errors as `WARN` instead of `ERROR` in NS RPC calls
  - This caused application layer failures (pushing downlinks above the limit) to be logged as `ERROR`, eating through our Sentry subscription
  - I've checked the individual devices from the Sentry events - there is an application that pushes downlinks after every uplink (instead of maybe 'replace'), so probably someone tried to schedule an OTAA update and didn't realize that the queues may not be empty. We also don't really communicate this limit (and errors are not visible via MQTT).
- Log cancelled errors as `WARN` instead of `ERROR` in NS RPC calls
  - Gateways may disconnect (thus cancelling the connection context) while RPCs as running. The NS shouldn't consider those to be errors.
-  Do not log connection registry errors if they are cancelled errors
   - Again, if the gateways disconnect really quick (before the connection stats goroutine starts), this error would be logged

#### Testing

N/A

##### Regressions

N/A

#### Checklist

### DIFF
--- a/config/messages.json
+++ b/config/messages.json
@@ -6821,9 +6821,9 @@
       "file": "config.go"
     }
   },
-  "error:pkg/networkserver:downlink_queue_capacity_exceeded": {
+  "error:pkg/networkserver:downlink_queue_capacity": {
     "translations": {
-      "en": "Downlink queue capacity exceeded"
+      "en": "downlink queue capacity exceeded"
     },
     "description": {
       "package": "pkg/networkserver",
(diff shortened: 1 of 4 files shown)


KrishnaIyer (approved):
LGTM

johanstokking (approved):

PR #4662, opened 2021-09-23
Remove embed and nullable options from gateway.proto

htdvisser:
#### Summary

This pull request addresses part of #2798 by removing some of the embed and nullable options from `gateway.proto`.

#### Changes

- Remove some `embed` options
- Remove some `nullable` options one by one

#### Testing

Hopefully everything is covered by existing unit tests

##### Regressions

Same as in https://github.com/TheThingsNetwork/lorawan-stack/pull/4639

Removing the embed option could result in some messages no longer implementing certain interfaces. As far as I could tell, this was mostly the `ttnpb.IDStringer` interface, so I've done my best to make sure that all messages still implement that.

Removing the nullable option could result in nil pointer panics. I've done my best to make sure that we use the generated getter methods where possible, and that we no longer set fields of GatewayIdentifiers directly.

#### Notes for Reviewers

There's unfortunately a lot of manual work here, so please review all non-generated files.

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #4674, opened 2021-09-27
Remove gogoproto options from gateway.proto

KrishnaIyer:
#### Summary

Remove gogoproto options from `gateway.proto` and `gateway_services.proto`. Please check `Notes for Reviewers` for details.

This is based off https://github.com/TheThingsNetwork/lorawan-stack/pull/4662 and is hence targeting the `issue/2798-gogo-gtw-identifiers` branch. Once that PR is merged, I'll rebase this on `v3.15`.

#### Changes

- Remove `gogoproto.nullable` and `gogoproto.embed` options from `gateway.proto` and `gateway_services.proto`
- Adapt interfaces and logic
- Add tests.

#### Testing

UT

##### Regressions

I think I've caught them mostly in UT.

#### Notes for Reviewers

- `PullGatewayConfigurator` doesn't seem to be used anywhere. I've added the interfaces for completeness nonetheless.
- The `gateway.version_ids` message was set as required `(validate.rules).message.required = true` but it's not. The `required` was effective only after removing the nullable option so I removed that as well.
- The `(gogoproto.nullable) = true` on `schedule_anytime_delay` had no effect.
- There's more than one way of handling the `GatewayAntennas` option. I've chosen the easiest one which requires the smallest amount of changes.
- This is a lot of manual work so please check your code owned files.

#### Checklist


pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -753,7 +758,7 @@ func TestGatewayServer(t *testing.T) {
 									UpdateLocation: false,
 									Up: &ttnpb.GatewayUp{
 										GatewayStatus: &ttnpb.GatewayStatus{
-											Time: time.Unix(424242, 0),
+											Time: timePtrUTC(time.Unix(424242, 0).UTC()),
```
adriansmares: Do you need to `UTC` twice ? The pointer function does it too.

pkg/gatewayserver/io/mqtt/format_protobufv2.go
```diff
@@ -66,6 +66,11 @@ type protobufv2 struct {
 	topics.Layout
 }
 
+// TimePtr converts time to a pointer.
+func TimePtr(x time.Time) *time.Time {
+	return &x
```
adriansmares, line 71: Does this need `UTC` ? Maybe that's how we end up with these weirdly deformed timestamps of gateway times (not saying this call site necessarily, but _maybe_...)

adriansmares (approved):

pkg/gatewayserver/io/mqtt/format_protobufv2.go
KrishnaIyer: Good question. The reason I added UTC in the other test was that when it was dereferenced by the assertions package, it takes local time. Let me check what the GS does in general with time and get back to you.

htdvisser (approved):
LGTM

pkg/gatewayserver/io/mqtt/format_protobufv2.go
htdvisser: I'd say this is out of scope for this PR, but indeed good to check.

KrishnaIyer:
@kschiffer: Could you please take a look at [this](https://github.com/TheThingsNetwork/lorawan-stack/pull/4674/checks?check_run_id=3731017663#step:9:238) test case?

The `GatewayAntennas` field of the `Gateway` message is now an array of pointers so that's where I think this assertion is failing. Cypress takes forever on my laptop so I'd appreciate it if you could take a look.

htdvisser:
This one missed the deadline for v3.15.1, so moving it to v3.15.2.

Issue #4478, opened 2021-08-02
Remove data rate indices from `TxSettings` and `TxRequest`


johanstokking:
#### Summary

Remove data rate indices from `TxSettings` and `TxRequest`.

#### Why do we need this?

Because intermediate components, like Gateway Server, gateways, other Network Servers in peering scenarios etc, do not need the data rate index.

Instead, the data rate index is part of the contract between the end device and the serving Network Server only.

Deleting these fields will lead to quite some code cleanups and forces us not to rely on the data rate index.

#### What is already there? What do you see now?

We already use the data rate indices in `TxRequest` and `TxSettings` for informative purposes. As far as I know, we don't depend on it anymore since Packet Broker also supports fully described data rates in combination with 3.14.x (since #4479).

#### What is missing? What do you want to see?

Add `rx1_data_rate` and `rx2_data_rate` to `TxRequest`. When we have that, we need another minor bump.

Remove `data_rate_index`, `rx1_data_rate_index` and `rx2_data_rate_index`.

#### How do you propose to implement this?

1. Add fully described data rates to `TxRequest` and make GS, NS and PBA use them with fallbacks (can be done from 3.15.0)
2. Remove usage of indexes from GS, NS and PBA (needs minor bump from step (1))

#### How do you propose to test this?

CI and main end-to-end flows, including testing through Packet Broker

#### Can you do this yourself and submit a Pull Request?

Yes

johanstokking:
The first part for 3.15.x is addressed in #4679.

When that is merged, we'll keep this issue open and we move it to a 3.16.x milestone where we can break the internal API compatibility.

johanstokking:
#4679 is now merged, moving this to 3.16.0. It can be implemented in any version >= 3.16.0. No hurries, just good to resolve some technical debt.

PR #4679, opened 2021-09-29
Use fully described data rates in `TxRequest`

johanstokking:
#### Summary

References #4478 

#### Changes

This adds fully described data rates to `TxRequest`:

- NS now sets both the index and the fully described data rate
- GS now prefers the fully described data rate but falls back to looking up the data rate by index
- PBA sets the fully described data rate on outgoing downlink and prefers the fully described data rate over incoming downlink

#### Testing

CI

##### Regressions

Shouldn't be any; we keep compatibility within 3.15.x

#### Notes for Reviewers

The next step is the easiest; removing all the fallbacks. We need a minor bump for that. This is also tracked in #4478.

#### Checklist


pkg/networkserver/downlink_test.go
```diff
@@ -852,16 +852,26 @@ func TestProcessDownlinkTask(t *testing.T) {
...
+							DownlinkPaths: paths,
+							Priority:      ttnpb.TxSchedulePriority_HIGHEST,
+							Rx1DataRate:   &lastUp.Settings.DataRate,
+							Rx1Delay:      dev.MacState.CurrentParameters.Rx1Delay,
+							Rx1Frequency:  dev.MacState.CurrentParameters.Channels[lastUp.DeviceChannelIndex].DownlinkFrequency,
+							Rx2DataRate: &ttnpb.DataRate{
+								Modulation: &ttnpb.DataRate_Lora{
+									Lora: &ttnpb.LoRaDataRate{
+										Bandwidth:       125000,
+										SpreadingFactor: 12 - uint32(dev.MacState.CurrentParameters.Rx2DataRateIndex),
```
johanstokking: This isn't really beautiful but I think this is acceptable in testing. Otherwise we'd need a utility function for this. Note that in not all cases we have PHY available and we only do this for RX2 which is always LoRa (not FSK and not LR-FHSS)

adriansmares: It's acceptable but at the same time we will look back at this in the future and feel slightly disappointed. 

Consider https://github.com/TheThingsNetwork/lorawan-stack/commit/78844dbc531d6d9b4853b4bf64bfe8b94a08b093 - maybe you like this more.

adriansmares (approved):

PR #4684, opened 2021-10-01
Fix/4558 expand toggle sections when validation fails

### LINKED ISSUE #4558: Automatically expand form toggle sections when a validation within fails
#### Summary
Our forms should automatically expand sections that contain fields that had validation errors when submitting a form.

#### Steps to Reproduce
1. Within an application, register a new device and use the manual registration mode
2. Fill out all required fields
3. Expand the `Show advanced activation, LoRaWAN class and cluster settings` section
4. Uncheck `Use network's default MAC settings`
5. For `Rx2 data rate` enter the invalid value `16`
6. Collapse the advanced section again
7. Submit the form
8. Observe the missing feedback since the invalid field is hidden:
image

#### What do you see now?
Missing feedback when submitting forms that have invalid fields which are currently hidden

All hidden `<Form.CollapseSection />`'s being expanded so that the field error is visible

Probably trivial to check within `<Form.CollapseSection />` for the presence of any validation errors using the form context and triggering an expand.

Manual testing should be enough, but possible to create a new test case for this.

@bafonins

brycechampaign:
#### Summary

Closes #4558 

#### Changes

- Extend `FormCollapseSection` component to expand when a section contains fields that had validation errors when submitting a form.
- Fix errors within collapse sections on the manual device-add form blocking Formik submission and validation handlers.
- Pass `errorTitles` props to FormCollapseSection instances to specify what errors should trigger an expand.

#### Testing

Manual testing was performed, as recommended by the original issue.

##### Regressions

On the manual device form, which directly accesses the Formik submission handler and bypasses native HTML validation that was previously performed when the form is submitted, modifications may need to be done to the validation schema to make sure all edge cases are caught. For example, it is possible to enter the letter "e" into number fields, but native HTML validation would raise an error if the entered number was invalid when the form is submitted. But since native submit behavior is prevented, it may be necessary to make sure the validation schema provided to the Formik handlers catches these cases.

Testing was performed manually by comparing what values could be entered into certain inputs within collapse sections and submitted before and after native form submission behavior was prevented.
#### Notes for Reviewers

I was hoping to keep changes limited to the reusable components such as `FormCollapseSection`, but I had trouble implementing a solution that would automatically work for all forms using the `FormCollapseSection` component. I implemented the `errorTitles` prop to be passed to `FormCollapseSection` instances in order to specify what errors caught by Formik would cause an expand, but it would be cleaner to find a way to automatically determine if a child field produced an error. 

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ For details about compatibility between different releases, see the **Commitment
 
 - Emails to admins about requested OAuth clients.
 - `session` handling for joined OTAA end devices in the Console.
+- `FormCollapseSection` components that contain fields that had validation errors when submitting a form should automatically expand.
 
 ### Security
 
--- a/pkg/webui/components/form/section/index.js
+++ b/pkg/webui/components/form/section/index.js
@@ -33,15 +33,24 @@ const m = defineMessages({
 // `<FormCollapseSection />` aggregates a set of form fields under common title as well as adds
 // functionality to hide or show the fields.
 const FormCollapseSection = props => {
-  const { className, id, title, onCollapse, isCollapsed, initiallyCollapsed, children } = props
+  const {
+    className,
+    id,
+    title,
+    onCollapse,
+    isCollapsed,
+    initiallyCollapsed,
+    children,
+    errorTitles,
+  } = props
   const { formatMessage } = useIntl()
-  const { disabled } = useFormContext()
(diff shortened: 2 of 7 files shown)


kschiffer:
Thanks for this contribution @brycechampaign. To move forward with this quickly, I will close this now to open up a new PR from a local branch together with the adjustments we discussed.

Issue #4703, opened 2021-10-04
Support Packet Broker authentication on interop server


johanstokking:
#### Summary

Support Packet Broker authentication on interop server

Follow up on #4677 
Preparation for #4678  

#### Why do we need this?

For Packet Broker to authenticate with Identity Server to call `HomeNSReq` using LoRaWAN Backend Interfaces.

#### What is already there? What do you see now?

Only TLS client authentication.

#### What is missing? What do you want to see?

Token-based authentication.

#### How do you propose to implement this?

1. Prefer TLS client authentication. If the client presents a certificate, only authenticate with this method. There is no fallback.
2. See if there's an `Authorization` header set and check if header type is `Bearer`. If not, cancel
3. Try parse JWT token. If that fails, cancel
4. Check issuer to see if it's the configured PB token issuer. If not, cancel
5. Check audience and see if it's the configured PB token audience. If not, cancel
6. Load PB public key from config
7. Parse PB token claims
8. If the PB token is cluster auth, use NetID `000000`. If the PB token is network auth, take the first NetID as authenticated NetID
9. Do not fill address

#### How do you propose to test this?

CI

#### Can you do this yourself and submit a Pull Request?

Yes

Issue #4678, opened 2021-09-29
Expose interop for Backend Interfaces on Identity Server


johanstokking:
#### Summary

Expose interop for Backend Interfaces on Identity Server; at least to answer to `HomeNSReq`.

Blocked by https://github.com/TheThingsNetwork/lorawan-stack/issues/4677

#### Why do we need this?

For external services to get the home Network Server's NetID, tenant ID and cluster ID.

#### What is already there? What do you see now?

Interop is only exposed by Join Server to handle the join flow.

#### What is missing? What do you want to see?

Expose interop from Identity Server as well.

#### How do you propose to implement this?

Use existing `pkg/interop` server when it got a nice update (see referenced issue).

#### How do you propose to test this?

CI and integration testing

#### Can you do this yourself and submit a Pull Request?

Yes

johanstokking:
Discussion point is what we do with the NetID and NSID that Identity Server returns.

For the NetID, there are three options:

1. Introduce new `is.network.net-id`. This is implemented in #4726 
2. Introduce `net-id` at the top-level and make NS's NetID fallback to that if empty. I think this is less clean than (1), as we still have two configuration fields, and now with weird config if the top-level `net-id` is different than the lower level `ns.net-id`. It's very much the same with all the host names that we have in TTS; they all have to be set likely to the same host for things to work.
3. Somehow get the value for `ns.net-id` when `is.net-id` is unset. This is dirty and adds coupling between IS and NS

The NSID needs to include the Tenant ID (optional) and Cluster ID. The implementation in #4726 returns the Network Server address as NSID, which may or may not include a Tenant ID. The tenant-less part is the Cluster ID.

In The Things Stack Open Source, we do not have a Tenant ID in the top-level config. We do have a Tenant ID in the Packet Broker Agent configuration. NSID has to include the Tenant ID if the TTSOS deployment uses a tenant in a host NetID. So, basically same issue as with NetID:

1. Introduce new `is.network.tenant-id`
2. Introduce `tenancy.default-id` to harmonize with TTSE config. PBA can also fallback to this value

@htdvisser thoughts? My preference would be (1) for NetID and (2) for Tenant ID.

htdvisser:
I think we just need to accept that our configuration is getting messier as we add features or move them around. Eventually we'll just have to re-think our configuration system and make sure we do it better when we bump to V4.

[...]

Issue #4677, opened 2021-09-29
Enable interop HTTP based authentication and resolve technical debt


johanstokking:
#### Summary

Enable HTTP-based authentication for interop. Whenever we touch this, we should also resolve technical debt.

References https://github.com/TheThingsNetwork/lorawan-stack/issues/717
References https://github.com/TheThingsNetwork/lorawan-stack/issues/2627

#### Why do we need this?

To support Backend Interfaces clients that do not use TLS client authentication, but that we allow authenticators that use an API key or token.

#### What is already there? What do you see now?

Currently, `pkg/interop`'s server only supports TLS client authentication.

The aforementioned package is also built around Echo.

#### What is missing? What do you want to see?

TLS client authentication should become optional and we should support dynamic authentication mechanisms. One authentication mechanism could be checking an API key against a configured value or validating a token.

Use the more generic HTTP middleware instead of Echo.

#### How do you propose to implement this?

1. Add non-TLS endpoint for interop
2. On the TLS endpoint, do not require a TLS client certificate
3. Replace Echo middleware by generic HTTP middleware
4. Add a dynamic authenticator mechanism that sets `pkg/auth` info on the request context
5. Turn Join Server's authorizer based on X.509 Distinguished Name in an authenticator that checks the authenticated name in a more generic way, as it may also come from a token or API key

#### How do you propose to test this?

CI

#### Can you do this yourself and submit a Pull Request?

Yes

johanstokking:
Blocking #4678

PR #4707, opened 2021-10-04
Refactor interop server

johanstokking:
#### Summary

Main part of #4677 
Addresses #2627 (interop server and testing)
Addresses #2798 (`joinserver.proto`)
References #717 
References #718 

#### Changes

- Refactor `pkg/interop` server part
- Introduce more generic authorizer that replaces implementation in `pkg/joinserver`. This prepares for usage in IS (#4678) and allows for multiple authentication methods (like Packet Broker #4703)
- Resolved quite some technical debt: Echo dependency and gogoproto tags in `joinserver.proto`
- Support DNS SANs in client certificates instead of using CommonName for the origin address. Regenerated test certs to test this
- Interpret Network Server address as `HNSID` in `HomeNSAns`
- Remove other usages of NSID; it's 1.1 only and we have yet to see to which extent we're going to use `SenderNSID` and `ReceiverNSID` here
- Send correct message type depending on the protocol version, also some validation for incoming requests
- Cleaned up test cases

#### Testing

CI covers all logic. Also local integration testing:

```bash
$ mkdir -p .dev/interop
$ CAROOT=.dev/interop mkcert -cert-file .dev/interop/cert.pem -key-file .dev/interop/key.pem -client localhost "*.localhost"
$ echo "000001: rootCA.pem" > .dev/interop/config.yml
```

Configure TTS with:

```yaml
interop:
  source: directory
  directory: .dev/interop
```

Example `HomeNSReq` (notice `HNSID` that is set to `network_server_address`):

```bash
$ curl --cacert ca.pem --cert .dev/interop/cert.pem --key .dev/interop/key.pem -H "Content-Type: application/json" -d '{"ProtocolVersion":"1.1","MessageType":"HomeNSReq","SenderID":"000001","ReceiverID":"70B3D57ED0000000","DevEUI":"1122334455667788"}' https://localhost:8886

{"ProtocolVersion":"1.1","TransactionID":0,"MessageType":"HomeNSAns","SenderID":"70B3D57ED0000000","ReceiverID":"000001","Result":{"ResultCode":"Success"},"HNSID":"localhost","HNetID":"000001"}
```

Example `JoinReq`:

```
$ curl --cacert ca.pem --cert .dev/interop/cert.pem --key .dev/interop/key.pem -H "Content-Type: application/json" -d '{"ProtocolVersion":"1.0","MessageType":"JoinReq","SenderID":"000001","ReceiverID":"70B3D57ED0000000","DevEUI":"1122334455667788","TransactionID":42,"MACVersion":"1.0.3","PHYPayload":"00000000d07ed5b37088776655443322111100e6ed0bb7","DevAddr":"26ABCDEF","DLSettings":"FF","RxDelay":5,"CFList":"FFFFFFFFFFFFFFFFFFFFFFFFFFFFFF01"}' https://localhost:8886/

{"ProtocolVersion":"1.0","TransactionID":42,"MessageType":"JoinAns","SenderID":"70B3D57ED0000000","ReceiverID":"000001","PHYPayload":"205697687B6E09FC8FB45D3CA4A8B4C740D05DF36B6B7634D37A356556798D9712","Result":{"ResultCode":"Success"},"Lifetime":0,"NwkSKey":{"KEKLabel":"","AESKey":"9D9F8AFDD16C43975580831A92DAAFB3"},"AppSKey":{"KEKLabel":"","AESKey":"D958DC9AC061A4458C6422A5C531E815"},"SessionKeyID":"017C4BE9138646EB4A4C6F3A1853E28D"}
```

##### Regressions

None expected

#### Notes for Reviewers

Since it's effectively a new implementation of the server part of `pkg/interop`, the diff view isn't very useful. You might want to check out the code, reset the index to the target branch and see in your code editor where the changes are and if the current code makes sense.

Note that there are no major changes in testing so the existing tests act as regression test.

@adriansmares is the main reviewer.

#### Checklist


pkg/interop/server_errors.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	msg := ErrorMessage{
+		MessageHeader: answerHeader,
+		Result: Result{
+			ResultCode:  code,
+			Description: desc,
+		},
+	}
+	json.NewEncoder(w).Encode(msg)
```
adriansmares, line 53: Nit: `w.WriteHeader(http.StatusBadRequest)` on non-`nil` error. At least we get infrastructure errors if this happens.

pkg/joinserver/http_interop.go
```diff
@@ -119,26 +124,40 @@ func (srv interopServer) JoinRequest(ctx context.Context, in *interop.JoinReq) (
 func (srv interopServer) HomeNSRequest(ctx context.Context, in *interop.HomeNSReq) (*interop.HomeNSAns, error) {
 	ctx = log.NewContextWithField(ctx, "namespace", "joinserver/interop")
 
-	netID, err := srv.JS.GetHomeNetID(ctx, types.EUI64(in.ReceiverID), types.EUI64(in.DevEUI), X509DNAuthorizer)
+	netID, nsID, err := srv.JS.GetHomeNetID(ctx, types.EUI64(in.ReceiverID), types.EUI64(in.DevEUI), InteropAuthorizer)
 	if err != nil {
+		switch {
+		case errors.Resemble(err, errRegistryOperation):
+			if errors.IsNotFound(errors.Cause(err)) {
+				return nil, interop.ErrUnknownDevEUI.WithCause(err)
```
adriansmares, line 132: Could also be `ReceiverID` / `JoinEUI` - maybe unknown EUIs ?

adriansmares (approved):
Did you find out what was causing the not found errors during our testing of `HomeNSReq` ?

pkg/interop/server_errors.go
johanstokking: The headers are already sent as soon as the encoder emitted the first characters. I don't think we can do much here, other than first encoding the response in memory and then writing it. TBH, I don't see a case here where JSON marshaling fails; these structs are under our control, there are no functions being marshaled or `interface{}` fields etc

johanstokking:
@adriansmares I appended a few commits, you can only look at those.

pkg/joinserver/http_interop.go
johanstokking: Yes, in The Things Stack only the combination of JoinEUI and DevEUI is supposed to be unique. Generally speaking, LoRaWAN Network Servers keep DevEUIs unique, regardless of the JoinEUI.

Therefore there is no error code in Backend Interfaces that says `UnknownEUIs`; it's either `UnkonwnDevEUI` or `UnknownReceiver`.

Now, `UnkownReceiver` (sic) should be returned when the `ReceiverID` isn't known, i.e. when you're contacting a Join Server that is not handling this JoinEUI. We might want to support this through config (like a list of JoinEUI prefixes), but I don't really see a need for it.

johanstokking:
> Did you find out what was causing the not found errors during our testing of `HomeNSReq` ?

No, but things changed dramatically already. I didn't put time in figuring out what the reason was in the current version. It's probably something really stupid. I did some integration testing with this code to see if it really is resolved and it looks like it.

pkg/interop/messages.go
```diff
@@ -58,6 +58,18 @@ type ErrorMessage struct {
 	Result Result
 }
 
+// NsMessageHeader contains the message header for NS messages.
+type NsMessageHeader struct {
+	MessageHeader
+	SenderID   NetID
```
adriansmares, line 64: Is the `0x` handling supposed to be removed ? I expected it to be in the unmarshaller maybe.

johanstokking: Yeah it's there, it goes via `Buffer`: https://github.com/TheThingsNetwork/lorawan-stack/blob/feature/4677-interop/pkg/interop/types.go#L279-L287 and https://github.com/TheThingsNetwork/lorawan-stack/blob/26e787cdd48bf43ce7372d6a48a1a92bc35aebb7/pkg/interop/types.go#L345-L348

adriansmares (approved):

pkg/config/shared.go
```diff
@@ -414,6 +414,7 @@ func (c SenderClientCA) Fetcher(ctx context.Context) (fetch.Interface, error) {
 // InteropServer represents the server-side interoperability through LoRaWAN Backend Interfaces configuration.
 type InteropServer struct {
 	ListenTLS                string            `name:"listen-tls" description:"Address for the interop server to listen on"`
+	TrustedProxies           []string          `name:"trusted-proxies" description:"CIDRs of trusted reverse proxies"`
```
htdvisser, line 417: This should be mentioned in CHANGELOG.

htdvisser (approved):
Looks good to me 👍

PR #4721, opened 2021-10-07
Expand toggle sections with invalid fields

### LINKED ISSUE #4558: Automatically expand form toggle sections when a validation within fails

kschiffer:
#### Summary
Follow up from #4684, with added fixes and improvements. Many thanks to @brycechampaign for co-authoring this 👍!

Closes #4558.

#### Changes
- Remove HTML5 form validation form our forms (we use our own validation system)
- Make form sections aware of the form fields they render and automatically expand the section when it has an invalid field

#### Testing

Manual testing. See reproduction steps in #4558.

#### Notes for Reviewers
The HTML5 validation was previously preventing the submit logic from being fired when a field was deemed as invalid by the browser. E.g. when entering a `e` into a number field (this is possible because it can be used to enter exponents) and trying to submit. Instead of working around that, I concluded that we don't need the HTML5 validation since we have our own validation system in place which takes care of any input error anyway and does not trigger arbitrary side-effects such as blocking the `onSubmit` handler. This also streamlines the validation UX.

#### Checklist

### DIFF
--- a/pkg/webui/components/form/section/index.js
+++ b/pkg/webui/components/form/section/index.js
@@ -22,8 +22,20 @@ import Icon from '@ttn-lw/components/icon'
 import Message from '@ttn-lw/lib/components/message'
 
 import PropTypes from '@ttn-lw/lib/prop-types'
+import getByPath from '@ttn-lw/lib/get-by-path'
+
+import Field from '../field'
 
 import style from './section.styl'
+
+const recursiveMap = (children, fn) =>
+  React.Children.map(children, child => {
+    if (child && child.props && child.props.children) {
+      return [fn(child), recursiveMap(child.props.children, fn)]
+    }
+
+    return fn(child)
+  })
 
 const m = defineMessages({
   expand: 'Expand section',
@@ -35,13 +47,13 @@ const m = defineMessages({
 const FormCollapseSection = props => {
   const { className, id, title, onCollapse, isCollapsed, initiallyCollapsed, children } = props
   const { formatMessage } = useIntl()
-  const { disabled } = useFormContext()
+  const { disabled, errors, isSubmitting, isValid } = useFormContext()
 
   // Check if the component is 'controlled'. When the `isCollapsed` prop is passed the component
   // is considered as 'uncontrolled' and it's state must be controlled from the outside.
   const isControlled = typeof isCollapsed === 'undefined'
-
   const [collapsed, setCollapsed] = React.useState(initiallyCollapsed)
+
   const onExpandedChange = React.useCallback(() => {
     if (isControlled) {
       setCollapsed(collapsed => !collapsed)
@@ -49,6 +61,26 @@ const FormCollapseSection = props => {
       onCollapse()
     }
   }, [isControlled, onCollapse])
+
+  // Trigger an expand if there is an error in a field within the section when submitting.
+  // Uses `errorTitles` to determine what types of errors should trigger an expand.
+  React.useEffect(() => {
+    if (collapsed) {
+      const fieldNames = recursiveMap(children, child => {
+        if (child !== null && child.type === Field && 'name' in child.props) {
+          return child.props.name
+        }
+      })
+      if (isSubmitting && !isValid) {
+        for (const fieldName of fieldNames) {
+          if (getByPath(errors, fieldName) !== undefined) {
+            setCollapsed(false)
+            break
+          }
+        }
+      }
+    }
+  }, [isSubmitting, isValid, children, errors, collapsed])
 
   const isSectionClosed = isControlled ? collapsed : isCollapsed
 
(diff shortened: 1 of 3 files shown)


pkg/webui/components/form/section/index.js
```diff
@@ -22,9 +22,21 @@ import Icon from '@ttn-lw/components/icon'
...
 import PropTypes from '@ttn-lw/lib/prop-types'
+import getByPath from '@ttn-lw/lib/get-by-path'
+
+import Field from '../field'
 
 import style from './section.styl'
 
+const recursiveMap = (children, fn) =>
+  React.Children.map(children, child => {
+    if (child && child.props && child.props.children) {
```
bafonins, line 33: In this case recursion goes further than the `Field` component which must be the base case. Consider the following example:
```js
      <Form.Field
        component={Radio.Group}
      >
        <Radio />
        <Radio />
        <Radio />
      </Form.Field>
```
With current implementation `recursiveMap` also visits each `<Radio />` component which is not needed at all.

pkg/webui/components/form/section/index.js
```diff
@@ -50,6 +62,26 @@ const FormCollapseSection = props => {
     }
   }, [isControlled, onCollapse])
 
+  // Trigger an expand if there is an error in a field within the section when submitting.
+  // Uses `errorTitles` to determine what types of errors should trigger an expand.
+  React.useEffect(() => {
+    if (collapsed) {
+      const fieldNames = recursiveMap(children, child => {
```
bafonins, line 69: This is executed on every interaction with the form, however we are interested in form fields that failed validation only on submission.

pkg/webui/components/form/section/index.js
```diff
@@ -22,9 +22,21 @@ import Icon from '@ttn-lw/components/icon'
...
 
 import style from './section.styl'
 
+const recursiveMap = (children, fn) =>
+  React.Children.map(children, child => {
+    if (child && child.props && child.props.children) {
+      return [fn(child), recursiveMap(child.props.children, fn)]
+    }
+
+    return fn(child)
```
bafonins, line 37: Do we really need recursion here?

pkg/webui/components/form/section/index.js
```diff
@@ -50,6 +62,26 @@ const FormCollapseSection = props => {
...
+    if (collapsed) {
+      const fieldNames = recursiveMap(children, child => {
+        if (child !== null && child.type === Field && 'name' in child.props) {
+          return child.props.name
+        }
+      })
+      if (isSubmitting && !isValid) {
+        for (const fieldName of fieldNames) {
+          if (getByPath(errors, fieldName) !== undefined) {
+            setCollapsed(false)
```
bafonins, line 77: The only reason why this works is because of [this](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.15/pkg/webui/console/views/device-add/manual/form/form.js#L359-L363). If you remove that part the section wont expand on validation errors.
This PR should replace [this check](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.15/pkg/webui/console/views/device-add/manual/form/form.js#L359-L363) and make sure that the form "knows" when to expand section components without explicitly specifying this for each field.

pkg/webui/components/form/section/index.js
```diff
@@ -50,6 +62,26 @@ const FormCollapseSection = props => {
     }
   }, [isControlled, onCollapse])
 
+  // Trigger an expand if there is an error in a field within the section when submitting.
+  // Uses `errorTitles` to determine what types of errors should trigger an expand.
+  React.useEffect(() => {
+    if (collapsed) {
+      const fieldNames = recursiveMap(children, child => {
+        if (child !== null && child.type === Field && 'name' in child.props) {
```
bafonins, line 70:
```suggestion
        if (React.isValidElement(child) && child.type === Field) {
```

```js
React.isValidElement(null) // => false
```

bafonins (changes requested):

kschiffer:
Closing this for now until we have capacity to work on this.

PR #4723, opened 2021-10-07
Support Packet Broker authentication through Backend Interfaces

### LINKED ISSUE #4677: Enable interop HTTP based authentication and resolve technical debt

### LINKED ISSUE #4703: Support Packet Broker authentication on interop server

johanstokking:
#### Summary

Support Packet Broker authentication through Backend Interfaces.

Closes #4677 
Closes #4703 
Unblocks #4678 

#### Changes

Next to TLS mutual authentication, the interop server for Backend Interfaces now also optionally verifies tokens via the HTTP Authorization header.

The only token issuer supported is currently Packet Broker. The way this works is that if a JWT is detected and the issuer matches with the configured issuer (by default `https://iam.packetbroker.net`), its signature is verified against the public keys hosted by this issuer. Also the audience is verified against the public TLS address of the interop server, to avoid that TTS deployment A can use a token from Packet Broker on TTS deployment B.

#### Testing

Unit testing and local integration testing.

I happen to have Packet Broker running locally (issuer `https://localhost:1916`) so I can do something like:

```bash
$ client_id=...
$ client_secret=...
$ curl -d 'client_id=$client_id&client_secret=$client_secret&grant_type=client_credentials&audience=https://localhost:8886' http://localhost:1916/token
```

The returned token is a Packet Broker cluster token, that is, not a network but a Packet Broker routing cluster. The claims of such a token look like this:

```json
{
  "aud": [
    "https://localhost:8886"
  ],
  "exp": 1633625722,
  "https://iam.packetbroker.net/claims": {
    "c": true,
    "rights": [0, 1, 4, 5, 6]
  },
  "iat": 1633622122,
  "iss": "http://localhost:1916",
  "jti": "01FHDQEQHYGK6VDN7F8FH60V68",
  "sub": "localhost"
}
```

With the compact serialized token in `$token`, this now works:

```bash
curl -H "Content-Type: application/json" \
 	-H "Authorization: Bearer $token" \
	-d '{"ProtocolVersion":"1.1","MessageType":"HomeNSReq","SenderID":"000000","SenderNSID":"localhost","ReceiverID":"70B3D57ED0000000","DevEUI":"1122334455667788"}' \
	https://localhost:8886

{"ProtocolVersion":"1.1","TransactionID":0,"MessageType":"HomeNSAns","SenderID":"70B3D57ED0000000","ReceiverID":"000000","ReceiverNSID":"localhost","Result":{"ResultCode":"Success"},"HNSID":"localhost","HNetID":"000001"}
```

The following checks are in place:
1. The token's issuer `iss` matches the configured Packet Broker token issuer
2. The token is signed by a private key belonging to a public key that is advertised at `<issuer>/.well-known/jwks.json`. This is refreshed at most every 10 minutes
3. The token's audience `aud` contains the public TLS address of The Things Stack interop server
4. The token's Packet Broker claims has cluster (`c`) set to `true` (that is, it's Packet Broker itself, like Packet Broker Router; not a token issued for a Packet Broker network or CLI)
5. The token is valid in time (now is between `iat` and `exp`)
6. The token's subject `sub` has to match the `SenderNSID`. In production, this will be something like `eu.packetbroker.io`
7. The `SenderID` must be `000000`

##### Regressions

None expected

#### Notes for Reviewers

@adriansmares please review everything

@htdvisser please review:
- https://github.com/TheThingsNetwork/lorawan-stack/commit/1485f6906ddc30d846d4881e4e5694a1ae8a2cd9
- https://github.com/TheThingsNetwork/lorawan-stack/commit/5520e1b2df1d0ea43cc414663a76075de7a54460

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,13 @@ For details about compatibility between different releases, see the **Commitment
 ### Added
 
 - Support for enhanced security policies of Packet Broker services.
-- Backend Interfaces middleware to extract proxy headers from trusted proxies. This adds a configuration `interop.trusted-proxies` that is similar to the existing `http.trusted-proxies` option.
 - Handling of MAC and PHY versions in end device forms based on selected frequency plan in the Console.
 - Support for scheduling downlink messages as JSON in the Console.
+- Backend Interfaces middleware to extract proxy headers from trusted proxies. This adds a configuration `interop.trusted-proxies` that is similar to the existing `http.trusted-proxies` option.
+- Support for Packet Broker authentication through LoRaWAN Backend Interfaces. This adds the following configuration options:
+  - `interop.public-tls-address`: public address of the interop server. The audience in the incoming OAuth 2.0 token from Packet Broker is verified against this address to ensure that other networks cannot impersonate as Packet Broker;
+  - `interop.packet-broker.enabled`: enable Packet Broker to authenticate;
+  - `interop.packet-broker.token-issuer`: the issuer of the incoming OAuth 2.0 token from Packet Broker is verified against this value.
 
 ### Changed
 
--- a/pkg/config/shared.go
+++ b/pkg/config/shared.go
@@ -411,12 +411,22 @@ func (c SenderClientCA) Fetcher(ctx context.Context) (fetch.Interface, error) {
 	}
 }
 
+// PacketBrokerInteropAuth respresents authen
+type PacketBrokerInteropAuth struct {
+	Enabled     bool   `name:"enabled" description:"Enable Packet Broker to authenticate"`
+	TokenIssuer string `name:"token-issuer" description:"Required issuer of Packet Broker tokens"`
+}
+
 // InteropServer represents the server-side interoperability through LoRaWAN Backend Interfaces configuration.
 type InteropServer struct {
-	ListenTLS                string            `name:"listen-tls" description:"Address for the interop server to listen on"`
-	TrustedProxies           []string          `name:"trusted-proxies" description:"CIDRs of trusted reverse proxies"`
+	ListenTLS        string   `name:"listen-tls" description:"Address for the interop server to listen on"`
+	PublicTLSAddress string   `name:"public-tls-address" description:"Public address of the interop server"`
+	TrustedProxies   []string `name:"trusted-proxies" description:"CIDRs of trusted reverse proxies"`
+
 	SenderClientCA           SenderClientCA    `name:"sender-client-ca"`
 	SenderClientCADeprecated map[string]string `name:"sender-client-cas" description:"Path to PEM encoded file with client CAs of sender IDs to trust; deprecated - use sender-client-ca instead"`
+
+	PacketBroker PacketBrokerInteropAuth `name:"packet-broker"`
 }
 
 // ServiceBase represents base service configuration.
--- a/pkg/config/shared.go
+++ b/pkg/config/shared.go
@@ -419,9 +419,8 @@ type PacketBrokerInteropAuth struct {
 
 // InteropServer represents the server-side interoperability through LoRaWAN Backend Interfaces configuration.
 type InteropServer struct {
-	ListenTLS        string   `name:"listen-tls" description:"Address for the interop server to listen on"`
-	PublicTLSAddress string   `name:"public-tls-address" description:"Public address of the interop server"`
-	TrustedProxies   []string `name:"trusted-proxies" description:"CIDRs of trusted reverse proxies"`
+	ListenTLS        string `name:"listen-tls" description:"Address for the interop server to listen on"`
+	PublicTLSAddress string `name:"public-tls-address" description:"Public address of the interop server"`
 
 	SenderClientCA           SenderClientCA    `name:"sender-client-ca"`
 	SenderClientCADeprecated map[string]string `name:"sender-client-cas" description:"Path to PEM encoded file with client CAs of sender IDs to trust; deprecated - use sender-client-ca instead"`
@@ -438,7 +437,7 @@ type ServiceBase struct {
 	Events           Events               `name:"events"`
 	GRPC             GRPC                 `name:"grpc"`
 	HTTP             HTTP                 `name:"http"`
-	Interop          InteropServer        `name:"interop"`
+	Interop          InteropServer        `name:"interop" description:"LoRaWAN Backend Interfaces interoperability configuration"`
 	TLS              tlsconfig.Config     `name:"tls"`
 	Sentry           Sentry               `name:"sentry"`
 	Blob             BlobConfig           `name:"blob"`
--- a/pkg/interop/server.go
+++ b/pkg/interop/server.go
@@ -204,6 +204,7 @@ func (s *Server) handle() http.Handler {
 		}
 		ctx, err = senderAuthenticator.Authenticate(ctx, r, data)
 		if err != nil {
+			logger.WithError(err).Warn("Failed to authenticate")
 			writeError(w, r, header, err)
 			return
 		}
--- a/pkg/interop/server.go
+++ b/pkg/interop/server.go
@@ -67,6 +67,8 @@ type Server struct {
 	senderClientCAs    map[string][]*x509.Certificate
 	senderClientCAPool *x509.CertPool
 
+	tokenVerifiers map[string]tokenVerifier
+
 	js JoinServer
 }
 
@@ -74,13 +76,15 @@ type Server struct {
 type Component interface {
 	Context() context.Context
 	RateLimiter() ratelimit.Interface
+	HTTPClient(context.Context) (*http.Client, error)
 }
 
 // NewServer builds a new server.
 func NewServer(c Component, contextFillers []fillcontext.Filler, conf config.InteropServer) (*Server, error) {
-	logger := log.FromContext(c.Context()).WithField("namespace", "interop")
+	ctx := c.Context()
+	logger := log.FromContext(ctx).WithField("namespace", "interop")
 
-	senderClientCAs, err := fetchSenderClientCAs(c.Context(), conf)
+	senderClientCAs, err := fetchSenderClientCAs(ctx, conf)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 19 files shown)


pkg/config/shared.go
```diff
@@ -428,7 +437,7 @@ type ServiceBase struct {
 	Events           Events               `name:"events"`
 	GRPC             GRPC                 `name:"grpc"`
 	HTTP             HTTP                 `name:"http"`
-	Interop          InteropServer        `name:"interop"`
+	Interop          InteropServer        `name:"interop" description:""`
```
adriansmares: Is this needed ?

pkg/interop/server.go
```diff
@@ -67,20 +67,24 @@ type Server struct {
...
 type Component interface {
 	Context() context.Context
 	RateLimiter() ratelimit.Interface
+	HTTPClient(context.Context) (*http.Client, error)
 }
 
 // NewServer builds a new server.
 func NewServer(c Component, contextFillers []fillcontext.Filler, conf config.InteropServer) (*Server, error) {
-	logger := log.FromContext(c.Context()).WithField("namespace", "interop")
+	ctx := c.Context()
```
adriansmares: You may put the fields in the context already such that subsequent calls have a namespace too (`fetchSenderClientCAs` for example).

pkg/interop/server_authn.go
```diff
@@ -88,102 +66,167 @@ nextPattern:
...
+		func(ctx context.Context) (*NetworkServerAuthInfo, error) {
+			logger := log.FromContext(ctx).WithField("authenticator", "packetbroker")
+			authz := strings.SplitN(r.Header.Get("Authorization"), " ", 2)
+			if len(authz) < 2 || strings.ToLower(authz[0]) != "bearer" {
+				return nil, nil
+			}
+			token, err := jwt.ParseSigned(authz[1])
+			if err != nil {
+				logger.WithError(err).Debug("Failed to parse token")
+				return nil, nil
```
adriansmares, line 181: This is pretty bad remote developer UX, but I guess we're swallowing the error such that we allow other authorization strategies to still run.

Another possibility is to aggregate the errors and to provide them as causes, and only return the aggregated error if none of the strategies match. If the error details are not rendered just skip this comment.

https://github.com/TheThingsNetwork/lorawan-stack/blob/658c8cef46fb7c397c19ee8f12446502fe8b68b3/pkg/applicationserver/io/pubsub/provider/connection.go#L98-L124

pkg/packetbroker/token.go
```diff
@@ -0,0 +1,253 @@
...
+		return key, err
+	})
+}
+
+var errFetchToken = errors.DefineAborted("fetch_token", "fetch token")
+
+// PublicKeyFromURL loads the public keys from the given URL.
+func PublicKeyFromURL(client *http.Client, url string) PublicKeyProvider {
+	return PublicKeyProviderFunc(func(ctx context.Context) (*jose.JSONWebKeySet, error) {
+		req, err := http.NewRequest(http.MethodGet, url, nil)
```
adriansmares: You can use `http.NewRequestWithContext` directly.

adriansmares (approved):
Nice

pkg/config/shared.go
johanstokking: No, this is what you get when you have 2 minutes spare between two calls. Will remove.

pkg/interop/server_authn.go
johanstokking: Hmm it is really best effort. We may add more token verifications and we would be trying them in no particular order.

The result is the same anyway: `UnknownSender` with status code 200.

The only difference is the logging really. Maybe we should warn here if the issuer matches but the verification doesn't?

johanstokking: But if we support API keys or other token formats in the future, we don't want previous verification attempts to always warn, right?

adriansmares: > The result is the same anyway: UnknownSender with status code 200. 

I was afraid this was the case. Then it doesn't help if we aggregate the errors or something, since the caller cannot hear the reasoning anyway.

> The only difference is the logging really. Maybe we should warn here if the issuer matches but the verification doesn't?

Not needed given that indeed we'll `WARN` even when another authorization strategy matches.

CHANGELOG.md
```diff
@@ -12,7 +12,10 @@ For details about compatibility between different releases, see the **Commitment
 ### Added
 
 - Support for enhanced security policies of Packet Broker services.
-- Backend Interfaces middleware to extract proxy headers from trusted proxies. This adds a configuration `interop.trusted-proxies` that is similar to the existing `http.trusted-proxies` option.
+- Support for Packet Broker authentication through LoRaWAN Backend Interfaces. This adds the following configuration options:
+  - `interop.public-tls-address`: public address of the interop server. The audience in the incoming OAuth 2.0 token from Packet Broker is verified against this address to ensure that other networks cannot impersonate as Packet Broker;
```
htdvisser, line 19: This _must_ be set on existing deployments, right? Otherwise they'll be configured with the default `https://localhost:8886`, and will accept tokens that are not for them.

johanstokking: Yes indeed. We don't have a global setting for this, right?

Should we leave this value empty and should we fail the interop server init (and stack start) when Packet Broker token authentication is enabled? That forces users to enter something sane here.

johanstokking: Then we get something like this:

```diff
diff --git a/cmd/internal/shared/config.go b/cmd/internal/shared/config.go
index 4c318c47aa..2135ac2eff 100644
--- a/cmd/internal/shared/config.go
+++ b/cmd/internal/shared/config.go
@@ -73,10 +73,9 @@ var DefaultHTTPConfig = config.HTTP{
 
 // DefaultInteropServerConfig is the default interop server config.
 var DefaultInteropServerConfig = config.InteropServer{
-	ListenTLS:        ":8886",
-	PublicTLSAddress: "https://" + DefaultPublicHost + ":8886",
+	ListenTLS: ":8886",
 	PacketBroker: config.PacketBrokerInteropAuth{
-		Enabled:     true,
+		Enabled:     false,
 		TokenIssuer: packetbroker.DefaultTokenIssuer,
 	},
 }
diff --git a/pkg/interop/errors.go b/pkg/interop/errors.go
index 445d581c45..efb319ac96 100644
--- a/pkg/interop/errors.go
+++ b/pkg/interop/errors.go
@@ -19,6 +19,7 @@ import (
 )
 
 var (
+	errNoPublicTLSAddress  = errors.DefineFailedPrecondition("no_public_tls_address", "no public TLS address configured for interop")
 	errUnknownMACVersion   = errors.DefineInvalidArgument("unknown_mac_version", "unknown MAC version")
 	errInvalidLength       = errors.DefineInvalidArgument("invalid_length", "invalid length")
 	errInvalidRequestType  = errors.DefineInvalidArgument("invalid_request_type", "invalid request type `{type}`")
diff --git a/pkg/interop/server.go b/pkg/interop/server.go
index 31f0569ff6..13a01eab4c 100644
--- a/pkg/interop/server.go
+++ b/pkg/interop/server.go
@@ -98,7 +98,11 @@ func NewServer(c Component, contextFillers []fillcontext.Filler, conf config.Int
 	tokenVerifiers := make(map[string]tokenVerifier)
 	if conf.PacketBroker.Enabled {
 		iss := conf.PacketBroker.TokenIssuer
-		aud := conf.PublicTLSAddress // The token audience must match the configured public TLS address.
+		// The token audience must match the configured public TLS address. Therefore, a non-empty value must be set.
+		aud := conf.PublicTLSAddress
+		if aud == "" {
+			return nil, errNoPublicTLSAddress.New()
+		}
 		tokenVerifier, err := newPacketBrokerTokenVerifier(ctx, iss, aud, c)
 		if err != nil {
 			return nil, err
```

johanstokking:
@htdvisser only check https://github.com/TheThingsNetwork/lorawan-stack/pull/4723#discussion_r724877869 please.

CHANGELOG.md
htdvisser: Yes, that looks okay to me.

johanstokking:
I'm taking earlier review and [approval on the only comment](https://github.com/TheThingsNetwork/lorawan-stack/pull/4723#discussion_r724952035) as an approval.

PR #4724, opened 2021-10-07
console: Fix consecutive desconnect/reconnect messages

### LINKED ISSUE #4685: Avoid consecutive disconnect/reconnect messages in the Live Data view
#### Summary
The live data view currently shows many consecutive connect and reconnect events when the internet connection is lost. This is frequently the case when e.g. closing the laptop or leaving the browser tab unfocused for a long time. We should combine such messages so that only two are shown.

image

#### Steps to Reproduce

1. Go to live data view
2. Either leave the browser tab unfocused for a long while or (easier) disconnect from Wi-Fi or internet a couple of times
3. See many synthetic events being generated about that consecutively for each disconnect event

#### What do you see now?
Many disconnect/reconnect errors being generated consecutively.

Such consecutive events to be combined into just two events:
1. The first disconnect event of the consecutive group
2. The last reconnect event of the consecutive group

This also means that such events should not be combined if there was an actual stream event in between them.

It's probably best to solve this at the event reducer level, by changing the reducer to only add the synthetic reconnect/disconnect event if there haven't been the same events immediately preceding. In that case, we can simply remove the last reconnect event and replace it with the more recent one.

Manual testing will suffice for now.

This is a good first issue.

ryaplots:
#### Summary
Closes #4685 - Avoid consecutive disconnect/reconnect messages in the Live Data view

...

#### Changes
file changed: pkg/webui/console/store/reducers/events.js: 
- added 12 lines, 1 if 
- added the logic to fix the show of consecutive messages

- ...
- ...

#### Testing

Disconnected and reconnected the internet/WiFi to verify that only the fist disconnect and last reconnect is shown

...

##### Regressions
I think no features will be affected. It's something very focused on the reducer.

...

#### Notes for Reviewers

...

#### Checklist

- [v] Scope: The referenced issue is addressed, there are no unrelated changes.
- [v] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [v] Documentation: Relevant documentation is added or updated.
- [v] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [v] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

### DIFF
--- a/pkg/webui/console/store/reducers/events.js
+++ b/pkg/webui/console/store/reducers/events.js
@@ -75,7 +75,21 @@ const addEvent = (state, event) => {
     return {}
   }
 
-  const currentEvents = events
+  let nonConsecutiveEvents = events
+
+  // Avoid consecutive disconnect/reconnect messages in the Live Data view.
+  if (events.length > 1 && (event.name === events[0].name || event.name === events[1].name)) {
+    if (event.name === 'synthetic.status.reconnected') {
+      const lastTwoEvents = [events[0], events[1]]
+      const prevIndex = lastTwoEvents.map(e => e.name).indexOf(event.name)
+
+      nonConsecutiveEvents = prevIndex === 0 ? events.slice(1) : events[0] + events.slice(2)
+    } else if (event.name === 'synthetic.error.unknown') {
+      return {} // If the disconnect already exists we don't want to add the event
+    }
+  }
+
+  const currentEvents = nonConsecutiveEvents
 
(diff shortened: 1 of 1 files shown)


kschiffer:
Thanks for the contribution @ryaplots. I will close this and reopen it with some modifications as discussed.

PR #4734, opened 2021-10-11
Lock based concurrency

### LINKED ISSUE #2696: Lock-based concurrency in registries
#### Summary

Replaces #33 
Lock values in registries on SET.

#### Why do we need this?

To ensure correctness and optimal performance.

In all registries except NS(after #2662) we use optimistic concurrency, where the modified keys are `WATCH`'ed and in case the transaction fails, we simply fail it without retrying.
In #2662 an efficient Redis mutex lock implementation is added. It is essentially a drop-in replacement for `redis.Watch`. The implementation only locks 1 key, but can be extended to support multiple easily if there's a use case for it.

Use `redis.LockedWatch` from #2662 (and, perhaps, lock other important keys depending on the implementation) to avoid possibility of transaction failure and remove the need for retrying transactions. Failed transactions in this case represent an implementation error.
Regarding the "transaction failures only happen in NS" or "the request count is small enough, that such issues rarely happen"(e.g. for JS) argument: concurrency-induced transaction failures shall not be possible. The overhead of locking is extremely small and so there's no reason not to do it.
At scale it ensures performance (versus the optimistic concurrency and retry solution) and correctness. At small scale it ensures correctness and has a negligible effect on performance.
Consider the case where the device would be updated in JS registry at the same time a join-request arrives - perfect example of a component, where clashes are very rare - with current implementation one of the operations would fail, but there's no reason for that to happen. Even though it's improbable for this to happen - it *can* happen and the more customers we have, the higher possibility of such. Even accidental concurrent set to a registry is a possible occasion and we can easily handle this case gracefully and give the user expected output instead of failing.

Please also see table at https://redislabs.com/ebook/part-2-core-concepts/chapter-6-application-components-in-redis/6-2-distributed-locking/6-2-4-fine-grained-locking/

Use `redis.LockedWatch`, go through each implementation probably adding additional locking logic. See #2662 for an example of NS part.

Integration testing, flow tests

yes

adriansmares:
#### Summary

Closes #2696 

#### Changes

- Add lock based concurrency to registries where optimistic concurrency was used
  - What this means in general is that we moved from `Watch` to `LockedWatch`, which wraps `Watch` with a mutex 
  - This has the same pattern:
    - Add `LockTTL` and mutex / entropy source
    - Ensure `Init` is called before the registry is used
    - Use `ttnredis.LockedWatch` instead of simple `Watch`
- Lock (mutex) TTL has been standardized to 10 seconds
  - The previous 1 second TTL has been a bit too optimistic for high load environments

#### Testing

Unit testing covers this.

##### Regressions

If I did something wrong, this may deadlock. With that being said, I'm pretty optimistic.

#### Checklist

### DIFF
--- a/pkg/applicationserver/io/pubsub/redis/registry.go
+++ b/pkg/applicationserver/io/pubsub/redis/registry.go
@@ -16,10 +16,14 @@ package redis
 
 import (
 	"context"
+	"io"
+	"math/rand"
+	"sync"
 	"time"
 
 	"github.com/go-redis/redis/v8"
 	"github.com/gogo/protobuf/proto"
+	"github.com/oklog/ulid/v2"
 	"go.thethings.network/lorawan-stack/v3/pkg/applicationserver/io/pubsub"
 	"go.thethings.network/lorawan-stack/v3/pkg/errors"
 	ttnredis "go.thethings.network/lorawan-stack/v3/pkg/redis"
@@ -52,7 +56,21 @@ func applyPubSubFieldMask(dst, src *ttnpb.ApplicationPubSub, paths ...string) (*
 
 // PubSubRegistry is a Redis pub/sub registry.
 type PubSubRegistry struct {
-	Redis *ttnredis.Client
+	Redis   *ttnredis.Client
+	LockTTL time.Duration
+
+	entropyMu *sync.Mutex
+	entropy   io.Reader
+}
+
+// Init initializes the PubSubRegistry.
+func (r *PubSubRegistry) Init(ctx context.Context) error {
+	if err := ttnredis.InitMutex(ctx, r.Redis); err != nil {
+		return err
+	}
+	r.entropyMu = &sync.Mutex{}
+	r.entropy = ulid.Monotonic(rand.New(rand.NewSource(time.Now().UnixNano())), 1000)
+	return nil
 }
 
 func (r *PubSubRegistry) allKey(ctx context.Context) string {
@@ -144,8 +162,13 @@ func (r PubSubRegistry) Set(ctx context.Context, ids ttnpb.ApplicationPubSubIden
 	appUID := unique.ID(ctx, ids.ApplicationIdentifiers)
 	ik := r.uidKey(appUID, ids.PubSubId)
 
+	lockerID, err := ttnredis.GenerateLockerID(r.entropy, r.entropyMu)
+	if err != nil {
+		return nil, err
+	}
+
 	var pb *ttnpb.ApplicationPubSub
-	err := r.Redis.Watch(ctx, func(tx *redis.Tx) error {
+	err = ttnredis.LockedWatch(ctx, r.Redis, ik, lockerID, r.LockTTL, func(tx *redis.Tx) error {
 		cmd := ttnredis.GetProto(ctx, tx, ik)
 		stored := &ttnpb.ApplicationPubSub{}
 		if err := cmd.ScanProto(stored); errors.IsNotFound(err) {
@@ -262,7 +285,7 @@ func (r PubSubRegistry) Set(ctx context.Context, ids ttnpb.ApplicationPubSubIden
 			return err
 		}
 		return nil
-	}, ik)
+	})
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/applicationserver/io/pubsub/redis/registry.go
+++ b/pkg/applicationserver/io/pubsub/redis/registry.go
@@ -16,14 +16,10 @@ package redis
 
 import (
 	"context"
-	"io"
-	"math/rand"
-	"sync"
 	"time"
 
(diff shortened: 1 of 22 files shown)


pkg/networkserver/redis/registry.go
```diff
@@ -905,10 +903,6 @@ func (r *DeviceRegistry) SetByID(ctx context.Context, appID ttnpb.ApplicationIde
 				}
 				if pb.JoinEui != nil && pb.DevEui != nil {
 					ek := r.euiKey(*pb.JoinEui, *pb.DevEui)
-
```
adriansmares, line 908: This was slightly overzealous. In theory it is safe, since the order of acquiring locks is `uidKey`, followed by `euiKey`, but unlocking the mutex is problematic (cannot be done using `tx`, because that would unlock it too soon, cannot be done with `p`, since then it wouldn't be unlocked if the transaction fails).

This can be fixed (create a function outside of the transaction and `defer` that) but it's quite ugly and doesn't protect much.

pkg/joinserver/redis/registry.go
```diff
@@ -313,10 +331,15 @@ func (r *DeviceRegistry) SetByEUI(ctx context.Context, joinEUI types.EUI64, devE
...
+	lockerID, err := ttnredis.GenerateLockerID(r.entropy, r.entropyMu)
+	if err != nil {
+		return nil, err
+	}
+
 	defer trace.StartRegion(ctx, "set end device by eui").End()
 
 	var pb *ttnpb.ContextualEndDevice
-	err := r.Redis.Watch(ctx, func(tx *redis.Tx) error {
+	err = ttnredis.LockedWatch(ctx, r.Redis, ek, lockerID, r.LockTTL, func(tx *redis.Tx) error {
```
adriansmares, line 333: We lock only in `SetByEUI`, but not in `SetByID` due to the way the keys work: you need to know the UID in order to find out the EUI, and vice versa. This means that you need to acquire the locks in a different order based on the call site, which is a textbook deadlock.

We can live with locking only `SetByEUI`, since that endpoint is the one that is hit the most (since most JS communication is based on EUIs, not device IDs).

pkg/applicationserver/io/pubsub/redis/registry.go
```diff
@@ -52,7 +56,21 @@ func applyPubSubFieldMask(dst, src *ttnpb.ApplicationPubSub, paths ...string) (*
...
-	Redis *ttnredis.Client
+	Redis   *ttnredis.Client
+	LockTTL time.Duration
+
+	entropyMu *sync.Mutex
+	entropy   io.Reader
+}
+
+// Init initializes the PubSubRegistry.
+func (r *PubSubRegistry) Init(ctx context.Context) error {
```
johanstokking, line 60: Is it not feasible to turn this in a `New()` that returns an initialized `PubSubRegistry`?

pkg/redis/redis.go
```diff
@@ -983,6 +985,18 @@ outer:
 	}
 }
 
+// GenerateLockerID generates a unique locker ID from the provided entropy source.
+// The mutex is acquired and held during entropy extraction.
+func GenerateLockerID(entropy io.Reader, mu *sync.Mutex) (string, error) {
+	mu.Lock()
+	defer mu.Unlock()
+	lockID, err := ulid.New(ulid.Timestamp(time.Now()), entropy)
```
johanstokking: This adds quite some synchronization. Is this faster than using `crypto/rand` and not lock?

pkg/applicationserver/io/pubsub/redis/registry.go
adriansmares: It's slightly awkward because the structs are already exposed, so you can construct the object without `New`. At least `Init` signals that you need to call `Init` first.

pkg/redis/redis.go
adriansmares: This is a carry over from the NS implementation, in order to ensure that we don't accidentally generate the same locker ID for two separate goroutines / processes. I'll think of a benchmark.

Edit: Actually yeah, I think `rand.Reader` is fine.

johanstokking (approved):
What happens if the callback panics that does the actual work? There's no panic recovery in `LockedWatch` – does Redis do the magic?

adriansmares:
> 
> 
> What happens if the callback panics that does the actual work? There's no panic recovery in `LockedWatch` – does Redis do the magic?

The panic is propagated through the stack, while the mutex is unlocked:
https://github.com/TheThingsNetwork/lorawan-stack/blob/8c3891a18b9f42dbadb5763bdda39a1c6c2c067e/pkg/redis/redis.go#L811-L815

htdvisser (approved):
Code-owned part lgtm.

KrishnaIyer (approved):

PR #4736, opened 2021-10-11
Remove embed and nullable from applicationserver_packages.proto

adriansmares:
#### Summary

References #2798 

#### Changes

- Remove the gogoproto `embed` and `nullable` options from `applicationserver_packages.proto`

#### Testing

Unit testing and some local testing.

##### Regressions

This touches most of the storage / RPC of the application packages frontend. Nonetheless, there are no expected regressions.

#### Checklist


pkg/ttnpb/applicationserver_packages.go
```diff
@@ -0,0 +1,141 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ttnpb
+
+// All EntityType methods implement the IDStringer interface.
```
johanstokking, line 17: I missed that this was necessary. What is the reasoning for this?

pkg/ttnpb/applicationserver_packages.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (m *ListApplicationPackageDefaultAssociationRequest) EntityType() string {
+	return m.Ids.EntityType()
+}
+
+func (m *SetApplicationPackageDefaultAssociationRequest) EntityType() string {
+	return m.Default.Ids.EntityType()
+}
+
+// All IDString methods implement the IDStringer interface.
```
johanstokking, line 59: Idem

pkg/ttnpb/applicationserver_packages.go
adriansmares: Rate limitting and resource limiting.

https://github.com/TheThingsNetwork/lorawan-stack/blob/8c3891a18b9f42dbadb5763bdda39a1c6c2c067e/pkg/ratelimit/grpc.go#L39-L50

I considered that it is fine to act as if the application package applies to the end-device / application limits.

johanstokking: OK. So I suppose we also need this in the NS/AS/JS end device registries?

adriansmares: Previously the interface was implemented because the whole hierarchy was flat due to `embed`.

I've implemented it for consistency (requests used to implemented it before), and because maybe we would like to have more aggressive rate limiting in the future (and the keys used for rate limiting are based on this).

cc @htdvisser if you have anything to add

johanstokking: Yes I understand. It's just wasn't part of [the instructions](https://github.com/TheThingsNetwork/lorawan-stack/issues/2798#issuecomment-929102166) so I wasn't aware that was necessary. But sure we need rate limiting on end device registries and other services.

johanstokking (approved):

pkg/ttnpb/applicationserver_packages.go
htdvisser: Ok, I guess I should have linked to the earlier comment https://github.com/TheThingsNetwork/lorawan-stack/issues/2798#issuecomment-916737093. I'll edit "the instructions".

htdvisser (approved):
Looks good to me 👍

Issue #4641, opened 2021-09-20
Console - device/gateway - open in new tab


jpmeijers:
#### Summary

Currently one can't right click or middle click on a device/gateway in the end device and gateway lists, and then choose to open it in a new tab. On the V2 console this is possible.

When debugging devices (normally multiple at the same time), it is useful to have them open in different tabs and keeping it open for a long period to see live data. On the V2 console my workflow was to go to an application, list the devices, and then right click, open in new tab. This is not possible on the V3 console, as the devices in the device list are not `<a>`-links. It means one need to manually open a new tab and manually browse to the other device in the new tab. This becomes tedious if debugging or checking up on many devices/gateways.

#### Why do we need this?

Makes debugging multiple devices/gateways easier.

#### What is already there? What do you see now?

Currently we see a list of device, but can't open then in new tabs.

![Screenshot from 2021-09-20 14-59-14](https://user-images.githubusercontent.com/948278/134006480-00ef790c-9834-4168-b306-99b71c88902a.png)

#### What is missing? What do you want to see?

Add a `<a>`-tag around every device so that my browser will allow me to open a device in a new tab.

![Screenshot from 2021-09-20 15-01-16](https://user-images.githubusercontent.com/948278/134006667-bec00267-970d-4b58-9ee1-00cd9261841f.png)

#### Environment

Ubuntu 20.04
Google Chrome Version 93.0.4577.63 (Official Build) (64-bit)
TTS CE v3.15.0

#### How do you propose to implement this?

Add `<a>`-tag around a device.

#### How do you propose to test this?

Open console, go to list of end devices or gateways, right click and see if the open in new tab option is available.

#### Can you do this yourself and submit a Pull Request?

No

kschiffer:
Thanks for reporting this @jpmeijers. This is indeed a bug and needs to be fixed.

@bafonins can you take a look? We'd have to extend the `<Table />` component to render plain `<a />` within the table row instead of using a click handler. We'll probably want to retain both modes of interaction though (either/or), as I could imagine there are situations where the click handler is still necessary, but please check that. Likely a `basePath` prop will be enough so that the `href` can be generated via `${basePath}/${id}`. We are sort of doing that composition already but on the `<FetchTable />` level.

kschiffer:
Using the middle mouse (or function key) to open the entity in a new row has been added in 3.15.2. For the remaining issue of refactoring the table component to use a `div` layout, see #4787

[...]

PR #4749, opened 2021-10-15
Allow opening table entries in new tab in the Console

bafonins:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/4641
Listen for `metaKey` when clicking on table rows.

#### Changes

- Listen for `metaKey` and open a new tab when necessary
- Listen for middle clicks and open a new tab

#### Testing

Manual in chrome/firefox/safari

#### Notes for Reviewers

Unfortunately it is not possible to render `<a />` tag within `<table />` element according to html content model. IMO, this is the best we can do to allow easier debugging of multiple entities.

Another solution would be to change semantics and add a new `<List />` component and render it instead of `<Table />`.

I will follow up with full description of the problem in the issue.

#### Checklist

### DIFF
--- a/pkg/webui/containers/fetch-table/index.js
+++ b/pkg/webui/containers/fetch-table/index.js
@@ -28,6 +28,7 @@ import Tabs from '@ttn-lw/components/tabs'
 import Overlay from '@ttn-lw/components/overlay'
 import ErrorNotification from '@ttn-lw/components/error-notification'
 
+import { selectApplicationRootPath } from '@ttn-lw/lib/selectors/env'
 import debounce from '@ttn-lw/lib/debounce'
 import PropTypes from '@ttn-lw/lib/prop-types'
 import sharedMessages from '@ttn-lw/lib/shared-messages'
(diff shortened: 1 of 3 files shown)


kschiffer (changes requested):
I get the reasoning but this does not close #4641 then because it does still not allow to use middle click or right click -> Open in new tab to open the page in a new tab, as described in the issue.

I think ultimately we can't get around letting go of the table model and using actual a tags. But as we discussed, this is a way larger refactor that we cannot do short term.

I'm still fine with merging this since it is an improvement, but since we're already using workarounds, could you also add functionality to open in a new tab on middle-click programmatically?

pkg/webui/containers/fetch-table/index.js
```diff
@@ -289,7 +294,11 @@ class FetchTable extends Component {
       entityPath = `${itemPathPrefix}/${item_id}`
     }
 
-    dispatch(push(`${pathname}${entityPath}`))
+    if (evt.metaKey || evt.button === 1) {
```
kschiffer, line 299: You have to change the handler to `onMouseDown` to make the handler trigger on middle clicks as well.

kschiffer (changes requested):

kschiffer (approved):

PR #4752, opened 2021-10-18
Add content-security policy to console,account,oauth

### LINKED ISSUE #4212: Add Content-Security-Policy header to HTML pages
#### Summary

I think we should add a `Content-Security-Policy` header to the HTML pages of the Account app and the Console.

- `default-src 'self' assets.cloud.thethings.industries [...];` (constructed from `AssetsBaseURL` and `BrandingBaseURL`)
- `connect-src 'self' eu1.cloud.thethings.industries [...] *.ingest.sentry.io;` (constructed from the API `BaseURL`s)
- `base-uri 'self';`
- `frame-ancestors 'none';`

We shouldn't need `unsafe-inline` and `unsafe-eval`, so if it doesn't already work without those, we should probably try and make it work.

We also need to allow outgoing calls

It should be possible to test this locally if you configure `AssetsBaseURL` and `BrandingBaseURL`.

pgalic96:
#### Summary

Closes #4212 

#### Changes

- Add middleware that adds `Content-Security-Policy` header
- Generate and pass nonce to the application

#### Testing

Manual

##### Regressions

If CSP is set too strict, our scripts and assets won't load which will break UI or functionality.

#### Notes for Reviewers

As discussed with some of you, in the bundled console/account file there are many expressions `Function("return this")` which need `unsafe-eval` policy in `default-src` or `script-src` to be set to work. The `unsafe-eval` is not recommended but I don't think there are immediate security concerns regarding this. However we should figure out why are these expressions necessary and how can we replace them. I will file an issue for this.

#### Checklist

### DIFF
--- a/pkg/account/server.go
+++ b/pkg/account/server.go
@@ -49,6 +49,7 @@ type server struct {
 	config  oauth.Config
 	store   Store
 	session sess.Session
+	csp     map[string][]string
 }
 
 // Store used by the account app.
@@ -60,12 +61,13 @@ type Store interface {
 }
 
 // NewServer returns a new account app on top of the given store.
-func NewServer(c Component, store Store, config oauth.Config) Server {
+func NewServer(c Component, store Store, config oauth.Config, csp map[string][]string) Server {
 	s := &server{
 		c:       c,
 		config:  config,
 		store:   store,
 		session: sess.Session{Store: store},
+		csp:     csp,
 	}
 
 	if s.config.Mount == "" {
@@ -94,6 +96,15 @@ func (s *server) RegisterRoutes(server *web.Server) {
 	csrfMiddleware := middleware.CSRF("_csrf", "/", s.config.CSRFAuthKey)
 	root := server.Group(
 		s.config.Mount,
+		func(next echo.HandlerFunc) echo.HandlerFunc {
+			return func(c echo.Context) error {
+				nonce := webui.GenerateNonce()
+				c.Set("csp_nonce", nonce)
+				cspString := webui.GenerateCSPString(s.csp, nonce)
+				c.Response().Header().Set("Content-Security-Policy", cspString)
+				return next(c)
+			}
+		},
 		ratelimit.EchoMiddleware(s.c.RateLimiter(), "http:account"),
 		func(next echo.HandlerFunc) echo.HandlerFunc {
 			return func(c echo.Context) error {
--- a/pkg/console/console.go
+++ b/pkg/console/console.go
@@ -1,4 +1,4 @@
-// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
+// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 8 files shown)


pkg/account/server.go
```diff
@@ -66,6 +67,25 @@ func NewServer(c Component, store Store, config oauth.Config) Server {
 		config:  config,
 		store:   store,
 		session: sess.Session{Store: store},
+		csp: map[string][]string{
+			"default-src": {
+				"'self'",
+				config.UI.AssetsBaseURL,
+				config.UI.BrandingBaseURL,
```
htdvisser: Should we de-duplicate these if they're the same, and omit them if empty?

pkg/console/console.go
```diff
@@ -48,6 +49,32 @@ func New(c *component.Component, config Config) (*Console, error) {
 		Component: c,
 		oc:        oc,
 		config:    config,
+		csp: map[string][]string{
+			"default-src": {
+				"'self'",
+				config.UI.AssetsBaseURL,
+				config.UI.BrandingBaseURL,
```
htdvisser, line 56: Same. Should we de-duplicate these if they're the same, and omit them if empty?

pkg/console/console.go
```diff
@@ -48,6 +49,32 @@ func New(c *component.Component, config Config) (*Console, error) {
...
+			"connect-src": {
+				"'self'",
+				config.UI.StackConfig.GS.BaseURL,
+				config.UI.StackConfig.IS.BaseURL,
+				config.UI.StackConfig.JS.BaseURL,
+				config.UI.StackConfig.NS.BaseURL,
+				config.UI.StackConfig.AS.BaseURL,
+				config.UI.StackConfig.EDTC.BaseURL,
+				config.UI.StackConfig.QRG.BaseURL,
+				config.UI.StackConfig.GCS.BaseURL,
```
htdvisser, line 68: Same

pkg/webui/csp.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+	"strings"
+)
+
+// GenNonce returns a nonce used for inline scripts
+func GenerateNonce() string {
```
htdvisser: Godoc "GenNonce" does not match func name "GenerateNonce"

pkg/webui/csp.go
```diff
@@ -0,0 +1,42 @@
...
+	var b [20]byte
+	if _, err := rand.Read(b[:]); err != nil {
+		panic(err)
+	}
+	return base64.StdEncoding.EncodeToString(b[:])
+}
+
+func GenerateCSPString(csp map[string][]string, nonce string) string {
```
htdvisser, line 50: Missing Godoc

pgalic96:
We need a CSP for `oauth` app as well since it is using the same template as `console` and `account`. Since `account` and `oauth` share the config they can share the csp as well.

htdvisser (approved):

kschiffer (approved):
LGTM. Please test this thoroughly once deployed on staging.

PR #4754, opened 2021-10-19
Allow toggling visibility of inputs with sensitive data

kschiffer:
#### Summary
This PR extends our input component with the ability to toggle visibility of sensitive values. It also applies the new `sensitive` prop to respective fields throughout our forms.

image
image
image

cc @adriansmares 

#### Changes
- Add `text-security-disc` fonts
- Extend `<Input />` component to enable hiding values when using the `sensitive` prop
  - This includes the `type="byte"` component variant
- Add `sensitive` prop to fields that hold sensitive data
- Some related styling fixes

#### Testing

Manual, also using storybook.

#### Notes for Reviewers
Simply using `type="password"` is not possible due to annoying side effects of browsers that will regard the input values as passwords and suggest generating or storing their values as passwords that are impossible to reliably supress. The only viable alternative I could find is using a [special font](https://github.com/noppa/text-security) that renders all characters as a dot. This also allows us some nice ways to keep the masking functionality for byte inputs when hiding the value, by providing different characters for the placeholder and the dot.

#### Checklist


pkg/webui/components/input/index.js
```diff
@@ -150,38 +178,39 @@ class Input extends React.Component {
...
       showPerChar,
-      forwardedRef,
-      inputRef,
-      actionDisable,
-      children,
+      title,
+      type,
+      valid,
+      value,
+      warning,
```
kschiffer, line 205: This diff comes from sorting the props alphabetically.

pkg/webui/components/input/index.js
```diff
@@ -223,17 +252,18 @@ class Input extends React.Component {
 
     const passedProps = {
       ...rest,
-      ...(type === 'byte' ? { showPerChar } : {}),
       ref: inputRef ? combineRefs([this.input, inputRef]) : this.input,
     }
 
+    const inputProps = classnames(style.input, { [style.hidden]: hidden })
```
bafonins: `inputCls` ?

pkg/webui/components/input/index.js
```diff
@@ -109,27 +119,45 @@ class Input extends React.Component {
...
+    }
+
+    return `${width}rem`
+  }
+
+  @bind
+  handleHideToggleClick() {
+    const { hidden } = this.state
+
+    this.setState({
```
bafonins:
```suggestion
    this.setState(({ hidden }) => ({ hidden: !hidden }))
```

bafonins (commented):
Looks good! 

One problem I noticed is that with with `join_eui` field. It is now possible to go beyong 16 characters up to 32. Could you check this?

kschiffer:
> One problem I noticed is that with with `join_eui` field. It is now possible to go beyong 16 characters up to 32. Could you check this?

Quite a catch! Was due to an erroneously removed line. Fixed now.

bafonins (approved):

PR #4779, opened 2021-10-25
Fix antenna comparison

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/4674

#### Changes

- Use `[]*ttnpb.GatewayAntenna` everywhere instead of moving to `[]ttnpb.GatewayAntenna`
- Fix `sameAntennaLocations` comparison when more than 1 antenna is present

#### Testing

Local testing

##### Regressions

This is already broken.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -854,14 +854,13 @@ func sameAntennaLocations(a, b []*ttnpb.GatewayAntenna) bool {
 	if len(a) != len(b) {
 		return false
 	}
-	for _, a := range a {
-		for _, b := range b {
-			if a.Location != nil && b.Location != nil && !sameLocation(*a.Location, *b.Location) {
-				return false
-			}
-			if (a.Location == nil) != (b.Location == nil) {
-				return false
-			}
+	for i := range a {
+		a, b := a[i], b[i]
+		if a.Location != nil && b.Location != nil && !sameLocation(*a.Location, *b.Location) {
+			return false
+		}
+		if (a.Location == nil) != (b.Location == nil) {
+			return false
 		}
 	}
 	return true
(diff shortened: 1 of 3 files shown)


pkg/gatewayserver/gatewayserver.go
```diff
@@ -547,14 +547,7 @@ func (gs *GatewayServer) GetConnection(ctx context.Context, ids ttnpb.GatewayIde
 }
 
 func requireDisconnect(connected, current *ttnpb.Gateway) bool {
-	fromPtrArray := func(in []*ttnpb.GatewayAntenna) []ttnpb.GatewayAntenna {
-		ret := make([]ttnpb.GatewayAntenna, len(in))
```
adriansmares, line 551: Since we do `make(type, len)`, the slice is already initialized, so we append after these 0-initialized structs. This causes the antenna comparison to fail

KrishnaIyer (approved):

pkg/gatewayserver/gatewayserver.go
```diff
@@ -857,18 +850,17 @@ func sameLocation(a, b ttnpb.Location) bool {
 		math.Abs(a.Longitude-b.Longitude) <= allowedLocationDelta
 }
 
-func sameAntennaLocations(a, b []ttnpb.GatewayAntenna) bool {
+func sameAntennaLocations(a, b []*ttnpb.GatewayAntenna) bool {
 	if len(a) != len(b) {
 		return false
 	}
-	for _, a := range a {
```
adriansmares, line 864: For `a=[a1, a2]` and `b=[a1,a2]` (identical), `a=a2` will be compared with `b=a1` and fail, even though the two slices are identical.

Normally `fromPtrArray` should have been a noop from a comparison perspective (as you add these empty antennas to both `a` and `b`), but since the comparison was bugged too, it actually caused errors. One bug has revealed another :joy: .

johanstokking (approved):

PR #4780, opened 2021-10-25
NS and JS component cleanup

### LINKED ISSUE #14: Clean up component registries when applications got deleted

pgalic96:
#### Summary

Closes #14 

#### Changes

- Add `Range` to NS and JS registries
- Add `ns-db cleanup` and `js-db cleanup`

#### Testing

Manual (not sure if tests are necessary, pretty much copy paste of AS cleanup tests)

##### Regressions

New functionality

#### Checklist

### DIFF
--- a/cmd/ttn-lw-stack/commands/js_db.go
+++ b/cmd/ttn-lw-stack/commands/js_db.go
@@ -1,0 +1,138 @@
+// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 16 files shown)


htdvisser (commented):
Code-owned changes LGTM.

I do think it would be good to cover this with tests. When we decide to refactor these stores (we may migrate to a different storage backend than Redis) we want to be confident that we're not breaking anything essential.

adriansmares (approved):
LGTM so far. Looking forward to the tests.

pgalic96:
Tests added. We should also probably start thinking about when/how we want to run this.

htdvisser (approved):
Nothing to add

adriansmares (approved):

Issue #3321, opened 2020-10-06
Use links in table rows in the Console


bafonins:
#### Summary

Currently, table rows are just `tr` elements with `onlick` listener.

#### Why do we need this?

In certain cases I want to open multiple items in new browser tabs. Links allow doing this by holding the commands/control key and clicking on a link. 

Consider:
1. Opening the device overview/events page while still staying at the application overview page (looking at events, quickly editing an end device)
2. Opening multiple device/gateway overview pages

#### What is already there? What do you see now?

Each row of the table component is just a `tr` element with `onclick` listener that changes the url via `connected-react-router` actions.

#### What is missing? What do you want to see?

Regular `react-router` links in table rows.

#### How do you propose to implement this?

tba

#### How do you propose to test this?

manual testing

#### Can you do this yourself and submit a Pull Request?

yes

bafonins:
Closing this. Will keep track of the issue in https://github.com/TheThingsNetwork/lorawan-stack/issues/4641

PR #4789, opened 2021-10-26
Refactor `<Table />` component to allow for `<a />` wrapped rows

### LINKED ISSUE #4641: Console - device/gateway - open in new tab

### LINKED ISSUE #4784: Buttons on the gateway table cannot be pressed
#### Summary

Clicking on the buttons in the gateways table opens the overview page for the gateway. I've tested this with the EUI copy button, and the admin purge / restore. 

#### Steps to Reproduce

1. Go to the gateways table in the Console.
2. Attempt to click on a gateway related button (let's say EUI)

#### What do you see now?

3. The action is not taken, and you are redirected to the gateway overview page

4. The button action is done (i.e. the EUI is copied)

#### Environment

`3.15.2` on `staging1`
Firefox Developer 94.0b9

I don't know actually.

Check if the buttons work again. Also check other tables (applications seems to have the same behavior).

Can review.

### LINKED ISSUE #4787: Refactor table component to enable usage of `<a />` tags
#### Summary
We currently use a `<table />`-based layout for our `<Table />` component. While that seems to be semantically correct, it brings some issues that are specific to how the table model works in HTML. We should hence refactor the table component to use a regular `<div />` layout and reproduce table cells using flex-box.

#### Why do we need this?
- To use proper `</a>` tags for table rows instead of recreating link functionality programmatically. See also https://github.com/TheThingsNetwork/lorawan-stack/issues/4641

- `<Table />` component using `<table />` markup

- `<Table />` component using `<div />` layout

Changing the layout itself shouldn't be a big issue in itself. Recreating a similar visual layout might be a bit trickier, but I think it should be possible using flexbox.

Existing end-to-end test should suffice.

@bafonins

kschiffer:
#### Summary
This PR changes the rendered markup of the `<Table />` component to use no more `<table />` elements (`td`, `tr`, `th`, etc.). It also adds functionality to wrap rows around `<a />` tags so that we can avoid recreating native link functionality.

Closes #4641 
Closes #4787
Closes #4784
References #3321 

#### Changes
- Remove any rendering of table based HTML elements from the `<Table />` component
- Add styling to recreate table layout on replaced `<div />`'s
- Add `rowHrefSelector` prop to calculate the row link based on the row's item
- Refactor `<FetchTable />` to use `rowHrefSelector` instead of `onRowClick`

#### Testing

Manually but also existing end-to-end tests should catch any significant issue.

##### Regressions

I could imagine that this might cause display issues caused by discrepancies of how browsers deal with `display: table-*` props. I haven't noticed any issues though.

#### Notes for Reviewers
I did not plan to work on this but it escalated from fixing a bug introduced via #4749 as described [here](https://github.com/TheThingsNetwork/lorawan-stack/issues/4641#issuecomment-951882025)

#### Checklist


bafonins (approved):
Please take a look at https://github.com/TheThingsNetwork/lorawan-stack/issues/4784 this issue was introduced with https://github.com/TheThingsNetwork/lorawan-stack/pull/4749 however it is still present with these changes as well.

.github/workflows/e2e.yml
```diff
@@ -81,7 +81,7 @@ jobs:
           path: |
             .cache/sqldump.sql
             .env/admin_api_key.txt
-          key: db-cache-${{ hashFiles('pkg/identityserver/store/**/*.go', 'cmd/ttn-lw-stack/commands/is-db.go') }}
+          key: -db-cache-${{ hashFiles('pkg/identityserver/store/**/*.go', 'cmd/ttn-lw-stack/commands/is-db.go') }}
```
htdvisser: Why?

kschiffer: Please disregard. This was testing only.

Issue #4715, opened 2021-10-06
Buffer full warning while under load


virtualguy:
#### Summary
When under significant load the application server reports dropping uplink packets as the buffer is full. 

#### Steps to Reproduce
1. Send lots of uplinks using the simulator OR subscribe to mqtt on a throttled connection

#### What do you see now?
```
WARN	Failed to publish message	{"device_uid": "halter.000-load49-00377", "error": "error:pkg/applicationserver/io:buffer_full (buffer is full)", "grpc.method": "HandleUplink", "grpc.service": "ttn.lorawan.v3.NsAs", "namespace": "applicationserver/distribution", "protocol": "applicationpackages", "request_id": "01FHAKXVRR3NECCP0D75DNBTHQ"}
```
  
#### What do you want to see instead?
No failed publish logs
...

#### Environment

TTS v3.14.2

#### How do you propose to implement this?

Is there any ability to tune or increase the buffers on mqtt?

...

#### How do you propose to test this?

Happy to run test branches on our load simulator

#### Can you do this yourself and submit a Pull Request?
Need guidance

virtualguy:
I wonder if the issue is the underlying MQTT server, perhaps the bigger size needs to be larger/configurable to handle higher packet volumes or high latency between AS and the subscribed client. 

See: https://github.com/TheThingsIndustries/mystique/blob/24778eddf8e34b7da1abc83bc14011285cbb4a2b/pkg/session/session.go#L24

adriansmares:
The `"protocol": "applicationpackages"` related `buffer_full` errors should have been reduced heavily during `3.15.0`/`3.15.1` - we've had quite a lot of related work such as https://github.com/TheThingsNetwork/lorawan-stack/pull/4607, https://github.com/TheThingsNetwork/lorawan-stack/pull/4609, https://github.com/TheThingsNetwork/lorawan-stack/pull/4680. You will still probably want to tune up the number of workers, and please let me know if you still see them occur (they will be called `pool_full` now, and you can track the pools using [metrics](https://www.thethingsindustries.com/docs/reference/telemetry/#general-metrics))

On the `"protocol":"mqtt"` side I would just recommend against using MQTT with high traffic volumes in general. More buffering won't make the connection faster if the subscriber cannot keep up with the publisher.

virtualguy:
[...]

PR #4790, opened 2021-10-26
Improve Application Server distribution observability

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/4715

#### Changes

- Change `TestSkipPayloadCrypto` to sleep _before_ starting to publish traffic, in order to allow the subscriber to actually subscribe. Previously no matter how much we would sleep the message would be lost since it would be published before the subscriber subscribed.
- Increase the default number of webhook, application packages and UDP workers to 1024
  - The workers are part of a dynamic worker pool, and are spawned on demand
- Increase the default maximum number of workers of a worker pool to 1024
  - This affects two pools currently: `webhooks_fanout` and `application_packages_fanout`, which we may move to a specific configuration option later
- Add metrics for subscription sets publishing status (success / failure)
  - The idea here is to observe basically how often `buffer_full` errors are observed, as failures can occur only on `io.Subscription.Publish`
- Allow advanced configuration of the Application Server traffic distribution
  - For each distribution level, allow the network operators to configure the buffer sizes, and if publishing should block
  - If publish blocks, the buffers no longer drop messages that arrive when the buffers are full
- The Application Server Global (PubSub based) distributor no longer buffers messages, and instead allows the underlying subscription set / subscription to do it

#### Testing

`GOMAXPROCS=1 TEST_SLOWDOWN=8 go test -v -failfast -timeout=1h -count=1000 -run TestSkipPayloadCrypto` _many_ times

##### Regressions

This mainly touches piping. If CI is happy, everything should be fine.
#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,19 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Added
 
+- `ttn_lw_as_subscription_sets_publish_success_total` and `ttn_lw_as_subscription_sets_publish_failed_total` metrics to track the number of subscription set publish attempts.
+- Application Server advanced distribution settings:
+  - `as.distribution.global.individual.subscription-blocks` controls if the Application Server should block while publishing traffic to individual global subscribers (such as MQTT clients).
+  - `as.distribution.global.individual.subscription-queue-size` controls how many uplinks the Application Server should buffer for an individual global subscriber. Note that when the buffer is full, the Application Server will drop the uplinks if `--as.distribution.global.individual.subscription-blocks` is not enabled. Use a negative value in order to disable the queue.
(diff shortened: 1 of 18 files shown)


KrishnaIyer (approved):
LGTM

> Change TestSkipPayloadCrypto to sleep before starting to publish traffic, in order to allow the subscriber to actually subscribe. Previously no matter how much we would sleep the message would be lost since it would be published before the subscriber subscribed.

Hopefully that fixes the CI flake.

htdvisser (approved):
Code-owned changes LGTM.

Please don't forget to document telemetry in https://github.com/TheThingsIndustries/lorawan-stack-docs/blob/master/doc/content/reference/telemetry/_index.md

Issue #4766, opened 2021-10-20
Merge session fields correctly


kschiffer:
#### Summary
When creating an ABP end device, the `session.started_at` property in the AS is set to `0001-01-01T00:00:00Z`. This is not a helpful value.

#### Steps to Reproduce

1. Go to the console
2. Using the manual creation, create an end device
3. Observe network tab of the browser, observe the response for the AS query of the end device
4. See that the `session.started_at` value is set to `0001-01-01T00:00:00Z`

#### What do you see now?
`started_at` value being stored as `0001-01-01T00:00:00Z`

#### What do you want to see instead?
`started_at` value being either omitted, or stored using the current timestamp of the creation time.

#### How do you propose to implement this?
Trivial.

#### How do you propose to test this?
I can test using the Console.

#### Can you do this yourself and submit a Pull Request?
@johanstokking or @adriansmares please coordinate and reassign accordingly.

johanstokking:
`started_at` is currently non-nullable (refs #2798), so we can't really avoid returning an empty time. What we can do, however, is fix that AS sets `started_at` and returns a non-empty value for `started_at`.

According to API specification, `started_at` should only be observed from Network Server.

johanstokking:
@kschiffer the best approach is to merge the session in a smarter way.

NS is in charge of lifetime (so `started_at`), `dev_addr`, `keys.{f,s}_nwk_s_int_key`, `keys.nwk_s_enc_key`, `last_f_cnt_up`, `last_n_f_cnt_down`, `last_conf_f_cnt_down`

AS is in charge of `keys.app_s_key`, `last_a_f_cnt_down` and you can best fetch `queued_application_downlinks` from AS.

The fix in #4791 only makes sure that AS never returns a non-empty value for `session.started_at` and `pending_session.started_at`.

johanstokking:
I think this is resolved.

PR #4791, opened 2021-10-27
Unset session `started_at` in Application Server

johanstokking:
#### Summary

Unset session `started_at`

Addresses #4766

#### Changes

- Do not set session `started_at` in AS
- Unset `started_at` as part of `as-db cleanup`

#### Testing

1. Made local device
2. Temporarily allowed setting `session.started_at` in AS (by allowing field mask validation, not part of this PR)
3. Set the started at to a value: 
```
TTN_LW_NETWORK_SERVER_ENABLED=false ttn-lw-cli dev set app1 dev1 --session.started_at "2021-01-01T12:00:00Z" --session.dev_addr ABCDABCD
```
4. Decode the value from the database:
```json
{"ids":{"device_id":"dev1","application_ids":{"application_id":"app1"},"dev_eui":"1122334455667788","join_eui":"0000000000000000","dev_addr":"ABCDABCD"},"created_at":"2021-10-27T08:34:35.761063Z","updated_at":"2021-10-27T08:37:21.740897Z","session":{"dev_addr":"ABCDABCD","keys":{},"started_at":"2021-01-01T12:00:00Z"}}
```
5. Run `as-db cleanup`
6. Decode the value from the database:
```json
{"ids":{"device_id":"dev1","application_ids":{"application_id":"app1"},"dev_eui":"1122334455667788","join_eui":"0000000000000000","dev_addr":"ABCDABCD"},"created_at":"2021-10-27T08:34:35.761063Z","updated_at":"2021-10-27T08:44:47.863279Z","session":{"dev_addr":"ABCDABCD","keys":{},"started_at":"0001-01-01T00:00:00Z"}}
```
Notice that it's reset to an empty value, the best we can do now

##### Regressions

None, unless clients were relying on faulty behavior

#### Notes for Reviewers

@kschiffer I'm leaving the target issue open and will provide some additional information on what the Console should do when merging sessions.

#### Checklist

### DIFF
--- a/cmd/ttn-lw-stack/commands/as_db.go
+++ b/cmd/ttn-lw-stack/commands/as_db.go
@@ -59,11 +59,12 @@ var (
 			defer func() { logger.Debugf("%d keys migrated", migrated) }()
 
 			const (
-				idRegexpStr  = `([a-z0-9](?:[-]?[a-z0-9]){2,}){1,36}?`
-				uidRegexpStr = idRegexpStr + `\.` + idRegexpStr
+				idRegexpStr        = `([a-z0-9](?:[-]?[a-z0-9]){2,}){1,36}?`
+				uidRegexpStr       = idRegexpStr
+				deviceUIDRegexpStr = idRegexpStr + `\.` + uidRegexpStr
 			)
 
-			uidRegexp := regexp.MustCompile(cl.Key("devices", "uid", uidRegexpStr+"$"))
+			deviceUIDRegexp := regexp.MustCompile(cl.Key("devices", "uid", deviceUIDRegexpStr+"$"))
 
 			lockerID, err := ttnredis.GenerateLockerID()
 			if err != nil {
@@ -72,7 +73,7 @@ var (
 			if err := ttnredis.RangeRedisKeys(ctx, cl, cl.Key("*"), 1, func(k string) (bool, error) {
 				logger := logger.WithField("key", k)
 				switch {
-				case uidRegexp.MatchString(k):
+				case deviceUIDRegexp.MatchString(k):
 					if err := ttnredis.LockedWatch(ctx, cl, k, lockerID, defaultLockTTL, func(tx *redis.Tx) error {
 						dev := &ttnpb.EndDevice{}
 						if err := ttnredis.GetProto(ctx, tx, k).ScanProto(dev); err != nil {
@@ -87,11 +88,11 @@ var (
 							sess.StartedAt = time.Time{}
 							any = true
 						}
-						_, err := ttnredis.SetProto(ctx, tx, k, dev, 0)
-						if err != nil {
-							return err
-						}
 						if any {
+							_, err := ttnredis.SetProto(ctx, tx, k, dev, 0)
+							if err != nil {
+								return err
+							}
 							migrated++
 						}
 						return nil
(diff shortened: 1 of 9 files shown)


adriansmares (approved):

pgalic96 (commented):
The changes LGTM. However, I assume we will need to run this only once and then there will be no non-zero `started_at` values anymore in the registry (might be a wrong assumption). Then it shouldn't be necessary to run this periodically with the rest of the cleanup, and maybe we can hide it behind a flag (`--unset-dev-fields` or something like that).

adriansmares:
> The changes LGTM. However, I assume we will need to run this only once and then there will be no non-zero `started_at` values anymore in the registry (might be a wrong assumption). Then it shouldn't be necessary to run this periodically with the rest of the cleanup, and maybe we can hide it behind a flag (`--unset-dev-fields` or something like that).

Or as part of a migration, albeit I do agree that we need to this would be longer if we do that.

johanstokking:
> The changes LGTM. However, I assume we will need to run this only once and then there will be no non-zero `started_at` values anymore in the registry (might be a wrong assumption). Then it shouldn't be necessary to run this periodically with the rest of the cleanup, and maybe we can hide it behind a flag (`--unset-dev-fields` or something like that).

Best would be to have some sort of migration version of the AS end device registry.

@adriansmares don't we have something like this for NS?

adriansmares:
> @adriansmares don't we have something like this for NS?

We do:
https://github.com/TheThingsNetwork/lorawan-stack/blob/8edf8c81c5e8e771e778b67fb8acf4696ee0205f/cmd/ttn-lw-stack/commands/ns_db.go#L468-L491

What needs to happen for that to be useful:
- Bring `as-db migrate` from TTI, make the migration there have version 1
- Mark this cleanup as version 2
- If the recorded version is lower than `asredis.SchemaVersion`, run the missing migrations in order

htdvisser (commented):
Nothing to add here. I agree that this belongs in a migration, and not in the periodic cleanup command.

johanstokking:
@adriansmares I replaced the cleanup logic by the `as-db migrate` command that looks a lot like NS's. I updated the original comment and verified that the migration produces the desired outcome.

Output:

```
$ go run ./cmd/ttn-lw-stack as-db migrate
INFO	Connecting to Redis database...
INFO	Setting schema version	{"version": 1}
DEBUG	1 keys migrated
DEBUG	No new version available
```

And again:

```
$ go run ./cmd/ttn-lw-stack as-db migrate
INFO	Connecting to Redis database...
INFO	Existing database schema version	{"version": 1}
INFO	Database schema version is already in latest version
DEBUG	No new version available
```

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -29,6 +33,76 @@ var (
...
+				uidRegexpStr = idRegexpStr + `\.` + idRegexpStr
+			)
+
+			uidRegexp := regexp.MustCompile(cl.Key("uid", uidRegexpStr+"$"))
+
+			if err := ttnredis.RangeRedisKeys(ctx, cl, cl.Key("*"), 1, func(k string) (bool, error) {
+				logger := logger.WithField("key", k)
+				switch {
+				case uidRegexp.MatchString(k):
+					if err := cl.Watch(ctx, func(tx *redis.Tx) error {
```
adriansmares: I'd do a `LockedWatch` here, otherwise we race with the real Application Server instances.

adriansmares (approved):

PR #4795, opened 2021-10-27
Add additional Gateway Server transmission metrics

adriansmares:
#### Summary

This short PR fixes a number of issues with GS traffic metrics, and adds extra metrics.

#### Changes

- Fix the events and metrics related to dropped uplinks when the upstream handlers are full
  - The type of the message is `GatewayUplinkMessage`, not `UplinkMessage`
- Dropped (rejected by upstream handlers) uplink messages and status messages now have the error cause attached to the metric
  - It's only one error for now, don't be afraid of high cardinality
- Remove the 'failed' events/metrics related to uplinks/status messages
  - These events were used when the upstream handlers were full
  - We now have a reason (error) attached to the dropped metrics, which should suffice
  - The old equilibrium equation was `received = forwarded + dropped + failed`, now it is `received = forwarded + dropped`
- Add transmission acknowledgements received/forwarded/dropped metrics/events

#### Testing

:monkey: 

##### Regressions

It was already broken. We can only go up from here.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -701,7 +701,9 @@ func (host *upstreamHost) handlePacket(ctx context.Context, item interface{}) {
 		}
 	case *ttnpb.TxAcknowledgment:
 		if err := host.handler.HandleTxAck(ctx, *gtw.Ids, msg); err != nil {
-			logger.WithField("host", host.name).WithError(err).Debug("Drop Tx acknowledgment")
+			registerDropTxAck(ctx, gtw, msg, host.name, err)
+		} else {
+			registerForwardTxAck(ctx, gtw, msg, host.name)
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


pkg/gatewayserver/gatewayserver.go
```diff
@@ -789,10 +792,14 @@ func (gs *GatewayServer) handleUpstream(conn connectionEntry) {
 			}
 			logger.WithField("name", host.name).WithError(err).Warn("Upstream handler publish failed")
 			switch msg := val.(type) {
-			case *ttnpb.UplinkMessage:
-				registerFailUplink(ctx, gtw, msg, host.name)
+			case *ttnpb.GatewayUplinkMessage:
```
adriansmares, line 795: I was quite dumbfounded that this metric was empty in production, even though the associated pool for the `ttnv2` upstream did mention that traffic was dropped. The type of `msg` was wrong so this path was never taken.

KrishnaIyer (approved):

PR #4798, opened 2021-10-28
Remove gogoproto options

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2798

#### Changes

- Remove `message_services.proto`, as these messages / service have been absorbed into `applicationserver.proto` during implementation
- Make the interface implementations (`IDStringer` etc.) use getters for consistency for some old migrated proto files
- Remove `embed` and `nullable` from `events.proto`, `gateway.proto`

#### Testing

CI.

##### Regressions

Changes are quite minimal, none are expected.

#### Checklist


johanstokking (approved):

pkg/events/events.go
```diff
@@ -38,7 +38,7 @@ type Event interface {
 	UniqueID() string
 	Context() context.Context
 	Name() string
-	Time() time.Time
+	Time() *time.Time
```
htdvisser: I don't think I agree with this change. Our internal (non-proto) events can still keep a non-pointer time, right?

htdvisser (approved):
LGTM with a small comment

pkg/events/events.go
adriansmares: Since this is marked as required at API level, I'll revert this.

PR #4799, opened 2021-10-28
Parse frame payload only once

adriansmares:
#### Summary

This short PR changes the Gateway Server to parse the frame payload only once in order to build the end device identifiers

#### Changes

- Extract the device identifiers from the parsed payload
  - We are always parsing the payload, but previously if parsing failed we would just let the message reach the backends (NS/PBA/`ttnv2`) where it would be dropped because the parsing would fail
- Use component tasks for gateway tasks (connection stats updates, location updates etc.)
  - Mainly in order to have a recovery mechanism in place, for panics

#### Testing

Unit testing.

##### Regressions

This is just plumbing.

#### Notes for Reviewers

This optimization should lower our load, since in production we have a pretty large amount of non-LoRaWAN frames that we receive, and parse at least 3 times for no reason.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -651,7 +651,7 @@ func (host *upstreamHost) handlePacket(ctx context.Context, item interface{}) {
 		}
 		msg.CorrelationIds = append(make([]string, 0, len(msg.CorrelationIds)+1), msg.CorrelationIds...)
 		msg.CorrelationIds = append(msg.CorrelationIds, host.correlationID)
-		drop := func(ids ttnpb.EndDeviceIdentifiers, err error) {
+		drop := func(ids *ttnpb.EndDeviceIdentifiers, err error) {
 			logger := logger.WithError(err)
 			if ids.JoinEui != nil {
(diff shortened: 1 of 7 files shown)


htdvisser (approved):
LGTM

pkg/gatewayserver/gatewayserver.go
```diff
@@ -1005,3 +1045,18 @@ func (gs *GatewayServer) GetMQTTConfig(ctx context.Context) (*config.MQTT, error
 	}
 	return &config.MQTT, nil
 }
+
+type endDeviceIdentifiersKeyType struct{}
+
+var endDeviceIdentifiersKey = endDeviceIdentifiersKeyType{}
```
KrishnaIyer: Can you put these at the top of the file for consistency? You could also move it to a separate file if you want.

KrishnaIyer (approved):
Looks fine to me.

pkg/gatewayserver/gatewayserver.go
adriansmares: I'll even remove them. I just realized that I can extract them on the spot, for free.

PR #4800, opened 2021-10-29
all: Remove environmental variables from Dockerfile

michalborkowski96:
#### Summary
Removes environmental variables from the Dockerfile.

Regarding these:

```
TTN_LW_BLOB_LOCAL_DIRECTORY
TTN_LW_IS_DATABASE_URI
TTN_LW_REDIS_ADDRESS
```

I see that we also set them in docker-compose, so no further action needed.

Regarding this:

```
TTN_LW_HEALTHCHECK_URL
```

Looks like this isn't an actual config variable, per this:

```
$ go run main.go --help | grep -i health
      --http.health.enable                                                Enable health check endpoint on HTTP server (default true)
      --http.health.password string                                       Password to protect health endpoint (username is health)
exit status 2
```

#### Changes
Removed environmental variables

#### Testing
Didn't test. If this is broken, some automatic tests should fail.

##### Regressions
AWS Cloud: we override `TTN_LW_IS_DATABASE_URI` and `TTN_LW_REDIS_ADDRESS`, so no need to change here. I don't see any mention of `blob` in the templates, so I assume we'd need to add that environmental variable, right? Or `blob` doesn't apply there?

AWS BYOL/PAYG: they build their own AMIs, so this change is irrelevant

Someone please confirm.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ For details about compatibility between different releases, see the **Commitment
 - The `gs.status.fail` and `gs.up.fail` events. `gs.status.drop` and `gs.up.drop` should be used instead, as they contain the failure cause.
 - The `data_rate_index` field in uplink message metadata. Observe the fully described data rate in the `data_rate` field instead.
 - LoRaWAN data rate index reported to LoRa Cloud DMS.
+- Dockerfile doesn't define environmental variables `TTN_LW_BLOB_LOCAL_DIRECTORY`, `TTN_LW_IS_DATABASE_URI` and `TTN_LW_REDIS_ADDRESS` anymore. They need to be set when running the container: please refer to `docker-compose.yml` for example values.
 
 ### Fixed
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,10 +37,7 @@ RUN mkdir /srv/ttn-lorawan/public/blob
 
 VOLUME ["/srv/ttn-lorawan/public/blob"]
 
-ENV TTN_LW_BLOB_LOCAL_DIRECTORY=/srv/ttn-lorawan/public/blob \
-    TTN_LW_IS_DATABASE_URI=postgres://root@cockroach:26257/ttn_lorawan?sslmode=disable \
-    TTN_LW_REDIS_ADDRESS=redis:6379 \
-    TTN_LW_HEALTHCHECK_URL=http://localhost:1885/healthz/live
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -37,11 +37,6 @@ RUN mkdir /srv/ttn-lorawan/public/blob
 
 VOLUME ["/srv/ttn-lorawan/public/blob"]
 
-ENV TTN_LW_BLOB_LOCAL_DIRECTORY=/srv/ttn-lorawan/public/blob \
-    TTN_LW_IS_DATABASE_URI=postgres://root@cockroach:26257/ttn_lorawan?sslmode=disable \
-    TTN_LW_REDIS_ADDRESS=redis:6379 \
-    TTN_LW_HEALTHCHECK_URL=http://localhost:1885/healthz/live
```
htdvisser, line 43: We still need ENV for the HEALTHCHECK URL, which needs to be configurable in case someone changes the port number, or sets auth on that endpoint.

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ For details about compatibility between different releases, see the **Commitment
 
 - The `ttn_lw_gs_status_failed_total`, `ttn_lw_gs_uplink_failed_total` metrics. `ttn_lw_gs_status_dropped_total` and `ttn_lw_gs_uplink_dropped_total` should be used instead, as they contain the failure cause.
 - The `gs.status.fail` and `gs.up.fail` events. `gs.status.drop` and `gs.up.drop` should be used instead, as they contain the failure cause.
+- Dockerfile doesn't define environmental variables `TTN_LW_BLOB_LOCAL_DIRECTORY`, `TTN_LW_IS_DATABASE_URI`, `TTN_LW_REDIS_ADDRESS` and `TTN_LW_HEALTHCHECK_URL` anymore. `TTN_LW_BLOB_LOCAL_DIRECTORY`, `TTN_LW_IS_DATABASE_URI` and `TTN_LW_REDIS_ADDRESS` need to be set when running the container: please refer to `docker-compose.yml` for example values. `TTN_LW_HEALTHCHECK_URL` isn't used by the stack.
```
htdvisser: Please update. This still refers to the now unchanged `TTN_LW_HEALTHCHECK_URL`.

htdvisser (approved):
LGTM

michalborkowski96:
@htdvisser can you answer the questions in the "regressions" section?

htdvisser:
On AWS deployments, we indeed don't use a local blob directory. Instead we store blobs on S3.

PR #4802, opened 2021-10-29
Remove data rate index and LoRaWAN PHY version from API

### LINKED ISSUE #4478: Remove data rate indices from `TxSettings` and `TxRequest`

johanstokking:
#### Summary

Closes #4478

My contribution to this week's technical debt removal cycle.

#### Changes

- Remove `data_rate_index` from uplink message
- Remove `rx{1,2}_data_rate_index` from downlink request
- Remove `lorawan_phy_version` from downlink request (was only needed for mapping)
- Remove usages and obsolete mapping in GS, NS and PBA

#### Testing

CI

##### Regressions

The `data_rate_index` field in uplink messages has been marked deprecated in a previous minor.

The `rx{1,2}_data_rate_index` is internal API and already supports fallback behavior.

The data rate index is no longer reported to LoRa Cloud DMS as it is unavailable in that context. This field is marked optional so it can be omitted.

#### Notes for Reviewers

@htdvisser please check API
@adriansmares please check NS and AS
@KrishnaIyer please check GS

#### Checklist


adriansmares (approved):

pkg/gatewayserver/io/io.go
```diff
@@ -421,64 +420,50 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
...
-	// Backwards compatibility. If there's no PHY version defined, assume that the band is already versioned.
-	// TODO: Remove (https://github.com/TheThingsNetwork/lorawan-stack/issues/4478).
-	if lwPhyVersion == ttnpb.PHY_UNKNOWN {
-		lwPhyVersion = ttnpb.RP001_V1_1_REV_B
-	}
-
-	phy, err := band.Get(fp.BandID, lwPhyVersion)
+	// Gateway Server does not take the LoRaWAN Regional Parameters version into account as it is a transparent forwarder
+	// between the Network Server and the end device. However, Gateway Server does enforce spectrum regulations that are
+	// defined in Regional Parameters. These include maximum EIRP and maximum payload length. These are taken from the
```
KrishnaIyer, line 425: However, **for gateways**, the Gateway Server does enforce spectrum regulations that are defined in Regional Parameters.

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -272,7 +270,7 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 					continue
 				}
 				sessionCtx := NewContextWithSession(ctx, &session)
-				dnmsg, err := s.formatter.FromDownlink(sessionCtx, uid, *down, concentratorTime, dlTime)
+				dnmsg, err := s.formatter.FromDownlink(sessionCtx, uid, *down, conn.BandID(), concentratorTime, dlTime)
```
KrishnaIyer, line 273: Since LBS does not allow us to specify the RP version, there could be a mismatch between this and the RP version in the downlink message. For a device we can define the PHY Version but there's one for gateways.
Should we maybe add that (in a separate PR) as a config to the gateway object?

KrishnaIyer (approved):
GS LGTM.

pkg/gatewayserver/io/io.go
johanstokking: For what else in the downlink routine of the Gateway Server?

pkg/gatewayserver/io/ws/ws.go
johanstokking: Yes, there could be. But the LoRa Alliance TC will not be breaking the data rate indices once again.

What may happen, is that the gateway or the Gateway Server do not understand the same set of data rates. For instance, LR-FHSS gateways. There's little to be done about that. When the gateway uses a newer Regional Parameters version and sends uplink messages with a data rate index that is unknown, GS will drop the frame with a warning. The other way around, that the Gateway Server uses a data rate that the gateway doesn't recognize, is highly unlikely as it would be a downlink-only data rate. If that situation may happen, indeed we need to record the RP version implemented by the gateway, or better yet, map it with the reported Basic Station version.

htdvisser (approved):
API changes LGTM. 

But I'm not sure how this affects existing entities in the (NS/AS/JS)EndDevice stores.

PR #4806, opened 2021-11-01
Remove nullable option from timestamps

htdvisser:
#### Summary

This PR removes `gogoproto.nullable` from timestamps in `application.proto`, `client.proto`, `oauth.proto`, `organization.proto`, `rights.proto` and `user.proto`.

#### Changes

- Remove the option, fix related code

#### Testing

Existing unit tests

##### Regressions

None expected

#### Notes for Reviewers

Since the `expires_at` columns in the database are NULLable, I've also turned DB model fields into pointers.

#### Checklist


pkg/identityserver/store/application.go
```diff
@@ -80,8 +80,8 @@ var applicationColumnNames = map[string][]string{
 
 func (app Application) toPB(pb *ttnpb.Application, fieldMask *pbtypes.FieldMask) {
 	pb.Ids = &ttnpb.ApplicationIdentifiers{ApplicationId: app.ApplicationID}
-	pb.CreatedAt = cleanTime(app.CreatedAt)
-	pb.UpdatedAt = cleanTime(app.UpdatedAt)
+	pb.CreatedAt = cleanTimePtr(&app.CreatedAt)
```
adriansmares, line 83: In case any reviewer wants to ask: shouldn't this become nullable at model level too ?
The answer is no, since they are controlled by the stores, and they cannot be modified via RPCs/store updates.

adriansmares (approved):

adriansmares:
> Since the `expires_at` columns in the database are NULLable, I've also turned DB model fields into pointers.

Will this need a manual migration ? IIRC gorm `Migrate` won't touch existing database columns.

pkg/identityserver/store/application.go
htdvisser: Actually, it's because they are NOT NULL columns in the database.

htdvisser:
Those fields are already columns with NULL in the database schema

PR #4812, opened 2021-11-02
Add extended gateway sorting algorithm

### LINKED ISSUE #13: Improve gateway sorting algorithm in NS
**Summary:** 

Gateway sorting algorithm in NS can be improved

**Why do we need this?**

To make better gateway choices

A very simplistic gateway sorting algorithm

A more efficient gateway sorting algorithm

Based on experience with v2:

- SNR isn't the only metric; RSSI is also important.
- We may need to filter "incorrect data" (SNR=0, RSSI=0 for example)
- We may want to include gateway uplink utilization to avoid Tx on "valuable uplink gateways"
- We may want to include gateway downlink utilization to be a little bit smarter about downlink gtw 
selection than always selecting one with a full duty cycle
- We may benefit from a little bit of randomness

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/729 by @rvolosatovs

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/13

#### Changes

- Add a wanted RSSI sorting algorithm (based on https://github.com/packetbroker/router/issues/28#issuecomment-923944546)
  - Can be enabled using the feature flag `ns.down.path.use_wanted_rssi`
- Gateways with no RSSI/SNR are considered to be lower than everyone else
  - Can be enabled using the feature flag `ns.down.path.ignore_invalid_metadata`

#### Testing

Local testing, I'll test this on `staging1` too.

##### Regressions

To be looked into.

#### Notes for Reviewers

As mentioned offline, PB having lower priority than cluster local gateways is already implemented.

#### Checklist

### DIFF
--- a/pkg/networkserver/downlink.go
+++ b/pkg/networkserver/downlink.go
@@ -28,6 +28,7 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/encoding/lorawan"
 	"go.thethings.network/lorawan-stack/v3/pkg/errors"
 	"go.thethings.network/lorawan-stack/v3/pkg/events"
+	"go.thethings.network/lorawan-stack/v3/pkg/experimental"
 	"go.thethings.network/lorawan-stack/v3/pkg/frequencyplans"
 	"go.thethings.network/lorawan-stack/v3/pkg/log"
 	. "go.thethings.network/lorawan-stack/v3/pkg/networkserver/internal"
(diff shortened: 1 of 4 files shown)


pkg/networkserver/downlink.go
```diff
@@ -637,12 +638,55 @@ type downlinkPath struct {
...
+func wantedRSSIMetadataComparator(mds []*ttnpb.RxMetadata) func(int, int) bool {
+	wantedRSSI := func(k int) float32 { return computeWantedRSSI(mds[k].Snr, mds[k].ChannelRssi) }
+	return func(i, j int) bool {
+		return wantedRSSI(i) >= wantedRSSI(j)
+	}
+}
+
+var (
+	ignoreInvalidMDFeatureFlag = experimental.DefineFeature("ns.down.path.ignore_invalid_metadata", false)
+	useWantedRSSIFeatureFlag   = experimental.DefineFeature("ns.down.path.use_wanted_rssi", false)
```
johanstokking, line 666: Would it be good to declare these in a `<pkg>/experiments.go` file so that we can better keep track of them? Would also be useful to add godoc or an issue reference so we know in the future what it does, where it comes from, why it is an experiment, what makes us decide to either integrate it or remove the feature as I think we don't want permanent experiments, etc.

@htdvisser did you have some thoughts on this?

johanstokking (approved):
LGTM but see comment

pkg/networkserver/downlink.go
htdvisser: I don't think a `pkg/experiments.go` is necessary; we can already quite easily find references to `experimental.DefineFeature`.

I do agree that we should keep track of experimental features. Perhaps we can just track those in GitHub issues?

adriansmares: I've opened https://github.com/TheThingsNetwork/lorawan-stack/pull/4819 regarding an issue template for experimental feature tracking.

PR #4818, opened 2021-11-03
Fix GS payloads for forward events

KrishnaIyer:
#### Summary

Fix GS payloads for forward events.

#### Changes

- The payload should be the message and not the host name.

#### Testing

Local via E2E test.

##### Regressions

NA

#### Checklist

### DIFF
--- a/pkg/gatewayserver/observability.go
+++ b/pkg/gatewayserver/observability.go
@@ -64,6 +64,7 @@ var (
 	evtForwardUp = events.Define(
 		"gs.up.forward", "forward uplink message",
 		events.WithVisibility(ttnpb.RIGHT_GATEWAY_TRAFFIC_READ),
+		events.WithDataType(&ttnpb.UplinkMessage{}),
 	)
 	evtSendDown = events.Define(
 		"gs.down.send", "send downlink message",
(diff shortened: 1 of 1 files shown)


adriansmares:
This is intentional - you just get the same event 3 times otherwise, and the payload itself is found in the `received` event. 

If you want to find when an uplink has been forwarded, use correlation identifiers instead (the events should have a number of common ones per message).

KrishnaIyer:
Ok right then I think the payload shouldn't be there at all. 

When I try to unmarshal this event, since this message does not define a data type but sends data (the `host`), the data type becomes a` "@type":"type.googleapis.com/google.protobuf.Value"` and  the receiver doesn't know how to unmarshal this.

```
illegal base64 data at input byte 4
```

```json
{"result":{"name":"gs.up.forward","time":"2021-11-04T14:34:50.305429Z","identifiers":[{"gateway_ids":{"gateway_id":"1111111111111111","eui":"1111111111111111"}}],"data":{"@type":"type.googleapis.com/google.protobuf.Value","value":"cluster"},"correlation_ids":["gs:conn:01FKNNZAX8KXG9RNE9SP51QG91","gs:up:host:01FKNNZAXX2VG2FGN8YGCNBAS7","gs:uplink:01FKNNZCVCBQZM0473ZG58STNQ"],"origin":"Krishnas-MacBook-Pro.local","context":{"tenant-id":"Cgl0aGV0aGluZ3M="},"visibility":{"rights":["RIGHT_GATEWAY_TRAFFIC_READ"]},"unique_id":"01FKNNZD21YM07FG42NB8GBTJ0"}}
```

adriansmares:
What about `WithDataType("")` - does this work ?

KrishnaIyer:
I tried that indeed already, didn't...
It doesn't recognise that it's a string.

KrishnaIyer:
This isn't necessary anymore since the original issue has been solved.

Issue #2661, opened 2020-06-04
Speed up frontend builds


rvolosatovs:
#### Summary

Frontend builds take an enormous amount of time, that should be fixed.

#### Why do we need this?

For faster CI and better developer experience

#### What is already there? What do you see now?

Slow builds
```
mage js:clean js:build  533.08s user 27.53s system 140% cpu 6:38.00 total
```

#### What is missing? What do you want to see?

Faster builds.
These really should not take more than 1 minute.

#### How do you propose to implement this?

No idea

#### How do you propose to test this?

`time`

#### Can you do this yourself and submit a Pull Request?

@bafonins is better suited for the job

kschiffer:
FTR: \`mage js:clean js:build  533.08s user 27.53s system 140% cpu 6:38.00 total\`

suggests that your are using a fresh checkout without any dependencies and caches. The build step has thus to fetch all dependencies and run babel transpilations from scratch which takes the majority of time for the build command. Subsequent builds will be around 2mins:

```
mage js:clean js:build  111.87s user 8.71s system 122% cpu 1:38.47 total
```

We have no control over the speed of dependency downloads (other than using less dependencies). In CI, we already cache the dependencies between runs (with exception of release builds). As far as babel and some webpack processes goes, there might be some margin to reduce duration but getting the initial build to around 1 minute will be next to impossible IMO, but I'm happy to be proven otherwise.

PR #4824, opened 2021-11-04
Increase frontend build speed

kschiffer:
#### Summary
This PR will increase the frontend build speed by making use of webpacks DLL feature in production, as well as removing a calculation heavy file hashing plugin.

References #2661

```
Before ---
time mage js:clean js:cleanDeps jsSdk:Clean jsSdk:cleanDeps js:build
351.66s user 57.73s system 169% cpu 4:01.81 total

time mage js:build
91.81s user 12.12s system 119% cpu 1:27.14 total

After ---
time mage js:clean js:cleanDeps jsSdk:Clean jsSdk:cleanDeps js:build
230.47s user 33.66s system 198% cpu 2:12.87 total

time mage js:build
28.38s user 3.05s system 141% cpu 22.191 total
```
Note that these are measures from my M1 Mac, so the times on the CI machines will be significantly slower, however, the relative delta is about the same.

#### Changes
- Enable DLL referencing in production as well (currently development only)
- Enable caching of DLL module in GitHub action
- Remove `webpack-plugin-hash-output`
- Enable source map generation for release builds only
- Remove some unnecessary GitHub action job steps

#### Testing
Tested this in a separate PR #4769 

##### Regressions

In production we will now, additionally to the app bundle, also include a bundle that includes external libraries. It is possible that this leads to issues. I could however not find any when running this locally and also end-to-end tests work properly.

#### Notes for Reviewers
Note that this includes a config update, where the default values for the `js-files` config has to be updated to include the additional `libs.*.bundle.js` file that includes the external modules.

There is more speed up potential which I cannot work on currently and includes: 
- Removing number of modules (e.g. stylus files that use a lot of duplicate styles)
- Improving the mage tooling to avoid running the `yarn install` command multiple times during build
  - This could e.g. be done using yarn workspaces, so that the dependencies for webui and SDK can be installed in one go
- Treeshaking and sharing external dependencies
- Migrating to webpack 5, which (seemingly) has better build caching capabilities

#### Checklist

### DIFF
--- a/cmd/internal/shared/console/config.go
+++ b/cmd/internal/shared/console/config.go
@@ -41,7 +41,7 @@ var DefaultConsoleConfig = console.Config{
 			AssetsBaseURL: shared.DefaultAssetsBaseURL,
 			IconPrefix:    "console-",
 			CSSFiles:      []string{"console.css"},
-			JSFiles:       []string{"console.js"},
+			JSFiles:       []string{"libs.bundle.js", "console.js"},
 		},
 		FrontendConfig: console.FrontendConfig{
 			DocumentationBaseURL: "https://thethingsindustries.com/docs",
--- a/cmd/internal/shared/identityserver/config.go
+++ b/cmd/internal/shared/identityserver/config.go
@@ -38,7 +38,7 @@ var DefaultIdentityServerConfig = identityserver.Config{
 				AssetsBaseURL: shared.DefaultAssetsBaseURL,
 				IconPrefix:    "oauth-",
 				CSSFiles:      []string{"account.css"},
-				JSFiles:       []string{"account.js"},
+				JSFiles:       []string{"libs.bundle.js", "account.js"},
 			},
 			FrontendConfig: oauth.FrontendConfig{
 				DocumentationBaseURL: "https://thethingsindustries.com/docs",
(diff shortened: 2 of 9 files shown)


cmd/internal/shared/console/config.go
```diff
@@ -41,7 +41,7 @@ var DefaultConsoleConfig = console.Config{
 			AssetsBaseURL: shared.DefaultAssetsBaseURL,
 			IconPrefix:    "console-",
 			CSSFiles:      []string{"console.css"},
-			JSFiles:       []string{"console.js"},
+			JSFiles:       []string{"libs.bundle.js", "console.js"},
```
htdvisser, line 44: This may need a mention in CHANGELOG, because deployments that customized this (I can't imagine there are any, but still) would need to update their config.

cmd/internal/shared/identityserver/config.go
```diff
@@ -38,7 +38,7 @@ var DefaultIdentityServerConfig = identityserver.Config{
 				AssetsBaseURL: shared.DefaultAssetsBaseURL,
 				IconPrefix:    "oauth-",
 				CSSFiles:      []string{"account.css"},
-				JSFiles:       []string{"account.js"},
+				JSFiles:       []string{"libs.bundle.js", "account.js"},
```
htdvisser, line 41: Same

htdvisser (approved):

htdvisser (approved):
Every improvement helps. Keep up the good work!

bafonins (approved):

adriansmares (approved):

PR #4825, opened 2021-11-04
Remove some gogoproto options from metadata.proto

KrishnaIyer:
#### Summary

Remove some gogoproto options from metadata.proto. Refs #2798 

#### Changes

- Remove embed and nullable from `RxMetadata.GatewayIDs` and `PacketBrokerRouteHop`.
- Update occurences and tests.

#### Testing

I think UT should suffice.

##### Regressions

Should be ideally caught with UT.

#### Notes for Reviewers

1. I've made a few comments in some places, please check those.
2. There are no instances where `IDstringer` is necessary for `RxMetadata` so I haven't implemented that.
3. I'll add the enterprise counterpart (mostly in the GS) before merging this.
4. @adriansmares: The uplink dedup implementation [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/ad470517e11cc2e5f4691b64a6e18133db7315df/pkg/networkserver/redis/uplink_deduplicator.go#L58) uses the `RxMetadata` message. Now one of it's fields (Gateway ID) is a nullable, non-embedded entity. Does this cause a mismatch with existing entities in the registry. If yes, since dedup window is short (200 ms afaik), is this change Ok since we will get new uplinks once this is deployed?

#### Checklist


pkg/networkserver/internal/utils.go
```diff
@@ -104,8 +104,10 @@ func RXMetadataStats(ctx context.Context, mds []*ttnpb.RxMetadata) (gateways int
 	for _, md := range mds {
 		if md.PacketBroker != nil {
 			gtws[fmt.Sprintf("%s@%s/%s", md.PacketBroker.ForwarderClusterId, md.PacketBroker.ForwarderNetId, md.PacketBroker.ForwarderTenantId)] = struct{}{}
+		} else if md.GatewayIds != nil {
+			gtws[unique.ID(ctx, md.GatewayIds)] = struct{}{}
 		} else {
-			gtws[unique.ID(ctx, md.GatewayIdentifiers)] = struct{}{}
+			continue // Metadata without PB or Gateway IDs should be invalid so skipping.
```
KrishnaIyer, line 110: I think it's right to skip here but please check @adriansmares.

pkg/applicationserver/applicationserver.go
```diff
@@ -883,16 +883,18 @@ func (as *ApplicationServer) storeUplink(ctx context.Context, ids ttnpb.EndDevic
...
-			GatewayIdentifiers: ttnpb.GatewayIdentifiers{
-				GatewayId: md.GatewayId,
-			},
-			AntennaIndex:  md.AntennaIndex,
-			FineTimestamp: md.FineTimestamp,
-			Location:      md.Location,
-			Rssi:          md.Rssi,
-			Snr:           md.Snr,
-		})
+		if md.GetGatewayIds() != nil {
```
adriansmares: Nit: turn this around: on `==`, do `continue`. Saves a nesting level.

I also wouldn't use the getter and as we access every other field directly, so it if `md` is `nil`, we're not gaining much.

pkg/applicationserver/io/packages/loragls/v3/api/objects.go
```diff
@@ -119,7 +119,7 @@ func BuildSingleFrameRequest(ctx context.Context, metadata []*ttnpb.RxMetadata)
 		Frame:    Frame{},
 	}
 	for _, m := range metadata {
-		if m.Location == nil {
+		if m.Location == nil || m.GetGatewayIds() == nil {
```
adriansmares: No need for the getter here as if `m == nil`, `m.Location` will panic anyway. Same below.

adriansmares (approved):
This LGTM.

My only comment is that in places where you have something like `m.Field` being actively read/wrote to, there is no point to use `m.GetGatewayIds()` instead of `m.GatewayIds`, since if `m` is `nil` everything around will panic anyway.

pkg/applicationserver/applicationserver.go
htdvisser: Agree with the `continue` on `== nil`, this is always the preferred style in Go.

And indeed, we also need to be more consistent about when to use getters and when to write our own nil checks and directly access fields.

htdvisser (approved):
Code-owned files LGTM.

pkg/applicationserver/io/packages/loragls/v3/api/objects.go
KrishnaIyer: Ok I'll remove the getter but `m.GatewayIds` can still be nil even if `m != nil`.

adriansmares: You're correct. It's just that there is no point in using the getter if we access the other fields without it.

PR #4851, opened 2021-11-09
Fix skip payload crypto setting for applications and end devices

### LINKED ISSUE #4782: Fix skip payload crypto settings in the Console
#### Summary
The Console currently only allows setting `Skip payload crypto` on the device level. It should also be possible to set the payload cypto skipping on the link (Application) level. Additionally, the field should allow three states -> Enabled, disabled, default.

Please see also: https://github.com/TheThingsIndustries/lorawan-stack-support/issues/485

#### Why do we need this?
The incomplete feature in the Console currently causes confusion and it also produces irreversible conditions since there is no way to reset the option (`Use default`).

We currently have a checkbox option in the end device general settings `Skip payload crypto`.

- We can add `Form.CollapseSection`'s to the application general settings for `Basic settings` (for everything that is currently there) and `Link settings`. The link settings will contain the new `Skip payload crypto` field with three radio buttons: 
- Skip payload crypto
- Don't skip payload crypto
- Use default

The form will make a request to the [`SetLink` rpc of the `As` service](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.16/api/applicationserver.proto#L128-L133). For this we will need to add an `setApplicationLinkLogic()` to `/pkg/webui/console/store/middleware/logics/applications.js`, along with the necessary redux primitives (actions, reducers, selectors). The form can then make the request to the backend via a store dispatch.

Since this setting is a bit interconnected, it makes sense to add an end-to-end test case here as well, which
- Checks whether the settings can be made and unset again, both on end device and link level
- Checks whether the message tab of the end device shows a disabled form with the warning that messaging is disabled for devices that skip crypto
  - Check this for various combinations of end device level and link level settings, e.g. keeping in mind that the end device level setting has precedence.

@ryaplots

ryaplots:
…pdate app, e2e test

#### Summary

Referring to #4782
To create a skip payload crypto (radio buttons on the device general settings and checkbox on the application general settings)

#### Changes

- pkg/webui/console/views/device-general-settings/application-server-form/index.js: added the radio buttons and form validation with default values.
- pkg/webui/console/views/application-general-settings/index.js: added a checkbox, its initial value (/mapping.js) and validation schema
- pkg/webui/console/store/middleware/logics/applications.js: middleware logic for updating an application
- pkg/webui/console/store/actions/link.js: action for updating an application
- cypress/integration/console/applications/edit.spec.js: checkbox e2e test

#### Testing

Trying to set up the test env correctly (TO EDIT)

##### Regressions

Application updating, skip payload crypto. 

#### Notes for Reviewers

In the device general settings I am validating the payload crypto twice. I did it this way because:
- the payload crypto should be sent as a boolean (these are inferred from the chosen radio button) 
- radio buttons are part of the form and they don't accept boolean as value
- I need to set the initial value as default which would not be possible if the form was validated with only a boolean.

#### Checklist


pkg/webui/console/views/application-general-settings/index.js
```diff
@@ -126,6 +127,7 @@ const validationSchema = Yup.object().shape({
       sharedMessages.attributeValueValidateTooLong,
       attributeValueTooLongCheck,
     ),
+    skip_payload_crypto: Yup.boolean().default(false),
```
kschiffer:
```suggestion
    skip_payload_crypto: Yup.boolean(),
```

I don't think you need that.

pkg/webui/console/views/device-general-settings/application-server-form/index.js
```diff
@@ -163,13 +177,17 @@ const ApplicationServerForm = React.memo(props => {
...
       {showResetNotification && <Notification content={messages.keysResetWarning} info small />}
       <Form.Field
-        autoFocus
+        required
         title={sharedMessages.skipCryptoTitle}
-        name="skip_payload_crypto"
+        name="skip_payload_crypto_value"
         description={sharedMessages.skipCryptoDescription}
-        component={Checkbox}
+        component={RadioButton.Group}
```
kschiffer, line 187: Have a look into the `encode`/ `decode` props of the `<Form.Field />`. You can use that to marshal values from the backend shape to the frontend shape and vice versa. Iirc the backend expects either `true`, `false` or `null`. If you use a `decode` function you can convert this into `skip`, `include` and `default` and the other way around.

pkg/webui/console/views/device-general-settings/application-server-form/index.js
```diff
@@ -31,11 +31,21 @@ import { generate16BytesKey, isNonZeroSessionKey } from '@console/lib/device-uti
...
+})
+
 const validationSchema = Yup.object()
   .shape({
-    skip_payload_crypto: Yup.boolean().default(false),
+    skip_payload_crypto_value: Yup.string()
+      .oneOf(['skip', 'include', 'default'])
+      .required(sharedMessages.validateRequired)
+      .default('default'),
+    skip_payload_crypto_overrride: Yup.boolean().nullable().default(null),
```
kschiffer: You don't need to use two values here. You can use `decode` and `encode` props of the `<Form.Field />` to marshal between notations. See my other comment below. In the validation schema you should think about validating the values in a way that they can be sent straight to the backend.

kschiffer (commented):
Good start! See my remarks.

cypress/integration/console/devices/messaging/edit.spec.js
```diff
@@ -70,6 +70,14 @@ describe('End device messaging', () => {
     })
 
     it('succeeds sending uplink message', () => {
+      cy.visit(
+        `${Cypress.config('consoleRootPath')}/applications/${applicationId}/general-settings`,
+      )
+
+      cy.findByRole('checkbox').uncheck()
```
kschiffer: This will potentially generate a false positive / false negative when we add another checkbox input to the application form. You should use `findByLabelText()` rather.

Generally, when writing cypress tests, try to mimic the conscious action of the user. E.g. a user would not just click on a checkbox, they would scan for a text that is about skipping payload crypto and check the checkbox connected to it. Writing tests in this way makes them more sustainable even when we change the format or structure of the form.

See also our development guidelines [about end-to-end tests](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.16/DEVELOPMENT.md#guiding-principle) and the docs of [testing library](https://testing-library.com/docs/).

cypress/integration/console/devices/messaging/edit.spec.js
```diff
@@ -88,6 +96,131 @@ describe('End device messaging', () => {
       cy.findByTestId('error-notification').should('not.exist')
     })
 
+    it('combination skip payload crypto enabled/disabled on application level and default on device level', () => {
```
kschiffer: Make sure to [name and structure your tests correctly](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.16/DEVELOPMENT.md#structurizing-tests).

cypress/integration/console/devices/messaging/edit.spec.js
```diff
@@ -88,6 +96,131 @@ describe('End device messaging', () => {
...
+
+      cy.intercept(
+        'GET',
+        `as/applications/${applicationId}/devices/${endDeviceId}?field_mask=version_ids,formatters,skip_payload_crypto_override,session,pending_session`,
+        response,
+      )
+
+      cy.findByTestId('notification')
+        .should('be.visible')
+        .findByText(`Simulation is disabled for devices that skip payload crypto`)
```
kschiffer:
```suggestion
        .findByText('Simulation is disabled for devices that skip payload crypto')
```

pkg/webui/console/views/application-general-settings/index.js
```diff
@@ -254,14 +277,21 @@ export default class ApplicationGeneralSettings extends React.Component {
     // If there is a change in attributes, copy all attributes so they don't get
     // overwritten.
     const update =
-      'attributes' in changed ? { ...changed, attributes: appValues.attributes } : changed
+      'attributes' in changed
+        ? {
+            ...changed,
+            attributes: appValues.attributes,
+            skip_payload_crypto: appValues.skip_payload_crypto,
```
kschiffer:
```suggestion
```
Is this necessary? The new value should be part of `...changed` anyway (?)

pkg/webui/console/views/device-general-settings/application-server-form/index.js
```diff
@@ -163,13 +180,19 @@ const ApplicationServerForm = React.memo(props => {
...
-        autoFocus
+        required
         title={sharedMessages.skipCryptoTitle}
-        name="skip_payload_crypto"
+        name="skip_payload_crypto_override"
         description={sharedMessages.skipCryptoDescription}
-        component={Checkbox}
+        component={RadioButton.Group}
+        decode={decodeSkipPayloadCrypto}
+        encode={encodeSkipPayloadCrypto}
```
kschiffer, line 189:
```suggestion
        horizontal
        encode={encodeSkipPayloadCrypto}
```

Better layout.

cypress/integration/console/devices/messaging/edit.spec.js
```diff
@@ -88,6 +96,131 @@ describe('End device messaging', () => {
...
     })
 
+    it('combination skip payload crypto enabled/disabled on application level and default on device level', () => {
+      cy.visit(
+        `${Cypress.config('consoleRootPath')}/applications/${applicationId}/general-settings`,
+      )
+
+      cy.findByRole('checkbox').check()
+
+      cy.findByRole('button', { name: 'Save changes' }).click()
```
kschiffer: The functionality of the application general settings page is not relevant for this test. You should instead make the required API calls directly to achieve the modifications directly (and more quickly).
If you want to verify the integrity of a complete flow of actions of different domains, you should write a smoke test instead.

cypress/integration/console/devices/messaging/edit.spec.js
```diff
@@ -88,6 +96,131 @@ describe('End device messaging', () => {
...
+      const response = {
+        skip_payload_crypto_override: null,
+        session: {},
+      }
+
+      cy.intercept(
+        'GET',
+        `as/applications/${applicationId}/devices/${endDeviceId}?field_mask=version_ids,formatters,skip_payload_crypto_override,session,pending_session`,
+        response,
+      )
```
kschiffer: You are mixing intercepting and actual modifications (like above). You should try to avoid API mocks as much as possible, since they lower the fidelity of the test. You can do an action POST request here as well to achieve the desired modifications in the registry.

cypress/integration/console/devices/messaging/edit.spec.js
```diff
@@ -88,6 +96,131 @@ describe('End device messaging', () => {
...
+      cy.visit(
+        `${Cypress.config(
+          'consoleRootPath',
+        )}/applications/${applicationId}/devices/${endDeviceId}/messaging/uplink`,
+      )
+
+      cy.findByLabelText('FPort').should('be.enabled')
+      cy.findByLabelText('Payload').should('be.enabled')
+
+      cy.findByRole('button', { name: 'Simulate uplink' }).should('be.enabled')
```
kschiffer: Do not combine multiple distinct tests into one. If you do that, the other modifications done before could interfere with the test and lower its expressiveness. Try to chunk tests into distinct bits that test a single thing in isolation.

A good test setup would resemble sth like:

```js
describe('Uplink', () => {
  …
  describe('when application link skips payload crypto', () => {
    beforeEach(() => { … })
    it('disables messaging when not using a SPC overwrite', () => {})
    it('still allows messaging when device disabled SPC via overwrite', () => {})
  })
  describe('when application link doesn't skip payload crypto', () => {
    beforeEach(() => { … })
    it('enables messaging when not using a SPC overwrite', () => {})
    it('still disables messaging when device disabled SPC via overwrite', () => {})
  })
})
```

All of these above should use direct API calls to set up the application and end device for the test scenario and only then run the assertions.

kschiffer (changes requested):
Looking good, but the cypress tests need some restructuring.

pkg/webui/console/views/application-general-settings/index.js
ryaplots: I think you are right and it is not necessary.

pkg/webui/console/views/application-general-settings/index.js
```diff
@@ -254,14 +277,20 @@ export default class ApplicationGeneralSettings extends React.Component {
...
+          }
+        : changed
 
     const {
       ids: { application_id },
     } = application
 
     try {
       await updateApplication(application_id, update)
+      await updateApplicationLink(application_id, update)
```
kschiffer: You need to distinguish `update` so that only the relevant fields are included. If you send the whole patch to the `updateApplicationLink` rpc you will get an error because you use invalid field masks.
```suggestion
      const { skip_payload_crypto, ...applicationUpdate } = update
      const linkUpdate = { skip_payload_crypto }
      await updateApplication(application_id, applicationUpdate)
      await updateApplicationLink(application_id, linkUpdate)
```

cypress/integration/console/shared/skip-payload-crypto/edit.spec.js
```diff
@@ -0,0 +1,211 @@
...
+          },
+        }
+        cy.request({
+          method: 'PUT',
+          url: `api/v3/as/applications/${applicationId}/link?field_mask=skip_payload_crypto`,
+          body: linkRequestBody,
+          headers: {
+            Authorization: `Bearer ${adminApiKey}`,
+          },
+        })
```
kschiffer: I guess this is WIP?

cypress/integration/console/shared/skip-payload-crypto/edit.spec.js
```diff
@@ -0,0 +1,250 @@
...
+          )}/applications/${applicationId}/devices/${endDeviceId}/messaging/uplink`,
+        )
+
+        cy.findByTestId('notification')
+          .should('be.visible')
+          .findByText(`Simulation is disabled for devices that skip payload crypto`)
+          .should('be.visible')
+
+        cy.findByLabelText('FPort').should('be.disabled')
+        cy.findByLabelText('Payload').should('be.disabled')
```
kschiffer:
```suggestion
```
I think it's enough to check for the button only.

cypress/integration/console/shared/skip-payload-crypto/edit.spec.js
```diff
@@ -0,0 +1,250 @@
...
+        const response = {
+          skip_payload_crypto_override: null,
+          session: {},
+        }
+
+        cy.intercept(
+          'GET',
+          `as/applications/${applicationId}/devices/${endDeviceId}?field_mask=version_ids,formatters,skip_payload_crypto_override,session,pending_session`,
+          response,
+        )
```
kschiffer: Only use intercepts if absolutely necessary. In the end-to-end tests we try to reproduce usage as closely as possible, this includes relying on actual API responses wherever possible. 
In this case, you could either actually update the end device using a `PUT` request, or initially create the end device with the desired configuration directly.

cypress/integration/console/shared/skip-payload-crypto/edit.spec.js
```diff
@@ -0,0 +1,250 @@
...
+        const response = {
+          skip_payload_crypto_override: false,
+          session: {},
+        }
+
+        cy.intercept(
+          'GET',
+          `as/applications/${applicationId}/devices/${endDeviceId}?field_mask=version_ids,formatters,skip_payload_crypto_override,session,pending_session`,
+          response,
+        )
```
kschiffer: See my comment above on using actual API responses.

cypress/integration/console/shared/skip-payload-crypto/edit.spec.js
```diff
@@ -0,0 +1,250 @@
...
+        const response = {
+          skip_payload_crypto_override: null,
+          session: {},
+        }
+
+        cy.intercept(
+          'GET',
+          `as/applications/${applicationId}/devices/${endDeviceId}?field_mask=version_ids,formatters,skip_payload_crypto_override,session,pending_session`,
+          response,
+        )
```
kschiffer: Also here.

cypress/integration/console/shared/skip-payload-crypto/edit.spec.js
```diff
@@ -0,0 +1,250 @@
...
+
+        cy.visit(
+          `${Cypress.config(
+            'consoleRootPath',
+          )}/applications/${applicationId}/devices/${endDeviceId}/messaging/uplink`,
+        )
+
+        cy.findByLabelText('FPort').should('be.enabled')
+        cy.findByLabelText('Payload').should('be.enabled')
```
kschiffer:
```suggestion
```

cypress/integration/console/shared/skip-payload-crypto/edit.spec.js
```diff
@@ -0,0 +1,250 @@
...
```
kschiffer:
```suggestion
```

cypress/integration/console/shared/skip-payload-crypto/edit.spec.js
```diff
@@ -0,0 +1,250 @@
...
+        const response = {
+          skip_payload_crypto_override: true,
+          session: {},
+        }
+
+        cy.intercept(
+          'GET',
+          `as/applications/${applicationId}/devices/${endDeviceId}?field_mask=version_ids,formatters,skip_payload_crypto_override,session,pending_session`,
+          response,
+        )
```
kschiffer: See above.

cypress/integration/console/shared/skip-payload-crypto/edit.spec.js
```diff
@@ -0,0 +1,250 @@
...
+          )}/applications/${applicationId}/devices/${endDeviceId}/messaging/uplink`,
+        )
+
+        cy.findByTestId('notification')
+          .should('be.visible')
+          .findByText(`Simulation is disabled for devices that skip payload crypto`)
+          .should('be.visible')
+
+        cy.findByLabelText('FPort').should('be.disabled')
+        cy.findByLabelText('Payload').should('be.disabled')
```
kschiffer:
```suggestion
```

cypress/integration/console/applications/edit.spec.js
```diff
@@ -65,6 +65,20 @@ describe('Application general settings', () => {
...
+
+    cy.findByLabelText('Skip payload encryption and decryption').click()
+
+    cy.findByRole('button', { name: 'Save changes' }).click()
+
+    cy.findByTestId('error-notification').should('not.exist')
+    cy.findByTestId('toast-notification')
+      .should('be.visible')
+      .findByText(`Application updated`)
+      .should('be.visible')
```
kschiffer:
```suggestion
      .findByText(`Application updated`)
      .should('be.visible')
```
I know we use the assertions like that in other places, but this is actually an unnecessary duplicate check, so let's not do that.

pkg/webui/console/views/device-general-settings/application-server-form/index.js
```diff
@@ -31,9 +32,15 @@ import { generate16BytesKey, isNonZeroSessionKey } from '@console/lib/device-uti
 
 import messages from '../messages'
 
+const m = defineMessages({
+  skip: 'Skip payload crypto',
+  include: 'Do not skip payload crypto',
+  default: 'Use default',
```
kschiffer:
```suggestion
  default: 'Use application default',
```

pkg/webui/console/views/device-general-settings/application-server-form/index.js
```diff
@@ -31,9 +32,15 @@ import { generate16BytesKey, isNonZeroSessionKey } from '@console/lib/device-uti
 
 import messages from '../messages'
 
+const m = defineMessages({
+  skip: 'Skip payload crypto',
```
kschiffer:
```suggestion
  skip: 'Enforce skipping payload crypto',
```

pkg/webui/console/views/device-general-settings/application-server-form/index.js
```diff
@@ -31,9 +32,15 @@ import { generate16BytesKey, isNonZeroSessionKey } from '@console/lib/device-uti
...
+  include: 'Do not skip payload crypto',
```
kschiffer:
```suggestion
  include: 'Enforce payload crypto',
```

pkg/webui/console/views/device-general-settings/application-server-form/index.js
```diff
@@ -163,13 +181,20 @@ const ApplicationServerForm = React.memo(props => {
...
       <Form.Field
-        autoFocus
+        required
         title={sharedMessages.skipCryptoTitle}
         name="skip_payload_crypto_override"
         description={sharedMessages.skipCryptoDescription}
-        component={Checkbox}
+        component={RadioButton.Group}
+        decode={decodeSkipPayloadCrypto}
+        horizontal
```
kschiffer:
```suggestion
```

Given the longer option names (see above), I think we should use the vertical layout after all.

pkg/webui/console/views/device-general-settings/application-server-form/index.js
```diff
@@ -163,13 +181,20 @@ const ApplicationServerForm = React.memo(props => {
     >
       {showResetNotification && <Notification content={messages.keysResetWarning} info small />}
       <Form.Field
-        autoFocus
+        required
```
kschiffer:
```suggestion
```
Radio button options can never be `required` since there is always an option selected already.

pkg/webui/console/views/device-general-settings/application-server-form/index.js
```diff
@@ -163,13 +181,20 @@ const ApplicationServerForm = React.memo(props => {
...
         title={sharedMessages.skipCryptoTitle}
```
kschiffer: Let's use a different message here, reading `Payload crypto override`.

pkg/webui/console/views/device-general-settings/application-server-form/index.js
```diff
@@ -163,13 +181,20 @@ const ApplicationServerForm = React.memo(props => {
...
+        component={RadioButton.Group}
+        decode={decodeSkipPayloadCrypto}
+        horizontal
+        encode={encodeSkipPayloadCrypto}
         onChange={handleSkipCryptoChange}
-      />
+      >
+        <RadioButton label={m.skip} value="skip" />
+        <RadioButton label={m.include} value="include" />
+        <RadioButton label={m.default} value="default" />
```
kschiffer:
```suggestion
        <RadioButton label={m.default} value="default" />
        <RadioButton label={m.skip} value="skip" />
        <RadioButton label={m.include} value="include" />
```

kschiffer (changes requested):
Almost there. Have some remarks mainly about cypress and wording.

pkg/webui/console/views/device-general-settings/application-server-form/index.js
kschiffer: Let's also add a tooltip:

> Skip payload crypto disables the application layer encryption of LoRaWAN frames. This causes the Application Server to forward the messages without any processing, such as payload formatters, to the integrations.  When doing so, the integrations are responsible for decryption and processing of the binary format in order to understand the message. 
> 
> This application-wide setting can be overwritten per end device using this overwrite setting.

htdvisser:

pkg/webui/lib/field-description-messages.js
```diff
@@ -250,6 +251,8 @@ const m = defineMessages({
 
   statusTimePeriodicityDescription:
     'Interval to request the end device status. Set to 0 to disable requesting the end device status on an interval.',
+  skipPayloadCryptoOverride:
```
kschiffer:
```suggestion
  skipPayloadCryptoOverrideDescription:
```

kschiffer (approved):
LGTM, see my remaining remark on the message naming.

PR #4828, opened 2021-11-05
Synchronize GPS time via uplink token

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2798
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4823

#### Changes

- Remove `embed` and `nullable` from `GatewayAntennaIdentifiers` and `UplinkToken`
- Add the gateway absolute time (GPS based) to `ttnpb.UplinkToken`, and make synchronization take it into account
- The Basic Station frontend now uses the reported `GPSTime` in order to compute the gateway time at which an uplink has been received

#### Testing

Unit testing. Will also run a snapshot on `staging1`.

##### Regressions

- The Network Server changes are strictly about `nullable`/`embed`, and I expect that unit tests cover these.
- The Gateway Server changes about GPS time should be covered once again by existing unit tests, and some of the refactoring is based on deduplication (moving time related formulas to `time.go`).

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

Issue #4852, opened 2021-11-09
Basic Station time transfer propagation


adriansmares:
#### Summary

The Basic Station time transfer algorithm should transfer `gpstime` from PPS/GPS enabled gateways to other gateways on frame reception.

#### Why do we need this?

The current time transfer algorithm is subject to RTT variations, while the propagation variant is not.

#### What is already there? What do you see now?

The server attempts to transform a server timestamp into a concentrator timestamp by taking into account the observed RTTs.

#### What is missing? What do you want to see?

The actual idea behind Basic Station time transfer is ingenious, and it's not subject to RTT variations:
Consider two gateways, `G1` which has a PPS available and knows the current GPS time, and `G2`, which does not.
If we have a frame which has been received by both `G1`, and `G2`, we now know the GPS time based on the `gpstime` reported by `G1`, and we know the `xtime` at which `G2` has received the frame. We can now transfer this pair to `G2` in order to allow it to have GPS time synchronization.

#### Environment

N/A.

#### How do you propose to implement this?

The challenge is that in our architecture, frame deduplication occurs in the Network Server, not in the Gateway Server. As such we would need a mini-deduplication process to occur at Gateway Server layer.

Another variant is to have the Network Server return the metadata of the frames after deduplication. The Gateway Server can store this in some per-gateway buffer that would allow the Basic Station frontend to search for a `(xtime, gpstime)` pair. 

#### How do you propose to test this?

Using a PPS enabled gateway and another Basic Station gateway. Receive an uplink via both gateways, and check that the second gateway receives the GPS synchronization afterwards. 

#### Can you do this yourself and submit a Pull Request?

Yes.

PR #4854, opened 2021-11-09
BasicStation timesync fixes

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/4805
References https://github.com/TheThingsNetwork/lorawan-stack/pull/4828

#### Changes

- Lower the `timesync` interval to 200 seconds, from 30 minutes
  - This ensures that both the `gpstime`, and the `MuxTime` are fresh
- Disable time transfer to gateways that requested a `timesync` round with the Gateway Server
  - This ensures that we don't push time synchronization messages to gateways that do not require them
- Add a 10% jitter to `ping-pong` and `timesync` tickers.
  - This should lower the pressure on the Gateway Server when multiple gateways reconnect at the same time.

#### Testing

Local testing + `staging1`.

##### Regressions

These changes have been made in the presence of a PPS-enabled gateway, and have been tested both with PPS enabled, and disabled.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/ws/config.go
+++ b/pkg/gatewayserver/io/ws/config.go
@@ -28,6 +28,8 @@ type Config struct {
 var DefaultConfig = Config{
 	UseTrafficTLSAddress: false,
 	WSPingInterval:       30 * time.Second,
-	TimeSyncInterval:     30 * time.Minute,
+	// Assuming 5ppm of drift, this means a drift of 5 microseconds in one second.
+	// A drift of 1 millisecond would occur every 200 seconds in such a situation.
+	TimeSyncInterval:     200 * time.Second,
 	AllowUnauthenticated: false,
 }
--- a/pkg/gatewayserver/io/ws/lbslns/downstream.go
+++ b/pkg/gatewayserver/io/ws/lbslns/downstream.go
@@ -136,8 +136,6 @@ func (dnmsg *DownlinkMessage) ToDownlinkMessage(bandID string) (*ttnpb.DownlinkM
 }
 
 const (
-	// transferTimeRTTPercentile is the percentile of round-trip times that is considered for determining the gateway GPS time.
-	transferTimeRTTPercentile = 90
 	// transferTimeMinRTTCount is the minimum number of observed round-trip times that are taken into account before using
 	// their statistics to determine the gateway GPS time.
 	transferTimeMinRTTCount = 5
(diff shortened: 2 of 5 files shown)


pkg/gatewayserver/io/ws/config.go
```diff
@@ -28,6 +28,8 @@ type Config struct {
 var DefaultConfig = Config{
 	UseTrafficTLSAddress: false,
 	WSPingInterval:       30 * time.Second,
-	TimeSyncInterval:     30 * time.Minute,
+	// Assuming 5ppm of drift, this means a drift of 5 microseconds in one second.
+	// A drift of 1 millisecond would occur every 200 seconds in such a situation.
```
johanstokking, line 32: And we need 1 millisecond precision because of class B, right?

pkg/gatewayserver/io/ws/lbslns/downstream.go
```diff
@@ -154,15 +152,15 @@ func (*lbsLNS) TransferTime(ctx context.Context, serverTime time.Time, conn *io.
 		return nil, nil
 	}
 
-	_, _, _, np, n := conn.RTTStats(transferTimeRTTPercentile, serverTime)
+	_, _, median, _, n := conn.RTTStats(100, serverTime)
```
johanstokking, line 155: We are no longer accepting outliers? What is median about p100?
Isn't this basically considering the max value?

pkg/gatewayserver/io/ws/config.go
adriansmares: Yes, trying to keep it as small as possible, given that [real time transfer is slightly more complicated](https://github.com/TheThingsNetwork/lorawan-stack/issues/4852).

pkg/gatewayserver/io/ws/lbslns/downstream.go
adriansmares: The percentile parameters only affects `np`, the last returned element in the tuple, right ?

https://github.com/TheThingsNetwork/lorawan-stack/blob/bb54ef7af244e85b55c5311e64f41376fffbe919/pkg/gatewayserver/io/rtts.go#L63-L92

`min`,`max`,`median` are over the full sample.

P90 is a good upper bound - 90% of the samples will be bellow it, but it's not a good estimator for the expected value itself.

johanstokking: Ah right I missed that here, thanks

johanstokking (approved):

KrishnaIyer (commented):
I don't have anything to add to Johan's comments.

PR #4859, opened 2021-11-10
Disconnect WS gateways that stop sending WS pong

### LINKED ISSUE #4783: Gateway status is slow to be updated using LBS
#### Summary
When using LoRa Basics Station the status in the console is slow to update if the gateway gets disconnected.

#### Steps to Reproduce
1. Turn on gateway using LBS

2. Unplug the gateway and wait

3. Wait for the status in the console to update or use 
ttn-lw-cli gateway get-connection-stats <gtw_id> or 
TTI API: v3/gs/gateways/<gtw_id>/connection/stats 

to check when gateway goes offline.

#### What do you see now?
It takes around 18 min for the gateway to show up as disconnected after it has been unplugged.

A faster way to check if a gateway is offline. The problem does not exist when using udp.

#### Environment

The Things Stack Cloud v3.15.1
LBS both using LNS and CUPS.
This has been tested on multiple different gateway brands.

WisGate Edge Lite 2
MultiTech Conduit AP
Tektelic Kona Micro
Gemtek Indoor Femto Gateway

All running Basics Station v2.0.5

Don't know

Turn on a gateway and then unplug it.

no

KrishnaIyer:
#### Summary

Disconnect WS gateways that stop sending WS pong. Closes #4783.

#### Background of the issue and the solution

- This issue only occurs with unclean TCP disconnects, for ex yanking the power cable of the gateway.
- This is not visible with `wscat` which does a clean disconnect.
- With local testing, I found that the gateway connection exists for about 2-3 mins after I've removed the power.
- In addition [AWS ELB's keep alive](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html#connection-idle-timeout ) is 350 seconds (5.8 mins) and that cannot be changed. The end result being that the connection is kept alive for 10-12 mins. On one test, I even got 18 mins. I'm not sure how to account for that.
>  Elastic Load Balancing sets the idle timeout value for TCP flows to 350 seconds. You cannot modify this value. Clients or targets can use TCP keepalive packets to reset the idle timeout. Keepalive packets sent to maintain TLS connections cannot contain data or payload. 
- So logically, we disconnect the gateway based on WS ping-pong.
  - Many gateways don't send ping to the server. We cannot use gateway pings as the reference.
  - A few of them don't even respond with pong to server ping.
  - So I've designed the logic to disconnect only if the gateway is capable of sending pongs but doesn't at some point in the connection.

#### Changes

- Use a new ticker to wait for pongs after a ping. If the gateway doesn't send pongs, disconnect.
- Here are the cases
  - Gateway doesn't send pong; don't do anything. Such gateways will continue to have the issue and we can't do anything. We should reach out to those gateway vendors to support WS pong.
  - Gateway sends pongs but stops sending at some point. At least allow 2 failed pings before disconnecting.
- Add UT.

#### Testing

UT and tested locally with my TTIG.
- I've also tested that this does not interfere with the reconnection.

##### Regressions

Covered by UT. Let's test this in staging.

#### Notes for Reviewers

We should do some extended testing for a few days in staging (and TTN) since we don't want to be randomly disconnecting gateways.
Should we put this under a flag so that we can get some feedback from TTS CE? Or should can move this to the next milestone and test on TTS CE first and update TTSC in `v3.16.1`.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/ws/ws.go
+++ b/pkg/gatewayserver/io/ws/ws.go
@@ -283,12 +283,12 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 				if gatewayCanSendPong {
 					missedPongsMu.Lock()
 					missedPongs++
+					missedPongsMu.Unlock()
 					if missedPongs == s.cfg.MissedPongThreshold {
 						logger.WithError(errMissedTooManyPongs).Warn("Disconnect gateway")
-						conn.Disconnect(errMissedTooManyPongs)
+						conn.Disconnect(errMissedTooManyPongs.New())
 						return
 					}
-					missedPongsMu.Unlock()
 				}
 			case <-timeSyncTickerC:
 				b, err := s.formatter.TransferTime(ctx, time.Now(), conn)
--- a/pkg/gatewayserver/io/ws/ws.go
+++ b/pkg/gatewayserver/io/ws/ws.go
@@ -43,8 +43,9 @@ import (
 )
 
 var (
-	errGatewayID      = errors.DefineInvalidArgument("invalid_gateway_id", "invalid gateway ID `{id}`")
-	errNoAuthProvided = errors.DefineUnauthenticated("no_auth_provided", "no auth provided `{uid}`")
+	errGatewayID          = errors.DefineInvalidArgument("invalid_gateway_id", "invalid gateway ID `{id}`")
+	errNoAuthProvided     = errors.DefineUnauthenticated("no_auth_provided", "no auth provided `{uid}`")
+	errMissedTooManyPongs = errors.Define("missed_too_many_pongs", "gateway missed too many pongs")
 )
 
 type srv struct {
@@ -137,10 +138,16 @@ func (s *srv) handleConnectionInfo(c echo.Context) error {
 var euiHexPattern = regexp.MustCompile("^eui-([a-f0-9A-F]{16})$")
 
 func (s *srv) handleTraffic(c echo.Context) (err error) {
-	id := c.Param("id")
-	auth := c.Request().Header.Get(echo.HeaderAuthorization)
-	ctx := c.Request().Context()
-	eps := s.formatter.Endpoints()
+	var (
+		id                 = c.Param("id")
+		auth               = c.Request().Header.Get(echo.HeaderAuthorization)
+		ctx                = c.Request().Context()
+		eps                = s.formatter.Endpoints()
+		gatewayCanSendPong bool
+		missedPongs        int
+		// Ping-pong is expected to be sychronous. However, it's theoretically possible for a pong to arrive when we're about to send a ping.
+		missedPongsMu sync.RWMutex
+	)
 
 	ctx = log.NewContextWithFields(ctx, log.Fields(
 		"endpoint", eps.Traffic,
@@ -243,6 +250,10 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 
 	// Not all gateways support pongs to the server's pings.
 	ws.SetPongHandler(func(data string) error {
+		gatewayCanSendPong = true
+		missedPongsMu.Lock()
+		missedPongs = 0
+		missedPongsMu.Unlock()
 		logger.Debug("Received pong from gateway")
 		return nil
 	})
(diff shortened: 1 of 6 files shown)


pkg/gatewayserver/io/ws/ws.go
```diff
@@ -226,9 +226,26 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
 		err = nil // Errors are sent over the websocket connection that is established by this point.
 	}()
 
+	// pingTicker is used to send periodic pings to the gateway.
+	// Ping direction; Server --> Gateway.
 	pingTicker := time.NewTicker(s.cfg.WSPingInterval)
 	defer pingTicker.Stop()
 
+	// pongTimeoutTicker is used to check when gateways don't send a pong to the server ping.
+	pongTimeoutTicker := NewCustomTicker(s.cfg.WSPingInterval * 2)
```
KrishnaIyer: The ticker interval `s.cfg.WSPingInterval * 2` is a magic number based on the fact that we should disconnect only after missing 2 consecutive pings. I didn't see any reason for making it more than that. We could make this configurable if necessary.

pkg/gatewayserver/io/ws/ticker.go
```diff
@@ -0,0 +1,54 @@
...
+
+package ws
+
+import "time"
+
+// CustomTicker wraps time.Ticker with some additional methods.
+type CustomTicker struct {
+	d    time.Duration
+	tick *time.Ticker
+	c    <-chan time.Time
```
adriansmares: Reading from the `nil` channel blocks indefinitely, so it is safe for `C()` to just return `nil` (or `tick.C`).

pkg/gatewayserver/io/ws/ticker.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// NewCustomTicker creates a new custom CustomTicker.
+// The ticker isn't started until the first time Reset is called.
+func NewCustomTicker(d time.Duration) *CustomTicker {
```
adriansmares: Would be nice to mention that this implementation is not safe for concurrent use (it doesn't have to be - ping and pong handlers are called from `ReadMessage`).

pkg/gatewayserver/io/ws/ws.go
adriansmares: The handlers are within the same goroutine (they get executed in `ReadMessage`) so you can actually just use a counter and the existing ping ticker.

Would the following also work ?
A boolean that maintains if we ever saw a pong from the gateway, and a counter of missed pongs. If the gateway exceeds the limit of pongs, and we did see it respond to them in the past, disconnect on the next ping. On pong, we reset the fault counter.

adriansmares (commented):
Some small comments - I think the implementation can be smaller.

pkg/gatewayserver/io/ws/ws.go
johanstokking: > The handlers are within the same goroutine (they get executed in `ReadMessage`) so you can actually just use a counter and the existing ping ticker.
> 
> Would the following also work ? A boolean that maintains if we ever saw a pong from the gateway, and a counter of missed pongs. If the gateway exceeds the limit of pongs, and we did see it respond to them in the past, disconnect on the next ping. On pong, we reset the fault counter.

I also think this is a simpler way and this is also how I initially envisaged this to work.

KrishnaIyer:
The reason for extracting the timer to a separate interface is that this file is already full of conditions/variables/branches.
I don't really mind rewriting this based on the suggestion, so here you go.

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -269,6 +280,16 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
...
 				}
+				if gatewayCanSendPong {
+					missedPongsMu.Lock()
+					missedPongs++
+					if missedPongs == s.cfg.MissedPongThreshold {
+						logger.WithError(errMissedTooManyPongs).Warn("Disconnect gateway")
+						conn.Disconnect(errMissedTooManyPongs)
+						return
+					}
+					missedPongsMu.Unlock()
```
adriansmares: `Unlock` after increment, otherwise you don't unlock when you disconnect the gateway. Also `.New`.

KrishnaIyer: The mutex is local to the function stack and will get GCed when we exit the function but this is good.

adriansmares: If due to some cosmic misfortune `ReadMessage` was running the pong handler while you return an error here, the main goroutine would not have returned, wouldn't it ? It would be stuck on `Lock`.

KrishnaIyer: That's exceedingly rare but not impossible, so it's good to do this.

adriansmares (approved):

KrishnaIyer:
Anything to add @johanstokking ?

johanstokking (approved):

PR #4865, opened 2021-11-12
Support `ftime` and `foff` for SX1303 using V1 UDP protocol

johanstokking:
#### Summary

The SX1303 produces fine timestamps and frequency offset, yet may use the legacy UDP protocol version 1. See https://github.com/Lora-net/sx1302_hal/blob/4b42025d1751e04632c0b04160e0d29dbbb222a5/packet_forwarder/PROTOCOL.md?plain=1#L129-L147

#### Changes

- This adds support to reading `ftime` and `foff` if they are set
- The V2 format (`rsig` set) still takes precedence for these values
- Few bugfixes along the way; coding rate wasn't reset properly in the loop and coding rate is taken for LR-FHSS

#### Testing

Unit testing should do it.

##### Regressions

We don't send `rxpk` to the gateway so there's no compatibility issue expected.

... unless, of course, some gateways are already sending values with these fields and they don't use the `uint32` and `int32` types respectively.

#### Notes for Reviewers

@htdvisser no need for review
@adriansmares please review instead

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Added
 
+- Support for fine timestamps and frequency offsets sent by gateways with SX1303 concentrator using the legacy UDP protocol.
+
 ### Changed
 
 ### Deprecated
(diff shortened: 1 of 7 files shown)


adriansmares (approved):
:tada:

PR #4867, opened 2021-11-15
Remove peer identifier stickiness

adriansmares:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/2974

#### Changes

- Remove peer identifiers stickiness in all of the components, except Gateway Claims (for downlink claims)
  - The context remains.

#### Testing

Unit testing, and some extra testing on `staging1`.

##### Regressions

None are expected. 

#### Checklist

### DIFF
--- a/pkg/applicationserver/applicationserver.go
+++ b/pkg/applicationserver/applicationserver.go
@@ -545,7 +545,7 @@ func (as *ApplicationServer) downlinkQueueOp(ctx context.Context, ids ttnpb.EndD
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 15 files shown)


htdvisser (approved):

johanstokking (approved):
Should we keep that third argument?

adriansmares:
> Should we keep that third argument?

Yes - it is used by claims (so NS basically selects the GS peer that has the downlink path claimed). For other call sites it is indeed not as useful.

htdvisser:
Good point actually. Maybe we can refactor the cluster interface some time to take (variadic) options to modify peer selection behavior. I've added it to my Technical-Debt-Thursday list.

PR #4873, opened 2021-11-16
Support resetting device state in the Console

### LINKED ISSUE #2434: Resetting frame counts for end devices in the Console

bafonins:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2434

<details>

Screenshot 2021-11-17 at 17 28 31
Screenshot 2021-11-17 at 17 28 22
Screenshot 2021-11-17 at 16 57 27

</details>

#### Changes

- Extend api definition to include `ResetFactoryDefaults`
- Add reset modal to NS section in device general settings

#### Testing

Manual testing on staging with OTAA and ABP devices

#### Notes for Reviewers

~Currently, when an OTAA end device is reset we reset last seen, session keys and frame counter values on the frontend since the end device has to rejoin. For ABP I noticed that only downlink count is reset, while `last_f_cnt_up` is still present in the payload returned from NS. Is this correct behavior~ @adriansmares @johanstokking ?

The console resets up/down counters and the last seen value when resetting end device state.

#### Checklist


adriansmares:
I think that for `supports_join` = `false` devices, `session.last_a_f_cnt_down` should also be reset to `0` in the AS when this button is pressed (and perhaps via CLI too). The reason is that since the AS is not aware of the reset, it will start counting the downlinks from the previous downlink frame counter, which is a huge waste.

Also perhaps we should reset the currently known JS nonces ? `used_dev_nonces` / `last_dev_nonce` 

Edit: I've opened https://github.com/TheThingsNetwork/lorawan-stack/issues/4874 about the other stores. Let's keep this PR only about the NS MAC state reset.

pkg/webui/console/store/reducers/devices.js
```diff
@@ -104,6 +105,27 @@ const devices = (state = defaultState, { type, payload, event }) => {
...
+      const isOTAA = Boolean(device.supports_join)
+
+      const resetDevice = { ...device }
+      if (isOTAA && resetDevice.session) {
+        // Reset session keys and last seen information for joined OTAA end devices.
+        resetDevice.session.keys = undefined
+        return mergeDerived({ ...state, [combinedId]: resetDevice }, combinedId, {
+          lastSeen: undefined,
+          uplinkFrameCount: undefined,
+          downlinkFrameCount: undefined,
```
kschiffer: Let's set this within `defaultState.derived` and then re-reference it here.

pkg/webui/console/store/reducers/devices.js
```diff
@@ -104,6 +105,27 @@ const devices = (state = defaultState, { type, payload, event }) => {
...
+        })
+      }
+
+      return mergeDerived(state, combinedId, {
+        lastSeen: undefined,
+        uplinkFrameCount: undefined,
+        downlinkFrameCount: undefined,
```
kschiffer: See above.

pkg/webui/console/views/device-general-settings/network-server-form/index.js
```diff
@@ -51,14 +54,22 @@ import {
 
 import validationSchema from './validation-schema'
 
+const m = defineMessages({
+  resetTitle: 'Reset to factory defaults',
```
```suggestion
  resetTitle: 'Reset session and MAC state',
```

pkg/webui/console/views/device-general-settings/network-server-form/index.js
```diff
@@ -51,14 +54,22 @@ import {
...
+  resetSuccess: 'End device reset',
+  resetFailure: 'There was an error resetting the end device',
```
```suggestion
  resetFailure: 'There was an error and the end device session and MAC state could not be reset',
```

pkg/webui/console/views/device-general-settings/network-server-form/index.js
```diff
@@ -51,14 +54,22 @@ import {
...
+  modalMessage:
+    'Are you sure you want to reset session context and MAC state of this end device? For OTAA all data is wiped and end device must rejoin, for ABP session keys, device address and downlink queue are preserved, while MAC state is reset.',
```
```suggestion
    'Are you sure you want to reset the session context and MAC state of this end device?<br /><br />This will have the following consequences:<ul><li>For <DocLink>OTAA</DocLink>-activated end devices <b>all session and MAC data is wiped and the end device MUST rejoin</b></li><li>For <DocLink>ABP</DocLink>-activated end devices, session keys, device address and downlink queue are preserved, while <b>the MAC state is reset</b></li></ul>',
```

Let's be as clear as possible here. @htdvisser @adriansmares please review this sentence and let us know if anything else should be brought to the users attention when resetting.

`DocLink` should point to `/devices/abp-vs-otaa`

pkg/webui/console/views/device-general-settings/network-server-form/index.js
```diff
@@ -348,6 +374,16 @@ const NetworkServerForm = React.memo(props => {
...
+        <ModalButton
+          type="button"
+          warning
+          naked
+          message={m.resetTitle}
+          modalData={{
+            message: m.modalMessage,
+          }}
+          onApprove={handleMacReset}
+        />
```
kschiffer: Let's add this as a regular button within the form rather. I'm afraid that people will miss or misinterpret the button when it is part of the submit bar. Also, using a regular field will enable us to use a tooltip description.

You could use a `<Form.InfoField />` for this. It should have the label `Session and MAC state reset`.

Additionally, please add a tooltip that reads as follows:

> Resetting the session context and MAC state will reset the end device to its initial (factory) state. This includes \<b>resetting the frame counters\<b/> and any other persisted MAC setting on the end device. Over The Air Activation (OTAA) end devices will also lose their session context. This means that such devices \<b>must rejoin\</b> the network to continue operation. Activation-by-personalization (ABP) end devices will only reset the MAC state, while preserving up/downlink queues.

Again, @htdvisser, @adriansmares please double check the wording here.

Additionally, in this case, we should use a field description reading:

> You can use this e.g. to cause a frame counter reset

kschiffer (changes requested):
Works well. I only have some wording/structure remarks to ensure good UX.

pkg/webui/console/views/device-general-settings/network-server-form/index.js
adriansmares: Text LGTM, but `Over The Air Activation (OTAA)-activated` is extremely verbose, also we don't give the full name for ABP.

Maybe show `Over The Air Activation` / `Activation By Personalization` on hover ?

kschiffer: I've updated this to include a link to the documentation about activation modes and then use the abbreviation only.

pkg/webui/console/store/reducers/devices.js
```diff
@@ -104,6 +105,23 @@ const devices = (state = defaultState, { type, payload, event }) => {
...
+      const isOTAA = Boolean(device.supports_join)
+
+      const resetDevice = { ...device }
+      if (isOTAA && resetDevice.session) {
+        // Reset session keys and last seen information for joined OTAA end devices.
+        resetDevice.session.keys = undefined
+        return mergeDerived(
+          { ...state, [combinedId]: resetDevice },
+          combinedId,
+          defaultState.derived,
```
kschiffer:
```suggestion
          defaultState.derived,
```

Haven't tested but from seeing this, this should not work since you're now spreading an empty object over the existing one, which should have no effect. Or am I missing something?

kschiffer:
```javascript
const defaultState = { … }
const defaultDerived = {
  lastSeen: undefined,
  uplinkFrameCount: undefined,
  downlinkFrameCount: undefined,
}
…
// Within the reducer:
return mergeDerived(
  { ...state, [combinedId]: resetDevice },
  combinedId,
  defaultDerived,
}
```

kschiffer (approved):
See my remark on the derived value.

PR #4831, opened 2021-11-05
Event filtering using the backend

kschiffer:
#### Summary
This PR introduces filtering logic for events in the backend, as well as the Console making use of the new feature. Client-side filtering is hence retired.

Referencing #2231 

#### Changes
- Add server-side filtering via the `names` prop
- Removing client-side filtering from the Console
- Updating structure of event filters, and associated handling logic in the Console
  - Using specific per-entity filters to avoid server overhead
  - Using a more sustainable

#### Testing

Manual testing.

##### Regressions

Possible but hard to check right now. We should check this closely once on staging.

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [x] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #4881, opened 2021-11-17
Respect event request tail if after is unspecified

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/4831/files#diff-b5098da68ed744be59ce6adb1993e9e522e75c14c7f63b12992157fd643d7bddR144

#### Changes

- Return only newer events if a tail is not specified

#### Testing

Local testing.

##### Regressions

This was a regression in itself.

#### Checklist

### DIFF
--- a/pkg/events/grpc/grpc.go
+++ b/pkg/events/grpc/grpc.go
@@ -141,7 +141,7 @@ func (srv *EventsServer) Stream(req *ttnpb.StreamEventsRequest, stream ttnpb.Eve
 	store, hasStore := srv.pubsub.(events.Store)
 	var group *errgroup.Group
 	if hasStore {
-		if req.After == nil {
+		if req.After == nil && req.Tail == 0 {
 			now := time.Now()
 			req.After = &now
 		}


pkg/events/grpc/grpc.go
```diff
@@ -141,7 +141,7 @@ func (srv *EventsServer) Stream(req *ttnpb.StreamEventsRequest, stream ttnpb.Eve
 	store, hasStore := srv.pubsub.(events.Store)
 	var group *errgroup.Group
 	if hasStore {
-		if req.After == nil {
+		if req.After == nil && req.Tail == 0 {
```
adriansmares, line 144: I initially wanted to come back to the old diff, where `srv.pubsub.Subscribe` would have been used if both of these are `nil`/`0`, but I think that there was some thought behind the original change. 

Is listening via the streams cheaper / more optimal than using the PubSub ? I can see that happening since we don't force the whole instance to subscribe to these events, but at the same time we allow an unbounded amount of connections to the underlying storage.

htdvisser: My main reasoning was that I wanted to follow the same subscription method for all subscriptions in a deployment.

Secondary reason was that with the streams (instead of the pub/sub) we can do some rate limiting on each stream. We wouldn't be able to do that with the other subscription method because it relies on a single `redis.(*PubSub).subscribeTask` routine.

htdvisser (approved):

PR #4884, opened 2021-11-17
Non-react route error template

### LINKED ISSUE #3261: Provide a static error HTML template for non-react handled routes
#### Summary
Following up from https://github.com/TheThingsNetwork/lorawan-stack/issues/233 we still need a Go HTML template that will render a generic and static error page for all errors that are not in Console or OAuth app context. See also: https://github.com/TheThingsIndustries/lorawan-stack/issues/1315#issuecomment-440656553

#### Why do we need this?
To display proper error pages for backend errors that are not in the scope of Console or OAuth app. E.g. https://github.com/TheThingsIndustries/lorawan-stack/issues/2268

Error page designs and error boundaries that render error views for routes that are handled by our frontend react apps.

Templates that render static HTML/CSS for any route that is not handled by our frontend react apps.

Implementing the design that we currently use for the react apps as static HTML/CSS template that renders for error responses.

Manual.

Yes but anyone please feel free to pick up.

pgalic96:
#### Summary

Closes #3261 

#### Changes

- Add Go error template
- Move assets necessary for rendering the error template to a separate directory used by the template
- Fix bug where not found routes were not handled properly in the main router

#### Testing

...

##### Regressions

None

#### Checklist


kschiffer (approved):

pkg/webhandlers/error_template.go
```diff
@@ -0,0 +1,210 @@
...
+	BackendErrorDetails string
+	Year                int
+	IsGenericNotFound   bool
+}
+
+const errorHTML = `
+<!doctype html>
+<html lang="en">
+  <head>
+    <title>Error - The Things Stack Enterprise</title>
```
johanstokking: We're in open source here.

Can we embed a file with embedded files? https://pkg.go.dev/embed

pkg/webhandlers/error_template.go
```diff
@@ -0,0 +1,210 @@
...
+                  <span class="logo">file_copy</span>
+                  Copy to clipboard
+                </button>
+              </details>
+              {{ end }}
+            </div>
+          </div>
+        </div>
+      </div>
+      <footer class="tti-footer">
```
johanstokking: Same

pgalic96:
The reason for replacing `RootRouter` with main `Router` is that the `Router` handles the `"/"` route, but it doesn't handle `NotFound` error. Adding `NotFoundHandler` to the `Router` causes all the other root routes (`/api`, `/assets` , `/debug/pprof`, etc.) to result in `NotFound`, since they are registered at the `RootRouter`. Registering all these routes directly at the main `Router` seems to resolve all the failing tests, but please let me know if this could cause some other issues.

htdvisser:

johanstokking:
I see many review requests. I did a pass on code style only falready, as I have little knowledge about the subtleties around the mux routers, intention and registered middleware, nor about the html and css. If you or reviewers need my opinion, mention me explicitly.

pkg/component/web.go
```diff
@@ -72,7 +72,7 @@ func (c *Component) initWeb() error {
 	}
 
 	if c.config.HTTP.PProf.Enable {
-		g := web.RootRouter().NewRoute().Subrouter()
+		g := web.Router().NewRoute().Subrouter()
```
htdvisser: We should keep using the root router here, otherwise this will break in Enterprise. Same below.

pkg/web/web.go
```diff
@@ -251,7 +252,7 @@ func New(ctx context.Context, opts ...Option) (*Server, error) {
 			}),
 		),
 	)
-	root.PathPrefix("/api/").Handler(apiRouter)
+	router.PathPrefix("/api/").Handler(apiRouter)
```
htdvisser: This should stay on the root router, otherwise we risk breaking some Enterprise APIs.

pkg/webhandlers/error_template.go
```diff
@@ -0,0 +1,89 @@
...
+	ErrorTitle          string
+	ErrorMessage        string
+	ErrorID             string
+	CorrelationID       string
+	BackendErrorDetails string
+	Year                int
+	IsGenericNotFound   bool
+}
+
+//go:embed "error_template.html"
```
htdvisser: The convention is to use `.html.tmpl` extension.

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -82,7 +82,7 @@ func New(ctx context.Context, server io.Server, formatter Formatter, cfg Config)
 		return nil, err
 	}
 
-	router := web.RootRouter()
+	router := web.Router()
```
htdvisser: I don't know if we can just do this, it may break in Enterprise.

adriansmares: Indeed - API endpoints do not need this kind of rendering.

adriansmares (approved):

htdvisser (approved):

PR #4891, opened 2021-11-18
Allow frontends to manually synchronize gateway clocks

### LINKED ISSUE #4853: Gateway absolute time synchronization should taken into consideration timestamp granularity
#### Summary

Our current Gateway absolute time synchronization logic is rather unapologetic - if `TxSettings.Time` is available, we consider that to be a gateway absolute timestamp:
https://github.com/TheThingsNetwork/lorawan-stack/blob/0b83c294b96458a47382fcdd373292187f6763f4/pkg/gatewayserver/io/io.go#L210-L224

We should not synchronize the gateway time if the accuracy is not known to be close to millisecond.

#### Why do we need this?

The problem is that this `Time` has undefined accuracy. TTKG has second accuracy associated with the reported `Time` - there is no fractional second. This will cause problems when the Network Server has to schedule an absolute time downlink without a pre-defined gateway. It will consider the recent downlink paths, and it may choose this TTKG. The TTKG has little chances of actually respecting the absolute timestamp, although it will try.

The Gateway Server always considers the `TxSettings.Time` to be a fine grained timestamp.

The protocol frontends should communicate explicitly this fine grained timestamp to the `io.Server`. The frontends should also get to choose if synchronization should take place at all, since some frontends (i.e. BasicStation) handle synchronization on their own.

The NS also should have access to a fine grained timestamp somehow. I think that `Time` should be reserved only for GPS timestamps.

#### Environment

`v3.16`

Add some options/settings to `io.Server.HandleUp` that allow the frontend to specify the time settings explicitly, or require that `TxSettings.Time` is always a GPS timestamp.

UT.

Yes.

adriansmares:
#### Summary

Closes #4853

#### Changes

- Gateway Server frontends may now manually provide clock synchronization information
  - We no longer sync the Basic Station frontend twice (once in the frontend, once in `io.HandleUp`)
- BasicStation `timetransfer` is now used only to transfer `MuxTime`
  - See https://github.com/TheThingsNetwork/lorawan-stack/issues/4852 for how GPS time transfer should actually work
  - Synchronizing `MuxTime` is enough in order to actually have accurate `RxTime` and `RefTime` (modulo RTT variations), which is enough for our usage

#### Testing

Unit testing.

##### Regressions

Not expected.

#### Notes for Reviewers

We already didn't include `TxSettings.Time` for the MQTT v2 protocol, so the granularity mentioned in the original issue is not a huge problem. For the rest of the frontends, they are in the order of milliseconds, which is enough for absolute time downlinks.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/io.go
+++ b/pkg/gatewayserver/io/io.go
@@ -201,26 +201,48 @@ var errBufferFull = errors.DefineInternal("buffer_full", "buffer is full")
 // Interval between emitting consecutive gs.up.repeat events for the same gateway connection.
 const consecutiveRepeatUpEventsInterval = time.Minute
 
+// ManualClockSynchronization contains the clock synchronization
+// timestamps provided by a frontend for manual synchronization.
+type ManualClockSynchronization struct {
+	Timestamp        uint32
+	ServerTime       time.Time
+	GatewayTime      *time.Time
+	ConcentratorTime scheduling.ConcentratorTime
+}
+
 // HandleUp updates the uplink stats and sends the message to the upstream channel.
-func (c *Connection) HandleUp(up *ttnpb.UplinkMessage) error {
+func (c *Connection) HandleUp(up *ttnpb.UplinkMessage, manualSync *ManualClockSynchronization) error {
 	if c.discardRepeatedUplink(up) {
 		return nil
 	}
 
 	var ct scheduling.ConcentratorTime
-	if up.Settings.Time != nil {
+	switch {
+	case manualSync != nil:
+		ct = c.scheduler.SyncWithGatewayConcentrator(manualSync.Timestamp, manualSync.ServerTime, manualSync.GatewayTime, manualSync.ConcentratorTime)
+		log.FromContext(c.ctx).WithFields(log.Fields(
+			"timestamp", manualSync.Timestamp,
+			"concentrator_time", manualSync.ConcentratorTime,
+			"server_time", manualSync.ServerTime,
+			"gateway_time", manualSync.GatewayTime,
+		)).Info("Gateway clocks have been synchronized by the frontend")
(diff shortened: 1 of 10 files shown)


johanstokking (commented):
Manual means by hand. What do you mean with manual here? That the time is determined by the frontend protocol?

adriansmares:
> Manual means by hand. What do you mean with manual here? That the time is determined by the frontend protocol?

In my mind, automatic is that the `io` package determines the clock values automatically from `TxSettings`/`up.ReceivedAt`/rollover detection, while manual means that the frontend decides what values to use while doing the synchronization. I agree that this is forcing it a bit - do you have a better suggestion ?

pkg/gatewayserver/io/io.go
```diff
@@ -201,26 +201,48 @@ var errBufferFull = errors.DefineInternal("buffer_full", "buffer is full")
...
-	if up.Settings.Time != nil {
+	switch {
+	case manualSync != nil:
+		ct = c.scheduler.SyncWithGatewayConcentrator(manualSync.Timestamp, manualSync.ServerTime, manualSync.GatewayTime, manualSync.ConcentratorTime)
+		log.FromContext(c.ctx).WithFields(log.Fields(
+			"timestamp", manualSync.Timestamp,
+			"concentrator_time", manualSync.ConcentratorTime,
+			"server_time", manualSync.ServerTime,
+			"gateway_time", manualSync.GatewayTime,
+		)).Info("Gateway clocks have been synchronized by the frontend")
```
johanstokking, line 228: So then `pkg/gatewayserver/io.FrontendClockSynchronization` sounds like a great name to me

johanstokking (approved):
Approved with one rename request

KrishnaIyer (approved):
Nothing more to add

PR #4892, opened 2021-11-18
Check if dev SQL database is running after starting it

### LINKED ISSUE #4911: Update `webui` e2e tests documentation
#### Summary

After https://github.com/TheThingsNetwork/lorawan-stack/pull/4813 we changed the way tests are set up.
If one follows setup steps from [docs](https://github.com/TheThingsNetwork/lorawan-stack/blob/bab6bb5414345708d7e42077bab750623d03111e/DEVELOPMENT.md#end-to-end-tests) `cypress` displays the following error:
```
Error: ERROR:  template database "tts_seed" does not exist

    at Socket.<anonymous> (/Users/bog/golang/src/go.thethings.network/lorawan-stack/cypress/plugins/tasks.js:105:17)
    at Socket.emit (events.js:315:20)
    at addChunk (_stream_readable.js:295:12)
    at readableAddChunk (_stream_readable.js:271:9)
    at Socket.Readable.push (_stream_readable.js:212:10)
    at Pipe.onStreamRead (internal/stream_base_commons.js:186:23)
```

#### Why do we need this?

We need complete documentation on how to run e2e tests locally

Incomplete documentation

How to avoid the error:
```
Error: ERROR:  template database "tts_seed" does not exist

    at Socket.<anonymous> (/Users/bog/golang/src/go.thethings.network/lorawan-stack/cypress/plugins/tasks.js:105:17)
    at Socket.emit (events.js:315:20)
    at addChunk (_stream_readable.js:295:12)
    at readableAddChunk (_stream_readable.js:271:9)
    at Socket.Readable.push (_stream_readable.js:212:10)
    at Pipe.onStreamRead (internal/stream_base_commons.js:186:23)
```

@kschiffer please assign milestone

kschiffer:
#### Summary
This quickfix PR will ensure that the database is running after running `tools/bin/mage dev:dbStart`, so that other targets that depend on the database can run right after it.

Closes #4911

#### Changes
- Add a connection check to `tools/bin/mage dev:dbStart` using `pg_isready`
- Mute `dropdb` output
- Update documentation accordingly to clarify usage of `dev:sqlCreateSeedDb`

#### Testing

Manual and CI

#### Notes for Reviewers
For reasons that I wasn't able to figure out, `pg_isready` does not run in CI but will always return an error code 1. Additionally, the problem of the postgres server not running immediately after the `dev:dbStart` target seem to be only present in local environments, not the CI. If anyone has an idea why this is, we might be able to avoid having to distinguish between CI in the mage target.

#### Checklist

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -787,7 +787,7 @@ We use [Cypress](https://cypress.io) for running frontend-based end-to-end tests
 
 #### Running frontend end-to-end tests locally
 
-Make sure to [build the frontend assets](#building-the-frontend), [start The Things Stack](#starting-the-things-stack) and run `tools/bin/mage dev:sqlDump` after running `tools/bin/mage dev:initStack` to save database dump used as database seed when running end-to-end tests. To run the stack when working on end-to-end tests, use the `tools/bin/mage dev:startDevStack` command. This will run the stack in proper configuration for the end-to-end tests, note that this is an endless process that essentially runs the The Things Stack process so you don't have to wait for it to finish.
+Make sure to [build the frontend assets](#building-the-frontend), [start The Things Stack](#starting-the-things-stack) and run `tools/bin/mage dev:initStack dev:sqlDump dev:sqlCreateSeedDb` to create a seed database which the tests can reset the database to in between runs. To run the stack when working on end-to-end tests, use the `tools/bin/mage dev:startDevStack` command. This will run the stack in proper configuration for the end-to-end tests, note that this is an endless process that essentially runs the The Things Stack process in the background so you don't have to wait for it to finish.
 
 `Cypress` provides two modes for running tests: headless and interactive.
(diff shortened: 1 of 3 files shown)


adriansmares:
> does not run in CI but will always return an error code 1

Per https://www.postgresql.org/docs/9.3/app-pg-isready.html - `1 if the server is rejecting connections (for example during startup)`

This was returned even after the 30 second wait time ?

htdvisser (commented):
We never had any issues with Postgres not being ready in the Go workflow. Can't we just do the same here?

https://github.com/TheThingsNetwork/lorawan-stack/blob/319669a28dfef56da14f0395ab9348d097a1fb19/.github/workflows/go.yml#L60-L90

kschiffer:
> This was returned even after the 30 second wait time ?

Yes.

> Can't we just do the same here?

I think that would be hard because we need to interact with the container during cypress runs to reseed the DB. It's currently all based on the docker image that we spin up using mage.

htdvisser (approved):

adriansmares (approved):

Issue #2627, opened 2020-05-29
Remove dependency on Echo


htdvisser:
#### Summary

This issue tracks the removal of our dependency on Echo. We've already done some work, but we didn't have an issue to track it.

#### Why do we need this?

https://github.com/TheThingsNetwork/lorawan-stack/pull/2382#issuecomment-621032344

#### What is already there? What do you see now?

A lot of work already done in https://github.com/TheThingsNetwork/lorawan-stack/pull/2382 and https://github.com/TheThingsNetwork/lorawan-stack/pull/2561

More upcoming removals in https://github.com/TheThingsNetwork/lorawan-stack/pull/2564

Some other pull requests removing other parts

#### What is missing? What do you want to see?

Complete removal

htdvisser:
Let's add this to the quarterly milestone. We haven't been able to upgrade Echo since v4.1.2 (released Apr 29, 2019) which is a really long time. At the same time, we don't actually _need_ Echo, because we also use Gorilla mux router, so let's just remove what's left of Echo.

johanstokking:
So we have:

- `pkg/account`, `pkg/oauth`, `pkg/web/oauthclient`: @kschiffer
- `pkg/basicstation/cups`: @KrishnaIyer 
- `pkg/console` and `pkg/webui`: @pgalic96 (touched these files last week and usage is very minimal)
- `pkg/gatewayserver/io/ws`: @KrishnaIyer 
- `pkg/web/middleware`: @pgalic96 (needs non-echo variants)

General cleanup, may depend on the above (?)
- `pkg/errors/web`: @htdvisser 
- `pkg/ratelimit`: @htdvisser (this is probably just deleting the middleware that has become obsolete)
- `pkg/util/test/assertions`: @htdvisser (independent; the test assertions aren't being used)
- `pkg/web`: @htdvisser and/or @johanstokking (definitely depends on removing usage above)

htdvisser:
> `pkg/web/middleware`: @pgalic96 (needs non-echo variants)

Those should all already exist in `pkg/webmiddleware`.

adriansmares:
>     * `pkg/basicstation/cups`: @KrishnaIyer
>     * `pkg/gatewayserver/io/ws`: @KrishnaIyer

I'll steal this.

PR #4893, opened 2021-11-18
Remove BasicStation-related Echo usage

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2627

#### Changes

- Remove Echo usage in CUPS
- Remove Echo usage in BasicStation

#### Testing

Unit testing.

##### Regressions

Unit tests should cover this.

#### Checklist

### DIFF
--- a/pkg/basicstation/cups/update_info.go
+++ b/pkg/basicstation/cups/update_info.go
@@ -19,6 +19,7 @@ import (
 	"crypto"
 	"crypto/rand"
 	"crypto/sha512"
+	"encoding/json"
 	"fmt"
 	"hash/crc32"
 	"net"
@@ -34,6 +35,7 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/rpcmetadata"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 	"go.thethings.network/lorawan-stack/v3/pkg/unique"
+	"go.thethings.network/lorawan-stack/v3/pkg/webhandlers"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
 )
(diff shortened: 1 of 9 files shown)


pkg/basicstation/cups/update_info.go
```diff
@@ -141,29 +143,29 @@ var getGatewayMask = pbtypes.FieldMask{Paths: []string{
 }}
 
 // UpdateInfo implements the CUPS update-info handler.
-func (s *Server) UpdateInfo(c echo.Context) (err error) {
-	// This is to account for older LBS gateways that don't set this header.
-	c.Request().Header.Set(echo.HeaderContentType, "application/json")
```
johanstokking, line 146: We don't need this anymore?

johanstokking (approved):

pkg/basicstation/cups/update_info.go
adriansmares: This was there for Echo's `Bind`, such that it knows that it should unmarshal the body using the JSON unmarshaler. We manually unmarshal the body now, so there is no need to pre-fill this anymore.

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -60,91 +58,94 @@ func (s *srv) Protocol() string            { return "ws" }
...
 	_, data, err := ws.ReadMessage()
 	if err != nil {
 		logger.WithError(err).Debug("Failed to read message")
-		return err
+		return
 	}
 
 	scheme := "ws"
-	if c.IsTLS() || s.cfg.UseTrafficTLSAddress {
+	if s.cfg.UseTrafficTLSAddress {
```
KrishnaIyer: The only reason for having this option `UseTrafficTLSAddress` was if TLS was terminated in a load balancer. 

We check if the connection is TLS  for SH users who directly use the stack ACME (PAYG/BYOL). `UseTrafficTLSAddress` is false in those templates so we will break those deployments.

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -254,8 +254,13 @@ func (s *srv) handleTraffic(c echo.Context) (err error) {
...
 	}
 
-	go func() {
+	go func() (err error) {
 		defer ws.Close()
+		defer func() {
+			if err != nil {
+				conn.Disconnect(err)
+			}
+		}()
```
KrishnaIyer, line 269: the `ws.Close()` causes the uplink flow to disconnect since `ws.ReadMessage` will error. This returns to the defer in `L217` which does a disconnect. I think that this explicit disconnection is not needed. 
If you mean to sync the downstream and upstream loops, i.e., exit the upstream loop on downstream error explicitly, then the upstream loop needs to listen on the connection context right?

pkg/gatewayserver/io/ws/ws.go
adriansmares: Good catch. Adding `r.TLS != nil` (that's the Echo definition for `IsTLS`).

pkg/gatewayserver/io/ws/ws.go
adriansmares: The intent here is to have the error of the downstream messages propagated to the error of the connection itself. If we do only `ws.Close()`, then the error that cancels the connection context will be that the connection has been closed, not that an error was encountered during the downstream flow. Does this make sense ?

KrishnaIyer: Ok yeah that sounds good.

KrishnaIyer (approved):

htdvisser (approved):

kschiffer (approved):
Codeowned LGTM

KrishnaIyer (approved):
Looks good with a small rename. Can you change `noPongReceived` to `numPongReceived`?

PR #4897, opened 2021-11-19
Do not specify bandwidth for FSK

adriansmares:
#### Summary

References https://github.com/lorabasics/basicstation/issues/138

#### Changes

- Do not provide bandwidth for FSK channels in the Basic Station frontend.

#### Testing

Unit testing.

##### Regressions

Not expected.

#### Notes for Reviewers

Validation on BasicStation is a bit weird - see:

Allows bandwidth:
https://github.com/lorabasics/basicstation/blob/60fd9788763ea11ec80ec1b1f683c0374792f71e/src/sx1301v2conf.c#L228
Rejects bandwidth:
https://github.com/lorabasics/basicstation/blob/60fd9788763ea11ec80ec1b1f683c0374792f71e/src/sx1301v2conf.c#L519

#### Checklist


htdvisser:

adriansmares:
@KrishnaIyer it seems that `antenna_gain` is not legal either - see https://doc.sm.tc/station/tcproto.html#router-config-message

Apparently v2 designs will also error on the `antenna_gain` - see https://github.com/lorabasics/basicstation/blob/60fd9788763ea11ec80ec1b1f683c0374792f71e/src/sx1301v2conf.c#L466

johanstokking (commented):
Leaving this up to @KrishnaIyer

KrishnaIyer:
FSK looks good to me. For `antenna_gain`, I think we need to have a way for the user to specify the concentrator reference and switch the fields accordingly. I really don't like the idea of removing a user-requested functionality because a few rare gateways break.

adriansmares:
> FSK looks good to me. For `antenna_gain`, I think we need to have a way for the user to specify the concentrator reference and switch the fields accordingly. I really don't like the idea of removing a user-requested functionality because a few rare gateways break.

The problem is that specifying the antenna gain is not standard behavior - `station` accepts it for `v1.5` designs out of historical reasons, but our implementation is not really standard if we send it - the protocol does not allow it.

One way to go through this is to basically mark `antenna_gain` as `omitempty` and to document that setting it for gateways which are using `v2` / `corecell` designs this will not work. Does this sound sensible ?

KrishnaIyer:
> One way to go through this is to basically mark antenna_gain as omitempty and to document that setting it for gateways which are using v2 / corecell designs this will not work. Does this sound sensible ?

That sounds good to me indeed.

adriansmares:
> That sounds good to me indeed.

Updated.

KrishnaIyer (approved):
LGTM

PR #4899, opened 2021-11-22
Account for implementation errors while parsing GPS timestamps

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack-support/issues/638

#### Changes

- Account for the fact that certain BasicStation implementations use millisecond precision instead of microsecond precision for GPS timestamps
   - The [standard](https://doc.sm.tc/station/tcproto.html#radio-metadata) explicitly says that microsecond precision must be used
   - This is probably caused by the fact that the UDP packet forwarder was using [millisecond precision](https://github.com/Lora-net/packet_forwarder/blob/d0226eae6e7b6bbaec6117d0d2372bf17819c438/PROTOCOL.TXT#L134), and bridges/implementations just thought that the field has the same meaning

Logs from a particular bridge that show this error (observe that `tmms` and `gpstime` are equal, even though they should not be):
```
Nov 22 12:36:13 kona-micro-outdoor tek_bstn[2110]: [DEBUG ] bridge_data_conv.c:437 PUSH_DATA message body: {"rxpk":[{"tmst":137070715,"tmms":1321619791991,"chan":4,"rfch":0,"freq":867.300000,"stat":1,"modu":"LORA","datr":"SF7BW125","codr":"4/5","lsnr":11.0,"rssi":-42,"size":20,"data":"QOAHACeAVi0FhAfwk3ryro2909A="}]}
Nov 22 12:36:13 kona-micro-outdoor tek_bstn[2110]: [DEBUG ] bridge_bstn_lns.c:1728 DR index found: [5]
Nov 22 12:36:13 kona-micro-outdoor tek_bstn[2110]: [DEBUG ] bridge_bstn_lns.c:448 sending [353] bytes to WS stream
Nov 22 12:36:13 kona-micro-outdoor tek_bstn[2110]: [DEBUG ] bridge_bstn_lns.c:449 payload:
{ "DR": 5, "Freq": 867300000, "upinfo": { "rctx": 0, "rxtime": 1637584573, "xtime": 24769944116496507, "gpstime": 1321619791991, "rssi": -42, "snr": 11.0, "chan": 4 }, "msgtype": "updf", "MHdr": 64, "DevAddr": 654313440, "FCtrl": 128, "FCnt": 11606, "FOpts": "", "FPort": 5, "FRMPayload": "8407f0937af2ae", "MIC": -791429747, "RefTime": 1637584573.983 }
```

#### Testing

Wrote a unit test based on the log.

##### Regressions

New unit tests for `TimePtrFromUpInfo` should cover the old and new behavior.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ For details about compatibility between different releases, see the **Commitment
 - The Content-Security-Policy header (that was previously behind the `webui.csp` feature flag) is now enabled by default.
 - Default `Cache-Control: no-store` headers.
 - `Cache-Control: public, max-age=604800, immutable` headers for hashed static files.
+- Experimental support for BasicStation GPS timestamps which use the wrong precision (milliseconds instead of microseconds).
+  - The Gateway Server will attempt to determine the correct GPS timestamp from the provided `gpstime` based on the time at which the upstream message has been received.
+  - This workaround will be available until the related gateway vendors will release patches for this issue.
 
 ### Changed
 
--- a/pkg/gatewayserver/io/ws/lbslns/time.go
+++ b/pkg/gatewayserver/io/ws/lbslns/time.go
@@ -25,7 +25,7 @@ import (
 // TimeFromUnixSeconds constructs a time.Time from the provided UNIX fractional timestamp.
 func TimeFromUnixSeconds(tf float64) time.Time {
 	sec, nsec := math.Modf(tf)
-	return time.Unix(int64(sec), int64(nsec*1e9))
+	return time.Unix(int64(sec), int64(nsec*1e9)).UTC()
 }
 
 // TimeFromUnixSeconds constructs a *time.Time from the provided UNIX fractional timestamp.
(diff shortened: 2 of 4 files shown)


johanstokking:
Shouldn't we be more strict here, and just not accept this?

How realistic is that we push this problem back to where it belongs (the gateway vendors) for them to distribute a fix?

This falls under our strict API compatibility commitment because this is towards gateways. So when we merge and release this, we can't go back. It doesn't hurt, but experience with V2 showed that the migration to a new major was much easier had we been more strict. Another reason to push this back to gateway vendors is not creating a precedence of us working around their bugs.

adriansmares:
> Shouldn't we be more strict here, and just not accept this?

I'm conflicted to be honest, as we are at the tail of this. The gateway vendors fixes this after an amount of time, release this fix after another amount of time, customers read about the update some time afterwards, and they deploy it after another period of time. The end devices and gateways don't have our flexibility when it comes to deploying fixes.

> How realistic is that we push this problem back to where it belongs (the gateway vendors) for them to distribute a fix?

I've already received confirmation of this particular bug and the expectation that a hotfix will come out of this, so forward looking  we are doing well. I was concerned about existing deployments when I wrote this PR.

> This falls under our strict API compatibility commitment because this is towards gateways. So when we merge and release this, we can't go back. It doesn't hurt, but experience with V2 showed that the migration to a new major was much easier had we been more strict. Another reason to push this back to gateway vendors is not creating a precedence of us working around their bugs.

My only argument _for_ this PR is that existing deployments are going to be slow in adoption. But I can agree that in the long term maybe pushing this towards the source of the problem is better.

johanstokking (approved):
Got some extra context

KrishnaIyer (commented):
I understand why but I don't like this to be honest. This is very V2-ish.
In my opinion, we'd rather let these gateways fail and make the vendors update and push a patch rather than take on the responsibility here.

johanstokking:
Can we mark this as an experimental feature and remove in six months or so?

adriansmares:
> Can we mark this as an experimental feature and remove in six months or so?

Yes. Will do that.

PR #4906, opened 2021-11-23
Update TTKG firmware version from status message

### LINKED ISSUE #4836: Update Gateway attributes for TTKG version
#### Summary

We should update the version information of The Things Kickstarter Gateways in the Identity Server when it changes.

We update `station`, `firmware` and `package` attributes for Basic Station gateways, so that we can show gateway owners what firmware versions their gateways are running.

We don't track the firmware versions of Kickstarter gateways.

When receiving a status message with version info, compare it to the gateway attributes, and update them if changed.

Just plug in a TTKG.

KrishnaIyer:
#### Summary

Closes #4836 

#### Changes

- Update `model` and `firmware` attributes from the TTKG platform field.
- Add some delay/jitter to prevent sudden updates.
- Make sure to update only on change.
- Tests

#### Testing

UT

##### Regressions

Don't expect any.

#### Notes for Reviewers

1. The one edge case here is when we connect the gateway in `unregistered` mode. The `Update` call would then fail. We could just print the error (as this PR does). The other option is to add a `registered` flag to the connection and update only if set. Let me know which one you prefer.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/config.go
+++ b/pkg/gatewayserver/config.go
@@ -57,7 +57,7 @@ type Config struct {
 	UpdateGatewayLocationDebounceTime time.Duration `name:"update-gateway-location-debounce-time" description:"Debounce time for gateway location updates from status messages"`
 	UpdateConnectionStatsDebounceTime time.Duration `name:"update-connection-stats-debounce-time" description:"Time before repeated refresh of the gateway connection stats"`
 
-	UpdateVersionInfoDelay time.Duration `name:"update-version-info-delay" description:"Maximum time to wait to update version information. A Jitter of 100% is applied for randomization"`
+	UpdateVersionInfoDelay time.Duration `name:"update-version-info-delay" description:"Maximum time to wait to update version information. A Jitter of 25% is applied for randomization"`
 
 	Forward      map[string][]string `name:"forward" description:"Forward the DevAddr prefixes to the specified hosts"`
 	PacketBroker PacketBrokerConfig  `name:"packetbroker" description:"Packet Broker upstream configuration"`
--- a/pkg/gatewayserver/config.go
+++ b/pkg/gatewayserver/config.go
@@ -57,6 +57,8 @@ type Config struct {
 	UpdateGatewayLocationDebounceTime time.Duration `name:"update-gateway-location-debounce-time" description:"Debounce time for gateway location updates from status messages"`
 	UpdateConnectionStatsDebounceTime time.Duration `name:"update-connection-stats-debounce-time" description:"Time before repeated refresh of the gateway connection stats"`
 
+	UpdateVersionInfoDelay time.Duration `name:"update-version-info-delay" description:"Maximum time to wait to update version information. A Jitter of 100% is applied for randomization"`
+
 	Forward      map[string][]string `name:"forward" description:"Forward the DevAddr prefixes to the specified hosts"`
 	PacketBroker PacketBrokerConfig  `name:"packetbroker" description:"Packet Broker upstream configuration"`
 
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -1002,6 +1002,12 @@ func (gs *GatewayServer) handleVersionInfoUpdates(ctx context.Context, conn conn
 				"model":    status.Versions["model"],
 				"firmware": status.Versions["firmware"],
 			}
+			d := random.Jitter(gs.config.UpdateVersionInfoDelay, 1)
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(d):
+			}
 			err := gs.entityRegistry.UpdateAttributes(conn.Context(), *conn.Gateway().Ids, attributes)
 			if err != nil {
 				log.FromContext(ctx).WithError(err).Debug("Failed to update version information")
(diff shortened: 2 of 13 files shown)


pkg/gatewayserver/config.go
```diff
@@ -57,6 +57,8 @@ type Config struct {
 	UpdateGatewayLocationDebounceTime time.Duration `name:"update-gateway-location-debounce-time" description:"Debounce time for gateway location updates from status messages"`
 	UpdateConnectionStatsDebounceTime time.Duration `name:"update-connection-stats-debounce-time" description:"Time before repeated refresh of the gateway connection stats"`
 
+	UpdateVersionInfoDelay time.Duration `name:"update-version-info-delay" description:"Maximum time to wait to update version information. A Jitter of 100% is applied for randomization"`
```
adriansmares: 100% ? I'd suggest 15-25%.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -802,6 +819,7 @@ func (gs *GatewayServer) handleUpstream(ctx context.Context, conn connectionEntr
 		case msg := <-conn.Status():
 			ctx = events.ContextWithCorrelationID(ctx, fmt.Sprintf("gs:status:%s", events.NewCorrelationID()))
 			val = msg
+
```
adriansmares: Remove.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -972,6 +990,36 @@ func (gs *GatewayServer) handleLocationUpdates(ctx context.Context, conn connect
 	}
 }
 
+// handleVersionInfoUpdates updates gateway attributes with version info.
+// This function runs exactly once; only for the first status message of each connection, since version information does (should) not change within the same connection.
```
adriansmares: Just leave it as `should`.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -1012,6 +1060,7 @@ func (gs *GatewayServer) UnclaimDownlink(ctx context.Context, ids ttnpb.GatewayI
 	return gs.UnclaimIDs(ctx, &ids)
 }
 
+// ValidateGatewayID implements IO.
```
adriansmares: `io.Server`.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -972,6 +990,36 @@ func (gs *GatewayServer) handleLocationUpdates(ctx context.Context, conn connect
...
 
+// handleVersionInfoUpdates updates gateway attributes with version info.
+// This function runs exactly once; only for the first status message of each connection, since version information does (should) not change within the same connection.
+func (gs *GatewayServer) handleVersionInfoUpdates(ctx context.Context, conn connectionEntry) {
+	select {
+	case <-ctx.Done():
+		return
+	case <-conn.VersionInfoChanged():
+		status, _, ok := conn.StatusStats()
+		if ok && status.Versions["model"] != "" && status.Versions["firmware"] != "" {
```
adriansmares: Add alias for `status.Versions` and `conn.Gateway().Attributes`.

If `conn.Gateway().Attributes` is `nil`, this panics.
Also make sure that `status.Versions` is not `nil` (I'm not sure if we initialize this everywhere).

pkg/gatewayserver/gatewayserver.go
```diff
@@ -972,6 +990,36 @@ func (gs *GatewayServer) handleLocationUpdates(ctx context.Context, conn connect
...
+				"model":    status.Versions["model"],
+				"firmware": status.Versions["firmware"],
+			}
+			d := random.Jitter(gs.config.UpdateVersionInfoDelay, 1)
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(d):
+			}
+			err := gs.entityRegistry.UpdateAttributes(conn.Context(), *conn.Gateway().Ids, *&conn.Gateway().Attributes, attributes)
```
adriansmares: `*&` does not copy the object. Make a clone explicitly before the call, or inside `UpdateAttributes`.

adriansmares:
Also wouldn't this be shorter and easier if you compare the versions in `updateConnStats` ? Then you don't need a whole separate flow just for the first status. 

Regarding unregistered: Printing the error suffices.

pkg/gatewayserver/gatewayserver.go
KrishnaIyer: I don't think so. As far as I'm aware, you can read from an uninitialised map.
https://go.dev/play/p/TGaXMKqyTXc

KrishnaIyer:
> Also wouldn't this be shorter and easier if you compare the versions in updateConnStats ? Then you don't need a whole separate flow just for the first status.

Yes the reason is that these two update different registries, different debouncing, different life time (only once v/s multiple). I think this is cleaner than adding more if/else blocks in the `updateConnStats`.

pkg/gatewayserver/config.go
KrishnaIyer: The default delay is 5s. So -100 is 0 and +100 is 10s. That's gives us a 10s range. Using 25%, the range is 3.75 - 6.25. This means that this call will always wait for at least 3.75s and I don't want that.

pkg/gatewayserver/gatewayserver.go
adriansmares: Correct, only write panics on `nil` maps.

Also turn the `if` around and return if it holds

```
if !ok || versions["model"] == "" || versions["firmware"] == "" {
    return
}
```

pkg/gatewayserver/config.go
adriansmares: The average user will wait 5 seconds anyway. The fact that the lower bound is 'lower' does not change the expected value. But ok, I'm fine with a `[0s, 10s]` range.

adriansmares (approved):

PR #4913, opened 2021-11-24
Disable webhooks after too many failures

### LINKED ISSUE #4526: Disable webhook configuration after too many failures
#### Summary

I think we should build functionality into the AS Webhooks that disables webhook configurations after too many failures.

The AS is wasting a lot of time attempting webhooks that never succeed.

Disable webhook configurations after a (configurable) number of failures.

- Some configuration for number of failures, types of failures, when to reset the counters, etc.
- Add fields in the webhook configuration message for keeping track of the number of failures, maybe a field for one or multiple error messages.
- Keep track of the number of failures for a webhook configuration.
- If this number crosses a certain threshold, don't even bother trying.
- Display a warning in the Console if there's a high number of failures.
- Allow users to re-enable the webhook configuration.
- Maybe re-enable the webhook configuration automatically?

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4526

#### Changes

- Introduce validation for webhook headers
- Introduce the concept of webhook health status
  - A webhook is either healthy (i.e. the last HTTP request succeeded) or it is unhealthy (i.e. the recent attempts to run the HTTP requests failed)
  - An unhealthy webhook is allowed to generate HTTP requests until a threshold is met (`UnhealthyAttemptsThreshold`). After the threshold is passed, newer webhooks are generated only after a cooldown period (`UnhealthyRetryInterval`) passes between attempts
  - If a HTTP request succeeds while the webhook is unhealthy, it becomes healthy
- The implementation contains the following optimizations:
  - The initial health status is retrieved at the same time with the webhook itself (only one round trip)
  - If the webhook is healthy, and the HTTP requests succeed, no additional store calls are made
  - We only contact the underlying store only to update the failure status / move a webhook to the healthy state
- The following optimization exists in the commit history but introduces quite some complexity so I've removed it
  - Run the health check before attempting to enqueue the HTTP request for execution

On the `TODO` list:
- ~Expose the health check configuration such that the frontend can tell the users how many failures are left, or how much time has to pass until the next request may be done~ - done.
- ~Frontend support (show health status in the table, provide option to reset health status)~ - static status is now available

#### Testing

Local testing. I have to write a test battery but I'm releasing this for an early review.

##### Regressions

TBA.

#### Checklist


api/applicationserver_web.proto
```diff
@@ -101,6 +102,19 @@ message ApplicationWebhookTemplates {
   repeated ApplicationWebhookTemplate templates = 1;
 }
 
+message ApplicationWebhookHealth {
+  message WebhookHealthStatusHealthy { }
+  message WebhookHealthStatusUnhealthy {
+    uint64 failed_attempts = 1;
+    google.protobuf.Timestamp last_failed_attempt_at = 2 [(validate.rules).timestamp.required = true];
+    google.protobuf.Any last_failed_attempt_details = 3 [(validate.rules).any.required = true];
```
johanstokking: How is a frontend going to display this? Why wouldn't this be error details or a string (but not any of them)?

johanstokking (commented):
Looks good. Please append fixup commits for further review.

api/applicationserver_web.proto
adriansmares: The data of events is marshaled in the same way, using `Any`, so we already have logic available that walks through the `Any` object and checks if it is indeed an `ErrorDetails` (and even knows how to walk through the `cause` field recursively), or at least an error string.

I think we can go for `ErrorDetails` since it's always a stack error (or at least a wrap).

htdvisser:

api/applicationserver_web.proto
johanstokking: I would prefer `ErrorDetails` indeed.

> so we already have logic available that walks through [...]

When you say "we" do you mean TTS CLI and/or Console? I'm thinking from the perspective of third party API clients; what do they do with `Any`?

adriansmares: Third party clients could have used the `@type` field in order to determine how to handle the error, and that is consistent with most event handling. But indeed, let's make this one easy to use.

https://github.com/TheThingsNetwork/lorawan-stack/pull/4913/commits/30cc798654c0531369cd6e1ab4d1e0b380e1cd28

adriansmares:
This is now ready for review, with both the underlying registry and the logic unit tests added. 

Console screenshots:
<details>

image
image
image

</details>

bafonins (approved):

johanstokking (approved):
Nice

johanstokking:
Don't forget the changelog

api/applicationserver_web.proto
```diff
@@ -109,7 +123,13 @@ message ApplicationWebhook {
   // Base URL to which the message's path is appended.
   string base_url = 4 [(validate.rules).string.uri = true];
   // HTTP headers to use.
-  map<string,string> headers = 5;
+  map<string,string> headers = 5 [
+    (validate.rules).map = {
+      max_pairs: 50,
+      keys: { string: { max_len: 64 } },
```
htdvisser, line 135: We could also consider adding a regex here

api/applicationserver_web.proto
```diff
@@ -138,7 +158,9 @@ message ApplicationWebhook {
   Message location_solved = 14;
   Message service_data = 18;
 
-  // next: 20
+  ApplicationWebhookHealth health_status = 20;
```
htdvisser, line 167: Maybe mention that this is read-only? Or can users also use this to clear the health status?

htdvisser (approved):

api/applicationserver_web.proto
adriansmares: That is indeed the intent - the Console (soon) will have a button that nulls this in order to reset the health status.

kschiffer:
As discussed internally with @adriansmares, @ryaplots will open up a PR with some frontend adjustments to this PR, we can merge after this has been integrated.

kschiffer (approved):

PR #585, opened 2019-04-30
Display of empty config slices

furtiman:
**Summary:**

Closes #576 

**Changes:**

- Added empty string print in case a config with a certain flag is empty

I don't have rights to add reviewers and assign myself etc., so:
cc @htdvisser @johanstokking

PR #4929, opened 2021-11-26
Display warning if import file is too large

ryaplots:
#### Summary

The console is displaying a warning is the file size is larger than 3MB.
Related to support [#585](https://github.com/TheThingsIndustries/lorawan-stack-support/issues/585)

#### Changes

- Add an isLarger prop to the state and set it to true is file larger than 3mb;
- Do this onChange.
- This is how it looks:

Screenshot 2021-11-29 at 12 13 00

#### Checklist


pkg/webui/components/file-input/index.js
```diff
@@ -101,8 +104,15 @@ export default class FileInput extends Component {
   handleChange(event) {
     const { maxSize } = this.props
     const { files } = event.target
+    const threeMb = 3 * 1024 * 1024
```
kschiffer:
```suggestion
    const warningSize = 10 * 1024 * 1024 // 10MB.
```

See here: https://github.com/TheThingsIndustries/lorawan-stack-support/issues/585#issuecomment-976298265

pkg/webui/components/file-input/index.js
```diff
@@ -168,37 +178,47 @@ export default class FileInput extends Component {
...
+            disabled={disabled}
+            tabIndex="-1"
+          />
+        </div>
+        <div>
+          {this.state.isLarger && (
+            <React.Fragment>
+              <Icon className={style.warningIcon} icon="error" />
+              <Message className={style.warning} content={m.largeFileWarning} />
+            </React.Fragment>
```
kschiffer:
```suggestion
            <Notification warning content={m.largeFileWarning}  small />
```

kschiffer: Also, the structure should rather be

```javascript
<div className={style.container}>
  … // controls
  <div> // It might also be possible to omit the wrapping `div`, depending on the styling.
    <Notification />
  </div>
</div>
```

pkg/webui/components/file-input/file-input.styl
```diff
@@ -36,6 +36,13 @@
     color: $c-error
     margin-right: $cs.xxs
 
+.warning
+  color: $c-warning
+  line-height: $lh.s
+  &-icon
+    color: $c-warning
+    margin-right: $cs.xxs
+
```
kschiffer: This can be removed when using the `<Notification />` component.

kschiffer (changes requested):
See my remarks.

Please always provide a screenshot for PRs with visual changes to facilitate review.

pkg/webui/components/file-input/index.js
ryaplots: With the structure you are suggesting the notification doesn't seem to be in the right place. Because the div has a container style with display flex

Screenshot 2021-11-29 at 10 26 50
.

ryaplots: If I do it like so:
```
<div>
    <Notification />
    <div className={style.container}>
    … // controls
    </div>
</div>
```

it will work like this:

Screenshot 2021-11-29 at 10 45 27

kschiffer: Ok yeah, this is indeed not working then as I suggested. The thing is though that classes with the name `container` should always be the outermost container. So what you should do here is adapt the CSS so that this is retained.

pkg/webui/components/file-input/index.js
```diff
@@ -31,6 +31,8 @@ const m = defineMessages({
   changeFile: 'Change file…',
   noFileSelected: 'No file selected',
   fileProvided: 'A file has been provided',
+  largeFileWarning:
+    'Providing files larger than 3MB can cause issues during the import process. We recommend you to split such files up into multiple smaller files and importing them one by one.',
```
kschiffer: I realize now that the feature cannot be hardcoded like that into the component. Keep in mind that the component is just an input component. The file size warning is contextual to the device import feature. The `<FileInput />` component is not aware of any split-up requirements.

As such, both the warning threshold (10MB) and the warning message itself should be passed as props.

pkg/webui/components/file-input/index.js
```diff
@@ -63,12 +67,14 @@ export default class FileInput extends Component {
...
     imageClassName: undefined,
-    maxSize: 10 * 1024 * 1024, // 10 MB
+    maxSize: 16 * 1024 * 1024, // 16 MB
     mayRemove: true,
     message: m.selectAFile,
     changeMessage: m.changeFile,
     providedMessage: m.fileProvided,
     value: undefined,
+    warningSize: 10 * 1024 * 1024, // 10 MB
+    warningThreshold: '10MB',
```
kschiffer:
```suggestion
```
Better to not use defaults here. Usually warning messages for large files won't be necessary when using the file input.

pkg/webui/components/file-input/index.js
```diff
@@ -99,10 +106,16 @@ export default class FileInput extends Component {
...
+    const { maxSize, warningSize } = this.props
     const { files } = event.target
 
     if (files && files[0] && files[0].size <= maxSize) {
+      if (files && files[0] && files[0].size <= warningSize) {
+        this.setState({ isLarger: true })
+      } else {
+        this.setState({ isLarger: false })
+      }
+
```
kschiffer: Keep in mind that it is possible to use the file input without a warning size.

pkg/webui/components/file-input/index.js
kschiffer: Also, ideally `warningThreshold` should not be necessary, as this can be derived from `warningSize`. You could use one of the [many snippets](https://stackoverflow.com/questions/10420352/converting-file-size-in-bytes-to-human-readable-string/10420404) to convert bytes into a human-readable format.

kschiffer (changes requested):

pkg/webui/components/file-input/index.js
```diff
@@ -48,13 +49,15 @@ export default class FileInput extends Component {
...
     imageClassName: PropTypes.string,
+    largeFileWarningMessage: PropTypes.string.isRequired,
     maxSize: PropTypes.number,
     mayRemove: PropTypes.bool,
     message: PropTypes.message,
     name: PropTypes.string.isRequired,
     onChange: PropTypes.func.isRequired,
     providedMessage: PropTypes.message,
     value: PropTypes.oneOfType([PropTypes.string, PropTypes.shape({})]),
+    warningSize: PropTypes.number.isRequired,
```
kschiffer:
```suggestion
    warningSize: PropTypes.number.isRequired,
```

It should not be required. It's perfectly fine to use the `<FileInput />` without any warning popping up for large file sizes.

pkg/webui/components/file-input/index.js
```diff
@@ -99,10 +103,14 @@ export default class FileInput extends Component {
...
   handleChange(event) {
-    const { maxSize } = this.props
+    const { maxSize, warningSize } = this.props
     const { files } = event.target
 
     if (files && files[0] && files[0].size <= maxSize) {
+      if (files && files[0] && files[0].size >= warningSize) {
+        this.setState({ isLarger: true })
+      }
+
```
kschiffer:
```suggestion
      if (typeof warningSize === 'number' && files && files[0] && files[0].size >= warningSize) {
        this.setState({ isLarger: true })
      }
```

pkg/webui/components/file-input/index.js
```diff
@@ -124,6 +132,12 @@ export default class FileInput extends Component {
     onChange(dataTransform(''), true)
   }
 
+  @bind
+  humanFileSize(size) {
+    const i = size === 0 ? 0 : Math.floor(Math.log(size) / Math.log(1024))
+    return `${(size / Math.pow(1024, i)).toFixed(2) * 1} ${['B', 'kB', 'MB', 'GB', 'TB'][i]}`
+  }
+
```
kschiffer: Consider outsourcing this as a `lib/util/human-file-size.js` since it might be useful in other places as well.

kschiffer (changes requested):

pkg/webui/components/file-input/index.js
```diff
@@ -99,10 +106,13 @@ export default class FileInput extends Component {
...
+    const { maxSize, warningSize } = this.props
     const { files } = event.target
 
     if (files && files[0] && files[0].size <= maxSize) {
+      if (typeof warningSize === 'number' && files && files[0] && files[0].size >= warningSize) {
+        this.setState({ isLarger: true })
+      }
       this.setState({ filename: files[0].name, error: undefined })
       this.reader.readAsDataURL(files[0])
     } else {
```
kschiffer:
```suggestion
    if (files && files[0]) {
      if (files[0].size <= maxSize) {
        this.setState({ error: m.tooBig })
      } else {
      this.reader.readAsDataURL(files[0])
      this.setState({ filename: files[0].name, error: undefined, isLarger: files[0].size >= warningSize })      
      }
    }
```

Can be simplified.

kschiffer (approved):
👍

CHANGELOG.md
```diff
@@ -15,8 +15,11 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
+
+- Increased the maximum gRPC message size to 16MB.
 - Gateways which have been deleted are now automatically disconnected by the Gateway Server.
 
+
```
adriansmares: No need for two lines.

adriansmares (approved):

htdvisser (approved):

PR #4932, opened 2021-11-29
Disconnect gateway on delete

### LINKED ISSUE #4901: Disconnect gateway when it gets deleted
#### Summary

Disconnect gateway when it gets deleted.

#### Why do we need this?

Because it shouldn't stay connected when it's gone.

The gateway isn't disconnected.

Functionality that we do already have, is that if some fields are changed, the gateway gets disconnected.

Disconnect the gateway when it's deleted.

In the logic that fetches gateway fields from Identity Server, check the returned error. If it's `NotFound`, disconnect the gateway.

CI

Can review, can do

KrishnaIyer:
#### Summary

Closes #4901 

#### Changes

- Disconnect gateway on delete. 
- Add tests

#### Testing

- [x] UT
- [x] Local stack test

1. Connection with Auth
WithAuth

2. Connection without Auth
WithoutAuth

##### Regressions

Don't think there should be any. The original disconnection logic is not touched.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
-- Disconnect gateways that are deleted on the Identity Server.
+- Gateways which have been deleted are now automatically disconnected by the Gateway Server.
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
+- Disconnect gateways that are deleted on the Identity Server.
+
 ### Deprecated
 
 ### Removed
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -583,7 +583,6 @@ func (gs *GatewayServer) startDisconnectOnChangeTask(conn connectionEntry) {
 				return ctx.Err()
 			case <-time.After(d):
 			}
-
 			gtw, err := gs.entityRegistry.Get(ctx, &ttnpb.GetGatewayRequest{
 				GatewayIds: conn.Gateway().GetIds(),
 				FieldMask: &pbtypes.FieldMask{
@@ -604,7 +603,12 @@ func (gs *GatewayServer) startDisconnectOnChangeTask(conn connectionEntry) {
 				},
 			})
 			if err != nil {
-				log.FromContext(ctx).WithError(err).Warn("Failed to get gateway")
+				if errors.IsNotFound(err) {
+					// Gateway was deleted. Disconnect.
+					conn.Disconnect(err)
+				} else {
+					log.FromContext(ctx).WithError(err).Warn("Failed to get gateway")
+				}
 				return err
 			}
 			if requireDisconnect(conn.Gateway(), gtw) {
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -608,7 +608,7 @@ func (gs *GatewayServer) startDisconnectOnChangeTask(conn connectionEntry) {
 					// Since there is an active connection, the `Get` request will not return a 404 as the gateway existed during the connect. Instead,
 					// 1. If the gateway is connected with an API key and is deleted, the IS returns an `Unauthenticated`, since the API Key is also deleted.
 					// 2. If the gateway is connected without an API key (UDP, LBS in unauthenticated mode) and is deleted the IS returns an `PermissionDenied` as there are no rights for these IDs.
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
+- Disconnect gateways that are deleted on the Identity Server.
```
adriansmares: I think `Gateways which have been deleted are now automatically disconnected by the Gateway Server` sounds a bit better.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -604,7 +604,15 @@ func (gs *GatewayServer) startDisconnectOnChangeTask(conn connectionEntry) {
 				},
 			})
 			if err != nil {
-				log.FromContext(ctx).WithError(err).Warn("Failed to get gateway")
+				if errors.IsUnauthenticated(err) || errors.IsPermissionDenied(err) {
+					// Since there is an active connection, the `Get` request will not return a 404 as the gateway existed during the connect. Instead,
+					// 1. If the gateway is connected with an API key and is deleted, the IS returns an `Unauthenticated`, since the API Key is also deleted.
+					// 2. If the gateway is connected without an API key (UDP, LBS in unauthenticated mode) and is deleted the IS returns an `PermissionDenied` as there are no rights for these IDs.
+					log.FromContext(ctx).WithError(err).Debug("Gateway was deleted and/or the API key used to link the gateway was invalidated. Disconnect.")
```
adriansmares: I would drop the `Disconnect.`.

adriansmares (approved):
Good stuff, some nits.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -604,7 +604,15 @@ func (gs *GatewayServer) startDisconnectOnChangeTask(conn connectionEntry) {
 				},
 			})
 			if err != nil {
-				log.FromContext(ctx).WithError(err).Warn("Failed to get gateway")
+				if errors.IsUnauthenticated(err) || errors.IsPermissionDenied(err) {
+					// Since there is an active connection, the `Get` request will not return a 404 as the gateway existed during the connect. Instead,
```
johanstokking: Why would it not when the gateway gets deleted while it is connected?

Nit; 404 is HTTP which isn't applicable here

KrishnaIyer: The IS fails on the rights assertion before it checks if the entity exists. Updated the comment.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -604,7 +604,16 @@ func (gs *GatewayServer) startDisconnectOnChangeTask(conn connectionEntry) {
 				},
 			})
 			if err != nil {
-				log.FromContext(ctx).WithError(err).Warn("Failed to get gateway")
+				if errors.IsUnauthenticated(err) || errors.IsPermissionDenied(err) {
+					// Since there is an active connection, the `Get` request will not return a `NotFound` error as the gateway existed during the connect, since the rights assertion fails first.
+					// Instead,
+					// 1. If the gateway is connected with an API key and is deleted, the IS returns an `Unauthenticated`, since the API Key is also deleted.
+					// 2. If the gateway is connected without an API key (UDP, LBS in unauthenticated mode) and is deleted the IS returns an `PermissionDenied` as there are no rights for these IDs.
```
johanstokking, line 611: Sorry I still don't get this. If there is no authentication to forward, GS uses cluster authentication (https://github.com/TheThingsIndustries/lorawan-stack/blob/v3.16/pkg/gatewayserver/is.go#L67-L69). IS will just return `NotFound` right?

KrishnaIyer: This is quite non-intuitive. I had the same expectation as you did and was surprised when the tests failed. So here it goes
1. On `Get`, the IS asserts the rights using [`RequireGateway`](https://github.com/TheThingsNetwork/lorawan-stack/blob/a053461febe1813ad537ef7ba38533e026c7b6cc/pkg/auth/rights/require.go#L159)
2. The IS then looks for the gateway store with the IDs. It returns a [`NotFound`](https://github.com/TheThingsNetwork/lorawan-stack/blob/c2ce76a770e8ac689a63508711e51e16a1519af2/pkg/identityserver/rights.go#L166-L170). Since we are using cluster auth and not doing a `Get` as admin, **the cause of the error (NotFound) is suppressed** and this function returns empty rights.
3. Now back in `RequireGateway`, since the returned rights are empty, we return a [PermissionDenied](https://github.com/TheThingsNetwork/lorawan-stack/blob/a053461febe1813ad537ef7ba38533e026c7b6cc/pkg/auth/rights/require.go#L165-L167).

I think the reason for this behaviour is to not leak whether a gateway is present or not for someone who wants to scan the database.

johanstokking (approved):

PR #4946, opened 2021-12-01
Include user defined antenna gain in LBS router config

### LINKED ISSUE #4191: Support user defined Antenna Gain in LBS router config
#### Summary

Support user defined Antenna Gain in LBS router config

#### Why do we need this?

To allow users to tweak the antenna gain for specific gateways.

LBS router config does not handle Antenna Gain.

The LBS spec moves the `antenna_gain` field to the RFCONF object.
https://doc.sm.tc/station/gw_v1.5.html#rfconf-object

I still need to figure out the relation between the radio and the antenna and also how to map antennas with multiple concentrators.

#### Environment

TTS v3.13.0

Not sure yet. Will keep this in backlog until relevant.

GTW testing

Yup

KrishnaIyer:
#### Summary

Closes #4191 

#### Changes

- Add `AntennaGain` to `RFCONF` object; https://doc.sm.tc/station/gw_v1.5.html#rfconf-object
- Update code and tests
- Also fix a small bug in the CLI while setting Gateway Antennas

#### Testing

- [x] UT 
- [x] Stack Test
   - [x] Check that LBS doesn't fail parsing
TTIG Parsing

   - [x] Check that Antenna Gain reflects in the RouterConfig   
RC
RC2
   
- [x] Test downlinks work
  
Downlink

join

- [ ] Check with Tektelic gateway ⚠️ 

##### Regressions

I don't expect any but depends on the gateway.

#### Notes for Reviewers

1. I can't really verify if the gateway is offsetting the Tx power since I'll need a spectrum analyzer. I assume that if there are no errors in the logs, this should be ok. I've checked the LBS source code and this is a logical assumption.
   - LBS [stores](https://github.com/lorabasics/basicstation/blob/60fd9788763ea11ec80ec1b1f683c0374792f71e/src/sx130xconf.c#L119) the gain value from config.
   - It's [offset](https://github.com/lorabasics/basicstation/blob/60fd9788763ea11ec80ec1b1f683c0374792f71e/src/ral_lgw.c#L211) during a downlink.
   - But when printing the Tx Ack, it prints the [absolute Pow value](https://github.com/lorabasics/basicstation/blob/60fd9788763ea11ec80ec1b1f683c0374792f71e/src/s2e.c#L294) and not the offset one.
2. The AntennaGain is an INT value for LBS so this will truncate fractional gain values.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/ws/lbslns/version.go
+++ b/pkg/gatewayserver/io/ws/lbslns/version.go
@@ -64,7 +64,7 @@ func (v Version) IsProduction() bool {
 }
 
 // GetRouterConfig gets router config for the particular version message.
-func (f *lbsLNS) GetRouterConfig(ctx context.Context, msg []byte, bandID string, fps map[string]*frequencyplans.FrequencyPlan, receivedAt time.Time) (context.Context, []byte, *ttnpb.GatewayStatus, error) {
+func (f *lbsLNS) GetRouterConfig(ctx context.Context, msg []byte, bandID string, fps map[string]*frequencyplans.FrequencyPlan, antennaGain int, receivedAt time.Time) (context.Context, []byte, *ttnpb.GatewayStatus, error) {
 	var version Version
 	if err := json.Unmarshal(msg, &version); err != nil {
 		return nil, nil, nil, err
@@ -74,7 +74,7 @@ func (f *lbsLNS) GetRouterConfig(ctx context.Context, msg []byte, bandID string,
 	// to gateways that signal the presence of a PPS.
 	// References https://github.com/lorabasics/basicstation/issues/135.
 	updateSessionTimeSync(ctx, true)
-	cfg, err := pfconfig.GetRouterConfig(bandID, fps, version.IsProduction(), time.Now())
+	cfg, err := pfconfig.GetRouterConfig(bandID, fps, version.IsProduction(), time.Now(), antennaGain)
 	if err != nil {
 		return nil, nil, nil, err
 	}
--- a/pkg/pfconfig/lbslns/lbslns.go
+++ b/pkg/pfconfig/lbslns/lbslns.go
@@ -84,9 +84,11 @@ type DataRates [16][3]int
 // The other fields of RFConfig (in pkg/pfconfig/shared) are hardware specific and are left out here.
 // - `type`, `rssi_offset`, `tx_enable` and `tx_notch_freq` are set in the gateway.
 // - `tx_freq_min` and `tx_freq_max` are defined in the  `freq_range` parameter of `router_config`.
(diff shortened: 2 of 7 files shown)


pkg/pfconfig/lbslns/lbslns.go
```diff
@@ -231,7 +234,7 @@ func (conf RouterConfig) MarshalJSON() ([]byte, error) {
 
 // GetRouterConfig returns the routerconfig message to be sent to the gateway.
 // Currently as per the basic station docs, all frequency plans have to be from the same region (band) https://doc.sm.tc/station/tcproto.html#router-config-message.
-func GetRouterConfig(bandID string, fps map[string]*frequencyplans.FrequencyPlan, isProd bool, dlTime time.Time) (RouterConfig, error) {
+func GetRouterConfig(bandID string, fps map[string]*frequencyplans.FrequencyPlan, isProd bool, dlTime time.Time, antennaGain int) (RouterConfig, error) {
```
adriansmares, line 237: This is getting quite crowded, we should consider a `struct` / `Options` in the future.

adriansmares (approved):

pkg/gatewayserver/io/ws/lbslns/version.go
```diff
@@ -64,7 +64,7 @@ func (v Version) IsProduction() bool {
 }
 
 // GetRouterConfig gets router config for the particular version message.
-func (f *lbsLNS) GetRouterConfig(ctx context.Context, msg []byte, bandID string, fps map[string]*frequencyplans.FrequencyPlan, receivedAt time.Time) (context.Context, []byte, *ttnpb.GatewayStatus, error) {
+func (f *lbsLNS) GetRouterConfig(ctx context.Context, msg []byte, bandID string, fps map[string]*frequencyplans.FrequencyPlan, antennaGain int, receivedAt time.Time) (context.Context, []byte, *ttnpb.GatewayStatus, error) {
```
htdvisser, line 67: This func signature is getting a bit nasty, eventually we should probably take a struct instead of all these arguments.

htdvisser (approved):
LGTM

pkg/gatewayserver/io/ws/lbslns/version.go
htdvisser: Oh haha, just noticed that @adriansmares wrote exactly the same comment

johanstokking (approved):

pkg/pfconfig/lbslns/lbslns.go
KrishnaIyer: Yup I've considered this as part of the multiple antenna downlink issue.

PR #4952, opened 2021-12-01
Fix application layer crash and add smoke test for device general settings

ryaplots:
#### Summary

Fixed the application layer crash spotted on staging.
Added smoke test to prevent this from happening in the future

#### Changes

- Description message
- Added a smoke test

#### Testing

Run smoke tests

#### Checklist


cypress/integration/smoke/index.js
```diff
@@ -23,6 +24,7 @@ import authorizationTests from './authorization'
 import profileSettingsTests from './profile-settings'
 
 export default [
+  ...checkCollapsingFields,
```
kschiffer: I think putting this after the `applicationTests` is better, since these tests also depend on the application tests to pass.

cypress/fixtures/console/devices/device.ns.json
```diff
@@ -0,0 +1,49 @@
+{
+  "end_device":{
+    "frequency_plan_id":"EU_863_870_TTN",
+    "lorawan_phy_version":"PHY_V1_0_2_REV_B",
+    "multicast":false,
+    "supports_join":false,
```
kschiffer: ABP devices are quite rare in the real world, so I think we should not use them by default in the device mock. Two things could be done here:

1.  Create a separate mock for ABP devices
2.  Create a way to pass overwrites per component to the mock device within `createMockDeviceAllComponents`, eg. `createMockDeviceAllComponents({ ns: { end_device: { supports_join: false }}})`

Then merging those like

```js
  cy.fixture('console/devices/device.ns.json').then(body => {
    cy.request({
      method: 'PUT',
      url: `${baseUrl}/api/v3/ns/applications/${applicationId}/devices/${body.end_device.ids.device_id}`,
      { ...body, ...overwrites.ns },
      headers,
    })
  })
```

I think I'd prefer 2. since it is more flexible and requires fewer fixtures.

kschiffer (changes requested):
Two small remarks, see above.

cypress/support/commands.js
```diff
@@ -275,7 +275,7 @@ Cypress.Commands.add('createEndDeviceAllComponents', applicationId => {
     cy.request({
       method: 'PUT',
       url: `${baseUrl}/api/v3/ns/applications/${applicationId}/devices/${body.end_device.ids.device_id}`,
-      body,
+      body: { ...body, ...overwrites.ns },
```
kschiffer: Why only the NS? The same mechanism should be used for all components.

cypress/support/commands.js
```diff
@@ -257,7 +257,7 @@ Cypress.Commands.add('createEndDevice', (applicationId, endDevice) => {
 })
 
 // Helper function to create a mock device in all components.
-Cypress.Commands.add('createEndDeviceAllComponents', applicationId => {
+Cypress.Commands.add('createMockDeviceAllComponents', (applicationId, overwrites = {}) => {
```
kschiffer:
```suggestion
Cypress.Commands.add('createMockDeviceAllComponents', (applicationId, overwrites = { ns: {}, as: {}, js: {}, is: {} }) => {
```

kschiffer (changes requested):

kschiffer (approved):

PR #4954, opened 2021-12-01
Sentry error fixes

adriansmares:
#### Summary

References https://sentry.io/organizations/the-things-industries/issues/2780300004/?project=2682566
References https://sentry.io/organizations/the-things-industries/issues/2406140221/?project=2682566
References https://sentry.io/organizations/the-things-industries/issues/2474217529/?project=2682566
References https://sentry.io/organizations/the-things-industries/issues/2828875503/?project=2682566
References https://sentry.io/organizations/the-things-industries/issues/2232797687/?project=2682566

#### Changes

- Disable `nil` channels via validation of the MAC state
  - This was possible due to missing validation
  - Existing devices are to be found and edited on demand unfortunately, I'll think of a way of doing this systematically perhaps via a migration
  - We may want to add this kind of validation to other `repeated` messages too - please let me know if you want this to be done systematically.
- Adjust some error codes that were unknown / corruption
- Gateways are now claimed using a decoupled context, in order to avoid spurious errors because the gateway disconnected sporadically. Keep in mind that we don't actually know if the claim succeeded, if the gateway disconnects while `ClaimIDs` is running, so we may actually miss an unclaim if we don't use an 'uncancellable context'. 

#### Testing

Unit testing.

##### Regressions

N/A

#### Checklist

### DIFF
--- a/pkg/gatewayserver/upstream/ns/ns.go
+++ b/pkg/gatewayserver/upstream/ns/ns.go
@@ -74,15 +74,15 @@ func (h *Handler) ConnectGateway(ctx context.Context, ids ttnpb.GatewayIdentifie
 	if conn.Frontend().SupportsDownlinkClaim() {
 		return nil
 	}
+	decoupledCtx := h.contextDecoupler.FromRequestContext(ctx)
 	logger := log.FromContext(ctx)
-	if err := h.cluster.ClaimIDs(ctx, &ids); err != nil {
+	if err := h.cluster.ClaimIDs(decoupledCtx, &ids); err != nil {
 		logger.WithError(err).Error("Failed to claim downlink path")
 		return err
 	}
 	logger.Info("Downlink path claimed")
 	defer func() {
-		ctx := h.contextDecoupler.FromRequestContext(ctx)
-		if err := h.cluster.UnclaimIDs(ctx, &ids); err != nil {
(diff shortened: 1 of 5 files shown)


KrishnaIyer (approved):
Code owned LGTM

api/end_device.proto
```diff
@@ -122,7 +122,10 @@ message MACParameters {
     bool enable_uplink = 5;
   }
   // Configured uplink channels and optionally Rx1 frequency.
-  repeated Channel channels = 19 [(validate.rules).repeated.min_items = 1];
+  repeated Channel channels = 19 [(validate.rules).repeated = {
+    min_items: 1,
+    items: { message: { required: true } }
```
htdvisser: Wait, is it possible to receive `nil`s in repeated messages received over RPC? Because then we probably need to check this in more places.

htdvisser (approved):

johanstokking (approved):

api/end_device.proto
adriansmares: No, it is not. I'm actually in the wrong here, it is not possible neither via gRPC, nor via the gRPC bridge. The `nil` channel must be from somewhere else.

Edit: Nullable channels are a common use case for channel deletion / creation. Will just add a `nil` check.

PR #4970, opened 2021-12-05
Asynchronous replication of end device locations

### LINKED ISSUE #4927: Asynchronous replication of end device metadata in Application Server
#### Summary

The Application Server should replicate end device metadata asynchronously from the Identity Server, instead of stalling the hot path.

#### Why do we need this?

Currently in the Application Server hot path we contact the Identity Server in order to retrieve the end device locations. 
The hit rate is pretty bad (citation needed, this lacks metrics), because the frequency at which end device messages arrive is very low, and we sometimes need to cross the Atlantic - this is again very expensive.

A large family of end device fetchers in the Application Server, which stall message processing until the Identity Server is contacted.

We should no longer stall the processing of messages while contacting the Identity Server. Instead, we should asynchronously replicate these fields in a hot store. 

I initially wanted to propose that we just hold these in the Application Server store, but maybe it's not a bad idea to have an actual separate store, on the `cache` Redis. On uplink we check if we have the metadata available, and if yes we fill it. Otherwise, without stalling the hot path, we contact the IS and inject the metadata in the hot store.

#### Environment

`v3.16.1`

Remove any other fetcher except the one that contacts the Identity Server. On uplink, check the cache, and if a miss occurred, contact the IS asynchronously.

The AS should also update the end device locations stored in the cache, in order to maintain a degree of coherency.

Unit testing.

Yes. cc @johanstokking @htdvisser

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4927

#### Changes

- Remove and deprecate the end device fetchers, and replace them with end device metadata registry
- Introduce a location registry that is backed by the Entity Registry
- Introduce a location cache that is backed by Redis
- Introduce a location registry that fills the location cache asynchronously
  - The cache is always filled asynchronously
  - Tune knobs are the minimum and maximum interval between two asynchronous cache refreshes, and a general TTL for the data
- Mark end devices as activated using worker pool instead of using separate tasks, for better monitoring
- Process uplinks published by integrations asynchronously 
  - Mainly affects integrations that that publish location solved messages, as those generally were blocking until the location solved message was processed, and this wasn't ideal as application packages have execution limits (context deadlines) that are meant for individual message processing
- Retrieve end device identifiers from the local device registry, instead of contacting the Entity Registry

#### Testing

##### Regressions

Unit tests should cover this.

#### Notes for Reviewers

This PR grew a bit in context since it tries to tackle the hot path limitations as a whole. @johanstokking main reviewer, but this PR may wait until the `3.17.0` cycle if we want to avoid having an eventful Christmas.

#### Checklist


pkg/applicationserver/metadata/location_registry.go
```diff
@@ -0,0 +1,246 @@
...
+	replicationPool workerpool.WorkerPool
+}
+
+// Get implements EndDeviceLocationRegistry.
+func (c *cachedEndDeviceLocationRegistry) Get(ctx context.Context, ids ttnpb.EndDeviceIdentifiers) (map[string]*ttnpb.Location, error) {
+	locations, ttl, err := c.cache.Get(ctx, ids)
+	switch {
+	case err != nil && !errors.IsNotFound(err):
+		return nil, err
+	case err == nil:
```
johanstokking: What if `errors.IsNotFound(err)`?

adriansmares: We don't enter the `switch` and trigger an asynchronous replication. The result of the `Get` is empty (no locations).

johanstokking: So if `errors.IsNotFound()` then one of the return values is still used; it could be nil or non-nil, or even have dirty values, depending on the implementation of `Get()`. We usually don't rely on other return values if `err` isn't `nil`.

johanstokking (approved):
LGTM. What are the potential consequences at scale? What is the downside of waiting for 3.17.0, other than diffs?

pkg/applicationserver/metadata/location_registry.go
adriansmares: Added a `nil` setup.

adriansmares:
> LGTM. What are the potential consequences at scale? What is the downside of waiting for 3.17.0, other than diffs?

- The 24 hour hard TTL is counter intuitive, but I would argue that it matches the expectation that 'stable' locations such as the user set location in the Console won't change often. I have prepared a cache invalidation mechanism (the `Delete` in the cache), so if someone really wants to purge the cache, we can add an RPC. With that being said, the invalidation will happen automatically once we take some time to implement https://github.com/TheThingsIndustries/lorawan-stack/issues/3001, so I wouldn't add this RPC unless we really have someone screaming and shouting.
- Devices that send messages at an interval larger than the 24 hour TTL will never have their locations filled from the Entity Registry, as the cache will expire before it is used.
- The cache Redis will see more usage, one key per end device, but with minimal contents (only the `_cached` key will be used for most end devices).
- The IS usage will plummet, as per metrics the end device location retrievals are the most request information from the IS.
- P95 and P99 NS-AS latency is expected to drop, as the injection of the location on the end path is zero cost both if the cache is hit or not.

We're not in a critical position (tenant retrieval is more dangeous as it actually affects the potential of an RX1=1 transmission), so minus the rebasing this can wait indeed.

johanstokking:
> * [...] unless we really have someone screaming and shouting.

It does affect the consistency of the data; an API client may update the locations in IS, and it's only eventually consistent (up to 1 day) in uplink traffic. So indeed there might be someone screaming and shouting as it's a behavioral API change in a patch.

What is 24 hours TTL based on, actual data?

adriansmares:
> What is 24 hours TTL based on, actual data?

Minimizing the chance that we return an empty map of locations due to the fact that the end device transmits at high intervals of time. My target was that we don't surprise anyone with this change, outside of the very long caching time. 

Perhaps that's the wrong target, and we can instead block on cache miss, ensure that we have the latest version cached, and this allows us to use a way smaller set of TTLs - something like 15 minutes soft TTL and 1 hour of hard TTL.

Regarding the eventual consistency - it was already eventually consistent, as we did cache the retrieved locations for 5 minutes. I can agree that 5 minutes and 24 hours are orders of magnitude different, so maybe we really should retrieve the locations on miss - then the only behavior that is changed is the TTL from a user perspective. For us, it ensures that we don't block the hot path for devices that send messages often enough, which is already an improvement perhaps.

I will run some tests with the blocking behavior and come back with more concrete data.

johanstokking:
>I can agree that 5 minutes and 24 hours are orders of magnitude different, so maybe we really should retrieve the locations on miss - then the only behavior that is changed is the TTL from a user perspective.

I meant that the largely increased TTL leads to the behavioral API change. Indeed, it was already eventual consistent.

I think that unblocking the hot path is good aim. I think that a longer TTL and asynchronous recache is the best solution. I think that 24 hours is a bit long though. As long as we can configure it and progressively increase it based on performance, I'm fine with it, also merging it in 3.16.

---

We would need a soft TTL to distribute recaches: either 1) many items at the same time because of mass eviction, or 2) one item that gets recached concurrently. I fail to see this flow meeting either criteria; (1) eviction would go in waves if we use an LFU cache and we're evicting all frequently used items with the same TTL, but in this case, the cache time and the eviction is distributed by incoming traffic, and (2) isn't met because deduplicated uplink traffic comes in sequentially for the same end device on the AS level. Or am I missing something? I know we discussed this but I don't remember ...

pkg/applicationserver/metadata/location_registry.go
```diff
@@ -0,0 +1,248 @@
...
+		remaining := window - ttl
+		if remaining <= 0 {
+			return locations, nil
+		}
+		threshold := time.Duration(random.Int63n(int64(window)))
+		// remaining is the remaining window of the soft TTL in the (0, window) interval.
+		// threshold is a uniformly distributed duration in the [0, window) interval.
+		if remaining >= threshold {
+			return locations, nil
+		}
```
adriansmares, line 190: Here is the soft TTL flow, that occurs on cache hit. `window` is the window of time in which a soft TTL asynchronous cache refresh can occur. `remaining` is how much of the window is left. At remaining `ttl`=`window`, the probability is 0% that we do an asynchronous cache retrieval, and at remaining `ttl`=`0` the probability is 100% that we do an asynchronous cache retrieval.

We don't block when this occurs - if the cache was hit (so we still have these values) we return them immediately. The cache refresh occurs in the background.

adriansmares:
I gave this some more thought and we can change a bit the strategy. Instead of setting the TTL of the keys in Redis, we store the time at which we stored the metadata. In this way we don't need a very large TTL, and the soft TTL / hard TTL control only the probabilities.

pkg/applicationserver/metadata/location_registry.go
```diff
@@ -0,0 +1,245 @@
...
+			// refresh the cache.
+			window := c.hardTTL - c.softTTL
+			threshold := time.Duration(random.Int63n(int64(window)))
+			// remaining is the remaining window of the soft TTL in the (0, window) interval.
+			// threshold is a uniformly distributed duration in the [0, window) interval.
+			if remaining >= threshold {
+				return locations, nil
+			}
+		}
+	}
```
adriansmares: I've updated the refreshing logic. Now instead of actually having expiring keys, we just guarantee that we will do a cache refresh after the maximum refresh interval has passed. So now:

- Items younger than the minimum refresh interval are just served as is.
- Items whose age is between the minimum and maximum refresh interval may trigger an asynchronous cache refresh.
- Items whose age is higher than the maximum refresh interval are guaranteed to trigger an asynchronous cache refresh.

johanstokking:
OK please request another review round if needed, otherwise my approval stands.

htdvisser (approved):
Code-owned change LGTM

PR #4976, opened 2021-12-07
Fix Redis namespace for non-general Redis instances

adriansmares:
#### Summary

This PR fixes the way fallbacks are handled for Redis settings, specifically namespacing: the events and caching Redis instances do not have a `RootNamespace` set, so the keys look as follows:

image

This happens because the fallback occurs only when the configuration is empty, but if the address is provided, then the `RootNamespace` is not copied, and the keys have an empty one.

#### Changes

- Introduce a default cache configuration. This will backfill the Redis configuration, including the root namespace, but the cache backend won't be set - the administrators need to manually set this to `redis` for caching to occur.
- Use the default Redis configuration for the events storage.

#### Testing

Local testing.

##### Regressions

This is a pretty breaking change. Specifically, since the root namespace won't be empty anymore, the old events won't be visible anymore when this occurs. I'm opening up this PR to ask if we want to fix this for a minor.

#### Checklist

### DIFF
--- a/cmd/internal/shared/config.go
+++ b/cmd/internal/shared/config.go
@@ -95,11 +95,17 @@ var DefaultRedisConfig = redis.Config{
 	RootNamespace: []string{"ttn", "v3"},
 }
 
+// DefaultCacheConfig is the default cache configuration.
+var DefaultCacheConfig = config.Cache{
+	Redis: DefaultRedisConfig,
+}
+
 // DefaultEventsConfig is the default config for Events.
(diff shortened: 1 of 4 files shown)


htdvisser (changes requested):
I think this is acceptable for caching and events. 

But please point this PR at a `v3.17` branch, so that we don't accidentally merge it into v3.16.

htdvisser (approved):

PR #4977, opened 2021-12-08
Add cluster hints on end devices

### LINKED ISSUE #3831: Add cluster hints to End Devices
#### Summary

I think it would be helpful to add cluster hints to end device lists and pages.

Related to https://github.com/TheThingsNetwork/lorawan-stack/issues/3830

With explicit linking gone, Applications in v3 are no longer explicitly linked (between Application Server and Network Server) in a specific cluster. This means that Applications can now contain devices in different clusters.

This cluster information is already known, as it's registered in the Identity Server. However, the Console doesn't display this information in places where I think it should. From what I can see, the only place where this information is used, is the "general settings" view:

Screen Shot 2021-02-19 at 09 12 18

The End Device lists and overview pages, don't show this information:

| **List** | **Single** |
| --- | --- |
| Screen Shot 2021-02-19 at 09 10 39 | Screen Shot 2021-02-19 at 09 12 38 |

While the Gateway views use this information better:

| **List** | **Single** |
| --- | --- |
| Screen Shot 2021-02-19 at 09 15 21 | Screen Shot 2021-02-19 at 09 15 49 |

- It would be nice if we could show "Other cluster" in the End Device list, like we do for Gateways.
- Instead of "Last seen info unavailable", I'd like to see "Other cluster", like we do for Gateways.

#### Environment

The Things Network, The Things Stack Cloud and other multi-cluster deployments.

I guess some copy-paste from the Gateway UI?

ryaplots:
#### Summary

Closes #3831.

#### Changes

- On device lists, disable clicks
- Show Other host as a status
- Add a tooltip to the status reading
- Redirect back to applications when navigating manually to such end devices
Screenshot 2021-12-07 at 16 02 29

#### Testing

Manually and e2e test.

#### Checklist


ryaplots:
Sill have to fix commit trail

cypress/integration/console/devices/device-on-othercluster.spec.js
```diff
@@ -0,0 +1,147 @@
...
+          application_server_address: 'tti.staging1.cloud.thethings.industries',
+          join_server_address: 'tti.staging1.cloud.thethings.industries',
+        },
+      ],
+    }
+
+    cy.visit(`${Cypress.config('consoleRootPath')}/applications/${applicationId}/devices`)
+    cy.intercept(
+      'GET',
+      `/api/v3/applications/${applicationId}/devices?page=1&limit=20&field_mask=name,application_server_address,network_server_address,join_server_address`,
```
bafonins: Can we avoid specifying all query parameters? See https://docs.cypress.io/api/commands/intercept 
May be we can simplify this to `/api/v3/applications/${applicationId}/devices*`

cypress/integration/console/devices/device-on-othercluster.spec.js
```diff
@@ -0,0 +1,147 @@
...
+      supports_join: true,
+      lorawan_version: 'MAC_V1_0_2',
+      ids: {
+        device_id: deviceId,
+        dev_eui: '70B3D57ED8000019',
+        join_eui: '0000000000000000',
+      },
+      supports_class_c: false,
+      supports_class_b: false,
+      mac_settings: {
```
bafonins, line 46: I think we can remove `mac_settings` to simplify the test

pkg/webui/console/containers/devices-table/index.js
```diff
@@ -57,6 +61,11 @@ import {
 
 import style from './devices-table.styl'
 
+const m = defineMessages({
+  otherClusterTooltip:
+    'This end device is registered on a different cluster or host. It cannot be accessed using this Console. To access this device, use the Console of the cluster that this device was registered on.',
```
bafonins: @kschiffer  Does it makes sense to tell the user where the end device is registered?

pkg/webui/console/containers/devices-table/index.js
```diff
@@ -146,18 +170,32 @@ class DevicesTable extends React.Component {
...
+    const { base_url: stackAsUrl } = asConfig
+    const currentHost = getHostnameFromUrl(stackAsUrl)
+
     for (const device of devices) {
       decoratedDevices.push({
         ...device,
-        _derivedLastSeen: selectDeviceDerivedLastSeen(state, appId, device.ids.device_id),
+        status: {
+          _derivedLastSeen: selectDeviceDerivedLastSeen(state, appId, device.ids.device_id),
+          otherCluster:
```
bafonins: Here it makes sense to create a new selector `isOtherClusterDevice` and implement this logic within that selector so you can reuse in the end device overview

bafonins (commented):
See my comments for the initial review

pkg/webui/console/containers/devices-table/index.js
kschiffer: Yeah, lets add it like this

```suggestion
    'This end device is registered on a different cluster or host (`{host}`). To access this device, use the Console of `{host}`.',
```

ryaplots: It makes sense indeed. 
I have a question related this. Some devices don't provide all servers. For example, a GET request to this end device: `/api/v3/applications/adrian-test-lab/devices/eui-0004a310001ab0b6` with the field masks: `name,description,version_ids,network_server_address,application_server_address,join_server_address,locations,attributes` responds with an object where "ns" and "as" are the same as host but that doesn't contain a `join_server_address`. 

That's why according to the logic implemented on device general settings the join server field appears blocked with the message "Not registered in this cluster" as you can see here:
Screenshot 2021-12-10 at 12 44 08

Does it make sense to block the device in the device table, implementing the same logic? For me a check for if the ns, as and js are defined before comparing them to the current host would make more sense. Because the information of join server address not being provided could mean that the device is simply not registered on that server and not that it's on another cluster. Or am I getting something wrong? @kschiffer @bafonins

kschiffer: This is a Generic Node that is provisioned on an external Join Server. Such end devices should not be hidden because they are still originating from the current cluster or host (unless both AS and NS are registered on different clusters).

See @htdvisser's remark here on when to determine end devices to be on "Other cluster": [https://github.com/TheThingsIndustries/lorawan-stack-support/issues/385#issuecomment-879677908](https://github.com/TheThingsIndustries/lorawan-stack-support/issues/385#issuecomment-879677908)

> For end devices this would happen if all of \`network\_server\_address\`, \`application\_server\_address\` and \`join\_server\_address\` are a different cluster than current

Note the logical AND here. Any server address not being present should remove them from the check.

E.g., an end device that does not have `join_server_address` set should only check for `network_server_address` and `application_server_address` to be on different hosts than the console.

cypress/integration/console/devices/device-on-othercluster.spec.js
```diff
@@ -0,0 +1,147 @@
...
+          },
+          network_server_address: 'tti.staging1.cloud.thethings.industries',
+          application_server_address: 'tti.staging1.cloud.thethings.industries',
+          join_server_address: 'tti.staging1.cloud.thethings.industries',
+        },
+      ],
+    }
+
+    cy.visit(`${Cypress.config('consoleRootPath')}/applications/${applicationId}/devices`)
+    cy.intercept(
```
kschiffer: Why do you use `intercept` as opposed to creating an actual device that is registered on a different cluster?

You could do that with the `cy.createEndDevice()` call, since you will only need to register the device in the IS.

Again, intercept should really be a last resort whenever a request-response cannot be obtained naturally, e.g.

*   when requesting from an external service
*   when requesting from an API that depends on reconfiguring the stack during a test run

If we only mock requests then we are effectively only checking frontend functionality in isolation, while bringing the whole system into the relevant state also allows us to catch issues originating from the backend.

cypress/integration/console/devices/device-on-othercluster.spec.js
```diff
@@ -0,0 +1,147 @@
...
+      'GET',
+      `/api/v3/applications/${applicationId}/devices?page=1&limit=20&field_mask=name,application_server_address,network_server_address,join_server_address`,
+      response,
+    )
+  })
+
+  it('succeeds disabling click on devices that are on another cluster', () => {
+    cy.findByText(deviceId).click()
+    cy.findByText('End devices (1)')
```
kschiffer: This feels a bit like an unnatural assumption to me. I think checking

```javascript
cy.location('pathname').should(
  eq',
  `${Cypress.config('consoleRootPath')}/applications/${appId}`,
)
cy.findByTestId('full-error-view').should('not.exist')
```

is better.

cypress/integration/console/devices/device-on-othercluster.spec.js
```diff
@@ -0,0 +1,147 @@
...
+      network_server_address: 'tti.staging1.cloud.thethings.industries',
+    }
+
+    cy.intercept(
+      'GET',
+      `api/v3/applications/${applicationId}/devices/${deviceId}?field_mask=name,description,version_ids,network_server_address,application_server_address,join_server_address,locations,attributes`,
+      response,
+    )
+
+    cy.findByText('Applications (1)')
```
kschiffer: Same here:

```javascript
cy.location('pathname').should(
  eq',
  `${Cypress.config('consoleRootPath')}/applications/${appId}`,
)
cy.findByTestId('full-error-view').should('not.exist')
```

pkg/webui/console/views/device-overview/index.js
```diff
@@ -45,12 +48,27 @@ const m = defineMessages({
...
-  device: selectSelectedDevice(state),
-}))
+@connect(state => {
+  const device = selectSelectedDevice(state)
+  const asConfig = selectAsConfig()
+  const currentHost = getHostnameFromUrl(asConfig.base_url)
+  const redirect =
+    currentHost !== device.application_server_address ||
+    currentHost !== device.network_server_address ||
+    currentHost !== device.join_server_address
```
kschiffer:
```suggestion
const isThisCluster = 
  (!('application_server_address' in device) || asHost === device.application_server_address) ||
  (!('network_server_address' in device) || nsHost === device.network_server_address) ||
  (!('join_server_address' in device) || jsHost === device.join_server_address) ||
const shouldRedirect = !isThisCluster
```

It should be noted that AS or NS addresses missing are weird edge cases. Also note @bafonins comment on integrating this into a selector. Also note our (implicit) convention of marking boolean values with `should/may/is/…` prefixes.

@htdvisser do you agree to the above as way to determine devices that are on "Other cluster"?

pkg/webui/console/views/device-overview/index.js
```diff
@@ -219,8 +237,13 @@ class DeviceOverview extends React.Component {
   }
 
   render() {
-    const { device } = this.props
+    const { device, redirect } = this.props
     const devIds = device && device.ids
+
+    if (redirect) {
+      return <Redirect to="/applications" />
```
kschiffer: Let's additionally also launch a toast notification reading:

> The end device you attempted to visit is registered on a different host and needs to be accessed using its host Console.

pkg/webui/components/table/index.js
```diff
@@ -118,7 +118,7 @@ class Tabular extends React.Component {
         key={appliedRowKeySelector(row)}
         id={rowIndex}
         onClick={onRowClick}
-        clickable={clickable}
+        clickable={row.status?.otherCluster ? !row.status.otherCluster : clickable}
```
kschiffer: You cannot include end device-related logic into the generic `<Table />` component. The table is a generic UI component that does not care about console-related implementation logic. If you need to make a row unclickable, you'll have to extend the functionality of the `<Table />` component to allow doing so. Some ideas for implementing that would be:

*   Introduce a `_meta` prop for the `row` object that allows setting some row-specific modifiers, e.g. `_meta.clickable = false`
*   Introduce a `rowOptions` prop that has the same length as `rows` to map options to specific indexes
*   Same thing could be done as an object that maps to the row ID.

kschiffer (changes requested):

pkg/webui/console/views/device-overview/index.js
ryaplots: Why OR isn't it supposed to be an AND?

kschiffer: According to Hylke's comment, we should regard devices and registered on other clusters when AS NS and JS are on different clusters. Note that the if statement is basically covering the reverse case, checking whether the device is NOT on another cluster. So whenever any of the component hosts is actually on the current host, the device will be regarded as registered on this cluster.
You can also write it as the reverse case if you want.

ryaplots: I was simply writing:
```
const isOtherCluster =
    asHost !== device.application_server_address &&
    nsHost !== device.network_server_address &&
    jsHost !== device.join_server_address
```
Because if one of them does not exist and therefore the statement evaluates to true, the other statements could evaluate to false (if the cluster is not the same), and isOtherCluster would be false. Which is correct. Also, if the remaining statements both evaluate to true (as not being on the same cluster) isOtherCluster is going to be true. Or is this way ambiguous in what concerns the existence of the address in the device?

kschiffer: Your condition would have marked an end device that is not registered on an AS, NS and JS (no `application_server_address`, `network_server_address`, `join_server_address`) as "other cluster".

However according to the discussion on slack, the correct condition should now be "if none of `*_server_address` are on the current host, consider device as 'On different cluster'". So your condition is correct now.

pkg/webui/components/table/index.js
```diff
@@ -111,29 +111,32 @@ class Tabular extends React.Component {
...
-          return (
-            <Table.DataCell key={index} align={header.align} small={small}>
-              {headers[index].render ? headers[index].render(value) : value}
-            </Table.DataCell>
-          )
-        })}
-      </Table.Row>
-    ))
+    const rows = paginatedData.map((row, rowIndex) => {
+      const rowClickable = !clickable ? false : row._meta?.clickable ?? clickable // If the whole table is disabled each row should be as well.
```
kschiffer:
```suggestion
  // If the whole table is disabled each row should be as well.
  const rowClickable = !clickable ? false : row._meta?.clickable ?? clickable
```

Please extend the proptype definition in L177 for the `data` prop to inform component consumers about the `_meta` prop and possible options in there.

pkg/webui/console/views/device-overview/index.js
```diff
@@ -219,8 +241,17 @@ class DeviceOverview extends React.Component {
...
+    const { device, shouldRedirect } = this.props
     const devIds = device && device.ids
+
+    if (shouldRedirect) {
+      toast({
+        type: toast.types.WARNING,
+        message: m.failedAccessOtherHostDeviceMessage,
+      })
+      return <Redirect to="/applications" />
+    }
```
kschiffer: I think we should rather solve this with the `withFeatureRequirement()` HOC in `console/lib/components/with-feature-requirement.js`. This will allow us to better control redirect behavior as a whole.

pkg/webui/console/containers/devices-table/index.js
```diff
@@ -53,10 +57,16 @@ import {
...
   selectDevicesFetching,
   selectDeviceDerivedLastSeen,
+  isOtherClusterDevice,
 } from '@console/store/selectors/devices'
 
 import style from './devices-table.styl'
 
+const m = defineMessages({
+  otherClusterTooltip:
+    'This end device is registered on a different cluster or host (`{host}`). It cannot be accessed using this Console. To access this device, use the Console of the cluster that this device was registered on.',
```
kschiffer:
```suggestion
    'This end device is registered on a different cluster or host (`{host}`). It cannot be accessed using this Console. To access this device, use the Console of the cluster that this end device was registered on.',
```

Sorry, overlooked this one.

kschiffer (changes requested):

pkg/webui/console/views/device-overview/index.js
```diff
@@ -219,21 +242,30 @@ class DeviceOverview extends React.Component {
...
 
   render() {
-    const { device } = this.props
+    const { device, shouldRedirect } = this.props
     const devIds = device && device.ids
+    const otherwise = {
+      redirect: '/applications',
+      render: toast({
+        type: toast.types.WARNING,
+        message: m.failedAccessOtherHostDeviceMessage,
```
kschiffer: This won't work as expected. The way that the `<Redirect />`\-component currently works is either `redirect` or `render`. Specifying both will give precedence to the redirect. Please make sure to check implementation details or test manually before using modules. I think it makes sense to add a `message` prop to the `otherwise` object to launch a toast message in any case if the requirement is not met.

```javascript
const otherwise = {
  redirect: '/applications',
  message: m.failedAccessOtherHostDeviceMessage,
}
```

You'll probably also want to update the respective cypress test to check for the toast notification.

pkg/webui/console/store/selectors/devices.js
```diff
@@ -50,6 +50,14 @@ export const selectSelectedDevice = state =>
...
 export const selectDeviceFetching = createFetchingSelector(GET_DEV_BASE)
 export const selectDeviceError = createErrorSelector(GET_DEV_BASE)
+export const isOtherClusterDevice = (asHost, nsHost, jsHost, device) => {
+  const isOtherCluster =
+    asHost !== device.application_server_address &&
+    nsHost !== device.network_server_address &&
+    jsHost !== device.join_server_address
+
+  return isOtherCluster
+}
```
kschiffer:
```suggestion
export const isOtherClusterDevice = device => {
  const isOtherCluster =
    getHostnameFromUrl(selectAsConfig().base_url) !== device.application_server_address &&
    getHostnameFromUrl(selectAsConfig().base_url) !== device.network_server_address &&
    getHostnameFromUrl(selectAsConfig().base_url) !== device.join_server_address

  return isOtherCluster
}
```

pkg/webui/components/table/index.js
```diff
@@ -170,8 +174,17 @@ class Tabular extends React.Component {
...
+   * Among them a `_meta` prop for options like clickable or
+   * any other that concern only one row.
+   */
+  data: PropTypes.arrayOf(
+    PropTypes.shape({
+      _meta: PropTypes.shape({
+        clickable: PropTypes.bool,
+      }),
+    }),
+  ),
```
kschiffer:
```suggestion
  /** A list of data entries to display within the table body. */
  data: PropTypes.arrayOf(
    PropTypes.shape({
      /** A meta config object used to control the behavior of individual rows. */
      _meta: PropTypes.shape({
        /** A flag specifying whether the row should be clickable. */
        clickable: PropTypes.bool,
      }),
    }),
  ),
```

kschiffer (changes requested):

pkg/webui/console/views/device-overview/index.js
ryaplots: That's weird. I tested it manually just to make sure and a toast appeared right after the redirect. It doesn't work for you? What could be the reason for the precedence to not be working correctly?

kschiffer: It indeed _seems_ to be working because as part of defining the `otherwise` prop, you are invoking the `toast()` function, which does not return a compnent but in itself will triggers a toast. This also means that the toast will appear for any render of regular end device overviews that do not redirect.

kschiffer:
Please also apply the following diff to fix several issues with the tooltip:

*   Doc link missing leading `/`
*   Tooltip appearance breaking layout
*   Tooltip being styled in right text alignment
*   Tooltip message not converting backticks

```diff
diff --git a/data/lorawan-devices b/data/lorawan-devices
--- a/data/lorawan-devices
+++ b/data/lorawan-devices
@@ -1 +1 @@
-Subproject commit a44436c81b04b62723b5bc428e807090dfc1656a
+Subproject commit a44436c81b04b62723b5bc428e807090dfc1656a-dirty
diff --git a/pkg/webui/components/tooltip/doc.js b/pkg/webui/components/tooltip/doc.js
index e02088ae7..acc91e7ae 100644
--- a/pkg/webui/components/tooltip/doc.js
+++ b/pkg/webui/components/tooltip/doc.js
@@ -27,6 +27,7 @@ import Tooltip from '.'
 
 const DocTooltip = ({ content, docPath, docTitle, interactive, ...rest }) => (
   <Tooltip
+    {...rest}
     interactive
     content={
       <>
@@ -38,7 +39,6 @@ const DocTooltip = ({ content, docPath, docTitle, interactive, ...rest }) => (
         </div>
       </>
     }
-    {...rest}
     appendTo={document.body}
   />
 )
diff --git a/pkg/webui/components/tooltip/index.js b/pkg/webui/components/tooltip/index.js
index 99b7998b9..57738a4a7 100644
--- a/pkg/webui/components/tooltip/index.js
+++ b/pkg/webui/components/tooltip/index.js
@@ -73,7 +73,7 @@ const Tooltip = props => {
       popperOptions={{ modifiers: popperModifiers }}
       delay={delay}
       onShow={handleShow}
-      appendTo={appendTo || interactive ? 'parent' : document.body}
+      appendTo={appendTo ? appendTo : interactive ? 'parent' : document.body}
       animation="fade"
       duration={250}
       hideOnClick={hideOnClick}
diff --git a/pkg/webui/console/containers/devices-table/index.js b/pkg/webui/console/containers/devices-table/index.js
index df22ae97b..692dcce57 100644
--- a/pkg/webui/console/containers/devices-table/index.js
+++ b/pkg/webui/console/containers/devices-table/index.js
@@ -64,7 +64,7 @@ import style from './devices-table.styl'
 
 const m = defineMessages({
   otherClusterTooltip:
-    'This end device is registered on a different cluster or host (`{host}`). It cannot be accessed using this Console. To access this device, use the Console of the cluster that this end device was registered on.',
+    'This end device is registered on a different cluster or host (`{host}`). To access this device, use the Console of the cluster that this end device was registered on.',
 })
 
 const headers = [
@@ -110,8 +110,9 @@ const headers = [
         const host = status.host
         return (
           <DocTooltip
-            docPath="getting-started/cloud-hosted"
-            content={<Message content={m.otherClusterTooltip} values={{ host }} />}
+            docPath="/getting-started/cloud-hosted"
+            content={<Message content={m.otherClusterTooltip} values={{ host }} convertBackticks />}
+            placement="top-end"
           >
             <Status status="unknown" label={sharedMessages.otherCluster}>
               <Icon icon="help_outline" textPaddedLeft small nudgeUp className="tc-subtle-gray" />
```

kschiffer (changes requested):
See above.

kschiffer (commented):
Also please don't forget to add an entry to `CHANGELOG.md`

pkg/webui/console/views/device-overview/index.js
ryaplots: It makes sense. Thanks for explaining. I will add a `message` prop to the `otherwise` object.

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ For details about compatibility between different releases, see the **Commitment
 
 - Increased the maximum gRPC message size to 16MB.
 - Gateways which have been deleted are now automatically disconnected by the Gateway Server.
+- Disabled devices that are registered on a different cluster.
```
```suggestion
- Mark off and hide and restrict access to end devices that are registered on a different cluster in the Console.
```

pkg/webui/console/containers/devices-table/index.js
```diff
@@ -53,10 +56,16 @@ import {
...
   selectDevicesFetching,
   selectDeviceDerivedLastSeen,
+  isOtherClusterDevice,
 } from '@console/store/selectors/devices'
 
 import style from './devices-table.styl'
 
+const m = defineMessages({
+  otherClusterTooltip:
+    'This end device is registered on a different cluster or host (`{host}`). To access this device, use the Console of the cluster that this end device was registered on.',
```
kschiffer:
```suggestion
    'This end device is registered on a different cluster (`{host}`). To access this device, use the Console of the cluster that this end device was registered on.',
```

pkg/webui/console/views/device-overview/index.js
```diff
@@ -43,14 +45,26 @@ const m = defineMessages({
   latestData: 'Latest data',
   rootKeys: 'Root keys',
   keysNotExposed: 'Keys are not exposed',
+  failedAccessOtherHostDeviceMessage:
+    'The end device you attempted to visit is registered on a different host and needs to be accessed using its host Console.',
```
kschiffer:
```suggestion
    'The end device you attempted to visit is registered on a different cluster and needs to be accessed using its host Console.',
```

CHANGELOG.md
kschiffer: Avoid speaking of "devices" only since this can be ambiguous (could also mean gateways). Also, we typically mark Console-related changes with "in the Console" in the changelog.

pkg/webui/console/views/device-overview/index.js
```diff
@@ -43,14 +45,26 @@ const m = defineMessages({
   latestData: 'Latest data',
   rootKeys: 'Root keys',
   keysNotExposed: 'Keys are not exposed',
+  failedAccessOtherHostDeviceMessage:
```
kschiffer:
```suggestion
  failedAccessOtherClusterDevice:
```

Avoid suffixing i18n messages with `Message`, as this is redundant.

kschiffer (approved):
Thank you. Please resolve the remaining comments before merging.

PR #4983, opened 2021-12-10
ci: Add GitHub Actions Workflow for checking Enterprise mergeability

htdvisser:
#### Summary

This adds a GitHub Actions workflow that checks pull requests to TheThingsNetwork/lorawan-stack for merge conflicts against TheThingsIndustries/lorawan-stack

#### Testing

With separate (non-TTS) repositories.

##### Regressions

This shouldn't break any code, and we can ignore CI failures if it doesn't work as expected.

### DIFF
--- a/.github/workflows/mergeability.yml
+++ b/.github/workflows/mergeability.yml
@@ -1,0 +1,34 @@
+name: Pull Request Mergeability
+
+on:
+  pull_request_target:
+    branches:
+      - "v3.*"
+
+jobs:
+
+  # IMPORTANT: THIS JOB USES AN UNTRUSTED CHECKOUT!
+  # DO NOT RUN ANY CODE FROM THE REPOSITORY!
+  check_mergeability:
+    name: Check Mergeability
+    if: ${{ github.repository == 'TheThingsNetwork/lorawan-stack' }}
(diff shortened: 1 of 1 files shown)


adriansmares (approved):

KrishnaIyer (approved):
Beautiful

johanstokking (approved):

.github/workflows/mergeability.yml
```diff
@@ -0,0 +1,34 @@
...
+  # IMPORTANT: THIS JOB USES AN UNTRUSTED CHECKOUT!
+  # DO NOT RUN ANY CODE FROM THE REPOSITORY!
+  check_mergeability:
+    name: Check Mergeability
+    if: ${{ github.repository == 'TheThingsNetwork/lorawan-stack' }}
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Configure Git
+        run: |
+          git config user.name github-actions
```
adriansmares, line 19: I think this needs to be global,  since you're not in a git repository before the checkout.

https://github.com/TheThingsNetwork/lorawan-stack/runs/4519748526?check_suite_focus=true

PR #4985, opened 2021-12-13
Isolate FrequencyPlans from Component definition

nicholaspcr:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/issues/3005

#### Changes

- Update `FrequencyPlans` variable from Component struct, not exported anymore
- Add `FrequencyPlans(ctx)` method to Component

#### Testing

Local testing

##### Regressions

N/A

#### Notes for Reviewers

The objective of this PR is to make way for the implementation of custom frequency plans in the enterprise version

#### Checklist

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -84,7 +84,7 @@ type Component struct {
 
 	fillers []fillcontext.Filler
 
-	FrequencyPlans *frequencyplans.Store
+	frequencyPlans *frequencyplans.Store
 	KeyVault       crypto.KeyVault
 
 	rightsFetcher rights.Fetcher
@@ -178,7 +178,7 @@ func New(logger log.Stack, config *Config, opts ...Option) (c *Component, err er
 	if err != nil {
 		return nil, err
 	}
-	c.FrequencyPlans = frequencyplans.NewStore(fpsFetcher)
+	c.frequencyPlans = frequencyplans.NewStore(fpsFetcher)
 
 	if c.clusterNew == nil {
 		c.clusterNew = cluster.New
@@ -263,6 +263,11 @@ func (c *Component) FromRequestContext(ctx context.Context) context.Context {
 		valueCtx:  ctx,
 		cancelCtx: c.ctx,
 	}
+}
+
+// FrequencyPlansStore returns the component's frequencyPlans Store
+func (c *Component) FrequencyPlansStore(ctx context.Context) (*frequencyplans.Store, error) {
+	return c.frequencyPlans, nil
 }
 
 // Start starts the component.
--- a/pkg/gatewayconfigurationserver/v2/http_frequency_plan.go
+++ b/pkg/gatewayconfigurationserver/v2/http_frequency_plan.go
@@ -24,8 +24,14 @@ import (
 )
 
 func (s *Server) handleGetFrequencyPlan(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
 	frequencyPlanID := mux.Vars(r)["frequency_plan_id"]
-	plan, err := s.component.FrequencyPlans.GetByID(frequencyPlanID)
+	fps, err := s.component.FrequencyPlansStore(ctx)
+	if err != nil {
+		webhandlers.Error(w, r, err)
+		return
+	}
+	plan, err := fps.GetByID(frequencyPlanID)
 	if err != nil {
 		webhandlers.Error(w, r, err)
 		return
--- a/pkg/networkserver/grpc_gsns.go
+++ b/pkg/networkserver/grpc_gsns.go
@@ -192,7 +192,12 @@ func applyCFList(cfList *ttnpb.CFList, phy *band.Band, chs ...*ttnpb.MACParamete
 
 // matchAndHandleDataUplink handles and matches a device prematched by CMACF check.
 func (ns *NetworkServer) matchAndHandleDataUplink(ctx context.Context, dev *ttnpb.EndDevice, up *ttnpb.UplinkMessage, deduplicated bool, cmacFMatchResult cmacFMatchingResult) (*matchResult, bool, error) {
-	phy, err := DeviceBand(dev, ns.FrequencyPlans)
+	fps, err := ns.FrequencyPlansStore(ctx)
+	if err != nil {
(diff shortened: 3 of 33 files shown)


pkg/util/test/frequencyplans.go
```diff
@@ -417,3 +417,14 @@ var (
 func FrequencyPlan(id string) *frequencyplans.FrequencyPlan {
 	return Must(FrequencyPlanStore.GetByID(id)).(*frequencyplans.FrequencyPlan)
 }
+
+func StaticFrequencyPlans() map[string][]byte {
```
adriansmares: Why not a variable that you can then reference in `FrequencyPlansFetcher` above ?

nicholaspcr:
I still have to squash the two commits into one but currently looking for the reason of a test failure

pkg/component/component.go
```diff
@@ -265,6 +258,15 @@ func (c *Component) FromRequestContext(ctx context.Context) context.Context {
 	}
 }
 
+// FrequencyPlansStore returns a FrequencyPlans Store based on the component config and the context provided
+func (c *Component) FrequencyPlansStore(ctx context.Context) (*frequencyplans.Store, error) {
+	fpsFetcher, err := c.config.FrequencyPlansFetcher(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return frequencyplans.NewStore(fpsFetcher), nil
```
htdvisser: We should figure out how we're going to cache this stuff, because like this we're instantiating a new (empty) cache every time something asks for the frequency plans store.

adriansmares: Right now the HTTP fetcher itself does the caching, but I believe that it would be better if `component.HTTPClient` has an option that allows it to attach a cache to the transport, and the cache is global to the component. In that case, even though we're recreating everything, the transport cache is stable.

pkg/gatewayconfigurationserver/v2/http_frequency_plan.go
```diff
@@ -24,8 +24,13 @@ import (
 )
 
 func (s *Server) handleGetFrequencyPlan(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
 	frequencyPlanID := mux.Vars(r)["frequency_plan_id"]
-	plan, err := s.component.FrequencyPlans.GetByID(frequencyPlanID)
+	fp, err := s.component.FrequencyPlansStore(ctx)
+	if err != nil {
+		webhandlers.Error(w, r, err)
```
adriansmares, line 31: Always `return` after `webhandlers.Error`, otherwise execution continues.

pkg/networkserver/grpc_gsns.go
```diff
@@ -287,7 +292,12 @@ func (ns *NetworkServer) matchAndHandleDataUplink(ctx context.Context, dev *ttnp
 				}
 				ctx = log.NewContextWithField(ctx, "f_cnt_reset", true)
 
-				macState, err := mac.NewState(dev, ns.FrequencyPlans, ns.defaultMACSettings)
+				fps, err := ns.FrequencyPlansStore(ctx)
```
adriansmares: You're in the same function (`matchAndHandleDataUplink`), do you need to refresh this here ?

pkg/packetbrokeragent/agent.go
```diff
@@ -303,7 +308,7 @@ func New(c *component.Component, conf *Config, opts ...Option) (*Agent, error) {
 			messageEncrypter:    a,
 			contextDecoupler:    a,
 			tenantIDExtractor:   a.tenantIDExtractor,
-			frequencyPlansStore: a.FrequencyPlans,
+			frequencyPlansStore: fps,
```
adriansmares: This needs to become a function that takes a `context.Context` and returns `(frequencyPlanStore, error)`. The context here won't contain the tenant ID.

pkg/packetbrokeragent/agent.go
```diff
@@ -291,6 +291,12 @@ func New(c *component.Component, conf *Config, opts ...Option) (*Agent, error) {
 	} else {
 		a.grpc.pba = &disabledServer{}
 	}
+
+	// get Frequency Plans Store
+	getFPS := func(ctx context.Context) (frequencyPlansStore, error) {
```
adriansmares: `getFrequencyPlanStore` without a comment is ok too.

pkg/packetbrokeragent/grpc_gspba.go
```diff
@@ -153,18 +156,20 @@ func (s *gsPbaServer) UpdateGateway(ctx context.Context, req *ttnpb.UpdatePacket
...
-		fps := make([]*frequencyplans.FrequencyPlan, 0, len(req.Gateway.FrequencyPlanIds))
-		var err error
-		for _, fpID := range req.Gateway.FrequencyPlanIds {
-			var fp *frequencyplans.FrequencyPlan
-			fp, err = s.frequencyPlansStore.GetByID(fpID)
-			if err != nil {
-				break
+		fpSlice := make([]*frequencyplans.FrequencyPlan, 0, len(req.Gateway.FrequencyPlanIds))
+		fps, err := s.frequencyPlansStore(ctx)
+		if err == nil {
```
adriansmares: Propagate the error here. This is not the kind of failure that should be hidden (since it shouldn't occur at all).

adriansmares (commented):
So modulo some naming nits, this looks good to me now. With that being said, as @htdvisser added, we need to solve the constant recreation and caching. I'll comment in the original issue in order to keep the trail here clean.

pkg/packetbrokeragent/agent.go
nicholaspcr: I choose a smaller name since is a function that is created and used in a small scope, but I'll agree that it's probably more readable if it is a more complete name. So I'll change it

kschiffer (approved):
Codeowned LGTM

pkg/component/component.go
nicholaspcr: From last change the `frequencyPlans` is now a unexported variable that still is inside the component but only accessible via the `component.FrequencyPlansStore` method. Basically like it was before, so the recreation of the cache isn't happening anymore but it will be something to look at in the TTI implementation.

adriansmares (approved):

htdvisser (approved):

Issue #3513, opened 2020-11-20
Add support for new RP002 regional parameters specifications


kschiffer:
#### Summary
We don't offer support for the [recently released regional parameters specifications](https://lora-alliance.org/search/RP2) `RP002-1.0.0`, `RP002-1.0.1`, `RP002-1.0.2` yet. Notably, these are applicable to all MAC versions. This is especially relevant for LoRaWAN 1.0.4, for which we currently don't provide a fitting PHY version yet (it needs at least `RP002-1.0.1`).

#### Why do we need this?
To support all possible end device certifications.

#### What is already there? What do you see now?
Support for `RP001` specifications (and the coupled one of `TS001`).

#### What is missing? What do you want to see?
Support for the new specifications `RP002-1.0.0`, `RP002-1.0.1`, `RP002-1.0.2`, `RP002-1.0.3`

#### How do you propose to implement this?
Backend
- Add enums to our lorawan protos
- Add definitions to our `pkg/band` package
- Likely some other things

Frontend
- Update PHY version constants in the Console
- Enable selection of new RPs in the end device creation
- Move towards using the official specification name, where applicable

#### How do you propose to test this?
Unit tests.

#### Can you do this yourself and submit a Pull Request?
No.

johanstokking:
We would also like to see LoRa FHSS support. This means four new data rates for EU868:

![image](https://user-images.githubusercontent.com/13334001/110661083-7ddada80-81c4-11eb-80a4-facb4fe0bbf9.png)

And two in US915. Other regions are coming.

kschiffer:
Also note that we really should adopt the naming of the official specification document `RP002-1.0.0`, `RP002-1.0.1`, `RP002-1.0.2`, as opposed to arbitrary `MAC 1.0.0` `PHY 1.0.0` etc.

johanstokking:
Yes, I agree

htdvisser:
This is blocking https://github.com/TheThingsNetwork/lorawan-frequency-plans/pull/30 and https://github.com/TheThingsNetwork/lorawan-frequency-plans/pull/32

johanstokking:
Blocking https://github.com/TheThingsNetwork/lorawan-stack/issues/3806

htdvisser:
Blocking https://github.com/TheThingsNetwork/lorawan-frequency-plans/issues/33 as well

johanstokking:
@adriansmares would be great if you can pick this up.

[...]

PR #4986, opened 2021-12-13
RP2 1.0.1 support

### LINKED ISSUE #4963: Error: RP002_V1_0_1 does not exist for EU868
#### Summary

It seems that RP002-1.0.1 is not supported for LoRaWAN v1.0.4. Note that the LoRaWAN certification v1.0.4 is based on this version of the Regional Parameter document.
According to the example files for the "device repository", this specification should be  the correct one for LoRaWAN v1.0.4. Please refer to "windsensor-profile.yaml":
LoRaWAN Regional Parameters version. Values depend on the LoRaWAN version:
1.0:   TS001-1.0
1.0.1: TS001-1.0.1
1.0.2: RP001-1.0.2 or RP001-1.0.2-RevB
1.0.3: RP001-1.0.3-RevA
1.0.4: RP002-1.0.0 or RP002-1.0.1
1.1:   RP001-1.1-RevA or RP001-1.1-RevB

#### Steps to Reproduce

Try to create a new device from the device repository configured with those versions: e.g. IMST GmbH, Wireless M-Bus Range Extender, FW V1.1

image

...

#### Environment

...

RP002-1.0.1 should be supported for LoRaWAN v1.0.4-

...

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/3513
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4963

#### Changes

- `AS923 modified to support multiple groups of default/join channels. Each country/band supports a specific configuration based on an offset from the original AS923 default/join channels. Country summary table updated to indicate support.`
- `Maximum Payload Size for AS923, DataRate 2 was increased from 59 to 123 for UplinkDwellTime = 0 and DownlinkDwellTime = 0.`
- `CN470-510 modified to reflect most recent regulatory requirements. Specifically, SF12 is no longer available and maximum payload sizes for several other datarates were modified to comply with the 1 second dwell time. Further, a 500kHz LoRa datarate and an FSK datarate were added.`

#### Testing

Unit testing.

##### Regressions

N/A. This is purely new functionality.

#### Checklist


pkg/band/all.go
```diff
@@ -143,10 +163,17 @@ func Get(id string, version ttnpb.PHYVersion) (Band, error) {
...
-
 // GetLatest returns the latest version of the band if it was found,
 // and returns an error otherwise.
 func GetLatest(id string) (Band, error) {
-	return Get(id, latestSupportedVersion)
+	versions, ok := All[id]
+	if !ok {
+		return Band{}, errBandNotFound.WithAttributes("id", id, "version", "latest")
+	}
+	for i := ttnpb.RP002_V1_0_3; i >= ttnpb.TS001_V1_0; i-- {
```
johanstokking: I think we need to keep this in a slice. There may be patches needed in older versions which are not necessarily the latest version for the band.

johanstokking (approved):
LGTM but see comment

johanstokking:
Officially, AS923 has been renamed to AS923-1. I'm fine keeping AS923 for compatibility reasons, but only if that is necessary.

pkg/band/all.go
```diff
@@ -143,10 +163,17 @@ func Get(id string, version ttnpb.PHYVersion) (Band, error) {
...
+		if b, ok := versions[i]; ok {
+			return b, nil
+		}
+	}
+	return Band{}, errBandNotFound.WithAttributes("id", id, "version", "latest")
```
htdvisser: Wouldn't it be better to explicitly indicate the "latest" version somewhere? I don't think we should keep using proto numbering.

htdvisser (approved):
LGTM

pkg/band/all.go
htdvisser: Okay I guess this is the same comment as @johanstokking just wrote a minute ago

adriansmares:
The reason for the loop is that we don't want to break for example 'legacy' CN470-510 gateways / frequency plans and so forth, since for these there never will be a `CN470_510@RP2-1.0.1`. Same reason why I didn't add AS923-1 as a new band but instead appended to AS923. I will keep a mapping for each band ID to the latest version.

johanstokking:
I mean that you shouldn't rely on the enum value when comparing old to new.

> Same reason why I didn't add AS923-1 as a new band but instead appended to AS923.

It's effectively a rename:
Screenshot 2021-12-14 at 11 32 19



> I will keep a mapping for each band ID to the latest version.

That's also fine

johanstokking (approved):

PR #5001, opened 2021-12-15
Cleanup HTTP clients usage

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/issues/3005

#### Changes

- Extract the component TLS configuration generators (`GetTLSServerConfig` / `GetTLSClientConfig`) to `pkg/config/tlsconfig`
   - This allows other packages to use TLS configuration options without creating a cycle with the `component` package
- Extract the component HTTP client generators (`HTTPClient` / `HTTPTransport`) to `pkg/httpclient`
   - Same rationale
- Remove the custom `*http.Client` that we've been adding to configs over time. We just propagate a `httpclient.Provider` down to the call site now.

#### Testing

Unit testing.

##### Regressions

Shouldn't occur.

#### Checklist

### DIFF
--- a/pkg/config/tlsconfig/provider.go
+++ b/pkg/config/tlsconfig/provider.go
@@ -1,0 +1,100 @@
+// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tlsconfig
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/log"
+)
+
+// Option provides customization for TLS configuration.
+type Option interface {
+	apply(*tls.Config)
+}
+
+// ConfigOptionFunc is a Option.
+type ConfigOptionFunc func(*tls.Config)
+
+func (fn ConfigOptionFunc) apply(c *tls.Config) {
+	fn(c)
+}
+
+// WithTLSClientAuth sets TLS client authentication options.
+func WithTLSClientAuth(auth tls.ClientAuthType, cas *x509.CertPool, verify func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error) Option {
+	return ConfigOptionFunc(func(c *tls.Config) {
+		c.ClientAuth = auth
+		c.ClientCAs = cas
+		c.VerifyPeerCertificate = verify
+	})
+}
+
+// WithTLSCertificates sets TLS certificates.
+func WithTLSCertificates(certificates ...tls.Certificate) Option {
+	return ConfigOptionFunc(func(c *tls.Config) {
+		c.Certificates = certificates
+	})
+}
+
+// WithNextProtos appends the given protocols to NextProtos.
+func WithNextProtos(protos ...string) Option {
+	return ConfigOptionFunc(func(c *tls.Config) {
+		c.NextProtos = append(c.NextProtos, protos...)
+	})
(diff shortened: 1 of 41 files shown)


pkg/httpclient/http.go
```diff
@@ -0,0 +1,130 @@
...
+
+// Option is an option for HTTP clients.
+type Option func(*httpClientOptions)
+
+type httpClientOptions struct {
+	cache            bool
+	tlsConfig        *tls.Config
+	tlsConfigOptions []tlsconfig.Option
+}
+
```
nicholaspcr, line 51: Why not make the `httpClientOptions` exported? That way other packages could implement their own private funcs that could be passed in the instantiation of a new client. I understand that most operations functions that alter the http client would stay in the package but it does bring extensibility to it.

Don't know if it's necessary, but the following code example might give a better idea of what I'm saying.
https://gist.github.com/nicholaspcr/e647607ad7ca20c6bb14451ad62665a8

pkg/httpclient/http.go
```diff
@@ -0,0 +1,130 @@
...
+		Timeout:   defaultHTTPClientTimeout,
+		Transport: rt,
+	}, nil
+}
+
+// NewProvider constructs a Provider on top of the provided TLS configuration provider.
+func NewProvider(tlsConfigProvider TLSClientConfigurationProvider) Provider {
+	return &provider{tlsConfigProvider: tlsConfigProvider}
+}
+
```
nicholaspcr, line 119: Wouldn't it be possible to have the tls configuration as an `Option` for the `HTTPClient` method? That way you wouldn't need the Provider interface, but rather the `tls` package could have a public function that would create the `Option` func for the creation of the `httpclient` instance.

It's a small change but you would remove the `tlsconfig` definitions/import from the package.

Something like:
```go
// in the `tlsconfig` pkg 
func (p ConfigurationProvider) GetTLSClientConfig(ctx context.Context, opts ...Option) (httpclient.Option, error)

// in component/http.go
func (c *Component) HTTPClient(ctx context.Context, opts ...httpclient.Option) (*http.Client, error) {
        tlsOpt, err := GetTLSClientConfig(ctx, opts...)
	return httpclient.HTTPClient(ctx, tlsOpt)
}
```

pkg/httpclient/http.go
adriansmares: Since we basically have a monoropo, if we need new options we just add them to the package itself - we don't really extend via 'inheritance'. Also keep in mind that the options that you provided can be applied even after the client has been built, which is not concurrency safe. This is the reason why they generally operate using an unexported function / struct.

pkg/httpclient/http.go
adriansmares: The currently implemented variant allows callers to request different `tlsconfig.Option` directly on `HTTPClient` using the `WithTLSConfigOptions` option. If we explicitly construct the `*tls.Config` in the component, you cannot actually pass these without reinventing `WithTLSConfigOptions`, which makes `httpclient` depened on `tlsconfig`.

pkg/httpclient/http.go
nicholaspcr: True, if exported there would be the potential of using the function outside the `httpclient` method, specially since it might not be intuitive to use. Then the current implementation looks good to me.

pkg/httpclient/http.go
nicholaspcr: I suppose my question was an extension of the idea of making the options struct exported, that way these functions wouldn't need to exist. Since the initial idea is quite troublesome due to the capability of being used outside the pkg, I believe this comment doesn't make sense anymore.

pkg/config/tlsconfig/provider.go
```diff
@@ -0,0 +1,100 @@
...
+type ConfigurationProvider func(context.Context) Config
+
+// GetTLSServerConfig gets the component's server TLS config and applies the given options.
+func (p ConfigurationProvider) GetTLSServerConfig(ctx context.Context, opts ...Option) (*tls.Config, error) {
+	conf := p(ctx)
+	cipherSuites, err := conf.GetCipherSuites()
+	if err != nil {
+		return nil, err
+	}
+	if cipherSuites != nil {
```
johanstokking, line 70: Nit; probably a length check is more appropriate here

johanstokking (approved):
Great stuff

htdvisser (approved):

kschiffer (approved):
codeowned LGTM

PR #5011, opened 2021-12-17
Use peer IP for rate-limiting if x-real-ip is not set

KrishnaIyer:
#### Summary

Currently, we just assume that all requests coming in from "trusted IP" ranges are from proxies and we get the `x-real-ip` from `x-forwarded-for`.
However, when running the stack locally, requests from the cli are from `127.0.0.1` which is a trusted IP but is not from a proxy. Without this header, rate-limiting doesn't work.
This PR checks that x-real-ip is always set by the proxy interceptor.

#### Changes

- Check that x-real-ip is always set by the proxy interceptor.

#### Testing

local

##### Regressions

I don't expect any.

#### Checklist

### DIFF
--- a/pkg/ratelimit/grpc.go
+++ b/pkg/ratelimit/grpc.go
@@ -25,6 +25,7 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/unique"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/peer"
 )
 
 // grpcRemoteIP retrieves the remote IP address by the X-Real-IP header. The header is set by rpcmiddleware.ProxyHeaders
@@ -32,6 +33,9 @@ func grpcRemoteIP(ctx context.Context) string {
 	md, _ := metadata.FromIncomingContext(ctx)
 	if v := md.Get("x-real-ip"); len(v) > 0 {
 		return v[0]
+	}
+	if p, ok := peer.FromContext(ctx); ok {
+		return p.Addr.String()
 	}
 	return ""
 }


adriansmares:
I think that the correct solution is to actually make the ratelimitting middleware take into account the peer IP if the header was not set:

```diff
diff --git a/pkg/ratelimit/grpc.go b/pkg/ratelimit/grpc.go
index a0d2ade9a..e5e30bd61 100644
--- a/pkg/ratelimit/grpc.go
+++ b/pkg/ratelimit/grpc.go
@@ -25,6 +25,7 @@ import (
        "go.thethings.network/lorawan-stack/v3/pkg/unique"
        "google.golang.org/grpc"
        "google.golang.org/grpc/metadata"
+       "google.golang.org/grpc/peer"
 )
 
 // grpcRemoteIP retrieves the remote IP address by the X-Real-IP header. The header is set by rpcmiddleware.ProxyHeaders
@@ -33,6 +34,9 @@ func grpcRemoteIP(ctx context.Context) string {
        if v := md.Get("x-real-ip"); len(v) > 0 {
                return v[0]
        }
+       if p, ok := peer.FromContext(ctx); ok {
+               return p.Addr.String()
+       }
        return ""
 }
```
Otherwise every client will look like its behind a proxy, which is not necessarily true.

htdvisser (commented):
I agree with @adriansmares. There isn't really a specification for the `X-Real-Ip` header, but usually it indicates that "the remote IP of the TCP connection isn't the actual IP". So things that operate on the caller IP should not rely only on the `X-Real-Ip` header, but fall back to the actual remote address when there is no `X-Real-Ip`.

adriansmares (approved):

htdvisser (approved):
The pull request description is out of date, but the code changes lgtm.

PR #5049, opened 2021-12-27
Make CLI retry when it fails due to rate limitations

### LINKED ISSUE #4902: Make CLI retry requests when they failed due to rate limitations
#### Summary
The CLI currently errors out on rate-limiting errors when it should rather wait for the cooldown and retry the request.

#### Why do we need this?
To allow for batch processes under the limitations of rate limiting rules.

Various CLI commands that call RPCs

The CLI reacting appropriately to rate limits.

In the frontend we already apply a middleware that checks for rate-limiting errors and based on the provided information in the response header, waits for a retry of the request.

Manually (?)

I think people more invested into the stack CLI are better suited for this

nicholaspcr:
#### Summary

Closes #4902 

#### Changes

- Add retry interceptor to cli client
- Add retry settings to cli config

#### Testing

- Manual testing - default timeout provided by the cli config
Screenshot from 2021-12-27 13-18-23
- Manual testing - testing the timeout generated from the rate limiter header
*Still have to finish testing it to get the screenshot*

#### Notes for Reviewers

Originally the issue specifies that a retry should happened only when the requests fails due to rate limitations but upon looking into it I opted to go for a more broad approach, the retry triggers according to the grcp status code returned in the request (`ResourceExhausted` and `Unavailable`).

More details will be on the issue.

#### Checklist


htdvisser (commented):
Can't we use the `x-rate-limit-limit`, `x-rate-limit-available`, `x-rate-limit-reset` and `x-rate-limit-retry` headers that the server returns to make a more informed decision about when we can retry the call?

Can't we also use those headers to slow down repeated (asBulk) calls _before_ we hit the rate limit?

nicholaspcr:
> Can't we use the `x-rate-limit-limit`, `x-rate-limit-available`, `x-rate-limit-reset` and `x-rate-limit-retry` headers that the server returns to make a more informed decision about when we can retry the call?
> 
> Can't we also use those headers to slow down repeated (asBulk) calls _before_ we hit the rate limit?

Using those headers is a better solution to determine the timeout in a efficient way, will work on that.

The bulk idea is not clear to me, so I want to confirm if my interpretation is correct. 
The idea would be to add a timeout not only on a request failure but in between the commands of the `asBulk`, is that correct?

cmd/ttn-lw-cli/internal/api/grpc.go
```diff
@@ -107,6 +125,34 @@ func GetDialOptions() (opts []grpc.DialOption) {
 	if withDump {
 		opts = append(opts, grpc.WithChainUnaryInterceptor(requestInterceptor))
 	}
+
+	callOpts := []grpc_retry.CallOption{
+		grpc_retry.WithBackoffContext(
+			func(ctx context.Context, attempt uint) time.Duration {
+				logger := log.FromContext(ctx)
+				logger.Debugf("Failed request, retry attempt %v", attempt)
```
nicholaspcr, line 133: This is before getting the feedback on the use of jitter but I'm rethinking on how to log this information.

My current idea is to do something like:
`				logger.WithField("attempt", attempt).Debug("Failed request, waiting approximately %v until next attempt", attempt)`

Would like to ask if it is appropriate to leave the `approximately` in the log to give an idea about the jitter.

nicholaspcr:
> > Can't we use the `x-rate-limit-limit`, `x-rate-limit-available`, `x-rate-limit-reset` and `x-rate-limit-retry` headers that the server returns to make a more informed decision about when we can retry the call?
> > Can't we also use those headers to slow down repeated (asBulk) calls _before_ we hit the rate limit?
> 
> Using those headers is a better solution to determine the timeout in a efficient way, will work on that.
> 
> The bulk idea is not clear to me, so I want to confirm if my interpretation is correct. The idea would be to add a timeout not only on a request failure but in between the commands of the `asBulk`, is that correct?

While thinking on the idea of slowing down the bulk calls I opted to use the Exponential backoff and use the jitter to avoid having multiple requests being retried at the same time. 

@htdvisser If possible, let me know what you think about this implementation.

nicholaspcr:
I was testing a bit more and realised that the current implementation might have a bit of problem to obtain values from the headers when retrying a request. My bad, while I look into it I'll leave the PR as a draft to make it clear that is not ready.

nicholaspcr:
Closing the PR to avoid more notifications every time I push it, will open again after the issue #5080 is solved and everything is tested.

PR #5077, opened 2022-01-10
Configure basic auth

### LINKED ISSUE #4835: Make it easier to configure Basic Auth for webhook integrations in the Console
#### Summary

We need to make it easier to configure Basic auth for webook integrations.

#### Why do we need this?

Basic Auth is a very common auth method for (self-built) integrations, so it would be nice if we could make the Console handle this as a special case. 

All headers are treated the same.

The `Authorization` header should get special treatment for `Basic` auth.

- If `headers` doesn't already have an `Authorization` key with a value that starts with `Basic`, show a button to add one.
- If `headers` does have one, base64-decode the value, split on the first `:` and display Username and Password fields.
- When the Username and/or Password fields change, update the header to the base64-encoded concatenation of Username, `:` and Password.

ryaplots:
#### Summary

This pull request closes #4835, by adding a basic auth configuration field to webhook form 

#### Changes

- Add basic auth configuration field
Screenshot 2022-01-04 at 17 29 35
Screenshot 2022-01-04 at 17 27 40

- Add warning when the user is trying to add another Authorization header
Screenshot 2022-01-10 at 09 52 25

#### Testing

Manually

#### Checklist


ryaplots:
Will clean commit trail before merging

pkg/webui/components/key-value-map/index.js
```diff
@@ -109,20 +108,26 @@ class KeyValueMap extends React.PureComponent {
...
+                    index={index}
+                    onRemoveButtonClick={this.removeEntry}
+                    onChange={this.handleEntryChange}
+                    onBlur={onBlur}
+                    indexAsKey={indexAsKey}
+                  />
+                )
+              }
+
+              return <></>
```
kschiffer:
```suggestion
              return null
```

pkg/webui/components/key-value-map/index.js
```diff
@@ -109,20 +108,26 @@ class KeyValueMap extends React.PureComponent {
...
-                valuePlaceholder={valuePlaceholder}
-                index={index}
-                onRemoveButtonClick={this.removeEntry}
-                onChange={this.handleEntryChange}
-                onBlur={onBlur}
-                indexAsKey={indexAsKey}
-              />
-            ))}
+            value.map((value, index) => {
+              if (!(value.key === 'Authorization' && value.value?.startsWith('Basic'))) {
```
kschiffer: You cannot implement this as part of the component. This is again putting business logic into a generic component. We are using the key-value map for other purposes than entering request headers as well, e.g. for attributes. With your change, it would be impossible to set an `Auth` `Basic xxx` attribute for no reason.

pkg/webui/console/components/webhook-form/index.js
```diff
@@ -87,6 +93,11 @@ const validationSchema = Yup.object().shape({
     .matches(webhookIdRegexp, Yup.passValues(sharedMessages.validateIdFormat))
     .required(sharedMessages.validateRequired),
   format: Yup.string().required(sharedMessages.validateRequired),
+  basic_auth: Yup.object().shape({
+    value: Yup.boolean(),
```
kschiffer:
```suggestion
    enabled: Yup.boolean(),
```

pkg/webui/console/components/webhook-form/index.js
```diff
@@ -87,6 +93,11 @@ const validationSchema = Yup.object().shape({
     .matches(webhookIdRegexp, Yup.passValues(sharedMessages.validateIdFormat))
     .required(sharedMessages.validateRequired),
   format: Yup.string().required(sharedMessages.validateRequired),
+  basic_auth: Yup.object().shape({
```
kschiffer:
```suggestion
  _basic_auth: Yup.object().shape({
```

We usually mark meta fields (as in fields that do not map directly to final data shape) with an underscore. See my comment below on using encode/decode on the field level to possibly avoid using this altogether.

pkg/webui/console/components/webhook-form/index.js
```diff
@@ -208,17 +223,29 @@ export default class WebhookForm extends Component {
...
 
   @bind
   async handleSubmit(values, { resetForm }) {
     const { appId, onSubmit, onSubmitSuccess, onSubmitFailure, existCheck, update } = this.props
+    // Ensure that the basic auth header is deleted properly.
+    if (values.basic_auth && values.basic_auth.value === false) {
+      values.headers = values.headers.filter(
+        header => header.key !== 'Authorizarion' && !header.value.startsWith('Basic'),
+      )
+    }
```
kschiffer: This would wrongfully also remove a basic auth header that was added manually.

What would be best is to still show the basic auth header, but as read-only field.

pkg/webui/console/components/webhook-form/index.js
```diff
@@ -208,17 +223,29 @@ export default class WebhookForm extends Component {
...
-    existingId: undefined,
+  constructor(props) {
+    super(props)
+    const { initialWebhookValue } = this.props
+
+    this.state = {
+      error: undefined,
+      displayOverwriteModal: false,
+      existingId: undefined,
+      mayShowCredentialsInput: initialWebhookValue.headers.Authorization?.startsWith('Basic'),
```
kschiffer:
```suggestion
      mayShowCredentialsInput: initialWebhookValue.headers.Authorization?.startsWith('Basic'),
```

Appears to me like this will error if `Authorization` evaluates to `undefined`.

pkg/webui/console/components/webhook-form/index.js
```diff
@@ -363,12 +406,42 @@ export default class WebhookForm extends Component {
...
+          <Form.Field
+            autoFocus
+            title={m.requestBasicAuth}
+            name="basic_auth.value"
+            label={m.basicAuthCheckbox}
+            onChange={this.handleRequestAuthenticationChange}
+            component={Checkbox}
+            tooltipId={tooltipIds.BASIC_AUTH}
+          />
+          {this.state.mayShowCredentialsInput && (
```
kschiffer:
```suggestion
          {this.state.shouldShowCredentials && (
```

pkg/webui/console/components/webhook-form/mapping.js
```diff
@@ -59,35 +86,51 @@ const mapHeadersTypeFormValueToWebhookHeadersType = formValue =>
...
-  downlink_queued: mapMessageTypeFormValueToWebhookMessageType(values.downlink_queued),
-  downlink_queue_invalidated: mapMessageTypeFormValueToWebhookMessageType(
-    values.downlink_queue_invalidated,
-  ),
-  location_solved: mapMessageTypeFormValueToWebhookMessageType(values.location_solved),
-  service_data: mapMessageTypeFormValueToWebhookMessageType(values.service_data),
-})
+    base_url: values.base_url,
+    format: values.format,
+    basic_auth: values.basic_auth,
```
kschiffer:
```suggestion
```
It's not necessary to add this here, since the webhook structure is unaware of a `basic_auth` prop.

pkg/webui/console/components/webhook-form/mapping.js
```diff
@@ -27,10 +27,37 @@ const mapWebhookHeadersTypeToFormValue = headersType =>
...
+
+    return decodedBasicAuth
+  }
+}
+
 export const mapWebhookToFormValues = webhook => ({
   webhook_id: webhook.ids.webhook_id,
   base_url: webhook.base_url,
   format: webhook.format,
+  basic_auth: decodeWebhookBasicAuth(webhook.headers),
```
kschiffer:
```suggestion
  _basic_auth: decodeWebhookBasicAuth(webhook.headers),
```

pkg/webui/console/components/webhook-form/mapping.js
```diff
@@ -27,10 +27,37 @@ const mapWebhookHeadersTypeToFormValue = headersType =>
...
   headers: mapWebhookHeadersTypeToFormValue(webhook.headers),
```
kschiffer: You will also need to extract the possible basic auth header here.

pkg/webui/console/components/webhook-form/index.js
```diff
@@ -363,12 +406,42 @@ export default class WebhookForm extends Component {
             sensitive
             code
           />
+          {this.state.hasAuthorization && (
+            <Notification content={m.hasAuthorization} small warning />
+          )}
+          <Form.Field
+            autoFocus
+            title={m.requestBasicAuth}
+            name="basic_auth.value"
```
kschiffer:
```suggestion
            name="headers"
            encode={...}
            decode={...}
```

I believe it would make things a lot easier if you use the encode decode functions here to marshal the different data shapes.

pkg/webui/console/components/webhook-form/index.js
```diff
@@ -363,12 +406,42 @@ export default class WebhookForm extends Component {
             sensitive
             code
           />
+          {this.state.hasAuthorization && (
+            <Notification content={m.hasAuthorization} small warning />
+          )}
```
kschiffer: The warning is not needed, though it can still make sense to generally show a warning when using duplicate warnings.

kschiffer (changes requested):
There are still quite some issues with the implementation. After looking at this more in-depth, I suggest the following:
- Use value encoding/decoding on field level to realize basic auth fields
- Show the generated basic auth header as read-only fields on the `headers`-field

pkg/webui/console/components/webhook-form/mapping.js
```diff
@@ -59,30 +59,34 @@ const mapHeadersTypeFormValueToWebhookHeadersType = formValue =>
     )) ||
   null
 
-export const mapFormValuesToWebhook = (values, appId) => ({
```
kschiffer, line 62: The diff in this file can be removed entirely.

cypress/integration/console/integrations/webhooks/edit.spec.js
```diff
@@ -30,6 +30,10 @@ describe('Application Webhook', () => {
   const webhookBody = {
     webhook: {
       base_url: 'https://example.com/edit-webhooks-test',
+      basic_auth: {
+        password: '',
+        username: '',
+      },
```
kschiffer: This is not necessary anymore.

kschiffer:
Just testing and noticed that the page crashes when manually entering a header with the key `Authorization`.

ryaplots:
> Just testing and noticed that the page crashes when manually entering a header with the key `Authorization`.

Will check it out as soon as I can

pkg/webui/console/components/webhook-form/index.js
```diff
@@ -363,13 +480,46 @@ export default class WebhookForm extends Component {
             sensitive
             code
           />
+          <Form.Field
+            autoFocus
```
kschiffer:
```suggestion
```
Autofocus only makes sense for fields that are at the beginning of a form (or likely to be entered first)

pkg/webui/console/components/webhook-form/index.js
```diff
@@ -165,6 +169,96 @@ const validationSchema = Yup.object().shape({
...
+      decodedCredentials.indexOf(':') + 1,
+      decodedCredentials.length,
+    )
+    return decodedPassword
+  }
+
+  return ''
+}
+
+const makeBasicAuthDecoder = forPassword => value => mapCredentialsToAuthHeader(forPassword, value)
```
kschiffer: I think all of this can be moved to the `./mapping` module.

pkg/webui/console/components/webhook-form/index.js
```diff
@@ -165,6 +169,96 @@ const validationSchema = Yup.object().shape({
...
+  if (encodedCredentials) {
+    const decodedCredentials = atob(encodedCredentials)
+    const decodedUsername = decodedCredentials.slice(0, decodedCredentials.indexOf(':'))
+    return decodedUsername
+  }
+
+  return ''
+}
+
+const mapCredentialsToAuthHeader = (forUsername, fieldValue) => {
```
kschiffer:
```suggestion
const createCredentialsToAuthHeaderMapper = forUsername => fieldValue => {
```

See my other comment below.

pkg/webui/console/components/webhook-form/index.js
kschiffer:
```suggestion
const mapCredentialsToPassword = createCredentialsToAuthHeaderMapper(true)
const mapCredentialsToUsername = createCredentialsToAuthHeaderMapper(false)
```

pkg/webui/console/components/webhook-form/index.js
```diff
@@ -363,13 +480,46 @@ export default class WebhookForm extends Component {
...
+            tooltipId={tooltipIds.BASIC_AUTH}
+          />
+          {this.state.shouldShowCredentialsInput && (
+            <Form.FieldContainer horizontal>
+              <Form.Field
+                required
+                title={sharedMessages.username}
+                name="headers"
+                decode={mapBasicAuthHeaderToUsername}
+                encode={makeBasicAuthDecoder(true)}
```
kschiffer: You should define that outside of the component. Otherwise, the function will be recreated for every rerender.
See my comments above.

kschiffer (changes requested):

kschiffer (changes requested):
Please adjust as discussed internally.

kschiffer (changes requested):
There are a couple of merge issues in the diff. Please resolve those and re-request review.

kschiffer (changes requested):
There are still issues, see my video below

https://user-images.githubusercontent.com/5710611/152558429-5a34c47c-e57f-4898-a9e4-0a09cc1c9663.mov

Additionally, the credential fields would not appear automatically, if the auth header was entered manually.

I've set up a [branch with some fix-suggestions of my own](https://github.com/TheThingsNetwork/lorawan-stack/compare/fix/4835-configure-basic-auth-mods). Please consider those for resolving these issues. While working on this I also noticed that keeping the `headers` value as a plain object without any intermediate format is actually not desirable, since this prevents users from typing multiple headers with the same key. While this is also not possible to store such a setup, it should at least be possible to temporarily enter such value. Currently, whenever two headers with the same keys are added, the old value disappears and is instantly replaced with the value that is currently entered.
In the end, I decided to reintroduce the intermediate array form and to de/encode it before sending the data to the server.

pkg/webui/components/key-value-map/index.js
```diff
@@ -103,26 +103,31 @@ class KeyValueMap extends React.PureComponent {
...
+                  name={name}
+                  value={value}
+                  keyPlaceholder={keyPlaceholder}
+                  valuePlaceholder={valuePlaceholder}
+                  index={index}
+                  onRemoveButtonClick={this.removeEntry}
+                  onChange={this.handleEntryChange}
+                  onBlur={onBlur}
+                  indexAsKey={indexAsKey}
+                  readOnly={readOnly}
```
kschiffer:
```suggestion
                  readOnly={isReadonly(value)}
```

No need to break up the arrow short hand for that variable definition.

pkg/webui/console/components/webhook-form/mapping.js
```diff
@@ -59,33 +58,86 @@ const mapHeadersTypeFormValueToWebhookHeadersType = formValue =>
...
+
+export const decodeBasicAuthHeaderPassword = value => {
+  if (hasBasicAuth(value)) {
+    const encodedCredentials = getBasicAuthValue(value)
+    if (encodedCredentials) {
+      const decodedCredentials = atob(encodedCredentials)
+      const decodedPassword = decodedCredentials.slice(
+        decodedCredentials.indexOf(':') + 1,
+        decodedCredentials.length,
+      )
```
kschiffer: What actually happens when the password or username contains a `:`? We need to catch that in the form validation.

> 1.  The username and password are combined with a single colon (:). This means that the username itself cannot contain a colon.

https://en.wikipedia.org/wiki/Basic_access_authentication

pkg/webui/console/components/webhook-form/mapping.js
```diff
@@ -59,33 +58,86 @@ const mapHeadersTypeFormValueToWebhookHeadersType = formValue =>
...
+        newHeaders[basicAuthIndex].value = `Basic ${btoa(`${value}:${password}`)}`
+      } else if (!forUsername && (value !== '' || username !== '')) {
+        newHeaders[basicAuthIndex].value = `Basic ${btoa(`${username}:${value}`)}`
+      } else {
+        newHeaders[basicAuthIndex].value = 'Basic '
+      }
+      return newHeaders
+    }
+  }
+  return currentHeaders
```
kschiffer: Nit pick (I think I introduced that): We usually add an empty line before `return` statements.

kschiffer (changes requested):
Almost there.

When going through the cypress tests for webhooks, I noticed there are some test gaps with regards to saving changes. Can you extend the current test cases for adding and editing webhooks with an assertion on whether the saved changes are actually present in the updated form?

That would be for `succeeds adding webhook` and `succeeds editing webhook`. Additionally, the existing test should be updated with some assertions around basic auth.

I need to be picky about this because webhooks are critical functionality and we need to make sure we don't introduce any issues with this PR.

pkg/webui/console/components/webhook-form/mapping.js
ryaplots: I think the password can contain a `:` because we break the string at the first occurrence of `:`. So the username does indeed need to have a validation that does not allow a semicolon to be part of it. Or am I missing something?

kschiffer: True, this is an issue for the username only, like also the bit on WP says.

kschiffer:
I applied some changes as discussed on slack. The branch should be ready to merge now. Please double-check.

My changes were:
- Remove usage of field level encoders, since as we determined, they cause too many side effects and constraints wrt field validation
- Move encoding/decoding to submit stage
- Remove form/field component adjustments that have become unnecessary after letting go of field level encoding
- Fix validation logic for basic auth credentials
- Some minor wording fixes
- Rebasing and fixup the commit trail

kschiffer (approved):

PR #5087, opened 2022-01-11
Use `NwkKey` when set, regardless of session version

### LINKED ISSUE #2600: Root key usage and compatibility of LoRaWAN 1.1 capable devices
#### Summary

This is a technical decision on the root key usage in LoRaWAN 1.0.x activations on a LoRaWAN 1.1 compatible LNS like The Things Stack V3.

#### Situation

Our Join Server stores NwkKey and AppKey. In LoRaWAN 1.0.x device activations, our Join Server uses the AppKey. However, the LoRaWAN 1.1 specification states that this should be the NwkKey:

Screenshot 2020-05-26 at 17 10 39

This can be interpreted as that AppKey was renamed to NwkKey, and a new AppKey has been introduced that is used for LoRaWAN 1.1 devices.

This means that if an end device sends a join-request, and we know it is a LoRaWAN 1.1 device, but we activate it as a LoRaWAN 1.0.x device (`OptNeg` is not set), we must use NwkKey and not AppKey.

#### Scope

Join Server activations of LoRaWAN 1.0.x.

#### Criteria

1. We should be compatible with LoRaWAN 1.1.x
2. We should not break existing devices that are stored

#### Alternatives

1. No change

This means that we keep using AppKey for LoRaWAN 1.0.x activations. We would not support the case where we know the end device is LoRaWAN 1.1, but we activate it as LoRaWAN 1.0.x device.

This can become problematic if users start onboarding LoRaWAN 1.1 capable devices, that they configure to use as LoRaWAN 1.0.x. In that case, the end device uses NwkKey, while our Join Server uses AppKey for session key derivation. We cannot get away with saying they must configure the highest capable LoRaWAN version, because it is LoRaWAN compliant (also for certification) to activate devices in a lower LoRaWAN version.

---

2. Support activating devices that we know are LoRaWAN 1.1 compliant in LoRaWAN 1.0.x compatibility mode

In this case, we would need to use NwkKey. Also, we would need to separate the capable LoRaWAN version (i.e. 1.1) from the desired LoRaWAN version (i.e. 1.0.3) for activation:

Screenshot 2020-05-26 at 17 26 05

#### Consideration

| Criteria | Alternative 1 | Alternative 2 |
--- | --- | ---
LoRaWAN compliant | Not entirely; we cannot activate LoRaWAN 1.1 capable devices as 1.0.x | Yes
Storage compatible | Yes | Yes, need additional field, like `lorawan_version_capabilities` (stored in JS), next to the desired `lorawan_version` in NS (like now) |

#### Decision

TBD

TBD

johanstokking:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2600

#### Changes

When a `NwkKey` is configured for the end device in the Join Server, it is used when the device is activating in 1.0.x mode.

In other words, when `NwkKey` is present, the end device is assumed to be capable of LoRaWAN 1.1. In that case, the end device will also use its `NwkKey`, not `AppKey`.

#### Testing

Unit testing

##### Regressions

As far as there are LoRaWAN 1.1 devices on the market, they are probably being activated in 1.1 mode. However, if there are 1.1 capable end devices registered in the Join Server, but they are activated by the Network Server in 1.0.x mode, now the `NwkKey` is used and not the `AppKey`. This is a breaking change. 1.1 capable end devices activating in 1.0.x mode relying on usage of `AppKey` rely on faulty behavior.

#### Checklist


pkg/joinserver/joinserver.go
```diff
@@ -389,28 +389,43 @@ func (js *JoinServer) HandleJoin(ctx context.Context, req *ttnpb.JoinRequest, au
...
 
-			var networkCryptoService cryptoservices.Network
-			if req.SelectedMacVersion.UseNwkKey() && dev.RootKeys != nil && dev.RootKeys.NwkKey != nil {
-				// LoRaWAN 1.1 and higher use a NwkKey.
+			// The root keys are used according to the following table:
+			//
+			//  Has NwkKey | Activation | Network uses | Application uses
+			//  ---------- | ---------- | ------------ | ----------------
+			//  No         | 1.0.x      | AppKey       | AppKey
+			//  Yes        | 1.0.x      | NwkKey       | NwkKey
```
johanstokking, line 397: This is the new case; NwkKey should be used here.

The rest is mostly renaming by using NwkKey instead of AppKey for network layer crypto. That follows LoRaWAN 1.1 terminology. When there's no NwkKey and the device activates as 1.0.x, AppKey is filled in for NwkKey.

adriansmares (approved):

adriansmares:
Don't forget a changelog entry.

htdvisser (approved):

PR #5094, opened 2022-01-12
Filter end device events that propagate to application event streams

### LINKED ISSUE #4834: Filter end device event types that are forwarded to applications
#### Summary

We need to filter which end device events are forwarded to the application event stream.

#### Why do we need this?

Because currently event streams can't keep up with large applications.

All end device events are forwarded to application event streams. This means that for applications with thousands of devices the event stream won't be able to keep up, and drop events. This makes the event stream pretty useless.

We can think about a (possibly configurable) set of end device events that get forwarded to the application. My proposal would be to only forward application-layer events (starting with `as.`).

#### Environment

All deployments, but specifically The Things Stack Cloud.

Define a list of end device event names that get forwarded to the application.

Sure, but this first needs some discussion.

htdvisser:
#### Summary

This pull request adds additional filtering to the functionality that propagates end device events to the application event streams.

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4834

#### Changes

- Remove (very chatty) "duplicate dropped" event from NS completely
- Add options to end device events that need to get forwarded to application streams

#### Testing

Existing unit tests, manual testing in the Console.

##### Regressions

@kschiffer to confirm that the Console doesn't rely on events that are no longer in the application stream.

#### Notes for Reviewers

nil

#### Checklist

### DIFF
--- a/pkg/events/redis/store.go
+++ b/pkg/events/redis/store.go
@@ -385,6 +385,7 @@ func (ps *PubSubStore) Publish(evt events.Event) {
 		if len(ids) == 0 {
 			tx.Publish(ps.ctx, ps.eventChannel(evt.Context(), evt.Name(), nil), m)
 		}
+		definition := events.GetDefinition(evt)
 		for _, id := range ids {
 			tx.Publish(ps.ctx, ps.eventChannel(evt.Context(), evt.Name(), id), m)
 			streamValues, err := encodeEventMeta(evt, id)
@@ -399,7 +400,7 @@ func (ps *PubSubStore) Publish(evt events.Event) {
 				Values:       streamValues,
 			})
 			tx.Expire(ps.ctx, eventStream, ps.entityHistoryTTL)
-			if devID := id.GetDeviceIds(); devID != nil {
+			if devID := id.GetDeviceIds(); devID != nil && definition != nil && definition.PropagateToParent() {
 				eventStream := ps.eventStream(evt.Context(), devID.ApplicationIds.GetEntityIdentifiers())
 				tx.XAdd(ps.ctx, &redis.XAddArgs{
 					Stream:       eventStream,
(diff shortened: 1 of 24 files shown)


pkg/events/redis/store.go
```diff
@@ -399,7 +400,7 @@ func (ps *PubSubStore) Publish(evt events.Event) {
 				Values:       streamValues,
 			})
 			tx.Expire(ps.ctx, eventStream, ps.entityHistoryTTL)
-			if devID := id.GetDeviceIds(); devID != nil {
+			if devID := id.GetDeviceIds(); devID != nil && definition != nil && definition.PropagateToParent() {
```
adriansmares, line 403: In which situation wouldn't a `nil` definition be an implementation error ? (i.e. publishing something with an unregistered name). 

I can understand how between versions you may not be able to have definition while _receiving an event_, but how would that happen on `Publish` ?

adriansmares (approved):

pkg/events/redis/store.go
htdvisser: In practice this shouldn't happen, but it is possible:

```go
events.Publish(events.New(ctx, "some.event", "some event", events.WithIdentifiers(...)))
```

johanstokking (approved):

kschiffer:
This change would break live updates of the frame counts in the Console. I need to make adjustments to the Console code to source this data from the event stream instead.

htdvisser:
Re-assigning since my part is done. What's still TODO is to update the Console accordingly.

kschiffer:
Updated the Console accordingly. @bafonins/@ryaplots please review code-owned files.

htdvisser:
I'll do a rebase and fix the conflicts. @bafonins @ryaplots please review.

bafonins (approved):

PR #5105, opened 2022-01-14
Use payload inference for LoRa Cloud

### LINKED ISSUE #4630: Pick up LoRa Cloud GNSS data from payload formatter field
#### Summary

We should ask device makers to break out Semtech GNSS data for the LR1110 in a payload formatter, and then pick it up in the LoRa Cloud integration by looking for that key

#### Why do we need this?

- The GNSS data does not come in uplinks in a standardized form. On the yabby edge, for example, it is contained in variable length wifi/gnss uplinks as the last 8 bytes. This needs to be defined in a payload formatter
- By looking for that key, we can stop doing unnecessary LoRa Cloud requests for devices with no GNSS data (currently, enabling the LoRa Cloud integration creates an error on every uplink)

When the LoRa Cloud integration is enabled for an application, all uplinks are forwarded to LoRa Cloud regardless of whether they contain valid gnss data

Only make LoRa cloud requests based on the presence of a gnss data key in the decoded output

@adriansmares

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4630
References https://github.com/TheThingsNetwork/lorawan-application-payload/pull/4

#### Changes

- Upgrade `lorawan-application-payload` to the newer version (needs to be merged, right now it refers to the commit in the PR)
- Remove manual WiFi access points logic from the AS parser
- Use `lorawan-application-payload` to infer the WiFi and GNSS payloads

#### Testing

Tested with a simulated Yabby Edge, using the payload `0200A4A0B3CCD358E600A4A0B3CCD358E6`
Screen Shot 2022-01-14 at 19 12 39

GNSS payload still to be tested (waiting on physical device).

##### Regressions

The previous behavior (i.e. previous parser) has been tested as part of the `lorawan-application-payload` library.

#### Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -109,7 +109,7 @@ require (
 	go.packetbroker.org/api/mapping/v2 v2.1.14-tts
 	go.packetbroker.org/api/routing v1.8.7-tts
 	go.packetbroker.org/api/v3 v3.10.7-tts
-	go.thethings.network/lorawan-application-payload v0.0.0-20211109090704-a9a0a6022856
+	go.thethings.network/lorawan-application-payload v0.0.0-20220125153912-1198ff1e403e
 	go.thethings.network/lorawan-stack-legacy/v2 v2.0.2
 	go.uber.org/zap v1.19.1
 	// NOTE: gocloud.dev is actually a different version (see above).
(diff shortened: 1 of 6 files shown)


johanstokking (approved):

htdvisser (approved):
Changes to my code-owned `go.mod` and `go.sum` look fine

PR #5111, opened 2022-01-18
Storage limitation for DevNonces

### LINKED ISSUE #5069: Store at most 10 used DevNonces
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/issues/2229

#### Why do we need this?

To reduce the amount of used DevNonces in our database. Also, many end devices do not have a true random number generator, leading to values being reused more often than should. Since the session key derivation also takes into account Join Nonce, this isn't too much of a problem.

The number of used DevNonces is infinite.

It should be limited to a configured maximum, by default 10.

On successful join, append the latest DevNonce and only store up to N used DevNonces back in the registry.

CI

Will review. @nicholaspcr let me know if you need any pointers or assistance. Relevant code is: https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.17/pkg/joinserver/joinserver.go#L333-L336

nicholaspcr:
#### Summary

Closes #5069 

The idea is to limit the amount of DevNonces stored in the the device registry, leaving the size to be something configurable.

#### Changes

- Add `devNonceLimit` to JS config
- Add config validation to JS
- Add limitation of DevNonces
- Add test cases confirming behaviour of the limitation process
- Update JS config for tests, added defaultDevNonceLimit to their JS config

#### Testing

Unit tests and manually testing, the images below show a bit of the behaviour for the devNonces where the limit per device is 5.
- [DevNonce with limit of 5](https://user-images.githubusercontent.com/28912302/150012589-6fed3762-ae91-4f22-a0c7-99b26445a0ad.png)
- [Same array after two new requests](https://user-images.githubusercontent.com/28912302/150012594-7bb5c876-bb6b-48f6-ac0b-d23d6a88f2f7.png)
- [curl to metrics](https://user-images.githubusercontent.com/28912302/150148215-02f5c631-7721-4f2c-bae0-62a24ac0d1e8.png)

##### Regressions

This changes the amount of DevNonces stored, since its not unlimited, the configurable amount of devNonces stored have a hard cap of the maximum value of an integer. Besides this there shouldn't be any changes in other features.

#### Notes for Reviewers

After Johan's observations, the requests with MacVersion  <= `1.0.3` will have their devNonces treated as random, therefore the slice that stores them will not be sorted, being just a FIFO queue.

#### Checklist

### DIFF
--- a/pkg/joinserver/config.go
+++ b/pkg/joinserver/config.go
@@ -24,4 +24,5 @@ type Config struct {
 	JoinEUIPrefixes               []types.EUI64Prefix                  `name:"join-eui-prefix" description:"JoinEUI prefixes handled by this Join Server"`
 	DefaultJoinEUI                types.EUI64                          `name:"default-join-eui" description:"Default JoinEUI for this Join Server"`
 	DeviceKEKLabel                string                               `name:"device-kek-label" description:"Label of KEK used to encrypt device keys at rest"`
+	DevNonceLimit                 int                                  `name:"dev-nonce-limit" description:"Amount of DevNonces stored per device"`
 }
--- a/pkg/joinserver/errors.go
+++ b/pkg/joinserver/errors.go
@@ -23,6 +23,7 @@ var (
 	errDeriveNwkSKeys                 = errors.Define("derive_nwk_s_keys", "failed to derive network session keys")
 	errDeviceNotFound                 = errors.DefineNotFound("device_not_found", "device not found")
 	errDevNonceTooSmall               = errors.DefineInvalidArgument("dev_nonce_too_small", "DevNonce is too small")
+	errDevNonceLimitInvalid           = errors.DefineInvalidArgument("dev_nonce_limit_invalid", "DevNonce limit can not be less than 1")
 	errDuplicateIdentifiers           = errors.DefineAlreadyExists("duplicate_identifiers", "a device identified by the identifiers already exists")
 	errEncodePayload                  = errors.DefineInvalidArgument("encode_payload", "failed to encode payload")
 	errEncryptPayload                 = errors.Define("encrypt_payload", "failed to encrypt JoinAccept")
--- a/pkg/joinserver/joinserver.go
+++ b/pkg/joinserver/joinserver.go
@@ -333,6 +333,7 @@ func (js *JoinServer) HandleJoin(ctx context.Context, req *ttnpb.JoinRequest, au
 			if req.SelectedMacVersion.IncrementDevNonce() {
 				if (dn != 0 || dev.LastDevNonce != 0 || dev.LastJoinNonce != 0) && !dev.ResetsJoinNonces {
 					if dn <= dev.LastDevNonce {
+						registerDevNonceTooSmall(ctx, req)
 						return nil, nil, errDevNonceTooSmall.New()
 					}
 				}
@@ -354,6 +355,7 @@ func (js *JoinServer) HandleJoin(ctx context.Context, req *ttnpb.JoinRequest, au
 					}
 					paths = append(paths, "used_dev_nonces")
 				} else if !dev.ResetsJoinNonces {
+					registerDevNonceReuse(ctx, req)
 					return nil, nil, errReuseDevNonce.New()
 				}
 			}
--- a/pkg/joinserver/joinserver.go
+++ b/pkg/joinserver/joinserver.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/binary"
-	"sort"
 	"time"
 
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
@@ -54,6 +53,7 @@ type JoinServer struct {
 
 	euiPrefixes    []types.EUI64Prefix
 	defaultJoinEUI types.EUI64
+	devNonceLimit  int
 
 	grpc struct {
 		nsJs                          nsJsServer
@@ -71,8 +71,18 @@ func (js *JoinServer) Context() context.Context {
 	return js.ctx
 }
 
+func validateConfig(conf *Config) error {
+	if conf.DevNonceLimit <= 0 {
+		return errDevNonceLimitInvalid.New()
+	}
+	return nil
+}
+
 // New returns new *JoinServer.
 func New(c *component.Component, conf *Config) (*JoinServer, error) {
+	if err := validateConfig(conf); err != nil {
+		return nil, err
+	}
 	js := &JoinServer{
 		Component: c,
 		ctx:       log.NewContextWithField(c.Context(), "namespace", "joinserver"),
@@ -83,6 +93,7 @@ func New(c *component.Component, conf *Config) (*JoinServer, error) {
 
 		euiPrefixes:    conf.JoinEUIPrefixes,
 		defaultJoinEUI: conf.DefaultJoinEUI,
+		devNonceLimit:  conf.DevNonceLimit,
 	}
 
 	js.grpc.applicationActivationSettings = applicationActivationSettingsRegistryServer{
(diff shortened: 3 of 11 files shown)


pkg/joinserver/config.go
```diff
@@ -24,4 +24,5 @@ type Config struct {
 	JoinEUIPrefixes               []types.EUI64Prefix                  `name:"join-eui-prefix" description:"JoinEUI prefixes handled by this Join Server"`
 	DefaultJoinEUI                types.EUI64                          `name:"default-join-eui" description:"Default JoinEUI for this Join Server"`
 	DeviceKEKLabel                string                               `name:"device-kek-label" description:"Label of KEK used to encrypt device keys at rest"`
+	DevNonceLimit                 int                                  `name:"dev-nonce-limit" description:"Sets the amount of the DevNonces to be stored per device"`
```
adriansmares: Consistency: `Amount of DevNonces stored per device` - the description is for the field, not for the configuration that you see with `--help`

pkg/joinserver/errors.go
```diff
@@ -23,6 +23,7 @@ var (
 	errDeriveNwkSKeys                 = errors.Define("derive_nwk_s_keys", "failed to derive network session keys")
 	errDeviceNotFound                 = errors.DefineNotFound("device_not_found", "device not found")
 	errDevNonceTooSmall               = errors.DefineInvalidArgument("dev_nonce_too_small", "DevNonce is too small")
+	errDevNonceLimitInvalid           = errors.DefineInvalidArgument("dev_nonce_limit_invalid", "DevNonce limit can't be less than 1")
```
adriansmares: Nit: `can not`

pkg/joinserver/joinserver.go
```diff
@@ -333,6 +345,13 @@ func (js *JoinServer) HandleJoin(ctx context.Context, req *ttnpb.JoinRequest, au
 					dev.UsedDevNonces = append(dev.UsedDevNonces, 0)
 					copy(dev.UsedDevNonces[i+1:], dev.UsedDevNonces[i:])
 					dev.UsedDevNonces[i] = dn
+
+					if len(dev.UsedDevNonces) > js.devNonceLimit {
+						dev.UsedDevNonces = dev.UsedDevNonces[1:]
```
adriansmares: The behavior here is a bit weird, or rather, unexpected. Evicting the first nonce like this will actually evict the lowest nonce, not necessarily the oldest one. I think that maybe we should drop the sorting.

On a separate note: Why is the `i == 0` comparison needed ? `limit == 0` is not allowed per validation

nicholaspcr: Yeah the weird behaviour is the biggest problem with keeping the array sorted, I could remove it but I'm wondering if the the impact of leaving like this outweighs the benefit of being a lot faster in situations with higher limits.

The `i == 0` refers to the result of the quicksort, which in this section of the code implies that it's different than the smallest value that was previously stored.

An idea came up to me that it might be possible to do an unintended behaviour by passing alternating small values, tested locally and confirmed that is possible.

adriansmares (approved):
Nice.

I think this variant is going to be faster than previous attempts. For such small N (10), searching iteratively is probably faster than sort + binary search.

nicholaspcr:
>I think this variant is going to be faster than previous attempts. For such small N (10), searching iteratively is probably faster than sort + binary search.

Agree but my worry is when N is close to maximum value of the integer, that could easily lead to a very slow search but I imagine it won't be happening and if it does the person will be aware of the implications in the time complexity.

pkg/joinserver/observability.go
```diff
@@ -67,16 +85,26 @@ func init() {
 type messageMetrics struct {
 	joinAccepted *metrics.ContextualCounterVec
 	joinRejected *metrics.ContextualCounterVec
+	devNonce     *devNonceMetrics
```
adriansmares, line 88: You can use the same struct, no need to compose here.

nicholaspcr: Could be, just opted to group the metrics within their own struct to make it a bit more organised

htdvisser (approved):
Code-owned (config) looks fine to me. Please don't forget to add it to configuration reference in the documentation. I think there are some other options missing there as well: https://github.com/TheThingsIndustries/lorawan-stack-docs/blob/master/doc/content/reference/configuration/join-server.md

pkg/joinserver/joinserver.go
johanstokking: Unresolving this comment, although it looks like the code changed since your earlier discussion.

I have two comments here:
1. Indeed there is no need to sort. But it is more likely that the last DevNonce are reused than that the first DevNonces are reused. So on L344, loop the used DevNonces backwards and keep things unsorted
2. Since you have `[1:]` here, you basically pop the first one only. But what happens if the limit was 100 and the operator reconfigured to a limit of 10? Since you only pop 1 every time, the number of DevNonces stored will stay 100. See suggestion below

pkg/joinserver/joinserver.go
```diff
@@ -322,19 +333,29 @@ func (js *JoinServer) HandleJoin(ctx context.Context, req *ttnpb.JoinRequest, au
...
+					if v == dn {
+						isReuse = true
+						break
+					}
+				}
+
+				if !isReuse {
+					dev.UsedDevNonces = append(dev.UsedDevNonces, dn)
+					if len(dev.UsedDevNonces) > js.devNonceLimit {
+						dev.UsedDevNonces = dev.UsedDevNonces[1:]
```
johanstokking:
```suggestion
					if n := len(dev.UsedDevNonces) - js.devNonceLimit; n > 0 {
						dev.UsedDevNonces = dev.UsedDevNonces[n:]
```

pkg/joinserver/observability.go
```diff
@@ -58,6 +58,24 @@ var jsMetrics = &messageMetrics{
 		},
 		[]string{"error"},
 	),
+	devNonce: &devNonceMetrics{
+		tooSmall: metrics.NewContextualCounterVec(
+			prometheus.CounterOpts{
+				Subsystem: subsystem,
+				Name:      "dev_nonce_too_small",
+				Help:      "Total number of devNonces too small errors",
```
johanstokking:
```suggestion
				Help:      "Total number of DevNonce too small errors",
```

pkg/joinserver/observability.go
```diff
@@ -58,6 +58,24 @@ var jsMetrics = &messageMetrics{
...
+			},
+			[]string{"mac_version"},
+		),
+		reuse: metrics.NewContextualCounterVec(
+			prometheus.CounterOpts{
+				Subsystem: subsystem,
+				Name:      "dev_nonce_reuse",
+				Help:      "Total number of devNonces reuse errors",
```
johanstokking:
```suggestion
				Help:      "Total number of DevNonce reuse errors",
```

pkg/joinserver/joinserver.go
```diff
@@ -322,19 +333,29 @@ func (js *JoinServer) HandleJoin(ctx context.Context, req *ttnpb.JoinRequest, au
...
-					dev.UsedDevNonces = append(dev.UsedDevNonces, 0)
-					copy(dev.UsedDevNonces[i+1:], dev.UsedDevNonces[i:])
-					dev.UsedDevNonces[i] = dn
+				isReuse := false
+				for _, v := range dev.UsedDevNonces {
+					if v == dn {
+						isReuse = true
+						break
+					}
+				}
```
johanstokking:
```suggestion
				for i := len(dev.UsedDevNonces) - 1; i >= 0; i-- {
					if dev.UsedDevNonces[i] == dn {
						isReuse = true
						break
					}
				}
```

nicholaspcr: True, it's more likely for the reused DevNonce to be in the start of the queue. Updated it.

johanstokking (approved):

PR #5115, opened 2022-01-19
Remove embed and nullability of end device requests

adriansmares:
#### Summary

References #2798 

#### Changes

- Remove `embed` and `nullable` from the remaining end device messages

#### Testing

Unit testing.

##### Regressions

As usual.

#### Notes for Reviewers

If CI passes, no breakage is expected.

#### Checklist


pkg/identityserver/end_device_registry.go
```diff
@@ -58,22 +58,22 @@ var errEndDeviceEUIsTaken = errors.DefineAlreadyExists(
 )
 
 func (is *IdentityServer) createEndDevice(ctx context.Context, req *ttnpb.CreateEndDeviceRequest) (dev *ttnpb.EndDevice, err error) {
-	if err = rights.RequireApplication(ctx, *req.Ids.ApplicationIds, ttnpb.RIGHT_APPLICATION_DEVICES_WRITE); err != nil {
+	if err = rights.RequireApplication(ctx, *req.EndDevice.Ids.ApplicationIds, ttnpb.RIGHT_APPLICATION_DEVICES_WRITE); err != nil {
 		return nil, err
 	}
-	if err = blacklist.Check(ctx, req.Ids.DeviceId); err != nil {
+	if err = blacklist.Check(ctx, req.EndDevice.Ids.DeviceId); err != nil {
```
KrishnaIyer, line 64: Don't we prefer a `req.EndDevice.GetIds().GetApplicationIds().GetApplicationId()` here or is that not necessary?

KrishnaIyer (approved):

pkg/identityserver/end_device_registry.go
htdvisser: We don't really have a strict rule about that, but perhaps it would be good to use the getters, just to be safe. I don't know if we need to fix that here, we can also run [gettercheck](https://github.com/saiskee/gettercheck) in the future and fix it then.

htdvisser (approved):

pkg/identityserver/end_device_registry.go
adriansmares: All of the fields are required in this chain, that's why I avoid the getter - `EndDevice` is a required field, `Ids` are required too, and so are `ApplicationIds`. I'm ok with adding getters for consistency in another PR.

Issue #5080, opened 2022-01-10
Rate Limiting documentation not clear about header values and behaviour


nicholaspcr:
#### Summary

The values of the `rate limiting` headers are not compliant with the [documentation](https://www.thethingsindustries.com/docs/reference/rate-limiting/) available. The behaviours observed were:
- the `x-rate-limit-reset` value increases through time instead of being a decreasing value.
- repeated values for the `x-rate-limit-available` header in different requests within the same minute.

#### Steps to Reproduce

1. Log of the result value in  [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/7a908e9f732cab737c36980fcde240e5cda7a942/pkg/ratelimit/grpc.go#L77). Something like `log.FromContext(ctx).Debug("Result: ", result)` would suffice.
2. Log the values from the rate-limit headers in any of the cli interceptors. Example below.
3. Start stack with `rate-limiting` enabled.
4. Create a gateway.
5. Run `go run ./cmd/ttn-lw-cli g get <gateway_id>` with intervals of 10 seconds between each request. Script example below.
  

##### stack config - enables rate-limiting
```yml
tls:
  source: file
  root-ca: ./ca.pem
  certificate: ./cert.pem
  key: ./key.pem

rate-limiting:
  profiles:
    - name: HTTP servers
      max-per-min: 30
      associations:
        - http
    - name: Application downlink traffic
      max-per-min: 10
      associations:
        - as:down:web
        - as:down:mqtt
        - grpc:method:/ttn.v3.lorawan.v3.AppAs/DownlinkQueuePush
        - grpc:method:/ttn.v3.lorawan.v3.AppAs/DownlinkQueueReplace
    - name: Gateway connections
      max-per-min: 2
      associations:
        - gs:accept:mqtt
        - gs:accept:ws
        - grpc:stream:accept:/ttn.lorawan.v3.GtwGs/LinkGateway
    - name: gRPC API
      max-per-min: 3
      associations:
        - grpc:method
        - grpc:stream:accept
    - name: Override rate for uplink simulation
      max-per-min: 5
      associations:
        - grpc:method:/ttn.lorawan.v3.AppAs/SimulateUplink
    - name: Gateway uplink traffic
      max-per-min: 1000
      max-burst: 1500
      associations:
       - gs:up
```

##### Template script to run go run command multiple times

```bash
#!/bin/bash
for run in {1..5}; do
[...]

PR #5121, opened 2022-01-21
Make CLI retry when it fails due to rate limitations

### LINKED ISSUE #4902: Make CLI retry requests when they failed due to rate limitations

nicholaspcr:
#### Summary

Closes #4902 

#### Changes

- Add unary interceptor for retry request
- Add unary interceptor test cases
- Add interceptor to CLI, disabled by default

#### Testing

Unit testing and local tests

Config and prints:
- [Print of concurrent unary requests with jitter](https://user-images.githubusercontent.com/28912302/150571680-d933fa3c-9b6f-4afb-ae1e-53d9dd815d8b.png)
- cli:
```yml
retry-config:
  max: 5
  jitter: 0.2
  enable_metadata: true
```
- stack
```yml
tls:
  source: file
  root-ca: ./ca.pem
  certificate: ./cert.pem
  key: ./key.pem

rate-limiting:
 profiles:
   - name: gRPC API
     max-per-min: 5
     max-burst: 7
     associations:
       - grpc:method
       - grpc:stream:accept
```
##### Regressions

Might increase the amount of requests made to services that usually return `resourceExhausted`, concurrent request could receive the same timeout but this can be mitigated with the jitter.

#### Notes for Reviewers

The implementation takes as a base the implementation that can be found [here](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/retry/retry.go), however there are a few small decisions made in this PR that changes some of the behaviours, all of these are welcomed to be discussed and changed. 

A few things that were changed:
- The validator that trigger the retry are determined by `func (error) bool`, this by default englobes `resourceExhausted` and `unavailable`. While it can be changed through grpcretry Option functions, there isn't any CLI settings to set which errors trigger the retry.
- In the `getHeaderLimiterTimeout` the reason for only using the `reset` header is due to its value being a incremental value. From the discussion that occured [here](https://github.com/TheThingsNetwork/lorawan-stack/issues/5080), the retry header will only contain a positive value when there are no available request to be made but to have a more dynamic timeout between requests the `reset` header is, in my opinion, ideal.
- After some discussion it was agreed that a stream interceptor is inappropriate for most of the requests that exist, therefore this PR only adds the unary interceptor.

Documentation PR - [here](https://github.com/TheThingsIndustries/lorawan-stack-docs/pull/741).
Will make the PR ready for review after everything here is approved, its just to have it as a reference until then.

#### Checklist

### DIFF
--- a/pkg/rpcmiddleware/rpcretry/client_interceptors.go
+++ b/pkg/rpcmiddleware/rpcretry/client_interceptors.go
@@ -1,0 +1,120 @@
+// Copyright © 2022 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rpcretry
+
+import (
+	"context"
+	"strconv"
+	"time"
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/util/backoffutils"
+	"go.thethings.network/lorawan-stack/v3/pkg/log"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// UnaryClientInterceptor returns a new unary client interceptor that retries the execution of external gRPC calls, the
+// retry attempt will only occur if any of the validators define the error as a trigger.
+func UnaryClientInterceptor(opts ...Option) grpc.UnaryClientInterceptor {
+	callOpts := evaluateClientOpt(opts...)
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		var md metadata.MD
+		var err error
+		logger := log.FromContext(ctx)
+
+		for attempt := uint(0); attempt <= callOpts.max; attempt++ {
+			retryTimeout := callOpts.timeout
+			if callOpts.enableMetadata {
+				if headerTimeout := getHeaderLimiterTimeout(ctx, md, callOpts); headerTimeout > 0 {
+					retryTimeout = headerTimeout
+				}
+			}
+
+			err = waitRetryBackoff(ctx, retryTimeout, attempt)
+			if err != nil {
+				logger.WithError(err).Error("An unexpected error occurred while in the timeout for the next request retry")
+				return err
+			}
+
+			err = invoker(ctx, method, req, reply, cc, append(opts, grpc.Header(&md))...)
+			if err == nil {
+				return nil
+			}
+
+			if !isRetriable(err, callOpts) {
+				return err
+			}
+		}
+
+		return err
+	}
+}
+
+func evaluateClientOpt(opts ...Option) *options {
+	optCopy := &options{}
+	*optCopy = *defaultOptions
+
+	for _, f := range opts {
+		f(optCopy)
+	}
+	return optCopy
+}
+
+func isRetriable(err error, opt *options) bool {
+	if err == nil {
+		return false
+	}
+	for _, check := range opt.validators {
+		if check(err) {
+			return true
+		}
+	}
+	return false
+}
+
+func waitRetryBackoff(ctx context.Context, timeout time.Duration, attempt uint) error {
+	if attempt == 0 {
+		return nil
+	}
+	log.FromContext(ctx).WithFields(log.Fields(
+		"attempt", attempt,
+		"timeout", timeout,
+	)).Debug("Failed request, waiting until next attempt")
+	timer := time.NewTicker(timeout)
+	select {
+	case <-ctx.Done():
+		timer.Stop()
+		return ctx.Err()
+	case <-timer.C:
+	}
+	return nil
+}
+
+func getHeaderLimiterTimeout(ctx context.Context, md metadata.MD, opt *options) time.Duration {
+	if len(md.Get("x-rate-limit-reset")) <= 0 {
+		return -1
+	}
+	var reset int
+	reset, err := strconv.Atoi(md.Get("x-rate-limit-reset")[0])
+	if err != nil {
+		return -1
+	}
+
+	timeout := time.Duration(reset) * time.Second
+	if opt.jitter > 0 {
+		timeout = backoffutils.JitterUp(timeout, opt.jitter)
+	}
+	return timeout
+}
--- a/pkg/rpcmiddleware/rpcretry/client_interceptors_test.go
+++ b/pkg/rpcmiddleware/rpcretry/client_interceptors_test.go
@@ -1,0 +1,174 @@
+// Copyright © 2022 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rpcretry_test
+
+import (
+	"context"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/smartystreets/assertions"
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/ratelimit"
+	"go.thethings.network/lorawan-stack/v3/pkg/rpcmiddleware/rpcretry"
+	"go.thethings.network/lorawan-stack/v3/pkg/util/rpctest"
+	"go.thethings.network/lorawan-stack/v3/pkg/util/test"
+	"go.thethings.network/lorawan-stack/v3/pkg/util/test/assertions/should"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+var (
+	exhaustedErr = errors.DefineResourceExhausted("rpcretry_unary_resource_exhausted", "mock error of a resource exhausted scenario")
+	internalErr  = errors.DefineInternal("rpcretry_unary_internal", "mock error that represents an error that should not be retried")
+)
+
+type unaryService struct {
+	rpctest.FooBarServer
+	md metadata.MD
+
+	counter uint
+	err     error
+	sleep   time.Duration
+}
+
+func (fs *unaryService) Unary(ctx context.Context, foo *rpctest.Foo) (*rpctest.Bar, error) {
+	fs.counter++
+	time.Sleep(fs.sleep)
+
+	if err := grpc.SendHeader(ctx, fs.md); err != nil {
+		return nil, err
+	}
+	return &rpctest.Bar{Message: "bar"}, fs.err
+}
+
+func Test_UnaryClientInterceptor(t *testing.T) {
+	type Service struct {
+		err   error
+		sleep time.Duration
+		md    metadata.MD
+	}
+	type Client struct {
+		retries       uint
+		retryTimeout  time.Duration
+		useMetadata   bool
+		jitter        float64
+		contextFiller func(ctx context.Context) context.Context
+	}
+
+	for _, tt := range []struct {
+		name              string
+		service           Service
+		client            Client
+		errAssertion      func(error) bool
+		expectedReqAmount int
+	}{
+		{
+			name:              "no error",
+			client:            Client{retries: 5, retryTimeout: 3 * test.Delay},
+			service:           Service{sleep: 0},
+			expectedReqAmount: 1,
+		},
+		{
+			name:              "unretriable error",
+			client:            Client{retries: 5, retryTimeout: 3 * test.Delay},
+			service:           Service{err: internalErr.New()},
+			errAssertion:      errors.IsInternal,
+			expectedReqAmount: 1,
+		},
+		{
+			name:              "retriable error",
+			client:            Client{retries: 5, retryTimeout: 3 * test.Delay},
+			service:           Service{err: exhaustedErr.New()},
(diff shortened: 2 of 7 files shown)


pkg/rpcmiddleware/rpcretry/client_interceptors.go
```diff
@@ -0,0 +1,182 @@
...
+type AttemptMetadataKey struct{}
+
+const (
+	headerXRateLimit     = "x-rate-limit-limit"
+	headerXRateAvailable = "x-rate-limit-available"
+	headerXRateReset     = "x-rate-limit-reset"
+	headerXRateRetry     = "x-rate-limit-retry"
+)
+
+var errInvalidStreamType = errors.DefineUnavailable("unsupported_stream_type_grpcretry", "Client streams are not supported in the grpc retry interceptor")
```
adriansmares: Very nit: `gRPC`.

pkg/rpcmiddleware/rpcretry/client_interceptors.go
```diff
@@ -0,0 +1,182 @@
...
+	"time"
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/util/backoffutils"
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/log"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+type AttemptMetadataKey struct{}
```
adriansmares: We generally name these as `*KeyType` and then have a single object called `*Key`. Also I don't think this should be accessible outside of the package.

If you want callers to have access to this information, create a specific function that allows the number of retries to be retrieved from a context - don't give them access to the key.

pkg/rpcmiddleware/rpcretry/client_interceptors.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	return false
+}
+
+func waitRetryBackoff(ctx context.Context, timeout time.Duration, attempt uint) error {
+	if attempt == 0 {
+		return nil
+	}
+
+	log.FromContext(ctx).WithField("attempt", attempt).Infof("Failed request, waiting %v until next attempt", timeout)
```
adriansmares: I would put the timeout in the fields.

pkg/rpcmiddleware/rpcretry/client_interceptors_test.go
```diff
@@ -0,0 +1,333 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/rpcmiddleware/rpcretry"
+	"go.thethings.network/lorawan-stack/v3/pkg/util/rpctest"
+	"go.thethings.network/lorawan-stack/v3/pkg/util/test"
+	"go.thethings.network/lorawan-stack/v3/pkg/util/test/assertions/should"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+var (
+	exaushtedErr       = errors.DefineResourceExhausted("rpcretry_unary_resource_exausted", "mock error of a resource exaushted scenario")
```
adriansmares: `exhausted`.

adriansmares (approved):

pkg/rpcmiddleware/rpcretry/client_interceptors.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	return nil
+}
+
+func hasRateLimitHeaders(md metadata.MD) bool {
+	return md != nil &&
+		len(md.Get(headerXRateLimit)) > 0 &&
+		len(md.Get(headerXRateAvailable)) > 0 &&
+		len(md.Get(headerXRateReset)) > 0 &&
+		len(md.Get(headerXRateRetry)) > 0
```
htdvisser: Why check for all of these if we only seem to care about `headerXRateReset`?

pkg/rpcmiddleware/rpcretry/client_interceptors.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func getHeaderLimiterTimeout(ctx context.Context, md metadata.MD, opt *options) time.Duration {
+	if !hasRateLimitHeaders(md) {
+		return -1
+	}
+	reset, _ := strconv.Atoi(md.Get(headerXRateReset)[0])
```
htdvisser: Are we sure that this will always be a number indicating a duration? I've also seen implementations that send a full date in this header.

pkg/rpcmiddleware/rpcretry/client_interceptors.go
```diff
@@ -0,0 +1,183 @@
...
+	timeout := time.Duration(reset) * time.Second
+	if opt.jitter > 0 {
+		rand.Seed(time.Now().UnixNano())
```
htdvisser: I don't think it's our job to be concerned about implementation details of `backoffutils`. I suppose it makes sense to seed the PRNG once when the program starts, but that's also not our concern here.

pkg/rpcmiddleware/rpcretry/client_interceptors.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	return false
+}
+
+func waitRetryBackoff(ctx context.Context, timeout time.Duration, attempt uint) error {
+	if attempt == 0 {
+		return nil
+	}
+
+	log.FromContext(ctx).WithField("attempt", attempt).WithField("timeout", timeout).Info("Failed request, waiting until next attempt")
```
htdvisser: 1. Prefer `WithFields` instead of multiple calls to `WithField`
2. Is `Info` appropriate here? Especially considering that the unexpected error above is only a `Debug`?

pkg/rpcmiddleware/rpcretry/client_interceptors.go
```diff
@@ -0,0 +1,183 @@
...
+// calls, the retry attempt will only occur if any of the validators define the error as a trigger.
+func StreamClientInterceptor(opts ...Option) grpc.StreamClientInterceptor {
+	callOpts := evaluateClientOpt(opts...)
+	return func(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+		if callOpts.max == 0 {
+			return streamer(ctx, desc, cc, method, opts...)
+		}
+		if desc.ClientStreams {
+			return nil, errInvalidStreamType.New()
+		}
```
htdvisser: It's way too easy to break a client this way. If the interceptor doesn't support certain stream types, just pass them onto the next interceptor.

```suggestion
		if desc.ClientStreams || callOpts.max == 0 {
			return streamer(ctx, desc, cc, method, opts...)
		}
```

pkg/rpcmiddleware/rpcretry/stream_client.go
```diff
@@ -0,0 +1,148 @@
...
+func (s *retriableStreamClient) incrementCounter() uint {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.counter++
+	return s.counter
+}
+
+func (s *retriableStreamClient) SendMsg(m interface{}) error {
+	s.mu.Lock()
+	s.bufferedSends = append(s.bufferedSends, m)
```
htdvisser: This looks like a memory leak waiting to happen.

pkg/rpcmiddleware/rpcretry/stream_client.go
```diff
@@ -0,0 +1,148 @@
...
+	bufferedSends := s.bufferedSends
+	s.mu.Unlock()
+
+	newStream, err := s.streamerCall(ctx)
+	if err != nil {
+		log.FromContext(ctx).WithError(err).Error("Failed to reconnect to server stream")
+		return nil, err
+	}
+
+	for _, msg := range bufferedSends {
```
htdvisser: How can we be sure that these buffered messages weren't already processed by the server? There are so many unexpected things that can happen if we re-send already processed messages, especially since the interceptor doesn't know anything about idempotency in the application.

htdvisser (changes requested):
While the unary retry interceptor looks okay to me, I don't think we should include the streaming retry interceptor. There are just too many things that can go wrong when you don't know anything about the application, even if the retry implementation itself were correct.

pkg/rpcmiddleware/rpcretry/client_interceptors.go
nicholaspcr: In `ratelimit`, the gRPC header it is set as an integer that is calculated from `time.Duration / time.Second`. There might be an implementation where a date is sent on the header but from the code I can only infer that it would be outside of the gRPC requests.

https://github.com/TheThingsNetwork/lorawan-stack/blob/4a442bf7d5f784ed6a632edd72d0a202027ba011/pkg/ratelimit/result.go#L38-L49

pkg/rpcmiddleware/rpcretry/stream_client.go
nicholaspcr: True. Will make it so that when attempting to reconnect it will only send the last send message, that way it shouldn't really be a problem.

nicholaspcr:
> While the unary retry interceptor looks okay to me, I don't think we should include the streaming retry interceptor. There are just too many things that can go wrong when you don't know anything about the application, even if the retry implementation itself were correct.

I agree that it can have unintended consequences. I'm thinking of leaving the implementation but not including the interceptor to the CLI, this way the stream interceptor can be adapted to work between two services and might be used in the future for specific cases.

pkg/rpcmiddleware/rpcretry/client_interceptors.go
nicholaspcr: It's not really necessary, it was in my mind a form of validation that confirms that is indeed from the `ratelimit.GRPCHeaders`. 
Since it doesn't make much sense anymore, I removed it

pkg/rpcmiddleware/rpcretry/client_interceptors.go
nicholaspcr: Added the check if there is an error while parsing it uses the default timer, it shouldn't fail but I agree that its good to have that check

pkg/rpcmiddleware/rpcretry/client_interceptors.go
nicholaspcr: The `backoffutils` implementation is just a two liner but since we already import it I didn't see a reason to not use it. The `rand.Seed` was originally here since its okay to use it in a concurrent setting but I agree that it might be best to leave it where the program starts so I moved it to the `commands/root.go` file.

pkg/rpcmiddleware/rpcretry/client_interceptors.go
nicholaspcr: 1 - done
2 - I was initially thinking that it would be appropriate to inform the user (outside the debug scope) that a retry attempts being made, the reason for this was that it might be a good amount of time without any feedback or information. 

Reflecting on it the worst case would be that the user would try the cli command again with `Debug` logs and see that his requests is failing and that it is retrying, so I opted to change the log to `Debug`. The unexpected error log was a mistake and is also fixed

pkg/rpcmiddleware/rpcretry/stream_client.go
nicholaspcr: I didn't notice the obvious before pushing the change but the biggest problem of not buffering the all the previous messages would be that a possible send that causes the error can easily be lost. 
For example:
send msgA (causes error) -> send msgB -> received error from msgA (uses msgB content to retry).

Disregard and read below

nicholaspcr: @htdvisser I'll admit that I was thrown in for a loop, especially since I'm thinking in terms of a bidiStream. I might be mistaken but after reading the code and revising my knowledge on the gRPC communications, the memory leak would only happen in a BidiStream where the client can send more than one message. In a `1:n` communication that problem wouldn't happen, since this interceptor rejects connections that has `ClientStreams` as true, neither `n:1` or `n:n` exchanges are possible.

With that in mind, it makes sense to keep the new change where is no longer a slice but just an `interface{}`, it would be a problem in the future when trying to do an interceptor that works for bidirectional stream but I don't think that is going to be the case soon.

htdvisser: I still think we should only do retries for unary RPCs, and remove the streaming interceptor completely.

It's not difficult to come up with an example for this going wrong in a server-streaming RPC (`1:n`). Take for example our `events.Stream` RPC. The client sends the stream request, including `after=2021-01-26T00:00:00Z`. At some point the stream breaks, the client retries, and suddenly receives duplicate messages that it already received before the stream broke. This is unexpected behavior, and instead I think the client should implement its own retry logic; in this case, retry the request with `after` set to the timestamp of the last event it did receive.

I imagine there will be a lot of other examples like this, so I just don't think that it's possible to build a one-size-fits-all interceptor for streaming RPCs.

htdvisser (commented):
I don't think that it's possible to build a one-size-fits-all interceptor for streaming RPCs, and I don't think we should build one that works in some cases and breaks in other cases, because that will be really difficult to debug.

nicholaspcr:
> I don't think that it's possible to build a one-size-fits-all interceptor for streaming RPCs, and I don't think we should build one that works in some cases and breaks in other cases, because that will be really difficult to debug.

I agree, the example that you gave makes things a bit more clear to me. Will remove the stream interceptor then.

pkg/rpcmiddleware/rpcretry/client_interceptors.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/util/backoffutils"
+	"go.thethings.network/lorawan-stack/v3/pkg/log"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+type attemptMetadataKeyType struct{}
+
+var attemptMetadataKey attemptMetadataKeyType
```
htdvisser: Seems unused.

htdvisser (approved):
I don't have time to thoroughly test this, but the code looks good to me.

pkg/rpcmiddleware/rpcretry/client_interceptors.go
nicholaspcr: Was originally intended as a bit of extra information for outgoing context, that way in the future if need be services could take that bit of information if need be. Since I considered mostly useful for the stream interceptor is something that I should have removed, will do so

PR #3051, opened 2020-08-07
Various small NS fixes and improvements

rvolosatovs:
#### Summary

Various small NS fixes and improvements
Upcoming NS testing PR (and following registry optimization one) is huge, so opening this up for easier review later

#### Changes

- Export registry utilities for usage from other packages
- Fix logging statement

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [x] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #5125, opened 2022-01-21
Integrate Ent diff into OS

ryaplots:
#### Summary

This PR merges unnecessary diff between Enterprise and OS. Refers to [#3051](https://github.com/TheThingsIndustries/lorawan-stack/issues/3051)

#### Testing

e2e and manually.

#### Checklist


pkg/webui/console/components/gateway-data-form/index.js
```diff
@@ -148,7 +166,30 @@ class GatewayDataForm extends React.Component {
           placeholder={sharedMessages.gatewayEUI}
           component={Input}
           tooltipId={tooltipIds.GATEWAY_EUI}
+          onBlur={this.onDisplayClaimChange}
         />
+        {shouldDisplayClaimWarning && (
```
bafonins, line 170: I dont think we need this in OS since there is no claiming view. Basiclaly, `console/gateways/claim` is not reachable.

bafonins (approved):
See my comment. the rest looks good

pkg/webui/console/components/gateway-data-form/index.js
kschiffer: Indeed. Until we figured out what to do here, we should change the message to

> It appears like you are trying to register a \<b>The Things Indoor Gateway\</b>. Such gateways are already preregistered and need to be claimed instead. Claiming can only be performed on a deployment that supports gateway claiming.

pkg/webui/components/notification/index.js
```diff
@@ -71,15 +68,7 @@ const Notification = ({
...
             />
-            {action && (
-              <Button
-                secondary
-                icon={buttonIcon}
-                onClick={action}
-                message={actionMessage}
-                type="button"
-              />
-            )}
```
kschiffer, line 82: `action` is actually still used here `console/components/webhook-form/index.js`

Make sure you refactor that before merging. This should also mean that this crashes in Enterprise.

kschiffer (changes requested):

pkg/webui/components/notification/index.js
ryaplots: Will resolve this for Enterprise in another PR

kschiffer (approved):
LGTM, please only merge after incorporating https://github.com/TheThingsIndustries/lorawan-stack/pull/3067

adriansmares (approved):

Issue #4847, opened 2021-11-09
End device onboarding flow


johanstokking:
#### Summary

New end device onboarding flow

Replaces #3770 
Blocked by #4840 
Blocked by #4841 
Blocked by #4845

#### Why do we need this?

To make it even easier to onboard new end devices by scanning QR codes.

Also we need to remove the creation on an external Join Server and integrate device claiming in the onboarding process.

For most end users, device creation and claiming is conceptually the same. We should put this in one nice device onboarding experience.

#### What is already there? What do you see now?

1. Device creation via Device Repository and manual
2. Device claiming by importing a manifest
3. QR code scanning app (TTSE only)

#### What is missing? What do you want to see?

Onboarding flow with QR code scanning, claiming, manual creation and retrieving info from the Device Repository integrated.

#### How do you propose to implement this?

For **onboarding**:

We have the following properties in an end device onboarding state:

- JoinEUI (to check with DCS if claiming is supported – #4841)
- DevEUI
- Vendor and Profile ID
- LoRaWAN device profile (includes activation mode)
- Brand, Model, Hardware and Firmware Version and Region (lookup end device model through Device Repository)
- Claim Authentication Code

Allow user to choose between: scan QR code, choose from device repository and manual creation:

0. Ask the user whether they want to scan a QR code. If yes, go to step 1, if not, go to step 2.
1. Scan QR code: this contains at least the JoinEUI and DevEUI, potentially also the numeric Vendor and Profile ID and the Claim Authentication Code. Call the `QRCodeParser.Parse()` rpc (https://github.com/TheThingsNetwork/lorawan-stack/issues/4845)
   1. Put the JoinEUI and DevEUI in the onboarding state
   2. If the QR code contains a claim authentication code, put this in the onboarding state
   3. If the QR code contains a non-zero Vendor ID and Profile ID, lookup the LoRaWAN device profile (#4842) and put this in the onboarding state. Otherwise, the LoRaWAN device profile is empty, but the activation mode can be preset to OTAA (ABP devices won't have QR codes)
   4. Go to step 2
2. Choose between selecting from the Device Repository or manual creation
   1. Choose from Device Repository. This gives us the Brand and Model ID, Hardware and Firmware Version and Region
      1. Get the LoRaWAN device profile from the Device Repository and put this in the onboarding state
   2. Manual creation, this basically proceeds to step 3 with an empty LoRaWAN device profile
[...]

Issue #4842, opened 2021-11-09
Get device information by vendor ID and profile ID


johanstokking:
#### Summary

Get device information by (numeric) vendor ID and profile ID

#### Why do we need this?

To fill out fields when onboarding devices with a QR code that contains the vendor ID and profile ID per LoRa Alliance TR005.

#### What is already there? What do you see now?

We keep track of vendor IDs and model IDs in the Device Repository, but we don't have an API to retrieve an end device model by vendor ID and profile ID.

So this is by the numeric vendor ID and profile ID, not by the string brand ID and model ID that the Device Repository uses.

This will only return the brand ID and the LoRaWAN device profile; _not_ the model ID, versions and region. This is because device profiles may be shared among multiple devices.

#### What is missing? What do you want to see?

Add rpc `GetProfileByID()` or something to the `DeviceRepository` service.

#### How do you propose to implement this?

Add rpc, implement service. The Device Repository may need to be indexed by the numerical identifiers.

#### How do you propose to test this?

Unit testing

#### Can you do this yourself and submit a Pull Request?

No

PR #5134, opened 2022-01-24
Implement QR Code Parser Service

### LINKED ISSUE #4845: Parse QR code
#### Summary

Parse QR code via a new gRPC service

#### Why do we need this?

To not rely on QR code format implementations in the front-end

We have a `EndDeviceQRCodeGenerator` service that generates QR codes for end devices.

We need a `QRCodeParser` service that can parse QR codes for anything it knows: end devices and gateways. This would basically parse known QR codes and return the result.

Add `QRCodeParser` service with one `Parse` rpc. It returns a model that is a `oneof` with information about an end device or gateway. This should be extendible, not just identifiers. For instance, for end devices, it may contain the JoinEUI, DevEUI, numeric vendor ID, numeric profile ID, claim authentication code, checksum and serial number, per LoRa Alliance TR005 specification.

Unit testing

Yes

KrishnaIyer:
#### Summary

Implement QR Code Parser Service. Closes #4845 .

#### Changes

- Refactor `qrcode` and `qrcodegenerator` package.
- Add a new `EntityOnboardingData` message with `EndDeviceOnboardingData` as the first type.
- Add a `QRCodeParser` service an implement the server.
- Adapt DCS to dial QRG.
- Adjust tests.

#### Testing

Unit tested.

##### Regressions

I don't expect any.

#### Notes for Reviewers

1. The QRG will now be the single point of truth for QR code formats. So I've moved the `qrcode` package into `qrcodegenerator`, so that components don't call the functions separately but dial QRG instead.
2. I don't see a use case where someone wants to parse a QR Code via the CLI; so I haven't added a command for this. We also don't claim using QR codes in the CLI. However, if we need a command for completeness, I can add that.

#### Checklist


CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ For details about compatibility between different releases, see the **Commitment
 ### Added
 - Add configurable storage limit to device's DevNonce in the JoinServer. Can be configured using the option `js.dev-nonce-limit`.
 
+- Service to parse QRCodes and return the data within.
+  - See the `Parse` RPC of the `QRCodeParser` service.
```
```suggestion
- Service to parse QR codes and return the data within.
  - See the `Parse` RPC of the `QRCodeParser` service.
```

Not sure if we need to mention rpc names in the changelog though

api/qrcodegenerator.proto
```diff
@@ -90,3 +90,52 @@ service EndDeviceQRCodeGenerator {
...
+  // claim_authentication_code ids referred to as;
+  // - Owner Token in LoRa Alliance TR005 specification.
+  // - Device Validation Code in the LoRa Alliance TR005 draft 2 and draft 3 specs.
+  string claim_authentication_code = 3;
+  bytes vendor_id = 4;
+  // model_id is referred to as VendorProfileID in LoRa Alliance TR005 specification.
+  bytes model_id = 5;
+  string checksum = 6;
+  string serial_number = 7;
+  string proprietary = 8;
```
johanstokking: Let's put some sane validation on these fields

api/qrcodegenerator.proto
```diff
@@ -90,3 +90,52 @@ service EndDeviceQRCodeGenerator {
...
+
+// EndDeviceOnboardingData contains fields necessary for onboarding a LoRaWAN End Device.
+message EndDeviceOnboardingData {
+	bytes join_eui = 1 [(gogoproto.customtype) = "go.thethings.network/lorawan-stack/v3/pkg/types.EUI64"];
+  bytes dev_eui = 2 [(gogoproto.customtype) = "go.thethings.network/lorawan-stack/v3/pkg/types.EUI64"];
+  // claim_authentication_code ids referred to as;
+  // - Owner Token in LoRa Alliance TR005 specification.
+  // - Device Validation Code in the LoRa Alliance TR005 draft 2 and draft 3 specs.
+  string claim_authentication_code = 3;
+  bytes vendor_id = 4;
```
johanstokking: This is a numeric field

api/qrcodegenerator.proto
```diff
@@ -90,3 +90,52 @@ service EndDeviceQRCodeGenerator {
...
+  // model_id is referred to as VendorProfileID in LoRa Alliance TR005 specification.
+  bytes model_id = 5;
```
johanstokking: This one too

api/qrcodegenerator.proto
johanstokking: The checksum isn't a string. Also these fields are very specific to TR005 and are not strictly necessary for onboarding.

Did you consider returning an end device template instead of this onboarding data?

api/qrcodegenerator.proto
```diff
@@ -90,3 +90,52 @@ service EndDeviceQRCodeGenerator {
...
+
+message ParseQRCodeResponse {
+  EntityOnboardingData entity_onboarding_data = 1;
+}
+
+service QRCodeParser {
+  // Parse QR Codes of known formats and return the information contained within.
+  rpc Parse(ParseQRCodeRequest) returns (ParseQRCodeResponse){
+    option (google.api.http) = {
+      post: "/qr-code",
```
johanstokking:
```suggestion
      post: "/qr-code/parse",
```

api/qrcodegenerator.proto
```diff
@@ -90,3 +90,52 @@ service EndDeviceQRCodeGenerator {
...
+  }
+}
+
+message ParseQRCodeRequest {
+  // QR code format identifier. Enumerate available formats with 
+  // - End Devices: rpc `ListFormats` in the EndDeviceQRCodeGenerator service.
+  // If this field is not specified, the server will attempt to parse the data with each known format.
+  string format_id = 1 [(validate.rules).string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$|^$", max_len: 36}];
+  // Raw QR code contents.
+  // The miminum
```
johanstokking: This doesn't look right

api/qrcodegenerator.proto
```diff
@@ -90,3 +90,52 @@ service EndDeviceQRCodeGenerator {
...
+  oneof data {
+    option (validate.required) = true;
+    EndDeviceOnboardingData end_device_onboarding_data = 2;
+  }
+}
+
+message ParseQRCodeRequest {
+  // QR code format identifier. Enumerate available formats with 
+  // - End Devices: rpc `ListFormats` in the EndDeviceQRCodeGenerator service.
+  // If this field is not specified, the server will attempt to parse the data with each known format.
```
johanstokking: How are these things related exactly? This is a generic service, right? What is the namespace of these formats?

Why allow one format? Why not multiple?

Why not limit results to entities, like end devices or gateways (or both)?

Do we need this at all?

johanstokking (commented):
Reviewed API only for now

api/qrcodegenerator.proto
KrishnaIyer: Ah I see ok. Well no because I didn't know until now that `EndDeviceTemplate` exists. I'll update that here and re-request.

pkg/deviceclaimingserver/grpc_end_devices.go
```diff
@@ -67,21 +66,28 @@ var (
...
+			client := ttnpb.NewQRCodeParserClient(conn)
+			data, err := client.Parse(ctx, &ttnpb.ParseQRCodeRequest{
+				QrCode: qrCode,
+			})
 			if err != nil {
 				return nil, errParseQRCode.WithCause(err)
 			}
-			authIDs, ok := data.(qrcode.AuthenticatedEndDeviceIdentifiers)
-			if !ok {
+			if onboardingData := data.EntityOnboardingData.GetEndDeviceOnboardingData(); onboardingData != nil && onboardingData.JoinEui != nil {
```
nicholaspcr: Nit:
```suggestion
			if onboardingData := data.GetEntityOnboardingData().GetEndDeviceOnboardingData(); onboardingData != nil && onboardingData.JoinEui != nil {
```

nicholaspcr (approved):
The DCS section LGTM.

htdvisser (commented):
Only looked at `api`. Not much to add on top of @johanstokking's comments at this point.

KrishnaIyer:
@johanstokking  Regarding the [comment on EndDeviceTemplate](https://github.com/TheThingsNetwork/lorawan-stack/pull/5134#discussion_r790849589).

Step 1 of the new onboarding flow says; https://github.com/TheThingsNetwork/lorawan-stack/issues/4847#issue-1048748654 (line numbers are mine)
```
a. Scan QR code: this contains at least the JoinEUI and DevEUI, potentially also the numeric Vendor and Profile ID and the Claim Authentication Code. Call the QRCodeParser.Parse() rpc (https://github.com/TheThingsNetwork/lorawan-stack/issues/4845)
b. Put the JoinEUI and DevEUI in the onboarding state
c. If the QR code contains a claim authentication code, put this in the onboarding state
d. If the QR code contains a non-zero Vendor ID and Profile ID, lookup the LoRaWAN device profile (https://github.com/TheThingsNetwork/lorawan-stack/issues/4842) and put this in the onboarding state. Otherwise, the LoRaWAN device profile is empty, but the activation mode can be preset to OTAA (ABP devices won't have QR codes)
```

So the QR Code is scanned and we only need the EUIs, CAC and the Vendor/Profile IDs. We then use this data in separate RPCs.

Theoretically, if we have the AppIDs and the above Vendor/Profile IDs, we can dial the Device Repository and get the End Device Template but this will be done further in the end device creation flow when the user chooses the region/firmware etc (Step 2 of onboarding flow).

a. So do we really need to use End Device Templates here? Because, if the QR Code does not have Vendor ID and Model ID, in that case wouldn't return a mostly empty End Device (Template) with just the EUIs and CAC? 
b. I can update the fields of the `EndDeviceOnboardingData` to remove the TR005-only fields. But what do we do with  those fields? Do we just discard them and only return what's needed for onboarding? Ex: Do we verify this checksum anywhere?

The way I understood the issue, we only need to parse the QR Code and return the data in a generic format and let the 
caller call other RPCs with the returned data (For example use Vendor ID/ Model ID and the DR for the device profile).

johanstokking:
> a. So do we really need to use End Device Templates here? Because, if the QR Code does not have Vendor ID and Model ID, in that case wouldn't return a mostly empty End Device (Template) with just the EUIs and CAC?

There may be other QR codes in the future with additional fields. Also, as we don't have a formal serial number field currently, we probably want to put it in an attribute. End device templates are really designed for this.

>b. I can update the fields of the `EndDeviceOnboardingData` to remove the TR005-only fields. But what do we do with those fields? Do we just discard them and only return what's needed for onboarding? Ex: Do we verify this checksum anywhere?

See TR005 1.0 spec:

>210 The Checksum is used to validate the data integrity. Even though QR codes have their own
>211 built-in integrity checks, this explicit checksum is useful when the content of the QR code is
>212 presented as plain text. Checksum is generated using the CRC-16-MODBUS [CRC] of the
>213 full QR content except the Checksum field itself, and presented in hexadecimal format
>214 without the leading “0x”.

So the TR005 parser should verify it, and after that, it's not of any use anymore.

> The way I understood the issue, we only need to parse the QR Code and return the data in a generic format and let the caller call other RPCs with the returned data (For example use Vendor ID/ Model ID and the DR for the device profile).

Yes. So the response would be a `oneof` with `EndDeviceTemplate` as one of the options.

api/qrcodegenerator.proto
KrishnaIyer: So the idea is that if the user already knows the format, then we can parse it directly with that format and we don't have to trial and error with all the formats. 

> Why allow one format? Why not multiple?

When a QR code is generated, it's generated for one format right?. If that format is known, then we use that to parse it. Why would we need to allow multiple formats?

> This is a generic service, right?

Indeed. That's why I added the line 
```
- End Devices: rpc `ListFormats` in the EndDeviceQRCodeGenerator service.
```
If we add gateway QR Codes, then I suppose we'll also add a new `GatewayQRCodeGenerator` which would be option 2 in this description. Ex:
```
- End Devices: rpc `ListFormats` in the EndDeviceQRCodeGenerator service.
- Gateways: rpc `ListFormats` in the GatewayQRCodeGenerator service.
```

> Why not limit results to entities, like end devices or gateways (or both)? 
> What is the namespace of these formats?

I think these are related questions.

I suppose we could add an `entity_type` field to the request and users can choose `enddevices`, `gateways` or something else we support later. If the user then provides the `format_id` (ex: We know that it's TR005), then we use that format or we loop through each possible case within the formats supported for the entity.  

I don't know if we need to be more generic than that. Do you see a case where the client (ex: Console) reads an unspecified QR code and we need to determine the entity also?

pkg/qrcodegenerator/qrcode/enddevice/lora_alliance_tr005.go
```diff
@@ -143,14 +151,52 @@ func (m *LoRaAllianceTR005) UnmarshalText(text []byte) error {
...
-// AuthenticatedEndDeviceIdentifiers implements the AuthenticatedEndDeviceIdentifiers interface.
-func (m *LoRaAllianceTR005) AuthenticatedEndDeviceIdentifiers() (joinEUI, devEUI types.EUI64, authenticationCode string) {
-	return m.JoinEUI, m.DevEUI, m.OwnerToken
+// GetEntityOnboardingData implements the Data interface.
+func (m *LoRaAllianceTR005) GetEntityOnboardingData() *ttnpb.EntityOnboardingData {
+	paths := []string{
+		"ids",
+		"claim_authentication_code",
+	}
+	attributes := make(map[string]string)
```
KrishnaIyer: > If the QR code contains a non-zero Vendor ID and Profile ID, lookup the LoRaWAN device profile (https://github.com/TheThingsNetwork/lorawan-stack/issues/4842) and put this in the onboarding state

I'm using attributes here since these fields are not defined in the `EndDevice` message. The next step of the onboarding flow for the client need these int values  so I'm using hex encoded string. The caller can then do a `strings.Atoi` and use that to fetch the device profile by numeric ID.

pkg/qrcodegenerator/qrcode/enddevice/lora_alliance_tr005.go
```diff
@@ -143,14 +151,52 @@ func (m *LoRaAllianceTR005) UnmarshalText(text []byte) error {
...
+	}
+	if m.ModelID != [2]byte{} {
+		attributes[profileIDAttribute] = strings.ToUpper(hex.EncodeToString(m.ModelID[:]))
+	}
+	if len(attributes) > 0 {
+		paths = append(paths, "attributes")
+	}
+	return &ttnpb.EntityOnboardingData{
+		FormatId: formatIDLoRaAllianceTR005,
+		Data: &ttnpb.EntityOnboardingData_EndDeviceTempate{
```
KrishnaIyer: As far as I've seen, the `MappingKey` field of the `EndDeviceTempate` is for format conversions so I've left that out here.

KrishnaIyer:
BTW for this one, 
> Checksum is generated using the CRC-16-MODBUS [CRC] of the
213 full QR content except the Checksum field itself, and presented in hexadecimal format
214 without the leading “0x”.

The only reference I found was this; https://www.modbustools.com/modbus_crc16.html. There are a couple of Go packages but they are super old.

Should I implement this checksum check in this PR or follow that up later?

johanstokking:
> Should I implement this checksum check in this PR or follow that up later?

I took this one too and tested against the spec but I got a different response. I think there's gonna be some clarification. Don't bother for now.

api/qrcodegenerator.proto
johanstokking: > If we add gateway QR Codes, then I suppose we'll also add a new `GatewayQRCodeGenerator` which would be option 2 in this description. Ex:
> 
> ```
> - End Devices: rpc `ListFormats` in the EndDeviceQRCodeGenerator service.
> - Gateways: rpc `ListFormats` in the GatewayQRCodeGenerator service.
> ```

I see, but that's what I meant with the namespace. TR005 is now about end devices, but TR042 could be about end devices and gateways. The format IDs are unique within `EndDeviceQRCodeGenerator` but not in `QRCodeParser`.

> I don't know if we need to be more generic than that. Do you see a case where the client (ex: Console) reads an unspecified QR code and we need to determine the entity also?

Not really, actually. Maybe we should keep this really simple and make a service specifically for end devices and add gateways later.

pkg/qrcodegenerator/qrcode/qrcode.go
```diff
@@ -0,0 +1,31 @@
...
+	"encoding"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+// Data represents QR code data.
+type Data interface {
+	Validate() error
+	// EndDeviceInfo returns the format ID as a string and the End Device Template corresponding to the QR code data.
+	EndDeviceInfo() (string, *ttnpb.EndDeviceTemplate)
```
KrishnaIyer: Future QR codes for other entities will have a specific function here.

api/qrcodegenerator.proto
KrishnaIyer: Ok so this turns out pretty neat with a new RPC to the existing `EndDeviceQRCodeGenerator`.

pkg/deviceclaimingserver/grpc_end_devices.go
```diff
@@ -67,21 +66,28 @@ var (
...
 		} else if qrCode := req.GetQrCode(); qrCode != nil {
-			data, err := qrcode.Parse(qrCode)
+			conn, err := srv.DCS.GetPeerConn(ctx, ttnpb.ClusterRole_QR_CODE_GENERATOR, nil)
+			if err != nil {
+				return nil, err
+			}
+			client := ttnpb.NewEndDeviceQRCodeGeneratorClient(conn)
+			data, err := client.Parse(ctx, &ttnpb.ParseEndDeviceQRCodeRequest{
+				QrCode: qrCode,
+			})
```
johanstokking: Is there no auth here at all, like not even checking whether there is any authentication?

If the API call is open, does it matter if we use cluster auth here to bypass the rate limiter?

pkg/qrcodegenerator/qrcode/enddevices/enddevices.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil
+	}
+	return res.(Format)
+}
+
+// RegisterEndDeviceFormat registers the given end device QR code format.
+// Existing registrations with the same ID will be overwritten.
+func (s *Server) RegisterEndDeviceFormat(id string, f Format) {
+	s.endDeviceFormats.Store(id, f)
+}
```
johanstokking: Is this ever going to be dynamic?

pkg/qrcodegenerator/qrcode/qrcode.go
johanstokking: Can we please use an interface and test the interface instead?

pkg/qrcodegenerator/qrcode/enddevices/enddevices.go
KrishnaIyer: This is used in the QRG tests; https://github.com/TheThingsNetwork/lorawan-stack/blob/954d67784785a52373699951e7ca627b88d4409a/pkg/qrcodegenerator/grpc_end_devices_test.go#L45
Though I've now updated this to register the known formats in the constructor.

pkg/deviceclaimingserver/grpc_end_devices.go
KrishnaIyer: The value that's passed as input (QR code contents) need to more secure than the data that comes out of this RPC so I don't think auth is required here. I've removed the cluster auth since that's indeed not necessary.

KrishnaIyer:
@johanstokking: What do you think about using https://github.com/TheThingsNetwork/lorawan-stack/pull/5323 here instead of exporting the numeric profile IDs? We can query the DR as part of the Parse request and return a EDT with the ED profile (if available) instead of clients making a separate call.

pkg/deviceclaimingserver/grpc_end_devices.go
johanstokking: I meant that if we don't use any auth, we can't do rate limiting other than on originating IP.

johanstokking:
> We can query the DR as part of the Parse request and return a EDT with the ED profile (if available) instead of clients making a separate call.

Yes that's possible but loading the device from the device repository is a different concern than parsing the QR code. Let's keep those concerns separate and make sure the primary keys are usable (including vendor and profile ID).

We might need additional fields in `EndDeviceVersionIdentifiers`?

api/qrcodegenerator.proto
```diff
@@ -67,6 +67,21 @@ message GenerateQRCodeResponse {
   Picture image = 2;
 }
 
+message ParseEndDeviceQRCodeRequest {
+  // QR code format identifier.
+  // Enumerate available formats with the rpc `ListFormats`.
+  // If this field is not specified, the server will attempt to parse the data with each known format.
+  string format_id = 1 [(validate.rules).string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$|^$", max_len: 36}];
+  // Raw QR code contents.
+  bytes qr_code = 3 [(validate.rules).bytes = {min_len: 10, max_len: 1024}];
```
johanstokking: Where's `2`?

pkg/deviceclaimingserver/enddevices/enddevices.go
```diff
@@ -179,15 +178,19 @@ func (upstream *Upstream) Claim(ctx context.Context, req *ttnpb.ClaimEndDeviceRe
...
-			return nil, errParseQRCode.WithCause(err)
+			return nil, err
 		}
-		authIDs, ok := data.(qrcode.AuthenticatedEndDeviceIdentifiers)
-		if !ok {
-			return nil, errQRCodeData.New()
+		qrg := ttnpb.NewEndDeviceQRCodeGeneratorClient(conn)
+		data, err := qrg.Parse(ctx, &ttnpb.ParseEndDeviceQRCodeRequest{
+			QrCode: qrCode,
+		})
```
johanstokking: So I still would like to see some form of authentication here, even if it was forwarded auth or cluster auth or whatever, so we can do proper rate limiting on the authenticated party.

pkg/deviceclaimingserver/enddevices/ttjs/ttjs.go
```diff
@@ -137,15 +164,24 @@ func (client *TTJS) Claim(ctx context.Context, req *ttnpb.ClaimEndDeviceRequest)
...
-			return nil, errQRCodeData.New()
+
+		dev := data.GetEndDeviceTempate().GetEndDevice()
+		if dev == nil || dev.ClaimAuthenticationCode == nil {
+			return nil, errParseQRCode.New()
 		}
-		joinEUI, devEUI, ownerToken = authIDs.AuthenticatedEndDeviceIdentifiers()
+		joinEUI = *dev.GetIds().JoinEui
+		devEUI = *dev.GetIds().DevEui
+		ownerToken = dev.ClaimAuthenticationCode.Value
```
johanstokking: I don't think that the The Things Join Server implementation should be bothered with parsing QR codes. This should be nothing more than a client to a Join Server.

If anything, DCS should do this before it let's JS clients claim.

In previous review rounds I already stressed on decoupling stuff. It would have been helpful if claimers are not taking any `pkg/ttnpb` types so that we are sure things aren't getting mixed up. You don't really need `ClaimEndDeviceRequest` here; JS claimers just need EUIs and a secret. It doesn't matter how does values were determined at this level.

pkg/qrcodegenerator/qrcode/enddevices/enddevices.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/qrcodegenerator/qrcode"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+const (
+	serialNumberAttribute = "serial-number"
+	vendorIDAttribute     = "vendor-id"
+	profileIDAttribute    = "profile-id"
```
johanstokking: I think this need to become `EndDevice` and/or `EndDeviceVersionIdentifiers` fields

pkg/qrcodegenerator/qrcode/enddevices/enddevices.go
```diff
@@ -0,0 +1,123 @@
...
+	Encode(*ttnpb.EndDevice) error
+	// FormatID returns the ID of the format used to parse the QR Code data.
+	FormatID() string
+	// EndDeviceTemplate returns the End Device Template corresponding to the QR code data.
+	EndDeviceTemplate() *ttnpb.EndDeviceTemplate
+}
+
+// Server provides methods for end device QR codes.
+type Server struct {
+	endDeviceFormats sync.Map
```
johanstokking: Still, why is this dynamic? When does this change at runtime?
Why isn't this fixed at init time? You can also add one at test init, right?

Apart from options, avoid codebase pollution for tests.

johanstokking (changes requested):

pkg/qrcodegenerator/qrcode/enddevices/enddevices.go
KrishnaIyer: > Why isn't this fixed at init time?

It is. https://github.com/TheThingsNetwork/lorawan-stack/blob/4f678e0e27566c47886df58a0ab234ca9845ccdf/pkg/qrcodegenerator/qrcode/enddevices/enddevices.go#L55-L63

This is similar to what we already have. In the current codebase;

- Formats have an option to be updated dynamically; https://github.com/TheThingsNetwork/lorawan-stack/blob/d168a51ebba39db4586790effda102af2fdda536/pkg/qrcode/qrcode.go#L72-L102
- A function is exposed to register formats from outside the package and this is used in the `QRGenerator` tests; https://github.com/TheThingsNetwork/lorawan-stack/blob/d168a51ebba39db4586790effda102af2fdda536/pkg/qrcodegenerator/grpc_test.go#L40

The only change I've made is to use a struct and a `sync.Map`.

johanstokking: If this isn't dynamic at runtime, why use a `sync.Map`?

What prevents you from registering the test format at test init time?

pkg/deviceclaimingserver/enddevices/enddevices.go
```diff
@@ -42,7 +41,7 @@ type EndDeviceClaimer interface {
 	// SupportsJoinEUI returns whether the Join Server supports this JoinEUI.
 	SupportsJoinEUI(joinEUI types.EUI64) bool
 	// Claim claims an End Device.
-	Claim(ctx context.Context, req *ttnpb.ClaimEndDeviceRequest) (*ttnpb.EndDeviceIdentifiers, error)
+	Claim(ctx context.Context, joinEUI *types.EUI64, devEUI *types.EUI64, cac string, hNSAddress string) error
```
KrishnaIyer: This could be done with an interface but then we need a different package to define the interface to avoid import cycles. This is a simple implementation that works so I've used this.

pkg/deviceclaimingserver/enddevices/enddevices.go
KrishnaIyer: Ok then I think that the same auth check must be applied for the other RPCs in the service for the same reasons. I've done that as well.

api/qrcodegenerator.proto
```diff
@@ -67,6 +67,21 @@ message GenerateQRCodeResponse {
...
+  // If this field is not specified, the server will attempt to parse the data with each known format.
+  string format_id = 1 [(validate.rules).string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$|^$", max_len: 36}];
+  // Raw QR code contents.
+  bytes qr_code = 2 [(validate.rules).bytes = {min_len: 10, max_len: 1024}];
+}
+
+message ParseEndDeviceQRCodeResponse {
+  // Identifier of the format used to successfully parse the QR code data.
+  string format_id = 1;
+  EndDeviceTemplate end_device_tempate = 2;
```
adriansmares: `end_device_template`.

adriansmares (commented):
Please revert the diff to `data/lorawan-frequency-plans`.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Added
 
+- Service to parse QRCodes and return the data within.
```
```suggestion
- Service to parse QR codes and return the data within.
```

api/identifiers.proto
```diff
@@ -97,6 +97,11 @@ message EndDeviceVersionIdentifiers {
   string hardware_version = 3 [(validate.rules).string.max_len = 32];
   string firmware_version = 4 [(validate.rules).string.max_len = 32];
   string band_id = 5 [(validate.rules).string.max_len = 32];
+  // VendorID managed by the LoRa Alliance, as defined in TR005.
+  uint32 vendor_id = 6;
+  // ID of the LoRaWAN end device profile assigned by the vendor.
+  uint32 vendor_profile_id = 7;
+  string serial_number = 8 [(validate.rules).string = {pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$", max_len: 36, ignore_empty: true}];
```
johanstokking, line 104: We might need to relax this pattern in the future

pkg/deviceclaimingserver/enddevices/enddevices.go
johanstokking:
```suggestion
	Claim(ctx context.Context, joinEUI, devEUI types.EUI64, claimAuthenticationCode string, networkServerAddress string) error
```

pkg/deviceclaimingserver/grpc_end_devices.go
```diff
@@ -86,15 +93,67 @@ type endDeviceClaimingServer struct {
...
+	if err := rights.RequireApplication(ctx, *targetAppID,
+		ttnpb.Right_RIGHT_APPLICATION_DEVICES_WRITE,
+	); err != nil {
+		return nil, err
+	}
+
+	var (
+		joinEUI *types.EUI64
+		devEUI  *types.EUI64
+		cac     string
```
johanstokking: Idem

pkg/deviceclaimingserver/grpc_end_devices.go
```diff
@@ -86,15 +93,67 @@ type endDeviceClaimingServer struct {
...
+			return nil, errParseQRCode.New()
+		}
+		joinEUI = dev.GetIds().JoinEui
+		devEUI = dev.GetIds().DevEui
+		cac = dev.ClaimAuthenticationCode.Value
+	} else {
+		return nil, errNoJoinEUI.New()
+	}
+
+	hNSAddress, _, err := net.SplitHostPort(req.TargetNetworkServerAddress)
```
johanstokking: Idem

pkg/qrcodegenerator/grpc_end_devices.go
```diff
@@ -19,34 +19,65 @@ import (
...
 }
 
+// GetFormat implements EndDeviceQRCodeGenerator.
 func (s *endDeviceQRCodeGeneratorServer) GetFormat(ctx context.Context, req *ttnpb.GetQRCodeFormatRequest) (*ttnpb.QRCodeFormat, error) {
-	format := qrcode.GetEndDeviceFormat(req.FormatId)
+	_, err := rpcmetadata.WithForwardedAuth(ctx, s.QRG.AllowInsecureForCredentials())
+	if errors.IsUnauthenticated(err) {
+		if clusterauth.Authorized(ctx) != nil {
+			return nil, errUnauthenticated.New()
+		}
```
johanstokking, line 40: Don't we have a simple check to see if the caller is authenticated?

pkg/qrcodegenerator/qrcode/enddevices/enddevices.go
```diff
@@ -0,0 +1,106 @@
...
+// New returns a new Server.
+func New(ctx context.Context) *Server {
+	s := &Server{
+		endDeviceFormats: make(map[string]Format),
+	}
+
+	// Register known formats.
+	s.endDeviceFormats[formatIDLoRaAllianceTR005] = new(LoRaAllianceTR005Format)
+	s.endDeviceFormats[formatIDLoRaAllianceTR005Draft2] = new(LoRaAllianceTR005Draft2Format)
+	s.endDeviceFormats[formatIDLoRaAllianceTR005Draft3] = new(LoRaAllianceTR005Draft3Format)
```
johanstokking: I still don't get why this isn't a package global initialized at init time

johanstokking: You can even expose the package global with a nice godoc saying it's read-only, but write to it in tests init

KrishnaIyer: I see what you mean but I think the implementation and your suggestion are isomorphic.  
Unless there is a (functional/maintainability) issue with the implementation (which I don't really see), can we say that this is a reviewer nitpick and I prefer to keep my implementation?

pkg/deviceclaimingserver/grpc_end_devices.go
```diff
@@ -16,16 +16,23 @@ package deviceclaimingserver
...
 	"go.thethings.network/lorawan-stack/v3/pkg/types"
 	"go.thethings.network/lorawan-stack/v3/pkg/web"
 )
 
-var errMethodUnavailable = errors.DefineUnimplemented("method_unavailable", "method available")
+var (
+	errParseQRCode       = errors.Define("parse_qr_code", "parse QR code failed")
+	errQRCodeData        = errors.DefineInvalidArgument("qr_code_data", "invalid QR code data")
+	errNoJoinEUI         = errors.DefineInvalidArgument("no_join_eui", "failed to extract JoinEUI from request")
+	errMethodUnavailable = errors.DefineUnimplemented("method_unavailable", "method available")
```
nicholaspcr: shouldn't it be `method unavailable` in the error message?

If so, the same applies to the file `pkg/deviceclaimingserver/enddevices/util_test.go`. It's only used in the mockEDCS but probably worth a change.

pkg/deviceclaimingserver/enddevices/enddevices.go
```diff
@@ -42,7 +41,7 @@ type EndDeviceClaimer interface {
 	// SupportsJoinEUI returns whether the Join Server supports this JoinEUI.
 	SupportsJoinEUI(joinEUI types.EUI64) bool
 	// Claim claims an End Device.
-	Claim(ctx context.Context, req *ttnpb.ClaimEndDeviceRequest) (*ttnpb.EndDeviceIdentifiers, error)
+	Claim(ctx context.Context, joinEUI *types.EUI64, devEUI *types.EUI64, claimAuthenticationCode string, networkServerAddress string) error
```
nicholaspcr: nit: `... joinEUI, devEUI *types.EUI64, claimAuthenticationCode, networkServerAddress string) error`

johanstokking: This isn't resolved, and we don't have to (and probably shouldn't) take pointers for EUIs here

pkg/deviceclaimingserver/enddevices/ttjs/ttjs.go
```diff
@@ -119,51 +118,19 @@ var (
...
-
-	hNSAddress, _, err := net.SplitHostPort(req.TargetNetworkServerAddress)
-	if err != nil {
-		// TargetNetworkServerAddress is already validated by the API.
-		// An error here means that it does not contain a port, so we use it directly.
-		hNSAddress = req.TargetNetworkServerAddress
-	}
-
-	hNSID, ok := client.hsNSIDs[hNSAddress]
+	hNSID, ok := client.hsNSIDs[networkServerAddress]
```
johanstokking, line 127: Don't we have to `SplitHostPort` anymore?

pkg/deviceclaimingserver/grpc_end_devices.go
```diff
@@ -86,15 +93,67 @@ type endDeviceClaimingServer struct {
...
+	}
+
+	networkServerAddress, _, err := net.SplitHostPort(req.TargetNetworkServerAddress)
+	if err != nil {
+		// TargetNetworkServerAddress is already validated by the API.
+		// An error here means that it does not contain a port, so we use it directly.
+		networkServerAddress = req.TargetNetworkServerAddress
+	}
+
+	err = edcs.DCS.endDeviceClaimingUpstream.Claim(ctx, joinEUI, devEUI, claimAuthenticationCode, networkServerAddress)
```
johanstokking, line 141: Please put a comment here for our future selves why we're taking the host here and not the address.

Like:

> External services, including Join Servers, typically identify Network Servers by host instead of by host and port.

pkg/qrcodegenerator/grpc_end_devices.go
johanstokking: Can you come up with something very simple in the rights package? We really only have to check the incoming context for anything non-empty in `AuthType` and `AuthValue`.

pkg/qrcodegenerator/qrcode/enddevices/enddevices.go
johanstokking: OK, but there's another issue with this, see below

pkg/qrcodegenerator/qrcode/enddevices/enddevices.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+var (
+	errUnknownFormat = errors.DefineInvalidArgument("unknown_format", "format `{format_id}` unknown")
+)
+
+// Parse attempts to parse the given QR code data.
+// It returns the parser and the format ID that successfully parsed the QR code.
+func (s *Server) Parse(formatID string, data []byte) (ret Data, err error) {
+	for id, format := range s.endDeviceFormats {
```
johanstokking: A map isn't ordered; this range is random access. This means that older QR code format versions (like drafts) may be attempted first, and they may even succeed (ignoring unknown fields that were introduced later, or different meaning to extensions etc), while we should try the latest version first. So the registration needs to via a slice (like it was previously).

pkg/qrcodegenerator/qrcode/enddevices/enddevices.go
```diff
@@ -0,0 +1,106 @@
...
+			continue
+		}
+		edFormat := format.New()
+		if err := edFormat.UnmarshalText(data); err == nil {
+			return edFormat, nil
+		} else if formatID == id {
+			return nil, err
+		}
+	}
+	return nil, errUnknownFormat.WithAttributes("format_id", formatID)
```
johanstokking: `formatID` may be empty here

pkg/deviceclaimingserver/enddevices/ttjs/ttjs.go
KrishnaIyer: It's already done at the entry in `grpc_end_devices.go`

pkg/deviceclaimingserver/enddevices/enddevices.go
KrishnaIyer: Ok for the pointers part. As for the original nit, for my eyes the original was more readable and hence didn't change it. But next time I'll add a note to that effect.

pkg/qrcodegenerator/qrcode/enddevices/enddevices.go
KrishnaIyer: Good point. Didn't think about that. Updated.

pkg/qrcodegenerator/grpc_end_devices.go
KrishnaIyer: > We really only have to check the incoming context for anything non-empty in AuthType and AuthValue.

`WithForwardedAuth` already does that actually. https://github.com/TheThingsNetwork/lorawan-stack/blob/9d58a370c158a60072ed841370a5e7a97b67f419/pkg/rpcmetadata/request.go#L40-L44

This doesn't cover ClusterAuth which is not set in the Metadata.

Should I maybe collect these lines in a single function to check "any" auth?

johanstokking (approved):

Issue #691, opened 2019-05-15
Event streaming for devices


kschiffer:
#### Summary
The console needs to support displaying event data for devices, both as widget in the overview page and as dedicated "Data" view.
References #28, #25 

#### Why do we need this?
Basic feature.

#### What is already there? What do you see now?
Event streaming for applications and gateways, event components and containers.

#### What is missing? What do you want to see?
A container for device events, event widget in device overview, device data sub-view.

#### Environment
Console in browser.

#### How do you propose to implement this?
Complementary to how it was solved with gateways and applications.

#### Can you do this yourself and submit a Pull Request?
Yes.

cc @bafonins

PR #5136, opened 2022-01-24
Improve PF UI

### LINKED ISSUE #2463: Improve payload formatters UI in the Console
#### Summary

I think it could be beneficial to change UI of payload formatters in the Console.

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2232

#### Why do we need this?

Better user experience. Currently, it is quite difficult to determine which formatter is used for a particular end device when both app/device formatters are set. Also, payload formatters options are not documented anywhere which is confusing (`None` vs `Repository`?). Additionally, refactoring the code for payload formatters (in the Console) would be nice as well.

Screenshot 2020-04-30 at 16 49 46

1. Inform the user when on the end device payload formatters page if there is app formatter set and which
2. Inform the user which payload formatter is used for a particular device
3. Overall improved UI
4. Separate UI's for app/device payload formatters pages

#### Environment

v3.7.2

yes

@laurensslats  @benolayinka  - documentation, suggestions, feedback
@kschiffer  @pierrephz - design, suggestions

cc @johanstokking

ryaplots:
#### Summary

- Closes [#691](https://github.com/TheThingsIndustries/lorawan-stack-support/issues/691)
- Closes #2463 

#### Changes

- Add paste buttons
Screenshot 2022-01-24 at 12 40 33
Screenshot 2022-01-24 at 12 40 46

- Change notification position and add information about the app PF
Screenshot 2022-01-24 at 12 40 59
Screenshot 2022-01-24 at 12 41 32

#### Testing

e2e tests and manually

#### Checklist


pkg/webui/components/notification/index.js
```diff
@@ -66,7 +67,7 @@ const Notification = ({
               content={content}
               values={messageValues}
               firstToUpper
-              convertBackticks={Boolean(error)}
+              convertBackticks={Boolean(error) || convertBackticks}
```
kschiffer: Let's actually just enable this by default in the `<Message />` component. I think that will save us a lot of hassle of propagating the prop through all the time.

pkg/webui/console/components/payload-formatters-form/index.js
```diff
@@ -43,15 +54,20 @@ const m = defineMessages({
   formatterType: 'Formatter type',
   formatterParameter: 'Formatter parameter',
   grpcFieldDescription: 'The address of the service to connect to',
-  appFormatter: 'Use application payload formatter',
+  appFormatter: 'Use application payload formatter  ',
```
kschiffer:
```suggestion
  appFormatter: 'Use application payload formatter',
```

pkg/webui/console/components/payload-formatters-form/index.js
```diff
@@ -86,10 +102,29 @@ const validationSchema = Yup.object().shape({
...
+    }
+  },
+  dispatch => ({
+    getRepositoryPayloadFormatters: (appId, versionIds) =>
+      dispatch(getRepositoryPayloadFormatters(appId, versionIds)),
+  }),
+)
+@withRequest(({ appId, versionIds, getRepositoryPayloadFormatters }) =>
+  getRepositoryPayloadFormatters(appId, versionIds),
+)
```
kschiffer: If you connect the component to the state, it will [no longer be a regular component but a container component](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.18/DEVELOPMENT.md#react-component-types). It should the be moved to the `container/` folder. I think it's better though to create a separate container that takes care of fetching and then passes down the data as props to the form component.

pkg/webui/console/components/payload-formatters-form/index.js
```diff
@@ -212,23 +279,58 @@ class PayloadFormattersForm extends React.Component {
...
-        return null
+          {type === TYPES.JAVASCRIPT && (
+            <>
+              <Button
+                type="button"
+                message={m.pasteApplicationFormatter}
+                secondary
+                icon={'payload_formats'}
+                onClick={pasteApplicationFormatter}
+              />
```
kschiffer: Keep in mind that the application PF can be set to `None`. In that case, the button should be hidden.

pkg/webui/console/components/payload-formatters-form/index.js
```diff
@@ -212,23 +279,58 @@ class PayloadFormattersForm extends React.Component {
...
-        )
-      default:
-        return null
+          {type === TYPES.JAVASCRIPT && (
+            <>
+              <Button
+                type="button"
+                message={m.pasteApplicationFormatter}
+                secondary
+                icon={'payload_formats'}
```
kschiffer:
```suggestion
```
Better without icon.

pkg/webui/console/components/payload-formatters-form/index.js
```diff
@@ -212,23 +279,58 @@ class PayloadFormattersForm extends React.Component {
...
+                onClick={pasteApplicationFormatter}
+              />
+              {Boolean(repositoryPayloadFormatters) && (
+                <Button
+                  type="button"
+                  message={m.pasteRepositoryFormatter}
+                  secondary
+                  icon={'payload_formats'}
```
kschiffer:
```suggestion
```

pkg/webui/console/components/payload-formatters-form/index.js
```diff
@@ -212,23 +279,58 @@ class PayloadFormattersForm extends React.Component {
...
+          type="text"
+          placeholder={sharedMessages.addressPlaceholder}
+          description={m.grpcFieldDescription}
+          autoComplete="on"
+        />
+      )
+    } else if (showRepositoryParameter) {
+      return (
+        <Form.Field
+          readOnly={isReadOnly}
```
kschiffer:
```suggestion
          readOnly
```

?

pkg/webui/console/containers/device-payload-formatters/downlink.js
```diff
@@ -88,10 +86,12 @@ class DevicePayloadFormatters extends React.PureComponent {
       }),
     }).isRequired,
     updateDevice: PropTypes.func.isRequired,
+    version_ids: PropTypes.shape({}),
```
kschiffer: I think this can be defined with at least the props that the console expects.

pkg/webui/console/containers/device-payload-formatters/downlink.js
```diff
@@ -60,6 +57,7 @@ import messages from './messages'
       link: selectApplicationLink(state),
       formatters,
       encodeDownlink: tts.As.encodeDownlink,
+      version_ids: selectVersionIds(state),
```
kschiffer: The new state props can be retrieved here and then passed on. See my comment above on moving state logic to the container prop.

pkg/webui/console/store/middleware/logics/device-repository.js
```diff
@@ -84,10 +84,44 @@ const getTemplateLogic = createRequestLogic({
...
+        version,
+      )
+      const repositoryPayloadFormatters = {
+        ...uplinkDecoder,
+        ...downlinkDecoder,
+        ...downlinkEncoder,
+      }
+      return repositoryPayloadFormatters
+    } catch (error) {
+      // There is no codec for this end device.
```
kschiffer: I believe it's also possible to have only one the codecs. Like this, no PF is recognized if only one is not existing.

Also, the catch is too general, there could be network or server errors that do not mean that the codecs do not exist. You'll probably want to check for `isNotFoundError(error)` only and rethrow any other error.

pkg/webui/console/store/middleware/logics/devices.js
```diff
@@ -34,7 +34,7 @@ const getDeviceLogic = createRequestLogic({
     } = action
     const dev = await tts.Applications.Devices.getById(appId, deviceId, selector)
     dispatch(devices.startDeviceEventsStream(dev.ids))
-    return dev
+    return { ...dev, version_ids: dev.version_ids }
```
kschiffer: Remove this change.

pkg/webui/console/store/reducers/devices.js
```diff
@@ -66,10 +67,11 @@ const devices = (state = defaultState, { type, payload, event }) => {
       return {
         ...state,
         selectedDevice: combineDeviceIds(payload.appId, payload.deviceId),
+        version_ids: payload.version_ids,
       }
     case UPDATE_DEV_SUCCESS:
     case GET_DEV_SUCCESS:
-      const updatedState = { ...state }
+      const updatedState = { ...state, version_ids: payload.version_ids }
```
kschiffer: I don't get these changes. The `version_ids` should be part of the device entity stored in the store.

kschiffer (changes requested):
Review is WIP but here are some change requests already.

Functionality-wise it's working well.

pkg/webui/console/store/middleware/logics/device-repository.js
```diff
@@ -104,15 +106,18 @@ const getRepositoryPayloadFormattersLogic = createRequestLogic({
...
       }
       return repositoryPayloadFormatters
     } catch (error) {
-      // There is no codec for this end device.
-      return null
+      if (isNotFoundError(error) && typeof repositoryPayloadFormatters !== 'undefined') {
+        return repositoryPayloadFormatters
+      }
+
+      throw error
```
kschiffer: Still not happy with this logic:

1. It runs three requests successively that could be run in parallel
2. If the first request errors out with a 404, the other requests will also not run (a thrown exception will stop further execution of the current `try`-block)

The if condition here can never be true because the error would occur before the assignment of `repositoryPayloadFormatters`

You will likely want something like this:
```js
  const ignoreNotFound = async (action, ...params) => {
    try {
      return await action(...params)
    } catch (e) {
      if (!isNotFound(e)) {
        throw e
      }
    }

    return undefined
  }

  const results = await Promise.all([
    ignoreNotFound(tts.Applications.Devices.Repository.getUplinkDecoder, ...),
    ignoreNotFound(tts.Applications.Devices.Repository.getDownlinkDecoder, ...),
    ignoreNotFound(tts.Applications.Devices.Repository.getDownlinkEncoder, ...),
  ])
  
  return {
    ...results[0],
    ...results[1],
    ...results[2],
  }
```

pkg/webui/console/components/payload-formatters-form/index.js
```diff
@@ -197,13 +204,46 @@ class PayloadFormattersForm extends React.Component {
...
+      return () =>
+        this.formRef?.current?.setFieldValue(FIELD_NAMES.JAVASCRIPT, applicationFormatter)
+    }
+
+    return () =>
+      this.formRef?.current?.setFieldValue(
+        FIELD_NAMES.JAVASCRIPT,
+        repoFormatters.formatter_parameter,
+      )
+  }
```
kschiffer: Mind the complexity/readability/maintainability costs of making unified handlers vs explicitly writing distinct handlers:

```javascript
  @bind
  pasteAppPayloadFormatter() {
    const { defaultParameter, uplink } = this.props
    this.formRef?.current?.setFieldValue(FIELD_NAMES.JAVASCRIPT, defaultParameter || getDefaultJavascriptFormatter(uplink))
  }

  @bind
  pasteRepoPayloadFormatter() {
    const { repoFormatters } = this.props
    this.formRef?.current?.setFieldValue(
      FIELD_NAMES.JAVASCRIPT,
      repoFormatters.formatter_parameter,
    )
  }
```

Don't sacrifice code clarity for a tiny memory improvement (if any).

pkg/webui/console/components/payload-formatters-form/index.js
```diff
@@ -197,13 +204,46 @@ class PayloadFormattersForm extends React.Component {
     }
   }
 
+  @bind
+  pastePayloadFormatter(app) {
+    const { defaultParameter, uplink, repoFormatters } = this.props
+    const applicationFormatter = defaultParameter
+      ? defaultParameter
+      : getDefaultJavascriptFormatter(uplink)
```
kschiffer:
```suggestion
    const applicationFormatter = defaultParameter || getDefaultJavascriptFormatter(uplink)
```

You can just use a logical OR to achieve the same thing.

pkg/webui/console/components/payload-formatters-form/index.js
kschiffer: Again, there is no reason to not hardcode the `readOnly` value as `true` here. The displayed repo formatter will always be read-only.

pkg/webui/console/components/payload-formatters-form/index.js
```diff
@@ -261,6 +336,7 @@ class PayloadFormattersForm extends React.Component {
         initialType === TYPES.GRPC ? initialParameter : getDefaultGrpcServiceFormatter(uplink),
     }
     const options = allowReset ? formatterOptionsWithReset : formatterOptions
+    const defaultFormatter = defaultType.replace('FORMATTER_', '').toLowerCase()
```
kschiffer:
```suggestion
    const defaultFormatter = defaultType.replace('FORMATTER_', '').toLowerCase()
```

Not a fan of solutions like these. It abuses our internal naming, which is completely arbitrary, as user-facing communication. Someone working on this at a later stage might not be aware of this liability and choose namings that are not OK for being displayed to users. Additionally, these strings are not internationalized and only work in English.

A better solution would be to create a mapping of formatter types to i18n messages and use that instead, similar to what we are doing with `console/lib/field-description-messages.js`.

pkg/webui/console/store/middleware/logics/devices.js
```diff
@@ -34,7 +34,7 @@ const getDeviceLogic = createRequestLogic({
     } = action
     const dev = await tts.Applications.Devices.getById(appId, deviceId, selector)
     dispatch(devices.startDeviceEventsStream(dev.ids))
-    return dev
+    return { ...dev }
```
kschiffer:
```suggestion
    return dev
```

pkg/webui/console/store/reducers/link.js
```diff
@@ -37,9 +37,8 @@ const getLinkFailure = (state, { payload }) => ({
   link: payload.link || {},
 })
 
-const updateLinkSuccess = (state, { link }) => {
-  const newLink = { ...state.link, ...link }
-
+const updateLinkSuccess = (state, link) => {
+  const newLink = { ...state.link, ...link.payload }
```
kschiffer, line 41: What's the rationale behind this change? Has this been broken all along?

kschiffer (changes requested):

pkg/webui/console/store/reducers/link.js
ryaplots: It was broken all along indeed. `{ link }` came as undefined every time. At least when I was testing. This is the reason behind that [support issue](https://github.com/TheThingsIndustries/lorawan-stack-support/issues/691)

pkg/webui/console/components/payload-formatters-form/index.js
ryaplots: This repoFormatter will also we shown when JavaScript type is selected. Shouldn't the user be able to edit it then? I thought the idea was that it is only read-only when the default type is JS and when the type is Repository. So there is one case when it should not be read-only - when the chosen type is JavaScript and then the user can paste and appFormatter or repoFormatter and adjust it as they wish. For that the editor cannot be read-only. Or did I misunderstand something?

pkg/webui/console/containers/devices-table/index.js
```diff
@@ -172,7 +172,13 @@ class DevicesTable extends React.Component {
       getDevicesList(
         props.appId,
         filters,
-        ['name', 'application_server_address', 'network_server_address', 'join_server_address'],
+        [
+          'name',
+          'application_server_address',
+          'network_server_address',
+          'join_server_address',
+          'version_ids',
```
kschiffer: The list request is not the right place to source the version ids, since it is only executed when accessing a device via the device table. When accessing an end device directly (e.g. by clicking on a link that goes to a specific end device), the version information will be missing since the list request did not run. To get the version IDs safely, they need to be sourced from the single device request in `pkg/webui/console/views/device/index.js`.

pkg/webui/console/store/middleware/logics/device-repository.js
```diff
@@ -84,10 +85,44 @@ const getTemplateLogic = createRequestLogic({
...
+      try {
+        return await action(...params)
+      } catch (e) {
+        if (!isNotFoundError(e)) {
+          throw e
+        }
+      }
+
+      return undefined
+    }
```
kschiffer: It's not necessary to reassign this function every time the logic runs. You can rather define the function outside of the logic scope.

ryaplots:
Button spacing now looks like:
Screenshot 2022-02-22 at 11 41 15

kschiffer (approved):

PR #5138, opened 2022-01-24
Dwell time boot configuration

### LINKED ISSUE #725: Record if UplinkDwellTime is enabled on boot
#### Summary

Record if `UplinkDwellTime` is enabled after device boot in band structure.
Refs #709 

#### Why do we need this?

Network Server can avoid sending a redundant `TxParamSetupReq`, in cases, where `UplinkDownlinkTime` is set on boot and required by the frequency plan.

Not this

A field indicating whether the `UplinkDwellTime` is enabled after device boot, if such information is recorded in the spec. E.g. in `1.0.3` AU915:
2019-05-21-17:18:12-screenshot
Network Server should use this value to construct the MAC state of device on reset.

`UplinkDwellTimeOnBoot *bool` on `band.Band` (nicer naming is welcome)
Update `newMACState` in NS

yes

### LINKED ISSUE #4971: Maximum downlink payload size exceeded when dwell time is activated
#### Summary

Maximum downlink payload size is exceeded when the end device has dwell time enabled, but the Network Server's frequency plan does not.

#### Steps to Reproduce

1. Register device with `AS_920_923_TTN_AU` (dwell time is deactivated)
2. Configure device with AS923 band, which boots with dwell time activated
3. Let end device join; this works successfully
4. Observe first downlink sent by Network Server

#### What do you see now?

That downlink message can be too long for the end device, which assumes that dwell time is activated. Consequently, the end device may reject the downlink, leading to an invalid state and potential loss of the downlink path.

For instance, it may carry FOpts `06 07 07 a0 af 8c 50 03 31 ff 00 01 09 05`, i.e. `TxParamSetupReq`, `DevStatusReq`, `LinkADRReq` and `NewChannelReq`. This `TxParamSetupReq` deactivates dwell time, but this downlink frame is discarded.

Network Server should not exceed the maximum payload size for downlink, even when dwell time is deactivated in the frequency plan, because it may still be activated in the end device.

In RP1 1.0.2 Rev. B band AS923, there is no default value specified for end device boot parameters. This means that NS cannot assume that dwell time is activated or not when the device joins. Wrongly using dwell time activated breaks the RX1 DR offset; wrongly using dwell time deactivated breaks the maximum payload size.

#### Environment

TTS 3.16.1, AS923 band, AS_920_923_TTN_AU frequency plan, LoRaWAN 1.0.2, RP1 1.0.2 rev. B.

NS always considers the dwell time setting from the frequency plan: https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.16/pkg/networkserver/downlink.go#L1243

To support AS923 RP1 1.0.2 Rev. B end devices with dwell time activated and deactivated on boot, we need a new `MACSettings` field (`BoolValue`). This should take precedence when configuring current parameters on MAC state reset.

Since LoRaMac-node has dwell time activated in this band and version, NS should assume that dwell time is activated when the new MAC settings field is not set and false.

In short:
1. Add `BoolValue {uplink,downlink}_dwell_time` to `MACSettings`
2. On MAC state reset, for configuring the current parameters, use the value from the new MAC setting (if set). If unset, use the new band version assumed value, which will be `true` for RP1 1.0.2 Rev. B band AS923. It should be commented that the assumed value comes from LoRaMac-node
3. Make sure that frequency plan's dwell time settings are only used on MAC state reset to configure desired parameters, not in other places
4. The maximum payload size is always the most restrictive between: MAC settings boot value, band version default and frequency plan. This is basically the most restrictive between the current and desired parameters, even if the desired parameters have not been conveyed yet to the end device.

Unit testing

Can review

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4971
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/725

#### Changes

- Add the boot dwell time configuration options to `MACSettings`
- Use the boot dwell time configuration in order to determine the current parameters of the end device after join/reset
  - Desired parameters are unchanged, only a bit of styling added
- Add the boot dwell time configuration to specific versions of the ~AS923 and~ AU915 bands
  - ~I've added these only to the ones which have a LoRaMAC node release~

#### Testing

Unit testing.

##### Regressions

I'm a bit uncomfortable about the AS923 having default uplink dwell time enabled at boot time. Per spec, the device is not forced to have this dwell time enabled, so we may reject uplinks which have dwell time disabled at boot time. Are we sure we want to have these boot time options available in that band ?

#### Checklist


pkg/networkserver/mac/utils.go
```diff
@@ -222,6 +222,32 @@ func DeviceDefaultPingSlotPeriodicity(dev *ttnpb.EndDevice, defaults ttnpb.MACSe
...
+	case dev.GetMacSettings().GetDownlinkDwellTime() != nil:
+		return &ttnpb.BoolValue{Value: dev.MacSettings.DownlinkDwellTime.Value}
+	case defaults.DownlinkDwellTime != nil:
+		return &ttnpb.BoolValue{Value: defaults.DownlinkDwellTime.Value}
+	case phy.BootDwellTime.Downlinks != nil:
+		return &ttnpb.BoolValue{Value: *phy.BootDwellTime.Downlinks}
+	default:
+		return nil
+	}
+}
```
johanstokking: Shouldn't we prioritize boot settings over default settings?

Do we need this part of default settings in the first place?

pkg/networkserver/networkserver_flow_test.go
```diff
@@ -39,7 +39,7 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/util/test/assertions/should"
 )
 
-func frequencyPlanMACCommands(macVersion ttnpb.MACVersion, phyVersion ttnpb.PHYVersion, fpID string, otaa bool) ([]MACCommander, []events.Builder) {
+func frequencyPlanMACCommands(macVersion ttnpb.MACVersion, phyVersion ttnpb.PHYVersion, fpID string, otaa bool) ([]MACCommander, []events.Builder, bool) {
```
johanstokking, line 46: Please good what that last return value means, because it isn't `ok`

htdvisser (commented):
I don't think we should base our assumptions on a specific MAC implementation (LoRaMac-node). Unless it's explicit in the spec (or in the device template), we should assume that on reset the end device is unaware of dwell time restrictions, and uses the worst-case boot configuration. So unless we're certain that the end device has the correct configuration, we should make sure that we send a `TxParamSetupReq` just in case.

> we may reject uplinks which have dwell time disabled at boot time

I don't think this is the right approach. If we reject uplinks from an end device that is unaware of the uplink dwell time restriction, we also won't be able to send a `TxParamSetupReq` that enables that dwell time limitation, so I think we need to be a bit more flexible here.

pkg/networkserver/mac/utils.go
adriansmares: Default settings are the Network Server default settings - in pretty much any parameter generation site we put them just bellow the device MAC settings, but above the band/frequency plan.

Keep in mind that this is for consistency - I didn't actually expose the setting in the NS config.

adriansmares:
I have reverted the boot settings for the AS923 band, because as was mentioned above, there is no guarantee per spec that the end device will actually have the same settings as LoRaMac-node. Should these end devices actually have this issue, it needs to be solved via MAC settings instead (and ideally indeed via device templates).

I've also tested this in the flow tests, and the previous commits (8d6b644317958246e9d08de0b3aa45e710d14629, fa1f2134145f49588f22a57664d8bc524fc438b9) do show that the NS will avoid sending the `LinkADRReq` if the dwell time MAC setting is enabled.

Should I follow up with an issue in the device repository schema and Console ?

johanstokking:
I agree that we should move forward now with assuming the most restrictive config in the end device if RP does not specify a boot value (like 1.0.2 rev B). That is, better safe than sorry and send `TxParamSetupReq` unless we're certain. That should fix the issue.

What was intended to be avoided when coming to the proposed implementation in #4971 was 1) having NS not transmit a dwell time instruction because there was no difference and 2) having NS not transmit unnecessary `TxParamSetupReq`s.

If we still want to support (2), we can do that optimization later, and I think it would be best to do that via the Device Repository indeed.

> Should I follow up with an issue in the device repository schema and Console ?

Yes please.

pkg/networkserver/mac/utils.go
johanstokking: So, in case of 1.0.2 rev. B with no per-device setting, this method returns `nil`. Then, only if the frequency plan has dwell time set (true or false), it will lead to a `TxParamSetupReq` per https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.17/pkg/networkserver/mac/tx_param_setup.go#L47-L56. Right?

Doesn't this mean that all frequency plans in bands where dwell time may be applicable, need to have a dwell time set explicitly?

Currently, in our general plans like `AS_923`, `AS_920_923` and `AS_923_925`, we don't set dwell time. That means that `TxParamSetupReq` is not sent.

Note that we cannot configure a dwell time because we don't know in which countries those plans are being used.

pkg/networkserver/utils.go
```diff
@@ -378,6 +379,32 @@ func publishEvents(ctx context.Context, evs ...events.Event) {
...
+	case dev.GetMacState().GetCurrentParameters().GetDownlinkDwellTime() != nil:
+		return dev.MacState.CurrentParameters.DownlinkDwellTime.Value
+	case fp.DwellTime.Downlinks != nil:
+		return *fp.DwellTime.Downlinks
+	case phy.BootDwellTime.Downlinks != nil:
+		return *phy.BootDwellTime.Downlinks
+	default:
+		return false
+	}
+}
```
johanstokking: That leads me here. In those Asian countries that use the generic AS frequency plans that do not configure dwell time, and where we do not know a boot value because (RP1), we assume that dwell time is deactivated here. That may very well lead to downlinks that are too long and end devices rejected the downlink frame.

Or am I missing anything?

adriansmares: You are correct, and not missing anything. 

We will not control the dwell time setting here, because frankly we don't know what value to set this to. I think that the frequency plan should include dwell time settings, explicitly, if it really wants a particular state - enabled or disabled - to be achieved. If we force the desired parameters to always include a conservative value (uplink disabled, downlink enabled), we're also affecting other bands. 

I propose the following:
- The frequency plans should explicitly enable or disable dwell time limitations
- If end devices have restrictive downlink dwell time limitations enabled at boot time, they should explicitly set these boot time settings via the MAC settings
- We assume the worst case scenario here and instead of defaulting to false, optimistically, we default to true, pessimistically. I don't feel this is very efficient, but it is what it is.

johanstokking: > I propose the following:
> 
> * The frequency plans should explicitly enable or disable dwell time limitations

I agree, but in generic plans we can't define dwell time. Those plans are used by RP2 versions too, that _do_ set a default. So if we set a default in the frequency plan, it should only be for RP1, and so that would mean we would need to prioritize the PHY defaults over the FP, which we shouldn't do.

So what we are looking for is a sane default if the end device uses a generic AS band plan (with dwell time unset), the device uses RP1 (with dwell time unset) and the dwell time has not been configured through MAC settings. This criteria match nearly 100% of AS923 devices today (not including AS923 in Australia).

This was the reason to go with a default that is widely deployed in the field, namely the one used by LoRaMac-node. That library or derivatives thereof are effectively used in most end devices, or are used as reference for implementations.

Other examples:
- Arm MbedOS: https://github.com/ARMmbed/mbed-os/blob/master/connectivity/lorawan/lorastack/phy/LoRaPHYAS923.cpp#L287-L288
- Arduino LMiC: https://github.com/mcci-catena/arduino-lmic/blob/master/src/lmic/lorabase_as923.h#L99-L102

@htdvisser do you still think we shouldn't go with these boot parameters in RP1?

> * If end devices have restrictive downlink dwell time limitations enabled at boot time, they should explicitly set these boot time settings via the MAC settings

All AS923 devices based on LoRaMac-node, MbedOS or MCCI's LMiC will have dwell time enabled at boot time, and we are introducing a new MAC setting that should be enabled for these end devices. So you are suggesting to explicitly enable downlink dwell time for all these devices, in retrospect as well.

We don't have support for this in the Device Repository and not in the Console, let alone as default value with tooltip etc. The CLI flag wouldn't also be enabled by default. In the Device Repository we would need to verify with device makers that their AS923 devices use LoRaMac-node, MbedOS or LMiC and if so, enable dwell time in the MAC settings. Still, we would need to apply that to all existing devices created that match these criteria, as part of a database migration.

> * We assume the worst case scenario here and instead of defaulting to false, optimistically, we default to true, pessimistically. I don't feel this is very efficient, but it is what it is.

Yes, so if we don't go with a sane boot parameter for RP1, we should return true here as default and use the most restrictive setting. We should also _always_ send `TxParamSetupReq` to make sure the end device then gets dwell time enabled. If we don't do that, and the end device does not use downlink dwell time and we do (because of `true`), downlink may be broken because the RX1 DR offset isn't correct (RX1 response to DR0/1 with DR2 I believe).

johanstokking: So we have two options when using the generic AS923 plans:

1. Use dwell time boot parameters in RP1 based on the default settings by all major LoRaWAN end device libraries. End devices that do not use these defaults should use the new MAC settings (to override the current params) or a frequency plan that explicitly sets the dwell time (to override the desired params)
2. Always enable dwell time and always configure this explicitly with `TxParamSetupReq`. This is inefficient but it is not impossible and it is safe to do.

---

I'm still more inclined to go with option 1.

htdvisser: > This was the reason to go with a default that is widely deployed in the field, namely the one used by LoRaMac-node. That library or derivatives thereof are effectively used in most end devices, or are used as reference for implementations.
> [...]
> @htdvisser do you still think we shouldn't go with these boot parameters in RP1?

With V2 we've made the mistake of following the defaults of certain implementations instead of the specification. Let's not make the same mistake now.

I don't know the details of the implementation in the NS, but perhaps it's possible to just accept that we don't know, so accepting that `UplinkDwellTime` and `DownlinkDwellTime` can be `1`, `0` or `unknown` until we actually need to know.

It's clear what we need to do for `1` and `0`. For `unknown`, if we only receive uplinks <400ms, we don't need to do anything. If we receive an uplink >400ms, we know that the device thinks that `UplinkDwellTime=0`, and we can adjust if needed. If we only need to send downlinks <400ms, we don't need to do anything. If we need to send a downlink >400ms to an RP1 device (nb: RP2 devices accept >400ms downlinks even if `DownlinkDwellTime=1`) we first need to adjust the params so that we're sure the device will accept the downlink.

And finally, we can choose to piggy-back the `TxParamSetupReq` onto existing downlinks only (if there is space), because I agree that it's undesirable to send downlink carrying only an unnecessary `TxParamSetupReq`.

johanstokking: > With V2 we've made the mistake of following the defaults of certain implementations instead of the specification. Let's not make the same mistake now.

There is no "instead of the specification" here, because there is nothing in the specification. In RP1 this was left open, and we see that the three major LoRaWAN end device stacks use a different implementation than The Things Stack.

A thing that we learned from V2 is that we should not guess things based on perceived behavior. Uplink and downlink dwell time are not necessarily the same. For downlink dwell time, even if we have less than 400 ms to send, we have to deal with the RX1 data rate offset. NS doesn't know which DR to use when the end device used DR0 or DR1 for uplink:

Screenshot 2022-01-27 at 18 08 29

For uplink, guessing whether uplink dwell time is enabled or not based on uplink message length isn't ideal. But apart from regional compliance checking, this isn't too important.

For downlink, we really need to send `TxParamSetupReq` and wait for `TxParamSetupAns` before we have a reliable downlink path.

---

I'd propose the following combination:

1. Reinstate the defaults based on boot parameters. This makes RX1 more reliable when the device uses DR0 or DR1. We need this RX1 for `TxParamSetupReq`. This configures current parameters. The value in the device's MAC settings takes priority. I still think this is a better default than what we have now, and there's no way around choosing a setting before we got `TxParamSetupAns`
2. When the frequency plan does not have a value for dwell time set, assume it is `true` for both directions: the most restrictive case. This configures the desired parameters
3. Send `TxParamSetupReq` after all MAC state resets, unless the dwell time settings of the device's MAC settings (so _not_ the current parameters) match the desired parameters already
4. Add frequency plans for HK and Taiwan with dwell time activated

adriansmares: 1. :heavy_check_mark:
2. Since we're going to default to `true` in the desired parameters, instead of `nil`, this will enable bidirectional dwell time explicitly in AU915 for 1.0.3+ devices. Do we want this ?
For reference, past behavior so far was to assume false bidirectionally.
3. :heavy_check_mark:
4. I assume this is for the Gateway Server mainly, since the NS will use `true` anyway due to point 2. I can follow up on this.

Extra question: what value do we assume for devices for which the current parameters are `nil`, in case there are no boot defaults ? `true` or `false` ?
For uplinks `false` is sane, but for downlink neither value is useful due to the RX1 offset issue.
For reference, before this change we assumed the value in the frequency plan, and defaulted to false otherwise. Also keep in mind that we're affecting bands such as `US915` with this decision, so maybe falling back to the frequency plan if there are no boot time settings is not evil here.

adriansmares:
Based on the discussions here, the behavior of the last two commits is as follows:

Current parameters - MAC settings provided value, `nil` otherwise
Desired parameters - Frequency Plan provided value,  `true` otherwise

Expected parameters (the value used for payload size limitation and RX1 DR computations) - MAC state current parameters if they exist, falling back to the boot time defaults in the band, falling back to the frequency plan settings, falling back to `false` for uplink and `true` for downlink (in order to avoid having the downlink rejected due to size).

This allows us to use the boot time defaults before the first `TxParamSetupReq`/`Ans` dance happens, while also sending a setup request to devices without a MAC setting provided.

Edit: updated based on the last commit.

pkg/networkserver/utils.go
johanstokking: > 2\. Since we're going to default to `true` in the desired parameters, instead of `nil`, this will enable bidirectional dwell time explicitly in AU915 for 1.0.3+ devices. Do we want this ?
>     For reference, past behavior so far was to assume false bidirectionally.

Isn't that an issue in the frequency plan then?

> 4\. I assume this is for the Gateway Server mainly, since the NS will use `true` anyway due to point 2. I can follow up on this.

Yeah indeed, this is less important. It's good to be explicit. No need to address this now.

> Extra question: what value do we assume for devices for which the current parameters are `nil`, in case there are no boot defaults ? `true` or `false` ?
> [...]
> so maybe falling back to the frequency plan if there are no boot time settings is not evil here.

Yes I think we should that in this case.

But aren't there always boot defaults?

adriansmares: > Isn't that an issue in the frequency plan then?

The RP document indeed marks this as a 'regional dependence', so indeed we will fix this using new frequency plans.

> But aren't there always boot defaults?

I've added boot defaults only to bands that explicitly have the text `<band> end-devices SHALL always consider *DwellTime = <value>`, and US915. 

For some AS923 combinations for which there never was a release of LoRaMac-node, we don't really have boot defaults, nor any frequency plan dwell time settings (in the generic plan). We will enable dwell time now for these versions, due to the desired `true` fallback, but we have to fit into RX2 worst case (due to the possible offset mismatch).

johanstokking: OK that sounds great.

johanstokking (approved):

PR #5171, opened 2022-02-03
Remove old session keys

### LINKED ISSUE #5086: Cleanup Join Server session keys
#### Summary

Cleanup the Join Server session keys

References https://github.com/TheThingsIndustries/lorawan-stack/issues/2229

#### Why do we need this?

To avoid unbound growth of the Join Server session key store.

They are currently persisted without limits. Also there isn't an index.

We should keep 3 session keys. That should be enough for Application Servers to recover the AppSKey.

1. Store the session keys in two lists. The first list is for join-requests and rejoin-requests type 2. In that case, the end device is either on join mode (has no session keys anymore) or is requested to force rejoin via `ForceRejoinReq` so its existing session keys are void. The second list is for rejoin-requests type 0 and 1. The end device may continuously send rejoin-requests type 0 and 1 which are not necessarily answered or may vanish in the air. That should not lead to rolling over the first list. Rejoin type 0 may also be the result of `ForceRejoinReq`, but it is unknown to the Join Server what the reason was for the end device to transmit the rejoin-request, so it's safe to store that with rejoin-request type 1
2. In a database migration script, enumerate all session keys per end device. The JoinEUI and DevEUI can only be parsed from the Redis key. It's safe to sort the keys ascending as the session key ID is an ULID
   - Since the existing key registry doesn't record the rejoin-request type, we should store the last 3 keys in both lists
   - The other keys can be discarded

Local testing

Yes

johanstokking:
#### Summary

Closes #5086 

#### Changes

- Record the last N (default 10) session keys in a Redis list. New key that are pushed lead to pruning the oldest ones
- Add `js-db migrate` to enumerate session key IDs for end devices and puts them in the list. Old session key IDs are pruned

#### Testing

Local testing of the migration and unit testing

##### Regressions

None expected

#### Notes for Reviewers

@adriansmares the JS session keys Redis implementation is likely not the fastest one as it does some reading and writing that can't be pipelined. We could work with a Lua script but this join flow isn't time critical enough for that in my opinion.

What is currently not being pruned are session keys of devices that are deleted or that are being deleted. Automatic deletion of session keys on device delete is a bit trickier because the JS device registry isn't configured with the key namespace of the session keys. We could do this with a separate `js-db prune` or `cleanup` analog to NS, but I don't see that in the scope of closing this issue.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Added
 
+- Retain at most 10 recent session keys in the Join Server. This avoids a slowly growing number of session keys in the Join Server's database.
+  - This requires a database migration (`ttn-lw-stack js-db migrate`).
+
 ### Changed
 
 ### Deprecated
--- a/cmd/internal/shared/joinserver/config.go
+++ b/cmd/internal/shared/joinserver/config.go
@@ -24,5 +24,6 @@ var DefaultJoinServerConfig = joinserver.Config{
 	JoinEUIPrefixes: []types.EUI64Prefix{
 		{},
 	},
-	DevNonceLimit: 10,
+	DevNonceLimit:   10,
+	SessionKeyLimit: 10,
 }
--- a/cmd/ttn-lw-stack/commands/as_db.go
+++ b/cmd/ttn-lw-stack/commands/as_db.go
@@ -45,11 +45,11 @@ var (
 			cl := ttnredis.New(config.Redis.WithNamespace("as"))
 
 			if force, _ := cmd.Flags().GetBool("force"); !force {
-				needMigration, err := checkLatestSchemaVersion(cl, asredis.SchemaVersion)
+				schemaVersion, err := getSchemaVersion(cl)
 				if err != nil {
 					return err
 				}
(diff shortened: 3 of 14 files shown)


adriansmares (approved):
> @adriansmares the JS session keys Redis implementation is likely not the fastest one as it does some reading and writing that can't be pipelined. We could work with a Lua script but this join flow isn't time critical enough for that in my opinion.

Agreed.

> What is currently not being pruned are session keys of devices that are deleted or that are being deleted. Automatic deletion of session keys on device delete is a bit trickier because the JS device registry isn't configured with the key namespace of the session keys. We could do this with a separate js-db prune or cleanup analog to NS, but I don't see that in the scope of closing this issue.

I'm ok with this, but please follow up with a `prune` command issue.

johanstokking:
> I'm ok with this, but please follow up with a `prune` command issue.

I decided to add it here. Let's close issues instead of making new ones.

Please only review https://github.com/TheThingsNetwork/lorawan-stack/pull/5171/files/82dbe71dde2652aef7f78b5e7b4ca2dce96a702b..a2be3a9531e2dfc2c90c357badb0f79dc27c5951

adriansmares (approved):

PR #5174, opened 2022-02-03
Add gRPC implementation to GCS API

### LINKED ISSUE #3280: Make GCS API available via gRPC
#### Summary
The Gateway Configuration Server currently does not use gRPC so it's API is plain HTML. In the JS SDK we somewhat rely on functionality being available as RPC so adding GCS functionality requires us to break out of our usual implementation.

#### Why do we need this?
- Retain implementation consistency in the JS SDK
- Avoid harcoded API routes and arbitrary implementations

gRPC API of all other stack components

gRPC API for the GCS

Likely recreate the implementation for the other stack components.

Unit tests.

No, @KrishnaIyer @adriansmares please reassign as appropriate.

nicholaspcr:
#### Summary

Closes #3280

#### Changes

- Add proto definition of GCS service
- Add grpc implementation of GCS service
- Add gRPC tests

#### Testing

Unit tests and manually testing by downloading the global config in the gateway page of console.

##### Regressions

The gRPC service is adjacent to the http routes that already exists, meaning that the route generated by the `grpc-gateway` is distinct from the previous routes. 

#### Notes for Reviewers

After the discussion in this PR thread, it was decided to create the gRPC http routes alongside with the previous http routes. This avoid the API breaking, in practical terms the http routes become something similar to: 
- (old_http) : `http://localhost:1885/api/v3/gcs/gateways/{gateway_id}/semtechudp/global_conf.json`
- (grcp_http): `http://localhost:1885/api/v3/gcs/gateways/configuration/{gateway_id}/{format}/{filename}`

#### Checklist


api/gateway_configuration.proto
```diff
@@ -0,0 +1,40 @@
...
+import "lorawan-stack/api/identifiers.proto";
+
+package ttn.lorawan.v3;
+
+option go_package = "go.thethings.network/lorawan-stack/v3/pkg/ttnpb";
+
+
+message GetGatewayConfigurationRequest {
+  GatewayIdentifiers gateway_ids = 1;
+  string format = 2;
```
htdvisser: This probably needs some validation.

htdvisser (commented):
I wasn't involved in the original discussion about building this feature, so this may be a bit late, but still. I don't think that "in the JS SDK we expect everything to be gRPC" is a good argument for forcing something that should be a simple file download into a gRPC API. Are we sure we really need to do this here? Why can't the SDK just support the file downloads that we already have?

api/gateway_configuration.proto
nicholaspcr: I'm not sure what should be allowed in the format so I opted to just leave lowercase letters with hyphens. Let me know what you think.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ For details about compatibility between different releases, see the **Commitment
 ## [Unreleased]
 
 ### Added
+- Add gRPC service to GatewayConfigurationServer, gateway configurations can now be obtained via gRPC requests.
 
 - Retain at most 10 recent session keys in the Join Server. This avoids a slowly growing number of session keys in the Join Server's database.
```
```suggestion

- gRPC service to Gateway Configuration Server so that gateway configurations can be obtained via gRPC requests.
- Retain at most 10 recent session keys in the Join Server. This avoids a slowly growing number of session keys in the Join Server's database.
```

api/gateway_configuration.proto
```diff
@@ -0,0 +1,41 @@
...
+
+import "github.com/envoyproxy/protoc-gen-validate/validate/validate.proto";
+import "google/api/annotations.proto";
+import "lorawan-stack/api/identifiers.proto";
+
+package ttn.lorawan.v3;
+
+option go_package = "go.thethings.network/lorawan-stack/v3/pkg/ttnpb";
+
+
```
johanstokking:
```suggestion

```

api/gateway_configuration.proto
```diff
@@ -0,0 +1,41 @@
...
+message GetGatewayConfigurationRequest {
+  GatewayIdentifiers gateway_ids = 1 [(validate.rules).message.required = true];
+  string format = 2 [(validate.rules).string = {pattern: "^[a-z]+(-[a-z]+)*$"}];
```
johanstokking: Maybe use the ID regex we use everywhere?
For sure include a length check

pkg/gatewayconfigurationserver/grpc.go
```diff
@@ -0,0 +1,100 @@
+// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
```
johanstokking:
```suggestion
// Copyright © 2022 The Things Network Foundation, The Things Industries B.V.
```

johanstokking (commented):
Please rebase and `make init` because `data/lorawan-devices` diff doesn't belong here

pkg/gatewayconfigurationserver/grpc.go
johanstokking: Also at least one other place

api/gateway_configuration.proto
nicholaspcr: was going to do that initially but wasn't sure if format should include numbers or not. Since its fine will use that same regex as the others

CHANGELOG.md
johanstokking: This isn't addressed

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ For details about compatibility between different releases, see the **Commitment
 ## [Unreleased]
 
 ### Added
+- gRPC service to GatewayConfigurationServer, gateway configurations can now be obtained via gRPC requests.
```
johanstokking: See my previous review comment

johanstokking (approved):

htdvisser (approved):
LGTM, but please also wait for review from @KrishnaIyer as main GCS code-owner

pkg/gatewayconfigurationserver/grpc.go
```diff
@@ -0,0 +1,100 @@
...
+
+	types "github.com/gogo/protobuf/types"
+	"go.thethings.network/lorawan-stack/v3/pkg/auth/rights"
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/pfconfig/cpf"
+	"go.thethings.network/lorawan-stack/v3/pkg/pfconfig/semtechudp"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+var errInvalidConfigurationFormat = errors.DefineInvalidArgument("invalid_configuration_format", "configuration format is not supported")
```
KrishnaIyer: You can add the format as an attribute to the error (ex: `{format}`).

pkg/gatewayconfigurationserver/server.go
```diff
@@ -166,6 +69,9 @@ func New(c *component.Component, conf *Config) (*Server, error) {
 	v2GCS := gcsv2.New(c, gcsv2.WithTheThingsGatewayConfig(conf.TheThingsGateway))
 	_ = v2GCS
 
+	hooks.RegisterUnaryHook("/ttn.lorawan.v3.GCs", rpclog.NamespaceHook, rpclog.UnaryNamespaceHook("gatewayserver"))
+	hooks.RegisterUnaryHook("/ttn.lorawan.v3.GCs", cluster.HookName, c.ClusterAuthUnaryHook())
```
KrishnaIyer: The service is `GatewayConfigurationService` so shouldn't that be registered in the hook?

pkg/gatewayconfigurationserver/grpc.go
```diff
@@ -0,0 +1,100 @@
...
+		configContent, err = json.Marshal(config)
+
+	case "lorafwd":
+		config, err := cpf.BuildLorafwd(gtw)
+		if err != nil {
+			return nil, err
+		}
+		configContent, err = config.MarshalText()
+	default:
+		return nil, errInvalidConfigurationFormat.New()
```
KrishnaIyer: And use this format here as an attribute.

pkg/gatewayconfigurationserver/http.go
```diff
@@ -0,0 +1,117 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/pfconfig/semtechudp"
+	"go.thethings.network/lorawan-stack/v3/pkg/ratelimit"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	"go.thethings.network/lorawan-stack/v3/pkg/web"
+	"go.thethings.network/lorawan-stack/v3/pkg/webhandlers"
+	"go.thethings.network/lorawan-stack/v3/pkg/webmiddleware"
+)
+
+// RegisterRoutes registers the web frontend routes.
+func (s *Server) RegisterRoutes(server *web.Server) {
```
KrishnaIyer, line 38: Aren't we replacing these routes with the gRPC service? 

The gRPC gateway paths are defined in your new proto as 
```
"/gcs/gateways/{gateway_ids.gateway_id}/configuration/{format}"
```

If you want a Semtech UDP file, you'd make a request like

1. Through gRPC gateway
```
/gcs/gateways/{gateway_ids.gateway_id}/configuration/semtechudp
```

2. Through the existing web route
```
/gcs/gateways/{gateway_id}/semtechudp/global_conf.json
```

These look like they do the same thing. 

If you add a `string filename` field to your `GetGatewayConfigurationRequest` message and remove the "configuration" word, then the gRPC request would become
```
/gcs/gateways/{gateway_ids.gateway_id}/{format}/{filename}
```

This is the same as what's duplicated in these web routes right? So why not make harmonize these and get rid of the web routes?

pkg/gatewayconfigurationserver/grpc.go
```diff
@@ -0,0 +1,100 @@
...
+		},
+	}, s.WithClusterAuth())
+	if err != nil {
+		return nil, err
+	}
+	return gtw, nil
+}
+
+func (s *Server) GetGatewayConfiguration(ctx context.Context, req *ttnpb.GetGatewayConfigurationRequest) (*ttnpb.GetGatewayConfigurationResponse, error) {
+	if s.config.RequireAuth {
```
KrishnaIyer, line 58: I can't really think of a valid use case for disabling auth here. The configuration for a specific gateway is never public right?  We had this already in the codebase but I can't remember why. @htdvisser ?

KrishnaIyer (changes requested):

pkg/gatewayconfigurationserver/http.go
nicholaspcr: I see it, the http and grpc routes could be merged. In addition of the `filename` field could open space for more configuration files without adding extra routes. 

Will work on that change.

nicholaspcr: It slipped by me when looking it at first  but there is a problem with this solution, it becomes more clear when looking at the previously existent paths:
- `/gcs/gateways/{gateway_id}/semtechudp/global_conf.json`
- `/gcs/gateways/{gateway_id}/kerlink-cpf/lorad/lorad.json`
- `/gcs/gateways/{gateway_id}/kerlink-cpf/lorafwd/lorafwd.toml`

The suggestion `/gcs/gateways/{gateway_ids.gateway_id}/{format}/{filename}` becomes something where `{format}` has to include something like `kerlink-cpf/loradfwd` which from my tests locally doesn't seem possible to do.

I'm a fan of making a gRPC gateway that replaces the the three previously defined http routes but I'm struggling to see a solution that doesn't change the api. So in my mind it might be more worth it to leave it as is.

pkg/gatewayconfigurationserver/server.go
nicholaspcr: Wasn't sure if you were referring to the namespace name or the `GCs` in the filter. 

Either way I changed the filter to refer to the `GatewayConfigurationService` instead of `GCs` and changed the namespace from `gatewayserver` to `gatewayconfigurationserver`.

pkg/gatewayconfigurationserver/http.go
nicholaspcr: @KrishnaIyer small reminder of this discussion

KrishnaIyer: Ok in that case can you add a `type` parameter? 
Then you can have `/gcs/gateways/{gateway_ids.gateway_id}/{format}/{type}/{filename}`.

For the Semtech UDP, since the type is not needed, you can add an [additional binding](https://github.com/TheThingsNetwork/lorawan-stack/blob/1c018c71892ef2c258b99fc751ca44efa1abb8a3/api/application_services.proto#L165) to that path without the type.

nicholaspcr: From our discussion and tests, the problem with replacing the existent routes with the ones that are generated by the gRPC-gateway is the fact that the response body changes in such a way that the json of the configuration stays as the base64 of the byte array. So when doing a request to the GCS gRPC service the response will alway be:
```json
{
   "contents": "<base64_byte_array>"
}
```

Since this changes the API, is best to keep the old implementation.

pkg/gatewayconfigurationserver/grpc.go
htdvisser: At some point in the past we allowed Basic Station gateways to connect without auth. We don't do that anymore on any of our production deployments, but it's not impossible that some private deployments still allow it.

htdvisser (approved):
Nothing to add

pkg/gatewayconfigurationserver/grpc.go
```diff
@@ -0,0 +1,127 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/frequencyplans"
+	"go.thethings.network/lorawan-stack/v3/pkg/pfconfig/cpf"
+	"go.thethings.network/lorawan-stack/v3/pkg/pfconfig/semtechudp"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+var errInvalidConfigurationFormat = errors.DefineInvalidArgument("invalid_configuration_format", "configuration format `{format}` is not supported")
+var errInvalidConfigurationType = errors.DefineInvalidArgument("invalid_configuration_type", "configuration type `{type}` is not supported")
+var errInvalidConfigurationFilename = errors.DefineInvalidArgument("invalid_configuration_filename", "configuration filename `{filename}` is not supported")
```
KrishnaIyer: Nit. Since the error type is already `InvalidArgument`, I'd recommend replacing the word `Invalid` with `Unsupported` or `NotSupported` in the variable and the error name.

pkg/gatewayconfigurationserver/grpc.go
KrishnaIyer: Yes but these routes are not for LBS. These are those json/toml files for UDP gateways. So now anyone can just insert a gateway ID here and get the configuration for a particular gateway. 
Not that it's a huge security issue but I think we can avoid some confusion here.

pkg/gatewayconfigurationserver/grpc.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	var configContent []byte
+	switch req.Format {
+	case "semtechudp":
+		configContent, err = handleSemtechUDP(gtw, fps, req)
+
+	case "kerlink-cpf":
+		configContent, err = handleKerlinkCPF(gtw, fps, req)
+
```
KrishnaIyer: Nit: You can remove these extra new lines.

pkg/gatewayconfigurationserver/grpc.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, err
+	}
+	return json.Marshal(config)
+}
+
+func handleKerlinkCPF(gtw *ttnpb.Gateway, fps *frequencyplans.Store, req *ttnpb.GetGatewayConfigurationRequest) ([]byte, error) {
+	switch req.Type {
+	case "lorad":
+		if req.Filename != "lorad.json" {
+			return nil, errInvalidConfigurationFilename.WithAttributes("filename", req.Filename)
```
KrishnaIyer: I think it would be good to also log the type here.
```
configuration filename `{filename}` for type `{type}` is not supported
```

Also in some of the other log statements.

pkg/gatewayconfigurationserver/http.go
```diff
@@ -0,0 +1,133 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/pfconfig/cpf"
+	"go.thethings.network/lorawan-stack/v3/pkg/pfconfig/semtechudp"
+	"go.thethings.network/lorawan-stack/v3/pkg/ratelimit"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	"go.thethings.network/lorawan-stack/v3/pkg/web"
+	"go.thethings.network/lorawan-stack/v3/pkg/webhandlers"
+	"go.thethings.network/lorawan-stack/v3/pkg/webmiddleware"
+)
+
+// RegisterRoutes registers the web frontend routes.
```
KrishnaIyer, line 34: Can you please add some notes here as to why these routes exists even though we have gRPC routes?

KrishnaIyer (commented):
Looks really good. I only have tiny comments.

Have you tested that the old routes (currently used by the console) work with these changes?

pkg/gatewayconfigurationserver/http.go
nicholaspcr: Left in the godoc a bit of explanation of why this isn't the only route capable of return the gateway configuration and the reason for such occurrence. Let me know what you think

nicholaspcr:
> Looks really good. I only have tiny comments.
> 
> Have you tested that the old routes (currently used by the console) work with these changes?

Just to be sure I tested again, the gateway configuration is returning the expected object in the old routes and in the new routes the response is the bytes of such object. Also checked if the `Download global_conf.json` button in the console returns what is expected, which it does.

pkg/gatewayconfigurationserver/http.go
```diff
@@ -0,0 +1,137 @@
...
+//
```
KrishnaIyer, line 35: I don't think it's Go convention to leave an empty line but I might be wrong.

pkg/gatewayconfigurationserver/http.go
```diff
@@ -0,0 +1,137 @@
...
+// These routes are not replaced by the ones generated in the gRPC-gateway, the reason for that is due to the break in
+// the API that would occur if such replacement was made. Therefore, for legacy reasons, there are two distinct groups
+// of http routes that returns the gateway configuration.
```
KrishnaIyer: I suggest the following
```
The gateway configuration value returned by the `grpc-gateway` routes are not content formatted, but a stream of bytes. This would be a breaking change for the consumers of this API and hence these routes are retained.
```

KrishnaIyer (approved):
LGTM

pkg/gatewayconfigurationserver/http.go
nicholaspcr: I believe it is part of the convention, specially since there are some sections of the standard library that utilizes that style of comment. For example the `Seek` method in the `os` pkg:
```go
// Seek sets the offset for the next Read or Write on file to offset, interpreted
// according to whence: 0 means relative to the origin of the file, 1 means
// relative to the current offset, and 2 means relative to the end.
// It returns the new offset and an error, if any.
// The behavior of Seek on a file opened with O_APPEND is not specified.
//
// If f is a directory, the behavior of Seek varies by operating
// system; you can seek to the beginning of the directory on Unix-like
// operating systems, but not on Windows.
func (f *File) Seek(offset int64, whence int) (ret int64, err error) {
	if err := f.checkValid("seek"); err != nil {
		return 0, err
	}
	r, e := f.seek(offset, whence)
	if e == nil && f.dirinfo != nil && r != 0 {
		e = syscall.EISDIR
	}
	if e != nil {
		return 0, f.wrapErr("seek", e)
	}
	return r, nil
}
```

I believe it's commonly used for package documentation, I opted to do that here to add a line just to separate the comment regarding the api in its own paragraph.

PR #5191, opened 2022-02-07
Fix `NwkKey` generation and usage in 1.0.x activation

johanstokking:
#### Summary

This is a quickfix to address the regression from 3.17.2 with 1.0.x devices that have root keys generated with the CLI not joining anymore.

#### Changes

- CLI no longer generates NwkKey when LoRaWAN version is specified and it's below 1.1. We do not require the LoRaWAN version to be specified, in case a device is not created on NS. That would be a breaking CLI change
- Work-around for JS to use AppKey even when a NwkKey is set but it was generated with the CLI

#### Testing

Tested CLI locally, JS work-around is unit tested

#### Notes for Reviewers

This doesn't win the beauty contest for sure, but I don't really see a good circumvention for this.

We need the CLI fix because when generating two root keys, they according to spec they are 1.1.

We need the JS work-around because of the end devices created with CLI. We can't just remove NwkKey from these devices in JS, because they may later want to use 1.1 activation. Also which LoRaWAN version is used is known by NS only so it's not that straight forward to remove NwkKey in the JS database. On top of that, most users won't update the CLI immediately.

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/end_devices.go
+++ b/cmd/ttn-lw-cli/commands/end_devices.go
@@ -476,11 +476,21 @@ var (
 					if device.ProvisionerId != "" {
 						return errEndDeviceKeysWithProvisioner.New()
 					}
-					// TODO: Set JoinEUI and DevEUI (https://github.com/TheThingsNetwork/lorawan-stack/issues/47).
 					device.RootKeys = &ttnpb.RootKeys{
 						RootKeyId: "ttn-lw-cli-generated",
 						AppKey:    &ttnpb.KeyEnvelope{Key: generateKey()},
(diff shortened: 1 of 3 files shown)


adriansmares (approved):
LGTM. 

For the end devices that really have both keys provisioned, the only solution that is reasonable solution remains to unset the `NwkKey`, right ?

johanstokking:
> For the end devices that really have both keys provisioned, the only solution that is reasonable solution remains to unset the `NwkKey`, right ?

Yes.

Actually, 3.17.2 was a fix for that. It shouldn't have worked when AppKey did not equal NwkKey, both were present, and AppKey was expected to be used.

PR #5199, opened 2022-02-09
Allow for multiple block creation for the same eui block type

### LINKED ISSUE #5045: EUIStore.CreateEUIBlock Updates Existing EUI Block
#### Summary

Unlike the name suggests, the `CreateEUIBlock` method of the current `EUIStore` implementation may change an existing EUI block.

#### What do you see now?

https://github.com/TheThingsNetwork/lorawan-stack/blob/dfa369c4457a74b0c04684684a6b8f522e83fb4d/pkg/identityserver/store/eui_block_store.go#L115-L140

A method that suggests we're creating something, shouldn't update something.

Re-implement.

Sure

pgalic96:
#### Summary

Closes #5045 

#### Changes

- Remove unique constraint and index for the `eui_type` field
- Updating config creates a new block, not overwriting the old one
- Issuing a dev block now selects all blocks that did not reach the limit, and picks a first one to issue DevEUI from

#### Testing

unit

##### Regressions

should be covered by tests

#### Notes for Reviewers

This will require a migration because the `type` was unique

#### Checklist

### DIFF
--- a/cmd/ttn-lw-stack/commands/is_db.go
+++ b/cmd/ttn-lw-stack/commands/is_db.go
@@ -25,6 +25,7 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/identityserver/gormstore/migrations"
 	"go.thethings.network/lorawan-stack/v3/pkg/identityserver/store"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	ttntypes "go.thethings.network/lorawan-stack/v3/pkg/types"
 )
 
 var (
@@ -291,6 +292,52 @@ var (
 			return nil
 		},
 	}
+	isDBEUIBlockCreationCommand = &cobra.Command{
+		Use:   "create-eui-block",
+		Short: "Create an EUI block in IS db (currently only DevEUI block supported)",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			logger.Info("Connecting to Identity Server database...")
+			db, err := gormstore.Open(ctx, config.IS.DatabaseURI)
+			if err != nil {
+				return err
+			}
+			defer db.Close()
+			useConfig, err := cmd.Flags().GetBool("use-config")
+			if err != nil {
+				return err
+			}
+			if useConfig {
+				logger.Info("Using config values...")
+				return gormstore.GetEUIStore(db).CreateEUIBlock(ctx, config.IS.DevEUIBlock.Prefix, config.IS.DevEUIBlock.InitCounter, "dev_eui")
+			}
+			prefix, err := cmd.Flags().GetString("prefix")
+			if err != nil {
+				return err
+			}
+			euiPrefix := &ttntypes.EUI64Prefix{}
+			if err := euiPrefix.UnmarshalConfigString(prefix); err != nil {
+				return err
+			}
+			counter, err := cmd.Flags().GetInt64("init-counter")
+			if err != nil {
+				return err
+			}
+			euiType, err := cmd.Flags().GetString("eui-type")
+			if err != nil {
+				return err
+			}
+			switch euiType {
+			case "dev_eui":
+				if err := gormstore.GetEUIStore(db).CreateEUIBlock(ctx, *euiPrefix, counter, euiType); err != nil {
+					return err
+				}
+				logger.Info("Block created successfully")
+			default:
+				logger.Error("Unsupported eui type")
+			}
+			return nil
+		},
+	}
 )
 
 func init() {
@@ -299,4 +346,9 @@ func init() {
 	isDBCommand.AddCommand(isDBMigrateCommand)
 	isDBCleanupCommand.Flags().Bool("dry-run", false, "Dry run")
 	isDBCommand.AddCommand(isDBCleanupCommand)
+	isDBEUIBlockCreationCommand.Flags().Bool("use-config", false, "Create block using values from config")
+	isDBEUIBlockCreationCommand.Flags().String("eui-type", "dev_eui", "EUI block type")
+	isDBEUIBlockCreationCommand.Flags().String("prefix", "", "Block prefix (format: 1234567800000000/32)")
+	isDBEUIBlockCreationCommand.Flags().Int64("init-counter", 0, "Initial counter (determines first address to be issued from block)")
+	isDBCommand.AddCommand(isDBEUIBlockCreationCommand)
 }
--- a/pkg/identityserver/gormstore/eui_block_store.go
+++ b/pkg/identityserver/gormstore/eui_block_store.go
@@ -73,27 +73,30 @@ func (s *euiStore) incrementApplicationDevEUICounter(ctx context.Context, ids *t
 	return nil
 }
 
+const atomicReadAndUpdateDevEUI = `
+UPDATE "eui_blocks" SET "current_counter" = "current_counter" + 1
+WHERE "start_eui" = (
+	SELECT "start_eui"
+	FROM "eui_blocks"
+	WHERE "current_counter"<="end_counter" AND "type"='dev_eui'
+	LIMIT 1
+) RETURNING "start_eui","current_counter"`
+
 func (s *euiStore) issueDevEUIAddressFromBlock(ctx context.Context) (*types.EUI64, error) {
-	var devEUIResult types.EUI64
-	var euiBlock EUIBlock
+	var euiResult []struct {
+		StartEui       EUI64
+		CurrentCounter int64
+	}
 	for {
-		// Get current value from DevEUI db.
-		query := s.query(ctx, EUIBlock{}).Where(EUIBlock{Type: "dev_eui"})
-		if err := query.First(&euiBlock).Error; err != nil {
+		// Atomically update valid block and return the values.
+		if err := s.query(ctx, EUIBlock{}).Raw(atomicReadAndUpdateDevEUI).Scan(&euiResult).Error; err != nil {
 			return nil, err
 		}
-		// Check if DevEUI block limit exists.
-		result := s.query(ctx, EUIBlock{}).
-			Where(`"eui_blocks"."type"='dev_eui' AND "eui_blocks"."current_counter" <= "eui_blocks"."end_counter"`).
-			Update("current_counter", gorm.Expr("current_counter + ?", 1))
-		if err := result.Error; err != nil {
-			return nil, err
-		}
-		if result.RowsAffected == 0 {
+		if len(euiResult) == 0 {
 			return nil, errMaxGlobalEUILimitReached.New()
 		}
-
-		devEUIResult.UnmarshalNumber(euiBlock.StartEUI.toPB().MarshalNumber() | uint64(euiBlock.CurrentCounter))
+		var devEUIResult types.EUI64
+		devEUIResult.UnmarshalNumber(euiResult[0].StartEui.toPB().MarshalNumber() | uint64(euiResult[0].CurrentCounter-1))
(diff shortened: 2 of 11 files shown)


pkg/identityserver/gormstore/migrations/eui_blocks.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Apply implements Migration.
+func (m EUIBlocks) Apply(ctx context.Context, db *gorm.DB) error {
+	db.Model(store.EUIBlock{}).RemoveIndex("eui_block_index")
+	return nil
+}
+
+// Rollback implements migration. For the EUIBlocks migration this is a no-op.
+func (m EUIBlocks) Rollback(ctx context.Context, db *gorm.DB) error {
+	return nil
```
htdvisser: Why no rollback? Shouldn't the rollback re-create the index?

pkg/identityserver/gormstore/migrations/eui_blocks.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"context"
+
+	"github.com/jinzhu/gorm"
+	store "go.thethings.network/lorawan-stack/v3/pkg/identityserver/gormstore"
+)
+
+// EUIBlocks removes the unique index constraint from `type` field.
+type EUIBlocks struct{}
```
htdvisser: I don't think this is descriptive enough. Maybe good to add "Index".

pkg/identityserver/gormstore/migrations/eui_blocks.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Name implements Migration.
+func (EUIBlocks) Name() string {
+	return "eui_blocks"
```
htdvisser:
```suggestion
	return "eui_blocks_index"
```

pkg/identityserver/identityserver.go
```diff
@@ -172,12 +175,11 @@ func New(c *component.Component, config *Config) (is *IdentityServer, err error)
...
-			err = gormstore.GetEUIStore(db).CreateEUIBlock(is.Context(), "dev_eui", is.config.DevEUIBlock.Prefix, is.config.DevEUIBlock.InitCounter)
-			if err != nil {
-				return err
-			}
-			return nil
+		if is.config.DevEUIBlock.Prefix.IsZero() {
+			return nil, errNoDevEUIPrefix.New()
+		}
+		err := is.withDatabase(is.Context(), func(db *gorm.DB) error {
+			return gormstore.GetEUIStore(db).InitializeDevEUIBlock(is.Context(), is.config.DevEUIBlock.Prefix, is.config.DevEUIBlock.InitCounter)
```
htdvisser: Maybe we should reconsider initializing the block in the database on _every_ IS instance start

pkg/identityserver/storetest/eui_store.go
```diff
@@ -83,24 +83,18 @@ func (st *StoreTest) TestEUIStore(t *T) {
...
-		// TODO: See CreateEUIBlock_Extra.
-		t.Skip("See CreateEUIBlock_Extra")
-		// a, ctx := test.New(t)
-		// eui, err := s.IssueDevEUIForApplication(ctx, app1.GetIds(), 3)
-		// if a.So(err, should.BeNil) && a.So(eui, should.NotBeNil) {
-		// 	a.So(*eui, should.Equal, types.EUI64{1, 1, 1, 1, 1, 1, 1, 0})
-		// }
+	t.Run("IssueDevEUIForApplication_Other", func(t *T) {
+		a, ctx := test.New(t)
+		_, err := s.IssueDevEUIForApplication(ctx, app1.GetIds(), 3)
```
htdvisser: Shouldn't we verify that this issues the expected EUI?

pkg/identityserver/identityserver.go
pgalic96: since configuring new blocks will happen extremely rarely, my suggestion is we move it to `is-db` command. `is-db create-block-from-config` or just move the block config from config into cli flags and have a enable/disable switch in config (not sure how our compatibility commitment works in this case).

pgalic96: Check https://github.com/TheThingsNetwork/lorawan-stack/pull/5199/commits/05d485f62738a3cc3ac13c2255f8473016dfe77e

cmd/ttn-lw-stack/commands/is_db.go
```diff
@@ -291,6 +292,56 @@ var (
 			return nil
 		},
 	}
+	isDBEUIBlockCreationCommand = &cobra.Command{
+		Use:   "create-eui-block",
+		Short: "Create an eui block in IS db (currently only DevEUI block supported)",
```
adriansmares: `EUI block`

pkg/identityserver/gormstore/eui_block_store.go
```diff
@@ -78,22 +78,15 @@ func (s *euiStore) issueDevEUIAddressFromBlock(ctx context.Context) (*types.EUI6
...
-			Update("current_counter", gorm.Expr("current_counter + ?", 1))
-		if err := result.Error; err != nil {
-			return nil, err
-		}
-		if result.RowsAffected == 0 {
 			return nil, errMaxGlobalEUILimitReached.New()
 		}
-
-		devEUIResult.UnmarshalNumber(euiBlock.StartEUI.toPB().MarshalNumber() | uint64(euiBlock.CurrentCounter))
+		euiBlock.CurrentCounter++
```
adriansmares: This is not atomic. Why is the previous `Update` approach no longer used ?

adriansmares (commented):
I'm missing the migration - did you remove that accidentally in the last commit ?

pkg/identityserver/gormstore/eui_block_store.go
htdvisser: Well, we both read and write in a single transaction, so in the worst case, the transaction just fails. In a future implementation (with Gorm v2 or a handwritten query) we can restore the atomic approach.

pgalic96: Yeah in current implementation this is still locking the whole table. Once we update to Gorm v2 we could easily use `FOR UPDATE` (https://gorm.io/docs/advanced_query.html#Locking-FOR-UPDATE) and just lock the specific row of the table. This is more relevant for the application rows since there will be more concurrent queries on the same table.

adriansmares:
```
diff --git a/pkg/identityserver/storetest/eui_store.go b/pkg/identityserver/storetest/eui_store.go
index d3c6f201d..cb0859279 100644
--- a/pkg/identityserver/storetest/eui_store.go
+++ b/pkg/identityserver/storetest/eui_store.go
@@ -15,6 +15,7 @@
 package storetest
 
 import (
+       "sync"
        . "testing"
 
@@ -100,4 +101,26 @@ func (st *StoreTest) TestEUIStore(t *T) {
                }
                a.So(err, should.BeNil)
        })
+
+       t.Run("IsThisSafe", func(t *T) {
+               a, ctx := test.New(t)
+
+               var wg sync.WaitGroup
+               var used sync.Map
+               for i := 0; i < 32; i++ {
+                       wg.Add(1)
+
+                       go func() {
+                               defer wg.Done()
+                               eui, err := s.IssueDevEUIForApplication(ctx, app1.GetIds(), 100)
+                               if a.So(err, should.BeNil) && a.So(eui, should.NotBeNil) {
+                                       euiStr := eui.String()
+                                       _, loaded := used.LoadOrStore(euiStr, true)
+                                       a.So(loaded, should.BeFalse)
+                               }
+                       }()
+               }
+
+               wg.Wait()
+       })
 }
```

This test fails immediately on my machine - am I misunderstanding the expected semantics here ? For me it seems that it is trivial to reuse the counter twice.

Edit: Or is it actually expected that the store method is safe only if the store was created on top of a running transaction ?

pgalic96: I reverted this because indeed it seems it could race.

pkg/identityserver/gormstore/migrations/eui_blocks_index.go
```diff
@@ -0,0 +1,44 @@
...
+func (m EUIBlocksIndex) Apply(ctx context.Context, db *gorm.DB) error {
+	return db.Model(store.EUIBlock{}).RemoveIndex("eui_block_index").Error
+}
+
+// Rollback implements migration. It recreates the removed "eui_block_index".
+func (m EUIBlocksIndex) Rollback(ctx context.Context, db *gorm.DB) error {
+	return db.Model(store.EUIBlock{}).AddUniqueIndex("eui_block_index", "type").Error
+}
+
+func init() {
```
adriansmares: I see now that we do this for the other migration too, but I'm not a huge fan as the order of initialization is undefined. Maybe we should put these in order in https://github.com/TheThingsNetwork/lorawan-stack/blob/54c6efee2a192950b5ae535cdf2f6a3722775fcf/pkg/identityserver/gormstore/migrations/migration.go#L62

Enterprise would be compatible with this 
https://github.com/TheThingsIndustries/lorawan-stack/blob/e5c42dec7a58ba8c261f05f5129bd03f7775538b/pkg/identityserver/gormstore/migrations/migrations.tti.go#L5

adriansmares (approved):

cmd/ttn-lw-stack/commands/is_db.go
```diff
@@ -291,6 +292,56 @@ var (
...
+			logger.Info("Connecting to Identity Server database...")
+			db, err := gormstore.Open(ctx, config.IS.DatabaseURI)
+			if err != nil {
+				return err
+			}
+			defer db.Close()
+			if !config.IS.DevEUIBlock.Enabled {
+				logger.Error("DevEUIBlock configuration disabled")
+				return nil
+			}
```
htdvisser: How is this relevant? Wouldn't it be possible to first create the block, and then later enable issuing from that block?

pkg/identityserver/gormstore/eui_block_store.go
```diff
@@ -107,41 +108,29 @@ func (s *euiStore) issueDevEUIAddressFromBlock(ctx context.Context) (*types.EUI6
...
-				CurrentCounter: initCounterValue,
-			},
-		).Error
-	} else if gorm.IsRecordNotFoundError(err) {
-		// If no block found, create a new block in the database.
-		return s.query(ctx, EUIBlock{}).Save(
-			&EUIBlock{
+// CreateUIBlock creates the block of appropriate type in IS db.
+func (s *euiStore) CreateEUIBlock(ctx context.Context, prefix types.EUI64Prefix, initCounter int64, euiType string) error {
+	defer trace.StartRegion(ctx, "initialize eui block").End()
```
htdvisser:
```suggestion
	defer trace.StartRegion(ctx, "create eui block").End()
```

pkg/identityserver/identityserver.go
```diff
@@ -142,7 +142,10 @@ func createAccountAppStore(db *gorm.DB) account_store.Interface {
 	}
 }
 
-var errDBNeedsMigration = errors.Define("db_needs_migration", "the database needs to be migrated")
+var (
+	errDBNeedsMigration = errors.Define("db_needs_migration", "the database needs to be migrated")
+	errNoDevEUIPrefix   = errors.Define("dev_eui_block_prefix_missing", "issuing DevEUI's from a block enabled but no prefix configured")
```
htdvisser: Unused?

htdvisser (approved):

PR #5203, opened 2022-02-11
Use FNV-1a for payload hashing

adriansmares:
#### Summary

Partially reverts https://github.com/TheThingsNetwork/lorawan-stack/pull/4574

#### Changes

- Revert back to FNV-1a from SHA3-224
  - At the time when this change was introduced, the line of thought was that perhaps we encounter hash collisions for the payloads. This proved to be false - https://github.com/TheThingsIndustries/lorawan-stack/pull/2925 .
  - This change should raise the uplink throughput per NS instance, as SHA3 is an expensive hash function to use due to its cryptographic properties. Expected raw hashing throughput is 6-7x (per https://github.com/rurban/smhasher , not 100% applicable but gives a raw idea).
- Explicitly implement and check that the `ttnpb.IDStringer` interface is implemented by the storage integration requests
  - This is required for rate limiting per entity

#### Testing

The old hash function has been battle tested for a year before the change, and the interface checks verify that the message explicitly implements the interface.

##### Regressions

N/A.

#### Checklist

### DIFF
--- a/pkg/networkserver/redis/redis.go
+++ b/pkg/networkserver/redis/redis.go
@@ -17,10 +17,10 @@ package redis
 
 import (
 	"encoding/base64"
+	"hash/fnv"
 
 	"go.thethings.network/lorawan-stack/v3/pkg/errors"
 	ttnredis "go.thethings.network/lorawan-stack/v3/pkg/redis"
-	"golang.org/x/crypto/sha3"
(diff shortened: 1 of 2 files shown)


htdvisser (approved):
Looks like unrelated changes, but LGTM

PR #5209, opened 2022-02-14
JavaScript ahead of time compilation support

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack-support/issues/722

#### Changes

- Add support for ahead of time compilation of JavaScript scripts in the JavaScript scripting engine
- Introduce the concept of _compilable_ payload encoder/decoders
  - The JavaScript payload encoder/decoders implement this concept, using the JS engine support
- Device Repository payload encoding / decoding now supports compiling the scripts ahead of time
  - The scripts are compiled upon retrieval, then the compiled program is cached - the parsing cost is paid only once

#### Testing

Unit testing. Reworked the device repository payload encoding / decoding tests.

##### Regressions

This affects the payload encoding / decoding chain, and unit tests and local testing should cover this.

#### Notes for Reviewers

- Why isn't running a script just a compile + run ? Why are there two different paths ?
  - In order to enforce the execution time limit of 100ms of compile + run. Otherwise, the 100ms would apply only to the run part, not the compilation, which would be a regression from the previous behavior.
  - Note that it is impossible to actually interrupt the compilation of a script, even when using `RunString`, because the compilation part of `RunString` is not affected by the fact that the VM was interrupted. Our current semantics are that if the compilation takes more than 100ms, we won't run the resulting compiled script, but the time per uplink still has the compilation time as an upper bound, not 100ms.
- What about JavaScript that is not coming from the Device Repository ?
  - I have a plan here, but it is outside the scope of this issue (timeouts for Device Repository scripts)
  - Basically, if we store the hash for the parameter, we could cache by hash. In this way, we don't have to pay the hashing cost each uplink, because we know the hash and update it in the store. We only need some minimal changes in order to achieve this, such as generating the hash on `Set`, and having a migration that generates the hash for existing entries.

#### Checklist

### DIFF
--- a/pkg/messageprocessors/cayennelpp/cayennelpp.go
+++ b/pkg/messageprocessors/cayennelpp/cayennelpp.go
@@ -32,7 +32,7 @@ type host struct{}
 type decodedMap map[string]interface{}
 
 // New creates and returns a new CayenneLPP payload encoder and decoder.
-func New() messageprocessors.PayloadEncodeDecoder {
+func New() messageprocessors.PayloadEncoderDecoder {
 	return &host{}
 }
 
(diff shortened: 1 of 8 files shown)


KrishnaIyer (approved):
Don't have a lot of knowledge in this area but didn't find anything obvious.

PR #5212, opened 2022-02-15
Strengthen identifiers nil check

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/pull/3083

#### Changes

- Add a `nil` check before `IsZero`, as `IsZero` requires a value receiver, not a pointer receiver, and as such is not usable directly.


htdvisser (approved):
Now that we don't use `gogoproto.nullable=false` anymore, we really need to get rid of those value methods on proto types, and any use of non-pointer protos.

PR #5213, opened 2022-02-15
Update changelog with minimum Redis version change

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/5171

#### Changes

- Mention that Redis 6.2 is now required, as the above mentioned changes use a parameter available from 6.2 onwards

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
+- The minimum required Redis version is now 6.2.
+
 ### Deprecated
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
+- The minimum Redis version required by the stack is now 6.2.
```
htdvisser: I don't think we refer to The Things Stack as "the stack"

htdvisser (approved):

PR #5222, opened 2022-02-17
Fix webhook status UX

### LINKED ISSUE #5150: Improve webhook health status Console UX
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/4526
References https://github.com/TheThingsNetwork/lorawan-stack/pull/5002

The Console should show information in a more contextual manner to the user. Specifically, the webhook should not be shown as 'pending' when the health checks are completely disabled.

#### Why do we need this?

Right now, even though the checks are completely disabled, the webhooks are shown as pending, and users are told that they will move from pending to healthy when the webhook will be executed. This is wrong, as no health check will take place if they are disabled.

The health status is pending in the webhook list, and editing the webhook tells the user that the status will change soon.

I have described the UX that I had in mind at the bottom of [this](https://github.com/TheThingsNetwork/lorawan-stack/issues/4526#issuecomment-988697068) comment, specifically:

> The frontend may provide quite some contextual help.  For example, if `failed_attempts` is smaller than `unhealthy_attempts_threshold`, the Console may convey that the next HTTP request will be attempted, but if it fails the number of attempts will increase.

> If `failed_attempts` is greater or equal to `unhealthy_attempts_threshold`, then the `last_failed_attempt_at` can be used to relay what will happen to the next request. If `last_failed_attempt_at` + `unhealthy_retry_interval` is a moment in the past, the request will be executed, but if it is a moment in the future, the request will be dropped (no attempt will be made).

> It can also be the case that `unhealthy_attempts_threshold` is `0`. This means that the Application Server will be doing strictly book keeping - the number of attempts will increase, the timestamp will be updated, the last error will be stored, but the requests will always be made - basically we're only monitoring the webhook. In such cases I would recommend that the Console conveys a friendly message that tells the users to get their webhook fixed asap or at least delete it if it's not used.

To add to this, if both `unhealthy_attempts_threshold` and `unhealthy_retry_interval` are zero, the checks are completely disabled - we may put status as 'disabled there, and show no extra information in the webhook itself.

#### Environment

`3.17.2`

First of all, backport the AS settings retrieval from enterprise:

https://github.com/TheThingsIndustries/lorawan-stack/blob/0906e8b33239cd4eff2769151fbf22e352a97835/pkg/webui/console/store/middleware/logics/application-server.js

This will allow the UI to have access to the health configuration, and in turn to provide this contextual information.

Edit the stack settings and send some successful/failed webhooks. I've given some advice in my above mentioned comment.

I can review. @kschiffer if the proposed UX looks good, please remove the label.

kschiffer:
#### Summary

This PR fixes UX issues with the new webhook health status in the Console.

Closes #5150 

image

#### Changes

- Change wording to `Healthy`, `Pending` and `Requests failing`
- Check whether webhook health checks are enabled and do not show any status indication when health status is disabled
- Add AS config service to JS SDK

#### Testing

Manual testing on staging environment.

#### Checklist


pkg/webui/console/components/webhook-form/webhook-form.styl
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
+// Copyright © 2019 The Things Network Foundation, The Things Industries B.V.
```
adriansmares: ⌚

adriansmares (approved):

ryaplots (approved):

PR #5229, opened 2022-02-18
Update json decoder to comport array input via stdin

### LINKED ISSUE #5228: Create end-devices in bulk returns an error regarding the decoding
#### Summary

The creation of end-devices in bulk via CLI doesn't work when using the format specified in [the docs](https://www.thethingsindustries.com/docs/getting-started/migrating/device-json/). It does in fact create the end-devices but after processing the input it always return a decoding error.

#### Steps to Reproduce

1. Create an instance of the TTN stack
2. login
3. create an application
4. Import devices in bulk via the cli, using the format described in the documentation

Relevant information:
- The config used was the default one.
- sequence of steps and logs  -> [steps.txt](https://github.com/TheThingsNetwork/lorawan-stack/files/8098080/steps.txt)
- input file used for the creation of devices -> [input.txt](https://github.com/TheThingsNetwork/lorawan-stack/files/8098050/input.txt)

#### What do you see now?

After handling the last device it returns an error regarding an invalid character `]`.
image

No error message at the end of the creation in bulk

#### Environment

image
image

when using the `inputDecoder` [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/75d581c437153a4bcd484b4607f8608aa6d4bec9/cmd/ttn-lw-cli/commands/end_devices.go#L364-L367), Decode the information into a slice of devices.

Manually test it. 
There is the issue https://github.com/TheThingsIndustries/lorawan-stack/issues/3029 that includes the automatic tests of the cli, so there is no need to implement an extra layer of tests besides the manual test.

Yes

nicholaspcr:
#### Summary

Closes #5228 

In short when using cli commands to create end-devices, users etc in bulk, there was an error of invalid character `]`. This happened due to the custom `JSONDecoder` of the `io` pkg, which always excluded the `[`.

#### Changes

- Update `JSONDecoder` to read all the content if need the object passed is an array.

#### Testing

All unit tests pass and manual testing. The second consists of creating users and end-devices by passing a json file with an array of objects and a file with just one object.

##### Regressions

N/A

#### Notes for Reviewers

I wrote the main reasons for this in the issue but the main problem is that running bulk operations in the cli integration tests will always fail. 

#### Checklist

### DIFF
--- a/pkg/util/io/io.go
+++ b/pkg/util/io/io.go
@@ -20,6 +20,7 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"io"
 	"strings"
 
@@ -33,8 +34,9 @@ type Decoder interface {
 }
 
 type jsonDecoder struct {
-	rd  *bufio.Reader
-	dec *json.Decoder
+	rd   *bufio.Reader
+	dec  *json.Decoder
+	objs []json.RawMessage
 }
 
 // NewJSONDecoder returns a new Decoder on top of r, and that uses the common JSON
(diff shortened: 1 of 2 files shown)


pkg/util/io/io.go
```diff
@@ -33,25 +34,55 @@ type Decoder interface {
...
 	}
+
+	if t == '[' {
+		if err := r.rd.UnreadRune(); err != nil {
+			return nil, err
+		}
+		if err := r.dec.Decode(&r.objs); err != nil {
+			return nil, err
+		}
+		fmt.Println(r.objs)
```
johanstokking, line 80: Debugging artifact?

johanstokking (commented):
Okay so this is a nice hack but this recursiveness isn't really easy to read.

Actually I don't think we did allow JSON arrays to be passed in. I think that the documentation is wrong.

pkg/util/io/io.go
nicholaspcr: Indeed it is, will remove it

nicholaspcr:
> Okay so this is a nice hack but this recursiveness isn't really easy to read.
> 
> Actually I don't think we did allow JSON arrays to be passed in. I think that the documentation is wrong.

If it were to allow json arrays I can try to make it more readable, maybe separate the queue of raw messages from the struct or create methods to operate on top of it. 

Outside the readability, it might have been me misinterpreting the documentation. Take the [end-device](https://www.thethingsindustries.com/docs/getting-started/migrating/device-json/) documentation, initially I was thinking it might just apply for the console but when looking into the `asBulk` implementation it does seem to have some sort of intentionality in regards to processing multiple elements in a single cli command. If that is not the case, then I'll just remove the test case from the cli integration tests in the enterprise repository.

adriansmares:
Indeed, the input is meant to be a JSON stream (JSON objects delimited by white space). I think fixing the docs would be better, although I do agree that it is annoying that we cannot pipe an array into `asBulk`.

nicholaspcr:
With the discussion in mind I'll close this PR and its corresponding issue. 
The documentation PR can be found here https://github.com/TheThingsIndustries/lorawan-stack-docs/pull/783

PR #5241, opened 2022-02-22
Publish encoded and encrypted downlink in forwarded event

johanstokking:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack-docs/issues/710#issuecomment-1042783544

#### Changes

The downlink forwarded event now contains the FPort and FRMPayload output from the encoding and encryption steps.

#### Testing

Tested locally:

Screen Shot 2022-02-22 at 16 16 50

In the second instance (here on top), I scheduled a downlink message that was encoded by the encoder. It appears that the JSON is not shown in the Console.

##### Regressions

If applications rely on the content of the `as.down.data.forward` payload or the contents of the `DownlinkQueued` payload, they may break. However, I think that it is acceptable that what AS forwards to NS is encrypted and encoded.

#### Checklist

### DIFF
--- a/pkg/applicationserver/applicationserver.go
+++ b/pkg/applicationserver/applicationserver.go
@@ -499,6 +499,7 @@ type downlinkQueueOperation struct {
 	Items             []*ttnpb.ApplicationDownlink
 	Operation         func(ttnpb.AsNsClient, context.Context, *ttnpb.DownlinkQueueRequest, ...grpc.CallOption) (*pbtypes.Empty, error)
 	SkipSessionKeyIDs [][]byte
+	ResultFunc        func(decrypted, encrypted []*ttnpb.ApplicationDownlink, err error)
 }
 
 func (d downlinkQueueOperation) shouldSkip(sessionKeyID []byte) bool {
@@ -512,12 +513,19 @@ func (d downlinkQueueOperation) shouldSkip(sessionKeyID []byte) bool {
 
 const maxDownlinkQueueOperationAttempts = 50
 
-func (as *ApplicationServer) attemptDownlinkQueueOp(ctx context.Context, dev *ttnpb.EndDevice, link *ttnpb.ApplicationLink, peer cluster.Peer, op downlinkQueueOperation) ([]string, error) {
+func (as *ApplicationServer) attemptDownlinkQueueOp(ctx context.Context, dev *ttnpb.EndDevice, link *ttnpb.ApplicationLink, peer cluster.Peer, op downlinkQueueOperation) (mask []string, err error) {
+	var encryptedItems []*ttnpb.ApplicationDownlink
+	if op.ResultFunc != nil {
+		defer func() {
+			op.ResultFunc(op.Items, encryptedItems, err)
+		}()
+	}
+
 	pc, err := peer.Conn()
 	if err != nil {
 		return nil, err
 	}
-	mask := make([]string, 0, 2)
+	mask = make([]string, 0, 2)
 	attempt := 1
 	for {
 		ctx := log.NewContextWithField(ctx, "attempt", attempt)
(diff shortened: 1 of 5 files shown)


johanstokking:
Hmm, maybe the `DownlinkQueued` should contain the decoded and decrypted downlink?

If so, we should do that really as a separate step. That is necessary because some applications may publish binary `frm_payload`, while there is a downlink decoder present that can and should still represent the downlink as object.

pkg/applicationserver/applicationserver.go
```diff
@@ -512,35 +513,46 @@ func (d downlinkQueueOperation) shouldSkip(sessionKeyID []byte) bool {
...
-func (as *ApplicationServer) attemptDownlinkQueueOp(ctx context.Context, dev *ttnpb.EndDevice, link *ttnpb.ApplicationLink, peer cluster.Peer, op downlinkQueueOperation) ([]string, error) {
+func (as *ApplicationServer) attemptDownlinkQueueOp(ctx context.Context, dev *ttnpb.EndDevice, link *ttnpb.ApplicationLink, peer cluster.Peer, op downlinkQueueOperation) (mask []string, err error) {
+	resultCallbackItems := op.Items
+	if op.ResultFunc != nil {
+		defer func() {
+			op.ResultFunc(resultCallbackItems, err)
+		}()
+	}
+
+	sessions := make([]*ttnpb.Session, 0, 2)
```
adriansmares: This was in the loop because the sessions may be updated by `as.buildSessionsFromError`

Also you cannot encrypt ahead of time, since the frame counter may have advanced during session recovery.

pkg/applicationserver/applicationserver.go
```diff
@@ -512,35 +513,46 @@ func (d downlinkQueueOperation) shouldSkip(sessionKeyID []byte) bool {
...
+	}
+
+	encryptedItems, err := as.encodeAndEncryptDownlinks(ctx, dev, link, op.Items, sessions)
+	if err != nil {
+		return nil, err
+	}
+	// When encoding and encryption succeeds, pass the encoded and encrypted items to the result callback. This is useful
+	// for diagnostics as error or successful forwarding events contains the actual FRMPayload and FPort and not the
+	// original input.
+	resultCallbackItems = encryptedItems
```
adriansmares: Are you really sure this does what it says ? `encodeAndEncryptDownlinks` erases the decoded payload.

adriansmares:
> Hmm, maybe the `DownlinkQueued` should contain the decoded and decrypted downlink?

Yes, and the encoding should occur only once. Right now we re-encode multiple times which is not really efficient.

johanstokking:
OK, so we need to write encrypted payload in the downlink forwarded event, and decrypted payload in the downlink queued upstream message.

Also, we need to account for rewriting the FPort by the encoder. If we keep the decrypted payload in the downlink queued upstream message as is, we don't have the right FPort, because it may not have been part of the scheduled message.

This is now addressed in https://github.com/TheThingsNetwork/lorawan-stack/pull/5241/commits/de79d905c79a7d0625297b194d1e7c8e0f8d3330

johanstokking:
So when scheduling a JSON object with FPort 1, and an encoder that sets FPort to 42, this is the event in the Console:

Screen Shot 2022-02-22 at 17 16 10

So that is encrypted FRMPayload. The FPort of the received message is already rewritten.

And this is the published downlink queued upstream message:

```json
{
  "end_device_ids": {
    "device_id": "dev1",
    "application_ids": {
      "application_id": "app1"
    }
  },
  "correlation_ids": [
    "as:downlink:01FWH315JWAQ1210VD92H28S30",
    "rpc:/ttn.lorawan.v3.AppAs/DownlinkQueueReplace:01FWH315J3TA6SJWAC11903TSM"
  ],
  "downlink_queued": {
    "f_port": 42,
    "decoded_payload": {
      "color": "blue"
    },
    "correlation_ids": [
      "as:downlink:01FWH315JWAQ1210VD92H28S30",
      "rpc:/ttn.lorawan.v3.AppAs/DownlinkQueueReplace:01FWH315J3TA6SJWAC11903TSM"
    ]
  }
}
```

adriansmares (commented):
This looks good now, but we still don't have the `frm_payload` in the downlink enqueued event, right ? Since the generate payload is part of `encryptedItem` and that one is encrypted after.

I still think that perhaps we should break down the encode-and-encrypt into encode once and encrypt possibly many times.

adriansmares (approved):

johanstokking:
So now it's even better; the downlink queued upstream message now looks like this:

```json
{
  "end_device_ids": {
    "device_id": "dev1",
    "application_ids": {
      "application_id": "app1"
    }
  },
  ...
  "downlink_queued": {
    "f_port": 42,
    "frm_payload": "AQ==",
    "decoded_payload": {
      "color": "blue"
    },
    ...
  }
}
```

And the Console shows the encoded and encrypted FRMPayload separately:

Screen Shot 2022-02-23 at 10 06 20

adriansmares:
Indeed, this is exactly what I wanted to see.

johanstokking:
This needs a changelog entry but the target branch isn't updated with the `v3.18.0` changelog yet, so I'm holding the merge.

PR #5251, opened 2022-02-23
Fix xtime sent to LBS gateways for Class C downlinks

KrishnaIyer:
#### Summary

Refs: https://github.com/TheThingsIndustries/lorawan-stack-support/issues/714#issuecomment-1048549330

#### Changes

- Introduce a new `concentrator_timestamp` field in the TxSettings message.
- Set the concentrator timestamp of the computed emission and use that to calculate `xtime` for LBS gateways.
- XTime now is fixed to actually send 48 bits of timestamp instead of 32.
- Add a mock clock for testing.

#### Testing

- [x] UT
- [x] Gateway tests

##### Regressions

- [x] Verify that Class A downlinks work.
- [ ] Verify that 72 min (32 bit) rollover works properly.

#### Checklist


pkg/gatewayserver/io/ws/ws.go
```diff
@@ -304,7 +305,16 @@ func (s *srv) handleTraffic(w http.ResponseWriter, r *http.Request) (err error)
 					return err
 				}
 			case down := <-conn.Down():
-				concentratorTime, ok := conn.TimeFromTimestampTime(down.GetScheduled().Timestamp)
```
johanstokking, line 307: Okay this was indeed the issue. `Timestamp` is not set for absolute timed downlink.

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -304,7 +305,16 @@ func (s *srv) handleTraffic(w http.ResponseWriter, r *http.Request) (err error)
...
+				var (
+					concentratorTime scheduling.ConcentratorTime
+					ok               bool
+				)
+				// If absolute time is set, use that for computing xtime. If not, use the concentrator timestamp.
+				if down.GetScheduled().Time != nil {
+					concentratorTime, ok = conn.TimeFromGatewayTime(*ttnpb.StdTime(down.GetScheduled().Time))
```
adriansmares: I'm not sure this is correct. Keep in mind that `ScheduleAt` is not just converting the gateway time, but rather allowing some server time leeway too. 

I think the correct way of handling this is to propagate the `ConcentratorTime` from the `ScheduleAt`/`ScheduleAnytime` instead of reconstructing it in the frontends. 

It is lossy otherwise, because the timestamp does not account for rollovers, and you need to basically construct the logic in `ScheduleAt` if you want to account for everything.

KrishnaIyer: I think you're right but this is more generic than LBS. We do the same for UDP as well; https://github.com/TheThingsNetwork/lorawan-stack/blob/2d719336fa53d29019e7acee419d5893024a9a09/pkg/ttnpb/udp/translation.go#L396-L401

The bug is here; https://github.com/TheThingsNetwork/lorawan-stack/blob/2d719336fa53d29019e7acee419d5893024a9a09/pkg/gatewayserver/io/io.go#L599-L604. If it's `timestamp` based, we replace the NS timestamp with the emission but for absolute time, we just forward it as is.

KrishnaIyer: So while we can indeed propagate the `ConcentratorTime` to the frontend, that works only for LBS. For UDP, we need the abs time there.

My suggestion would be then to offset the `settings.Time` with the calculated emission value in `io.go`. What do you think?

adriansmares: In UDP it is actually possible to use the absolute time (`scheduled.Time`). Unless we actually implement class B scheduling, this is not an option for LBS - we need to convert either time, or timestamp, to a concentrator time.

If we convert time, we need to reinvent the logic of `ScheduleAt`.
If we convert timestamp (which we don't even have due to `settings.Timestamp = 0`), we're doing a lossy conversion.

Edit: 
> My suggestion would be then to offset the settings.Time with the calculated emission value in io.go. What do you think?

But then you're affecting UDP too, right ?
While I'm proposing that we're adding a concentrator time field to the transmission settings and always put the concentrator time there. The other frontends can ignore this field, while LBS uses it to lossely create the `XTime`.

johanstokking: How sure are we that gateways have gateway time? From what I've seen they do, either via NTP or via time transfer.

The scheduler determines the concentrator time by gateway time otherwise by server time: https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.18/pkg/gatewayserver/scheduling/scheduler.go#L310-L321. That concentrator time is only used for duty-cycle enforcement and conflict detection.

What we really want to have here is the same concentrator time that the scheduler used too. With relative time scheduling, this is already in `Timestamp`. We explicitly don't set `Timestamp` with absolute time scheduling, see https://github.com/TheThingsNetwork/lorawan-stack/pull/1889/files#diff-9c87734043add0ade22bca71f2a6641c1d37830e1aae533fbc6066899c479d1b. Even if we had `Timestamp`, it's not the full concentrator time that we need here.

We cannot extend `TxSettings` with a `ConcentratorTime` field because downlink may not be scheduled by the Gateway Server (but by Packet Broker Agent).

So yeah, I don't know either what's the best solution here. I'm not against this approach, but there can be a discrepancy between the concentrator time here and the concentrator time that the scheduler determined. That may lead to a conflict or duty-cycle enforcement that is observed at the gateway. The only way to avoid that would be to extract that time determination logic from the scheduler to a method that can be called here.

adriansmares: > We cannot extend `TxSettings` with a `ConcentratorTime` field because downlink may not be scheduled by the Gateway Server (but by Packet Broker Agent).

We could always fallback, right ? Use the concentrator time if available, otherwise check time and use the common function for deduction, otherwise fallback to timestamp.

Edit: Actually wait, I don't see how PBA is relevant here. PBA goes through the same scheduler which must have the concentrator time available, the downlinks are not just pushed to the frontends.

The flow is NS/PBA -> `SchduleDownlink` -> `ScheduleAt` which has the concentrator timestamp -> frontend.

KrishnaIyer: Right and by `ConcentratorTime`, you mean the one from the emission right? I see how that could work actually.

adriansmares: > Right and by `ConcentratorTime`, you mean the one from the emission right? I see how that could work actually.

Indeed. That's exactly the time the scheduler expects the emission to occur.

johanstokking: Ah right. Yes indeed that would mean we can put the concentrator time in `TxSettings`.

pkg/gatewayserver/io/ws/lbslns/time.go
```diff
@@ -128,11 +128,13 @@ func TimePtrToUpInfoTime(t *time.Time) (float64, int64) {
 // ConcentratorTimeToXTime contructs the XTime associated with the provided
 // session ID and concentrator timestamp.
 func ConcentratorTimeToXTime(id int32, t scheduling.ConcentratorTime) int64 {
-	return int64(id)<<48 | (int64(t) / int64(time.Microsecond) & 0xFFFFFFFFFF)
+	return int64(id)<<48 | (int64(t) / int64(time.Microsecond) & 0xFFFFFFFFFFFF)
```
KrishnaIyer: This makes the `xtime` now correctly have upto 48 bits of timestamp.

pkg/gatewayserver/io/io.go
```diff
@@ -602,6 +602,7 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
 		} else {
 			settings.Timestamp = 0
 		}
+		settings.ConcentratorTimestamp = uint64(time.Duration(em.Starts()))
```
KrishnaIyer: We don't need to convert this to microseconds since the LBS frontend will cast it to `scheduling.ConcentratorTime` and that's nanoseconds.

pkg/gatewayserver/io/ws/ws_test.go
```diff
@@ -1092,7 +1119,46 @@ func TestTraffic(t *testing.T) {
...
+							t.Fatalf("Invalid RawPayload: %v", up.Message.RawPayload)
+						}
+						up.Message.RawPayload = nil
+						up.Message.RxMetadata[0].UplinkToken = nil
+						expectedUp := tc.ExpectedNetworkUpstream.(ttnpb.UplinkMessage)
+
+						// Set the correct xtime and timestamps for the assertion.
+						expectedUp.RxMetadata[0].Timestamp = timestamp
+						expectedUp.RxMetadata[0].Time = ttnpb.ProtoTime(&now)
+						expectedUp.RxMetadata[0].Time.Nanos = 0
```
KrishnaIyer: The nano second value is not used by the GS but is added in the expected message. I don't know if there is a more beautiful way to do this.

pkg/gatewayserver/io/ws/ws_test.go
```diff
@@ -1186,8 +1291,19 @@ func TestTraffic(t *testing.T) {
...
-						msg.MuxTime = tc.ExpectedBSDownstream.(lbslns.DownlinkMessage).MuxTime
-						if !a.So(msg, should.Resemble, tc.ExpectedBSDownstream.(lbslns.DownlinkMessage)) {
+						expected := tc.ExpectedBSDownstream.(lbslns.DownlinkMessage)
+						msg.MuxTime = expected.MuxTime
+						if dlClass == ttnpb.Class_CLASS_A {
+							expected.XTime = clock.GetXTimeForTimestamp(timeStamp)
+						} else {
+							expected.XTime = clock.GetXTimeForTime(dlTime)
+							diff := expected.XTime - msg.XTime
+							if !a.So(diff, should.BeGreaterThanOrEqualTo, 0) && !a.So(diff, should.BeLessThan, 20000000) { //  Tolerance of 2s
```
KrishnaIyer: I ran some tests locally and this diff is anywhere between 0.2s to 1.2s. Since this whole PR is meant to essentially test this one value, I can't leave it out. But I can't guarantee that this won't be flaky with the CI.
I think the only way is to see how CI works and extend the tolerance if necessary.

api/lorawan.proto
```diff
@@ -315,6 +315,10 @@ message TxSettings {
   google.protobuf.Timestamp time = 7;
   // Transmission settings for downlink.
   Downlink downlink = 8;
+  // Concentrator timestamp for the downlink as calculated by the Gateway Server scheduler.
+  // This value takes into account necessary offsets such as the RTT (Round Trip Time) and TOA (Time Of Arrival).
+  // This field is set and used only by the Gateway Server.
+  uint64 concentrator_timestamp = 9;
```
adriansmares: Why unsigned ? The type definition for `ConcentratorTime` is signed.

pkg/gatewayserver/io/io.go
adriansmares: You can use `int64(em.Starts())` directly then, right ?

pkg/gatewayserver/io/ws/lbslns/time.go
adriansmares: Ouch, ouch, ouch.

This happens in multiple places - please patch those (either now or in a separate PR) too.
https://github.com/TheThingsIndustries/lorawan-stack/search?q=0xFFFFFFFFFF

pkg/gatewayserver/io/ws/ws_test.go
```diff
@@ -1092,7 +1119,46 @@ func TestTraffic(t *testing.T) {
...
+						expectedUp.Settings.Timestamp = timestamp
+						expectedUp.Settings.Time = ttnpb.ProtoTime(&now)
+						expectedUp.Settings.Time.Nanos = 0
+
+						a.So(up.Message, should.Resemble, &expectedUp)
+					case <-time.After(timeout):
+						t.Fatalf("Read message timeout")
+					}
+				case lbslns.JoinRequest:
+					now := time.Now()
```
adriansmares: Maybe erase the nanos from `now` here (divide by nano then multiply) ? Then you don't need to do it in other places.

pkg/gatewayserver/io/ws/ws_util_test.go
```diff
@@ -38,6 +39,38 @@ import (
...
+
+// GetTimestamp returns the current timestamp.
+func (c *mockClock) GetTimestamp() uint32 {
+	return uint32(int64(c.startTimeStamp) + time.Since(c.startTime).Microseconds())
+}
+
+// GetXTime returns the timestamp in the LoRa Basics Station `xtime` format for the given time.
+func (c *mockClock) GetXTimeForTime(t time.Time) int64 {
+	ts := uint32(int64(c.startTimeStamp) + t.Sub(c.startTime).Microseconds())
+	return int64(c.sessionID)<<48 | int64(ts)&0xFFFFFFFFFFFF
```
adriansmares: Maybe the helpers for time conversion should be moved such that they are accessible here. We're copy pasting too much these formulas and if we're some `FF` off it is bad.

pkg/gatewayserver/io/ws/ws_test.go
adriansmares: Where is the difference coming from at all ?

adriansmares (commented):
Nice ! This is what I had in mind. Added some comments/questions.

api/lorawan.proto
KrishnaIyer: For some reason I thought protobuf doesn't support signed ints. Fixed now.

pkg/gatewayserver/io/ws/ws_test.go
KrishnaIyer: This was quite easy to do in the end.

pkg/gatewayserver/io/ws/lbslns/time.go
KrishnaIyer: I've fixed the other one here. I'll fix Tabs Hubs when I work on the merge PR.

pkg/gatewayserver/io/ws/ws_test.go
KrishnaIyer: I thought it was the TOA but that's pretty constant. I'll investigate this.

pkg/gatewayserver/io/ws/lbslns/time.go
johanstokking: Damn this is a nasty one indeed. So after 12 hours of connectivity downlink breaks ...

pkg/gatewayserver/io/ws/ws_test.go
KrishnaIyer: So after some checks, the issue here seems to be that the mock gateway's concentrator timestamp and the concentrator timestamp of the GS are slightly out of sync each time and I'm unable to figure out why. Their absolute times are correct and the GS correctly calculates the xTime for Class A. I'll look at this fresh on Monday but the rest is good to go.

htdvisser (approved):
Code-owned (API) lgtm.

pkg/gatewayserver/io/ws/ws_util_test.go
KrishnaIyer: Yeah so I can't just copy the helpers. This mock uses a microsecond timestamp that simulates a gateway and `scheduling.ConcentratorTime` is in nanoseconds. I can move it to the helper package but this is too specific to the test code that I prefer to leave this here.

adriansmares: You can move the constant (`0xFFFFFFFFFFFF`).

adriansmares (approved):
Very good job.

pkg/gatewayserver/io/ws/ws_test.go
KrishnaIyer: FTR: The flake (which exposed an underlying bug) was caused by the fact that `RxTime` does not use microsecond precision but the GS scheduler does.  So we now use `GatewayTime` only if there's GPSTime https://github.com/TheThingsNetwork/lorawan-stack/pull/5251/commits/c2106c5f20d88d391a5182985cf9b9d7ce434867.

PR #5265, opened 2022-02-28
Add Redis connection pool idle timeout

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/commit/1c520263e8680754ab26b145201572c109a48fac ( the original change )

#### Changes

- Allow the Redis connection pool idle timeout to be configured.
- Increase the default maximum size of the connection pool.

#### Testing

Local, `staging1`, and another one.

##### Regressions

Deployments may see more connections per component, but that's not necessarily a bad thing, as the dial operation can be expensive.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ For details about compatibility between different releases, see the **Commitment
 - Add HTTP basic authentication configuration to the webhooks form in the Console.
 - Show repository formatter code in the payload formatter form in the Console and allow pasting the application and payload formatter code when using the JavaScript option.
 - gRPC service to Gateway Configuration Server so that gateway configurations can be obtained via gRPC requests.
-
+- The option to configure the Redis idle connection pool timeout, using the `redis.idle-timeout` setting.
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ For details about compatibility between different releases, see the **Commitment
 - Add HTTP basic authentication configuration to the webhooks form in the Console.
 - Show repository formatter code in the payload formatter form in the Console and allow pasting the application and payload formatter code when using the JavaScript option.
 - gRPC service to Gateway Configuration Server so that gateway configurations can be obtained via gRPC requests.
-
+- The option to configure the Redis idle connection pool timeout, using the `--redis.idle-timeout` setting.
```
htdvisser: Nit: we (used to) write settings without the `--`.

cmd/internal/shared/config.go
```diff
@@ -93,6 +93,8 @@ var DefaultRedisConfig = redis.Config{
 	Address:       "localhost:6379",
 	Database:      0,
 	RootNamespace: []string{"ttn", "v3"},
+	PoolSize:      1024,
+	IdleTimeout:   time.Hour + 5*time.Minute,
```
htdvisser: These defaults of 1024 connections and 1h5m seem a bit arbitrary. Why don't we use the existing defaults (in order to not change anything for existing deployments) and add configurations to specific deployments where it makes sense?

htdvisser (approved):

PR #5267, opened 2022-02-28
NS matching optimizations

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/commit/f84dd7e5a4e99ab8c50339db0a63b2a9455164ae (original change)
References https://github.com/TheThingsIndustries/lorawan-stack/commit/5b74e49257296f1100f909d5bb32c828eaa4fc1a (original change)

#### Changes

- Deduplicate the uplinks before matching
  - This means that we lose the receive event for the duplicates, but that's perhaps not a bad thing
  - This ensures that a singular instance has to run the matching process, the other instances just store their metadata and return instantly.
- Retrieve the session information in a single round trip
  - Before this change, we would copy the session information keys, then slowly pop them one by one.
  - We now 'download' everything and process it locally. This reduces the amount of round trips massively for large sets.

#### Testing

Local testing, `staging1`, and more.

##### Regressions

This touches one of the most important part of our stack, matching. Extensive testing has been done in order to ensure that we are not regressing (i.e. failing matches that previously would work). Performance wise, the evaluation order (lower frame counter sessions high to low, then highest frame counter sessions high to low) has been kept, so the average number of checks is the same. Metrics confirm this too.

#### Notes for Reviewers

Please take your time to nit pick this one.

#### Checklist


pkg/networkserver/redis/registry.go
```diff
@@ -480,305 +536,110 @@ func (r *DeviceRegistry) RangeByUplinkMatches(ctx context.Context, up *ttnpb.Upl
...
+
+		ses := &UplinkMatchSession{}
+		err = msgpack.Unmarshal([]byte(session.Session), ses)
 		if err != nil {
-			log.FromContext(ctx).WithError(err).Error("Failed to parse match result UID as device identifiers")
-			return errDatabaseCorruption.WithCause(err)
+			continue
+		}
+
+		if uint16(ses.LastFCnt) > lsb {
```
adriansmares, line 597: This condition comes from the 'greater-than' scan script (negated, almost).
https://github.com/TheThingsNetwork/lorawan-stack/blob/e0b838c1f4ec534cb4d1b97190f0da72b22a0e01/pkg/networkserver/redis/lua/deviceMatchScanGT.lua#L25-L27

pkg/networkserver/redis/registry.go
```diff
@@ -480,305 +536,110 @@ func (r *DeviceRegistry) RangeByUplinkMatches(ctx context.Context, up *ttnpb.Upl
...
+		if err != nil {
+			return err
+		}
 	}
-	matchType, ok := vs[0].(string)
-	if !ok {
-		panic(fmt.Sprintf("expected first match script return value element to be a string, got '%v'(%T)", vs[0], vs[0]))
+
+	fillContext := func(ctx context.Context, uid string) (context.Context, *ttnpb.EndDeviceIdentifiers, error) {
+		ids, err := unique.ToDeviceID(uid)
```
johanstokking, line 574: We can do `WithContext` here too right?

pkg/networkserver/redis/registry.go
```diff
@@ -480,305 +536,110 @@ func (r *DeviceRegistry) RangeByUplinkMatches(ctx context.Context, up *ttnpb.Upl
...
+		ctx, ids, err := fillContext(ctx, session.UID)
+		if err != nil {
+			return err
 		}
-		ctx = log.NewContextWithField(ctx, "match_key", matchResultKey)
-		res := &UplinkMatchResult{}
-		if err := msgpack.Unmarshal([]byte(s), res); err != nil {
-			log.FromContext(ctx).WithError(err).Error("Failed to unmarshal match result")
-			return errDatabaseCorruption.WithCause(err)
+		if ids == nil {
```
johanstokking, line 587: When is this happening?

pkg/networkserver/redis/registry.go
```diff
@@ -480,305 +536,110 @@ func (r *DeviceRegistry) RangeByUplinkMatches(ctx context.Context, up *ttnpb.Upl
...
-			panic(fmt.Sprintf("expected match script return value of `result` type to contain 2 elements, got %d", len(vs)))
+	for _, session := range pendingSessionSet {
+		ctx, ids, err := fillContext(ctx, session.UID)
+		if err != nil {
+			return err
 		}
-		s, ok := vs[1].(string)
-		if !ok {
-			panic(fmt.Sprintf("expected second element of match script return value of `result` type to be a string, got '%v'(%T)", vs[1], vs[1]))
+		if ids == nil {
```
johanstokking, line 621: Idem

johanstokking (approved):
Great stuff!

pkg/networkserver/redis/registry.go
adriansmares: This actually comes with a sanity check in Enterprise - https://github.com/TheThingsIndustries/lorawan-stack/blob/d514264344268b5294cf26fa6bdacb454000fd60/pkg/networkserver/redis/registry.go#L596-L607.

pkg/networkserver/redis/registry.go
adriansmares: In Enterprise, when you have traffic coming from tenant `t1`, you don't want to match end devices in tenant `t2` unless the traffic was routed via PB.

https://github.com/TheThingsIndustries/lorawan-stack/blob/afb7ca7dd431b6a8447ef5c1078307efcc1abf27/pkg/networkserver/redis/registry.go#L710-L716

`t1` being the caller (in the context of a gateway), `t2` being the one who did register the possible match.

adriansmares: See https://github.com/TheThingsIndustries/lorawan-stack/blob/afb7ca7dd431b6a8447ef5c1078307efcc1abf27/pkg/networkserver/redis/registry.go#L710-L716 . We basically ensure that we don't match the traffic without the packet broker reflecting it to the instance. The situation when this would happen when two tenants share the same cluster, so both are going to have the session registered under the packet broker tenant. When `t1` sees the traffic, we don't want to match it in `t2` unless PB actually sent it back.

PR #5268, opened 2022-02-28
Debounce stats updates with jitter

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/commit/336777c6fb6ffaa9c1646a01e2ce677a5f84dfae (original change)

#### Changes

- Reintroduce stat updates debouncing, with a different strategy - the debounce period occurs before the write, in order to de-correlate writes.
- Add jitter to the gateway location writes debouncing.

#### Testing

Local testing, `staging1`, and more.

##### Regressions

Stat updates will be slightly delayed, but this is fine.

#### Checklist

### DIFF
--- a/cmd/internal/shared/gatewayserver/config.go
+++ b/cmd/internal/shared/gatewayserver/config.go
@@ -32,6 +32,7 @@ var DefaultGatewayServerConfig = gatewayserver.Config{
 	FetchGatewayInterval:              10 * time.Minute,
 	FetchGatewayJitter:                0.2,
 	UpdateGatewayLocationDebounceTime: time.Hour,
+	UpdateConnectionStatsDebounceTime: 5 * time.Second,
 	ConnectionStatsTTL:                12 * time.Hour,
 	ConnectionStatsDisconnectTTL:      48 * time.Hour,
 	UpdateVersionInfoDelay:            5 * time.Second,
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -906,6 +906,16 @@ func (gs *GatewayServer) updateConnStats(ctx context.Context, conn connectionEnt
 		case <-ctx.Done():
 			return
 		case <-conn.StatsChanged():
+			if duration := gs.config.UpdateConnectionStatsDebounceTime; duration > 0 {
+				// We debounce the updates before the actual write to the store in order
+				// to de-correlate updates over time.
+				duration := random.Jitter(duration, debounceJitter)
+				select {
+				case <-ctx.Done():
+					return
+				case <-time.After(duration):
+				}
+			}
 		case <-refreshTTLTimer.C:
 		}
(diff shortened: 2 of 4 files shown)


htdvisser (commented):
What happens on gateway disconnect? From what I can see, it looks like this could un-set the `disconnected_at` if the gateway disconnects within `DebounceTime` of sending an uplink.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -906,6 +906,16 @@ func (gs *GatewayServer) updateConnStats(ctx context.Context, conn connectionEnt
 		case <-ctx.Done():
 			return
 		case <-conn.StatsChanged():
+			if duration := gs.config.UpdateConnectionStatsDebounceTime; duration > 0 {
+				// We debounce the updates before the actual write to the store in order
+				// to de-correlate updates over time.
+				duration := random.Jitter(duration, 0.25)
```
KrishnaIyer: I think it's better to to export the jitter percentage as a const.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -1000,11 +1010,13 @@ func (gs *GatewayServer) handleLocationUpdates(ctx context.Context, conn connect
 				}
 			}
 
-			timeout := time.After(gs.config.UpdateGatewayLocationDebounceTime)
-			select {
-			case <-ctx.Done():
-				return
-			case <-timeout:
+			if duration := gs.config.UpdateGatewayLocationDebounceTime; duration > 0 {
+				duration := random.Jitter(duration, 0.25)
```
KrishnaIyer: Same here.

cmd/internal/shared/gatewayserver/config.go
```diff
@@ -32,6 +32,7 @@ var DefaultGatewayServerConfig = gatewayserver.Config{
 	FetchGatewayInterval:              10 * time.Minute,
 	FetchGatewayJitter:                0.2,
 	UpdateGatewayLocationDebounceTime: time.Hour,
+	UpdateConnectionStatsDebounceTime: 10 * time.Second,
```
KrishnaIyer: Isn't 10 seconds a lot considering some of the heavy duty gateways?

adriansmares:
> What happens on gateway disconnect? From what I can see, it looks like this could un-set the `disconnected_at` if the gateway disconnects within `DebounceTime` of sending an uplink.

The debounce time is waited in parallel with `context.Done`, and that's the gateway connection context, so I don't see why the write could 'bounce back' the gateway.

cmd/internal/shared/gatewayserver/config.go
adriansmares: At heart that's the point - reducing the writes that result due to these heavy duty gateways. But I think we can live with a 5 too.

KrishnaIyer (approved):

johanstokking (approved):

PR #5290, opened 2022-03-07
Add Device LastSeen

### LINKED ISSUE #4937: Let NS/AS update Last Seen of End Device in IS
#### Summary

It would be nice if the Network Server or Application Server could update the "LastSeen" of end devices in the Identity Server. 

#### Why do we need this?

Mostly because users want to sort their end devices on LastSeen, but it's also nice to have for network statistics.

LastSeen is tracked in the session of the Network Server.

Either the Network Server or the Application Server should update the LastSeen of the end device in the Identity Server

- We can add a `last_seen` to the EndDevice API and DB model.
- Since the Application Server already regularly fetches the end device from the Identity Server, I think it's easiest if the Application Server updated the LastSeen in the Identity Server using Cluster Auth. 
  - Updating should happen asynchronously (so not in hot-path)
  - Granularity should be limited; we shouldn't update on every uplink, perhaps only if the LastSeen is more than X hours ago.

pgalic96:
#### Summary

Closes #4937 

#### Changes

- Add `LastSeen` field to `EndDevice` proto and db model
- Add `BatchUpdateEndDevice` to update end devices in batch
- Add `LastSeenProvider` that stores last seen timestamps and updates them in IS periodically and upon the max capacity of the map

#### Testing

- Unit tests

##### Regressions

- New functionality, shouldn't be any.

#### Notes for Reviewers

Does this need to end up in 3.19? There is a db and api change but its all additions.

#### Checklist


api/end_device.proto
```diff
@@ -570,6 +570,8 @@ message EndDevice {
   // The field cannot be unset once set.
   google.protobuf.Timestamp activated_at = 53;
 
+  google.protobuf.Timestamp last_seen = 54;
```
adriansmares: A short description would help, and this probably should be `last_seen_at` (everything ends in `_at` for timestamps in this proto).

pkg/applicationserver/applicationserver.go
```diff
@@ -748,8 +753,10 @@ func (as *ApplicationServer) handleUp(ctx context.Context, up *ttnpb.Application
 	}
 	switch p := up.Up.(type) {
 	case *ttnpb.ApplicationUp_JoinAccept:
+		as.lastSeenMap.storeLastSeenFromUplink(ctx, up)
```
adriansmares: It's better to do those inside `handle*`, preferably in a `defer` if no error occurred.

pkg/applicationserver/config.go
```diff
@@ -309,3 +310,20 @@ func (c EndDeviceLocationStorageConfig) NewRegistry(ctx context.Context, comp *c
...
+
+// LastSeenConfig defines configuration for the device last seen map which stores timestamps for batch updates.
+type LastSeenConfig struct {
+	MapCapacity int           `name:"map-capacity" description:"Capacity of last seen map"`
+	Interval    time.Duration `name:"flush-interval" description:"Interval at which lat seen timestamps are sent to Identity Server"`
+}
+
+// NewLastSeen defines a new batch update map.
+func (c LastSeenConfig) NewLastSeen(ctx context.Context, comp *component.Component) (*LastSeenProvider, error) {
+	if c.MapCapacity == 0 {
```
adriansmares: I would prefer this to be explicit default configuration (since it is visible when you use `--help` or `config`). Both should error on 0.

pkg/applicationserver/last_seen.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	ls.lastSeenMap[unique.ID(ctx, up.GetEndDeviceIds())] = currentTimestamp
+
+	return ls.batchUpdateEndDeviceLastSeen(ctx, endDevices)
+}
+
+// StoreInMap stores the timestamp in map if its younger than the current timestamp.
+// If map is full it doesn't store the new entry and returns false.
+// This method is not safe.
```
adriansmares: `safe for concurrent use`.

pkg/applicationserver/last_seen.go
```diff
@@ -0,0 +1,161 @@
...
+	ctx     context.Context
+	cluster metadata.ClusterPeerAccess
+	ticker  <-chan time.Time
+
+	mapCapacity int
+	lastSeenMap map[string]time.Time
+	lsMu        sync.Mutex
+}
+
+func (ls *LastSeenProvider) storeLastSeenFromUplink(ctx context.Context, up *ttnpb.ApplicationUp) error {
```
adriansmares: I have to admit that the method visibility here is weird - the methods actually used by the AS are hidden, but the ones which are implementation details are open.

api/end_device.proto
```diff
@@ -595,6 +597,13 @@ message UpdateEndDeviceRequest {
   google.protobuf.FieldMask field_mask = 2;
 }
 
+message BatchUpdateEndDeviceRequest {
```
adriansmares: This is going to be troublesome in multi tenant environments. We either have to provide the tenant identifiers too (cumbersome diff, but can be done), or we have to partition on this on the AS side, which will result in multiple calls (possibly falling back to 1 call per end device).

pkg/applicationserver/last_seen.go
```diff
@@ -0,0 +1,161 @@
...
+		return err
+	}
+	if len(endDevices) == 0 {
+		return nil
+	}
+	return ls.batchUpdateEndDeviceLastSeen(ctx, endDevices)
+}
+
+func (ls *LastSeenProvider) batchUpdateEndDeviceLastSeen(ctx context.Context, endDevices []*ttnpb.EndDevice) error {
+	conn, err := ls.cluster.GetPeerConn(ctx, ttnpb.ClusterRole_ENTITY_REGISTRY, nil)
```
adriansmares: This will stall the uplink hot path with a potentially long duration call. It would be best if the submission here is done asynchronously. You're also holding the lock while doing the call.

pkg/applicationserver/last_seen.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		log.FromContext(ls.ctx).WithError(err).Warn("Failed to get Identity Server peer")
+	}
```
adriansmares: `return err` otherwise you fall through.

pkg/applicationserver/last_seen.go
```diff
@@ -0,0 +1,161 @@
...
+	cl := ttnpb.NewEndDeviceRegistryClient(conn)
+	_, err = cl.BatchUpdate(ctx, &ttnpb.BatchUpdateEndDeviceRequest{
```
adriansmares: `ctx` here also is ideally a tenant-less context (component context perhaps).

pkg/identityserver/end_device_registry.go
```diff
@@ -228,6 +228,37 @@ func (is *IdentityServer) updateEndDevice(ctx context.Context, req *ttnpb.Update
 	return dev, nil
 }
 
+func (is *IdentityServer) batchUpdateEndDevice(ctx context.Context, req *ttnpb.BatchUpdateEndDeviceRequest) (*ttnpb.EndDevices, error) {
+	if err := clusterauth.Authorized(ctx); err != nil {
+		return nil, err
+	}
+	req.FieldMask = cleanFieldMaskPaths([]string{"activated_at", "locations", "last_seen"}, req.FieldMask, nil, getPaths)
```
adriansmares: I would just allow only `last_seen_at` for now, also would simplify the logic below.

pkg/identityserver/gormstore/end_device.go
```diff
@@ -56,6 +56,7 @@ type EndDevice struct {
 	PictureID *string `gorm:"type:UUID;index:end_device_picture_index"`
 
 	ActivatedAt *time.Time `gorm:"default:null"`
+	LastSeen    *time.Time `gorm:"default:null"`
```
adriansmares: `LastSeenAt`

htdvisser: Please extend the tests in `pkg/identityserver/storetest/end_device_store.go` to cover this field.

api/end_device.proto
htdvisser: I think that the logic of per-tenant batching should be in the AS.

api/end_device_services.proto
```diff
@@ -80,6 +80,9 @@ service EndDeviceRegistry {
     };
   };
 
+  // Update a batch of end devices, changing the fields specified by the field mask to the provided values.
+  rpc BatchUpdate(BatchUpdateEndDeviceRequest) returns (EndDevices);
```
htdvisser: Do we really want to introduce an RPC that is able to update many fields of many end devices? Isn't it enough to have an RPC that only updates the LastSeenAt?

pkg/applicationserver/config.go
htdvisser: Or batching should be disabled on `0`.

pkg/identityserver/end_device_registry.go
```diff
@@ -228,6 +228,37 @@ func (is *IdentityServer) updateEndDevice(ctx context.Context, req *ttnpb.Update
...
+				fieldMask = cleanFieldMaskPaths(ttnpb.EndDeviceFieldPathsNested, fieldMask, nil, []string{"activated_at"})
+			}
+			dev, err = st.UpdateEndDevice(ctx, dev, fieldMask.GetPaths())
+			if err != nil {
+				return err
+			}
+			endDevices = append(endDevices, dev)
+		}
+		return nil
+	})
```
htdvisser: This isn't really a batch update. It will just be a large number of individual updates in a single transaction, where each update still requires a round-trip to the database. A real batch update would look more like:

```sql
WITH updates (application_id, device_id, last_seen_at) AS (
  VALUES
    ('foo', 'bar', '2022-03-08T...'),
    ('baz', 'qux', '2022-03-08T...'),
    ...
)
UPDATE end_devices AS end_device
SET last_seen_at = updates.last_seen_at
FROM updates
WHERE end_device.application_id = updates.application_id
AND end_device.device_id = updates.device_id
```

pkg/identityserver/gormstore/end_device_store.go
```diff
@@ -178,6 +179,15 @@ func (s *deviceStore) UpdateEndDevice(ctx context.Context, dev *ttnpb.EndDevice,
...
 	oldAttributes, oldLocations, oldPicture := devModel.Attributes, devModel.Locations, devModel.Picture
+	if ttnpb.HasAnyField(fieldMask, "last_seen") && dev.GetLastSeen() != nil {
+		if timestamp, err := types.TimestampFromProto(dev.GetLastSeen()); err != nil {
+			return nil, err
+		} else {
+			if timestamp.Before(*devModel.LastSeen) {
+				dev.LastSeen, _ = types.TimestampProto(*devModel.LastSeen)
+			}
+		}
+	}
```
htdvisser: Please use `ttnpb.StdTime` and `ttnpb.ProtoTime` so that we have less refactoring to do when we switch away from gogo.

htdvisser (changes requested):
And this indeeds need to go to v3.19 because of DB compatibility.

api/end_device_services.proto
pgalic96: @adriansmares suggested creating a more generic batch update RPC as he had some other use cases for it

adriansmares: Indeed, I've proposed this one for batch location updates and activation statuses.

htdvisser: Okay, but then we can't really write a properly optimized query for it (like in https://github.com/TheThingsNetwork/lorawan-stack/pull/5290#discussion_r821447695) and we can ask ourselves what's the point of batching updates in a single RPC in the first place

adriansmares: I'm ok with specializing the RPC then, and we add subsequent RPCs in the future perhaps. 

Edit: or maybe we should whitelist the allowed field masks and then based on those use an optimized query ? (and allow one field at a time max)

pkg/identityserver/end_device_registry.go
pgalic96: Did it like this: https://github.com/TheThingsNetwork/lorawan-stack/blob/2ef3388754d6aa6009f30da8f59b4d570cd58dd1/pkg/identityserver/gormstore/end_device_store.go#L232-L263

It looks super hacky but I am afraid there is no better way. GORM (especially v1) doesn't really provide a nice way to do this, which is confusing to me because it seems like a common operation. Maybe experimenting with creating a temporary `Table` but I am not sure it's worth it. 

In Gorm V2 there is an option of inserting and then updating on conflict, but this is also something we don't want to do as we don't want to create entries for possibly deleted devices. Also in this way we can add extra restrictions (such as that the last seen timestamp in the db must be older than the one we want to replace it with).

At this point I think we should probably specialize this method for LastSeen updates only, but I left some room for whitelisting as @adriansmares suggested. There should be a way of adding other WHERE clauses as well.

api/end_device_services.proto
pgalic96: see https://github.com/TheThingsNetwork/lorawan-stack/pull/5290#discussion_r822703050

pgalic96:
I believe this is ready for review, tests are failing and I need to add a few more of them but the main functionality is there.

pkg/applicationserver/applicationserver.go
```diff
@@ -786,8 +791,13 @@ var errFetchAppSKey = errors.Define("app_s_key", "failed to get AppSKey")
 // handleJoinAccept handles a join-accept message.
 // If the application or device is not configured to skip application crypto, the InvalidatedDownlinks and the AppSKey
 // in the given join-accept message is reset.
-func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids *ttnpb.EndDeviceIdentifiers, joinAccept *ttnpb.ApplicationJoinAccept, link *ttnpb.ApplicationLink) error {
+func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids *ttnpb.EndDeviceIdentifiers, joinAccept *ttnpb.ApplicationJoinAccept, link *ttnpb.ApplicationLink) (err error) {
 	defer trace.StartRegion(ctx, "handle join accept").End()
+	defer func() {
+		if err == nil {
+			go as.lastSeenMap.storeLastSeenFromUplink(ctx, ids, joinAccept.ReceivedAt)
```
adriansmares: Please use a `workerpool.Workerpool` - look at how the device activations are currently sent to the IS

pkg/applicationserver/config.go
```diff
@@ -106,6 +106,7 @@ type Config struct {
 	Packages                 ApplicationPackagesConfig      `name:"packages" description:"Application packages configuration"`
 	Interop                  InteropConfig                  `name:"interop" description:"Interop client configuration"`
 	DeviceKEKLabel           string                         `name:"device-kek-label" description:"Label of KEK used to encrypt device keys at rest"`
+	LastSeen                 LastSeenConfig                 `name:"device-last-seen" description:"End Device last seen batch update config"`
```
adriansmares: `DeviceLastSeen`.

pkg/applicationserver/last_seen.go
```diff
@@ -0,0 +1,180 @@
...
+			Ids:        ids,
+			LastSeenAt: lastSeen,
+		}); err != nil {
+			log.FromContext(ctx).WithError(err).Warn("Failed to update device last seen timestamp")
+			return
+		}
+	}
+	endDevices, err := ls.StoreInMap(unique.ID(ctx, ids), lastSeen)
+	if err != nil {
+		log.FromContext(ctx).WithError(err).Warn("Failed to update device last seen timestamp")
```
adriansmares: `Failed to store in map device last seen timestamp`.

pkg/applicationserver/last_seen.go
```diff
@@ -0,0 +1,180 @@
...
+	ticker  <-chan time.Time
+
+	mapCapacity int
+	lastSeenMap map[string]time.Time
+	lsMu        sync.Mutex
+}
+
+// StoreInMap stores the timestamp in map if its younger than the current timestamp.
+// If map is full it doesn't store the new entry and returns false.
+func (ls *LastSeenProvider) StoreInMap(ids string, lastSeen *types.Timestamp) ([]*ttnpb.EndDevice, error) {
```
adriansmares: I think that:
- `ids` should be `uid`, and really be a unique ID.
- The return should be a `map[string]*ttnpb.EndDevice` in order to allow for easier partitioning in Enterprise.

pkg/applicationserver/last_seen.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// ClearMap returns an array of end devices with timestamps from the last seen map.
+// This method is not safe for concurrent use.
+func (ls *LastSeenProvider) ClearMap() ([]*ttnpb.EndDevice, error) {
+	if len(ls.lastSeenMap) == 0 {
+		return nil, nil
+	}
+	endDevices := make([]*ttnpb.EndDevice, 0)
+	for devID, timestamp := range ls.lastSeenMap {
```
adriansmares: `devUID`

pkg/applicationserver/last_seen.go
adriansmares: This is still not handled.

pkg/identityserver/gormstore/fieldmask_paths.go
```diff
@@ -39,6 +39,7 @@ const (
 	grantsField                         = "grants"
 	hardwareVersionField                = "version_ids.hardware_version"
 	joinServerAddressField              = "join_server_address"
+	lastSeenField                       = "last_seen_at"
```
adriansmares: Nit: `lastSeenAtField`

pkg/ttnpb/field_mask_validation.go
```diff
@@ -497,6 +499,7 @@ var RPCFieldMaskPaths = map[string]RPCFieldMaskPathValue{
 	"/ttn.lorawan.v3.EndDeviceRegistry/Get":                    {All: EndDeviceFieldPathsNested, Allowed: isEndDeviceReadFieldPaths},
 	"/ttn.lorawan.v3.EndDeviceRegistry/List":                   {All: EndDeviceFieldPathsNested, Allowed: isEndDeviceReadFieldPaths},
 	"/ttn.lorawan.v3.EndDeviceRegistry/Update":                 {All: EndDeviceFieldPathsNested, Allowed: isEndDeviceWriteFieldPaths, Set: true},
+	"/ttn.lorawan.v3.EndDeviceRegistry/BatchUpdate":            {All: EndDeviceFieldPathsNested, Allowed: isEndDeviceWriteFieldPaths, Set: true},
```
adriansmares: The `Allowed` paths here should contain only `last_seen_at`.

adriansmares (commented):
I like the progress - I've added some nits.

Please also open the Enterprise PR with the required changes there before merging this.

pgalic96:
@adriansmares I fixed the feedback, going to start on the enterprise part now

pkg/applicationserver/config.go
```diff
@@ -309,3 +310,14 @@ func (c EndDeviceLocationStorageConfig) NewRegistry(ctx context.Context, comp *c
 	}
 	return registry, nil
 }
+
+// LastSeenConfig defines configuration for the device last seen map which stores timestamps for batch updates.
+type LastSeenConfig struct {
+	MapCapacity int           `name:"map-capacity" description:"Capacity of last seen map"`
+	Interval    time.Duration `name:"flush-interval" description:"Interval at which lat seen timestamps are sent to Identity Server"`
```
adriansmares: Nit: `FlushInterval`.

pkg/applicationserver/config.go
```diff
@@ -309,3 +310,14 @@ func (c EndDeviceLocationStorageConfig) NewRegistry(ctx context.Context, comp *c
...
+}
+
+// NewLastSeen defines a new batch update map.
+func (c LastSeenConfig) NewLastSeen(ctx context.Context, comp *component.Component) (*LastSeenProvider, error) {
+	return NewLastSeen(ctx, c.MapCapacity, time.NewTicker(c.Interval).C, comp)
```
adriansmares: Validate `MaxCapacity`/`Interval` here. `NewTicker(0)` panics for example.

Having both equal to `0` could return a no-op implementation (and you can use an interface).

pkg/applicationserver/last_seen.go
adriansmares: This is still not handled.

pgalic96: ah, I thought I handled all places, sorry

api/end_device.proto
```diff
@@ -570,6 +570,10 @@ message EndDevice {
   // The field cannot be unset once set.
   google.protobuf.Timestamp activated_at = 53;
 
+  // Timestamp when a device uplink has been last observed.
+  // This field is set by the Application Server.
+  google.protobuf.Timestamp last_seen_at = 54;
+
   // next: 54;
```
htdvisser: You've now used 54.

```suggestion
  // next: 55;
```

pkg/applicationserver/config.go
```diff
@@ -309,3 +310,14 @@ func (c EndDeviceLocationStorageConfig) NewRegistry(ctx context.Context, comp *c
 	}
 	return registry, nil
 }
+
+// LastSeenConfig defines configuration for the device last seen map which stores timestamps for batch updates.
+type LastSeenConfig struct {
+	MapCapacity int           `name:"map-capacity" description:"Capacity of last seen map"`
```
htdvisser: I don't think we should expose implementation details (the fact that we use a map) to users. From what I understand this is the maximum batch size, then let's call it that.

pkg/identityserver/end_device_registry.go
```diff
@@ -228,6 +228,33 @@ func (is *IdentityServer) updateEndDevice(ctx context.Context, req *ttnpb.Update
 	return dev, nil
 }
 
+var errMultipleFieldBatchUpdate = errors.DefineInvalidArgument(
+	"multiple_update_fields_specified",
+	"multiple fieldmask paths specified while batch update allows only one field update",
+)
```
htdvisser: What's the reason for not allowing multiple fields to be updated in a batch update?

pkg/identityserver/gormstore/end_device_store.go
```diff
@@ -178,6 +178,12 @@ func (s *deviceStore) UpdateEndDevice(ctx context.Context, dev *ttnpb.EndDevice,
 		return nil, err
 	}
 	oldAttributes, oldLocations, oldPicture := devModel.Attributes, devModel.Locations, devModel.Picture
+	if ttnpb.HasAnyField(fieldMask, "last_seen_at") && dev.GetLastSeenAt() != nil {
+		timestamp := ttnpb.StdTime(dev.GetLastSeenAt())
+		if devModel.LastSeenAt != nil && timestamp.Before(*devModel.LastSeenAt) {
+			dev.LastSeenAt = ttnpb.ProtoTime(devModel.LastSeenAt)
+		}
+	}
```
htdvisser: I don't know if this should be the responsibility of `UpdateEndDevice`. This is the storage layer, not the business layer.

pkg/identityserver/gormstore/end_device_store.go
```diff
@@ -217,3 +223,41 @@ func (s *deviceStore) DeleteEndDevice(ctx context.Context, id *ttnpb.EndDeviceId
 	defer trace.StartRegion(ctx, "delete end device").End()
 	return s.deleteEntity(ctx, id)
 }
+
+var errUnknownField = errors.DefineInvalidArgument(
+	"unknown_field_update",
+	"unknown field update in end device batch update",
+)
+
+func batchUpdateQuery(values []string, field string) string {
```
htdvisser: I hope `values` means `placeholders` here.

pkg/identityserver/store/store_interfaces.go
```diff
@@ -65,6 +65,7 @@ type EndDeviceStore interface {
 	GetEndDevice(ctx context.Context, id *ttnpb.EndDeviceIdentifiers, fieldMask FieldMask) (*ttnpb.EndDevice, error)
 	UpdateEndDevice(ctx context.Context, dev *ttnpb.EndDevice, fieldMask FieldMask) (*ttnpb.EndDevice, error)
 	DeleteEndDevice(ctx context.Context, id *ttnpb.EndDeviceIdentifiers) error
+	BatchUpdateEndDevice(ctx context.Context, devs []*ttnpb.EndDevice, updateField string) ([]*ttnpb.EndDevice, error)
```
htdvisser: I actually think it would be better to just add a `BatchUpdateEndDeviceLastSeen` instead of some generic `BatchUpdateEndDevice` that in the current implementation only supports `last_seen`.

pkg/identityserver/gormstore/end_device_store.go
pgalic96: The problem (with this implementation as well) is that it could race (multiple AS instances updating at the same time). 
Now that I am thinking I guess we don't really care about these timestamps being the most accurate (in case of race it will be a ms difference, I suppose). So I think it would be acceptable to have some data races here?

adriansmares: Even the same Application Server will race with itself in the model that we have. It is fine to have eventual consistency.

pkg/identityserver/end_device_registry.go
```diff
@@ -205,6 +205,21 @@ func (is *IdentityServer) updateEndDevice(ctx context.Context, req *ttnpb.Update
 		req.FieldMask = cleanFieldMaskPaths(ttnpb.EndDeviceFieldPathsNested, req.FieldMask, nil, []string{"activated_at"})
 	}
 
+	if ttnpb.HasAnyField(req.FieldMask.GetPaths(), "last_seen_at") {
+		err = is.store.Transact(ctx, func(ctx context.Context, st store.Store) (err error) {
+			dev, err = st.GetEndDevice(ctx, req.EndDevice.Ids, store.FieldMask{"last_seen_at"})
```
adriansmares: Just write the value, no need to add these checks and round trips. The value can be eventually consistent.

pkg/identityserver/end_device_registry.go
```diff
@@ -228,6 +243,26 @@ func (is *IdentityServer) updateEndDevice(ctx context.Context, req *ttnpb.Update
...
+	"multiple fieldmask paths specified while batch update allows only one field update",
+)
+
+func (is *IdentityServer) batchUpdateEndDeviceLastSeen(ctx context.Context, req *ttnpb.BatchUpdateEndDeviceLastSeenRequest) (*ttnpb.EndDevices, error) {
+	if err := clusterauth.Authorized(ctx); err != nil {
+		return nil, err
+	}
+	endDevices := make([]*ttnpb.EndDevice, len(req.EndDevices.EndDevices))
+	err := is.store.Transact(ctx, func(ctx context.Context, st store.Store) (err error) {
+		endDevices, err = st.BatchUpdateEndDeviceLastSeen(ctx, req.GetEndDevices().EndDevices)
```
adriansmares: Why use a getter if you will access the field directly afterwards ? It would panic for `nil`.

adriansmares (approved):
I still see time ordering behavior here - it is not needed.

We will see, at application level, uplinks from a singular end device only every couple of seconds at most. 

A LoRaWAN end device cannot push application uplinks faster than the RX2 downlink window. For 99% of the end device in our stack, that means a gap of at least 5+1 seconds, to which you would need to add some extra latency that pushes you towards 6.5 seconds. We are going to be fine if the last seen timestamp is ~10 seconds off.

AS looks good to me for now.

api/end_device.proto
```diff
@@ -595,6 +599,10 @@ message UpdateEndDeviceRequest {
   google.protobuf.FieldMask field_mask = 2;
 }
 
+message BatchUpdateEndDeviceLastSeenRequest {
+  EndDevices end_devices = 1 [(validate.rules).message.required = true];
```
htdvisser: I don't think it's appropriate to use full `EndDevices` here. Why not use a specialized message that only contains what we actually use?

```suggestion
  message EndDeviceLastSeenUpdate {
  	EndDeviceIdentifiers ids = 1 [(validate.rules).message.required = true];	
	google.protobuf.Timestamp last_seen = 2 [(validate.rules).message.required = true];
  }
  repeated EndDeviceLastSeenUpdate updates = 1 [(validate.rules).message.required = true];
```

api/end_device_services.proto
```diff
@@ -80,6 +80,9 @@ service EndDeviceRegistry {
     };
   };
 
+  // Update a batch of end devices, changing the fields specified by the field mask to the provided values.
+  rpc BatchUpdateLastSeen(BatchUpdateEndDeviceLastSeenRequest) returns (EndDevices);
```
htdvisser: What interest does the caller have in those returned EndDevices? If it doesn't care, then let's just return a `google.protobuf.Empty`; we can then always change the returned type in the future.

pkg/identityserver/end_device_registry.go
htdvisser: This error definition is now unused, right?

pkg/identityserver/store/store_interfaces.go
```diff
@@ -65,6 +65,7 @@ type EndDeviceStore interface {
 	GetEndDevice(ctx context.Context, id *ttnpb.EndDeviceIdentifiers, fieldMask FieldMask) (*ttnpb.EndDevice, error)
 	UpdateEndDevice(ctx context.Context, dev *ttnpb.EndDevice, fieldMask FieldMask) (*ttnpb.EndDevice, error)
 	DeleteEndDevice(ctx context.Context, id *ttnpb.EndDeviceIdentifiers) error
+	BatchUpdateEndDeviceLastSeen(ctx context.Context, devs []*ttnpb.EndDevice) ([]*ttnpb.EndDevice, error)
```
htdvisser: Same as my comment on API. I don't think using full `[]*ttnpb.EndDevice`s is appropriate here.

pkg/identityserver/gormstore/end_device_store.go
```diff
@@ -217,3 +217,38 @@ func (s *deviceStore) DeleteEndDevice(ctx context.Context, id *ttnpb.EndDeviceId
 	defer trace.StartRegion(ctx, "delete end device").End()
 	return s.deleteEntity(ctx, id)
 }
+
+var errUnknownField = errors.DefineInvalidArgument(
```
htdvisser: This error definition is now unused, right?

pkg/identityserver/gormstore/end_device_store.go
```diff
@@ -217,3 +217,38 @@ func (s *deviceStore) DeleteEndDevice(ctx context.Context, id *ttnpb.EndDeviceId
...
+)
+
+func batchUpdateLastSeenAtQuery(placeholders []string) string {
+	return fmt.Sprintf(`
+	WITH updates (application_id, device_id, last_seen_at) AS (VALUES %s)
+	UPDATE end_devices AS end_device
+	SET last_seen_at = updates.last_seen_at
+	FROM updates
+	WHERE end_device.application_id = updates.application_id
+	AND end_device.device_id = updates.device_id
```
htdvisser, line 228: And if you still want that "only update to more recent timestamps", then I think you can add something like

```sql
AND (end_device.last_seen_at IS NULL OR end_device.last_seen_at < updates.last_seen_at)
```

api/end_device.proto
```diff
@@ -595,6 +599,16 @@ message UpdateEndDeviceRequest {
   google.protobuf.FieldMask field_mask = 2;
 }
 
+message BatchUpdateEndDeviceLastSeenRequest {
+  message EndDeviceLastSeenUpdate {
+  	EndDeviceIdentifiers ids = 1 [(validate.rules).message.required = true];
+	  google.protobuf.Timestamp last_seen_at = 2;
```
htdvisser: Strange mix of tabs+spaces

pkg/identityserver/end_device_registry.go
```diff
@@ -191,7 +191,7 @@ func (is *IdentityServer) setFullEndDevicePictureURL(ctx context.Context, dev *t
 
 func (is *IdentityServer) updateEndDevice(ctx context.Context, req *ttnpb.UpdateEndDeviceRequest) (dev *ttnpb.EndDevice, err error) {
 	if clusterauth.Authorized(ctx) == nil {
-		req.FieldMask = cleanFieldMaskPaths([]string{"activated_at", "locations"}, req.FieldMask, nil, getPaths)
+		req.FieldMask = cleanFieldMaskPaths([]string{"activated_at", "locations", "last_seen_at"}, req.FieldMask, nil, getPaths)
```
htdvisser, line 270: This change is no longer necessary, but I'm fine with keeping it

pkg/identityserver/gormstore/end_device_store.go
```diff
@@ -217,3 +217,31 @@ func (s *deviceStore) DeleteEndDevice(ctx context.Context, id *ttnpb.EndDeviceId
...
+	defer trace.StartRegion(ctx, "batch update end devices").End()
+
+	valueStrings := []string{}
+	valueArgs := []interface{}{}
+
+	for _, dev := range devsLastSeen {
+		valueStrings = append(valueStrings, "(?, ?, ?::timestamptz)")
+		valueArgs = append(valueArgs, dev.Ids.ApplicationIds.ApplicationId)
+		valueArgs = append(valueArgs, dev.Ids.DeviceId)
+		valueArgs = append(valueArgs, ttnpb.StdTime(dev.LastSeenAt))
```
htdvisser: Nit: why multiple `append`s if you can combine them into one with multiple args?

pkg/identityserver/storetest/end_device_store.go
```diff
@@ -460,3 +462,81 @@ func (st *StoreTest) TestEndDeviceStorePagination(t *T) {
...
+
+		validDevTime := time.Now()
+		dev1.LastSeenAt = ttnpb.ProtoTimePtr(validDevTime)
+		dev2.LastSeenAt = ttnpb.ProtoTimePtr(validDevTime)
+		dev3.LastSeenAt = ttnpb.ProtoTimePtr(validDevTime.Add(10 * time.Second))
+
+		batch := []*ttnpb.BatchUpdateEndDeviceLastSeenRequest_EndDeviceLastSeenUpdate{
+			{Ids: dev1.Ids, LastSeenAt: dev1.LastSeenAt},
+			{Ids: dev2.Ids, LastSeenAt: dev2.LastSeenAt},
+			{Ids: dev3.Ids, LastSeenAt: dev3.LastSeenAt},
```
htdvisser, line 500: Should we also check that it doesn't crash if the same device is updated multiple times?

pkg/identityserver/storetest/end_device_store.go
```diff
@@ -460,3 +462,81 @@ func (st *StoreTest) TestEndDeviceStorePagination(t *T) {
...
+		}
+		err := s.BatchUpdateEndDeviceLastSeen(ctx, batch)
+		a.So(err, should.BeNil)
+		got, err := s.ListEndDevices(ctx, app1.GetIds(), []string{"last_seen_at"})
+		a.So(err, should.BeNil)
+		a.So(got, should.HaveLength, 3)
+
+		for _, dev := range got {
+			a.So(dev.LastSeenAt, should.NotBeNil)
+		}
```
htdvisser: Nit:

```suggestion
		if a.So(err, should.BeNil) && a.So(got, should.NotBeNil) && a.So(got, should.HaveLength, 3) {
			for _, dev := range got {
				a.So(dev.LastSeenAt, should.NotBeNil)
			}
		}
```

That way, if something doesn't go as expected, we'll just see the first failure and won't get confused by panics caused by subsequent assertions.

---

And should we make sure that the correct timestamp is set for the correct device (just copy-paste from below)?

pkg/identityserver/storetest/end_device_store.go
```diff
@@ -460,3 +462,81 @@ func (st *StoreTest) TestEndDeviceStorePagination(t *T) {
...
+		for _, dev := range got {
+			a.So(dev.LastSeenAt, should.NotBeNil)
+			if dev.Ids.DeviceId == dev1.Ids.DeviceId {
+				a.So(dev.LastSeenAt, should.Resemble, ttnpb.ProtoTimePtr(validDevTime))
+			} else if dev.Ids.DeviceId == dev2.Ids.DeviceId {
+				a.So(dev.LastSeenAt, should.Resemble, ttnpb.ProtoTimePtr(validDevTime))
+			} else if dev.Ids.DeviceId == dev3.Ids.DeviceId {
+				a.So(dev.LastSeenAt, should.Resemble, ttnpb.ProtoTimePtr(validDevTime.Add(10*time.Second)))
+			}
+		}
```
htdvisser: Nit: also wrap this in `if` like I suggested above.

pkg/identityserver/end_device_registry.go
pgalic96: This is still needed in case the batch update is disabled, then the devices get updated one by one by using `Update` RPC.

htdvisser (approved):
Code-owned changes LGTM

pgalic96:
Test cache is broken for this PR, PR from a separate branch (https://github.com/TheThingsNetwork/lorawan-stack/pull/5385) passes all tests. This can be merged.

PR #5311, opened 2022-03-16
Add protoc-gen-go-flags generated flag sets to CLI

pgalic96:
#### Summary

References #2798

#### Changes

- Add `protoc-gen-go-flags` plugin to protos
- Replace reflect based flag creation with the new generated flags for proto structs

#### Testing

🐒 

##### Regressions

anything could break.

#### Notes for Reviewers

Probably this could be merged after the CLI integration tests?

#### Checklist


htdvisser (commented):
> Regressions
>
> anything could break.

Yeah, sorry, but you're going to have to do better than that.

If I `go run ./cmd/ttn-lw-cli gen-md-doc` on v3.18, and then diff that with the result of running it on this branch, I already see some changes that may be problematic:

- `--pubsub-id` flag missing in `ttn-lw-cli applications pubsubs get`
- missing description for `--grants` and `--rights` in `ttn-lw-cli clients create`: do those still work?
- `ttn-lw-cli organizations create` suddenly having an `--email` flag.
- `ttn-lw-cli users create` and `update` suddenly having `--password-updated-at` and `--temporary-password-created-at` flags.

And in order to test for further regressions, you could for example run a bunch of commands with the `--dump-requests` flag and compare the requests against the ones generated for the same commands on the v3.18 branch.

cmd/ttn-lw-cli/commands/applications.go
```diff
@@ -80,22 +73,22 @@ var (
...
 		RunE: func(cmd *cobra.Command, args []string) error {
+			req := &ttnpb.ListApplicationsRequest{}
+			_, err := req.SetFromFlags(cmd.Flags(), "")
+			if err != nil {
+				return err
+			}
 			paths := util.SelectFieldMask(cmd.Flags(), selectApplicationFlags)
 			paths = ttnpb.AllowedFields(paths, ttnpb.RPCFieldMaskPaths["/ttn.lorawan.v3.ApplicationRegistry/List"].Allowed)
-
+			if req.FieldMask == nil {
```
adriansmares, line 84: Why would the field mask not be `nil` ?

cmd/ttn-lw-cli/commands/clients.go
```diff
@@ -60,13 +59,13 @@ func getClientID(flagSet *pflag.FlagSet, args []string) *ttnpb.ClientIdentifiers
...
 }
 
-var searchClientsFlags = func() *pflag.FlagSet {
-	flagSet := &pflag.FlagSet{}
-	flagSet.AddFlagSet(searchFlags)
-	// NOTE: These flags need to be named with underscores, not dashes!
-	util.AddField(flagSet, "state", reflect.TypeOf([]ttnpb.State{}), false)
-	return flagSet
-}()
+// var searchClientsFlags = func() *pflag.FlagSet {
```
adriansmares: This probably should be deleted if it is not used ?

go.mod
```diff
@@ -2,6 +2,7 @@ module go.thethings.network/lorawan-stack/v3
 
 go 1.17
 
+replace github.com/TheThingsIndustries/protoc-gen-go-flags => ../../github.com/TheThingsIndustries/protoc-gen-go-flags
```
adriansmares: Is this still needed ? Isn't the latest version tagged ?

pgalic96: yep, I am using this to make ad-hoc fixes and accidentally committed it.

cmd/ttn-lw-cli/commands/applications.go
pgalic96: initially the idea was that the user could set `--field-mask` flag directly, but I think now we would rather go with skipping this flag.

pgalic96:
@htdvisser I went through all changed commands and fixed issues I could encounter (and also did some diffs with prev implementation). I think the only way we feel 100% confident that this won't introduce any bugs is that we merge it after https://github.com/TheThingsIndustries/lorawan-stack/pull/3097 .

go.mod
htdvisser: Please remove this replacement. Tip: if you're already using Go 1.18 for development, you can setup a Go Workspace (`go.work`) where you `use` your local clone of the dependency.

go.mod
```diff
@@ -237,3 +238,10 @@ require (
 	gopkg.in/alexcesaro/quotedprintable.v3 v3.0.0-20150716171945-2caba252f4dc // indirect
 	gopkg.in/ini.v1 v1.66.4 // indirect
 )
+
+require github.com/TheThingsIndustries/protoc-gen-go-flags v0.0.0-20220307091919-193eaeff17e7
```
htdvisser: Let's create a tagged version of protoc-gen-go-flags and use that

htdvisser (commented):
Then I'd say we move forward with this now. Even with https://github.com/TheThingsIndustries/lorawan-stack/pull/3097 we won't cover 100% of the (edge) cases, especially considering that that mostly works with JSON input, so it won't properly cover the CLI flags anyway.

So I'm happy if you tested that simple (scalar) flags work, as well as our common edge cases:

1. Custom types (DevAddr, EUI + prefixes)
2. (Wrapped) enums
3. That weird custom `BoolValue` we defined
4. Oneofs

If there's still anything we missed, we can just fix that later. The compatibility commitment is about not _intentionally_ breaking things, but mistakes happen, and we'll fix those when we discover them.

htdvisser (approved):

adriansmares (approved):
LGTM.

I don't think that we need to block this PR on this, but I realized while reviewing that if I have two proto files, `end_device.proto` and `end_devices_flags.proto`, the flags Go code of the first file conflicts with the standard proto definition file of the second file. I think it would have been better for the naming structure to be `{file}.pb.flags.go`, in retrospective.

htdvisser:
We didn't come up with the naming of generated files, we just follow https://github.com/grpc/grpc-go/blob/master/cmd/protoc-gen-go-grpc/grpc.go

PR #5320, opened 2022-03-17
CSV device template converter

### LINKED ISSUE #5139: Import end devices from CSV
#### Summary

It looks like this was available before (https://github.com/TheThingsNetwork/lorawan-stack/pull/1448) but now only json is supported.

When we buy off-the-shelf LoRaWAN sensor, the vendors normally send us an Excel file with the DevEUI, AppEUI, AppKey, etc. A file like this can contain hundreds of entries, and I need to register these devices in TTS. The current (planned) solution is to write a python script to iterate through the Excel-csv, and call `ttn-lw-cli` for each one to register the device.

It would be nicer if TTS Console allowed importing devices directly from CSV. After uploading the csv, one would need to choose which columns contain which fields, and then enter default values for the fields that do not exist in the csv. A pre-defined format/column order would also be acceptable as Excel allows easily swapping columns. For default values one could fill the column with the default value in Excel.

I see TTS has a templating feature, but this seems too complicated and I'm not sure this is exactly what I need.

#### Why do we need this?

Make registering a batch of devices from a vendor provided Excel file easier.

I can import a batch of devices, but the imported file has to be in TTS JSON format.

Also have a CSV import file format, with the ability to choose which columns contain which values.

#### Environment

TTS v3.17

Purely on the console using Javascript. But maybe it can also be a ttn-lw-cli feature, but needs to be documented then.

Import a csv file that contains DevEUI, AppEUI, AppKey values.

No

johanstokking:
#### Summary

Closes #5139 
Documented at https://github.com/TheThingsIndustries/lorawan-stack-docs/pull/821

#### Changes

- Rename the existing JSON converter in the codebase
- Add CSV template converter. This works with the same format as used on The Things Join Server

#### Testing

Local testing and unit tests

#### Checklist


KrishnaIyer (approved):
Nice.

adriansmares (approved):

PR #5323, opened 2022-03-21
Fetch End Device Template by numeric IDs

### LINKED ISSUE #4842: Get device information by vendor ID and profile ID

KrishnaIyer:
#### Summary

Closes #4842 

#### Changes

- Create a new index in bleve for the profiles.
- Extend the `GetTemplate` RPC to allow numeric IDs and get them.
- Update tests
- Remove the unused `no-op` implementation.

#### Testing

UT + local

##### Regressions

- The current `GetTemplate` function (which loads the template on the console) is tested and works as expected.

#### Notes for Reviewers

1. Afaik, none of the device profiles in the repository have a correct `vendorProfileID`. This field is either not present or filled with the same value for all devices for a brand. Until now this wasn't necessary, but for customers to use this PR properly, those need to be adjusted. cc @Jaime-Trinidad 

#### Checklist


pkg/devicerepository/store/remote/schema.go
```diff
@@ -300,155 +286,6 @@ func (d EndDeviceModel) ToPB(brandID, modelID string, paths ...string) (*ttnpb.E
 	return res, nil
 }
 
-// EndDeviceProfile is the format of the `vendor/<vendor-id>/<profile-id>.yaml` file.
```
KrishnaIyer, line 303: These are moved to `pkg/devicerepository/store/profiles.go`

api/devicerepository.proto
```diff
@@ -298,11 +298,24 @@ message ListEndDeviceModelsRequest {
...
+
   // Application identifiers.
   ApplicationIdentifiers application_ids = 1 [(validate.rules).message.required = true];
 
   // End device version information.
+  // EndDeviceProfileIdentifiers take precedence over this field if both are specified.
   EndDeviceVersionIdentifiers version_ids = 2;
+
+  // End device profile identifiers.
+  EndDeviceProfileIdentifiers end_device_profile_ids = 3;
```
htdvisser: You could also move these two into a oneof; proto-wise that's a backwards compatible change.

api/devicerepository.proto
```diff
@@ -400,6 +413,9 @@ service DeviceRepository {
   rpc GetTemplate(GetTemplateRequest) returns (EndDeviceTemplate) {
     option (google.api.http) = {
       get: "/dr/applications/{application_ids.application_id}/brands/{version_ids.brand_id}/models/{version_ids.model_id}/{version_ids.firmware_version}/{version_ids.band_id}/template"
+      additional_bindings {
+        get: "/dr/applications/{application_ids.application_id}/vendor-id/{end_device_profile_ids.vendor_id}/vendor-profile-id/{end_device_profile_ids.vendor_profile_id}/template"
```
htdvisser: I don't think we use `-id` anywhere in our API.

```suggestion
        get: "/dr/applications/{application_ids.application_id}/vendors/{end_device_profile_ids.vendor_id}/profiles/{end_device_profile_ids.vendor_profile_id}/template"
```

api/devicerepository.proto
adriansmares: The `oneof` won't be compatible with the gRPC gateway I think. iirc we had the same discussion / suggestion for the Storage Integration requests.

KrishnaIyer: Indeed. I didn't think that it was backwards compatible. I'd very much prefer to use a oneof.

htdvisser: @adriansmares can you give an example of this breaking compatibility? Because I can't think of any.

The only thing that changes here is the generated code. On the wire (both as protobuf and as JSON) setting `version_ids` (or not setting) looks exactly the same, regardless of whether it's in a oneof or not.

adriansmares: Oneof fields cannot be used in the paths for the HTTP paths of the gRPC gateway, or at least so it seems. We've had the same discussion for the storage integration, where you could get data for the whole application, or for a single end device (so a `oneof` made perfect sense from a protobuf perspective).

Take the following diff that introduces the `oneof`:
```diff
diff --git a/api/devicerepository.proto b/api/devicerepository.proto
index f203d27d3..18c64b12d 100644
--- a/api/devicerepository.proto
+++ b/api/devicerepository.proto
@@ -305,17 +305,19 @@ message GetTemplateRequest {
 
     // ID of the LoRaWAN end device profile. The naming of this field follows the TR005 specification.
     uint32 vendor_profile_id = 2;
-    }
+  }
 
   // Application identifiers.
   ApplicationIdentifiers application_ids = 1 [(validate.rules).message.required = true];
 
-  // End device version information.
-  // EndDeviceProfileIdentifiers take precedence over this field if both are specified.
-  EndDeviceVersionIdentifiers version_ids = 2;
+  oneof ids {
+    // End device version information.
+    // EndDeviceProfileIdentifiers take precedence over this field if both are specified.
+    EndDeviceVersionIdentifiers version_ids = 2;
 
-  // End device profile identifiers.
-  EndDeviceProfileIdentifiers end_device_profile_ids = 3;
+    // End device profile identifiers.
+    EndDeviceProfileIdentifiers end_device_profile_ids = 3;
+  }
 }
 
 message GetPayloadFormatterRequest {
```

It works exactly as you expect, and the PBs compile happily. Then at runtime we try to actually use one of the HTTP paths:

```
adriansmares@Adrian-Stefans-MacBook-Pro lorawan-stack % curl 'http://localhost:1885/api/v3/dr/applications/dev1/brands/adeunis/models/analog/1.03/EU_863_870/template' -H 'Authorization: Bearer MFRWG.FW64YQR7SU5UG52S7IGILBES675MXJGYHJMVKCQ.XJXQQBPARCOS7JLICFKBSOEDUGODLUJ5COJOAHSJOL5C2AI675OA'
{"code":3,"message":"type mismatch, parameter: version_ids.model_id, error: field already set for ids oneof"}%
```

This looks like a bug in `grpc-gateway`, but I don't know if it ever got fixed.

Edit: allegedly this issue is fixed in the v2 version of `grpc-gateway`, per https://github.com/grpc-ecosystem/grpc-gateway/issues/753#issuecomment-741662557 , so maybe we can come back later and add the `oneof` then ? If we make the fields mutually exclusive already via validation at RPC level, it should be fine with regards to the compatibility commitment I think.

htdvisser: > Oneof fields cannot be used in the paths for the HTTP paths of the gRPC gateway, or at least so it seems. 

We also use oneof fields for the the collaborator in the IS, and that has always worked just fine:

https://github.com/TheThingsNetwork/lorawan-stack/blob/fd41a7df32f47c0cf3bf4af6b5f65e4ea3a239d0/api/application_services.proto#L36-L45

adriansmares: Based on https://github.com/grpc-ecosystem/grpc-gateway/issues/753#issuecomment-741662557 this happens because there are more fields set in the path. So for example `organization_id` is the only field of the `oneof` which is set here, which apparently works.

But in our case `EndDeviceVersionIdentifiers` has multiple fields; the first one is successfully parsed (`version_ids.brand_id`), and then the second one, `version_ids.model_id`, fails. 

What we can do is that we ensure that the fields are mutually exclusive in the `GetTemplate` implementation (i.e. one field being not `nil` implies the other one is `nil`). Then we add the `oneof` when we upgrade to `grpc-gateway` v2.

htdvisser (approved):

api/devicerepository.proto
```diff
@@ -298,11 +298,24 @@ message ListEndDeviceModelsRequest {
...
 
 message GetTemplateRequest {
+  // Identifiers to uniquely identify a LoRaWAN end device profile.
+  message EndDeviceProfileIdentifiers {
+    // VendorID managed by the LoRa Alliance, as defined in TR005.
+    uint32 vendor_id = 1 [(validate.rules).uint32 = {gte: 1}];
+
+    // ID of the LoRaWAN end device profile. The naming of this field follows the TR005 specification.
+    uint32 vendor_profile_id = 2;
+    }
```
johanstokking: Spacing

api/devicerepository.proto
```diff
@@ -298,11 +298,24 @@ message ListEndDeviceModelsRequest {
 }
 
 message GetTemplateRequest {
+  // Identifiers to uniquely identify a LoRaWAN end device profile.
+  message EndDeviceProfileIdentifiers {
+    // VendorID managed by the LoRa Alliance, as defined in TR005.
+    uint32 vendor_id = 1 [(validate.rules).uint32 = {gte: 1}];
+
+    // ID of the LoRaWAN end device profile. The naming of this field follows the TR005 specification.
```
johanstokking:
```suggestion
    // ID of the LoRaWAN end device profile assigned by the vendor.
```

pkg/devicerepository/store/bleve/store.go
```diff
@@ -179,8 +183,60 @@ func (s *bleveStore) GetModels(req store.GetModelsRequest) (*store.GetModelsResp
...
+
+		if len(result.Hits) == 0 {
+			return nil, errEndDeviceProfileNotFound.WithAttributes(
+				"vendor_id", endDeviceProfileIds.VendorId,
+				"vendor_profile_id", endDeviceProfileIds.VendorProfileId)
+		}
+
+		if len(result.Hits) != 1 {
+			// There can only be one profile for a given tuple.
+			return nil, errInvalidNumberOfProfiles.New()
```
johanstokking, line 224: I don't think this is validated at the moment

@Jaime-Trinidad please file an issue to validate this

johanstokking (approved):

pkg/devicerepository/store/bleve/store.go
KrishnaIyer: https://github.com/TheThingsNetwork/lorawan-devices/issues/405

Issue #5092, opened 2022-01-12
Allow overriding the maximum number of retransmissions


adriansmares:
#### Summary

The end device MAC settings should allow the user to limit the max `NbTrans` used during ADR.

#### Why do we need this?

The `NbTrans` parameter controls the number of transmissions done for a singular LoRaWAN frame. It is by default one, and is increased when the Network Server detects gaps in the uplink frame counters. It adds a degree of redundancy which is meant to help with the loss rate experienced by the end device. 

This mechanism is abstracted away from the application layer - the application sees the message only one time. But, if the application layer protocol does retransmissions on its own, the `NbTrans` mechanism is itself redundant and can cause significant costs in battery life.

#### What is already there? What do you see now?

A static maximum of 3 transmissions
https://github.com/TheThingsNetwork/lorawan-stack/blob/10da1157cb27216d89e651a43750686bb9ea5ea4/pkg/networkserver/mac/adr.go#L71-L72

#### What is missing? What do you want to see?

1. A MAC setting that allows the user to limit the `NbTrans`. I don't think we should allow the rates themselves to be editable, so this setting should be at most 3.
2. Take into account the MAC setting during the `NbTrans` computation
https://github.com/TheThingsNetwork/lorawan-stack/blob/10da1157cb27216d89e651a43750686bb9ea5ea4/pkg/networkserver/mac/adr.go#L287-L301

#### Environment

`3.17.0`

#### How do you propose to implement this?

Add the API option and adapt the ADR code to take it into account, similar to how other MAC settings are taken into account (separate function that takes into account overrides such as the NS settings).

#### How do you propose to test this?

Unit testing.

#### Can you do this yourself and submit a Pull Request?

Yes. cc @htdvisser / @johanstokking if there are any comments on this.

johanstokking:
Wouldn't we want to implement this with https://github.com/TheThingsNetwork/lorawan-stack/issues/40?

I think it becomes tedious to configure this per end device.

adriansmares:
But the profiles are on a per-device basis too, right ?  While these kind of things could become part of the profile, you still need to configure this on a per-device basis.

johanstokking:
[...]

Issue #3820, opened 2021-02-17
Option to set persistent LinkADR settings when using Custom ADR option


i3mSuperMan:
#### Summary
The custom ADR option allows to set the ADR parameter values, and NS communicates them to the device via `LinkADRReq`. But these values are not persistent and lost on reset of MAC State during re-join, thus requiring re-configuration every time device joins the network.

#### Why do we need this?
To make the parameters' values persistent and communicate to the device in case of MAC State reset after Join, to avoid manual intervention to set them with every `re-join` of the device.

#### What is already there? What do you see now?
The custom ADR option currently allows configuring the `adr-data-rate-index`, `adr-tx-power-index` & `adr-nb-trans` values using the below flags. These values `reset` when the MAC State of the device `reset` on re-join and not communicated to the device.
```
ttn-lw-cli end-devices set <app-id> <dev-id> --mac-state.desired-parameters.adr-data-rate-index
ttn-lw-cli end-devices set <app-id> <dev-id> --mac-state.desired-parameters.adr-tx-power-index
ttn-lw-cli end-devices set <app-id> <dev-id> --mac-state.desired-parameters.adr-nb-trans
```

#### What is missing? What do you want to see?
An option to make the changes to `adr-data-rate-index`, `adr-tx-power-index` & `adr-nb-trans` parameters persistent.

#### Environment
TTES V3.10.9 CH

#### How do you propose to implement this?
Add relevant `mac-settings` flags at the NS end.

#### How do you propose to test this?

...

#### Can you do this yourself and submit a Pull Request?
No. @rvolosatovs

rvolosatovs:
Good to note that this is mostly relevant for pre-1.1 devices in US915-like bands

rvolosatovs:
Also, these fields should be mutually exclusive with `enable_adr`

baranisrc:
[...]

PR #3604, opened 2020-12-21
Allow custom ADR in Network Server

rvolosatovs:
#### Summary

Allow custom ADR in Network Server
Refs https://github.com/TheThingsIndustries/lorawan-stack-support/issues/242
Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/3564

#### Changes

- Send `LinkADRReq` with ADR parameters even if `use-adr` is `false`

#### Testing

Unit tests

##### Regressions

Higely unlikely

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [x] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

Issue #40, opened 2019-01-29
ADR profiles


johanstokking:
**Summary:** 

ADR profiles give fine grained control to which strategy is being followed per end device.

**What is already there? What do you see now?**

Basic ADR algorithm. This is essentially the "low power" profile.

**What is missing? What do you want to see?**

I'd suggest three profiles, something like this:

1. Low power: highest data rate, still downlink path available (even if it were RX2). Useful for applications that use battery powered sensors (i.e. uplink oriented)
2. Downlink availability: reach gateways that have a downlink path available in RX1. Useful for applications that use quite some downlink. Parameters:
   - target number of gateways with downlink available in RX1 (i.e. 2)
   - minimum allowed data rate (i.e. SF10)
3. Reach: reach gateways for more critical applications or geolocation. Useful for applications that need geolocation (tags that enable ADR when the device becomes stationary) or if you want to reduce packet loss as much as possible. Parameters:
   - target number of gateways with decent reception (i.e. 3)
   - minimum allowed data rate (i.e. SF10)

----

Original issue: https://github.com/TheThingsIndustries/lorawan-stack/issues/1258 by @johanstokking

PR #5326, opened 2022-03-21
Add ADR settings API

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/3820
References https://github.com/TheThingsNetwork/lorawan-stack/issues/5092

#### Changes

- Refactor the ADR settings API that is part of the end device MAC settings

#### Testing

This is purely API, N/A.

##### Regressions

N/A.

#### Notes for Reviewers

My proposal is to add the API mentioned in this PR, with the following semantics:
| |`use_adr` = `true` | `use_adr` = `false` | `use_adr`= `null`|
| --- | --- | --- | --- |
| `adr.mode` = `null` | Adapt the data rate, with the margin `adr_margin` (current behavior) | Use the desired parameters values (current behavior) | Adapt the data rate, with the margin `adr_margin` (current behavior) |
| `adr.mode` = `dynamic` | Adapt the data rate, with the margin `adr.mode.dynamic.margin` | Same as left hand side | Same as left hand side |
| `adr.mode` = `static` | Use the `adr.mode.static` parameters | Same as left hand side | Same as left hand side |
| `adr.mode` = `disabled` | Do not adapt the data rates, just enable/disable channels | Same as left hand side | Same as left hand side |

The idea is to change nothing for existing end devices, while introducing settings which are far more flexible for newer configurations. Once the end device is configured using `adr.mode`, the `use_adr` / `adr_margin` fields are ignored completely.

Note that the band has veto powers and may disable ADR completely. The ISM2400 band does this. For such bands we want to basically disable `adr.mode` = `dynamic`, since the algorithm is not compatible.

@htdvisser main reviewer, as @johanstokking is off for now.

#### Checklist


api/end_device.proto
```diff
@@ -167,6 +167,39 @@ message EndDeviceVersion {
...
+    // Transmission power index to use.
+    uint32 tx_power_index = 2 [(validate.rules).uint32.lte = 15];
+    // Number of retransmissions.
+    uint32 nb_trans = 3 [(validate.rules).uint32.lte = 15];
+  }
+
+  // Configuration options for dynamic ADR.
+  message DynamicMode {
+    // The ADR margin tells the network server how much margin it should add in ADR requests.
+    // A bigger margin is less efficient, but gives a better chance of successful reception.
```
htdvisser: Maybe add that if this is not set explicitly, the network default is used.

api/end_device.proto
```diff
@@ -167,6 +167,39 @@ message EndDeviceVersion {
...
+    google.protobuf.FloatValue margin = 1;
+    // Maximum number of retransmissions.
+    google.protobuf.UInt32Value max_nb_trans = 2 [(validate.rules).uint32 = {gte: 1, lte: 3}];
```
htdvisser: 1. Shouldn't we have both a `min_nb_trans` and `max_nb_trans`?
2. Should we then also have `{min,max}_{data_rate_index,tx_power_index}`?
3. Maybe add that if this is not set explicitly, the network default is used.

htdvisser (commented):
I don't know if we need to define new messages, and "ADR modes". We can also start by adding additional constraints (`{min,max}_{data_rate_index,tx_power_index,nb_trans}`). That way we can also cover `static` and `dynamic`. Setting the existing `use_adr` to `false` should be equivalent to the `disabled` mode, right?

adriansmares:
> I don't know if we need to define new messages, and "ADR modes". We can also start by adding additional constraints (`{min,max}_{data_rate_index,tx_power_index,nb_trans}`). That way we can also cover `static` and `dynamic`. Setting the existing `use_adr` to `false` should be equivalent to the `disabled` mode, right?

There are some problems with this, which I've described in the PR body. First of all, `use_adr` to `false` is not actually disabling ADR. It is currently impossible to disable ADR. Ever since https://github.com/TheThingsNetwork/lorawan-stack/pull/3604 , `use_adr` = `false` means that some external application needs to change the desired parameters to the current parameters in order to avoid having everything fall back to 0.

So `use_adr` setting currently has some very bad semantics, and there are customers in the wild that use them. At the same time, we cannot disable ADR at all using these settings. 

I don't think that we should add more boolean options in order to reach the state space that we want. Instead, we should prepare the API for future changes already - if you want to use static/custom values, just use the `static` field. The NS then doesn't have to look at historical traffic, build maps or anything like that - it will just attempt to send the MAC command to the end device until `current` = `desired`. The same argument comes about experimental / newer ADR implementations (ADR that lowers the data rate). Adding more values then doing a nested switch is worse in my opinion than making the `oneof` option clear.

For dynamic, we can also add more `min/max` values, but again these settings make sense only when you are actually using ADR - having them stored while ADR is disabled/static just makes validation harder. I can already reserve their indices, and we can implement them later on.

htdvisser (commented):
Okay then I'm good with this.

api/end_device.proto
```diff
@@ -167,6 +167,59 @@ message EndDeviceVersion {
...
+    // Number of retransmissions.
+    uint32 nb_trans = 3 [(validate.rules).uint32.lte = 15];
+  }
+
+  // Configuration options for dynamic ADR.
+  message DynamicMode {
+    // The ADR margin tells the network server how much margin it should add in ADR requests.
+    // A bigger margin is less efficient, but gives a better chance of successful reception.
+    // If unset, the default value from Network Server configuration will be used.
+    google.protobuf.FloatValue margin = 1;
```
johanstokking, line 187: This is dB, right?

api/end_device.proto
```diff
@@ -167,6 +167,59 @@ message EndDeviceVersion {
...
+    // Maximum transmission power index.
+    // If unset, the default value from Network Server configuration will be used.
+    google.protobuf.UInt32Value max_tx_power_index = 5 [(validate.rules).uint32.lte = 15];
+
+    // Minimum number of retransmissions.
+    // If unset, the default value from Network Server configuration will be used.
+    google.protobuf.UInt32Value min_nb_trans = 6 [(validate.rules).uint32 = {gte: 1, lte: 3}];
+    // Maximum number of retransmissions.
+    // If unset, the default value from Network Server configuration will be used.
+    google.protobuf.UInt32Value max_nb_trans = 7 [(validate.rules).uint32 = {gte: 1, lte: 3}];
```
johanstokking, line 208: `NbTrans` has range `[1:15]`

johanstokking (approved):
API generally looks good, two remarks though that I'd like to see changed but no need for an additional review round.

Will there be a migration to move away from `use_adr`?

Also references https://github.com/TheThingsNetwork/lorawan-stack/issues/40 in the sense that those profiles could be a preset for the ADR mode. We may need additional (dynamic) options, like a soft minimum/maximum number of gateways, but we can do that later.

adriansmares:
> Will there be a migration to move away from `use_adr`?

No, as there are probably API users in the wild. Instead, we will make the fields mutually exclusive (so using the API here requires that you don't use `use_adr`/`adr_margin`).

api/end_device.proto
adriansmares: The algorithm that we use does not go over `3`, so allowing values over 3 is going to have no effect, while being confusing. The static settings allow the constant value to go to 15.

PR #5330, opened 2022-03-22
Set default port if it doesn't exist in LBS discover message

### LINKED ISSUE #5197: Wrong URL format returned to LNS
#### Summary
When using Basics Station LNS, the wrong URL can be provided to the gateway for MUXS. Basics Station seems to always expect the port to be specified, however TTS can return no port under some circumstances.

For example, if I set the `gcs.basic-station.default.lns-uri` or the gateway server address to `wss://stack.example.com:443`, TTS returns `wss://stack.example.com/traffic/eui-blabblah`, when it should have actually returned `wss://stack.example.com:443/traffic/eui-blabblah`.

#### Steps to Reproduce
1. Set `gcs.basic-station.default.lns-uri` to `wss://stack.example.com:443`, 
2. Follow the normal setup steps for a Gateway to connect it via Basics Station and CUPS, but make sure to leave the Gateway Server address field blank on the gateway
3. Basics Station tries to connect, but returns the error
`[TCE:ERRO] Illegal muxs URI (no port/path etc.): wss://example.com/traffic/eui-blahblah`

I'm not familiar with Go, but I have had a look through the code to see why the port isn't included. It seems to come from `info.Address`, which seems to come from this line: https://github.com/TheThingsNetwork/lorawan-stack/blob/95b5591a304e28e199d39a9754fc03777d1d7d17/pkg/gatewayserver/io/ws/ws.go#L128

Looking at the docs for Request.Host (https://pkg.go.dev/net/http#Request), it seems that this only returns the port _if the port is non-default for the scheme that was used_, so for example if you're connecting via `wss://` and you're connected via 443, it will omit the port, but if you use the scheme `wss://` but you're connected via 8887, the port will be included. The issue is that TTS currently assumes that it always returns the port.

(See [RFC 7230, section 5.4](https://rfc-editor.org/rfc/rfc7230.html#section-5.4): `if the connection's incoming TCP port number differs from the default port for the effective request URI's scheme, then a colon (":") and the incoming port number (in decimal form) are appended to the authority component`).

#### Environment
TTS v3.17.2 in Docker on a Ubuntu host (on EC2).

I have Application Load Balancer in front of the EC2 instance which I am using for SSL termination and re-routing LNS to port 443.

I am using a Dragino LPS8, running the latest firmware. 

I don't have much experience with Go, but I've quickly made a fix here: https://github.com/jameseden1/lorawan-stack/commit/a151a2c6d82ccea07c9ebd8613626763c9f883de

I'm not sure if there is a better way to fix this, but I think this commit demonstrates the issue well. 

If my linked commit above is an acceptable solution, just let me know and I can create a Pull Request

KrishnaIyer:
#### Summary

Closes #5197 

The reasoning is in the issue
> Looking at the docs for Request.Host (https://pkg.go.dev/net/http#Request), it seems that this only returns the port if the port is non-default for the scheme that was used, so for example if you're connecting via wss:// and you're connected via 443, it will omit the port, but if you use the scheme wss:// but you're connected via 8887, the port will be included. The issue is that TTS currently assumes that it always returns the port.

#### Changes

- Use default port values and set ports only if present in `r.Host`.

#### Testing

Locally 
```
» wscat -c ws://localhost:80/router-info
Connected (press CTRL+C to quit)
> {"router":"11-11-11-11-11-11-11-11"}
< {"router":"1111:1111:1111:1111","muxs":"muxs-::0","uri":"ws://localhost:80/traffic/eui-1111111111111111"}
Disconnected (code: 1006, reason: "")
```

##### Regressions

Tested
```
 » wscat -c ws://localhost:1887/router-info
Connected (press CTRL+C to quit)
> {"router":"11-11-11-11-11-11-11-11"}
< {"router":"1111:1111:1111:1111","muxs":"muxs-::0","uri":"ws://localhost:1887/traffic/eui-1111111111111111"}
Disconnected (code: 1006, reason: "")
```

#### Notes for Reviewers

This is quite difficult to UT since the tests need to bind port 80/443 and it's not easy to locally test since I need to setup a proxy. But I think the current tests are sufficient.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/ws/ws.go
+++ b/pkg/gatewayserver/io/ws/ws.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"net"
 	"net/http"
 	"regexp"
 	"strings"
@@ -129,14 +130,18 @@ func (s *srv) handleConnectionInfo(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// If port is retrievable from the host, use it.
-	address := strings.Split(r.Host, ":")
-	if len(address) == 2 {
-		port = address[1]
+	host, p, err := net.SplitHostPort(r.Host)
+	if err == nil {
+		// Both `host` and `p` are valid, since we have no error.
+		port = p
+	} else {
+		// `host` and `p` are unknown/empty. Reset `host` to `r.Host` since it does not contain the port.
+		host = r.Host
 	}
 
 	info := ServerInfo{
 		Scheme:  scheme,
-		Address: fmt.Sprintf("%s:%s", address[0], port),
+		Address: net.JoinHostPort(host, port),
 	}
 
 	resp := s.formatter.HandleConnectionInfo(ctx, data, s.server, info, time.Now())
--- a/pkg/gatewayserver/io/ws/ws.go
+++ b/pkg/gatewayserver/io/ws/ws.go
@@ -119,14 +119,24 @@ func (s *srv) handleConnectionInfo(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	scheme := "ws"
+	var (
+		scheme = "ws"
+		port   = "80"
+	)
 	if r.TLS != nil || s.cfg.UseTrafficTLSAddress {
 		scheme = "wss"
+		port = "443"
+	}
+
+	// If port is retrievable from the host, use it.
+	address := strings.Split(r.Host, ":")
+	if len(address) == 2 {
+		port = address[1]
 	}
 
 	info := ServerInfo{
 		Scheme:  scheme,
-		Address: r.Host,
+		Address: fmt.Sprintf("%s:%s", address[0], port),
 	}
 
 	resp := s.formatter.HandleConnectionInfo(ctx, data, s.server, info, time.Now())
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ For details about compatibility between different releases, see the **Commitment
 - Fix issues in the webhook forms causing webhooks to be created with all message types enabled and no way of deactivating message types.
 - Fix validation issue in the webhook form not detecting message type paths with more than 64 characters.
 - Fix "reactivate"-webhook button in the Console.
+- Port returned by the LBS LNS discovery message if standard 80/443 ports are used.
 
 ### Security
 


pkg/gatewayserver/io/ws/ws.go
```diff
@@ -119,14 +119,24 @@ func (s *srv) handleConnectionInfo(w http.ResponseWriter, r *http.Request) {
...
+		scheme = "ws"
+		port   = "80"
+	)
 	if r.TLS != nil || s.cfg.UseTrafficTLSAddress {
 		scheme = "wss"
+		port = "443"
+	}
+
+	// If port is retrievable from the host, use it.
+	address := strings.Split(r.Host, ":")
```
adriansmares: https://pkg.go.dev/net#SplitHostPort and check if `err == nil`

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -119,14 +120,24 @@ func (s *srv) handleConnectionInfo(w http.ResponseWriter, r *http.Request) {
...
+		scheme = "ws"
+		port   = "80"
+	)
 	if r.TLS != nil || s.cfg.UseTrafficTLSAddress {
 		scheme = "wss"
+		port = "443"
+	}
+
+	// If port is retrievable from the host, use it.
+	host, p, err := net.SplitHostPort(r.Host)
```
adriansmares, line 133: `host` is an unknown value if `err != nil`. You probably want to keep using `r.Host` if an error occurs (and the error means that the port is not there probably).

KrishnaIyer: Yeah well I think that my original implementation was more straightforward.

adriansmares: Unfortunately it wouldn't have worked on IPv6, where `:` can be part of the host.

KrishnaIyer: Ah yeah fair.

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -119,14 +120,25 @@ func (s *srv) handleConnectionInfo(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		port = p
+		host = r.Host
```
adriansmares:
```suggestion
	if err == nil {
	  // Both `host` and `p` are valid, since we have no error. 
		port = p
	} else {
	  // `host` and `p` are unknown/empty. Reset `host` to `r.Host` since it does not contain the port.
	  host = r.Host
	}
```

adriansmares (changes requested):

pkg/gatewayserver/io/ws/ws.go
```diff
@@ -119,14 +120,28 @@ func (s *srv) handleConnectionInfo(w http.ResponseWriter, r *http.Request) {
...
+		port = p
+	} else {
+		// `host` and `p` are unknown/empty. Reset `host` to `r.Host` since it does not contain the port.
+		host = r.Host
 	}
 
 	info := ServerInfo{
 		Scheme:  scheme,
-		Address: r.Host,
+		Address: fmt.Sprintf("%s:%s", host, port),
```
adriansmares: `net.JoinHostPort`

adriansmares (approved):

PR #5334, opened 2022-03-23
Add configurable session expiry time

kschiffer:
#### Summary
References https://github.com/TheThingsIndustries/lorawan-stack/issues/1411

#### Changes
- Add config option `--is.user-session.duration` (with a default of 365 days)
- Set session expiry based on this value (both `expires_at` and cookie max age)
- Add flag to the OAuth client schema `tie_access_to_session`
- When set, the OAuth server will not issue access tokens that are longer than the connected user session (e.g. when using refresh tokens)
- Add session expiration date to both `authInfo` and `/oauth/me` endpoints to make the console aware of session expiration times
- Refactor the way that "Log back in"-modals are shown
  - Some internal store refactor
  - Prevent them from showing before the initialization
  - Automatically show the modal when the session has expired (instead of only after the next auth action)

#### Testing

Manual testing.

#### Notes for Reviewers
- I've set the default session duration to one year now, which effectively equals the situation we currently have (no expiry)
- When the client has `tie_access_to_session` set to `true` (default `false`), the access token will never outlive the user session expiry. It is however still possible to use the access token after the user session was manually deleted. In order to prevent that as well, we'd need to add another flag to the access token structure and run a DB query for the session every time an access-token-authorized request is made. Alternatively, we can also delete associated access tokens automatically when a user session is deleted. I decided to not add any of that for now since that might need to be discussed first.

#### Checklist


pkg/account/session/config.go
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package session
+
+import "time"
+
+// SessionConfig is the configuration for user sessions.
+type Config struct {
+	Duration time.Duration `name:"duration" description:"Time after which user sessions will automatically expire (default 8760h)"`
```
adriansmares: You don't need to put the default in the description. It is generated automatically as long as it was set in the `cmd` package.

pkg/webui/lib/store/logics/status.js
```diff
@@ -115,4 +117,33 @@ const connectionCheckFailLogic = createLogic({
...
+    const { isLoggedIn, sessionExpiresAt } = action.payload
+    if (!isLoggedIn || !Boolean(sessionExpiresAt)) {
+      done()
+      return
+    }
+
+    const expiryDate = new Date(sessionExpiresAt)
+    const expiryDuration = expiryDate - new Date()
+
+    if (expiryDuration >= 0 && expiryDuration < 60 * 60 * 24 * 1000) {
```
adriansmares, line 133: Should this period be hard coded here ?

sdk/js/generated/api-definition.json
```diff
@@ -802,7 +802,6 @@
         "up.uplink_message.session_key_id",
         "up.uplink_message.settings",
         "up.uplink_message.settings.coding_rate",
-        "up.uplink_message.settings.concentrator_timestamp",
```
adriansmares: This diff probably shouldn't be here.

api/oauth.proto
```diff
@@ -78,6 +78,7 @@ message OAuthAccessTokenIdentifiers {
 message OAuthAccessToken {
   UserIdentifiers user_ids = 1 [(validate.rules).message.required = true];
   string user_session_id = 9 [(validate.rules).string.max_len = 64];
+  google.protobuf.Timestamp session_expires_at = 10;
```
htdvisser: 1. The ordering is a bit strange here. Why not put it together with the other `expires_at`?
2. Please also add a little comment to explain the difference between `expires_at` and `session_expires_at`

pkg/account/server.go
```diff
@@ -50,27 +56,30 @@ type Component interface {
...
+	config        Config
 	store         account_store.TransactionalInterface
 	session       sess.Session
 	generateCSP   func(config *oauth.Config, nonce string) string
 	schemaDecoder *schema.Decoder
 }
 
 // NewServer returns a new account app on top of the given store.
-func NewServer(c *component.Component, store account_store.TransactionalInterface, config oauth.Config, cspFunc func(config *oauth.Config, nonce string) string) (Server, error) {
+func NewServer(c *component.Component, store account_store.TransactionalInterface, oauthConfig oauth.Config, sessionConfig sess.Config, cspFunc func(config *oauth.Config, nonce string) string) (Server, error) {
```
htdvisser: This func signature isn't getting much better if we keep adding arguments. Maybe it's time to start taking a config struct (you're already introducing a new `Config` struct) or options instead.

pkg/account/server.go
```diff
@@ -42,6 +42,12 @@ type Server interface {
 	Logout(w http.ResponseWriter, r *http.Request)
 }
 
+// Config represents the Account app configuration
+type Config struct {
+	OAuth   oauth.Config `name:"oauth"`
+	Session sess.Config  `name:"session"`
+}
```
htdvisser: We typically put config structs in `config.go` so that the right CODEOWNERS are requested for review when making changes that could affect config compatibility.

pkg/account/user.go
```diff
@@ -58,9 +58,11 @@ func (s *server) CurrentUser(w http.ResponseWriter, r *http.Request) {
 	webhandlers.JSON(w, r, struct {
 		User       json.RawMessage `json:"user"`
 		LoggedInAt *time.Time      `json:"logged_in_at"`
+		ExpiresAt  *time.Time      `json:"expires_at"`
```
htdvisser: 1. Shouldn't this be `SessionExpiresAt` /  `session_expires_at`?
2. While we're at it, we may also want to render the `session.SessionId`.

pkg/identityserver/client_registry.go
```diff
@@ -121,6 +121,7 @@ func (is *IdentityServer) createClient(ctx context.Context, req *ttnpb.CreateCli
 		req.Client.StateDescription = "admin approval required"
 		req.Client.SkipAuthorization = false
 		req.Client.Endorsed = false
+		req.Client.TieAccessToSession = false
```
htdvisser: I don't think we need to force this to `false` for user-created clients _on create_.

But: your description in the API says that it can only be modified by admins, and `updateClient` doesn't prevent non-admins from updating the field.

pkg/identityserver/gormstore/client.go
```diff
@@ -52,6 +52,8 @@ type Client struct {
 
 	Grants Grants `gorm:"type:INT ARRAY"`
 	Rights Rights `gorm:"type:INT ARRAY"`
+
+	TieAccessToSession bool `b	gorm:"not null"`
```
htdvisser:
```suggestion
	TieAccessToSession bool `gorm:"not null"`
```

htdvisser: And please update `pkg/identityserver/storetest/client_store.go` to cover this new field.

pkg/oauth/oauth.go
```diff
@@ -291,6 +292,28 @@ func (s *server) Token(w http.ResponseWriter, r *http.Request) {
...
+			ar.UserData.(userData).UserSessionIdentifiers.GetSessionId(),
+		)
+		if err != nil {
+			webhandlers.Error(w, r, err)
+			return
+		}
+		if session == nil {
+			ar.Authorized = false
+		} else {
+			secondsUntilExpiry := int32(ttnpb.StdTime(session.ExpiresAt).Sub(time.Now()).Seconds())
```
htdvisser: We prefer to use `time.Duration` instead of "seconds".

```suggestion
			timeUntilExpiry := time.Until(ttnpb.StdTime(session.ExpiresAt))
			// etc.
```

htdvisser (changes requested):

pkg/webui/lib/store/logics/status.js
kschiffer: It's a quite arbitrary period to prevent setting long-running timeouts. I don't mind hardcoding it here. Do you have a specific concern?

pkg/account/server_test.go
```diff
@@ -29,6 +29,7 @@ import (
 	"github.com/smartystreets/assertions"
 	"github.com/smartystreets/assertions/should"
 	"go.thethings.network/lorawan-stack/v3/pkg/account"
+	sess "go.thethings.network/lorawan-stack/v3/pkg/account/session"
```
htdvisser: Nit: I don't think it's necessary to alias the import here.

pkg/account/user.go
```diff
@@ -56,13 +56,15 @@ func (s *server) CurrentUser(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	webhandlers.JSON(w, r, struct {
-		User       json.RawMessage `json:"user"`
-		LoggedInAt *time.Time      `json:"logged_in_at"`
-		SessionId  string          `json:"session_id"`
+		User             json.RawMessage `json:"user"`
+		LoggedInAt       *time.Time      `json:"logged_in_at"`
+		SessionId        string          `json:"session_id"`
+		SessionExpiresAt *time.Time      `json:"expires_at"`
```
htdvisser:
```suggestion
		SessionExpiresAt *time.Time      `json:"session_expires_at"`
```

pkg/account/user.go
```diff
@@ -56,13 +56,15 @@ func (s *server) CurrentUser(w http.ResponseWriter, r *http.Request) {
...
 	}{
-		User:       userJSON,
-		LoggedInAt: ttnpb.StdTime(session.CreatedAt),
-		SessionId:  session.SessionId,
+		User:             userJSON,
+		LoggedInAt:       ttnpb.StdTime(session.CreatedAt),
+		SessionExpiresAt: ttnpb.StdTime(session.ExpiresAt),
+		SessionId:        session.SessionId,
```
htdvisser: Nit: Prefer to keep the original ordering of the struct fields.

```suggestion
		SessionId:        session.SessionId,
		SessionExpiresAt: ttnpb.StdTime(session.ExpiresAt),
```

pkg/identityserver/gormstore/client.go
```diff
@@ -52,6 +52,8 @@ type Client struct {
 
 	Grants Grants `gorm:"type:INT ARRAY"`
 	Rights Rights `gorm:"type:INT ARRAY"`
+
+	TieAccessToSession bool `gorm:"not null"`
```
htdvisser: If we add a `not null` field, we need to set a `default`, otherwise the migration will fail.

htdvisser:
Just realized that this requires a minor bump, since we're adding fields to the database, which requires a migration, and we only do that in minor bumps. So I'm afraid this will have to wait until v3.20.

pkg/account/server.go
```diff
@@ -50,27 +50,27 @@ type Component interface {
...
+	config        Config
 	store         account_store.TransactionalInterface
 	session       sess.Session
 	generateCSP   func(config *oauth.Config, nonce string) string
 	schemaDecoder *schema.Decoder
 }
 
 // NewServer returns a new account app on top of the given store.
-func NewServer(c *component.Component, store account_store.TransactionalInterface, config oauth.Config, cspFunc func(config *oauth.Config, nonce string) string) (Server, error) {
+func NewServer(c *component.Component, store account_store.TransactionalInterface, config Config, cspFunc func(config *oauth.Config, nonce string) string) (Server, error) {
```
nicholaspcr: I believe it generates a lint error with the new ci, since the line is too big.
If it is of interest, the link to Hylke's comment on how he breaks the methods: https://github.com/TheThingsNetwork/lorawan-stack/pull/5487#discussion_r880555535

pkg/account/config.go
```diff
@@ -0,0 +1,26 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package account
+
+import (
+	sess "go.thethings.network/lorawan-stack/v3/pkg/account/session"
```
nicholaspcr: Nit: `session` instead of `sess`

ryaplots (approved):

johanstokking:
What is the status here?

kschiffer:
Had to lay this aside,since it was consuming too much time fixing CI and wouldn't make the last minor anyways. Planning to get back to this to merge it into v3.21.0

PR #5358, opened 2022-03-31
Add default order to sortable tables

### LINKED ISSUE #4886: Applications list in console should set sorted by ID by default in column header
#### Summary

When you open the applications list in the console, it is sorted in ascending order by ID by default, but the column header sort widget isn't set to show this.

This means to get descending sort you must click the ID column header twice - the first click puts it into ascending order (so is a waste of a click and time), the second click gets you the descending order.

#### Steps to Reproduce

1. Open the applications list in the console.
2. Note the application IDs are in ascending order but the header does not show the table is sorted.
3. Click the ID column header and see the table redraws, but in the same order and the header now shows the sort order.
4. Click a second time and you get the table in descending ID order.

My guess it other tables in the console suffer the same problem.

The table sort order should be set to ascending by default and the first click on the ID column reverses the sort order.

ryaplots:
#### Summary

Closes #4886 

#### Changes

- The tables that are sortable are now sorted by default with a new prop

#### Testing

Manually

#### Notes for Reviewers

I could try and derive the name of the column with is initially sorted, but that seemed like a huge work around. When adding a new prop it is possible to choose with column is the table sorted by. 

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Added
 
 ### Changed
-
+- Sortable tables are now sorted by default in the Console
 ### Deprecated
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 - Sortable tables are now sorted by default in the Console
+
 ### Deprecated
 
 ### Removed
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -12,7 +12,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Added
 
 ### Changed
-
+- Sortable tables are now sorted by default in the Console
```
kschiffer, line 15:
```suggestion
- Sortable tables are now sorted by default in the Console

```

pkg/webui/containers/fetch-table/index.js
```diff
@@ -302,7 +304,6 @@ class FetchTable extends Component {
     } = this.props
     const { page, query, tab, order, initialFetch, error } = this.state
     let orderDirection, orderBy
-
```
kschiffer: Please remove this change.

kschiffer (approved):

PR #5383, opened 2022-04-13
Make the OAuth client useful in server-to-server scenarios

johanstokking:
#### Summary

This makes the OAuth client (`pkg/web/oauthclient`) useful for server-to-server scenarios.

Needed for https://github.com/TheThingsIndustries/lorawan-stack/issues/3161

#### Changes

- Expose a method to get a fresh token, extracted from a HTTP handler that returned the token to web clients
- Expose the logout flow, extracted from a HTTP handler that did the logout and redirect
- Cleaned up some artifacts from removing Echo; the Echo context was replaced by the HTTP response writer and request, while what was only necessary was the actual (request) context
- Other minor code cleanups

#### Testing

Integration testing

##### Regressions

None expected

#### Checklist

### DIFF
--- a/pkg/web/oauthclient/callback.go
+++ b/pkg/web/oauthclient/callback.go
@@ -102,7 +102,7 @@ func (oc *OAuthClient) HandleCallback(w http.ResponseWriter, r *http.Request) {
 		webhandlers.Error(w, r, err)
 		return
 	}
-	conf, err := oc.oauth(w, r)
+	conf, err := oc.oauth(r.Context())
 	if err != nil {
 		webhandlers.Error(w, r, err)
 		return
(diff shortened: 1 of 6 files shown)


pkg/web/oauthclient/callback.go
```diff
@@ -134,7 +134,10 @@ func (oc *OAuthClient) HandleCallback(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := oc.callback(w, r, token, stateCookie.Next); err != nil {
-		webhandlers.Error(w, r, errExchange.WithCause(err))
+		if _, ok := errors.From(err); !ok {
+			err = errExchange.WithCause(err)
+		}
```
johanstokking, line 139: The default callback never returns an error. So this errors only when there's a custom callback configured. In that case, return the error without wrapping it in an "token exchange refused" permission denied error. The callback is not about the token exchange, and the callback may fail because of other reasons than permission denied.

kschiffer (approved):
LGTM

PR #5386, opened 2022-04-13
Fix gateway status

ryaplots:
#### Summary

This PR fixes the Gateway status being not visible in the Gateways list.

#### Changes

- Removed the getHostFromUrl()

#### Testing

Open gateways list

#### Checklist

### DIFF
--- a/pkg/webui/lib/host-from-url.js
+++ b/pkg/webui/lib/host-from-url.js
@@ -26,6 +26,6 @@ export default url => {
 
     return new URL(`http://${url}`).hostname
   } catch (error) {
-    return undefined
+    return url
   }
 }
--- a/pkg/webui/lib/host-from-url.js
+++ b/pkg/webui/lib/host-from-url.js
@@ -19,8 +19,19 @@
  * @returns {string?} - The hostname of the `url` or undefined.
  */
 export default url => {
+  const IPv6RegexExp =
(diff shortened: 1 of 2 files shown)


adriansmares:
Does everything still work if you edit the gateway server address to `localhost:1234` ? That was the use case here

ryaplots:
> Does everything still work if you edit the gateway server address to `localhost:1234` ? That was the use case here

If I do that the status of all gateways appears as "Other cluster"

adriansmares:
The problem is that `getHostFromUrl` won't correctly parse `gateway.gateway_server_address` since it is not really a fully fledged URL. I think the correct solution here is to adapt `getHostFromUrl` to match the logic that we use in the CLI:

https://github.com/TheThingsNetwork/lorawan-stack/blob/911078b6cde10c3d2c319efcd3e90e47355332aa/cmd/ttn-lw-cli/commands/utils.go#L31-L45

Basically `new URL` should be used only if a scheme is present (that's what the `://` check does). Note that splitting between host and port should be handled with care as you cannot just use `:` since `:` appears in IPv6 addresses. I'm not sure what the correct equivalent for `net.SplitHostPort` is in JavaScript.

ryaplots:
> The problem is that `getHostFromUrl` won't correctly parse `gateway.gateway_server_address` since it is not really a fully fledged URL. I think the correct solution here is to adapt `getHostFromUrl` to match the logic that we use in the CLI:
> 
> https://github.com/TheThingsNetwork/lorawan-stack/blob/911078b6cde10c3d2c319efcd3e90e47355332aa/cmd/ttn-lw-cli/commands/utils.go#L31-L45

You are right indeed. I removed it because I remember there was a PR that added the `getHostFromUrl` in a lot of places where it wasn't before. So I presumed this was one of those places and it does not work here. But indeed extending it would make much more sense. I am on it

ryaplots:
@adriansmares I did some research and couldn't find an equivalent to `net.SplitHostPort` besides handling the actual `location` or using `new URL()`. So what I did was match the string to a `regex` to handle IPv6 addresses. And after that if the url contains a `:` I can add `://` before and then `new URL()` does not return undefined anymore. I tested this with URLs like 'tti.staging1.cloud.thethings.industries', 'localhost', 'localhost:1234' and 'https://example' and it worked.

pkg/webui/lib/host-from-url.js
```diff
@@ -19,8 +19,19 @@
...
+    if (url.match(IPv6RegexExp)) {
+      return url
+    } else if (url.includes('://')) {
+      return new URL(url).hostname
+    } else if (url.includes(':')) {
+      return new URL(`https://${url}`).hostname
+    }
+
+    return url
   } catch (error) {
```
kschiffer:
```suggestion
  try {
    if (url.matches(/^[a-zA-Z0-9]+:\/\/.*/)) {
      return new URL(url).hostname
    }
    return new URL(`http://${url}`).hostname
  } catch (error) {
    return url
  }
```

Let's simplify this.

kschiffer (changes requested):

pkg/webui/lib/host-from-url.js
```diff
@@ -20,7 +20,11 @@
...
 export default url => {
   try {
-    return new URL(url).hostname
+    if (url.match(/^[a-zA-Z0-9]+:\/\/.*/)) {
+      return new URL(url).hostname
+    }
+
+    return new URL(`http://${url}`).hostname
   } catch (error) {
     return undefined
```
kschiffer:
```suggestion
    return url
```

I think that's more resilient.

kschiffer (approved):

adriansmares (approved):
I'll backport this one into `rc1` everywhere.

PR #5412, opened 2022-04-26
Decode downlink when scheduling binary payload

### LINKED ISSUE #5260: Decode downlink when scheduling binary payload
#### Summary

Decode downlink when scheduling binary payload. This is for better consistency.

References https://github.com/TheThingsIndustries/lorawan-stack-docs/issues/710

#### Why do we need this?

When a downlink decoder is present but payload gets scheduled in its binary form, it should be presented in its decoded form in the downlink queued event.

Downlink is already decoded when it's received (encrypted) from the Network Server, e.g. on downlink sent, failed, ack, nack events.

Decoded downlink in all downlink events, including the downlink queued event that is only published by Application Server.

When scheduling downlink (push or replace), decode downlink.

While we're at it, remove the downlink queued event propagation from Network Server as Network Server doesn't send that message.

UT

Yes

johanstokking:
#### Summary

Closes #5260 

#### Changes

- Decode downlink when binary payload gets scheduled
- Remove processing downlink queued event from NS as it's never sent

#### Testing

Seeing the downlink being decoded is a bit hard to test with unit tests, as tests currently don't observe the "downlink queued" event when queuing downlink; the tests observe the entire downlink queue and then the downlink was also decoded.

So ran locally:

```bash
$ ttn-lw-cli dev downlink push app1 dragino1 --frm-payload DEADBEEF --f-port 42
```

The downlink queued event has the result of the downlink decoder (which sums the bytes):

Screen Shot 2022-04-26 at 10 45 17

##### Regressions

None expected

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
+- Application Server now decodes downlink if a downlink decoder is present and binary payload is scheduled.
+
 ### Deprecated
 
 ### Removed
--- a/pkg/applicationserver/applicationserver.go
+++ b/pkg/applicationserver/applicationserver.go
@@ -787,8 +787,6 @@ func (as *ApplicationServer) handleUp(ctx context.Context, up *ttnpb.Application
 		return true, as.handleUplink(ctx, up.EndDeviceIds, p.UplinkMessage, link)
 	case *ttnpb.ApplicationUp_DownlinkQueueInvalidated:
 		return as.handleDownlinkQueueInvalidated(ctx, up.EndDeviceIds, p.DownlinkQueueInvalidated, link)
-	case *ttnpb.ApplicationUp_DownlinkQueued:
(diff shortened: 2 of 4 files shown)


adriansmares (approved):
Cool.

Should we deprecate `downlink_queued` completely at some point in the future ? Or should we emit it when a downlink queue operation occurs ?

johanstokking:
> Should we deprecate `downlink_queued` completely at some point in the future ? Or should we emit it when a downlink queue operation occurs ?

It is being emitted by AS and it stays that way. That is very useful for applications to observe downlink being queued by other applications. This change contributes to that; if one application schedules binary payload, the other receives it also in decoded for now.

What has been removed is the downlink queued upstream event handling from NS to AS. That shouldn't have been there, because AS emits these events, not NS.

Or do I misunderstand your question?

adriansmares:
I misunderstood the behavior - I didn't see that we generate them in `registerForwardDownlinks` (`observability.go`) when I wrote the question. 

All clear now.

PR #5415, opened 2022-04-28
cli: Use cache for version check

### LINKED ISSUE #5394: Cache result of version check
#### Summary

`ttn-lw-cli` should cache the result of the version check on disk, so that it doesn't have to do version check requests on _every_ call.

#### Why do we need this?

I'm working on a script that makes a batch of 100 devices claimable and generates QR codes for them. If the version check is enabled (which it is by default), I'm making 200 calls to the version check API, which is excessive.

#### Environment

ttn-lw-cli v3.18 on macOS

Cache a successful version check response, together with a timestamp, in `os.UserCacheDir`. Only do a version check (1) if that file does not exists or (2) if the timestamp is too old.

michalborkowski96:
#### Summary
Closes #5394 

#### Changes
Made version check use cache

#### Testing
1. Run one time - information about version check. Run second time - no information.
2. Edit the cache file to an early date and run - version check gets performed.
3. Edit the cache file to make it incorrect and run - warning shows up, then version check gets performed, subsequent run doesn't show version check information.

##### Regressions
None

#### Notes for Reviewers
Maybe instead of coding the cache logic in `pkg/version/remote.go`, it'd be better to create an all-purpose caching interface?

#### Checklist

### DIFF
--- a/pkg/version/remote.go
+++ b/pkg/version/remote.go
@@ -19,6 +19,8 @@ import (
 	"encoding/json"
 	"io"
 	"net/http"
+	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -26,6 +28,10 @@ import (
 	"github.com/blang/semver"
 	"go.thethings.network/lorawan-stack/v3/pkg/errors"
 	"go.thethings.network/lorawan-stack/v3/pkg/log"
+)
+
+const (
+	versionCheckCacheTTL = 11 * time.Hour // Cache is valid for one hour less than component.version_check.versionCheckPeriod.
 )
 
 type timestamp struct {
@@ -103,6 +109,11 @@ type Update struct {
 	Minor, Patch bool
 }
 
+type cachedUpdate struct {
+	Result    *Update
+	Timestamp timestamp
+}
+
 type checkOptions struct {
 	client  *http.Client
 	current semver.Version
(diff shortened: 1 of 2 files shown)


htdvisser (commented):
Could you explain why this was not implemented as in the proposal in the issue? Why don't we store the actual _result_ of the version check request, and use that if it's not too old?

michalborkowski96:
@htdvisser can you explain the advantages of your approach?

htdvisser:
The advantage of my approach is that we can print warnings without having to make HTTP requests all the time.

adriansmares:
To add to the difference - if you don't cache the response, you can't say that the CLI is old - you only know that a check occured. You would do the check and report it only every TTL. But if you cache the response, on each command run, you can report that an update is available, without doing a version check request.

pkg/version/remote.go
```diff
@@ -142,9 +153,64 @@ func WithReference(current semver.Version) CheckOption {
 	})
 }
 
+// cachedCheck returns previously saved result from cache
+func cachedCheck() (*cachedUpdate, error) {
+	cacheDir, err := os.UserCacheDir()
+	if err != nil {
+		return nil, err
+	}
+	cacheFile := filepath.Join(cacheDir, "tts", "version-check")
```
htdvisser: For the sake of DRY, maybe pull this into a little func? Then you can also fall back to `os.TempDir` on error.

You can ever consider making it a var:

```go
var cacheFile = func() string {
	cacheDir, err := os.UserCacheDir()
	if err != nil {
		cacheDir = os.TempDir()
	}
	return filepath.Join(cacheDir, "tts", "version-check")
}()
```

pkg/version/remote.go
```diff
@@ -28,6 +30,10 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/log"
 )
 
+const (
+	versionCheckCacheTTL = 11 * time.Hour // cache is valid for one hour less than component.version_check.versionCheckPeriod
```
htdvisser: Comments are sentences; they start with an uppercase letter and end with a period. More below.

pkg/version/remote.go
```diff
@@ -142,9 +153,64 @@ func WithReference(current semver.Version) CheckOption {
...
+// The function uses cache.
 func CheckUpdate(ctx context.Context, opts ...CheckOption) (*Update, error) {
+	cachedCheck, err := cachedCheck()
+	if err != nil {
+		log.FromContext(ctx).WithError(err).Warn("Failed to access cache for version check")
+	} else {
+		if cachedCheck.Timestamp.Add(versionCheckCacheTTL).After(time.Now()) {
+			return cachedCheck.Result, nil
+		}
+	}
```
htdvisser:
```suggestion
	} else if cachedCheck.Timestamp.Add(versionCheckCacheTTL).After(time.Now()) {
		return cachedCheck.Result, nil
	}
```

pkg/version/remote.go
```diff
@@ -173,13 +239,18 @@ func CheckUpdate(ctx context.Context, opts ...CheckOption) (*Update, error) {
...
+	update := &Update{
 		Current: o.current,
 		Latest:  latest,
 		DocsURL: o.docsURL,
 		Minor:   minor,
 		Patch:   patch,
-	}, nil
+	}
+	err = saveCheck(update)
+	if err != nil {
```
htdvisser: Nit:

```suggestion
	if err = saveCheck(update); err != nil {
```

htdvisser (approved):
LGTM

PR #5419, opened 2022-04-29
Add `last_seen_at` support and automatically update session info

### LINKED ISSUE #5387: Add support for device last seen timestamp in the console
#### Summary

With https://github.com/TheThingsNetwork/lorawan-stack/pull/5290 the devices now have a `last_seen_at` timestamp in the IS db. We need to show this in the console.

#### Why do we need this?

To inform users of the last time their device was observed in AS. This can be used as additional debug data.

`last_seen_at` timestamp field in end devices.

Console support.

#### Environment

v3.19

I suppose it should be implemented in a similar way to gateway last seen.

E2E

I suppose I could but I don't think I have time.

ryaplots:
#### Summary

Closes #5387.

#### Changes

- Add `last_seen_at` field mask in `getDevice`;
- Use `last_seen_at` in the initial fetch of the end device.

#### Testing

E2E and manually

#### Notes for Reviewers

Is this -> [`session.started_at` issue](https://github.com/TheThingsNetwork/lorawan-stack/issues/4766) still relevant now that we only use `last_seen_at`? @kschiffer 

#### Checklist


CHANGELOG.md
```diff
@@ -10,7 +10,8 @@ For details about compatibility between different releases, see the **Commitment
 ## [Unreleased]
 
 ### Added
-
+- Support for end device `last_seen_at` timestamp in the Console.
```
```suggestion

- Support for end device `last_seen_at` timestamp in the Console.
```

pkg/webui/console/containers/devices-table/index.js
```diff
@@ -172,8 +172,14 @@ class DevicesTable extends React.Component {
...
-        ['name', 'application_server_address', 'network_server_address', 'join_server_address'],
-        { withLastSeen: true, withStatus: true },
+        [
+          'name',
+          'application_server_address',
+          'network_server_address',
+          'join_server_address',
+          'last_seen_at',
+        ],
+        { withStatus: true },
```
kschiffer:
```suggestion
```
We're not using that anymore.

pkg/webui/console/store/middleware/logics/devices.js
```diff
@@ -128,11 +94,25 @@ const getDeviceTemplateFormatsLogic = createRequestLogic({
...
 
+const getDeviceSessionLogic = createLogic({
+  type: devices.GET_DEVICE_EVENT_MESSAGE_SUCCESS,
+  process: async ({ action }, dispatch, done) => {
+    const { event, id } = action
+    const appId = id.application_ids.application_id
+
+    if (event.name === 'as.up.join.forward') {
+      dispatch(devices.getDevice(appId, id.device_id, ['session']))
+      done()
```
kschiffer:
```suggestion
```
I believe you don't need to use `done()` when doing only one dispatch.

pkg/webui/console/store/middleware/logics/devices.js
```diff
@@ -128,11 +94,25 @@ const getDeviceTemplateFormatsLogic = createRequestLogic({
...
+    }
```
kschiffer, line 133: I think it would be nice to actually send a toast notification when the device joins as well. You can dispatch the toast from within the logic:
```js
        toast({
          title: id.device_id,
          message: m.joinSuccess,
          type: toast.types.INFO,
        })
```
> The device has successfully joined the network

pkg/webui/console/store/reducers/devices.js
```diff
@@ -91,10 +91,7 @@ const devices = (state = defaultState, { type, payload, event }) => {
 
       // Update derived last seen value if possible.
       const derived = {}
-      const lastSeen = getLastSeen(payload)
-      if (lastSeen) {
-        derived.lastSeen = lastSeen
-      }
+      derived.lastSeen = payload.last_seen_at
```
kschiffer: Since it is not a derived value anymore, we can remove `lastSeen` completely from the derived store. Just make sure that no other logic is dependent on `lastSeen` anymore (in which case it should use the `last_seen_at` in the device store directly).

Check the `selectDeviceDerivedLastSeen` selector which should be removed and its calls should be replaced with a new selector `selectDeviceLastSeen` which will get the value from the entity directly.

kschiffer (changes requested):
Please also update the device reducer that updates the last seen value, so that the value is updated on the device record directly and not using the value within `derived`. The `derived` store should not store any last seen value anymore.
`pkg/webui/console/store/reducers/devices.js:203`

kschiffer:
> Is this -> [`session.started_at` issue](https://github.com/TheThingsNetwork/lorawan-stack/issues/4766) still relevant now that we only use `last_seen_at`? @kschiffer 

The underlying issue of component priority when merging the device record in the SDK is still present, so that still needs to be fixed. It is however not making any problems currently.

pkg/webui/console/store/reducers/devices.js
ryaplots: I wanted to indeed remove `lastSeen` from the derived state, but then I realized that it is constantly being changed by each event that comes in [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/14ebc66bff863ad1685f6601100b53aa27fdf9b8/pkg/webui/console/store/reducers/devices.js#L208). So if we remove `lastSeen` in this situation, we'd need to update the `last-seen-at` directly in the device, would that be okay - to constantly change a device property?

kschiffer: Yes, that's fine. We're modifying the end device but effectively only by recreating live how the end device record is updated in the backend anyway. It also makes implementation a lot easier than keeping track of a second derived value. There's a slight chance of having a mismatch with our updated value and the actual value but I think that is offset by the advantages of having a more dynamic end device record in the store.

kschiffer:
I've added a couple of commits on top to fix live session updates and used the opportunity to clean up the end device overview a bit.

Just using `as.up.join.forward` to trigger a session request is not enough since this will be at the start of the join procedure when no session is yet set. The session is only deemed as established once the first uplink was received until then we have a `pending_session` only.

So I made some updates to also show pending session info in the overview and to check for the right moment to update the device session.

kschiffer (approved):

PR #5420, opened 2022-04-29
Gateway Server connection stats event and event type

johanstokking:
#### Summary

Gateway Server groundwork for https://github.com/TheThingsIndustries/lorawan-stack/issues/3161

#### Changes

- Add new `gs.gateway.connection.stats` event that publishes connection stats changes, besides being persisted in the stats registry
- Increased the debounce time; writing and publishing stats every 5 seconds is too often
- Change event data type of `gs.up.receive` to `GatewayUplinkMessage` instead of `UplinkMessage`

#### Testing

Local testing

Console event row:

Screen Shot 2022-04-29 at 18 09 30

##### Regressions

None expected, unless clients rely on the data type of `gs.up.receive`.

#### Notes for Reviewers

@htdvisser @kschiffer  @adriansmares @pgalic96 no need to review

@KrishnaIyer @nicholaspcr  please review GS
@ryaplots please review Console

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -649,9 +649,6 @@ func (gs *GatewayServer) startHandleUpstreamTask(conn connectionEntry) {
 }
 
 func (gs *GatewayServer) startUpdateConnStatsTask(conn connectionEntry) {
-	if gs.statsRegistry == nil {
-		return
-	}
 	conn.tasksDone.Add(1)
 	gs.StartTask(&task.Config{
 		Context: conn.Context(),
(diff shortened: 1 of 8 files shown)


nicholaspcr (approved):
LGTM

KrishnaIyer (commented):
Changes look good but I'm not so sure of the event name. From what I see, we generally use 
```
{component}.{entity}....{action}
```
The last part is usually an action (verb) ex; receive, update, fail etc.

We could use something along the lines of
```
gs.gateway.connection.stats.update
```
Or hyphenated if that's preferred
```
gs.gateway-connection-stats.update
```

htdvisser (approved):
Code-owned changes LGTM

johanstokking:
> Changes look good but I'm not so sure of the event name. From what I see, we generally use
> 
> ```
> {component}.{entity}....{action}
> ```
> 
> The last part is usually an action (verb) ex; receive, update, fail etc.
> 
> We could use something along the lines of
> 
> ```
> gs.gateway.connection.stats.update
> ```
> 
> Or hyphenated if that's preferred
> 
> ```
> gs.gateway-connection-stats.update
> ```

Yeah it's a type of event that we don't really have anywhere else. It's more like a heartbeat. It's not like anything happens that we have a verb for it. Because of the debouncing, it's not that the stats are updated (they are updated more often), plus if the stats didn't change this event is still fired (while an "update" implies a content change). 

I think it should start with `gs.gateway.` because we also have `gs.gateway.{connect|disconnect}`. I made this `gs.gateway.connection.stats` to clarify it relates to the connection. Also with a group `connection` we can add other connection events in the future, like `upgrade` or `ping` or `log` or whatever LoRa Alliance GW-NS task force comes up with.

KrishnaIyer (approved):
I see. OK clear then

pkg/gatewayserver/gatewayserver.go
```diff
@@ -649,9 +649,6 @@ func (gs *GatewayServer) startHandleUpstreamTask(conn connectionEntry) {
 }
 
 func (gs *GatewayServer) startUpdateConnStatsTask(conn connectionEntry) {
```
adriansmares, line 651: Why does this check need to be removed ?

If the caching is not enabled at stack level (Redis caching specifically), `gs.statsRegistry` will be `nil` and the updates will panic.

johanstokking:
@ryaplots can you review please? It's very minor

ryaplots (approved):
Code owned LGTM

PR #5435, opened 2022-05-04
Get NS address from TTJS configuration

### LINKED ISSUE #5404: Configure NS address while claiming in DCS config
#### Summary

Configure NS address while claiming in DCS config

#### Why do we need this?

There is no real need for users to provide this field during claiming. The DCS will already deployed to each Cluster and hence knows to which NS to set for the device in TTJS.

`networkserveraddress` is part of the claiming API.

Add an option to get this from the TTJS config and deprecate this field in the API.

#### Environment

TTS v3.19.1

I think this can be easily done in the configuration. We currently have `home-ns-ids` fields as a map whose key is the Home NS.
```
home-ns-ids:
  thethings.localhost: "0000000000000000"
```

Instead of this being dynamically filled at config time and compared to the `networkserveraddress` from the client request, we just set the correct NSID to the NS and use that directly.

So there will be only one key in this map for each cluster and the value will be the globally agreed up hNSID for that Home NS.

Local + staging.

Yes cc @johanstokking / @kschiffer

KrishnaIyer:
#### Summary

Closes #5404 

#### Changes

- Fetch NS address and HomeNSID from config instead of ascertaining it from user input.
- Add new config and validate it.
- Update interface
- Update tests

#### Testing

Local

##### Regressions

NA

#### Notes for Reviewers

- This technically counts as breaking configuration but no one else currently uses the DCS for claiming with TTJS and this only exists in staging for now. So I think this is fine.

#### Checklist

### DIFF
--- a/pkg/deviceclaimingserver/enddevices/ttjs/ttjs.go
+++ b/pkg/deviceclaimingserver/enddevices/ttjs/ttjs.go
@@ -21,8 +21,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
+	"regexp"
 
 	"go.thethings.network/lorawan-stack/v3/pkg/cluster"
 	"go.thethings.network/lorawan-stack/v3/pkg/config"
@@ -41,16 +43,22 @@ type BasicAuth struct {
 	Password string `yaml:"password"`
 }
 
+// NS contains information related to the Network Server
+type NS struct {
+	Address  string `yaml:"address"`
+	HomeNSID string `yaml:"home-ns-id"`
+}
+
 // Config is the configuration to communicate with The Things Join Server End Device Claming API.
 type Config struct {
 	NetID           types.NetID         `yaml:"-"`
 	JoinEUIPrefixes []types.EUI64Prefix `yaml:"-"`
 
 	BasicAuth          `yaml:"basic-auth"`
-	ClaimingAPIVersion string            `yaml:"claiming-api-version"`
-	URL                string            `yaml:"url"`
-	TenantID           string            `yaml:"tenant-id"`
-	HomeNSIDs          map[string]string `yaml:"home-ns-ids"`
+	ClaimingAPIVersion string `yaml:"claiming-api-version"`
+	URL                string `yaml:"url"`
+	TenantID           string `yaml:"tenant-id"`
+	NS                 NS     `yaml:"ns"`
 }
 
 // Component abstracts the underlying *component.Component.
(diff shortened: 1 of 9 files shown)


nicholaspcr (approved):
LGTM. Since TTJS its only on staging I don't see much of a problem with the change in the API.

pkg/deviceclaimingserver/enddevices/ttjs/ttjs.go
```diff
@@ -65,15 +73,18 @@ type Component interface {
...
 	Component
 
-	hsNSIDs     map[string]types.EUI64
 	httpClient  *http.Client
 	baseURL     *url.URL
 	config      Config
 	ttiVendorID OUI
 }
 
+var nsAddressRegexp = regexp.MustCompile("^(?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*(?:[A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])(?::[0-9]{1,5})?$")
```
johanstokking: Hmm this isn't enforced by TTJS

pkg/deviceclaimingserver/enddevices/ttjs/ttjs.go
```diff
@@ -65,15 +73,18 @@ type Component interface {
...
+
+var errInvalidNSAddress = errors.DefineInvalidArgument("invalid_ns_address", "invalid NS address `{address}`")
```
johanstokking: Let's spell out Network Server

pkg/deviceclaimingserver/enddevices/ttjs/ttjs.go
```diff
@@ -82,22 +93,33 @@ func (config Config) NewClient(ctx context.Context, c Component) (*TTJS, error)
...
-		res[nsAddress] = nsID
+
+	// External services, including Join Servers, typically identify Network Servers by host instead of by host and port.
+	// So we need to drop the port here if provided. We first need to check that the configured address is valid before splitting the port.
+	s := nsAddressRegexp.FindStringSubmatch(config.NS.Address)
+	if len(s) != 1 {
+		return nil, errInvalidNSAddress.WithAttributes("address", config.NS.Address)
 	}
+	address, _, err := net.SplitHostPort(s[0])
+	if err != nil {
```
johanstokking: Why is the validation needed?

pkg/deviceclaimingserver/enddevices/ttjs/ttjs.go
```diff
@@ -41,16 +43,22 @@ type BasicAuth struct {
 	Password string `yaml:"password"`
 }
 
+// NS contains information related to the Network Server.
+type NS struct {
+	Address  string `yaml:"address"`
+	HomeNSID string `yaml:"home-ns-id"`
```
johanstokking: Since we're deploying DCS in each cluster, why isn't this DCS component config?

pkg/deviceclaimingserver/enddevices/ttjs/ttjs.go
KrishnaIyer: Well this needs to be validated somewhere. If you add that to the TTJS, I can drop it here.

KrishnaIyer: Or you mean the `host:port` splitting thing? That may not be necessary if document to use only a hostname and omit the port.

pkg/deviceclaimingserver/enddevices/ttjs/ttjs.go
KrishnaIyer: I see it. Since the NSID or the Address doesn't change regardless of the upstream, I think this indeed can be added to the DCS config.

KrishnaIyer:
I've made the config simple and dropped the validation here. I assume that operators know not to use a port in the `hostname` field and I think it's sufficient to validate this in TTJS.

pkg/deviceclaimingserver/enddevices/ttjs/ttjs.go
```diff
@@ -45,12 +45,14 @@ type BasicAuth struct {
 type Config struct {
 	NetID           types.NetID         `yaml:"-"`
 	JoinEUIPrefixes []types.EUI64Prefix `yaml:"-"`
+	NetworkServer   struct {
+		Hostname, HomeNSID string
```
johanstokking: At this level, declare home NSID as `types.EUI64`

johanstokking (commented):
LGTM but please don't use the anonymous struct in TTJS

johanstokking (approved):

PR #5445, opened 2022-05-09
Include entity IDs in API key query if available

htdvisser:
#### Summary

This updates the IS EndDevice store interface to add a method for getting an API key by _both_ entity ID and API key ID. In most places where we get the API key, we have both available, so we should query by both.

Refs https://github.com/TheThingsIndustries/lorawan-stack-support/issues/783

#### Changes

- Duplicate `GetAPIKey` into `GetAPIKey` and `GetAPIKeyByID`, and only use `GetAPIKeyByID` when no entity IDs available.

#### Testing

Updated unit tests

##### Regressions

nil

#### Notes for Reviewers

While this is a security issue, I don't think it's severe enough to require a patch release before the next scheduled one. @johanstokking ACK?

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Security
 
+- Security fix for an issue where the description and list of rights of arbitrary API keys could be retrieved by any logged-in user if the 24-bit random API key ID was known.
+
 ## [3.19.1] - 2022-05-04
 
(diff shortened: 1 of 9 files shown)


johanstokking:
> While this is a security issue, I don't think it's severe enough to require a patch release before the next scheduled one. @johanstokking ACK?

ACK

nicholaspcr (approved):
LGTM

PR #5452, opened 2022-05-11
Add schedule attempt and failure events

johanstokking:
#### Summary

Groundwork for https://github.com/TheThingsIndustries/lorawan-stack/issues/3161

#### Changes

- Add `gs.down.{schedule,fail}.attempt` events and `downlink_schedule_{attempted,failed}_total` metrics, analoguous to Network Server
- Change `gs.down.send` event data to the actually sent downlink message (with scheduled payload instead of request payload)
- Add `gs.update-connection-stats-interval` config (default 1 minute) which is the minimum interval that connection stats are updated / events are sent. This is usually not necessary, but for LBS gateways not receiving status messages and not receiving messages, this is used to refresh stats and notify upstream services that the gateway is still connected

#### Testing

CI

##### Regressions

Clients that depend on the data structure of `gs.down.send` may fail if they expect scheduled payload. I don't think this is an issue and it isn't covered by our compatibility commitment.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -875,8 +875,6 @@ func (gs *GatewayServer) updateConnStats(ctx context.Context, conn connectionEnt
 		ConnectedAt: ttnpb.ProtoTimePtr(connectTime),
 		Protocol:    conn.Connection.Frontend().Protocol(),
 	}
-	refreshTTLTimer := time.NewTicker(gs.config.ConnectionStatsTTL / 2)
-	defer refreshTTLTimer.Stop()
 
 	// Initial update, so that the gateway appears connected.
 	registerGatewayConnectionStats(ctx, *ids, stats)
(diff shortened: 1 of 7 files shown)


pkg/gatewayserver/gatewayserver.go
```diff
@@ -917,7 +917,7 @@ func (gs *GatewayServer) updateConnStats(ctx context.Context, conn connectionEnt
 				case <-time.After(duration):
 				}
 			}
-		case <-refreshTTLTimer.C:
+		case <-updateStatsTimer.C:
```
adriansmares: I think we should reset the ticker after exiting the `select` here, in order to avoid having two stats updates in rapid succession (`conn.StatsChanged()` triggers at second 59, then 1 second later `updateStatsTimer.C` triggers again).

pkg/gatewayserver/grpc_nsgs.go
```diff
@@ -72,17 +72,22 @@ func (gs *GatewayServer) ScheduleDownlink(ctx context.Context, down *ttnpb.Downl
...
 		if err != nil {
 			logger.WithField("gateway_uid", uid).WithError(err).Debug("Failed to schedule on path")
 			pathErrs = append(pathErrs, errSchedulePath.WithCause(err).WithAttributes("gateway_uid", uid))
+			registerScheduleDownlinkFail(ctx, conn.Gateway(), err, conn.Frontend().Protocol())
 			continue
 		}
+
 		ctx = events.ContextWithCorrelationID(ctx, events.CorrelationIDsFromContext(conn.Context())...)
 		down.CorrelationIds = append(down.CorrelationIds, events.CorrelationIDsFromContext(ctx)...)
 		registerSendDownlink(ctx, conn.Gateway(), down, conn.Frontend().Protocol())
```
adriansmares: `connDown` instead of `conn`, right ?

johanstokking: Yeah the actual code is in a local stash, thanks

pkg/gatewayserver/gatewayserver.go
johanstokking: Yeah good point. I reworked it a bit. We were waiting on the debounce time even if the last time was long ago, which isn't really necessary.

The new code should update anywhere between `UpdateConnectionStatsInterval` and `UpdateConnectionStatsDebounceTime`.

adriansmares (approved):

PR #2339, opened 2020-04-10
Fix FOpts encryption and decryption in LoRaWAN 1.1 frames

sidpat:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/2171
The encryption scheme for FOpts is slightly different from that of the frame-payload, in the sense that the last byte in the one and only A block of FOpts is 0x00 and in the case of frame-payload the last byte in Ai block is 0x01,0x02,0x03...k, where k is the total number of blocks in frame-payload. Currently FOpts is considered as the 1st Ai block which meant the last byte in its A block was 0x01 rather than 0x00.

image

#### Changes

- Added one more boolean argument to `encryptMessage` function to see if its used for FOpts or frame-payload encryption and modified the function
- Refactored function accordingly in other files
- Added test scenario for FOpts encryption
- (@rvolosatovs) Fix uplink FOpts handling
- (@rvolosatovs) Fix data downlink MIC computation in `simulate`
- (@rvolosatovs) Add support for decoding `FOpts` in `simulate`

#### Notes for Reviewers

Feel free to make changes as required

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, database and configuration, according to the stability commitments in `README.md`.
- [x] Testing: The changes are covered with unit tests. The changes are tested manually as well.
- [ ] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #5460, opened 2022-05-16
Support FOpts encryption usage errata

### LINKED ISSUE #5360: Add support for the FOpts, FCntDwn erratum to LoRaWAN 1.1.0
#### Summary
I have been trying to connect a LoRaWAN 1.1.0 device to TTN. The device is based on LoRaMac-node v4.6.0 which implements LoRaWAN 1.1.0 together with the [FOpts,FCntDwn erratum](https://resources.lora-alliance.org/technical-specifications/fopts-encryption-usage-of-fcntdwn-errata-on-the-lorawan-l2-1-1-specification). The erratum fixes a security vulnerability in LoRaWAN 1.1.0. In order to do so, it changes the encryption algorithm for the FOpts header.

Since TTN v3.18.2 and earlier does not seem to implement the erratum, MAC commands transmitted via FOpts do not work. The device can join successfully and application layer communication works, but MAC commands transmitted via FOpts (ADR, LinkCheck, DeviceStatus, etc..) are ignored on both the device and the TTN network server. This is due to the mutually incompatible FOpts encryption scheme. When either end-point tries to decrypt FOpts, it gets bogus MAC commands and ignores the entire header.

I have been debating whether to submit this as a bug or an improvement request. Since the erratum is four years old at this point and critical for secure LoRaWAN 1.1.0 operation, I consider this a bug. In my opinion, TTN network server should implement the erratum when LoRaWAN 1.1.0 is selected for a device.

The erratum will be probably rolled into LoRaWAN 1.1.1 when it is published. However, I would encourage the TTN team to consider implementing the fix proposed in the erratum even before LoRaWAN 1.1.1 is published.

CC'ing @johanstokking based on an earlier discussion in a different thread.

#### Steps to Reproduce

1. Provision a new LoRaWAN 1.1.0 device in TTN
2. Use [LoRaMac-node v4.6.0](https://github.com/Lora-net/LoRaMac-node) on the device
3. Join TTN in OTAA mode (this will work)
4. Send a LinkCheckReq to TTN (this will fail)

Not very easy to reproduce I admit, unless you already use LoRaMac-node, but please feel free to reach out to me if you need help reproducing it.

#### What do you see now?

1. The device sends a LinkCheckReq encoded in FRMPayload. TTN NS correctly processes this.
2. TTN NS generates LinkCheckAns and transmits the answer encrypted via FOpts
3. The device decrypts FOpts, but gets bogus MAC commands because it uses a different encryption scheme
4. The entire FOpts header is ignored by the device

The same thing happens in the opposite direction as well, i.e., TTN NS ignores encrypted FOps headers from the device. MAC commands transmitted via FRMPayload work.

I would like the TTN network server to use the [FOpts, FCntDown erratum](https://resources.lora-alliance.org/technical-specifications/fopts-encryption-usage-of-fcntdwn-errata-on-the-lorawan-l2-1-1-specification) when operating in LoRaWAN 1.1.0 mode.

#### Environment
- LoRaMac-node v4.6.0
- TTN network server v3.18.2
- nam1 cluster

If you wish to retain compatibility with existing LoRaWAN 1.1.0 devices that do not implement the erratum, it might be best to introduce a new MAC version, e.g., MAC_V1_1_0_FOPTS. In the console, you could present this as "LoRaWAN Specification 1.1.0 + FOpts, FCntDown erratum". The end-user would then be able to select the appropriate MAC version, depending on whether their devices implement the changes proposed in the erratum.

When MAC_V1_1_1_0_FOPTS is selected, the network server would encrypt the FOpts header field differently, [according to the erratum](https://resources.lora-alliance.org/technical-specifications/fopts-encryption-usage-of-fcntdwn-errata-on-the-lorawan-l2-1-1-specification).

See [this ChirpStack function](https://github.com/brocaar/lorawan/blob/master/phypayload.go#L904-L947) for an implementation in Go. Or see [this function](https://github.com/Lora-net/LoRaMac-node/blob/master/src/mac/LoRaMacCrypto.c#L190-L259) in LoRaMac-node for an implementation in C.

(see above)

No. I am not familiar with the TTN code base.

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/5360

#### Changes

- Remove support for LoRaWAN version comparisons in the codebase
  - These were hard to maintain, since we had lots of checks spread throughout the Network Server and Join Server
  - They have been replaced by the `macspec` package, which based on a LoRaWAN version can decide if certain behavior is available
- Support the FOpts encryption erratum
  - This one is achieved by introducing the concept of encryption options, which manipulate how data is encrypted
  - The `macspec` package can generate encryption options based on the LoRaWAN version
- Avoid starting the downlink task dispatcher in some tests
  - Fixes some annoying flaky tests that had the `Dispatch called, but not set` error

#### Testing

Tested with the 4.6.0 release of LoRaMac-node. I've tested basically all 3 situations:
- FOpts attached to network (FPort = 0) downlink - start with ADR margin 60 dBm, allow device to send some uplinks in SF12. Change margin to 15 dBm. This forces a `LinkADRReq` which will be part of the FOpts. The corresponding `LinkADRAns` confirms that the `LinkADRReq` was received.
- FOpts attached to application (FPort != 0) downlink - same instructions, but also add an application downlink before you change the margin to 15 dBm.
- FOpts attached to uplink - both of the above tests confirm that this works, as `LinkADRAns` arrives for both in the FOpts field if the end device always sends data uplinks.

##### Regressions

I could not find a LoRaMac-node version that implements the pre-erratum version. I don't know if these devices actually exist in the wild.

#### Checklist


pkg/networkserver/downlink.go
```diff
@@ -229,60 +230,52 @@ func (ns *NetworkServer) generateDataDownlink(ctx context.Context, dev *ttnpb.En
...
-				func(ctx context.Context, dev *ttnpb.EndDevice, maxDownLen uint16, maxUpLen uint16) mac.EnqueueState {
-					// NOTE: LinkADRReq must be enqueued after NewChannelReq.
-					st, err := mac.EnqueueLinkADRReq(ctx, dev, maxDownLen, maxUpLen, phy)
-					if err != nil {
-						logger.WithError(err).Error("Failed to enqueue LinkADRReq")
-						return mac.EnqueueState{
-							MaxDownLen: maxDownLen,
-							MaxUpLen:   maxUpLen,
-						}
+		enqueuers = append(enqueuers,
```
adriansmares, line 233: I've removed the version checks here since each individual  `mac.Enqueue*` command does the version checks as well. I can bring them back, but I'm not sure how much value that would bring (especially since arithmetic comparisons are no longer what we should do).

johanstokking:
Conceptually this `macspec` is a fantastic idea.

Regarding 1.1 vs 1.1 + errata: the original 1.1 is considered broken and can only work correctly with the errata. So this would rather be a fix for 1.1; no need for another version. Right?

I'll review tomorrow.

adriansmares:
> Regarding 1.1 vs 1.1 + errata: the original 1.1 is considered broken and can only work correctly with the errata. So this would rather be a fix for 1.1; no need for another version. Right?

I actually don't know. I've read this statement, that 1.1 was 'not possible' without the errata, but I don't really understand how development actually occurred since 2018 related to 1.1 - I would have expected to receive more bug reports regarding this in the last 2 years. 

We can definitely just add this behavior to the 1.1 version, but this is a breaking change and I would like to understand better if we really have and will have 0 original 1.1 end devices.

For example https://github.com/TheThingsNetwork/lorawan-stack/pull/2339 came in April 2020, and it basically introduces the 'vanilla' 1.1 encryption behavior. Did we have collective amnesia with respect to the errata (which came in January 2018) at the time ? Because both LoRaMac-node and mbed-OS did not use the errata at the time, and apparently people were testing MIC checks and MAC commands 🤔 .

htdvisser (approved):
Code-owned files LGTM.

Package `macspec` seems like a good idea. Please make sure to add it to `CODEOWNERS`.

johanstokking:
> I actually don't know. I've read this statement, that 1.1 was 'not possible' without the errata, but I don't really understand how development actually occurred since 2018 related to 1.1 - I would have expected to receive more bug reports regarding this in the last 2 years.
> 
> We can definitely just add this behavior to the 1.1 version, but this is a breaking change and I would like to understand better if we really have and will have 0 original 1.1 end devices.

I don't think anyone uses LoRaWAN 1.1 seriously, especially not the non-errata.

This is the timeline:

- October 2017: LoRaWAN 1.1
- January 2018: LoRaWAN 1.1 FCnt errata (with Christmas and New Year in between, plus TC meeting schedule, this is effectively an immediate fix)
- July 2018: LoRaWAN 1.0.3
- October 2020: LoRaWAN 1.0.4

It was pretty clear in 2018 that LoRaWAN 1.1 was too far fetched; the device maker community wanted a simpler stack with minor additions/fixes compared to 1.0.2 and more regions. So from what I know, nothing happened to LoRaWAN 1.1 since the spring of 2018 and all effort went in LoRaWAN 1.0.3, 1.0.4 and RP2.

> For example #2339 came in April 2020, and it basically introduces the 'vanilla' 1.1 encryption behavior. Did we have collective amnesia with respect to the errata (which came in January 2018) at the time ?

Yes, I think so.

> Because both LoRaMac-node and mbed-OS did not use the errata at the time, and apparently people were testing MIC checks and MAC commands 🤔 .

Early LoRaWAN 1.1 implementations have mostly been implementations without actual interoperability testing on actual LoRaWAN networks. I confirm this specifically for the Mbed OS stack.

I think we should consider erratas really as fixes. Otherwise there would have been a 1.1.1 version instead.

adriansmares:
Per above conversation I've removed the `1_1_FCNT` version and incorporated the changes into the currently existing `1_1` version. Feel free to review @johanstokking .

pkg/crypto/data_messages.go
```diff
@@ -24,7 +24,26 @@ import (
...
+// EncryptionOption encapsulates custom LoRaWAN encryption options.
+type EncryptionOption func(*encryptionOptions)
+
+// WithFrameIdentifier specifies which bytes are to be used for
+// positions [1:5] during encryption.
+func WithFrameIdentifier(identifier [4]byte) EncryptionOption {
+	return func(encOpts *encryptionOptions) {
+		copy(encOpts.frameIdentifier[:], identifier[:])
+	}
+}
```
johanstokking: Can you document this a bit more? What is the LoRaWAN aspect of this?

pkg/specification/macspec/specification.go
```diff
@@ -0,0 +1,242 @@
...
+	return compareMACVersion(v, ttnpb.MACVersion_MAC_V1_0_4) >= 0
+}
+
+// UseNwkKey reports whether v uses a root NwkKey.
+func UseNwkKey(v ttnpb.MACVersion) bool {
+	return compareMACVersion(v, ttnpb.MACVersion_MAC_V1_1) >= 0
+}
+
+// UseLegacyMIC reports whether v uses legacy MIC computation algorithm.
+func UseLegacyMIC(v ttnpb.MACVersion) bool {
```
johanstokking, line 99: Legacy is a bit of an unfortunate name in The Things Stack. Initially, in 2018 when we started with The Things Stack, LoRaWAN 1.1 was the future and so LoRaWAN 1.0.x was legacy. That didn't really turn out to be the case.

With LoRaWAN 1.2, the MIC is going to change again. I think therefore that these shouldn't be boolean switches but rather per-version functions that compute the MIC. If we do that in `pkg/crypto`, do we need this function here?

This is not really a need for change now, but some foreseeable technical debt with LoRaWAN 1.2 on the horizon.

pkg/specification/macspec/specification.go
```diff
@@ -0,0 +1,242 @@
...
+	return compareMACVersion(v, ttnpb.MACVersion_MAC_V1_1) < 0
+}
+
+// UseSharedFCntDown reports whether v uses the same frame counter
+// for both network and application downlinks.
+func UseSharedFCntDown(v ttnpb.MACVersion) bool {
```
johanstokking, line 105: `UseSingleFCntDown` ?

pkg/specification/macspec/specification.go
```diff
@@ -0,0 +1,242 @@
...
+	return compareMACVersion(v, ttnpb.MACVersion_MAC_V1_1) < 0
+}
+
+// UseDLChannel reports whether v is allowed to use the
+// DLChannel{Req|Ans} MAC commands.
+func UseDLChannel(v ttnpb.MACVersion) bool {
```
johanstokking:
```suggestion
// UseDLChannelReq reports whether v is allowed to use the
// DLChannel{Req|Ans} MAC commands.
func UseDLChannelReq(v ttnpb.MACVersion) bool {
```

All versions have downlink channels, but not all versions support the setup with `DLChannelReq`. I found the `UseDLChannelReq(v)` check a bit weirdly looking. This is a bit of a personal preference.

pkg/specification/macspec/specification.go
```diff
@@ -0,0 +1,242 @@
...
+	return compareMACVersion(v, ttnpb.MACVersion_MAC_V1_0_2) >= 0
+}
+
+// UseTxParamSetup reports whether v is allowed to use the
+// TxParamSetup{Req|Ans} MAC commands.
+func UseTxParamSetup(v ttnpb.MACVersion) bool {
```
johanstokking: Idem and onwards

pkg/specification/macspec/specification.go
```diff
@@ -0,0 +1,242 @@
...
+
+// AllowDuplicateLinkADRAns reports whether v is allowed to use
+// duplicate LinkADRAns MAC responses within the same message.
+func AllowDuplicateLinkADRAns(v ttnpb.MACVersion) bool {
+	return compareMACVersion(v, ttnpb.MACVersion_MAC_V1_0_2) >= 0 && compareMACVersion(v, ttnpb.MACVersion_MAC_V1_1) < 0
+}
+
+// AtomicLinkADRAns reports whether v will accept or reject all
+// channel masks in a singular LinkADRAns MAC response.
+func AtomicLinkADRAns(v ttnpb.MACVersion) bool {
```
johanstokking: Then I would also call this `SingularLinkADRAns`

pkg/specification/macspec/specification.go
```diff
@@ -0,0 +1,242 @@
...
+
+// LimitConfirmedTransmissions reports whether v must limit
+// the number of confirmed uplink transmissions.
+func LimitConfirmedTransmissions(v ttnpb.MACVersion) bool {
+	return compareMACVersion(v, ttnpb.MACVersion_MAC_V1_0_4) < 0
+}
+
+// RequireDevEUIForABP reports whether v requires ABP devices to have a DevEUI associated.
+func RequireDevEUIForABP(v ttnpb.MACVersion) bool {
+	return compareMACVersion(v, ttnpb.MACVersion_MAC_V1_0_4) >= 0 && compareMACVersion(v, ttnpb.MACVersion_MAC_V1_1) < 0
```
johanstokking: Let's also require this for LoRaWAN 1.1. I'm aware it's not with that many words in the spec, but let's please require this.

pkg/specification/macspec/specification.go
```diff
@@ -0,0 +1,242 @@
...
+		panic("unreachable")
+	}
+}
+
+// ValidateUplinkPayloadSize reports whether v requires that the uplink
+// payload size must be valid with respect to the regional parameters
+// and local limitations.
+func ValidateUplinkPayloadSize(v ttnpb.MACVersion) bool {
+	return compareMACVersion(v, ttnpb.MACVersion_MAC_V1_1) >= 0
+}
```
johanstokking: Is this a spec change since 1.1?

johanstokking (approved):
Looks great, just minor code quality things which you can ignore if you disagree.

pkg/crypto/data_messages.go
adriansmares: This is a bit of poetic license from my side. The (updated) specification does not have names for the parts of the block:
image
The 4 bytes after the `0x01` are what I call the frame identifier. The naming itself comes from the erratum body:
image
They are called 'constants' there, but their role is to separate between different frame types - so maybe 'frame type constant' is closer to what they really do ?

pkg/specification/macspec/specification.go
adriansmares: We can have the equivalent of `crypto.EncryptionOption` for MIC - `crypto.MICOption` that basically controls how the MIC is generated, and then the `macspec` package can choose based on the LoRaWAN version what options to provide.

What is a bit different is that the LoRaWAN 1.1 MIC is basically derived from the 1.0.x MIC, so for now we're always computing the 1.0.x MIC. I guess we will see when 1.2 is out what kind of changes to introduce.

pkg/specification/macspec/specification.go
adriansmares: I named it 'shared' since this makes it clear that both entities (application layer and network layer) basically fight/negotiate which value to use. The fact that the frame counter is single could also be understood as only one layer needs to manage it.

pkg/specification/macspec/specification.go
adriansmares: Full disclosure - I added this because it existed before:
https://github.com/TheThingsNetwork/lorawan-stack/blob/70975ab7e96936103f2e1ca70b53cd68e1a64898/pkg/networkserver/downlink.go#L1825-L1832

The spec itself is quite straight forward about why we do this (page 29):
image

pkg/crypto/data_messages.go
johanstokking: Yes that sounds better to me.

pkg/specification/macspec/specification.go
johanstokking: The 1.1 MIC is already different, right, taking two integrity keys, ConfFCnt and TX parameters into account?

What will happen with 1.2 is that there comes a companion spec that specifies cipher suites. They will declare the MIC input and keys to use. Indeed let's see how that comes out of the washing machine and dryer and not address anything here and now.

adriansmares: It is different but it acts as an extension of the 1.0.x MIC. So each time when you compute the 1.1 MIC, your first step is to actually compute the 'legacy' mic (which is 2 bytes), then you compute the 1.1 extension from there (the other two bytes).

This is an optimization used by the NS right now:
- You always compute the expected legacy MIC
- You always compare the expected legacy MIC with the provided legacy MIC. If they mismatch, you can already drop the frame since extending it with the 1.1 procedure won't make it match.
- If the end device is a 1.1 device, you actually compute the other two bytes and compare the MIC.

ryaplots (approved):
code-owned lgtm

pkg/specification/macspec/specification.go
johanstokking: Okay so this is for downlink. That should be more clear.

PR #5466, opened 2022-05-19
Fix handling of dropped events

htdvisser:
#### Summary

This should fix an issue where events would be silently dropped. With these changes is should get less likely that events get dropped from the Stream RPC, and if they do get dropped, we'll be able to see that in telemetry.

#### Changes

- Fix metric that wasn't incrementing
- Use larger event channels in Stream RPC

#### Testing

I couldn't reliably reproduce the problem locally, but I think this should fix the issue.

##### Regressions

none expected

#### Notes for Reviewers

Let's try to still get this in with v3.19.2

#### Checklist


adriansmares (approved):

htdvisser:
I'm going to ignore the flaky test and just merge this.

PR #5473, opened 2022-05-20
Replace Issue Templates with Issue Forms

htdvisser:
#### Summary

This replaces some of our issue templates with issue forms.

#### Notes for Reviewers

See https://github.com/htdvisser/lorawan-stack-fork/issues/new/choose

Let me know what you think about the new questions / sections. Is anything missing? Anything unnecessary that should be removed?


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,100 @@
...
+      options:
+        - Linux (64-bit Intel)
+        - Linux (other)
+        - Windows (64-bit Intel)
+        - Windows (other)
+        - MacOS (64-bit Intel)
+        - MacOS (Apple Silicon)
+        - Other
+    validations:
+      required: true
```
johanstokking: Not sure what the relevance is of this.

I would indeed ask for what the client is "using", but it'll depend on the context what is relevant: browser (when using Console), OS (when using TTSOS, TTSE or CLI).

Since many issues are filed by end users, @kschiffer @ryaplots what kind of information do you need here?

johanstokking (commented):
In general this is very nice

.github/ISSUE_TEMPLATE/bug_report.yml
NicolasMrad: it looks really nice from my side, and agree with Johan maybe we can just mention the usage and people can mention the OS or browser or the thing that may affect situation.

KrishnaIyer (approved):
Don't have any different comments. Looks good 👍

adriansmares (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,100 @@
...
+  - type: textarea
+    attributes:
+      label: Expected Result
+      description: Explain what you think should happen instead.
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Relevant Logs
+      description: Copy and paste any relevant logs (don't screenshot your logs). This will be automatically formatted into code, so no need for backticks.
```
kschiffer:
```suggestion
      description: Copy and paste any relevant logs. If the error occurred in the Console, please include the full error message and browser logs if applicable. This will be automatically formatted into code, so no need for backticks.
```

kschiffer (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
kschiffer: Browser (name and version) would be nice to know indeed. I agree with Johan that bugs can be reported from two perspectives: user and operator, which makes the template ambiguous. Maybe we want to split that up into two separate templates in the future.

PR #5474, opened 2022-05-20
Validate complete entity on update

adriansmares:
#### Summary

References https://sentry.io/organizations/the-things-industries/issues/3288249317/?project=2682566
References https://sentry.io/organizations/the-things-industries/issues/3128395476/?project=2682566

I've started with the above mentioned issues while looking into how `ttnpb.EndDevice.Session.Keys` could be `nil`, given that the field is marked as `required`. The rabbit whole led to this PR.

#### Changes

- The Redis registry implementations now validate the complete object upon updates, instead of just the updated paths.
  - The reasoning is that a `SetFields` operation may create previously `nil` fields. Consider the following situation:
    1. You create an end device entity by providing a valid entity. You don't set any `session` or `pending_session`.
    2. You update the entity by setting `session.dev_addr`. This creates both the `session` and the `session.dev_addr`.
    3. At this point, the registry code will validate only `session.dev_addr`. `session.keys` is `nil`, even though it is a field marked as `required` - no error occurs !
  - In general it doesn't seem like we can avoid having to do a full object validation round. Possible proposals include doing the full object check only for user facing operations (coming from user-facing RPCs).
  - Validation itself is quite light on most entities, but the Network Server end device registry contains rather large entities (specifically due to traffic history). ~Since the performance hit may be too high, the change here is guarded by the experimental flag `ns.registry.redis.full_validation`.~

#### Testing

Local testing.

##### Regressions

There are two kinds of regressions that we may encounter here:
- The performance regression because we validate the complete object instead of just the updated fields.
  - The expectation here is that the this will come mainly from the Network Server end device registry, which is why this is guarded by a feature flag.
  - The other registries are light as is, and should not cause issues.
- Previously valid entity updates and creation requests will now result in errors.
  - The result object is invalid, so this is covered by the compatibility agreement.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,13 +13,16 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
+- Entities are now fully validated when updated in the stores.
+  - Previously only the updated paths where validated. This lead to situations in which a partial update could cause the entity as a whole to reach an invalid state.
+
 ### Deprecated
 
 ### Removed
 
 ### Fixed
 
-- End devices running on MAC versions lower than 1.1 showing network downlink frame counters instead of application downlink frame counters.
+- End devices running on MAC versions higher or equal to 1.1 showing network downlink frame counters instead of application downlink frame counters.
 
 ### Security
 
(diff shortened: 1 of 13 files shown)


KrishnaIyer (approved):
LGTM. Should we run some "migration" to see how many entities may be flagged as invalid with this?

ryaplots (approved):
code-owned changes look good to me

kschiffer (approved):

PR #5480, opened 2022-05-23
Update withRequest handling

### LINKED ISSUE #5448: Avoid rendering children in `withRequest` when an error occurred
#### Summary
An issue in the `withRequest`-HOC causes the passed chidren to be rendered before the passed error is thrown. This caused us to use an unnecessary workaround when passing the fetching prop to the HOC.

#### Why do we need this?
- Remove unnecessary and fragile workarounds for determining fetched status
- Clean up code and ensure consistency when using `withRequest`-HOC

A check for errors in the `withRequest()` HOC that will throw the error, to be caught by the nearest error boundary.

Due to the React component lifecycle, the component renders the new props before running [`componentDidUpdate()`](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.19/pkg/webui/lib/components/with-request.js#L67-L75). This means that before the error is thrown, there is a short timeframe where the [child components are rendered](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.19/pkg/webui/lib/components/with-request.js#L87). This usually causes an error when the child components are rendered without the fetched data. This means that a different generic error is rendered instead of the relevant backend error.

We used to solve this by using [an extra check to see whether the fetched entity is still empty](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.19/pkg/webui/console/views/application/connect.js#L67).
```js
  // pkg/webui/console/views/application/connect.js:65-69
  withRequest(
    ({ appId, loadData }) => loadData(appId),
    ({ fetching, application }) => fetching || !Boolean(application),
  ),
```

This should not actually be required when avoiding rendering children if an error occurred:
```js
diff --git a/pkg/webui/lib/components/with-request.js b/pkg/webui/lib/components/with-request.js
index 12437c4d7..402903e00 100644
--- a/pkg/webui/lib/components/with-request.js
+++ b/pkg/webui/lib/components/with-request.js
@@ -84,6 +84,13 @@ const withRequest =
           )
         }
 
+        // Prevent rendering the component when an error was received
+        // but not thrown yet via `componentDidUpdate()`.
+        if (mapPropsToError(this.props)) {
+          return null
+        }
+
         return <Component {...this.props} />
       }
     }

```

1. Apply the diff as described above
2. Remove all unnecessary logic we used when using `withRequest()` throughout the codebase. In most cases just the default `mapPropsToFetching()` should suffice. It needs to be made sure though that a selector for `fetching` is set.
3. It's a good opportunity also to make sure that we always correctly pass both the `fetching` and the `error` prop (so it can be consumed by the `mapPropsToError()`-function when there was an error during the request). I noticed that this is [not always the case](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.19/pkg/webui/console/views/admin-user-management-edit/index.js#L46-L57).

I don't think we need additional tests for this.

@ryaplots

ryaplots:
#### Summary

Closes #5448 

#### Changes

- Add diff to withRequest component
- Remove unnecessary logic
- Add `fetching` and `error` selectors where needed

#### Testing

cypress tests

#### Checklist

### DIFF
--- a/pkg/webui/console/views/admin-user-management-add/index.js
+++ b/pkg/webui/console/views/admin-user-management-add/index.js
@@ -36,12 +36,14 @@ import { getIsConfiguration } from '@console/store/actions/identity-server'
 import {
   selectPasswordRequirements,
   selectIsConfigurationFetching,
+  selectIsConfigurationError,
 } from '@console/store/selectors/identity-server'
 
 @connect(
   state => ({
     passwordRequirements: selectPasswordRequirements(state),
     fetching: selectIsConfigurationFetching(state),
+    error: selectIsConfigurationError(state),
   }),
   {
     createUser: attachPromise(createUser),
--- a/pkg/webui/console/views/admin-user-management-add/index.js
+++ b/pkg/webui/console/views/admin-user-management-add/index.js
@@ -49,7 +49,7 @@ import {
     getConfiguration: () => getIsConfiguration(),
   },
 )
-@withRequest(({ getConfiguration }) => getConfiguration(), ({ fetching }) => fetching)
+@withRequest(({ getConfiguration }) => getConfiguration())
 @withBreadcrumb('admin.user-management.add', () => (
   <Breadcrumb path={`/admin/user-management/add`} content={sharedMessages.add} />
 ))
--- a/pkg/webui/console/views/application-general-settings/connect.js
+++ b/pkg/webui/console/views/application-general-settings/connect.js
@@ -107,6 +107,7 @@ const mapStateToProps = state => {
     mayViewCollaborators,
     mayViewLink,
     fetching,
+    error,
     mayPurge: mayPurgeApp,
     shouldConfirmDelete: !isPristine || !mayViewCollaborators || !mayViewApiKeys || Boolean(error),
     mayDeleteApplication: mayDeleteApp,
--- a/pkg/webui/console/views/application-general-settings/connect.js
+++ b/pkg/webui/console/views/application-general-settings/connect.js
@@ -162,10 +162,7 @@ const addHocs = pipe(
   withFeatureRequirement(mayEditBasicApplicationInfo, {
     redirect: ({ appId }) => `/applications/${appId}`,
   }),
-  withRequest(
-    ({ loadData }) => loadData(),
-    ({ fetching }) => fetching,
-  ),
+  withRequest(({ loadData }) => loadData()),
 )
 
 export default ApplicationGeneralSettings => addHocs(ApplicationGeneralSettings)
--- a/pkg/webui/console/views/device-add/connect.js
+++ b/pkg/webui/console/views/device-add/connect.js
@@ -37,12 +37,9 @@ export default DeviceAdd =>
     mapStateToProps,
     mapDispatchToProps,
   )(
-    withRequest(
-      ({ getPrefixes, jsEnabled }) => {
-        if (jsEnabled) {
-          return getPrefixes()
-        }
-      },
-      ({ fetching }) => fetching,
-    )(DeviceAdd),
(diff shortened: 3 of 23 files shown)


pkg/webui/console/views/admin-user-management-add/index.js
```diff
@@ -49,7 +49,7 @@ import {
     getConfiguration: () => getIsConfiguration(),
   },
 )
-@withRequest(({ getConfiguration }) => getConfiguration(), ({ fetching }) => fetching)
+@withRequest(({ getConfiguration }) => getConfiguration())
```
kschiffer, line 54: The `error`\-prop is not selected in the `connect()` above. The `withRequest` will then not be aware of any errors that the action might have resulted in.

pkg/webui/console/views/application-general-settings/connect.js
```diff
@@ -162,10 +162,7 @@ const addHocs = pipe(
   withFeatureRequirement(mayEditBasicApplicationInfo, {
     redirect: ({ appId }) => `/applications/${appId}`,
   }),
-  withRequest(
-    ({ loadData }) => loadData(),
-    ({ fetching }) => fetching,
-  ),
+  withRequest(({ loadData }) => loadData()),
```
kschiffer, line 166: `error` is not selected here as well, even though it gets composed in L86.

pkg/webui/console/views/device-add/connect.js
```diff
@@ -37,12 +37,9 @@ export default DeviceAdd =>
...
   )(
-    withRequest(
-      ({ getPrefixes, jsEnabled }) => {
-        if (jsEnabled) {
-          return getPrefixes()
-        }
-      },
-      ({ fetching }) => fetching,
-    )(DeviceAdd),
+    withRequest(({ getPrefixes, jsEnabled }) => {
```
kschiffer, line 44: `error` is not selected here.

pkg/webui/console/views/gateway-general-settings/index.js
```diff
@@ -134,7 +134,7 @@ import m from './messages'
 @withFeatureRequirement(mayEditBasicGatewayInformation, {
   redirect: ({ gtwId }) => `/gateways/${gtwId}`,
 })
-@withRequest(({ loadData }) => loadData(), ({ fetching }) => fetching)
+@withRequest(({ loadData }) => loadData())
```
kschiffer, line 138: `error` is not selected here.

pkg/webui/console/views/organization-general-settings/connect.js
```diff
@@ -115,9 +115,4 @@ export default GeneralSettings =>
     mapStateToProps,
     mapDispatchToProps,
     mergeProps,
-  )(
-    withRequest(
-      ({ loadData }) => loadData(),
-      ({ fetching }) => fetching,
-    )(GeneralSettings),
-  )
+  )(withRequest(({ loadData }) => loadData())(GeneralSettings))
```
kschiffer, line 119: `error` is not selected here.

kschiffer (changes requested):
There are still quite some missing `error` props in there. Please also go through the `account` codebase.

pkg/webui/console/views/application-general-settings/connect.js
ryaplots: Would it be enough to add `error` to the returned values here https://github.com/TheThingsNetwork/lorawan-stack/blob/b1e0a11a620ba04a1e8bf4edd7d6c3cca2904147/pkg/webui/console/views/application-general-settings/connect.js#L103-L114?

kschiffer: Yes exactly.

kschiffer (approved):
Thank you!

PR #5487, opened 2022-05-24
Resolve lint issues in pkg/config

htdvisser:
#### Summary

This pull request resolves most lint issues in `pkg/config`. There are still a couple of issues that will need to be fixed in the future:

- `goconst` issue about repeated `"directory"`, `"blob"` etc. around fetcher configs
- `gocyclo` issue about the config manager
- `copylocks` issue about the mutex in TLS config

### DIFF
--- a/pkg/config/tlsconfig/config.go
+++ b/pkg/config/tlsconfig/config.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package tlsconfig provides configuration for TLS clients and servers.
 package tlsconfig
 
 import (
@@ -32,7 +33,7 @@ type ACME struct {
 	manager *autocert.Manager
 
 	// TODO: Remove Enable (https://github.com/TheThingsNetwork/lorawan-stack/issues/1450)
-	Enable      bool     `name:"enable" description:"Enable automated certificate management (ACME). This setting is deprecated; set the TLS config source to acme instead"`
+	Enable      bool     `name:"enable" description:"Enable automated certificate management (ACME). This setting is deprecated; set the TLS config source to acme instead"` //nolint:lll
 	Endpoint    string   `name:"endpoint" description:"ACME endpoint"`
 	Dir         string   `name:"dir" description:"Location of ACME storage directory"`
 	Email       string   `name:"email" description:"Email address to register with the ACME account"`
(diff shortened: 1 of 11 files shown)


pkg/config/tlsconfig/config.go
```diff
@@ -177,31 +177,39 @@ func readCert(fileReader FileReader, certFile, keyFile string) (*tls.Certificate
...
-	CipherSuites []string   `name:"cipher-suites" description:"DEPRECATED: List of IANA names of TLS cipher suites to use"`
+	CipherSuites []string   `name:"cipher-suites" description:"DEPRECATED: List of IANA names of TLS cipher suites to use"` //nolint:lll
 }
 
 var (
-	errInvalidTLSConfigSource = errors.DefineFailedPrecondition("tls_config_source_invalid", "invalid TLS configuration source `{source}`")
-	errEmptyTLSSource         = errors.DefineFailedPrecondition("tls_source_empty", "empty TLS source")
-	errTLSKeyVaultID          = errors.DefineFailedPrecondition("tls_key_vault_id", "invalid TLS key vault ID")
-	errInvalidCipherSuite     = errors.DefineFailedPrecondition("tls_cipher_suite_invalid", "invalid TLS cipher suite {cipher}")
+	errInvalidTLSConfigSource = errors.DefineFailedPrecondition(
```
adriansmares, line 188: Ok so I've skipped `lll` on such long errors in my PRs. In the future I'll refactor to this style instead.

pkg/config/tlsconfig/provider.go
```diff
@@ -35,7 +35,11 @@ func (fn ConfigOptionFunc) apply(c *tls.Config) {
 }
 
 // WithTLSClientAuth sets TLS client authentication options.
-func WithTLSClientAuth(auth tls.ClientAuthType, cas *x509.CertPool, verify func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error) Option {
+func WithTLSClientAuth(
+	auth tls.ClientAuthType,
```
adriansmares, line 39: Do we have a clear style here ?

The rules I've been able to infer so far are:
- If you can put all parameters on a line, do it.
- Otherwise put the return and name+receiver on separate lines
- Otherwise do the above, and give each parameter a new line.

adriansmares (approved):

pkg/config/tlsconfig/provider.go
htdvisser: We don't have a clear style (yet), but here's what I've been doing:

0. If it fits, put it on one line:

```go
func (s *Server) Handle(ctx context.Context, proxyReq events.APIGatewayProxyRequest) (events.APIGatewayProxyResponse, error) {
```

1. If it doesn't fit, split arguments on separate line:

```go
func (s *Server) Handle(
	ctx context.Context, proxyReq events.APIGatewayProxyRequest,
) (events.APIGatewayProxyResponse, error) {
```

2. If it still doesn't fit, split arguments onto multiple lines:

```go
func (s *Server) Handle(
	ctx context.Context,
	proxyReq events.APIGatewayProxyRequest,
) (events.APIGatewayProxyResponse, error) {
```

3. If it still doesn't fit, and you're sure you shouldn't just rewrite your func, it's also possible to split the return values:

```go
func (s *Server) Handle(
	ctx context.Context,
	proxyReq events.APIGatewayProxyRequest,
) (
	events.APIGatewayProxyResponse,
	error,
) {
```

PR #5490, opened 2022-05-25
Protect error definitions with a mutex

htdvisser:
#### Summary

This is a fix for issues where (parallel) tests would panic because either (1) they would register error definitions concurrently or (2) they would re-register error definitions if `-count` is greater than one. 

#### Changes

- Allow re-registering error definitions if they're the same
- Protect error definitions with a mutex

#### Testing

It's all about testing

##### Regressions

None expected

#### Notes for Reviewers

None

#### Checklist

### DIFF
--- a/pkg/errors/definitions.go
+++ b/pkg/errors/definitions.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"sync"
 
 	"github.com/gotnospirit/messageformat"
 	"go.thethings.network/lorawan-stack/v3/pkg/i18n"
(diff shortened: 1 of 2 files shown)


pkg/errors/definitions.go
```diff
@@ -169,17 +168,46 @@ nextArg:
 
 	def.setGRPCStatus() // store the (marshaled) gRPC status message.
 
-	Definitions[fullName] = &def
+	definitionsMu.Lock()
```
adriansmares, line 187: I don't see why you wouldn't `defer` here. You want to avoid holding the lock while defining the `i18n` definition ?

adriansmares (approved):
I'm ok with this as is, but shouldn't we long term move away from definitions defined inside tests to global definitions anyway ?

pkg/errors/definitions.go
htdvisser: Yes, exactly. The `i18n` part should probably (in a separate PR) get its own mutex, and we know all too well how frustrating it is to debug those. But I can extract locked map access into a separate func, so that we can do a deferred unlock.

PR #5494, opened 2022-05-30
Add CLI commands for Notification Service

### LINKED ISSUE #5457: Add support for reading notifications in the CLI
#### Summary

As a follow-up of https://github.com/TheThingsNetwork/lorawan-stack/issues/4935 we should add CLI support for reading notifications and marking them as seen.

#### Why do we need this?

Although I expect that most people will use the Console to work with notifications (TODO: issue reference), it should also be possible to manage notifications with the CLI.

Backend support, but no frontend support yet.

A CLI command `ttn-lw-cli notifications list`, perhaps with a flag `--mark-as-seen`.

Yes

htdvisser:
#### Summary

This pull request adds the CLI commands `ttn-lw-cli notifications list` and `ttn-lw-cli notifications set-status`.

Closes #5457.

#### Changes

- Define CLI flags on the proto request messages
- Hide the `email` flag from `UserIdentifiers`, since that flag isn't used anywhere.
- Upgrade the flags generator to fix an issue where enums (in this case the notification statuses) specified by flags were appended to the default value instead of replacing the default value.
- Implement the `ttn-lw-cli notifications list` command.
	- For convenience, I also added a `--mark-as-seen` flag that immediately marks the listed notifications as seen.
- Implement the `ttn-lw-cli notifications set-status` command.

#### Testing

🐒 

##### Regressions

None expected.

#### Notes for Reviewers

nil.

#### Checklist

	- This will be auto-generated


cmd/ttn-lw-cli/commands/notifications.go
```diff
@@ -0,0 +1,144 @@
...
+		if err != nil {
+			return err
+		}
+		getTotal()
+
+		if err = io.Write(os.Stdout, config.OutputFormat, res.Notifications); err != nil {
+			return err
+		}
+
+		if markAsSeen, _ := cmd.Flags().GetBool("mark-as-seen"); !markAsSeen && len(res.Notifications) > 0 {
```
adriansmares: I miss why the logic is reversed here. So you're marking them as unseen if the flag is false, but where are they marked as seen ? I see that the store implementation uses the read only store, so `List` cannot have side effects.

adriansmares (approved):

cmd/ttn-lw-cli/commands/notifications.go
htdvisser: Oh yeah this is completely wrong. Don't know what I was thinking.

PR #5500, opened 2022-05-30
Add generic net.Error timeout fallback

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack-support/issues/745

#### Changes

- Add a generic `net.Error` fallback to `pkg/errors` conversion, only for timeout errors.
- Change the description of `errNetOperation`, as we never provided a `message` attribute anyway, and using `Error()` is dubious as we cannot control the chain of contents.
- Add a generic error for `os.DeadlineExceeded`

#### Testing

Local testing.

##### Regressions

None are expected. Specifically this fallback just ensures that if a timeout error is encountered, we actually output this information to the user.

#### Notes for Reviewers

This is a very conservative implementation, since we would like to avoid outputting the error message (`Error()`) for any error that matches the interface. The original issue would require a form of type checking for the `net/http` errors, but unfortunately these are [unexported](https://github.com/golang/go/blob/690ac4071fa3e07113bf371c9e74394ab54d6749/src/net/http/transport.go#L2503).

Let me know if you think that we can actually always expose the message.

#### Checklist

### DIFF
--- a/pkg/errors/conversion.go
+++ b/pkg/errors/conversion.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"net"
 	"net/url"
+	"os"
 
 	"github.com/golang/protobuf/proto"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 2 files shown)


pkg/errors/conversion.go
```diff
@@ -20,12 +20,15 @@ import (
...
 	"net"
 	"net/url"
+	"os"
 
 	"github.com/golang/protobuf/proto"
 	"google.golang.org/grpc/status"
 )
 
 var (
+	errDeadlineExceeded = DefineDeadlineExceeded("deadline_exceeded", "deadline exceeded")
```
htdvisser: Can't we just treat `os.ErrDeadlineExceeded` as `context.DeadlineExceeded`?

adriansmares: It also applies to things like read / writes timeout (caused by `SetReadDeadline` and so forth), but I'm ok with reusing the definition.

htdvisser (approved):

PR #5505, opened 2022-06-01
Add typing to pkg/workerpool

adriansmares:
#### Summary

This PR adds type information to `pkg/workerpool`, in order to avoid type conversions to and from `interface{}`.

#### Changes

- Make worker pools typed, with the worker item type being carried to the handler
  - This works very well for pools that have a singular item type, even if it is an interface.
  - There is a pool in `pkg/gatewayserver` which still requires `any` because it may carry 3 different work items. There are no sum types in Go yet (only type set constrains, which cannot be used here), so for now it will have to stay as is.

#### Testing

Unit testing.

##### Regressions

No new functionality has been added or changed - this is just a tech debt change that moves away from the unsafe `interface{}` to a stronger typing. As such, existing unit tests should cover all of the uses.

#### Checklist

### DIFF
--- a/pkg/applicationserver/applicationserver.go
+++ b/pkg/applicationserver/applicationserver.go
@@ -87,9 +87,9 @@ type ApplicationServer struct {
 	interopClient InteropClient
 	interopID     string
 
-	activationPool     workerpool.WorkerPool
-	processingPool     workerpool.WorkerPool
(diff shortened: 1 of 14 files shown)


htdvisser (approved):
Cool stuff

PR #5506, opened 2022-06-01
Fix concurrency of login requests in cypress

kschiffer:
#### Summary
This quickfix PR is resolving flaky behavior in the cypress tests, when login request are not completed in the right order.\

#### Changes
- Chain requests done for the programmatic login utility correctly

#### Testing

Manual.

#### Notes for Reviewers
This PR is presented by @adriansmares !

#### Checklist

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -135,6 +135,8 @@ jobs:
         uses: ./.github/actions/build-mage
       - name: Install Node and Dependencies
         uses: ./.github/actions/install-node-and-deps
+      - name: Fix chrome version
+        uses: browser-actions/setup-chrome@latest
       - uses: actions/download-artifact@v2
         name: Download build artifacts
         with:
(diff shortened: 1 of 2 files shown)


adriansmares (approved):
For context: we were looking at 'access token not found' flakes and realized that `cy.request` does not block until the request is done, so even though the 3 requests are done in order here, they actually occur in parallel.

This means that it is possible that the `Obtain access token.` step occurs before the `Login to Account App provider` does. When this happens, the `Obtain access token.` obviously fails, because the login did not occur yet.

.github/workflows/e2e.yml
```diff
@@ -135,6 +135,7 @@ jobs:
         uses: ./.github/actions/build-mage
       - name: Install Node and Dependencies
         uses: ./.github/actions/install-node-and-deps
+      - uses: browser-actions/setup-chrome@stable
```
adriansmares: For reference: This is required because the end to end tests run on multiple machines, and the expectation is that all of these machines run the same Chrome version. Apparently GitHub runners at the time of writing are split between Chrome 101 and Chrome 102, so in order to avoid flakes we need to make sure that Chrome is updated.

Even this PR had this particular flake - see the run at https://github.com/TheThingsNetwork/lorawan-stack/actions/runs/2422495024 .

htdvisser (approved):

ryaplots (approved):

PR #5517, opened 2022-06-07
Use expected uplink dwell time for 1.0.4 checks

adriansmares:
#### Summary

This PR ensures that we use the expected uplink dwell time while performing the LoRaWAN 1.0.4 max uplink payload size checks. The change is done for consistency reasons.

For the reference the difference between the expected uplink dwell time and the current parameters uplink dwell time is fallback handling. 

#### Changes

- Use expected uplink dwell time during matching, instead of only the current parameters.

#### Testing

Unit tests.

##### Regressions

N/A.

#### Checklist


adriansmares:
Decided against this because this would cause the NS to reject uplinks from end device which may not have received a `TxParamSetupReq` yet.

PR #5369, opened 2022-04-03
Add ClassBCGatewayIdentifiers with group index

onizmx:
#### Summary

Reference: [#5183](https://github.com/TheThingsNetwork/lorawan-stack/issues/5183)

#### Changes

* Added ClassBCGatewayIdentifiers with group index to API for sending a downlink message to multiple gateways using the same multicast session.

#### Testing

N/A. Let me know if needed.

##### Regressions

I don't think fields prefixed with `XXX_` are used. This change won't handle values from those fields.

#### Notes for Reviewers

This change is mostly for the client side as agreed. Things I would like to get some help with are:
1. If group index validation (not_set cannot be mixed with valid group index values) should go with this PR. If yes, which places require the validation.
2. Whether validation is needed at all. I have deployed a draft version of the group index implementation, and it looks like group index without any values have 0 by default in networkserver/downlink.go. I have done something like below, and perhaps we won't need any validations as long as the implementation looks okay
```
--- a/pkg/networkserver/downlink.go
+++ b/pkg/networkserver/downlink.go
@@ -1429,22 +1429,31 @@ func (ns *NetworkServer) attemptNetworkInitiatedDataDownlink(ctx context.Context
                        QueuedApplicationUplinks: genState.appendApplicationUplinks(nil, false),
                }
        }
+       numPaths, pathsGroup := 0, make(map[uint32][]downlinkPath)
 
-       var paths []downlinkPath
        if fixedPaths := genState.ApplicationDownlink.GetClassBC().GetGateways(); len(fixedPaths) > 0 {
-               paths = make([]downlinkPath, 0, len(fixedPaths))
+               numPaths += len(fixedPaths)
                for i := range fixedPaths {
-                       paths = append(paths, downlinkPath{
+                       groupIndex := fixedPaths[i].GroupIndex
+                       pathsGroup[groupIndex] = append(pathsGroup[groupIndex], downlinkPath{
                                GatewayIdentifiers: fixedPaths[i].GatewayIds,
                                DownlinkPath: &ttnpb.DownlinkPath{
                                        Path: &ttnpb.DownlinkPath_Fixed{
-                                               Fixed: fixedPaths[i],
+                                               Fixed: &ttnpb.GatewayAntennaIdentifiers{
+                                                       GatewayIds:   fixedPaths[i].GatewayIds,
+                                                       AntennaIndex: fixedPaths[i].AntennaIndex,
+                                               },
                                        },
                                },
                        })
                }
        }
```

#### Checklist

[...]

PR #5520, opened 2022-06-07
Support transmissions of the same frame through multiple gateways

### LINKED ISSUE #5183: Simulcast support for Class C multicast downlink
#### Summary

Simulcast support for Class C multicast device.

#### Why do we need this?

To cover some cases where we need to send exact same payload + absolute time to a single multicast device with a list of multiple gateways. For more context, we need to ensure full coverage of end devices, which are mapped into a single multicast group. Those devices are geographically spread across a large area served by multiple gateways.

`scheduleDownlinkByPaths` returns upon a successful attempt while iterating over downlink paths [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/c52be1c2b65902568db228a15c7eae23b416a524/pkg/networkserver/downlink.go#L998). We are currently sending a Class C multicast downlink to up to 8 gateways, Adding 1 second to `absolute_time` for each. This isn't very scalable for our business model. A workaround for this is creating one multicast device per gateway (or a gateway cluster), which have the same app & network key as well as frame counter & device address. This also required the gateways' GPS time synced. Since end device decoding was successful, we would prefer this to be the stack's official feature if possible.

Class C multicast downlink with `simulcast` or `use-simulcast` flag gets sent to all gateways included in `gateway_ids`, not just one.

#### Environment

3.17.2, but not related to a specific environment.

Add `use_simulcast` boolean flag to ClassBC. We already have a barebone / draft of the implementation which will be tested both locally & in our development environment. A general guidance of the implementation will be much appreciated (especially how errors should be returned on a partial success & not interfering with Class B / beaconing delay). Please have a look at the last 2 commits of this [forked branch](https://github.com/halter-corp/lorawan-stack/commits/feat/class-c-simulcast).

Test CLI with the newly added flag.
* Check both multicast / gateway events to make sure downlink is sent to all gateways.
* End device decoding.
* Run the stack in our development / staging environment to make sure nothing breaks with or without the flag.

Yes. But I would need a bit of guidance.

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/5183
References https://github.com/TheThingsNetwork/lorawan-stack/pull/5369

#### Changes

- Add downlink path group index API (courtesy of @onizmx) 
  - I've had to rebase the original PR due to the changes that got aggregated over time.
- Support transmitting the same frame from multiple gateways.

#### Testing

##### Regressions

The unit tests should cover this.

#### Notes for Reviewers

@johanstokking main reviewer for the NS changes. @htdvisser no need to review.

#### Checklist


pkg/networkserver/downlink.go
```diff
@@ -881,14 +882,15 @@ func (ns *NetworkServer) scheduleDownlinkByPaths(ctx context.Context, req *sched
 		}
 
 		var a *attempt
-		if len(attempts) > 0 {
-			if last := attempts[len(attempts)-1]; last.Equal(target) {
-				a = last
+		for _, attempt := range attempts {
```
adriansmares: Previously we would merge these attempts only if they were equal to the last target. But this doesn't really make much sense to me, since it would result in the NS doing 3 GS RPCs instead of 2 for the following set of paths:
- Gateway 1 on peer 1.
- Gateway 2 on peer 2.
- Gateway 3 on peer 1.

I understand that the previous implementation was linear with respect to the number of peers, but I think things got prematurely optimized.

Now gateway 1 and 3 are merged on the same attempt. As a result of this change the downlinks tests have been updated to reflect this merging behavior.

pkg/networkserver/downlink.go
```diff
@@ -952,7 +962,12 @@ func (ns *NetworkServer) scheduleDownlinkByPaths(ctx context.Context, req *sched
 			errs = append(errs, err)
 			continue
 		}
-		transmitAt := time.Now().Add(delay)
+		var transmitAt time.Time
+		if absTime := req.AbsoluteTime; absTime != nil {
```
adriansmares, line 966: Adding the `delay` to `time.Now` is not taking into account the round trip times. For arbitrary time scheduling this is acceptable, but for absolute time scheduling we can be more precise.

pkg/networkserver/downlink.go
johanstokking: We did that to preserve order: gateway 1, 2 and 3. Now it could be 1 and 3, then 2, while 3 may have a lower link budget or a "prefer other" downlink path constraint.

What is the optimization here really? How long do these rpcs take? There's no gateway roundtrip in `ScheduleDownlink`.

pkg/networkserver/downlink.go
```diff
@@ -967,18 +987,31 @@ func (ns *NetworkServer) scheduleDownlinkByPaths(ctx context.Context, req *sched
...
+		latestScheduledDownlink = &scheduledDownlink{
 			Message:    down,
 			TransmitAt: transmitAt,
-		}, queuedEvents, nil
+		}
+		downlinkEvents = req.DownlinkEvents.New(ctx, eventIDOpt)
+	}
+	// NOTE: A singular successful scheduling attempt will render the downlink as a whole as
+	// being successfully transmitted, even if certain groups did not have a successful scheduling
+	// attempt.
```
johanstokking: I don't understand this comment. What is a singular scheduling attempt?

pkg/networkserver/downlink.go
adriansmares: You're entirely correct that this change breaks the ordering of the paths - my bad. I've fixed this in 7cf7e5b6b5e59ea041106448dcf13987997e873f.

With respect to how much any round trip optimizations are worth - in P95 we actually can see 20ms at times, due to cross-availability-zone communication.

pkg/networkserver/downlink.go
adriansmares: This is related to the following section of the 1.0.4 spec:
image

I've updated the wording in 7cf7e5b6b5e59ea041106448dcf13987997e873f - let me know if this is more clear.

johanstokking (approved):

pkg/networkserver/downlink.go
johanstokking: Right, clear now.

PR #5524, opened 2022-06-08
Fix duplicated identifiers and visibility items in events

adriansmares:
#### Summary

This PR fixes a bug that caused certain slices in events to be duplicated.

image
image

#### Changes

1. When using the Redis events store, always unmarshal the event data completely, instead of merging the representations.
   - As mentioned in the code comment, `proto.UnmarshalMerge` has append semantics for slices, and as such duplicates the identifiers. The first identifiers come from the event sparse representation found in the streams, while the full identifiers (with EUIs) come from the actual event data.
2. Remove the deprecated `GetCorrelationIDs` getter syntax.
   - This is a bit of debt from the transitory period in which we had both custom names and non-custom names fields.

#### Testing

Local testing.

##### Regressions

1. I've looked everywhere and I couldn't find a case in which the merging behavior was actually useful. `encodeEventData` will always store a full representation of the event, so merging cannot add any missing information.
2. This is just debt and we don't have any code that actually uses `GetCorrelationIDs` (case sensitive !)

#### Checklist

### DIFF
--- a/pkg/events/correlation_context.go
+++ b/pkg/events/correlation_context.go
@@ -25,10 +25,10 @@ import (
 type correlationKey struct{}
 
 // ContextWithCorrelationID returns a derived context with the correlation IDs if they were not already in there.
-// NOTE: This function assumes that the provided correlation IDs do not contain any duplicates.
 func ContextWithCorrelationID(ctx context.Context, cids ...string) context.Context {
 	cids = append(cids[:0:0], cids...)
 	sort.Strings(cids)
+	cids = uniqueStrings(cids)
 
 	existing, ok := ctx.Value(correlationKey{}).([]string)
 	if !ok {
@@ -49,6 +49,26 @@ func CorrelationIDsFromContext(ctx context.Context) []string {
 // NewCorrelationID returns a new random correlation ID.
 func NewCorrelationID() string {
 	return ulid.MustNew(ulid.Now(), rand.Reader).String()
+}
+
+// uniqueStrings returns a slice with the unique elements of
+// the provided slice. The provided slice must be sorted,
+// and the returning slice will be sorted as well.
+// The provided slice will be modified in place, so it must
+// not be reused after uniqueStrings has been called.
+func uniqueStrings(s []string) []string {
+	if len(s) < 2 {
+		return s
+	}
+	u := 1
+	for i := 1; i < len(s); i++ {
+		if s[i] == s[i-1] {
+			continue
+		}
+		s[u] = s[i]
+		u++
+	}
+	return s[:u]
 }
 
 // mergeStrings merges 2 sorted string slices and returns the resulting slice
--- a/pkg/events/correlation_context.go
+++ b/pkg/events/correlation_context.go
@@ -25,6 +25,7 @@ import (
 type correlationKey struct{}
 
 // ContextWithCorrelationID returns a derived context with the correlation IDs if they were not already in there.
+// NOTE: This function assumes that the provided correlation IDs do not contain any duplicates.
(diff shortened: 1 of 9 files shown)


KrishnaIyer (approved):
Code owned LGTM

pkg/events/correlation_context.go
```diff
@@ -25,6 +25,7 @@ import (
 type correlationKey struct{}
 
 // ContextWithCorrelationID returns a derived context with the correlation IDs if they were not already in there.
+// NOTE: This function assumes that the provided correlation IDs do not contain any duplicates.
```
htdvisser: It's not very expensive to check for + remove duplicates, especially since we're already sorting the thing. Maybe consider doing that instead of making the caller responsible for it.

pkg/events/redis/codec.go
```diff
@@ -47,8 +47,11 @@ func decodeEventData(enc string, evt *ttnpb.Event) error {
 	if err != nil {
 		return err
 	}
-	// NOTE: We're merging additional event data into an event that may already contain fields.
-	return proto.UnmarshalMerge(bpb, evt)
+	// NOTE: We override the contents of `evt` with the contents `enc` explicitly.
+	// proto.UnmarshalMerge has append semantics for slices, and as such identifiers
+	// and event visibility will be duplicated, as `evt` most likely contains them
+	// from the sparse event representation.
+	return proto.Unmarshal(bpb, evt)
```
htdvisser, line 54: Doesn't this unset some of the fields that are only in the metadata, and not in the sparse event? As far as I can tell, the sparse event doesn't contain `name`, `unique_id`, so aren't we un-setting those fields now?

adriansmares: The sparse event is `evt`, not `enc`. When we `proto.Unmarshal`, we inherently do a `evt.Reset()`. As such, we are replacing the sparse representation with the full representation.

The function here (`decodeEventData`) has 3 call sites:
- `subscribeTask` - it unmarshals the contents that come via the Redis channels into an empty proto - there is nothing to merge:
https://github.com/TheThingsNetwork/lorawan-stack/blob/127cf5ad55bb0eacda53741ea472a204acf26054/pkg/events/redis/redis.go#L152-L158
- `LoadEvent` - same behavior, load into empty proto without a real merge happening:
https://github.com/TheThingsNetwork/lorawan-stack/blob/127cf5ad55bb0eacda53741ea472a204acf26054/pkg/events/redis/store.go#L110-L121
- `loadEventData` - this is the only one that actually has some partial event representations, but the binary data is complete, as it comes from the event stored by the unique ID:
https://github.com/TheThingsNetwork/lorawan-stack/blob/127cf5ad55bb0eacda53741ea472a204acf26054/pkg/events/redis/store.go#L79-L101

In all 3 cases, the `bpb` here contains a full binary representation of the event - no fields are going to be lost, as we're always storing the full event:
https://github.com/TheThingsNetwork/lorawan-stack/blob/127cf5ad55bb0eacda53741ea472a204acf26054/pkg/events/redis/store.go#L378-L388
https://github.com/TheThingsNetwork/lorawan-stack/blob/127cf5ad55bb0eacda53741ea472a204acf26054/pkg/events/redis/store.go#L64-L69
https://github.com/TheThingsNetwork/lorawan-stack/blob/127cf5ad55bb0eacda53741ea472a204acf26054/pkg/events/redis/codec.go#L28-L38

pkg/events/correlation_context.go
adriansmares: e84bcc956c18f783bb3bd235c6adc604c29cdea5

htdvisser (approved):

PR #5471, opened 2022-05-20
all: Remove some non-pointer usage of proto structures

michalborkowski96:
#### Summary
Open-source version of https://github.com/TheThingsIndustries/lorawan-stack/pull/3206

#### Changes
Remove some non-pointer usage of proto structures

#### Testing
Automatic tests

##### Regressions
None

#### Notes for Reviewers
Open-source version of https://github.com/TheThingsIndustries/lorawan-stack/pull/3206

#### Checklist

- [ ] Scope: The referenced issue is addressed, there are no unrelated changes.
- [ ] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [ ] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #5535, opened 2022-06-13
Fix correlation IDs race in the Gateway Server

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/5471
References https://sentry.io/organizations/the-things-industries/issues/3346136069
References https://sentry.io/organizations/the-things-industries/issues/3345934395

#### Changes

- Ensure that we actually shallow clone the uplink message in the Gateway Server upstream handler
  - The context here is that each upstream handler needs to edit the uplink message in order to add its correlation IDs. Since this would be a concurrent write, we used to do a shallow clone of the uplink message.
  - Doing shallow clones is no longer supported in the new protobuf library, so we have to manually copy each fields.
  - If we do not do the shallow clone, bad things happen during marshalling (the library pre-allocates space for the contents of the message, the message is concurrently modified, panic ensues)
- Do not append correlation IDs using `append`, but use the `events` package function instead
  - This is mostly cosmetical, but it ensures that appended entries are sorted correctly. 
  - We were and still are a bit inconsistent here. I will do some small fixes in the future, but this should suffice for now.

#### Testing

Unit testing.

##### Regressions

This was already broken.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -769,6 +769,25 @@ func (host *upstreamHost) handlePacket(ctx context.Context, item any) {
 			registerForwardStatus(ctx, gtw, msg, host.name)
 		}
 	case *ttnpb.TxAcknowledgment:
+		correlationIDs := make([]string, 0, len(msg.CorrelationIds)+len(msg.DownlinkMessage.GetCorrelationIds()))
+		correlationIDs = append(correlationIDs, msg.CorrelationIds...)
+		correlationIDs = append(correlationIDs, msg.DownlinkMessage.GetCorrelationIds()...)
+		ctx := events.ContextWithCorrelationID(ctx, correlationIDs...)
+		msg = &ttnpb.TxAcknowledgment{
+			CorrelationIds:  events.CorrelationIDsFromContext(ctx),
+			Result:          msg.Result,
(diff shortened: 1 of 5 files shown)


pkg/gatewayserver/gatewayserver.go
```diff
@@ -705,15 +705,26 @@ func (host *upstreamHost) handlePacket(ctx context.Context, item any) {
 	ctx = events.ContextWithCorrelationID(ctx, host.correlationID)
 	logger := log.FromContext(ctx)
 	gtw := host.gtw
+	// Each concurrent upstream host will receive the message and edit it in order
+	// to append the correlation IDs. This would be a concurrent write, so we are
+	// creating a shallow message copy in order to safely edit the correlation IDs.
 	switch msg := item.(type) {
 	case *ttnpb.GatewayUplinkMessage:
-		up := msg.Message
```
adriansmares, line 710: This used to be a shallow clone using `*msg.Message`, but these are not allowed anymore. As such, we need to manually clone the message.

KrishnaIyer (approved):
LGTM

johanstokking (approved):
Looks good to me too.

I think it's good that some of us are more explicitly adding code comments so our future selves are aware of why we're doing certain things.

PR #5536, opened 2022-06-15
Add RPC to batch retrieve gateway connection stats

### LINKED ISSUE #5398: Add gateway stats retrieval batch RPC
#### Summary

We should have a batch retrieval operation for the gateway connection stats. 

#### Why do we need this?

Clients which need to retrieve multiple gateway stats at the same time (i.e the Console in the gateway list view) currently retrieve the status of each individual gateway over multiple roundtrips. While the roundtrips themselves are done in parallel, they are rather inefficient in the long term, as one page retrieval amplifies to at least 20 parallel reads. 

The `GetGatewayConnectionStats` RPC, which returns the stats for a singular gateway:
https://github.com/TheThingsNetwork/lorawan-stack/blob/afeb84b0ac3155f94bcb175bcfe4c418daba660f/api/gatewayserver.proto#L102-L106

A batch variant of the RPC, which can return multiple statuses at once.

#### Environment

`v3.19`, but not relevant.

Add a batch variant of `GetGatewayConnectionStats`. The request contains a list of `GatewayIdentifiers` (which should be validated to have a maximum length), and the response contains a list of `GatewayConnectionStats`, which match the stats with the identifiers.
The HTTP binding probably should be `/gs/gateways/connection/stats`.

At RPC implementation level, we need to check the rights for each individual gateway.
At store level, we can just build a pipeline that reads the protos associated with these gateway identifiers in a single roundtrip.

Unit testing.

Can review.

KrishnaIyer:
#### Summary

Closes #5398 

#### Changes

- Add new `BatchGetGatewayConnectionStats`.
- Implement GS registry interface
- Add CLI support

#### Testing

- [x] UT
- [x] E2E

```
End to End test with two (simulated) gateways
----------------------------------------------

1. Gateways not created

$./ttn-lw-cli gtw get-connection-stats test-gtw-2 --gateway-ids test-gtw,test-gtw-1
WARN    Using insecure connection to OAuth server
WARN    Using insecure connection to API
error:pkg/identityserver/gormstore:gateway_not_found (gateway `test-gtw-1` not found)
    gateway_id=test-gtw-1
    correlation_id=2905f102c2aa498aaa53223bd22dcf30

2. Gateways created but not connected

$ ./ttn-lw-cli gtw get-connection-stats test-gtw-2 --gateway-ids test-gtw,test-gtw-1
WARN    Using insecure connection to OAuth server
WARN    Using insecure connection to API
{}

3. Gateway 1 connected

$./ttn-lw-cli gtw get-connection-stats test-gtw-2 --gateway-ids test-gtw,test-gtw-1
WARN    Using insecure connection to OAuth server
WARN    Using insecure connection to API
{
  "entries": {
    "test-gtw": {
      "connected_at": "2022-07-13T11:44:14.364400Z",
      "protocol": "ws",
      "gateway_remote_address": {
        "ip": "127.0.0.1"
      }
    }
  }
}

4. Both gateways connected

$ ./ttn-lw-cli gtw get-connection-stats test-gtw-2 --gateway-ids test-gtw,test-gtw-1
WARN    Using insecure connection to OAuth server
WARN    Using insecure connection to API
{
  "entries": {
    "test-gtw": {
      "connected_at": "2022-07-13T11:44:14.364400Z",
      "protocol": "ws",
      "gateway_remote_address": {
        "ip": "127.0.0.1"
      }
    },
    "test-gtw-1": {
      "connected_at": "2022-07-13T11:44:34.619971Z",
      "protocol": "ws",
      "gateway_remote_address": {
        "ip": "127.0.0.1"
      }
    }
  }
}


5. Gateway 2 disconnected

$ ./ttn-lw-cli gtw get-connection-stats test-gtw-2 --gateway-ids test-gtw,test-gtw-1
WARN    Using insecure connection to OAuth server
WARN    Using insecure connection to API
{
  "entries": {
    "test-gtw": {
      "connected_at": "2022-07-13T11:44:14.364400Z",
      "protocol": "ws",
      "gateway_remote_address": {
        "ip": "127.0.0.1"
      }
    },
    "test-gtw-1": {
      "disconnected_at": "2022-07-13T11:44:45.462092Z",
      "protocol": "ws",
      "gateway_remote_address": {
        "ip": "127.0.0.1"
      }
    }
  }
}

6. Gateway 1 disconnected

$ ./ttn-lw-cli gtw get-connection-stats test-gtw-2 --gateway-ids test-gtw,test-gtw-1
WARN    Using insecure connection to OAuth server
WARN    Using insecure connection to API
{
  "entries": {
    "test-gtw": {
      "disconnected_at": "2022-07-13T11:44:49.544901Z",
      "protocol": "ws",
      "gateway_remote_address": {
        "ip": "127.0.0.1"
      }
    },
    "test-gtw-1": {
      "disconnected_at": "2022-07-13T11:44:45.462092Z",
      "protocol": "ws",
      "gateway_remote_address": {
        "ip": "127.0.0.1"
      }
    }
  }
}


7. Regression

$ ./ttn-lw-cli gtw get-connection-stats test-gtw
WARN    Using insecure connection to OAuth server
WARN    Using insecure connection to API
{
  "disconnected_at": "2022-07-13T11:44:49.544901Z",
  "protocol": "ws",
  "gateway_remote_address": {
    "ip": "127.0.0.1"
  }
}


8. Filter on paths

./ttn-lw-cli gtw get-connection-stats --gateway-ids test-gtw,test-gtw-1 --paths protocol
WARN  Using insecure connection to OAuth server
WARN  Using insecure connection to API
{
  "entries": {
    "test-gtw": {
      "protocol": "ws"
    },
    "test-gtw-1": {
      "protocol": "ws"
    }
  }
}

$ ./ttn-lw-cli gtw get-connection-stats --gateway-ids test-gtw,test-gtw-1 --paths protocol,disconnected_at
WARN  Using insecure connection to OAuth server
WARN  Using insecure connection to API
{
  "entries": {
    "test-gtw": {
      "disconnected_at": "2022-07-13T11:44:49.544901Z",
      "protocol": "ws"
    },
    "test-gtw-1": {
      "disconnected_at": "2022-07-13T11:44:45.462092Z",
      "protocol": "ws"
    }
  }
}


9. Check invalid paths
 
./ttn-lw-cli gtw get-connection-stats --gateway-ids test-gtw,test-gtw-1 --paths protocol,disconnected_at,wrong_path
WARN  Using insecure connection to OAuth server
WARN  Using insecure connection to API
    forbidden_paths=[wrong_path]
    correlation_id=ceebc010321842b4b35f310f67db2b98
```

##### Regressions

Possible regression in `get-connection-stats` CLI command. Test case 7 covers this.

#### Checklist


cmd/ttn-lw-cli/commands/gateways.go
```diff
@@ -471,6 +472,44 @@ var (
 			return io.Write(os.Stdout, config.OutputFormat, res)
 		},
 	}
+	gatewaysBatchConnectionStats = &cobra.Command{
+		Use:     "batch-get-connection-stats",
+		Aliases: []string{"batch-connection-stats", "batch-cnx-stats"},
+		Short:   "Get connection stats for a batch of gateways",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ids, err := cmd.Flags().GetString("gateway-ids")
```
adriansmares: You can use `GetStringSlice` (and register the parameter as a string slice) instead.

References https://github.com/spf13/pflag/blob/85dd5c8bc61cfa382fecd072378089d4e856579d/string_slice.go#L88-L95 .

pkg/gatewayserver/redis/registry.go
```diff
@@ -86,3 +86,32 @@ func (r *GatewayConnectionStatsRegistry) Get(ctx context.Context, ids *ttnpb.Gat
...
+// or is connected to a different cluster, are ignored.
+func (r *GatewayConnectionStatsRegistry) BatchGet(ctx context.Context,
+	ids []*ttnpb.GatewayIdentifiers,
+) ([]*ttnpb.GatewayConnectionStatsEntry, error) {
+	ret := make([]*ttnpb.GatewayConnectionStatsEntry, 0)
+	if _, err := r.Redis.TxPipelined(ctx, func(p redis.Pipeliner) error {
+		for _, gtwIDs := range ids {
+			uid := unique.ID(ctx, gtwIDs)
+			result := &ttnpb.GatewayConnectionStats{}
+			if err := ttnredis.GetProto(ctx, r.Redis, r.key(uid)).ScanProto(result); err != nil {
```
adriansmares: This won't use the pipeline, and the smoking gun is that nothing references `p`.

The way Redis pipelines work is as follows:
1. You call `{Tx}Pipelined` with a function that will _queue_ commands.
2. The function will be called with this `redis.Pipeliner` to which you add commands, but do not read responses from yet, as the commands have not been sent to the server.
3. The commands are then sent to the server in a batched manner, then the responses come back and `TxPipelined` returns.
4. You can now access the commands that you queued and read their result. 

The description above is what I meant when writing the issue description (sorry it wasn't detailed enough). You can see an example of this in the NS matching registry:
https://github.com/TheThingsNetwork/lorawan-stack/blob/f6e9d73bd6d6ccf69474dcc47f34c5a68c70dc44/pkg/networkserver/redis/registry.go#L539-L555
1. The commands are declared.
2. Inside the pipeline function we just assign them (but we don't read the result).
3. After the `TxPipelined` runs, we actually read the result.

With all of that being said, you can actually simplify your life massively using [`MGET`](https://redis.io/commands/mget/). You just have to build a slice of keys (so basically calling `r.key` on each unique ID), call `MGET` on them, then walk the result slice of results and unmarshal the protos using `redis.UnmarshalProto`. 

You can see an example of how this would work here: https://github.com/TheThingsIndustries/lorawan-stack/blob/afc07d9b357086064cfb46b26ae06cfaf6cdfbcc/pkg/redis/redis.tti.go#L48-L66 .

KrishnaIyer: Oh nice. Thanks for the detailed explanation. I'm definitely a noob with Redis.

pkg/gatewayserver/redis/registry.go
```diff
@@ -86,3 +87,45 @@ func (r *GatewayConnectionStatsRegistry) Get(ctx context.Context, ids *ttnpb.Gat
 	}
 	return result, nil
 }
+
+// BatchGet returns the connection stats for a batch of gateways.
+// NotFound errors indicating that the gateway is either not connected
+// or is connected to a different cluster, are ignored.
+func (r *GatewayConnectionStatsRegistry) BatchGet(ctx context.Context,
```
adriansmares: Nit: One parameter per line.

pkg/gatewayserver/redis/registry.go
```diff
@@ -86,3 +87,45 @@ func (r *GatewayConnectionStatsRegistry) Get(ctx context.Context, ids *ttnpb.Gat
...
+	ids []*ttnpb.GatewayIdentifiers,
+) ([]*ttnpb.GatewayConnectionStatsEntry, error) {
+	ret := make([]*ttnpb.GatewayConnectionStatsEntry, 0)
```
adriansmares: Very minor nit: Since you know the length of the result, both of those can become `make([]*ttnpb.GatewayConnectionStatsEntry, 0, len(ids))` and `make([]string, 0, len(ids))`

adriansmares (approved):
Cool, now it is really clean.

api/gateway.proto
```diff
@@ -432,3 +432,8 @@ message GatewayConnectionStats {
 
   // next: 13
 }
+
+message GatewayConnectionStatsEntry{
+  GatewayIdentifiers gateway_ids = 1;
+  GatewayConnectionStats gateway_connection_stats = 2;
+}
```
htdvisser: You can also move this message inside `BatchGetGatewayConnectionStatsResponse`

cmd/ttn-lw-cli/commands/gateways.go
```diff
@@ -471,6 +471,38 @@ var (
...
+				&ttnpb.BatchGetGatewayConnectionStatsRequest{
+					GatewayIds: gtwIDs,
+				})
+			if err != nil {
+				return err
+			}
+
+			return io.Write(os.Stdout, config.OutputFormat, res)
+		},
+	}
```
htdvisser: Why not make this part of the existing `get-connection-stats`? We can just adjust the logic if more than one gateway ID is passed, right?

KrishnaIyer: Good one 👍

api/gateway.proto
KrishnaIyer: That would make the type a bit long to use; `ttnpb.BatchGetGatewayConnectionStatsResponse_GatewayConnectionStatsEntry`.

api/gateway.proto
```diff
@@ -432,3 +432,8 @@ message GatewayConnectionStats {
 
   // next: 13
 }
+
+message GatewayConnectionStatsEntry{
```
johanstokking:
```suggestion
message GatewayConnectionStatsEntry {
```

api/gatewayserver.proto
```diff
@@ -96,6 +96,14 @@ service NsGs {
   rpc ScheduleDownlink(DownlinkMessage) returns (ScheduleDownlinkResponse);
 }
 
+message BatchGetGatewayConnectionStatsRequest{
+  repeated GatewayIdentifiers gateway_ids = 1 [(validate.rules).repeated.min_items = 1, (validate.rules).repeated.max_items = 100];
+}
+
+message BatchGetGatewayConnectionStatsResponse{
+  repeated GatewayConnectionStatsEntry entries = 1;
+}
```
johanstokking:
```suggestion
message BatchGetGatewayConnectionStatsRequest {
  repeated GatewayIdentifiers gateway_ids = 1 [(validate.rules).repeated.min_items = 1, (validate.rules).repeated.max_items = 100];
}

message BatchGetGatewayConnectionStatsResponse {
  repeated GatewayConnectionStatsEntry entries = 1;
}
```

api/gateway.proto
```diff
@@ -432,3 +432,8 @@ message GatewayConnectionStats {
...
+  GatewayIdentifiers gateway_ids = 1;
+  GatewayConnectionStats gateway_connection_stats = 2;
```
johanstokking: Can't we do a `map<string, GatewayConnectionStats>` with UID as key? Or is this better this way?

cmd/ttn-lw-cli/commands/gateways.go
```diff
@@ -434,41 +434,68 @@ var (
...
-
 			gs, err := api.Dial(ctx, config.GatewayServerGRPCAddress)
 			if err != nil {
 				return err
 			}
 
-			res, err := ttnpb.NewGsClient(gs).GetGatewayConnectionStats(ctx, gtwID)
-			if err != nil {
-				return err
+			var result interface{}
```
johanstokking:
```suggestion
			var result any
```

api/gatewayserver.proto
johanstokking: Why no field mask?

cmd/ttn-lw-cli/commands/gateways.go
```diff
@@ -434,41 +434,68 @@ var (
...
+				var gtwIDs []*ttnpb.GatewayIdentifiers
+				for _, id := range ids {
+					gtwIDs = append(gtwIDs, &ttnpb.GatewayIdentifiers{
+						GatewayId: id,
+					})
+				}
+				res, err := ttnpb.NewGsClient(gs).BatchGetGatewayConnectionStats(ctx,
+					&ttnpb.BatchGetGatewayConnectionStatsRequest{
+						GatewayIds: gtwIDs,
+					})
```
johanstokking: Why no field mask?

pkg/component/grpc_test.go
```diff
@@ -71,6 +71,15 @@ func (gs *gsImplementation) GetGatewayConnectionStats(ctx context.Context, _ *tt
 	return &ttnpb.GatewayConnectionStats{}, nil
 }
 
+func (*gsImplementation) BatchGetGatewayConnectionStats(ctx context.Context,
+	_ *ttnpb.BatchGetGatewayConnectionStatsRequest,
+) (*ttnpb.BatchGetGatewayConnectionStatsResponse, error) {
+	if err := clusterauth.Authorized(ctx); err != nil {
+		return nil, err
+	}
```
johanstokking, line 79: Why test with cluster auth?

johanstokking (commented):
Looks good but few questions

api/gateway.proto
KrishnaIyer: Yup this ends up being cleaner.

api/gatewayserver.proto
KrishnaIyer: Yeah that's also a good suggestion.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ For details about compatibility between different releases, see the **Commitment
 
 - More fields were added to the csv end-device migration procedure. The details on which fields were added can be found [here](https://www.thethingsindustries.com/docs/getting-started/migrating/device-csv/).
 - Authorization management in the Account app.
+- `BatchGetGatewayConnectionStats` RPC to fetch Gateway Connection Stats for a batch of gateways.
+- Gateway remote address to gateway connection statistics.
```
KrishnaIyer: This is a fix for a wrong entry in a previous PR.

pkg/component/grpc_test.go
KrishnaIyer: This is a copy of https://github.com/TheThingsNetwork/lorawan-stack/blob/34b15a6e0e3e7a3735707e56c971b775fbc1542d/pkg/component/grpc_test.go#L67-L72 
I'm satisfying the interface. I didn't implement this so I don't have the context.

cmd/ttn-lw-cli/commands/gateways.go
```diff
@@ -434,41 +435,76 @@ var (
...
 			}
 
-			is, err := api.Dial(ctx, config.IdentityServerGRPCAddress)
+			gtwID, err := getGatewayID(cmd.Flags(), args, true)
 			if err != nil {
-				return err
+				if !errors.IsInvalidArgument(err) || len(ids) == 0 {
+					return err
+				}
 			}
```
htdvisser: I think it's better to solve this as follows:

- Define the new `gateway-ids` flag
- Replace the existing `gateway-id` flag with an alias to the `gateway-ids` flag
- Also support multiple IDs in arguments

cmd/ttn-lw-cli/commands/gateways.go
```diff
@@ -434,41 +435,76 @@ var (
...
+					&ttnpb.BatchGetGatewayConnectionStatsRequest{
+						GatewayIds: gtwIDs,
+						FieldMask: &types.FieldMask{
+							Paths: paths,
+						},
+					})
+				if err != nil {
+					return err
+				}
+				result = res
```
htdvisser: Why not just `return io.Write(os.Stdout, config.OutputFormat, res)`? Then you also don't need the `else`.

KrishnaIyer:
Test for CLI Combinations
```
$ ./ttn-lw-cli gtw get-connection-stats test-gtw
WARN  Using insecure connection to OAuth server
WARN  Using insecure connection to API
{
  "disconnected_at": "2022-07-13T11:44:49.544901Z",
  "protocol": "ws",
  "gateway_remote_address": {
    "ip": "127.0.0.1"
  }
}


$ ./ttn-lw-cli gtw get-connection-stats --gateway-id test-gtw
WARN  Using insecure connection to OAuth server
WARN  Using insecure connection to API
{
  "disconnected_at": "2022-07-13T11:44:49.544901Z",
  "protocol": "ws",
  "gateway_remote_address": {
    "ip": "127.0.0.1"
  }
}

$ ./ttn-lw-cli gtw get-connection-stats --gateway-ids test-gtw
WARN  Using insecure connection to OAuth server
WARN  Using insecure connection to API
{
  "disconnected_at": "2022-07-13T11:44:49.544901Z",
  "protocol": "ws",
  "gateway_remote_address": {
    "ip": "127.0.0.1"
  }
}


$ ./ttn-lw-cli gtw get-connection-stats --gateway-ids test-gtw,test-gtw-1
WARN  Using insecure connection to OAuth server
WARN  Using insecure connection to API
{
  "entries": {
    "test-gtw": {
      "disconnected_at": "2022-07-13T11:44:49.544901Z",
      "protocol": "ws",
      "gateway_remote_address": {
        "ip": "127.0.0.1"
      }
    },
    "test-gtw-1": {
      "disconnected_at": "2022-07-13T11:44:45.462092Z",
      "protocol": "ws",
      "gateway_remote_address": {
        "ip": "127.0.0.1"
      }
    }
  }
}

$ ./ttn-lw-cli gtw get-connection-stats test-gtw test-gtw-1
WARN  Using insecure connection to OAuth server
WARN  Using insecure connection to API
{
  "entries": {
    "test-gtw": {
      "disconnected_at": "2022-07-13T11:44:49.544901Z",
      "protocol": "ws",
      "gateway_remote_address": {
        "ip": "127.0.0.1"
      }
    },
    "test-gtw-1": {
      "disconnected_at": "2022-07-13T11:44:45.462092Z",
      "protocol": "ws",
      "gateway_remote_address": {
        "ip": "127.0.0.1"
      }
    }
  }
}


$ ./ttn-lw-cli gtw get-connection-stats
WARN  Using insecure connection to OAuth server
WARN  Using insecure connection to API
error:cmd/ttn-lw-cli/commands:no_gateway_id (no gateway ID set)
    correlation_id=640a684b72354a5b93db492076374666
```

api/gatewayserver.proto
```diff
@@ -104,4 +116,15 @@ service Gs {
...
     };
   };
+
+  // Get statistics about gateway connections to the Gateway Server of a batch of gateways.
+  // This is not persisted between reconnects.
+  // Gateways that are not connected or are part of a different cluster are ignored.
+  // It's upto the client to make sure that the gateways are in the requested cluster.
+  rpc BatchGetGatewayConnectionStats(BatchGetGatewayConnectionStatsRequest) returns (BatchGetGatewayConnectionStatsResponse) {
+    option (google.api.http) = {
+      post: "/gs/gateways/connection/stats"
```
johanstokking, line 127: Did we blacklist the gateway ID `connection`?

api/gatewayserver.proto
```diff
@@ -96,6 +97,17 @@ service NsGs {
...
 
+message BatchGetGatewayConnectionStatsRequest {
+  repeated GatewayIdentifiers gateway_ids = 1 [(validate.rules).repeated.min_items = 1, (validate.rules).repeated.max_items = 100];
+  // The names of the gateway stats fields that should be returned.
+  // This mask will be applied on each entry returned.
+  google.protobuf.FieldMask field_mask = 2;
+}
+
+message BatchGetGatewayConnectionStatsResponse {
+  map<string,GatewayConnectionStats> entries = 1;
```
johanstokking, line 109: Actually apologies for the confusion but UID isn't necessary here; it can just be the gateway ID. Please also document this in godoc.

pkg/gatewayserver/redis/registry.go
```diff
@@ -86,3 +87,62 @@ func (r *GatewayConnectionStatsRegistry) Get(ctx context.Context, ids *ttnpb.Gat
...
+					return nil, err
+				}
+			}
+
+			// The result of MGet is in the same order as the input keys passed to it.
+			// MGet inserts "nil" values for keys that don't have values, thereby maintaining the order.
+			// So we can use the index of the result to correlate the gateway IDs.
+			ret[unique.ID(ctx, ids[i])] = stats
+		default:
+			log.FromContext(ctx).Warnf("Invalid %T element in stats payloads", val)
```
johanstokking: Please always use log fields and no formatted log messages

api/gatewayserver.proto
KrishnaIyer: That's a good one.

api/gatewayserver.proto
```diff
@@ -104,4 +117,15 @@ service Gs {
       get: "/gs/gateways/{gateway_id}/connection/stats"
     };
   };
+
+  // Get statistics about gateway connections to the Gateway Server of a batch of gateways.
+  // This is not persisted between reconnects.
+  // Gateways that are not connected or are part of a different cluster are ignored.
+  // It's upto the client to make sure that the gateways are in the requested cluster.
```
nicholaspcr: `It is up to the client ...`

cmd/ttn-lw-cli/commands/gateways.go
```diff
@@ -432,42 +435,81 @@ var (
...
+Alternatively use the --gateway-ids flag.
+If both the parameter and the flag are provided, the flag is ignored.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			gtwID, err := getGatewayID(cmd.Flags(), args, true)
+			var (
+				ids = args
+				err error
+			)
+			if len(ids) == 0 {
+				raw, err := cmd.Flags().GetString("gateway-ids")
```
nicholaspcr, line 451: Couldn't this be `GetStringSlice`?

pkg/gatewayserver/grpc.go
```diff
@@ -48,3 +48,60 @@ func (gs *GatewayServer) GetGatewayConnectionStats(ctx context.Context, ids *ttn
...
+// FieldMask paths can be used directly since they are sanitized by the middleware.
+func (gs *GatewayServer) BatchGetGatewayConnectionStats(
+	ctx context.Context,
+	req *ttnpb.BatchGetGatewayConnectionStatsRequest,
+) (*ttnpb.BatchGetGatewayConnectionStatsResponse, error) {
+	for _, ids := range req.GatewayIds {
+		if err := gs.entityRegistry.AssertGatewayRights(ctx, ids, ttnpb.Right_RIGHT_GATEWAY_STATUS_READ); err != nil {
+			return nil, err
+		}
+	}
```
nicholaspcr, line 74: Returning an error at the first encounter of a gateway without `STATUS_READ` right forces the client into a process of trial and error until he has the right of all the requested gateways. I imagine this wouldn't be a problem for the console but might be for the CLI user.

Wouldn't it be best to have a list of gtws in which he doesn't have the right to read the stats?

nicholaspcr (commented):
Small comments but looks pretty good.

Also, I think the commit messages for `52b94bc` and `3e824d3` are swapped.

cmd/ttn-lw-cli/commands/gateways.go
KrishnaIyer: Please see https://github.com/TheThingsNetwork/lorawan-stack/pull/5536#discussion_r920080643

pkg/gatewayserver/grpc.go
KrishnaIyer: This will not scale. If a user makes a request with 100 gateway IDs and  doesn't have the rights to any of them, we will make 100 API calls to the IS and then return that to the caller.

cmd/ttn-lw-cli/commands/gateways.go
htdvisser: Getting the flag as a string slice would actually make sense

pkg/gatewayserver/grpc.go
htdvisser: I do think it would be a good idea (especially if we want more bulk operations in the future) to also implement a bulk rights check.

cmd/ttn-lw-cli/commands/gateways.go
nicholaspcr: I'm falling to see how the comment relates to the usage of `GetStringSlice` on the flag.  It should work with just one value and would be fetched only when ids aren't passed as an argument.

pkg/gatewayserver/grpc.go
nicholaspcr: +1 on the idea of having a bulk rights check.

>This will not scale. If a user makes a request with 100 gateway IDs and doesn't have the rights to any of them, we will make 100 API calls to the IS and then return that to the caller.

True, indeed it would be a quite bad behaviour.

cmd/ttn-lw-cli/commands/gateways.go
KrishnaIyer: `gateway-ids` is an alias to `gateway-id` which is an existing flag.

```
./ttn-lw-cli gtw get-connection-stats --help
Get connection stats for a (group of) gateway(s).
The command arguments support multiple gateway IDs.
Alternatively use the --gateway-ids flag.
If both the parameter and the flag are provided, the flag is ignored.

Usage:
  ttn-lw-cli gateways get-connection-stats [gateway-ids] [flags]

Aliases:
  get-connection-stats, connection-stats, cnx-stats, stats

Flags:
      --gateway-eui string
      --gateway-id string
      --gateway-ids string
```

If I reverse that alias to use `StringSlice`

```
Flags:
      --gateway-id strings    comma separated list of gateway IDs to batch get stats (default [])
      --gateway-ids strings   comma separated list of gateway IDs to batch get stats
```

You're either wrong labelling `gateway-id` as a slice or `gateway-ids` as a string value. I don't really prefer one over the other. If I really have to use `StringSlice` here, I can. But I don't think it adds much.

pkg/gatewayserver/grpc.go
KrishnaIyer: Yes bulk rights is the preferred way and I believe we have an issue for that somewhere. The implementation is out of scope for this PR.

cmd/ttn-lw-cli/commands/gateways.go
nicholaspcr: I see what is the problem now. I also don't have a preference of one over the other, I'll leave it up to you. Either one is fine to me.

johanstokking (approved):
Nothing to add to the above

api/gatewayserver.proto
nicholaspcr: I think I mixed a `nit` and an actual comment, the `It's` -> `It is` was in fact a `nit`.

@KrishnaIyer I'm not so sure anymore but I think `upto` cannot be used as one word, has to be `up to`.

PR #5547, opened 2022-06-20
Restore development database from file

johanstokking:
#### Summary

Restore development database from file
Open source backport from https://github.com/TheThingsIndustries/lorawan-stack/pull/3273

#### Changes

- Format files; follow conventions from `/prettierrc.yaml`
- Instead of recreating the database from a template, restore from a (cached) file. This is needed for the referenced work
- Support Postgres to restart during its first initialization. This happens when TimescaleDB loads; it tunes and reconfigures Postgres and then restarts it

#### Testing

Local testing and CI

##### Regressions

None expected

#### Notes for Reviewers

@kschiffer is the main reviewer

#### Checklist

### DIFF
--- a/cypress/plugins/tasks.js
+++ b/cypress/plugins/tasks.js
@@ -22,7 +22,6 @@ const codeCoverageTask = require('@cypress/code-coverage/task')
 
 const isCI = process.env.CI === 'true' || process.env.CI === '1'
 const devDatabase = 'ttn_lorawan_dev'
-const seedDatabase = 'tts_seed'
 
 const pgConfig = {
   user: 'root',
@@ -93,7 +92,9 @@ const sqlTask = on => {
     },
     dropAndSeedDatabase: () => {
       postgresContainer.stdin.write(
-        `dropdb --if-exists --force ${devDatabase}; createdb -T ${seedDatabase} ${devDatabase};\n`,
+        `dropdb --if-exists --force ${devDatabase} 2> /dev/null; ` +
+          `createdb ${devDatabase}; ` +
+          `pg_restore -Fc -d ${devDatabase} /var/lib/ttn-lorawan/cache/database.pgdump;\n`,
       )
       return Promise.all([
         new Promise(resolve => exec('tools/bin/mage dev:redisFlush', resolve)),
--- a/cypress/plugins/tasks.js
+++ b/cypress/plugins/tasks.js
@@ -12,16 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-const { execSync, exec, spawn } = require('child_process')
+const childProcess = require('child_process')
 const fs = require('fs')
 const path = require('path')
 
+const util = require('util')
 const { Client } = require('pg')
 const yaml = require('js-yaml')
 const codeCoverageTask = require('@cypress/code-coverage/task')
 
 const isCI = process.env.CI === 'true' || process.env.CI === '1'
-const devDatabase = 'ttn_lorawan_dev'
 
 const pgConfig = {
   user: 'root',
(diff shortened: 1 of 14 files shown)


tools/mage/dev.go
```diff
@@ -168,18 +159,50 @@ func (Dev) DBStart() error {
...
+		for j := 0; j < 30; j++ {
+			time.Sleep(time.Second)
+			_, err := sh.Exec(nil, nil, nil, "docker-compose", flags...)
+			isReady := err == nil
+			switch {
+			case wasReady && !isReady:
+				if mg.Verbose() {
+					fmt.Println("Postgres is not ready anymore")
+				}
+				continue nextCycle
```
adriansmares, line 186: Shouldn't you reset the `successiveReady` counter here ? Otherwise it is not really successive.

adriansmares (approved):

tools/mage/dev.go
johanstokking: No because `nextCycle` is the outer loop, and `successiveReady` is in the scope of that outer loop body.

cypress/plugins/tasks.js
```diff
@@ -84,27 +71,26 @@ const stackConfigTask = (_, config) => {
...
-          postgresContainer.on('close', resolve)
-        }),
+    dropAndSeedDatabase: async () => {
+      const exec = util.promisify(childProcess.exec)
+      const res = await Promise.all([
+        exec('tools/bin/mage dev:sqlRestore'),
+        exec('tools/bin/mage dev:redisFlush'),
       ])
+      const err = res
+        .filter(e => !!e.stderr)
```
kschiffer:
```suggestion
        .filter(e => Boolean(e.stderr))
```
More conventional.

cypress/plugins/tasks.js
```diff
@@ -84,27 +71,26 @@ const stackConfigTask = (_, config) => {
...
+        .map(e => e.stderr)
```
kschiffer, line 87:
```suggestion
        .map(e => e.stderr.toString())
```
The outs are byte arrays that need to be converted to strings.

kschiffer (changes requested):

cypress/plugins/tasks.js
johanstokking: I've got them as `string`s actually. Where do you see this?

kschiffer: Oh ok, seems like `exec` is already taking care of that then. Disregard.

cypress/plugins/tasks.js
```diff
@@ -84,27 +70,26 @@ const stackConfigTask = (_, config) => {
...
       return res
     },
-    dropAndSeedDatabase: () => {
-      postgresContainer.stdin.write(
-        `dropdb --if-exists --force ${devDatabase}; psql template1 -c "CREATE DATABASE ${devDatabase} TEMPLATE ${seedDatabase};"\n`,
-      )
-      return Promise.all([
-        new Promise(resolve => exec('tools/bin/mage dev:redisFlush', resolve)),
-        new Promise(resolve => {
-          postgresContainer.stdout.on('data', resolve)
```
johanstokking, line 101: Another issue with this is that the promises resolves on the first data. That happened to work because the first and only output was from the `psql` command and after that everything was done. However, the `dev:sqlRestore` mage step may do multiple things with output in various stages, and all that has to be awaited before continuing. So I don't think there's going around spawning a new process because that's the only way to await everything.

kschiffer: If you use the mage command then I guess you can just use `.on('close', () => { … })`.

kschiffer (approved):

PR #5554, opened 2022-06-22
Backport fix for getting gateway rights

johanstokking:
#### Summary

Backports https://github.com/TheThingsIndustries/lorawan-stack/pull/3268

cc @KrishnaIyer 

#### Changes

Backport gateway EUI reference issue.

#### Testing

CI

##### Regressions

None expected

#### Notes for Reviewers

I don't really get what this resolves, but this should be backported.

#### Checklist


KrishnaIyer:
Ah I was sure I had a backport PR for this.
> I don't really get what this resolves,

When checking rights, we should use only the ID to look up the gateway so we discarded the EUI. But setting the EUI of the argument to nil set the EUI in the original RPC request to nil since this was passed as a pointer during one of our GoGo proto migration. This means that if you want to update the EUI of a gateway (ex: during the gateway claiming flow), the EUI becomes nil during the rights check.

PR #5582, opened 2022-07-06
Add additional fields for csv converter

### LINKED ISSUE #5456: Support additional fields for importing end devices from CSV file format.
#### Summary
The Things Stack currently supports importing end devices from CSV (comma-separated values) files with limited device fields. Users are seeking support for additional columns (e.g., device class and Rx2 Data Rate, etc).
#### Why do we need this?
Adding the ability to import devices via CSV with additional fields, similarly to the JSON format for end devices, will eliminate the need for separate configuration of other fields via CLI/API, etc.
We can currently import an end device with the few fields specified in the [CSV File Reference](https://www.thethingsindustries.com/docs/getting-started/migrating/device-csv/).
Importing devices with more specific fields similar to the [JSON End Device Format](https://www.thethingsindustries.com/docs/getting-started/migrating/device-json/#json-end-device-format).
#### Environment
TTS v3.19.1
Add support for additional fields in the CSV import feature.
...
No

nicholaspcr:
#### Summary

Closes #5456.

While loading the device profile from the DR solves a lot of the problems in devices migration, there is interest in having at least some of the most common [JSON](https://www.thethingsindustries.com/docs/getting-started/migrating/device-json/#json-end-device-format) fields settable via CSV.

#### Changes

- Added more csv field setters.
- Updated unit tests in `devicetemplates` pkg.

#### Testing

Unit testing and a bit of 🐒 

##### Regressions

There shouldn't be one as this is just adding more fields in which are accepted on the csv import.

#### Notes for Reviewers

Firstly I'm not so sure if my approach to the csv field names is the correct one. Taking the existent field names as example the `ids.dev_eui` is listed as only `dev_eui`, I followed the same logic for the fields added, for example the `session.dev_addr` was listed as `dev_ddr`. This might be confusing as it doesn't follow the standard defined in the json naming but it might be too verbose if we were to apply the same standard to the csv naming.

Secondly, the fields added were listed in the documentation as the most commonly used. With my current knowledge I'm unaware if there are other fields which should be included here, if you think there are any feel free to reference them and I'll add it.

Thirdly, I'll update the documentation after the two previous topics are discussed and before merging the PR.

#### Checklist


pkg/devicetemplates/ttscsv.go
```diff
@@ -177,6 +182,116 @@ var csvFieldSetters = map[string]csvFieldSetterFunc{
...
+		}
+		if dst.Session == nil {
+			dst.Session = &ttnpb.Session{}
+		}
+		if dst.Session.Keys == nil {
+			dst.Session.Keys = &ttnpb.SessionKeys{}
+		}
+		if dst.Session.Keys.AppSKey == nil {
+			dst.Session.Keys.AppSKey = &ttnpb.KeyEnvelope{}
+		}
```
KrishnaIyer: I don't think this field gets set outside this function. You can skip this and directly assume that `AppSKey` is nil

pkg/devicetemplates/ttscsv.go
```diff
@@ -177,6 +182,116 @@ var csvFieldSetters = map[string]csvFieldSetterFunc{
...
+		if dst.Session == nil {
+			dst.Session = &ttnpb.Session{}
+		}
+		if dst.Session.Keys == nil {
+			dst.Session.Keys = &ttnpb.SessionKeys{}
+		}
+		if dst.Session.Keys.FNwkSIntKey == nil {
+			dst.Session.Keys.FNwkSIntKey = &ttnpb.KeyEnvelope{}
+		}
```
KrishnaIyer: Is the `FNwkSIntKey` set somewhere else? If not you can skip this and assume it's nil.

pkg/devicetemplates/ttscsv_test.go
```diff
@@ -77,13 +77,23 @@ func TestTTSCSVConverter(t *testing.T) {
...
 					"version_ids.hardware_version",
 					"version_ids.model_id",
+					"root_keys.app_key.key",
+					"root_keys.nwk_key.key",
+					"mac_settings.rx1_delay",
+					"mac_settings.supports_32_bit_f_cnt",
+					"session.dev_addr",
+					"session.last_f_cnt_up",
+					"session.last_n_f_cnt_down",
+					"session.last_a_f_cnt_down",
```
KrishnaIyer: Nit; Please keep this alphabetically sorted.

pkg/devicetemplates/ttscsv.go
nicholaspcr: you are right, doesn't make much sense in testing if `AppSKey` is `nil` here.

pkg/devicetemplates/ttscsv.go
nicholaspcr: you are right, doesn't make much sense in testing if `FNwkSIntKey ` is `nil` here.

KrishnaIyer (approved):

johanstokking (commented):
Already merged, but approved in retrospect

PR #5594, opened 2022-07-13
Device provisioning section of the New Device Onboarding

ryaplots:
#### Summary

Refers to #4847 
The provisioning part of the device onboarding

#### Changes

- Add joinEUI logic
Screenshot 2022-07-13 at 15 19 16

- Add claiming
Screenshot 2022-07-13 at 15 19 03

- Add regular registration
Screenshot 2022-07-13 at 15 18 24

- Refactor validation schemas

#### Testing

Manually for now

##### Regressions

--

#### Checklist


pkg/webui/console/components/dev-eui-component/dev-eui.styl
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+.indicator
+  margin-left: $cs.s
+  align-self: center
+  color: $tc-subtle-gray
+
+.error
+  color: $tc-error
```
kschiffer:
```suggestion
```
You can use the global `tc-error` class.

pkg/webui/console/components/dev-eui-component/index.js
```diff
@@ -0,0 +1,183 @@
...
+
+  const indicatorContent = Boolean(errorMessage)
+    ? errorMessage
+    : {
+        ...sharedMessages.used,
+        values: {
+          currentValue: applicationDevEUICounter,
+          maxValue: env.devEUIConfig.applicationLimit,
+        },
+      }
```
kschiffer:
```suggestion
  const indicatorContent = errorMessage ||
  {
     ...sharedMessages.used,
     values: {
        currentValue: applicationDevEUICounter,
        maxValue: env.devEUIConfig.applicationLimit,
      },
    }
```

pkg/webui/console/components/dev-eui-component/index.js
```diff
@@ -0,0 +1,183 @@
+// Copyright © 2022 The Things Network Foundation, The Things Industries B.V.
```
kschiffer: This should be a container rather, since it connects to the store and makes requests.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/device-claiming-form-section/validation-schema.js
```diff
@@ -12,14 +12,30 @@
...
+    dev_eui: Yup.string()
+      .length(8 * 2, Yup.passValues(sharedMessages.validateLength))
+      .required(sharedMessages.validateRequired),
+  }),
+  authentication_code: Yup.string().when('_claim', claim => {
+    if (claim) {
+      return Yup.string()
+        .matches(/^[A-Z0-9]{1,32}$/, Yup.passValues(m.validateCode))
+        .required(sharedMessages.validateRequired)
+    }
```
kschiffer:
```suggestion
  authentication_code: Yup.string().when('_claim', { 
  is: true, 
  then: schema => schema
    .matches(/^[A-Z0-9]{1,32}$/, Yup.passValues(m.validateCode))
    .required(sharedMessages.validateRequired),
  }
```

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/index.js
```diff
@@ -80,8 +116,16 @@ const DeviceTypeFormSection = () => {
...
           component={Input}
+          actionDisable={!Boolean(ids?.join_eui)}
+          action={{
+            type: 'button',
+            disable: !Boolean(ids?.join_eui),
+            title: mayConfirm ? msg.confirm : msg.reset,
+            message: mayConfirm ? msg.confirm : msg.reset,
+            onClick: isClaimable,
+            raw: true,
+          }}
```
kschiffer: Let's not use this prop anymore. It is sort of a relict and makes the prop structure really messy.
Instead you can do
```js
<Form.Field
  …
>
  { _claim === undefined ? <Button disabled={mayConfirm} … /> : <Button … />
</Form.Field>
```

I'll take note to remove the `action` prop from `<Input />`.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/index.js
```diff
@@ -38,36 +40,70 @@ import DeviceRegistrationFormSection, {
...
+    }
+  }, [ids, setFieldValue, _claim, dispatch, setValues])
+
+  React.useEffect(() => {
+    // Merge the new device template with other form values.
+    if (template && hasValidDeviceRepositoryType && _inputMethod === 'device-repository') {
+      const { end_device } = template
+      setValues(merge(values, end_device), false)
+    }
+  }, [template, _inputMethod, setValues, values])
```
kschiffer: This is already done in https://github.com/TheThingsNetwork/lorawan-stack/blob/feature/4847-device-onboarding/pkg/webui/console/containers/device-onboarding-form/device-type-form-section/device-type-repository-form-section/index.js#L112-L121

It shouldn't be required to merge again, or am I missing something here?

pkg/webui/console/containers/device-onboarding-form/index.js
```diff
@@ -33,50 +44,159 @@ import DeviceTypeFormSection, {
...
-    },
+    values: { frequency_plan_id, lorawan_version, lorawan_phy_version, ids },
   } = useFormContext()
+
   const maySubmit =
     Boolean(frequency_plan_id) &&
     Boolean(lorawan_version) &&
     Boolean(lorawan_phy_version) &&
-    join_eui.length === 16
+    ids?.join_eui?.length === 16
```
kschiffer:
```suggestion
    typeof _claim === 'boolean'
```
Otherwise the submit button appears already before confirming the EUI.

pkg/webui/console/containers/device-onboarding-form/index.js
```diff
@@ -33,50 +44,159 @@ import DeviceTypeFormSection, {
...
+      }
+
+      try {
+        const device = await dispatch(
+          attachPromise(claimDevice(appId, qr_code, authenticatedIdentifiers)),
+        )
+        const { ids } = device
+        navigateToDevice(appId, ids.device_id)
+      } catch (error) {
+        setSubmitting(false)
```
kschiffer:
```suggestion
        setError(error)
```
The error needs to be stored and passed to the form. `setSubmitting()` is not required when the submit handler is asynchronous. We still unnecessarily use it in many submit handlers though.

pkg/webui/console/containers/device-onboarding-form/index.js
```diff
@@ -33,50 +44,159 @@ import DeviceTypeFormSection, {
...
+            toast({
+              type: toast.types.SUCCESS,
+              message: m.createSuccess,
+            })
+          default:
+            navigateToDevice(appId, ids.device_id)
+        }
+      } catch (error) {
+        setErrors(error)
+        setSubmitting(false)
```
kschiffer:
```suggestion
```

pkg/webui/console/containers/device-onboarding-form/index.js
```diff
@@ -33,50 +44,159 @@ import DeviceTypeFormSection, {
...
+            resetForm({ values: initialValues })
+            toast({
+              type: toast.types.SUCCESS,
+              message: m.createSuccess,
+            })
+          default:
+            navigateToDevice(appId, ids.device_id)
+        }
+      } catch (error) {
+        setErrors(error)
```
kschiffer:
```suggestion
        setError(error)
```

`setErrors()` from the form context is used to set errors for individual fields and is not what you want here. Instead, we need a separate state `error` and `setError()` which is then passed to the form as `error` prop.

pkg/webui/console/containers/device-onboarding-form/index.js
```diff
@@ -33,50 +44,159 @@ import DeviceTypeFormSection, {
...
+  const handleSubmit = useCallback(
+    async (values, { setSubmitting, resetForm, setErrors }, cleanedValues) => {
+      if (values._claim) {
+        return handleClaim(cleanedValues, setSubmitting)
+      }
+
+      return handleRegister(values, setSubmitting, resetForm, setErrors, cleanedValues)
+    },
+    [handleClaim, handleRegister],
+  )
```
kschiffer: Consider a layout like this:
```suggestion
  const handleSubmit = useCallback(
    async (values, _, cleanedValues) => {
      try {
        let result
        if (values._claim) {
          result = await handleClaim(cleanedValues)
        } else {
          result = await handleRegister(values, setSubmitting)
        }
        return result
      } catch (error) {
        setError(error)
      }
    },
    [handleClaim, handleRegister],
  )
```
You can then remove the try/catch from the other handlers.

pkg/webui/console/components/dev-eui-component/dev-eui.styl
```diff
@@ -0,0 +1,21 @@
...
```
kschiffer: Use `tc-error` global class instead.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/device-claiming-form-section/index.js
```diff
@@ -14,8 +14,30 @@
...
+
+const DeviceClaimingFormSection = () => (
+  <>
+    <DevEUIComponent />
+    <Form.Field
+      title={sharedMessages.claimAuthCode}
+      name="authentication_code"
+      component={Input}
+      sensitive
+    />
```
kschiffer: The field for the end device ID is missing here.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/index.js
```diff
@@ -38,36 +40,70 @@ import DeviceRegistrationFormSection, {
   initialValues as registrationInitialValues,
 } from './device-registration-form-section'
 
+const msg = defineMessages({
```
kschiffer, line 44:
```suggestion
const m = defineMessages({
```

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/index.js
```diff
@@ -38,36 +40,70 @@ import DeviceRegistrationFormSection, {
...
+  registration: '✔ This end device can be registered on the network',
+  claiming: '✔ This end device can be claimed from its current owner',
+})
+
 const initialValues = merge(
   {
     ids: {
       join_eui: '',
     },
+    _claim: '',
```
kschiffer:
```suggestion
    _claim: undefined,
```
An empty string is not a good default value for this. You can use `undefined` here since the field does not map to any actual field and it is more fitting for this scenario.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/index.js
```diff
@@ -38,36 +40,70 @@ import DeviceRegistrationFormSection, {
...
+    lorawan_version,
+    lorawan_phy_version,
+    ids,
+  } = values
+
   const template = useSelector(selectDeviceTemplate)
   const mayProvisionDevice =
     (Boolean(frequency_plan_id) && Boolean(lorawan_version) && Boolean(lorawan_phy_version)) ||
     (_inputMethod === 'device-repository' && hasValidDeviceRepositoryType(version, template))
+  const mayConfirm = _claim === '' || _claim === undefined
```
kschiffer:
```suggestion
  const mayConfirm = ids.join_eui.length === 16
```
The button should be disabled when the Join EUI hasn't been fully entered. This was not precise in the wireframes. Sorry about that.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/index.js
```diff
@@ -38,36 +40,70 @@ import DeviceRegistrationFormSection, {
...
+        },
+        _claim: '',
+      }))
+    } else {
+      const claim = await dispatch(attachPromise(getInfoByJoinEUI({ join_eui: ids?.join_eui })))
+      const supportsClaiming = claim.supports_claiming ?? false
+
+      setFieldValue('_claim', supportsClaiming)
+    }
+  }, [ids, setFieldValue, _claim, dispatch, setValues])
```
kschiffer: This took me quite a while to understand.

1. `isClaimable` is not a good name for an event handler, since it suggests that this is a function that returns a boolean whether the device is claimable or not
2. Combining both fetching and resetting into one handler makes it hard to maintain. I'd suggest to render two distinct buttons with two distinct handlers.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/index.js
```diff
@@ -38,36 +40,70 @@ import DeviceRegistrationFormSection, {
...
+
+  React.useEffect(() => {
+    // Merge the new device template with other form values.
+    if (template && hasValidDeviceRepositoryType && _inputMethod === 'device-repository') {
+      const { end_device } = template
+      setValues(merge(values, end_device), false)
+    }
+  }, [template, _inputMethod, setValues, values])
 
   return (
```
kschiffer, line 123: Consider a structure like this:
```js
const isClaiming = _claim === true
const isRegistration = _claim === false
const isDetermined = _claim !== undefined
const isRepository = _inputMethod === 'device-repository'
let joinEuiConfirmationMessage

if (isDetermined) {
  joinEuiConfirmationMessage = isClaiming ? m.claiming : m.registration
}

if (!mayProvisionDevice) {
  return <Message content={isDeviceRepository ? m.continueDeviceRepo : m.continueManual} … />
}

return (
  <>
    <Form.Field name="ids.join_eui" description={joinEuiConfirmationMessage} … />
    { !isDetermined && <Message content={m.continue} … /> }
    { isClaming && < DeviceClaimingFormSection />}
    { isRegistration && < DeviceRegistrationFormSection />}
  </>
)
```
I'm suggesting this because I believe that keeping conditional rendering to a minimum helps with maintainability. You can achieve that by assigning helper variables beforehand and by merging shared conditions as much as possible.

pkg/webui/console/components/dev-eui-component/index.js
```diff
@@ -0,0 +1,183 @@
...
+
+import {
+  selectApplicationDevEUICount,
+  selectSelectedApplicationId,
+} from '@console/store/selectors/applications'
+
+import style from './dev-eui.styl'
+
+const DevEUIComponent = props => {
+  const { values, setFieldValue, initialValues, devEUISchema } = props
```
kschiffer:
```suggestion
  const { devEUISchema } = props
  const { values, setFieldValue, initialValues } = useFormContext()
```
If you are setting this up as a `<Form.Field />` then you can assume that this component will be rendered within a `<Form />` in which you will have access to the form context.
Also, [you were not passing any of these props anyway](https://github.com/TheThingsNetwork/lorawan-stack/pull/5594/files#diff-b5fea30b9eed7fac720e3f08220be51470dc6a0cc19546b9699db48f1bbdf1a7R33).

pkg/webui/console/components/dev-eui-component/index.js
```diff
@@ -0,0 +1,183 @@
...
+        ...sharedMessages.used,
+        values: {
+          currentValue: applicationDevEUICounter,
+          maxValue: env.devEUIConfig.applicationLimit,
+        },
+      }
+
+  const indicatorCls = classnames(style.indicator, {
+    [style.error]:
+      applicationDevEUICounter === env.devEUIConfig.applicationLimit || Boolean(errorMessage),
```
kschiffer:
```suggestion
    'tc-error':
      applicationDevEUICounter === env.devEUIConfig.applicationLimit || Boolean(errorMessage),
```

pkg/webui/console/components/dev-eui-component/index.js
```diff
@@ -0,0 +1,183 @@
...
+
+  const handleGenerate = React.useCallback(async () => {
+    try {
+      const result = await handleDevEUIRequest()
+      setDevEUIGenerated(true)
+      euiInputRef.current.focus()
+      setErrorMessage(undefined)
+      return result
+    } catch (error) {
+      if (error.details[0].name === 'global_eui_limit_reached') {
```
kschiffer:
```suggestion
      if (getBackendErrorName(error) === 'global_eui_limit_reached') {
```
`import { getBackendErrorName } from '@ttn-lw/lib/errors/utils'`

pkg/webui/console/components/dev-eui-component/index.js
```diff
@@ -0,0 +1,183 @@
...
+  )
+
+  const handleIdPrefill = React.useCallback(() => {
+    if (values) {
+      // Do not overwrite a value that the user has already set.
+      if (values.ids.device_id === initialValues.ids.device_id) {
+        const generatedId = generateDeviceId(values)
+        setFieldValue('ids.device_id', generatedId)
+      }
+    }
```
kschiffer:
```suggestion
    // Do not overwrite a value that the user has already set unless it is already in the EUI format.
    if (values?.ids?.eui.length === 16 && (values?.ids?.device_id === '' || /eui-\d{16}/.test(values?.ids?.device_id)) {
      const generatedId = `eui-${values.ids.eui.toLowerCase()}`
      setFieldValue('ids.device_id', generatedId)
    }
```

pkg/webui/console/components/dev-eui-component/index.js
```diff
@@ -0,0 +1,183 @@
...
+        devEUISchema.validateSync(idsValues.dev_eui)
+        return `eui-${idsValues.dev_eui.toLowerCase()}`
+      } catch (e) {
+        // We dont want to use invalid `dev_eui` as `device_id`.
+      }
+
+      return initialValues.ids.device_id || undefined
+    },
+    [initialValues.ids.device_id, devEUISchema],
+  )
```
kschiffer: This feels overly complicated to me. After all, we only care about whether the EUI has been entered fully. With the way the Byte Input works, there is virtually no way to otherwise enter an invalid value.
```suggestion
```

pkg/webui/console/components/dev-eui-component/index.js
```diff
@@ -0,0 +1,183 @@
...
+
+  const devEUIGenerateDisabled =
+    applicationDevEUICounter === env.devEUIConfig.applicationLimit ||
+    !env.devEUIConfig.devEUIIssuingEnabled ||
+    devEUIGenerated
+
+  return env.devEUIConfig.devEUIIssuingEnabled ? (
+    <Form.Field
+      title={sharedMessages.devEUI}
+      name="ids.dev_eui"
```
kschiffer:
```suggestion
      name={props.name}
```
This way, you can pass the relevant field name `authenticated_identifiers.dev_eui` / `ids.dev_eui` based on whether you are claiming or not.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/index.js
```diff
@@ -80,8 +116,16 @@ const DeviceTypeFormSection = () => {
           max={8}
```
kschiffer, line 132: You can use
```js
name="ids.join_eui,authenticated_identifiers.join_eui"
```
here to save the value in both places so that you don't have to assemble the proper values structure when submitting.

I'm doing a [similar thing](https://github.com/TheThingsIndustries/lorawan-stack/pull/3304/files#diff-932a550e758bd839831097546834b5cdf7e7726810861ed6625d05a89515a626R114) on gateway onboarding.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/device-registration-form-section/validation-schema.js
```diff
@@ -13,13 +13,168 @@
...
+      if (supports_joins) {
+        return devEUISchema.required(sharedMessages.validateRequired)
+      }
+
+      return devEUISchema.default(null).nullable()
+    }),
+    device_id: Yup.string().when('_claim', claim => {
+      if (!claim) {
+        return Yup.string().required(sharedMessages.validateRequired)
+      }
```
kschiffer: A check like this shouldn't be necessary since the whole schema is only applied when `!_claim` anyway.

kschiffer (changes requested):
Good stuff so far. The devil is always in the details so this is a first round of review. I'll review the validation schemas a bit more in-depth but you can already start with addressing these comments.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/index.js
ryaplots: On L35 I am already importing the messages from `../messages` and naming it `m`, that's why here I used `msg`

pkg/webui/console/containers/dev-eui-component/index.js
```diff
@@ -0,0 +1,164 @@
...
+  }),
+}
+
+DevEUIComponent.defaultProps = {
+  values: {
+    ids: {
+      device_id: undefined,
+    },
+  },
+}
```
kschiffer:
```suggestion
}
```

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/device-claiming-form-section/index.js
```diff
@@ -14,8 +14,78 @@
...
+
+    try {
+      devEUISchema.validateSync(authenticated_identifiers.dev_eui)
+      return `eui-${authenticated_identifiers.dev_eui.toLowerCase()}`
+    } catch (e) {
+      // We dont want to use invalid `dev_eui` as `device_id`.
+    }
+
+    return initialValues.target_device_id || ''
+  }
```
kschiffer:
```suggestion
```
Just get rid of this whole bit and generate the Dev EUI on the fly. See below.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/device-claiming-form-section/index.js
```diff
@@ -14,8 +14,78 @@
...
+        const { setSelectionRange } = idInputRef.current
+
+        const generatedId = generateDeviceId(values)
+        if (generatedId.length > 0 && generatedId === values.target_device_id) {
+          setSelectionRange.call(idInputRef.current, 0, generatedId.length)
+        }
+      }
+    },
+    [values],
+  )
```
kschiffer:
```suggestion
```

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/device-registration-form-section/index.js
```diff
@@ -13,9 +13,199 @@
...
+  )
+
+  let appKeyPlaceholder = undefined
+  let nwkKeyPlaceholder = undefined
+  if (!mayEditKeys) {
+    appKeyPlaceholder = sharedMessages.insufficientAppKeyRights
+    nwkKeyPlaceholder = sharedMessages.insufficientNwkKeyRights
+  }
+
+  const isOTAA = template?.end_device?.supports_join
```
kschiffer: This shouldn't be necessary

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/index.js
```diff
@@ -38,62 +41,121 @@ import DeviceRegistrationFormSection, {
...
           required
           component={Input}
           tooltipId={tooltipIds.JOIN_EUI}
-          onBlur={onIdPrefill}
-        />
+          encode={joinEuiEncoder}
+          decode={joinEuiDecoder}
+        >
+          {_claim === undefined ? (
+            <Button disabled={!mayConfirm} onClick={getClaiming} message={msg.confirm} raw />
```
kschiffer:
```suggestion
            <Button disabled={!mayConfirm} onClick={getClaiming} message={msg.confirm} />
```

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/index.js
```diff
@@ -38,62 +41,121 @@ import DeviceRegistrationFormSection, {
...
+          ) : (
+            <Button onClick={resetJoinEui} message={msg.reset} raw />
```
kschiffer:
```suggestion
            <Button onClick={resetJoinEui} message={msg.reset} />
```

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/validation-schema.js
```diff
@@ -13,18 +13,32 @@
...
-  .concat(claimValidationSchema)
-  .concat(registrationValidationSchema)
+const validationSchema = Yup.object({
+  ids: Yup.object().when(['supports_join'], (isOTAA, schema) => {
+    if (isOTAA) {
+      return schema.shape({
+        join_eui: joinEUISchema.required(sharedMessages.validateRequired),
+      })
+    }
+  }),
```
kschiffer: Please move this to the manual registration validation schema.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/validation-schema.js
```diff
@@ -13,18 +13,32 @@
...
+  supports_join: Yup.bool().default(false),
```
kschiffer:
```suggestion
```

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/validation-schema.js
```diff
@@ -13,18 +13,32 @@
...
+  _claim: Yup.mixed().oneOf(['', true, false]),
```
kschiffer:
```suggestion
```

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/device-type-manual-form-section/advanced-settings-section.js
```diff
@@ -328,7 +328,7 @@ const AdvancedSettingsSection = () => {
           <>
             <Form.FieldContainer horizontal>
               <Form.Field
-                required={!isUndefined(mac_settings.rx1_data_rate_offset)}
+                required={!isUndefined(mac_settings?.rx1_data_rate_offset)}
```
kschiffer:
```suggestion
                required={!isUndefined(mac_settings.rx1_data_rate_offset)}
```

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/index.js
```diff
@@ -63,8 +62,12 @@ const DeviceTypeFormSection = () => {
...
         setShowModal(false)
         if (approved) {
-          resetForm()
-          return setFieldValue(name, value)
+          return resetForm({
+            errors: {},
+            values: {
+              [name]: value,
+            },
+          })
```
kschiffer: Should be good as it was.

sdk/js/src/service/claim.js
```diff
@@ -0,0 +1,56 @@
...
+
+  // Claim
+  async claim(
+    applicationId,
+    qrCode,
+    values,
+    targetNetworkServerAddress = this._stackConfig.nsHost,
+    targetApplicationServerAddress = this._stackConfig.asHost,
+  ) {
+    const deviceToClaim = qrCode ? { qr_code: btoa(qrCode) } : { authenticated_identifiers: values }
```
kschiffer: You don't need to decode this.

pkg/webui/console/containers/device-onboarding-form/index.js
```diff
@@ -33,50 +44,154 @@ import DeviceTypeFormSection, {
...
+    async values => {
+      const { ids, authentication_code, qr_code } = values
+      let authenticatedIdentifiers
+      if (!qr_code) {
+        authenticatedIdentifiers = {
+          join_eui: ids.join_eui,
+          dev_eui: ids.dev_eui,
+          authentication_code,
+        }
+      }
```
kschiffer: No need to compose the object again.

pkg/webui/console/containers/device-onboarding-form/index.js
```diff
@@ -33,50 +44,154 @@ import DeviceTypeFormSection, {
...
+  )
+
+  const handleSubmit = useCallback(
+    async (values, { resetForm }, cleanedValues) => {
+      try {
+        let result
+        if (values._claim) {
+          result = handleClaim(cleanedValues)
+        } else {
+          result = handleRegister(values, resetForm, cleanedValues)
```
kschiffer:
```suggestion
          result = await handleClaim(cleanedValues)
        } else {
          result = await handleRegister(values, resetForm, cleanedValues)
```

pkg/webui/console/containers/device-onboarding-form/index.js
```diff
@@ -33,50 +44,154 @@ import DeviceTypeFormSection, {
...
       initialValues={initialValues}
-      hiddenFields={['network_server_address', 'application_server_address', 'join_server_address']}
+      hiddenFields={[
+        'network_server_address',
+        'application_server_address',
+        'join_server_address',
+        'frequency_plan_id',
+        'lorawan_phy_version',
+        'lorawan_version',
+        'supports_join',
```
kschiffer: Please try to register the fields of the device template manually.

kschiffer (changes requested):

pkg/webui/console/containers/dev-eui-component/index.js
ryaplots: Here I still need `name`, to pass the correct field name, depending on where the dev_eui component is used, after all.

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/device-type-manual-form-section/advanced-settings-section.js
```diff
@@ -284,6 +286,17 @@ const AdvancedSettingsSection = () => {
...
   }, [_default_ns_settings, defaultMacSettings, mayChangeToDefaultSettings, setValues])
 
+  // Register hidden fields so they don't get cleaned.
+  useEffect(() => {
+    const hiddenFields = [
+      'network_server_address',
+      'application_server_address',
+      'join_server_address',
+    ]
+    addToFieldRegistry(hiddenFields)
```
kschiffer:
```suggestion
    addToFieldRegistry(...hiddenFields)
```
You can spread the array. This way you don't have to flatten the field registry.

pkg/webui/components/form/index.js
```diff
@@ -129,7 +129,9 @@ const Form = props => {
       try {
         // Compose clean values as well, which do not contain values of unmounted
         // fields, as well as pseudo values (starting with `_`).
-        const cleanedValues = omitBy(pick(values, fieldRegistry), (_, key) => key.startsWith('_'))
+        const cleanedValues = omitBy(pick(values, fieldRegistry.flat(Infinity)), (_, key) =>
+          key.startsWith('_'),
+        )
```
kschiffer: Please remove this diff. See my comment on spreading entries below.

pkg/webui/console/containers/dev-eui-component/index.js
```diff
@@ -94,7 +94,11 @@ const DevEUIComponent = props => {
         ) {
           const generatedId = `eui-${values.authenticated_identifiers.dev_eui.toLowerCase()}`
           setFieldValue('target_device_id', generatedId)
+
+          return generatedId
```
kschiffer: Why does it have to return anything?

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/device-type-manual-form-section/advanced-settings-section.js
```diff
@@ -284,6 +286,17 @@ const AdvancedSettingsSection = () => {
...
+    return () => removeFromFieldRegistry(hiddenFields)
```
kschiffer:
```suggestion
    return () => removeFromFieldRegistry(...hiddenFields)
```

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/index.js
```diff
@@ -38,62 +41,127 @@ import DeviceRegistrationFormSection, {
...
 
-const DeviceTypeFormSection = () => {
-  const [isClaiming, setIsClaiming] = useState(undefined)
+// Save EUI in both fields.
+const joinEuiEncoder = value => ({
+  ids: { join_eui: value },
+  authenticated_identifiers: {
+    join_eui: value,
+  },
+})
```
kschiffer, line 73:
```
const joinEuiEncoder = value => ({
  ids: { ...value.ids, join_eui: value },
  authenticated_identifiers: {
    ...value.authenticated_identifiers,
    join_eui: value,
  },
})
```
Otherwise, you lose the existing values there which will cause issues when submitting.

kschiffer: Please disregard this change actually. It should work as you put it. The issue is that the base branched is not rebased on latest master.

pkg/webui/console/containers/dev-eui-component/index.js
ryaplots: It was a way of preventing it from getting into both ifs, but I think I found a better way

kschiffer: You don't have to even make the condition. You can just set the field in both paths:
```js
setFieldValue('target_device_id', generatedId)
setFieldValue('ids.device_id', generatedId)
```
The irrelevant one will be stripped when submitting anyway.

ryaplots: Actually I just realized why I put that if there. If we are setting the dev eui in claiming, it will be stored in `authenticated_identifiers`, while if it's regular registration it will be stored in `ids`. It's two separate fields. In this if, I have to `setFieldValue('ids.device_id', generatedId)`, if it's stored in `ids` and `setFieldValue('target_device_id', generatedId)` if it's stored in `authenticated_identifiers`.

kschiffer (approved):

Issue #5586, opened 2022-07-08
DeviceTimeAns timestamp is not corresponding to the specification


adriansmares:
### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/2471
References https://github.com/TheThingsNetwork/lorawan-stack/pull/2478
References https://github.com/TheThingsNetwork/lorawan-stack/issues/5585

The Network Server should not use the Network Server receive time as a response to `DeviceTimeReq`, but instead the timestamp of the end of the transmission.

### Current Situation

The Network Server received timestamp is used:

https://github.com/TheThingsNetwork/lorawan-stack/blob/70d88dd86a9b17bf8faf1bbea6961d6c5d3f2ae7/pkg/networkserver/mac/device_time.go#L34-L43

The standard is both misleading and clear regarding this in both the 1.0.3 and 1.0.4 specification:
<img width="937" alt="image" src="https://user-images.githubusercontent.com/36161392/177996171-fbe9ab25-8ae3-40a1-b7d4-a455c947ed2f.png">

I think that https://github.com/TheThingsNetwork/lorawan-stack/issues/2471 is caused by the phrasing of 'current Network time', but the below sentence (GPS time at the end of the transmission) should be hopefully clearer.

### Why do we need this? Who uses it, and when?

We need this for proper class B support.

### Proposed Implementation

This requires https://github.com/TheThingsNetwork/lorawan-stack/issues/5585 first.

I think we need the following fallback system:
1. In the presence of a `settings.gps_time` field, use that.
2. If not found, walk through the `rx_metadata` slice and use the first `gps_time` found.
3. If not found, use the Network Server receive time.

I think that option 2 can become more sophisticated later on, but for now this should suffice.

cc @johanstokking 

Edit:
Also https://github.com/TheThingsNetwork/lorawan-stack/pull/2478/files#diff-9ba913e113eb0e0a51b8c6b091c237d7af5d797e29720b3fea84410ece5a99c5L666-L671 should be reverted, as the handler should run after deduplication.

### Contributing

- [X] I can help by doing more research.
- [X] I can help by implementing the feature after the proposal above is approved.
- [X] I can help by testing the feature before it's released.

### Code of Conduct

- [X] I agree to follow TTN's [Community Code of Conduct](https://github.com/TheThingsNetwork/Manifesto/blob/master/CodeOfConduct.md).

johanstokking:
[...]

PR #5598, opened 2022-07-14
Clarify the semantics of the uplink message timestamps

### LINKED ISSUE #5585: Clarify the semantics of the uplink message timestamps
### Summary

The timestamps found in the `UplinkMessage` message should have their semantics clarified.

### Current Situation

We currently have multiple timestamps available:

1. The timestamp at which the Gateway Server / Network Server received the frame:
https://github.com/TheThingsNetwork/lorawan-stack/blob/70d88dd86a9b17bf8faf1bbea6961d6c5d3f2ae7/api/messages.proto#L83-L85

2. The timestamp at which a singular transmission has been received, from the perspective of the gateway:
https://github.com/TheThingsNetwork/lorawan-stack/blob/70d88dd86a9b17bf8faf1bbea6961d6c5d3f2ae7/api/lorawan.proto#L326-L328

3. A set of timestamps of each individual transmission metadata:
https://github.com/TheThingsNetwork/lorawan-stack/blob/70d88dd86a9b17bf8faf1bbea6961d6c5d3f2ae7/api/metadata.proto#L41
https://github.com/TheThingsNetwork/lorawan-stack/blob/70d88dd86a9b17bf8faf1bbea6961d6c5d3f2ae7/api/messages.proto#L82

The semantics of timestamps 2 and 3 are not clear unfortunately. Specifically it is unclear what is the clock source of this timestamp (is it a GPS timestamp that the gateway is _sure_ of, or did it do some NTP and has only a bit of accuracy ?).

This has the following effects:
- Some uplinks are marked as coming from the 80s, because the gateway did some kind of NTP failure, or lacks a GPS fix.
- Some uplinks are buffered and the timestamp is mis-attributed (time of buffer submission vs actual receive timestamp).
- Correctly supporting `DeviceTimeAns` is not really possible (if we have 3 timestamps, which one do we pick ? per spec, it should be a GPS one)

### Proposed Implementation

_The proposals below are targeting the `RxMetadata` message_

Option 1: Change the semantics of the `time`, for uplink messages, such that the presence of the value signals that it is provided by GPS, and leave it empty otherwise.

PROS:
- It was always nullable, so taking it out is by the book correct.
- We guarantee that applications that use it have a useful timestamp.

CONS:
- De facto, this is a breaking change. Gateways that had 0 issues will suddenly no longer have a timestamp.

I don't think that we should go this way, but I'm documenting this here.

Option 2: Add a new field, `gps_time`, that is filled when the timestamp is really sourced by GPS. If we have both an NTP sourced timestamp, and a GPS sourced timestamp, `time` contains the GPS one.

PROS:
- Not a breaking change.
- It will clarify the semantics.

CONS:
- We add a new field to store. Keep in mind that this is both something we will carry around (data transfer between components) and store (event storage, NS recent uplinks storage maybe, AS storage integration), so we should be mindful.

I'm more in favor of this variant, but it's not entirely 'free' (storage / transfer costs). 

cc @johanstokking

### Contributing

### Code of Conduct

happyRip:
#### Summary

Closes #5585

#### Changes

- Added `gps_time` and `received_at` fields to `RxMetadata`
- Improved uplink message's `GatewayReceiveTime` accuracy (when possible)

#### Testing

Adjust relevant test cases to feature expected `received_at` and `gps_time` values.

##### Regressions

Possible regressions listed in #5585 and #5586.

#### Notes for Reviewers

nil

#### Checklist


api/metadata.proto
```diff
@@ -71,12 +71,16 @@ message RxMetadata {
   uint32 hopping_width = 19;
   // Frequency drift in Hz between start and end of an LR-FHSS packet (signed).
   int32 frequency_drift = 20;
+  // GPS sourced timestamp at the end of the transmission.
+  google.protobuf.Timestamp gps_time = 21;
+  // Time when the service received the message.
+  google.protobuf.Timestamp received_at = 22;
```
adriansmares, line 80: I would add slightly more details to the comments here. We now are going to have 3 timestamps and that will be confusing in the absence of the explanation.

I would add a comment to `time` saying `Timestamp at the end of the transmission, provided by the gateway. The accuracy is undefined.`

For `gps_time` I would mention `Timestamp at the end of the transmission, provided by the gateway. Guaranteed to be based on a GPS PPS signal, with an accuracy of 1 millisecond.`.

For `received_at` I would mention `Timestamp at which the Gateway Server has received the message.`

pkg/gatewayserver/gatewayserver_test.go
```diff
@@ -1365,6 +1365,7 @@ func TestGatewayServer(t *testing.T) {
 										for i, md := range msg.RxMetadata {
 											a.So(md.UplinkToken, should.NotBeEmpty)
 											md.UplinkToken = nil
+											expected.RxMetadata[i].ReceivedAt = md.ReceivedAt
```
adriansmares: I would say that it is better to erase it from `md` (set to `nil`).

pkg/gatewayserver/io/io.go
```diff
@@ -297,6 +297,12 @@ func (c *Connection) HandleUp(up *ttnpb.UplinkMessage, frontendSync *FrontendClo
 		} else if !c.gateway.LocationPublic {
 			md.Location = nil
 		}
+
+		r := *ttnpb.StdTime(up.ReceivedAt)
+		if _, _, median, _, count := c.RTTStats(100, time.Now()); count > 0 {
+			r.Add(-median / 2)
```
adriansmares: `time.Time.Add` does not modify `r` - you need to reassign`r`.

pkg/gatewayserver/io/io.go
```diff
@@ -297,6 +297,12 @@ func (c *Connection) HandleUp(up *ttnpb.UplinkMessage, frontendSync *FrontendClo
 		} else if !c.gateway.LocationPublic {
 			md.Location = nil
 		}
+
+		r := *ttnpb.StdTime(up.ReceivedAt)
```
adriansmares: It would be better to compute this outside of the loop, since `up.ReceivedAt` or the return of `RTTStats` does not change between iterations (mostly).

Also be careful do this write at the start of the loop - you have some `continue` statements above that will make the metadata not contain the `received_at` when triggered.

pkg/packetbrokeragent/translation.go
```diff
@@ -80,15 +80,15 @@ func fromPBDataRate(dataRate *packetbroker.DataRate) (dr *ttnpb.DataRate, coding
...
-	// 	return &ttnpb.DataRate{
-	// 		Modulation: &ttnpb.DataRate_Lrfhss{
-	// 			Lrfhss: &ttnpb.LRFHSSDataRate{
-	// 				ModulationType:        mod.Lrfhss.ModulationType,
-	// 				OperatingChannelWidth: mod.Lrfhss.OperatingChannelWidth,
-	//  			CodingRate:            mod.Lrfhss.CodingRate,
-	// 			},
-	// 		},
-	// 	}, mod.Lrfhss.CodingRate, true
+	//	return &ttnpb.DataRate{
```
adriansmares: Revert the changes here.

pkg/packetbrokeragent/translation.go
```diff
@@ -121,15 +121,15 @@ func toPBDataRate(dataRate *ttnpb.DataRate, codingRate string) (*packetbroker.Da
...
-	// 	return &packetbroker.DataRate{
-	// 		Modulation: &packetbroker.DataRate_Lrfhss{
-	// 			Lrfhss: &packetbroker.LRFHSSDataRate{
-	// 				ModulationType:        mod.Lrfhss.ModulationType,
-	// 				OperatingChannelWidth: mod.Lrfhss.OperatingChannelWidth,
-	// 				CodingRate:            mod.Lrfhss.CodingRate,
-	// 			},
-	// 		},
-	// 	}, true
+	//	return &packetbroker.DataRate{
```
adriansmares: Revert the changes here.

pkg/packetbrokeragent/translation.go
```diff
@@ -373,11 +373,17 @@ func toPBUplink(ctx context.Context, msg *ttnpb.GatewayUplinkMessage, config For
...
 			}
 
-			if md.Time != nil {
-				t := ttnpb.StdTime(md.Time)
-				if gatewayReceiveTime == nil || t.Before(*gatewayReceiveTime) {
-					gatewayReceiveTime = t
-				}
+			var t *time.Time
+			switch {
+			case md.GpsTime != nil:
```
adriansmares, line 383: So this is what I requested in the issue, but maybe we need to do better - I think that if we have a GPS timestamp, we should override whatever we stored in `gatewayReceiveTime`, and no future updates to `gatewayReceiveTime` should be done.

Consider the following example:
2 metadata - one which contains a `gps_time` of `14:02:01.123123123` and one which contains a `time` of `14:02:01.0000000`. Since the secondary metadata is earlier we should use that, even though the GPS time is more precise and correct.

pkg/gatewayserver/io/io.go
happyRip: Turns out `receivedAt := *ttnpb.StdTime(up.ReceivedAt)` was declared outside of the loop, thus I [used it](https://github.com/TheThingsNetwork/lorawan-stack/pull/5598/commits/14ec453bfdafcd503df611ea71d9d9a2955aac89#diff-9c87734043add0ade22bca71f2a6641c1d37830e1aae533fbc6066899c479d1bR277).

pkg/packetbrokeragent/translation.go
happyRip: Had multiple ideas on how to handle this case, [finally went with this](https://github.com/TheThingsNetwork/lorawan-stack/pull/5598/commits/14ec453bfdafcd503df611ea71d9d9a2955aac89#diff-a8a8e5ff5879a789dca4cdcd50830e79b514bf8eb5bd0ad553450bf7dcbb2a82R384).

```go
var t *time.Time
switch {
case md.GpsTime != nil:
  t = ttnpb.StdTime(md.GpsTime)
case md.Time != nil:
  t = ttnpb.StdTime(md.Time)
default:
  t = ttnpb.StdTime(md.ReceivedAt)
}
gatewayReceiveTime = t
```
`gatewayReceiveTime` at the very end emphasises the fact that's the variable I'd like to set here and it's going to be reassigned under every condition.

Other possible solutions:
- ```go
  switch {
  case md.GpsTime != nil:
    gatewayReceiveTime = ttnpb.StdTime(md.GpsTime)
  case md.Time != nil:
    gatewayReceiveTime = ttnpb.StdTime(md.Time)
  default:
    gatewayReceiveTime = ttnpb.StdTime(md.ReceivedAt)
  }
  ```
  - seems less clean due to extensive usage of a long variable
- ```go
  gatewayReceiveTime = func() *time.Time {
    switch {
    case md.GpsTime != nil:
      return ttnpb.StdTime(md.GpsTime)
    case md.Time != nil:
      return ttnpb.StdTime(md.Time)
    default:
      return ttnpb.StdTime(md.ReceivedAt)
    }
  }()
  ```
  - non-standard

adriansmares: But this does not solve the issue that I've mentioned - if the order of the metadata elements is 'low precision', 'high precision', 'low precision' you will keep the last timestamp, which is the low precision one.

happyRip: I see, so first choose most precise timestamp(s) and then choose among those in `gps` → `time` → `received_at` order?
Or should the earliest be chosen each case?

adriansmares: The axiom (which should be true in practice, but you never know) is that all `gps_time` is equivalent modulo 1 millisecond. So if you find a GPS time, then you don't need anything else. But if all of the metadata is missing the GPS time, then indeed we should take the earliest `time` and fallback to `received_at` (in case `time` is `nil`, compare using `received_at`).

happyRip: Solution proposal:

```go
gatewayReceiveTime = func() *time.Time {
	if t := md.GpsTime; t != nil {
		return ttnpb.StdTime(t)
	}

	t, r := ttnpb.StdTime(md.Time), ttnpb.StdTime(md.ReceivedAt)
	if t.IsZero() || r.Before(*t) {
		return r
	}
	return t
}()
```

Always use `gps_time` if available
```go
if t := md.GpsTime; t != nil {
  return ttnpb.StdTime(t)
}
```

Otherwise choose earliest from `time` and `received_at`.
Fallback to `received_at` if `time` is not available.
```go
t, r := ttnpb.StdTime(md.Time), ttnpb.StdTime(md.ReceivedAt)
if t.IsZero() || r.Before(*t) {
  return r
}
return t
```

adriansmares: The problem is that when you do this, you do not take into consideration that `gatewayReceiveTime` could already be set to a `gps_time`.

Take the following:
Metadata 1 contains `gps_time`, metadata 2 contains `time`.
When you process metadata 1 you will just set `gatewayReceiveTime` to `gps_time`.
Then you process metadata 2, there is no `gps_time`, so you go into the secondary condition and overwrite the timestamp.

What I have in mind is:
```
switch {
case hasGPSTime:
case md.GpsTime != nil:
  gatewayReceiveTime = ttnpb.StdTime(md.GpsTime)
  hasGPSTime = true
case gatewayReceiveTime == nil || md.Time != nil && minimal condition between gatewayReceiveTime and md.Time:
  gatewayReceiveTime = ttnpb.StdTime(md.Time)
case gatewayReceiveTime == nill || minimal condition between gatewayReceiveTime and md.ReceivedAt:
  gatewayReceiveTime = ttnpb.StdTime(md.ReceivedAt)
}
```

happyRip: Pushed a new solution:

Add `var hasGpsTime bool` before the loop begins

Inside the loop:
```go
timestampCase := func(t *pbtypes.Timestamp) bool {
	g := gatewayReceiveTime
	return g == nil || t != nil && ttnpb.StdTime(t).Before(*g)
}
switch {
case hasGpsTime:
case md.GpsTime != nil:
	gatewayReceiveTime = ttnpb.StdTime(md.GpsTime)
	hasGpsTime = true
case timestampCase(md.Time):
	gatewayReceiveTime = ttnpb.StdTime(md.Time)
case timestampCase(md.ReceivedAt):
	gatewayReceiveTime = ttnpb.StdTime(md.ReceivedAt)
}
```

pkg/gatewayserver/io/io.go
```diff
@@ -274,6 +274,12 @@ func (c *Connection) HandleUp(up *ttnpb.UplinkMessage, frontendSync *FrontendClo
 	}
 
 	for _, md := range up.RxMetadata {
+		r := receivedAt
+		if _, _, median, _, count := c.RTTStats(100, time.Now()); count > 0 {
```
johanstokking: Two things:
1. We want the message's received time as reference time. The difference between `time.Now()` and `receivedAt` is probably negligible, but still, it's cleaner to use `receivedAt` here I think
2. The result of the `c.RTTStats()` does not depend on the loop variable `md`, so this calculation should be moved to before the loop, and the loop is only needed if there's a known RTT to apply

pkg/packetbrokeragent/translation.go
```diff
@@ -373,11 +374,19 @@ func toPBUplink(ctx context.Context, msg *ttnpb.GatewayUplinkMessage, config For
 				localization.Antennas = append(localization.Antennas, locAnt)
 			}
 
-			if md.Time != nil {
-				t := ttnpb.StdTime(md.Time)
-				if gatewayReceiveTime == nil || t.Before(*gatewayReceiveTime) {
-					gatewayReceiveTime = t
-				}
+			timestampCase := func(t *pbtypes.Timestamp) bool {
```
johanstokking:
```suggestion
			earlierGatewayReceiveTime := func(t *pbtypes.Timestamp) bool {
```

pkg/gatewayserver/io/io.go
```diff
@@ -273,12 +273,13 @@ func (c *Connection) HandleUp(up *ttnpb.UplinkMessage, frontendSync *FrontendClo
 		panic("unreachable")
 	}
 
+	r := receivedAt
+	if _, _, median, _, count := c.RTTStats(100, receivedAt); count > 0 {
+		r = receivedAt.Add(-median / 2)
+	}
+	mdReceivedAtProto := ttnpb.ProtoTimePtr(r)
```
johanstokking:
```suggestion
	receivedAtGateway := receivedAt
	if _, _, median, _, count := c.RTTStats(100, receivedAt); count > 0 {
		receivedAtGateway = receivedAt.Add(-median / 2)
	}
```

pkg/gatewayserver/io/io.go
```diff
@@ -273,12 +273,13 @@ func (c *Connection) HandleUp(up *ttnpb.UplinkMessage, frontendSync *FrontendClo
...
 	for _, md := range up.RxMetadata {
-		r := receivedAt
-		if _, _, median, _, count := c.RTTStats(100, time.Now()); count > 0 {
-			r = r.Add(-median / 2)
-		}
-		md.ReceivedAt = ttnpb.ProtoTimePtr(r)
+		md.ReceivedAt = mdReceivedAtProto
```
johanstokking:
```suggestion
		md.ReceivedAt = ttnpb.ProtoTimePtr(receivedAtGateway)
```

pkg/gatewayserver/io/io.go
johanstokking: Better not declare single character variables in a large scope to avoid confusion. In small scopes, like short loops etc, it's obviously permitted.

johanstokking (approved):

adriansmares (approved):
LGTM. Squash the last commit that was commited from the web UI.

PR #5599, opened 2022-07-14
Lower stored MAC state data

### LINKED ISSUE #5127: Lower data stored for recent traffic
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/issues/2256
References https://github.com/TheThingsNetwork/lorawan-stack/pull/4968

The Network Server should store less data in the MAC state `recent_uplinks`/`recent_downlinks` fields. Specifically, only fields which are actually used later should be stored, but nothing else, in a similar manner to how the Application Server achieves the same filtering:

https://github.com/TheThingsNetwork/lorawan-stack/blob/6e8763164f1edab0207afac50c791efcc5c4e3e8/pkg/applicationserver/applicationserver.go#L882-L908

#### Why do we need this?

In order to store less data that we don't use.

Right now we filter out only the correlation IDs.

Any unused field should be cleared out.

#### Environment

`v3.17.2`

Filter out the fields in the same manner as the correlation IDs are removed.

Existing unit tests can cover this.

Yes.

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/5127

#### Changes

- As part of the `MACState` message, add minimized versions of `UplinkMessage`, `DownlinkMessage`,  `Message`, `TxSettings` and `RxMetadata`
   - The scope here is to ensure that fields which are not going to be available from the 'real' `UplinkMessage`/`DownlinkMessage` are not accidentally used (i.e. using zero values as if they were intentional)
- Use the minimized versions as part of the Network Server MAC state handling
- Fix LoRaWAN 1.1 `AFCntDown` sanity check
    - The old check did not take into account the fact that the frame header frame counter is 16 bit, and instead the full (32bit) frame counter should be used for checks

#### Testing

Unit testing.

##### Regressions

The code that actually uses these fields is unchanged in this PR - if a field was skipped in the minimal representation, it would result in a compile error. The _only_ error prone operation is the conversion between the normal message and the minimal message, which I've double checked and also covered via existing unit tests.

#### Checklist


johanstokking (approved):
I would also be fine with keep using the original messages but define a field mask for storage.

Code quality wise I don't think it's necessary to declare all fields reserved with the original message's field name as they diverge. If it's just to make sure that our future selves don't add fields with the same numbers and old binary data gets unmarshaled and lead to corrupted data, we can just declare the reservations [in short form](https://developers.google.com/protocol-buffers/docs/proto3#reserved).

adriansmares:
I've minimised the `reserved` by removing the comments and joining the statements.

Regarding the field mask - The field mask tooling does not support applying field masks to repeated fields - see https://github.com/TheThingsIndustries/protoc-gen-fieldmask/issues/37 , and a lot of the dropped fields are in `rx_metadata`.

Offtopic: We probably should fix that before we add field masks to webhooks, since most of the filtering that I expect will be in the `rx_metadata` of `ApplicationUp`.

PR #5609, opened 2022-07-18
Add workflow dispatch to snapshot action

adriansmares:
#### Summary

This PR allows us to trigger the snapshot creation action via workflow dispatch.

#### Changes

- Add `workflow_dispatch` as a possible trigger to `release-snapshot.yml`.

#### Testing

I've forked the repo, merged this PR, and tested the output. The flow fails due to missing Docker credentials, but you can see that the checkout action succeeds and the output is expected:

Triggered via push: https://github.com/adriansmares/lorawan-stack/runs/7391807504?check_suite_focus=true - checks out the merge commit `8bacfbf62e669081fb38d7c6e49b44eb2da8d1fe`
Triggered manually: https://github.com/adriansmares/lorawan-stack/runs/7391839450?check_suite_focus=true - checks out the manually inputted commit `02b89ea7aa7ef7e5b4cc9c7ee31264d69b66d3c5`

##### Regressions

N/A

#### Checklist

### DIFF
--- a/.github/workflows/release-snapshot.yml
+++ b/.github/workflows/release-snapshot.yml
@@ -4,6 +4,11 @@ on:
   push:
     branches:
       - 'v3.*'
+  workflow_dispatch:
+    inputs:
+      commit:
+        description: Commit hash
+        required: true
 
 env:
   NODE_ENV: production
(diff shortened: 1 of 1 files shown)


.github/workflows/release-snapshot.yml
```diff
@@ -17,6 +22,7 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v2
         with:
+          ref: ${{ github.event.inputs.commit }}
```
adriansmares, line 27: On workflow dispatch, this evaluates to `''`, which is the default value for `ref` when not explicitly defined.

htdvisser: Maybe good to add that as a comment in the YAML

htdvisser (approved):

PR #5621, opened 2022-07-19
Fix packet broker network policy page being unaccessible

kschiffer:
#### Summary
Closes https://github.com/TheThingsIndustries/lorawan-stack/issues/3301

This PR fixes an issue that caused some packet broker network pages to be unaccessible due to a bug in how the frontend paginates through the list of available networks.

#### Changes
- Fix missing zero-index when paginating through List RPCs
- Change the pagination limit from `1000` to `100`, which appears to be the maximum. Using 1000 then messes with the pagination calculation.
- Fix errors persisting until a refresh when a packet broker network error occurred

#### Testing
Manual testing on the staging environment.

#### Notes for Reviewers
The Console has to paginate through the list of available networks, since the network information can only be obtained through a paginated RPC. The while loop that cycled through the list had a faulty break condition causing the networks on the last page to not be retrievable and hence causing the 404.

#### Checklist

### DIFF
--- a/pkg/webui/console/views/admin-packet-broker/network-routing-policy.js
+++ b/pkg/webui/console/views/admin-packet-broker/network-routing-policy.js
@@ -27,7 +27,6 @@ import Link from '@ttn-lw/components/link'
 import Message from '@ttn-lw/lib/components/message'
 import DateTime from '@ttn-lw/lib/components/date-time'
 import RequireRequest from '@ttn-lw/lib/components/require-request'
-import { FullViewErrorInner } from '@ttn-lw/lib/components/full-view-error'
 
 import RoutingPolicy from '@console/components/routing-policy'
 import RoutingPolicyForm from '@console/components/routing-policy-form'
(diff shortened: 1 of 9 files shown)


pkg/webui/console/views/admin-packet-broker/network-routing-policy.js
```diff
@@ -236,11 +236,6 @@ const NetworkRoutingPolicyView = props => {
   const tenantId = props.match.params.tenantId
   const combinedId = tenantId ? `${netId}/${tenantId}` : netId
   const registered = useSelector(selectRegistered)
-  const networkError = useSelector(selectPacketBrokerNetworkError)
-
-  if (networkError) {
-    return <FullViewErrorInner error={networkError} />
-  }
```
kschiffer, line 243: When the error is returned here, it will be before a new request has been launched, causing the error to be persistent even when navigating to other networks.

adriansmares (approved):

ryaplots (approved):

PR #5630, opened 2022-07-20
Add Account Search

### LINKED ISSUE #5027: Enable selecting users via autosuggest
#### Summary
We currently only allow specifying users via regular text input. This is inconvenient because it forces users to remember and/or copy user ids from other places. Instead, the user/org input should autosuggest available users when typing into it.

#### Why do we need this?
Improve UX of adding collaborators, contacts, owners, etc.

- User/org input via simple text input
- Search RPC for searching users/orgs `SearchOrganizations` and `SearchUsers`

We should introduce a `<UserInput />` component specifically for selecting users and/or orgs. It should 
- allow specifying whether it is possible to select only users or both users and organizations
- also allow selecting collaborators of an entity (instead of using the search RPC)
- allow fetching some initial options before anything is typed (useful for finding collaborators)

I think we can introduce a utility factory/HOC for this which allows us to generate specific auto-suggested inputs for various use cases. This would be similar to the `<FetchSelect />` container we currently use. It should:
1. Take a handler that takes in the current (debounced) value and return a list of suggestions (typically by piping it through n API request like `SearchUsers` or `ListCollaborators` based on use case)
2. Pipes the values from 1. and passes them as options to a `<Select />`

We can then use such component generator to define a `<UserInput />` that will autosuggest users and orgs. Similar to how we use `<FetchSelect />` with e.g. `<FreqPlansSelect />`.

End-to-end tests.

@ryaplots

### LINKED ISSUE #5504: "autocomplete / autosuggest" public API
#### Summary

A separate "account ID autocomplete / autosuggest" public API. 
References #5027 

#### Why do we need this?

In order to enable selecting users via auto-suggest (#5027), we need a public API with user names.

A search RPC restricted to admins only.

An "account ID autocomplete / autosuggest" API that understands the constraints.

-

-

@htdvisser

htdvisser:
#### Summary

This (draft) pull request adds account search to the search API and search implementation in the IS.

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/5504

Handing this over to @ryaplots who can add frontend so that this also closes https://github.com/TheThingsNetwork/lorawan-stack/issues/5027.

#### Changes

- [x] Define account search API
- [x] Implement account search in the IS store
- [x] Implement account search RPC
- [ ] ~Use account search in collaborator management in the Console https://github.com/TheThingsNetwork/lorawan-stack/issues/5027. Note that the current search implementation has some restrictions for non-admin users (see `pkg/identityserver/registry_search.go`).~

#### Testing

Unit tests only

##### Regressions

n/a

#### Notes for Reviewers

~At the moment this is a draft that only includes backend code. Frontend is still TODO.~

We've decided that frontend comes later.

#### Checklist


api/search_services.proto
```diff
@@ -262,6 +262,29 @@ message SearchUsersRequest {
   bool deleted = 10;
 }
 
+message SearchAccountsRequest {
+  option (thethings.flags.message) = { select: false, set: true };
+
+  string query = 11 [
```
nicholaspcr: shouldn't this be just `1`?

pkg/identityserver/bunstore/entity_search.go
```diff
@@ -345,3 +345,80 @@ func (s *entitySearch) SearchUsers(
...
+	// Count the total number of results.
+	count, err := selectQuery.Count(ctx)
+	if err != nil {
+		return nil, wrapDriverError(err)
+	}
+	store.SetTotal(ctx, uint64(count))
+
+	// Apply ordering, paging and field masking.
+	selectQuery = selectQuery.
+		Apply(selectWithLimitAndOffsetFromContext(ctx))
```
nicholaspcr: The comment doesn't seem appropriate for the section, since it just specifies the `limit` and `offset`.

nicholaspcr (approved):

pkg/identityserver/registry_search.go
```diff
@@ -446,6 +446,47 @@ func (rs *registrySearch) SearchUsers(ctx context.Context, req *ttnpb.SearchUser
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Limit the amount of data we provide to non-admin users when searching across all accounts.
+	if req.CollaboratorOf == nil && !authInfo.IsAdmin {
+		// Require a search query of at least 3 characters.
+		if len(req.Query) < 3 {
+			return &ttnpb.SearchAccountsResponse{}, nil
+		}
```
kschiffer: I get the rationale but this also means that users with 2 character IDs cannot be found.

kschiffer (approved):

pkg/identityserver/registry_search.go
htdvisser: Fair point, updated this to require a minimum length of 2

PR #5634, opened 2022-07-22
Allow disabling downlink message scheduling

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack-docs/issues/926#issuecomment-1192460307

#### Changes

- Add the `schedule_downlinks` MAC settings field.
   - Nullable, `nil` being interpreted as `true`.
- Skip processing downlink tasks if `schedule_downlinks` is set and `false`.
   - Pushing / replacing the downlink queue, and updating the end device, already cause the downlink tasks to be refreshed, so updating this option should be seamless.

#### Testing

Local testing.

##### Regressions

This functionality allows the users to disable scheduling downlinks. One of the possible risks here is that re-enabling downlink scheduling may cause downlink scheduling to not resume. This is not an issue: each time the `mac_settings.schedule_downlinks` / `mac_settings.schedule_downlinks.value` paths are updated, the NS will schedule a new downlink tasks that will check if some kind of downlink (join accept, class A downlink, network initiated downlink) has to be scheduled.

#### Checklist


pkg/networkserver/grpc_gsns.go
```diff
@@ -1166,7 +1166,11 @@ func (ns *NetworkServer) handleJoinRequest(ctx context.Context, up *ttnpb.Uplink
 
 	if !matched.SupportsJoin {
 		log.FromContext(ctx).Warn("ABP device sent a join-request, drop")
-		queuedEvents = append(queuedEvents, evtDropJoinRequest.NewWithIdentifiersAndData(ctx, matched.Ids, errABPJoinRequest))
+		queuedEvents = append(queuedEvents, evtDropJoinRequest.NewWithIdentifiersAndData(ctx, matched.Ids, errABPJoinRequest.New()))
+		return nil
+	}
+	if !mac.DeviceScheduleDownlinks(matched, ns.defaultMACSettings) {
```
adriansmares, line 1172: Since we're dropping the join request so early, we will not consume the `nonce`. For me this seems safe (it is equivalent to us not hearing the join request), but are there any other arguments against this ?

pkg/networkserver/grpc_deviceregistry.go
```diff
@@ -875,6 +877,8 @@ var (
 
 	// downlinkInfluencingSetFields contains fields that can influence downlink scheduling, e.g. trigger one or make a scheduled slot obsolete.
 	downlinkInfluencingSetFields = [...]string{
+		"mac_settings.schedule_downlinks",
+		"mac_settings.schedule_downlinks.value",
 		"last_dev_status_received_at",
```
johanstokking: This isn't ordered

johanstokking (approved):

PR #5636, opened 2022-07-22
Fix DeviceTimeAns timestamp not corresponding to the specification

### LINKED ISSUE #5586: DeviceTimeAns timestamp is not corresponding to the specification

happyRip:
#### Summary

Closes #5586 

#### Changes

- Improved [HandleDeviceTimeReq](https://github.com/TheThingsNetwork/lorawan-stack/blob/70d88dd86a9b17bf8faf1bbea6961d6c5d3f2ae7/pkg/networkserver/mac/device_time.go#L34-L43) timestamp
- Reverted the change that made the MAC handler run before the deduplication occurred

#### Testing

- Added test cases regarding the improvement

##### Regressions

n/a

#### Notes for Reviewers

nil

#### Checklist

### DIFF
--- a/pkg/networkserver/mac/device_time.go
+++ b/pkg/networkserver/mac/device_time.go
@@ -17,6 +17,8 @@ package mac
 import (
 	"context"
 
+	pbtypes "github.com/gogo/protobuf/types"
+
 	"go.thethings.network/lorawan-stack/v3/pkg/events"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 )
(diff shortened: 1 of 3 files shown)


pkg/networkserver/mac/device_time.go
```diff
@@ -33,11 +35,27 @@ var (
...
 	}, nil
 }
+
+func messageTimestamp(msg *ttnpb.UplinkMessage) *pbtypes.Timestamp {
+	var ts *pbtypes.Timestamp
+	for _, md := range msg.RxMetadata {
+		if t := md.GpsTime; t != nil {
+			return t
+		}
+		if ts == nil && md.Time != nil {
```
adriansmares: This should be `md.ReceivedAt`, and the tests should be updated.

`md.Time` is the timestamp from the gateway with undefined accuracy, so using it is not ideal.

adriansmares (approved):

PR #5640, opened 2022-07-25
Add manual device type form (Device Onboarding)

kschiffer:
#### Summary
This PR adds the manual device type form section to the new device onboarding form.

<details><summary>Screenshots</summary>

image

image

image

</details>

#### Changes
- Update form scaffold
  - Remove `connect()` HOCs and use selector hooks instead
  - Remove duplicate sumbit button
  - Remove component switches (DO will now only work when all components are present)
- Refactor `<Form />` and `<Form.Field />`-components
  - Remove `stripUnusedFields` and spread cleaned form values onto the submit handler instead (third argument)
  - Add seperate field registry (to enable stripping of unmounted fields)
  - Allow passing `connectedFields` to the `Field` component to avoid stripping indirect fields
  - Pass old value to the field encoder to enable composing new field values based on the old one
- Refactor `<UnitInput.Factored />` to enable storing the last unit when no value is present
- Update the Device Type Form Section
  - Fix the warning modal when changing input method
- Update the Device Repository Type Form
  - Fetch and apply the device template when the fields are filled
  - Use encoder/decoder for setting `version_id` values form `brand_id` and `model_id`
  - Fix use of array as dependant for `useEffect` which caused infinite render loops
  - Remove superfluous style module
  - Improve logic of when to show info/hints a bit
- Add Manual Device Type Form Section
  - Refactor validation schema to not include any transforms/skips and to dynamically build the validation schema for `mac_settings` based on the current default mac settings

#### Testing

For now only manual testing.

##### Regressions

The form changes might break other forms but that would be caught by the end-to-end tests.

#### Notes for Reviewers
I'll make some comments in the diff.

Note that this does not target master but our dedicated feature branch.

#### Checklist


pkg/webui/components/unit-input/common/duration.js
```diff
@@ -27,10 +27,13 @@ const units = [
   { label: sharedMessages.hours, value: 'h' },
 ]
 
-const encoder = (value, unit) => (value ? `${value}${unit}` : unit)
```
kschiffer, line 30: Saving just the unit would require transformation logic if the fields should allow empty values. Rather, the field should correctly output null and at the same time keep the last used unit in the state. This way, the unit can be changed even when no integer was entered.

pkg/webui/console/containers/device-onboarding-form/device-onboarding-form.js
```diff
@@ -1,67 +0,0 @@
...
-  }, [])
-
-  return (
-    <Form
-      onSubmit={handleSubmit}
-      initialValues={initialValues}
-      validationSchema={validationSchema}
-      stripUnusedFields
-    >
-      <DeviceTypeFormSection appId={appId} getRegistrationTemplate={getRegistrationTemplate} />
```
kschiffer, line 48: Instead of passing state-related props through as component props, it is preferable to do the necessary selection in the child component itself. Keep in mind that being able to avoid prop waterfalls is one of the main reasons to use a global state engine like redux.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/index.js
```diff
@@ -42,8 +48,21 @@ const initialValues = merge(
...
+  const {
+    values: {
+      _inputMethod,
+      version_ids: version,
+      frequency_plan_id,
+      lorawan_version,
+      lorawan_phy_version,
+    },
+  } = useFormContext()
+  const template = useSelector(selectDeviceTemplate)
```
kschiffer, line 62:
```suggestion
  const template = useSelector(selectDeviceTemplate)
```

Note my use of selector hooks here and in the other components. I think we should move towards using those by default instead of the `connect()`-HOC. The problem I have with `connect()` are:
- It's not following the hooks paradigm.
- It generates a new component only for passing through state data.
- I also don't really like outsourcing the `connect()` into it's own module. It can be nice for complex components but I typically find myself looking for the origin of certain props which is effectively hidden when looking at the component alone.

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/device-type-manual-form-section/advanced-settings-section.js
```diff
@@ -0,0 +1,521 @@
...
+          _device_class: isMulticast
+            ? values._device_class === DEVICE_CLASS_MAP.CLASS_A
+              ? ''
+              : _device_class
+            : _device_class,
+        },
+        isUndefined,
+      ),
+    )
+  }, [
```
kschiffer: Note how applying field modifications/side effects in this centralized manner really helps to improve clarity over how fields are behaving.

We used to do these modifications as part of the validation schema which was really hard to maintain IMO as it would be convoluted with the validation rules and other `Yup`-specific idioms.

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/device-type-manual-form-section/advanced-settings-section.js
```diff
@@ -0,0 +1,521 @@
...
+    mayChangeToDefaultSettings,
+    setValues,
+  ])
+
+  return (
+    <Form.CollapseSection id="advanced-settings" title={m.advancedSectionTitle}>
+      <Form.Field
+        title={sharedMessages.activationMode}
+        name="_activation_mode"
+        connectedFields={['supports_class_b', 'supports_class_c', 'multicast']}
```
kschiffer: Connected fields are used to prevent such fields from being stripped when the values are cleaned.

In this case, the `_activation_mode` field actually corresponds to these three fields that it controls. These fields should only be stripped when the activation mode field is stripped (which never happens in this case).

pkg/webui/components/form/index.js
```diff
@@ -60,13 +60,13 @@ const Form = props => {
     validateOnChange,
     validateOnMount,
     validateSync,
-    validationContext,
+    validationContext: initialValidationContext,
     validationSchema,
-    stripUnusedFields,
```
kschiffer: Instead of using the prop, we can now just use the third argument of the `onSubmit()`-callback.

```js
const handleSubmit useCallback((values, { resetForm }, cleanedValues) => { … }, […])
```

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/device-type-manual-form-section/validation-schema.js
```diff
@@ -12,14 +12,148 @@
...
+  _default_ns_settings: Yup.bool(),
+  _activation_mode: Yup.mixed().when(['$mayEditKeys'], (mayEditKeys, schema) => {
+    // If the user may not set any root keys, the only activation option is OTAA.
+    if (!mayEditKeys) {
+      return schema.oneOf([ACTIVATION_MODES.OTAA]).required(sharedMessages.validateRequired)
+    }
+
+    return schema.oneOf(Object.values(ACTIVATION_MODES)).required(sharedMessages.validateRequired)
+  }),
+  _skip_js_registration: Yup.bool(),
```
kschiffer: Note how validation schemas become a lot cleaner now that they can focus on validation only and not on any transformation logic.

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/device-type-repository-form-section/device-selection/band-select/index.js
```diff
@@ -12,9 +12,64 @@
...
+
+const formatOptions = (profiles = []) =>
+  profiles
+    .map(profile => ({
+      value: profile,
+      label: profile,
+    }))
+    .concat([{ value: SELECT_OTHER_OPTION, label: messages.otherOption }])
+
+const BandSelect = props => {
```
kschiffer, line 42: For this component I refactored it to use just a single module and use `useSelector`-hooks directly. I think this makes the component a lot easier to grasp.

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/device-type-repository-form-section/index.js
```diff
@@ -38,6 +49,30 @@ import BandSelect from './device-selection/band-select'
...
+        brand_id: oldValue.brand_id,
+        model_id: value,
+        hardware_version: undefined,
+        firmware_version: undefined,
+        band_id: undefined,
+      }
+    : oldValue
+
+const decodeBrand = value => (isPlainObject(value) ? value.brand_id : '')
+const decodeModel = value => (isPlainObject(value) ? value.model_id : '')
```
kschiffer: I used the decoder/encoder pattern for the `brand_id` and `model_id` fields because using the `onChange`-handlers would cause the form values to be updated with faulty values for a short time until the `setValues()` hook was executed. This would not be a big issue usually but in this case, it was since the added `useEffect()` hook on line 131 would fire whenever a completed set of version IDs was detected. If a version ID that contained a faulty combination of values was there, it would fire request for these values which would fail.

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/index.js
```diff
@@ -38,61 +37,74 @@ const initialValues = merge(
...
       <Form.SubTitle title={m.endDeviceType} />
       <Form.Field
         title={m.inputMethod}
-        name="_inputMethod"
         component={Radio.Group}
         onChange={handleMethodChange}
+        value={_inputMethod}
+        name="__inputMethod"
+        connectedFields={['_inputMethod']}
       >
```
kschiffer: We were originally only using the `onChange` handler to control the modal here. The problem with this is that the `onChange` handler only gets executed after the value was already changed in the form. So regardless of the modals decision, the value would already be changed regardless.
To fix this, I added an intermediate field value `__inputMethod` which is not really used but allows us to control the actual value `_inputMethod` based on the result of the modal.

I also changed the conditions for when the modal would show, since it should only show when some progress in the form was actually already made which would be lost.

pkg/webui/console/containers/device-onboarding-form/index.js
```diff
@@ -12,9 +12,75 @@
...
+      <DeviceTypeFormSection />
+      <DeviceProvisioningFormSection />
+      {maySubmit && (
+        <SubmitBar>
+          <Form.Submit message={sharedMessages.addDevice} component={SubmitButton} />
+        </SubmitBar>
+      )}
+    </>
+  )
+}
```
kschiffer, line 64: I had to seperate the topmost form component into two components to be able to tap into the form context.
Alternatively one could have also done something like
```js
<Form
  ...
/>{({values}) => {
}}
</Form>
```
But I decided against that for consistency.

kschiffer (commented):
Adding some explanatory comments of my own

pkg/webui/components/form/field/index.js
```diff
@@ -103,9 +106,23 @@ const FormField = props => {
...
   })
 
+  // Apply any connected names to the field registry.
+  useEffect(() => {
+    if (connectedFields) {
+      addToFieldRegistry(connectedFields)
+      return () => {
+        removeFromFieldRegistry(connectedFields)
+      }
+    }
```
ryaplots: Why do you add to field registry and then remove right after?

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/device-type-manual-form-section/advanced-settings-section.js
```diff
@@ -0,0 +1,521 @@
...
+  const [defaultMacSettings, setDefaultMacSettings] = useState(initialValues.mac_settings)
+
+  const mayEditKeys = useSelector(state => checkFromState(mayEditApplicationDeviceKeys, state))
+
+  // Fetch and apply default MAC settings, when FP or PHY version changes.
+  useEffect(() => {
+    let isMounted = true
+    const updateMacSettings = async () => {
+      try {
+        const settings = await tts.Ns.getDefaultMacSettings(frequency_plan_id, lorawan_phy_version)
```
ryaplots: Is it okay to use the API directly here? Wouldn't it be more consistent if we used the store?

pkg/webui/console/containers/device-onboarding-form/messages.js
```diff
@@ -20,6 +20,53 @@ export default defineMessages({
   inputMethod: 'Input Method',
   inputMethodDeviceRepo: 'Select the end device in the LoRaWAN Device Repository',
   inputMethodManual: 'Enter end device specifics manually',
+  registerEndDevice: 'Register end device',
+  claimEndDevice: 'Claim end device',
+  continueManual: 'Please enter versions and frequency plan information above to continue',
+  continueDeviceRepo: 'Please specify the your device above to continue',
```
ryaplots:
```suggestion
  continueDeviceRepo: 'Please specify your device above to continue',
```

pkg/webui/console/containers/device-onboarding-form/messages.js
```diff
@@ -20,6 +20,53 @@ export default defineMessages({
...
+  submitTitle: 'Register end device',
+  activationModeNone: 'Do not configure activation',
+  afterRegistration: 'After registration',
+  singleRegistration: 'View registered end device',
+  multipleRegistration: 'Register another end device of this type',
+  createSuccess: 'End device registered',
+  deviceIdDescription: 'This value is automatically prefilled using the DevEUI',
+  // Manual messages.
+  basicTitle: 'Basic settings',
+  basicDescription: "End device ID's, Name and Description",
```
ryaplots, line 44:
```suggestion
  basicDescription: 'End device ID's, Name and Description',
```

pkg/webui/console/containers/device-onboarding-form/messages.js
```diff
@@ -20,6 +20,53 @@ export default defineMessages({
...
+  frequencyPlaceholder: 'e.g. 869525000 for 869,525 MHz',
+  pingSlotPeriodicityTitle: 'Ping slot periodicity',
+  pingSlotPeriodicityValue: '{count, plural, one {every second} other {every {count} seconds}}',
+  pingSlotDataRateTitle: 'Ping slot data rate',
+  pingSlotFrequencyTitle: 'Ping slot frequency',
+  rx2DataRateIndexTitle: 'Rx2 data rate',
+  rx2FrequencyTitle: 'Rx2 frequency',
+  classBTimeout: 'Class B timeout',
+  classCTimeout: 'Class C timeout',
+  defaultNetworksSettings: "Use network's default MAC settings",
```
ryaplots, line 66:
```suggestion
  defaultNetworksSettings: 'Use network's default MAC settings',
```

ryaplots (commented):
Your comments were very useful. I took notes on what to change in my PR. I just have a few comments.

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/device-type-repository-form-section/device-selection/band-select/index.js
ryaplots: I will use those in the provisioning part as well

pkg/webui/components/form/field/index.js
kschiffer: If you return a function from `useEffect()`, it is used as a cleanup function when the component unmounts:
https://reactjs.org/docs/hooks-effect.html#effects-with-cleanup

Essentially this means to add the field to the registry on mount and removing it from the registry on unmount.

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/device-type-manual-form-section/advanced-settings-section.js
kschiffer: Yeah, this should be wrapped around a store action indeed.

pkg/webui/console/containers/device-onboarding-form/messages.js
kschiffer: Note the `'` within the message which warrants the switch to double quotes.

pkg/webui/console/containers/device-onboarding-form/messages.js
ryaplots: Nevermind this as well

ryaplots (approved):
Looks good so far

kschiffer:
I've made a couple of changes after finding some issues with the old implementation These changes are based on top of #5678
This includes:
- Getting rid of `connectedFields` and using composite fields instead
- Making use of `valueSetters` to achieve complex form value updates in one go, without the need for effect hooks
- A couple of small fixes unrelated to the form updates

PR #5645, opened 2022-07-26
Fix panic on deletion of ABP devices

### LINKED ISSUE #5560: SIGSEGV on `ttn-lw-cli device delete` 3.20.0
#### Summary

Deleting an end-device using `ttn-lw-cli` can throw a SIGSEGV.

#### Steps to Reproduce

Not 100% sure what the trigger is (have deleted a device with this code before).

What I did in this case:
1. create the test device using `ttn-lw-cli`
2. deleted the test device in the console
3. created the test device in the console (same name, but in ABP mode)
4. attempt to run a script that identified ABP mode device, deletes it before recreating as OTAA.

The command run was:

``` bash
ttn-lw-cli --config "${ipath}/ttnconfig/ttn-lw-cli-lcs.yml" device delete --application-id ${app_id} --device-id ${dev_id}
```
_NB: I just changed absolute values to bash syntax variables to abstract personal details out of the invocation. The values were valid and correct.  I've used them on the CLI in other contexts to create this device before._

The response was:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x11acb74]

goroutine 1 [running]:
go.thethings.network/lorawan-stack/v3/cmd/ttn-lw-cli/commands.glob..func85(0x2544958, {0x2dbc3c0, 0x0, 0x6})
  /home/runner/work/lorawan-stack/lorawan-stack/cmd/ttn-lw-cli/commands/end_devices.go:1014 +0x498
github.com/spf13/cobra.(*Command).execute(0x2544958, {0x2dbc390, 0x6, 0x6})
  /home/runner/go/pkg/mod/github.com/spf13/cobra@v1.2.1/command.go:856 +0x6b0
github.com/spf13/cobra.(*Command).ExecuteC(0x254b4f8)
  /home/runner/go/pkg/mod/github.com/spf13/cobra@v1.2.1/command.go:974 +0x414
main.main()
  /home/runner/work/lorawan-stack/lorawan-stack/cmd/ttn-lw-cli/main.go:27 +0x20
```

#### Environment

This was using the TTN community server: the eu1.cloud.thethings.network:8884

Test host is running 'Raspbian':
```
Version:             3.20.0
Build date:          2022-06-17T09:38:56Z
Git commit:          899adb340
Go version:          go1.18.3
OS/Arch:             linux/arm
```

Someone familiar with the code looking at the stack trace and working out how a nil pointer is being abused with no error checking or appropriate try/catch.

Should  further logs or access to the test device be useful I'll work with an interested party to get access.

Sorry not to fix it myself.  Hope this report is still a helpful clue!

Thanks for the hard work!

nicholaspcr:
#### Summary

Closes #5560

Effectively what happened is that when creating an end-device that uses ABP the `JoinEUI` field was set as nil and the `.Bytes()` method caused it to panic. I tried looking at other places in which this would happen but couldn't find any that aren't validated through a `joinEui != nil`.

#### Changes

- Added `OrZero()` to generate an empty `EUI64` variable instead of `nil`

#### Testing

Manual testing

##### Regressions

Bug fix, there should be none.

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/end_devices.go
+++ b/cmd/ttn-lw-cli/commands/end_devices.go
@@ -1004,7 +1004,7 @@ var (
 				return errAddressMismatchEndDevice.New()
 			}
 
-			if existingDevice.JoinServerAddress == "" {
+			if existingDevice.JoinServerAddress == "" && devID.GetJoinEui() != nil {
 				// Attempt to unclaim device via the DCS.
 				dcs, err := api.Dial(ctx, config.DeviceClaimingServerGRPCAddress)
 				if err != nil {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Fixed
 
 - Console showing `404 Not Found` errors for pages containing user IDs in the path, when the user ID has a length of two.
+- CLI no longer panics when deleting a device without JoinEUI, this scenario only occurred when deleting a device that uses ABP.
 
 ### Security
 


cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -1011,7 +1011,7 @@ var (
 					return err
 				}
 				claimInfoResp, err := ttnpb.NewEndDeviceClaimingServerClient(dcs).GetInfoByJoinEUI(ctx, &ttnpb.GetInfoByJoinEUIRequest{
-					JoinEui: devID.JoinEui.Bytes(),
+					JoinEui: devID.GetJoinEui().OrZero().Bytes(),
```
adriansmares: This is not ideal due to the fact that a Join EUI of all zeroes is considered valid.

I think it would be better to check for `dev.GetJoinEui() != nil` on line 1007 (in addition to the empty address check).

nicholaspcr: I was under the impression that an JoinEUI fileld with zeros is a valid field but it would be discarded. But reflecting on it, if it was `nil` it really should't even be doing the request in the first place.

adriansmares (approved):

PR #5655, opened 2022-07-29
Add `field_mask` to webhooks and webhook templates

### LINKED ISSUE #5597: Enable field masks for webhook messages
### Summary

We should provide a way to pass field masks to the event streams and event data that is passed to integrations so that any unwanted data is stripped and bandwidth is saved.

### Current Situation

Webhook data is always passed in full. There is no way to control what (relevant) data is being passed. This increases the required bandwidth as well as noise.

- Make it easier to retrieve relevant information / remove noise from event streams and integrations
- Decrease bandwidth for streams and integrations
- To configure a default set of field masks and only show all data in a verbose mode

This feature will be used by users relying on webhook integrations, that are only interested in some data points within the message data.

### Proposed Implementation

Since the event types that are being passed through webhooks are well known, it should be relatively easy to allow configuring a set of field masks and filtering the messages accordingly. I cannot propose a concrete implementation plan since I'm not familiar with how webhooks work in the backend.

In the Console, we would need an additional config panel for such field masks, which should work on a per-event-type level.

### Contributing

### Code of Conduct

happyRip:
#### Summary

Closes #5597

#### Changes

- Add `field_mask` field to webhook templates and entities
- Apply the field mask (if present) before marshalling the uplink
- Console integration

#### Testing

Tested manually by adding `field_mask` to a webhook with

```
$ ttn-lw-cli app webhooks set appname webhookname --field-mask received_at
```

When no `field_mask` is set all fields are sent.

When there is a `field_mask` it filters the fields to be sent.

---

Added `FieldMask` field to relevant unit tests.

##### Regressions

n/a

#### Notes for Reviewers

No frontend integration of the new `field_mask` field was added.

#### Checklist


pkg/applicationserver/io/web/webhooks.go
```diff
@@ -332,6 +333,13 @@ func (w *webhooks) newRequest(ctx context.Context, msg *ttnpb.ApplicationUp, hoo
 	if !ok {
 		return nil, errFormatNotFound.WithAttributes("format", hook.Format)
 	}
+	if paths := hook.FieldMask.GetPaths(); len(paths) > 0 {
```
adriansmares, line 371: This is what was requested in the issue, and it will work, but it will not be ideal. 

The `ApplicationUp` message has a `oneof` field called `up` - it contains the main contents of the message - the `JoinAccept`, the `UplinkMessage` and so forth. You can see a full list of viable field mask paths here:
https://github.com/TheThingsNetwork/lorawan-stack/blob/dbd03c4e5f3639dae5432b23c999a79937d37471/pkg/ttnpb/messages.pb.paths.fm.go#L752-L923

The problem is that by design `SetFields` won't allow two `oneof` fields to be provided. Specifically the following field mask is invalid:
- `up.join_accept.received_at`
- `up.uplink_message.decoded_payload`

Since `up` is a `oneof` field, it is impossible for it to be both `join_accept` and `uplink_message`, so providing both basically renders the webhook unusable (`SetFields` will always return an error).

We can avoid this issue by doing the following thing:
- You can retrieve all of the fields in the mask which are _not_ `up` using `ttnpb.ExcludeSubFields(paths, "up")`. These are to be kept in the field mask that we want to apply.
1. You then have to check which `oneof` field of `up` is set. This will be a switch akin to the one on L283.
2. Based on the active message, you can exclude paths which are not part of the active `oneof`. Consider building a function in `pkg/ttnpb/fieldmask_utils.go` for this.
3. Take the union between the fields from (1) and (2). Use that as a field mask.

adriansmares (commented):
LGTM. I have a longer comment bellow regarding how to make this functionality be more useful with our current `oneof` structure.

Could you also please add/update existing unit tests to cover this ?

pkg/applicationserver/io/web/webhooks.go
happyRip: Regarding the testing:
According to
```
$ tools/bin/mage go:Cover && go tool cover -html coverage.out
```
there is no coverage of the file at least in go tests. I had trouble producing those from the ground up on my own, but I can do that with some guidance if necessary.

Regarding the improvement the description is clear, I'm on it.

adriansmares: image

Make sure that `go:cover` actually did finish successfully. You can find the tests in `webhooks_test.go`.

adriansmares (approved):

adriansmares:
Don't forget to update the documentation at https://github.com/TheThingsIndustries/lorawan-stack-docs/tree/master/doc/content/integrations/webhooks/webhook-templates to also include the field masks.

adriansmares (approved):
LGTM.

Please don't forget to add Console support for this. All you need to do is to add `field_mask: template.field_mask` here:
https://github.com/TheThingsNetwork/lorawan-stack/blob/09314345106cebbb8113a6eb77d9b6d4ca260333/pkg/webui/console/containers/webhook-add/webhook-add.js#L52-L71

kschiffer:
When trying to set the field mask via the API, I get:
```
{
  "code": 3,
  "message": "json: cannot unmarshal array into Go value of type map[string]json.RawMessage"
}
```
Here's the request:
```
curl 'http://localhost:8080/api/v3/as/webhooks/232342/rstrst' \
  -X 'PUT' \
  -H 'Accept: application/json, text/plain, */*' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/json' \
  … [redacted for brevity]
  --data-raw '{"webhook":{"service_data":null,"location_solved":null,"downlink_queue_invalidated":null,"downlink_queued":null,"downlink_failed":null,"downlink_sent":null,"downlink_nack":null,"downlink_ack":null,"join_accept":null,"downlink_api_key":"","field_mask":["rstrst"],"headers":{}},"field_mask":{"paths":["downlink_ack","downlink_api_key","downlink_failed","downlink_nack","downlink_queue_invalidated","downlink_queued","downlink_sent","field_mask","headers","join_accept","location_solved","service_data"]}}' \
  --compressed

**Comment by adriansmares** [comment]
Good catch - since it is a `google.protobuf.FieldMask`, we need to provide it as 
```
		"field_mask": {
			"paths": ["rstst"]
		},
```
instead.

**Comment by kschiffer** [comment]
image

I've checked Console as well and setting field masks via templates is working. I've also added functionality to add field masks for custom webhooks.
While doing that I also fixed a bug which lead to the webhook form not displaying certain errors.

File: pkg/applicationserver/io/web/webhooks.go [hunk 2]
```diff
@@ -244,6 +244,7 @@ func (w *webhooks) handleUp(ctx context.Context, msg *ttnpb.ApplicationUp) error
 			"service_data",
 			"uplink_message",
 			"health_status",
+			"field_mask",
```
johanstokking: nit; sort ascending (this was already wrong)

pkg/ttnpb/fieldmask_utils.go
```diff
@@ -317,3 +317,17 @@ func ApplySessionKeysFieldMask(dst, src *SessionKeys, paths ...string) (*Session
 func FilterGetSessionKeys(pb *SessionKeys, paths ...string) (*SessionKeys, error) {
 	return ApplySessionKeysFieldMask(nil, pb, AddImplicitSessionKeysGetFields(paths...)...)
 }
+
+// IncludeFields returns the given paths with the given search paths to include.
+func IncludeFields(paths []string, includePaths ...string) []string {
```
johanstokking, line 322: How is this different from `AllowedFields()`?

adriansmares: `AllowedFields` uses only full paths (`ContainsField`), while this one allows prefix paths (`HasAnyField`).

happyRip: `AllowedFields()` uses `ContainsField()` while `IncludeFields()` uses `HasAnyField()`

johanstokking: Ah ok

johanstokking (approved):

adriansmares (approved):

pkg/webui/locales/en.json
```diff
@@ -479,6 +482,93 @@
   "console.containers.device-importer.index.abortWarningTitle": "Device import aborted",
   "console.containers.device-importer.index.abortWarningMessage": "The end device import was aborted and the remaining {count} {count, plural, one {end device} other {end devices}} have not been imported",
   "console.containers.device-importer.index.largeFileWarningMessage": "Providing files larger than {warningThreshold} can cause issues during the import process. We recommend you to split such files up into multiple smaller files and importing them one by one.",
+  "console.containers.device-onboarding-form.device-provisioning-form-section.device-claiming-form-section.validation-schema.validateCode": "Claim authentication code must consist only of numbers and letters",
```
adriansmares: This file needs to be regenerated (`./tools/bin/mage go:messages js:translations`)

happyRip: on it

kschiffer: Thank you!

philippdzm:
Hi @kschiffer 

> I've also added functionality to add field masks for custom webhooks.

This does not work for me. 

I think, I am the first wanting to use `field-mask` with custom webhook template. Can you try updating the `dezem.yml` and test this? Here is my change: https://github.com/deZemLabs/lorawan-webhook-templates/commit/189b2d99a2456e6cfa5cd990d84be89ba8940f4b

My tests showed that the lorawan-stack seem to ignore `field-mask`. I Assume this is a bug.

Can you check that? That'd be nice

Edit: To be more precise on what is not working for me: When applying a webhook template in the console, I expect to see entries in "Filter event data". This was not the case.

kschiffer:
Hi @philippdzm! By "custom webhooks" I meant the ones that are set in the Console when using the "Custom Webhook" button, meaning webhooks without template. If field masking in your custom webhook is not working, then @happyRip might want to take another look at it. Iirc though, some fields cannot be filtered out from the webhook data at all, but I might be wrong. In any case, @happyRip / @adriansmares can assist you better with this.

philippdzm:
Thanks @kschiffer . I edited my question to be more precise on what was not working. Hopefully your colleagues can have a look and help.

kschiffer:
I actually identified the problem, which was indeed coming from the Console. I have opened a PR and it should land in the next version next week.

#5655 

Thanks for notifying us of this issue!

philippdzm:
@kschiffer amazing! I assume, you meant https://github.com/TheThingsNetwork/lorawan-stack/pull/6204 instead of https://github.com/TheThingsNetwork/lorawan-stack/pull/5655 (which is this PR) 👍

philippdzm:
Another question.
How come that the paths in field-mask diverge from the paths in the Webhook's payload?

**Example for the spreading factor**
```
field-mask
    - mask up.uplink_message.settings.data_rate.modulation.lora.spreading_factor
```
and in the payload of the Webhook
`uplink_message->settings->data_rate->lora->spreading_factor`

Based on the documentation, I was not able to understand this.

adriansmares:
Field masks are centered around the `protobuf` format used to actually pass these messages around inside The Things Stack. 

Fields such as the `data_rate` have `oneof` fields such as `modulation` - you cannot have more than one modulation at the same time. The wrapper (`modulation`) is not rendered in JSON, but it exists at `protobuf` level.

PR #5656, opened 2022-07-29
Remove root CA caching in client config

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/issues/3187#issuecomment-1197902496

#### Changes

- Remove the caching mechanism for the root CA in the TLS client configuration
  - The caching was always unsafe due to the fact that we were copying `sync.Once`.
  - I've looked at the call sites for `ApplyTo` - they are always part of the initialization of an object / component - we can bear the file read cost there. 
  - The only runtime configuration situation is in `ConfigurationProvider.GetTLSClientConfig`, but that one always generated the `tlsconfig.Config` every time anyway, so no real caching was involved.

#### Testing

UT.

##### Regressions

N/A.

#### Checklist

### DIFF
--- a/pkg/redis/redis.go
+++ b/pkg/redis/redis.go
@@ -153,25 +153,22 @@ func (c FailoverConfig) Equals(other FailoverConfig) bool {
 
 func (c Config) makeDialer() func(ctx context.Context, network, addr string) (net.Conn, error) {
 	var (
-		tlsConfig    *tls.Config
+		dialer interface {
+			DialContext(ctx context.Context, network, addr string) (net.Conn, error)
+		} = &net.Dialer{}
 		tlsConfigErr error
 	)
 	if c.TLS.Require {
-		tlsConfig = &tls.Config{
+		tlsConfig := &tls.Config{
 			MinVersion: tls.VersionTLS12,
 		}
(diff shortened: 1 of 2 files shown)


pkg/redis/redis.go
```diff
@@ -153,25 +153,22 @@ func (c FailoverConfig) Equals(other FailoverConfig) bool {
...
 		defer trace.StartRegion(ctx, "dial redis").End()
-		var dialer interface {
-			DialContext(ctx context.Context, network, addr string) (net.Conn, error)
-		} = &net.Dialer{}
-		if c.TLS.Require {
-			if tlsConfigErr != nil {
-				return nil, tlsConfigErr
-			}
-			dialer = &tls.Dialer{NetDialer: dialer.(*net.Dialer), Config: tlsConfig}
+		if tlsConfigErr != nil {
```
adriansmares, line 170: I wanted to pull this error handling out of the dialer function but it proved to have very deep roots. I'll come back to this in another PR.

htdvisser (approved):
I thought this was necessary to allow users to use a custom Root CA for interop, webhooks and that sort of stuff. But if we don't, there's indeed no reason to have all this complexity.

Issue #5512, opened 2022-06-02
Load device profile from DR when using CSV import


kschiffer:
#### Summary
When importing end devices in bulk using the end device import feature, the device profile should be applied as provided from the DR.

#### Why do we need this?

- Facilitate device import, allow importing devices from keys only, without the need of providing settings manually
- Make better use of data we have in the DR

#### What is already there? What do you see now?

- Device profiles in the DR
- End device import feature
- The device selector from the single device onboarding

#### What is missing? What do you want to see?

- Automatically apply device profile (mac settings, MAC/PHY, FP, etc) to the imported devices
- Add the device model selector from the single device onboarding to the device import UI (so it can be used as fallback)

#### How do you propose to implement this?

I see two ways:

1. Load the profile in the frontend and apply it before running the creation API requests
2. Load the profile in the backend (EDTC) and apply it to the output

I believe that is more desirable because it would be applicable for CLI as well. We might also want to allow disabling this automatic device profile application if wished.

#### Can you do this yourself and submit a Pull Request?

I can do the frontend part but I need help for adding the backend functionality.

cc @adriansmares

nicholaspcr:
When working on this I had an error when making a request to the Device Repository in which it states that `rights` are needed for the request.
```
// FetchProfile provides the end device profile.
func (dtc *DeviceTemplateConverter) FetchProfile(device *ttnpb.EndDevice) (*ttnpb.EndDeviceTemplate, error) {
	conn, err := grpc.Dial(dtc., grpc.WithInsecure())
	if err != nil {
		logger.Error("NICK -> ", err)
		return nil, err
	}

	return ttnpb.NewDeviceRepositoryClient(conn).GetTemplate(context.Background(), &ttnpb.GetTemplateRequest{
		ApplicationIds: device.GetIds().GetApplicationIds(),
		VersionIds:     device.GetVersionIds(),
	})
}
// Generates
ERROR  rpc error: code = PermissionDenied desc = error:pkg/auth/rights:no_application_rights (no rights for application `test-app`)
```
After having a talk with @adriansmares the request of application rights on the DR seems to be a bit strange, when the device repository was specified by https://github.com/TheThingsNetwork/lorawan-stack/issues/3433, one of the comments were `The API in principle doesn't require rights, but do require any authentication`.

There are two options in which we can change this:
[...]

PR #5659, opened 2022-07-29
Deprecate ApplicationID from Device Repository

nicholaspcr:
#### Summary

References #5512 

#### Changes

- Deprecate ApplicaitonID on DR
- Update unit tests on DR
- Update mockis
  - Added a mutex for the registries
  - Added a simple entity access mock
- Update IS's entity access to return `nil` when `AuthType` isn't specified

#### Testing

Manual testing. 
- Requested a template via curl to see if I would get an authentication error.
- Validated that console still works with this change. 

##### Regressions

The main question for me is if the IS change will have any repercussion on other components, from looking into code and tests there seems to always have a validation if the `authResponse` is `nil` or not so I believe this should not impact any other component. 

#### Notes for Reviewers

This is motivated by the discussion that occurred in the #5512, where the usage of the `appID` in the DR was used solely as an authentication process. 

The reason for adding the `entity_access` mock in the `mockis` is due to the pending change in the `rpcmetadata.WithForwardedAuth`. In the issue there is a bigger discussion on the topic but the short end of the stick is, the method currently only validates if the fields are not empty, in another PR this will be changed to validate if the user is authenticated/has_any_rights and most likely tests will need to be updated to include a mock of the entity_access.

#### Checklist


pkg/devicerepository/grpc.go
```diff
@@ -60,13 +61,31 @@ func (dr *DeviceRepository) ensureBaseAssetURLs(models []*ttnpb.EndDeviceModel)
 
 const defaultLimit = 1000
 
-func requireApplication(ctx context.Context, ids *ttnpb.ApplicationIdentifiers) error {
-	return rights.RequireApplication(ctx, ids, ttnpb.Right_RIGHT_APPLICATION_DEVICES_READ)
+var errNotAuthenticated = errors.DefineUnauthenticated(
```
adriansmares: The name of the error (`errNotAuthenticated`) should be very close to the ID (`no_rights_in_context`). You probably want just `errUnauthenticated`/`unauthenticated`.

pkg/devicerepository/grpc.go
```diff
@@ -60,13 +61,31 @@ func (dr *DeviceRepository) ensureBaseAssetURLs(models []*ttnpb.EndDeviceModel)
...
+	"no_rights_in_context",
+	"Not authenticated, need to be logged in to make this request",
```
adriansmares: The error description should not start with upper case letter, and can be shorter: `unauthenticated`.

pkg/devicerepository/grpc.go
```diff
@@ -60,13 +61,31 @@ func (dr *DeviceRepository) ensureBaseAssetURLs(models []*ttnpb.EndDeviceModel)
...
+)
+
+func requireAuthentication(ctx context.Context) error {
+	log.FromContext(ctx).Debug("Validating request authentication")
+	authInfo, err := rights.AuthInfo(ctx)
+	if err != nil {
+		log.FromContext(ctx).WithError(err).Debug("Error validating user's auth")
```
adriansmares: `Failed to validate authentication information`.

pkg/devicerepository/grpc.go
```diff
@@ -60,13 +61,31 @@ func (dr *DeviceRepository) ensureBaseAssetURLs(models []*ttnpb.EndDeviceModel)
...
+		return err
+	}
+	if authInfo == nil {
+		log.FromContext(ctx).Debug("User not authenticated")
```
adriansmares: `No authentication response available`. 

Also see my comment below on why I'd prefer not to use `nil` but instead a different check.

pkg/identityserver/entity_access.go
```diff
@@ -75,7 +75,7 @@ func (is *IdentityServer) authInfo(ctx context.Context) (info *ttnpb.AuthInfoRes
 
 	md := rpcmetadata.FromIncomingContext(ctx)
 	if md.AuthType == "" {
-		return &ttnpb.AuthInfoResponse{}, nil
+		return nil, nil
```
adriansmares: I think that the `nil` here is not ideal. In general returning `nil` over gRPC is not encouraged.

I think that a user being authenticated means that either:
- `res.UniversalRights` is not empty OR
- `res.AccessMethod` is not `nil`

The empty message (that was used here before) would be compatible with this check (since `AccessMethod` would be `nil`).

johanstokking (commented):
Don't we need variants of https://github.com/TheThingsNetwork/lorawan-stack/pull/5659/files#diff-15fab987d2f4f47384e0f77d8848a1958a0271981ef1c25bedf9d1cb2340279dR388 that do not take the application ID?

nicholaspcr:
>Don't we need variants of https://github.com/TheThingsNetwork/lorawan-stack/pull/5659/files#diff-15fab987d2f4f47384e0f77d8848a1958a0271981ef1c25bedf9d1cb2340279dR388 that do not take the application ID?

Apologies on my end but the link just points to the PR diff for me, so its a bit unclear to me which section you are talking about.

adriansmares (approved):
I guess @johanstokking would like `devicerepository.requireAuthentication` to be promoted to `pkg/auth/rights` as `RequireAuthenticated` such that other components may use it too.

This LGTM otherwise.

CHANGELOG.md
```diff
@@ -187,6 +187,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Changed
 
 - Sortable tables are now sorted by default in the Console
+- Device Repository no longer uses the ApplicationID for validating requests, authentication is still necessary but the ApplicationID field has been deprecated in the DR api.
```
KrishnaIyer:
```
Device Repository no longer uses the `ApplicationID` for validating requests. Authentication is still necessary, but the `ApplicationID` field has been deprecated in the Device Repository API.
```

pkg/auth/rights/auth_info.go
```diff
@@ -47,3 +48,22 @@ func AuthInfo(ctx context.Context) (authInfo *ttnpb.AuthInfoResponse, err error)
 	}
 	return authInfo, nil
 }
+
+var errUnauthenticated = errors.DefineUnauthenticated("unauthenticated", "unauthenticated")
+
+// RequireAuthentication confirms if the authentication information within a context contains any rights, if so,
+// the user is considered to be authenticated.
+func RequireAuthentication(ctx context.Context) error {
+	log.FromContext(ctx).Debug("Validating request authentication")
```
KrishnaIyer: Thou shalt use imperative. Also you can simply log `Authenticate request`.

pkg/auth/rights/auth_info.go
```diff
@@ -47,3 +48,22 @@ func AuthInfo(ctx context.Context) (authInfo *ttnpb.AuthInfoResponse, err error)
...
+	authInfo, err := AuthInfo(ctx)
+	if err != nil {
+		log.FromContext(ctx).WithError(err).Debug("Failed to validate authentication information")
+		return err
+	}
+
+	if authInfo.GetAccessMethod() == nil && len(authInfo.GetUniversalRights().GetRights()) == 0 {
+		log.FromContext(ctx).Debug("User not authenticated")
```
KrishnaIyer: Is this really a user authentication? Wouldn't call authentication or request authentication be more generic?

Also since the error is being returned to the caller and there's no really ambiguity for the cause, I don't think this needs to be logged.

pkg/identityserver/mock/device_registry.go
```diff
@@ -25,28 +25,31 @@ import (
 type mockISEndDeviceRegistry struct {
 	ttnpb.EndDeviceRegistryServer
 
-	endDevicesMu sync.RWMutex
-	endDevices   map[string]*ttnpb.EndDevice
+	endDevices map[string]*ttnpb.EndDevice
+	sync.Mutex
```
KrishnaIyer: Any reason for not using `sync.Map`?

pkg/identityserver/mock/gateway_registry.go
```diff
@@ -56,13 +57,27 @@ type mockISGatewayRegistry struct {
 	gatewayRights map[string]authKeyToRights
 
 	registeredGateway *ttnpb.GatewayIdentifiers
+
+	sync.Mutex
```
KrishnaIyer: We generally don't prefer this syntax. Please define the mutex variable explicitly. Also in the other mocks.

KrishnaIyer (commented):
Implementation looks good. A few comments

pkg/auth/rights/auth_info.go
nicholaspcr: Regarding the question of if this is really an user authentication, not sure if what you meant is similar to what I brought it up on the issue [discussion](https://github.com/TheThingsNetwork/lorawan-stack/issues/5512#issuecomment-1200517714). If so I indeed agree, the reason for not changing it in this PR was it that the validation seems to be sufficient for the DR and it will be changed to be a tad bit more generic when working on `Updating the rpcmetadata.WithForwardedAuth`.

Or perhaps you are talking about the rights validation? If so I'm not too sure what you meant

pkg/identityserver/mock/gateway_registry.go
nicholaspcr: Made it be similar to how it is in other packages, `mu sync.Mutex`.

pkg/identityserver/mock/device_registry.go
nicholaspcr: Just didn't come to mind, I ended up making the change only on the device registry.

pkg/auth/rights/auth_info.go
nicholaspcr: I thank thee for noticing this mistake.

pkg/auth/rights/auth_info.go
```diff
@@ -47,3 +48,21 @@ func AuthInfo(ctx context.Context) (authInfo *ttnpb.AuthInfoResponse, err error)
...
+var errUnauthenticated = errors.DefineUnauthenticated("unauthenticated", "unauthenticated")
+
+// RequireAuthentication confirms if the authentication information within a context contains any rights, if so,
+// the user is considered to be authenticated.
+func RequireAuthentication(ctx context.Context) error {
+	log.FromContext(ctx).Debug("Authenticate request")
+	authInfo, err := AuthInfo(ctx)
+	if err != nil {
+		log.FromContext(ctx).WithError(err).Debug("Failed to validate authentication information")
+		return err
```
KrishnaIyer: Shouldn't this be `errUnauthenticated.WithCause(err)`?

pkg/auth/rights/auth_info.go
```diff
@@ -47,3 +48,21 @@ func AuthInfo(ctx context.Context) (authInfo *ttnpb.AuthInfoResponse, err error)
 	}
 	return authInfo, nil
 }
+
+var errUnauthenticated = errors.DefineUnauthenticated("unauthenticated", "unauthenticated")
+
+// RequireAuthentication confirms if the authentication information within a context contains any rights, if so,
+// the user is considered to be authenticated.
```
KrishnaIyer: My only question is the term "user"; are you authenticating a user? Is this call always on behalf of a user? 
Or can it be made more generic by saying `call authentication` or `request authentication`.

KrishnaIyer (commented):
Looks great. Final tiny comments.

pkg/auth/rights/auth_info.go
KrishnaIyer: See https://github.com/TheThingsNetwork/lorawan-stack/pull/5659#discussion_r936615531

pkg/auth/rights/auth_info.go
nicholaspcr: You are correct, It can be made a bit more generic.
Probably was a bit too focused on the interaction of importing the end devices via csv/json and ended up thinking the interaction only happens through a user.

KrishnaIyer (approved):

nicholaspcr:
Ignoring the lint error regarding the `identityserver/mock/unexported-return`

PR #5661, opened 2022-08-01
Fix empty keys regression

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/5657

#### Changes

- Verify if the key envelopes are empty, not if the unencrypted key is empty.

#### Testing

Traffic end to end tests.

##### Regressions

N/A

#### Checklist

### DIFF
--- a/pkg/joinserver/grpc_application_activation_settings_registry.go
+++ b/pkg/joinserver/grpc_application_activation_settings_registry.go
@@ -81,7 +81,7 @@ func (srv applicationActivationSettingsRegistryServer) Set(ctx context.Context, 
 	}
 
 	sets := req.FieldMask.GetPaths()
-	if ttnpb.HasAnyField(sets, "kek.key") && !types.MustAES128Key(reqKEK.GetKey()).OrZero().IsZero() {
+	if ttnpb.HasAnyField(sets, "kek.key") && len(reqKEK.GetKey()) > 0 {
 		kek, err := cryptoutil.WrapAES128Key(ctx, *types.MustAES128Key(reqKEK.Key), srv.kekLabel, srv.JS.KeyVault)
 		if err != nil {
 			return nil, errWrapKey.WithCause(err)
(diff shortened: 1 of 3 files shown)


pkg/joinserver/grpc_application_activation_settings_registry.go
```diff
@@ -81,7 +81,7 @@ func (srv applicationActivationSettingsRegistryServer) Set(ctx context.Context,
 	}
 
 	sets := req.FieldMask.GetPaths()
-	if ttnpb.HasAnyField(sets, "kek.key") && !types.MustAES128Key(reqKEK.GetKey()).OrZero().IsZero() {
+	if ttnpb.HasAnyField(sets, "kek.key") && len(reqKEK.GetKey()) > 0 {
```
adriansmares, line 84: Semantically this allows the unencrypted key to be all zeroes.

pkg/networkserver/downlink.go
```diff
@@ -504,7 +504,7 @@ func (ns *NetworkServer) generateDataDownlink(ctx context.Context, dev *ttnpb.En
 	ctx = log.NewContext(ctx, logger)
 
 	if len(cmdBuf) > 0 && (!cmdsInFOpts || macspec.EncryptFOpts(dev.MacState.LorawanVersion)) {
-		if types.MustAES128Key(dev.GetSession().GetKeys().GetNwkSEncKey().GetKey()).OrZero().IsZero() {
+		if dev.GetSession().GetKeys().GetNwkSEncKey() == nil {
```
adriansmares, line 507: So this check has a pretty funny history.

The old check was `len(dev.Session.GetKeys().GetSNwkSIntKey().GetKey()) == 0` which was always going to be false since `GetKey()` returns `*types.AES128Key`, which is always going to have length 16.

We then changed this to the `IsZero` variant, but this was wrong too, because the session keys are stored in their wrapped form at rest, so `GetKey()` will always return `nil` actually.

We now check that the envelope exists, and `Unwrap` will take care that `key` is filled (either just because it is already there, or by unwrapping).

--- 

This bug was caught only in the ABP tests, which makes me think that we actually do not store the session keys encrypted at rest for OTAA. From what I can see, we just take the key envelopes received from the JS and roll with them. Perhaps they should be wrapped too.

KrishnaIyer (approved):
LGTM. Don't see any obvious implementation issues.

johanstokking:
LGTM too

PR #5649, opened 2022-07-27
Support sticky MAC command answers

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/5608

#### Changes

- Add the `ns.mac.command.parse.fail` (failed to parse MAC command), `ns.mac.command.unknown` (unknown MAC command received), `ns.mac.command.process.fail` (failed to process MAC command), `ns.mac.command.unanswered` (MAC command unanswered by end device) events
- Lower the log level of some Network Server messages which are not actionable by deployment administrators
   - `Failed to read MAC command`
   - `Unknown MAC command received`
   - `MAC command buffer not fully answered`
   - `Failed to process MAC command after deduplication`
- Introduce the concept of 'recent MAC command identifiers'
   - It is a small `repeated` field which contains the MAC command identifiers for the sticky MAC commands.
   - In theory we could decrypt + decode the last uplink of `recent_uplinks`, but I consider that duplicating that code would be too much, and this slice will be empty most of the time (so 0 storage cost).
- For `RxParamSetup`, `RxTimingSetup`, `TxParamSetup` and `DlChannel`, avoid scheduling a request if the last uplink received from the end device contains a sticky response. 
  - Consider the following trace:
     a. Receive uplink message and generate a `DlChannelReq`.
     b. Receive uplink message with `DlChannelAns`.
     c. Receive uplink message with `DlChannelAns` (sticky).
     d. Receive uplink message with `DlChannelAns` (sticky) and generate a `DlChannelReq` (due to some changes in the desired parameters).
     e. Receive uplink message with `DlChannelAns`.
     At this point it is not possible to check if this `DlChannelAns` corresponds to the request in (a), or the one in (d).

#### Testing

Local testing + unit testing.

##### Regressions

The event related changes should not cause issues - they just lower the log levels and generate new events instead.
The recent MAC command identifiers are always reset after an uplink, and will not grow indefinitely (as expected).
The new sticky command request condition is covered by unit tests.

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
[...]

PR #5675, opened 2022-08-03
Fix MAC commands marshalling in events

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/5649

#### Changes

- Wrap `[]*ttnpb.MACCommands` in a message in order to facilitate marshaling
  - The full story is below.

#### Testing

UT.

##### Regressions

N/A.

#### Notes for Reviewers

Over the past couple of months we've come to believe that the Network Server generates quite a lot of events during uplink frame processing. The MAC command events (one per request, one per response) is a big source of those.

As such, I've experimented with using slices of MAC commands and 'compress' a bit the resulting events.
The catch is that event data should be either:
1. A proto message (`proto.Message`)
2. An error (preferably from `pkg/errors`)
3. A standard Go type.

A slice of proto messages is a bit of a mix between (1) and (3). This causes `pkg/events` to marshal it using `gogoproto.Value`, which does not play nicely with `oneof` fields.

As such, in the future if we want to have aggregates of messages (slices, maps) in event data, we should always have a wrapper type for them.

#### Checklist


api/lorawan.proto
```diff
@@ -739,6 +739,10 @@ message MACCommand {
   }
 }
 
+message MACCommands {
```
KrishnaIyer, line 742: Maybe add a note that this is used (currently) only for internal purposes.
(If we use it in the future as part of incoming request, we'll probably need to add a limit).

KrishnaIyer (approved):
LGTM

PR #5680, opened 2022-08-05
Fix cypress tests

ryaplots:
#### Summary

This PR fixes the failing end-to-end tests for the packer broker.

#### Changes

- Add correct intercepts

#### Testing

cypress

#### Notes for Reviewers

...

#### Checklist

### DIFF
--- a/cypress/integration/console/packet-broker/routing-policies.spec.js
+++ b/cypress/integration/console/packet-broker/routing-policies.spec.js
@@ -25,14 +25,14 @@ describe('Packet Broker routing policies', () => {
     cy.intercept('/api/v3/pba/forwarders/policies*', {
       fixture: 'console/packet-broker/policies-forwarder.json',
     })
+    cy.intercept('GET', '/api/v3/pba/home-networks/gateway-visibilities/default', {
+      statusCode: 404,
+    })
 
     cy.loginConsole({ user_id: 'admin', password: 'admin' })
   })
 
   it('succeeds setting a default routing policy', () => {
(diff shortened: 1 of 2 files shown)


adriansmares (commented):
FTR this is PR is about this flake: https://github.com/TheThingsNetwork/lorawan-stack/runs/7634188066?check_suite_focus=true

It seems to be caused by the fact that we do two requests in parallel here:
https://github.com/TheThingsNetwork/lorawan-stack/blob/ead0a3b40dc6d9bd3f2a524a72d473a34c4b6ccb/pkg/webui/console/views/admin-packet-broker/admin-packet-broker.js#L288-L294

If we don't intercept both calls in every test, there is a chance that the unintercepted call reaches the real PBA, which returns a `400` error. That 400 error is then rendered and causes the tests to fail.

cypress/integration/console/packet-broker/routing-policies.spec.js
```diff
@@ -87,6 +87,9 @@ describe('Packet Broker routing policies', () => {
   })
 
   it('succeeds setting individual per-network routing policy', () => {
+    cy.intercept('GET', '/api/v3/pba/home-networks/gateway-visibilities/default', {
+      statusCode: 404,
+    })
     cy.intercept('GET', '/api/v3/pba/home-networks/policies/default', {
```
kschiffer: Since all tests in this suite use this intercept, we can just add it do `.beforeEach()` on LL20.

kschiffer (approved):
Thanks! This was a regression from #5621.

PR #5686, opened 2022-08-09
Manually upgrade Google Chrome

adriansmares:
#### Summary

This PR changes the way we install Google Chrome to use `apt` instead of a separate GitHub action.


kschiffer (approved):
Nice, thank you

PR #5701, opened 2022-08-11
Expose band definitions via api

### LINKED ISSUE #5403: Expose band definitions via API
#### Summary

We should expose band definitions via a gRPC API such that clients (mainly the Console) can provide a better UX.

#### Why do we need this?

Currently MAC settings in the Console are very low level - users input data rate _indices_ , data rate offsets and raw frequencies. While this is fine, it is very error prone, as indices have different meanings between bands, even within the same band over different versions. As such, it would be better if the Console could show direct definitions using dropdowns instead of allowing free input. In order to be able to render such dropdowns, the Console needs a way to retrieve the band definition.

A frequency plans store exposed by all of the components.

A versioned band store exposed by all of the components.

#### Environment

v3.19.0

Add the API for the band definitions and expose the service. The idea for the band definition is that it does not replace `band.Band`, but just provides a serializable form similar to the ground truth testing that was used in previous `pkg/band` refactorings.

UT.

Yes. cc @kschiffer

ysmilda:
#### Summary

Closes #5403 

#### Changes

- Add rpc to list all bands with filtering on `BandID` and `PhyVersion`

#### Testing

Unit tests

##### Regressions

n/a

#### Checklist


api/configuration_services.proto
```diff
@@ -57,6 +58,89 @@ message GetPhyVersionsResponse {
...
+}
+
+message BandDescription {
+  string id = 1;
+
+  message Beacon {
+    DataRateIndex data_rate_index = 1;
+    string coding_rate = 2;
+    bool inverted_polarity = 3;
+    reserved 4; //ComputeFrequency
```
adriansmares: Space after `//`, and let's use the converted casing (so `compute_frequency` since that would be field name if we actually filled it). 

Same for the rest of the reserved fields.

api/configuration_services.proto
```diff
@@ -57,6 +58,89 @@ message GetPhyVersionsResponse {
   repeated VersionInfo version_info = 2;
 }
 
+message ListBandsRequest {
+  // Optional Band ID to filter the results.
+  // If unused, all supported Bands and their versions are returned.
+  string band_id = 1;
```
adriansmares, line 65: We should also provide a `PHYVersion` filter, with the same semantics (applied if present).

api/configuration_services.proto
```diff
@@ -57,6 +58,89 @@ message GetPhyVersionsResponse {
...
+  message DwellTime {
+    bool uplinks = 1;
+    bool downlinks = 2;
+  }
+
+  DwellTime boot_dwell_time = 28;
+}
+
+message ListBandsResponse {
+  repeated BandDescription bands = 1;
```
adriansmares: While the `ListFrequencyPlansResponse` indeed returns a repeated result, here we have two levels of indirection: first the band ID, then the `PHYVersion`.

Since in protobuf we cannot have map of map, let's wrap `map<PHYVersion, BandDescription>` into a message and then make `map<string, that-wrapper>` part of the response.

api/configuration_services.proto
```diff
@@ -70,4 +154,10 @@ service Configuration {
       get: "/configuration/phy-versions"
     };
   }
+
+  rpc ListBands(ListBandsRequest) returns (ListBandsResponse) {
+    option (google.api.http) = {
+      get: "/configuration/bands"
```
adriansmares, line 171: Also here we should have two additional bindings 

```
      additional_bindings {
        get: "/configuration/bands/{band_id}"
      }
      additional_bindings {
        get: "/configuration/bands/{band_id}/{phy_version}"
      }
```

api/configuration_services.proto
```diff
@@ -57,6 +58,89 @@ message GetPhyVersionsResponse {
...
+
+  message Rx2Parameters {
+    DataRateIndex data_rate_index = 1; 
+    uint64 frequency = 2;
+  }
+
+  Rx2Parameters default_rx2_parameters = 27;
+
+  message DwellTime {
+    bool uplinks = 1;
```
adriansmares: This actually needs to be `google.protobuf.BoolValue` since this can be `null`.

api/configuration_services.proto
```diff
@@ -57,6 +58,89 @@ message GetPhyVersionsResponse {
...
+
+  message Beacon {
+    DataRateIndex data_rate_index = 1;
+    string coding_rate = 2;
+    bool inverted_polarity = 3;
+    reserved 4; //ComputeFrequency
+  }
+
+  Beacon beacon = 2;
+  uint64 ping_slot_frequency = 3;
```
adriansmares: `google.protobuf.UInt64Value` since it is nullable.

api/configuration_services.proto
ysmilda: Would the same also apply to the `GetPhyVersions` rpc?

api/configuration_services.proto
ysmilda: Changed the `Wrapper` to be of the kind `map<string, BandDescription>` as there is no support for an `enum` as key in maps.

api/configuration_services.proto
adriansmares: Not for now.

api/configuration_services.proto
```diff
@@ -57,6 +59,96 @@ message GetPhyVersionsResponse {
   repeated VersionInfo version_info = 2;
 }
 
+message ListBandsRequest {
+  // Optional Band ID to filter the results.
+  // If unused, all supported Bands are returned.
+  string band_id = 1;
+  // Optional PHY version to filter the results.
+  // If unused, all supported versions are returned.
+  string phy_version = 2;
```
adriansmares: Can't this be the enum type `PHYVersion` ?

api/configuration_services.proto
```diff
@@ -57,6 +59,96 @@ message GetPhyVersionsResponse {
...
+  message DwellTime {
+    google.protobuf.BoolValue uplinks = 1;
+    google.protobuf.BoolValue downlinks = 2;
+  }
+
+  DwellTime boot_dwell_time = 28;
+}
+
+message ListBandsResponse {
+  message Wrapper {
```
adriansmares: Let's call `Wrapper` `VersionedBandDescriptions` instead.

api/configuration_services.proto
adriansmares: Not for now.

api/configuration_services.proto
```diff
@@ -57,6 +59,96 @@ message GetPhyVersionsResponse {
...
+    google.protobuf.BoolValue uplinks = 1;
+    google.protobuf.BoolValue downlinks = 2;
+  }
+
+  DwellTime boot_dwell_time = 28;
+}
+
+message ListBandsResponse {
+  message VersionedBandDescription {
+    // map<PHYVersion, BandDescription>
```
adriansmares: Drop the comment.

pkg/band/rpc.go
```diff
@@ -58,3 +59,138 @@ func GetPhyVersions(ctx context.Context, req *ttnpb.GetPhyVersionsRequest) (*ttn
...
+		versions, ok := All[req.BandId]
+		if !ok {
+			return nil, errBandNotFound.WithAttributes(req.BandId, req.PhyVersion.String())
+		}
+
+		filteredVersions[req.BandId] = versions
+	}
+
+	if len(filteredVersions) == 0 {
+		filteredVersions = All
```
adriansmares: Clone the map using `maps.Clone`, otherwise the `delete` and the assignment below will edit the global variable.

pkg/band/rpc.go
```diff
@@ -58,3 +59,138 @@ func GetPhyVersions(ctx context.Context, req *ttnpb.GetPhyVersionsRequest) (*ttn
...
+	}
+
+	if req.PhyVersion != 0 {
```
adriansmares: Nit: Compare with `ttnpb.PHYVersion_PHY_UNKNOWN`.

pkg/band/rpc.go
```diff
@@ -58,3 +59,138 @@ func GetPhyVersions(ctx context.Context, req *ttnpb.GetPhyVersionsRequest) (*ttn
...
+// convertToBandDescription parses a band into a ttnpb.BandDescription
+func (band Band) convertToBandDescription() *ttnpb.BandDescription {
+	bandDescription := &ttnpb.BandDescription{
+		Id: band.ID,
+		Beacon: &ttnpb.BandDescription_Beacon{
+			DataRateIndex:    band.Beacon.DataRateIndex,
+			CodingRate:       band.Beacon.CodingRate,
+			InvertedPolarity: band.Beacon.InvertedPolarity,
+		},
+		PingSlotFrequency:      &types.UInt64Value{},
```
adriansmares: Leave this `nil` here and initialize the full value below. Otherwise you will report a `0`/`false` when the real value is `null` / nonexistant.

pkg/band/rpc.go
```diff
@@ -58,3 +59,138 @@ func GetPhyVersions(ctx context.Context, req *ttnpb.GetPhyVersionsRequest) (*ttn
...
+		AdrAckLimit:            band.ADRAckLimit,
+		TxParamSetupReqSupport: band.TxParamSetupReqSupport,
+		DefaultMaxEirp:         band.DefaultMaxEIRP,
+		LoraCodingRate:         band.LoRaCodingRate,
+		DefaultRx2Parameters: &ttnpb.BandDescription_Rx2Parameters{
+			DataRateIndex: band.DefaultRx2Parameters.DataRateIndex,
+			Frequency:     band.DefaultRx2Parameters.Frequency,
+		},
+		BootDwellTime: &ttnpb.BandDescription_DwellTime{
+			Uplinks:   &types.BoolValue{},
```
adriansmares: Same comment here and for the downlinks.

pkg/band/rpc.go
ysmilda: Did not know about this behaviour, fixed it

pkg/band/rpc.go
```diff
@@ -58,3 +60,141 @@ func GetPhyVersions(ctx context.Context, req *ttnpb.GetPhyVersionsRequest) (*ttn
...
 }
+
+// ListBands returns the list of supported bands.
+func ListBands(ctx context.Context, req *ttnpb.ListBandsRequest) (*ttnpb.ListBandsResponse, error) {
+	filteredVersions := make(map[string]map[ttnpb.PHYVersion]Band)
+
+	if req.BandId != "" {
+		versions, ok := All[req.BandId]
+		if !ok {
+			return nil, errBandNotFound.WithAttributes(req.BandId, req.PhyVersion.String())
```
adriansmares: The `WithAttributes` expects key value pairings, so you want to use actually `("id", req.BandId, "version", req.PhyVersion)` here.

pkg/band/rpc.go
```diff
@@ -58,3 +60,145 @@ func GetPhyVersions(ctx context.Context, req *ttnpb.GetPhyVersionsRequest) (*ttn
...
+		}
+
+		res.Descriptions[bandID] = versionedBandDescription
+	}
+
+	return res, nil
+}
+
+// convertToBandDescription parses a band into a ttnpb.BandDescription
+func (band Band) convertToBandDescription() *ttnpb.BandDescription {
```
KrishnaIyer: `BandDescription` (the parsing/conversion is implied).

KrishnaIyer (commented):
Looks good to me. Didn't check all the details. Leaving that upto @adriansmares.

adriansmares (approved):

johanstokking (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Added
 
 - New `SearchAccounts` RPC on the `EntityRegistrySearch` service.
+- New `Bands` RPC on the `Configuration` service.
```
adriansmares: `ListBands`.

pkg/component/configuration_grpc.go
```diff
@@ -44,11 +44,11 @@ func (c *ConfigurationServer) RegisterServices(s *grpc.Server) {
 
 // RegisterHandlers registers the Configuration service handler.
 func (c *ConfigurationServer) RegisterHandlers(s *runtime.ServeMux, conn *grpc.ClientConn) {
-	ttnpb.RegisterConfigurationHandler(c.component.Context(), s, conn)
+	_ = ttnpb.RegisterConfigurationHandler(c.component.Context(), s, conn)
 }
 
 // ListFrequencyPlans implements the Configuration service's ListFrequencyPlans RPC.
-func (c *ConfigurationServer) ListFrequencyPlans(ctx context.Context, req *ttnpb.ListFrequencyPlansRequest) (*ttnpb.ListFrequencyPlansResponse, error) {
+func (c *ConfigurationServer) ListFrequencyPlans(ctx context.Context, req *ttnpb.ListFrequencyPlansRequest) (*ttnpb.ListFrequencyPlansResponse, error) { //nolint:lll
```
adriansmares: Do not use `nolint:lll` here and below.

When the signatures get too big, we prefer to use the following rule set:
1. Give the parameters their own individual line, like so:
```
func (c *ConfigurationServer) ListFrequencyPlans(
	ctx context.Context, req *ttnpb.ListFrequencyPlansRequest,
) (*ttnpb.ListFrequencyPlansResponse, error) {
```
2. If this is still too long, give each parameter a new line:
```
func (c *ConfigurationServer) ListFrequencyPlans(
	ctx context.Context,
	req *ttnpb.ListFrequencyPlansRequest,
) (*ttnpb.ListFrequencyPlansResponse, error) {
```

Same rules for return values.

ysmilda: Tried that but found it not helping the readability. But if that's the standard I'll gladly follow it

PR #5706, opened 2022-08-15
Update device repository service's bindings

nicholaspcr:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/5659#pullrequestreview-1056571153
Which was misinterpreted by me in a earlier PR, the idea is since the applicationID was deprecated in the Device Repository, we should update the services URLs while still accepting the old ones to maintain compatibility.

#### Changes

- Device Repository service methods have been updated to remove `{application_ids.application_id}`

#### Testing

- Ran the TTS locally to see if the console still loaded models when creating an end-device.
- Did some curls to see if the paths returns `unauthorized` (expected response), instead of `Not implemented`.

##### Regressions

This is a fix, small one in fact. This effectively should only broaden the accepted paths by the DR service.

#### Checklist


johanstokking (approved):
LGTM but wait for final blessing from @adriansmares

adriansmares (approved):
👼

PR #5715, opened 2022-08-18
Put coding rate in band's data rate

### LINKED ISSUE #4466: Put coding rate in band's data rate

ysmilda:
#### Summary

Closes #4466 

#### Changes

- Define `coding_rate` in `LoRaDataRate` instead of `Band` definition
- Remove `coding_rate` from `TxSettings` as it is now defined in `DataRate`

#### Testing

UT

#### Checklist


CHANGELOG.md
```diff
@@ -19,10 +19,12 @@ For details about compatibility between different releases, see the **Commitment
 ### Changed
 
 - Gateway registration in the Console has been updated to simplify the onboarding experience.
+- LoRa coding rate now defined in `DataRate` instead of `Band`
 
 ### Deprecated
 
 ### Removed
+- Removed coding rate from `TxSettings` as it is now defined in `DataRate`
```
adriansmares: New line after `Removed`, and dot at the end of the sentences.

api/lorawan.proto
```diff
@@ -446,7 +447,7 @@ message TxSettings {
   DataRate data_rate = 1 [(validate.rules).message.required = true];
   reserved 2; // data_rate_index
   // LoRa coding rate.
```
adriansmares: Remove the old field comment.

pkg/gatewayserver/io/mqtt/format_protobufv2.go
```diff
@@ -60,7 +60,7 @@ func (protobufv2) FromDownlink(down *ttnpb.DownlinkMessage, _ *ttnpb.GatewayIden
 	switch dr := settings.DataRate.Modulation.(type) {
 	case *ttnpb.DataRate_Lora:
 		lorawan.Modulation = ttnpbv2.Modulation_LORA
-		lorawan.CodingRate = settings.CodingRate
+		lorawan.CodingRate = settings.DataRate.GetLora().CodingRate
```
adriansmares: You can use `dr.Lora.CodingRate`, right ?

pkg/gatewayserver/io/mqtt/format_protobufv2.go
```diff
@@ -112,7 +112,7 @@ func (protobufv2) ToUplink(message []byte, ids *ttnpb.GatewayIdentifiers) (*ttnp
 			return nil, err
 		}
 		settings.DataRate = loraDr.DataRate
-		settings.CodingRate = lorawanMetadata.CodingRate
+		settings.DataRate.Modulation.(*ttnpb.DataRate_Lora).Lora.CodingRate = lorawanMetadata.CodingRate
```
adriansmares, line 115: I would move this inside the switch, just to make sure that `settings.DataRate.Modulation` really has type `*ttnpb.DataRate_Lora` (and you can edit `loraDr` directly).

pkg/ttnpb/udp/translation.go
```diff
@@ -220,8 +220,10 @@ func convertUplink(rx RxPacket, md UpstreamMetadata) (*ttnpb.UplinkMessage, erro
 	}
 
 	switch up.Settings.DataRate.Modulation.(type) {
```
adriansmares: You can do `mod := up.Setings...` in the switch and then you can edit the coding rate via `mod.Lora.CodingRate` / `mod.Lrfhss.CodingRate`.

pkg/ttnpb/udp/translation.go
```diff
@@ -411,7 +409,6 @@ func FromDownlinkMessage(msg *ttnpb.DownlinkMessage) (*TxPacket, error) {
 	tx.DatR.DataRate = scheduled.DataRate
 	switch scheduled.DataRate.GetModulation().(type) {
 	case *ttnpb.DataRate_Lora:
```
adriansmares, line 422: This is a conversion function to the UDP format. Do not remove the `tx.CodR` assignment, but instead use the same `mod := ...` in the switch and fill it with `mod.Lora.CodingRate`.

pkg/gatewayserver/io/mqtt/format_protobufv2.go
ysmilda: If I'm not mistaken it is already positioned within the switch statement, so the type is set at that point. As `loraDr` is also of type `DataRate` we would still need to imply a type on it to edit it. 

Might be a nicer solution to adjust the `ParseLoRa` function to also take a coding rate. But this would cause an issue with the `UnmarshallJSON` function which has no information about the coding rate.

CHANGELOG.md
```diff
@@ -33,6 +33,11 @@ For details about compatibility between different releases, see the **Commitment
 ### Changed
 
 - Gateway registration in the Console has been updated to simplify the onboarding experience.
+- LoRa coding rate now defined in `DataRate` instead of `Band`.
```
adriansmares: Make sure that this ends up under `Unreleased`, not `3.21.1`.

pkg/packetbrokeragent/translation.go
```diff
@@ -57,7 +57,7 @@ var toPBRegion = map[string]packetbroker.Region{
 	band.ISM_2400:        packetbroker.Region_WW_2G4,
 }
 
-func fromPBDataRate(dataRate *packetbroker.DataRate) (dr *ttnpb.DataRate, codingRate string, ok bool) {
+func fromPBDataRate(dataRate *packetbroker.DataRate) (dr *ttnpb.DataRate, ok bool) {
 	switch mod := dataRate.GetModulation().(type) {
 	case *packetbroker.DataRate_Lora:
 		// TODO: Set coding rate from data rate (https://github.com/TheThingsNetwork/lorawan-stack/issues/4466)
```
johanstokking: This comment can be removed right?

johanstokking (commented):
Great stuff.

Shouldn't this go in 3.22 as it's an API change; the LoRa coding rate can now only be retrieved through the data rate and not as part of the `settings` field (in uplink metadata)?

@adriansmares `(pkg/band.Band).FindUplinkDataRate` uses `proto.Equal` for finding the uplink data rate based on equality. This PR implicitly includes the coding rate as part of the predicate. A potential problem is that only since RP2 1.0.0, there's normative text to use coding rate 4/5 for uplink for all bands. In RP1, it was only defined for US915. This means that technically, end devices using something other than 4/5 (like 4/6 or 4/7, if any) will now get unmatched uplink channel errors. Please confirm. If so, we should first analyze in our observed data (recent uplinks?) to which extent end devices use something other than 4/5 to measure the impact of this change. If that is neglible, we can accept these changes as-is, otherwise we may have to make an exception in `FindUplinkDataRate()` for LoRa modulation. For safety, we can do that anyway, by applying a field mask or something elegant.

adriansmares:
> Shouldn't this go in 3.22 as it's an API change; the LoRa coding rate can now only be retrieved through the data rate and not as part of the `settings` field (in uplink metadata)?

Yes. I've created `v3.22` and made this PR target it.

> @adriansmares `(pkg/band.Band).FindUplinkDataRate` uses `proto.Equal` for finding the uplink data rate based on equality. This PR implicitly includes the coding rate as part of the predicate. A potential problem is that only since RP2 1.0.0, there's normative text to use coding rate 4/5 for uplink for all bands. In RP1, it was only defined for US915. This means that technically, end devices using something other than 4/5 (like 4/6 or 4/7, if any) will now get unmatched uplink channel errors. Please confirm. If so, we should first analyze in our observed data (recent uplinks?) to which extent end devices use something other than 4/5 to measure the impact of this change. If that is neglible, we can accept these changes as-is, otherwise we may have to make an exception in `FindUplinkDataRate()` for LoRa modulation. For safety, we can do that anyway, by applying a field mask or something elegant.

The expectation is correct - the NS will fail to find the data rate used by the uplink. 

It is an small part of the traffic on the gateway layer (~0.5% of traffic, per NOC), and at application layer (virtually 0, but not 0, per Storage Integration). See the data [here](https://docs.google.com/document/d/1iHZnKlw5yrewco9kVhOK1EFCKmmnV3w8tEYbX51vUWg/edit). The presence of empty values at application layer means that there are some gateways out there that for some reason don't fill the coding rate, but the applications still expect these uplinks.

I think that for RP001 variants we can reasonably leave out the data rate out of the comparison, except for the `US915` variants. For RP002 I think that we need a field in the band that should specify that the coding rate comparisons are strict (`StrictCodingRate`), and based on that indeed `FindUplinkDataRate` should either take the coding rate into account, or not.

ysmilda:
Changed out `proto.Equal` for a dedicated compare function and added a `StrictCodingRate` boolean to the band definitions based on the corresponding `LoRaWan` specifications. This should mitigate the problem at hand.

pkg/band/band.go
```diff
@@ -154,11 +155,54 @@ func (b Band) FindDownlinkDataRate(dr *ttnpb.DataRate) (ttnpb.DataRateIndex, Dat
...
+
+// compareDatarates checks if two given *ttnpb.DataRate's are equal. If the strict
+// bit is not set the *ttnpb.DataRate_Lora check will skip the `CodingRate`.
+// Make sure to update this check when altering any of the `DataRate` types.
+func compareDatarates(a, b *ttnpb.DataRate, strict bool) bool {
+	switch modA := a.Modulation.(type) {
+	case *ttnpb.DataRate_Lora:
+		switch modB := b.Modulation.(type) {
+		case *ttnpb.DataRate_Lora:
+			if modA.Lora.Bandwidth != modB.Lora.Bandwidth ||
```
adriansmares: You can turn these into singular return statements by negating them and removing the `if`.

ysmilda: Done, unfortunately the linter is not letting me remove the `default` cases as then it starts complaining about that it could have been an `if` statement even though that is not the case in this instance.

adriansmares: Please also apply the negation to the expression itself - do not just wrap the old `if` with it. It becomes hard to parse for other developers if we have too many symbols there, and it is also error prone.

ysmilda: My bad, negated the logic and that cleans it up quite nicely.

adriansmares (approved):
The changes look good to me (the ones which I did not commit myself).

@KrishnaIyer please take a look at the last two commits.

pkg/band/band_test.go
```diff
@@ -678,3 +679,35 @@ func TestBeacon(t *testing.T) {
 		})
 	}
 }
+
+func TestStrictCodingRateSanityCheck(t *testing.T) {
+	t.Parallel()
+	for band_id, versions := range All {
+		for version, band := range versions {
+			band_id := band_id
```
adriansmares: You can use `bandId, version, band := bandId, version, band`.

KrishnaIyer (approved):
Didn't check `pkg/band` but other code-owned files LGTM.

PR #5720, opened 2022-08-19
Remove GatewayAntenna attributes

### LINKED ISSUE #5614: Support (or remove) GatewayAntenna attributes
### Summary

The GatewayStore in the Identity Server should support attributes on GatewayAntennas.

Or, if we decide not to do that, we should remove the `attributes` field from the API.

### Current Situation

The GatewayStore implementations in the Identity Server do not handle the attributes field of gateway antennas.

It's in the API, so it's confusing that the database doesn't support it.

### Proposed Implementation

Add a JSONB field to the model (in hindsight I think the `attributes` table that we use for other entities was perhaps not the best solution) and store the attributes in there.

### Contributing

### Code of Conduct

ysmilda:
#### Summary

Closes #5614 

#### Changes

- Removed `attributes` from `GatewayAntenna` definition.

#### Testing

Unit testing

#### Checklist


api/gateway.proto
```diff
@@ -337,13 +337,7 @@ message GatewayAntenna {
...
   // location is the antenna's location.
   Location location = 2;
-  map<string,string> attributes = 3 [
-    (validate.rules).map = {
-      max_pairs: 10,
-      keys: { string: { pattern: "^[a-z0-9](?:[-]?[a-z0-9]){2,}$", max_len: 36 } },
-      values: { string: { max_len: 200 } }
-    }
-  ];
+  reserved 3; // attributes
```
nicholaspcr: I don't think it can be dropped like this, the problem before was that it is supported by the API and not by store. Since this would change the API, considering the commitments on the `README.md`, this sort of change would be considered a breaking change on the API since it no longer accepts the field itself.

>We will not break the API towards gateways and applications within the major version. This includes how gateways communicate (with Gateway Server) and how applications work with data (with Application Server)

It might be best to do something like use the `deprecated` option. Like in [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/cca5e6bd0dad0220cd7b487f60449e273823afe9/api/user.proto#L70-L72), that way the field is still accepted but not used in the internal logic.

nicholaspcr (commented):
Should probably remove: https://github.com/TheThingsNetwork/lorawan-stack/blob/8d824e3c51befd5c8e8c4b616575d0cbde875de9/pkg/identityserver/bunstore/gateway_antenna_store.go#L40-L41

Since its no longer supported.

api/gateway.proto
ysmilda: Okay good to know! Changed it to match

kschiffer (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Removed
 
+- Removed `attributes` from `GatewayAntenna` definition.
```
adriansmares: You can move under `Changed` now, since we basically only deprecated it. You may also add a small explanation to this change and mention that the underlying store never stored the values here.

sdk/js/generated/api-definition.json
```diff
@@ -812,14 +812,14 @@
...
         "up.uplink_message.settings.concentrator_timestamp",
         "up.uplink_message.settings.data_rate",
         "up.uplink_message.settings.data_rate.modulation",
         "up.uplink_message.settings.data_rate.modulation.fsk",
         "up.uplink_message.settings.data_rate.modulation.fsk.bit_rate",
         "up.uplink_message.settings.data_rate.modulation.lora",
         "up.uplink_message.settings.data_rate.modulation.lora.bandwidth",
+        "up.uplink_message.settings.data_rate.modulation.lora.coding_rate",
         "up.uplink_message.settings.data_rate.modulation.lora.spreading_factor",
         "up.uplink_message.settings.data_rate.modulation.lrfhss",
```
johanstokking: This file contains unrelated changes right?

Please see my review comment in #5715; this may break API

pkg/identityserver/bunstore/gateway_antenna_store.go
```diff
@@ -37,17 +37,13 @@ type GatewayAntenna struct {
 
 	Location
 
-	// TODO: Add attributes as JSONB field (https://github.com/TheThingsNetwork/lorawan-stack/issues/5614).
```
adriansmares, line 40: Probably we should remove them from the `GatewayAntenna` model of `gormstore` too.

adriansmares (approved):

pkg/identityserver/bunstore/gateway_antenna_store.go
ysmilda: The `gormstore` `GatewayAntenna` model has no mention of the `attributes`. My assumption is that it was added with the switch to `bunstore` as a preparation step.

sdk/js/generated/api-definition.json
ysmilda: This was indeed unrelated, when regenerating via `tools/bin/mage jsSDK:definitions` there seems to be some caching which caused this to remain. Running `tools/bin/mage jsSDK:definitionsClean jsSDK:definitions` was necessary to fix this.

nicholaspcr (approved):
Looks great!

pkg/identityserver/bunstore/gateway_antenna_store.go
adriansmares: https://github.com/TheThingsNetwork/lorawan-stack/blob/22db57b2fe35ee8f93281b9f6a9a737f97954288/pkg/identityserver/gormstore/gateway_antenna.go#L19-L27

ysmilda: I see, in this PR I had already removed them; that's why I couldn't find it in my local version

johanstokking (approved):

nicholaspcr:
I'm not sure if there is a strict rule against merging the "master" branch into your branch instead or rebasing it, so I'm asking this more out of curiosity since I skimmed through the `DEVELOPMENT.MD` and couldn't find it. @adriansmares @KrishnaIyer  Is there a pattern or both are ok?

adriansmares:
> I'm not sure if there is a strict rule against merging the "master" branch into your branch instead or rebasing it, so I'm asking this more out of curiosity since I skimmed through the `DEVELOPMENT.MD` and couldn't find it. @adriansmares @KrishnaIyer Is there a pattern or both are ok?

I prefer to rebase in order to avoid the extra merge commit, but we don't explicitly deny the merge option.

ysmilda:
> I prefer to rebase in order to avoid the extra merge commit, but we don't explicitly deny the merge option.

Will keep that in mind for the next time, made use of GitHub's build in merge conflict solver here.

Issue #3886, opened 2021-03-04
Add authorization management to the Account App


kschiffer:
#### Summary
The Account App should have functionality that lets users manage their current OAuth authorizations. See also #1422

#### Why do we need this?
Users will want to review and revoke authorizations for security reasons.

#### What is already there? What do you see now?
- Account App was introduced with 3.11.0
- API for OAuth authorizations

#### What is missing? What do you want to see?
- Account app with authorization management UI 
  - OAuth client authorizations (listing, reviewing, revoking)
  - OAuth access tokens (listing, reviewing, revoking)

#### How do you propose to implement this?
Need to think about how to structure this properly, UI-wise.

#### How do you propose to test this?
End-to-end test.

#### Can you do this yourself and submit a Pull Request?
Yes.

Issue #3885, opened 2021-03-04
Add session management to Account App


kschiffer:
#### Summary
The Account App should allow to manage the currently active sessions of the user.

#### Why do we need this?
It is security relevant for users to review and delete active sessions and hence managing resource access.

#### What is already there? What do you see now?
- The Account App has been introduced with 3.11.0. As of now it still lacks session management functionality.
- Session management API
- There's an option to revoke active sessions when using the "forgot password flow"

#### What is missing? What do you want to see?
The Account App providing functionality to manage sessions (including reviewing details and deleting sessions)

#### How do you propose to implement this?
#1422 has some basic wireframes. Implementation shouldn't be a big deal, as it's mostly plumbing.

#### How do you propose to test this?
This should be tested by end-to-end tests

#### Can you do this yourself and submit a Pull Request?
Yes.

PR #5725, opened 2022-08-23
Fix pagination in session and tokens tables

ryaplots:
#### Summary

Fix pagination in sessions table and tokens table (these views were added in #3886  and #3885 )

#### Changes

- Add `params` to request

#### Testing

Manually

##### Regressions

-

#### Checklist


pkg/webui/account/containers/tokens-table/tokens-table.js
```diff
@@ -162,7 +164,7 @@ const TokensTable = props => {
   )
 
   const getItems = React.useCallback(
-    () => getAuthorizationTokensList(userId, clientId, []),
+    filters => getAuthorizationTokensList(userId, clientId, { ...filters }),
```
kschiffer: Could you also update the names to refer to access tokens instead of authorization tokens? It's kind of confusing otherwise.

kschiffer (changes requested):
Some more things:
- I saw that the pagination for Authorizations is also not setup. Please add that as well.
- Please also add a table title for the access token table `Access tokens`.
- Please remove all `handlesSorting` props from the fech tables. `handlesSorting` only makes sense for unpaginated data. Please also double check sorting while doing that. You can [refer to the protos](https://github.com/TheThingsIndustries/lorawan-stack/blob/v3.21/api/oauth.proto#L52) to check on which fields sorting is allowed and then add `sortable: true` to those.

ryaplots:
Let me fix the tests first. They were passing locally so I have to see what's up here

pkg/webui/account/containers/authorizations-table/index.js
```diff
@@ -76,7 +76,7 @@ const OAuthClientAuthorizationsTable = props => {
     [],
   )
 
-  const getItems = React.useCallback(() => getAuthorizationsList(userId, []), [userId])
+  const getItems = React.useCallback(filters => getAuthorizationsList(userId, ...filters), [userId])
```
kschiffer:
```suggestion
  const getItems = React.useCallback(filters => getAuthorizationsList(userId, filters), [userId])
```
You cannot spread an object into function arguments. Pagination actions expect the `filters`-object as second argument.

pkg/webui/account/containers/sessions-table/index.js
```diff
@@ -54,7 +54,10 @@ const UserSessionsTable = props => {
   const { pageSize, user, handleDeleteSession, sessionId } = props
   const dispatch = useDispatch()
 
-  const getSessions = React.useCallback(filters => getUserSessionsList(filters, user), [user])
+  const getSessions = React.useCallback(
+    filters => getUserSessionsList(user, { ...filters }),
```
kschiffer:
```suggestion
    filters => getUserSessionsList(user, filters),
```
Why do you clone the object again via spread?

pkg/webui/account/containers/tokens-table/connect.js
```diff
@@ -44,9 +44,9 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => ({
   ...stateProps,
   ...dispatchProps,
   ...ownProps,
-  deleteToken: id => dispatchProps.deleteToken(stateProps.userId, stateProps.clientId, id),
+  deleteToken: id => dispatchProps.deleteToken([stateProps.userId, stateProps.clientId], id),
```
kschiffer: I don't see the need for this change. See also my other comment below.

pkg/webui/account/store/middleware/authorizations.js
```diff
@@ -34,47 +35,48 @@ const getAuthorizationsLogic = createRequestLogic({
...
-const deleteAuthorizationTokenLogic = createRequestLogic({
-  type: authorizations.DELETE_AUTHORIZATION_TOKEN,
+const deleteAccessTokenLogic = createRequestLogic({
+  type: authorizations.DELETE_ACCESS_TOKEN,
   process: async ({ action }) => {
-    const { userId, clientId, id } = action.payload
+    const { routeParams, id } = action.payload
 
-    return await tts.Authorizations.deleteToken(userId, clientId, id)
+    return await tts.Authorizations.deleteToken(...routeParams, id)
```
kschiffer: I'd prefer to not use spread here, since it only adds an unnecessary abstraction.

Generally, spreading arguments makes sense when working with a dynamic number of arguments. Otherwise, it only makes it harder to follow what arguments are being passed.

pkg/webui/account/containers/tokens-table/tokens-table.js
```diff
@@ -162,7 +165,7 @@ const TokensTable = props => {
   )
 
   const getItems = React.useCallback(
-    () => getAuthorizationTokensList(userId, clientId, []),
+    filters => getAccessTokensList(userId, clientId, { ...filters }),
```
kschiffer:
```suggestion
    filters => getAccessTokensList(userId, clientId, filters),
```

sdk/js/src/service/authorizations.js
```diff
@@ -23,10 +23,13 @@ class Authorizations {
...
-  async getAllAuthorizations(userId) {
-    const result = await this._api.OAuthAuthorizationRegistry.List({
-      routeParams: { 'user_ids.user_id': userId },
-    })
+  async getAllAuthorizations(userId, params) {
+    const result = await this._api.OAuthAuthorizationRegistry.List(
+      {
+        routeParams: { 'user_ids.user_id': userId },
+      },
+      { ...params },
```
kschiffer:
```suggestion
      params,
```

sdk/js/src/service/authorizations.js
```diff
@@ -39,10 +42,13 @@ class Authorizations {
...
-  async getAllTokens(userId, client_id) {
-    const result = await this._api.OAuthAuthorizationRegistry.ListTokens({
-      routeParams: { 'user_ids.user_id': userId, 'client_ids.client_id': client_id },
-    })
+  async getAllTokens(userId, client_id, params) {
+    const result = await this._api.OAuthAuthorizationRegistry.ListTokens(
+      {
+        routeParams: { 'user_ids.user_id': userId, 'client_ids.client_id': client_id },
+      },
+      { ...params },
```
kschiffer:
```suggestion
      params,
```

sdk/js/src/service/sessions.js
```diff
@@ -23,10 +23,13 @@ class Sessions {
...
-  async getAllSessions(userId) {
-    const result = await this._api.UserSessionRegistry.List({
-      routeParams: { 'user_ids.user_id': userId },
-    })
+  async getAllSessions(userId, params) {
+    const result = await this._api.UserSessionRegistry.List(
+      {
+        routeParams: { 'user_ids.user_id': userId },
+      },
+      { ...params },
```
kschiffer:
```suggestion
      params,
```
Please make sure you understand when spreading objects is required. Typically in two situations:
1. You want to compose a new object consisting of data from multiple objects 
2. You want to shallow clone an object (usually to avoid the original object from being modified)

kschiffer (changes requested):

pkg/webui/account/containers/tokens-table/connect.js
ryaplots: Currently, in the global-pagination actions, there are defined delete actions that take pagination into consideration, but none of them allows several params as arguments. There are only these: [https://github.com/TheThingsNetwork/lorawan-stack/blob/4f42706237e34063a5cdcd60b0e19a24bbf67c8e/pkg/webui/lib/store/actions/pagination.js#L61-L72](https://github.com/TheThingsNetwork/lorawan-stack/blob/4f42706237e34063a5cdcd60b0e19a24bbf67c8e/pkg/webui/lib/store/actions/pagination.js#L61-L72 ) But for deleting a token I need more than `id` and/or `targetId`. So I decided to create a generic `pagination delete` action that would take any number of route params in a single array and the id of the item to be deleted. This is why here I put the `userId` and the `cliendId` in an array, that will be treated by that action and later spread in the request in the middleware and used as route parameters in the actual request. I could also use an object and do like: `{userId: stateProps.userId, clientId: stateProps.clientId}` and avoid spreading and using the array, this may be better to avoid making it harder to follow what arguments are being passed

pkg/webui/account/containers/tokens-table/tokens-table.js
```diff
@@ -38,6 +41,7 @@ import {
 } from '@account/store/selectors/authorizations'
 
 const m = defineMessages({
+  tableTitle: 'Access Tokens',
```
kschiffer:
```suggestion
  accessTokens: 'Access tokens',
```
[No title case](url).

pkg/webui/account/containers/tokens-table/tokens-table.js
```diff
@@ -52,8 +56,8 @@ const TokensTable = props => {
     props
 
   useBreadcrumbs(
-    'client-authorizations.single.tokens',
-    <Breadcrumb path={`/client-authorizations/${clientId}/tokens`} content={'tokens'} />,
+    'client-authorizations.single.access-tokens',
+    <Breadcrumb path={`/client-authorizations/${clientId}/access-tokens`} content={'tokens'} />,
```
kschiffer:
```suggestion
    <Breadcrumb path={`/client-authorizations/${clientId}/access-tokens`} content={m.accessTokens} />,
```

pkg/webui/account/containers/tokens-table/tokens-table.js
kschiffer: Please also check other messages, I see more title case being used in:
`OAuth Clients` -> `OAuth clients`
`OAuth Client Authorizations` -> `OAuth client authorizations`

pkg/webui/account/store/actions/sessions.js
```diff
@@ -29,7 +29,7 @@ export const [
     success: getUserSessionsListSuccess,
     failure: getUserSessionsListFailure,
   },
-] = createPaginationByParentRequestActions('USER_SESSIONS')
+] = createPaginationByIdRequestActions('USER_SESSIONS')
```
kschiffer: You don't need the pagination actions here. You can use the regular `createRequestActions`.

kschiffer (changes requested):
Can you change one more thing please:
image
image

These should use the `name` (`Console`) instead of the ID.

pkg/webui/account/views/oauth-client-authorization-overview/index.js
```diff
@@ -57,6 +61,9 @@ const AuthorizationOverview = props => {
     siteName,
   } = props
 
+  const client = useSelector(state => selectClientById(state, clientId))
+  const clientName = client?.name || capitalize(clientId)
```
kschiffer: Don't use capitalize. The ID should stay as it is if it is used.

kschiffer (approved):

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Fixed
 
+- Pagination in `sessions` and `access tokens` tables in the Console.
```
adriansmares: Please make sure this goes into the `v3.22` section of the changelog.

PR #5727, opened 2022-08-23
Miscellaneous LR-FHSS fixes

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/5704
References https://github.com/TheThingsNetwork/lorawan-stack/pull/5181

#### Changes

- Fix the handling of block `LinkADRAns` MAC command answers
  - Previously if we had a command buffer that looked like `[LinkADRAns, LinkADRAns, MiscCommand, OtherCommand]` the `MiscCommand` would be skipped
  - This has been broken for [4 years](https://github.com/TheThingsNetwork/lorawan-stack/commit/e2e3c2c5c9f4f16a4d63b708fc5e11dc1a3ffd48#diff-57380479952dc4fff38b87e5ee1c741504e82125eb9dbbe1551dc0dd627ec01cR30)
- The `operating_channel_width` of the LR-FHSS data rates has been updated to be used in Hz, not KHz.
  - It is already in Hz in the Packet Broker API and in `pkg/band`, but the UDP translation layer filled it in as if it was in KHz
- The Network Server can now generate `LinkADRReq` requests with data rates which are not LoRa based
  - Combined with the static ADR mode, this can be used to switch the end device between modulations
- Enable LR-FHSS modulation in Packet Broker Agent
- Fix Console rendering of non-LoRa modulations
- Show `decoded_payload` of application downlinks when available

#### Testing

Unit testing + device testing for the block `LinkADRAns` handling.
Device testing with the `v5.0.0-branch` release of LoRaMac-node.

For improved UX, LR-FHSS will require a (draft) frequency plan, but we can handle that in a separate PR. For integration testing I patch the `mac_state.desired_parameters.channels` and then command the end device to use LR-FHSS using the static ADR API.

##### Regressions

Testing already covers most of the changes, and the block MAC command handling has a new test suite that should ensure that we don't regress in the future.

#### Checklist

### DIFF
--- a/pkg/ttnpb/udp/packet_data.go
+++ b/pkg/ttnpb/udp/packet_data.go
@@ -53,11 +53,10 @@ type RxPacket struct {
 	Aesk  uint         `json:"aesk"`            // AES key index used for encrypting fine timestamps (unsigned integer)
 }
 
-// reduceCodingRate reduces the coding rate fraction returned by the packet forwarder.
-// The packet forwarder is concentrator centric - and the concentrator expresses coding
-// rates as a fraction of form `4/(4 + x)`. But the RP002-1.0.x series of documents uses
-// the irreducible form of the fraction.
-func reduceCodingRate(cr string) string {
+// reduceLRFHSSCodingRate reduces the coding rate fraction returned by the packet forwarder.
+// The packet forwarder will render the coding rates used by LR-FHSS in their `4/x` form, even
+// though the real coding rates are irreducible fractions.
+func reduceLRFHSSCodingRate(cr string) string {
 	switch cr {
 	case "4/6":
 		return "2/3"
@@ -83,9 +82,7 @@ func (p *RxPacket) UnmarshalJSON(data []byte) error {
 	}
 	switch mod := p.DatR.DataRate.Modulation.(type) {
 	case *ttnpb.DataRate_Lrfhss:
-		// The normalization should be applied only for LR-FHSS modulations
-		// since they are defined in the RP002 as irreducible fractions.
-		mod.Lrfhss.CodingRate = reduceCodingRate(p.CodR)
+		mod.Lrfhss.CodingRate = reduceLRFHSSCodingRate(p.CodR)
 	}
 	return nil
 }
--- a/pkg/ttnpb/udp/packet_data.go
+++ b/pkg/ttnpb/udp/packet_data.go
@@ -128,6 +128,11 @@ type Stat struct {
 	FPGA *uint32       `json:"fpga,omitempty"` // Version of Gateway FPGA (unsigned integer)
 	DSP  *uint32       `json:"dsp,omitempty"`  // Version of Gateway DSP software (unsigned interger)
 	HAL  *string       `json:"hal,omitempty"`  // Version of Gateway driver (format X.X.X)
+	HVer *struct {
+		FPGA *uint32 `json:"fpga,omitempty"` // Version of FPGA (unsigned integer)
+		DSP0 *uint32 `json:"dsp0,omitempty"` // Version of DSP 0 software (unsigned interger)
+		DSP1 *uint32 `json:"dsp1,omitempty"` // Version of DSP 1 software (unsigned interger)
+	} `json:"hver,omitempty"` // Gateway hardware versions
 }
 
 // TxError is returned in the TxPacketAck
--- a/pkg/ttnpb/udp/packet_data.go
+++ b/pkg/ttnpb/udp/packet_data.go
@@ -53,6 +53,23 @@ type RxPacket struct {
 	Aesk  uint         `json:"aesk"`            // AES key index used for encrypting fine timestamps (unsigned integer)
 }
 
+// reduceCodingRate reduces the coding rate fraction returned by the packet forwarder.
+// The packet forwarder is concentrator centric - and the concentrator expresses coding
+// rates as a fraction of form `4/(4 + x)`. But the RP002-1.0.x series of documents uses
+// the irreducible form of the fraction.
+func reduceCodingRate(cr string) string {
+	switch cr {
+	case "4/6":
+		return "2/3"
+	case "4/8":
+		return "1/2"
+	case "2/6":
+		return "1/3"
+	default:
+		return cr
+	}
+}
+
 // UnmarshalJSON implements json.Unmarshaler.
 func (p *RxPacket) UnmarshalJSON(data []byte) error {
 	type Alias RxPacket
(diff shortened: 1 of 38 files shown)


pkg/webui/console/components/events/previews/gateway-uplink-message.js
```diff
@@ -46,22 +46,8 @@ const GatewayUplinkMessagePreview = React.memo(({ event }) => {
...
-        rssi = data.message.rx_metadata[0].rssi
-      }
-    }
-
-    if ('settings' in data.message && 'data_rate' in data.message.settings) {
-      const bandwidth = getByPath(data, 'message.settings.data_rate.lora.bandwidth')
-      const spreadingFactor = getByPath(data, 'message.settings.data_rate.lora.spreading_factor')
-      dataRate = `SF${spreadingFactor}BW${bandwidth / 1000}`
-    }
+    ;({ snr, rssi } = getSignalInformation(data.message))
```
kschiffer:
```suggestion
    const { snr, rssi } = getSignalInformation(data.message)
```

kschiffer (changes requested):

pkg/webui/console/components/events/previews/gateway-uplink-message.js
adriansmares: Are you sure this is correct ? The variables are created on L31, and declaring the constants inside the scope of the `if` on L33 does not make them available when they are used on L66/67.

image

kschiffer: You're right indeed. This deconstruction into existing variables is a very uncommon sight but ultimately legitimate. I would write it as:
```js
const signalInfo = getSignalInformation(data.message)
snr = signalInfo.snr
rssi = signalinfo.rssi
```
to avoid the notation which also requires a semicolon.

johanstokking (dismissed):
Beautiful

kschiffer (approved):

pkg/ttnpb/udp/packet_data.go
```diff
@@ -53,6 +53,20 @@ type RxPacket struct {
...
+	case "4/6":
+		return "2/3"
+	case "4/8":
+		return "1/2"
+	case "2/6":
+		return "1/3"
+	default:
+		return cr
+	}
+}
```
johanstokking: If this is generally speaking always the same, why is `4/8` even a notation?

Note that this breaks TOA if you don't incorporate this in `pkg/toa`.

adriansmares: Settled offline - the reasoning behind this `4/x` notation has been added to the function documentation.

pkg/ttnpb/udp/packet_data.go
```diff
@@ -53,6 +53,23 @@ type RxPacket struct {
...
```
johanstokking: Isn't this `4/(4 + X)` notation specific to LoRa modulation?

The coding rate has the useful bits on the numerator and the total bits on the denominator. So `4/(4 + X)` means that the redundant bits are `X`. My interpretation is that in LoRa modulation, the useful bits are always a multiple of 4, and so is the nominator. For example, the coding rate is `4/8` and there are 9 useful bits to send. `4/8` would make it explicit that `ceil(9/4)*8 = 24` bits are sent, while with `1/2` it would be `9*2 = 18` bits. But I might be wrong here.

To this function name; this is very generic and seems to apply only to LR-FHSS.

pkg/ttnpb/udp/packet_data.go
```diff
@@ -128,6 +147,11 @@ type Stat struct {
 	FPGA *uint32       `json:"fpga,omitempty"` // Version of Gateway FPGA (unsigned integer)
 	DSP  *uint32       `json:"dsp,omitempty"`  // Version of Gateway DSP software (unsigned interger)
 	HAL  *string       `json:"hal,omitempty"`  // Version of Gateway driver (format X.X.X)
+	HVer *struct {
+		FPGA *uint32 `json:"fpga,omitempty"` // Version of FPGA (unsigned integer)
+		DSP0 *uint32 `json:"dsp0,omitempty"` // Version of DSP 0 software (unsigned interger)
+		DSP1 *uint32 `json:"dsp1,omitempty"` // Version of DSP 1 software (unsigned interger)
+	} `json:"hver,omitempty"` // Gateway hardware versions
```
johanstokking:
```suggestion
	DSP  *uint32       `json:"dsp,omitempty"`  // Version of Gateway DSP software (unsigned integer)
	HAL  *string       `json:"hal,omitempty"`  // Version of Gateway driver (format X.X.X)
	HVer *struct {
		FPGA *uint32 `json:"fpga,omitempty"` // Version of FPGA (unsigned integer)
		DSP0 *uint32 `json:"dsp0,omitempty"` // Version of DSP 0 software (unsigned integer)
		DSP1 *uint32 `json:"dsp1,omitempty"` // Version of DSP 1 software (unsigned integer)
	} `json:"hver,omitempty"` // Gateway hardware versions
```

Where does `hver` come from, where is it specified?

I seem to remember there were fields very specific to Kerlink gateways, which we may not want to include here. If these are (still proprietary) generic UDP packet forwarder fields, then I'm okay adding them.

It's just that any mistakes here, because some (future) packet forwarders use a different data type or anything, will result in unmarshaling errors. They're not worth it because nobody really uses this data I believe.

pkg/ttnpb/udp/packet_data.go
adriansmares: I've added some new details in [this](https://github.com/TheThingsIndustries/lorawan-stack/issues/3400#issuecomment-1227693736) issue, given that the packet forwarder is not public yet. 

TLDR is that I think you are right, this is an packet forwarder quirk related to how data rates are rendered, and seems to apply only to LR-FHSS.

pkg/ttnpb/udp/packet_data.go
adriansmares: The Kerlink Common Packet Forwarder will indeed fill the top level `dsp` and `hal` fields. You can find their definition of the fields [here](https://github.com/TheThingsIndustries/lorawan-stack/issues/3400#issuecomment-1227698944). It is indeed the case that the public source code of `packet_forwarder` [does not include them](https://github.com/Lora-net/packet_forwarder/blob/d0226eae6e7b6bbaec6117d0d2372bf17819c438/PROTOCOL.TXT#L214-L244).

The `hver` field is indeed part of the still proprietary [packet forwarder statistics message definition](https://github.com/TheThingsIndustries/lorawan-stack/issues/3400#issuecomment-1227700752).

Regarding the use of the data - it depends. We do have some custom logic to save the custom data for TTKGs and LBS gateways in the gateway attributes. This information in turn is used to check what firmware versions are running in the wild. We can use this information (by parsing the status message) if needed.

johanstokking: Okay, well if it's worth unmarshaling errors because of other gateways that use the same fields but with different types, I'm fine with it. Let's do add metrics though for UDP unmarshaling errors so at least we can see how much traffic we drop.

adriansmares: https://github.com/TheThingsNetwork/lorawan-stack/pull/5727/commits/0257604af8920b0d0479f3709432a30e308a8cea should do it. It will also record the field which has a type mismatch and the observed value in a minimal fashion - see [the docs](https://pkg.go.dev/encoding/json#UnmarshalTypeError).

johanstokking (approved):

PR #5730, opened 2022-08-23
Support payload normalization

### LINKED ISSUE #5429: Support for `normalized_payload` field
#### Summary

The new field `normalized_payload` will allow integrations to rely on a standardized data model for all devices in the device repository that support this new normalization step, regardless of the manufacturer, the units, and the wording they use, etc.

#### Why do we need this?

This feature request has already been largely discussed in https://github.com/TheThingsNetwork/lorawan-devices/issues/395

However, defining the data model right it's a different issue and does not block the implementation of this one (https://github.com/TheThingsNetwork/lorawan-devices/issues/395#issuecomment-1115023964), so this can be done before getting “the perfect standardized format”, which is a difficult task to get right, specially for less common quantities.

The implementation in `lorawan-stack` also does not depend on each specific device supporting the new `normalizeUplink(data)` function. This step should be completely optional (though encouraged for manufacturers).

There is no support for a standard data format.

Support for a new field in uplinks with a normalized format.

Has already been described in the previous referenced issue. I can't think of the specifics of `lorawan-stack` itself, since I'm not familiar with the codebase.

I suppose unit testing is sufficient, but I'm not certain.

Maybe, I'm willing to contribute to get this done as swiftly as I can, but I'll need guidance on how to proceed since I don't know the internals of the system in detail, I may miss important stuff and waste a lot of time.

Let me bring @johanstokking into discussion to see if he can assign this to someone appropriate or give me assistance to achieve this.

johanstokking:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/5429
References https://github.com/TheThingsNetwork/lorawan-devices/issues/395

#### Changes

- This adds `normalized_payload` to upstream messages
- Also includes initial parsing of normalized payload per the draft schema (see referenced Device Repository PR)

#### Testing

Unit testing and local integration testing

##### Regressions

None expected

#### Notes for Reviewers

Skip the first commit; all lint warning removals are squashed in there

This includes some commits from https://github.com/TheThingsNetwork/lorawan-stack/pull/5727 to show non-LoRa data rates in the event preview

@kschiffer please review `console:` commit only

We can probably do improve, but this is already pretty cool:

Screen Shot 2022-08-27 at 16 13 09

#### Checklist


api/messages.proto
```diff
@@ -204,6 +204,14 @@ message ApplicationUplink {
   // Warnings generated by the message processor while decoding the frm_payload.
   repeated string decoded_payload_warnings = 12;
 
+  // The normalized frame payload of the uplink message.
+  // This field is set by the message processor that is configured for the end device (see formatters) or application (see default_formatters).
+  // If the message processor is custom script and the decoded payload output is a valid normalized payload structure,
+  // this field is set to the decoded payload.
```
pablojimpas: So if the `decodeUplink` function returns a result that it's already normalized, the `normalizeUplink` function should not run at all even if it exits? And both fields (`decoded_payload` and `normalized_payload`) will be equal? Where is this logic implemented?

johanstokking: > So if the `decodeUplink` function returns a result that it's already normalized, the `normalizeUplink` function should not run at all even if it exits?

If it exists, it will run. The output is either the output from the normalizer, or from the decoder if that output is already normalized. I'll see if I can clarify that in this comment.

> And both fields (`decoded_payload` and `normalized_payload`) will be equal?

Yes

> Where is this logic implemented?

In the Application Server.

pkg/messageprocessors/javascript/javascript.go
```diff
@@ -59,45 +61,72 @@ var (
...
 	}
-	return h.encodeDownlink(ctx, ids, version, msg, run)
+	return h.encodeDownlink(ctx, msg, run)
 }
 
-func (h *host) encodeDownlink(ctx context.Context, ids *ttnpb.EndDeviceIdentifiers, version *ttnpb.EndDeviceVersionIdentifiers, msg *ttnpb.ApplicationDownlink, run func(context.Context, string, ...interface{}) (func(interface{}) error, error)) error {
+func (*host) encodeDownlink(
+	ctx context.Context,
+	msg *ttnpb.ApplicationDownlink,
+	run func(context.Context, string, ...interface{},
```
adriansmares: Nit - Do you need the new line for the return values ? Can we do the parameters of `run` in the same style as we do in other places (so the list goes on its own line).

adriansmares (commented):
I like the direction in which this is going.

Now that the uplink payload formatter parameter encapsulates two functions at once, should we consider increasing the character limit, or do we keep the one we have so far ?

pkg/messageprocessors/javascript/javascript.go
johanstokking: No I wasn't paying attention here

johanstokking:
> Now that the uplink payload formatter parameter encapsulates two functions at once, should we consider increasing the character limit, or do we keep the one we have so far ?

Let's not do that until we run into problems. I think we're already quite generous in the Device Repository with 64KB.

Inline codecs in The Things Stack have a lower limit, but since we support detecting whether the payload is already normalized, what is really needed is that codec developers just return normalized payload going forward.

pkg/messageprocessors/javascript/javascript.go
```diff
@@ -205,21 +278,72 @@ func (h *host) decodeUplink(ctx context.Context, ids *ttnpb.EndDeviceIdentifiers
...
+		// If the normalizer is not set, the decoder may return already normalized payload.
+		// This is a best effort attempt to parse the decoded payload as normalized payload.
+		// If that does not return an error, the decoded payload is assumed to be normalized.
+		normalizedPayload := []*pbtypes.Struct{
+			decodedPayload,
+		}
+		_, err := normalizedpayload.Parse(normalizedPayload)
+		if err == nil {
+			msg.NormalizedPayload, msg.NormalizedPayloadWarnings = normalizedPayload, nil
+		}
```
johanstokking, line 345: @adriansmares you expressed concerns with a performance hit of parsing each and every decoded uplink through the normalized payload validator (https://github.com/TheThingsNetwork/lorawan-stack/issues/5429#issuecomment-1196598347) (1/3)

pkg/messageprocessors/normalizedpayload/uplink.go
```diff
@@ -0,0 +1,235 @@
...
+	return res, nil
+}
+
+func parse(dst *Measurement, src *pbtypes.Struct, prefix string) error {
+	for k, v := range src.GetFields() {
+		path := fmt.Sprintf("%s%s", prefix, k)
+		parser, ok := fieldParsers[path]
+		if !ok {
+			return errUnknownField.WithAttributes("path", path)
+		}
```
johanstokking, line 252: What basically happens is that because [`additionalProperties` aren't allowed](https://github.com/TheThingsNetwork/lorawan-devices/pull/508/files#diff-9fcde326d127f74194f70e563bdf2c118c51b719c308f015b8eb0204a9a552fbR296), if any field is present that is not a known normalized field, there's an immediate error (2/3)

pkg/messageprocessors/normalizedpayload/uplink.go
```diff
@@ -0,0 +1,235 @@
...
+	return func(dst *Measurement, src *pbtypes.Value, path string) error {
+		val, ok := src.Kind.(*pbtypes.Value_NumberValue)
+		if !ok {
+			return errFieldType.WithAttributes("path", path)
+		}
+		n := val.NumberValue
+		if err := validate(n, vals, path); err != nil {
+			return err
+		}
+		*selector(dst) = &n
```
johanstokking, line 124: If the field exists, there's an early exit on a mismatched type. Only if the field name and type are normalized, the actual validations are run. That is not gonna be more than bounds checks, maybe regex, maybe enums.

What we could do, is first check if the structure is normalized, before validating the fields. That would make things a bit faster. In any case, this all happens in locally so I expect the performance impact is negligble. (3/3)

adriansmares (approved):

pkg/applicationserver/applicationserver.go
```diff
@@ -145,6 +141,11 @@ func New(c *component.Component, conf *Config) (as *ApplicationServer, err error
 		interopClient: interopCl,
 		interopID:     conf.Interop.ID,
 	}
+	as.formatters = messageprocessors.MapPayloadProcessor{
+		ttnpb.PayloadFormatter_FORMATTER_JAVASCRIPT: javascript.New(),
+		ttnpb.PayloadFormatter_FORMATTER_CAYENNELPP: cayennelpp.New(),
+		ttnpb.PayloadFormatter_FORMATTER_REPOSITORY: devicerepository.New(as.formatters, as),
```
pablojimpas: Is this line correct here instead of after initializing `as.formatters` as it was before? `as.formatters` is an empty map at this point when it's passed to `devicerepository.New`, right?

pkg/messageprocessors/javascript/javascript.go
```diff
@@ -205,21 +278,72 @@ func (h *host) decodeUplink(ctx context.Context, ids *ttnpb.EndDeviceIdentifiers
...
+
+	if normalized := output.Normalized; normalized != nil {
+		if errs := normalized.Errors; len(errs) > 0 {
+			return errOutputErrors.WithAttributes("errors", strings.Join(errs, ", "))
+		}
+		if normalized.Data == nil {
+			return errOutput.New()
+		}
+		// The returned data can be an array of measurements or a single measurement object.
+		var ms []map[string]interface{}
```
pablojimpas: Pardon my nitpicking, but I'm curious, why do you prefer a short variable name for this one that has a lifespan of about 20 lines? It seems more readable to me to use `measurments` here and change `measurment` to `m` or something like that, since the later ones have a shorter scope.

pkg/applicationserver/applicationserver.go
johanstokking: Great catch!

johanstokking:
@adriansmares this is ready for another full pass

You can skip the first commit; all lint warning removals are squashed in there, no functional changes otherwise

johanstokking:
@adriansmares strangely enough everything works, except there's no `as.up.data.forward` event published with the normalized payload. The message is being published on webhooks and MQTT integrations, but when I'm attached to the event stream (CLI or Console), it doesn't show. Any idea?

My suspicion is that the message is a duplicate of the uplink in the (Redis) pub/sub or that I'm still missing a typecheck somewhere.

pkg/messageprocessors/normalizedpayload/uplink.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// parseTime parses and validates the time. The input value must be RFC3339.
+func parseTime(selector func(dst *Measurement) **time.Time, vals ...fieldValidator[time.Time]) fieldParser {
+	return func(dst *Measurement, src *pbtypes.Value, path string) error {
+		val, ok := src.Kind.(*pbtypes.Value_StringValue)
+		if !ok {
+			return errFieldType.WithAttributes("path", path)
+		}
+		t, err := time.Parse(time.RFC3339, val.StringValue)
```
adriansmares: Maybe it should be `RFC3339Nano` ? The parsing behaves equivalently to `RFC3339` when there is no fractional part, but the fractional part itself may be specified (for whatever reason) and won't generate a parsing error compared to `RFC3339`.

This is how `encoding/json` works, for reference.

pkg/messageprocessors/normalizedpayload/uplink.go
```diff
@@ -117,12 +139,68 @@ func minimum[T constraints.Ordered](min T) fieldValidator[T] {
...
 	"time": parseTime(func(dst *Measurement) **time.Time { return &dst.Time }),
 	"air":  object(func(dst *Measurement) **Air { return &dst.Air }),
 	"air.temperature": parseNumber(func(dst *Measurement) **float64 { return &dst.Air.Temperature },
 		minimum(-273.15),
 	),
+	"air.relativeHumidity": parseNumber(func(dst *Measurement) **float64 { return &dst.Air.RelativeHumidity },
+		minimum(0.0),
+		maximum(1.0),
+	),
+	"air.pressure": parseNumber(func(dst *Measurement) **float64 { return &dst.Air.Pressure },
```
adriansmares: Schema PR related: We call this air here and in the schema, but shouldn't we call this actually environment ?

Think of a sensor that measures the temperature outside of a factory, vs one that measures the temperature/ RH / pressure of a container / chamber - the validation probably differs between the two, and if the container has for some reasons air in it, it becomes a bit weird.

pkg/applicationserver/applicationserver.go
```diff
@@ -1134,27 +1174,27 @@ func (as *ApplicationServer) handleUplink(
...
+		lastSeenAt: info.uplink.ReceivedAt,
 	}); err != nil {
 		log.FromContext(ctx).WithError(err).Warn("Failed to publish last seen event")
 	}
 
 	return nil
 }
 
-func (as *ApplicationServer) handleSimulatedUplink(ctx context.Context, ids *ttnpb.EndDeviceIdentifiers, uplink *ttnpb.ApplicationUplink, link *ttnpb.ApplicationLink) error {
+func (as *ApplicationServer) handleSimulatedUplink(ctx context.Context, item uplinkInfo) error {
```
adriansmares: Nit: `info` instead of `item`, since that is what you used above.

adriansmares (commented):
Webhook template store schema (the YAML files that describe the templates) require an update with the new message type too. See https://github.com/TheThingsNetwork/lorawan-stack/blob/6db9affc2812a12e982256080d19a60cf16874a3/pkg/applicationserver/io/web/templates.go#L225-L236

pkg/messageprocessors/normalizedpayload/uplink.go
johanstokking: Environment is also relative to the sensor, just like outside a temperature sensor can be in the shade or in the sun.

This positioning and hierarchy is to be done upstream. The sensor is an air temperature sensor, which could be placed on a ship or in a container; the same device model. We're not intending to make the firmware aware whether it's indoor or outdoor. In the digital twin of that ship, the sensors should be modeled correctly.

adriansmares (approved):

pkg/webui/console/components/events/previews/application-uplink-normalized.js
```diff
@@ -0,0 +1,58 @@
...
+const ApplicationUplinkNormalizedPreview = React.memo(({ event }) => {
+  const { data, identifiers } = event
+  const deviceIds = identifiers[0].device_ids
+  const { snr, rssi } = getSignalInformation(data)
+  const dataRate = getDataRate(data)
+
+  return (
+    <DescriptionList>
+      <DescriptionList.Byte title={messages.devAddr} data={deviceIds.dev_addr} />
+      {data.normalized_payload.air && (
```
kschiffer, line 36: Note that this will break if the data structure does not contain `normalized_payload`. If it's safe to assume that this will never be the case then that's fine. Otherwise, you might want to use `data?.normalized_payload?.air`. Same below.

kschiffer (approved):

pkg/webui/console/components/events/previews/application-uplink-normalized.js
johanstokking: This event type will always contain a non-nil `normalized_payload`.

PR #5731, opened 2022-08-24
Cypress tests for repository registration

ryaplots:
#### Summary

Add repository cypress tests.

#### Changes

- Update tests
- Update schemas
- Fix some rendering conditions

#### Testing

e2e tests

#### Checklist

### DIFF
--- a/cypress/integration/console/devices/create-repository-registration.spec.js
+++ b/cypress/integration/console/devices/create-repository-registration.spec.js
@@ -138,6 +138,7 @@ describe('End device repository manual registration', () => {
             ).should('be.visible')
             cy.findByRole('button', { name: /Change input method/ }).click()
           })
+        cy.findByLabelText('Enter end device specifics manually').should('be.checked')
       })
 
       it('succeeds hiding provisioning information on JoinEUI reset', () => {
--- a/cypress/integration/console/devices/create-repository-registration.spec.js
+++ b/cypress/integration/console/devices/create-repository-registration.spec.js
@@ -12,12 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {
-  generateHexValue,
-  disableNetworkServer,
-  disableApplicationServer,
-  disableJoinServer,
-} from '../../../support/utils'
+import { generateHexValue } from '../../../support/utils'
 
 describe('End device repository manual registration', () => {
   const user = {
@@ -145,7 +140,7 @@ describe('End device repository manual registration', () => {
           })
       })
 
-      it('succeeds hiding provisioing information on JoinEUI reset', () => {
+      it('succeeds hiding provisioning information on JoinEUI reset', () => {
         cy.findByLabelText('End device brand').selectOption('test-brand-otaa')
         cy.findByLabelText('Model').selectOption('test-model2')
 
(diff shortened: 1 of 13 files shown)


pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/device-registration-form-section/index.js
```diff
@@ -61,10 +62,14 @@ const DeviceRegistrationFormSection = () => {
     ((isABP || isMulticast) && values._inputMethod === 'manual') ||
     (!isOTAA && values._inputMethod === 'device-repository')
 
+  const jsEnabled = selectJsConfig().enabled
```
kschiffer: The form does not support any disabled components, so this should not be added. We have to add a top layer checks that disables the form altogether if any component is disabled.
If an old test spec was testing unusual stack configs we can remove that bit from the test.

kschiffer (changes requested):
I'm a bit confused with this diff. ~~It does not add any tests, only removes a test spec and changes implementation. Did something go wrong here?~~
EDIT: Sorry overlooked the collapsed diff.

pkg/webui/console/containers/device-onboarding-form/device-provisioning-form-section/device-registration-form-section/index.js
ryaplots: Will update this

cypress/integration/console/devices/create-repository-registration.spec.js
```diff
@@ -154,49 +124,81 @@ describe.skip('End device repository create', () => {
         cy.findByTestId('device-registration').should('not.exist')
       })
 
+      it('succeeds shwoing modal on device type change', () => {
```
kschiffer:
```suggestion
      it('succeeds showing modal on device type change', () => {
```

cypress/integration/console/devices/create-repository-registration.spec.js
```diff
@@ -154,49 +124,81 @@ describe.skip('End device repository create', () => {
...
+        cy.findByRole('button', { name: 'Add end device' }).click()
 
         cy.location('pathname').should(
           'eq',
-          `${Cypress.config('consoleRootPath')}/applications/${appId}/devices/${devId}`,
+          `${Cypress.config(
+            'consoleRootPath',
+          )}/applications/${appId}/devices/eui-${devEui.toLocaleLowerCase()}`,
         )
         cy.findByTestId('full-error-view').should('not.exist')
```
kschiffer: You have to wait here until the page was actually loaded before checking whether an error is in the page.
```suggestion
        cy.findByRole('heading', { name: `eui-${devEui}` }).should('be.visible')
        cy.findByTestId('full-error-view').should('not.exist')
```

Please update other tests to do the same.

cypress/integration/console/devices/create-repository-registration.spec.js
```diff
@@ -209,27 +211,33 @@ describe.skip('End device repository create', () => {
...
-        const devId = 'test-model3-device'
+        const devEui = generateHexValue(16)
 
         // End device selection.
-        cy.findByLabelText('Brand').selectOption('test-brand-otaa')
+        cy.findByLabelText('End device brand').selectOption('test-brand-otaa')
         cy.findByLabelText('Model').selectOption('test-model3')
         cy.findByLabelText('Hardware Ver.').selectOption('2.0')
         cy.findByLabelText('Firmware Ver.').selectOption('1.0.1')
         cy.findByLabelText('Profile (Region)').selectOption('EU_863_870')
```
kschiffer: Can we reuse the `selectDevice()` here?

cypress/integration/console/devices/create-repository-registration.spec.js
```diff
@@ -601,7 +532,7 @@ describe.skip('End device repository create', () => {
         cy.findByLabelText('End device ID').type(devId2)
         cy.findByLabelText('View registered end device').check()
 
-        cy.findByRole('button', { name: 'Register end device' }).click()
+        cy.findByRole('button', { name: 'Add end device' }).click()
```
kschiffer: I think we should actually try doing this three times. Rish was notifying me about an issue there where it wouldn't work after the third time. I guess it has to do with how the values are reset after the first time.

kschiffer (changes requested):
Very nice! I think we should add another test device profile that runs on LW/PHY 1.1.0, since that is currently still a bit of a blank spot. To add that, you need to add another fixture in `cypress/fixtures/console/devices/repository` as well as update `cypress/fixtures/console/devices/repository/test-brand-otaa.models.json`

Also, let's structure the new end device tests like this:
```
cypress/integration/console/devices/onboarding/repository-registration.spec.js
cypress/integration/console/devices/onboarding/repository-claiming.spec.js
cypress/integration/console/devices/onboarding/manual-registration.spec.js
cypress/integration/console/devices/onboarding/manual-claiming.spec.js
```

ryaplots:
About the claiming parts - claiming is not active on TTOS, which means I cannot write and run tests related to claiming. I can make a branch on tti to include the progress on the device onboarding so far and write the e2e tests there.

pkg/webui/console/containers/device-onboarding-form/index.js
```diff
@@ -126,10 +126,9 @@ const DeviceOnboardingForm = () => {
             errors: {},
             values: {
               ...values,
-              ...initialValues,
```
kschiffer, line 129: Did this break the multi-onboarding? Any idea why?

ryaplots: It was overwriting the template values to empty initial values, if I recall correctly.

kschiffer: Oh wait. This obviously makes sense, since we want to retain all configs except the IDs.

kschiffer (approved):

PR #5732, opened 2022-08-24
Cypress tests for manual registration

ryaplots:
#### Summary

Add cypress tests for manual registration

#### Changes

- Update UI
- Remove stackConfig-related test cases

#### Testing

e2e

#### Checklist


cypress/integration/console/devices/create-manual.spec.js
```diff
@@ -44,36 +39,39 @@ describe.skip('End device manual create', () => {
...
+      cy.findByLabelText('Frequency plan').should('be.visible')
+      cy.findByLabelText('LoRaWAN version').should('be.visible')
+      cy.findByLabelText('Regional Parameters version').should('be.visible')
+
+      cy.findByText(
+        'Please enter versions and frequency plan information above to continue',
+      ).should('be.visible')
+    })
+
+    it('validates empty form before submitting', () => {
```
kschiffer, line 59: Let's add an additional test here that checks whether missing MAC Settings are also validated. So that would mean unchecking the `Network defaults` checkbox and removing a value from a required field, submitting, and then checking whether the validation message appears.

cypress/integration/console/devices/create-manual.spec.js
```diff
@@ -390,12 +363,14 @@ describe.skip('End device manual create', () => {
         cy.findByLabelText('Network defaults').uncheck()
```
kschiffer:
```suggestion
        cy.findByLabelText('Network defaults').should('not.be.checked').and('have.attr', 'disabled')
```
The form will now actually automatically uncheck this when there is a required field within.

cypress/integration/console/devices/create-manual.spec.js
```diff
@@ -541,12 +523,14 @@ describe.skip('End device manual create', () => {
         cy.findByLabelText('Define multicast group (ABP & Multicast)').check()
         cy.findByLabelText('LoRaWAN class for multicast downlinks').selectOption('class-b')
```
kschiffer, line 700: Please add a check here that `cy.findByLabelText('Network defaults').should('not.be.checked').and('have.attr', 'disabled')`

pkg/webui/console/containers/device-onboarding-form/device-type-form-section/device-type-manual-form-section/validation-schema.js
```diff
@@ -108,7 +108,7 @@ const macSettingsSchema = Yup.object({
           class_c_timeout: Yup.string(),
         }
 
-        // Each MAC setting that does have a corresponding default setting is required.
+        // Each MAC setting that does not have a corresponding default setting is required.
```
kschiffer:
```suggestion
        // Each MAC setting that does have a corresponding default setting is required.
```
Sorry for the confusion, but this was actually right. What I mean here is that any value that has a default should not be cleared, since it will be prefilled with the default and I want to make the settings as explicit as possible. Note that this does not mean that all fields that do not have a default setting are automatically not required. Eg. the ping slot periodicity has no default value but needs to be provided explicitly by the user and these exceptions are defined explicitly in the validation schema.

kschiffer (changes requested):
Happy with this already but let's make it more complete by:
- Adding checks for the `Generator` buttons. I'm fine with integrating this into one of the existing tests to avoid too much duplication.
- For each created device, wait until the device page was loaded and check against the values in the device overview. It's fine to only check against the distinct values that the test focused on.
- Add a test that checks whether MAC settings are validated as expected when they are not provided
- Check that creating multiple devices in a row works as expected (using the `After registration` option)
- A check that retroactively resetting the FP, LW or PHY will reset subsequent fields.
- Add at least one simple test for each missing LW version `1.0.1`, `1.0.3`, and `1.0.4`

cypress/integration/console/devices/onboarding/manual-registration.spec.js
```diff
@@ -82,10 +86,49 @@ describe.skip('End device manual create', () => {
...
+      cy.findByLabelText('AppKey').type(device.app_key)
+
+      cy.findByLabelText('Frequency plan').selectOption(device.frequency_plan_id_2)
+
+      cy.findByLabelText('JoinEUI').should('not.exist')
+      cy.findByLabelText('DevEUI').should('not.exist')
+      cy.findByLabelText('AppKey').should('not.exist')
+    })
+
+    it('succeeds generation keys by clicking on `generate` button', () => {
```
kschiffer:
```suggestion
    it('succeeds generating keys by clicking on `Generate` button', () => {
```

kschiffer (approved):

PR #5737, opened 2022-08-28
Synchronize gateway time only with GPS timestamps

adriansmares:
#### Summary

References https://github.com/Lora-net/LoRaMac-node/issues/1348

This PR fixes some GS scheduling issues which I've encountered while re-testing class B support.

#### Changes

- Sync gateway time only with timestamps which are guaranteed to come from a GPS source.
  - This ensures that gateways that may provide a `Time` that has undefined accuracy are not considered for class B ping slots.
  - `Time` has undefined accuracy, while `GpsTime` is guaranteed (by the frontends) to have millisecond accuracy.

#### Testing

Local testing with real end devices.

##### Regressions

This change causes absolute time class C downlinks to fail in the absence of a GPS sync, or at least of some RTTs being observed between the NS and the gateway. 

Previously scheduling the absolute time downlinks would always work, but the timestamp would have an accuracy equivalent to the accuracy of the forwarder (UDP in some implementations has seconds accuracy, MQTTv2 is always second accuracy, while BasicStation has a fractional timestamps of undefined accuracy).

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/io.go
+++ b/pkg/gatewayserver/io/io.go
@@ -22,6 +22,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/gogo/protobuf/types"
 	"go.thethings.network/lorawan-stack/v3/pkg/band"
 	"go.thethings.network/lorawan-stack/v3/pkg/config"
 	"go.thethings.network/lorawan-stack/v3/pkg/errorcontext"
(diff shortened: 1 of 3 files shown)


KrishnaIyer (approved):
LGTM.

johanstokking:
> This change causes absolute time class C downlinks to fail in the absence of a GPS sync, or at least of some RTTs being observed between the NS and the gateway.

What do you mean with the second part of this sentence?

adriansmares:
I wanted to refer specifically to the following part of `ScheduleAt`:
https://github.com/TheThingsNetwork/lorawan-stack/blob/0bd5b7a84e2368ffa8b8f4d886d500b2ca2f5ef3/pkg/gatewayserver/scheduling/scheduler.go#L310-L323
When a gateway GPS time sync is not available, the backend allows the downlink to be scheduled using only the server time by subtracting half of the median RTT. 
The `mediatRTT` is set if we have an RTT enabled gateway with at least 5 RTTs observed:
https://github.com/TheThingsNetwork/lorawan-stack/blob/0bd5b7a84e2368ffa8b8f4d886d500b2ca2f5ef3/pkg/gatewayserver/scheduling/scheduler.go#L294-L300
`errNoAbsoluteGatewayTime` will now be visible more often since we no longer sync with `TxSettings.Time`. Basically any uplink had a `TxSettings.Time`, but this timestamp is often far off.

---

Since `settings.Time` has an unclear source and accuracy, the messages were sent, but very often at the wrong time.

This becomes bad when we have two downlink paths for an absolute time downlink. If we don't disqualify the downlink paths where the gateway time is not well-sourced (i.e. not from GPS), we may end up sending the message at the wrong time. Downlink paths where the gateway time may be wrong must be disqualified.

You may argue that we need this rigor only for class B, and these gateways are guaranteed to be GPS time enabled. But if the same message is picked up by a gateway that is not GPS enabled, before this change the non-GPS downlink path may have been used and caused the downlink to be transmitted at the wrong time (more often than not).

For class C absolute time scheduling the same issue appears - if you have two paths, one with 'bad quality' and one with 'good quality', it is hard to disqualify the 'bad quality' one without having some extra settings in `TxSettings`. In the future the `ClassBC` message maybe can be adapted to have a 'loose timestamp' option, but since this diverts us from Backend Interfaces, we should think more about it.

johanstokking:
I see, yeah I agree. I think it's indeed acceptable to observe a couple of round trip delays in order to guestimate the gateway transmission time on the server, instead of relying on a time provided by the gateway.

PR #5738, opened 2022-08-29
Fix snapshot release workflow

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/5609

#### Changes

- Remove the commit input, as it is not needed, and also we should require that a branch exists for a snapshot to be created.

#### Testing

Tested in a fork.

##### Regressions

N/A

#### Checklist

### DIFF
--- a/.github/workflows/release-snapshot.yml
+++ b/.github/workflows/release-snapshot.yml
@@ -5,10 +5,6 @@ on:
     branches:
       - 'v3.*'
   workflow_dispatch:
-    inputs:
-      commit:
-        description: Commit hash
-        required: true
 
 env:
   NODE_ENV: production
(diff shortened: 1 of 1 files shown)


.github/workflows/release-snapshot.yml
```diff
@@ -5,10 +5,6 @@ on:
     branches:
       - 'v3.*'
   workflow_dispatch:
```
KrishnaIyer, line 7: Isn't this an empty key now?

KrishnaIyer (approved):

.github/workflows/release-snapshot.yml
adriansmares: [This is the expected syntax apparently](https://github.com/orgs/community/discussions/26098#discussioncomment-3250374).

PR #5742, opened 2022-08-30
Add notification receivers validation

nicholaspcr:
#### Summary

In short while working on the Entity Limit Notification on TTES. I found a behaviour in which if you specify through code that an email notification does not have receivers, the email will be sent to all users. So in short this validation serves as a check in order to avoid having multiple users being spammed if there is an error on the future updates.

Obs: Did not open an issue for this.

#### Changes

- Validate if the `Receivers` field is `nil` when storing an email.

#### Testing

Tested in the same scenario where I found the behaviour, picked this commit put it on my ttes branch and ran the same script that generate the email spam.

##### Regressions

I don't  think there is a regression, is an addition of a simple if on the `storeNotification` and this shouldn't be empty unless we specify it so.

#### Notes for Reviewers

Ideally this should be an error that should be made to be catched at compile time or on the general CI but I couldn't think of basic way of doing so, the very next things was the current solution where the integration tests would fail due to the email error. 

#### Checklist

### DIFF
--- a/pkg/identityserver/notification_registry.go
+++ b/pkg/identityserver/notification_registry.go
@@ -21,6 +21,7 @@ import (
 	clusterauth "go.thethings.network/lorawan-stack/v3/pkg/auth/cluster"
 	"go.thethings.network/lorawan-stack/v3/pkg/auth/rights"
 	"go.thethings.network/lorawan-stack/v3/pkg/email"
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
 	"go.thethings.network/lorawan-stack/v3/pkg/identityserver/store"
 	"go.thethings.network/lorawan-stack/v3/pkg/log"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
@@ -62,6 +63,9 @@ func filterUserIdentifiers(ids []*ttnpb.OrganizationOrUserIdentifiers) []*ttnpb.
 }
 
 func (is *IdentityServer) notifyInternal(ctx context.Context, req *ttnpb.CreateNotificationRequest) error {
+	if err := req.ValidateFields(); err != nil {
+		panic(err)
+	}
 	ctx = is.FromRequestContext(ctx)
 	if authInfo, err := is.authInfo(ctx); err == nil {
 		if userIDs := authInfo.GetEntityIdentifiers().GetUserIds(); userIDs != nil {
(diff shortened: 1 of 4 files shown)


pkg/identityserver/notification_registry.go
```diff
@@ -165,8 +166,13 @@ func (is *IdentityServer) lookupNotificationReceivers(ctx context.Context, req *
 	return receiverUserIDs, nil
 }
 
+var errNoEmailReceiver = errors.DefineInvalidArgument("no_email_receiver", "no email receiver")
+
 func (is *IdentityServer) storeNotification(ctx context.Context, req *ttnpb.CreateNotificationRequest, receiverUserIDs ...*ttnpb.UserIdentifiers) (*ttnpb.Notification, error) {
 	var notification *ttnpb.Notification
+	if req.Receivers == nil {
```
adriansmares: I think it would be better in `is.notifyInternal` to run `req.ValidateFields()` and return the error there than check this here (or even `panic` for that matter).

Also `is.createNotification` should return an error if after `is.lookupNotificationReceivers` there are no receivers.

Also in general when you have slices, use a `len == 0` check instead of a `nil` check. There are non-`nil` slices with length 0.

nicholaspcr: Agree, best to use the `ValidateFields` method in the `notifyInternals` methods.

> Also is.createNotification should return an error if after is.lookupNotificationReceivers there are no receivers.

Ended up adding the validation inside the `lookupNotificationReceivers`

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ For details about compatibility between different releases, see the **Commitment
 - Support for loading end device template from Device Repository when importing devices using a CSV file.
 - Experimental support for normalized payload.
 - Decoded payloads are now visible for downlinks in the Console.
+- Extra validations were added before sending internal notifications.
```
adriansmares: This is strictly for us - no need to add it to changelog (end users do not 'feel' it).

adriansmares (approved):

CHANGELOG.md
nicholaspcr: true, wasn't sure I needed to add it or not but opted to do it since you would review it.

Technically the only other group that would need this information are people that fork this repository, so indeed might not be necessary to include it here.

kschiffer (approved):

PR #5746, opened 2022-08-31
Fix AU915 and US915 LR-FHSS channels max data rate indices

adriansmares:
#### Summary

This PR fixes the max data rate index used by the 500kHz channels in the US915 and AU915. Starting with RP002-1.0.2 (when LR-FHSS was first introduced), these channels are no longer only meant for SF7BW500, but also for LR-FHSS - see the second paragraph:

image

image

#### Changes

- Fix US915 and AU915 max data rate indices for the wide channels.
- Fix CN470-510 channel min data rate index.
  - I've added a sanity test that checks if the data rate range of a channel is well defined for a band, and caught this error.

#### Testing

Unit testing and also tested with a physical device the data rate changes.

##### Regressions

N/A. This is bleeding edge.

#### Checklist

### DIFF
--- a/pkg/band/us_902_928.go
+++ b/pkg/band/us_902_928.go
@@ -22,7 +22,7 @@ const (
 )
 
 var (
-	us902928UplinkChannels = func() []Channel {
+	us902928UplinkChannels = func(wideChannelDelta ttnpb.DataRateIndex) []Channel {
 		uplinkChannels := make([]Channel, 0, 72)
 		for i := 0; i < 64; i++ {
 			uplinkChannels = append(uplinkChannels, Channel{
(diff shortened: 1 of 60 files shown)


pkg/band/us_902_928.go
```diff
@@ -22,7 +22,7 @@ const (
 )
 
 var (
-	us902928UplinkChannels = func() []Channel {
+	us902928UplinkChannels = func(wideChannelOffset ttnpb.DataRateIndex) []Channel {
```
johanstokking: offset vs delta; what's the difference?

johanstokking (approved):

pkg/band/us_902_928.go
adriansmares: Just inconsistency - renamed to `wideChannelDelta` like in AU915.

PR #5752, opened 2022-08-31
Treat normalized payload validation errors as warnings

johanstokking:
#### Summary

Follow up on #5730 
References https://github.com/TheThingsNetwork/lorawan-stack/issues/5429
References https://github.com/TheThingsNetwork/lorawan-devices/issues/395

cc @pablojimpas

#### Changes

This changes the way field validation errors are treated. Before, the error was propagated to the caller and the entire payload wouldn't be considered normalized. This means that one out-of-bounds field renders the entire payload invalid. I think this isn't a nice developer experience: we should still not set these fields to out-of-bounds values but we should accept the rest of the fields and set the errors as warnings.

There's also a small unrelated change to see the `normalized_payload{_warnings}` in the payload simulation view.

#### Testing

CI and simulated uplink:
Screen Shot 2022-08-31 at 17 59 13

##### Regressions

None expected

#### Notes for Reviewers

- No need for changelog entry because this is still unreleased
- @kschiffer only review code owned files. I didn't add a code editor for the normalized payload. I think that would be nice though (although it's in the complete payload too), but just putting another editor in there didn't feel right UX wise so I'll leave this up to you. I consider this out of scope for this PR though

#### Checklist

### DIFF
--- a/pkg/messageprocessors/javascript/javascript_test.go
+++ b/pkg/messageprocessors/javascript/javascript_test.go
@@ -367,24 +367,49 @@ func TestDecodeUplink(t *testing.T) {
 		err := host.DecodeUplink(ctx, ids, nil, message, script)
 		a.So(err, should.BeNil)
 
-		decodedPayload, err := gogoproto.Map(message.DecodedPayload)
-		a.So(err, should.BeNil)
-		a.So(decodedPayload, should.Resemble, map[string]interface{}{
-			"temperature": -21.3,
+		a.So(message.DecodedPayload, should.Resemble, &pbtypes.Struct{
+			Fields: map[string]*pbtypes.Value{
+				"temperature": {
+					Kind: &pbtypes.Value_NumberValue{
+						NumberValue: -21.3,
+					},
+				},
+			},
 		})
 		a.So(message.DecodedPayloadWarnings, should.Resemble, []string{"it's cold"})
 
-		a.So(message.NormalizedPayload, should.HaveLength, 1)
+		a.So(message.NormalizedPayload, should.Resemble, []*pbtypes.Struct{
+			{
+				Fields: map[string]*pbtypes.Value{
+					"air": {
+						Kind: &pbtypes.Value_StructValue{
+							StructValue: &pbtypes.Struct{
+								Fields: map[string]*pbtypes.Value{
+									"temperature": {
+										Kind: &pbtypes.Value_NumberValue{
+											NumberValue: -21.3,
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		})
+		a.So(message.NormalizedPayloadWarnings, should.BeEmpty)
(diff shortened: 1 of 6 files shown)


pkg/messageprocessors/javascript/javascript_test.go
```diff
@@ -539,8 +543,23 @@ func TestDecodeUplink(t *testing.T) {
...
+		a.So(message.NormalizedPayload, should.HaveLength, 1)
+		parsedMeasurements, err := normalizedpayload.Parse(message.NormalizedPayload)
+		a.So(err, should.BeNil)
+		measurements := make([]normalizedpayload.Measurement, len(parsedMeasurements))
+		for i, m := range parsedMeasurements {
+			measurements[i] = m.Measurement
+		}
+		a.So(measurements, should.Resemble, []normalizedpayload.Measurement{
+			{
+				Air: &normalizedpayload.Air{},
```
adriansmares: I think the measurement should be skipped completely, otherwise you're providing the subscribers with a measurement that would not pass validation (think of a measurement which has an invalid zero value).

johanstokking: Yeah actually I just deleted the commit that omitted empty objects cause I think it's not a good idea to do that as part of validation.

All values are optional anyway, so default values will not cause invalid values.

The issue is that if a developer returns:

```js
return {
  data: {
    air: {}
  }
}
```

We would keep `air`, but if there was an invalid temperature in `air` (and no other valid field), we would remove `air`. Hence I deleted that commit because that is ambiguous.

I'm not against omitting empty objects, but if we do it, it would come after validation. I just don't know what the value is, except that it looks cleaner. Would there be objects in the future which presence already indicate something? Like `location`; the device _tried_ getting a GPS fix but didn't get any, vs not having `location` at all?

adriansmares: For location, historically payload formatters had a `fix` or `hasFix` field that would be false when an attempt was made, but was unsuccessful. I don't think that we should give 'presence' semantics to these empty measurements - I don't consider the empty measurement to be an error, but I do think it should be stripped (even in the case that you mentioned where it was empty before validation).

pablojimpas: I agree with @adriansmares, emitting empty fields will be useless for integrations. Imagine that an application developer stores the normalized payloads in a database as they arrive, without any checking or format modification, that database will be wasting storage on data without any meaningful value.

johanstokking: Okay, fine with me as well!

adriansmares (approved):

kschiffer (approved):
Changes look good. I wouldn't want to add something ad-hoc in here, so if you could create an issue for showing the normalized payload that would be great.

PR #5765, opened 2022-09-02
Add more specific client rights

### LINKED ISSUE #5447: Use more specific client rights to make it consistent with other entities
#### Summary
We should use specific client rights and not just rely on a single `RIGHT_CLIENT_ALL`, which looks like a pseudo-right but is actually an explicit right.

#### Why do we need this?
To correctly assign all client rights to admins. Currently all rights with the suffix `_ALL` will not be granted to admins, since they are considered to be pseudo-rights. This causes admins to not be allowed to watch some non-public data of foreign clients when `--is.admin-rights.all` is not set to true.

Rights composition of other clients, where several explicit rights are combined by an `_ALL` pseudo right.

- A start point for explicit rights. Right now that could be a single `RIGHT_CLIENT_GENERAL_INFO`
- Evaluating the current `RIGHT_CLIENT_ALL` as a pseudo right
- Checking for `RIGHT_CLIENT_GENERAL_INFO` instead of `RIGHT_CLIENT_ALL` when accessing clients

Can be followed from our other entities.

Unit tests should cover this.

I don't really have time so I will assign to @htdvisser who might want to reassign.

This is currently blocking #5414

nicholaspcr:
#### Summary

Closes #5447

#### Changes

- Add specific rights for clients
- Update tests using client rights

#### Testing

Existing UT

##### Regressions

Considering that all clients currently have `RIGHTS_CLIENTS_ALL` the creations of specific clients should not have an impact on existing clients.

#### Notes for Reviewers

...

#### Checklist


adriansmares (approved):

kschiffer (approved):
Nice!

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ For details about compatibility between different releases, see the **Commitment
 - Decoded payloads are now visible for downlinks in the Console.
 - Support for dynamic ping slot frequencies, as used by the US915 and AU915 bands.
 - Support for LoRa Basics Station beaconing.
+- Add more specific rights for OAuth clients.
```
adriansmares: Please make sure this goes into the `v3.22` section of the changelog.

PR #5766, opened 2022-09-02
Fix device import on JS-only deployments

kschiffer:
#### Summary
This quickfix PR fixes device import when using JS-only configs.

#### Changes

- Hide the frequency plan input for setting fallbacks
- Respect allowed field mask paths when doing device operations
- Update e2e tests

#### Testing

Manual and e2es

##### Regressions

This affects device update and create requests. E2Es should unearth any significant issue there.

#### Checklist


kschiffer:
cc @furtiman

adriansmares (commented):
LGTM.

ryaplots (approved):

kschiffer:
The whole issue escalated quite a bit further. Here's some more info for the record:
- We used to not do any field mask checks for end device operations (create and update). I added those to make sure that only valid field masks can be requested. For this, I used a concatenated list of the allowed field masks for all device registry setters.
- We already had logic in device update/create to make sure that only the deepest possible field mask is added (e.g. `mac_settings.supports_class_b`, will not add `mac_settings`).
- I also added logic to avoid unnecessary requests to NS/AS/JS when doing device requests, based on whether the IS has stored them

Issue #3658, opened 2021-01-12
Network Server logging


rvolosatovs:
#### Summary

Reorganize Network Server logging

#### Why do we need this?

To simplify debugging. e.g. see https://github.com/TheThingsNetwork/lorawan-stack/issues/3637#issuecomment-758672243

#### What is already there? What do you see now?

Some important log statements are missing and some less relevant, low-level messages are printed

#### What is missing? What do you want to see?

Clear and simple log statements, which are easily understood by us and our customers

#### How do you propose to implement this?

Think about what we want to see from NS as a whole and go through existing statements.

#### How do you propose to test this?

Run the stack locally and send some traffic

#### Can you do this yourself and submit a Pull Request?

yes

nejraselimovic:
@htdvisser I guess this is still relevant, but needs input from others too. Can you pick it up?

htdvisser:
Requested input from @michalborkowski96 @i3mSuperMan @ymgupta.

michalborkowski96:
I'll copy-paste what I wrote on Slack:

When something breaks, there are typically two scenarios:
1. Component completely dead
2. Component mostly working, broken for a small number of users

When case 1 happens usually in the logs you see a few messages repeated endlessly. In this case the more messages the merrier, because it gives you more information about the actual problem. From my perspective it's very simple, because I already have a clear starting point.

When case 2 happens you need to catch useful information in a sea of messages related to normal operation. This is very difficult when I have only vague idea about how TTS works, so the less messages the better. This is why I was pushing for https://github.com/TheThingsIndustries/lorawan-stack/issues/2702 because then I would be able to import these messages somewhere, and filter out normal ones.

I don't really know what to say besides the fact that both types of problems occur and we just need to look for some kind of balance.

Regarding messages themselves:
1. If a message keeps appearing all the time, and looks like normal operation, maybe it's worth it to remove it, or somehow aggregate (print once in a few seconds that an event happened that many times). In particular I think that aggregation sounds like a good idea, as it'd reduce the number of messages without much information loss.
2. If fetching messages to my local computer takes longer than 15 minutes for the whole stack for last 24h, then we have way too many messages. If we're under this, or a similar sensible limit, we don't really need to remove messages. Otherwise we do, because this means that logs can't be searched through even by a machine.
3. If an event is rare, it's worth logging, even if it's seemingly normal (for example completely valid, but super-rare device behaviour).

i3mSuperMan:
[...]

PR #5776, opened 2022-09-06
Make ADR commands generation report errors via events

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/3658

#### Changes

- Refactor the `LinkADRReq` command generation to report errors via events, as opposed to log messages which are not accessible for hosted instances.
- Improve `LinkADRReq` validation:
  - Validate the current channels for to check that they are well formed (not only the desired channels).
  - Check individual data rate indices against the available data rate index intervals, instead of just checking `min <= index <= max`. This is relevant for cases in which there is a gap of data rate indices between `min` and `max`.
  - The NS will avoid now scheduling a channel mask which would render the end device impossible to send uplinks. This happens if for example the end device was using LR-FHSS channels and we disable uplinks over those. If we instruct the end device to keep its data rate, we are asking it to mute itself, since no channels are going to have uplinks available with the old LR-FHSS data rate.
- Refactor coding style for better visibility.

#### Testing

Unit testing.

##### Regressions

`generateLinkADRReq` no longer attempts to lower the data rate used by the end device when the desired data rate index or transmission power has been rejected. This is fine, as the ADR algorithm was already doing it, and in cases in which ADR is disabled we do not need to do this for the end device - we will just maintain the current parameters.

#### Checklist

### DIFF
--- a/config/messages.json
+++ b/config/messages.json
@@ -7046,9 +7046,18 @@
       "file": "errors.go"
     }
   },
-  "error:pkg/networkserver/internal:downlink_channel": {
+  "error:pkg/networkserver/internal:incompatible_channel_mask": {
     "translations": {
-      "en": "channel does not allow uplinks"
+      "en": "the channel mask is not compatible with the current parameters"
+    },
+    "description": {
+      "package": "pkg/networkserver/internal",
+      "file": "errors.go"
+    }
+  },
+  "error:pkg/networkserver/internal:invalid_data_rate_index": {
+    "translations": {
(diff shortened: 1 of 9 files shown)


johanstokking:
What happened here?

adriansmares:
I've opened this PR while you were away, as part of technical debt refactors. Since it was hard to review for people that are not really familiar with the NS, I decided to close it and open it in a form that is more reviewable commit by commit.

I will open a new PR that should be easier to digest.

PR #5795, opened 2022-09-12
Fix form bugs related to ADR

ryaplots:
#### Summary

Fix some network server section bugs in the device general settings form.

#### Changes

- The `settings` have to be spread inside `mac_settings` otherwise other values will not be shown correctly
- In adr static fields, if the user set them to 0, after refresh the values went back to being empty, because the backend did not include these fields in the response. This behavior happened due to adr static fields being either `enums` or `integers`, where '0' corresponds to 'default' or 'missing'. So in order to maintain some consistency and not have empty values when the user has set the value to '0'. I decided, after speaking to Adrian, to simply set the field value to '0' every time it is not present in the backend response.

https://user-images.githubusercontent.com/56658938/189659669-be653fbb-71d3-47c1-b96f-9f50ac5ac625.mov

#### Testing

manually

#### Checklist

### DIFF
--- a/pkg/webui/console/components/mac-settings-section/index.js
+++ b/pkg/webui/console/components/mac-settings-section/index.js
@@ -120,8 +120,6 @@ const maxDutyCycleOptions = [
 const encodeAdrMode = value => ({ [value]: {} })
 const decodeAdrMode = value => (value !== undefined ? Object.keys(value)[0] : null)
 
-const decodeStaticFields = value => (value ? value : 0)
-
 const MacSettingsSection = props => {
   const {
     activationMode,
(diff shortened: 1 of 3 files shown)


kschiffer (changes requested):
We typically use the [`_emitDefaults` function within the SDK](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.22/sdk/js/src/service/devices/index.js#L60-L135) to make swallowed null values explicit, so let's stick to that rather.

kschiffer (approved):

mjamescompton (approved):

PR #5796, opened 2022-09-12
Support factory preset frequencies in fixed channel plans

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack-support/issues/335

#### Changes

- Factory preset frequencies may now be provided for fixed channel plans bands.
  - Semantically, the frequencies are interpreted as the uplink frequencies enabled at boot time in these bands.
- In bands with variable channel plans, factory preset frequencies are now validated to be within a sub band.
- The channel masks that we generate for LoRaWAN 1.0 and 1.0.1 end devices are now valid when evaluated individually.
  - Starting with LoRaWAN 1.0.2, `LinkADRReq` MAC commands are meant to be evaluated in atomic blocks - you have to accept all of the commands or reject all of the commands. This is not the case in LoRaWAN 1.0 and 1.0.1.
  - Our generator generates a `Cntl=7` command which disables all of the channels, then a `Cntl=n` command that enables an individual FSB. This is space efficient - it saves about 15 bytes per downlink which does matter at high SF.
  - Unfortunately the individual `Cntl=7` command that disables all of the channels is interpreted by the end device as being invalid (since when you interpret it as a singular command, it tries to mute the end device).
  - We're more friendly to such devices now - specifically we will now enable the individual FSBs that we do use, then disable the remaining FSBs. Since we start with the enabled channels, all of the intermediary states through which the end device goes while enabling the channels are valid (you always have _some_ channels enabled).

#### Testing

Unit testing. 

I don't have a physical device with me at the moment, but one of our The Things Uno US915 variants can be used to test this change actually - the initial batches come with a 1.0.1 firmware, which shows the same behavior (reject the `Cntl=7`).

##### Regressions

N/A. LoRaWAN 1.0 and 1.0.1 end device would previously reject these commands like there is no tomorrow anyway.

#### Notes for Reviewers

@KrishnaIyer please review all of the backend changes.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ For details about compatibility between different releases, see the **Commitment
 
 - `--mac-settings.adr.mode.disabled`, `--mac-settings.adr.mode.dynamic` and `--mac-settings.adr.mode.static` flags of the `end-device update` command.
 - Pagination in `sessions` and `access tokens` tables in the Console.
+- `LinkADRReq` MAC command generation for LoRaWAN 1.0 and 1.0.1 end devices.
 
 ### Security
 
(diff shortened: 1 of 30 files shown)


KrishnaIyer (approved):
Backend changes look great!

ryaplots (approved):
Front-end changes look good to me

kschiffer:
Nice!

Issue #5393, opened 2022-04-19
AS923: NewChannelReq rejection, ChMask discrepancy, LinkADRReq rejection / repetition


jbiveroni:
#### Summary

After joining, the device is setup to use channels 0...6 (ChMask 0x7F). The LNS sends a NewChannelReq (channel 7). The device rejects DrRange, and therefore does not enable the new channel (ChMask 0x7F). However, in subsequent LinkADRReqs, the LNS uses ChMask 0xFF. The devices therefore rejects these LinkADRReqs. The LNS repeats LinkADRReq commands "forever", although they are always rejected. This creates unnecessary downlink traffic.

#### Steps to Reproduce

TTS, v3.18.2: Asia 920-923 MHz, LoRaWAN Specification 1.0.2, RP001 Regional Parameters 1.0.2 revision B.
Device: Dwell time enabled (DR2...5), 2 default channels enabled.

1. Joining, channels 0...6 enabled.
2. LNS: NewChannelReq: MinDR: 0, MaxDR: 5, freq: 924xxx; LinkADRReq: DR 5, channel mask 0x00FF (8 channels).
3. Device: NewChannelAns: Accept frequency, REJECT Min/MaxDR. Probably, because MinDR = 0 is not permitted, minimum DR is 2 (SF10). New channel is not enabled on device. LinkADRAns: Accept DR, PWRIDX, but REJECT channel mask (8 channels, but only 7 are set up now!)
4. LNS: LinkADRReq: DR 5, channel mask 0x00FF (8 channels).
5. Device: LinkADRAns: Accept DR, PWRIDX, but REJECT channel mask (8 channels, but only 7 are set up now!)
6. LNS: LinkADRReq (above)
7. Device: LinkADRAns (above)
8. ... Repeats "forever"

#### What is already there? What do you see now?

Probably 2 separate issues:
- NewChannelReq rejected -> ChMask LNS/device do not match
- LinkADRReq rejected -> LNS repeats LinkADRReq "forever"

Workaround (not satisfactory):
Use frequency plan "Asia 915–928 MHz (AS923 Group 1) with only default channels". NewChannelReq is not used. This way, the channel mask bits agree and the LinkADR is accepted.

More observations:
- Disable ADR on the device: behavior stays the same.
- Disable ADR in LNS (--mac-settings.use-adr=false): behavior stays the same (??).

#### What is missing? What do you want to see?

[...]

PR #5811, opened 2022-09-19
Reorganise MAC command priorities

### LINKED ISSUE #4041: MAC command NewChannelAns can't delete channel
#### Summary
The channels a device uses to communicate  can be modified by the network server using MAC commands (NewChannelReq & NewChannelAns). 
This should work to add new channels but also to remove some 

But it seems that removing a channel isn't implemented 

#### Steps to Reproduce

Add this test case to `pkg/networkserver/mac/new_channel_test.go#TestHandleNewChannelAns` : 

```
{
	Name: "Delete channel",
	Device: &ttnpb.EndDevice{
		MACState: &ttnpb.MACState{
			PendingRequests: []*ttnpb.MACCommand{
				(&ttnpb.MACCommand_NewChannelReq{
					ChannelIndex:     4,
					Frequency:        0,
					MinDataRateIndex: 0,
					MaxDataRateIndex: 0,
				}).MACCommand(),
			},
			CurrentParameters: ttnpb.MACParameters{
				Channels: []*ttnpb.MACParameters_Channel{
					nil,
					nil,
					nil,
					nil,
					{
						DownlinkFrequency: 42,
						UplinkFrequency:   42,
						MinDataRateIndex:  2,
						MaxDataRateIndex:  3,
						EnableUplink:      true,
					},
				},
			},
		},
	},
	Expected: &ttnpb.EndDevice{
		MACState: &ttnpb.MACState{
			CurrentParameters: ttnpb.MACParameters{
				Channels: []*ttnpb.MACParameters_Channel{
					nil,
					nil,
					nil,
					nil,
				},
			},
			PendingRequests: []*ttnpb.MACCommand{},
		},
	},
	Payload: &ttnpb.MACCommand_NewChannelAns{
		FrequencyAck: true,
		DataRateAck:  true,
	},
	Events: events.Builders{
		EvtReceiveNewChannelAccept.With(events.WithData(&ttnpb.MACCommand_NewChannelAns{
			FrequencyAck: true,
			DataRateAck:  true,
		})),
	},
}
```

#### What do you see now?
The test fails, the channel is still here and was not removed 

The test suceeds and the channel is removed 
#### Environment
Observed in v3.9 (self-hosted) 
Can reproduce in v3.12

In `pkg/networkserver/mac/new_channel.go#HandleNewChannelAns`, add a case if the frequency requested was 0 (meaning we wanted to delete the channel), and if  `len(DesiredParameters.Channels) < len(CurrentParameters.Channels)`

The test should pass 
No, I don't know how to solve the following issues:

- HandleNewChannelAns can be called multiple times if there is duplicated uplinks, but we should only remove once 
- How to keep `req.ChannelIndex` coherent  with other mac commands (that may be in the same uplink) if we want to delete a channel that isn't the last one

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/4041
References https://github.com/TheThingsNetwork/lorawan-stack/issues/5393 (This PR fixes the channel mask loop, but the data rate index range 'bug' is still open)

#### Changes

- Send `TxParamSetupReq` as early as possible.
  - The reason to do this is to ensure that the end device and the Network Server are synchronized with respect to dwell time limitations. This also enables the Network Server to use the RX1 slot in places where the usage is ambiguous.
- `BeaconFreqReq` is now sent even outside of class B operation mode, if a drift is found.
  - The condition for this command did not actually make sense - an end device needs to acquire a beacon before it switches to class B, but it cannot acquire the beacon if it listens for it on the wrong channel.
- `DevStatusReq` is now lower priority than other MAC commands.
  - The synchronization of MAC parameters (channels, data rates, frequencies) is more important than an up to date `DevStatusReq` at the beginning of a session.
- `LinkADRReq` channel mask generation now takes into account the fact that the end device may not actually have a channel available. Consider the following scenario:
  - The end device joins. It knows only the 2 default channels of AS923.
  - The NS enqueues a `NewChannelReq` and a `LinkADRReq`.
  - Before this fix, the `LinkADRReq` would attempt to enable all 8 channels of the AS923 frequency plan, even though the end device knows at most 3 channels after processing the singular `NewChannelReq`.
  - After this fix, the `LinkADRReq` attempts to enable only the 3 channels (2 known defaults + 1 from the enqueud `NewChannelReq`).
- `NewChannelReq` will no longer perpetually attempt to delete a channel.
  - This used to occur if the number of current channels was larger than the number of desired channels.
- NS flow tests have been adapted to the MAC command reordering changes.
- NS flow tests now correctly render the MAC commands attached to a downlink when the MAC version encrypts the MAC commands.

#### Testing

Unit testing + physical `AS923` end device tested on `staging1`.

##### Regressions

Physical end device testing covered pretty much all of the changes:
- `TxParamSetup` is among the first downlink commands scheduled.
- The first `LinkADRReq` now correctly contains a mask that refers only to the channels which have been configured on the end device.
- `NewChannelReq` is working correctly in adding the remaining 6 channels.
- `DevStatusReq` is not enqueued among the first MAC commands anymore.

#### Notes for Reviewers

@KrishnaIyer please take a look at the changes. Let me know if you need more context somewhere, but the description above should cover everything hopefully. 

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ For details about compatibility between different releases, see the **Commitment
 - LoRa coding rate now defined in `DataRate` instead of `Band`.
 - The Network Server will now schedule a potentially empty downlink in order to stop end devices from sending sticky MAC commands.
 - Factory preset frequencies may now be provided for bands with fixed channel plans, such as US915 or AU915. The factory preset frequencies are interpreted as the only channels which are enabled at boot time.
+- `TxParamSetupReq` MAC command priority has been increased.
+- `DevStatusReq` MAC command priority has been lowered.
 
 ### Deprecated
 
@@ -33,6 +35,7 @@ For details about compatibility between different releases, see the **Commitment
 - `--mac-settings.adr.mode.disabled`, `--mac-settings.adr.mode.dynamic` and `--mac-settings.adr.mode.static` flags of the `end-device update` command.
 - Pagination in `sessions` and `access tokens` tables in the Console.
 - `LinkADRReq` MAC command generation for LoRaWAN 1.0 and 1.0.1 end devices.
+- `LinkADRReq` no longer attempts to enable channels which have not yet been negotiated with the end device.
 
 ### Security
 
--- a/pkg/networkserver/downlink.go
+++ b/pkg/networkserver/downlink.go
@@ -234,8 +234,7 @@ func (ns *NetworkServer) generateDataDownlink(ctx context.Context, dev *ttnpb.En
 		dev.MacState.QueuedResponses = nil
 		dev.MacState.PendingRequests = dev.MacState.PendingRequests[:0]
 
-		enqueuers := make([]func(context.Context, *ttnpb.EndDevice, uint16, uint16) mac.EnqueueState, 0, 13)
-		enqueuers = append(enqueuers,
+		enqueuers := []func(context.Context, *ttnpb.EndDevice, uint16, uint16) mac.EnqueueState{
 			mac.EnqueueDutyCycleReq,
 			func(ctx context.Context, dev *ttnpb.EndDevice, maxDownLen uint16, maxUpLen uint16) mac.EnqueueState {
 				return mac.EnqueueTxParamSetupReq(ctx, dev, maxDownLen, maxUpLen, phy)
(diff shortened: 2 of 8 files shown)


KrishnaIyer (approved):
Also LGTM 👍

PR #5824, opened 2022-09-27
Fix downlink path computation for non-LoRa modulations

adriansmares:
#### Summary

This PR downlink path computation for non-LoRa modulated uplinks (FSK/LR-FHSS). 

The context is as follows: when an uplink is received by multiple gateways, we need to choose through which gateway we may send a potential RX1/RX2 downlink. This selection is traditionally done by sorting the gateways by their LoRa-adjusted RSSI. LoRa-adjusted RSSI is a combination of channel RSSI and SNR. The problem is that modulations which are not LoRa (FSK and LR-FHSS for now) do not have an SNR (it is effectively either 0, or -127). This causes the paths to be sorted incorrectly since the SNR is invalid.

#### Changes

- Downlink paths from non-LoRa modulated uplinks are now sorted descending by channel RSSI (which is always defined).
- Reformat all Go files with `tools/bin/mage go:fmt`.

#### Testing

Unit testing.

##### Regressions

The path for LoRa modulated uplinks (i.e. 100% of the devices in the wild) is unchanged. Only non-LoRa modulated downlink paths are affected.

#### Notes for Reviewers

@KrishnaIyer please take a look.

#### Checklist

### DIFF
--- a/pkg/identityserver/user_registry_test.go
+++ b/pkg/identityserver/user_registry_test.go
@@ -54,8 +54,8 @@ func TestValidatePasswordStrength(t *testing.T) {
 			"password1":    false, // Too common.
 			"āaaAAA123!@#": true,
 			"       1A":    true,
-			"āaa	AAA123 ": true,
-			"āaaAAA123 ": true,
+			"āaa	AAA123 ":  true,
+			"āaaAAA123 ":   true,
 		} {
 			t.Run(p, func(t *testing.T) {
 				a, ctx := test.New(t)
--- a/pkg/networkserver/downlink.go
+++ b/pkg/networkserver/downlink.go
@@ -642,21 +642,37 @@ type downlinkPath struct {
 	*ttnpb.DownlinkPath
 }
 
-func buildMetadataComparator(mds []*ttnpb.MACState_UplinkMessage_RxMetadata) func(int, int) bool {
-	invalidMD := func(k int) bool { return mds[k].Snr == 0.0 || mds[k].ChannelRssi == 0.0 }
-	wantedRSSI := func(k int) float32 { return lora.AdjustedRSSI(mds[k].ChannelRssi, mds[k].Snr) }
-	return func(i, j int) bool {
-		lhsInvalid, rhsInvalid := invalidMD(i), invalidMD(j)
-		if lhsInvalid {
-			return lhsInvalid == rhsInvalid
+func buildMetadataComparator(
+	settings *ttnpb.MACState_UplinkMessage_TxSettings, mds []*ttnpb.MACState_UplinkMessage_RxMetadata,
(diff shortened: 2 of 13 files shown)


pkg/identityserver/user_registry_test.go
```diff
@@ -54,8 +54,8 @@ func TestValidatePasswordStrength(t *testing.T) {
 			"password1":    false, // Too common.
 			"āaaAAA123!@#": true,
 			"       1A":    true,
-			"āaa	AAA123 ": true,
-			"āaaAAA123 ": true,
+			"āaa	AAA123 ":  true,
+			"āaaAAA123 ":   true,
```
kschiffer, line 58: Doesn't look like the right spacing.
```suggestion
			"āaaAAA123 ":    true,
```

kschiffer: ?

kschiffer (approved):
Codeowned OK

pkg/identityserver/user_registry_test.go
nicholaspcr: Was gonna approve and write the same, my question would be if its a mistake on the part of `go:fmt` or not.

nicholaspcr (approved):
Gave a look into the changes non IS related and they also look good to me.

pkg/identityserver/user_registry_test.go
adriansmares: This is what `gofumpt` prefers - I didn't format this manually. I suppose does not match our intuition because the character spacing is weird for these esoteric letters. Note that in the middle of that 'word' there is an actual tab character, not a space.

image

KrishnaIyer (approved):
NS implementation LGTM

pkg/networkserver/downlink.go
```diff
@@ -642,21 +642,37 @@ type downlinkPath struct {
...
+				return lhsInvalid == rhsInvalid
+			}
+			return measure(i) >= measure(j)
 		}
-		return wantedRSSI(i) >= wantedRSSI(j)
+	}
+	switch settings.DataRate.Modulation.(type) {
+	case *ttnpb.DataRate_Lora:
+		return build(
+			func(k int) bool { return mds[k].Snr == 0.0 || mds[k].ChannelRssi == 0.0 },
```
johanstokking, line 660: An SNR of `0.0` is not invalid

adriansmares: This check has always been here since we first added this filtering (https://github.com/TheThingsNetwork/lorawan-stack/pull/4812).

I ran some queries on TTSCE `eu1` and indeed does occur (about 1% of the time). Removed in https://github.com/TheThingsNetwork/lorawan-stack/pull/5877.

PR #5829, opened 2022-09-27
Reset Application and Join Server state in CLI reset

### LINKED ISSUE #4874: The end device reset command should also reset the Application Server and Join Server state
#### Summary

The end device `reset` command should also reset the AS and JS states.

#### Why do we need this?

Because right now we reset the NS state, which is not sufficient.

Only the Network Server state is reset.

For devices that do not `supports_join`, we should reset `session.last_a_f_cnt_down` in the AS end device, otherwise we may lose a significant chunk of downlinks.
For devices that do `supports_join`, we probably can reset the nonces stored in the JS end device.

We may consider adding a `ResetFactoryDefaults` RPC to the AS and JS end device stores, similar to how the NS one works.
For the AS, it should just reset the `session.last_a_f_cnt_down`, while for the JS it probably should reset `last_dev_nonce`, `used_dev_nonces`,`last_join_nonce`, `last_rj_count_0` and `last_rj_count_1`.

#### Environment

`v3.16.0`

Make sure that the field mask contains `supports_join` while returning the NS end device. This allows the AS/JS logic to take place. Don't forget to remove the field at the end if it was not explicitly requested by the user.

Based on this, update the AS/JS end devices.

Reset an end device.

Can review.

happyRip:
#### Summary

Closes #4874

#### Changes

- added `EndDevice.UpdateTimestamps()` method
- `reset` cli command now resets values for `js`/`as` servers as well

#### Testing

Tested locally

- only necessary and requested fields are shown
image

##### Regressions

n/a

#### Notes for Reviewers

- [util: Add UpdateTimestamps method to ttnpb.EndDevice](https://github.com/TheThingsNetwork/lorawan-stack/pull/5829/commits/8e6427586eacf1ece14cd306b34e8830b17d3131) makes the `EndDevice.UpdateTimestamps()` method standardised and possibly common with `lorawan-stack-migrate` instead of a proprietary [updateDeviceTimestamps()](https://github.com/TheThingsNetwork/lorawan-stack-migrate/blob/master/pkg/source/ttnv3/util.go#L49-L56)

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/end_devices.go
+++ b/cmd/ttn-lw-cli/commands/end_devices.go
@@ -1024,6 +1024,13 @@ var (
 				device.UpdateTimestamps(jsDevice)
 			}
 
+			// Remove temporary fields (e.g. "supports_join") that were not selected by user
+			joinedPaths := ttnpb.AddFields(isPaths, ttnpb.AddFields(nsPaths, ttnpb.AddFields(asPaths, jsPaths...)...)...)
+			if diff := ttnpb.ExcludeFields(joinedPaths, paths...); len(diff) > 0 {
+				if err := device.SetFields(nil, diff...); err != nil {
+					return err
+				}
+			}
 			return io.Write(os.Stdout, config.OutputFormat, device)
 		},
 	}
--- a/cmd/ttn-lw-cli/commands/end_devices.go
+++ b/cmd/ttn-lw-cli/commands/end_devices.go
@@ -922,7 +922,7 @@ var (
 			}
 			paths := util.SelectFieldMask(cmd.Flags(), selectEndDeviceFlags)
 
-			isPaths, nsPaths, _, _ := splitEndDeviceGetPaths(paths...)
+			isPaths, nsPaths, asPaths, jsPaths := splitEndDeviceGetPaths(paths...)
 
 			is, err := api.Dial(ctx, config.IdentityServerGRPCAddress)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


happyRip:
Adding a function like
```go
// JoinFieldSlices joins all paths and returns them in a single slice.
func JoinFieldSlices(paths ...[]string) []string {
	var out []string
	// out := make([]string, 0, flattened_paths_length)
	for _, p := range paths {
		out = AddFields(out, p...)
	}
	return out
}
```
to `pkg/ttnpb/field_mask_utils.go` could improve readability of code fragments like https://github.com/TheThingsNetwork/lorawan-stack/pull/5829/files#diff-4f8da12a18316f8cb1c1024e2bd3984abd87ed523672003e1fead5abb4b65580R1016 making it
```go
joinedPaths := ttnpb.JoinFieldSlices(isPaths, nsPaths, asPaths, jsPaths)
```
instead of
```go
joinedPaths := ttnpb.AddFields(isPaths, ttnpb.AddFields(nsPaths, ttnpb.AddFields(asPaths, jsPaths...)...)...)
```

pkg/ttnpb/end_device.go
```diff
@@ -1552,3 +1552,14 @@ func (m *GetEndDeviceRequest) ExtractRequestFields(dst map[string]interface{}) {
 func (m *EndDevice) ExtractRequestFields(dst map[string]interface{}) {
 	m.GetIds().ExtractRequestFields(dst)
 }
+
+// Utility
```
adriansmares: Remove this comment and actually comment on the function itself with a description of the functionality.

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -949,22 +949,87 @@ var (
...
+				return err
+			}
+			device.UpdateTimestamps(nsDevice)
+
+			as, err := api.Dial(ctx, config.ApplicationServerGRPCAddress)
+			if err != nil {
+				return err
+			}
+			logger.WithField("paths", asPaths).Debug("Reset end device to factory defaults on Application Server")
+			asDevice, err := ttnpb.NewAsEndDeviceRegistryClient(as).Get(ctx, &ttnpb.GetEndDeviceRequest{
```
adriansmares, line 970: Why is the `Get` operation needed here ? If you provide the field masks with an empty source, you are in effect unsetting that particular value.
There is no need for `Get` -> `SetFields` -> `Set`, you can do just the `Set`.

cmd/ttn-lw-cli/commands/end_devices.go
```diff
@@ -949,22 +949,87 @@ var (
...
+					EndDeviceIds: devID,
+					FieldMask:    ttnpb.FieldMask(jsPaths...),
+				})
+				if err != nil {
+					return err
+				}
+				if err = jsDevice.SetFields(nil, "last_dev_nonce", "used_dev_nonces", "last_join_nonce", "last_rj_count_0", "last_rj_count_1"); err != nil {
+					return err
+				}
+				_, err = ttnpb.NewJsEndDeviceRegistryClient(js).Set(ctx, &ttnpb.SetEndDeviceRequest{
```
adriansmares, line 1014: Same as above.

adriansmares (commented):
Please don't forget to update the PR names when releasing them for review.

pkg/ttnpb/end_device.go
```diff
@@ -1553,8 +1553,7 @@ func (m *EndDevice) ExtractRequestFields(dst map[string]interface{}) {
 	m.GetIds().ExtractRequestFields(dst)
 }
 
-// Utility
-
+// UpdateTimestamps sets correct CreatedAt and UpdatedAt timestamps for EndDevice based on src device.
```
adriansmares: Specifically correct here means that the `CreateAt` is the earliest timestamp, while the `UpdatedAt` is the latest timestamp.

adriansmares (approved):

adriansmares (approved):

PR #5846, opened 2022-10-03
Gateway EUI constraint

### LINKED ISSUE #5615: Update gateway EUI database constraint to allow re-use after delete
### Summary

We should recreate the constraint we have on the `eui` field in the `gateways` table so that it excludes deleted gateways, similar to what we did for users (https://github.com/TheThingsNetwork/lorawan-stack/pull/5570).

### Current Situation

The database only allows a single gateway with the same EUI in the database. In order to allow EUI re-use after deleting a gateway, we unset the EUI. As a result, when restoring the gateway, we will have lost its EUI.

### Desired Situation

An updated constraint in the database, similar to what we did for users in https://github.com/TheThingsNetwork/lorawan-stack/pull/5570 will make this work better.

### Code of Conduct

nicholaspcr:
#### Summary

Closes #5615

#### Changes

- Update `gateway_eui_index` to only account for non deleted gateways.
- Add more test cases regarding new gtw-eui behaviour.

#### Testing

Manual testing and UT

##### Regressions

Couldn't think of any, it is a bit of an extension of what an user could do. Previously the EUI was cleared automatically but now there is a possibility of recovering but that is also not a guarantee.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
+- Gateway EUI is no longer unset when deleting a gateway, meaning it could be recovered if no other gateway claimed it. This requires a schema migration (`ttn-lw-stack is-db migrate`) because of the change in the database's `gateway_eui_index`.
+
 ### Deprecated
 
 ### Removed
--- a/pkg/identityserver/gateway_registry.go
+++ b/pkg/identityserver/gateway_registry.go
@@ -625,6 +625,7 @@ func (is *IdentityServer) restoreGateway(ctx context.Context, ids *ttnpb.Gateway
 		if time.Since(*deletedAt) > is.configFromContext(ctx).Delete.Restore {
 			return errRestoreWindowExpired.New()
 		}
+		ids = ttnpb.Clone(gtw.Ids)
 		return st.RestoreGateway(ctx, ids)
 	})
 	if err != nil {
--- a/pkg/identityserver/gormstore/gateway_store.go
+++ b/pkg/identityserver/gormstore/gateway_store.go
@@ -213,6 +213,30 @@ func (s *gatewayStore) DeleteGateway(ctx context.Context, id *ttnpb.GatewayIdent
 
 func (s *gatewayStore) RestoreGateway(ctx context.Context, id *ttnpb.GatewayIdentifiers) error {
 	defer trace.StartRegion(ctx, "restore gateway").End()
+	// Fetches the deleted gateway by ID.
+	queryGtwByID := s.query(ctx, Gateway{}, withGatewayID(id.GatewayId), withSoftDeleted())
+	queryGtwByID = selectGatewayFields(ctx, queryGtwByID, []string{"ids.eui"})
+	var gtwByIDModel Gateway
+	if err := queryGtwByID.First(&gtwByIDModel).Error; err != nil && !gorm.IsRecordNotFoundError(err) {
+		return err
+	}
+
+	// If deleted gateway contains an EUI, validates if there is another gateway with the same value.
+	if gtwByIDModel.GatewayEUI != nil {
+		queryByEUI := s.query(ctx, Gateway{}, withSoftDeleted(), withGatewayEUI(*gtwByIDModel.GatewayEUI))
+		queryByEUI = selectGatewayFields(ctx, queryByEUI, []string{"ids.gateway_id", "ids.eui"})
+		var total uint64
+		if err := queryByEUI.Count(&total).Error; err != nil {
+			return err
+		}
+
+		if total > 1 {
+			gtwByIDModel.GatewayEUI = nil
+			if err := s.updateEntity(store.WithSoftDeleted(ctx, true), &gtwByIDModel, "gateway_eui"); err != nil {
+				return err
+			}
+		}
+	}
 	return s.restoreEntity(ctx, id)
 }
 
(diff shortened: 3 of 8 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Added
 
 ### Changed
+- Gateway EUI is no longer unset when deleting a gateway, meaning it could be recovered if no other gateway claimed it.
```
adriansmares: New line between message and  header. Also needs warning for `is-db` migrate, and this should target `3.23`.

pkg/identityserver/gateway_registry.go
```diff
@@ -625,6 +626,7 @@ func (is *IdentityServer) restoreGateway(ctx context.Context, ids *ttnpb.Gateway
 		if time.Since(*deletedAt) > is.configFromContext(ctx).Delete.Restore {
 			return errRestoreWindowExpired.New()
 		}
+		ids = proto.Clone(gtw.Ids).(*ttnpb.GatewayIdentifiers)
```
adriansmares: `ids = ttnpb.Clone(gtw.Ids)`.

pkg/identityserver/storetest/gateway_store.go
```diff
@@ -458,6 +456,45 @@ func (st *StoreTest) TestGatewayStoreCRUD(t *T) {
...
+		a.So(deletedGtw.Ids, should.Resemble, gtwCreated.Ids)
+
+		// creates new gateway with same EUI.
+		newGtw, err := s.CreateGateway(ctx, &ttnpb.Gateway{
+			Ids: &ttnpb.GatewayIdentifiers{GatewayId: "bar", Eui: eui.Bytes()},
+		})
+		a.So(err, should.BeNil)
+		a.So(newGtw.Ids.Eui, should.Resemble, deletedGtw.Ids.Eui)
+
+		err = s.RestoreGateway(ctx, &ttnpb.GatewayIdentifiers{GatewayId: "foo", Eui: eui.Bytes()})
```
adriansmares: I believe this should work without the gateway EUI being provided on restore.

pkg/identityserver/gormstore/gateway_store.go
```diff
@@ -213,6 +213,20 @@ func (s *gatewayStore) DeleteGateway(ctx context.Context, id *ttnpb.GatewayIdent
...
+		query := s.query(ctx, Gateway{}, withGatewayEUI(EUI64(types.MustEUI64(id.Eui).OrZero())))
+		query = selectGatewayFields(ctx, query, []string{"ids.gateway_id", "ids.eui"})
+		var gtwModel Gateway
+		if err := query.First(&gtwModel).Error; err != nil {
+			if gorm.IsRecordNotFoundError(err) {
+				return s.restoreEntity(ctx, id)
+			}
+			return err
+		}
+		if gtwModel.GatewayID != id.GatewayId {
```
adriansmares: Is this the right thing to compare ? What if they are from different tenants with the same ID ?

pkg/identityserver/gormstore/gateway_store.go
```diff
@@ -213,6 +213,20 @@ func (s *gatewayStore) DeleteGateway(ctx context.Context, id *ttnpb.GatewayIdent
...
+			ctx = context.WithValue(ctx, cleanGtwEuiKey, true)
```
adriansmares: You can already update the EUI to `nil` here, right ? No need to do the context enrichment and do this in `restoreEntity`.

nicholaspcr: Removed it

pkg/identityserver/gormstore/gateway_store.go
nicholaspcr: True, ended up removing this and doing an `Unscoped` search for any gateway that contains the EUI. If more than one if found then the restored gateway will have its own EUI cleaned.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
+- Gateway EUI is no longer unset when deleting a gateway, meaning it could be recovered if no other gateway claimed it. This requires the `20220913000000_update_gtw_eui_index` migration to be executed in order to be functional.
```
adriansmares: Look into how we referenced this in the past changelog entries:
`- This requires a database schema migration (`ttn-lw-stack is-db migrate`) because of added columns and views.`

The migration names are not visible to users, nor relevant.

pkg/identityserver/storetest/gateway_store.go
```diff
@@ -458,6 +456,45 @@ func (st *StoreTest) TestGatewayStoreCRUD(t *T) {
...
+
+		// Deletes it.
+		err = s.DeleteGateway(ctx, &ttnpb.GatewayIdentifiers{GatewayId: "foo"})
+		a.So(err, should.BeNil)
+
+		deletedGtw, err := s.GetGateway(
+			store.WithSoftDeleted(ctx, true), &ttnpb.GatewayIdentifiers{GatewayId: "foo"}, mask,
+		)
+		a.So(err, should.BeNil)
+		// validates if deleted gateway has the same Eui as the new one.
```
adriansmares: Nit: Uppercase here and below. Also grammar.

adriansmares (approved):
Don't forget the Enterprise counter parts.

kschiffer (approved):

PR #5850, opened 2022-10-05
Add support for suppressing logs for selected errors on selected gRPC methods

### LINKED ISSUE #5803: Allow disabling RPC logs for explicit errors
### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/2863
References https://github.com/TheThingsNetwork/lorawan-stack/pull/5721

We need a mechanism to disable RPC logs even when the RPC call has returned an error.

### Current Situation

Currently we can disable RPC logs for individual RPCs, as long as no error has occurred (see the first referenced PR). The problem is that we would like to disable RPC logs even when an error is occurring, provided that we filter by the error.

Our production environments [avoid emitting RPC logs](https://github.com/TheThingsIndustries/lorawan-stack-aws/blob/adcd609cfe6cb7c7b2c16f75b6be6ed2f6599578/deploy/src/4-2a-configuration.py#L1624) for `/ttn.lorawan.v3.GsNs/HandleUplink`. With this change, we could drop logging for duplicate and unmatched traffic.

### Proposed Implementation

I propose that the semantics are changed a bit. Currently we have only the RPC full method (`/ttn.lorawan.v3.GsNs/HandleUplink`). I propose that we add some extra meaning:

- `/ttn.lorawan.v3.GsNs/HandleUplink` means that only when no error occurs the log is skipped.
- `/ttn.lorawan.v3.GsNs/HandleUplink:pkg/networkserver:duplicate_uplink` means that on `pkg/networkserver:duplicate_uplink` (the full name obtained via `Definition.FullName()`) the log is skipped.
- `/ttn.lorawan.v3.GsNs/HandleUplink:pkg/networkserver:duplicate_uplink;pkg/networkserver:device_not_found` means that on both `pkg/networkserver:device_not_found` and `pkg/networkserver:duplicate_uplink` the log is skipped.
- `/ttn.lorawan.v3.GsNs/HandleUplink:;pkg/networkserver:duplicate_uplink;pkg/networkserver:device_not_found` means that if there is no error, or the error is `pkg/networkserver:duplicate_uplink` / `pkg/networkserver:device_not_found` the log is skipped.

Right now we carry a set around (`map[string]struct{}`) - we probably want to carry a mapping between the full method and an error matching function.

### Contributing

### Code of Conduct

cvetkovski98:
#### Summary
Added support for disabling logs for specific errors that occur on specific gRPC method.

Closes #5803.

#### Changes
- extended the `--grpc.log-ignore-methods` option to accept error names in addition to the method name as described in #5803.
- if any error occurs that matches the selected error names, do not log anything

#### Testing
- UT for the option parser
- Tested locally for the actual log filtering

##### Regressions
There are not any

#### Notes for Reviewers
...

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
+- Option to ignore logs from selected gRPC methods now supports ignoring logs for selected errors on method.
+    Examples:
+    - `--grpc.log-ignore-methods="/ttn.lorawan.v3.GsNs/HandleUplink"`: log is skipped when no error occurs.
+    - `--grpc.log-ignore-methods="/ttn.lorawan.v3.GsNs/HandleUplink:pkg/networkserver:duplicate_uplink;pkg/networkserver:device_not_found"`: log is skipped when either `pkg/networkserver:duplicate_uplink` or `pkg/networkserver:device_not_found` error occurs (but not on success).
+    - `--grpc.log-ignore-methods="/ttn.lorawan.v3.GsNs/HandleUplink:;pkg/networkserver:duplicate_uplink"`: log is skipped on success or when `pkg/networkserver:duplicate_uplink` error occurs.
+
 ### Deprecated
 
 ### Removed
--- a/pkg/rpcmiddleware/rpclog/rpclog.go
+++ b/pkg/rpcmiddleware/rpclog/rpclog.go
@@ -118,3 +118,11 @@ func (l *ttnGrpcLogger) V(int) bool {
 func shouldSuppressError(err error) bool {
 	return errors.IsResourceExhausted(err)
 }
+
+func shouldSuppressLog(cfg methodLogConfig, err error) bool {
+	if err != nil {
+		wrapped, ok := errors.From(err)
+		return ok && cfg.shouldIgnorError(wrapped)
+	}
+	return cfg.IgnoreSuccess
+}
--- a/pkg/rpcmiddleware/rpclog/server_interceptors.go
+++ b/pkg/rpcmiddleware/rpclog/server_interceptors.go
@@ -36,11 +36,10 @@ func UnaryServerInterceptor(ctx context.Context, opts ...Option) grpc.UnaryServe
 		startTime := time.Now()
 		resp, err := handler(newCtx, req)
 
-		if err == nil {
-			if _, ok := o.ignoreMethods[info.FullMethod]; ok {
-				return resp, err
-			}
+		if cfg, ok := o.ignoreMethods[info.FullMethod]; ok && shouldSuppressLog(cfg, err) {
+			return resp, err
 		}
+
 		if shouldSuppressError(err) {
 			return resp, err
 		}
@@ -87,11 +86,10 @@ func StreamServerInterceptor(ctx context.Context, opts ...Option) grpc.StreamSer
 		startTime := time.Now()
 		err := handler(srv, wrapped)
 
-		if err == nil {
-			if _, ok := o.ignoreMethods[info.FullMethod]; ok {
-				return err
-			}
+		if cfg, ok := o.ignoreMethods[info.FullMethod]; ok && shouldSuppressLog(cfg, err) {
+			return err
 		}
+
 		if shouldSuppressError(err) {
 			return err
 		}
--- a/pkg/rpcmiddleware/rpclog/utils.go
+++ b/pkg/rpcmiddleware/rpclog/utils.go
@@ -107,3 +107,24 @@ func commit(i log.Interface, level log.Level, msg string) {
 		panic(fmt.Sprintf("rpclog: unknown log level %d", level))
 	}
 }
+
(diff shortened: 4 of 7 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Changed
 
+- Option to ignore logs from selected gRPC methods now supports ignoring logs for selected errors on method.
```
adriansmares, line 16: We may want to provide a minimized example here on how things should look like.

pkg/rpcmiddleware/rpclog/server_interceptors.go
```diff
@@ -36,11 +37,22 @@ func UnaryServerInterceptor(ctx context.Context, opts ...Option) grpc.UnaryServe
 		startTime := time.Now()
 		resp, err := handler(newCtx, req)
 
-		if err == nil {
-			if _, ok := o.ignoreMethods[info.FullMethod]; ok {
-				return resp, err
+		if methodLogCfg, ok := o.ignoreMethods[info.FullMethod]; ok {
+			// There is a log configuration for this method.
```
adriansmares: The comments describe the code here, so they may be elided.

pkg/rpcmiddleware/rpclog/server_interceptors.go
```diff
@@ -36,11 +37,22 @@ func UnaryServerInterceptor(ctx context.Context, opts ...Option) grpc.UnaryServe
 		startTime := time.Now()
 		resp, err := handler(newCtx, req)
 
-		if err == nil {
-			if _, ok := o.ignoreMethods[info.FullMethod]; ok {
-				return resp, err
+		if methodLogCfg, ok := o.ignoreMethods[info.FullMethod]; ok {
```
adriansmares: Nit: just `cfg` suffices.

pkg/rpcmiddleware/rpclog/server_interceptors.go
```diff
@@ -36,11 +37,22 @@ func UnaryServerInterceptor(ctx context.Context, opts ...Option) grpc.UnaryServe
...
+			// There is a log configuration for this method.
+			if err != nil {
+				wrapped, ok := errors.From(err)
+				if ok && methodLogCfg.shouldIgnorError(wrapped) {
```
adriansmares: You can do `if wrapped, ok := errors.From(err); ok && cfg.shouldIgnore(wrapped) {`.

pkg/rpcmiddleware/rpclog/utils.go
```diff
@@ -107,3 +107,24 @@ func commit(i log.Interface, level log.Level, msg string) {
...
+	isSuccessIgnored := ignoredErrors[0] == ""
+	if isSuccessIgnored {
+		ignoredErrors = ignoredErrors[1:]
+	}
+	ignoredErrorSet := make(map[string]struct{}, len(ignoredErrors))
+	for _, ignoredMethodError := range ignoredErrors {
+		ignoredErrorSet[ignoredMethodError] = struct{}{}
+	}
+	return methodName, methodLogConfig{
+		IsSuccessIgnored: isSuccessIgnored,
```
adriansmares: Nit: `IgnoreSuccess`.

pkg/rpcmiddleware/rpclog/server_interceptors.go
cvetkovski98: If done like that, in the case where: `IgnoreSuccess == true` and we have an error that shouldn't be ignored, I believe the check will fail and the log will be suppressed?

The change `if err != nil {` -> `if wrapped, ok := errors.From(err); ok {` should be fine tho. What do you think?

adriansmares: I didn't refer to merging the `if`s, just having the second `if` have the declarations merged into its body, since neither `ok` nor `wrapped` are used outside of the second `if`s body.

cvetkovski98: Got it, my bad.

pkg/rpcmiddleware/rpclog/rpclog.go
```diff
@@ -118,3 +118,12 @@ func (l *ttnGrpcLogger) V(int) bool {
 func shouldSuppressError(err error) bool {
 	return errors.IsResourceExhausted(err)
 }
+
+func shouldSuppressLog(cfg methodLogConfig, err error) bool {
+	if err != nil {
+		wrapped, ok := errors.From(err)
+		return ok && cfg.shouldIgnorError(wrapped)
+	} else {
```
adriansmares: Return without else.

CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ For details about compatibility between different releases, see the **Commitment
...
+    Examples:
+    - `--grpc.log-ignore-methods="/ttn.lorawan.v3.GsNs/HandleUplink"`: log is skipped when no error occurs
```
adriansmares: Nit: dot at the end of the sentences.

adriansmares (approved):

PR #5858, opened 2022-10-06
Fix Application Server panic on filter without identifiers

adriansmares:
#### Summary

This quickfix ensures that the end device identifiers are still available after we filter out the webhook fields. Otherwise downlink header generation fails (since it needs device identifiers) and panics.

#### Changes

- Execute webhooks in tasks to avoid stopping the full process.
- Save the identifiers before the field mask is applied.

#### Testing

Local.

##### Regressions

N/A.

#### Notes for Reviewers

@happyRip please review.

#### Checklist


happyRip (approved):
Looks good

PR #5862, opened 2022-10-10
BasicStation dutycycle alignment fixes

### LINKED ISSUE #5844: BasicStation duty cycle computation mismatch
### Summary

References https://github.com/lorabasics/basicstation/issues/164

BasicStation computes the sub band utilization in a different way compared to our sub band scheduler. This means that downlinks that we consider to be legal (i.e. within the duty cycle usage requirement), can be silently rejected by BasicStation.

[Production builds of BasicStation cannot be instructed to disable duty cycle checks](https://github.com/lorabasics/basicstation/blob/ba4f85d80a438a5c2b659e568cd2d0f0de08e5a7/src-linux/sys_linux.c#L755-L764). This means that we cannot just tell BasicStation to 'trust the LNS'.

### Current Situation

Downlink scheduling is a multi component affair. Fundamentally it goes between NS, GS and finally the packet forwarder. The flow is as follows:
1. NS tries to schedule a downlink on the GS.
2. GS accepts or rejects it. If rejected, go to 1 and try different gateway.
3. GS forwards it to the packet forwarder. If ack/nack, signal this to the LNS. This is just passed to the application layer, no retries are done (https://github.com/TheThingsNetwork/lorawan-stack/issues/18).

There are nuances to this. BasicStation does not actually have a nack message - only acknowledgements, and the ack is sent at the actual transmission time (so 5 seconds after the uplink for RX1=5), so even in the absence of the acknowledgement we cannot actually try to schedule again (since the window is closed).

Since BasicStation is non-cooperative in this regard, we need the GS to be 'as sure as possible' that if it schedules a downlink, the forwarder will accept it too. Otherwise the GS accepts it, the NS thinks that things are done, and then the forwarder silently drops it.

One way packet forwarder scheduling can fail is duty cycle. As described in the summary, there are actual differences between how we compute duty cycle and how BasicStation does it.

[BasicStation does not have rolling windows at all](https://github.com/lorabasics/basicstation/issues/164). BasicStation computes air time * 100/duty-cycle and considers that the next moment in time after which transmissions can take place. Any downlink between the current time and that point in the future is discarded silently. 

Consider the following scenario.
1. The current time is 00:00:00 (hh:mm:ss).
2. We want to schedule a downlink whose air time is 0.2 seconds, in a band with 1% duty cycle.

We compute 0.2 seconds * 100/1 = 20 seconds. This means that the next downlink in that particular sub band can only be transmitted after 00:00:20. Our scheduler is a bit more lenient - it will allow downlinks in this 20 seconds period, if the overall utilization in the rolling hour is under 1%.

Environments in which confirmed downlinks are used are affected by this issue the most. Due to the unfortunately symmetrical nature of this setup (every uplink airtime generates some downlink airtime), it is annoyingly easy to hit the duty cycle of a sub band.

In such situations, the GS will schedule (accept) the downlink, but the forwarder will silently drop it. This is very bad because generally there is overlapping coverage between multiple gateways, so if the GS would actually reject the scheduling attempt, the NS has the chance to try another gateway (which may not be duty cycled).

Confirmed downlinks are not the only ones that will hit this. Having duty cycle windows actually allows short bursts to occur - these are actually common ! Think of the initial moments of a session - you have a downlink for the join accept, then one or two downlinks for extra MAC commands (or even application layer setup). These are usually done in a short time frame, shorter than the Basic Station style allows.

### Proposed Implementation

Even if BasicStation would receive a bucket update, and we would align bucket sizes, most of the gateways in the field will not be upgraded for long periods of time. I think we need to accept this as a reality, and design and implement this in a pragmatic manner.

1. As part of `io.Frontend`, expose a `DutyCycleStyle() DutyCycleStyle` function that tells the GS which sub band scheduler to use. Valid styles are `Agnostic` (UDP, MQTT and gRPC are like this, since they don't do their own duty cycling) and `Blocking` (the BasicStation style mentioned above).
2. Abstract away the sub band scheduler under an interface, and turn the current `scheduling.SubBand` into a `scheduling.bucketSubBand`.
3. Introduce `scheduling.blockingSubBand` that follows the BasicStation rules. Statistics such as utilization are still based on hourly buckets, but scheduling attempts are computed in this 'sub band blocking' style.

### Contributing

### Code of Conduct

adriansmares:
#### Summary

Closes #5844 

#### Changes

- Introduce the concept of dutycycle styles for the Gateway Server sub band scheduler.
  - We have two styles - window (what we have been using so far) and blocking window.
  - The blocking window style adds an additional check that ensures that BasicStation-based forwarders do not reject downlinks, since in BasicStation duty cycle is enforced by 'blocking' the sub band for the inverse duration of the duty cycle. What this 'inverse' means in plain words is that a 1 second transmission keeps a 1% duty cycle sub band blocked for 100 seconds, or for 10 seconds in a sub band with 10% duty cycle.

#### Testing

Unit testing + `staging1` testing.

##### Regressions

Users will now observe that we do not schedule downlinks on their duty-cycled Basic Station gateways which is an event visible in the Console. This may seem to them as a regression, but note that Basic Station would previously just silently drop the downlink. At least now the Network Server has the chance to try another gateway.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/scheduling/sub_band.go
+++ b/pkg/gatewayserver/scheduling/sub_band.go
@@ -117,6 +117,20 @@ func (sb *SubBand) prioritizedDutyCycle(p ttnpb.TxSchedulePriority) float32 {
 	return sb.DutyCycle * ceiling
 }
 
+// computeWindowUsage verifies if the emission will not exceed the usable amount of the duty cycle in the
+// [t + toa - window, t + toa] and [t, t + window] windows, where `t` is the start of the emission and
+// `toa` is the duration of the emission.
+// This method requires that the read lock is held.
+func (sb *SubBand) computeWindowUsage(em Emission, usable float32) (used float32, ok bool) {
+	for _, to := range []ConcentratorTime{em.Ends(), em.t + ConcentratorTime(DutyCycleWindow)} {
+		used := float32(sb.sum(to-ConcentratorTime(DutyCycleWindow), to)+em.d) / float32(DutyCycleWindow)
+		if used > usable {
+			return used, false
+		}
+	}
+	return 0, true
+}
+
 var errDutyCycle = errors.DefineResourceExhausted("duty_cycle", "utilization `{used}%` would be higher than the available `{usable}%` for priority `{priority}`")
 
 // Schedule schedules the given emission with the priority.
@@ -129,16 +143,12 @@ func (sb *SubBand) Schedule(em Emission, p ttnpb.TxSchedulePriority) error {
 		return nil
 	}
 	usable := sb.prioritizedDutyCycle(p)
-	// Check the window before and after the emission for availability.
-	for _, to := range []ConcentratorTime{em.Ends(), em.t + ConcentratorTime(DutyCycleWindow)} {
-		used := float32(sb.sum(to-ConcentratorTime(DutyCycleWindow), to)+em.d) / float32(DutyCycleWindow)
-		if used > usable {
-			return errDutyCycle.WithAttributes(
-				"used", fmt.Sprintf("%.1f", used*100),
-				"usable", fmt.Sprintf("%.1f", usable*100),
-				"priority", fmt.Sprintf("%v", p),
-			)
-		}
+	if used, ok := sb.computeWindowUsage(em, usable); !ok {
+		return errDutyCycle.WithAttributes(
+			"used", fmt.Sprintf("%.1f", used*100),
+			"usable", fmt.Sprintf("%.1f", usable*100),
+			"priority", fmt.Sprintf("%v", p),
+		)
 	}
 	sb.emissions = sb.emissions.Insert(em)
 	return nil
@@ -165,13 +175,7 @@ func (sb *SubBand) ScheduleAnytime(d time.Duration, next func() ConcentratorTime
 		)
 	}
 	for {
-		conflicts := false
-		// Check the window before and after the emission for availability.
-		for _, to := range []ConcentratorTime{em.Ends(), em.t + ConcentratorTime(DutyCycleWindow)} {
-			sum := float32(sb.sum(to-ConcentratorTime(DutyCycleWindow), to)+em.d) / float32(DutyCycleWindow)
-			conflicts = conflicts || sum > usable
-		}
-		if !conflicts {
+		if _, ok := sb.computeWindowUsage(em, usable); ok {
 			break
 		}
 		if t := next(); t != em.t {
--- a/pkg/gatewayserver/scheduling/sub_band.go
+++ b/pkg/gatewayserver/scheduling/sub_band.go
@@ -124,18 +124,20 @@ var errDutyCycle = errors.DefineResourceExhausted("duty_cycle", "utilization `{u
 func (sb *SubBand) Schedule(em Emission, p ttnpb.TxSchedulePriority) error {
 	sb.mu.Lock()
 	defer sb.mu.Unlock()
-	if sb.DutyCycle < 1 {
-		usable := sb.prioritizedDutyCycle(p)
-		// Check the window before and after the emission for availability.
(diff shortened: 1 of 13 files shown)


adriansmares:
@KrishnaIyer - could you please take a look here ? I would like to include this in `3.22.1`.

pkg/gatewayserver/scheduling/sub_band.go
```diff
@@ -117,26 +135,74 @@ func (sb *SubBand) prioritizedDutyCycle(p ttnpb.TxSchedulePriority) float32 {
...
+// checkDutyCycle verifies if the emission complies with the duty cycle limitations, based on the style.
+//
+// For the simple window style, it verifies if the emission will not exceed the usable amount of the duty
+// cycle in the [t + toa - window, t + toa] and [t, t + window] windows, where `t` is the start of the
+// emission and `toa` is the duration of the emission.
+//
+// For the blocking window style, in addition to the above mentioned check it verifies if the inverse of
+// the duty cycle has passed from the last emission.
+// Given a duty cycle of p%, and an emission of d time units at timestamp t, the duty cycle check will
+// allow the next transmission to occur only after the t + d * (100%/p%) timestamp.
```
KrishnaIyer, line 158: shouldn't this be ` t + d * ((100%/p%) -1)`?

adriansmares: https://github.com/lorabasics/basicstation/blob/ba4f85d80a438a5c2b659e568cd2d0f0de08e5a7/src/s2e.c#L208-L212
https://github.com/lorabasics/basicstation/blob/ba4f85d80a438a5c2b659e568cd2d0f0de08e5a7/src/s2e.c#L375
https://github.com/lorabasics/basicstation/blob/ba4f85d80a438a5c2b659e568cd2d0f0de08e5a7/src/s2e.c#L-479-L480

The way I interpret it is that if a duration `d` can block the air only 1% of the time, the rest of 99% of the time should not be used either (in order to ensure that in this period we only used the air time only 1% of the time).

KrishnaIyer: The way I'm thinking is; ON + OFF makes the total duty cycle. With an example duty cycle of 1%, the radio is ON (transmitting) for 1% then it's OFF (blocked) for 99%. 
If we take a simple time period of 100us, then it's 1us ON and 99us OFF.
So if a transmission takes 3us (`d`) , then the waiting time would be 3 x 99us right? 
I'm thinking that we should account for the ON time (d) also in the waiting period.

adriansmares: If the transmission takes 3us at t=5us, then the waiting time should be 3 * 100 = 300us. No new transmissions until t=305us. In the [5us, 305us] we have used only 3us of airtime, which is 1%, and this is how BasicStation validates dutycycle - there are no sample windows.

But I think that the confusion comes here from this `t` - `t` is the timestamp _at the start_ of the transmission, not the end of it. So if we block `[t, t + d * (100%/dc%)]`, we did take the duration into consideration.

We are sort of saying the same thing here - you say that we should wait 297us after the transmission has ended, I am saying we should wait 300us from the start of the transmission, which for a 3us transmission means that we should wait 297us after the end of the transmission (the same value as yours).

KrishnaIyer: > t is the timestamp at the start of the transmission, not the end of it. So if we block [t, t + d * (100%/dc%)], we did take the duration into consideration.

Ok I see it now, this is indeed correct.

KrishnaIyer (approved):
Implementation LGTM.
> Users will now observe that we do not schedule downlinks on their duty-cycled Basic Station gateways which is an event visible in the Console. This may seem to them as a regression, but note that Basic Station would previously just silently drop the downlink. At least now the Network Server has the chance to try another gateway.

Can you add (a brief version of) this note to the Changelog and/or docs?

PR #5871, opened 2022-10-13
Revert `CodingRate` removal

ysmilda:
#### Summary

Quickfix PR to revert the removal of the `CodingRate` from the `TxSettings` message and move its deprecation to `v3.23.0`.

#### Changes

- Revert removal of `CodingRate` field from `TxSettings` message.

#### Testing

UT

#### Checklist


johanstokking:
Although this is great and correct, I think it's too late to merge this.

@adriansmares what do you think?

adriansmares:
I think we should just be really mindful about this going forward, and adding it back is two weeks late now. Applications that were broken are already fixed, and at this point we’re adding it back, just to take it out again in another minor. 99.999999% of our users do not actually use the proto/generate proto API so they don’t know that the `TxSettings` field is deprecated. They will see it there and accidentally use it. Then we’ll manage to let them down twice with the same field once we remove it.

Going forward deletions of external facing API should just be skipped period, and we’ll have to think of how do we want to fill those (most likely at AS level only - no point in carrying these values everywhere).

johanstokking:
Yep agreed. At some point we probably need to maintain a copy of the `ApplicationUp` data structure tree just to ensure backward compatibility. Lessons learned, let's not bother with reverting it now.

PR #5884, opened 2022-10-20
Store Redis events with jittered TTL

adriansmares:
#### Summary

This PR ensures that event storage TTLs are de-correlated in order to avoid 'expiration bursts'. This is achieved by adding a 1% jitter to the event history and payload TTL.

#### Changes

- Add jitter to the event storage TTL, in order to ensure that events don't correlate accidentally.
- Add jitter to the blocking period of the event subscriptions done via `XRead`. This ensures that multiple streams reconnecting at the same time do not become correlated over time.
- Expire keys using `PEXPIRE` instead of `EXPIRE`. This has the advantage of having millisecond instead of second accuracy, thus reducing the correlation between events.

#### Testing

UT.

##### Regressions

The history TTL (i.e. duration for which the event metadata, but not data, is visible) is 24 hours by default. 1% jitter for 24 hours is about 14.4 minutes - some events will expire a little earlier, some a little later. I think this is reasonable.

For the event data storage, which is only 15 minutes by default, 1% jitter is 9 seconds. Again, this should be reasonable.

#### Notes for Reviewers

The background here is that operations that touch the event storage Redis cluster inherently cause some events to become correlated over time. This happens via the following sequence of events:

1. The cluster is unavailable for a short while (say due to an infrastructure upgrade) - can be in the tens of seconds.
2. Some events are queued to be stored - TTL is relative (i.e. it is a duration, not a time in the future) so when all of them are stored, all of their expiration times are also virtually equal.
3. Redis has to expire these keys in the future, so it has to keep the cluster busy while deleting the keys. From the perspective of the application this looks like the downtime described in (1).
4. If more keys are queued than deleted, over time this will cause Redis to correlate more and more events.

You may wonder 'why does Redis need to block while doing this' and the answer is that normally you don't have to block, but in order to ensure that replicas also have the same expired keys, Redis will synthesize `DEL` commands in order to ensure that the keys also expired on the replica:
image

#### Checklist

### DIFF
--- a/pkg/events/redis/store.go
+++ b/pkg/events/redis/store.go
@@ -26,6 +26,7 @@ import (
 	"github.com/go-redis/redis/v8"
 	"go.thethings.network/lorawan-stack/v3/pkg/events"
 	"go.thethings.network/lorawan-stack/v3/pkg/log"
+	"go.thethings.network/lorawan-stack/v3/pkg/random"
 	ttnredis "go.thethings.network/lorawan-stack/v3/pkg/redis"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 	"go.thethings.network/lorawan-stack/v3/pkg/unique"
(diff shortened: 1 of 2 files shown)


pkg/events/redis/store.go
```diff
@@ -298,49 +302,53 @@ func (ps *PubSubStore) SubscribeWithHistory(
...
-			return ctx.Err()
-		default:
-			args := &redis.XReadArgs{
-				Count: 10,
-				Block: time.Second,
+		args := &redis.XReadArgs{
+			Count: eventCountLimit,
+			Block: random.Jitter(8*time.Second, 0.2),
+		}
+		for _, s := range state {
```
adriansmares, line 319: The diff here looks bigger than it is. I have removed the outer `select` (since `XRead` will return `ctx.Err()` on `ctx.Done()`), but the only changes are to `args.Count` and `args.Block`.

johanstokking (approved):

KrishnaIyer (approved):

PR #5887, opened 2022-10-23
Use a thread safe `sync.Map` to store rights

### LINKED ISSUE #5534: Ttn-lw-cli applications subscribe HANG
#### Summary
When I execute **ttn-lw-cli events --gateway_id [_whatever_]**
the cli will (as expected) run until the access token expires and returns to the caller.

When I’m not authenicated I can just do **ttn-lw-cli login** to get a new access token.
That all works fine :smiley:

However, when I execute **ttn-lw-cli applications subscribe --application-id [_whatever_]**
the cli (unexpectedly) will not return after the access token has expired :frowning:

#### Steps to Reproduce
The symptom I see is that I get the desired application output data while the access token is valid but no clue that the token has expired.
Visually it is as if the device had stopped sending data to the application, but it has been broadcasting fine, as witnessed by the gateway events.

I can use **--log.level debug** to get an indication of when the token will expire, along the lines of
`DEBUG Using access token (valid until 8:06AM)`

… but I would expect the **ttn-lw-cli** to actually terminate at 8:06AM and not just hang!

#### What do you see now?
I feel ths behaviour is inconsistent with **ttn-lw-cli events** which sympathetically returns when the token expires.

If I have a script wrapping **ttn-lw-cli** I don't want it to hang and never return.
Once the access token has expired there is no further data retuned from the subscription.
However this data stream stops silently :frowning:

#### Environment
Version:             3.19.2
Build date:          2022-05-25T07:14:30Z
Git commit:          2dfab4a75
Go version:          go1.18.2
OS/Arch:             windows/amd64

...

I'm not in a position to fix it myself.
I currently have a workaround that has another process that monitors the script running **ttn-lw-cli**
When the token has expired I  kill the hung  **ttn-lw-cli**
I determine the expiry timestamp based on scanning the debug output

...

It should be simple to determine if the ttn-lw-cli unblocks after (typically) 1 hour allowing the wrapper script to continue.
All being well I would not need for another process to kill **ttn-lw-cli**

...

I'm not in a position to fix it myself.

happyRip:
#### Summary

References #5534 

#### Changes

- Use `sync.Map` to store rights (instead of `map[string]*ttnpb.Rights`)

#### Testing

- Added unit tests for `rights.Map`

##### Regressions

Using `rights.Map` requires to use it's interface instead of relying on a native `map`.

#### Notes for Reviewers

nil

#### Checklist

### DIFF
--- a/pkg/auth/rights/rights.go
+++ b/pkg/auth/rights/rights.go
@@ -147,15 +147,15 @@ type rightsKeyType struct{}
 
 var rightsKey rightsKeyType
 
-func fromContext(ctx context.Context) (Rights, bool) {
-	if rights, ok := ctx.Value(rightsKey).(Rights); ok {
+func fromContext(ctx context.Context) (*Rights, bool) {
+	if rights, ok := ctx.Value(rightsKey).(*Rights); ok {
 		return rights, true
 	}
-	return Rights{}, false
+	return &Rights{}, false
 }
 
 // NewContext returns a derived context with the given rights.
-func NewContext(ctx context.Context, rights Rights) context.Context {
+func NewContext(ctx context.Context, rights *Rights) context.Context {
 	return context.WithValue(ctx, rightsKey, rights)
 }
 
@@ -175,11 +175,11 @@ func cacheInContext(ctx context.Context, f func(*Rights)) {
 	}
 }
 
-func cacheFromContext(ctx context.Context) (Rights, bool) {
+func cacheFromContext(ctx context.Context) (*Rights, bool) {
 	if rights, ok := ctx.Value(rightsCacheKey).(*Rights); ok {
-		return *rights, true
+		return rights, true
 	}
-	return Rights{}, false
+	return &Rights{}, false
 }
 
 type authInfoKeyType struct{}
--- a/pkg/auth/rights/rights.go
+++ b/pkg/auth/rights/rights.go
@@ -28,8 +28,8 @@ type Map struct {
 }
 
 // NewMap ...
-func NewMap(rights map[string]*ttnpb.Rights) Map {
-	m := Map{}
+func NewMap(rights map[string]*ttnpb.Rights) *Map {
+	m := &Map{}
 	for k, v := range rights {
 		m.SetRights(k, v)
 	}
--- a/pkg/auth/rights/rights.go
+++ b/pkg/auth/rights/rights.go
@@ -17,106 +17,129 @@ package rights
 
 import (
 	"context"
+	"sync"
 
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 )
 
+// Map ...
+type Map struct {
+	syncMap sync.Map
+}
+
+// NewMap ...
+func NewMap(rights map[string]*ttnpb.Rights) Map {
+	m := Map{}
+	for k, v := range rights {
+		m.SetRights(k, v)
+	}
+	return m
+}
+
+// SetRights sets the rights for the given UID.
+func (m *Map) SetRights(uid string, rights *ttnpb.Rights) {
+	m.syncMap.Store(uid, rights)
+}
+
+// GetRights returns the rights stored in the map for a given UID,
+// or nil if no value is present.
+// The ok result indicates whether value was found in the map.
+func (m *Map) GetRights(uid string) (*ttnpb.Rights, bool) {
+	if r, ok := m.syncMap.Load(uid); ok {
+		return r.(*ttnpb.Rights), ok
+	}
+	return nil, false
+}
+
+// MissingRights returns the rights that are missing for the given RightsMap.
+func (m *Map) MissingRights(uid string, rights ...ttnpb.Right) []ttnpb.Right {
+	if r, ok := m.GetRights(uid); ok {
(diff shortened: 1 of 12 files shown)


pkg/auth/rights/rights.go
```diff
@@ -17,106 +17,132 @@ package rights
 
 import (
 	"context"
+	"sync"
 
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 )
 
+// RightsMap ...
+type RightsMap struct {
```
adriansmares: Consider renaming to `Map`.

pkg/auth/rights/rights.go
```diff
@@ -17,106 +17,132 @@ package rights
...
+	syncMap *sync.Map
+}
+
+// NewRightsMap ...
+func NewRightsMap(uid string, rights *ttnpb.Rights) RightsMap {
```
adriansmares: Consider taking the old `map[string]*ttnpb.Rights` as an argument - then the updated sites will be better.

pkg/joinserver/grpc_deviceregistry_test.go
```diff
@@ -720,14 +695,13 @@ func TestDeviceRegistrySet(t *testing.T) {
...
-					ApplicationRights: map[string]*ttnpb.Rights{
-						unique.ID(test.Context(), &ttnpb.ApplicationIdentifiers{
-							ApplicationId: registeredApplicationID,
-						}): ttnpb.RightsFrom(
+					ApplicationRights: rights.NewRightsMap(
+						unique.ID(test.Context(), &ttnpb.ApplicationIdentifiers{ApplicationId: registeredApplicationID}),
+						ttnpb.RightsFrom(
 							ttnpb.Right_RIGHT_APPLICATION_DEVICES_WRITE,
-							ttnpb.Right_RIGHT_APPLICATION_DEVICES_WRITE_KEYS,
+							ttnpb.Right_RIGHT_APPLICATION_DEVICES_READ_KEYS,
```
adriansmares: The test fails because you changed the right here.

pkg/auth/rights/rights.go
```diff
@@ -17,106 +17,132 @@ package rights
...
 import (
 	"context"
+	"sync"
 
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 )
 
+// RightsMap ...
+type RightsMap struct {
+	syncMap *sync.Map
```
adriansmares: Instead of making `syncMap` a pointer, make `NewContext` and `FromContext` carry the `*Rights`

pkg/auth/rights/rights.go
happyRip: Renamed to `Map`

pkg/auth/rights/rights.go
happyRip: Fixed

pkg/joinserver/grpc_deviceregistry_test.go
happyRip: Fixed

pkg/auth/rights/rights.go
happyRip: Fixed https://github.com/TheThingsNetwork/lorawan-stack/pull/5887/commits/a9f7db7b915b193cddabf450fd57c84a2d05e886

adriansmares (approved):

KrishnaIyer (approved):
Code owned LGTM

pkg/auth/rights/rights.go
```diff
@@ -17,122 +17,145 @@ package rights
...
-	ApplicationRights  map[string]*ttnpb.Rights
-	ClientRights       map[string]*ttnpb.Rights
-	GatewayRights      map[string]*ttnpb.Rights
-	OrganizationRights map[string]*ttnpb.Rights
-	UserRights         map[string]*ttnpb.Rights
+	ApplicationRights  Map
+	ClientRights       Map
+	GatewayRights      Map
+	OrganizationRights Map
+	UserRights         Map
```
johanstokking, line 68: Is there a reason to store this by value and not by reference?

pkg/auth/rights/rights_test.go
```diff
@@ -19,6 +19,7 @@ import (
 
 	"github.com/smartystreets/assertions"
 	"github.com/smartystreets/assertions/should"
+
```
johanstokking, line 22: Remove empty line

pkg/auth/rights/rights.go
adriansmares: In order for the zero value (`&Rights{}`) to not require any additional initialization in places like `setApplicationRights`, `setClientRights` and so forth, which would race the initialization. 

Alternatively, we can have a factory function that takes these individual `*Map` and we expose only an interface, if that seems more appropriate to you.

pkg/auth/rights/rights_test.go
happyRip: Fixed

pkg/auth/rights/rights.go
johanstokking: OK, I see

johanstokking (approved):

Issue #3686, opened 2021-01-20
Support dynamic class B ping slot frequency


johanstokking:
#### Summary

Support dynamic class B ping slot frequency.

References https://github.com/TheThingsNetwork/lorawan-devices/pull/38#discussion_r555044700

#### Steps to Reproduce

Onboard a device with, for example, band CN470 that supports class B.

#### What do you see now?

There's no fixed ping slot frequency, which is a required value by NS. It is not possible to specify empty;

```json
{
  "code": 3,
  "message": "error:pkg/networkserver:field_mask (invalid field mask)",
  "details": [
    {
      "@type": "type.googleapis.com/ttn.lorawan.v3.ErrorDetails",
      "namespace": "pkg/networkserver",
      "name": "field_mask",
      "message_format": "invalid field mask",
      "correlation_id": "2b2b2538c031439d899f8793d604b3f8",
      "cause": {
        "namespace": "pkg/ttnpb",
        "name": "missing_field",
        "message_format": "field `{field}` is missing",
        "attributes": {
          "field": "mac_settings.ping_slot_frequency.value"
        },
        "code": 2
      },
      "code": 3
    }
  ]
}
```

#### What do you want to see instead?

This should be dynamic and depending on the band.

#### How do you propose to implement this?

Support dynamic ping slot frequency as defined in LoRaWAN Regional Parameters based on the band; CN470, US915 and AU915 seem to be affected.

#### How do you propose to test this?

Test creating devices.

#### Can you do this yourself and submit a Pull Request?

Can review

rvolosatovs:
Indeed, didn't see this in the spec before.
We'll have to do some refactoring to allow for this.

Also, how would this work for ABP? I suppose we will still have to force `mac_settings.ping_slot_frequency.value` to be set for those devices

johanstokking:
> Also, how would this work for ABP? I suppose we will still have to force `mac_settings.ping_slot_frequency.value` to be set for those devices

I guess so, yes.

janakj:
I just came across what I believe to might be the same or similar issue. I am in US915 and when I enable class B for my device and configure beacon and ping slot frequency to 923.3 MHz manually, I get the following error:

![Screen Shot 2022-04-04 at 01 42 13](https://user-images.githubusercontent.com/574979/161481025-33042ac3-0a8b-4925-8b4f-9eaba41c4ced.png)

Is there any way to enable class B in US915 with TTN v3.18.2?

PR #5892, opened 2022-10-26
Allow zero frequencies as part of the MAC settings

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/3686
References https://github.com/TheThingsIndustries/lorawan-stack-support/issues/872

The backend bug fixed here is that we return a default value(`mac_settings.ping_slot_frequency` of `0`) in US915, which cannot be used to create the end device (since `mac_settings.ping_slot_frequency` was previously a `FrequencyValue`, which does not allow zeroes).

#### Changes

- Add a new wrapper called `ZeroableFrequencyValue` - it is both binary and JSON compatible with `FrequencyValue`, but it allows zero values to be encoded.
  - This is required since in US915 and AU915 both the ping slot frequency and beacon frequencies may be `0`. The meaning of `0` is 'hop between 8 predefined frequencies'.
  - This allows users to explicitly request the hopping behavior, and ensures that the default MAC settings that we return actually can be used to create end devices.

#### Testing

Unit testing + tried to create a multicast US915 device as described in the original issue.

##### Regressions

N/A. The changes are binary compatible and JSON compatible.

#### Checklist


pkg/devicerepository/store/profiles.go
```diff
@@ -125,7 +125,7 @@ func (p EndDeviceProfile) ToTemplatePB(ids *ttnpb.EndDeviceVersionIdentifiers, i
 		paths = append(paths, "mac_settings.ping_slot_data_rate_index")
 	}
 	if p.PingSlotFrequency > 0 {
```
adriansmares, line 127: This is not ideal because it does not differentiate between 'unknown' and 0. It would have been great if this was nullable in the device repository I think.

johanstokking (approved):

PR #5894, opened 2022-10-27
Fix gateway stats debouncing doubling behavior

adriansmares:
#### Summary

This PR fixes the following double stats emission event:
image

It also removes the gateway EUI from the gateway downlink scheduling event:
image

#### Changes

- Fix timestamp saving during gateway stats debounce.
- Remove gateway EUI from downlink scheduling events.
- Fix modification of frequency plans as part of scheduler initialization.

#### Testing

Unit tests.

##### Regressions

N/A.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/gatewayserver.go
+++ b/pkg/gatewayserver/gatewayserver.go
@@ -952,14 +952,15 @@ func (gs *GatewayServer) updateConnStats(ctx context.Context, conn connectionEnt
 			DisconnectedAt: ttnpb.ProtoTimePtr(time.Now()),
 		}
 		registerGatewayConnectionStats(decoupledCtx, ids, stats)
-		if gs.statsRegistry != nil {
-			if err := gs.statsRegistry.Set(
-				decoupledCtx, ids, stats,
-				[]string{"connected_at", "disconnected_at"},
-				gs.config.ConnectionStatsDisconnectTTL,
-			); err != nil {
-				logger.WithError(err).Warn("Failed to clear connection stats")
-			}
+		if gs.statsRegistry == nil {
+			return
+		}
+		if err := gs.statsRegistry.Set(
+			decoupledCtx, ids, stats,
+			[]string{"connected_at", "disconnected_at"},
+			gs.config.ConnectionStatsDisconnectTTL,
+		); err != nil {
+			logger.WithError(err).Warn("Failed to clear connection stats")
 		}
 	}()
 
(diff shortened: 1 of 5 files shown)


pkg/gatewayserver/gatewayserver.go
```diff
@@ -982,22 +982,23 @@ func (gs *GatewayServer) updateConnStats(ctx context.Context, conn connectionEnt
...
+		if wait := gs.config.UpdateConnectionStatsDebounceTime - time.Since(lastUpdate); wait > 0 {
 			duration := random.Jitter(wait, debounceJitter)
 			select {
 			case <-ctx.Done():
 				return
 			case <-time.After(duration):
 			}
 		}
-		lastUpdate = now
+		lastUpdate = time.Now()
```
adriansmares, line 993: Saving `now` was wrong, since `now` is the timestamp before the wait. If a stats update came during the wait time, the whole debouncing would be skipped, since `now.Sub(lastUpdate)` would basically equal the last wait time (and thus wait - wait = 0).

pkg/gatewayserver/scheduling/scheduler.go
```diff
@@ -102,21 +102,21 @@ func NewScheduler(
...
 
-	var toa *frequencyplans.TimeOffAir
+	var timeOffAir *frequencyplans.TimeOffAir
 	for _, fp := range fps {
-		if toa != nil && fp.TimeOffAir != *toa {
+		if timeOffAir != nil && fp.TimeOffAir != *timeOffAir {
 			return nil, errFrequencyPlansTimeOffAir.New()
 		}
-		toa = &fp.TimeOffAir
+		timeOffAir = fp.TimeOffAir.Clone()
```
adriansmares, line 110: Since `toa` was a pointer, clamping it with `QueueDelay` below actually modified the frequency plans.

pkg/webui/console/components/events/previews/downlink-message.js
```diff
@@ -52,7 +51,6 @@ const DownLinkMessagePreview = React.memo(({ event }) => {
       const rx2DataRateIndex = getByPath(data, 'request.rx2_data_rate_index')
       return (
         <DescriptionList>
-          <DescriptionList.Byte title={messages.gatewayEUI} data={gatewayEUI} />
```
adriansmares, line 55: I don't really understand why this was added in the first place. If you see the live data of a gateway, you know which gateway you are seeing, so providing the gateway EUI is really weird.

KrishnaIyer (approved):
Good catch. Go part LGTM

johanstokking (approved):

PR #5913, opened 2022-11-04
Remove the required constraint from `server_url` in LoRa Cloud integration

cvetkovski98:
#### Summary
With the recent update to the LoRa Cloud integration the `server_url` field in the LoRa Cloud Modem & Geolocation Services form was marked as required with a default initial value. Existing integrations may have an empty value here, and the backend does not require it either, therefore this PR removes the `required` constraint.

#### Changes
- Removed the required constraint in the yup validation scheme for the `server_url` field in the forms for the LoRa Cloud Modem & Geolocation Services

#### Testing
Manual.

##### Regressions
There shouldn't be any.

#### Notes for Reviewers

...

#### Checklist

### DIFF
--- a/pkg/webui/console/containers/lora-cloud-das-form/index.js
+++ b/pkg/webui/console/containers/lora-cloud-das-form/index.js
@@ -68,9 +68,7 @@ const validationSchema = Yup.object()
     data: Yup.object().shape({
       token: Yup.string().required(sharedMessages.validateRequired),
       use_tlv_encoding: Yup.boolean(),
-      server_url: Yup.string()
-        .url(sharedMessages.validateUrl)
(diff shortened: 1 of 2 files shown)


kschiffer (changes requested):
The `server_url` should also be added to the initial values when updating an existing LRC integration. Otherwise validation will not show possible errors on the field.

```js
    // /console/containers/lora-cloud-gls-form/index.js:129
    const initialValues = validationSchema.cast({ ...defaultAssociation, server_url: '' })
```
And same in the DAS form.

You cannot use just the cast, because the default association will have stripped null values, but the form needs all values defined.

kschiffer (approved):

PR #5929, opened 2022-11-09
Fix email message error by `client` created by organization API key

nicholaspcr:
#### Summary

When creating an `client` via an organization API Key, there is no User to be attached to that, which was generating an email requesting for an admin approval in which the user responsible for the requests was not visible. [Slack thread](https://thethingsindustries.slack.com/archives/C06UQSMPB/p1666964454546849).

#### Changes

- Add email message proto as a wrapper for the API Key.
- Add validation of `SenderIDs` on the `client_requested` template generation.
- Update tests

#### Testing

Manual testing. Still have to update the tests in order to have more than one golden file per notification_type.

##### Regressions

With the current implementation, notifications generated by and API key do not have `sender_id` and `sender_uid` in the DB. The information regarding which API key is stored at the `data` field in the Notification table.

#### Notes for Reviewers

I'm aware that instead of having the proto message as a wrapper I could inject the value on the context, my question regarding this is what should be the standard when injecting values beyond the standard request proto message. Injecting values is all good and dandy but if it is too many values it would be confusing. @adriansmares 

#### Checklist


pkg/email/templates/client_requested.go
```diff
@@ -42,18 +42,41 @@ func init() {
...
+func (crd *ClientRequestedData) SenderTypeMidSentence() string {
+	if crd.Notification().GetSenderIds().IDString() == "" {
+		return "API key"
+	}
+	return "user"
+}
+
+func (crd *ClientRequestedData) Sender() string {
+	if crd.Notification().GetSenderIds().IDString() == "" {
+		return crd.ApiKey
```
adriansmares: I recommend `GetApiKey`.

pkg/identityserver/client_registry.go
```diff
@@ -154,11 +154,20 @@ func (is *IdentityServer) createClient( //nolint:gocyclo
 		return nil, err
 	}
 
+	var apikey string
+	if authInfo, err := is.authInfo(ctx); err == nil {
+		apikey = authInfo.GetApiKey().GetApiKey().GetId()
```
adriansmares: In which cases would the error be not `nil` and we would still like to continue ?

adriansmares (commented):
I am actually quite fine with a new protobuf message. What we do with these contexts carrying the smallest information around is on the border line between clever and hacky, so a new message is quite clean.

The only downside here is that we mix public API with internal API, but this was already happening long before this PR.

The changes look good to me, and I would indeed prefer an extra golden file if possible.

pkg/email/templates/client_requested.go
nicholaspcr: Instead of propagating the `CreateClientEmailMessage` I just added the ApiKey field on the `ClientRequestedData`, therefore there is no proto methods generated for that.

```go
// ClientRequestedData is the data for the client_requested email.
type ClientRequestedData struct {
	email.NotificationTemplateData
	*ttnpb.CreateClientRequest
	ApiKey string
}
```

adriansmares: In that spirit I would keep the whole `*ttnpb.APIKey` then - maybe you can use the key name if present.

pkg/identityserver/client_registry.go
nicholaspcr: Since it wasn't being validated before I initially ignored error related to the `authInfo`. Reflecting on it, the authInfo being unavailable should trigger an error on the client creation so I just return the error if there is one.

pkg/email/templates/client_requested.go
```diff
@@ -42,18 +42,51 @@ func init() {
 }
 
 func newClientRequestedData(_ context.Context, data email.NotificationTemplateData) (email.NotificationTemplateData, error) {
-	var nData ttnpb.CreateClientRequest
-	if err := ttnpb.UnmarshalAny(data.Notification().GetData(), &nData); err != nil {
+	var dd ttnpb.CreateClientEmailMessage
```
adriansmares: Just `data`.

adriansmares (approved):

pkg/email/templates/client_requested.go
nicholaspcr: changed to `emailMsg`, since the parameter is already named `data`

kschiffer (approved):

api/email_messages.proto
```diff
@@ -0,0 +1,29 @@
...
+
+import "lorawan-stack/api/client.proto";
+import "lorawan-stack/api/rights.proto";
+
+package ttn.lorawan.v3;
+
+option go_package = "go.thethings.network/lorawan-stack/v3/pkg/ttnpb";
+
+// CreateClientEmailMessage is used as a wrapper for handling the email
+// regarding the create client procedure.
```
johanstokking: nit; no need for Go-style comments

johanstokking (approved):
Only reviewed API via code ownership

api/email_messages.proto
nicholaspcr: It does propagate to the created struct, which is a bit useful to reinforce that it serves as a wrapper while on development ( mouse over variable definition on vscode ).

PR #5946, opened 2022-11-21
Partition Redis event streams on subscribe

adriansmares:
#### Summary

This PR introduces a partitioning mechanism to our Redis historical event stream subscriptions. The context here is that subscribers are allowed to subscribe to multiple streams at once (that is, a stream per entity), and this in turn causes issues when the number of subscribed streams is in the hundreds to thousands.

The problem is that subscribing to historical streams is inherently an `XRead` operation - we give Redis a set of streams in which we are interested, and Redis returns back when at least one of the streams (or the timeout) has a message available. 

The trouble is that when we provide hundreds of streams at once per call, we ask Redis to always walk through these streams and check if a message is available. This scales terribly when the arrival pattern of these streams is de-correlated (think gateway event streams for gateways without overlapping coverage), because every message causes a full set traversal, but with little chances of actually returning too many messages (due to de-correlation).

The proposed fix is to partition this set of streams and to listen to these subsets in parallel, but to process the partitions serially. This reduces the concurrency of a events subscribe call to a fixed size, while still allowing the caller to subscribe to multiple streams in a unified manner.

#### Changes

- Partition the set of event streams and listen to them in parallel while using the Redis Store events backend.

#### Testing

Edit: Also tested on `staging1`.

##### Regressions

This change affects the `SubscribeWithHistory` RPC which is used by the Console and some monitoring integrations.

The unit tests and local testing cover this, ~and I will run some tests on `staging1` and 2 while preparing this PR~.
Edit: Also tested on `staging1`.

#### Checklist

### DIFF
--- a/cmd/internal/shared/config.go
+++ b/cmd/internal/shared/config.go
@@ -110,6 +110,7 @@ var DefaultEventsConfig = func() config.Events {
 	c.Redis.Store.EntityTTL = 24 * time.Hour
 	c.Redis.Store.EntityCount = 100
 	c.Redis.Store.CorrelationIDCount = 100
+	c.Redis.Store.StreamPartitionSize = 64
 	c.Redis.Workers = 16
 	c.Redis.Publish.QueueSize = 8192
 	c.Redis.Publish.MaxWorkers = 1024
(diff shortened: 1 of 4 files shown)


KrishnaIyer (approved):
Looks generally good to me. Not an expert on this.

PR #5947, opened 2022-11-22
Do not report common task errors

adriansmares:
#### Summary

This PR makes our task mechanism not report `context.Cancelled`, `context.DeadlineExceeded` and `io.EOF` errors as warnings when they occur. They generally show that the RPC has been cancelled by the caller and do not add any additional information.

#### Changes

- Expose the `ErrContextCancelled` and `ErrContextDeadlineExceeded`, which are the wire versions of the standard errors.
- Do not log warnings for the above mentioned errors.

#### Testing

Local testing.

##### Regressions

N/A. These warnings are not really useful and only pollute the reporting.

#### Checklist

### DIFF
--- a/pkg/task/task.go
+++ b/pkg/task/task.go
@@ -17,6 +17,7 @@ package task
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"runtime/debug"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/task/task.go
```diff
@@ -165,7 +166,14 @@ func DefaultStartTask(conf *Config) {
...
 			executionDuration := time.Since(startTime)
-			if err != nil && err != context.Canceled {
+			switch {
+			case err == nil:
+			case errors.Is(err, context.Canceled),
+				errors.Is(err, errors.ErrContextCancelled),
+				errors.Is(err, context.DeadlineExceeded),
+				errors.Is(err, errors.ErrContextDeadlineExceeded):
+			case errors.Is(err, io.EOF):
+			default:
```
johanstokking: This magic is already supported by `pkg/errors`: https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.23/pkg/errors/codes.go#L45-L50

This is cleaner than testing error types as what you really want is knowing whether the error is canceled or a deadline has exceeded, i.e. the code.

Note also that we use US English "canceled" with one l.

adriansmares: The lack of error code checking was actually intentional - I wanted to skip the common/standard errors, but still show warnings for other errors with the same code. I am not sure if we should drop all canceled/deadline exceeded errors indiscriminately.

johanstokking: Okay please decorate this with a comment then.

johanstokking (approved):

PR #5972, opened 2022-12-08
Unclaim end device on delete

### LINKED ISSUE #5836: Add device un-claim to delete device in the console
### Summary

The new version of claiming requires the an end device to be un-claimed before being deleted. The should be done on behalf of the user in the console as it already is in the cli.

### Current Situation

End device is deleted but may still be claimed.

This is not critical but should be done soon as it leaves end devices un-claimable after being deleted and requires using the cli to fix which would be frustrating for console users. 

### Proposed Implementation

Un-claim during onDelete method in end-device setting.

### Contributing

### Code of Conduct

mjamescompton:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/5836

#### Changes

- Set all delete requests to use body instead of query params
- Add handleUnclaim to device general settings 

#### Testing

...


pkg/webui/console/views/device-general-settings/connect.js
```diff
@@ -76,4 +86,17 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => ({
...
+    mapDispatchToProps,
+    mergeProps,
+  )(
+    withRequest(
+      ({
+        getInfoByJoinEUI,
+        device: {
+          ids: { join_eui: joinEUI },
+        },
+      }) => getInfoByJoinEUI({ join_eui: joinEUI }),
```
kschiffer: You can use the `mergeProps` to compose the action with the join eui.

kschiffer: I'd suggest fetching the claim info on the `/views/device` level instead since we might use that info in other places eventually.

pkg/webui/console/views/device-general-settings/identity-server-form/index.js
```diff
@@ -136,13 +140,22 @@ const IdentityServerForm = React.memo(props => {
   )
 
   const onDeviceDelete = React.useCallback(async () => {
+    // Check if device is claimable if so try to unclaim.
```
```suggestion
    // Check if device is claimable and if so, try to unclaim.
```

pkg/webui/console/views/device-general-settings/messages.js
```diff
@@ -36,6 +36,7 @@ const messages = defineMessages({
   updateSuccess: 'End device updated',
   keysResetWarning:
     'You do not have sufficient rights to view end device keys. Only overwriting is allowed.',
+  unclaimFailure: 'An error occurred and the end device could not be unclaimed',
```
kschiffer:
```suggestion
  unclaimFailure: 'An error occurred and the end device could not be unclaimed and deleted',
```

kschiffer (changes requested):

mjamescompton:
@KrishnaIyer What is your advice on how to test this before putting it live, could you take over a test this on staging?

KrishnaIyer:
If this gets reviewed, unit tested and merged into TTS, I can deploy a snapshot and run a test. Please ping me when that's done.

kschiffer:
Ok, let's add a small cypress test-case then which tests against the unclaim API request shape.

ryaplots (approved):

cypress/integration/console/devices/unclaim.spec.js
```diff
@@ -0,0 +1,103 @@
+// Copyright © 2021 The Things Network Foundation, The Things Industries B.V.
```
kschiffer, line 1:
```suggestion
// Copyright © 2022 The Things Network Foundation, The Things Industries B.V.
```

kschiffer (approved):

PR #5981, opened 2022-12-19
Fix csv import first column skip

nicholaspcr:
#### Summary

Closes https://github.com/TheThingsIndustries/lorawan-stack-support/issues/892

Basically there were some cases where the CSV import would not consider the value of the ID column in the csv file, this in fact was not a problem with the ID column but the first column present in the header of any file that contained the BOM (Byte Order Mark) bytes.

#### Changes

- Trim bytes of BOM bytes if present
- ...

#### Testing

Manual testing, used the CSV file provided in the issue and also tested the import with a json file.

##### Regressions

This is a bug fix, the only thing that could be affected by this change would be the JSON import but that was tested.

#### Checklist

### DIFF
--- a/pkg/devicetemplates/ttscsv.go
+++ b/pkg/devicetemplates/ttscsv.go
@@ -367,6 +367,18 @@ func (t *ttsCSV) Convert(ctx context.Context, r io.Reader, ch chan<- *ttnpb.EndD
 	comma := ';'
 	const maxHeaderLength = 1024
 	buf := bufio.NewReaderSize(r, maxHeaderLength)
+
+	// eliminate BOM if its present
+	ru, _, err := buf.ReadRune()
+	if err != nil {
+		return err
+	}
+	if ru != '\uFEFF' {
+		if err := buf.UnreadRune(); err != nil {
+			return err
+		}
+	}
+
 	if head, _ := buf.Peek(maxHeaderLength); len(head) > 0 {
 		if c, ok := determineComma(head); ok {
 			comma = c
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Fixed
 
 - When searching for end-devices, specifying `last_seen_at` as the field in which the devices will be sorted by no longer returns an error.
+- Importing devices via CSV no longer skips the first header column when BOM bytes are present.
 
 ### Security
 


pkg/devicetemplateconverter/grpc.go
```diff
@@ -57,7 +57,8 @@ func (s *endDeviceTemplateConverterServer) Convert(
 	ctx = profilefetcher.NewContextWithFetcher(ctx, profilefetcher.NewTemplateFetcher(s.DTC.Component))
 	ch := make(chan *ttnpb.EndDeviceTemplate)
 	go func() {
-		if err := converter.Convert(ctx, bytes.NewReader(req.Data), ch); err != nil {
+		trimmedBytes := bytes.Trim(req.Data, "\xef\xbb\xbf")
+		if err := converter.Convert(ctx, bytes.NewReader(trimmedBytes), ch); err != nil {
```
KrishnaIyer: At this point, we don't yet know the encoding of `req.Data` right? 
The data is interpreted as UTF-8 in the [converter](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.23/pkg/devicetemplates/ttscsv.go#L361).

I think this trimming should be there as well.

KrishnaIyer (changes requested):

pkg/devicetemplateconverter/grpc.go
adriansmares: I think that no trimming should take place, and instead we should check why the `charset` reader is not removing the BOM.

nicholaspcr: Its not specified in https://www.rfc-editor.org/rfc/rfc4180.html which is what `enconding/csv` is defined on.
So its something added by excel and not exactly expected.

Indeed maybe instead of adding it in the dtc I could peek at the first bytes and remove them if they match the BOM

adriansmares: I am referring to the conversion at https://github.com/TheThingsNetwork/lorawan-stack/blob/55212be1049e8c85ab7c3b752172051a2d30e2f9/pkg/devicetemplates/ttscsv.go#L357-L364 - [charset](https://pkg.go.dev/golang.org/x/net/html/charset) should automatically remove the BOM and convert the text to UTF-8. You should not see the BOM in UTF-8 after the conversion.

adriansmares: By the time you have reached the decoder, the stream is guaranteed to be converted to UTF-8:
https://github.com/TheThingsNetwork/lorawan-stack/blob/55212be1049e8c85ab7c3b752172051a2d30e2f9/pkg/devicetemplates/ttscsv.go#L357-L364

What could happen is indeed that our header character parsing does not handle the BOM correctly - please fix that one instead.

pkg/devicetemplates/ttscsv.go
```diff
@@ -371,6 +376,9 @@ func (t *ttsCSV) Convert(ctx context.Context, r io.Reader, ch chan<- *ttnpb.EndD
 		if c, ok := determineComma(head); ok {
 			comma = c
 		}
+		if ok := isBOM(head); ok {
```
adriansmares: Could we instead peek the first rune (`buf.ReadRune()`) after we initialize `buf`, and then compare it with `'\uFEFF'` ? That's a bit more idiomatic than byte comparison and handling.

If the first rune is not `'\uFEFF'`, we can use `buf.UnreadRune()`.

adriansmares (approved):

nicholaspcr:
@KrishnaIyer, a gentle reminder of this.

KrishnaIyer (approved):
Ah sorry. I thought @adriansmares's review was enough. Good stuff.

PR #5983, opened 2022-12-19
Validate other admin existence on user delete

nicholaspcr:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack-support/issues/894#issuecomment-1357596803

Add validation to check if an user that is going to be deleted is the last admin of a tenant.

#### Changes

- Add invalid deletion error to store pkg.
- Add admin validation at the user deletion
- Add tests (TODO)

#### Testing

Manual testing

#### Checklist

### DIFF
--- a/pkg/identityserver/store/errors.go
+++ b/pkg/identityserver/store/errors.go
@@ -63,6 +63,9 @@ var (
 	ErrUserSessionNotFound = errors.DefineNotFound(
 		"user_session_not_found", "user session with id `{session_id}` not found", "user_id",
 	)
+	ErrLastAdmin = errors.DefineFailedPrecondition(
+		"last_admin", "user `{user_id}` is the last admin",
+	)
 
 	ErrAPIKeyNotFound = errors.DefineNotFound(
 		"api_key_not_found", "api key with id `{api_key_id}` not found", "entity_type", "entity_id",
--- a/pkg/identityserver/user_registry.go
+++ b/pkg/identityserver/user_registry.go
@@ -469,6 +469,12 @@ func (is *IdentityServer) updateUser(ctx context.Context, req *ttnpb.UpdateUserR
 	}
 
 	err = is.store.Transact(ctx, func(ctx context.Context, st store.Store) (err error) {
+		if ttnpb.HasAnyField(req.FieldMask.GetPaths(), "admin") {
+			if err := isLastAdmin(ctx, st, req.User.Ids); err != nil {
+				// Is updating the last admin to no longer be an admin.
+				return err
+			}
+		}
 		updatingContactInfo := ttnpb.HasAnyField(req.FieldMask.GetPaths(), "contact_info")
 		var contactInfo []*ttnpb.ContactInfo
 		updatingPrimaryEmailAddress := ttnpb.HasAnyField(req.FieldMask.GetPaths(), "primary_email_address")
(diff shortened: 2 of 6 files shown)


pkg/identityserver/store/errors.go
```diff
@@ -63,6 +63,9 @@ var (
 	ErrUserSessionNotFound = errors.DefineNotFound(
 		"user_session_not_found", "user session with id `{session_id}` not found", "user_id",
 	)
+	ErrUnauthorizedAdminRemoval = errors.DefinePermissionDenied(
```
adriansmares: I would not say this is unauthorized, as the rights of the user have been checked. It should be failed precondition - a (new) precondition of the delete RPC is that the user is not the last admin.

Compiled up:
`ErrLastAdmin = errors.DefineFailedPrecondition("last_admin", "user `{user_id}` is the last admin"`.

pkg/identityserver/user_registry.go
```diff
@@ -469,6 +469,11 @@ func (is *IdentityServer) updateUser(ctx context.Context, req *ttnpb.UpdateUserR
 	}
 
 	err = is.store.Transact(ctx, func(ctx context.Context, st store.Store) (err error) {
+		err = isLastAdmin(ctx, st, req.User.Ids.UserId)
```
adriansmares: Good catch. Indeed this is another possible removal.

pkg/identityserver/user_registry_test.go
```diff
@@ -316,5 +316,28 @@ func TestUsersCRUD(t *testing.T) {
...
+		_, err = reg.Update(ctx, &ttnpb.UpdateUserRequest{
+			User: &ttnpb.User{
+				Ids:   adminUsr.GetIds(),
+				Admin: false,
+			},
+			FieldMask: ttnpb.FieldMask("admin"),
+		}, adminUsrCreds)
+		a.So(errors.IsPermissionDenied(err), should.BeTrue)
+
+		// provides a false admin field but does not include it in the fieldmask.
```
adriansmares: Nit: Uppercase initial letter.

pkg/identityserver/utils.go
```diff
@@ -80,3 +81,16 @@ func idStrings(entityIDs ...*ttnpb.EntityIdentifiers) []string {
...
 }
+
+// isLastAdmin determines if the userID provided belongs to the last admin of the tenant. Returns an error if the
+// condition is true.
+func isLastAdmin(ctx context.Context, st store.Store, id string) error {
+	adminList, err := st.ListAdmins(ctx, store.FieldMask{"ids"})
+	if err != nil {
+		return err
+	}
+	if len(adminList) == 1 && adminList[0].Ids.UserId == id {
```
adriansmares: I am not very confident in user ID only comparisons. Can we use `proto.Equal` between `ids` instead ?

pkg/identityserver/store/errors.go
nicholaspcr: I went with Unauthorized as there is no authorization for anyone to delete the last user but I agree that failed precondition is more readable and fits better the situation. Will change

pkg/identityserver/user_registry.go
```diff
@@ -469,6 +469,11 @@ func (is *IdentityServer) updateUser(ctx context.Context, req *ttnpb.UpdateUserR
 	}
 
 	err = is.store.Transact(ctx, func(ctx context.Context, st store.Store) (err error) {
+		err = isLastAdmin(ctx, st, req.User.Ids)
```
adriansmares: I think we should do this check only if the field mask really contains `admin`.

adriansmares (approved):

pkg/identityserver/user_registry.go
nicholaspcr: Agreed, also removed the `!req.user.admin` validation since if the field mask is present the change is implied

PR #6004, opened 2023-01-06
Add OpenTelemetry Tracing

### LINKED ISSUE #5691: Add OpenTelemetry Tracing
### Summary

In https://github.com/TheThingsNetwork/lorawan-stack/pull/5642 I started adding OpenTelemetry trace regions to `pkg/identityserver/bunstore`. We should instrument more code in other packages, and set up exporters and such.

### Current Situation

We have some OpenTelemetry instrumentation in `pkg/identityserver/bunstore`, but the rest of the OpenTelemetry SDK has not been set up.

OpenTelemetry tracing is going to be useful for getting more insights into how messages pass trough The Things Stack and can help us understand how one component affects others, especially as the system grows.

### Proposed Implementation

- Instrument more code with OpenTelemetry trace spans ([docs](https://opentelemetry.io/docs/instrumentation/go/getting-started/#trace-instrumentation))
- Set up the OpenTelemetry SDK ([docs](https://opentelemetry.io/docs/instrumentation/go/getting-started/#sdk-installation))

### Contributing

### Code of Conduct

happyRip:
#### Summary

Closes #5691 

#### Changes

- Add telemetry configuration 
- propagate telemetry in component rpc context

#### Testing

To test this PR first an instance of `jaeger` is needed ([Jaeger All in One](https://www.jaegertracing.io/docs/1.41/getting-started/#all-in-one)).
After running `jaeger` access the UI at http://localhost:16686.

Then build and run `lorawan-stack` from `feature/opentelemetry-tracing` branch.
To use the telemetry feature a couple flags need to be set:

```bash
go run ./cmd/ttn-lw-stack -c ./config/stack/ttn-lw-stack.yml start \
  --tracing.enable \
  --tracing.exporter 'otlp' \
  --tracing.collector-config.insecure \
  --tracing.collector-config.endpoint-url 'localhost:4317' \
  --tracing.sample-probability 1.
```

The results should be visible in `jaeger` UI:
localhost_16686_search_end=1677073543230000 limit=100 lookback=5m maxDuration minDuration service=ttn-lw-stack start=1677073243230000(Laptop with HiDPI screen)

##### Regressions

n/a

#### Notes for Reviewers

nil

#### Checklist


happyRip:
Using OTLP (`go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc` or `go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp`) exporter instead of `go.opentelemetry.io/otel/exporters/jaeger` would be better, but all versions past [v1.3.0](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc@v1.3.0) require [`google.golang.org/grpc/credentials/insecure`](https://pkg.go.dev/google.golang.org/grpc/credentials/insecure) package, which was introduced with `https://pkg.go.dev/google.golang.org/grpc@v1.34.0`, but we are currently pinned at [v1.33.1](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.24/go.mod#L33).

https://opentelemetry.io/blog/2022/jaeger-native-otlp/

edit: packages are up-to-date on `v3.25` of the stack, so I'll upgrade to that

pkg/telemetry/tracer/tracer.go
```diff
@@ -0,0 +1,40 @@
...
+
+	otrace "go.opentelemetry.io/otel/trace"
+)
+
+type tracerKeyType struct{}
+
+var tracerKey = &tracerKeyType{}
+
+// NewContext returns a derived context with the tracer set.
+func NewContext(ctx context.Context, t otrace.Tracer) context.Context {
```
adriansmares: I think we should have a `NewContextWithTracer` in order to avoid this pattern of `NewContext` of `FromContext`.

pkg/applicationserver/applicationserver.go
```diff
@@ -104,7 +108,13 @@ var errListenFrontend = errors.DefineFailedPrecondition("listen_frontend", "fail
 
 // New returns new *ApplicationServer.
 func New(c *component.Component, conf *Config) (as *ApplicationServer, err error) {
-	ctx := log.NewContextWithField(c.Context(), "namespace", "applicationserver")
+	ctx := tracer.NewContext(c.Context(),
+		telemetry.FromContext(c.Context()).Tracer(packageName),
+	)
+	ctx, span := tracer.FromContext(ctx).Start(ctx, "New")
```
adriansmares: I think that tracing `New` calls is a bit overkill - we should focus on RPCs instead for this iteration. Same in other spots.

pkg/component/component.go
```diff
@@ -26,6 +26,7 @@ import (
 	"sync"
 	"syscall"
 
+	otrace "go.opentelemetry.io/otel/trace"
```
adriansmares: Do you need the alias given that you don't import `pkg/trace` or anything like that ?

pkg/telemetry/config.go
```diff
@@ -0,0 +1,43 @@
...
+	Enable            bool            `name:"enable" description:"Enable telemetry"`
+	ResourceConfig    ResourceConfig  `name:"resource-config" description:"Information about the running service"`
+	Exporter          string          `name:"exporter" description:"Telemetry exporter (jaeger, writer)"`
+	CollectorConfig   CollectorConfig `name:"collector-config" description:"Trace collector exporter configuration"`
+	WriterConfig      WriterConfig    `name:"writer-config" description:"Writer exporter configuration"`
+	SampleProbability float64         `name:"sample-probability" description:"Sampling probability. Fractions >= 1 will always sample. Fractions < 0 are treated as zero."`
+}
+
+// ResourceConfig represents information about the running service.
+type ResourceConfig struct {
```
adriansmares: I don't think this is relevant. The service version should always come from `pkg/version.String()`, and the service name is always `the-things-stack` for now.

This happens because our binary can run multiple components at once, and as such having a specific service name for one component is not really possible always.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,154 @@
...
+
+	// Fill empty values with defaults
+	return resource.Merge(resource.Default(), rsc)
+}
+
+var initOnce sync.Once
+
+// InitTelemetry initializes the telemetry package and returns the tracer provider.
+// If telemetry is not enabled it returns a noop tracer provider instead.
+func InitTelemetry(config *Config) (otrace.TracerProvider, error) {
```
adriansmares: I think you should take the context here and propagate it downwards, instead of using `context.Background`.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,154 @@
...
+	if !config.Enable {
+		return otrace.NewNoopTracerProvider(), nil
+	}
+
+	var firstTime bool
```
adriansmares: To be honest we're guaranteed to lose the race to set the global tracer provider before the other external packages, and that's fine. We can just avoid setting the global once and having this 'initialize only once' logic in place.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,154 @@
...
+
+	otel.SetTracerProvider(tp)
+	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(
+		propagation.TraceContext{},
+		propagation.Baggage{},
+	))
+	return tp, nil
+}
+
+var cleanupFuncs []func(context.Context) error
```
adriansmares: I think this shouldn't be global, and the caller of `InitTelemetry` should be responsible to `Shutdown` instead.

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -101,6 +108,26 @@ var startCommand = &cobra.Command{
...
+			return err
+		}
+		defer func(ctx context.Context) {
+			if err := telemetry.Shutdown(ctx); err != nil {
+				logger.WithError(err).Error("failed to shutdown telemetry")
+			}
+		}(ctx)
+		ctx := tracer.NewContext(ctx, tp.Tracer(packageName))
+
+		ctx, span := tracer.FromContext(ctx).Start(ctx, "startCommand", otrace.WithAttributes(
```
adriansmares: Same comment regarding the `span` - lets focus on RPCs instead.

adriansmares (commented):
Looks great. We should also make sure that the `bunstore` implementation of the Identity Server uses the contextual trace provider, not the global one.

pkg/telemetry/telemetry.go
happyRip: `otel/trace.TracerProvider` interface does not feature a `Shutdown()` method, but the `otel/sdk/trace.TracerProvider` struct does, but I preferred to stick with the interface. Would returning the struct pointer be preferred in this case?

adriansmares: You can also consider returning `(Trace.TraceProvider, func(context.Context) error, error)` then, and specify that the returned function is a cleanup function.

pkg/telemetry/tracer/tracer.go
```diff
@@ -0,0 +1,50 @@
...
+	return context.WithValue(ctx, tracerKey, t)
+}
+
+func NewContextWithTracer(ctx context.Context, name string, opts ...otrace.TracerOption) context.Context {
```
adriansmares: Nit: Missing comment.

pkg/telemetry/tracer/tracer.go
```diff
@@ -0,0 +1,50 @@
...
+// or returns a noop tracer if it does not exist.
+func FromContext(ctx context.Context) otrace.Tracer {
+	v := ctx.Value(tracerKey)
+	if v == nil {
+		return otrace.NewNoopTracerProvider().Tracer("")
+	}
+	return v.(otrace.Tracer)
+}
+
+func StartFromContext(ctx context.Context, name string, opts ...otrace.SpanStartOption) (context.Context, otrace.Span) {
```
adriansmares: Nit: Missing comment.

nicholaspcr (commented):
Overall the changes to IS are simple and they look good to me. 

Hope you don't mind me not approving the PR. But before doing it so, I want to take some time reading to understand the section regarding the opentelemetry in more detail and to test it locally.

Disregarding the above there is one thing that I noticed which isn't code relate but pkg related, the pkg name conflicts with what is introduced in the [telemetry exporter PR](https://github.com/TheThingsIndustries/lorawan-stack/pull/3574). I personally think its nice to have this and the telemetry exporter inside the `telemetry` folder but each on its own pkg, something like:
```bash
telemetry
├── tracer
    └── ...
└── ttn
    └── ...

```
 
Let me know what you think @happyRip

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -178,11 +180,23 @@ var startCommand = &cobra.Command{
 			start.DeviceClaimingServer = true
 		}
 
+		tp, shutdown, err := telemetry.InitTelemetry(ctx, &config.Telemetry)
+		if err != nil {
+			return err
+		}
+		defer func(ctx context.Context, f func(context.Context) error) {
+			if err := f(ctx); err != nil {
+				logger.WithError(err).Error("failed to shutdown telemetry")
```
johanstokking:
```suggestion
				logger.WithError(err).Error("Failed to shutdown telemetry")
```

pkg/applicationserver/applicationserver.go
```diff
@@ -105,7 +108,12 @@ var errListenFrontend = errors.DefineFailedPrecondition("listen_frontend", "fail
 
 // New returns new *ApplicationServer.
 func New(c *component.Component, conf *Config) (as *ApplicationServer, err error) {
-	ctx := log.NewContextWithField(c.Context(), "namespace", "applicationserver")
+	c.AddContextFiller(func(ctx context.Context) context.Context {
+		return tracer.NewContextWithTracer(ctx, packageName)
```
johanstokking: The context filling is component-wide. The component `c` can be shared with multiple roles; the single binary `ttn-lw-stack start all` uses one component. If you add context fillers, they will be used by all other roles. So the request context of `NetworkServer` will be filled with the tracers of all other roles in the process.

So if you need per-role context filling for the request context, you need to register a hook. See how `c.GRPC.Register{Unary,Stream}Hook()` is being used.

pkg/applicationserver/applicationserver.go
```diff
@@ -57,6 +58,8 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+const packageName = "go.thethings.network/lorawan-stack/pkg/applicationserver"
```
johanstokking: This looks like a great candidate to put in `observability.go`. Also isn't this more like a tracer namespace?

Another nit; the package name is `go.thethings.network/lorawan-stack/v3/pkg/applicationserver`. I'm not sure though if we need package names; why not namespaces as we use in logging?

pkg/telemetry/config.go
```diff
@@ -0,0 +1,40 @@
...
+
+import "go.thethings.network/lorawan-stack/v3/pkg/config/tlsconfig"
+
+// Config represents configuration for OpenTelemetry tracing.
+type Config struct {
+	Enable            bool            `name:"enable" description:"Enable telemetry"`
+	Exporter          string          `name:"exporter" description:"Telemetry exporter (otlp, writer)"`
+	CollectorConfig   CollectorConfig `name:"collector-config" description:"Trace collector exporter configuration"`
+	WriterConfig      WriterConfig    `name:"writer-config" description:"Writer exporter configuration"`
+	SampleProbability float64         `name:"sample-probability" description:"Sampling probability. Fractions >= 1 will always sample. Fractions < 0 are treated as zero."` //nolint:lll
```
johanstokking:
```suggestion
	SampleProbability float64         `name:"sample-probability" description:"Sampling probability. Fractions >= 1 will always sample. Fractions < 0 are treated as zero"` //nolint:lll
```

pkg/telemetry/exporter.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+func initOTLPExporter(ctx context.Context, config *CollectorConfig) (sdktrace.SpanExporter, error) {
+	var opts []grpc.DialOption
+	switch {
+	case config.Insecure:
+		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+
+	default:
```
johanstokking: If statement?

pkg/telemetry/exporter.go
```diff
@@ -0,0 +1,104 @@
...
+		case tlsConfig == nil:
+			tlsConfig = new(tls.Config)
+			fallthrough
+
+		case tlsConfig.RootCAs == nil:
+			var err error
+			if tlsConfig.RootCAs, err = x509.SystemCertPool(); err != nil {
+				tlsConfig.RootCAs = x509.NewCertPool()
+			}
+		}
```
johanstokking: I don't think this is necessary? See other usages of `tlsconfig.Client`

pkg/telemetry/exporter.go
```diff
@@ -0,0 +1,104 @@
...
+
+		if err := config.TLS.ApplyTo(tlsConfig); err != nil {
+			return nil, err
+		}
+		opts = append(opts, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)))
+	}
+
+	conn, err := grpc.DialContext(ctx, config.EndpointURL, append(opts, grpc.WithBlock())...)
```
johanstokking: Add `grpc.FailOnNonTempDialError(true)` as dial option here

pkg/telemetry/exporter.go
```diff
@@ -0,0 +1,104 @@
...
+func exporterFromConfig(ctx context.Context, config *Config) (sdktrace.SpanExporter, error) {
+	switch strings.ToLower(config.Exporter) {
+	case "otlp":
+		return initOTLPExporter(ctx, &config.CollectorConfig)
+
+	case "writer":
+		return initWriterExporter(&config.WriterConfig)
+
+	default:
+		return nil, errors.New("unknown telemetry exporter")
```
johanstokking: Use error definitions

pkg/telemetry/exporter.go
johanstokking: By default, this value is `localhost:<something>`; so does this mean TTS relies on this endpoint to be there?

cmd/internal/shared/config.go
```diff
@@ -141,6 +142,18 @@ var DefaultKeyVaultConfig = config.KeyVault{
...
 }
 
+// DefaultTelemetryConfig is the default config for telemetry tracing.
+var DefaultTelemetryConfig = telemetry.Config{
+	Exporter: "otlp",
+	WriterConfig: telemetry.WriterConfig{
+		Destination: "stderr",
+	},
+	CollectorConfig: telemetry.CollectorConfig{
+		EndpointURL: "localhost:4317",
```
johanstokking, line 153: I mean here

pkg/applicationserver/applicationserver.go
happyRip: The package name was matched to what was previously used, which didn't include the `...v3/...` part https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.24/pkg/identityserver/bunstore/observability.go#L19. `tracerNamespace` may be a better fitting name indeed. I will move this back to `observability.go` as well.

pkg/telemetry/exporter.go
happyRip: I set this as a 'sensible default', like here https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.24/cmd/internal/shared/config.go#L93
By default telemetry is not enabled, so it does not rely on the endpoint.

johanstokking: OK, but how can I see from `DefaultTelemetryConfig` that it isn't enabled?

I understand about the Redis reference, but Redis is required for TTS.

happyRip: Boolean `Enable` is false by default, but it could be set explicitly to false in the default struct.

nicholaspcr:
>Hope you don't mind me not approving the PR. But before doing it so, I want to take some time reading to understand the section regarding the opentelemetry in more detail and to test it locally.

Forgot to mention here, tested this locally last Friday. Its surprisingly easy to use and its pretty cool.

>Disregarding the above there is one thing that I noticed which isn't code relate but pkg related, the pkg name conflicts with what is introduced in the https://github.com/TheThingsIndustries/lorawan-stack/pull/3574. I personally think its nice to have this and the telemetry exporter inside the telemetry folder but each on its own pkg, something like:

Regarding this I saw that you did change a bit of the structure but I meant to suggest a complete isolation, almost like leaving `telemetry` folder with a file with functions that could be used by the tracer and the telemetry exporter. This is just for organization and to avoid conflict with the other telemetry PR.

happyRip:
> Regarding this I saw that you did change a bit of the structure but I meant to suggest a complete isolation, almost like leaving telemetry folder with a file with functions that could be used by the tracer and the telemetry exporter. This is just for organization and to avoid conflict with the other telemetry PR.

I haven't yet addressed it to be honest. I wanted to polish the functionality first.
Since `observability` name is used a couple of times, maybe that could be a fitting name to this package, wether nested in `telemetry` directory or not?

edit: Settled offline for `pkg/telemetry/tracing`

nicholaspcr:
>I haven't yet addressed it to be honest. I wanted to polish the functionality first.

My bad, misunderstood the changes then.

>Since observability name is used a couple of times, maybe that could be a fitting name to this package, wether nested in telemetry directory or not?

I personally prefer shorter names but that is really up to you, I think it works either way.

cmd/internal/shared/config.go
```diff
@@ -141,6 +142,19 @@ var DefaultKeyVaultConfig = config.KeyVault{
 	Provider: "static",
 }
 
+// DefaultTracingConfig is the default config for telemetry tracing.
+var DefaultTracingConfig = tracing.Config{
+	Enable:   false,
```
adriansmares, line 147: `Enable` will be false anyway.

cmd/ttn-lw-stack/commands/start.go
```diff
@@ -178,11 +180,23 @@ var startCommand = &cobra.Command{
 			start.DeviceClaimingServer = true
 		}
 
+		tp, shutdown, err := tracing.Initialize(ctx, &config.Tracing)
+		if err != nil {
+			return err
+		}
+		defer func(ctx context.Context, f func(context.Context) error) {
```
adriansmares: Why do you pass `ctx`/`shutdown` as parameters ?

pkg/networkserver/networkserver.go
```diff
@@ -177,7 +180,15 @@ const (
 
 // New returns new NetworkServer.
 func New(c *component.Component, conf *Config, opts ...Option) (*NetworkServer, error) {
-	ctx := log.NewContextWithField(c.Context(), "namespace", "networkserver")
+	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.Ns", cluster.HookName, func(uh grpc.UnaryHandler) grpc.UnaryHandler {
```
adriansmares: Is this needed ?

adriansmares (commented):
Looks great. In general a problem that we have is that we slowly added services over time, but we did not add the logging hook to the RPC middleware. Since I guess you followed that particular hook while adding the tracing hook, there are some services which lack the hook.

For example, in the Application Server we only mention the `NsAs` service, but if you look into `ApplicationServer.RegisterServices` you will observe that also there are the `As`, `AsEndDeviceRegistry`, `AppAs`, `ApplicationWebhookRegistry`,`ApplicationPubSubRegistry` services.

Can you take a look at the big components (IS, NS, AS, JS, GS, PBA) and ensure that they have the logging namespace and tracing namespace set via the gRPC hook ? You can create a new issue to investigate the remaining components.

We also need the same middleware for HTTP handles (like the ones in the Gateway Configuration Server), but I am ok with adding them in a subsequent PR.

cmd/internal/shared/config.go
happyRip: https://github.com/TheThingsNetwork/lorawan-stack/pull/6004#discussion_r1116698774

johanstokking: Let's be explicit here indeed

pkg/applicationserver/applicationserver.go
```diff
@@ -23,6 +23,10 @@ import (
 	"time"
 
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/emptypb"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
```
adriansmares: Remove the newline.

pkg/applicationserver/applicationserver.go
```diff
@@ -266,7 +273,28 @@ func New(c *component.Component, conf *Config) (as *ApplicationServer, err error
...
 	}
 
-	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.NsAs", cluster.HookName, c.ClusterAuthUnaryHook())
+	for _, hook := range []struct {
+		name       string
+		middleware hooks.UnaryHandlerMiddleware
+	}{
+		{rpctracer.TracerHook, rpctracer.UnaryTracerHook(tracerNamespace)},
+		{rpclog.NamespaceHook, rpclog.UnaryNamespaceHook(logNamespace)},
+		{cluster.HookName, c.ClusterAuthUnaryHook()},
```
adriansmares: You are now adding the cluster hook to services other than `NsAs` - this should not be done.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -28,6 +28,10 @@ import (
 	"time"
 
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
```
adriansmares: Remove new line.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -168,8 +174,22 @@ func New(c *component.Component, conf *Config, opts ...Option) (gs *GatewayServe
...
 	// Register gRPC services.
-	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.NsGs", rpclog.NamespaceHook, rpclog.UnaryNamespaceHook("gatewayserver"))
-	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.NsGs", cluster.HookName, c.ClusterAuthUnaryHook())
+	for _, hook := range []struct {
+		name       string
+		middleware hooks.UnaryHandlerMiddleware
+	}{
+		{rpctracer.TracerHook, rpctracer.UnaryTracerHook(tracerNamespace)},
+		{rpclog.NamespaceHook, rpclog.UnaryNamespaceHook(logNamespace)},
+		{cluster.HookName, c.ClusterAuthUnaryHook()},
```
adriansmares: Same comment as before - the cluster auth hook should not be added to existing services.

pkg/joinserver/joinserver.go
```diff
@@ -113,14 +118,28 @@ func New(c *component.Component, conf *Config) (*JoinServer, error) {
...
-	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.NsJs", cluster.HookName, c.ClusterAuthUnaryHook())
-	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.AsJs", cluster.HookName, c.ClusterAuthUnaryHook())
-	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.Js", cluster.HookName, c.ClusterAuthUnaryHook())
+	for _, hook := range []struct {
+		name       string
+		middleware hooks.UnaryHandlerMiddleware
+	}{
+		{rpctracer.TracerHook, rpctracer.UnaryTracerHook(tracerNamespace)},
+		{rpclog.NamespaceHook, rpclog.UnaryNamespaceHook(logNamespace)},
+		{cluster.HookName, c.ClusterAuthUnaryHook()},
```
adriansmares: Same here.

pkg/networkserver/networkserver.go
```diff
@@ -270,14 +275,30 @@ func New(c *component.Component, conf *Config, opts ...Option) (*NetworkServer,
...
-	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.AsNs", rpclog.NamespaceHook, rpclog.UnaryNamespaceHook("networkserver"))
-	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.Ns", rpclog.NamespaceHook, rpclog.UnaryNamespaceHook("networkserver"))
-	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.GsNs", cluster.HookName, c.ClusterAuthUnaryHook())
+	for _, hook := range []struct {
+		name       string
+		middleware hooks.UnaryHandlerMiddleware
+	}{
+		{rpctracer.TracerHook, rpctracer.UnaryTracerHook(tracerNamespace)},
+		{rpclog.NamespaceHook, rpclog.UnaryNamespaceHook(logNamespace)},
+		{cluster.HookName, c.ClusterAuthUnaryHook()},
```
adriansmares: Same here - `Ns` should not have it.

pkg/packetbrokeragent/agent.go
```diff
@@ -362,8 +368,22 @@ func New(c *component.Component, conf *Config, opts ...Option) (*Agent, error) {
...
+	for _, hook := range []struct {
+		name       string
+		middleware hooks.UnaryHandlerMiddleware
+	}{
+		{rpctracer.TracerHook, rpctracer.UnaryTracerHook(tracerNamespace)},
+		{rpclog.NamespaceHook, rpclog.UnaryNamespaceHook(logNamespace)},
+		{cluster.HookName, c.ClusterAuthUnaryHook()},
+	} {
+		for _, filter := range []string{
+			"/ttn.lorawan.v3.Pba",
```
adriansmares, line 379: Same here - `Pba` should not have it.

pkg/joinserver/joinserver.go
```diff
@@ -113,14 +118,27 @@ func New(c *component.Component, conf *Config) (*JoinServer, error) {
...
+			"/ttn.lorawan.v3.JsEndDeviceRegistry",
+			"/ttn.lorawan.v3.Js",
+			"/ttn.lorawan.v3.ApplicationActivationSettingRegistry",
+		} {
+			c.GRPC.RegisterUnaryHook(filter, hook.name, hook.middleware)
+		}
+	}
 	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.AsJs", cluster.HookName, c.ClusterAuthUnaryHook())
-	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.Js", cluster.HookName, c.ClusterAuthUnaryHook())
+	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.AppJs", cluster.HookName, c.ClusterAuthUnaryHook())
```
adriansmares: `AppJs` did not have `ClusterAuthUnaryHook` previously - I see only the logging hook.

pkg/networkserver/networkserver.go
```diff
@@ -270,14 +275,28 @@ func New(c *component.Component, conf *Config, opts ...Option) (*NetworkServer,
...
+		}
+	}
 	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.GsNs", cluster.HookName, c.ClusterAuthUnaryHook())
-	c.GRPC.RegisterStreamHook("/ttn.lorawan.v3.AsNs", cluster.HookName, c.ClusterAuthStreamHook())
 	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.AsNs", cluster.HookName, c.ClusterAuthUnaryHook())
-	c.GRPC.RegisterUnaryHook("/ttn.lorawan.v3.Ns", cluster.HookName, c.ClusterAuthUnaryHook())
+
+	c.GRPC.RegisterStreamHook("/ttn.lorawan.v3.AsNs", rpctracer.TracerHook, rpctracer.StreamTracerHook(tracerNamespace))
+	c.GRPC.RegisterStreamHook("/ttn.lorawan.v3.AsNs", rpclog.NamespaceHook, rpclog.StreamNamespaceHook(logNamespace))
+	c.GRPC.RegisterStreamHook("/ttn.lorawan.v3.AsNs", cluster.HookName, c.ClusterAuthStreamHook())
```
adriansmares, line 300: Where is the `Ns` `ClusterAuthUnaryHook` ?
Same for the `AsNs` `ClusterAuthStreamHook`.

adriansmares (approved):

PR #6007, opened 2023-01-09
Replace `heptiolabs/healthcheck` with `hellofresh/health-go`

### LINKED ISSUE #5997: Migrate away from github.com/heptiolabs/healthcheck
### Summary

We should migrate away from [github.com/heptiolabs/healthcheck](https://github.com/vmware-archive/healthcheck), as it is a deprecated library and we require dependency replacements because of it.

### Current Situation

Currently use this library in order to expose the `/healthz/live` and `/healthz/ready` endpoints.

[In theory](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) these endpoints are meant to give an insight into the container lifecycle status. In practice we don't have any checks and return 200 OK every time.

In order to avoid depending on this old library version. We also do dependency replacements which we should get rid of:
https://github.com/TheThingsNetwork/lorawan-stack/blob/1be5bc914afd36481f858c235756a59054ac41df/go.mod#L36-L38

### Proposed Implementation

I think we should look into using another library. A candidate that I found at a first glance is https://github.com/hellofresh/health-go .

We don't necessarily have to add new checks now, but we should adapt the functions exposed by `*component.Component` to be usable with the new library.

### Contributing

### Code of Conduct

cvetkovski98:
#### Summary
Introduce the `healthcheck` package to provide functionality for determining a Component's health based on given checks and an implementation using `hellofresh/health-go`

Closes #5997 

#### Changes
- `/healthz/ready` and `/healthz/live` are merged under `/healthz` since that's how `hellofresh/health-go` naturally works
- the component health handler now depends on an interface instead of concrete implementation to allow for easier changes in the future
- added tests for component health checking 

#### Testing
- Manually and UT

##### Regressions
It is indicated in the task that the implementation needs to be adapted to the already existing endpoints. The library does not provide means to separate the readiness from the liveliness checks.

#### Notes for Reviewers

...

#### Checklist


go.mod
```diff
@@ -268,9 +266,9 @@ require (
 	google.golang.org/api v0.61.0 // indirect
 	google.golang.org/appengine v1.6.7 // indirect
 	google.golang.org/protobuf v1.28.1 // indirect
-	// NOTE: gopkg.in/DATA-DOG/go-sqlmock.v1 is actually a different version (see above).
```
adriansmares, line 263: We should remove the replace as well.

go.mod
```diff
@@ -268,9 +266,9 @@ require (
...
-	gopkg.in/DATA-DOG/go-sqlmock.v1 v1.0.0-00010101000000-000000000000 // indirect
 	gopkg.in/alexcesaro/quotedprintable.v3 v3.0.0-20150716171945-2caba252f4dc // indirect
 	gopkg.in/ini.v1 v1.66.4 // indirect
 	mellium.im/sasl v0.3.0 // indirect
 )
+
+require github.com/hellofresh/health-go/v5 v5.0.0
```
adriansmares: Move this to the `require ()` block with the rest of the direct dependencies.

pkg/component/web.go
```diff
@@ -108,8 +108,7 @@ func (c *Component) initWeb() error {
 				webmiddleware.AuthUser(healthUsername, c.config.HTTP.Health.Password),
 			)))
 		}
-		g.HandleFunc("/healthz/live", c.healthHandler.LiveEndpoint)
-		g.HandleFunc("/healthz/ready", c.healthHandler.ReadyEndpoint)
+		g.Handle("/healthz", c.healthHandler.GetHandler())
```
adriansmares, line 111: Check if the Docker health status checks still works. It still uses the old URL:

https://github.com/TheThingsNetwork/lorawan-stack/blob/8a7bcd6c98c6757a2dd36f8090a4fdc8478be704/Dockerfile#L40

pkg/component/web.go
```diff
@@ -121,14 +120,9 @@ func (c *Component) RegisterWeb(s web.Registerer) {
...
 
-// RegisterLivenessCheck registers a liveness check for the component.
-func (c *Component) RegisterLivenessCheck(name string, check healthcheck.Check) {
-	c.healthHandler.AddLivenessCheck(name, check)
-}
-
-// RegisterReadinessCheck registers a readiness check for the component.
-func (c *Component) RegisterReadinessCheck(name string, check healthcheck.Check) {
-	c.healthHandler.AddReadinessCheck(name, check)
+// RegisterCheck registers a liveness check for the component.
```
adriansmares: I would say `RegisterHealthCheck` and `registers a health check`.

pkg/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,32 @@
...
+	"context"
+	"net/http"
+)
+
+// Check is a function that determines the health of the HealtChecker.
+type Check func(ctx context.Context) error
+
+// HealthChecker manages checks for determining the healhiness of a component.
+type HealthChecker interface {
+	AddHTTPCheck(name string, url string) error
```
adriansmares, line 28: The interface is a good idea but components don't really have access to it if you think about it. Only `AddCheck` is basically exposed by `Component`.

You can consider just having a `Component.HealthChecker() healthcheck.HealthChecker` function that returns this interface to implementation instead. In this way components gain access to `AddHTTPCheck`, `AddPgCheck` and so forth.

pkg/healthcheck/healthgo.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"net/http"
+
+	"github.com/hellofresh/health-go/v5"
+	healthHttp "github.com/hellofresh/health-go/v5/checks/http"
+	healthPg "github.com/hellofresh/health-go/v5/checks/postgres"
+)
+
+type healthGoHealthChecker struct {
```
adriansmares: `defaultHealthChecker` is also ok. I get that you wanted to signal the underlying implementation but you don't have to.

pkg/healthcheck/healthgo.go
```diff
@@ -0,0 +1,62 @@
...
+	})
+}
+
+// GetHandler implements HealthChecker.
+func (hghc *healthGoHealthChecker) GetHandler() http.Handler {
+	return hghc.h.Handler()
+}
+
+// NewHealthGoHealthChecker creates a new HealthCheker implementation using hellofresh/health-go.
+func NewHealthGoHealthChecker() HealthChecker {
```
adriansmares: I would just use `NewDefaultHealthChecker()`.

pkg/healthcheck/healthgo.go
```diff
@@ -0,0 +1,62 @@
...
+	h, _ := health.New(health.WithSystemInfo())
```
adriansmares: Please don't ignore the error, but propagate it.

pkg/component/web.go
cvetkovski98: Updated it and tested locally - works

adriansmares (approved):

PR #6017, opened 2023-01-13
Fix events not appearing in Console

kschiffer:
#### Summary
This is a preliminary fix to a regression regarding event ingestion in the Console.

References https://github.com/TheThingsNetwork/lorawan-stack/issues/6019

#### Changes
- Do not use event stop event as cancellation type for starting an event stream.

#### Testing

- Local testing on staging environment.

#### Notes for Reviewers

This is a preliminary fix as it does not resolve the underlying issue of the device view initially mounting, unmounting and remounting again. I'll get to the ground of that while we have this fix applied.

#### Checklist

### DIFF
--- a/pkg/webui/console/store/middleware/logics/events.js
+++ b/pkg/webui/console/store/middleware/logics/events.js
@@ -86,7 +86,7 @@ const createEventsConnectLogics = (reducerName, entityName, onEventsStart) => {
   return [
     createLogic({
       type: START_EVENTS,
-      cancelType: [STOP_EVENTS, START_EVENTS_FAILURE],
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Fixed
 
 - When searching for end-devices, specifying `last_seen_at` as the field in which the devices will be sorted by no longer returns an error.
+- Event data not appearing in the Console.
```
adriansmares: `v3.23.1` does not seem to have this behavior, so the changelog entry may be skipped.

PR #6036, opened 2023-01-25
LoRa Cloud Improvements

### LINKED ISSUE #6025: LoRa Cloud integration improvements
### Summary

We should automatically extra ports pre-fill the ports of the integration, and add support for additional message type and timestamp accuracy in the LoRa Cloud DAS integration.

### Current Situation

1. The ports field of the LoRa Cloud integration is by default empty.
2. We support only `algorithm_type` `gnss` and `wifi`.
3. The timestamp accuracy that we send to LoRa Cloud is in seconds.
4. The Console cannot render a location solved message whose source is unknown.

In order to improve the user experience while using the integration.

### Proposed Implementation

1.  Most likely the default ports should be added to the default values, in string form(?):
https://github.com/TheThingsNetwork/lorawan-stack/blob/acb6587e64d717f0bc5756514f9a509a78c1fdf3/pkg/webui/console/containers/lora-cloud-das-form/index.js#L94-L99
2. Add a new location solved type, `GNSSNGPositionSolutionType` inside the object definitions:
https://github.com/TheThingsNetwork/lorawan-stack/blob/acb6587e64d717f0bc5756514f9a509a78c1fdf3/pkg/applicationserver/io/packages/loradms/v1/api/objects/objects.go#L55-L65
We should then actually interpret it as just a GPS source:
https://github.com/TheThingsNetwork/lorawan-stack/blob/acb6587e64d717f0bc5756514f9a509a78c1fdf3/pkg/applicationserver/io/packages/loradms/v1/package.go#L205-L211
The string form is `gnssng`, and `ng` stands for new generation.
Also let's remove the algorithm from the service name, and use only the package name:
https://github.com/TheThingsNetwork/lorawan-stack/blob/acb6587e64d717f0bc5756514f9a509a78c1fdf3/pkg/applicationserver/io/packages/loradms/v1/package.go#L218
3. We should also take into account `UnixNano` and make it part of the timestamp fraction:
https://github.com/TheThingsNetwork/lorawan-stack/blob/acb6587e64d717f0bc5756514f9a509a78c1fdf3/pkg/applicationserver/io/packages/loradms/v1/package.go#L333-L339
It would also be great if we could use a bit more logic for timestamping the messages using the function above. Right now we use the message `ReceivedAt`, which is the timestamp at which the Application Server (the process) received, not necessarily the gateway itself. While it is very hard to correctly timestamp messages, we can do a bit of a better job.
We have this functionality right now in the Network Server:
https://github.com/TheThingsNetwork/lorawan-stack/blob/acb6587e64d717f0bc5756514f9a509a78c1fdf3/pkg/networkserver/mac/device_time.go#L47-L61
I would like to see it moved to `pkg/util/lora` and be used with the timestamps that we send to LoRa Cloud.
4. The source should be included only if it exists:
https://github.com/TheThingsNetwork/lorawan-stack/blob/acb6587e64d717f0bc5756514f9a509a78c1fdf3/pkg/webui/console/components/events/previews/application-location.js#L32
In unknown source messages, `source` will be undefined.

### Contributing

### Code of Conduct

cvetkovski98:
#### Summary
This PR adds several LoRa Cloud integration improvements.

Closes #6025.

#### Changes
- Add default F Port to LoRa Cloud DAS form
- Add `GNSSNGPositionSolutionType`
- Extract timestamp handling to `pkg/util/lora`

#### Testing
- Existing UT
- Manual

##### Regressions

...

#### Notes for Reviewers

...

#### Checklist

### DIFF
--- a/pkg/applicationserver/io/packages/loradms/v1/package.go
+++ b/pkg/applicationserver/io/packages/loradms/v1/package.go
@@ -215,7 +215,7 @@ func (p *DeviceManagementPackage) sendLocationSolved(ctx context.Context, ids *t
 		ReceivedAt:     ttnpb.ProtoTimePtr(time.Now()),
 		Up: &ttnpb.ApplicationUp_LocationSolved{
 			LocationSolved: &ttnpb.ApplicationLocation{
-				Service: fmt.Sprintf("%v-%s", PackageName, position.Algorithm),
+				Service: PackageName,
 				Location: &ttnpb.Location{
 					Latitude:  position.LLH[0],
 					Longitude: position.LLH[1],
--- a/pkg/applicationserver/io/packages/loradms/v1/package.go
+++ b/pkg/applicationserver/io/packages/loradms/v1/package.go
@@ -336,7 +336,7 @@ func float64PtrOfTimestamp(x *pbtypes.Timestamp) *float64 {
 	if x == nil {
 		return nil
 	}
-	f := float64(ttnpb.StdTime(x).UnixNano())
+	f := float64(ttnpb.StdTime(x).UnixNano()) / float64(1e9)
 	return &f
 }
 
--- a/pkg/applicationserver/io/packages/loradms/v1/package.go
+++ b/pkg/applicationserver/io/packages/loradms/v1/package.go
@@ -204,7 +204,7 @@ func (p *DeviceManagementPackage) sendLocationSolved(ctx context.Context, ids *t
 	}
 	source := ttnpb.LocationSource_SOURCE_UNKNOWN
 	switch position.Algorithm {
-	case objects.GNSSPositionSolutionType:
+	case objects.GNSSPositionSolutionType, objects.GNSSNGPositionSolutionType:
 		source = ttnpb.LocationSource_SOURCE_GPS
 	case objects.WiFiPositionSolutionType:
 		source = ttnpb.LocationSource_SOURCE_WIFI_RSSI_GEOLOCATION
--- a/pkg/applicationserver/io/packages/loradms/v1/package.go
+++ b/pkg/applicationserver/io/packages/loradms/v1/package.go
@@ -30,6 +30,7 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/log"
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 	"go.thethings.network/lorawan-stack/v3/pkg/types"
+	lorautil "go.thethings.network/lorawan-stack/v3/pkg/util/lora"
 	urlutil "go.thethings.network/lorawan-stack/v3/pkg/util/url"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/util/lora/lora.go
```diff
@@ -27,3 +33,38 @@ func AdjustedRSSI(channelRSSI, snr float32) float32 {
 	}
 	return rssi
 }
+
+// ExtractRxMetadataReceivedAt returns the ReceivedAt timestamp based on the
+// provided metadatas or nil if no md.GpsTime nor md.ReceivedAt is found.
+func ExtractRxMetadataReceivedAt(mds []*ttnpb.RxMetadata) *pbtypes.Timestamp {
```
adriansmares: You can merge these 3 functions into one if you accept an interface with the following contents:
```
GetRxMetadata() []*RxMetadata
GetReceivedAt() *timestamppb.Timestamp
```

pkg/applicationserver/io/packages/loradms/v1/package.go
```diff
@@ -334,7 +336,7 @@ func float64PtrOfTimestamp(x *pbtypes.Timestamp) *float64 {
 	if x == nil {
 		return nil
 	}
-	f := float64(ttnpb.StdTime(x).Unix())
+	f := float64(ttnpb.StdTime(x).UnixNano())
```
adriansmares: This is not correct. What I expected was `float64(ttnpb.StdTime(x).UnixNano()) / float64(1e9)` - the unit has to be seconds, not nano seconds.

pkg/webui/console/containers/lora-cloud-das-form/index.js
```diff
@@ -95,6 +95,7 @@ const defaultValues = {
   data: {
     token: '',
     server_url: LORA_CLOUD_DAS.DEFAULT_SERVER_URL,
+    f_port_set: LORA_CLOUD_DAS.DEFAULT_PORT,
```
adriansmares: This is a single port, while we should provide multiple of them per issue description.

pkg/util/lora/lora.go
```diff
@@ -27,3 +33,27 @@ func AdjustedRSSI(channelRSSI, snr float32) float32 {
 	}
 	return rssi
 }
+
+// ReceivedAtImprovable describes an entity that holds RxMetadata and timestamp.
+type ReceivedAtImprovable interface {
```
adriansmares: Nit: just leave it as `UplinkMessage` - since we're just looking for a shape, it's fine.

pkg/util/lora/lora.go
```diff
@@ -27,3 +33,27 @@ func AdjustedRSSI(channelRSSI, snr float32) float32 {
...
+	GetRxMetadata() []*ttnpb.RxMetadata
+	GetReceivedAt() *pbtypes.Timestamp
+}
+
+// GetRxMetadataImprovedReceivedAt tries to improve the ReceivedAt timestamp using
+// the message's GpsTime and metadata's ReceivedAt.
+func GetRxMetadataImprovedReceivedAt(rai ReceivedAtImprovable) *pbtypes.Timestamp {
```
adriansmares: Nit: `GetAdjustedReceivedAt`.

pkg/webui/console/containers/lora-cloud-das-form/index.js
```diff
@@ -95,6 +95,7 @@ const defaultValues = {
   data: {
     token: '',
     server_url: LORA_CLOUD_DAS.DEFAULT_SERVER_URL,
+    f_port_set: [192, 197, 198, 199],
```
adriansmares: I think that in the frontend we prefer that this becomes something like `LORA_CLOUD_DAS.DEFAULT_PORT_SET`.

adriansmares (approved):

adriansmares (approved):

kschiffer (approved):
Codeowned LGTM.

PR #6043, opened 2023-01-31
Add last invalidation keys removal migration

### LINKED ISSUE #4652: Remove obsolete last invalidation keys
#### Summary

https://github.com/TheThingsNetwork/lorawan-stack/pull/4651 removes the `ttn:v3:ns:application-uplinks:uid:<uid here>:last-invalidation` keys from the application uplink queue. These keys are now unused, and may be deleted.

#### Why do we need this?

In order to avoid storing unused keys.

Keys of the form `ttn:v3:ns:application-uplinks:uid:*:last-invalidation`

These keys should be deleted.

#### Environment

`v3.15.0`

`ns-db` migrate takes care of the `v3.10` -> `v3.11` migration. This migration needs to be extracted in a separate function / file, and a new one that ranges through the invalidation keys and deletes them should be added.

Manually set the key and make sure that the migration deletes it safely.

Yes.

cvetkovski98:
#### Summary
Replace the v3.10 to v3.11 Network Server db migration with last invalidation keys and add version warnings accordingly.

Closes #4652 

#### Changes
- Removed the v3.10 to v3.11 key migration from `ns-db migrate`
- Added last invalidation keys migration

#### Testing
- Added the keys to Redis and verified that they don't exist after the migration
- Verified that the version numbers are properly set after the migration under the correct namespace 

##### Regressions
- Direct migrations of version 0 to version 2 of the Network Server db are not supported. Tested by manually setting the versions in the db and verified that the correct warning is displayed.

#### Notes for Reviewers
...

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Removed
 
 - The device version identifiers no longer have the `serial_number`, `vendor_id` and `vendor_profile_id` fields.
+- Automatic migrations of the Network Server database using `ns-db migrate` from versions prior to v3.23 are removed. Migrating from prior versions should be done through v3.23 instead.
 
 ## [3.23.2] - 2023-01-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ For details about compatibility between different releases, see the **Commitment
 
 - Network Server ID (NSID) used for Backend Interfaces interoperability via the `ns.interop.id` and `dcs.edcs.ns-id` configuration options.
   - In the Network Server, `ns.interop.id` acts as a fallback value for `sender-ns-id` in Join Server interoperability configuration.
+- Optional Network Server database migration that removes obsolete last invalidation keys is now available.
 
 ### Changed
 
@@ -48,7 +49,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Removed
 
 - The device version identifiers no longer have the `serial_number`, `vendor_id` and `vendor_profile_id` fields.
-- Automatic migrations of the Network Server database using `ns-db migrate` from versions prior to v3.23 are removed. Migrating from prior versions should be done through v3.23 instead.
+- Automatic migrations of the Network Server database using `ns-db migrate` from versions prior to v3.24 are removed. Migrating from prior versions should be done through v3.24 instead.
 
 ## [3.23.2] - 2023-01-18
 
--- a/cmd/ttn-lw-stack/commands/ns_db.go
+++ b/cmd/ttn-lw-stack/commands/ns_db.go
@@ -104,8 +104,11 @@ var (
 					return nil
 				}
 				if schemaVersion < nsredis.UnsupportedMigrationVersionBreakpoint {
-					return fmt.Errorf(
-						"you are currently running version %d of the Network Server database. This version cannot be auto-migrated to latest. Use v3.23.0 of The Things Stack to migrate to a supported one before migrating to latest", //nolint:lll
+					return fmt.Errorf( //nolint:stylecheck
+						"You are currently running version schema %d of the Network Server database. "+
+							"This version cannot be auto-migrated to latest. "+
+							"Database needs to be upgraded using v3.19 of The Things Stack before upgrading to latest. "+
+							"If you are not upgrading from versions before v3.19 please run with `--force` flag",
 						schemaVersion,
 					)
 				}
--- a/cmd/ttn-lw-stack/commands/ns_db.go
+++ b/cmd/ttn-lw-stack/commands/ns_db.go
@@ -15,17 +15,15 @@
 package commands
 
 import (
+	"fmt"
 	"regexp"
 	"time"
 
-	"github.com/redis/go-redis/v9"
 	"github.com/spf13/cobra"
-	"github.com/vmihailenco/msgpack/v5"
 	"go.thethings.network/lorawan-stack/v3/pkg/cleanup"
 	nsredis "go.thethings.network/lorawan-stack/v3/pkg/networkserver/redis"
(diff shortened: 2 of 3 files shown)


johanstokking:
> * Direct migrations of version 0 to version 2 of the Network Server db are not supported. Tested by manually setting the versions in the db and verified that the correct warning is displayed.

This breaks our compatibility commitment.

Why was this necessary? The referenced issue doesn't mention this.

adriansmares:
> Why was this necessary? The referenced issue doesn't mention this.

The context here is that the 0 -> 1 migration was done as part of release 3.11, 2 years ago. The migration itself is relevant only while upgrading from a pre 3.11 version to a post 3.11 version. I would say that it is unlikely that such deployments still exist in the wild, and if they do, they have a lot of configuration to update as well in two years of changes.

Regarding asking people to upgrade to an intermediary version first - we do have precedent as part of the `bunstore` changes done a year ago:
https://github.com/TheThingsNetwork/lorawan-stack/blob/166acb9c4b9746b894709daf6c9191c70736701f/pkg/identityserver/store/migrations/20220520000000_v3_20.go#L48-L54

We can try to keep this 0 -> 1 migration in place, but it is a very large monolithic migration which I don't think is still relevant today. Also, someone upgrading from 3.9.0 still needs to use an intermediary 3.19 version for the Identity Server upgrade, so they can do the `ns-db` too if needed.

johanstokking:
OK. I think it is acceptable, especially in this case, but let's make sure we're consistent going forward.

As for the changes here, there's only a warning and no exit code if one attempts to migrate an ancient NS database. Like the bunstore migration, I think we should really fail with a concise instruction on what to do.

adriansmares:
> As for the changes here, there's only a warning and no exit code if one attempts to migrate an ancient NS database. Like the bunstore migration, I think we should really fail with a concise instruction on what to do.

I agree with this sentiment but I am not sure how we would nicely achieve this. Versioning has been introduced later on, in version 3.14 (https://github.com/TheThingsNetwork/lorawan-stack/pull/4410). We've had no NS database migrations since the 3.11 one, so most people don't actually have the schema version key set at all - `eu2` for example does not have a schema version. 

What we can do is that we always error when the schema key is missing, and tell people to use `--force` if they are not upgrading from a version preceding 3.11 or if they have setup their deployment after version 3.11.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Removed
 
 - The device version identifiers no longer have the `serial_number`, `vendor_id` and `vendor_profile_id` fields.
+- Automatic migrations of the Network Server database using `ns-db migrate` from versions prior to v3.23 are removed. Migrating from prior versions should be done through v3.23 instead.
```
adriansmares: We also should mention that there is an optional migration available which deletes unused keys.

cmd/ttn-lw-stack/commands/ns_db.go
```diff
@@ -92,377 +90,64 @@ var (
...
 				}
 				if schemaVersion >= nsredis.SchemaVersion {
 					logger.Info("Database schema version is already in latest version")
 					return nil
 				}
+				if schemaVersion < nsredis.UnsupportedMigrationVersionBreakpoint {
+					return fmt.Errorf( //nolint:stylecheck
+						"You are currently running version schema %d of the Network Server database. "+
+							"This version cannot be auto-migrated to latest. "+
+							"Database needs to be upgraded using v3.19 of The Things Stack before upgrading to latest. "+
```
adriansmares: In the changelog we say v3.23, and here we use v3.19. I think we should say v3.24 in both (given that this lands in v3.25).

cmd/ttn-lw-stack/commands/ns_db.go
```diff
@@ -92,377 +90,64 @@ var (
...
-					logger.Debug("Skip valid key")
-					return true, nil
+			uidLastInvalidationKey := regexp.MustCompile(uplinkClient.Key("uid", ".*", "last-invalidation$"))
 
-				default:
-					d, err := cl.TTL(ctx, k).Result()
-					if err != nil {
-						logger.WithError(err).Error("Failed to determine TTL of unmatched key")
+			pipeliner := uplinkClient.Pipeline()
+			err := ttnredis.RangeRedisKeys(ctx, uplinkClient, uplinkClient.Key("*"), ttnredis.DefaultRangeCount,
```
adriansmares: Using `Key("*")` is very wide - you will iterate through the whole uplinks space. Wouldn't something like `Key("uid", "*", "last-invalidations")` be better ?

pkg/networkserver/redis/registry.go
```diff
@@ -41,7 +41,12 @@ var (
 )
 
 // SchemaVersion is the Network Server database schema version. Bump when a migration is required.
-const SchemaVersion = 1
+const SchemaVersion = 2
```
adriansmares: I think we should have a schema version for the devices (which is in theory at version 1), and one for the uplinks (which is now at 1 too).

adriansmares (approved):

adriansmares (changes requested):
Looks good in general, with some comments.

cmd/ttn-lw-stack/commands/ns_db.go
cvetkovski98: True. Updated it and removed the RegEx checking as it was not needed anymore because of the specific uplink space.

cmd/ttn-lw-stack/commands/ns_db.go
```diff
@@ -92,377 +89,70 @@ var (
 				panic("Only Redis is supported by this command")
 			}
 
-			logger.Info("Connecting to Redis database...")
-			cl := NewNetworkServerDeviceRegistryRedis(config)
+			logger.Info("Connecting to Network Server db ...")
```
adriansmares: Nit: `database`.

adriansmares (approved):

PR #6060, opened 2023-02-14
Limit downlink retry

### LINKED ISSUE #6054: Avoid retrying confirmed downlink message indefinitely
### Summary

We currently retry the same confirmed downlink message indefinitely. We should reconsider this design decision.

### Current Situation

The flow for the downlinks works as follows:

1. The Application Server receives a confirmed downlink via some API (MQTT / Webhooks etc.)
2. The downlink is passed to the Network Server, which schedules it for transmission.
3. The downlink is transmitted.
4. The end device does not hear the transmission.
5. On the next uplink from the end device, the uplink does not contain an acknowledgement (since the downlink was not 'heard' by the end device).
6. The Network Server generates a `DownlinkNack` event.
7. The Application Server processes this `DownlinkNack` and always re-pushes the downlink to the end of the queue:

https://github.com/TheThingsNetwork/lorawan-stack/blob/b6a4ebaa12e62b09bf0853a538b0d7651e3dc142/pkg/applicationserver/applicationserver.go#L1321-L1339

The problem is that this process is infinite if the device never acknowledges the downlink - the Network Server is forced to queue it indefinitely, thus using precious air time for downlinks which likely may never be acknowledged.

We want to avoid sending the downlinks indefinitely to these end devices. Problematic end devices should not cause infinite downlink attempts.

### Proposed Implementation

1. Add a new message under the `ApplicationDownlink` message, with the following format:
```proto
  message ConfirmedRetry {
    option (thethings.flags.message) = { select: true, set: true };

    // The number of confirmed downlink acknowledgement attempts done so far.
    uint32 attempt = 1;
    // The maximum number of confirmed downlink acknowledgement attempts.
    // If null, the Application Server configuration is used instead.
    google.protobuf.UInt32Value max_attempts = 2;
  }

  ConfirmedRetry confirmed_retry = 11;

  // next: 12
```
2. Add an Application Server configuration section for `downlinks`, which should have a `confirmed` subsection.
3. Under this subsection, we need two settings - one for the default amount of confirmed retry attempts (`default-retry-attempts`), and one for the maximum amount of retry attempts that we allow users to configure (`max-retry-attempts`).
`default-retry-attempts` is used as `max_attempts` when `max_attempts` is `null`, and when `max_attempts` is present, we should validate that it is <= `max-retry-attempts`.
4. When a downlink is pushed/replaced, we should ensure that it has a `confirmed_retry` message set if it is a confirmed downlink. By default, the number of attempts is zero and the configured `max_attempts` is used.
We can do this injection as part of `as.downlinkQueueOp`, after we decode / encode the downlink:
https://github.com/TheThingsNetwork/lorawan-stack/blob/b6a4ebaa12e62b09bf0853a538b0d7651e3dc142/pkg/applicationserver/applicationserver.go#L647-L659
5. As part of the `DownlinkNack` event handling, we should stop retrying if `attempt >= max_attempts`, and increment the attempt number otherwise. We also should be conscious that existing downlinks were inserted before we started 'injecting' `confirmed_retry`, so we should fill the `confirmed_retry` field if it is `null` too.

### Contributing

### Code of Conduct

cvetkovski98:
#### Summary
Introduce the downlink retry config in the AS to avoid retrying downlinks indefinitely.

Closes #6054 

#### Changes

- Add `Downlinks` section in the AS config with appropriate defaults
- Extend the `ApplicationDownlink` proto message with the retry attempt

#### Testing
- Manual

##### Regressions
- Confirmed downlinks that are currently being retried will receive a default retry counter of 8 and drop afterwards.

#### Notes for Reviewers

...

#### Checklist


pkg/applicationserver/applicationserver.go
```diff
@@ -105,7 +106,22 @@ func (as *ApplicationServer) Context() context.Context {
...
+var (
+	errListenFrontend = errors.DefineFailedPrecondition(
+		"listen_frontend", "failed to start frontend listener `{protocol}` on address `{address}`",
+	)
+	errMaxRetriesReached = errors.DefineAborted("max_retries_reached", "maximum number of retries reached")
+)
+
+func validateDownlinkConfirmationConfig(c ConfirmationConfig) error {
+	if c.DefaultRetryAttempts == 0 && c.MaxRetryAttempts == 0 {
+		return errRetryAttemptCountMissing
```
adriansmares: `.New()`, and same below.

pkg/applicationserver/applicationserver.go
```diff
@@ -643,6 +673,24 @@ func (as *ApplicationServer) attemptDownlinkQueueOp(ctx context.Context, dev *tt
...
+	maxRetries := as.config.Downlinks.ConfirmationConfig.DefaultRetryAttempts
+	if as.config.Downlinks.ConfirmationConfig.MaxRetryAttempts != 0 {
+		maxRetries = as.config.Downlinks.ConfirmationConfig.MaxRetryAttempts
+	}
+
+	for _, item := range items {
+		if item.Confirmed && item.ConfirmedRetry == nil {
+			item.ConfirmedRetry = &ttnpb.ApplicationDownlink_ConfirmedRetry{
+				Attempt:     1,
+				MaxAttempts: wrapperspb.UInt32(uint32(maxRetries)),
```
adriansmares: I wouldn't initialize `max_attempts` and instead just build it dynamically in the nack handler.

pkg/applicationserver/applicationserver.go
```diff
@@ -643,6 +673,24 @@ func (as *ApplicationServer) attemptDownlinkQueueOp(ctx context.Context, dev *tt
 	}
 }
 
+// initializeConfirmationRetriesConfig initializes the confirmation retries
+// configuration for the given downlink items.
+func (as *ApplicationServer) initializeConfirmationRetriesConfig(items []*ttnpb.ApplicationDownlink) {
+	maxRetries := as.config.Downlinks.ConfirmationConfig.DefaultRetryAttempts
+	if as.config.Downlinks.ConfirmationConfig.MaxRetryAttempts != 0 {
```
adriansmares: This is wrong, because you always use max instead of default when max is not zero.

pkg/applicationserver/applicationserver.go
```diff
@@ -1352,6 +1403,15 @@ func (as *ApplicationServer) handleDownlinkNack(ctx context.Context, ids *ttnpb.
 			}
 
 			items := []*ttnpb.ApplicationDownlink{msg}
+
+			if msg.Confirmed && msg.ConfirmedRetry != nil {
+				msg.ConfirmedRetry.Attempt++
+				if msg.ConfirmedRetry.Attempt > msg.ConfirmedRetry.MaxAttempts.GetValue() {
```
adriansmares: I would build the max attempts dynamically here:
1. If `confirmed_retry.max_attempts` is not null, validate it against the configured `max_attempts`.
2. Fallback to the default max attempts if it is not zero.
3. If both default and max are zero, do not drop.

The point is to avoid storing `max_attempts` unless the user explicitly provided it. This allows us to retroactively introduce the `max_attempts` for messages later on by editing the AS configuration.

---

I would also say that existing messages with `confirmed_retry == nil` should have their attempt counter initialized retroactively.

pkg/ttnpb/messages.go
```diff
@@ -51,6 +51,20 @@ func (v *ApplicationDownlink_ClassBC) FieldIsZero(p string) bool {
...
 
+// FieldIsZero returns whether path p is zero.
+func (v *ApplicationDownlink_ConfirmedRetry) FieldIsZero(p string) bool {
+	if v == nil {
+		return true
+	}
+	switch p {
+	case "attempt":
+		return v.Attempt == 0
+	case "max_attempts":
```
adriansmares, line 62: This should be `v.MaxAttempts == nil`.

`FieldIsZero` is related to the zero value concept of Go, where the zero value of a pointer is `nil`, not the zero value of the value of the pointer.  Using `GetValue` conflates `&{Value: 0}` with `nil`.

pkg/applicationserver/applicationserver.go
```diff
@@ -105,7 +105,27 @@ func (as *ApplicationServer) Context() context.Context {
...
+	errListenFrontend = errors.DefineFailedPrecondition(
+		"listen_frontend", "failed to start frontend listener `{protocol}` on address `{address}`",
+	)
+	errMaxRetriesReached = errors.DefineAborted(
+		"max_retries_reached", "maximum number of retries ({max_retries}) reached",
+	)
+)
+
+func validateDownlinkConfirmationConfig(c ConfirmationConfig) error {
+	if c.DefaultRetryAttempts == 0 && c.MaxRetryAttempts == 0 {
```
adriansmares: We should add default configuration for these settings. I would say that by default 8 retries is more than sufficient to avoid people complaining about backwards incompatibility, and 32 as a max should be fine.

Also I would say that both values are required to be non zero.

pkg/applicationserver/applicationserver.go
```diff
@@ -105,7 +105,27 @@ func (as *ApplicationServer) Context() context.Context {
...
+		return errRetryAttemptCountMissing.New()
+	}
+	if c.MaxRetryAttempts < 0 {
```
adriansmares: We also could just mark these fields as unsigned.

pkg/applicationserver/applicationserver.go
```diff
@@ -105,7 +105,27 @@ func (as *ApplicationServer) Context() context.Context {
...
+		return errNegativeValue.WithAttributes("field", "max_retry_attempts")
+	}
+	if c.DefaultRetryAttempts < 0 {
+		return errNegativeValue.WithAttributes("field", "default_retry_attempts")
+	}
+	return nil
```
adriansmares, line 124: We should check that `default <= max`.

pkg/applicationserver/applicationserver.go
```diff
@@ -1352,6 +1414,28 @@ func (as *ApplicationServer) handleDownlinkNack(ctx context.Context, ids *ttnpb.
...
+				if msg.ConfirmedRetry == nil {
+					msg.ConfirmedRetry = &ttnpb.ApplicationDownlink_ConfirmedRetry{
+						Attempt: 0,
+					}
+				}
+				msg.ConfirmedRetry.Attempt++
+
+				maxRetries := uint32(as.config.Downlinks.ConfirmationConfig.DefaultRetryAttempts)
+				if msg.ConfirmedRetry.MaxAttempts.GetValue() > 0 {
+					maxRetries = msg.ConfirmedRetry.MaxAttempts.GetValue()
```
adriansmares: Can use `.Value` directly, and you may consider having a variable directly in the `if` definition, like `if maxAttempts := msg.ConfirmedRetry.MaxAttempts; maxAttempts != nil`.

api/messages.proto
```diff
@@ -384,7 +384,19 @@ message ApplicationDownlink {
...
 
-  // next: 11
+  message ConfirmedRetry {
+    option (thethings.flags.message) = { select: true, set: true };
+
+    // The number of confirmed downlink acknowledgement attempts done so far.
+    uint32 attempt = 1;
+    // The maximum number of confirmed downlink acknowledgement attempts.
+    // If null, the Application Server configuration is used instead.
+    google.protobuf.UInt32Value max_attempts = 2;
```
adriansmares: I think we should ensure that this is greater than zero, and probably limit this to some large value (like 100).

pkg/applicationserver/applicationserver.go
```diff
@@ -1352,6 +1414,28 @@ func (as *ApplicationServer) handleDownlinkNack(ctx context.Context, ids *ttnpb.
...
+				}
+				if maxRetries != 0 && msg.ConfirmedRetry.Attempt > maxRetries {
```
adriansmares: In which situations would this be zero, if we make the AS configuration not accept zeroes for the default or (validation) max ?

pkg/applicationserver/config.go
```diff
@@ -93,6 +93,17 @@ type FormattersConfig struct {
 	MaxParameterLength int `name:"max-parameter-length" description:"Maximum allowed size for length of formatter parameters (payload formatter scripts)"`
 }
 
+// ConfirmationConfig represents the configuration for downlink confirmations.
+type ConfirmationConfig struct {
+	DefaultRetryAttempts int `name:"default-retry-attempts" description:"Default number of retry attempts for downlink confirmation"` // nolint:lll
```
adriansmares: `for a confirmed downlink`.

pkg/applicationserver/config.go
```diff
@@ -93,6 +93,17 @@ type FormattersConfig struct {
...
+	MaxRetryAttempts     int `name:"max-retry-attempts" description:"Maximum number of retry attempts for downlink confirmation"`     // nolint:lll
```
adriansmares: `for a confirmed downlink`.

pkg/applicationserver/applicationserver.go
adriansmares: This is not solved, but I can see the confusion. Neither of the values should be zero (`||` not `&&`).

pkg/applicationserver/applicationserver.go
```diff
@@ -1352,6 +1412,28 @@ func (as *ApplicationServer) handleDownlinkNack(ctx context.Context, ids *ttnpb.
 			}
 
 			items := []*ttnpb.ApplicationDownlink{msg}
+
+			if msg.Confirmed {
```
adriansmares: I think this `if` is redundant - it is not possible (unless a bug occurs) that `msg.Confirmed` is false if it arrives via a `NACK`.

adriansmares (approved):
Looks great ! I've changed the base `v3.27` as this changes the API and behavior.

Please add a `CHANGELOG` entry regarding the old and new behavior, and the configuration options which have been added.

CHANGELOG.md
```diff
@@ -11,8 +11,13 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Added
 
+- Each `ApplicationDownlink` message now has the `Attempt` and `MaxAttempts` fields to indicate the current and maximum number of attempts for a specific **confirmed downlink**.
+- The Application Server configuration now has the `Downlinks` section and its `ConfirmedRetry` subsection that configures the amount of `MaxAttempts` for confirmed downlinks. The default values are `8` for the `default-retry-attempts` used when the `MaxAttempts` field is not set and `32` for the `max-retry-attempts` used to limit the `MaxAttempts` field.
```
KrishnaIyer: Though these items are "added", they are technically part of a "change" in behaviour. 
So you can move these items as sub-items to the `Changed` description below.

api/messages.proto
```diff
@@ -384,7 +384,19 @@ message ApplicationDownlink {
 
   repeated string correlation_ids = 9 [(validate.rules).repeated.items.string.max_len = 100];
 
-  // next: 11
+  message ConfirmedRetry {
+    option (thethings.flags.message) = { select: true, set: true };
+
+    // The number of confirmed downlink acknowledgement attempts done so far.
```
KrishnaIyer: `The number of attempted confirmed downlink acknowledgements.` is clearer.

cmd/internal/shared/applicationserver/config.go
```diff
@@ -100,4 +100,10 @@ var DefaultApplicationServerConfig = applicationserver.Config{
 		BatchSize:     1000,
 		FlushInterval: 10 * time.Second,
 	},
+	Downlinks: applicationserver.DownlinksConfig{
+		ConfirmationConfig: applicationserver.ConfirmationConfig{
+			DefaultRetryAttempts: 8,
+			MaxRetryAttempts:     32,
```
KrishnaIyer, line 106: What are these defaults based on?

api/messages.proto
```diff
@@ -384,7 +384,19 @@ message ApplicationDownlink {
...
+    uint32 attempt = 1;
+    // The maximum number of confirmed downlink acknowledgement attempts.
+    // If null, the Application Server configuration is used instead.
+    google.protobuf.UInt32Value max_attempts = 2 [(validate.rules).uint32 = {gt: 0, lte: 100}];
```
KrishnaIyer, line 394: Is this configurable per message to allow setting  a lower limit than the cluster config for specific devices?
What is the advantage of not directly using the value in the configuration as max?

pkg/applicationserver/applicationserver.go
```diff
@@ -105,7 +105,24 @@ func (as *ApplicationServer) Context() context.Context {
...
+)
+
+func validateDownlinkConfirmationConfig(c ConfirmationConfig) error {
+	if c.DefaultRetryAttempts == 0 || c.MaxRetryAttempts == 0 {
+		return errZeroAttemptCounts.New()
+	}
+	if c.DefaultRetryAttempts > c.MaxRetryAttempts {
+		return errDefaultRetryCountLargerThanMax.New()
+	}
+	return nil
```
KrishnaIyer, line 124: I don't get why `DefaultRetryAttempts` and `MaxRetryAttempts` are both user configurable. Shouldn't the default be hardcoded in the software and checked against a user-set `MaxRetryAttempts`?

pkg/applicationserver/applicationserver.go
```diff
@@ -1352,6 +1412,26 @@ func (as *ApplicationServer) handleDownlinkNack(ctx context.Context, ids *ttnpb.
 			}
 
 			items := []*ttnpb.ApplicationDownlink{msg}
+
+			if msg.ConfirmedRetry == nil {
+				msg.ConfirmedRetry = &ttnpb.ApplicationDownlink_ConfirmedRetry{
+					Attempt: 0,
+				}
+			}
+			msg.ConfirmedRetry.Attempt++
```
KrishnaIyer, line 1419: Though `msg` is a pointer, you're updating it after it's added to the `items` which is then passed to `attemptDownlinkQueueOp`.
Or am I understanding this wrongly?

cmd/internal/shared/applicationserver/config.go
adriansmares: > What are these defaults based on?

Nothing scientific. The old behavior was trying indefinitely, while these defaults try to add a limit without it being too visible to clients that use confirmed downlinks.

api/messages.proto
adriansmares: > Is this configurable per message to allow setting a lower limit than the cluster config for specific devices?

That is the point indeed. The cluster default may be higher, but the user can choose to limit this to 1 and do the retry themselves.

pkg/applicationserver/applicationserver.go
adriansmares: > I don't get why `DefaultRetryAttempts` and `MaxRetryAttempts` are both user configurable. Shouldn't the default be hardcoded in the software and checked against a user-set `MaxRetryAttempts`?

Default is used when the user did not provide an explicit value, while max ensures that the user cannot just do an explicit limit of `999999` - we don't want people to retry indefinitely.

CHANGELOG.md
```diff
@@ -11,8 +11,13 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Added
 
+- Each `ApplicationDownlink` message now has the `Attempt` and `MaxAttempts` fields to indicate the current and maximum number of attempts for a specific **confirmed downlink**.
```
adriansmares: People generally interact with our JSON API, where the field names are `attempt` and `max_attempts`. I would prefer to keep that convention in order to avoid confusion. I would also say just `application downlink`.

CHANGELOG.md
adriansmares: I would also explain the difference between default and max (i.e. why we have a default). Also `Downlinks` and `ConfirmedRetry` do not exist for users or stack operators - in YAML they see only `downlinks` and `confirmed-retry`, and generally we use the full path that would appear in `--help`.

I also have the same comment regarding the field naming for the proto fields.

CHANGELOG.md
```diff
@@ -109,7 +114,7 @@ For details about compatibility between different releases, see the **Commitment
 - Device claiming that transfer devices between applications is now deprecated and will be removed in a future version of The Things Stack. Device claiming on Join Servers, including The Things Join Server, remains functional. This deprecates the following components:
   - API for managing application claim authorization (`EndDeviceClaimingServer.AuthorizeApplication` and `EndDeviceClaimingServer.UnauthorizeApplication`)
   - CLI commands to manage application claim settings (`ttn-lw-cli application claim [authorize|unauthorize]`)
-  - CLI command to claim end devices  (`ttn-lw-cli devices claim`)
+  - CLI command to claim end devices (`ttn-lw-cli devices claim`)
```
adriansmares: These changes are not supposed to be here.

pkg/applicationserver/applicationserver.go
cvetkovski98: Yes, it's updated after being added to the array, but indeed, it is a pointer, so the value of `msg.ConfirmedRetry.Attempt` is updated before invoking `attemptDownlinkQueueOp`. It is also passed in the `registerDropDownlinks`.

Would you rather see the update before the addition to the array?

KrishnaIyer: Yes please. Though the end result is the same, in the current implementation the update is a side-effect rather than being intentional.

pkg/applicationserver/applicationserver.go
KrishnaIyer: Sure but my question is this; why is `DefaultRetryAttempts` user configurable?

adriansmares: There are 2 configuration options:
1. `DefaultRetryAttempts`, which is used as a limit when the user did not provide a maximum amount of retry attempts explicitly in the downlink. User here is the application which submits downlinks.
2. `MaxRetryAttempts`, which is used to validate the user provided maximum amount of retry attempts provided explicitly in a downlink. User here is the application which submits downlinks.

Since `DefaultRetryAttempts` is used when the user did not provide an explicitly limit of retry attempts, to be at least it makes sense to have it configurable. Hardcoding it forces us to re-release the stack in case we want to change the behavior that clients which do not provide an explicit limit have - I don't see why that is preferable.

---

Or is the naming the one that threw you off and you would prefer something like `DefaultMaxRetryAttempts` and `MaxRetryAttemptsLimit` ?

KrishnaIyer: > DefaultRetryAttempts, which is used as a limit when the user did not provide a maximum amount of retry attempts explicitly in the downlink.

Ok why not use `config.MaxRetryAttempts` in this case?

What would be the downside to the following: 
- If user sets the limit explicitly in `ApplicationDownlink`, check that it's not greater than `config.MaxRetryAttempts`
- If user does not set the limit explicitly in `ApplicationDownlink`, use `config.MaxRetryAttempts` as the limit.
And when we read in config, check that `config.MaxRetryAttempts` is not zero.

We set the default value of `config.MaxRetryAttempts`  in software and the cluster admin can change this value if necessary.

adriansmares: Ok I see, and this ties to your question regarding 'how you came up with these defaults'.

The problem in an abstract is that we lack insight into how many retries somebody needs, and probably that number cannot be universal. Before this change, we retried indefinitely and we didn't have any book keeping on how many attempts did we have so far for a particular downlink. Because of this, we do not really know how many attempts make for a good default.

If we have a singular configuration option, that value may end up being too small (and you cannot request more than the configuration option), or too big (and you retry the same downlink too many times even though it is a lost cause).

If we set the limit too low, the application developers need to add special handling that takes the failed downlink and enqueues it again. If we set it too high, we send the same downlink a lot of times even though it is probably a lost cause.

Having this _default_ in between tries to simplify the transition process between the infinite retries world and the finite retries world. My expectation is that most confirmed downlinks do not need that many retries, so having a smaller default is a sane default. If the user really needs a higher value, the configuration max attempts allows him to provide a higher explicit value without needing to add extra logic that would take the failed downlink and submit it again, manually.

Let me know your thoughts on this.

KrishnaIyer: Ok so the problem as I understand, is that you're trying to keep the default no of retransmissions as low as possible while at the same time trying to cap the maximum (per-message) value to something sane.

This looks good to me then.

An alternative approach would be that the max retries would be a configurable multiple of the default but that's ultimately the same idea.

KrishnaIyer (approved):

adriansmares (approved):

PR #6080, opened 2023-02-27
Add default error values for backend-origin messages

### LINKED ISSUE #6003: Missing attributes in live data for gateways
### Summary

In case you get a **Failed to schedule downlink for transmission by gateway** event, the attributes are not shown in the data preview in the console.

gateway_console

### Steps to Reproduce

1. Setup a gateway with a 3G connection
2. Use RX1_delay = 1s for end device
3. Schedule a downlink

This is the reason why I'm getting this message, maybe you have an easier way to trigger it. 

### Current Result

gateway_console

### Expected Result

Missing attributes are shown

### Relevant Logs

```shell
{
  "name": "gs.down.schedule.fail",
  "time": "2023-01-05T12:28:51.915770433Z",
  "identifiers": [
    {
      "gateway_ids": {
        "gateway_id": "gw135",
        "eui": "80029CFFFECFC527"
      }
    }
  ],
  "data": {
    "namespace": "pkg/gatewayserver/io",
    "name": "tx_schedule",
    "message_format": "failed to schedule",
    "correlation_id": "c761ec5ecc9945e7b89f159bec530563",
    "code": 10,
    "details": [
      {
        "@type": "type.googleapis.com/ttn.lorawan.v3.ScheduleDownlinkErrorDetails",
        "path_errors": [
          {
            "namespace": "pkg/gatewayserver/io",
            "name": "rx_window_schedule",
            "message_format": "schedule in Rx window `{window}` failed",
            "attributes": {
              "window": 1
            },
            "correlation_id": "0b9bf59c9d9348bb9f726b3d9f83b74f",
            "cause": {
              "namespace": "pkg/gatewayserver/scheduling",
              "name": "too_late",
              "message_format": "too late to transmission scheduled time (delta is `{delta}`, min is `{min}`)",
              "attributes": {
                "min": "1.771613169s"
              },
              "correlation_id": "69ed7acf7d0a424f860ec52971c97b78",
              "code": 9
            },
            "code": 9
          },
          {
            "namespace": "pkg/gatewayserver/io",
            "name": "rx_window_schedule",
            "message_format": "schedule in Rx window `{window}` failed",
            "attributes": {
              "window": 2
            },
            "correlation_id": "39d14f90c43549e5b005cc6404d424ca",
            "cause": {
              "namespace": "pkg/gatewayserver/scheduling",
              "name": "blocked",
              "message_format": "sub band is blocked for `{duration}`",
              "attributes": {
                "duration": "2.823088s"
              },
              "correlation_id": "28a0826f724a4b2f9e56ba7316b751be",
              "code": 8
            },
            "code": 8
          }
        ]
      }
    ]
  },
  "correlation_ids": [
    "gs:conn:01GNZCQR04GBQPQ0H1XK3GXQ38",
    "gs:up:host:01GNZCQR0838AFTRAJ7V2RZ297",
    "gs:uplink:01GP0YFH3C903ENMPZSPG37N8P",
    "ns:downlink:01GP0YFMTBAPB9JSH6DTFEBAXP",
    "ns:transmission:01GP0YFMTBENSZWZ0D4NTF6MPF",
    "ns:uplink:01GP0YFH3DDW1EKK1E0PW7DPVB",
    "rpc:/ttn.lorawan.v3.GsNs/HandleUplink:01GP0YFH3DF8GGDTQHYSD48RKE",
    "rpc:/ttn.lorawan.v3.NsGs/ScheduleDownlink:01GP0YFMTBCQYYT2SSMSWRMHZ2"
  ],
  "origin": "ip-10-23-14-132.eu-west-1.compute.internal",
  "context": {
    "tenant-id": "CglzYW5pbnVkZ2U="
  },
  "visibility": {
    "rights": [
      "RIGHT_GATEWAY_TRAFFIC_READ"
    ]
  },
  "unique_id": "01GP0YFMTBRNHR3YTVY59QB95R"
}
```

### URL

_No response_

### Deployment

The Things Stack Cloud

### The Things Stack Version

3.23.1

### Client Name and Version

_No response_

### Other Information

_No response_

### Proposed Fix

_No response_

### Contributing

- [ ] I can help by implementing a fix after the proposal above is approved.
- [ ] I can help by testing the fix before it's released.

### Code of Conduct

ryaplots:
#### Summary

Closes #6003
Complements #6078 

#### Changes

- Add default values for messages coming from the backend

#### Testing

e2e and manually (use payload from the [issue](https://github.com/TheThingsNetwork/lorawan-stack/issues/6003))

#### Notes for Reviewers

`Intl` throws an error if one of the values of a message is missing. The way I found to avoid this was to add a default, to which the value falls back if not set in attributes by the backend.

So for this payload:
```
"cause": {
       "namespace": "pkg/gatewayserver/scheduling",
       "name": "too_late",
       "message_format": "too late to transmission scheduled time (delta is `{delta}`, min is `{min}`)",
       "attributes": {
              "min": "1.771613169s"
       },
       "correlation_id": "69ed7acf7d0a424f860ec52971c97b78",
       "code": 9
},
```
if `delta` is not present it will have the value of the default fallback.

My question is: is this really necessary? Aren't these attributes required? Are there situations where the attribute won't be there, other than a bug? @adriansmares @kschiffer 

#### Checklist


adriansmares:
> My question is: is this really necessary? Aren't these attributes required? Are there situations where the attribute won't be there, other than a bug? @adriansmares @kschiffer

The attributes are unfortunately not required, and can be optional. In this case, it was a bug, but otherwise I think we should fail more softly (and maybe generate a Sentry error) if an attribute is missing.

Is it not possible to make `Intl` apply only present attributes and error for the remaining ones, instead of having defaults ? (which are also error prone, as they are just strings which are decoupled from the backend).

ryaplots:
I've spent some time looking through the [`Intl` documentation](https://formatjs.io/docs/react-intl/components/#message-formatting-fallbacks) and I couldn't find a way to fail softly.

We could generate default values if the placeholders in the message are more that the values in attributes, extracting `{value}` from the message, and also report the error to Sentry if this occurs and if `Intl` fails. See my last commits.
@kschiffer

pkg/webui/lib/components/with-locale.js
```diff
@@ -191,6 +193,10 @@ const LocaleLoader = ({ children }) => {
 
   const handleIntlError = useCallback(err => {
     error(err)
+    if (err.code === ReactIntlErrorCode.FORMAT_ERROR && !isDevelopment) {
+      ingestError(err, { ingestedBy: 'IntlFormat' })
+      return
```
kschiffer:
```suggestion
```

kschiffer: One problem with this is that this error will potentially fire hundreds of times when the message gets rerendered a lot. But let's keep it and see how much of a problem this is.

pkg/webui/lib/errors/utils.js
```diff
@@ -501,12 +501,22 @@ export const toMessageProps = (error, each = false) => {
...
         const rootCause = getBackendErrorRootCause(errorDetails)
+
+        // If the attributes are missing values, use the generated default values based on the message.
+        const messageValues = rootCause.message_format.match(/[^{}]+(?=})/g) || []
+        const hasMissingValues =
+          rootCause.attributes && Object.keys(rootCause.attributes).length < messageValues.length
+        const missingValues = messageValues.filter(val => !(val in rootCause.attributes))
+        const defaultErrorValues = hasMissingValues
+          ? missingValues.reduce((obj, val) => ((obj[val] = `{${val}}`), obj), {})
+          : {}
```
kschiffer:
```suggestion
        const messageValues = rootCause?.message_format?.match(/[^{}]+(?=})/g) || []
        const fallbackValues = messageValues.reduce((obj, val) => ((obj[val] = `{${val}}`), obj), { ...rootCause.attributes })
        const values = { ...fallbackValues, ...rootCause.attributes }
```
Let's just always spread the existing values over the missing values.

pkg/webui/lib/errors/utils.js
```diff
@@ -501,12 +501,22 @@ export const toMessageProps = (error, each = false) => {
...
+
         props.push({
           content: {
             id: getBackendErrorDetailsId(rootCause),
             defaultMessage: rootCause.message_format,
           },
-          values: rootCause.attributes,
+          values: { ...rootCause.attributes, ...defaultErrorValues },
```
kschiffer:
```suggestion
          values,
```

kschiffer (changes requested):
See my small suggestion. Please run some tests on staging. Any issues we introduce here can be very catastrophic since we use messages just about anywhere.

pkg/webui/lib/errors/utils.js
```diff
@@ -501,12 +501,20 @@ export const toMessageProps = (error, each = false) => {
       if (hasCauses(errorDetails)) {
         // Use the root cause if any.
         const rootCause = getBackendErrorRootCause(errorDetails)
+
+        // If the attributes are missing values, use the generated default values based on the message.
+        const messageValues = rootCause?.message_format?.match(/[^{}]+(?=})/g) || []
+        const fallbackValues = messageValues.reduce((obj, val) => ((obj[val] = `{${val}}`), obj), {
+          ...rootCause.attributes,
+        })
+        const values = { ...fallbackValues, ...rootCause.attributes }
```
kschiffer:
```suggestion
        const values = messageValues.reduce((obj, val) => {
          if (!(val in obj)) {
            obj[val] = val
          }
          return obj
        }, { ...rootCause.attributes })
```

kschiffer: Sorry, my last suggestion was odd.

kschiffer (changes requested):

kschiffer (approved):

PR #6082, opened 2023-02-28
Convert correlation ID sets to sorted sets

adriansmares:
#### Summary

This PR ensures that our correlation ID sets do not contain stale data for prolonged periods of time.

Certain correlation IDs are 'long lived' - for example each gateway connection has a stable correlation ID of form `gs:conn:xxx`. Events which reference this correlation ID may be long gone by the time someone calls `FindRelated`, but we do not have a mechanism to evict these references.

This PR changes the encoding of this 'correlation ID to event set mapping' to use sorted sets, where the key is the event time. This enables us to evict not only over space (last N events), but also over time (events from the last N minutes).

#### Changes

- Change the hash method for correlation ID to `fnv1a`, instead of `sha1`. This is more space efficient and time efficient.
- Store the correlation ID hash in base64 instead of base32.
- Store event IDs in sorted sets instead of lists, with the event time as key.
- Evict 'old' events on event push, based on the TTL.

#### Testing

Local testing.

##### Regressions

The indexing key has been changed, which means that events that precede this upgrade will not be available via `FindCorrelated`. As the TTL is 10 minutes, I believe this is a reasonable trade off.

#### Notes for Reviewers

This PR targets `v3.25` as changing the data types breaks the compatibility commitment: if I targeted `v3.24.2`, `v3.24.1` components would store the event IDs in a different key, and patch releases are meant to be backwards compatible.

#### Checklist

### DIFF
--- a/pkg/events/redis/store.go
+++ b/pkg/events/redis/store.go
@@ -16,9 +16,9 @@ package redis
 
 import (
 	"context"
-	"crypto/sha1"
-	"encoding/base32"
+	"encoding/base64"
 	"errors"
+	"hash/fnv"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


johanstokking:
> ##### Regressions
> 
> The indexing key has been changed, which means that events that precede this upgrade will not be available via `FindCorrelated`. As the TTL is 10 minutes, I believe this is a reasonable trade off.
> #### Notes for Reviewers
> 
> This PR targets `v3.25` as changing the data types breaks the compatibility commitment: if I targeted `v3.24.2`, `v3.24.1` components would store the event IDs in a different key, and patch releases are meant to be backwards compatible.

Perfect

johanstokking (approved):

PR #6090, opened 2023-03-03
TS003 LoRa Application Layer Clock Sync

### LINKED ISSUE #6065: Support Application Layer Clock Synchronization
### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/issues/3244

We should implement support for the Application Layer Clock Synchronization protocol via an application package.

### Current Situation

We do not support [TS003](https://resources.lora-alliance.org/document/ts003-2-0-0-application-layer-clock-synchronization) at all.

We need this in order to allow end devices to synchronize their clocks at application layer in a standardized fashion.

### Proposed Implementation

The initial implementation of the application package can be stateless. It should just respond to `AppTimeReq` via an `AppTimeAns` response.

Subsequent support for `DeviceAppTimePeriodicityReq`, `DeviceAppTimePeriodicityAns` and `ForceDeviceResyncCmd` can be added in a subsequent issue.

---

From a stack implementation perspective, the `alcsync-v1` is very close to existing packages such as the storage integration or LoRaCloud packages - you will receive the request on the application package port and you can automatically enqueue a downlink which responds to the request.

### Contributing

### Code of Conduct

cvetkovski98:
#### Summary
Implement LoRa Application Level Clock Sync 

Closes #6065

#### Changes
- Add the `alcsync-v1` package to AS

#### Testing
- UT
- Manual using Console simulate

##### Regressions
...

#### Notes for Reviewers

...

#### Checklist

### DIFF
--- a/pkg/applicationserver/io/packages/alcsync/v1/command.go
+++ b/pkg/applicationserver/io/packages/alcsync/v1/command.go
@@ -1,0 +1,33 @@
+// Copyright © 2023 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package alcsyncv1
+
+import (
+	"context"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+// Command is the interface for commands.
+type Command interface {
+	// Code returns the command code.
+	Code() uint8
+
+	// Execute runs the command logic.
+	Execute(
+		ctx context.Context,
+		onDownlink func(context.Context, *ttnpb.ApplicationDownlink) error,
+	) error
+}
--- a/pkg/applicationserver/io/packages/alcsync/v1/command.go
+++ b/pkg/applicationserver/io/packages/alcsync/v1/command.go
@@ -15,8 +15,6 @@
 package alcsyncv1
 
 import (
-	"context"
-
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 )
 
@@ -26,8 +24,5 @@ type Command interface {
 	Code() uint8
 
 	// Execute runs the command logic.
-	Execute(
-		ctx context.Context,
-		onDownlink func(context.Context, *ttnpb.ApplicationDownlink) error,
-	) error
+	Execute() ([]*ttnpb.ApplicationDownlink, error)
 }
--- a/pkg/applicationserver/io/packages/alcsync/v1/command.go
+++ b/pkg/applicationserver/io/packages/alcsync/v1/command.go
@@ -14,15 +14,16 @@
 
 package alcsyncv1
 
-import (
-	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
-)
+// Result is the result of a command execution.
+type Result interface {
+	// MarshalBinary marshals the result into a binary representation.
+	MarshalBinary() ([]byte, error)
+}
 
 // Command is the interface for commands.
 type Command interface {
 	// Code returns the command code.
 	Code() uint8
-
 	// Execute runs the command logic.
-	Execute() ([]*ttnpb.ApplicationDownlink, error)
+	Execute() (Result, error)
 }
--- a/pkg/applicationserver/io/packages/alcsync/v1/data.go
+++ b/pkg/applicationserver/io/packages/alcsync/v1/data.go
@@ -27,25 +27,12 @@ var errInvalidFieldType = errors.DefineCorruption("invalid_field_type", "field `
 var defaultThreshold = time.Duration(4) * time.Second
 
 type packageData struct {
-	FPort     uint32
 	Threshold time.Duration
 }
 
 func (d *packageData) fromStruct(st *structpb.Struct) error {
 	fields := st.GetFields()
-	value, ok := fields["f_port"]
-	if ok {
-		numberValue, ok := value.GetKind().(*structpb.Value_NumberValue)
-		if !ok {
-			return errInvalidFieldType.WithAttributes(
-				"field", "f_port",
-				"type", "number",
-			)
-		}
-		d.FPort = uint32(numberValue.NumberValue)
-	}
-
-	value, ok = fields["threshold"]
+	value, ok := fields["threshold"]
 	if ok {
 		numberValue, ok := value.GetKind().(*structpb.Value_NumberValue)
 		if !ok {
@@ -62,29 +49,27 @@ func (d *packageData) fromStruct(st *structpb.Struct) error {
 func mergePackageData(
 	def *ttnpb.ApplicationPackageDefaultAssociation,
 	assoc *ttnpb.ApplicationPackageAssociation,
-) (*packageData, error) {
+) (*packageData, uint32, error) {
 	var defaultData, associationData packageData
 	if err := defaultData.fromStruct(def.GetData()); err != nil {
-		return nil, err
+		return nil, 0, err
 	}
 	if err := associationData.fromStruct(assoc.GetData()); err != nil {
-		return nil, err
+		return nil, 0, err
 	}
 
-	merged := &packageData{}
+	merged := &packageData{
+		Threshold: defaultThreshold,
+	}
 	for _, data := range []packageData{defaultData, associationData} {
-		if data.FPort != 0 {
-			merged.FPort = data.FPort
-		}
 		if data.Threshold != 0 {
 			merged.Threshold = data.Threshold
 		}
 	}
-	if merged.Threshold == 0 {
-		merged.Threshold = defaultThreshold
+	fPort := def.GetIds().GetFPort()
+	assocFPort := assoc.GetIds().GetFPort()
+	if assocFPort != 0 {
+		fPort = assocFPort
 	}
-	if merged.FPort == 0 {
-		merged.FPort = def.Ids.FPort
-	}
-	return merged, nil
+	return merged, fPort, nil
 }
--- a/pkg/applicationserver/io/packages/alcsync/v1/data.go
+++ b/pkg/applicationserver/io/packages/alcsync/v1/data.go
@@ -1,0 +1,82 @@
+// Copyright © 2023 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package alcsyncv1
+
+import (
+	"time"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+var errInvalidFieldType = errors.DefineCorruption("invalid_field_type", "field `{field}` has the wrong type `{type}`")
+
+type packageData struct {
+	FPort     uint32
+	Threshold time.Duration
+}
+
+func (d *packageData) fromStruct(st *structpb.Struct) error {
+	fields := st.GetFields()
+	value, ok := fields["f_port"]
+	if ok {
+		numberValue, ok := value.GetKind().(*structpb.Value_NumberValue)
+		if !ok {
+			return errInvalidFieldType.WithAttributes(
+				"field", "f_port",
+				"type", "number",
+			)
+		}
+		d.FPort = uint32(numberValue.NumberValue)
+	}
+
+	value, ok = fields["threshold"]
+	if ok {
+		numberValue, ok := value.GetKind().(*structpb.Value_NumberValue)
+		if !ok {
+			return errInvalidFieldType.WithAttributes(
+				"field", "threshold",
+				"type", "number",
+			)
+		}
+		d.Threshold = time.Duration(numberValue.NumberValue) * time.Second
+	}
+	return nil
+}
+
+func mergePackageData(
+	def *ttnpb.ApplicationPackageDefaultAssociation,
+	assoc *ttnpb.ApplicationPackageAssociation,
+) (*packageData, error) {
+	var defaultData, associationData packageData
+	if err := defaultData.fromStruct(def.GetData()); err != nil {
+		return nil, err
+	}
+	if err := associationData.fromStruct(assoc.GetData()); err != nil {
+		return nil, err
+	}
+
+	merged := &packageData{}
+	for _, data := range []packageData{defaultData, associationData} {
+		if data.FPort != 0 {
+			merged.FPort = data.FPort
+		}
+		if data.Threshold != 0 {
+			merged.Threshold = data.Threshold
+		}
+	}
+	return merged, nil
+}
--- a/pkg/applicationserver/io/packages/alcsync/v1/data.go
+++ b/pkg/applicationserver/io/packages/alcsync/v1/data.go
@@ -24,6 +24,8 @@ import (
 
 var errInvalidFieldType = errors.DefineCorruption("invalid_field_type", "field `{field}` has the wrong type `{type}`")
 
+var defaultThreshold = time.Duration(4) * time.Second
+
 type packageData struct {
 	FPort     uint32
 	Threshold time.Duration
@@ -77,6 +79,12 @@ func mergePackageData(
 		if data.Threshold != 0 {
 			merged.Threshold = data.Threshold
 		}
+	}
+	if merged.Threshold == 0 {
+		merged.Threshold = defaultThreshold
+	}
+	if merged.FPort == 0 {
+		merged.FPort = def.Ids.FPort
 	}
 	return merged, nil
 }
--- a/pkg/applicationserver/io/packages/alcsync/v1/messages.go
+++ b/pkg/applicationserver/io/packages/alcsync/v1/messages.go
@@ -32,13 +32,17 @@ type AppTimeAns struct {
 	TokenAns       uint8
 }
 
+var _ Result = (*AppTimeAns)(nil)
+
 // MarshalBinary marshals the AppTimeAns into a byte slice.
 func (ans *AppTimeAns) MarshalBinary() ([]byte, error) {
-	// DeviceTime - bytes [0,3].
-	// Param - byte 4 (bits: RFU [7,4]; TokenAns [3,0]).
+	// CID - byte 0.
+	// DeviceTime - bytes [1,4].
+	// Param - byte 5 (bits: RFU [7,4]; TokenAns [3,0]).
 
-	cPayload := make([]byte, 5)
-	binary.LittleEndian.PutUint32(cPayload, uint32(ans.TimeCorrection))
-	cPayload[4] = ans.TokenAns & 0x0F
+	cPayload := make([]byte, 6)
+	cPayload[0] = 0x01
+	binary.LittleEndian.PutUint32(cPayload[1:5], uint32(ans.TimeCorrection))
+	cPayload[5] = ans.TokenAns & 0x0F
 	return cPayload, nil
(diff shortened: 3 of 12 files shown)


pkg/applicationserver/io/packages/alcsync/v1/commands.go
```diff
@@ -0,0 +1,51 @@
...
+	"time"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+)
+
+var errUnknownCommand = errors.DefineNotFound("unknown_command", "unknown command")
+
+// Converts GPS time to UTC time with second precision.
+func convertToUTCTime(gpsTime uint32) time.Time {
+	gpsEpoch := time.Date(1980, 1, 6, 0, 0, 0, 0, time.UTC)
```
adriansmares: Please note that this does not take the leap seconds into consideration and will output the wrong timestamps. Please use `pkg/gpstime`.

pkg/applicationserver/io/packages/alcsync/v1/commands.go
```diff
@@ -0,0 +1,51 @@
...
+func calculateTimeCorrection(deviceTime uint32, serverUTCTime time.Time) int32 {
+	deviceUTCTime := convertToUTCTime(deviceTime)
+	timeDifference := serverUTCTime.Sub(deviceUTCTime)
+	timeCorrection := int32(timeDifference.Seconds())
+	return timeCorrection
+}
+
+// TimeSynchronizationCommand implements the time synchronization command.
+func TimeSynchronizationCommand(_ context.Context, req *AppTimeReq) (*AppTimeAns, error) {
+	serverUTCTime := time.Now()
```
adriansmares: You probably should use the adjusted received at timestamp which we use in the Network Server / LoRaCloud package instead. Otherwise we're summing up lots of errors.

pkg/applicationserver/io/packages/alcsync/v1/messages.go
```diff
@@ -0,0 +1,66 @@
...
+// limitations under the License.
+
+package alcsyncv1
+
+import "encoding/binary"
+
+// AppTimeReq is the device request for time correction.
+type AppTimeReq struct {
+	// DeviceTime represents the device GPS time in seconds.
+	DeviceTime  uint32
```
adriansmares: Maybe you will have a bit of an easier time if you use `time.Time` here and marshal/unmarshal do the conversion to `uint32`.

cvetkovski98:
Really appreciate the remarks!

pkg/applicationserver/io/packages/alcsync/v1/data.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+var (
+	errFieldNotFound    = errors.DefineNotFound("field_not_found", "field `{field}` is not found")
+	errInvalidFieldType = errors.DefineCorruption("invalid_field_type", "field `{field}` has the wrong type `{type}`")
+)
+
+type packageData struct {
+	FPort            uint32
+	ThresholdSeconds uint32
```
adriansmares: Use a `time.Duration` please, and use `time.Duration.String()` / `time.ParseDuration` in the marshaling.

pkg/applicationserver/io/packages/alcsync/v1/factory.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"encoding/binary"
+	"time"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/gpstime"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	lorautil "go.thethings.network/lorawan-stack/v3/pkg/util/lora"
+)
+
+// CommandFactory is the factory for commands.
```
adriansmares: In general in Go we don't really use this 'factory' notation. I would remove `CommandFactory` and just leave the functions at the top level of the package.

pkg/applicationserver/io/packages/alcsync/v1/factory.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// newTimeSyncCommand builds a new TimeSyncCommand.
+func (CommandFactory) newTimeSyncCommand(
+	data []byte,
+	threshold uint32,
+	receivedAt time.Time,
+	fPort uint32,
+) (*TimeSyncCommand, []byte, error) {
+	// DeviceTime - bytes 0,1,2,3
```
adriansmares: Nit: dot at the end. Also you can use range notation (`[0, 3]`) to save a bit of space.

pkg/applicationserver/io/packages/alcsync/v1/factory.go
```diff
@@ -0,0 +1,104 @@
...
+	// Param - byte 4 (bits: RFU 7,6,5; AnsRequired 4; TokenReq 3,2,1,0)
+
+	lenBytes := 5
+	if len(data) < lenBytes {
+		return nil, data, errUnknownCommand
```
adriansmares: `errUnknownCommand.New()`.

pkg/applicationserver/io/packages/alcsync/v1/factory.go
```diff
@@ -0,0 +1,104 @@
...
+	cID, cPayload := frmPayload[0], frmPayload[1:]
+	commands := make([]Command, 0)
+	for {
+		cmd, rest, err := f.makeCommand(cID, cPayload, up, data)
+		if err != nil {
+			return commands, err
+		}
+		commands = append(commands, cmd)
+
+		if len(rest) == 0 { // no more commands
```
adriansmares: `No commands left.`

pkg/applicationserver/io/packages/alcsync/v1/factory.go
```diff
@@ -0,0 +1,104 @@
...
+	up *ttnpb.ApplicationUplink,
+	data *packageData,
+) (Command, []byte, error) {
+	receivedAt := lorautil.GetAdjustedReceivedAt(up)
+	fPort := data.FPort
+	switch cID {
+	case 0x01:
+		return f.newTimeSyncCommand(cPayload, data.ThresholdSeconds, receivedAt.AsTime(), fPort)
+	default:
+		return nil, cPayload, errUnknownCommand
```
adriansmares: `errUnknownCommand.New()`.

pkg/applicationserver/io/packages/alcsync/v1/messages.go
```diff
@@ -0,0 +1,44 @@
...
+
+// AppTimeAns is the answer to the device request for time correction.
+type AppTimeAns struct {
+	TimeCorrection int32
+	TokenAns       uint8
+}
+
+// Marshal marshals the AppTimeAns into a byte slice.
+func (ans *AppTimeAns) Marshal() []byte {
+	// DeviceTime - bytes 0,1,2,3
```
adriansmares: Same comment as for the request parser.

pkg/applicationserver/io/packages/alcsync/v1/messages.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// AppTimeAns is the answer to the device request for time correction.
+type AppTimeAns struct {
+	TimeCorrection int32
+	TokenAns       uint8
+}
+
+// Marshal marshals the AppTimeAns into a byte slice.
+func (ans *AppTimeAns) Marshal() []byte {
```
adriansmares: We use `MarshalBinary() ([]byte, error)` as a signature for this style of functions.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -0,0 +1,109 @@
...
+func (a *alcsyncpkg) HandleUp(
+	ctx context.Context,
+	def *ttnpb.ApplicationPackageDefaultAssociation,
+	assoc *ttnpb.ApplicationPackageAssociation,
+	up *ttnpb.ApplicationUp,
+) error {
+	ctx = log.NewContextWithField(ctx, "namespace", "applicationserver/io/packages/alcsync/v1")
+	logger := log.FromContext(ctx)
+
+	logger.Info("Handle uplink")
```
adriansmares: I would put this after the association check, as `Debug`.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -0,0 +1,109 @@
...
+
+	if def == nil && assoc == nil {
+		logger.Info("No association available")
```
adriansmares: This is an `Error` - `HandleUp` is not supposed to be called without one of these two.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -0,0 +1,109 @@
...
+		return errNoAssociation.New()
+	}
+
+	msg, ok := up.Up.(*ttnpb.ApplicationUp_UplinkMessage)
```
adriansmares: Use `GetUplinkMessage` instead, and compare with nil.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -0,0 +1,109 @@
...
+	if !ok {
+		logger.Debug("Uplink is not an uplink message; drop payload")
```
adriansmares: We don't really use `;` inside log messages. Use `.` as a sentence connector instead.

Also same below.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil
+	}
+
+	data, err := mergePackageData(def, assoc)
+	if err != nil {
+		logger.Debug("Failed to merge package data; drop payload")
+		return err
+	}
+
+	if up.GetUplinkMessage().GetFPort() != data.FPort {
```
adriansmares: You already have `msg`, so you can use that instead of `up.GetUplinkMessage()`.

pkg/applicationserver/io/packages/alcsync/v1/timesync.go
```diff
@@ -0,0 +1,59 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+var errUnknownCommand = errors.DefineNotFound("unknown_command", "unknown command")
+
+// TimeSyncCommand is the command for time synchronization.
+type TimeSyncCommand struct {
+	req              *AppTimeReq
+	receivedAt       time.Time
+	thresholdSeconds uint32
```
adriansmares: Please use `time.Duration` here.

pkg/applicationserver/io/packages/alcsync/v1/data.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (d *packageData) fromStruct(st *structpb.Struct) (err error) {
+	fields := st.GetFields()
+	value, ok := fields["f_port"]
+	if !ok {
+		return errFieldNotFound.WithAttributes("field", "f_port")
```
adriansmares: I don't think this should be a required field. You already have it in the associations, so it is a bit rough to require it.

pkg/applicationserver/io/packages/alcsync/v1/data.go
```diff
@@ -0,0 +1,85 @@
...
+	if !ok {
+		return errInvalidFieldType.WithAttributes(
+			"field", "f_port",
+			"type", "number",
+		)
+	}
+	d.FPort = uint32(numberValue.NumberValue)
+
+	value, ok = fields["threshold"]
+	if !ok {
```
adriansmares: I don't think this should be a required field. I think we should have a default here (maybe 4 seconds ? does the spec suggest a value ?).

pkg/applicationserver/io/packages/alcsync/v1/timesync_test.go
```diff
@@ -0,0 +1,219 @@
...
+			},
+		},
+	}
+
+	for _, _tC := range testCases {
+		tC := _tC
+		t.Run(tC.Name, func(t *testing.T) {
+			t.Parallel()
+			a, ctx := test.New(t)
+			// execute the command
```
adriansmares: Nit: Upper case first letter, dot at the end. Same for other comments.

pkg/applicationserver/io/packages/alcsync/v1/timesync_test.go
```diff
@@ -0,0 +1,219 @@
...
+				thresholdSeconds: 3,
+			},
+			Expected: AppTimeAns{
+				TimeCorrection: 10,
+				TokenAns:       1,
+			},
+		},
+	}
+
+	for _, _tC := range testCases {
```
adriansmares: `tc`. Same in other places.

pkg/applicationserver/io/packages/alcsync/v1/timesync_test.go
```diff
@@ -0,0 +1,219 @@
...
+		tC := _tC
```
adriansmares: `tc := tc` - in Go re-aliasing is allowed. Same in other places.

adriansmares (changes requested):
Logic wise this looks good to me.

Please pay more attention to the code style of your work. Things like comment format, logging message format and general variable naming should follow [DEVELOPMENT.md](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.25/DEVELOPMENT.md#code-style) and the general [Go styling guide](https://google.github.io/styleguide/go/decisions#variable-names).

pkg/applicationserver/io/packages/alcsync/v1/timesync.go
```diff
@@ -0,0 +1,59 @@
...
+	onDownlink func(context.Context, *ttnpb.ApplicationDownlink) error,
+) error {
+	difference := cmd.receivedAt.Sub(cmd.req.DeviceTime)
+	timeCorrectionSeconds := int32(difference.Seconds())
+	ans := &AppTimeAns{
+		TimeCorrection: timeCorrectionSeconds,
+		TokenAns:       cmd.req.TokenReq,
+	}
+	frmPayload := ans.Marshal()
+	if cmd.req.AnsRequired || ans.TimeCorrection >= int32(cmd.thresholdSeconds) {
```
adriansmares: Shouldn't we compare the absolute value of the time correction with the threshold ? Being in the past is an error too.

pkg/applicationserver/io/packages/alcsync/v1/factory.go
cvetkovski98: Should I leave the functions in the `factory.go` file or move them to the `alcsync.go` file altogether?

pkg/applicationserver/io/packages/alcsync/v1/data.go
cvetkovski98: There is no default value suggested by the spec - I'll use 4 as you suggest.

pkg/applicationserver/io/packages/alcsync/v1/factory.go
adriansmares: I am ok with moving it to `alcsync.go`.

pkg/applicationserver/io/packages/alcsync/v1/timesync.go
cvetkovski98: True.

pkg/applicationserver/io/packages/alcsync/v1/command.go
```diff
@@ -0,0 +1,33 @@
...
+// Command is the interface for commands.
+type Command interface {
+	// Code returns the command code.
+	Code() uint8
+
+	// Execute runs the command logic.
+	Execute(
+		ctx context.Context,
+		onDownlink func(context.Context, *ttnpb.ApplicationDownlink) error,
+	) error
```
adriansmares: I think it would be ok to just return a `[]*ttnpb.ApplicationDownlink` from `Execute` to be honest.

adriansmares (approved):
Looks great, thanks !

Can you create an issue to make this available in the Console, perhaps in the application settings ?

pkg/applicationserver/io/packages/alcsync/v1/command.go
cvetkovski98: Cleaner indeed.

pkg/applicationserver/io/packages/alcsync/v1/data.go
```diff
@@ -0,0 +1,90 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+var errInvalidFieldType = errors.DefineCorruption("invalid_field_type", "field `{field}` has the wrong type `{type}`")
+
+var defaultThreshold = time.Duration(4) * time.Second
+
+type packageData struct {
+	FPort     uint32
```
adriansmares: We shouldn't have the port both in the package data and in the (default) associations.

Let's remove it from the package data...

pkg/applicationserver/io/packages/alcsync/v1/data.go
```diff
@@ -0,0 +1,90 @@
...
+			merged.FPort = data.FPort
+		}
+		if data.Threshold != 0 {
+			merged.Threshold = data.Threshold
+		}
+	}
+	if merged.Threshold == 0 {
+		merged.Threshold = defaultThreshold
+	}
+	if merged.FPort == 0 {
```
adriansmares: ...and make sure that we go through `assoc` and `def` both here.

Otherwise have a grand total of 4 places in which we can set the FPort, which is too much.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -0,0 +1,111 @@
...
+	for _, cmd := range commands {
+		downlinks, err := cmd.Execute()
+		if err != nil {
+			logger.WithError(err).Debug("Failed to execute command")
+			continue
+		}
+		if len(downlinks) == 0 {
+			continue
+		}
+		if err := a.server.DownlinkQueuePush(ctx, up.EndDeviceIds, downlinks); err != nil {
```
adriansmares: You may benefit from accumulating them from each command and then doing a singular push after the loop. Less round trips, as `Push` needs to call the NS. You may also want to make it a single downlink containing multiple responses - otherwise the end device may re-send the requests while waiting for the second response.

pkg/applicationserver/io/packages/alcsync/v1/timesync.go
```diff
@@ -0,0 +1,57 @@
...
+func (*TimeSyncCommand) Code() uint8 {
+	return 0x01
+}
+
+// Execute implements commands.Command.
+func (cmd *TimeSyncCommand) Execute() (Result, error) {
+	difference := cmd.receivedAt.Sub(cmd.req.DeviceTime)
+	exceedsThreshold := math.Abs(difference.Seconds()) > cmd.threshold.Seconds()
+	if !cmd.req.AnsRequired && !exceedsThreshold {
+		return nil, errIgnoreDownlink
```
adriansmares: `.New()`.

pkg/applicationserver/io/packages/alcsync/v1/command.go
```diff
@@ -0,0 +1,29 @@
...
+
+// Result is the result of a command execution.
+type Result interface {
+	// MarshalBinary marshals the result into a binary representation.
+	MarshalBinary() ([]byte, error)
+}
+
+// Command is the interface for commands.
+type Command interface {
+	// Code returns the command code.
```
adriansmares, line 25: I don't see references to this - do we need it ?

pkg/applicationserver/io/packages/alcsync/v1/timesync.go
```diff
@@ -0,0 +1,57 @@
...
+	"math"
+	"time"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/errors"
+)
+
+var (
+	errUnknownCommand = errors.DefineNotFound("unknown_command", "unknown command")
+
+	// ErrIgnoreDownlink is a sentinel error returned when the command result should be ignored.
```
adriansmares, line 27: I would use `DefineUnavailable` instead, as the downlink is unavailable.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -0,0 +1,121 @@
...
+	commands, err := MakeCommands(msg, fPort, data)
+	if err != nil {
+		logger.WithError(err).Debug("Failed to parse frame payload into commands")
+		return err
+	}
+
+	results := make([]Result, 0, len(commands))
+	for _, cmd := range commands {
+		result, err := cmd.Execute()
+		if errors.Is(err, errIgnoreDownlink) {
```
adriansmares: I would use `errors.IsUnavailable` instead.

pkg/applicationserver/io/packages/alcsync/v1/command.go
cvetkovski98: It's used in pkg/applicationserver/io/packages/alcsync/v1/alcsync_test.go#L215 to assert that the correct Command is constructed. I can change this to a type assertion. Would that be preferable?

adriansmares: No, you're right. It's fine to keep this.

adriansmares (approved):

PR #5814, opened 2022-09-20
Fallback values for loading device profile from DR

ryaplots:
#### Summary

This PR completes front-end part of #5512 

#### Changes

- Refactor import form from class component to functional
- Add input method
- Add `versions_ids` selection section
- Add version ids as fallback

https://user-images.githubusercontent.com/56658938/191236645-cb91fafa-8c6a-437d-b6ed-eb09b6b0a044.mov

#### Testing

cypress

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #6114, opened 2023-03-14
Fallback values for loading device profile from DR

### LINKED ISSUE #5512: Load device profile from DR when using CSV import

ryaplots:
#### Summary

This PR re-opens #5814 
Closes #5512 

#### Changes

- Refactor import form from class component to functional
- Add input method
- Add versions_ids selection section
- Add version ids as fallback

#### Testing

cypress

#### Checklist


.github/workflows/e2e.yml
```diff
@@ -27,6 +27,7 @@ env:
   CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
   CYPRESS_FAIL_FAST: true
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  ACTIONS_RUNNER_DEBUG: true
```
adriansmares: This is no longer needed, right ?

CHANGELOG.md
```diff
@@ -176,6 +176,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Added
 
 - Add more specific rights for OAuth clients.
+- Fallback `version_ids` to load end device profile from DR when importing a device in the Console.
```
adriansmares: I would phrase this as `Fallback end device version identifiers to be used while importing end devices using the Console`.

pkg/webui/components/form/index.js
```diff
@@ -22,7 +22,7 @@ import {
 } from 'formik'
 import scrollIntoView from 'scroll-into-view-if-needed'
 import { defineMessages } from 'react-intl'
-import { isPlainObject, isFunction, pick, omitBy, pull, merge } from 'lodash'
+import { isPlainObject, isFunction, pick, omitBy, pull, merge, isEmpty } from 'lodash'
```
adriansmares: I think linting complains that this was added.

.github/workflows/e2e.yml
ryaplots: Indeed. Removing

pkg/webui/components/form/index.js
ryaplots: Indeed

pkg/webui/console/components/device-import-form/fallback-version-ids-section.js
```diff
@@ -0,0 +1,57 @@
...
+  const showOtherHint = hasSelectedOther
+
+  return (
+    <Row>
+      <Col>
+        <RequireRequest
+          requestAction={listBrands(appId, {}, ['name', 'lora_alliance_vendor_id'])}
+          errorRenderFunction={ErrorNotification}
+          spinnerProps={{ center: false, inline: true }}
+        >
```
kschiffer: I would prefer this to be loaded only once initially in the device import view. Otherwise, this results in flashes when switching the options, also it will cause the brands to be reloaded every time the device repository option is selected.
So that would be in `pkg/webui/console/views/device-import/index.js`, let's also refactor this component to a functional component and use `<RequireRequest />`.

pkg/webui/console/containers/device-profile-section/index.js
```diff
@@ -0,0 +1,138 @@
...
+  const hardwareVersion = version_ids?.hardware_version
+  const firmwareVersion = version_ids?.firmware_version
+
+  const hasBrand = Boolean(brand) && !isOtherOption(brand)
+  const hasModel = Boolean(model) && !isOtherOption(model)
+  const hasHwVersion = Boolean(hardwareVersion) && !isOtherOption(hardwareVersion)
+  const hasFwVersion = Boolean(firmwareVersion) && !isOtherOption(firmwareVersion)
+
+  return (
+    <>
```
kschiffer: You can remove the fragment.

kschiffer (changes requested):
Looks good. Only have one small comment on the fetching logic.

adriansmares (commented):
My comments are addressed, but I am not knowledgeable enough to comment on the `pkg/webui` changes.

pkg/webui/console/views/device-import/index.js
```diff
@@ -12,53 +12,55 @@
...
-    return (
+  const deviceTemplateFormats = useSelector(selectDeviceTemplateFormats)
+  const deviceTemplateFormatsFetching = useSelector(selectDeviceTemplateFormatsFetching)
+  const showEmptyWarning =
+    !deviceTemplateFormatsFetching && Object.keys(deviceTemplateFormats).length === 0
+
+  return (
+    <RequireRequest
+      requestAction={listBrands(appId, {}, ['name', 'lora_alliance_vendor_id'])}
+      errorRenderFunction={ErrorNotification}
```
kschiffer:
```suggestion
      errorRenderFunction={SubViewError}
```
I think this is what should be rendered in case of an error.

pkg/webui/lib/components/require-request.js
```diff
@@ -31,19 +31,20 @@ const RequireRequest = ({
...
-      <Spinner center>
+      <Spinner {...spinnerProps}>
         <Message content={sharedMessages.fetching} />
       </Spinner>
     )
   }
 
   if (error && handleErrors) {
-    return <ErrorRenderFunction error={error} />
+    return <ErrorRenderFunction error={error} content={error} />
```
kschiffer: I think you can remove that.

mjamescompton:
@ryaplots maybe I am late to the party as I know we had a lot of issues around e2e testing but should we have an e2e test for specifically fallback values from device repository instead only `cy.findByLabelText('Enter LoRaWAN versions and frequency plan manually')`

ryaplots:
> @ryaplots maybe I am late to the party as I know we had a lot of issues around e2e testing but should we have an e2e test for specifically fallback values from device repository instead only `cy.findByLabelText('Enter LoRaWAN versions and frequency plan manually')`

Yeah I think it does make sense. I will add it.

mjamescompton (approved):

kschiffer (approved):

PR #6119, opened 2023-03-21
ALCSync improvements

### LINKED ISSUE #6105: LoRa ALCS Improvements
### Summary

Follow up #6090 with some improvements:
- [ ] Add observability
  - [ ] Use `WithCause(err)` when propagating package errors
  - [ ] Use `evtPackageFail` event to signal package failures
  - [ ] Use events to give insight into the data after and before serialization (verbose events)
- [ ] Update CHANGELOG.md
- [ ] Fix typos

### Current Situation

We have no error trace when an error occurs in the `alcsync-v1` package and it is difficult to inspect the communication between The Stack and the end devices.

It would help in debugging when something goes wrong.

### Proposed Implementation

_No response_

### Contributing

### Code of Conduct

cvetkovski98:
#### Summary

Follow up improvements for the time correction command that's part of the Application Layer Clock Synchronization package.

Closes #6090.

#### Changes

- Properly wrap errors to build up a trace
- Improve error and debug logging
- Add observability events after command ingestion and processing 
- Declare commands through proto messages

#### Testing
- Manual and 
- Existing UT

#### Notes for Reviewers

...

#### Checklist


api/applicationserver_integrations_alcsync.proto
```diff
@@ -0,0 +1,52 @@
...
+import "google/protobuf/timestamp.proto";
+
+package ttn.lorawan.v3;
+
+option go_package = "go.thethings.network/lorawan-stack/v3/pkg/ttnpb";
+
+enum ALCSyncCommandIdentifier {
+  option (thethings.json.enum) = { marshal_as_string: true, prefix: "CID" };
+
+  CID_PKG_VERSION = 0;
```
adriansmares: So a pitty with the `enum` system in `protobuf` is that is subject to restrictions that appear in other languages, like C++.

In C++, enums have to be unique both within a single enum definition (that's quite reasonable), _and_ within a namespace (less obvious).

The namespace in this case is `ttn.lorawan.v3`, and we already use the `CID` notation for `MACCommandIdentifier`. If at some point we would have a MAC command with the same identifier, they would clash while generating C++ code.

The solution is to prefix the commands here (since we cannot break the `MACCommandIdentifier` API) with a prefix like `ALCSYNC`. I know it is verbose and doubles the naming, but it is what it is.

pkg/applicationserver/io/packages/alcsync/v1/alcsync.go
```diff
@@ -60,27 +65,58 @@ func newTimeSyncCommand(
 }
 
 // MakeCommands parses the uplink payload and returns the commands.
-func MakeCommands(up *ttnpb.ApplicationUplink, fPort uint32, data *packageData) ([]Command, error) {
-	cID, cPayload := up.FrmPayload[0], up.FrmPayload[1:]
+func MakeCommands(up *ttnpb.ApplicationUplink, fPort uint32, data *packageData) ([]Command, []events.Builder, error) {
+	cIDByte, cPayload := up.FrmPayload[0], up.FrmPayload[1:]
```
adriansmares: You can also cast here directly, and same below, if you want to avoid having two notations (`cIDByte` and `cID`).

pkg/applicationserver/io/packages/alcsync/v1/errors.go
```diff
@@ -0,0 +1,35 @@
...
+	// ErrIgnoreDownlink is a sentinel error returned when the command result should be ignored.
+	errIgnoreDownlink = errors.DefineUnavailable("downlink_unavailable", "downlink unavailable")
+
+	errCommandCreationFailed  = errors.Define("command_creation_failed", "failed to create command")
+	errDownlinkCreationFailed = errors.Define("downlink_creation_failed", "failed to create downlink")
+
+	errInvalidFieldType = errors.DefineCorruption("invalid_field_type", "field `{field}` has the wrong type `{type}`")
+	errPkgDataMerge     = errors.DefineCorruption("pkg_data_merge", "failed to merge package data")
+
+	errTimeSyncCommandPayloadInsufficientLength = errors.DefineInvalidArgument(
```
adriansmares: Errors in `pkg/errors` are namespaced - for example this is an error in the `pkg/applicationserver/io/packages/alcsync/alcsyncv1`. As such, I would drop the `TimeSync`/`time_sync` prefix.

pkg/applicationserver/io/packages/alcsync/v1/observability.go
```diff
@@ -0,0 +1,119 @@
...
+	"context"
+	"fmt"
+
+	"go.thethings.network/lorawan-stack/v3/pkg/events"
+	"go.thethings.network/lorawan-stack/v3/pkg/log"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
+func publishEvents(ctx context.Context, builders ...events.Builder) {
+	n := len(builders)
```
adriansmares, line 27: You could use `events.Builders` - It already has a `New` function that does exactly what you are doing here.

pkg/applicationserver/io/packages/alcsync/v1/observability.go
```diff
@@ -0,0 +1,119 @@
...
+
+	evts := make([]events.Event, n)
+	for i, builder := range builders {
+		evts[i] = builder.New(ctx)
+	}
+	log.FromContext(ctx).WithField("event_count", n).Debug("Publish events")
+	events.Publish(evts...)
+}
+
+func alcsyncEventOptions(extraOpts ...events.Option) []events.Option {
```
adriansmares: Same comment as before - since we're already in the ALCSync package, we can drop it as a prefix from both local and global functions in general.

Same for the helpers below.

pkg/applicationserver/io/packages/alcsync/v1/observability.go
```diff
@@ -0,0 +1,119 @@
...
+	return append([]events.Option{events.WithVisibility(ttnpb.Right_RIGHT_APPLICATION_TRAFFIC_READ)}, extraOpts...)
+}
+
+func defineALCSyncPkgFailEvent(name, desc string, opts ...events.Option) func() events.Builder {
+	return events.DefineFunc(
+		fmt.Sprintf("as.packages.alcsync.%s.failed", name), fmt.Sprintf("%s package failed", desc),
```
adriansmares: I would add the `v1` in the path, so it would be `as.packages.alcsync.v1.%s.failed`.

I would also say that `PropagateToParent` and `WithErrorDataType` should be added here by default. The general standard is that failure events _should_ always have these two options.

api/applicationserver_integrations_alcsync.proto
```diff
@@ -0,0 +1,52 @@
...
+  CID_APP_TIME = 1;
```
adriansmares: Is this discrepancy (`APP_TIME` vs `APP_DEV_TIME`) also present in the standard ?

pkg/applicationserver/io/packages/alcsync/v1/observability.go
```diff
@@ -0,0 +1,119 @@
...
+		alcsyncEventOptions(opts...)...,
+	)
+}
+
+func defineALCSyncCmdReceivedEvent(name, desc string, opts ...events.Option) func() events.Builder {
```
adriansmares: I would say that the spirit here is that there is one received event per request command type. So in this case you would have an `receive` for `app_time_req`.

pkg/applicationserver/io/packages/alcsync/v1/observability.go
```diff
@@ -0,0 +1,119 @@
...
+	return events.DefineFunc(
+		fmt.Sprintf("as.packages.alcsync.cmd.%s.accept", name), fmt.Sprintf("%s command received", desc),
+		alcsyncEventOptions(opts...)...,
+	)
+}
+
+func defineALCSyncCmdParsedEvent(name, desc string, opts ...events.Option) func() events.Builder {
```
adriansmares: To be honest I wouldn't go that deep as to generate events for when we _parsed_ a command - just having a `receive` (which means received + parsed) for that particular command suffices in my opinion, and it is consistent with the Network Server events.

pkg/applicationserver/io/packages/alcsync/v1/observability.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func defineALCSyncCmdParsedFailEvent(name, desc string, opts ...events.Option) func() events.Builder {
+	return events.DefineFunc(
+		fmt.Sprintf("as.packages.alcsync.cmd.%s.parse.failed", name), fmt.Sprintf("%s command failed when parsing", desc),
+		alcsyncEventOptions(opts...)...,
+	)
+}
+
+func defineALCSyncCmdHandledEvent(name, desc string, opts ...events.Option) func() events.Builder {
```
adriansmares: In the same spirit as `app_time_req.receive` I think we should have `app_time_ans.queued` for the handling in such cases.

pkg/applicationserver/io/packages/alcsync/v1/observability.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func defineALCSyncCmdParsedEvent(name, desc string, opts ...events.Option) func() events.Builder {
+	return events.DefineFunc(
+		fmt.Sprintf("as.packages.alcsync.cmd.%s.parsed", name), fmt.Sprintf("%s command parsed", desc),
+		alcsyncEventOptions(opts...)...,
+	)
+}
+
+func defineALCSyncCmdParsedFailEvent(name, desc string, opts ...events.Option) func() events.Builder {
```
adriansmares: The thing about parsing bytes is that it is actually hard to check if the command that failed was even _real_. 

Think of it like this - due to some mismatch in standards / misbehavior a command is now extended from one byte of payload to two bytes of payload. You parse the first byte, and you consumption function now uses the second byte of payload as a command identifier - if you are lucky, it's an invalid identifier, but if you are not, it will just look like this fictitious command has an invalid payload, while it was not there in the first place.

As such, I would recommend that we just have a generic parse failed event, not one per request type.

pkg/applicationserver/io/packages/alcsync/v1/observability.go
```diff
@@ -0,0 +1,119 @@
...
+func defineALCSyncCmdHandledEvent(name, desc string, opts ...events.Option) func() events.Builder {
+	return events.DefineFunc(
+		fmt.Sprintf("as.packages.alcsync.cmd.%s.handled", name), fmt.Sprintf("%s command handled", desc),
+		alcsyncEventOptions(opts...)...,
+	)
+}
+
+func defineALCSyncCmdHandledFailEvent(name, desc string, opts ...events.Option) func() events.Builder {
+	return events.DefineFunc(
+		fmt.Sprintf("as.packages.alcsync.cmd.%s.handled.failed", name), fmt.Sprintf("%s command handled", desc),
```
adriansmares: Just `app_time_req.failed` and `app_time_ans.failed` is fine for me. Also the description is wrong.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -59,12 +58,15 @@ func (a *alcsyncpkg) HandleUp(
 
 	data, fPort, err := mergePackageData(def, assoc)
 	if err != nil {
-		logger.Error("Failed to merge package data")
+		logger.WithError(err).Error("Failed to merge package data")
```
adriansmares: I wouldn't log this as `Error` to be honest - I think this should stay as debug.

The reason for this is that the package data is just a blob of JSON that comes from the user, and there is no validation on what they input here, so it is not a _server_ error if the contents are blaberred - what matters is that an event is generated.

For that matter, please take a look at how the `defer`ed error is used in other packages - you introduce a name return called `err` and inside the defer you generate the failed processing event.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -86,22 +101,35 @@ func (a *alcsyncpkg) HandleUp(
...
+			events.WithData(err),
+		))
+		return a.publishAndExit(ctx, eventBuilders, err)
 	}
 	if err := a.server.DownlinkQueuePush(ctx, up.EndDeviceIds, []*ttnpb.ApplicationDownlink{downlink}); err != nil {
-		logger.WithError(err).Debug("Failed to push downlinks to queue")
+		logger.WithError(err).Error("Failed to push downlinks to queue")
 	}
-	return nil
+	return a.publishAndExit(ctx, eventBuilders, nil)
```
adriansmares: You should use a `defer` instead. Otherwise you're not publishing on panics or other nasty cases.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -86,22 +101,35 @@ func (a *alcsyncpkg) HandleUp(
 			continue
 		}
 		if err != nil {
-			logger.WithError(err).Debug("Failed to execute command")
+			logger.WithError(err).WithField("command_id", cmd.Code()).Error("Failed to execute command")
```
adriansmares: I don't think this should be an `Error`, for the same reason - it's not a server error.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -86,22 +101,35 @@ func (a *alcsyncpkg) HandleUp(
...
 		if result != nil {
 			results = append(results, result)
+			eventBuilders = append(eventBuilders, result.GetEvtSuccessfullyExecuted())
 		}
 	}
 	downlink, err := MakeDownlink(results, fPort)
 	if err != nil {
-		logger.WithError(err).Debug("Failed to create downlink from results")
-		return err
+		logger.WithError(err).Error("Failed to create downlink from results")
```
adriansmares: Same comment.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -86,22 +101,35 @@ func (a *alcsyncpkg) HandleUp(
...
+		eventBuilders = append(eventBuilders, EvtALCSyncPkgFail.With(
+			events.WithIdentifiers(up.GetEndDeviceIds()),
+			events.WithData(err),
+		))
+		return a.publishAndExit(ctx, eventBuilders, err)
 	}
 	if err := a.server.DownlinkQueuePush(ctx, up.EndDeviceIds, []*ttnpb.ApplicationDownlink{downlink}); err != nil {
-		logger.WithError(err).Debug("Failed to push downlinks to queue")
+		logger.WithError(err).Error("Failed to push downlinks to queue")
```
adriansmares: Same comment regarding `Error`.

adriansmares (changes requested):

pkg/applicationserver/io/packages/alcsync/v1/alcsync.go
```diff
@@ -60,27 +65,36 @@ func newTimeSyncCommand(
...
-	cID, cPayload := up.FrmPayload[0], up.FrmPayload[1:]
+func MakeCommands(up *ttnpb.ApplicationUplink, fPort uint32, data *packageData) ([]Command, []events.Builder, error) {
+	cID, cPayload := ttnpb.ALCSyncCommandIdentifier(up.FrmPayload[0]), up.FrmPayload[1:]
 	commands := make([]Command, 0)
+	evts := make([]events.Builder, 0)
 	for {
 		cmd, rest, err := makeCommand(cID, cPayload, up, fPort, data)
 		if err != nil {
-			return commands, err
+			err := errCommandCreationFailed.WithCause(err).WithAttributes(
```
adriansmares, line 75: These attributes won't be visible anywhere as they are not part of the error message or public attributes.

If you would like to make them part of the error, you need to add them as part of the error definition. In effect, `errors.Define("name", "message", "public_attribute_1", "public_attribute_2")` and so forth.

Very nit: I don't think `up.ReceivedAt` is relevant to this error.

pkg/applicationserver/io/packages/alcsync/v1/command.go
```diff
@@ -14,16 +14,28 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+)
+
 // Result is the result of a command execution.
 type Result interface {
 	// MarshalBinary marshals the result into a binary representation.
 	MarshalBinary() ([]byte, error)
+
+	// GetAnsEnqueuedEvt returns the event that should be emitted when the answer is enqueued or executed.
+	GetAnsEnqueuedEvt() events.Builder
```
adriansmares: Nit: `AnswerEnqueuedEventBuilder`.

pkg/applicationserver/io/packages/alcsync/v1/command.go
```diff
@@ -14,16 +14,28 @@
...
 }
 
 // Command is the interface for commands.
 type Command interface {
 	// Code returns the command code.
-	Code() uint8
+	Code() ttnpb.ALCSyncCommandIdentifier
+
+	// GetCmdReceivedEvt returns the event that should be emitted when the command is successfully parsed.
```
adriansmares: Nit: `CommandReceivedEventBuilder`.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Added
 
+- Support for end-device clock correction through `AppTimeReq` and `AppTimeAns` as described in the LoRaWAN Application Layer Clock Synchronization Specification - TS003-2.0.0.
```
adriansmares: I've added a changelog entry as part of the release in 84eda6ec79d2e3ada371b0864c81d4bdc89221ba . If you think that one suffices, let's drop this one from here.

pkg/applicationserver/io/packages/alcsync/v1/observability.go
```diff
@@ -0,0 +1,78 @@
...
+	events.Publish(evts...)
+}
+
+func eventOptions(extraOpts ...events.Option) []events.Option {
+	return append([]events.Option{events.WithVisibility(ttnpb.Right_RIGHT_APPLICATION_TRAFFIC_READ)}, extraOpts...)
+}
+
+func defineCmdReceivedEvent(name, desc string, opts ...events.Option) func() events.Builder {
+	return events.DefineFunc(
+		fmt.Sprintf("as.packages.alcsync.v1.%s.cmd_received", name),
```
adriansmares, line 43: Nit: `and parsed successfully` is not needed since an error event would've been generated otherwise.

pkg/applicationserver/io/packages/alcsync/v1/observability.go
```diff
@@ -0,0 +1,78 @@
...
+		fmt.Sprintf("%s command received and parsed successfully", desc),
+		eventOptions(opts...)...,
+	)
+}
+
+func defineCmdAnsEnqueueEvent(name, desc string, opts ...events.Option) func() events.Builder {
+	return events.DefineFunc(
+		fmt.Sprintf("as.packages.alcsync.v1.%s.answer_enqueued", name),
```
adriansmares, line 51: Nit: same comment for ` executed succesfully and `.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -73,9 +75,27 @@ func (a *alcsyncpkg) HandleUp(
 		return nil
 	}
 
-	commands, err := MakeCommands(msg, fPort, data)
+	eventBuilders := make(events.Builders, 0)
```
adriansmares: I think you should defer this before merging the application data, such that errors coming from that generate an event too.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -73,9 +75,27 @@ func (a *alcsyncpkg) HandleUp(
...
+			eventBuilders = append(eventBuilders, EvtPkgFail.With(
+				events.WithIdentifiers(ids),
+				events.WithData(err),
+			))
+		}
+		publishEvents(ctx, eventBuilders...)
+	}(up.GetEndDeviceIds())
+
+	commands, evts, err := MakeCommands(msg, fPort, data)
 	if err != nil {
```
adriansmares, line 92: I don't think execution should continue here.

pkg/applicationserver/io/packages/alcsync/v1/package.go
```diff
@@ -73,9 +75,27 @@ func (a *alcsyncpkg) HandleUp(
...
 		logger.WithError(err).Debug("Failed to parse frame payload into commands")
+	}
+
+	eventBuilders = append(eventBuilders, evts...)
+
+	if err != nil && len(commands) == 0 {
```
adriansmares: I don't think this is an error, and if you return early you don't need to check `err`.

adriansmares (approved):

PR #5325, opened 2022-03-21
Validate uplink message payload

adriansmares:
#### Summary

This short PR adds validation to the uplink message payloads and improves logging of clock synchronization.

#### Changes

- Ensure that clock synchronization messages contain `time.Time` timestamps
- Validate messages after unmarshaling

#### Testing

Unit tests.

##### Regressions

N/A

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #6126, opened 2023-03-24
Drop uplink frame CRC failures

### LINKED ISSUE #6123: Discard CRC failures
### Summary

Discard CRC failures

### Current Situation

The default gateway packet forwarder configuration has forwarding frames with a CRC failure disabled. However, sometimes packet forwarders are sending frames with a CRC failure.

Processing frames that we know are wrong is useless; it's unnecessary compute and bandwidth and it screws up statistics.

Seeing frames with a CRC failure can be useful though, in the gateway traffic view, for debugging purposes.

### Proposed Implementation

Drop the frame with a CRC failure in Gateway Server. This should result in a `gs.up.drop` event with proper reason.

This probably requires an API extension as I think we don't track CRC failure in RX metadata.

The end result should be that a `gs.up.receive` is emitted, there's also a `gs.up.drop`, and the frame should not be forwarded to upstream handlers (NS and PBA).

### Contributing

### Code of Conduct

johanstokking:
#### Summary

Closes #6123 

#### Changes

- Add a field `crc_status` to `UplinkMessage`, which is propagated by the UDP decoding. Implicitly, there is MQTT and GRPC support as well as they use protobuf marshaling
- Reorder events; receive is now always before the result event
- Check that if CRC failed, do emit the receive event but drop immediately
- Remove bogus `validation` host for validation errors

#### Testing

Unit testing

##### Regressions

None expected

@adriansmares please verify:
- Did we need a non-empty `host` for some reason? Why was that `validation`; just for signaling? The prometheus label values can be empty, or is that the reason?
- We are now publishing `gs.up.receive` _before_ marshaling happened; event consumers should not expect unmarshaled payload to be in the event payload. I think we have checks, but this is a potential regression

#### Checklist


pkg/gatewayserver/gatewayserver.go
```diff
@@ -894,19 +896,23 @@ func (gs *GatewayServer) handleUpstream(ctx context.Context, conn connectionEntr
 			correlationIDs = append(correlationIDs, fmt.Sprintf("gs:uplink:%s", events.NewCorrelationID()))
 			ctx = events.ContextWithCorrelationID(ctx, correlationIDs...)
 			msg.Message.CorrelationIds = events.CorrelationIDsFromContext(ctx)
+			registerReceiveUplink(ctx, gtw, msg, protocol)
```
adriansmares: The subscribers (NOC and Console) really do depend on the payload being there in the event.

The whole point of the `UnmarshalMessage` done below is to allow the payload to be visible. If we emit the `gs.up.receive` before that, there is no event that actually shows the payload - `receive` has only the binary payload, `drop` has the error and `forward` has the host.

I think that it is better to have a useful `receive` than to move it such that you see what you've dropped. Since the CRC failed you're seeing garbage anyway.

adriansmares (commented):
I've added the non-empty host in https://github.com/TheThingsNetwork/lorawan-stack/pull/5325 , but after running some checks now, having an empty label should be valid, and our dashboards do not break the errors per target host, so things should be fine.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -897,16 +899,19 @@ func (gs *GatewayServer) handleUpstream(ctx context.Context, conn connectionEntr
 			if msg.Message.Payload == nil {
 				msg.Message.Payload = &ttnpb.Message{}
 				if err := lorawan.UnmarshalMessage(msg.Message.RawPayload, msg.Message.Payload); err != nil {
-					registerDropUplink(ctx, gtw, msg, "validation", err)
```
adriansmares, line 900: Any reason to drop this register ?

pkg/gatewayserver/gatewayserver.go
```diff
@@ -897,16 +899,19 @@ func (gs *GatewayServer) handleUpstream(ctx context.Context, conn connectionEntr
...
 					continue
 				}
 			}
+			registerReceiveUplink(ctx, gtw, msg, protocol)
```
adriansmares, line 905: I would still register the receive only at the end, after everything has been dropped.

pkg/gatewayserver/gatewayserver.go
```diff
@@ -921,13 +926,13 @@ func (gs *GatewayServer) handleUpstream(ctx context.Context, conn connectionEntr
 			if d := msg.DownlinkMessage; d != nil {
 				d.CorrelationIds = events.CorrelationIDsFromContext(ctx)
 			}
+			registerReceiveTxAck(ctx, gtw, msg, protocol)
```
adriansmares, line 929: Same comment.

adriansmares:
Also add the changelog item to `v3.25.0` - we'll backport and redeploy this.

pkg/gatewayserver/gatewayserver.go
johanstokking: Because it's before the receive.

Traditionally we had first receive, then drop or forward in GS, NS and AS. In GS, that drop or forward is per upstream. This way we wouldn't miss traffic in our metrics.

When reusing this `gs.up.drop` for validation _before_ a receive, the order is off, `count(upstreams) * receives != drops + forwards`, and we get a weird label value (`validation`) for host. I get that it's nice to have something here, but it's not clean. That is the reason this one gets removed here.

---

I think we need to decide what we want to achieve. I do value some sort of orderly events and also consistency. What we could do is:

1. Receive: everything, raw payload only. Could be non-LoRaWAN, could be CRC failure, could be unsupported LoRaWAN major version, etc
2. Parse and validation errors; new event. Includes CRC failure
3. Accept; new event
4. Forward and drop: per upstream only

Or we keep it simple:

1. Receive: only something that makes sense, i.e. something that passes LoRaWAN parsing. That is only relevant anyway
2. Drop if CRC or validation fails
3. Forward and drop: per upstream

This is what is implemented now. This is not entirely clean, but we wouldn't have to refactor the events and we stay close to what we have in NS and AS.

adriansmares: I am ok with keeping option 2 (what is implemented), and doing option 1 for the next minor. 

In general I am not sure if `receive` and fork to `drop` or `forward` is really consistent, at least body wise, between components, but I am ok with revisiting this.

adriansmares (approved):

KrishnaIyer (approved):
Nothing more to add.

PR #6127, opened 2023-03-27
Deprecate end devices claim command

KrishnaIyer:
#### Summary

Deprecate end devices claim command. Refs; https://github.com/TheThingsIndustries/lorawan-stack-support/issues/936#issuecomment-1484784745

#### Changes

- Deprecate `ttn-lw-cli devices claim`
- Add note to Changelog.

#### Testing

Locally

##### Regressions

NA. This feature was broken anyway.

#### Checklist


adriansmares (approved):
LGTM.

For the reference, it wasn't _that_ broken. I used it to claim my Generic Nodes on `staging1` when we moved from the old JS to the new one, during testing. It does work if want to claim something which is already registered (but this is an edge case).

Issue #6128, opened 2023-03-27
Deprecate NaN and (-)Infinity usage in payload decoders


adriansmares:
### Summary

Payload decoders may emit `NaN` and `-Infinity` or `Infinity` number values, as these are supported values in JavaScript. 

The most commonly used serialization format is JSON, and JSON cannot represent these values without creating a confusion between `NaN`, the float concept, and `NaN`, the string. 

We should consider whether or not allowing these values as part of the output is valuable or not.

### Current Situation

Historically we've rendered these values as `"NaN"` and `"Infinity"`, but our newer `protojson` library [does not](https://github.com/protocolbuffers/protobuf-go/commit/0fd4f3a506894c05a9cad2d52c521a8ce953b49d) like to render these values as strings anymore, as they break the round-trip concept (serialize -> deserialize should be the identity function).

The current workaround is to round-trip the values via `structpb.NewStruct` which [still](https://github.com/protocolbuffers/protobuf-go/blob/808c66411fe76c839a68168654228446fbdc1ecf/types/known/structpb/struct.pb.go#L401-L413) allows the special values to be rendered as strings. This is not ideal as it changes the protobuf serialization result, but it does maintain the JSON serialization.

Regarding the usage today:
1. `Infinity` is never observed as part of our stored messages.
2. `-Infinity` is virtually never observed (10^-9) as part of our stored messages.
3. `NaN` is observed in under 0.001% of cases as part of our stored messages.

This methodology is biased towards the storage integrations, but historically we haven't seen any `NaN` or `Infinity` usage in the wild which has been intentional. 

`NaN`, which is the most observed (but still barely observed overall) is general the result of a mathematical operation which interacts with a list and goes out of bounds. Consider the following snippet, which can occur while consuming bytes from an end device too eagerly (i.e. consuming what is not there).
```js
a = [1, 2, 3] 
console.log(a[4] * 5) // This outputs NaN.
```

### Desired Situation

We should remove support for these values. We have at least two options here:
1. Sanitize recursively the result object.
- PRO: if only some parts of the decoded payload are wrong, we still output the remaining values.
- CON: the remaining part of the message may still be garbage, since the logic may depend on evaluating such degenerate NaNs. It is also the case that we're no longer rendering fields which historically were there.
2. Drop the whole payload.
- PRO: downstream consumers do not see potentially invalid decoded payloads.
- CON: it could be the case that some parts of the decoded message are still useful.

[...]

PR #6129, opened 2023-03-27
Round trip decoded payloads

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/6128

#### Changes

- Round trip decoded payloads through `structpb.NewStruct` in order to convert special values (`NaN`, `Infinity`) to strings.
- Add a deprecation warning for the handling of such values by future values of The Things Stack.

#### Testing

CI and local.

##### Regressions

This change changes the binary format for the decoded payload. What this means is that previously `NaN` was encoded as a `Value_NumberValue`. Due to the round trip, it will now be rendered as a `Value_StringValue` instead.

The JSON format is unaffected, and the changes allow such values to be rendered.

#### Checklist

### DIFF
--- a/pkg/messageprocessors/javascript/javascript.go
+++ b/pkg/messageprocessors/javascript/javascript.go
@@ -379,6 +379,11 @@ func (*host) decodeUplink(
 			msg.NormalizedPayloadWarnings = appendValidationErrors(msg.NormalizedPayloadWarnings, normalizedMeasurements)
 		}
 	}
+
+	// Roundtrip the message in order to convert special number values such as NaN and Infinity to their string form.
+	// TODO: Clean up, or fail, the message (https://github.com/TheThingsNetwork/lorawan-stack/issues/6128).
+	msg.DecodedPayload, _ = structpb.NewStruct(msg.DecodedPayload.AsMap())
+
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


pkg/messageprocessors/javascript/javascript.go
```diff
@@ -379,6 +379,11 @@ func (*host) decodeUplink(
 			msg.NormalizedPayloadWarnings = appendValidationErrors(msg.NormalizedPayloadWarnings, normalizedMeasurements)
 		}
 	}
+
+	// Roundtrip the message in order to convert special number values such as NaN and Infinity to their string form.
+	// TODO: Clean up, or fail, the message (https://github.com/TheThingsNetwork/lorawan-stack/issues/6128).
+	msg.DecodedPayload, _ = structpb.NewStruct(msg.DecodedPayload.AsMap())
```
KrishnaIyer: This errors only on invalid UTF-8 it seems. Fine then.

KrishnaIyer (approved):
If this is the direction we are going, Implementation LGTM.

pkg/messageprocessors/javascript/javascript.go
adriansmares: I'll actually emit the error in order to avoid showing an empty decoded payload.

KrishnaIyer (approved):

johanstokking (approved):

PR #6139, opened 2023-04-03
Fix RPC ordering in PBA

adriansmares:
#### Summary

This PR fixes a regression introduced by our upgrade to `grpc-gateway/v2`. The `v2` version of `grpc-gateway` is using by default the ['last-match-wins'](https://grpc-ecosystem.github.io/grpc-gateway/docs/development/grpc-gateway_v2_migration_guide/#withlastmatchwins-and-allow_colon_final_segmentstrue-is-now-default-behaviour) behavior of path matching. This means that the last path in the service definition has a higher priority when it comes to matching.

Since we have both `/pba/home-networks/policies/default` and `/pba/home-networks/policies/{net_id}`, the `/default` path needs to appear after the `/{net_id}` inside the file, otherwise the matched route will be the wrong one.

For gRPC purposes this is a no-op, but the gRPC-gateway does care about RPC order in the service file.

#### Changes

- Move `GetHomeNetworkDefaultRoutingPolicy`, `SetHomeNetworkDefaultRoutingPolicy` and `DeleteHomeNetworkDefaultRoutingPolicy` after `GetHomeNetworkRoutingPolicy`, `SetHomeNetworkRoutingPolicy` and `DeleteHomeNetworkRoutingPolicy`.

#### Testing

Local testing.

##### Regressions

N/A. This fixes a regression on its own.

#### Notes for Reviewers

To be backported to `v3.25.0`.

#### Checklist


KrishnaIyer (approved):
Whoa nice one.

PR #6151, opened 2023-04-13
Validate transmission settings frequency

adriansmares:
#### Summary

This PR adds a validation rule to the transmission settings message, in order to allow us to drop uplinks with an invalid frequency (generally 0) at Gateway Server level.

The Network Server was already dropping these uplinks, as it was unable to find the appropriate end device channel for them:
Join request handling: https://github.com/TheThingsNetwork/lorawan-stack/blob/881ec4fc78e61e1e05280cff3dee0ea796a91877/pkg/networkserver/grpc_gsns.go#L1243-L1246
Uplink message handling: https://github.com/TheThingsNetwork/lorawan-stack/blob/881ec4fc78e61e1e05280cff3dee0ea796a91877/pkg/networkserver/grpc_gsns.go#L648-L652

#### Changes

- Validate transmission settings frequency in order to needlessly process invalid frames.

#### Testing

CI.

##### Regressions

As mentioned above, these frames were never matched with an end device, so we're just dropping invalid traffic.

#### Checklist


api/lorawan.proto
```diff
@@ -444,7 +444,7 @@ message TxSettings {
   reserved 2; // data_rate_index
   reserved 3; // coding_rate
   // Frequency (Hz).
-  uint64 frequency = 4;
+  uint64 frequency = 4 [(validate.rules).uint64.gte = 100000];
```
KrishnaIyer, line 447: Why 100khz?

adriansmares: This is our standard `gte` used for frequencies - if you check for `[(validate.rules).uint64.gte = 100000]` you will see that most of the times this is what we check for.

Why 100k ? Probably because this is the smallest possible LoRaWAN frequency in the specification. If you send a frequency to an end device via some MAC command, you don't send the frequency itself, but frequency/100k.

johanstokking (approved):

KrishnaIyer (approved):

PR #6157, opened 2023-04-14
Fix region name usage in Basics Station

adriansmares:
#### Summary

This PR fixes our usage of the `region` field in the `router_config` we send to Basics Station gateways.

Version 2.0.6 of Basics Station started being more thorough in the checks it does for the region names. Historically, you could provide any `region` and unknown regions would just be [silently accepted without any form of validation](https://github.com/lorabasics/basicstation/blob/817e6e1e7f0c3affce078bac7b060ba675043664/src/s2e.c#L957-L1000).

Version 2.0.6 [no longer does that](https://github.com/lorabasics/basicstation/blob/ba4f85d80a438a5c2b659e568cd2d0f0de08e5a7/src/s2e.c#L973-L1041). If we send an 'unknown' `region`, Basics Station [will refuse any downlinks](https://github.com/lorabasics/basicstation/blob/ba4f85d80a438a5c2b659e568cd2d0f0de08e5a7/src/s2e.c#L1794-L1798).

#### Changes

- Convert the band IDs to region IDs using explicit mappings.
   - The resulting region is the region which was previously used, except for `AS923` which was supposed to be `AS923JP` actually.
   - Since we send the legacy names, we ensure that versions predating 2.0.6 also continue to act the same.

#### Testing

Local testing with my physical gateway fleet, unit testing.

##### Regressions

Certain gateway vendors _may_ use the `region` field to limit the regions in which the gateway operates. If they were expecting the wrong value for `AS923` (`AS923` instead of `AS923JP`), we may have some compatibility issues.

Preliminary assessment of connected gateways connected to TTSC shows that we don't have large fleets of Basic Station gateways connected with `AS923`-based frequency plans.

#### Checklist

### DIFF
--- a/pkg/pfconfig/lbslns/lbslns.go
+++ b/pkg/pfconfig/lbslns/lbslns.go
@@ -30,9 +30,21 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
 )
 
-const (
-	configHardwareSpecPrefix = "sx1301"
-)
+const configHardwareSpecPrefix = "sx1301"
+
+// Based on
+// https://github.com/lorabasics/basicstation/blob/ba4f85d80a438a5c2b659e568cd2d0f0de08e5a7/src/s2e.c#L973-L1041 .
+// Note that versions 2.0.6 or higher will reject any downstream messages such as downlinks if the
+// region ID is unknown, while older versions will be more forgiving.
+// Non standard names are used for backwards compatibility reasons.
+var bandIDToRegionID = map[string]string{
+	band.EU_863_870: "EU863", // Non standard name, officially `EU868`.
+	band.AS_923_4:   "IL915", // Non standard name, officially `AS923-4`.
+	band.KR_920_923: "KR920",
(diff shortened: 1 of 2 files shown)


adriansmares:
This is waiting on clarifications on whether or not it will break region validation code for some vendors.

Edit: confirmed as not breaking.

pkg/pfconfig/lbslns/lbslns.go
```diff
@@ -30,9 +30,21 @@ import (
...
 )
 
-const (
-	configHardwareSpecPrefix = "sx1301"
-)
+const configHardwareSpecPrefix = "sx1301"
+
+// Based on
+// https://github.com/lorabasics/basicstation/blob/ba4f85d80a438a5c2b659e568cd2d0f0de08e5a7/src/s2e.c#L973-L1041 .
+// Note that versions higher than 2.0.6 will reject any downstream messages such as downlinks if the
```
KrishnaIyer: `2.0.6 or higher`

KrishnaIyer (approved):

Issue #6160, opened 2023-04-14
UDP frontend concurrency issues


adriansmares:
### Summary

The UDP frontend of the Gateway Server is designed to operate even in situations in which the push UDP flow is connected to one Gateway Server instance, while the pull flow is connected to another Gateway Server instance.

This functionality works and we do see in the wild uplinks received on one instance, while the corresponding downlink is sent via another instance which holds the identifiers claim. With that being said, there are some UX issues:

1. The same gateway is counted twice when the flows are connected to separate Gateway Server instances. This happens because each flow counts as a connection in the Gateway Server.
2. If for some reason both flows start by being connected to the same instance, but later on split (due to loss of downlink path since UDP packets were lost), the original instance will continue to push statistics for the flow which is no longer connected.

The following situation is possible:
1. Connect both flows to a Gateway Server instance.
2. The gateway sends some uplinks and receives some downlinks.
3. One flow (lets say the pull) expires.
4. The flow reconnects to another instance.
5. Both instances now push downlink statistics, because the original one had some traffic, so its connection state is not clean.

The same situation may occur for the uplink flow disconnecting.

### Steps to Reproduce

1. Be terribly unlucky.
2. Have the steps described above happen.

### Current Result

The two instances fight over the gateway stats in Redis. The downlink counter and timestamps go back and forth between a stale value and the real value. The gateway is also counted twice in the metrics.

### Expected Result

The downlink counter and timestamps should be correct, and we should not count gateways twice.

### Relevant Logs

```shell
N/A.
```

### Deployment

The Things Stack Enterprise (self-hosted)

### The Things Stack Version

3.25.0

### Proposed Fix

Frontends (the UDP frontend really) are aware of the staleness of their flows. This information should be made available to the underlying `io.Connection`, such that the connection can avoid overwriting the counters and timestamps.

### Contributing

- [X] I can help by doing more research.
- [X] I can help by implementing a fix after the proposal above is approved.
- [X] I can help by testing the fix before it's released.

### Code of Conduct

- [X] I agree to follow TTN's [Community Code of Conduct](https://github.com/TheThingsNetwork/Manifesto/blob/master/CodeOfConduct.md).

PR #6161, opened 2023-04-14
UDP Gateway Server frontend technical debt fixes

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/6160

#### Changes

- Repeated uplink detection now is safe for concurrent use, and stores only the hash of the payload, instead of the full uplink payload.
- Fix some linting issues in the UDP frontend:
  - `config` variable use renamed to `conf` in order to avoid `config` package shadowing.
  - `state` variable use renamed to `st` in order to avoid `state` struct shadowing.
  - `delete` variable use renamed to `del` in order to avoid `delete` standard function shadowing.
  - Use `binary.BigEndian` encoding to encode the downlink tokens, instead of manually doing the encoding via bitwise operations.
- Use `atomic.Pointer` instead of `atomic.Value` for type safety purposes.
- Do not override the `connected_at`, `disconnected_at`, `protocol` and `gateway_remote_address` on every stats call.
   - There can be multiple connections updating these timestamps, and we're better off updating them only once on connect/disconnect.
- Add transmission acknowledgement counters to the status messages.
  - This allows Basics Station and UDP users to compare submitted downlinks vs actually transmitted downlinks.

#### Testing

Unit testing.

##### Regressions

The existing structure of the code is generally unchanged, minus some subtle refactorings to avoid excessive nesting. Existing unit tests cover the changes.

#### Notes for Reviewers

This is a precursor to https://github.com/TheThingsNetwork/lorawan-stack/issues/6160 .

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/io.go
+++ b/pkg/gatewayserver/io/io.go
@@ -89,8 +89,9 @@ type Connection struct {
 	connectTime,
 	lastStatusTime,
 	lastUplinkTime,
-	lastDownlinkTime int64
-	lastStatus atomic.Value
+	lastDownlinkTime,
+	lastRepeatUpTime int64
+	lastStatus atomic.Value // *ttnpb.GatewayStatus
 
 	ctx       context.Context
 	cancelCtx errorcontext.CancelFunc
@@ -103,6 +104,7 @@ type Connection struct {
 	fps              *frequencyplans.Store
 	scheduler        *scheduling.Scheduler
 	rtts             *rtts
+	addr             *ttnpb.GatewayRemoteAddress
 
 	upCh     chan *ttnpb.GatewayUplinkMessage
 	downCh   chan *ttnpb.DownlinkMessage
@@ -114,10 +116,7 @@ type Connection struct {
 
 	versionInfoCh chan struct{}
 
-	lastUplink            *uplinkMessage
-	lastRepeatUpEventTime time.Time
-
-	addr *ttnpb.GatewayRemoteAddress
+	lastUplink atomic.Value // *uplinkMessage
 }
 
 type uplinkMessage struct {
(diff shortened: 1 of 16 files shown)


pkg/gatewayserver/io/io.go
```diff
@@ -103,27 +106,22 @@ type Connection struct {
...
+	statsChangedCh       chan struct{}
+	locChangedCh         chan struct{}
+	versionInfoChangedCh chan struct{}
 }
 
 type uplinkMessage struct {
-	payload   []byte
-	frequency uint64
-	antennas  []uint32
+	payloadHash uint64
```
adriansmares: Should we also store the data rate in order to avoid false positives on actual end device retransmissions ?

pkg/gatewayserver/io/udp/udp.go
```diff
@@ -383,16 +387,20 @@ func (s *srv) handleDown(ctx context.Context, state *state) error {
...
 				logger.WithError(err).Warn("Failed to marshal downlink message")
 				// TODO: Report to Network Server: https://github.com/TheThingsNetwork/lorawan-stack/issues/76
 				break
 			}
-			downlinkPath := state.lastDownlinkPath.Load().(downlinkPath)
+			downlinkPath := st.lastDownlinkPath.Load()
+			if downlinkPath == nil {
+				logger.Debug("Received downlink message without an active downlink path")
+				break
+			}
```
KrishnaIyer, line 403: How about?
```go
downlinkPath, ok := state.lastDownlinkPath.Load().(downlinkPath)
if !ok {
    logger.Debug("Received downlink message without an active downlink path")
}
```

pkg/gatewayserver/io/udp/udp.go
```diff
@@ -169,13 +170,16 @@ func (s *srv) handlePacket(ctx context.Context, packet encoding.Packet) {
...
-			if !errors.IsResourceExhausted(err) || ratelimit.Require(s.limitLogs, ratelimit.NewCustomResource(eui.String())) == nil {
-				logger.WithError(err).Warn("Packet filtered")
-			}
+	if err := s.firewall.Filter(packet); err != nil {
+		if !errors.IsResourceExhausted(err) {
+			goto filtered
+		}
+		if ratelimit.Require(s.limitLogs, ratelimit.NewCustomResource(eui.String())) != nil {
 			return
 		}
```
KrishnaIyer, line 179: So we silently drop rate-limited packets?

pkg/gatewayserver/io/udp/udp.go
adriansmares: The value stored is no longer `downlinkPath`, but `*downlinkPath`, and there is no `interface{}` in between - the result of `Load` is of type `*downlinkPath`.

pkg/gatewayserver/io/udp/udp.go
adriansmares: Yes, specifically we avoid printing the rate limiting logs _from the same gateway_. This was the old behavior too, but the line was so long that it triggered linting warnings.

pkg/gatewayserver/io/io.go
KrishnaIyer: Yes good one.

KrishnaIyer (approved):

pkg/gatewayserver/io/io.go
johanstokking: This duplication check uses a way shorter window than a possible retransmission, right?

johanstokking (approved):

pkg/gatewayserver/io/io.go
adriansmares: There is no time window check for this logic here - the gateway may send the 'duplicate' even after 2 days.

The original PR is https://github.com/TheThingsNetwork/lorawan-stack/pull/4181 and the original issue is https://github.com/TheThingsNetwork/lorawan-stack/issues/4095 . I think that the time window logic never got implemented, and for the past years we may have been dropping some legitimate retransmissions which had different data rates, but the same frequency. If we would compare data rates too, the only possible false positive is SF12 on the same frequency, which is less likely than the current frequency re-use chance of collision.

The current fields are based on your description [here](https://github.com/TheThingsNetwork/lorawan-stack/pull/4181#discussion_r635905772), but I cannot tell why you didn't include the data rate. Do you recall by any chance ?

johanstokking: I think I was assuming the duplicate check would be time-bound, as I wrote in #4095. Without diving in the code, we're not keeping track of all uplink messages in the connection for duplicate checks, right?

adriansmares: > Without diving in the code, we're not keeping track of all uplink messages in the connection for duplicate checks, right?

Correct - we always compare the last processed uplink only with the pending uplink (pending approval, since we would reject an uplink which is identical).

johanstokking: Oh okay. Yeah then we'd need to factor in data rate too. In fact, the 10 seconds I was talking about in #4095 wouldn't be enough for retransmissions either. I think I was assuming that a retransmission would only use another frequency, not a different data rate.

PR #6162, opened 2023-04-17
Support `administrative_contact` and `technical_contact` in the Console

### LINKED ISSUE #6112: Support `administrative_contact` and `technical_contact` in the Console
### Summary

References #4930

With the addition of  Administrative/Technical in #4996, there is the need to add support for it on the Console. 

This means that `Clients`, `Gateways`, `Applications` and `Organizations` would benefit from the support of the `administrative_contact` and `technical_contact` fields and that `Users` have their contact information deprecated. Furthermore this implies in organization not sending emails to a set of users but to a specific email determined by the admin user.

With this in mind, there is the need to update the entities mentioned to support the `administrative_contact` and the `technical_contact`.

### Current Situation

Currently there are pages withing these entities: `Application`, `Gateway`, `Client` which references to an user with an `primary_email` or a set of `users` (an organization).

Add the support for `administrative_contact` and the `technical_contact` within those entities, allowing to trigger emails to a specific target instead of the whole set of users that belong to that organization.

### Proposed Implementation

I lack the knowledge to know what is the best approach on the front-end, but I imagine that adding fields similar to the `email_address` the user's profile section but applied to administrative_contact and technical_contact would be appropriate.

### Contributing

### Code of Conduct

ryaplots:
#### Summary

Closes #6112 

#### Changes

- Add contacts to relevant entities.

<details>
Screenshot 2023-07-10 at 11 22 15
Screenshot 2023-07-10 at 11 22 05
</details>

#### Testing

Manually

#### Notes for Reviewers

I decided to add this in the entity general settings, as this is the place where I have access to the collaborator's rights for the entity, and not in the user settings as we initially thought.

#### Checklist


kschiffer:
I think we'll have to rework the UX here a bit. See also my comment here:
https://github.com/TheThingsNetwork/lorawan-stack/issues/6112#issuecomment-1522148730

I'll look into this later this week/next week. Should not affect the implementation as a whole too much though.

kschiffer:
Currently blocked, so reopening once #6112 is merged.

ryaplots:
This is unblocked.

pkg/webui/account/components/oauth-client-form/index.js
```diff
@@ -156,6 +184,40 @@ const OAuthClientForm = props => {
           />
         </>
       )}
+      {update && mayEditBasicInformation && (
+        <>
+          <Form.SubTitle title={sharedMessages.contactInformation} className="mb-cs-s" />
+          <Notification small warning content={m.contactWarning} />
```
kschiffer:
```suggestion
```
Let's remove these notifications. The user will see that the initial user is selected initially and it's not possible to leave the input empty.

pkg/webui/account/components/oauth-client-form/index.js
```diff
@@ -156,6 +184,40 @@ const OAuthClientForm = props => {
...
+            entityId={clientId}
+            encode={encodeContact}
+            decode={decodeContact}
+          />
+          <Message
+            content={m.adminContactDescription}
+            component="p"
+            className="mt-cs-xs tc-subtle-gray"
+          />
+          <CollaboratorSelect
```
kschiffer, line 208: If there is only one collaborator anyway, the input should be disabled and locked to this collaborator. This can be done inside the `CollaboratorSelect`.

pkg/webui/account/components/oauth-client-form/index.js
```diff
@@ -156,6 +184,40 @@ const OAuthClientForm = props => {
...
+            name="technical_contact"
+            title={sharedMessages.technicalContact}
+            placeholder={sharedMessages.contactFieldPlaceholder}
+            entity={'client'}
+            entityId={clientId}
+            encode={encodeContact}
+            decode={decodeContact}
```
kschiffer, line 215:
```suggestion
            decode={decodeContact}
            required
```
At least it's currently not possible to submit these forms the selector empty.

pkg/webui/console/components/application-general-settings-form/index.js
```diff
@@ -68,6 +85,16 @@ const validationSchema = Yup.object().shape({
...
+  administrative_contact: Yup.object().when(['organization_ids'], {
+    is: organizationIds => Boolean(organizationIds),
+    then: schema => schema.concat(organizationSchema),
+    otherwise: schema => schema.concat(userSchema),
+  }),
+  technical_contact: Yup.object().when(['organization_ids'], {
+    is: organizationIds => Boolean(organizationIds),
+    then: schema => schema.concat(organizationSchema),
+    otherwise: schema => schema.concat(userSchema),
+  }),
```
kschiffer: Generally, I would suggest making these fields required. It's only confusing to be able to remove the collaborator only to get them implicitly set to the first collaborator. There's always going to be a technical/admin contact so let's make it explicit.

pkg/webui/containers/collaborator-select/util.js
```diff
@@ -0,0 +1,54 @@
...
+      ? value.ids?.user_ids.user_id
+      : value.ids?.organization_ids.organization_id,
+  icon: 'user_ids' in value?.ids ? 'user' : 'organization',
+})
+
+export const composeContactOption = value => ({
+  value: 'user_ids' in value ? value.user_ids.user_id : value.organization_ids.organization_id,
+  label: 'user_ids' in value ? value.user_ids.user_id : value.organization_ids.organization_id,
+  icon: 'user_ids' in value ? 'user' : 'organization',
+})
```
kschiffer:
```suggestion
export const composeOption = value => {
  const data = value.ids || value;
  return {
    value: 'user_ids' in data 
      ? data.user_ids?.user_id 
      : data.organization_ids?.organization_id,
    label: 'user_ids' in data 
      ? data.user_ids?.user_id 
      : data.organization_ids?.organization_id,
    icon: 'user_ids' in data ? 'user' : 'organization',
  }
}
```

pkg/webui/containers/collaborator-select/index.js
```diff
@@ -31,7 +31,7 @@ import PropTypes from '@ttn-lw/lib/prop-types'
 import { getCollaboratorsList } from '@ttn-lw/lib/store/actions/collaborators'
 import { selectCollaborators } from '@ttn-lw/lib/store/selectors/collaborators'
 
-import composeOption from './util'
+import { composeOption } from './util'
 
 import styles from './collaborator-select.styl'
```
kschiffer, line 39: Can you change line 40 to:
```
    <Message content={m.suggestions} className="ml-cs-s mt-cs-xxs mb-cs-xxs" component="h4" />
```

pkg/webui/account/components/oauth-client-form/messages.js
```diff
@@ -54,4 +54,11 @@ export default defineMessages({
   grantTypeAndRights: 'Grant types and rights',
   stateDescriptionDesc:
     'You can use this field to save additional information about the state of this OAuth client, e.g. why it has been flagged',
+  contactWarning:
+    'Note that if no contact is provided, it will default to the first collaborator of the client.',
```
kschiffer: It's going to be removed anyways but for the record: there's no point in redefining this same message over and over for each entity. This is what we have shared messages for.

kschiffer (changes requested):

adriansmares:
Does this take in account that in certain deployments collaborators can set only themselves as contacts, but not others (https://github.com/TheThingsNetwork/lorawan-stack/pull/6378) ?

ryaplots:
> Does this take in account that in certain deployments collaborators can set only themselves as contacts, but not others (#6378) ?

Should we inform those users of that, with a notification? @kschiffer

kschiffer:
As far as I understand, we can determine this based on the IS configration fetched from the backend. If indeed the user fan only set itself, the console should deactivate the selector field and render a button 'Set yourself as contact' to the fields. Clicking the button will then put the user into the field.

There should also be a notification above the field saying:
> You can only set yourself as a contact. If you would like another collaborator set as contact, please contact this collaborator to self-assign as contact.

kschiffer (commented):
EDIT: Disregard; didn't pull latest changes.

pkg/webui/components/select/suggested-select.js
```diff
@@ -79,6 +79,11 @@ const SuggestedSelect = props => {
   const { formatMessage } = useIntl()
   const [inputValue, setInputValue] = useState(value)
 
+  // React memoizes the initial state value and does not update it unless explicitly told to do so.
+  useEffect(() => {
+    setInputValue(value)
+  }, [value])
+
```
kschiffer: Why do you need to store the value again in a state anyway?

pkg/webui/containers/collaborator-select/index.js
```diff
@@ -79,47 +92,70 @@ const Suggest = ({ initialOptions, userInputCustomComponent, entity, entityId, .
...
+        {...rest}
+        name={name}
+        defaultOptions={initialOptions}
+        component={Select.Suggested}
+        noOptionsMessage={handleNoOptions}
+        loadOptions={handleLoadingOptions}
+        showOptionIcon
+        openMenuOnFocus
+        className={styles.collaboratorSelect}
+        customComponents={{ SingleValue, Menu: customMenu }}
```
kschiffer, line 157: `Suggestions` should not show when the user typed in a value and the search had no results.

pkg/webui/containers/collaborator-select/index.js
```diff
@@ -79,47 +92,70 @@ const Suggest = ({ initialOptions, userInputCustomComponent, entity, entityId, .
...
+          await dispatch(attachPromise(searchAccounts(value, onlyUsers, collaboratorOf)))
+          const newOptions = searchResultsRef?.current?.map(account => ({
+            value:
+              'user_ids' in account
+                ? account.user_ids?.user_id
+                : account.organization_ids?.organization_id,
+            label:
+              'user_ids' in account
+                ? account.user_ids?.user_id
+                : account.organization_ids?.organization_id,
```
kschiffer, line 114: The label should include both the ID (in bold) and the full name (regular, subtle gray, if exists). Please take another look at the wireframe:
image

pkg/webui/account/components/oauth-client-form/messages.js
```diff
@@ -54,4 +54,9 @@ export default defineMessages({
   grantTypeAndRights: 'Grant types and rights',
   stateDescriptionDesc:
     'You can use this field to save additional information about the state of this OAuth client, e.g. why it has been flagged',
+  adminContactDescription:
+    'Administrative contact information for this client. Typically used to indicate who to contact with administrative questions about the client.',
+  techContactDescription:
+    'Technical contact information for this client. Typically used to indicate who to contact with technical/security questions about the client.',
+  contactPlaceholder: 'Type to choose a contact',
```
kschiffer:
```suggestion
```

pkg/webui/lib/shared-messages.js
```diff
@@ -115,6 +116,7 @@ export default defineMessages({
   connecting: 'Connecting',
   connectionIssues: 'Connection issues',
   contactInformation: 'Contact information',
+  contactFieldPlaceholder: 'Type to choose a contact',
```
```suggestion
  contactFieldPlaceholder: 'Type or choose a collaborator',
```
Diverting from the wireframe here as I think `collaborator` is more fitting.

pkg/webui/console/views/gateway-general-settings/basic-settings-form/validation-schema.js
```diff
@@ -84,6 +86,20 @@ const validationSchema = Yup.object().shape({
...
+      otherwise: schema => schema.concat(userSchema),
+    })
+    .required(sharedMessages.validateRequired),
+  technical_contact: Yup.object()
+    .when(['organization_ids'], {
+      is: organizationIds => Boolean(organizationIds),
+      then: schema => schema.concat(organizationSchema),
+      otherwise: schema => schema.concat(userSchema),
+    })
+    .required(sharedMessages.validateRequired),
```
kschiffer: There's a lot duplication now with this schema. I'd suggest to outsource this into a lib: `/webui/console/lib/shared-schemas.js`

kschiffer (changes requested):

kschiffer:
```diff
diff --git a/pkg/webui/lib/store/selectors/collaborators.js b/pkg/webui/lib/store/selectors/collaborators.js
index 8293e4f83..940b6f468 100644
--- a/pkg/webui/lib/store/selectors/collaborators.js
+++ b/pkg/webui/lib/store/selectors/collaborators.js
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { createSelector } from 'reselect'
+
 import { createFetchingSelector } from '@ttn-lw/lib/store/selectors/fetching'
 import { createErrorSelector } from '@ttn-lw/lib/store/selectors/error'
 import {
@@ -59,8 +61,10 @@ const createSelectCollaboratorsFetchingSelector = createFetchingSelector(
 )
 const createSelectCollaboratorsErrorSelector = createErrorSelector(GET_COLLABORATORS_LIST_BASE)
 
-export const selectCollaborators = state =>
-  createSelectCollaboratorsIdsSelector(state).map(id => selectCollaboratorById(state, id))
+export const selectCollaborators = createSelector(
+  [createSelectCollaboratorsIdsSelector, selectCollaboratorsEntitiesStore],
+  (collaboratorIds, collaborators) => collaboratorIds.map(id => collaborators[id]),
+)
 export const selectCollaboratorsTotalCount = state =>
   createSelectCollaboratorsTotalCountSelector(state)
 export const selectCollaboratorsFetching = state => createSelectCollaboratorsFetchingSelector(state)
```
Please also apply this change to remove the warning in the Console, and please make sure to not ignore warnings:
image

pkg/webui/components/select/suggested-select.js
ryaplots: This was how the original `<Select />` handled the value, so I just kept it. But now I see i don't to have a state. I can use the value from the form context.

pkg/webui/containers/collaborator-select/index.js
ryaplots: The search API only return the id, which is why I think I overlooked this. To get the name I would need to make another request to fetch the collaborator with that id and add the name, which to me sounds like too many requests for a single input field. Maybe editing the API to also return the name is a better solution? Or not include the name at all. What do you think?

kschiffer: @adriansmares Do you think we can add the full name to the account search API as well here?
Another option would be to add a `query` to the `List{Entity}Collaborator` API and then use this API instead here.

I also wouldn't mind launching a separate API call. You could traverse this page-wise using the ListCollaborators request until all names have been found. That should be just one additional API call for 99.9% of the time if you ramp up the limit to 1000.

Another thing I wondered is, if it is right that we enable organizations to be set as contacts (cc @adriansmares)

kschiffer: @ryaplots I'm did the same thing for a similar situation here:
https://github.com/TheThingsNetwork/lorawan-stack/blob/d068363dc0def9b48b95d17e2cf11d37c069b9b2/pkg/webui/console/store/middleware/logics/packet-broker.js#L43-L61

ryaplots: Isn't it better to just return the name as well in the original search request? @adriansmares @kschiffer

adriansmares: Organizations can be contacts for other entities, so it is not a mistake to also search for organizations when it comes to contacts. Think of the following use case - you have a fleet of gateways and you would like to set the technical contact for all of them. If you set a user, you will have to update all of them later on, but if you set an organization, you just need to update the technical contact in that organization.

---

Regarding the `name`: the problem is that these are two separate types of entities: accounts and users - the fact that they have some shared attributes (`name`, `description`, `attributes`) does not change that the entity types can be different - it is not as straight forward as it sounds at face value.

Now this doesn't mean that it is not a solvable problem - we can consider having in the API an entity which contains these common attributes, and then adding a field mask to the search request which would be able to return only these common fields. With that being said, I would prefer not to stall this functionality now and rather implement this in a subsequent PR. I do want to have this in the API however - I agree that it would be useful.

kschiffer: Alright. Then let's solve it by an additional API request like I suggested for now and create an issue for extending the account search API. @ryaplots can you do that?

You can just add this to the account search store logic. After fetching the accounts, collect the IDs, then cycle through the collaborator list and decorate the account search result and filter by the IDs to collect the collaborator names. Ideally, the collaborator names would be cached somehow so that we don't need to launch the request all the time.

kschiffer (approved):

Issue #6102, opened 2023-03-09
Use OpenTelemetry gorilla/mux middleware


adriansmares:
### Summary

We should add the OpenTelemetry gorilla/mux middleware to our routers.

### Current Situation

We do not trace HTTP server handlers at all.

### Why do we need this? Who uses it, and when?

In order to have some rudimentary tracing of web calls.

### Proposed Implementation

Use [otelmux](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/gorilla/mux/otelmux) middleware as part of the `router` and `apiRouter` in `pkg/web`. See [the example](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/gorilla/mux/otelmux/example/server.go) and as always don't forget to inject the trace provider.

### Contributing

- [ ] I can help by doing more research.
- [ ] I can help by implementing the feature after the proposal above is approved.
- [X] I can help by testing the feature before it's released.

### Code of Conduct

- [X] I agree to follow TTN's [Community Code of Conduct](https://github.com/TheThingsNetwork/Manifesto/blob/master/CodeOfConduct.md).

PR #6167, opened 2023-04-20
Use OpenTelemetry `gorilla/mux` middleware

happyRip:
#### Summary

References #6102

#### Changes

- Add `gorilla/mux` otel middleware

#### Testing

Tested the same way as in #6004 
localhost_16686_trace_90b9453b276798d36b30d0a29b167de6(Laptop with HiDPI screen)

##### Regressions

nil

#### Notes for Reviewers

nil

#### Checklist


pkg/web/web.go
```diff
@@ -188,6 +190,7 @@ func New(ctx context.Context, opts ...Option) (*Server, error) {
 		mux.MiddlewareFunc(webmiddleware.Log(logger, options.logIgnorePaths)),
 		mux.MiddlewareFunc(webmiddleware.Cookies(hashKey, blockKey)),
 		mux.MiddlewareFunc(webmiddleware.NoCache),
+		otelmux.Middleware("ttn-lw-stack", otelmux.WithTracerProvider(tracing.FromContext(ctx))),
```
johanstokking: Is it intentional that this is the last middleware?

The other middlewares are, or can be, request specific, so I don't see a reason why they would not be included in the trace.

happyRip: It wasn't intentional, moved it to the top spot.

johanstokking: It should be below Recover, because if this middleware panics, TTS panics.

Is there anything in the tracer that comes from the request context filled by TTS? In that case, it should also below `FillContext`. But it shouldn't be (much) lower.

happyRip: > Is there anything in the tracer that comes from the request context filled by TTS?

No, then just below panic middleware.

johanstokking (approved):

adriansmares (approved):

kschiffer (approved):

PR #6168, opened 2023-04-21
Check frontend stream status before emiting stats message

### LINKED ISSUE #6160: UDP frontend concurrency issues

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/6160

#### Changes

- Add a simple stream states mechanism which allows GS frontends to be queried if a stream is still active.
- Fix a race condition in the rate limiting UDP frontend.

#### Testing

Local testing.

##### Regressions

This functionality affects only stats book keeping, without affecting the hot path behavior.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/stream_state.go
+++ b/pkg/gatewayserver/io/stream_state.go
@@ -18,11 +18,11 @@ package io
 type MessageStream uint32
 
 const (
-	UplinkStream   MessageStream = 0 // UplinkStream is the uplink message stream.
-	DownlinkStream MessageStream = 1 // DownlinkStream is the downlink message stream.
-	TxAckStream    MessageStream = 2 // TxAckStream is the transmission acknowledgment stream.
-	StatusStream   MessageStream = 3 // StatusStream is the status message stream.
-	RTTStream      MessageStream = 4 // RTTStream is the round-trip times stream.
+	UplinkStream   MessageStream = iota // UplinkStream is the uplink message stream.
+	DownlinkStream                      // DownlinkStream is the downlink message stream.
+	TxAckStream                         // TxAckStream is the transmission acknowledgment stream.
+	StatusStream                        // StatusStream is the status message stream.
+	RTTStream                           // RTTStream is the round-trip times stream.
 )
 
 func alwaysOnStreamState(MessageStream) bool { return true }
--- a/pkg/gatewayserver/io/stream_state.go
+++ b/pkg/gatewayserver/io/stream_state.go
@@ -1,0 +1,28 @@
+// Copyright © 2023 The Things Network Foundation, The Things Industries B.V.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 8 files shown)


pkg/gatewayserver/io/stream_state.go
```diff
@@ -0,0 +1,28 @@
...
+
+// MessageStream is a message stream.
+type MessageStream uint32
+
+const (
+	UplinkStream   MessageStream = 0 // UplinkStream is the uplink message stream.
+	DownlinkStream MessageStream = 1 // DownlinkStream is the downlink message stream.
+	TxAckStream    MessageStream = 2 // TxAckStream is the transmission acknowledgment stream.
+	StatusStream   MessageStream = 3 // StatusStream is the status message stream.
+	RTTStream      MessageStream = 4 // RTTStream is the round-trip times stream.
```
johanstokking: nit;

```suggestion
	UplinkStream   MessageStream = iota // UplinkStream is the uplink message stream.
	DownlinkStream  // DownlinkStream is the downlink message stream.
	TxAckStream     // TxAckStream is the transmission acknowledgment stream.
	StatusStream    // StatusStream is the status message stream.
	RTTStream       // RTTStream is the round-trip times stream.
```

needs some spacing

johanstokking (approved):

PR #6181, opened 2023-04-28
Support claim in device import

### LINKED ISSUE #6131: Support claim in device import
### Summary

Support claim in device import

References https://github.com/TheThingsIndustries/lorawan-stack-support/issues/936

### Current Situation

Now, the Console uses Device Template Converter (DTC) to turn a binary file into an end device template. This template may then get decorated with additional fields (LoRaWAN and RP version, generate claim authentication code). The resulting end device structure is then simply created on IS, NS, AS and JS.

We also have support for claiming via Device Claiming Server (DCS).

The CLI already uses DCS as part of create; it will try to claim the device via DCS before the end device is actually created in IS, NS and AS (no JS).

We need this for UX and consistency; the claim flow is already integrated in creating devices in the Console. It would be handy to be able to import a CSV or JSON file with claim authentication codes to batch claim devices.

We also have CLI support for claiming in a batch.

### Proposed Implementation

Technically, it would have to work like the end device create functionality in CLI.

I would guess that this needs to be implemented mainly in https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.25/sdk/js/src/service/devices/index.js#L467-L480.

Otherwise it should follow mainly https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.25/cmd/ttn-lw-cli/commands/end_devices.go#L601-L675.

What should be removed is generating a claim authentication code. We should avoid that we randomly generate a CAC and then try to claim with that CAC. @KrishnaIyer is fixing this in the CLI.

### Contributing

### Code of Conduct

ryaplots:
#### Summary

Closes #6131 

#### Changes

- Try to claim a device if a CAC is present
- Do not generate a random CAC by default

#### Testing

Import a device with a CAC

#### Checklist


sdk/js/src/service/devices/index.js
```diff
@@ -513,6 +515,29 @@ class Devices {
...
+            authentication_code: devicePayload.end_device.claim_authentication_code.value,
+          },
+          target_device_id: deviceId,
+          target_application_ids: {
+            application_id: applicationId,
+          },
+        }
+        const response = await this._api.EndDeviceClaimingServer.Claim(undefined, payload)
+
+        return Marshaler.payloadSingleResponse(response)
```
kschiffer: If the device is claimed, it will only be provisioned in the Join Server. We still need to create the entity in the registries of the other components (IS, NS, AS), so we cannot return here but have to follow the rest of the process, without creating the device on the JS again.

It's basically following the submit logic of device onboarding, which also has to do the same thing:
https://github.com/TheThingsNetwork/lorawan-stack/blob/399efad7b7374cf9536f99959b9cf9924ab4816a/pkg/webui/console/containers/device-onboarding-form/index.js#L97

1. Check if CAC exists
2. If yes, claim
3. Extract the identifiers (`device.ids`) and merge them into the to-be-created device (taking precedence)
4. Create the entity in IS, NS and AS

If we change the device creation in the SDK now to include claiming then we also have to refactor the above submit logic to use the creation function solely.

sdk/js/src/service/devices/index.js
```diff
@@ -513,6 +515,29 @@ class Devices {
...
+      } catch (error) {
+        if (error.statusCode === 400) {
+          throw error
+        }
```
kschiffer: I don't think we should only care about 400s here. If the claiming fails for whatever reason while having a CAC, we probably don't want to fall back to normal registration. I think it's legitimate to stop the process altogether then and throw the error.

@johanstokking any thoughts on this? Also, is it right to use only the CAC as an indicator for claiming, or should this be done more explicitly, so we can also allow manual registration even when setting a CAC?

kschiffer (changes requested):

sdk/js/src/service/devices/index.js
KrishnaIyer: > If the claiming fails for whatever reason while having a CAC, we probably don't want to fall back to normal registration. I think it's legitimate to stop the process altogether then and throw the error.

ACK indeed. We should not support fallbacks for claimable devices. If rollback is tricky and we don't want to leave devices in a weird state, you can also first check if claiming succeeds before creating it in the NS/AS/IS.

>  Also, is it right to use only the CAC as an indicator for claiming, or should this be done more explicitly, so we can also allow manual registration even when setting a CAC?

I think a combination of checking the JoinEUI for claimability (`/api/v3/edcs/claim/info`) and if there is a CAC is the right way to go.

sdk/js/src/service/devices/index.js
KrishnaIyer: > If the device is claimed, it will only be provisioned in the Join Server.

More specifically, on an **External** Join Server such that if claimed, the flow should skip the cluster JS.

sdk/js/src/service/devices/index.js
johanstokking: @kschiffer there is no longer value in storing the CAC if it's not directly used for claiming.

The first step should indeed be checking whether the JoinEUI is supported with claiming. If so, the CAC is required and only then the CAC should be accepted.

sdk/js/src/service/devices/index.js
kschiffer: Yes indeed.

sdk/js/src/service/devices/index.js
```diff
@@ -486,14 +488,63 @@ class Devices {
...
+            target_application_ids: {
+              application_id: applicationId,
+            },
+          }
+      try {
+        const claimResponse = await this._api.EndDeviceClaimingServer.Claim(undefined, claimPayload)
+        claimDeviceIds = Marshaler.payloadSingleResponse(claimResponse)
+      } catch (error) {
+        throw error
+      }
```
kschiffer:
```suggestion
      } catch (error) {
        throw new Error('Unable to claim end device',  { cause: error })
      }
```

kschiffer: You probably just forgot to change this but just to be sure: rethrowing a caught error without modification is a no-op.

kschiffer (changes requested):

kschiffer (approved):

sdk/js/src/service/devices/index.js
ryaplots: indeed, I forgot

PR #6182, opened 2023-04-28
Do not disable the wide channels in fixed channel plans

### LINKED ISSUE #5551: CFList inconsistent with United States 902-928 MHz, FSB 2
#### Summary
The CFList in the Join-Accept received by the end-device is inconsistent with the frequency plan [United States 902-928 MHz, FSB 2](https://github.com/TheThingsNetwork/lorawan-frequency-plans/blob/master/US_902_928_FSB_2.yml). This frequency plan specifies that uplinks can be received on US915 channels 8-15 and 65, but the ChMask received from The Things Stack enables channels 8-15 only. It seems to me that it should enable channel 65 as well.

Note: The same inconsistency may affect the other US915 channel plans (FSB 1 and FSB 3-8), as well as the corresponding AU915 channel plans, but I have not checked that.

FYI: The symptom that alerted me to this inconsistency is that the end-device can join using DR4 (a.k.a. DR12), but is subsequently "unable" (i.e., not permitted) to transmit (Unconfirmed/Confirmed) uplinks using DR4.

#### Steps to Reproduce

1. Configure a LoRaWAN 1.0.4 US915 end-device for OTAA, and to use the channel plan identified in the summary (above).
2. Configure a US915 gateway accordingly and connect it to The Things Network.
3. Enable "Verbose stream" in the "Live Data" tab of the console for the aforementioned US915 end-device (see step 1).
4. Force that end-device to undergo OTAA (i.e., rejoin The Things Network), e.g., by resetting or power-cycling the end-device.
5. Inspect the "ch_masks" value for the (verbose) entry type "Schedule join-accept for transmission" in the "Live Data" tab.

Note: The attached screenshots may clarify some of the above steps.

#### What do you see now?
The "ch_masks" value enables US915 channels 8-15 only.

Note: This was confirmed by modifying the end-device software to print out the ChMask it received.

The "ch_masks" value should enable US915 channel 65 as well.

Alternatively: The frequency plan "United States 902-928 MHz, FSB 2" should be updated to match the "ch_masks" value. I.e., remove the "lora-standard-channel" entry therein.

#### Environment
TTS v3.20.0; LoRaWAN 1.0.4 end-device using OTAA; US915 as per RP002-1.0.3, using frequency plan "United States 902-928 MHz, FSB 2."

Note: I do not recall observing the same end-device symptom (described in "Summary") before last week, so perhaps the inconsistency was introduced in v3.20.0? (Upgrade to that version occurred on June 13, 2022.)

I do not.

N/A.

N/A.
End-Device-Overview
Join-Accept-ChMask-0-45
Join-Accept-ChMask-45-71

adriansmares:
#### Summary

Closes https://github.com/TheThingsNetwork/lorawan-stack/issues/5551

#### Changes

- The wide channel (i.e. the LoRa Standard Channel, or 500KHz, or whatever you want to call it) in the fixed channel plans such as US915 and AU915 is now enabled by default.
- Simplify the generation of desired channels - in fixed channel plans the data rates or frequencies cannot be edited anyway, so there is no point in editing these out based on the frequency plan. A nice corollary of this is that we can drop the whole 'translate data rate because AU915 before RP1-1.0.2 did not have SF11 and SF12'. 

#### Testing

Unit testing. Also tested on `staging1` with both LoRaMac-node and LoRa Basics Modem by forcing the joins to always go via the wide channels, on both `US915` and `AU915`.

##### Regressions

We have always disabled this 'wide channel' even though it is part of our channel plans and gateway do listen on it. Enabling the channel does not change the number of `LinkADRReq` commands that we need to send to the end device, and as long as the end device joined via the narrow channels (the 125KHz ones), nothing changes. 

End devices that may now join via the wide channel (which wasn't possible before, because we were sending an invalid CFList), will operate in the wide channel. A subsequent PR will add an option to the ADR algorithm to 'steer' the end devices to the narrow channels, if the users want this behavior.

#### Checklist


johanstokking (approved):
Great stuff

PR #6187, opened 2023-05-01
Add ADR channel steering configuration

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/pull/6182
References https://github.com/TheThingsNetwork/lorawan-stack/issues/5551

#### Changes

- Ensure that the dynamic ADR limits are based on narrow channel limits (so basically in EU868 the maximum value they can use for the extrema is 5, i.e. SF7BW125).
- Add note on why we don't don't adapt the margin based on the transmission power difference while changing data rates.
- Use _allowed_ data rates instead of skipping _rejected_ data rates. This makes the code reusable for the band steering mechanism.
- Add a channel steering mechanism and experimental API.
  - Has two operating modes - disabled and narrow.
  - The disabled mode simply keeps the currently active band, be it narrow, wide or non-LoRa modulated at all.
  - The narrow mode steers the end device towards the narrow (LoRa modulated, 125kHz bandwidth) channels.
  - The modes use `oneof` instead of just an `enum` for future extensions. I was thinking of adding an initial data rate (in case users do not want to jump from SF8BW500 directly to SF7BW125, in US bands for example), but we can do this later.

#### Testing

Unit tests have been added for the added functionality, and existing unit tests have been adapted if needed.
The `min`/`max` data rate index limits have been tested with `ttn-lw-cli dev set`.
The steering mode has been tested with LoRaMac-node, by forcing the end device to always join via the wide channel. After joining, the end device is steered towards the narrow channels. It is also the case that if steering is enabled for a device which is currently using the wide channels, it is still steered towards the narrow channels. Tested on both `US915` and `AU915`.

##### Regressions

Unit tests describe the new behavior and check it, while the manual testing confirms that the added code works end to end.

#### Checklist


api/end_device.proto
```diff
@@ -225,6 +225,31 @@ message ADRSettings {
...
+    // EXPERIMENTAL: Channel steering settings.
+    message ChannelSteeringSettings {
+      option (thethings.flags.message) = { select: true, set: true, semantical: true };
+
+      // Configuration options for narrow channels steering.
+      // The narrow mode attempts to steer the end device towards
+      // using the LoRa modulated, 125kHz bandwidth channels.
+      message NarrowMode {
+        option (thethings.flags.message) = { select: true, set: true };
+      }
```
johanstokking: I'm not so happy with the term "narrow". In LoRa terms, 125 kHz is more narrow than 500 kHz of course, but 125 kHz is still very wide for communications. For example, LR-FHSS uses narrow band signals spread in a band (whether that's narrow or wide is subjective).

Since this behavior, and the documentation (comment, changelog), implies that this is about LoRa 125 KHz channels, can you at least rename this to `LoRaNarrowMode` or even `LoRa125kHzMode` ?

adriansmares: I am not strongly attached to this naming, but to me it seems redundant. I agree that it is narrow with respect to LoRa, but not necessarily for radio communication as a whole.

We can do ADR only using the LoRa modulation (LR-FHSS and FSK don't have SNR), and the dynamic ADR functionality is heavily tied to the LoRa modulation. As such, this usage of narrow here is heavily tied to LoRa.

johanstokking: I'd not be surprised if we'd be able to support some ADR-like logic on LR-FHSS data rates in the future. Let's explicitly call this LoRa here so we know what we're talking about.

johanstokking (approved):

PR #6206, opened 2023-05-10
Fix connection timestamp monotonicity

adriansmares:
#### Summary

References https://www.thethingsnetwork.org/forum/t/offline-status-on-gateways-page/62935
References https://github.com/TheThingsNetwork/lorawan-stack/pull/6161

#### Changes

- Revert the periodic stats update function, such that it always updates the `connected_at`, `disconnected_at`.
  - The problem is that if we don't re-update the `connected_at` on every periodic stats a update, a disconnection from another instance can erase our `connected_at`.
- Always store the earliest `connected_at` in the stats registry.
  - This ensures that even if we have two streams connected to two GS instances, the `connected_at` does not flip-flop.

#### Testing

Unit testing.

##### Regressions

This fixes a regression introduced by the original PR, and a better fix for the original problem (flip flopping timestamps).

#### Notes for Reviewers

This will be backported to the `v3.25.2` release.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/redis/registry.go
+++ b/pkg/gatewayserver/redis/registry.go
@@ -35,10 +35,7 @@ type GatewayConnectionStatsRegistry struct {
 
 // Init initializes the GatewayConnectionStatsRegistry.
 func (r *GatewayConnectionStatsRegistry) Init(ctx context.Context) error {
-	if err := ttnredis.InitMutex(ctx, r.Redis); err != nil {
-		return err
-	}
-	return nil
+	return ttnredis.InitMutex(ctx, r.Redis)
 }
 
 func (r *GatewayConnectionStatsRegistry) key(uid string) string {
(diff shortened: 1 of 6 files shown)


pkg/gatewayserver/redis/registry.go
```diff
@@ -57,29 +60,71 @@ func (r *GatewayConnectionStatsRegistry) Set(ctx context.Context, ids *ttnpb.Gat
...
+		if stored == nil && pb == nil {
+			return nil
+		}
+		var pipelined func(redis.Pipeliner) error
+		if pb == nil {
+			pipelined = func(p redis.Pipeliner) error {
+				p.Del(ctx, uk)
+				return nil
+			}
+		} else {
```
KrishnaIyer, line 97: This branch wouldn't be used since these we never really delete stats but good to keep it consistent.

KrishnaIyer (approved):
LGTM!

PR #6212, opened 2023-05-11
Update to webpack 5

kschiffer:
#### Summary
This PR bumps webpack to the latest version.

#### Changes
- Update webpack and related modules
- Update webpack configs
- Replace dynamic module imports in the `with-locale.js` to static ones to avoid polluting the build folder with all existing language/locale combination (we want to support only English and Japanese at the moment anyway)
- Small related fixes

#### Testing

Tested this locally. Now let's see what CI says.

##### Regressions

Hard to tell but should probably be fine if build succeeds. Anything critical will be caught by the end-to-end tests.

#### Checklist


pkg/webui/console/lib/random-bytes.js
```diff
@@ -12,10 +12,34 @@
...
+ * @param {string} [type='hex'] - The type of the string to generate.
+ * @returns {string} The generated string.
+ * @example
+ * const randomString = randomBytes(16)
+ * @example
+ * const randomString = randomBytes(16, 'base64')
+ */
+export default (len, type = 'hex') => {
+  let bytes
+  if (!window.crypto) {
```
adriansmares: Isn't the logic reversed here ?

kschiffer: Oh indeed. Changed this during testing.

mjamescompton:
As a note, node 14 error 

```
error @storybook/react@7.0.9: The engine "node" is incompatible with this module. Expected version ">=16.0.0". Got "14.21.1"
error Found incompatible module.
TypeError: mkdirp is not a function
    at /Users/michaelcompton/code/lorawan-stack/tools/mage/translations.js:266:13
    at Array.map (<anonymous>)
    at writeLocales (/Users/michaelcompton/code/lorawan-stack/tools/mage/translations.js:263:26)
    at main (/Users/michaelcompton/code/lorawan-stack/tools/mage/translations.js:335:9)
Error: running "yarn install --no-progress --production=false --check-files" failed with exit code 1
running "node tools/mage/translations.js --backend-messages config/messages.json --locales pkg/webui/locales/.backend --backend-only" failed with exit code 1
```

mjamescompton:
```
(node:44294) [DEP_WEBPACK_COMPILATION_ASSETS] DeprecationWarning: Compilation.assets will be frozen in future, all modifications are deprecated.
BREAKING CHANGE: No more changes should happen to Compilation.assets after sealing the Compilation.
	Do changes to assets earlier, e. g. in Compilation.hooks.processAssets.
	Make sure to select an appropriate stage from Compilation.PROCESS_ASSETS_STAGE_*.
```

Updating html-webpack-plugin to `"html-webpack-plugin": "^5.5.1",` in removes this warning package.json

mjamescompton (approved):
lgtm

I have tested with node 14, 16, 18, 20

As I thought updating some of these packages has sorted out all the node 18 and 20 errors we where getting. 

We could at this point choose to update our supported version of node, or at least try the end 2 end tests using the different versions

kschiffer:
image
Github uses node16 by default, so since the maintenance will phase out soon, we should switch that to node 18.
I'll update that and also change DEVELOPMENT.md to state node 18 as preferred version.

kschiffer:
The node version in our GitHub Actions is controlled by an input. @adriansmares can you set this to 18 after we merge this?

adriansmares:
> The node version in our GitHub Actions is controlled by an input. @adriansmares can you set this to 18 after we merge this?

The input is used only if a value is provided, otherwise the default is used. We don't override the version, so most likely you can just update the default version in this PR and check if anything breaks.

kschiffer:
Alright. I just saw that the input [was marked as `required`](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.26/.github/actions/install-node-and-deps/action.yml#L6), so that got me confused.

adriansmares (approved):

PR #6213, opened 2023-05-12
Use DNS resolver for dial targets

adriansmares:
#### Summary

This PR makes all of the external gRPC calls use the `dns` scheme, as opposed to `passthrough`.

#### Changes

- Use the `dns` scheme by default.
- Use `passthrough` explicitly for the loopback connection.

#### Testing

Local.

##### Regressions

N/A. By default the load balancing policy for DNS is `pick_first`, which is equivalent to `passthrough`.

#### Checklist

### DIFF
--- a/pkg/rpcclient/rpcclient.go
+++ b/pkg/rpcclient/rpcclient.go
@@ -37,6 +37,7 @@ import (
 	"go.thethings.network/lorawan-stack/v3/pkg/version"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/keepalive"
+	"google.golang.org/grpc/resolver"
 )
 
 // DefaultDialOptions for gRPC clients
@@ -77,4 +78,8 @@ func DefaultDialOptions(ctx context.Context) []grpc.DialOption {
 			PermitWithoutStream: false,
 		}),
 	}
+}
+
+func init() {
+	resolver.SetDefaultScheme("dns")
 }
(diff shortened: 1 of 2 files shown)


johanstokking:
Did you consider [`SetDefaultScheme`](https://pkg.go.dev/google.golang.org/grpc/resolver#SetDefaultScheme)?

The gRPC client won't use `dns` as default because it would change behavior for some library users. In our case, it won't change anything material. We would just set this in the `rpcclient` package `init()`.

adriansmares:
> The gRPC client won't use `dns` as default because it would change behavior for some library users. In our case, it won't change anything material. We would just set this in the `rpcclient` package `init()`.

I did actually. I don't fully understand why they are so cautious (https://github.com/grpc/grpc-go/issues/3544, https://github.com/grpc/grpc-go/issues/2168) about making `dns` the default - to me `dns` acts like `passthrough` in all situations, unless the target has the `TXT` records set, which may add service configuration that changes the load balancing method.

If you think there are no issues, I am ok with overriding the default in an `init`.

johanstokking:
I think it's because of https://github.com/grpc/grpc-go/issues/2168#issuecomment-399491187 which makes sense. It doesn't apply to us. We can use `dns` always.

adriansmares:
> I think it's because of [grpc/grpc-go#2168 (comment)](https://github.com/grpc/grpc-go/issues/2168#issuecomment-399491187) which makes sense. It doesn't apply to us. We can use `dns` always.

The loopback connection depends on `passthrough`, because it uses a custom dialer :joy: . I'll try to fix this.

johanstokking (approved):
Right, that's what a library change of behavior would have done for us.

Great changeset this way.

PR #6220, opened 2023-05-17
Fix LBS PONG payload

adriansmares:
#### Summary

This PR fixes our `PING`/`PONG` response behavior such that `PONG`s contain the application payload of the associated `PING`s. 

This behavior is mandated by the [WebSockets specification](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.3), but the reference LoRa Basics Station implementation [does not](https://github.com/lorabasics/basicstation/blob/ba4f85d80a438a5c2b659e568cd2d0f0de08e5a7/src/net.c#LL595C1-L595C12) verify the `PONG` payload.

#### Changes

- Reply to `PING`s with `PONG`s which contain the same application payload.

#### Testing

Tested on `staging1` with the reference implementation, with both versions 2.0.4 and 2.0.6, from multiple vendors.
Also tested with non-reference implementation such as Tektelic and Kerlink. This fix allows Kerlink gateways to `PING`/`PONG` with the server properly.

##### Regressions

As this was tested with the reference implementations and the alternative ones, we do not expected functional regressions. 

There are slightly more bytes (I saw that some implementations choose to send 4 bytes as `PING` payload) being sent around, but at a low interval. If the extra bandwidth is problematic for the client, they can always turn off the client side `PING`/`PONG`.

#### Notes for Reviewers

Live on `staging1`.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/ws/ws.go
+++ b/pkg/gatewayserver/io/ws/ws.go
@@ -170,7 +170,7 @@ func (s *srv) handleTraffic(w http.ResponseWriter, r *http.Request) (err error) 
 		ctx          = r.Context()
 		eps          = s.formatter.Endpoints()
 		missedPongs  = noPongReceived
-		pongCh       = make(chan struct{}, 1)
+		pongCh       = make(chan []byte, 1)
 		downstreamCh = make(chan []byte, 1)
 	)
 
@@ -291,7 +291,7 @@ func (s *srv) handleTraffic(w http.ResponseWriter, r *http.Request) (err error) 
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
-		case pongCh <- struct{}{}:
+		case pongCh <- []byte(data):
 		}
 		return nil
 	})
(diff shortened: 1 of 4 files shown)


pkg/gatewayserver/io/ws/ws.go
```diff
@@ -170,7 +170,7 @@ func (s *srv) handleTraffic(w http.ResponseWriter, r *http.Request) (err error)
 		ctx          = r.Context()
 		eps          = s.formatter.Endpoints()
 		missedPongs  = noPongReceived
-		pongCh       = make(chan struct{}, 1)
+		pongCh       = make(chan string, 1)
```
KrishnaIyer: Nit: We can use a byte channel instead and cast the data at the sender.

KrishnaIyer (approved):

pkg/gatewayserver/io/ws/ws.go
adriansmares: I don't fully see the gain here - someone has to cast, and both casts have to copy the data (because you don't want the byte slice to edit your original string, and you don't want your slice edits to be visible in your new string). What do you have in mind ?

Issue #6218, opened 2023-05-14
Investigate Gateway Server long schedule anytime delays


adriansmares:
### Summary

We are sometimes experiencing very long transmission delays when class C downlinks are scheduled without an absolute time. We should investigate why this happens, as generally the numbers are absurd (days, to months and sometimes even years), while the maximum delay should be a little over an hour (that's the duty cycle limit).

### Steps to Reproduce

Unknown at this time.

### Current Result

Delays of days to months.

### Expected Result

Delays of at most one hour, and even those should be extremely rare. 1 hour delay means that the GS decided to schedule in the next possible window.

### Relevant Logs

```shell
N/A.
```

### Deployment

The Things Stack Cloud

### The Things Stack Version

3.25.1

### Contributing

- [X] I can help by doing more research.
- [X] I can help by implementing a fix after the proposal above is approved.
- [ ] I can help by testing the fix before it's released.

### Code of Conduct

- [X] I agree to follow TTN's [Community Code of Conduct](https://github.com/TheThingsNetwork/Manifesto/blob/master/CodeOfConduct.md).

johanstokking:
If the culprit is indeed that GS's `ScheduleAnytime()` schedules downlink far in the future, then the cause could still be an already scheduled downlink that GS is scheduling after. Like indicated offline, https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.25/pkg/gatewayserver/scheduling/scheduler.go#L438-L442.

The logs in #6221 will flag the downlink scheduled far in the future, but if the reason is that there's already one far in the future, we really would like to see that earlier downlink message logged.

Roaming could be another cause. Absolute time class C scheduling is not possible via roaming, but normal class C scheduling is.

Few things:

1. Is RX1 delay bounds checked? Packet Broker doesn't do it. On quick glance, PBA doesn't do it either (https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.25/pkg/packetbrokeragent/translation.go#L712-L713) nor is `rx1_delay` validated on maximum value (https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.25/api/lorawan.proto#L508-L509). This means that an RX1 delay of 1 year may be possible. I'm not sure, but I don't see why it isn't
2. Lots of class C "anytime" downlinks that cause multiple duty-cycle windows in the future to fill up? If our NS doesn't do it, that's ok, but roaming doesn't forbid this
[...]

PR #6221, opened 2023-05-17
Log long scheduling delay in the Gateway Server

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/6218

#### Changes

- Fix the log fields for the join EUI and device EUI in the Network Server join request handling.
- Remove unused `rxTime` field from Basic Station frontend.
- Log scheduling delays above the scheduling window size.
- Ensure that round trip times are positive in LoRa Basics Station.

#### Testing

Unit tests.

##### Regressions

This PR attempts to find out which gateways are affected by this behavior, and in which situations the delay is over calculated. No semantical changes are introduced.

#### Notes for Reviewers

@johanstokking can you take a look at `Scheduler.ScheduleAnytime` and `SubBand.ScheduleAnyTime` and try to think of a which through which the scheduling delay (`res.Starts() - now`) would have absurd values (weeks to months) ?

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/io.go
+++ b/pkg/gatewayserver/io/io.go
@@ -635,13 +635,16 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
 		default:
 		}
 		var f func(context.Context, scheduling.Options) (scheduling.Emission, scheduling.ConcentratorTime, error)
+		var maxDelay time.Duration
 		switch request.Class {
 		case ttnpb.Class_CLASS_A:
 			f = c.scheduler.ScheduleAt
 			if request.Rx1Delay == ttnpb.RxDelay_RX_DELAY_0 {
 				return false, false, 0, errNoRxDelay.New()
 			}
-			settings.Timestamp = uplinkToken.Timestamp + uint32((time.Duration(request.Rx1Delay)*time.Second+rx.delay)/time.Microsecond)
+			rxDelay := time.Duration(request.Rx1Delay)*time.Second + rx.delay
+			settings.Timestamp = uplinkToken.Timestamp + uint32(rxDelay/time.Microsecond)
+			maxDelay = rx.delay
 		case ttnpb.Class_CLASS_B:
 			if request.AbsoluteTime == nil {
 				return false, false, 0, errNoAbsoluteTime.New()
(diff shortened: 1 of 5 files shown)


KrishnaIyer (approved):

pkg/gatewayserver/io/io.go
```diff
@@ -688,13 +694,19 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
...
 			"now", now,
 			"delay", delay,
-		)).Debug("Scheduled downlink")
+			"max_delay", maxDelay,
+		))
+		if delay >= maxDelay {
+			delay = maxDelay
+			logger.Error("Scheduling delay above maximum delay")
+		}
+		logger.Debug("Scheduled downlink")
```
johanstokking: So it is an error that we log and that's it?

What are the side effects of setting `delay` to `maxDelay`, i.e. what does it change to NS?

adriansmares: The error would reach Sentry, and we can monitor it there. This is at least a start to try to find a pattern of these large delays.

The NS uses the delay in order to estimate when the transmission takes place, in order to avoid scheduling everything at once. Extremely large delay values basically stop downlinks completely, as the NS waits for the downlink to be transmitted before transmitting a new one (and if delay was 1 year, there will be no downlinks for a year).

If we limit the delay, at least we can stop the MAC state from being 'tainted', until we can find out more information about what causes these delays to be so large.

pkg/gatewayserver/io/io.go
```diff
@@ -652,12 +655,15 @@ func (c *Connection) ScheduleDown(path *ttnpb.DownlinkPath, msg *ttnpb.DownlinkM
...
 			settings.Time = request.AbsoluteTime
+			maxDelay = time.Until(request.AbsoluteTime.AsTime())
 		case ttnpb.Class_CLASS_C:
 			if request.AbsoluteTime != nil {
 				f = c.scheduler.ScheduleAt
 				settings.Time = request.AbsoluteTime
+				maxDelay = time.Until(request.AbsoluteTime.AsTime())
 			} else {
 				f = c.scheduler.ScheduleAnytime
+				maxDelay = scheduling.DutyCycleWindow + 5*time.Second
```
johanstokking: Lots of "anytime" schedules can fill up multiple duty-cycle windows.

That NS doesn't schedule these is great, but class C downlink can also be scheduled via roaming.

adriansmares: ~Are you sure that you can schedule class C downlinks via roaming ? There is no support for that in either Packet Broker, nor BI to the best of knowledge - you cannot schedule something without an uplink token, and these do not exist in class C.~

Regarding filling multiple duty cycle windows - I understand the argument that this will limit the delay to 'only' one hour, but how else could we track information about this issue ? We have no reproduction steps, and we only see effects. At least with this form of primitive logging we can try to settle a set of affected gateways and limit the damage that the invalid delay could do.

johanstokking: My point is that this is not just logging but actually setting `delay` and potentially changing behavior.

So let's make this PR focus on logging only, and then have another PR for letting GS return errors when the scheduling would happen too far in the future.

johanstokking (approved):

PR #6281, opened 2023-05-25
Remove cancellable promise

kschiffer:
#### Summary
This PR removes the `CancellablePromise` from the `useRequest`-hook. Since an update, the functionality was changed which caused crashes in the Console. Canceling promises is a topic on its own and it is generally recommended to not use promise cancellation, so I decided to remove this functionality entirely. In some edge cases, this could mean that promises resolve after they've become obsolete. This is only a performance issue though and should not noticeably affect the UX.

#### Changes
- Remove promise cancellation and `CancellablePromise`

#### Testing

Manual and e2e

##### Regressions

None

#### Notes for Reviewers
- This caused a crash when navigating away from any page that fetched data using `useRequest`.
- I did not remove the `CancellablePromise` from `package.json` because I'll do that as part of a larger depencency update PR that I'm currently working on.
- Changelog entry is not necessary because this regression is within v3.26.0.

#### Checklist

### DIFF
--- a/pkg/webui/lib/hooks/use-request.js
+++ b/pkg/webui/lib/hooks/use-request.js
@@ -14,7 +14,6 @@
 
 import { useState, useEffect } from 'react'
 import { useDispatch } from 'react-redux'
-import CancelablePromise from 'cancelable-promise'
 
 import attachPromise from '@ttn-lw/lib/store/actions/attach-promise'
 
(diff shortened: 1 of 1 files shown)


ryaplots (approved):

mjamescompton (approved):
Tested locally on buttons I shared, looks good

Issue #5587, opened 2022-07-11
Refactor Identity Server Store interfaces


htdvisser:
### Summary

It would be good to make some changes to some Identity Server Store interfaces to make methods more "single-purpose".

### Current Situation

- The `SetMember` method of the `MembershipStore` interface deletes a collaborator when passing zero rights.
- The `UpdateAPIKey` method of the `APIKeyStore` interface deletes an API key when passing zero rights.
- The `Validate` method of the `ContactInfoStore` does multiple things that should be "business logic" instead of "store logic":
	- Getting a `ContactInfoValidation`
	- Checking for already used / expired
	- Marking `ContactInfo` as "validated"
	- Validating the primary email address of a user
	- Marking the `ContactInfoValidation` as "used"
- Similarly for the `SetInvitationAcceptedBy` method of the `InvitationStore`

### Desired Situation

- The `MembershipStore` interface should have separate methods for updating and deleting collaborators.
- The `APIKeyStore` interface should have separate methods for updating and deleting API keys.
- The `ContactInfoStore` should have separate methods for each of the tasks described above. Other business logic belongs in the RPC implementation.
- Similar story for the `InvitationStore`.

### Code of Conduct

- [X] I agree to follow TTN's [Community Code of Conduct](https://github.com/TheThingsNetwork/Manifesto/blob/master/CodeOfConduct.md).

htdvisser:
Blocked on #3250, so for triage let's just assign someone, and maybe schedule it for Q4 or so.

cc: @NicolasMrad

adriansmares:
What is the status here ?

I saw that we have [a branch](https://github.com/TheThingsNetwork/lorawan-stack/tree/issue/5587-refactor-is-interfaces) already, and I am wondering how much is left in order to schedule this for review and merging.

nicholaspcr:
It was supposed to be finished on the Thursday two weeks ago but didn't managed to do it. 
I'm pretty confident in finishing this Thursday.

nicholaspcr:
This was actually finished on `v3.26.1` the related RPC implementation is the continuation based on the addition of the delete methods in the interfaces.

Therefore I'll be closing this issue.

PR #6310, opened 2023-06-05
Refactor Invitation store interface

nicholaspcr:
#### Summary

References #5587 

#### Changes

- Move `SetInvitationAcceptedBy` checks to `user_registry`.
- Add tests from store to user_registry

#### Testing

Unit tests

#### Checklist

### DIFF
--- a/pkg/identityserver/bunstore/invitation_store.go
+++ b/pkg/identityserver/bunstore/invitation_store.go
@@ -179,7 +179,6 @@ func (s *invitationStore) GetInvitation(ctx context.Context, token string) (*ttn
 		attribute.String("invitation_token", token),
 	))
 	defer span.End()
-
 	model, err := s.getInvitationModelBy(ctx, func(q *bun.SelectQuery) *bun.SelectQuery {
 		return q.
 			Where("token = ?", token).
@@ -189,27 +188,19 @@ func (s *invitationStore) GetInvitation(ctx context.Context, token string) (*ttn
 	})
 	if err != nil {
 		if errors.IsNotFound(err) {
-			return nil, store.ErrInvitationNotFound.WithAttributes(
-				"invitation_token", token,
-			)
+			return nil, store.ErrInvitationNotFound.WithAttributes("invitation_token", token)
 		}
 		return nil, err
 	}
-
-	pb, err := invitationToPB(model)
-	if err != nil {
-		return nil, err
-	}
-
-	return pb, nil
+	return invitationToPB(model)
 }
 
 func (s *invitationStore) SetInvitationAcceptedBy(
-	ctx context.Context, token string, accepter *ttnpb.UserIdentifiers,
+	ctx context.Context, token string, usrIDs *ttnpb.UserIdentifiers,
 ) error {
 	ctx, span := tracer.StartFromContext(ctx, "SetInvitationAcceptedBy", trace.WithAttributes(
 		attribute.String("invitation_token", token),
-		attribute.String("user_id", accepter.GetUserId()),
+		attribute.String("user_id", usrIDs.GetUserId()),
 	))
 	defer span.End()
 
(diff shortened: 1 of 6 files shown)


pkg/identityserver/bunstore/invitation_store.go
```diff
@@ -195,58 +194,31 @@ func (s *invitationStore) GetInvitation(ctx context.Context, token string) (*ttn
...
-		return nil, err
-	}
-
-	return pb, nil
+	return invitationToPB(model)
 }
 
 func (s *invitationStore) SetInvitationAcceptedBy(
-	ctx context.Context, token string, accepter *ttnpb.UserIdentifiers,
+	ctx context.Context, validation *ttnpb.Invitation, usrIDs *ttnpb.UserIdentifiers,
```
adriansmares: I don't fully understand what we gain by passing the full `*ttnpb.Invitation` here - we use only the token.

pkg/identityserver/user_registry.go
```diff
@@ -248,16 +247,6 @@ func (is *IdentityServer) createUser(ctx context.Context, req *ttnpb.CreateUserR
 	defer func() { is.setFullProfilePictureURL(ctx, usr) }()
 
 	err = is.store.Transact(ctx, func(ctx context.Context, st store.Store) (err error) {
-		if req.InvitationToken != "" {
```
adriansmares, line 251: Why shouldn't we keep the ordering here ? You are creating the user even if the invitation is expired. 

I know that the transaction will fail but I don't see why it shouldn't fail early.

pkg/identityserver/user_registry_test.go
```diff
@@ -122,13 +123,72 @@ func TestUserCreate(t *testing.T) {
...
 
-		is.config.UserRegistration.Enabled = true
-		is.config.UserRegistration.Invitation.Required = true
-
-		_, err = reg.Create(ctx, req)
-		if a.So(err, should.NotBeNil) {
-			a.So(errors.Resemble(err, errInvitationTokenRequired), should.BeTrue)
-		}
+		t.Run("InvitationRequired", func(t *testing.T) {
+			is.config.UserRegistration.Enabled = true
```
adriansmares: I find it wrong that you are moving the config update inside the test - it will affect future tests too, so it should stay outside - it makes it look like this is affecting only this test, while it does not.

pkg/identityserver/user_registry_test.go
```diff
@@ -54,8 +55,8 @@ func TestValidatePasswordStrength(t *testing.T) {
 			"password1":    false, // Too common.
 			"āaaAAA123!@#": true,
 			"       1A":    true,
-			"āaa	AAA123 ":  true,
-			"āaaAAA123 ":   true,
+			"āaa	AAA123 ": true,
```
adriansmares: Do not change this.

adriansmares (changes requested):

pkg/identityserver/user_registry.go
nicholaspcr: I took the transaction for granted and decided to eliminate the duplicate check in the beginning since the `SetInvitationAcceptedBy` used the `usr` response from the `CreateUser`.

Changed so that the validation happens before the user creation.

pkg/identityserver/bunstore/invitation_store.go
nicholaspcr: Agree, it was an unnecessary change. Revert to the original `token string` paramater.

pkg/identityserver/bunstore/invitation_store.go
```diff
@@ -195,58 +194,31 @@ func (s *invitationStore) GetInvitation(ctx context.Context, token string) (*ttn
...
-	if model.ExpiresAt != nil && model.ExpiresAt.Before(s.now()) {
-		return store.ErrInvitationExpired.WithAttributes("invitation_token", token)
-	}
-
-	if model.AcceptedByID != nil {
-		return store.ErrInvitationAlreadyUsed.WithAttributes("invitation_token", token)
-	}
-
-	_, userUUID, err := s.getEntity(ctx, accepter)
+	_, userUUID, err := s.getEntity(ctx, usrIDs)
```
adriansmares, line 216: The advantage of `getInvitationModelBy` is that it maintains tenant barriers based on the context. Our current update statement does not, and I am not sure if tokens are guaranteed to be globally unique.

I think we should keep the old logic which retrieved the entity, even if we take some of the business logic out of it.

nicholaspcr: Checked just in case and you are right, token aren't globally unique, they are separated by tenant barriers. 
Changed to use the model returned from the `getInvitationModelBy`

adriansmares: You are skipping the error handling for `getInvitationModelBy`.

nicholaspcr: true, my bad. Added the check and a test which checks that.

pkg/identityserver/user_registry_test.go
```diff
@@ -125,10 +126,70 @@ func TestUserCreate(t *testing.T) {
 		is.config.UserRegistration.Enabled = true
 		is.config.UserRegistration.Invitation.Required = true
 
-		_, err = reg.Create(ctx, req)
-		if a.So(err, should.NotBeNil) {
-			a.So(errors.Resemble(err, errInvitationTokenRequired), should.BeTrue)
-		}
+		t.Run("InvitationRequired", func(t *testing.T) { //nolint:paralleltest
```
adriansmares: The `nolint` should have a space after `//`.

adriansmares (approved):

PR #6316, opened 2023-06-08
Fix - tenant barrier error on contact_info refactor

nicholaspcr:
#### Summary

References #5587 

When fetching entities the tenant constraint is always applied via the `newSelectModel` method, this PR fixes the lack of said method when fetching an `user` or an `ContactInfo` entity. 

#### Changes

- Usage of `newSelectModel` when performing updates
- Added test cases to confirm that operations were done in the correct manner.
- Added an error in case the provided `ContactInfo` isn't found.
  - This is mostly a redundant check since `GetValidation` fills the contact info with the information in the `ContactInfoValidation` fetched from the DB.
  - Nevertheless this is just to have a check instead of leaving the responsibility exclusively to the `GetValidation` method 

#### Testing

Added more cases to the unit test

#### Notes for Reviewers

Looked at the other PRs for something similar but couldn't find anything similar to this error on them.

#### Checklist

### DIFF
--- a/pkg/identityserver/bunstore/contact_info_store.go
+++ b/pkg/identityserver/bunstore/contact_info_store.go
@@ -286,21 +286,45 @@ func (s *contactInfoStore) ValidateContactInfo(ctx context.Context, pb *ttnpb.Co
 	if len(pb.GetContactInfo()) != 1 {
 		return store.ErrValidationWithoutContactInfo.New()
 	}
-	contactInfo := pb.GetContactInfo()[0]
-	_, err := s.DB.NewUpdate().
-		Model(&ContactInfo{}).
-		Where("entity_type = ? AND entity_id = ?", pb.GetEntity().EntityType(), pb.GetEntity().IDString()).
-		Where("contact_method = ? AND LOWER(value) = LOWER(?)", contactInfo.GetContactMethod(), contactInfo.GetValue()).
+	contInfo := pb.GetContactInfo()[0]
+
+	entityType, entityUUID, err := s.getEntity(ctx, pb.GetEntity())
+	if err != nil {
+		return err
+	}
+
+	models, err := s.getContactInfoModelsBy(ctx, func(q *bun.SelectQuery) *bun.SelectQuery {
+		return q.
+			Where("entity_type = ? AND entity_id = ?", entityType, entityUUID).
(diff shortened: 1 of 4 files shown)


pkg/identityserver/bunstore/contact_info_store.go
```diff
@@ -386,6 +410,20 @@ func validationToPB(m *ContactInfoValidation) *ttnpb.ContactInfoValidation {
...
+	by func(*bun.SelectQuery) *bun.SelectQuery,
+) (*ContactInfoValidation, error) {
+	model := &ContactInfoValidation{}
+	selectQuery := newSelectModel(ctx, s.DB, model).Apply(by)
+	err := selectQuery.Scan(ctx)
+	if err != nil {
+		return nil, storeutil.WrapDriverError(err)
+	}
+	return model, nil
+
```
adriansmares: Remove new line.

pkg/identityserver/bunstore/contact_info_store.go
```diff
@@ -286,21 +286,45 @@ func (s *contactInfoStore) ValidateContactInfo(ctx context.Context, pb *ttnpb.Co
...
+	models, err := s.getContactInfoModelsBy(ctx, func(q *bun.SelectQuery) *bun.SelectQuery {
+		return q.
+			Where("entity_type = ? AND entity_id = ?", entityType, entityUUID).
+			Where("contact_method = ? AND LOWER(value) = LOWER(?)", contInfo.GetContactMethod(), contInfo.GetValue())
+	})
+	if err != nil {
+		return err
+	}
+	if len(models) != 1 {
```
adriansmares, line 304: Do we have any index that guarantees this ?

nicholaspcr: For the contact infos I don't think we do, speaking in a general term. 

But with the current way the method itself is used its guaranteed to be only one since its from a `ContactInfoValidation` which is fetched by providing `reference` and `token`

adriansmares: I don't see what stops us from having two pairs of `contact_method, lower(value)` with the same values (under the same entity). Then `len(models)` will be 2 and validation will continuously fail.

nicholaspcr: Plus the login in the `ContactInfo` creation which stops having duplicate elements.

So there isn't an index that guarantees this but store logic pretty much guarantees it.

adriansmares (approved):

PR #6317, opened 2023-06-08
Remove AS uplink storage

### LINKED ISSUE #6215: Remove Application Server uplink storage
### Summary

_NOTE: This is unrelated to the storage integration._

The Application Server currently has an uplink storage functionality which allows it to store the metadata of recent uplinks. The metadata is stored in order to allow multi frame queries in the LoRaCloud GLS application package, when multi frame queries are used.

### Current Situation

The GLS application package is rarely used, but we do store the metadata for every single end device.

We would benefit a lot if we could remove this 'always-on' storage, as it uses quite a chunk of space.

### Proposed Implementation

1. Deprecate the uplink storage registry:
  - [ ] Mark the configuration options as `DEPRECATED`.
  - [ ] Remove the `ApplicationUplinkRegistry` and its implementations and usages.
  - [ ] Remove the `UplinkStorage` interface in `io`.
  - [ ] Write a migration for `as-db` which removes the associated keys from the removed store.
2. Refactor the multi-frame queries in the GLS integration.
  - [ ] Store the metadata that used to be stored by the uplink storage mechanism inside the application package association.
3. Add `Clear` methods to the application registry store.
  - [ ] One that clears all of the application package associations for a specific end device.
     - Watch the UID key.
     - Retrieve all of the members of the set found at the UID key, build their keys and watch them.
     - Delete the keys in a `TxPipelined`.
  - [ ] One that clears all of the application package default associations for a specific application. Same instructions.
  - [ ] Use these methods for cleanup when end devices are deleted, and when an application link is deleted.

### Contributing

### Code of Conduct

cvetkovski98:
#### Summary
The AS Uplink Storage is used only for the LoRa Geolocation Services integration. This PR moves its functionality to the GLS association data, adds appropriate methods to the association registry for cleaning association when a device or application link is removed and an appropriate migration for the removed functionality.

This PR also removes the automatic determination of the window size based on the frame payload of the GLS integration and limits the frame size in the `[1, 16]` range. This can be set through the Console.

Closes #6215 

#### Changes
- Deprecate and remove the uplink storage registry
- Refactor the multi-frame queries in the GLS integration to use the association data
- Add `Clear` methods to the application registry store.

#### Testing
- UT
- Manual

##### Regressions
...

#### Notes for Reviewers
...

#### Checklist


cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -227,6 +227,41 @@ var (
...
+			cl := ttnredis.New(config.Redis.WithNamespace("as", "applicationups"))
+
+			logger.Info("Cleaning uplink registry data")
+
+			numCleared := 0
+			uidApplicationUpsKey := cl.Key("uid", "*")
+			pipeliner := cl.Pipeline()
+			err := ttnredis.RangeRedisKeys(ctx, cl, uidApplicationUpsKey, ttnredis.DefaultRangeCount,
+				func(k string) (bool, error) {
+					if err := pipeliner.Del(ctx, k).Err(); err != nil {
```
adriansmares: A pipeline is executed only when you call `Exec` - individual calls to the pipeline cannot fail.

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -227,6 +227,41 @@ var (
...
+						numCleared++
+						return false, err
+					}
+					return true, nil
+				})
+			if err != nil {
+				logger.Errorf("Failed to clean uplink registry data: %v", err)
```
adriansmares: Use `logger.WithError`.

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -227,6 +227,41 @@ var (
...
+				return err
+			}
+			if _, err := pipeliner.Exec(ctx); err != nil {
+				logger.Errorf("Failed to clean uplink registry data: %v", err)
```
adriansmares: Use `logger.WithError`.

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -227,6 +227,41 @@ var (
...
+				return err
+			}
+
+			logger.Infof("Cleaned %d records", numCleared)
```
adriansmares: Use `logger.WithField`.

pkg/applicationserver/applicationserver.go
```diff
@@ -1434,7 +1397,8 @@ func (as *ApplicationServer) GetMQTTConfig(ctx context.Context) (*config.MQTT, e
 
 // RangeUplinks ranges the application uplinks and calls the callback function, until false is returned.
 func (as *ApplicationServer) RangeUplinks(ctx context.Context, ids *ttnpb.EndDeviceIdentifiers, paths []string, f func(ctx context.Context, up *ttnpb.ApplicationUplink) bool) error {
-	return as.appUpsRegistry.Range(ctx, ids, paths, f)
+	// TODO: Implement this.
```
adriansmares: This can be dropped - no other upstreams outside of the GLS integration depend on it.

pkg/applicationserver/config.go
```diff
@@ -98,7 +98,7 @@ type Config struct {
 	LinkMode                 string                         `name:"link-mode" description:"Deprecated - mode to link applications to their Network Server (all, explicit)"`
 	Devices                  DeviceRegistry                 `name:"-"`
 	Links                    LinkRegistry                   `name:"-"`
-	UplinkStorage            UplinkStorageConfig            `name:"uplink-storage" description:"Application uplinks storage configuration"`
+	UplinkStorage            UplinkStorageConfig            `name:"uplink-storage" description:"Deprecated - Application uplinks storage configuration"`
```
adriansmares: Use `DEPRECATED` as the description - no need to keep the old description.

pkg/applicationserver/cleanup.go
```diff
@@ -57,9 +56,7 @@ func (cleaner *RegistryCleaner) DeleteDeviceData(ctx context.Context, devSet []s
 		if err != nil {
 			return err
 		}
-		if err := cleaner.AppUpsRegistry.Clear(ctx, devIds); err != nil {
-			return err
-		}
+		// TODO (Uplink Storage Removal): Implement Clear API.
```
adriansmares: I think that the application packages association cleaner does this already.

pkg/applicationserver/io/packages/loragls/v3/data.go
```diff
@@ -80,6 +81,12 @@ const (
 	QUERY_TOAWIFI
 )
 
+// UplinkMD contains the uplink metadata stored by the package.
+type UplinkMD struct {
```
adriansmares: Nit: `UplinkMetadata`.

pkg/applicationserver/io/packages/loragls/v3/data.go
```diff
@@ -80,6 +81,12 @@ const (
...
+	RxMetadata []*ttnpb.RxMetadata
```
adriansmares: I would use a local definition of `RxMetadata`, with only the fields that we need. Otherwise it is hard to keep up with what is needed and what is not needed.

pkg/applicationserver/io/packages/loragls/v3/data.go
```diff
@@ -96,6 +103,8 @@ type Data struct {
 	ServerURL *url.URL
 	// Token is the API token to be used when comunicating with the GLS server.
 	Token string
+	// UplinkMDs are the metadatas from the recent uplink messages received by the gateway.
+	UplinkMDs []*UplinkMD
```
adriansmares: Nit: `RecentMetadata`.

pkg/applicationserver/io/packages/loragls/v3/data.go
```diff
@@ -105,6 +114,7 @@ const (
 	multiFrameWindowAge  = "multi_frame_window_age"
 	serverURLField       = "server_url"
 	tokenField           = "token"
+	uplinksField         = "uplinks"
```
adriansmares: Nit: `recent_metadata`.

pkg/applicationserver/io/packages/loragls/v3/data.go
```diff
@@ -131,6 +141,95 @@ func toFloat64(f float64) *structpb.Value {
 	}
 }
 
+func toGatewayIDsStruct(gatewayIDs *ttnpb.GatewayIdentifiers) *structpb.Value {
```
adriansmares: I think we should cheat a little and use something like [gob](https://pkg.go.dev/encoding/gob). We can use our local struct definitions here as `gob` messages, and only store the binary as base64 in `data`.

The reason is that external users should not interact with the recent metadata (they don't insert it anyway), and having the long field names is inefficient space wise.

This has the advantage that we don't need to write encoders and decoders by hand - we just marshal to base64 and back.

pkg/applicationserver/io/packages/redis/registry.go
```diff
@@ -585,3 +585,44 @@ func (r ApplicationPackagesRegistry) Range(
 		return true, nil
 	})
 }
+
+func (r ApplicationPackagesRegistry) clearAssociations(ctx context.Context, id ttnpb.IDStringer) error {
+	uid := unique.ID(ctx, id)
+	uidKey := r.uidKey(uid)
+
+	return r.Redis.Watch(ctx, func(tx *redis.Tx) error {
+		// retrieve the list of fPorts from the uidKey set
```
adriansmares: Nit: uppercase first letter, dot at the end.

pkg/applicationserver/io/packages/redis/registry.go
```diff
@@ -585,3 +585,44 @@ func (r ApplicationPackagesRegistry) Range(
...
+			return err
+		}
+		// build all the association keys
+		keys := make([]string, 0, len(fPorts))
+		for _, fPort := range fPorts {
+			keys = append(keys, r.associationKey(uid, fPort))
+		}
+		keys = append(keys, uidKey)
+
+		if _, err := tx.Pipelined(ctx, func(p redis.Pipeliner) error {
```
adriansmares: Use `TxPipelined`, otherwise `WATCH` has no effect.

adriansmares (commented):
This is going in the right direction and looks correct so far. My only change to the design is to optimize a bit storage, as saving the field names has no benefit since only the AS has to be able to encode and decode the data.

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -227,6 +227,39 @@ var (
 			return nil
 		},
 	}
+	asDBUplinkRegistryCleanupCommand = &cobra.Command{
+		Use:   "cleanup-uplink-registry",
+		Short: "Clean stale Application Server uplink registry data",
```
johanstokking: Don't we call this "purge" in other places?

pkg/applicationserver/applicationserver.go
```diff
@@ -61,6 +61,8 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+const defaultLockTTL = 10 * time.Second
```
johanstokking: Remove?

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -227,6 +227,39 @@ var (
 			return nil
 		},
 	}
+	asDBUplinkRegistryCleanupCommand = &cobra.Command{
+		Use:   "cleanup-uplink-registry",
```
adriansmares: I would use `purge` - that is what we used in `dcs-db`.

pkg/applicationserver/applicationserver.go
adriansmares: Remove.

pkg/applicationserver/config.go
```diff
@@ -98,7 +98,7 @@ type Config struct {
 	LinkMode                 string                         `name:"link-mode" description:"Deprecated - mode to link applications to their Network Server (all, explicit)"`
 	Devices                  DeviceRegistry                 `name:"-"`
 	Links                    LinkRegistry                   `name:"-"`
-	UplinkStorage            UplinkStorageConfig            `name:"uplink-storage" description:"Application uplinks storage configuration"`
+	UplinkStorage            UplinkStorageConfig            `name:"uplink-storage" description:"DEPRECATED"`
```
adriansmares: In retrospective, this should've been `"-"` since it is not rendered anyway - only leafs (so final values inside the structs) are. We should mark these as deprecated since that is what shows up in `--help`.

pkg/applicationserver/io/packages/loragls/v3/package.go
```diff
@@ -136,7 +141,55 @@ func minInt(a int, b int) int {
 	return b
 }
 
-func (p *GeolocationPackage) multiFrameQuery(ctx context.Context, ids *ttnpb.EndDeviceIdentifiers, up *ttnpb.ApplicationUplink, data *Data, client *api.Client) (api.AbstractLocationSolverResponse, error) {
+// pushUplink updates the package data with the given uplink.
+func (p *GeolocationPackage) pushUplink(
+	ctx context.Context,
+	pkgAssocIDs *ttnpb.ApplicationPackageAssociationIdentifiers,
```
adriansmares: Since there are no other `ids`, you can get away with `ids`.

pkg/applicationserver/io/packages/loragls/v3/package.go
```diff
@@ -136,7 +141,55 @@ func minInt(a int, b int) int {
...
+	up *ttnpb.ApplicationUplink,
+	data *Data,
+) error {
+	maxUplinkCount := minInt(data.MultiFrameWindowSize, 16)
+
+	setter := func(apa *ttnpb.ApplicationPackageAssociation) (*ttnpb.ApplicationPackageAssociation, []string, error) {
```
adriansmares: `assoc` or `association` please.

pkg/applicationserver/io/packages/loragls/v3/package.go
```diff
@@ -136,7 +141,55 @@ func minInt(a int, b int) int {
...
+		data := &Data{}
+		if err := data.FromStruct(apa.Data); err != nil {
+			return nil, nil, err
+		}
+
+		if maxUplinkCount != 0 && len(data.RecentMetadata) >= maxUplinkCount {
+			data.RecentMetadata = data.RecentMetadata[1:]
+		}
+
+		storedMD := &UplinkMetadata{}
```
adriansmares: Just `md` will do.

pkg/applicationserver/io/packages/loragls/v3/package.go
```diff
@@ -136,7 +141,55 @@ func minInt(a int, b int) int {
...
+		st, err := data.Struct()
+		if err != nil {
+			return nil, nil, err
+		}
+		apa.Data = st
+
+		return apa, []string{"data"}, nil
+	}
+
+	_, err := p.registry.SetAssociation(ctx, pkgAssocIDs, nil, setter)
```
adriansmares: Do you actually have a `data` if you don't provide it explicitly ?

pkg/applicationserver/io/packages/loragls/v3/package.go
```diff
@@ -136,7 +141,55 @@ func minInt(a int, b int) int {
...
+	return err
```
adriansmares, line 233: We should return the result of `SetAssociation` and use it in `multiFrameQuery`, because otherwise `data` will be the old one - without the last uplink.

pkg/applicationserver/io/packages/loragls/v3/structs.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// ToProto converts the Location to a protobuf representation.
+func (l Location) ToProto() *ttnpb.Location {
+	return &ttnpb.Location{
+		Latitude:  l.Latitude,
+		Longitude: l.Longitude,
+		Altitude:  l.Altitude,
+		Accuracy:  l.Accuracy,
+		Source:    ttnpb.LocationSource(l.Source),
```
adriansmares: Do we use the source in the query ?

pkg/applicationserver/io/packages/loragls/v3/structs.go
```diff
@@ -0,0 +1,144 @@
...
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// GatewayIDs contains the gateway identifiers stored by the package.
+type GatewayIDs struct {
+	GatewayID string `json:"gateway_id"`
+}
+
+// ToProto converts the GatewayIDs to a protobuf representation.
+func (g GatewayIDs) ToProto() *ttnpb.GatewayIdentifiers {
```
adriansmares: When do we turn back to proto ?

pkg/applicationserver/io/packages/loragls/v3/structs.go
```diff
@@ -0,0 +1,144 @@
...
+	GatewayIDs    *GatewayIDs `json:"gateway_ids"`
+	AntennaIndex  uint32      `json:"antenna_index"`
+	FineTimestamp uint64      `json:"fine_timestamp"`
+	Location      *Location   `json:"location"`
+	RSSI          float32     `json:"rssi"`
+	SNR           float32     `json:"snr"`
+}
+
+// ToProto converts the RxMetadata to a protobuf representation.
+func (r RxMetadata) ToProto() *ttnpb.RxMetadata {
```
adriansmares: Do we need this ?

pkg/applicationserver/io/packages/redis/registry.go
```diff
@@ -585,3 +603,44 @@ func (r ApplicationPackagesRegistry) Range(
 		return true, nil
 	})
 }
+
+func (r ApplicationPackagesRegistry) clearAssociations(ctx context.Context, id ttnpb.IDStringer) error {
+	uid := unique.ID(ctx, id)
+	uidKey := r.uidKey(uid)
+
+	return r.Redis.Watch(ctx, func(tx *redis.Tx) error {
```
adriansmares: I would use `LockedWatch`, for consistency.

pkg/applicationserver/io/packages/redis/registry_test.go
```diff
@@ -0,0 +1,158 @@
...
+					"package_name",
+					"data",
+				}, nil
+		},
+	)
+	if !a.So(err, should.BeNil) {
+		t.FailNow()
+	}
+	actual, err := registry.ListDefaultAssociations(ctx, appIDs, nil)
+	if !a.So(err, should.BeNil) {
```
adriansmares, line 79: I would also compare the returned element at 0.

pkg/applicationserver/io/packages/loragls/v3/structs.go
cvetkovski98: I wanted to keep as much of the core logic intact. The `BuildMultiFrameRequest` method uses `ttnpb.RxMetadata`. I believe that is the only reason, so I convert to a proto before that. Should I refactor this?

pkg/applicationserver/io/packages/loragls/v3/structs.go
cvetkovski98: Yes, when converting back to proto before building the multi frame request.

pkg/applicationserver/io/packages/loragls/v3/structs.go
cvetkovski98: It was part of the fields for the previous `"cleanUplink"`. From what I can see, we don't use it when we build a multi frame request. Should I remove it?

pkg/applicationserver/io/packages/loragls/v3/structs.go
adriansmares: I think we should move some of the structs defined here to the `api` package, next to `BuildMultiFrameRequest`.

I am saying this because right now we do `ttnpb` -> `v3` -> `ttnpb` -> `api` which has a redundant transformation to and from `ttnpb` at the end. I would prefer that we remove that, because in this way we can track better which fields we kept or did not keep while doing the initial transformation from `ttnpb`.

pkg/applicationserver/io/packages/loragls/v3/structs.go
adriansmares: Yes, please remove it.

pkg/applicationserver/config.go
```diff
@@ -98,7 +98,7 @@ type Config struct {
 	LinkMode                 string                         `name:"link-mode" description:"Deprecated - mode to link applications to their Network Server (all, explicit)"`
 	Devices                  DeviceRegistry                 `name:"-"`
 	Links                    LinkRegistry                   `name:"-"`
-	UplinkStorage            UplinkStorageConfig            `name:"uplink-storage" description:"Application uplinks storage configuration"`
+	UplinkStorage            UplinkStorageConfig            `name:"-" description:"DEPRECATED"`
```
adriansmares: The description is the one you can skip, not the name - the name is part of the path.

pkg/applicationserver/io/packages/loragls/v3/package.go
```diff
@@ -73,6 +84,17 @@ func (p *GeolocationPackage) HandleUp(ctx context.Context, def *ttnpb.Applicatio
 
 	switch m := up.Up.(type) {
 	case *ttnpb.ApplicationUp_UplinkMessage:
+		assoc, err := p.pushUplink(ctx, assoc.Ids, m.UplinkMessage, data)
+		if err != nil {
+			return err
+		}
+		if assoc != nil {
```
adriansmares: In which situations would it be `nil` ?

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -227,6 +227,39 @@ var (
 			return nil
 		},
 	}
+	asDBUplinkRegistryCleanupCommand = &cobra.Command{
+		Use:   "purge-uplink-registry",
```
adriansmares: This should be simply `purge`, and we can use it to delete other keys later on if needed. We should also keep the command description general.

adriansmares (commented):
We should add a changelog entry for these changes. I would say that the migration is the main point to put an emphasis on, and the deprecated configuration options.

CHANGELOG.md
```diff
@@ -10,10 +10,12 @@ For details about compatibility between different releases, see the **Commitment
 ## [Unreleased]
 
 ### Added
+- The `as-db purge` command can now be used to purge unused data from the Application Server database.
```
adriansmares: New line before changelog entries. 

Also generally the changelog lines are meant to be read from the header. So for example if you look at other added entries, it goes like '(Added) Support for x' and so forth.

CHANGELOG.md
```diff
@@ -10,10 +10,12 @@ For details about compatibility between different releases, see the **Commitment
...
 
 ### Changed
 
 ### Deprecated
+- With the removal of the Application Uplink Registry the `--as.uplink-storage.limit` flag for `ttn-lw-stack` is deprecated.
```
adriansmares: New line at before the header. 

Also in general we have deprecated the `as.uplink-storage.limit` configuration as a whole - this means that it goes away also from the YAML file and the environment variables. I would just mention that the `as.uplink-storage.limit` configuration option has been deprecated.

adriansmares (approved):
LGTM. @johanstokking could you also take a look ?

CHANGELOG.md
johanstokking: Do we advise a database migration on the next .0 release?

cmd/ttn-lw-stack/commands/as_db.go
```diff
@@ -227,6 +227,39 @@ var (
...
+	asDBPurgeCommand = &cobra.Command{
+		Use:   "purge",
+		Short: "Purge unused Application Server data",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+
+			logger.Info("Connecting to Redis database...")
+			cl := ttnredis.New(config.Redis.WithNamespace("as", "applicationups"))
+
+			logger.Info("Cleaning uplink registry data")
```
johanstokking: Inconsistent terminology clean vs purge, go with purge/purging, also below

johanstokking (commented):
So this is only unit tested; are you unable to do a local integration test by purging data as well as seeing if GLS still works?

cmd/ttn-lw-stack/commands/as_db.go
johanstokking: Old review comment, was already addressed

pkg/applicationserver/applicationserver.go
johanstokking: Idem

CHANGELOG.md
adriansmares: `purge` commands can be run anytime, and possibly never. They are optional in that regard.

cvetkovski98:
> So this is only unit tested; are you unable to do a local integration test by purging data as well as seeing if GLS still works?

Now it's tested and additional changes have been made.

pkg/applicationserver/io/packages/loragls/v3/package.go
```diff
@@ -87,6 +109,22 @@ func (p *GeolocationPackage) Package() *ttnpb.ApplicationPackage {
...
 }
 
+// assocIDs returns the identifiers of the given association. If the association is nil, new identifiers are created.
+func (p *GeolocationPackage) assocIDs(
+	assoc *ttnpb.ApplicationPackageAssociation,
+	ids *ttnpb.EndDeviceIdentifiers,
+) *ttnpb.ApplicationPackageAssociationIdentifiers {
+	assocIDs := assoc.GetIds()
+	if assocIDs == nil {
+		pkg := p.Package()
```
adriansmares: You should use the port of the default association, and then you no longer need this function to be tied to `*GeolocationPackage`.

Due to checks in the uplink handler, you will always have at least the association, or the default association, or both.

pkg/applicationserver/io/packages/loragls/v3/package.go
```diff
@@ -136,7 +178,60 @@ func minInt(a int, b int) int {
...
 
-func (p *GeolocationPackage) multiFrameQuery(ctx context.Context, ids *ttnpb.EndDeviceIdentifiers, up *ttnpb.ApplicationUplink, data *Data, client *api.Client) (api.AbstractLocationSolverResponse, error) {
+// pushUplink updates the package data with the given uplink.
+func (p *GeolocationPackage) pushUplink(
+	ctx context.Context,
+	ids *ttnpb.ApplicationPackageAssociationIdentifiers,
+	up *ttnpb.ApplicationUplink,
+	data *Data,
+) (*ttnpb.ApplicationPackageAssociation, error) {
+	maxUplinkCount := minInt(data.MultiFrameWindowSize, 16)
```
adriansmares: I think zero should mean that the default is used too. I don't think we should allow infinite storage of metadata.

cvetkovski98: Should I remove the `maxUplinkCount != 0` check?

adriansmares: I would say that a `maxUplinkCount` of zero after this clamping (`minInt`) should turn `maxUplinkCount` into 16.

adriansmares (approved):

adriansmares (approved):

johanstokking (approved):

kschiffer (approved):

PR #6325, opened 2023-06-14
Fetch device from IS for Unclaim and GetClaimStatus RPCs

KrishnaIyer:
#### Summary

References; https://github.com/TheThingsIndustries/product-management/issues/6

#### Changes

- Remove the request body from `Unclaim`. We are doing this since the support for a request body for HTTP `Delete` requests is unclear.
  -  Also, we have to get the device from the IS to check that the EUIs are correct anyway.
  - This is a break in API but only the console/CLI currently use it and @adriansmares and I feel that this is justified.
- Get the EUIs for devices from the IS instead of using the request values for `Unclaim` and `GetClaimStatus`.
- Update tests
- Update console.

#### Testing

DCS locally tested. For the console part, I think if the Cypress tests pass, I think this is good.

##### Regressions

Will be tested on `staging1` since there's currently no support for local TTJS testing.

#### Notes for Reviewers

⚠️ We are (mildly) breaking API here. The request URI for `unclaim` always needed to include the App and Device ID. It's just that we no longer accept a body.

#### Checklist


sdk/js/src/service/claim.js
```diff
@@ -46,24 +46,25 @@ class DeviceClaim {
...
       routeParams: {
         'application_ids.application_id': applicationId,
         device_id: deviceId,
       },
     }
 
-    const response = await this._api.EndDeviceClaimingServer.Unclaim(params, {
-      dev_eui: devEui,
-      join_eui: joinEui,
+    const response = await this._api.EndDeviceClaimingServer.Unclaim({
```
adriansmares, line 56: Why not just use `params` ?

cypress/integration/console/devices/unclaim.spec.js
```diff
@@ -86,11 +86,6 @@ describe('Device un-claiming', () => {
         cy.findByRole('button', { name: /Unclaim and delete end device/ }).click()
       })
 
-    cy.wait('@unclaim-request').its('request.body').should('deep.equal', {
```
adriansmares, line 89: I am not sure this is correct, you could just check that the body is empty.

adriansmares (commented):
We should remove that `allow_delete_body` from the proto generator as well.

sdk/js/src/service/claim.js
KrishnaIyer: I believe these are equivalent, so I opt for lesser lines of code.

KrishnaIyer:
The console part is a "best effort". Let me know what else needs to be changed.

johanstokking (approved):
API and DCS LGTM

kschiffer (approved):
Codeowned LGTM.
Mildly related: #6333, but now we don't use any parameters here (other than route parameters) so it should be fine either way.

nicholaspcr (approved):
LGTM

johanstokking:
> * Also, we have to get the device from the IS to check that the EUIs are correct anyway.

As discussed offline; if the device does not exist, DCS should just proceed attempting to unclaim.

cypress/integration/console/devices/unclaim.spec.js
```diff
@@ -86,10 +86,7 @@ describe('Device un-claiming', () => {
...
-      dev_eui: ns.end_device.ids.dev_eui,
-      join_eui: '0000000000000000',
-    })
+    cy.wait('@unclaim-request').its('request.body').should('deep.equal', {})
```
mjamescompton: Use just `cy.wait('@unclaim-request')`

`.its('request.body').should('deep.equal', {})` will always be true as its your incepted response.

mjamescompton (changes requested):

KrishnaIyer:
I'll close this PR but keep the branch. I'll adapt some of these changes after #6381 is merged.

PR #6331, opened 2023-06-16
Automatically set the GOMEMLIMIT in container environments

adriansmares:
#### Summary

This PR adds support for [KimMachineGun/automemlimit](https://github.com/KimMachineGun/automemlimit), which automatically sets the [GOMEMLIMIT](https://tip.golang.org/doc/gc-guide) environment variable. This in turn allows us to use more aggressive `GOGC` values in containerized environments.

#### Changes

- Automatically set `GOMEMLIMIT` in container environments.

#### Testing

Tested locally in a constrained container.

##### Regressions

None. We did not set a `GOMEMLIMIT` before, and `cgroup` based limits are hard anyway.

#### Checklist

### DIFF
--- a/cmd/ttn-lw-stack/commands/root.go
+++ b/cmd/ttn-lw-stack/commands/root.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/KimMachineGun/automemlimit/memlimit"
 	"github.com/getsentry/sentry-go"
 	"github.com/spf13/cobra"
 	"go.thethings.network/lorawan-stack/v3/cmd/internal/commands"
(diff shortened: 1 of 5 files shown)


cmd/ttn-lw-stack/commands/root.go
```diff
@@ -109,6 +110,9 @@ var (
 			if _, err := maxprocs.Set(); err != nil {
 				logger.WithError(err).Debug("Failed to set GOMAXPROCS")
 			}
+			if _, err := memlimit.SetGoMemLimit(0.9); err != nil {
+				logger.WithError(err).Debug("Failed to set GOMEMLIMIT")
```
adriansmares: Compared to `automaxprocs`, `automemlimit` does return an error when no `cgroup` limits can be found. I believe that it is fine to log this error as a debug message.

johanstokking: Yes. I would log info level here though, also for max procs.

johanstokking (approved):

PR #6335, opened 2023-06-22
Fix claiming issues in frontend

kschiffer:
#### Summary
This quickfix PR addresses some issues with device claiming in the Console and JS SDK.

#### Changes

- Fix claim info request when claiming device (JS SDK)
- Improve error messaging for failed device claims

#### Testing

Manual testing.

#### Checklist


pkg/webui/console/containers/device-onboarding-form/index.js
```diff
@@ -113,6 +113,14 @@ const DeviceOnboardingForm = () => {
           delete mac_state.current_parameters
         }
 
+        if (values.authenticated_identifiers) {
+          cleanedValues.ids = {
+            ...cleanedValues.ids,
+            ...values.authenticated_identifiers,
+            device_id: values.target_device_id,
+          }
+        }
```
johanstokking: This causes a regression in the manual device onboarding flow, causing `ids.dev_eui` to be set to `""` and `ids.authentication_code` to be set, which doesn't exist in `ids`.

johanstokking (changes requested):

pkg/webui/console/containers/device-onboarding-form/index.js
johanstokking: I think there are two issues here: 1) `values.authenticated_identifiers` seems to be always truthy as it's an object, but the values `dev_eui` and `authentication_code` can be empty and 2) this sets `authentication_code` from `authenticated_identifiers` in `ids` which is wrong, it should be at most the `dev_eui`.

johanstokking:
```suggestion
        if (values.authenticated_identifiers.dev_eui) {
          cleanedValues.ids = {
            ...cleanedValues.ids,
            dev_eui: values.authenticated_identifiers.dev_eui,
            device_id: values.target_device_id,
          }
        }
```

adriansmares (approved):

sdk/js/src/service/devices/index.js
```diff
@@ -521,12 +521,8 @@ class Devices {
               application_id: applicationId,
             },
           }
-      try {
```
kschiffer, line 524: This try/catch and wrapping the error will make us loose the error shape we use in the Console to forward errors to the user.

ryaplots (approved):

PR #6378, opened 2023-07-05
Limit tech admin contact operations via IS config

### LINKED ISSUE #6323: IS config for limiting who can be a admin/tech contact for entities
### Summary

References #4930

Like suggested by Johan on the issue, instead of privacy flags there should be IdentityServer configuration option which restricts an user from setting another collaborator as a administrative/technical contact.

With this, operations that could abuse the privacy of other users are restricted.  In addition to that, it removes the necessity of privacy flags on the `Collaborator` entity.

### Current Situation

Administrative/Technical contacts are already implemented and they replace the previous `ContactInfo`, but with the current implementation any collaborator would be able to set another collaborator as one of these contacts. 

A IS config option which restricts whom an user can set as a contact to only himself.

### Proposed Implementation

Add an extra field to `IsConfiguration` which disallows the user to set others as the administrative/technical contact.

### Contributing

### Code of Conduct

nicholaspcr:
#### Summary

Closes #6323

#### Changes

- Add an contact info restriction error
- Add `restrictions` field to IS config, alongside the contact info restriction
- Update entities registries to validate contact if config is enabled

#### Testing

Unit tests and manual testing

##### Regressions

This is a new restriction on any flow that sets the admin/tech contacts which aren't done through the console. So I can't imagine a regression and most likely CLI users aren't setting those contacts since the old way still works (even with the deprecated warning).

#### Notes for Reviewers

Was going to do a testcase just for restrictions for each registry test file but it felt like overkill, let me know if the smaller tests found currently on the PR are appropriate enough.

If there are any lint errors on CI, I'll run the tests again after the format merge of the separate repository. Made the lint changes into a different PR just to remove it from this PR.

#### Checklist


api/identityserver.proto
```diff
@@ -125,6 +125,19 @@ message IsConfiguration {
   reserved 10; reserved "application_limits";
   reserved 11; reserved "organization_limits";
   reserved 12; reserved "user_limits";
+
+  // Restrictions contains constraints that are applied to the IS's business logic.
+  message Restrictions {
+    // Entity contains restrictions that are applied to entities.
+    message Entity {
+      // restrict_contacts forbids standard users from setting others as the administrative or technical contacts.
+      google.protobuf.BoolValue restrict_contacts = 1;
```
johanstokking: I would make this more explicit. The comment is clear but the field name is not really. Maybe `allow_set_others_as_contact` ?

Also nit; no need for Go style comments here, 
```suggestion
      // Allow anyone to set others as the administrative or technical contacts.
      google.protobuf.BoolValue allow_set_others_as_contact = 1;
```

nicholaspcr: Indeed the name could be better but I do disagree with the name suggestion slightly. Since this is within the restriction config, it makes more sense in my eyes to have variable which are capable of restriction rather than allowing it all operations.

TLDR: probably more appropriate to have something like `disallow_set_other_as_contact`

Will do the changes in a bit, regarding the `nit`, I know its rather insignificant but its mostly something for when someone wants to check the GoDoc of the generated code

pkg/identityserver/config.go
```diff
@@ -115,6 +115,11 @@ type Config struct {
 		TenantID string         `name:"tenant-id" description:"Tenant ID in the host NetID"`
 	} `name:"network"`
 	TelemetryQueue telemetry.TaskQueue `name:"-"`
+	Restrictions   struct {
+		Entity struct {
+			DisallowSetOtherAsContact bool `name:"disallow-set-other-as-contact" description:"Restrict users from setting others as the administrative or technical contact of any entity"`
```
adriansmares: How about `CollaboratorRights.SetOthersAsContacts` ? We already have `UserRights` and `AdminRights`, and `SetOthersAsContacts` (default `false`) would disallow collaborators from setting the contacts to other values other than their own identity.

pkg/identityserver/application_registry_test.go
```diff
@@ -187,6 +187,32 @@ func TestApplicationsCRUD(t *testing.T) {
 			a.So(updated.Name, should.Equal, "Updated Name")
 		}
 
+		t.Run("contact_info_restriction", func(t *testing.T) { // nolint:paralleltest
+			a, ctx := test.New(t)
+			is.config.Restrictions.Entity.DisallowSetOtherAsContact = true
+			t.Cleanup(func() { is.config.Restrictions.Entity.DisallowSetOtherAsContact = false })
```
adriansmares: You should set the old value, not `false`.

pkg/identityserver/store/errors.go
```diff
@@ -124,4 +124,8 @@ var (
 		"application_dev_eui_limit_reached",
 		"application issued DevEUI limit ({dev_eui_limit}) reached",
 	)
+
+	ErrContactInfoRestricted = errors.DefinePermissionDenied(
```
adriansmares, line 128: Maybe the description should be `contact information can only reference to the caller` ?

adriansmares (commented):
Implementation LGTM in general. I have some naming concerns only.

pkg/identityserver/config.go
nicholaspcr: I did think about doing something similar to `AdminRights` but for Colalborators like you mentioned but the way I'm tackling this mentally is more in regards to how classify these sort of operations. 

I was going to write that I'm seeing this more as a set of operations restricted by deployment configurations than `Rights` but it should be able to be bypassed by admins (which currently it doesn't) so indeed your suggestion probably makes more sense

api/identityserver.proto
```diff
@@ -125,6 +125,11 @@ message IsConfiguration {
   reserved 10; reserved "application_limits";
   reserved 11; reserved "organization_limits";
   reserved 12; reserved "user_limits";
+
+  message CollaboratorRights {
+    google.protobuf.BoolValue set_other_as_contact = 1;
```
adriansmares: `set_others_as_contacts`.

pkg/identityserver/config.go
```diff
@@ -83,6 +83,9 @@ type Config struct {
 	AdminRights struct {
 		All bool `name:"all" description:"Grant all rights to admins, including _KEYS and _ALL"`
 	} `name:"admin-rights"`
+	CollaboratorRights struct {
+		SetOthersAsContacts bool `name:"set-others-as-contacts" description:"Allow users to set other users as their contact"`
```
adriansmares: `as entity contacts`.

pkg/identityserver/gateway_registry_test.go
```diff
@@ -207,6 +207,56 @@ func TestGatewaysCRUD(t *testing.T) {
 			a.So(updated.Name, should.Equal, "Updated Name")
 		}
 
+		t.Run("contact_info_restrictions", func(t *testing.T) { // nolint:paralleltest
```
adriansmares: Spaces are automatically converted to underlines, so you can just leave the spaces in the naming, or use something like `ContactInfoRestrictions`. We don't have a consistent style in the codebase.

pkg/identityserver/store/errors.go
adriansmares: Apparently it should be `contact information can only reference the caller`, apologies.

adriansmares (approved):

pkg/identityserver/gateway_registry_test.go
nicholaspcr: Was using the underscore to make it easier to skip the whole test name under my text editor, since I don't want to create a third way of doing things I changed to the first suggestion. Thanks for catching that.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ For details about compatibility between different releases, see the **Commitment
 - RPCs and CLI command to delete a batch of end devices within an application.
   - Check `ttn-lw-cli end-devices batch-delete` for more details.
 - Add `UserInput` component to the Console to handle user id input fields by implementing an autosuggest.
+- The Identity Server configuration has a new optional configurable restriction regarding adminstrative and technical contacts of entities. This limits the action of an user or organization to set these contacts only to themselves, its enabled by default but it is possible to disable it by setting `is.collaborator-rights.set-others-as-contacts` as true.
```
adriansmares: I think we should default to `true` in order to avoid breaking changes.

nicholaspcr: yeah, I agree

PR #6381, opened 2023-07-05
Remove body from Unclaim request

KrishnaIyer:
#### Summary

Remove body from Unclaim request. Refs https://github.com/TheThingsNetwork/lorawan-stack/pull/6325

#### Changes

- Extract only the API and the tooling change for a backport.

#### Testing

Console E2E should cover this. Will test in staging.

##### Regressions

This fixes a bug.

#### Checklist


pkg/webui/console/views/device-general-settings/index.js
```diff
@@ -100,7 +102,9 @@ const DeviceGeneralSettings = () => {
     const {
       ids: { dev_eui: devEui, join_eui: joinEui },
     } = device
-    await dispatch(attachPromise(unclaimDevice(appId, devId, devEui, joinEui)))
+    await dispatch(
+      attachPromise(unclaimDevice(appId, devId, hexToBase64(devEui), hexToBase64(joinEui))),
```
adriansmares: Should we do this in the SDK instead ? We don't have to edit all of the callers to this format then.

adriansmares (approved):

pkg/webui/console/views/device-general-settings/index.js
KrishnaIyer: The thing is that the changes in https://github.com/TheThingsNetwork/lorawan-stack/pull/6325 will remove EUIs from this request.

adriansmares: That's right, but if we want to support the use case in which we allow unclaiming if the device is not registered, I think we should keep the identifiers around. Otherwise you cannot know what EUIs to unclaim.

KrishnaIyer:
@mjamescompton / @kschiffer: Can you quickly look at the console change? This is a bug that I'd like to patch in the current release.

pkg/webui/console/views/device-general-settings/index.js
kschiffer: I agree, if there's a requirement to convert to base64 then we should do that on SDK level so we keep handling of byte data consistent.

KrishnaIyer: Ok is that small enough to quickly add to this PR? Unclaiming is currently broken and I don't know where to look to update something in the SDK.

KrishnaIyer: Ok done.

sdk/js/src/util/bytes.js
```diff
@@ -0,0 +1,46 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Converts hex encoded string to base64.
+ *
+ * @param {string} str - Hex encoded string.
+ * @returns {string} - `str` base64 encoded.
+ */
```
KrishnaIyer: This file is a duplicate of `console/lib/bytes` but it doesn't seem wise to import the console functions here.

kschiffer (approved):

sdk/js/src/util/bytes.js
adriansmares: Maybe it would be wise to make the console use this SDK function instead of the console one in a tech debt PR in the future.

adriansmares (approved):

PR #6385, opened 2023-07-07
Implement End Device Batch Get RPC

KrishnaIyer:
#### Summary

Implement End Device Batch Get RPC. References https://github.com/TheThingsIndustries/lorawan-stack/issues/3632. This is a follow up of https://github.com/TheThingsIndustries/lorawan-stack/pull/3676

Contents are the same as https://github.com/TheThingsIndustries/lorawan-stack/pull/3729

#### Changes

- Add new Get RPC for the EndDeviceBatchRegistry.
- Add RPC implementation and reuse store.FindEndDevices under the hood.
- Modify `ttn-lw-cli batch-delete` to use `Get`.

#### Testing

Covered by unit tests.

##### Regressions

Tested that `batch-delete` works.

```
 ./ttn-lw-cli dev list test-app
WARN	Using insecure connection to OAuth server
WARN	Using insecure connection to API
[{
  "ids": {
    "device_id": "test-dev-1",
    "application_ids": {
      "application_id": "test-app"
    },
    "dev_eui": "0004A310001AB101",
    "join_eui": "70B3D57ED0000000"
  },
  "created_at": "2023-07-07T08:56:05.347245Z",
  "updated_at": "2023-07-07T08:56:05.347246Z"
}, {
  "ids": {
    "device_id": "test-dev-2",
    "application_ids": {
      "application_id": "test-app"
    },
    "dev_eui": "0004A310001AB102",
    "join_eui": "70B3D57ED0000000"
  },
  "created_at": "2023-07-07T08:56:06.181713Z",
  "updated_at": "2023-07-07T08:56:06.181713Z"
}, {
  "ids": {
    "device_id": "test-dev-3",
    "application_ids": {
      "application_id": "test-app"
    },
    "dev_eui": "0004A310001AB103",
    "join_eui": "70B3D57ED0000000"
  },
  "created_at": "2023-07-07T08:56:06.241736Z",
  "updated_at": "2023-07-07T08:56:06.241737Z"
}]
-----------------------------------------------------------------------------------------------
./ttn-lw-cli dev batch-delete test-app test-dev-1 test-dev-2 test-dev-3
WARN	Using insecure connection to OAuth server
WARN	Using insecure connection to API
-----------------------------------------------------------------------------------------------
./ttn-lw-cli dev list test-app
WARN	Using insecure connection to OAuth server
WARN	Using insecure connection to API
[]

```

#### Notes for Reviewers

1. I've reused `store.FindEndDevices` under the hood. This seems to me the easiest since the ordering/pagination is built in. If I need to implement a store interface and then internally use `listEndDevicesby` (as done with the search RPCs), I can do that as well.
2. This PR adds the Get RPC only to the `EndDeviceBatchRegistry` since that's what's relevant for subsequent PRs. This PR does not add a Get RPC to the `XsEndDeviceBatchRegistry` and consequently, there is no CLI command.

#### Checklist


pkg/identityserver/end_device_registry.go
```diff
@@ -510,6 +510,56 @@ func (is *IdentityServer) batchDeleteEndDevice(
...
+
+	ids := make([]*ttnpb.EndDeviceIdentifiers, 0, len(req.DeviceIds))
+	for _, id := range req.DeviceIds {
+		ids = append(ids, &ttnpb.EndDeviceIdentifiers{
+			ApplicationIds: req.ApplicationIds,
+			DeviceId:       id,
+		})
+	}
+
+	err = is.store.Transact(ctx, func(ctx context.Context, st store.Store) error {
```
KrishnaIyer, line 536: Note to self: This needs to be modified in enterprise after merging.

api/end_device.proto
```diff
@@ -941,3 +941,28 @@ message BatchDeleteEndDevicesRequest {
...
+  // This mask is applied on all the end devices in the result.
+  // See the API reference for which fields can be returned by the different services.
+  google.protobuf.FieldMask field_mask = 3;
+  // Order the results by this field path (must be present in the field mask).
+  // Default ordering is by ID. Prepend with a minus (-) to reverse the order.
+  string order = 4 [
+    (validate.rules).string = { in: ["", "device_id", "-device_id", "join_eui", "-join_eui", "dev_eui", "-dev_eui",
+      "name", "-name", "description", "-description", "created_at", "-created_at", "last_seen_at", "-last_seen_at"] }
+  ];
+  // Limit the number of results per page.
```
adriansmares: It's a bit weird that we have pagination in a request where the identifiers have been fully provided to be honest (and the limit is also small). Should we just keep these fields as `reserved` (id + name) ?

adriansmares (approved):

nicholaspcr (approved):

api/end_device.proto
KrishnaIyer: Without pagination, would the console then set a limit itself on how many items are fetched?
Also, we do plan to expand this limit at some point.

adriansmares: I don't understand the use case for the Console to send a lot of identifiers in one request then paginate (thus sending them again, and again, and again). The Console already has its own pages, and knows the identifiers, thus it does not need a second layer of pagination in place.

I have the same feeling for any client that knows the identifiers ahead of time - it does not need a second layer of pagination, since it knows the requested identifiers set already.

KrishnaIyer: Ok understood. 

Do you want me to remove the ordering as well?

> Should we just keep these fields as reserved (id + name) ?

Do you mean keep the api fields reserved but not use it?

adriansmares: The backend ordering matters a lot for pagination, since you cannot sort without having all of the pages available as a client. I think the ordering is not needed either.

For future proofing in general we reserve the field numbers and the field names, such that in the future we can come back and add the functionality if needed. You can see an example of this [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/7d93026eecbb8f20133794d5a375a93a8681a677/api/identityserver.proto#L124).

adriansmares (approved):

PR #6403, opened 2023-07-19
Run the CLI telemetry submission asynchronously

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/issues/6081

#### Changes

- Start the telemetry submission in the command pre run, and only wait in the post run.

#### Testing

Local testing.

##### Regressions

N/A, but may be susceptible to lambda slow starts. I think that for now this is an acceptable trade of, given that we don't want to slow down the CLI excessively. 

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [ ] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #6414, opened 2023-08-01
Fix telemetry async submission

nicholaspcr:
#### Summary

References #6403, the lack of a message sent to the channel implies in the telemetry always timing out.

#### Changes

- Add `telemetrySubmission <- struct{}{}`

#### Testing

Manual tests, also confirmed that the CLI interacts with the telemetry collector properly and that it shows on the Grafana dashboard.

#### Notes for Reviewers

I noticed that when deleting the telemetry's cache file I got a timeout even with a second delay, I believe its due to the initial creation of the telemetry cache file and the yaml parsing involved in that:
```go
// Write telemetry state into the provided path.
func (cst *cliStateTelemetry) Write(p string) error {
	b, err := yaml.Marshal(cst)
	if err != nil {
		return err
	}

	// Creates the file since it does not exist.
	if err := os.WriteFile(p, b, 0o644); err != nil { //nolint:gas
		return err
	}
	return nil
}
```
All subsequent send operations are within the time window (1s) regardless if its on a SSD or a normal hard drive, which shouldn't make a difference since the time window is so large. I'm thinking that the parsing is the main culprit for the slowdown, even if the data is quite small.

#### Checklist

### DIFF
--- a/cmd/ttn-lw-cli/commands/root.go
+++ b/cmd/ttn-lw-cli/commands/root.go
@@ -179,6 +179,7 @@ func preRun(tasks ...func() error) func(cmd *cobra.Command, args []string) error
 				cli.NewCLITelemetry(
 					cli.WithCLITarget(config.Telemetry.Target),
 				).Run(ctx)
+				telemetrySubmission <- struct{}{}
 			}(ctx)
 		} else {
 			close(telemetrySubmission)


nicholaspcr:
Never mind, my mistake. Fooled myself while testing the CLI

PR #4996, opened 2021-12-15
Add administrative and technical contacts to replace contact info

htdvisser:
#### Summary

This pull request adds `administrative_contact` and `technical_contact` to applications, OAuth clients, gateways and organizations. ~It also adds partial Console support.~

Refs #4930.

#### Changes

- Define API
- Define Database Models in the IS
- Modify stores to handle creates/updates with contacts
- Modify Update RPCs to handle creates/updates with contacts
- Update CLI to allow un-setting contacts

#### Testing

- Updated IS unit tests

##### Regressions

n/a

#### Notes for Reviewers

nil

#### Checklist

- [x] Scope: The referenced issue is addressed, there are no unrelated changes.
- [x] Compatibility: The changes are backwards compatible with existing API, storage, configuration and CLI, according to the compatibility commitments in `README.md` for the chosen target branch.
- [ ] Documentation: Relevant documentation is added or updated.
- [x] Changelog: Significant features, behavior changes, deprecations and fixes are added to `CHANGELOG.md`.
- [x] Commits: Commit messages follow guidelines in `CONTRIBUTING.md`, there are no fixup commits left.

PR #6430, opened 2023-08-02
Prioritize gateway's field contacts over Contact Info list

### LINKED ISSUE #4930: Replace ContactInfo with UserIdentifiers as Administrative/Technical/Security Contacts
#### Summary

I think we should remove the `repeated ContactInfo` from entities and instead add `UserIdentifiers` for administrative, technical and security contacts.

Users can configure `repeated ContactInfo` for applications, clients, gateways, organizations and users. For users we already have their contact info (primary_email_address). For the other entities it's practically unused. It's also very annoying to maintain email addresses on many entities.

Instead, we should allow selecting a user (or perhaps organization) as Administrative / Technical / Security contact on these entities.

- Deprecate `ContactInfo` in those entities
- Add `UserIdentifiers administrative_contact`, `technical_contact` and `security_contact` to entities (perhaps not to User, because there it doesn't really make sense anymore).

Question: Should it be `UserIdentifiers` or should it be `OrganizationOrUserIdentifiers`, because sometimes you can't really assign a single user.

nicholaspcr:
#### Summary

Closes #4930

While going through the files of PBA to check what are the necessary changes it was noticed the following:
- The PBA operates assuming that only admin/tech contacts are send, but uses the `ContactInfo` slice to pass those contacts around.
- The `UpdateGateway` method in `grpc_gspba` was only fetching contacts from the `gtw.ContactInfo` and not from the `gtw.AdministrativeContact` and `gtw.TechnicalContact`.
- GatewayServer's pba client doesn't include the gateway's admin/tech contacts in the requests.

This PR addresses the last two items mentioned above.

#### Changes

- Add `gtw.AdministrativeContact` and `gtw.TechnicalContact` to PBA requests in GS.
- Prioritize gtw's field contacts over emails present in the `ContactInfo` slice

#### Testing

Don't believe there are any unit tests that cover this specific change and I'm not sure how to test the change locally.

**Correction**: Never mind the comment regarding the lack of tests, the `agent_test` cover the gspba.

##### Regressions

I'm not sure but the change that would have the most impact would be the addition of the gateway's `AdministrativeContact` and `TechnicalContact`, where previously no contact information was sent. Tried to understand why but wasn't clear to me so I'm asking for the opinion of the reviewers on the subject.

#### Notes for Reviewers

Main worry was listed on the Regression section.

#### Checklist


pkg/gatewayserver/upstream/packetbroker/packetbroker.go
```diff
@@ -118,12 +118,14 @@ func (h *Handler) ConnectGateway(ctx context.Context, ids *ttnpb.GatewayIdentifi
...
-			FrequencyPlanIds: gtw.FrequencyPlanIds,
-			StatusPublic:     gtw.StatusPublic,
-			LocationPublic:   gtw.LocationPublic,
-			Online:           true,
+			Ids:                   pbIDs,
+			Antennas:              antennas,
+			FrequencyPlanIds:      gtw.FrequencyPlanIds,
+			StatusPublic:          gtw.StatusPublic,
+			LocationPublic:        gtw.LocationPublic,
+			AdministrativeContact: gtw.AdministrativeContact,
```
adriansmares, line 124: For the record I am not sure if `gtw.*Contact` won't always be null.

`conn.Gateway()` is retrieved when the gateway connects, and the field mask for that call is fixed here:
https://github.com/TheThingsNetwork/lorawan-stack/blob/c8e625243ba8106a5240d4c5d7b0feb0fc1e9a3c/pkg/gatewayserver/gatewayserver.go#L474-L491
You will have to add the `contact_info` both here and in the field mask, and also the `*_contact` fields.

pkg/gatewayserver/upstream/packetbroker/packetbroker.go
```diff
@@ -163,9 +165,11 @@ func (h *Handler) ConnectGateway(ctx context.Context, ids *ttnpb.GatewayIdentifi
 
 		req := &ttnpb.UpdatePacketBrokerGatewayRequest{
 			Gateway: &ttnpb.PacketBrokerGateway{
-				Ids:          pbIDs,
-				Online:       true,
-				StatusPublic: gtw.StatusPublic,
+				Ids:                   pbIDs,
+				Online:                true,
+				StatusPublic:          gtw.StatusPublic,
+				AdministrativeContact: gtw.AdministrativeContact,
```
adriansmares: I would do it only on connection to be honest, I don't think that we have to update the contact info contiguously.

pkg/packetbrokeragent/grpc_gspba.go
```diff
@@ -158,7 +158,19 @@ func (s *gsPbaServer) UpdateGateway(ctx context.Context, req *ttnpb.UpdatePacket
 	}
 
 	if ttnpb.HasAnyField(req.FieldMask.GetPaths(), "contact_info") {
+		adminUsrId := req.Gateway.GetAdministrativeContact().GetUserIds()
```
adriansmares: For the record this is wrong - it is possible that the administrative contact is an organization, and in that case you will need to follow the chain and check what the administrative contact of that organization will be (that one is guaranteed to be a user, since organizations cannot be collaborators to another organization).

pkg/packetbrokeragent/agent_test.go
```diff
@@ -461,63 +461,131 @@ func TestForwarder(t *testing.T) {
...
-			a.So(update.AdministrativeContact.GetValue().GetEmail(), should.Equal, "admin@example.com")
-			a.So(update.TechnicalContact.GetValue().GetEmail(), should.Equal, "tech@example.com")
-			a.So(update.FrequencyPlan.GetLoraMultiSfChannels(), should.HaveLength, 8)
-			a.So(update.Online.GetValue(), should.BeTrue)
-			a.So(update.GatewayLocation.GetLocation().GetTerrestrial().GetAntennaCount().GetValue(), should.Equal, 1)
-		case <-time.After(timeout):
-			t.Fatal("Expected gateway update timeout")
-		}
+		t.Run("Admin and tech contacts present", func(t *testing.T) {
+			adminId := &ttnpb.UserIdentifiers{UserId: "admin", Email: "newAdmin@example.com"}
```
adriansmares: Please follow the Go naming conventions and use `ID` both here and in other places.

pkg/packetbrokeragent/grpc_gspba.go
```diff
@@ -158,7 +158,19 @@ func (s *gsPbaServer) UpdateGateway(ctx context.Context, req *ttnpb.UpdatePacket
...
+		techUsrId := req.Gateway.GetTechnicalContact().GetUserIds()
 		adminContact, techContact := toPBContactInfo(req.Gateway.GetContactInfo())
+
+		// Prioritize administrative and technical contacts attached to the gateway instead of the ones present in
+		// ContactInfo slice.
+		if req.Gateway.AdministrativeContact != nil {
+			adminContact = &packetbroker.ContactInfo{Name: adminUsrId.GetUserId(), Email: adminUsrId.GetEmail()}
```
adriansmares: The `email` in the `UserIdentifiers` message is one of our API foot guns.

It exists solely for the `forget-password` functionality, and it is never used in other places. In fact, not even the forgotten password functionality uses it, if you check - `bunstore.GetUser` ignores it, and the implementation for `createTemporaryPassword` doesn't care about it either. I wish I could erase it from the API to be honest, but that's out of scope for this PR.

Back to this PR - you will need to retrieve the email via an RPC to the Identity Server in order to actually access the email. You will need to do this using cluster rights, and you will need to use the same tactic in order to retrieve the organization for the issue I've mentioned above (contact being an organization).

adriansmares (commented):
Please make sure that you follow the Go naming guidelines.

Regarding testing: @johanstokking is there some way to give developers access to PB staging in order to allow them to check changes like this one ?

pkg/packetbrokeragent/grpc_gspba.go
johanstokking: Right. I think this is the tricky part of addressing this issue in PBA. Let us know if you need help on this @nicholaspcr.

pkg/gatewayserver/upstream/packetbroker/packetbroker.go
johanstokking: Yes, this call is only meant to bump the online status.

pkg/packetbrokeragent/agent_test.go
```diff
@@ -461,63 +461,131 @@ func TestForwarder(t *testing.T) {
...
+			})
+			a.So(err, should.BeNil)
+			a.So(res.OnlineTtl.AsDuration(), should.NotBeZeroValue)
+
+			select {
+			case update := <-updateCh:
+				a.So(update.AdministrativeContact.GetValue().GetName(), should.Equal, "admin")
+				a.So(update.AdministrativeContact.GetValue().GetEmail(), should.Equal, "newAdmin@example.com")
+				a.So(update.TechnicalContact.GetValue().GetName(), should.Equal, "tech")
+				a.So(update.TechnicalContact.GetValue().GetEmail(), should.Equal, "newTech@example.com")
```
johanstokking: Very much a nit; keep email addresses lowercase

johanstokking:
I just documented how to use Packet Broker staging environment in our internal KB: https://discuss.thethingsindustries.com/t/packet-broker-staging-environment-for-local-development-use/983.

nicholaspcr:
@adriansmares @johanstokking  re-requesting your review.

Please ignore lint issues as I'll fix them before merging.

When working on this something came to mind. If I understood correctly, when a gateway first connect it has the `updateGatewayTimeout` (5 seconds) to first interact and at every change the time renews. With this in mind, if PB would notify the admin/tech contacts every time there is an update its quite easy for the notification to become spam, no?

johanstokking:
> With this in mind, if PB would notify the admin/tech contacts every time there is an update its quite easy for the notification to become spam, no?

The timeout is really used for Packet Broker to process the update itself. If Packet Broker is slow to process the gateway update, PBA just cancels it. This is to avoid that too many concurrent requests are outstanding.

Packet Broker does not send any notifications to anyone, certainly not on gateway updates. Packet Broker uses the contact information for internal use for now, so that we can contact folks if things are off. In the future, the contact info can be used for network operators to contact each other.

pkg/gatewayserver/upstream/packetbroker/packetbroker.go
```diff
@@ -139,6 +141,7 @@ func (h *Handler) ConnectGateway(ctx context.Context, ids *ttnpb.GatewayIdentifi
 			"rx_rate",
 			"status_public",
 			"tx_rate",
+			"contact_info",
```
johanstokking: This doesn't look right;

1. `contact_info` is not set as field
2. Field mask should be sorted (that's for us, no technical reason)

pkg/gatewayserver/gatewayserver.go
```diff
@@ -487,6 +487,9 @@ func (gs *GatewayServer) Connect(
 			"schedule_downlink_late",
 			"status_public",
 			"update_location_from_status",
+			"contact_info",
+			"administrative_contact",
+			"technical_contact",
```

pkg/gatewayserver/gatewayserver.go
```diff
@@ -487,6 +487,9 @@ func (gs *GatewayServer) Connect(
 			"schedule_downlink_late",
 			"status_public",
 			"update_location_from_status",
+			"contact_info",
```
johanstokking: Why contact info?

pkg/packetbrokeragent/grpc_gspba.go
```diff
@@ -159,12 +160,25 @@ func (s *gsPbaServer) UpdateGateway(ctx context.Context, req *ttnpb.UpdatePacket
 
 	if ttnpb.HasAnyField(req.FieldMask.GetPaths(), "contact_info") {
 		adminContact, techContact := toPBContactInfo(req.Gateway.GetContactInfo())
```
johanstokking: `GetContactInfo()` will always return `nil` here.

pkg/packetbrokeragent/grpc_gspba.go
```diff
@@ -159,12 +160,25 @@ func (s *gsPbaServer) UpdateGateway(ctx context.Context, req *ttnpb.UpdatePacket
...
-		updateReq.AdministrativeContact = &packetbroker.ContactInfoValue{
-			Value: adminContact,
+
+		// Only replaces the administrative and technical contacts if no error happens.
```
johanstokking: What kind of errors would occur that can just be swallowed?

pkg/packetbrokeragent/is.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// GetOrganization calls the Entity Registry with cluster auth and returns an organization.
+func (is IS) GetOrganization(ctx context.Context, req *ttnpb.GetOrganizationRequest) (*ttnpb.Organization, error) {
+	registry, err := is.newOrganizationRegistryClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return registry.Get(ctx, req, is.WithClusterAuth())
+}
```
johanstokking: Wouldn't we need some form of caching here? Otherwise it looks like worst case we get the following rpcs per gateway connection:

1. Get gateway
2. Get organization if admin contact is an org
3. Get administrative contact of the organization
4. Get organization if the tech contact is an org (could be same as 2)
5. Get technical contact of the organization (could be same as 3)

With thousands of gateways reconnecting, likely linked to the same contacts, I think that caching orgs and user contact info is crucial.

pkg/packetbrokeragent/agent_test.go
```diff
@@ -461,63 +488,129 @@ func TestForwarder(t *testing.T) {
...
-				Ids: &ttnpb.PacketBrokerGateway_GatewayIdentifiers{
-					GatewayId: "foo-gateway",
-					Eui:       types.EUI64{0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x88}.Bytes(),
-				},
-				ContactInfo: []*ttnpb.ContactInfo{
-					{
-						ContactType:   ttnpb.ContactType_CONTACT_TYPE_OTHER,
-						ContactMethod: ttnpb.ContactMethod_CONTACT_METHOD_EMAIL,
-						Value:         "admin@example.com",
+		t.Run("Only contactInfo present", func(t *testing.T) { // nolint:paralleltest
```
johanstokking: Check whether we need this test case as the `contact_info` field is not set in code

johanstokking: Also inconsistent casing `contactInfo`; it's `contact_info` or `ContactInfo`

pkg/packetbrokeragent/agent_test.go
```diff
@@ -461,63 +488,129 @@ func TestForwarder(t *testing.T) {
...
-		case update := <-updateCh:
-			a.So(update.AdministrativeContact.GetValue().GetEmail(), should.Equal, "admin@example.com")
-			a.So(update.TechnicalContact.GetValue().GetEmail(), should.Equal, "tech@example.com")
-			a.So(update.FrequencyPlan.GetLoraMultiSfChannels(), should.HaveLength, 8)
-			a.So(update.Online.GetValue(), should.BeTrue)
-			a.So(update.GatewayLocation.GetLocation().GetTerrestrial().GetAntennaCount().GetValue(), should.Equal, 1)
-		case <-time.After(timeout):
-			t.Fatal("Expected gateway update timeout")
-		}
+		t.Run("admin-tech contacts present", func(t *testing.T) { // nolint:paralleltest
```
johanstokking: Inconsistent field naming

pkg/gatewayserver/upstream/packetbroker/packetbroker.go
nicholaspcr: Basically added it in order to trigger
https://github.com/TheThingsNetwork/lorawan-stack/blob/e0f0719673c291b87b607712cf465f1179655308/pkg/packetbrokeragent/grpc_gspba.go#L161

But indeed it should be `administrative_contact` and `technical_contact`. 
Since neither the `contact_info` field mask or its value was there before I'm not adding them (specially since the field is deprecated)

adriansmares: I interpret @johanstokking's comment as 'why are you making it part of the field mask, but you are not making it part of the `PacketBrokerGateway` on line 120'.

pkg/gatewayserver/gatewayserver.go
nicholaspcr: When solving https://github.com/TheThingsNetwork/lorawan-stack/pull/6430#discussion_r1283092849 I just added the three of them, for some reason I still have it my mind that `contact_info` should still be in there since its total deprecation would only truly happen on `v4` but that is not the case. So I'll remove it and leave only the `administrative_contact` and `technical_contact`

pkg/packetbrokeragent/grpc_gspba.go
nicholaspcr: True and that also has been the previous state of the interaction between the GS and PBA before this PR. In the comment that you made [here](https://github.com/TheThingsNetwork/lorawan-stack/pull/6430#discussion_r1290989970) the question was in regards to why have the `contact_info` on the `gatewayserver.go`, the reason would be that we would be for fetching the gateway's `contact_info` in order for it to be sent to the PBA. 

But since the field is deprecated and the behavior wasn't present before, there is no point in adding it in.

pkg/packetbrokeragent/grpc_gspba.go
nicholaspcr: Since a gateway's admin/tech contact cannot be a user/org that doesn't exist there and we are using `WithClusterAuth`, there should not be an error in regards to business logic. With this in mind the reason for swallowing the error would be to not interrupt the `UpdateGateway` procedure due to error in the connection between the PBA and IS. Regardless I should at least log the error in case something unexpected happens. 

Let me know if you agree with my logic in this case

pkg/packetbrokeragent/is.go
nicholaspcr: Makes a lot of sense, will add an in memory cache (not Redis). That should still have some problems with `packetbroker` replicas generating more calls than it is needed but its miles better than no cache.

Note for myself: It should have a reasonable TTL, users don't update their `primary_email_address` or `contact_info` that often so a long time shouldn't be a problem. Also not marked in this comment but I didn't validate on the local tests if the email being set is the one we use for sending notifications, so have to check that as well

pkg/gatewayserver/upstream/packetbroker/packetbroker.go
nicholaspcr: I believe the reply answers that as well, the idea of the field mask was to trigger the usage of the admin/tech fields which is wrong by itself (it should've been their respective FieldMasks). In addition to that the `ContactInfo` field wasn't present before on the `PacketBrokerGateway` on line 120 and since its deprecated it shouldn't be there, right?.

adriansmares: The fact that it wasn't there before is a mistake in itself, as the code on the PBA side depends on the GS actually pushing the contact info. I would argue that the field is deprecated, not removed, so it should still be retrieved by the GS on connection and sent to PBA as well, since the PBA logic does use it as a failover when the explicit contacts are not present.

nicholaspcr: Okay, I was under the impression that it would be more correct to maintain the non usage of the ContactInfo. But your point makes sense, this should be removed in the PR which would eliminate the field completely.

pkg/packetbrokeragent/grpc_gspba.go
adriansmares: It's possible that the user is deleted after it is set as the contact, and in these cases it will be 'not found', while still being referenced in the contact fields.

As you have said, there are many reasons unrelated to the business logic itself that can cause this call to fail - connection issues, database issues, and others. While the caller cannot fix them, it should at least log them with a warning. You can consider skipping the logging for 'not found entities'.

pkg/packetbrokeragent/grpc_gspba.go
```diff
@@ -159,12 +160,25 @@ func (s *gsPbaServer) UpdateGateway(ctx context.Context, req *ttnpb.UpdatePacket
...
+		adminID, err := s.fetchUserIDFromContact(
+			ctx, req.Gateway.GetAdministrativeContact(),
+			func(o *ttnpb.Organization) *ttnpb.OrganizationOrUserIdentifiers { return o.GetAdministrativeContact() },
+		)
+		if err == nil {
+			adminContact = &packetbroker.ContactInfo{Name: adminID.GetUserId(), Email: adminID.GetEmail()}
```
adriansmares: You need to retrieve the `primary_email_address` of the `User` entity. The `UserIdentifiers.Email` is not usable.

pkg/packetbrokeragent/is.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/bluele/gcache"
+	"go.thethings.network/lorawan-stack/v3/pkg/cluster"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	"google.golang.org/grpc"
+)
+
+var (
+	cacheSize = 4096
+	cacheTTL  = 24 * 60 * 60 * time.Second // 24 hour
+)
```
nicholaspcr: Gave a look into the Grafana stats and I'm questioning myself regarding the cache size, the amount of gateways connected is a lot larger than this but I'm assuming that there are a few replicas of the PBA on production, so the cache size might be appropriate but I wanted to confirm that with you @adriansmares.

pkg/packetbrokeragent/grpc_gspba.go
```diff
@@ -211,3 +243,42 @@ func (s *gsPbaServer) UpdateGateway(ctx context.Context, req *ttnpb.UpdatePacket
...
+		org, err := s.entityRegistry.GetOrganization(ctx, &ttnpb.GetOrganizationRequest{
+			OrganizationIds: orgID,
+			FieldMask:       ttnpb.FieldMask("administrative_contact", "technical_contact"),
+		})
+		if err != nil {
+			return nil, err
+		}
+		usr, err := s.entityRegistry.GetUser(ctx, &ttnpb.GetUserRequest{
+			UserIds:   selectOrgContact(org).GetUserIds(),
+			FieldMask: ttnpb.FieldMask("ids", "primary_email_address"),
```
adriansmares: Nit: please also check `primary_email_address_validated_at` to be not null. This is important as we don't want to pass forward unverified emails.

You can avoid doing `GetUser` twice by having a top level variable with the user identifiers - initially `contact.GetUserIds()`, and the in the `if` you change it to the retrieved user if organization identifiers were actually provided.

pkg/packetbrokeragent/is.go
adriansmares: I think this would be a good moment to discuss why we care about caching, and what the mindset should be regarding caching.

In general we care a lot about caching in order to deal with pathological use cases, like gateways reconnecting _very often_ (multiple times per minute, generally in a loop). These cases are what you should imagine as a 'vocal minority' - there aren't that many of them (so you don't need a large cache size) and they are very noisy (so they will reconnect many times and thus trigger a retrieval many times).

You can get away with a lot by just having a 5 minute TTL and 1024 users/organizations.

pkg/packetbrokeragent/is.go
```diff
@@ -0,0 +1,118 @@
...
+	Cluster
+	userCache gcache.Cache
+	orgCache  gcache.Cache
+}
+
+// NewIS returns a new IS, which is a set of wrappers that perform calls to the Identity Server.
+func NewIS(c Cluster) *IS {
+	return &IS{
+		Cluster:   c,
+		userCache: gcache.New(cacheSize).LRU().Expiration(cacheTTL).Build(),
```
adriansmares: Prefer `ARC` to `LRU`.

pkg/packetbrokeragent/is.go
```diff
@@ -0,0 +1,118 @@
...
+		orgCache:  gcache.New(cacheSize).LRU().Expiration(cacheTTL).Build(),
+	}
+}
+
+// GetUser calls the Entity Registry with cluster auth and returns an user.
+func (is IS) GetUser(ctx context.Context, req *ttnpb.GetUserRequest) (*ttnpb.User, error) {
+	if is.userCache.Has(req.UserIds) {
```
adriansmares: This races between `Has` and `Get` - use only `Get`.

Another problem is that this doesn't take into account tenant boundaries - you will override the `adrian` in tenant 1 with the `adrian` of tenant 2. Use `unique.ID` as a cache key.

pkg/packetbrokeragent/is.go
```diff
@@ -0,0 +1,118 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return usr.(*ttnpb.User), nil
+	}
+	registry, err := is.newUserRegistryClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+	usr, err := registry.Get(ctx, req, is.WithClusterAuth())
```
adriansmares, line 78: We should cache errors too, with a different TTL between successful calls and error calls. Check our code base for `ErrorTTL` - there are a couple of examples on how to do this - I would just have a struct with both the entity and the error code.

pkg/packetbrokeragent/is.go
```diff
@@ -0,0 +1,118 @@
...
+
+	if err := is.userCache.Set(usr.GetIds().GetUserId(), usr); err != nil {
+		return nil, err
+	}
+	return usr, nil
+}
+
+// GetOrganization calls the Entity Registry with cluster auth and returns an organization.
+func (is IS) GetOrganization(ctx context.Context, req *ttnpb.GetOrganizationRequest) (*ttnpb.Organization, error) {
+	if is.orgCache.Has(req.OrganizationIds) {
```
adriansmares: Same comments as for the user part.

adriansmares (commented):
Great progress.

pkg/packetbrokeragent/grpc_gspba.go
```diff
@@ -157,16 +166,44 @@ func (s *gsPbaServer) UpdateGateway(ctx context.Context, req *ttnpb.UpdatePacket
...
+	if ttnpb.HasAnyField(req.FieldMask.GetPaths(), "administrative_contact") {
+		// Only replaces the administrative contact if no error happens.
+		adminUsr, err := s.fetchUserFromContact(
+			ctx, req.Gateway.GetAdministrativeContact(),
+			func(o *ttnpb.Organization) *ttnpb.OrganizationOrUserIdentifiers { return o.GetAdministrativeContact() },
+		)
+		if err == nil {
+			adminContact = &packetbroker.ContactInfo{Name: adminUsr.Ids.UserId, Email: adminUsr.PrimaryEmailAddress}
+		} else if !errors.IsNotFound(err) {
+			logger.WithField("gatewayID", req.Gateway.Ids).WithError(err).Warn("Failed to fetch administrative contact")
```
adriansmares: `snake_case` for log fields. I also don't think we should log this error for cases in which the email is not validated.

pkg/packetbrokeragent/grpc_gspba.go
```diff
@@ -157,16 +166,44 @@ func (s *gsPbaServer) UpdateGateway(ctx context.Context, req *ttnpb.UpdatePacket
...
+	if ttnpb.HasAnyField(req.FieldMask.GetPaths(), "technical_contact") {
+		// Only replaces the technical contact if no error happens.
+		techUsr, err := s.fetchUserFromContact(
+			ctx, req.Gateway.GetTechnicalContact(),
+			func(o *ttnpb.Organization) *ttnpb.OrganizationOrUserIdentifiers { return o.GetTechnicalContact() },
+		)
+		if err == nil {
+			techContact = &packetbroker.ContactInfo{Name: techUsr.Ids.UserId, Email: techUsr.PrimaryEmailAddress}
+		} else if !errors.IsNotFound(err) {
+			logger.WithField("gatewayID", req.Gateway.Ids).WithError(err).Warn("Failed to fetch technical contact")
```
adriansmares: `snake_case` and same error comment.

pkg/packetbrokeragent/grpc_gspba.go
```diff
@@ -211,3 +248,41 @@ func (s *gsPbaServer) UpdateGateway(ctx context.Context, req *ttnpb.UpdatePacket
...
+
+	usr, err := s.entityRegistry.GetUser(ctx, &ttnpb.GetUserRequest{
+		UserIds:   usrID,
+		FieldMask: ttnpb.FieldMask("ids", "primary_email_address", "primary_email_address_validated_at"),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if usr.PrimaryEmailAddressValidatedAt == nil {
```
adriansmares: We could force this check to be done in the caller code in order to avoid skipping all of the failed precondition error codes.

pkg/packetbrokeragent/is.go
```diff
@@ -0,0 +1,136 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/log"
+	"go.thethings.network/lorawan-stack/v3/pkg/ttnpb"
+	"go.thethings.network/lorawan-stack/v3/pkg/unique"
+	"google.golang.org/grpc"
+)
+
+var (
+	cacheSize     = 2048
+	cacheTTL      = 5 * time.Minute
+	cacheErrorTTL = 10 * time.Second
```
adriansmares: This can be longer, like a minute.

pkg/packetbrokeragent/is.go
```diff
@@ -0,0 +1,136 @@
...
+	expire := cacheTTL
+	usr, err := registry.Get(ctx, req, is.WithClusterAuth())
+	if err != nil {
+		expire = cacheErrorTTL
+	}
+
+	// Caches error in order to avoid calls in succession, time to live is significantly lower when an error occurs.
+	item := &cacheItem[*ttnpb.User]{entity: usr, err: err}
+	if err := is.cache.SetWithExpire(unique.ID(ctx, item.entity), item, expire); err != nil {
+		log.FromContext(ctx).WithError(err).Error("Failed to cache User")
```
adriansmares: `user`.

pkg/packetbrokeragent/is.go
```diff
@@ -0,0 +1,136 @@
...
+	expire := cacheTTL
+	org, err := registry.Get(ctx, req, is.WithClusterAuth())
+	if err != nil {
+		expire = cacheErrorTTL
+	}
+
+	// Caches error in order to avoid calls in succession, time to live is significantly lower when an error occurs.
+	item := &cacheItem[*ttnpb.Organization]{entity: org, err: err}
+	if err := is.cache.SetWithExpire(unique.ID(ctx, item.entity), item, expire); err != nil {
+		log.FromContext(ctx).WithError(err).Error("Failed to cache Organization")
```
adriansmares: `organization`.

adriansmares (approved):
This LGTM. Has it been tested with the staging PB environment ?

nicholaspcr:
> This LGTM. Has it been tested with the staging PB environment ?

It was tested on the initial implementation (without the cache). What was changed doesn't impact the interaction but I'll test it again before merging it.

nicholaspcr:
> > This LGTM. Has it been tested with the staging PB environment ?
> 
> It was tested on the initial implementation (without the cache). What was changed doesn't impact the interaction but I'll test it again before merging it.

Tested it again and everything is working, including the cache.

@johanstokking @KrishnaIyer requesting another code review

nicholaspcr:
@johanstokking and @KrishnaIyer requesting a review again.

pkg/gatewayserver/upstream/packetbroker/packetbroker.go
johanstokking: So we keep supporting both here?

pkg/packetbrokeragent/grpc_gspba.go
```diff
@@ -157,16 +163,44 @@ func (s *gsPbaServer) UpdateGateway(ctx context.Context, req *ttnpb.UpdatePacket
 		}
 	}
 
+	var adminContact, techContact *packetbroker.ContactInfo
 	if ttnpb.HasAnyField(req.FieldMask.GetPaths(), "contact_info") {
-		adminContact, techContact := toPBContactInfo(req.Gateway.GetContactInfo())
-		updateReq.AdministrativeContact = &packetbroker.ContactInfoValue{
-			Value: adminContact,
+		adminContact, techContact = toPBContactInfo(req.Gateway.GetContactInfo()) // nolint: staticcheck
+	}
```
johanstokking: Just verifying that it is essential that we carry `contact_info` from IS to GS to PBA?

pkg/gatewayserver/upstream/packetbroker/packetbroker.go
adriansmares: Yes, I would say so, since the fields are only deprecated, not removed. I would say that existing devices which have their contacts in `contact_info` but did not migrate to `*_contact` should keep on working.

pkg/packetbrokeragent/grpc_gspba.go
nicholaspcr: It isn't essential as the field has been deprecated.

There was a [previous discussion](https://github.com/TheThingsNetwork/lorawan-stack/pull/6430#discussion_r1291305443) in this PR which the topic wasn't exactly your question but it does bring up the reason as to why pass the `contact_info` around. The main argument is that while the field is deprecated the expected behavior was to have it propagated between these components, so it makes sense to have it included until we remove it completely.

pkg/gatewayserver/upstream/packetbroker/packetbroker.go
johanstokking: Why don't we migrate this and deprecate the field?

pkg/packetbrokeragent/grpc_gspba.go
johanstokking: Yes so it is deprecated, but apparently there's still valuable information in it. Why aren't we migrating this?

pkg/gatewayserver/upstream/packetbroker/packetbroker.go
adriansmares: I'll continue in the comment below in order to avoid having two threads about the same subject.

pkg/packetbrokeragent/grpc_gspba.go
adriansmares: So I was wrong, my bad. I thought that we still need this information, but that's not the case: https://github.com/TheThingsNetwork/lorawan-stack/pull/4996 already migrated the contacts on a best effort basis, and I should've checked that PR first. Notifications completely skip `contact_info` today, so it's really legacy data that is not used internally anymore (with a small exception for users).

We can already drop the reference to `contact_info` in this PR, and in a subsequent tech debt PR start removing the storage of the fields. We can keep the entity `contact_info` as a readonly view of the `technical_contact` and `administrative_contact` fields, while disallowing updates.

nicholaspcr: >So I was wrong, my bad. I thought that we still need this information, but that's not the case: https://github.com/TheThingsNetwork/lorawan-stack/pull/4996 already migrated the contacts on a best effort basis, and I should've checked that PR first.

Same, specially since the PR mentioned was linked on the discussion issue and I should've caught that. Regardless, will work on removing the `contact_info` in this PR  and will create the issue for the tech_debt suggested by Adrian

nicholaspcr: Marking this as solved since #6515 was created

nicholaspcr: And in addition to that the `contact_info` references were removed

nicholaspcr:
@johanstokking requesting your review again

johanstokking (approved):
Awesome

KrishnaIyer:
@nicholaspcr: This also needs testing steps.

nicholaspcr:
@KrishnaIyer 
I would say that all the alterations regarding the behavior change done in the Stack are validated by the new tests and therefore this should not be a part of the milestone's tests before release. The reason I say all this is that because we are only validating if the packetbroker properly received the fields specified in the API and if the packetbroker processed them.

The way I went about testing this is not completely replicable since I used logs which aren't present in the code on production, basically logged the fields which should've been changed after calling IS. I did configure the local Stack deployment to interact with Packetbroker on staging as detailed in https://discuss.thethingsindustries.com/t/packet-broker-staging-environment-for-local-development-use/983

Previously we were under the impression that a gateway's ContactInfo was being sent to the packetbroker but that wasn't the case since GS didn't provide the fields to the UpdateGateway request. Now the new Admin|Tech contact fields are fetched and propagated and that is validated via unit tests and the manual tests with the logs mentioned in the paragraph above.

nicholaspcr:
In addition to what has been written I tried to look for some way to validate the request body received by the packetbroker (like a debug log), that way we could just link the Stack to the packetbroker on staging and validate that but I'm still not super familiar with the packetbroker. Maybe @adriansmares has a suggestion on how to test this.

adriansmares:
Well `staging1` is connected to PB staging, so it is possible to just use the `https://staging-mapper.packetbroker.net/api/v2/gateways/netID=000013,tenantID={tenant ID here},id={gateway ID here}` URL to check the contact information. I see that my gateways do have the information, so I am pretty optimistic. This can be tested with a newly registered gateway for best repeatability.

nicholaspcr:
Thanks. Will do it locally and then write the test steps then.

KrishnaIyer:
I've tested this now with a newly connected "gateway".

https://staging-mapper.packetbroker.net/api/v2/gateways/netID=000013,tenantID=tti,id=eui-1111111111111113
```json
{"netID":"000013","tenantID":"tti","id":"eui-1111111111111113","clusterID":"staging1.cloud.thethings.industries","updatedAt":"2023-09-08T12:42:22.467757Z","online":true,"administrativeContact":{"name":"krishna","email":"krishna@thethingsindustries.com"},"technicalContact":{"name":"krishna","email":"krishna@thethingsindustries.com"},"frequencyPlan":{"region":"EU_863_870","loraMultiSFChannels":[868100000,868300000,868500000,867100000,867300000,867500000,867700000,867900000]},"rxRate":0,"txRate":0}
```


Screenshot 2023-09-08 at 14 44 36


@nicholaspcr: Please confirm the contents of the JSON and I can report to CE. I see that the `techincal_contact` is set.

nicholaspcr:
Yeah that matches with what is expected, thanks for doing it so fast.

PR #6444, opened 2023-08-04
Admins no longer triggers client creation notification

nicholaspcr:
#### Summary

This small PR fixes the bug found where a Client created by an admin where its not specified its state during its creation, will trigger the request to be approved by one of the tenant's admins. This is not the expected behavior since the user who created the Client is an admin himself.

#### Changes

- Add admin check in the email notification conditional.

#### Testing

Local test. 

In regards to tests that catch this sort of behavior, from what I checked (other unit tests and the integration test) we don't have tests that validate the contents of the emails generated by registry operations. Since this is something interesting to have, I'll create an issue to add those validations on the clicrud integrations test that we have (will add the triage label just to check before working on it). 

Steps for manually testing this bug fix:
- Create an OAuth Client with an admin account
  - Do not specify its state during its creation, simply giving it a name and nothing else.
- Validate if any emails requesting for an admin's approval were sent (there should be none)
- Validate the state of the created client (it should be approved)

##### Regressions

This is a bug fix.

#### Notes for Reviewers

Didn't do the tests but in the Test section of this PR there is a step to remedy the lack of tests in this regard

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ For details about compatibility between different releases, see the **Commitment
 
 ### Fixed
 
+- OAuth clients created by an admin no longer trigger an email requesting approval from one of the tenant's admins.
+
 ### Security
 
 ## [3.27.0] - 2023-07-31
--- a/pkg/identityserver/client_registry.go
+++ b/pkg/identityserver/client_registry.go
@@ -159,7 +159,7 @@ func (is *IdentityServer) createClient( //nolint:gocyclo
 		return nil, err
 	}
 	apikey := authInfo.GetApiKey().GetApiKey()
-	if cli.State == ttnpb.State_STATE_REQUESTED {
+	if !createdByAdmin && cli.State == ttnpb.State_STATE_REQUESTED {
 		go is.notifyAdminsInternal(ctx, &ttnpb.CreateNotificationRequest{
 			EntityIds:        req.GetClient().GetIds().GetEntityIdentifiers(),
 			NotificationType: "client_requested",


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ For details about compatibility between different releases, see the **Commitment
 ### Removed
 
 ### Fixed
+- OAuth clients created by an admin no longer trigger an email requesting approval from one of the tenant's admins.
```
adriansmares, line 25: New line before the header.

pkg/identityserver/client_registry.go
```diff
@@ -126,6 +126,8 @@ func (is *IdentityServer) createClient( //nolint:gocyclo
 		req.Client.StateDescription = "admin approval required"
 		req.Client.SkipAuthorization = false
 		req.Client.Endorsed = false
+	} else {
```
adriansmares: I don't think this is the right decision here. It means that as an admin you can only create approved clients, which is a bit weird - why do we have the fields if they cannot be set ?

I think we should just avoid sending the email notification if the entity is created by an admin.

nicholaspcr: I also thought about that but what bothered me was the possibility of an admin setting the OAuth client to pending when it can be approved by the creator. 

The two solutions that I see which are implementable on the back-end are:
- Set a conditional on the trigger for the email based on the state of the client and the state of who created the client.
- Determine default values for the client based on the state of the user requesting the creation of the client. 
- Third option that I thought about it while writing this, is this really a backend bug? an alternative could be limiting the operation from being done on the front-end, make it so that the admin user has two options: request approval from other admins or just have the client approved.

The second option seemed better to me but now I'm considering the third option, let me know what you think @adriansmares

nicholaspcr: just a note, the two options are basically the same thing as one conditional derives from the other. I'm just talking about the flow on which restrictions are imposed on the action for sending the email. 

The only real difference that the second option would provide is that an admin user could request the approval of other admins if he wished for it, even if that makes no sense

adriansmares: It doesn't matter if the OAuth client is created via the Console - you can do the same thing via the CLI, and also via the API directly. It is not the fault of any frontend that the backend decides to say, explicitly, `admin x is not an admin and requests that you approve his OAuth client`, which is a blatant error - the backend via email explicitly says false information.

The check there does not contain default values - there are no default values for the IS RPCs, which explicitly make a difference between `Create` and `Update` operations such that the semantics of a `Create` operation is that all of the provided values will be used while creating the entity.

The check there exists in order to force the approval of these OAuth clients when they are created by plain users - that's an override, not a default. No matter what you do explicitly as a normal user, you cannot skip this override.

Why an admin may want to create a `STATE_REQUESTED` client is not really relevant to this bug - the point is that the email should never fire in such situations. The admin didn't request any approval while creating the client - the backend just decided to spam every admin with a notification which contained false information, which is a bug.

The correct behavior is for the email to not fire when the client is created by an admin. The changes proposed in this PR basically make it impossible for an admin to create an OAuth client with different approval states.

nicholaspcr: I agree with pretty much everything that you said. 

>The only real difference that the second option would provide is that an admin user could request the approval of other admins if he wished for it, even if that makes no sense

Indeed the note I left is wrong as with the current changes I'm always overwriting the values of the client.

I honestly believe that I conveyed what I meant wrongly.
1. By using the word `default values`. Essentially I meant to convey that the conditional of being an admin sets the hard values to the client itself instead of limiting the procedure to send the email. So I referenced those hard coded values as `default` but they are meant to restrict the possible actions based on the type of user.

>The check there does not contain default values - there are no default values for the IS RPCs, which explicitly make a difference between Create and Update operations such that the semantics of a Create operation is that all of the provided values will be used while creating the entity.
2. Initially I was under the impression that an admin creating an already approved client is better than being able to create something not approved in a general way. But reading the comment referenced above I agree that it should use all the fields provided on the request's body and limit triggers(send the email)based on both the request data and user state.

The main point of the first comment that I made was more focused around the other in which should affect the way we handle the conditionals, in my head it made more sense to take the state of the user making the request and reflect certain behavior under the client data present on the body and take the email restrictions trigger upon the conditional of the client data values. But that isn't how we do these checks in other areas of the code base and as such your suggestion makes more sense.

nicholaspcr: Regarding the topic of being it done via the console instead of doing it on the back-end, the idea was in fact to restrict instead of eliminating the ability for admins to send that email to other admins, this only makes sense because I was not considering the contents of the email itself but the idea of an admin requesting the approval of other admins so the email message would have to change.

So it would be a possibility which isn't encouraged by the console but still possible. Reflecting on it, it is just an unnecessary change for overthinking the interactions between admins

nicholaspcr: Marking this as resolved since now there is only a check on the email notification trigger. Anything else related to the changes in the PR, like tests, are mentioned in the updated PR description.

adriansmares (approved):

PR #6450, opened 2023-08-10
Ensure complete identifiers are available in downlink events

adriansmares:
#### Summary

This PR addresses a bug in which the `ttnpb.ApplicationUp` generated in the downlink queued / acked / nacked / failed / sent events did not contain the full set of end device identifiers or the received at timestamp.

#### Changes

- Use the registry identifiers instead of the caller provided identifiers while registering events and generating uplinks.
- Propagate the received at timestamp.

#### Testing

Local testing. The feature can be tested further by attempting to push/replace the downlink queue and observing that the `received_at` is available, and that the EUIs are part of the `ids`.

##### Regressions

N/A. This fixes an inconsistency in our message filling behavior, but does not change the existing logic significantly.

#### Checklist

### DIFF
--- a/pkg/applicationserver/applicationserver.go
+++ b/pkg/applicationserver/applicationserver.go
@@ -754,16 +754,16 @@ func (as *ApplicationServer) downlinkQueueOp(ctx context.Context, ids *ttnpb.End
 				}
 			}
 
-			registerReceiveDownlinks(ctx, ids, items)
+			registerReceiveDownlinks(ctx, dev.Ids, items)
 
 			mask, err := as.attemptDownlinkQueueOp(ctx, dev, link, peer, downlinkQueueOperation{
 				Items:     items,
 				Operation: op,
 				ResultFunc: func(decrypted, encrypted []*ttnpb.ApplicationDownlink, err error) {
 					if err != nil {
-						as.registerDropDownlinks(ctx, ids, decrypted, now, err)
+						as.registerDropDownlinks(ctx, dev.Ids, decrypted, now, err)
 					} else {
-						as.registerForwardDownlinks(ctx, ids, decrypted, encrypted, now)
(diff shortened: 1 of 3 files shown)


pkg/applicationserver/applicationserver.go
```diff
@@ -753,16 +754,16 @@ func (as *ApplicationServer) downlinkQueueOp(ctx context.Context, ids *ttnpb.End
 				}
 			}
 
-			registerReceiveDownlinks(ctx, ids, items)
+			registerReceiveDownlinks(ctx, dev.Ids, items)
```
adriansmares, line 757: The registry always contains the full set of identifiers for an end device, while the caller provided `ids` may be only minimalistic (application ID + end device ID).

pkg/applicationserver/applicationserver.go
```diff
@@ -1016,7 +1023,7 @@ func (as *ApplicationServer) handleJoinAccept(ctx context.Context, ids *ttnpb.En
 				SkipSessionKeyIDs: [][]byte{items[0].SessionKeyId},
 				ResultFunc: func(decrypted, _ []*ttnpb.ApplicationDownlink, err error) {
 					if err != nil {
-						as.registerDropDownlinks(ctx, ids, decrypted, err)
+						as.registerDropDownlinks(ctx, dev.Ids, decrypted, receivedAt, err)
```
adriansmares, line 1026: This `receivedAt` is slightly misleading - it is the time at which the Join Accept which contains this invalidate downlink was received by the Application Server, not the original timestamp at which the downlink was enqueued.

Do we want to keep using the `nil` timestamps, because we cannot recover the original timestamp, or is it fine to always fill it with the closest timestamp to the message ?

johanstokking: I think it's fine to set it to the timestamp of a message that triggered the event, for application layer ordering purposes.

johanstokking (approved):

PR #6458, opened 2023-08-15
Pick device address prefixes in a weighted manner

adriansmares:
#### Summary

This PR makes the device address generation take into account the number of bits assigned for the actual device address. This ensures that when blocks of different sizes are used, the size of the block is used as a weight for the chosen block.

#### Changes

- Use a weighted random algorithm for the device address prefix selection.

#### Testing

Unit testing.

##### Regressions

N/A.

#### Checklist

- [ ] The steps/process to test this feature are clearly explained including testing for regressions.

### DIFF
--- a/pkg/networkserver/networkserver.go
+++ b/pkg/networkserver/networkserver.go
@@ -73,15 +73,27 @@ func makeWindowDurationFunc(d time.Duration) windowDurationFunc {
 }
 
 // newDevAddrFunc is a function, which is used by Network Server to derive new DevAddrs.
-type newDevAddrFunc func(ctx context.Context, dev *ttnpb.EndDevice) types.DevAddr
+type newDevAddrFunc func(ctx context.Context) types.DevAddr
 
 // makeNewDevAddrFunc returns a newDevAddrFunc, which derives DevAddrs using specified prefixes.
 func makeNewDevAddrFunc(ps ...types.DevAddrPrefix) newDevAddrFunc {
-	return func(ctx context.Context, dev *ttnpb.EndDevice) types.DevAddr {
+	weights := make([]int64, len(ps))
+	totalWeight := int64(0)
+	for i, p := range ps {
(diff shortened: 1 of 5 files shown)


pkg/networkserver/networkserver.go
```diff
@@ -77,11 +78,16 @@ type newDevAddrFunc func(ctx context.Context, dev *ttnpb.EndDevice) types.DevAdd
...
+	w := sampleuv.NewWeighted(weights, nil)
 	return func(ctx context.Context, dev *ttnpb.EndDevice) types.DevAddr {
 		var devAddr types.DevAddr
-		rand.Read(devAddr[:])
-		p := ps[random.Int63n(int64(len(ps)))]
-		return devAddr.WithPrefix(p)
+		_, _ = rand.Read(devAddr[:])
+		idx, _ := w.Take()
+		return devAddr.WithPrefix(ps[idx])
 	}
```
johanstokking: This is missing tests, we don't need floating point operations (one bitshift will do) and I don't think we need a dependency for this.

We can just:

1. Loop over `ps` and put them in a new slice with a field `weight` that is `1 << (32-Length)`, and keep a `totalWeight`
2. Pick a random number `r` between 0 and `totalWeight` (inclusive)
3. Loop over the slice, decrease `r` by the `weight` in the slice, and when it goes `0` or lower, take that `ps`

johanstokking (changes requested):

pkg/networkserver/networkserver.go
adriansmares: Based on https://github.com/packetbroker/router/blob/5645a6075d9e319691b9324bb6873dd0a697fbcc/internal/pubsub/grpc/server.go#L52 , I generated the number in the `[0, totalWeight)` range, and actually the `ps` which made `r` negative.

johanstokking (approved):

PR #6465, opened 2023-08-16
Refactor ping pong tests

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/actions/runs/5388909842/job/14586179127?pr=3745

#### Changes

- Refactor the flow used to detect that the client has missed responding to a PING. See inline comment for details.
- Use `WriteControl` instead of `WriteMessage` for ping and pong messages. This is how these messages are supposed to be sent.
- Simplify how we start HTTP servers.
- Remove superfluous `break`s.
- Ensure that we cancel contexts when test cases end.

#### Testing

`go test -tags tti -count 3000 -failfast -v -run TestPingPong`

I warn you, it will take a lot of time. But it was worth it as we waste 15 minutes on every single time this damn flake is triggered.

##### Regressions

This touches the code used to detect missing `PONG`s, but the unit tests should cover this issue. The race condition is virtually impossible to reproduce by hand, and requires thousands of tests until it occurs.

#### Checklist

### DIFF
--- a/pkg/gatewayserver/io/ws/ws.go
+++ b/pkg/gatewayserver/io/ws/ws.go
@@ -321,7 +321,7 @@ func (s *srv) handleTraffic(w http.ResponseWriter, r *http.Request) (err error) 
 			case <-conn.Context().Done():
 				return
 			case <-pingTicker.C:
-				if err := ws.WriteMessage(websocket.PingMessage, nil); err != nil {
+				if err := ws.WriteControl(websocket.PingMessage, nil, time.Time{}); err != nil {
 					logger.WithError(err).Warn("Failed to send ping message")
 					return err
 				}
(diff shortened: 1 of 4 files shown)


pkg/gatewayserver/io/ws/ws.go
```diff
@@ -321,23 +335,24 @@ func (s *srv) handleTraffic(w http.ResponseWriter, r *http.Request) (err error)
 			case <-conn.Context().Done():
 				return
 			case <-pingTicker.C:
-				if err := ws.WriteMessage(websocket.PingMessage, nil); err != nil {
```
adriansmares, line 324: In the old code, notice that between the moment in which we finish sending the ping, and the moment in which we increment the number of missing pongs, anything can happen, including receiving a pong. This is the crux of the flake - if you have zero missed pongs, you receive a pong so you replace zero with zero, you will accidentally create one missing pong.

The error flake is created by the following sequence of events.

1. Send a ping to the client: `missingPongs == -1`.
3. Receive a pong from the client: `missingPongs == 0`.
4. Increment the number of missing pongs: `missingPogs == 1`.
5. Send a ping to the client: `missingPongs == 1`.
6. Increment the number of missing pongs: `missingPongs == 2`.
7. You are now disconnecting the gateway while in reality only one pong is missing.

The proposed solution is to increment the number of missing pongs before sending the pings, and using the `missingPongs` counter as the difference between the outstanding pings and pongs. The counter is not allowed to go negative, as it implies that we have received more pongs than pings, which is a signal of client implementation error.

KrishnaIyer (approved):
Logic looks fine to me.

johanstokking (approved):
Hallelujah

PR #6474, opened 2023-08-18
Fix more Go test races

adriansmares:
#### Summary

References https://github.com/TheThingsNetwork/lorawan-stack/actions/runs/5902633208/job/16011029191?pr=6459

#### Changes

- Use atomics to increment call counters in the telemetry tests, as the handlers can be called in parallel.
- Do not use fixed ports for component gRPC listeners in tests. We don't have a global database of these listeners, and its easy to have overlapping tests.
- Do not use fixed ports for component HTTP listeners in tests. Same reasoning as before.
- Upgrade `golang.org/x/exp` to the last version.

#### Testing

Lots of local test runs.

##### Regressions

N/A. This changes only test code.

#### Checklist

- [ ] The steps/process to test this feature are clearly explained including testing for regressions.


pkg/band/channel_mask.go
```diff
@@ -446,8 +446,8 @@ func generateChMask72Generic(currentChs, desiredChs []bool, atomic bool) ([]ChMa
 		// We also sort the masks descending on the number of enabled channels in order to ensure
 		// that intermediary states through which the end device goes while parsing the masks
 		// are valid (i.e. the total number of enabled channels is always greater than 0).
-		slices.SortFunc(pairs, func(a, b ChMaskCntlPair) bool {
-			return trueCount(a.Mask[:]...) >= trueCount(b.Mask[:]...)
+		slices.SortFunc(pairs, func(a, b ChMaskCntlPair) int {
```
adriansmares, line 449: Go 1.21 will bring the `slices` and `maps` packages to the standard library, and the variant adopted there uses an `int` returning comparator instead of a `bool` one. The change here allows us to easily move to the standard library implementation once we move to Go 1.21.

KrishnaIyer (approved):
Nice. Old bad habits :)

johanstokking (approved):
Nice archeology work

PR #6478, opened 2023-08-23
Show gateway locations retrieved from status messages in the map

kschiffer:
#### Summary
This PR adds functionality to display gateway locations from status messages in the gateway map in the Console.
It also streamlines how markers are displayed for end devices and gateways (both with a tooltip that gives information about the location source and trust)

image
(removed the exact lat/long for privacy reasons)

#### Changes

- Add store logic to paste status message locations into the gateway store
- Fix issue with prop types
- Use same utility function for generating markers from antenna locations

#### Testing

This can be tested as follows:

In the Console, navigate to a Gateway that uses GPS location in its status messages. On staging, this if for example:
https://tti.staging1.cloud.thethings.industries/console/gateways/kerlink-ibts-compact-1

The location in the overview should appear as soon as a status message is received (or already initially, if the event data retention included a status message)

#### Notes for Reviewers

As of now, all status updates are sent through insecure channels, which is why they cannot be trusted entirely. This is also why these locations are currently not used to update the gateway's location entry. I'm still in favor of at least displaying the location in the Console with a little warning and also to use them as a starting point when setting the location manually.

cc @adriansmares 

#### Checklist


pkg/webui/console/containers/packet-broker-networks-table/index.js
ryaplots, line 1: Are these changes intentional? Seems like something from a previous PR.

pkg/webui/console/store/middleware/logics/gateways.js
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import { createLogic } from 'redux-logic'
+import { eq } from 'lodash'
```
ryaplots: is this being used anywhere? I don't see it in the changes.

pkg/webui/console/store/reducers/gateways.js
```diff
@@ -12,12 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { merge } from 'lodash'
+
```
ryaplots: same comment here

ryaplots (approved):
Looks good! Just some small comments.

pkg/webui/console/containers/packet-broker-networks-table/index.js
kschiffer: No, this is intentional because otherwise the location in the location form will update every time a new status message is received.

KrishnaIyer:
> Manual testing.

Please add the steps to test this feature, including instructions for the setup.

kschiffer:
Updated the description.

KrishnaIyer:
Thanks!

PR #6479, opened 2023-08-23
Support notifications in the Console

### LINKED ISSUE #5458: Add support for reading notifications in the Console
#### Summary

As a follow-up of https://github.com/TheThingsNetwork/lorawan-stack/issues/4935 we should add Console support for reading notifications and marking them as seen.

#### Why do we need this?

So that users can also see and manage notifications for which The Things Stack doesn't send emails.

API: `GET /users/{user_id}/notifications` to list notifications and `PATCH /users/{user_id}/notifications` to update the status of notifications. See `notification_service.proto` for more details.

@kschiffer should probably come up with the UI/UX for this, but I'd like to see something similar to GitHub notifications:

- A "bell" icon with an indicator for unread notifications
- A table view of all (not archived) notifications, and on each row:
	- Indicator if status is unread
	- Checkbox for bulk update (status can be updated to unseen / seen / archived)
	- Icon for entity type
	- Entity ID
	- Notification type (until we can sync the "registry" with the frontend, I think we should just use the raw text here)
	- Time (ago)
	- Some way for the user to get the details (the full JSON) of the notification

No. Let's start with some UI/UX ideas from @kschiffer.

ryaplots:
#### Summary

Closes #5458 

#### Changes

- Add notifications view
- Fetch notifications every 5 minutes
- Add more utility classes

<details>
<summary>Notifications screenshots</summary>
Screenshot 2023-08-23 at 16 00 59
Screenshot 2023-08-23 at 11 24 35
Screenshot 2023-08-23 at 11 24 50
</details>
<details>
<summary>Archive screenshots</summary>
Screenshot 2023-08-23 at 11 25 13
</details>
<details>
<summary>Responsive screenshots</summary>
Screenshot 2023-08-23 at 11 30 34
Screenshot 2023-08-23 at 11 30 49
</details>

#### Testing

To test notification view:
1. Login in the console.
2. Create an API key for each entity (gateways, applications, organizations).
3. Go to `console/notifications`.
4. See a list of notifications.
5. Click on a notifications with a blue dot (means unseen) and see the content of the notification and blue background for selected notification.
6. See the amount of unseen notifications in the blue bubble go down. 
7. Click on "Archive" and see notification disappearing from the list.
8. Click on "Mark all as read" and see all blue dots and unseen notifications bubble disappear.
9. Click on "See archived messages" and see the message you just archived in the list.
10. Click on "Unarchive" and see the notification disappear from the list.
11. Click on "See all messages" and see the notification is back in the list of notifications.

To test new notifications toast:
1. Go to any entity and create an API key, or add a collaborator.
2. Wait for at least 10 minutes and see a toast notification appear in the right bottom of the page saying "You have new notifications"

#### Notes for Reviewers

1. Let's decide here on the content for the notifications. The information I get from the backend is the following:
<details>
<summary>api key create</summary>
<code>
{
	"0": {
		"id": "df40d337-bfd8-4e8d-aa6b-4541a47b44f3",
		"created_at": "2023-08-16T12:31:23.987086Z",
		"entity_ids": {
			"application_ids": {
				"application_id": "darya"
			}
		},
		"notification_type": "api_key_created",
		"data": {
			"@type": "type.googleapis.com/ttn.lorawan.v3.APIKey",
			"id": "I337KND3ZIDFEUHTZ2T32UFULF4JVVQYLWEX6EI",
			"name": "not",
			"rights": [
				"RIGHT_APPLICATION_ALL"
			],
			"created_at": "2023-08-16T12:31:23.966830Z",
			"updated_at": "2023-08-16T12:31:23.966831Z"
		},
		"sender_ids": {
			"user_id": "admin"
		},
		"receivers": [
			"NOTIFICATION_RECEIVER_ADMINISTRATIVE_CONTACT"
		],
		"status_updated_at": "2023-08-16T12:31:23.987086Z"
	}
}
</code>
</details>
<details>
<summary>client requested</summary>
<code>
{
	"0": {
		"id": "78830284-c575-464b-8b13-eea55c4b01af",
		"created_at": "2023-08-16T10:23:11.320390Z",
		"entity_ids": {
			"client_ids": {
				"client_id": "triggetnot"
			}
		},
		"notification_type": "client_requested",
		"data": {
			"@type": "type.googleapis.com/ttn.lorawan.v3.CreateClientEmailMessage",
			"create_client_request": {
				"client": {
					"ids": {
						"client_id": "triggetnot"
					},
					"administrative_contact": {
						"user_ids": {
							"user_id": "darya"
						}
					},
					"technical_contact": {
						"user_ids": {
							"user_id": "darya"
						}
					},
					"secret": "PBKDF2$sha256$20000$BXM4H1FooPZqaQvo$bYvK0Aq_j8MgYkhf0XC8FMZkAzHNQaW6cDipWGSZ4j0",
					"state_description": "admin approval required",
					"grants": [
						"GRANT_AUTHORIZATION_CODE"
					],
					"rights": [
						"RIGHT_APPLICATION_ALL",
						"RIGHT_CLIENT_ALL",
						"RIGHT_GATEWAY_ALL",
						"RIGHT_ORGANIZATION_ALL",
						"RIGHT_USER_ALL"
					]
				},
				"collaborator": {
					"user_ids": {
						"user_id": "darya"
					}
				}
			}
		},
		"sender_ids": {
			"user_id": "darya"
		},
		"receivers": [
			"NOTIFICATION_RECEIVER_ADMINISTRATIVE_CONTACT"
		],
		"status_updated_at": "2023-08-16T10:23:11.320390Z"
	}
}
</code>
</details>
<details>
<summary>collaborator added or changed</summary>
<code>
{
	"0": {
		"id": "ad25f7b0-f29f-476d-a2e9-2990b6df5450",
		"created_at": "2023-08-16T13:10:14.388098Z",
		"entity_ids": {
			"application_ids": {
				"application_id": "darya2-0"
			}
		},
		"notification_type": "collaborator_changed",
		"data": {
			"@type": "type.googleapis.com/ttn.lorawan.v3.Collaborator",
			"ids": {
				"user_ids": {
					"user_id": "admin"
				}
			},
			"rights": [
				"RIGHT_APPLICATION_ALL"
			]
		},
		"sender_ids": {
			"user_id": "darya"
		},
		"receivers": [
			"NOTIFICATION_RECEIVER_ADMINISTRATIVE_CONTACT"
		],
		"status_updated_at": "2023-08-16T13:10:14.388098Z"
	}
}
</code>
</details>
<details>
<summary>entity state changed</summary>
<code>
{
	"0": {
		"id": "0dc087d8-9dbb-4511-bc4b-13d3ce720f48",
		"created_at": "2023-08-16T13:24:53.893772Z",
		"entity_ids": {
			"user_ids": {
				"user_id": "darya"
			}
		},
		"notification_type": "entity_state_changed",
		"data": {
			"@type": "type.googleapis.com/ttn.lorawan.v3.EntityStateChangedNotification",
			"state": "STATE_APPROVED"
		},
		"sender_ids": {
			"user_id": "admin"
		},
		"receivers": [
			"NOTIFICATION_RECEIVER_COLLABORATOR"
		],
		"status_updated_at": "2023-08-16T13:24:53.893772Z"
	}
}
</code>
</details>
<details>
<summary>password changed</summary>
<code>
{
	"0": {
		"id": "d7b0d00c-4ea6-4252-bdb9-f96a18b0f16f",
		"created_at": "2023-08-16T14:02:39.461763Z",
		"entity_ids": {
			"user_ids": {
				"user_id": "darya"
			}
		},
		"notification_type": "password_changed",
		"sender_ids": {
			"user_id": "darya"
		},
		"receivers": [
			"NOTIFICATION_RECEIVER_COLLABORATOR"
		],
		"status_updated_at": "2023-08-16T14:02:39.461763Z"
	}
}
</code>
</details>
<details>
<summary>user requested</summary>
<code>
{
	"3": {
		"id": "72a40b96-b5fb-4ab3-b1ca-302a1a5f34c6",
		"created_at": "2022-09-01T18:18:56.544863Z",
		"entity_ids": {
			"user_ids": {
				"user_id": "anton"
			}
		},
		"notification_type": "user_requested",
		"data": {
			"@type": "type.googleapis.com/ttn.lorawan.v3.CreateUserRequest",
			"user": {
				"ids": {
					"user_id": "anton"
				},
				"contact_info": [
					{
						"contact_method": "CONTACT_METHOD_EMAIL",
						"value": "abeitler@semtech.com"
					}
				],
				"primary_email_address": "abeitler@semtech.com",
				"password": "PBKDF2$sha256$20000$6vi-0PIQ1S5dYaKi$eetS9YYWOe4UHVfzpMJ73FipWc8EVQpaq8weTcoNbcI",
				"password_updated_at": "2022-09-01T18:18:56.518911135Z",
				"state_description": "admin approval required"
			},
			"invitation_token": "YJZEKJZ6EP24R2GD5BUO2AM3KTMOXAIIFRHP3B4BB6HLG3QZD4DA"
		},
		"status": "NOTIFICATION_STATUS_SEEN",
		"status_updated_at": "2023-08-16T14:30:46.961697Z"
	}
}
</code>
</details>

2. Where in the header should I put the link to the notifications for now?

#### Checklist


pkg/webui/components/button/index.js
```diff
@@ -220,7 +220,7 @@ const commonPropTypes = {
    */
   unstyled: PropTypes.bool,
   /** The html `value` prop passed to the <button /> element. */
-  value: PropTypes.string,
+  value: PropTypes.oneOfType([PropTypes.string, PropTypes.shape({})]),
```
kschiffer: Why?

pkg/webui/console/components/notifications/templates/api-key-changed.js
```diff
@@ -0,0 +1,159 @@
...
+
+import Message from '@ttn-lw/lib/components/message'
+
+import PropTypes from '@ttn-lw/lib/prop-types'
+
+import { getEntity } from '../utils'
+
+const m = defineMessages({
+  title: 'An API key of your {entityType} on your network has been changed',
+  greeting: 'Dear {receiverName},',
```
kschiffer: `greeting`, `rights` and `right` is now repeated many times, which will result in many identical messages. These should be outsourced to `sharedMessages`.

pkg/webui/console/components/notifications/templates/api-key-changed.js
```diff
@@ -0,0 +1,159 @@
...
+  body: 'An API key of your {entityType} <code>{id}</code> on your network has been changed.',
+  apikey: '<b>API Key ID:</b> <code>{apiKeyId}</code>',
+  rights: 'Rights:',
+  right: '<code>{right}</code>',
```
kschiffer: You don't need a message for that, it has no translatable content.

pkg/webui/console/components/notifications/templates/api-key-changed.js
kschiffer: I believe we can also strip the greeting/address altogether to keep it more concise and not imitating emails too much.

pkg/webui/console/components/notifications/templates/client-requested.js
```diff
@@ -0,0 +1,191 @@
...
+              {msg}
+            </Link.Anchor>
+          ),
+        }}
+      />
+    </>
+  )
+}
+
+ClientRequested.propTypes = {
```
kschiffer, line 122: There's a lot of duplication here as well which can be outsourced to the prop-types module.

kschiffer (changes requested):
Cutting my review off because I think this needs an iteration that addresses the code repetition. 
While the code works fine, I find that there is just way too much repetition with regard to how the templates work. Since the shape of the notifications is quite similar except for different messages and maybe slightly different markup, I think it would be good to refactor the code to use one templating component that accepts different messages and maybe a couple of configuration props but is then able to handle all notification templates at once, simply by switching the messages.

This will address the problem of repetition in messages, prop types, and markup and I think will serve us better in terms of maintainability as well.

I'm not through with the review but wanted to make this point so you can start with this already.

pkg/webui/components/button/index.js
ryaplots: Left over from refactoring. My bad.

pkg/webui/console/components/notifications/templates/api-key-changed.js
```diff
@@ -0,0 +1,137 @@
...
+  )
+}
+
+ApiKeyChangedPreview.propTypes = {
+  notificationData: PropTypes.shape({
+    data: PropTypes.shape({
+      id: PropTypes.string.isRequired,
+    }).isRequired,
+    entity_ids: PropTypes.shape({}).isRequired,
+  }).isRequired,
```
kschiffer: Please outsource this into the `propTypes` package, so it doesn't have to be redefined everytime.

```suggestion
  notificationData: PropTypes.notificationData.isRequired()
```

pkg/webui/console/components/notifications/templates/api-key-changed.js
```diff
@@ -0,0 +1,137 @@
...
+    body: m.body,
+    entities: sharedMessages.apiKeyId,
+    action: sharedMessages.viewLink,
+  }
+  const values = {
+    body: {
+      entityType: getEntity(entity_ids),
+      id: entity_ids[`${getEntity(entity_ids)}_ids`][`${getEntity(entity_ids)}_id`],
+      code: msg => <code>{msg}</code>,
+      b: msg => <b>{msg}</b>,
```
kschiffer: Instead of redefining these every time, I'd suggest defining a base set of markup (`b`, `code`, `lineBreak`) that can then be used throughout all notifications and then passing them by default to the rendering `<Message />` component of the template renderer.

pkg/webui/console/components/notifications/utils.js
```diff
@@ -0,0 +1,23 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import notificationMap from './constants'
+
+export const getNotification = notificationType => notificationMap[notificationType]
+
+export const getEntity = entityIds => {
+  const entity = Object.keys(entityIds)[0].split('_')[0]
```
kschiffer: I don't think it's good to base the entity string entirely off the object key here. It might break with new entities to be added in the future (e.g. if they have an underscore in their name). Another issue is that this results in "end device" being called just "device", which is against our [language guidelines](https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.27/DESIGN.md#general-considerations).

I believe it's better to just use a mapping instead:

```js
const idToEntityMap = {
  application_ids: 'application',
  device_ids: 'end device',
  gateway: 'gateway',
  …
}

const getEntity = ({ entity_ids }) => idToEntityMap[Object.keys(entity_ids)[0]] || 'entity'
```

pkg/webui/console/components/notifications/templates/entity-state-changed.js
```diff
@@ -0,0 +1,126 @@
...
+  const { data, entity_ids } = notificationData
+  const { formatMessage } = useIntl()
+
+  return (
+    <Message
+      content={m.body}
+      values={{
+        entityType: getEntity(entity_ids),
+        entityId: entity_ids[`${getEntity(entity_ids)}_ids`][`${getEntity(entity_ids)}_id`],
+        state: capitalizeMessage(formatMessage({ id: `enum:${data.state}` })),
```
kschiffer:
```suggestion
        state: formatMessage({ id: `enum:${data.state}` }),
```
Not a good idea to tinker with a translated message as you cannot know whether the capitalization would make sense in another language. Also, I don't think it necessary here to capitalize it anyway.

pkg/webui/console/components/notifications/templates/entity-state-changed.js
```diff
@@ -0,0 +1,126 @@
...
+
+import ContentTemplate from './template'
+
+const m = defineMessages({
+  title: 'The state of your {entityType} has been changed.',
+  body: 'The state of the {entityType} <code>{entityId}</code> on your network has been changed to "{state}".',
+  link: 'You can view this <Link>here</Link>.',
+})
+
+const EntityStateChangedtPreview = ({ notificationData }) => {
```
kschiffer:
```suggestion
const EntityStateChangedPreview = ({ notificationData }) => {
```

pkg/webui/console/components/notifications/templates/entity-state-changed.js
```diff
@@ -0,0 +1,126 @@
...
+        entityType: getEntity(entity_ids),
+        entityId: entity_ids[`${getEntity(entity_ids)}_ids`][`${getEntity(entity_ids)}_id`],
+        state: capitalizeMessage(formatMessage({ id: `enum:${data.state}` })),
+        code: msg => msg,
+      }}
+    />
+  )
+}
+
+EntityStateChangedtPreview.propTypes = {
```
kschiffer:
```suggestion
EntityStateChangedPreview.propTypes = {
```

pkg/webui/lib/prop-types.js
```diff
@@ -350,4 +350,45 @@ PropTypes.routingPolicy = PropTypes.shape({
 
 PropTypes.entity = PropTypes.oneOf([APPLICATION, GATEWAY, ORGANIZATION, USER, CLIENT])
 
+PropTypes.clientNotificationData = PropTypes.shape({
```
kschiffer: I don't think we need one proptype for each notification type. Instead, let's define one generic `NotificationData` proptype that includes all possible props of all notification types and use that then throughout.

pkg/webui/console/containers/notifications/notification-content/index.js
```diff
@@ -0,0 +1,145 @@
...
+  isArchive,
+  selectedNotification,
+  setShowContent,
+  fetchItems,
+  setShowListColumn,
+  showListColumn,
+}) => {
+  const userId = useSelector(selectUserId)
+  const dispatch = useDispatch()
+  const isMobile = window.innerWidth < 768
```
kschiffer: If you have to use breakpoint values, please use the ones defined in `pkg/webui/constants/layout.js` instead. But please see my other comment on using CSS only.

pkg/webui/console/containers/notifications/index.js
```diff
@@ -0,0 +1,149 @@
...
+  seeArchived: 'See archived messages',
+  seeAll: 'See all messages',
+})
+
+const pageSize = 5
+const DEFAULT_PAGE = 1
+
+const pageValidator = page => (!Boolean(page) || page < 0 ? DEFAULT_PAGE : page)
+
+const NotificationsContainer = ({ setPage, page }) => {
```
kschiffer:
```suggestion
const Notifications = ({ setPage, page }) => {
```

Not really conventional to add `Container` to the component name.

pkg/webui/console/containers/notifications/index.js
```diff
@@ -0,0 +1,149 @@
...
+  const userId = useSelector(selectUserId)
+  const totalNotifications = useSelector(selectTotalNotificationsCount)
+  const dispatch = useDispatch()
+  const [showContent, setShowContent] = useState(false)
+  const [selectedNotification, setSelectedNotification] = useState(undefined)
+  const [showListColumn, setShowListColumn] = useState(true)
+  const [isNotifications, setIsNotifications] = useQueryState('isNotifications', 'true')
```
kschiffer:
```suggestion
  const [showArchived, setShowArchived] = useQueryState('archived', 'true')
```

pkg/webui/console/containers/notifications/index.js
```diff
@@ -0,0 +1,149 @@
...
+            page,
+          }),
+        ),
+      )
+    },
+    [dispatch, userId, page, isNotifications],
+  )
+
+  useEffect(() => {
+    fetchItems()
```
kschiffer: You are not keeping track of the fetching status, this causes a jagged experience where new results will appear significantly after e.g. a new page number is already rendered. Ideally, the old results should be replaced by overlayed by a spinner until the new results appear.

Additionally, the UI should always select the first notification (if any) by default after fetching.

Also, I realize I haven't made this clear from my screen designs, but there should not actually be classic pagination but rather infinite scrolling with automatic fetching of new messages while scrolling.

pkg/webui/console/containers/notifications/index.js
```diff
@@ -0,0 +1,149 @@
...
+
+const NotificationsContainer = ({ setPage, page }) => {
+  const userId = useSelector(selectUserId)
+  const totalNotifications = useSelector(selectTotalNotificationsCount)
+  const dispatch = useDispatch()
+  const [showContent, setShowContent] = useState(false)
+  const [selectedNotification, setSelectedNotification] = useState(undefined)
+  const [showListColumn, setShowListColumn] = useState(true)
+  const [isNotifications, setIsNotifications] = useQueryState('isNotifications', 'true')
+  const isMobile = window.innerWidth < 768
```
kschiffer: I'm sure there is a more CSS-centric solution to this. For example:

1. Set a `notification-selected`-class on the container whenever a notification is selected
2. Use media queries to hide the notification list whenever a notification is selected (and enlarging the notification body)
3. When clicking back while in mobile mode, deselect the notification and remove the `notification-selected`-class from the container

This way you can handle the responsiveness purely on the CSS side and maintain a proper separation of concerns.

pkg/webui/console/containers/notifications/index.js
```diff
@@ -0,0 +1,149 @@
...
+
+  const handleShowArchived = useCallback(async () => {
+    setPage(DEFAULT_PAGE)
+    setShowContent(false)
+    setIsNotifications(isNotifications === 'false' ? 'true' : 'false')
+  }, [setIsNotifications, isNotifications, setPage])
+
+  return (
+    <Row className={classNames(style.notificationsContainer, 'm-0')}>
+      {showListColumn && (
```
kschiffer: See my other comment on solving this CSS-only.

pkg/webui/console/containers/notifications/index.js
```diff
@@ -0,0 +1,149 @@
...
+  )
+
+  const handleShowArchived = useCallback(async () => {
+    setPage(DEFAULT_PAGE)
+    setShowContent(false)
+    setIsNotifications(isNotifications === 'false' ? 'true' : 'false')
+  }, [setIsNotifications, isNotifications, setPage])
+
+  return (
+    <Row className={classNames(style.notificationsContainer, 'm-0')}>
```
kschiffer: Consider native props first before modifying the grid with classes:
https://sealninja.github.io/react-grid-system/#row

pkg/webui/console/containers/notifications/index.js
```diff
@@ -0,0 +1,149 @@
...
+      {showListColumn && (
+        <Col md={4.5} className={classNames(style.notificationList, 'mt-cs-l', 'mb-cs-l')}>
```
kschiffer: `4.5` Wow, I didn't know you could do that 😂 . I think we should stick with full integers for consistency though.

pkg/webui/console/containers/notifications/notification-list/index.js
```diff
@@ -0,0 +1,162 @@
...
+      setSelectedNotification(notifications.find(notification => notification.id === id))
+      if (!isArchive) {
+        await dispatch(
+          attachPromise(updateNotificationStatus(userId, [id], 'NOTIFICATION_STATUS_SEEN')),
+        )
+        setTimeout(async () => {
+          await fetchItems()
+          await dispatch(attachPromise(getUnseenNotifications(userId)))
+        }, 300)
+      }
```
kschiffer: Each notification selection now triggers three separate requests. As far as I can tell, you should really only need a single request, which is the one that updates the seen status **if** the notification was unseen before. If that request returns a 200 you can be sure enough that the message was marked correctly and set it as such in the update reducer.

What's the reason for refetching all notifications?

Also using arbitrary timeouts like here is only going to set you up for race conditions in slower environments.

pkg/webui/console/containers/notifications/notification-list/index.js
```diff
@@ -0,0 +1,162 @@
...
+  )
+
+  const handleMarkAllAsSeen = useCallback(async () => {
+    await dispatch(
+      attachPromise(updateNotificationStatus(userId, unseenIds, 'NOTIFICATION_STATUS_SEEN')),
+    )
+    setTimeout(async () => {
+      await fetchItems()
+      await dispatch(attachPromise(getUnseenNotifications(userId)))
+    }, 300)
```
kschiffer: Also here, you shouldn't have to launch two more requests to get the new items and the number of unseen items. From the response to the update request, you can assume that there will be no unseen items and otherwise the notifications should be unaffected.

pkg/webui/console/containers/notifications/notifications.styl
```diff
@@ -0,0 +1,153 @@
...
+  position: relative
+  border: none !important
+  border-bottom: 1px solid #E4E4E4 !important
+  border-radius: 0 !important
+  box-shadow: none !important
+  min-height: 7rem
+  padding: $cs.l !important
+  justify-content: start !important
+  flex-direction: column !important
+  align-items: start !important
```
kschiffer: This is a lot of `!important` here. I don't want to be too strict about it but it is definitely a code smell. Any chance you can resolve this by using better selector specificity?

kschiffer (changes requested):
This still needs some rework in terms of code repetition, fetching logic, and mobile styles. See my comments for details.

pkg/webui/console/containers/notifications/index.js
ryaplots: I will go through all the grid elements and remove any unnecessary classes. But in this particular situation I am adding a border radius, min-height and margin which I believe is not possible with native props. Or am I missing something?

pkg/webui/console/containers/notifications/notification-list/index.js
ryaplots: The reason for fetching all notifications was to update the store with the correct status and therefore update the visual, since the update request does not return the ids that were changed. But you are right, it is not necessary. Thanks for the suggestion. Haven't thought about this in that way.

pkg/webui/console/containers/notifications/index.js
kschiffer: No that's fine. Personally, I'd be more inclined to use a separate container for that then to keep the containers as pure as possible, but it's not a must.

KrishnaIyer:
This is blocked until we have a clear map of the console redesign https://github.com/TheThingsIndustries/lorawan-stack/issues/2663

KrishnaIyer:
Moved to Phase 3 of the console redesign: https://github.com/TheThingsIndustries/product-management/issues/29

ryaplots:
Added notifications dropdown:
Screenshot 2024-01-16 at 16 12 31

kschiffer:
Thanks!

Please make the following adjustments:
- The last three notifications of the dropdown should be loaded by default (as part of `logics/init.js`) so that there is no loading necessary when the dropdown is clicked
- The notification dropdown currently shows the first three notifications instead of most recent ones
- If there is an unseen notification, the dropdown button should render a blue dot to signify that
- The notification ID should be part of the URL to make it possible to link to individual notifications directly, eg: `/console/notifications/cdcf9cd0-fe2d-4ddc-ae01-39a8a96969b9`
  - The notification dropdown should make use of this feature instead of any other logic that determines the ID
  - Also note that the links in the notification dropdown do not work when the notification panel is already open
 - Please fix the following error:
   1. Navigate to the notifications panel
   2. Scroll down the list to trigger a reload of items
   3. Now click on the "See archived messages" button
   4. The console crashes
- The date in the title section of the message panel (right side) needs better responsive behavior.
  - I'll provide you with a design for this but please already work to change the date formatting from eg `29 January, 2024` to the shorter `29/01/2024` formatting in smaller screen-widths.

image

kschiffer:
Please also apply the following diff:
```diff
diff --git a/pkg/webui/components/spinner/spinner.styl b/pkg/webui/components/spinner/spinner.styl
index c2ae15a1da..c06779f6c0 100644
--- a/pkg/webui/components/spinner/spinner.styl
+++ b/pkg/webui/components/spinner/spinner.styl
@@ -54,10 +54,10 @@ $xs = $cs.m
   animation: rotate .6s infinite linear
 
 .stop
-  stop-color: var(--c-icon-neutral-light)
+  stop-color: var(--c-icon-brand-normal)
 
   .faded &
-    stop-color: var(--c-icon-neutral-normal)
+    stop-color: var(--c-icon-neutral-light)
 
 @keyframes rotate
   from
@@ -72,7 +72,7 @@ $xs = $cs.m
 .center
   &:not(.inline)
     position: absolute
-    top: ($l * -1) // To achieve visual centering.
+    top: 0
     left: 0
     right: 0
     bottom: 0
@@ -83,11 +83,11 @@ $xs = $cs.m
 
   &.small
     height: $s
-    top: ($s * -1) // To achieve visual centering.
+    top: 0
 
   &.micro
     height: $xs
-    top: ($xs * -1) // To achieve visual centering.
+    top: 0
 
 .small .spinner
   width: $s
diff --git a/pkg/webui/console/containers/notifications/notification-list/list-item.js b/pkg/webui/console/containers/notifications/notification-list/list-item.js
index 6fa0de090e..c68b1a3005 100644
--- a/pkg/webui/console/containers/notifications/notification-list/list-item.js
+++ b/pkg/webui/console/containers/notifications/notification-list/list-item.js
@@ -34,7 +34,7 @@ export const NotificationListItem = ({ notification, selectedNotification, handl
   const unseenIds = useSelector(selectUnseenIds)
   const showSelected = selectedNotification?.id === notification.id
   const showUnseenStatus = unseenIds.includes(notification.id)
-  const classes = classNames(style.notificationPreview, 'm-0 d-flex', {
+  const classes = classNames(style.notificationPreview, 'm-0 d-flex p-cs-m', {
     [style.notificationSelected]: showSelected,
   })
   const titleClasses = classNames(style.notificationPreviewTitle, {
@@ -105,14 +105,11 @@ NotificationListItem.defaultProps = {
 }
 
 export const NotificationListSpinner = () => {
-  const classes = classNames(style.notificationPreview, 'm-0')
-  const titleClasses = classNames(style.notificationPreviewTitle)
+  const classes = classNames(style.notificationPreview, 'm-0', 'p-0')
 
   return (
     <div className={classes}>
-      <div className={titleClasses}>
-        <Spinner faded small center />
-      </div>
+      <Spinner faded micro center />
     </div>
   )
 }
diff --git a/pkg/webui/console/containers/notifications/notifications.styl b/pkg/webui/console/containers/notifications/notifications.styl
index ea34a74256..4441f06f97 100644
--- a/pkg/webui/console/containers/notifications/notifications.styl
+++ b/pkg/webui/console/containers/notifications/notifications.styl
@@ -17,6 +17,7 @@
   flex-direction: column
   border-right: 1px solid var(--c-border-neutral-light)
   width: 33%
+  min-width: 25rem
 
   &-change-button
     margin-top: $cs.m
@@ -84,10 +85,9 @@
       display: block
 
 .notification-preview
-  height: 7rem
   width: 100%
+  height: 100%
   border-bottom: 1px solid var(--c-border-neutral-light)
-  padding: $cs.m
 
   &.notification-selected
     background-color: var(--c-bg-brand-normal)
@@ -100,6 +100,7 @@
     gap: $cs.xxs
     margin-bottom: $cs.xs
     font-weight: 600
+    position: relative
 
     &.notification-read
       margin-left: 0.857rem
```

kschiffer:
And these:
```diff
diff --git a/pkg/webui/console/containers/notifications/notification-list/index.js b/pkg/webui/console/containers/notifications/notification-list/index.js
index f2ab7b906d..ad5459cf2a 100644
--- a/pkg/webui/console/containers/notifications/notification-list/index.js
+++ b/pkg/webui/console/containers/notifications/notification-list/index.js
@@ -87,13 +87,20 @@ const NotificationList = ({
   }, [dispatch, userId, unseenIds])
 
   const classes = classNames(styles.notificationHeaderIcon)
+  const isSelected = notification =>
+    selectedNotification && selectedNotification.id === notification.id
+  const isNextSelected = notification => {
+    const index = items.findIndex(item => item.id === notification.id)
+    return index + 1 < items.length && isSelected(items[index + 1])
+  }
 
   const Item = ({ index, style }) =>
     isItemLoaded(index) ? (
       <div style={style}>
         <NotificationListItem
           notification={items[index]}
-          selectedNotification={selectedNotification}
+          isSelected={isSelected(items[index])}
+          isNextSelected={isNextSelected(items[index])}
           handleClick={handleClick}
         />
       </div>
@@ -133,7 +140,7 @@ const NotificationList = ({
           />
         )}
       </div>
-      <div className="flex-grow">
+      <div className="flex-grow pl-cs-xs">
         <AutoSizer>
           {({ height, width }) => (
             <InfiniteLoader
@@ -145,9 +152,10 @@ const NotificationList = ({
             >
               {({ onItemsRendered, ref }) => (
                 <List
+                  className={styles.notificationListList}
                   height={height}
                   width={width}
-                  itemSize={98}
+                  itemSize={88}
                   ref={ref}
                   itemCount={itemCount}
                   onItemsRendered={onItemsRendered}
diff --git a/pkg/webui/console/containers/notifications/notification-list/list-item.js b/pkg/webui/console/containers/notifications/notification-list/list-item.js
index c68b1a3005..06d6c28b5e 100644
--- a/pkg/webui/console/containers/notifications/notification-list/list-item.js
+++ b/pkg/webui/console/containers/notifications/notification-list/list-item.js
@@ -30,20 +30,18 @@ import { selectUnseenIds } from '@console/store/selectors/notifications'
 
 import style from '../notifications.styl'
 
-export const NotificationListItem = ({ notification, selectedNotification, handleClick }) => {
+export const NotificationListItem = ({ notification, isSelected, isNextSelected, handleClick }) => {
   const unseenIds = useSelector(selectUnseenIds)
-  const showSelected = selectedNotification?.id === notification.id
   const showUnseenStatus = unseenIds.includes(notification.id)
   const classes = classNames(style.notificationPreview, 'm-0 d-flex p-cs-m', {
-    [style.notificationSelected]: showSelected,
+    [style.notificationSelected]: isSelected,
+    [style.notificationNextSelected]: isNextSelected,
   })
   const titleClasses = classNames(style.notificationPreviewTitle, {
     [style.notificationRead]: !showUnseenStatus,
-    [style.notificationSelected]: showSelected,
   })
   const previewClasses = classNames(style.notificationPreviewContent, {
     [style.notificationRead]: !showUnseenStatus,
-    [style.notificationSelected]: showSelected,
   })
 
   return (
@@ -55,7 +53,7 @@ export const NotificationListItem = ({ notification, selectedNotification, handl
       data-test-id="notification-list-item"
     >
       {showUnseenStatus && <Status pulse={false} status="good" className={style.unseenMark} />}
-      <div>
+      <div className="w-full">
         <div className={titleClasses}>
           <div className={style.notificationPreviewTitleText}>
             <Notification.Title
@@ -89,19 +87,14 @@ export const NotificationListItem = ({ notification, selectedNotification, handl
 
 NotificationListItem.propTypes = {
   handleClick: PropTypes.func.isRequired,
+  isNextSelected: PropTypes.bool.isRequired,
+  isSelected: PropTypes.bool.isRequired,
   notification: PropTypes.shape({
     id: PropTypes.string,
     created_at: PropTypes.string,
     notification_type: PropTypes.string,
     status: PropTypes.string,
   }).isRequired,
-  selectedNotification: PropTypes.shape({
-    id: PropTypes.string,
-  }),
-}
-
-NotificationListItem.defaultProps = {
-  selectedNotification: undefined,
 }
 
 export const NotificationListSpinner = () => {
diff --git a/pkg/webui/console/containers/notifications/notifications.styl b/pkg/webui/console/containers/notifications/notifications.styl
index 69cf0eec86..33856d7275 100644
--- a/pkg/webui/console/containers/notifications/notifications.styl
+++ b/pkg/webui/console/containers/notifications/notifications.styl
@@ -85,34 +85,47 @@
       display: block
 
 .notification-preview
-  height: 100%
   width: 100%
+  height: 100%
   border-bottom: 1px solid var(--c-border-neutral-light)
 
   &.notification-selected
     background-color: var(--c-bg-brand-normal)
+    border-radius: $br.m
+    border: 0
     +media-query($bp.s)
       background-color: unset
 
+    .notification-preview-title
+      color: var(--c-text-neutral-min)
+      +media-query($bp.s)
+        color: var(--c-text-neutral-heavy)
+
+      .notification-time
+        color: var(--c-text-neutral-min)
+
+    .notification-preview-content
+      color: var(--c-text-neutral-min)
+      +media-query($bp.s)
+        color: var(--c-text-neutral-light)
+
+  &.notification-next-selected
+    border: 0
+
   &-title
     display: flex
     justify-content: space-between
     gap: $cs.xxs
-    margin-bottom: $cs.xs
+    margin-bottom: $cs.xxs
     font-weight: 600
     position: relative
+    overflow: hidden
+    text-overflow: ellipsis
+    white-space: nowrap
 
     &.notification-read
       margin-left: 0.857rem
 
-    &.notification-selected
-      color: var(--c-text-neutral-min)
-      +media-query($bp.s)
-        color: var(--c-text-neutral-heavy)
-
-      .notification-time
-        color: var(--c-text-neutral-min)
-
     .notification-time
       font-size: $fs.s
       font-weight: normal
@@ -128,18 +141,13 @@
     overflow: hidden
     display: -webkit-box
     -webkit-box-orient: vertical
-    -webkit-line-clamp: 2 // Start showing ellipsis when 2nd line is reached.
+    -webkit-line-clamp: 3 // Start showing ellipsis when 2nd line is reached.
     white-space: pre-wrap
     font-size: $fs.s
 
     &.notification-read
       margin-left: 0.857rem
 
-    &.notification-selected
-      color: var(--c-text-neutral-min)
-      +media-query($bp.s)
-        color: var(--c-text-neutral-light)
-
   &.notification-selected
     color: var(--c-text-neutral-min)
     +media-query($bp.s)
```

kschiffer:
image

Please see this screendesign for the responsive behavior of the title section.

- Date format becomes short format
- Archive button becomes icon only
- Title text uses ellipsis

ryaplots:
> image
> 
> Please see this screendesign for the responsive behavior of the title section.
> 
>     * Date format becomes short format
> 
>     * Archive button becomes icon only
> 
>     * Title text uses ellipsis

From what breakpoint should this happen? @kschiffer

kschiffer:
> > image
> > Please see this screendesign for the responsive behavior of the title section.
> > ```
> > * Date format becomes short format
> > 
> > * Archive button becomes icon only
> > 
> > * Title text uses ellipsis
> > ```
> 
> From what breakpoint should this happen? @kschiffer

Please decide for yourself based on some experiments.

pkg/webui/console/store/middleware/logics/notifications.js
```diff
@@ -23,11 +23,43 @@ import createRequestLogic from '@ttn-lw/lib/store/logics/create-request-logic'
...
 import { selectUserId } from '@console/store/selectors/logout'
+import { selectTotalUnseenCount } from '@console/store/selectors/notifications'
 
 const m = defineMessage({
   newNotifications: 'You have new notifications',
 })
 
+const updateThroughPagination = async (totalCount, userId) => {
+  let page = 1
+  const limit = 100
```
kschiffer: Use the highest limit value here, which is 1000. This will save requests.

pkg/webui/console/containers/notifications/notification-list/index.js
```diff
@@ -64,9 +64,19 @@ const NotificationList = ({
...
+  const handleClick = useCallback(
+    async (_, id) => {
+      const clickedNotification = items.find(notification => notification.id === id)
+      if (!isArchive && !('status' in clickedNotification) && totalUnseenCount > 0) {
+        await dispatch(attachPromise(updateNotificationStatus([id], 'NOTIFICATION_STATUS_SEEN')))
+      }
+    },
+    [items, dispatch, isArchive, totalUnseenCount],
+  )
+
```
kschiffer: Instead of using the click handler, you can use a `useEffect` in `containers/notifications/index.js` that watches the `selectedNotification` and sends a "seen" request if required. Let's actually put this behind a `setTimeout` so it only updates after the notification has been "seen" for two seconds.

pkg/webui/console/views/app/index.js
```diff
@@ -122,6 +125,13 @@ const Layout = () => {
   }, [isDrawerOpen, openDrawer, closeDrawer])
   // End of mobile side menu drawer functionality
 
+  useEffect(() => {
```
kschiffer: Your earlier iteration of setting the interval within the logic was good. Why did you change it?

pkg/webui/console/containers/side-bar/navigation/general-side-navigation.js
```diff
@@ -55,7 +55,7 @@ const GeneralSideNavigation = () => {
         )}
         <SideNavigation.Item
           title={sharedMessages.notifications}
-          path="/notifications"
+          path="/notifications/inbox"
```
kschiffer:
```suggestion
          path="/notifications"
```
I realize this has to be `/notifications` only, otherwise `/notifications/archive` will not match and the root route will forward to `/notifications/inbox` anyway.

kschiffer (changes requested):

pkg/webui/console/views/app/index.js
ryaplots: There two parts to this:
1. The problem changed a bit from the first iteration. Before, we needed to only show a toast when there were new notifications. However, now we also need to update the state in order to display the new notifications dot in the header dropdown. I solved it by adding a `dispatch(notifications.getUnseenNotificationsPeriodicallySuccess())` within the `setInterval`.
2. The problem with the above solution is that apparently, an action is not dispatched within a `setInterval()` in the middleware logic. So although the request happens every 5 minutes, the state is not updated, which means that the notification dot never shows on the header dropdown and the total unseen notifications state doesn't get updated unless the user opens the notifications view. I tried to understand why and used ChatGPT a lot for that, but I couldn't get to the root of it in the end. Whenever I would dispatch an action within a setInterval(), it would never happen. So I looked for other ways to do update the state every 5 minutes. And this was a solution that would update the state. @kschiffer

kschiffer: This has likely to do with the fact that you used `createRequestLogic()`, which is our wrapper for redux-logic's `createLogic()` which decorates it with many store and connection specifics, but it also automatically calls `done()` which will cause any subsequent dispatches to be ignored, so this cannot be used for any periodic logics.

So to make this work, please use a custom logic with `createLogic()`, which does not call `done()`. It should then also use the `timeout: 0` option.

ryaplots: For my understanding, what problems can useEffect() bring in this case? Why is it better to do it in the middleware logic?

kschiffer: The same thing can definitely be done as a `useEffect()` as well but it would make maintenance harder. E.g. `/console/views/app/index.js` is not a good place for setting this logic up as this component itself has nothing to do with the notifications, making it harder for anyone later on to find the source of these periodical dispatches. The more we keep related logic in the same place, the easier it is for anyone to reason about and maintain the code.

`redux-logic` just as much allows us to orchestrate dispatches and gives us other capabilities on top such as debouncing, validation, and cancellation. This is why we use it to centralize and streamline business logic, which should be separated from component logic as much as possible, and in this case, I don't see a good reason to break out of this convention.

kschiffer (approved):
🎉

pkg/webui/console/store/middleware/logics/notifications.js
```diff
@@ -0,0 +1,186 @@
...
+  process: async (_, dispatch) => {
+    // Fetch once initially.
+    dispatch(notifications.refreshNotifications())
+
+    setInterval(
+      async () => {
+        dispatch(notifications.refreshNotifications())
+      },
+      // Refresh notifications every 15 minutes.
+      1000 * 10 * 15,
```
kschiffer:
```suggestion
      1000 * 60 * 15,
```
mea culpa

adriansmares:
The notification events PR (https://github.com/TheThingsNetwork/lorawan-stack/pull/6917) has been merged - you can consider in this PR, or another PR, adding support for user subscriptions instead of periodical polling.

kschiffer:
Yes, thanks! We'll do this separately.

PR #6480, opened 2023-08-24
Fix extra limitless request

ryaplots:
#### Summary

This PR fixes a bug found in the entity tables. There was one request being made without a limit, which then messed up the pagination if it took longer than the other requests.

#### Changes

- `useRequest()` did not include the correct filters.
- Removed the `useRequest()` entirely as it triggers an extra request, which in the case of the Fetch Table already happens in `useEffect()` and we can derive fetching and errors from there.

#### Testing

1. Open any table.
2. See that all list requests have a limit field mask.

#### Checklist


kschiffer (approved):
I think it would be better to use just useRequest instead to have a streamlined way of fetching items. But lets fix this first and then look what's the best way to do this.

ryaplots:
> I think it would be better to use just useRequest instead to have a streamlined way of fetching items. But lets fix this first and then look what's the best way to do this.

Will open a PR with alternative fix.

PR #6482, opened 2023-08-25
Add RP2-1.0.4 support

adriansmares:
#### Summary

References https://github.com/TheThingsIndustries/lorawan-stack/issues/3743

#### Changes

- Deduplicate common band parameters.
  - Some band parameters, like the default RX1 delay, are not really region specific - the regional parameters specification uses them for all of the bands in the specification, and in general they have not really changed between regional parameters versions. 
  - Since we duplicate them a lot in the band definitions, but they are identical, I have extracted them into a `SharedParameters` struct that we can reference, and should the parameters change in a future RP document, we can use a single instance instead of defining them in each band.
- Add RP2-1.0.4 support.
  - The support itself is required for the relay specification, as the relay parameters are introduced by RP2-1.0.4.
  - The RP2-1.0.4 specification has two features that are not implemented in this PR: adding extra channels to fixed channel regions like US915/AU915, and using the channel mask CFList in dynamic channel plan regions. Our current frequency plans do not require these features, and we can revisit them later on in case we want to support this functionality.
- Fix CN4710-510 26 MHz beacon and ping channel frequencies.
  - This is mainly for completeness sake, as the 26 MHz bands have a simpler frequency description.
- Add relay specification WOR channels to the appropriate regions in the RP2-1.0.4 specification.
  - 2.4GHz and the Morocco draft are intentionally left out, as they don't have a provided set of relay parameters. EU433 apparently has no relay parameters in RP2-1.0.4 - I do not know the reason for this.
- Support the relay parameters in the band description RPCs.

#### Testing

Unit testing. Release specific testing is not required.

##### Regressions

The only retroactive changes affect the shared band parameters, which can be verified by inspecting the `universalSharedParameters` definition and comparing it against the RP2 document:

image

#### Notes for Reviewers

You can skip individual the Go band definitions and test data. Please only take a look at the structural changes.

#### Checklist


johanstokking (approved):

ryaplots (approved):
Console part looks good to me.

KrishnaIyer (approved):
LGTM. Looked only a few samples and skimmed through the rest.

Issue #6314, opened 2023-06-07
RPC implementation of deletion for interfaces refactored


nicholaspcr:
### Summary

Taking into consideration that a few interfaces on IS were changed to be more "single-purpose". Instead of calling the update route without passing rights it is interesting to have RPC services for the new delete methods.

Reference to refactor issue https://github.com/TheThingsNetwork/lorawan-stack/issues/5587

### Current Situation

Delete method for [MembershipStore](https://github.com/TheThingsNetwork/lorawan-stack/blob/d461e4653bf1cefe64c91cf740c71399590b2929/pkg/identityserver/store/store_interfaces.go#LL215C1-L215C1) is called at:
1. ClientAccess - "/clients/{client_ids.client_id}/collaborators"
2. GatewayAccess - "/gateways/{gateway_ids.gateway_id}/collaborators"
3. OrganizationAccess - "/organizations/{organization_ids.organization_id}/collaborators"
4. ApplicationAccess - "/applications/{application_ids.application_id}/collaborators"

Delete method for [APIKeyStore](https://github.com/TheThingsNetwork/lorawan-stack/blob/d461e4653bf1cefe64c91cf740c71399590b2929/pkg/identityserver/store/store_interfaces.go#L247) is called at:
1. UserAccess - "/users/{user_ids.user_id}/api-keys/{api_key.id}"
2. OrganizationAccess - "/organizations/{organization_ids.organization_id}/api-keys/{api_key.id}"
3. ApplicationAccess - "/applications/{application_ids.application_id}/api-keys/{api_key.id}"
5. GatewayAccess - "/gateways/{gateway_ids.gateway_id}/api-keys/{api_key.id}"

### Why do we need this? Who uses it, and when?

To not depend on calling Set/Update routes while passing no rights, the alternative is to have a defined route/rpc_service to operate the delete operations. 

### Proposed Implementation

Add new paths/services in the proto definitions of the `entities_access.proto` and do their implementation in a way where the new interface methods are called.

### Contributing

- [ ] I can help by doing more research.
- [X] I can help by implementing the feature after the proposal above is approved.
- [X] I can help by testing the feature before it's released.

### Code of Conduct

- [X] I agree to follow TTN's [Community Code of Conduct](https://github.com/TheThingsNetwork/Manifesto/blob/master/CodeOfConduct.md).

nicholaspcr:
As mentioned in the PRs, I'm writing the test steps bellow. 
[...]

PR #6489, opened 2023-08-29
Create RPC delete collaborator

nicholaspcr:
#### Summary

References #6314 

#### Changes

- Add deleteCollaborator for: [clients, gateways, organizations, applications]
- Update tests to validate deleteCollaborator behavior

#### Testing

Update on unit tests and manual tests via the changes done on PRs that will be linked shortly.

###### Steps to reproduce and test the changes. 

Since this will be merge only when all the subsequent PRs are approved, it means that the Console and CLI will be calling the new route. So in order to test just add and delete a collaborator to the entities mentioned in the changes section.

Nevertheless I will write some more detailed steps on the referenced issue.

##### Regressions

New rpc service and it does not change the behavior of the update route, there should not be any regressions.

#### Notes for Reviewers

Broke this into smaller PRs in order to avoid having huge PRs.

Didn't add anything on the `CHANGELOG` since this is mostly internal. It is probably worth adding that deletion of collaborators won't be able to be performed via the `update` route but that will only happen on a major release.

#### Checklist


pkg/identityserver/application_access_test.go
```diff
@@ -418,14 +418,37 @@ func TestApplicationCollaborators(t *testing.T) {
 				})
 			}
 
+			// TODO: Remove SetCollaborator test case (https://github.com/TheThingsNetwork/lorawan-stack/issues/6488).
```
adriansmares, line 421: You should also verify that the collaborator is no longer available, similar to the old test structure.

Same for other access stores.

adriansmares (commented):
LGTM in general.

adriansmares (approved):

johanstokking (commented):
Skimmed over it; you can merge

